PR #53, opened 2015-11-06
Commented Out SSL Proxy settings so runserver command can be used for…

devGregA:
… testing by default


devGregA:
summoning @grendel513  :D

grendel513:
Hey @devGregA  - 

I think we should comment out all of the following  in order to make sure the SSL options don't bite anyone:

```
SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTOCOL', 'https')
SECURE_SSL_REDIRECT = True
SECURE_BROWSER_XSS_FILTER = True
```

And in the `MIDDLEWARE_CLASSES` section the following class should be commented as well:

```
'django.middleware.security.SecurityMiddleware'
```

Also, in order to alert folks of these setting a message should be added to `setup.bash` and the Vagrant install to let them know that they should uncomment these when they are ready to go production.

Thoughts?

devGregA:
Hi @grendel513 why comment out the secure browser filter or the middleware?
Just from my experience, changing only the SSL settings allows a user to download, go through setup.bash, and immediately use manage.py runserver. We already warn users at line 131 in setup.bash We could make it more prominent though?

grendel513:
hey @devGregA - I was thinking for consistency's sake.  Since the middleware governs all the pieces you commented out and since the Django project recommends these features should be taken car of by the Web Server in production - we could comment them all out and provide a link to the documentation as well as with a warning that it is inherently insecure to run our set up in production.  See https://docs.djangoproject.com/en/1.8/ref/middleware/#module-django.middleware.security  for details.

Just my two cents - merging your pull request as-is is fine by me too ;)

devGregA:
To me I think easiest is best, but I do get what you're saying. It's just more work to go from dev to production and an opportunity for someone new to not get something right. Summoning @cneill for his thoughts? :D

cneill:
@devGregA @grendel513 : I'd tend to agree with Jay - nginx/etc. can handle these things better than Django, and I don't think we want to be in the business of trying to configure folks' production environments for them. I'd say comment out all the SSL stuff, and we can maybe write a docs page describing how we'd set up nginx for production, with a "not-recommended" section at the bottom explaining how to set these headers using Django.

devGregA:
@grendel513 @cneill so should we just remove all the security measures rather than comment them out?

grendel513:
@devGregA @cneill - I say comment them out, this will keep folks from having to look them up in the docs.

cneill:
Agreed, @grendel513 . I think we can leave them in there and reference them in a "not-recommended" doc page like I suggested above.

devGregA:
@grendel513 @cneill  sorry for the delay. Changes are made. I also updated setup.bash and tested setup.bash to make sure I didn't screw anything up.

PR #58, opened 2016-01-10
Support for importing Checkmarx Findings

aaronweaver:
Sample Checkmarx XML file for testing: https://github.com/aaronweaver/scannerxml/blob/master/checkmarx-sample.xml

### DIFF
--- a/dojo/fixtures/test_type.json
+++ b/dojo/fixtures/test_type.json
@@ -75,5 +75,12 @@
     },
     "model": "dojo.test_type",
     "pk": 11
+  },
+  {
+    "fields": {
+      "name": "Checkmarx"
+    },
+    "model": "dojo.test_type",
+    "pk": 12
   }
 ]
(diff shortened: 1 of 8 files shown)


devGregA:
Awesome stuff @aaronweaver! @grendel513 do you have some time to review with me tomorrow?

grendel513:
@aaronweaver this is great, thanks for helping make DefectDojo better.  The only thing I see that may have to be updated is: If a duplicate finding is discovered you do not capture the information about the duplicate.  In the other parsers, if a finding is believed to be a duplicate we concatenate the description  See https://github.com/rackerlabs/django-DefectDojo/blob/master/dojo/tools/nexpose/parser.py#L241 and https://github.com/rackerlabs/django-DefectDojo/blob/master/dojo/tools/nessus/parser.py#L187 for an example.  Do you think the Checkmarx parser should do the same?  Or is the dupe_key (NodeId) unique enough.  The Veracode parser (https://github.com/rackerlabs/django-DefectDojo/blob/master/dojo/tools/veracode/parser.py#L44) creates a dupe_key that includes more pieces, so it does not need to concatenate the description.  Thoughts?

aaronweaver:
@grendel513 I'll take a look and double check but I think NodeId is always unique.

aaronweaver:
@grendel513 I ran a few more tests and found the node ID changed so I've updated the pull request with a change similar to the Veracode duplication check.

grendel513:
Thanks @aaronweaver - this looks great!

Issue #79, opened 2016-04-14
Report generation issue


S1m0x81:
I'm having an issue generating a PDF report, once I click on "Save and Run" the report status remains "Running" for hours and the report is never generated.

Below is the debug log, any help would be appreciated. Thanks

Performing system checks...

System check identified no issues (0 silenced).
April 14, 2016 - 11:23:03
Django version 1.8.10, using settings 'dojo.settings'
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.

[14/Apr/2016 11:23:24] "POST /reports/custom HTTP/1.1" 302 0
[14/Apr/2016 11:23:24] "GET /reports HTTP/1.1" 200 27206
[14/Apr/2016 11:23:34] "GET /reports/7/status HTTP/1.1" 200 32
[14/Apr/2016 11:23:44] "GET /reports/7/status HTTP/1.1" 200 32
[14/Apr/2016 11:23:54] "GET /reports/7/status HTTP/1.1" 200 32
[14/Apr/2016 11:24:04] "GET /reports/7/status HTTP/1.1" 200 32
[14/Apr/2016 11:24:14] "GET /reports/7/status HTTP/1.1" 200 32
[14/Apr/2016 11:24:24] "GET /reports/7/status HTTP/1.1" 200 32
[14/Apr/2016 11:24:34] "GET /reports/7/status HTTP/1.1" 200 32
[14/Apr/2016 11:24:44] "GET /reports/7/status HTTP/1.1" 200 32

![defectdojo](https://cloud.githubusercontent.com/assets/18467094/14538033/5c8162be-023f-11e6-82be-bdc9e484be1f.png)

devGregA:
Hi @S1m0x81, 

Sorry to hear you're having this issue. @grendel513 knows the reporting section better than I do. Could you change debug to true in dojo/settings.py and see if there are any errors?

devGregA:
Also what operating system are you running? I'll try to reproduce the issue.

grendel513:
HI @S1m0x81 - Do you have celery running?  PDF reports depend on celery because they are asynchronous.  See our documentation at http://defectdojo.readthedocs.org/en/latest/features.html#reports

S1m0x81:
Thanks everyone for your help. I'm running Kali Linux 2.0 which is based on Debian Jessie and I do have celery running. Also I think it's worth to mention that your demo site has the same issue (See screenshot below)

[...]

PR #84, opened 2016-06-26
added reports script and updated tasks.py to work out of box

devGregA:
Will submit PR with doc updates shortly.


grendel513:
Fixes #79

PR #89, opened 2016-08-03
added deduplication of findings

devGregA:


grendel513:
hey @devGregA - I feel like ENABLE_DEDUPLICATION should be defaulted to False, thoughts?  Also, did you test code when there is no ENABLE_DEDUPLICATION entry at all in settings file?  I believe you may get a key error if you reference it and it does not exist.  This is important because for all existing installations the setting will not be there initially.

We also may need to add some documentation around this new setting and what it accomplishes.

@cneill - thoughts?

dojo/models.py
```diff
@@ -526,6 +527,20 @@ def get_numerical_severity(severity):
...
+
+    def save(self, *args, **kwargs):
+        super(Finding, self).save(*args, **kwargs)
+        if settings.ENABLE_DEDUPLICATION:
+            eng_findings_cwe = Finding.objects.filter(test__engagement=self.test.engagement, cwe=self.cwe).exclude(id=self.id).exclude(cwe=None)
+            eng_findings_title = Finding.objects.filter(test__engagement=self.test.engagement, title=self.title).exclude(id=self.id)
+            total_findings = eng_findings_cwe | eng_findings_title
+            for find in total_findings:
+               if set(find.endpoints.all()) == set(self.endpoints.all()):
+                        find.duplicate = True
```
cneill: This line and the line below it are over-indented

dojo/models.py
```diff
@@ -526,6 +527,20 @@ def get_numerical_severity(severity):
...
+        super(Finding, self).save(*args, **kwargs)
+
+    def save(self, *args, **kwargs):
+        super(Finding, self).save(*args, **kwargs)
+        if settings.ENABLE_DEDUPLICATION:
+            eng_findings_cwe = Finding.objects.filter(test__engagement=self.test.engagement, cwe=self.cwe).exclude(id=self.id).exclude(cwe=None)
+            eng_findings_title = Finding.objects.filter(test__engagement=self.test.engagement, title=self.title).exclude(id=self.id)
+            total_findings = eng_findings_cwe | eng_findings_title
+            for find in total_findings:
+               if set(find.endpoints.all()) == set(self.endpoints.all()):
```
cneill: This line isn't indented to 4 spaces

dojo/engagement/views.py
```diff
@@ -180,9 +180,10 @@ def view_engagement(request, eid):
         eng.save()
 
     add_breadcrumb(parent=eng, top_level=False, request=request)
-
+    findings = Finding.objects.filter(test__engagement=eng, duplicate=False)
```
cneill: This should check for `settings.ENABLE_DEDUPLICATION` so that users can control whether they want to filter these out or not

dojo/models.py
```diff
@@ -526,6 +527,20 @@ def get_numerical_severity(severity):
...
+                        find.duplicate = True
+                        find.legacy_save()
```
cneill: This doesn't pass `*args`/`**kwargs` to legacy_save.

dojo/models.py
```diff
@@ -526,6 +527,20 @@ def get_numerical_severity(severity):
         else:
             return 'S4'
 
+    def legacy_save(self, *args, **kwargs):
+        super(Finding, self).save(*args, **kwargs)
+
+    def save(self, *args, **kwargs):
+        super(Finding, self).save(*args, **kwargs)
```
cneill: Instead of calling self.legacy_save() on 542, you should call this at the bottom of save(), outside the check for ENABLE_DEDUPLICATION. I think we can do without legacy_save() entirely since it's not called anywhere else in the codebase, and the below lines would only affect the user if they explicitly enable deduplication. I can't think of a reason we'd want to ignore the ENABLE_DEDUPLICATION setting if the user potentially set it to True.

dojo/templates/dojo/view_eng.html
```diff
@@ -371,4 +371,64 @@ <h4 class="pull-left">
         </div>
 
     </div>
+ <div class="row">
+        <div id="finding" class="col-md-12">
+            <div class="panel panel-default">
+                <div class="panel-heading">
+                    <h4> Findings</h4>
+                </div>
+{% if findings %}
```
cneill: Nit: this should match the indentation of the "else"/"endif" blocks below

cneill:
I agree that we should default this to False and document it so that users know what they're getting and aren't surprised by Findings suddenly going missing.

See my other comments inline.

PR #95, opened 2016-08-18
Update setup.bash

davevs:
added dependecies for vanilla linux installes

### DIFF
--- a/setup.bash
+++ b/setup.bash
@@ -52,7 +52,7 @@ if [[ ! -z $YUM_CMD ]]; then
 	sudo yum install gcc libmysqlclient-dev python-devel mysql-server mysql-devel MySQL-python python-setuptools python-pip nodejs wkhtmltopdf npm -y
 	sudo yum groupinstall 'Development Tools'
 elif [[ ! -z $APT_GET_CMD ]]; then
-    sudo apt-get install gcc libssl-dev python-dev libmysqlclient-dev python-pip mysql-server nodejs-legacy wkhtmltopdf npm -y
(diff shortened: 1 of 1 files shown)


grendel513:
Thanks @davevs!  @devGregA can you please take a look to see if this is inline with the install process?

devGregA:
Summoning @cneill :D. I went through all the new libs. Couldn't find major vulnerabilities or issues. Doesn't brake the install for most common systems. @cneill will you weigh in?

devGregA:
Hey @davevs is there a specific OS you're trying to cover? I'm wondering if I can make the list a little lighter by targeting different libs to satisfy the requirements.

davevs:
@devGregA not really. In my view this tool might be tried by not the most tech-savy of people. Following that logic that kind of people probably uses something like mint or ubuntu. I wanted to make sure that the default installer would at least work on those systems to prevent people from abandoning it within a couple of minutes.

devGregA:
Hi @davevs, I've been looking through these packages for a while, and I haven't found a use case for adding these. Most people that we've spoken to are using Ubuntu. These packages aren't required for Dojo to run on Ubuntu and add overhead. If I've missed something please feel free to point it out to me.

davevs:
@devGregA I got errors when using the installer on Mint and had to install these packages manually to make the installer complete. Especially git is necessary to complete the last part of the install, but is not necessarily on the system. Other examples are libtiff and libjpg which are necessary to complete the install.

devGregA:
@davevs gotcha. I'm swamped until Wednesday but will re-examine Mint later in the week.

devGregA:
@davevs sorry for the delay. I haven't forgotten about this.

davevs:
@devGregA np, take your time :)

PR #114, opened 2016-09-09
added 200 status api check product, test, and finding

devGregA:


devGregA:
@aaronweaver @mtesauro any idea what is going wrong here guys?

aaronweaver:
It looks like the environment variables in Travis aren't set but the screenshot you showed has them set. If I look in prior builds I see Travis listing the environment variables (build #6) but in build #8 I don't see them.

build__6_-_owasp_django-defectdojo_-_travis_ci

build__8_-_owasp_django-defectdojo_-_travis_ci

mtesauro:
+1 on what Aaron said.  

I was about to write that but he's quicker then me.  The admin user and password are blank since those environment variables are not set and the test cannot log in whatever values it has for the admin creds.

Issue #110, opened 2016-09-07
setup.bash doesn't work


opsdisk:
Tried installing on vanilla 14.04.3 x64.

Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 303, in execute
    settings.INSTALLED_APPS
  File "/usr/local/lib/python2.7/dist-packages/django/conf/__init__.py", line 48, in __getattr__
    self._setup(name)
  File "/usr/local/lib/python2.7/dist-packages/django/conf/__init__.py", line 44, in _setup
    self._wrapped = Settings(settings_module)
  File "/usr/local/lib/python2.7/dist-packages/django/conf/__init__.py", line 92, in __init__
    mod = importlib.import_module(self.SETTINGS_MODULE)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/home/rack/Desktop/django-DefectDojo/dojo/__init__.py", line 5, in <module>
    from .celery import app as celery_app  # noqa
  File "/home/rack/Desktop/django-DefectDojo/dojo/celery.py", line 3, in <module>
    from celery import Celery
ImportError: No module named celery
==============================================================================

SUCCESS! Now edit your settings.py file in the 'dojo' directory to complete the installation.

We suggest you consider changing the following defaults:

    DEBUG = True  # you should set this to False when you are ready for production.
    Uncomment the following lines if you enabled SSL/TLS on your server:
        SESSION_COOKIE_SECURE = True
        CSRF_COOKIE_SECURE = True
        SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTOCOL', 'https')
        SECURE_SSL_REDIRECT = True
        SECURE_BROWSER_XSS_FILTER = True
        django.middleware.security.SecurityMiddleware

When you're ready to start the DefectDojo server, type in this directory:

    python manage.py runserver

Note: If git cannot connect using the git:// protocol when downloading bower artifacts, you can run the command 
below to switch over to https://
          git config --global url.https://.insteadOf git://

devGregA:
[...]

PR #115, opened 2016-09-10
setup.bash fix

opsdisk:
See "setup.bash doesn't work #110" ticket

Updated setup.bash (added libjpeg-dev) and changed the order of packages installed in setup.py (not sure if it matters, but it worked for me) and added packages uwsgi and watson.  

Had to comment out the last 3 lines in dojo/models.py as well

#watson.register(Product)
#watson.register(Test)
#watson.register(Finding)


devGregA:
@grendel513  I think it's interesting that libjpeg-dev corrected issues for them in 14.04. I remember Dav said the same thing about Mint. Any thoughts?

grendel513:
@opsdisk @devGregA i can see adding libjpeg-dev it will not add anything that is not needed.  I do not see why the watson registration lines are commented out.  They are needed in order for the search feature to work.  This pull request should not be merged with those lines commented out.  Thanks for looking into it.

PR #119, opened 2016-09-23
Some PEP8 linting

cneill:
Did some basic PEP8 linting across various files. I didn't bother with the 80 character limit since most of the lines in question would require significant re-writes. I tried not to make any changes that would affect the actual code, but there are a couple minor modifications like using .format() in a few places.


cneill:
Looking at failed build checks above, I don't think this patch has anything to do with the failure..

devGregA:
Nice! Great stuff @cneill. @aaronweaver I think this is because Travis only supplies proper env variables for commits which is a pain. Trying to figure out how to fix this.

devGregA:
found a fix. just a sec

devGregA:
@aaronweaver  https://docs.travis-ci.com/user/pull-requests I don't think at this time we have to worry about any of our envs being leaked?

devGregA:
I can add    - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then bash ./travis/run_on_pull_requests; fi'
to enable for PRs

aaronweaver:
@devGregA The only one that should be protected is the API key for SRC:CLR.

devGregA:
@cneill looks like you took out an import os that was required in settings.py.

cneill:
My bad - didn't look like it was referenced in the file.

devGregA:
going to roll back because I'm seeing other strange filter issues.

devGregA:
Hey @cneill  there was another issue that I just patched if you can resubmit.

PR #130, opened 2016-10-05
Added has attr to URLs

devGregA:
@grendel513, I'm in a catch 22 for testing. I don't have an old settings file, so if I comment out URL_PREFIX I get in error from settings.py. I ran the urls.py and didn't hit any errors. Please let me know if you see other problems.

### DIFF
--- a/dojo/templates/dojo/report_builder.html
+++ b/dojo/templates/dojo/report_builder.html
@@ -167,7 +167,7 @@
 
             $(document).on('submit', 'form.finding-list', function (event) {
                 var form = this;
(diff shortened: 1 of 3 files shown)


grendel513:
@devGregA it breaks because you left these lines in:
`print >>sys.stderr, 'this is prefix'

print >>sys.stderr, settings.URL_PREFIX`

There is no check around your debugging lines...I say take them out :)

devGregA:
Caught that and fixed here: https://github.com/OWASP/django-DefectDojo/pull/130/commits/de7d0bffc4ba0efcc3c83a0e579f2b7aca4ba120

PR #132, opened 2016-10-05
added datetime try/catch

devGregA:


devGregA:
@grendel513 temporary solution until I can trace down the inconsistency

devGregA:
https://travis-ci.org/devGregA/django-DefectDojo

Issue #62, opened 2016-01-15
Person Table


aaronweaver:
Create a person table with a role(s) and contact information so that a person can be associated with a application, line of business or organization.

grendel513:
@aaronweaver - Hi, the users table already exists, it is the default Django users (https://docs.djangoproject.com/es/1.9/topics/auth/default/#user-objects) and we do store email for them.  They are also tied in to Product authentication, see http://defectdojo.readthedocs.org/en/latest/features.html#users and http://defectdojo.readthedocs.org/en/latest/features.html#products for details.

aaronweaver:
@grendel513 I was referring to the Product Manager, Technical Contact and Manager fields. Right now it's a free form text field. I'd rather have that as a table associated with role which might be a person who doesn't login. For example I have the same owner for many of my products. It would be easier to associate with one person rather than having to re-enter the same information. Plus I'd like the ability to create a person with a role and associate it with a product rather than have pre-defined roles.

grendel513:
Thanks for the clarification @aaronweaver - I do think it makes sense.  I am not sure how soon we can get this implemented, but I will add it to our board.

devGregA:
@aaronweaver I think this is covered now with metadata. Would you please confirm?

aaronweaver:
@devGregA It would work but it would be easier to manage if it was a person table. One example I'd give is that often times there will be the same resource for several products. That means I'll have to update that person if they change in all those products.

grendel513:
@aaronweaver - do you envision these folks ever login on to the system?  Can we get away with extending Django's User model to store additional metadata?  This would allow for that table to store all humans.  You can restrict their access (not staff, not superuser and no authorized products) so there is no risk of them seeing or manipulating data they shouldn't.

A migration script will have to be provided for those that are already storing information in the Product Manager,  Technical Contact and Team Manager fields in Product, as they would go from plain character fields to User Model fields.  This script would have to exists even if we have a separate table.

Thoughts?

grendel513:
FYI @aaronweaver, @devGregA - see the accepted answer for steps needed to go from char field to model: http://stackoverflow.com/questions/35999186/change-type-of-django-model-field-from-charfield-to-foreignkey

aaronweaver:
@grendel513 - That might work and be the most flexible. I could see cases where the user might never login but others where the product manager might to see the overall security status.

grendel513:
[...]

PR #157, opened 2016-11-02
Takes care of #62

grendel513:
Takes care of #62. Product now has three new fields: product_manager, team_manager, and technical_contact.

These replace the fields prod_manager, manager, and tech_contact.
The new fields are foreign keys to User objects.

User is now extended by UserContactInfo which adds ability to store a title, phone, cell, twitter and github for each user.

In order for these changes to work correctly, updaters must run the following commands:
`./manage.py makemigrations dojo`
`./manage.py migrate`
`./manage.py migrate_product_contacts`

Fields prod_manager, manager, and tech_contact are no longer used anywhere, but left in Product model for backwards compatibility.  They will be removed in future updates.  References to these in the project have been replaced by new fields product_manager, team_manager, and technical_contact.


devGregA:
@aaronweaver

Issue #159, opened 2016-11-08
Severity Name Toggling


devGregA:

grendel513:
@devGregA Details?

devGregA:
@grendel513 right now in the metrics 'severity' is displayed as s3-s0 instead of low-critical. Should we axe this? Or make it toggle-able?

grendel513:
@devGregA - i see.  All the graphs (or most) do use the full severity name, it is the text based metrics that use the numerical severity value.  We can either show both "Critical (S0)" or we can have a toggle (javascript based) that can translate one to the other.

devGregA:
Sent you a Slack message, but I think it should be togglable from the settings. What are your thoughts?

PR #166, opened 2016-11-21
Takes care of #159

grendel513:
Now able to set the following `settings.py` attribute:

    S_FINDING_SEVERITY_NAMING = True/False

True will display S0, S1, S2, ect in most places
False will display Critical, High, Medium, etc

For new installations the attribute defaults to False.
For existing installations the attribute needs to be added.


devGregA:
Thanks for doing this Jay! Do you think we should have S Naming by default?

grendel513:
It's the current behavior, but I am okay with changing the default.

PR #168, opened 2016-11-23
Create ability to import scans from the tastypie API

CCoffie:
I was able to implement a importscan method to the API through tastypie. I probably did some things wrong because I'm not too familiar with tastypie. Tell me if there's anything I should change.

To test it you'll want to do a command like the following:
`curl -F "file=@report123.json" -H "Authorization: ApiKey admin:xxxxxxxxxxxxxxxxxxx" -F 'scan_type=Arachni Scan' -F 'tags=testing' -F 'active=true' -F 'scan_date=11/23/2016' -F 'eid=1' http://127.0.0.1:8000/api/v1/importscan/`

### DIFF
--- a/dojo/api.py
+++ b/dojo/api.py
@@ -4,15 +4,27 @@ from tastypie.authentication import ApiKeyAuthentication, MultiAuthentication, S
 from tastypie.authorization import Authorization
 from tastypie.authorization import DjangoAuthorization
 from tastypie.constants import ALL, ALL_WITH_RELATIONS
-from tastypie.exceptions import Unauthorized
-from tastypie.resources import ModelResource
+from tastypie.exceptions import Unauthorized, ImmediateHttpResponse, NotFound
+from tastypie.http import HttpCreated
+from tastypie.resources import ModelResource, Resource
 from tastypie.serializers import Serializer
-from tastypie.validation import CleanedDataFormValidation
+from tastypie.validation import CleanedDataFormValidation, Validation
+from django.core.exceptions import ObjectDoesNotExist
+from pytz import timezone
+from django.conf import settings
 
 from dojo.models import Product, Engagement, Test, Finding, \
-    User, ScanSettings, IPScan, Scan, Stub_Finding, Risk_Acceptance, Finding_Template
+    User, ScanSettings, IPScan, Scan, Stub_Finding, Risk_Acceptance, \
+    Finding_Template, Test_Type, Development_Environment, \
+    BurpRawRequestResponse, Endpoint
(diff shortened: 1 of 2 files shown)


aaronweaver:
@CCoffie Great PR! A few comments: 

- The TastyPie convention would be engagement rather than eid and the value would be: /api/v1/engagements/1/
- I think tags should be an optional value. (nice, I see that's done already)
- The response for a created object in tastypie is 201 with a location href of the new object id. For uploads it should be the test_id. I don't think you need to return anything, that's if we stick to the tastypie convention. Another API call to tests can be done for # of records imported.
- Change the date format to: "2016-11-01"

FYI - I have an API wrapper for Dojo, and have added uploads to test the functionality. https://github.com/aaronweaver/defectdojo_api

CCoffie:
- For the engagement parameter you would pass in the endpoint to the corresponding engagement? That's an interesting way of doing it. I can switch that around.
- Yeah I noticed that yesterday while testing it a little more so I made the tags parameter optional.
- I'm not to sure on how to do the responses with TastyPie but I'll give it a shot. If you have any pointers where I should look that would be assume.
- I kept running into problems with the date format stuff and it worked eventually but I think I know what I was doing wrong so I'll go back and fix it.

CCoffie:
Okay, I think I made all the necessary changes. To test it now you can use the following curl command.
```
curl -i -F "file=@report123.json" -H "Authorization: ApiKey admin:XXXXXXXXXXXXXXX" -F 'scan_type=Arachni Scan' -F 'tags=testing' -F 'verified=false' -F 'active=false' -F 'scan_date=2016/11/29' -F 'engagement=/api/v1/engagements/1/' http://127.0.0.1:8000/api/v1/importscan/
```

YSaxon:
Were you planning to add PUT/obj_update/re_import_scan support?

CCoffie:
No I didn't have plans to implement that. How does it differ from the import scan functionality?

aaronweaver:
I tested out importing several types of files and it works correctly. @devGregA @grendel513 Can you take a look?

YSaxon:
@CCoffie see [the reupload code](https://github.com/OWASP/django-DefectDojo/blob/master/dojo/test/views.py#L383). Basically, by rerunning/reimporting a scan you can automatically update old findings as mitigated or reintroduced by their absence or reappearance in the repeat scan.

CCoffie:
@YSaxon That actually sounds very useful. I can try looking into adding that.

devGregA:
This looks good to me. @aaronweaver I'm good with merging

PR #180, opened 2016-12-13
Fixing Re-Importing scan functionality

CCoffie:
I found a bug where when you went to re-import a scan it wouldn't add the endpoints or requests for the new findings.


devGregA:
Hi @CCoffie great find and fix!

PR #217, opened 2017-02-08
Add @mentions to notes

YSaxon:
As I mentioned in slack, it seems to me that notes ought to have an @username type notification system like jira or github etc. Here's a go at making one.

### DIFF
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -26,7 +26,7 @@ from dojo.forms import NoteForm, CloseFindingForm, FindingForm, PromoteFindingFo
     DeleteFindingTemplateForm, FindingImageFormSet, JIRAFindingForm, ReviewFindingForm, ClearFindingReviewForm
 from dojo.models import Product_Type, Finding, Notes, \
     Risk_Acceptance, BurpRawRequestResponse, Stub_Finding, Endpoint, Finding_Template, FindingImage, \
-    FindingImageAccessToken, JIRA_Issue, JIRA_PKey, JIRA_Conf, Dojo_User, User
+    FindingImageAccessToken, JIRA_Issue, JIRA_PKey, JIRA_Conf, Dojo_User
 from dojo.utils import get_page_items, add_breadcrumb, FileIterWrapper, send_review_email, process_notifications
 from dojo.tasks import add_issue_task, update_issue_task, add_comment_task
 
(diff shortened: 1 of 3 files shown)


devGregA:
@YSaxon awesome!

devGregA:
@aaronweaver @grendel513 I have reviewed this code and do not have any concerns, would either of you like to way in? Or @cneill?

devGregA:
@CCoffie @jpbowie if either of you are available to provide feedback, it would be greatly appreciated 😃

grendel513:
@YSaxon (cc: @devGregA, @aaronweaver) this is great!  before we merge, I would like to see it incorporated into the visual alerts on the upper right hand side.  This will be great for those users that do not have an email address associated with their account.  You can see how the alerts are harvested here https://github.com/OWASP/django-DefectDojo/blob/master/dojo/utils.py#L598

I think it will be possible to add an alert by querying all notes with an @mention and adding it to the list if it is for that user.  You can use the contains syntax https://docs.djangoproject.com/en/1.9/ref/models/querysets/#std:fieldlookup-contains

Thoughts?

devGregA:
@grendel513 would that work correctly when you wanted to dismiss the alert? If you just queried it as mentioned?

grendel513:
@devGregA - currently there is no way to dismiss an alert, the are timing based (last 7 days).  Some action on Findings for example (touching the finding, updating it, etc) will clear that alert.  However dismissing it is not a feature now.

YSaxon:
@grendel513 @devGregA Here's a first attempt at alerts.

YSaxon:
Another approach for alerts , in order to reduce number of alerts that could pop up from a single conversation
```
    atmentions=Notes.objects.filter(date__range=[start, now],
    entry__icontains='@'+user.username)
    findings_with_atmention=set([Finding.objects.get(notes=note)
                    for note in atmentions if Finding.objects.filter(notes=note).exists()])
    for f in findings_with_atmention:
        alerts.append(['You were mentioned in note(s) on Finding:{}'.format(f.title),
                       'Posted ' , #find the latest one posted
                       'file-text-o',
                       reverse('view_finding', args=(f.id,))])
```
Another thing I'd like is for it to not show an alert for any conversation in which the user poster after the latest @mention of him. Would need to check the order of notes posted on a finding.

grendel513:
@YSaxon - I like it, you can always check to see if there are notes posted after the one found with the @mention

jpbowie:
Looks good to me.

devGregA:
Merged. Thanks guys!

Issue #213, opened 2017-01-30
When creating a new user, enabling staff/superuser status is not effective


mindtap:
I just noticed in creating a new user (Users -> wrench icon -> New User) that selecting the checkbox to make the new user a Staff or SuperUser does nothing. The resultant user is still not those things

Not the biggest deal, since you can edit it and add those attributes immediately afterwards, but a bit confusing. If there is in fact some legitimate reason why not to allow creating a new user as a staff/superuser, then those checkboxes should be removed. Otherwise they should be made to work.

YSaxon:
That was actually me that posted that issue. Didn't realize I was signed in to GitHub on another account.

aaronweaver:
@YSaxon This is how it should behave:

**Active**
Designates whether this user should be treated as active. Unselect this instead of deleting accounts.
**Staff status**
Designates whether the user can log into this site.
**Superuser status**
Designates that this user has all permissions without explicitly assigning them.

Can you confirm that staff has superuser status?

devGregA:
@YSaxon is referring to when a user is created from the 'Users' sections the permissions aren't set correctly. You then have to navigate to /admin to set the password and permissions.

YSaxon:
Just to clarify:
Here's adding a new user and selecting that the user should be a staff and su: 
<img width="879" alt="screen shot 2017-02-11 at 9 59 11 pm" src="https://cloud.githubusercontent.com/assets/11711101/22859312/83cd274a-f0a5-11e6-86a9-cd81f93a1569.png">



And here is what you see after you click the add user button
<img width="872" alt="screen shot 2017-02-11 at 9 59 46 pm" src="https://cloud.githubusercontent.com/assets/11711101/22859317/97ed98a4-f0a5-11e6-9082-69c105c08f2a.png">


Note that the checkboxes have unchecked themselves because of [this code here](https://github.com/OWASP/django-DefectDojo/blob/master/dojo/user/views.py#L173).

Should you reselect the checkboxes and hit edit user, it _will_ work. But this is unnecessary and confusing. If the boxes are there, they should work the first time.

YSaxon:
Note that the same issue is true for Active status as well, except that it overrides the user's choice and forces Active to true. Not sure how often anyone will want to create an inactive account, but perhaps for consistency it should be addressed as well.

devGregA:
[...]

PR #221, opened 2017-02-12
Fix213 - make checkboxes work on add

YSaxon:
One of three alternatives I'm providing for fixing #213. This would make the checkboxes actually work as advertised the first time.


devGregA:
Thanks again @YSaxon the consensus was for this option.

PR #241, opened 2017-03-05
api syntax fix

szeist:
Hi,

There is typo in MultipartResource class name which breaks the API.
It would be great if it could be fixed in the master version. :)

All the best,
Istvan

### DIFF
--- a/dojo/api.py
+++ b/dojo/api.py
@@ -121,7 +121,7 @@ class MultipartResource(object):
             data.update(request.FILES)
             return data
 
-        return super(MultiPartResource, self).deserialize(request, data, format)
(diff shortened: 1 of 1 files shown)


aaronweaver:
Thanks @szeist

devGregA:
Hi @aaronweaver @szeist, greatly appreciate the PR, but I believe it was correct as is. Please this for other multipart examples http://stackoverflow.com/questions/25182131/multipart-form-data-post-put-patch-in-django-tastypie

devGregA:
If you saw my deleted comment. Never-mind! scratch that. I misread the files. Thank you @szeist! greatly appreciated.

PR #292, opened 2017-05-30
Update parser.py

jordanpotti:
Added support for DefectDojo to recognize mitigation status from Veracode XML imports.
Issue# 290

### DIFF
--- a/dojo/tools/veracode/parser.py
+++ b/dojo/tools/veracode/parser.py
@@ -52,27 +52,51 @@ class VeracodeXMLParser(object):
                             references = description[description.index('References:') + 13:].replace(')  ', ')\n')
                         else:
                             references = 'None'
-
+			false_p = 0
                         if 'date_first_occurrence' in flaw.attrib:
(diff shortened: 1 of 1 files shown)


devGregA:
Hi @jordanpotti thank you for your work on this! Would it not be better to mark a mitigated finding as `active=False` rather than `false_p=1`

jordanpotti:
I'm not sure. In my eyes, if I have it marked as mitigated in the Veracode platform, the finding is basically a false positive when consolidating with other findings in DefectDojo. However, that might not be the best option for most users, I can change that to leave `false_p=1` out and since `active=False` is default, it shouldn't require any other changes.

devGregA:
@grendel513 @aaronweaver @mtesauro can I get a second review please?

devGregA:
Hi @jordanpotti  really appreciate you working on this. If you don't mind sending me an address and a shirt size to greg.anderson@owasp.org. We'll send you some dojo swag.

PR #297, opened 2017-06-03
[WIP] Upgrade to Django 1.11

### LINKED ISSUE #250: Update Django Version to 1.11 LTS
### Make sure you assign this issue to yourself so folks do not duplicate effort.

Currently DefectDojo uses Django 1.8(LTS). Update all setup files necessary to make Dojo compatible with the new LTS version of Django (1.11). This includes but is not limited to setup.py and manage.py. All CI tests must pass and all supported install options must work.

narayanaditya95:
Fixes #250


dojo/models.py
```diff
@@ -832,11 +832,11 @@ def get_response(self):
 
 class Risk_Acceptance(models.Model):
     path = models.FileField(upload_to='risk/%Y/%m/%d',
-                            editable=False, null=False,
+                            editable=True, null=False,
```
narayanaditya95: Non-editable fields cannot be included in modelforms. Will try making them read-only in their respective modelforms. Is this the right approach?

dojo/filters.py
```diff
@@ -791,15 +768,15 @@ class LogEntryFilter(DojoFilter):
 
     class Meta:
         model = LogEntry
-        exclude = ['content_type', 'object_pk', 'object_id', 'object_repr', 'changes']
+        exclude = ['content_type', 'object_pk', 'object_id', 'object_repr', 'changes', 'additional_data']
```
narayanaditya95:
```
AssertionError: LogEntryFilter resolved field 'additional_data' with 'exact' lookup to an unrecognized field type JSONField.
```
What should the lookup look like?

setup.py
```diff
@@ -41,6 +41,7 @@
         'pycrypto'],
     dependency_links=[
         "https://github.com/grendel513/python-pdfkit/tarball/master#egg=pdfkit-0.5.0",
+        "https://github.com/aropan/django-tastypie-swagger/archive/fix-compatibility-django-1.10.zip#egg=django-tastypie-swagger-dev",
```
narayanaditya95: Mainline isn't Django 1.11 compatible. This is a PR branch.

dojo/filters.py
```diff
@@ -228,18 +228,6 @@ def __init__(self, *args, **kwargs):
...
-        except Finding.DoesNotExist:
-            earliest_finding = None
-
-        if earliest_finding:
-            start_date = local_tz.localize(datetime.combine(
-                earliest_finding.date, datetime.min.time())
-            )
-            self.start_date = _truncate(start_date - timedelta(days=1))
-            self.end_date = _truncate(now() + timedelta(days=1))
-
```
narayanaditya95: We cannot make DB queries in the __init__ method. Not sure why previous versions of Django did not complain about it.

dojo/models.py
devGregA: Hi @narayanaditya95  great catch. It looks like in 1.9 forward we use field.disabled

https://docs.djangoproject.com/en/1.9/ref/forms/fields/#disabled

dojo/filters.py
devGregA: @grendel513 this is the one I need your input on.

dojo/filters.py
devGregA: @grendel513 this one too please

grendel513:
I tried to take a look but:


​

PR #308, opened 2017-06-28
Fix queryset errors

patriknordlen:
This PR fixes a bug with a missing qs reference in the history view as well as a bug that makes the search view throw an error due to a double qs reference.


devGregA:
Merged. CI tests are currently not functioning. Thank you for the PR @patriknordlen!

PR #313, opened 2017-07-01
Migrate application settings to model

patriknordlen:
This PR addresses (or at least attempts to address :) ) https://github.com/OWASP/django-DefectDojo/issues/215.

It adds the model "System_Settings" for keeping system settings in the database, along with views and forms to change them in the web UI.
Also changes have been made to all components using these settings so that they refer to the settings from the model rather than settings.py.

Settings that have been added to the model and migrated in the application are:
- Enabling/disabling findings deduplication (ENABLE_DEDUPLICATOIN)
- Enabling/disabling JIRA integration (ENABLE_JIRA)
- Enabling/disabling S finding severity naming (S_FINDING_SEVERITY_NAMING)
- Specifying URL prefix (URL_PREFIX)
- Specifying timezone (TIME_ZONE)
- Specifying team name (TEAM_NAME)

Settings that are missing from this PR that were part of the original issue:
- Mail Server - from what I understand this is used internally by Django so might require another approach to get to work?

Caveats:
- Setting url_prefix and time_zone in the web UI needs a restart to take effect. 
- If the specified url_prefix does not point to a valid defectdojo instance this will render the web UI unusable and correct url_prefix needs to be set directly in the database.
- Database will be populated with default settings using a fixture on a new install but I haven't written anything to migrate existing settings in settings.py automatically.

There's, uh, also a couple of additional commits related to look and feel + bug fixes:
- I changed the behaviour for help_text to be displayed as a popover over a question mark next to the field name.
- The report builder's functionality to add hidden inputs when a user has tried to apply filter terms wasn't marked as safe so was HTML escaped, so any filter terms the user entered would not be applied to the generated report. Fixed by escaping individual parameters from the original query string but marking the final output as safe.
- The search was throwing an error due to wrong watson property being called. Fixed by calling the correct property instead. :)

I'm used to straight and honest PR feedback so don't hold back - I'm new at contributing to this project and I'm sure I've gotten a thing or two wrong.

### DIFF
--- a/dojo/templates/dojo/view_eng.html
+++ b/dojo/templates/dojo/view_eng.html
@@ -1,6 +1,6 @@
 {% extends "base.html" %}
 {% load display_tags %}
-{% load get_config_setting %}
+{% load get_system_setting %}
 {% load humanize %}
 {% load static from staticfiles %}
 {% block add_styles %}
--- a/dojo/templates/dojo/view_eng.html
+++ b/dojo/templates/dojo/view_eng.html
@@ -639,7 +639,7 @@
                                 <th>Mitigation Date</th>
                                 <th>Verified</th>
                                 <th>Active</th>
-                                {% if "ENABLE_JIRA"|get_config_setting %}
+                                {% if "enable_jira"|get_system_setting %}
                                     <th>Jira</th>
                                 {% endif %}
                                 <th>Actions</th>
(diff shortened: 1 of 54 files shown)


devGregA:
@patriknordlen thank you so much for doing this! Will clone and review shortly.

devGregA:
@patriknordlen just wanted to provide an update. Will be able to clone and review on Wednesday

patriknordlen:
Cool! Don't sweat it :)

devGregA:
Hi @patriknordlen thank you again for working on this!
The only thing I think would should add is a section in upgrades on readthedocs documenting what folks should due to upgrade their installs https://github.com/OWASP/django-DefectDojo/blob/master/docs/upgrading.rst. I think we should put a date entry so it is easier for people to see in a separate section such as:

-----
July 5th
load the systems settings fixture `python manage.py loaddata system_settings.json'
remove the following from settings.py
------

I'm happy to merge as is though and due the documentation if that is confusing / you don't want to :)
Also if you would like dojo swag as a reward for knocking this out please email greg.anderson@owasp.org with the address to mail to.

patriknordlen:
Thanks for the review! I'll go make the documentation changes tomorrow and submit, so let's do the merge after that. :)

patriknordlen:
@devGregA documentation added - please have a look.

dojo/templates/dojo/view_eng.html
```diff
@@ -649,14 +649,16 @@ <h4 class="pull-left">
                             {% for finding in findings %}
                                 <tr class="{% if finding.active %}active_finding{% else %}inactive_finding{% endif %}">
                                     <td><a href="{% url 'view_finding' finding.id %}">{{ finding.title }}</a></td>
-                                    <td>{% if finding.severity == "Critical" or finding.severity == "High" %}
-                                        <span class="text-error">
-                                    {% else %}<span>{% endif %}{{ finding.severity }}</span></td>
+                                    <td>
+                                        <span class="label severity severity-{{ finding.severity|safe }}">
```
devGregA: @patriknordlen last question I promise :joy, would you please explain this change? I understand the nav / label changes but I'm not sure why we're marking the severity as safe

patriknordlen: Haha, that's ok! :smile: 

That's a good catch, it's from my initial misunderstanding of what the safe function did :roll_eyes: 

I'll change that and go through any other uses of |safe in my commits to make sure they're actually needed!

patriknordlen:
There, done!

devGregA:
@patriknordlen merged :)

PR #315, opened 2017-07-05
Adds leads to tests

devGregA:
Adds leads to tests.
Modifies /engagement for easier access to tests
**Changes Product models for tech manager, product manager, etc back to the legacy text fields since I'm not aware of a good way to currently / easily add their information**

### DIFF
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -309,7 +309,7 @@ def add_tests(request, eid):
         if form.is_valid():
             new_test = form.save(commit=False)
             new_test.engagement = eng
-            # new_test.lead = User.objects.get(id=form['lead'].value())
+            new_test.lead = User.objects.get(id=form['lead'].value())
             new_test.save()
             tags = request.POST.getlist('tags')
             t = ", ".join(tags)
@@ -338,6 +338,8 @@ def add_tests(request, eid):
                 return HttpResponseRedirect(reverse('view_engagement', args=(eng.id,)))
     else:
         form = TestForm()
+	form.initial['target_start'] = eng.target_start
+	form.initial['target_end'] = eng.target_end
     add_breadcrumb(parent=eng, title="Add Tests", top_level=False, request=request)
     return render(request, 'dojo/add_tests.html',
                   {'form': form,
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -309,7 +309,11 @@ def add_tests(request, eid):
         if form.is_valid():
             new_test = form.save(commit=False)
             new_test.engagement = eng
-            new_test.lead = User.objects.get(id=form['lead'].value())
+	    try:
+            	new_test.lead = User.objects.get(id=form['lead'].value())
+	    except:
+		new_test.lead = None
+		pass
             new_test.save()
             tags = request.POST.getlist('tags')
             t = ", ".join(tags)
--- a/dojo/forms.py
+++ b/dojo/forms.py
@@ -555,7 +555,7 @@ class TestForm(forms.ModelForm):
                                      "Choose from the list or add new tags.  Press TAB key to add.")
     lead = forms.ModelChoiceField(
 	queryset=User.objects.exclude(is_staff=False),
-	required=True, label="Testing Lead")
+	required=False, label="Testing Lead")
 
 
     def __init__(self, *args, **kwargs):
--- a/dojo/forms.py
+++ b/dojo/forms.py
@@ -181,7 +181,7 @@ class ProductForm(forms.ModelForm):
 
     class Meta:
         model = Product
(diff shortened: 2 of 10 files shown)


devGregA:
@patriknordlen would you mind reviewing if you have time? :)

dojo/forms.py
```diff
@@ -553,6 +553,9 @@ class TestForm(forms.ModelForm):
                            required=False,
                            help_text="Add tags that help describe this test.  "
                                      "Choose from the list or add new tags.  Press TAB key to add.")
+    lead = forms.ModelChoiceField(
+	queryset=User.objects.exclude(is_staff=False),
+	required=True, label="Testing Lead")
```
patriknordlen: Setting this field to required causes problems in some parts of the app, for instance when trying to delete a test you're first required to set a test lead before being allowed to delete it. It's probably fairly common as well to know that a test is coming up but no decision has been made about who will take the lead on it, so I think this field shouldn't be required unless there are good reasons that outweigh the mentioned issues.

dojo/templates/dojo/engagement.html
```diff
@@ -60,8 +60,15 @@ <h3 class="has-filters">
...
                                     {% for e in p.engagement_set.all %}
                                         {% if e.active %}
-                                            <a class="eng_link" href="{% url 'view_engagement' e.id %}">
+						<div class="lineContainer">
+							<a  style="display: inline" class="eng_link" href="{% url 'view_engagement' e.id %}">
                                                 {% if e.name %}{{ e.name }} {% endif %}{{ e.target_start }}</a>
+							| Lead : {{ e.lead.first_name }}
+                                                        |
+                                                        {%  for test in e.test_set.all %}
+                                                                <a href="{% url 'view_test' test.id %}"> {{ test.test_type }} : {{ test.lead.first_name }} </a> |
```
patriknordlen: Maybe only add lead info if lead is set (if lead is not a required field, see other comment)

dojo/engagement/views.py
```diff
@@ -338,6 +338,8 @@ def add_tests(request, eid):
                 return HttpResponseRedirect(reverse('view_engagement', args=(eng.id,)))
     else:
         form = TestForm()
+	form.initial['target_start'] = eng.target_start
```
patriknordlen: Great :+1:

dojo/templates/dojo/engagement.html
```diff
@@ -60,8 +60,15 @@ <h3 class="has-filters">
                                 <td>
                                     {% for e in p.engagement_set.all %}
                                         {% if e.active %}
-                                            <a class="eng_link" href="{% url 'view_engagement' e.id %}">
+						<div class="lineContainer">
+							<a  style="display: inline" class="eng_link" href="{% url 'view_engagement' e.id %}">
                                                 {% if e.name %}{{ e.name }} {% endif %}{{ e.target_start }}</a>
+							| Lead : {{ e.lead.first_name }}
```
patriknordlen: I like that this is displayed directly in the overview. Nice!

dojo/forms.py
devGregA: Thanks @patriknordlen great catch. Will get it fixed, and also display conditionally.

devGregA:
@patriknordlen should be fixed :) I also added documentation on labels and did a Dojo in production guide.

dojo/forms.py
```diff
@@ -181,7 +181,7 @@ def __init__(self, *args, **kwargs):
 
     class Meta:
         model = Product
-        fields = ['name', 'description', 'tags', 'product_manager', 'technical_contact', 'team_manager', 'prod_type',
+        fields = ['name', 'description', 'tags', 'prod_manager', 'tech_contact', 'manager', 'prod_type',
```
patriknordlen: I missed this in my former review. The fields used in the product view page are product_manager, technical_contact and team_manager, so when adding a new product with these changes the "Team Manager" etc fields will show up as "Unknown".

When looking into this further it seems these fields introduced in this commit are marked as deprecated, personally I prefer keeping the old freetext fields rather than the 'new' ones that require these roles to be tied to users in the application.

devGregA:
Hi @patriknordlen  I went through product/views.py, engagement/views.py, templates/dojo/product.html, templates/dojo/engagement.html, and forms and didn't see any mistmatches

patriknordlen:
@devGregA, I found it on the product view page, here: https://github.com/OWASP/django-DefectDojo/blob/master/dojo/templates/dojo/view_product.html#L357

devGregA:
I was looking in the wrong template 😂 should be fixed now. Thanks @patriknordlen

patriknordlen:
:+1: LGTM!

devGregA:
Merged. Thanks!

Issue #228, opened 2017-02-21
Status Updates via Slack / Email


devGregA:
### Make sure you assign this issue to yourself so folks do not duplicate effort.

To complete this issue, Dojo users should be able to configure slack / email notifications for issues.

Implementation Requirements:

- [ ] Create a configuration page that has a checkbox matrix to assign when notifications will be sent out.
- [ ] Integrate with slack APIs to push issues to a **channel**
- [ ] Use the email integration present with scan profiles to send notifications via this new system.
- [ ] Add documentation to the /docs folder.

Resources:
Slack Documentation:  https://api.slack.com/custom-integrations

patriknordlen:
I worked on this tonight and have notification settings and slack integration working. Next up are mail and docs, as well as adding support for notifications for more events.

PR #318, opened 2017-07-11
Notifications

patriknordlen:
This PR addresses issue #228.

It adds a new model "notifications" which has a number of events that can be set up to trigger notifications to be sent by mail and/or as a Slack message. This can be done globally as well as on a per-user basis, so individual users can set up notifications of their own without needing to request assistance from an admin. Destinations for personal notifications are the user's mail address and Slack username, respectively.

Initial events that have been setup to offer notifications are:
- Engagement added
- Test added
- Results added to test
- Report created

Slack messages are sent using "django_slack" which has been added to the INSTALLED_APPS section of settings.dist.py.


devGregA:
@grendel513 do you have time to review?

devGregA:
Hi @patriknordlen awesome stuff as usual :) would you please add a section to https://github.com/OWASP/django-DefectDojo/blob/master/docs/features.rst documenting how to setup the integration?

https://github.com/OWASP/django-DefectDojo/blob/master/docs/features.rst generates
http://defectdojo.readthedocs.io/en/latest/features.html

grendel513:
Does this replace or integrate with the current alerts in dojo?

I think it should replace them, notifications and alerts seem redundant and
alerts needs a revamp anyway.

patriknordlen:
@grendel513 it does not replace alerts, it's a pure addition only. It wouldn't be too hard to add alert as a notification type in addition to the ones already in there...I haven't looked into revamping alerts though and build something that automatically does something the other way around (create notifications for anything that pushes an alert), that could possibly be more work. :)

grendel513:
I'm not too keen on having two separate alert/notification mechanisms in
dojo...we need to work on combining them and making sure both
functionalities remain.  Since your notification mechanism appears to be
more mature/robust, I recommend taking a look at the current alert
functionality and finding a way to combine them.

I think this will confuse users who are not aware they are separate an
unrelated.

my two cents

patriknordlen:
Definitely sounds fair. I don't mind working something out, just want to make sure I go for an approach that seems sensible to others as well.

My first thought would be my suggestion above, that is, have alerts work as another notification action. So users could then set up either mail, slack or alert as notification for events. Things that use alerts directly today would then need to be rewritten to use notifications instead, and the logic for which action to take would then be handled there.

grendel513:
I think that sounds like an appropriate approach.  See
https://github.com/OWASP/django-DefectDojo/issues/220 for additional
comments on alerts.  You can see what type of events/situations trigger an
alert in the code here
https://github.com/OWASP/django-DefectDojo/blob/master/dojo/utils.py#L612

I feel we should incorporate those events as well.

Thanks for your willingness to take this on!! We really appreciate it.

Jay

devGregA:
I agree that in the long run they should be one entity, but I personally would like slack integration sooner rather than later ¯\_(ツ)_/¯. @grendel513 would you be okay with merging as if functional and then we can create a separate issue?

devGregA:
I also think before we combine them we would need to decide what displays where. Right now alerts serve to notify you when 1. something fails 2. something hasn't been completed. As an admin I wouldn't want notifications every time something is opened / closed etc. I see a lot of granularity that needs to be defined that is outside of the issue @patriknordlen is working to solve.

grendel513:
the key phrase here is "in the long run" - it is my experience that those
things left undone become technical debt...if we are in there now, we
should get it done ;)

patriknordlen:
One thing about current alerts is that they're not necessarily triggered by events. Things like stale or upcoming engagements/scans are not "generated" so much as they're returned when specifically asking for that info. To move that over to run via notifications there would have to be something like a cron job that runs regularly and collects the same info instead. Unless there's some nifty Django feature that I don't know about that solves this.

devGregA:
@aaronweaver would you please share your thoughts?

patriknordlen:
I've reworked notifications and alerts a bit so they're integrated now. As for notifications that would require some kind of period check - I'm thinking celery periodic tasks might be possible to use for that? My knowledge of it is limited to that it came up when I googled for if that kind of thing is possible with celery. Opinions, thoughts?

grendel513:
Yes, celery is the right approach, and it is already a part of dojo for
reports and jira.  That should be the way to create tasks that are executed
at regular intervals -
http://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html

patriknordlen:
Thanks for the input @grendel513, I'll get right on it. :)

grendel513:
take a look at
https://github.com/OWASP/django-DefectDojo/blob/master/dojo/tasks.py to see
how these are implemented.  additional tasks can be added to this file to
take care of the notifications.  thanks again for attempting to get this
done

patriknordlen:
Just a short update on this. I've worked on it a bit and have some local changes that I haven't committed yet because I'm still rewriting things here and there.

- I've added alerts as another notification option.
- I've changed all parts of the application that notify the user by mail or alert in some way to use the new notification functionality instead.
- I've implemented periodic tasks with celery that will check things that were previously in get_alerts and send notifications based on that.

In other words, progress is being made, so no need to merge this in its current state :)

grendel513:
@patriknordlen just wondering how this was coming along?

patriknordlen:
@grendel513 making progress now - I'm back from a 4w vacation which is why there's been a lack of updates. Functionality is done, working now on testing functionality in various situations and ensuring a smooth migration.

patriknordlen:
Ok! Summary of the changes:

- All types of notifications in the application is now handled via the function 'create_notification' in utils.py. This function will in turn send notifications via mail, Slack and/or alerts if this has been enabled either on a global or personal scope in the Notifications page.
- Notifications are based on templates found in the dojo/notifications directory. Templates are matched using the 'event' argument to create_notification. If no template is found, the template 'other.tpl'  will be used.
- Alerts are no longer loaded in full when loading defectDojo pages. An alert count endpoint is called every 10 seconds, but full alert data is only loaded when the user clicks on the dropdown.
- The alerts page now has functionality to select and remove alerts.
- Alerts that were previously dynamically retrieved each time a user requested his/her alerts (upcoming engagements, stale engagements) have been moved to be run as periodic tasks using celery beat instead. These tasks will run every hour and create alerts according to their configuration.

patriknordlen:
@devGregA @grendel513 please have a look - with a PR of this size I'm sure there are problems that I've missed even after testing. For instance I don't have a jira instance handy to see if everything works properly there.

devGregA:
@aaronweaver would you also have time to test?

grendel513:
@devGregA, @aaronweaver, @patriknordlen - I tested the UI as well as the Alerts and email integrations.  All work great.  I have not tested integration with Slack, Hipchat, or Jira.

I think we need to add documentation for this new feature...especially the need to:
    `pip install django-slack`
    `pip install django-multiselectfield`
    `./manage.py makemigrations`
    `./manage.py migrate`

We also need to make sure we mention that the celery process needs to be running for all of this to work.

Thanks!

devGregA:
@grendel513 I have tested the JIRA changes, but not the notifications + JIRA. My suggestion is to get the documentation in as you noted, but then I would be okay with merging and looking at it as 'beta features' I have some code I want to submit relating to handling false positives, but I want to get this done first so @patriknordlen doesn't have to deal with conflicts. Would you be okay with merging once the docs are in place and treating them as 'beta' features for maybe a week? By then I will have plenty of time to thoroughly test the changes at my work.

patriknordlen:
Sounds good @devGregA! I'll work on the docs tonight.

patriknordlen:
@devGregA @grendel513 sorry for the delay - work etc. I've added a new section to the 'Features' page covering notifications in brief detail.

`django_slack` is now gone as a dependency (it's not used anymore, notifications are now sent directly using the `requests` library) and django-multiselectfield is added as a dependency in `setup.py`. Users are fine if they run `pip install .` which will take care of installing all dependencies for them, so I added that command to the list of default commands to run when upgrading. `makemigrations` and `migrate` are already there.

devGregA:
Merged. Will roll out at my employer this week for testing.

Issue #218, opened 2017-02-08
NMAP scan results import


bobdob42:
could DefectDojo support Import of NMAP scan results, similar to Nessus or Burp XML
The nmap scan results would be in .XML format https://nmap.org/book/output-formats-xml-output.html

devGregA:
Hi @bobdob42 we haven't supported nmap as a vulnerability scanner in the past because we didn't feel that open ports equal vulnerabilities. If you're using the  nmap plugins that is different, but we do support nmap / port scans http://defectdojo.readthedocs.io/en/latest/features.html#port-scans

bobdob42:
Hi Greg
was thinking in terms of extending that "port-scans" feature, the DefectDojo server may not have access to the same parts of the infrastructure to allow those scans or the scans maybe run by a 3rd party and we want to import the results.

devGregA:
@bobdob42 ah great point. Definitely would be a great enhancement. Unfortunately, I'm not sure when I'll be able to get to it due to my backlog.

patriknordlen:
I have a working prototype for this, it essentially creates informational findings for each open port. Issuing a PR shortly.

devGregA:
@bobdob42 done :) courtesy of @patriknordlen

PR #319, opened 2017-07-11
nmap XML file parsing support

patriknordlen:
This PR addresses issue #218.

It adds a parser for nmap XML files, and creates informational findings with basic details about identified services from them.

It also extends the Endpoint model to include port and fqdn in order to improve filtering and reporting possibilities in the future.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -416,10 +416,12 @@ class CWE(models.Model):
 
 class Endpoint(models.Model):
     protocol = models.CharField(null=True, blank=True, max_length=10,
-                                help_text="The communication protocl such as 'http', 'ftp', etc.")
+                                help_text="The communication protocol such as 'http', 'ftp', etc.")
     host = models.CharField(null=True, blank=True, max_length=500,
                             help_text="The host name or IP address, you can also include the port number. For example"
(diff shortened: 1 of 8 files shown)


dojo/models.py
```diff
@@ -438,14 +440,19 @@ def __unicode__(self):
...
         query = self.query
         fragment = self.fragment
 
+        if port:
+            netloc += ':%s' % port
+
         if netloc or (scheme and scheme in uses_netloc and url[:2] != '//'):
             if url and url[:1] != '/': url = '/' + url
-            if scheme:
+            if scheme and scheme in uses_netloc and url[:2] != '//':
```
patriknordlen: I've done some testing against local files on this and it seems to work but the logic here is complicated enough that I don't fully trust my own judgment here...so please test this :)

devGregA:
@patriknordlen will test first thing tomorrow

devGregA:
@patriknordlen  tested and verified even using options such as -A. @patriknordlen would you please capitalize the 'n' in nmap in dojo/templates/dojo/import_scan_results.html, dojo/forms.py, and dojo/tools/factory.py for consistency  :) oh! and please add a line to https://github.com/OWASP/django-DefectDojo/blob/master/docs/features.rst#imports adding nmap

patriknordlen:
Done @devGregA!

Issue #237, opened 2017-02-27
Long-term Engagement breakup


humor4fun:
One thing that I plan to do regularly is to run an engagement with a team: 'starting this week, I'll do a burp scan, web app pentesting for 2 weeks, and then in about 2 months we will circle back around and do another burp scan.'

It makes sense to include this sort of thing as _three_ **Tests** on _one_ **Engagement**, but the calendar will show the engagement running for the entire duration of the 3 months, when in reality it should only show the engagement as _active_ for the duration of the **Tests**.

I propose that the Calendar (and other date-operations) should read its information from the Test date fields if available rather than the Engagement date fields. This allows me to schedule long-term engagements and future tests/engagements without seeing engagements on the calendar that are not active. Currently in order to do this I'd create one engagement for now and one for later, call them Engagement (part 1/2) and Engagement part (2/2). 

Perhaps a multi-part engagement makes more sense, in that case, allowing Engagements to be linked together for metadata and reporting would solve the issue.

devGregA:
I agree, my thought is to create a toggle between tests / engagements for the calendar

patriknordlen:
I have a working prototype for this. Switch between engagements/tests for the calendar as well as filtering calendar by lead and showing info about who's the lead for the project. Will issue a PR when/if other current PR is accepted as some of the code relies on that. :)

devGregA:
@patriknordlen awesome stuff! After we get all your stuff merged I have a fork that adds leads to tests which I think should be helpful. My one issue is that it makes the 'active engagements' page look a little cluttered. Once we get this in it would be great to get your input :)

devGregA:
@humor4fun please see @patriknordlen's new code

PR #323, opened 2017-07-13
View and filter tests/engagements and users in calendar

patriknordlen:
This PR addresses issue #237.

It adds functionality in the calendar to view calendars by tests or engagements, as well as filtering those by a set of users or unassigned engagements/tests.

image


devGregA:
Hi @patriknordlen awesome stuff as always :) On my screen it looks a little tight. Could we add some padding?

screen shot 2017-07-13 at 4 11 34 pm

 I was also thinking that we should consider changing the link on the side menu to include /engagements? so that users immediately get the original functionality from the side menu. What are your thoughts? :)

PR #324, opened 2017-07-14
Changed dependencies list for YUM installation

WolfangAukang:
The nodejs and python-pip packages are available through epel-release. Also, libmysqlclient-dev package is not available on the repos and the nodejs installation script was installing a deprecated version.

This was tested on a CentOS 7 machine.


devGregA:
Merged @WolfangAukang thank you for your contribution!

PR #343, opened 2017-08-09
wrap pre-tags in the finding view

no-sec-marko:
The PR wraps the text in pre-tags in the finding view which improves the legibility of findings.

### DIFF
--- a/dojo/static/dojo/css/dojo.css
+++ b/dojo/static/dojo/css/dojo.css
@@ -642,6 +642,11 @@ td .fa-fw {
     background-color: transparent;
     border: 0px;
     padding: 0px;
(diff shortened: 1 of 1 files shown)


devGregA:
hey @no-sec-marko I merged, but now that I'm looking at the wrap info, why would we want to break words? `word-wrap: break-word;`

no-sec-marko:
Hi @devEricA , the code based on this code snippet [1] [2] and wraps the text for IE - untested ;)

[1] https://css-tricks.com/snippets/css/make-pre-text-wrap/
[2] https://css-tricks.com/snippets/css/make-pre-text-wrap/#comment-441142

PR #344, opened 2017-08-09
fix #338 missing tags in Template

### LINKED ISSUE #338: Missing tags in "Update Finding Template" and "Create Findings from Template"-function
Hi,

I setup a clean install of the defect dojo using the ./setup.bash (master branch) on the following system:

> DISTRIB_ID=CENTOS
> DISTRIB_RELEASE=7
> Django 1.11.2
> Dojo v1.1.4

and additionally setup a near-production conditions using nginx, celery and uwsgi following the guide "https://github.com/OWASP/django-DefectDojo/blob/master/docs/dojo-production.rst".

After adding some finding templates, I noticed that the tagging mechanism in the view is broken. The "add findings from template" and "edit template"-function doenst show the assigned tags on the finding.

A debug session identified the issue during the rendering process in the _dojo\finding\views.py_ file in class _def edit_template(request, tid)_ line 766.

The following code works fine and contains the assined tags:
`788:    form.initial['tags'] = [tag.name for tag in template.tags]`

But the _form_ object in _dojo/templates/dojo/form_fields.html_ does not contains the values anymore.  Unfortunately, I was not able to fix the described behaviour. Hopefully, you can reproduce the issue.

no-sec-marko:
Fix missing tags in "Update Finding Template" and "Create Findings from Template"-function

### DIFF
--- a/dojo/forms.py
+++ b/dojo/forms.py
@@ -781,7 +781,7 @@ class FindingTemplateForm(forms.ModelForm):
             'invalid_choice': 'Select valid choice: Critical,High,Medium,Low'})
 
     def __init__(self, *args, **kwargs):
-        tags = Tag.objects.usage_for_model(Finding)
+        tags = Tag.objects.usage_for_model(Finding_Template)
         t = [(tag.name, tag.name) for tag in tags]
         super(FindingTemplateForm, self).__init__(*args, **kwargs)
         self.fields['tags'].widget.choices = t


devGregA:
hi @no-sec-marko will review your PRs before the end of the week. Thank you for your contributions!

devGregA:
I **think** this is intentional @grendel513 can you confirm?

grendel513:
@devGregA honestly i cannot remember it tags were omitted from templates by design.  I can see an argument for bot including and omitting them.  Templates are meant to be general or more abstract, and adding tags to them makes them less so.  However, some tags may make sense and may not add a level of specificity that would detract from the abstract nature of a template.

Longwinded answer to say, I am good either way :)

no-sec-marko:
hi @grendel513 , @devGregA , I understand your objection but in this case the tag feature should be remove from the template view? In addition, after adding tags to a finding, an identical tag on an assigned template will suddenly appeared. The described behavior leads to the PR request ;)

PR #349, opened 2017-08-30
Use django's timezone mechanisms

patriknordlen:
This PR addresses an issue that possibly arose when timezone settings were moved to a model instead of settings.py, which is that this timezone setting wasn't honored.

The PR adds a middleware that will activate the proper timezone and changes lots and lots of datetime functions to instead use corresponding Django timezone aware functions using django.utils.timezone.


patriknordlen:
Caught a bug...issuing fix shortly.

devGregA:
@patriknordlen is this ready for testing / merging?

patriknordlen:
@devGregA it is!

devGregA:
@patriknordlen sanity checking now

PR #352, opened 2017-09-03
Fix several issues and documentation errors related to install/configuration

njfox:
This PR fixes several issues in both the install scripts and documentation and should make the install process a lot more robust. 

First, setup.bash would fail to set up the database properly because it did not grant itself permissions after creating it. It felt awkward to me to have the admin create a user but NOT a database, so I've modified the script to require the empty database to already have been created. This way, the admin creates the database the same time he creates the user, and then assigns privileges to the database, similar to the process involved in installing something like WordPress. Note that we could also solve this problem by modifying the script to assign itself necessary permissions after creating the database--feel free to discuss. I've also modified the script so that it will terminate with an error if it cannot access the database with the supplied information. Previously, it just continued to run and would throw a bunch of Python MySQL errors.

I've also updated the relevant documentation to reflect these changes and provided a walkthrough for running the MySQL commands.

Second, I've rearranged and revised several parts of the dojo-production.rst documentation, since there were errors, missing steps, etc. I've also fixed formatting issues to make it consistent throughout.

### DIFF
--- a/docs/dojo-production.rst
+++ b/docs/dojo-production.rst
@@ -1,7 +1,7 @@
 Running in Production
 =====================
 
-This guide will walk you through how to setup DefectDojo for running in production using Ubuntu 16.04, nginx, and uwsgi.
+This guide will walk you through how to setup DefectDojo for running in production using Ubuntu 16.04, MySQL, nginx, and uwsgi.
 
 **Install Required Packages**
 
--- a/docs/dojo-production.rst
+++ b/docs/dojo-production.rst
@@ -3,6 +3,28 @@ Running in Production
 
 This guide will walk you through how to setup DefectDojo for running in production using Ubuntu 16.04, nginx, and uwsgi.
 
+**Install Required Packages**
+
+.. code-block:: console
+
+  sudo apt-get install python-pip nginx mysql-server
+
+**Create MySQL Database and User**
+
+.. code-block:: console
+
+  mysql -u root -p
+
+  mysql> create user 'dojo'@'localhost' identified by '<password>';
+
+  mysql> create database defectdojo;
+
+  mysql> grant all privileges on defectdojo.* to 'dojo'@'localhost';
+
+  mysql> flush privileges;
+
+  mysql> quit
+
 *Install, Setup, and Activate Virtualenv*
 
 .. code-block:: console
@@ -11,7 +33,7 @@ This guide will walk you through how to setup DefectDojo for running in producti
 
   virtualenv dojo
 
-  source my_project/bin/activate
+  source dojo/bin/activate
 
 **Install Dojo**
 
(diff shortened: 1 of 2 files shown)


devGregA:
Hi @njfox I absolutely agree that how you have changed setup.bash is a better practice than what is currently in place, however, it increases setup overhead and is an additional barrier to entry for new users. @grendel513 @patriknordlen would you guys like to way in?

njfox:
I agree it adds a small bit of complexity when setting up MySQL but we're running into a bit of a chicken-and-egg problem with the database permissions currently, as the setup script runs and throws a bunch of MySQL errors, then appears to complete but hasn't actually done anything to the new database since it doesn't have any permissions on that database. I think this adds much greater overhead for new users than instructing them how to create a database and assign privileges for it.

There may be a safe way for the script to create the database and then assign itself privileges only on that database, but I'm not familiar enough with MySQL to say. For that to work, I'm guessing the user account would need additional grant privileges if those exist, which could have potential security implications if the Defect Dojo application was exploited.

grendel513:
@devGregA we have to maintain the balance here.  There are many Dojo users who are not devs/builders/sysadmins and will forgo even trying Dojo if the install process is too cumbersome.

njfox:
What if the script used the root MySQL user to create a user, assign the permissions, and create the database, then added the low-privileged user to settings.py?

devGregA:
@njfox that would be great! In my opinion that accomplishes both goals.

njfox:
a2d3523 ended up being a good deal more complicated than I was expecting, but I've tested it with a local MySQL database and it appears to be working. I have not tested it with a remote database. I'm also not an expert with bash so please let me know if there are any major errors or stylistic issues. Once everything looks good I will update the documentation with another commit before this branch is merged.

The new flow assumes the database and user do not exist and will offer to drop them or terminate if they do. It uses the mysql root account to create the database and user and assign permissions on the new database. If the MySQL database is on a remote server, it prompts the user for the IP of the local dojo server to allow the dojo user to connect remotely later. It then generates a random 20 character password from upper/lower/numeric to be used for the dojo account and stored in settings.py.

If we decide to go this route, we will need to update postgresql as well, but I'm not that familiar with PGSQL so it may be better if someone else wants to take that on.

devGregA:
hey @njfox I'm back from appsec usa and will be evaluating again shortly

devGregA:
@grendel513 @njfox finally getting back around to this. I really like where this is headed. I think there is however a  use case we haven't considered where a user setting up dojo might not have root access (ex: co-tenant on an RDS instance). Maybe we should create the option to create a new db or use one that already exists? I know it's been a while. I can modify this PR to realize this if the general flow is agreeable.

PR #353, opened 2017-09-03
Fix PDF Report Generation When Host Header Contains a Different External Port

njfox:
This patch fixes a bug that occurred when generating a report from the Engagements page when the client is using a different port in the host header than Defect Dojo is listening on locally (e.g., port forwarding/SSH tunneling, but it could happen in other scenarios). The backend task that attempts to generate the PDF report would try to connect to the external port and the connection would be refused. Now, we simply ignore the port by checking if there is a ":" in the host header and stripping it out if it does exist.


devGregA:
Tested with both report generation, and the custom report builder

PR #363, opened 2017-09-21
Fix credential links in getting started guide

spinkham:


devGregA:
Hi @spinkham, thank you for your PR! I agree it's wrong. However for some reason it does appear to work https://defectdojo.readthedocs.io/en/latest/getting-started.html let me consult with the other guys

spinkham:
Hmm, where I saw it was here: https://github.com/OWASP/django-DefectDojo/blob/master/docs/getting-started.rst 
If you read the wiki and get redirected to "read the docs" it's fine, if you just browse the /docs dir you get broken images.  I tend to head to /docs first...

devGregA:
Hi @spinkham we try to push people towards the readthedocs page as the /docs folder itself isn't intended to be used as the primary reference (readthedocs is populated off of /docs). I'll try to make that more clear in the actual folder. I double checked that it isn't possible to make it work for both.

PR #384, opened 2017-11-01
Updated/fixed Dockerfile and temporarily disabled Zap tests in Travis

mtesauro:
Docker file now works for reporting as it starts both celery processes needed by Dojo.  You can launch the docker like:

`$ docker run -d -p 8000:8000 --name=dojo mtesauro/defect-dojo:1.1`

And then go to http://localhost:8000 in a browser.  Default login is admin / admin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,8 @@ before_install:
     # - sleep 30 #allow dojo time to startup
   - sudo service docker restart
   - docker logs $CONTAINER_NAME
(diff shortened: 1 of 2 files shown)


devGregA:
@mtesauro https://travis-ci.org/OWASP/django-DefectDojo/builds/295957574?utm_source=github_status&utm_medium=notification

I think something still isn't quite right here.

mtesauro:
Will do a new PR to fix the curl connectivity issue.

PR #388, opened 2017-11-07
Updated/fixed Dockerfile and temporarily disabled Zap tests in Travis

mtesauro:
Docker file now works for reporting as it starts both celery processes needed by Dojo. You can launch the docker like:

$ docker run -d -p 8000:8000 --name=dojo mtesauro/defect-dojo:1.1

And then go to http://localhost:8000 in a browser. Default login is admin / admin

Note that you have to have -p 127.0.0.1:8000:8000 to keep Travis CI happy.


mtesauro:
Crap.  Had a 1 character typo.  Closing this.

PR #401, opened 2017-11-12
Add provisioning script for the production & development

secfigo:
This PR adds support for provisioning Defect Dojo on production and supports following features

1. HTTPS/TLS support for both the development and production environment with Let's Encrypt as CA.
2. Gunicorn as app server
3. Supervisor as process control (restarts automatically if something django goes down)
4. Supports both MySQL and PostgreSQL
5. Closely mirrors the installation script setup.bash
6. Follows all the best practices outlined in [Ansible best practices page](http://docs.ansible.com/ansible/latest/playbooks_best_practices.html) 

[README.md](ansible/prod-install/README.md) has all the instructions needed to get up and running.

checklist:


secfigo:
@aaronweaver @devGregA Aaron and Greg could you please review this PR and +1. Thanks.

secfigo:
If you are wondering about the docs, its available at https://github.com/secfigo/django-DefectDojo/blob/876398a64e42ef852a8ea63bbb6b5e0d5e94ceb7/ansible/prod-install/README.md 

I just wanted to see what could be the best place to put it, probably under docs/running-in-production.rst ?

aaronweaver:
@secfigo this is fantastic work. I'll look to test this installation out. A couple comments, where possible can we pull in the files for config and settings from the current location? For example, ansible/prod-install/files/requirements/production.txt can this instead be pull from our current python requirements? Same comment for the settings.py file.
Why I'm asking for this is to ensure going forward that we don't get out of synch with the files. If new requirements are added we want that propagated to all of our install scripts.

secfigo:
Thanks Aaron,

I agree with your comments.

Thats what I wanted to do but i couldn't find requirements.txt and
settings.py needs a overhaul to do that, for one, we need to pickup
variables from environment variables like passwords, port numbers etc., .

I intend to make defect dojo compatible with django cookie cutter in coming
weeks and that will smoothen this process of not having duplicates.

devGregA:
@secfigo this is some of the best work I've seen related to Dojo. I share Aaron's concerns about possibly breaking the ansible due to changes in settings.py. Let me look at how Matt did the dev ansible and I'll see if I can suggest a solution

devGregA:
@aaronweaver I looked at the original ansible Matt wrote, he does the same thing. It seems like it s a concern we need to solve as a project, but since we already have the issue present, I don't see any reason to not accept @secfigo as is. I'll do another read through to make sure I didn't miss anything else. Would also be great to get @mtesauro's eyes on it, but I don't think the settings / requirements issue is something to solve in this PR.

aaronweaver:
@devGregA I'm fine with merging it and I've created an issue to track this as the next enhancement for installation.

https://github.com/OWASP/django-DefectDojo/issues/402

secfigo:
Thanks @devGregA and @aaronweaver for the kind words. I will pick #402 next week but to give you guys heads up, it will be based on [Cookie Cutter](https://github.com/pydanny/cookiecutter-django)

devGregA:
@secfigo sounds great! If you would like a dojo shirt as a thank you, please send your size / address to greg.anderson@owasp.org

PR #409, opened 2017-11-29
Fixing SECURE_PROXY_SSL_HEADER setting

cneill:
Replacing `HTTP_X_FORWARDED_PROTOCOL` with `HTTP_X_FORWARDED_PROTO`.
Currently causes issues with how notifications are generated. Using the
Django `build_absolute_uri` helper, URLs are built with "http" instead
of "https" if the `SECURE_PROXY_SSL_HEADER` isn't found. The proper
header is X-Forwarded-Proto (see:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto)


cneill:
Need to edit other instances...

PR #417, opened 2017-12-07
run_dojo.sh: run 'python manage.py runserver' in background

xthk:
Hi all,

this is just a minor (and very small) fix for "run_dojo.sh". The way the shell script is written implies to run 'python manage.py runserver' in the background. If it should stay in the forground, lines 19-21 would be obsolete (https://github.com/OWASP/django-DefectDojo/blob/master/run_dojo.bash#L19-L21).

Thanks in advance,
Tilmann H.

----

- [N/A] Your code is pep8 compliant (Dojo's code isn't currently pep8 compliant, but we're trying to correct that)
- [N/A] If this is a new feature and not a bug fix, you've included the proper docuemntation under the /docs folder

### DIFF
--- a/run_dojo.bash
+++ b/run_dojo.bash
@@ -15,7 +15,7 @@ cpid=$!
 celery beat -A dojo -l info &
 bpid=$!
 
-python manage.py runserver
+python manage.py runserver &
 ppid=$!
 
 tail -f /dev/null


aaronweaver:
@mtesauro / @devGregA What's this script typically used for?

devGregA:
@aaronweaver I think we had issues collecting the PID when the process was in the foreground causing celery not to terminate properly. Will test when I can.

devGregA:
The scrip is for running both dojo and celery without two separate commands (for testing)  and terminating correctly when an interrupt signal is received (CTRL + C)

PR #445, opened 2018-01-23
more flexible bash path

tpltnt:
Hi there,

I adjusted the path to bash in the two shell scripts to also work in non-Linux environments. This does not touch the Python code.

Cheers
tpltnt

### DIFF
--- a/run_dojo.bash
+++ b/run_dojo.bash
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 function control_c()
 # run if user hits control-c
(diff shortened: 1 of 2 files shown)


cneill:
Which platforms are you interested in supporting that need this change? I'm inclined not to accept this one for the reasons outlined here: https://stackoverflow.com/questions/16365130/what-is-the-difference-between-usr-bin-env-bash-and-usr-bin-bash

> The downside is that since you aren't calling an explicit executable, it's got the potential for mistakes, and on multiuser systems security problems (if someone managed to get their executable called bash in your path, for example).

devGregA:
Closing for now.

PR #452, opened 2018-01-25
Updating readme

cneill:
- fixing images linking to the wrong repo
- removing references to github.com/OWASP/...
- removing HTML
- resizing images
- removing link to stickers since it no longer works
- adding https to links


devGregA:
Hi @cneill thanks for cleaning this up. What is the advantage of shrinking the images and using markdown as opposed to resizing with html?

devGregA:
Merged. I think we may want to shrink them a little more.

PR #456, opened 2018-01-31
Import OWASP SKF issues.

martinmarsicano:
With this commit you can import the failed requirements of a OWASP SKF sprint.


devGregA:
@martinmarsicano Code looks great. Any chance you have a sample scan we could use for testing?

devGregA:
Also please update the import list here: https://github.com/DefectDojo/django-DefectDojo/blob/master/docs/features.rst#imports so many places to update I know 😂

martinmarsicano:
Docs features, updated. I am having a problem with Finding.date, I am assigning that value (already have tried with various format), but then in the Dojo UI that date is not being the one I put.

And example of port of csv:
date,title,description,mitigation,notes
"2018-01-25 17:13:18","Verify that all application components are identified and are known to be needed","When you are building an application you first want to map where you are placing source files\, libraries and executables. With these components identified and mapped\, it becomes transparent where possible pitfalls might be in your application and increases the maintainability of the system. Also\, you have an indicator where possible reinforcements have to be implemented to avoid attacks.(i.e places where your application contains executable's)","Verify that all application components (either individual or groups of source files\, libraries\, and/or executables) that are present in the application are identified. When you identified these components you may want to map and document them in order to have a quick reference to this infrastructure when needed.","Some test pages were found."
"2018-01-29 19:50:04","Ensure forward secrecy ciphers are in use to mitigate passive attackers recording traffic.","In cryptography\, forward secrecy (FS; also known as perfect forward secrecy\, or PFS) is a property of keyagreement protocols ensuring that a session key derived from a set of longterm keys cannot be compromised if one of the longterm keys is compromised in the future. The key used to protect the transmission of data must not be used to derive any additional keys\, and if the key used to protect the transmission of data is derived from some other keying material\, then that material must not be used to derive any more keys. In this way\, compromise of a single key permits access only to data protected by that single key.","In the beginning SSL handshake\, the client sends a list of supported cipher suites (among other things). The server then picks one of the cipher suites\, based on a ranking\, and tells the client which one they will be using. This step is the one that determines whether or not the future connection will have perfect forward secrecy. Note that\, at this point\, certificates have not entered the picture at all. This is because whether or not a connection has perfect forward secrecy is determined by how the session key is derived. And how the session key is derived is determined by the cipher suite in use. So\, the cipher suites that use ephemeral DiffieHellman (DHE) or the elliptic curve variant (ECDHE) will have perfect forward secrecy while the other options will not.","There are some poor ciphers configured\, and no use of Forward Secrecy."

devGregA:
@martinmarsicano let me took art how the other importers do it. Will try to help with your date issue. Are you sure a data is actually being passed?

martinmarsicano:
Yes, is the first column of CSV and have printed it while testing.

devGregA:
Going to go ahead and merge. I'll add it to my backlog

PR #479, opened 2018-02-22
Predev

dkdewijer:
We've changed findings/views.py, dojo/models.py and dojo/templates/open_findings.html. Our goal in these changes was to add a "reviewer" field to the Finding model, DB and the finding creation screen. In the open findings screen, we've filtered those findings according to the user that's logged in. This way, two findings tables will be displayed, one with the findings that have been assigned to the user itself, and one table with the "general" findings that haven't directly been assigned to the user. So basically, we've created a functionality to assign a finding to a specific user, and filter those findings into two seperate tables.


devGregA:
@dkdewijer  could you close and resubmit the PR? It might have been an issue with Travis.

devGregA:
@dkdewijer please see my previous comment.

PR #481, opened 2018-02-27
Refactor/cleanup

maennel:
Hey there, 

I took up the code of DefectDojo and found it hard to get started. For example, I had trouble to get tests running locally. 

This PR addresses the entry hurdle by trying to collect unit-tests in a central location. It thus makes the code also a bit more "Django-compliant" if there's anything like this. Along with centralizing the unit-tests, I was able to identify at least one small bug. ;) 

I've also tried to centralize the logging config to a single settings file, instead of having it declared in every views.py file. The changed logging config should behave the same as before. 

I've also switched the "MySQL client" dependency to a more maintained library ("MySQL-python" to "mysqlclient"). I've made good experiences with this module in other projects, including migrations where the whole codebase was ported to Django2, while sticking to this client library.

In general, this PR does not change any function, but tries to improve the quality of the codebase. 

Thanks and best, 
-Manuel J.


devGregA:
@maennel Thank you for your PR! Will review shortly.

devGregA:
@cneill @aaronweaver would you have time to provide your thoughts?

mtesauro:
I gave this a very quick once-over when it was submitted but I need to actually **review** the changes before I'm OK with this non-trivial PR.

From the quick scroll through, it looks like stellar work - so props to @maennel for this PR.

I'm about to eat with my family but will give this a solid look over after our meal.

mtesauro:
OK.  Had a chance to really go through this PR, here's my thoughts:

- Loads of these changes are to move settings.py from its current location to a settings sub-directory.  I don't have strong feelings either way (move or keep where it is) so I'll defer to other's opinions
- The pep8 work is great and I'm glad to see it.  Not sure if this gets up to full pep8 compliance but great if it does
- The changes to the ansible stuff looks fine to me
- The changes to setup.bash are OK with one exception:

I wasn't sure at first why the boolean logic in [this line](https://github.com/DefectDojo/django-DefectDojo/blob/master/setup.bash#L197):
`if [ "$AUTO_DOCKER" != "yes" ]; then`

was changed to:
`if [ "$AUTO_DOCKER" == "yes" ]; then`
in the PR [here](https://github.com/DefectDojo/django-DefectDojo/pull/481/files#diff-963d3d20cd955e5295c643dc60a01d72R197)

The point of that != was to have the if clause run when the setup is interactive and the else clause to run when run non-interactively (like for creating the docker image).  I see now how @maennel  basically pulled the common code out of that If statement to after the if.  

So, it's a trade off of consistency of the interactive/non-interactive IF's for mildly simpler code that's one line shorter.  Fair enough - and its a matter of opinion that's not worth arguing about in my mind, especially since I violated that consistency in [line 288](https://github.com/DefectDojo/django-DefectDojo/blob/master/setup.bash#L288) already. ;-)

For my 2 cents, I'd accept this PR.

maennel:
Hi everyone,

Thanks for taking the time and reviewing this PR so quickly.

First of all, I realize moving the settings file is a major change and I
really don't want to break anyone's idea on how we want to use it in this
project. My main motivation to put it into a dedicated settings folder was,
to easily inherit from it (as done for the unittest settings). Let me know
if I should adapt this.

Regarding the PEP-8, I used the IDE's (PyCharm) shortcut. It usually does a
pretty decent job with some exceptions that need to be resolved manually -
which I didn't do...

Thanks for DefectDojo, a great piece of work!
-Manuel

PS: I joined the slack channel, so we can chat about the details. I
noticed, however, that different timezones may make this harder than it
should be :)

devGregA:
@maennel @cneill will merge tomorrow at close of business, unless there is an abjection.

devGregA:
Merged. @maennel this is a great PR. If you would like a dojo shirt as a thank you. Please email me: greg.anderson@owasp.org

PR #484, opened 2018-03-04
Let tests fail

maennel:
Add an exit statement triggered in case of failing tests, also letting travis know tests have failed.
Otherwise, tests may seem "green", even if problems have occurred during test execution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,12 +31,12 @@ script:
     case "$TEST" in
       unit-test)
         travis_fold start "unit-test"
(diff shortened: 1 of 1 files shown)


maennel:
@devGregA I noticed merging this PR made tests fail for real - that wasn't actually the intention :)
However, I noticed simply running the tests again brings them back to green: https://travis-ci.org/maennel/django-DefectDojo/jobs/349522945
This may be due to a timing issue in the integration-tests script, or something like that...?

PR #486, opened 2018-03-05
More refactor

maennel:
Hi everyone, 

first: this PR is intended more to get the discussion going rather than being a complete and finished change (even though I tried to keep things green).

The discussion's topic should be: How many deployment methods should be shipped with the actual DefectDojo repository, knowing that all of them need to be maintained.

Currently, there are three different ansible folders. Only one of them (legacy) seems referenced from within the Vagrantfile (but the path has changed and Vagrantfile wasn't updated). 
Also, there are several different scripts triggering Docker image builds (I think two or three of them). Some work, some don't.

The currently documented way of running the "all-in-one" container doesn't work, since the script has moved to a subfolder (from /django-DefectDojo to /opt/django-DefectDojo). A bit worse: Some of the scripts have the old path hardcoded, which makes the scripts fail out-of-the-box. Also, having to adjust paths in scripts at many places makes it hard to maintain. 

I've addressed some of these points with this PR (but not all of them - I'm not fit enough with Ansible, for example). So, my suggestion would be:
Let's remove unused deployment methods and move them to dedicated repos (including this repo as a submodule). Instead, let's provide a clean sample Dockerfile, which we can use at the same time for integration tests. 

So, here's what I did in this PR:

- I deleted seemingly unreferenced files in a quite aggressive manner
- I adapted the getting-started doc to point the docker-startup script's real path
- I moved most scripts to a folder ``entrypoint_scripts``. That folder has 4 subfolders: 

   - ``run``: to keep scripts that run things (running servers etc.)
   - ``test``: scripts to test something (i.e. for travis)
   - ``misc``: scripts that didn't fit elsewhere
   - ``common``: source-able scripts defining functions and variables which can (should?) be reused in other scripts

- I didn't modify a lot in the docker directory (except for the said file removal), since I was not sure about the logic that was followed when these scripts were contributed.
- I did change the Dockerfile towards running as few commands as root user as possible. Instead, I've added a (improveable) sudoers file for the dojo user. 

The approach I followed was to move as much bash-logic as possible to ``shared-functions`` and to eliminate duplicates there. This probably impacts ``setup.bash`` most.
With this, I think that we should keep DefectDojo installable as non-root user with some (listable) sudo privileges.

The refactoring process is not quite finished yet, but I find it hard and time consuming to go on without any feedback at this point.

Looking forward to hear from you - feedback welcome. 

PS: I realize this text and PR mix together a lot of stuff, so thanks already for your patience.


devGregA:
@maennel very interesting. Will review and discuss. Not too long ago we were discussing how to consolidate the install dependencies as much as possible (ansible, docker, bash, etc), but ran into architectural issues with ansible that ended up being left as a TODO. Will be back shortly with my thoughts

devGregA:
@maennel planning to meet Monday with @mtesauro and @aaronweaver to really review

maennel:
@devGregA Great to see you guys taking this so seriously. I, and I believe many others, appreciate this a lot. Thank you!

My 2 cents:
I was thinking about the matter again during the last couple of days. 
Maybe it would be an approach to keep this repo as clean as possible and focus on creating a clean python wheel, which can be uploaded to PyPi. This wheel would clearly state OS requirements that need to be present when installing and running DefectDojo. Initial configuration would take place through a config file (for example in json, xml or ini-format) at some predefined location.

Additional repositories could provide sample deployments, for example all-in-one Dockerfiles, a distributed Docker setup, a setup with Vagrant, etc. These deployments would have a common point in installing DefectDojo via pip.

Best, 
-Manuel 

PS: I think I was a bit quick in adding another commit to this branch - I'll take care of bringing this back to green ;)

aaronweaver:
@maennel If you have time next week maybe we can meet together with @devGregA and @mtesauro on a hangout and figure out a strategy.

devGregA:
Will clone & test

t3h2mas:
Is this going to be merged as is, or in feature specific chunks?

maennel:
@t3h2mas I didn't plan to split it up for now - so rather the first one, merge as is.

devGregA:
@maennel something introduced is breaking sed usage of setup.bash

sed: 1: "dojo/settings/settings.py": extra characters at the end of d command

t3h2mas:
Sounds good. Looks like I'll be waiting until after the merge to do any local or PR development. That way a big merge conflict is avoided.

maennel:
@devGregA Can you please indicate the version of sed and what system you're using it on? Thanks

devGregA:
@maennel I believe it is 10.6, but it is BSD sed which OSX uses. Currently the if clause here: https://github.com/DefectDojo/django-DefectDojo/blob/master/setup.bash#L204 handles the sed check.

maennel:
@devGregA Thanks for the info. 
Indeed, going into the man pages for [GNU sed](https://linux.die.net/man/1/sed) and [BSD sed](https://www.unix.com/man-page/FreeBSD/1/sed/) shows that the argument to the ``-i`` flag is optional for GNU and mandatory for BSD. Will adapt this.

devGregA:
@maennel testing

devGregA:
Talked with Aaron, merging.

PR #499, opened 2018-03-12
Update watson dependency

maennel:
Updated the watson dependency and optimized imports in files where it is being used. 
At the same time, made big parts of models.py and search/views.py (which are the files using watson) mostly PEP-8 compliant. 

- [ ] If this is a new feature and not a bug fix, you've included the proper docuemntation under the /docs folder


devGregA:
Will clone and try

PR #500, opened 2018-03-12
Remove unused python dependencies

maennel:
Removed dependencies that look like they are not used (i.e. not referenced in templates and not imported in python files). 

Also tried to update/remove ansible references, which should work out somehow as well. ;)

- [ ] Your code is pep8 compliant (Dojo's code isn't currently pep8 compliant, but we're trying to correct that) --> Nothing changed with PEP-8 relevance
- [x] If this is a new feature and not a bug fix, you've included the proper docuemntation under the /docs folder --> Adapted docs not to include removed dependencies anymore;


devGregA:
Add back overextends to avoid breaking plugins

maennel:
Didn't mean to let these checks fail, but got interrupted by a call.. following up...

Issue #495, opened 2018-03-08
Track Django migrations in GitHub


maennel:
### Issue Type 
- [ ] Bug
- [x] Enhancement

### Description
In order not to break existing installations and thus making the process of upgrading DefectDojo easier, I suggest to start keeping track of Django migration files in the GitHub repo. 

The following scenario currently breaks the application and may keep people from upgrading DefectDojo:
- Someone operates a DefectDojo instance
- A new version of DefectDojo including a new killer feature but also with some model changes is released.
- Someone deploys the new version of DefectDojo and generates a new 0001_initial_migration.py representing the new state of the models. The newly generated initial migration will be ignored, since Django keeps track, which migrations have been applied already (in the DB). So, the app will throw errors, telling some fields could not be found. The application will not be usable, unless the DefectDojo Admin figures out, which version they used previously and generates a 0002 migration based on the model changes.

Instead of this scenario, I suggest to keep track of model changes via migration files, which are incremental and which should improve the upgrading experience for DefectDojo Admins. 

EDIT: Removed the actual mention of _at_Someone :D

maennel:
Looks like the described scenario is the cause for issues #468, #473, and probably more. So this change would cause fewer issues.

maennel:
I'm thinking about a strategy to add migrations and I can think of the two approaches:
- Simply add an initial mi

Issue #473, opened 2018-02-20
add_findings doesn't work for user with not touched profile


tulsluper:
Please use the following template when submitting issues:

### Issue Type 
- [x] Bug
- [ ] Enhancement

### Description

add_findings page returns error when  user with not touched profile (maybe not created automatically) tries to add a finding.
After go to User Profile and save empty form - adding findings works without errors.

### Operating System 
RHEL 7.3

### Install

- [x] setup.bash
- [ ] docker

### Steps to Reproduce
create new user in admin with staff permissions
log in as new user
add new findings - <defectdojo_url>/test/<test_id>/add_findings

Error:

```
   RelatedObjectDoesNotExist at /test/45/add_findings
   User has no usercontactinfo.

    /opt/dojo/defect-dojo/dojo/test/views.py in add_findings
    236.    new_finding.save() ...

    /opt/dojo/defect-dojo/dojo/models.py in save
    781.    if self.reporter.usercontactinfo.block_execution:
```

devGregA:
@tulsluper I haven't been able to reproduce this. I've tried creating users via admin console, and the dojo create user function. I've also tried with an instance being migrated vs not being migrated. I'm not sure what else to try. The error you received definitely seems to lineup.

Issue #468, opened 2018-02-09
defectDojo failed after using latest docker image


Mason-Fan:
Please use the following template when submitting issues:

### Issue Type 
- [x] Bug
- [ ] Enhancement

### Description
use the defectdojo docker image from appsecpipeline/django-defectdojo docker hub for about 1month. yesterday update the image with latest and start a new container with the older mysql data volume mounted with command below
 docker run --name defectDojo -v defectDojo-mysql:/var/lib/mysql -it -p 8000:8000 appsecpipeline/django-defectdojo bash -c " bash /django-DefectDojo/docker/docker-startup.bash" 

get the error below after docker starts

Request Method: | GET
-- | --
OperationalError
(1054, "Unknown column 'dojo_finding.line' in 'field list'")
/usr/local/lib/python2.7/dist-packages/django/db/models/sql/compiler.py in execute_sql, line 886
/usr/bin/python
2.7.12
['/opt/django-DefectDojo',  '/usr/lib/python2.7',  '/usr/lib/python2.7/plat-x86_64-linux-gnu',  '/usr/lib/python2.7/lib-tk',  '/usr/lib/python2.7/lib-old',  '/usr/lib/python2.7/lib-dynload',  '/usr/local/lib/python2.7/dist-packages',  '/usr/lib/python2.7/dist-packages',  '/opt/django-DefectDojo']
Fri, 9 Feb 2018 04:46:28 +0000

so hope we can have detailed document about upgrade or migration

### Operating System 
Ubuntu 16.04

### Install

- [ ] setup.bash
- [x] docker

### Screenshots (optional)
![image](https://user-images.githubusercontent.com/8722967/36012297-2e91c842-0d98-11e8-97e1-acabc91cbf73.png)

devGregA:
@Mason-Fan on the docker system if you perform the following from the DD directory it should fix the issue: `python manage.py makemigrations` `python manage.py migrate` it should update the DB.

PR #511, opened 2018-03-16
Add migrations to version control

maennel:
As described in issue #495, Django migrations are now tracked under version control. 
An official documentation of migrations can be found at https://docs.djangoproject.com/en/dev/topics/migrations/#version-control

In my opinion, tracking migrations can prevent issues to happen in future releases and ensure data consistency across new releases. 

This PR includes:
- an initial migration
- a technical way to ensure contributors have added migrations (tests will fail otherwise)
- an upgrading "How-To" guide
- several small bugfixes (mainly wrong settings imports discovered during testing of migrations)
- renamed settings.dist.py to settings_dist.py to make it importable in unittest-settings (adapted all retrieveable occurrences of it, as well)

---

- [ ] Your code is pep8 compliant (Dojo's code isn't currently pep8 compliant, but we're trying to correct that)
- [x] If this is a new feature and not a bug fix, you've included the proper documentation under the /docs folder


devGregA:
@maennel @aaronweaver did some more testing. One more thought. What if you're developing a feature that requires a migration, but ends up not being permanent i.e. I think I need a manytomany when it turns out I just need a foreign key. That extra DB change will be erroneously tracked, would it not?

maennel:
@devGregA Yes, if the migration file makes it to a release, it will be applied on anyone's DB installing the "new" release. There's the possibility to revert (or change) this in a controlled manner with another, subsequent migration, transforming the M2M relationship to a ForeignKey in the example you've given. 

If there weren't any migrations, users would be faced with two conflicting models (assuming they don't keep track of their own migrations locally - which may be the case in a PaaS deployment or a distributed containerized environment). And they would have to figure out a resolution to the problem themselves - or file an issue here. 
As noted in #495, I believe issues #468 and #473 may be such cases.

devGregA:
Hi @maennel still discussing. I definitely see where this could solve some issues, but I think it will very likely cause some issues as well. The Django community seems to be out on whether or not migrations should be tracked. When an application is internal and moves through dev -> staging -> production the case for including migrations is compelling. However, with plugins, extensions, modifications, and changes made to models in dev, We're still not sure. I will try to get @grendel513 's opinion as I think he is stronger in Django core knowledge than any of the current maintainers. However, he is also one of the most busy. Will get back to you as soon as I can.

maennel:
Thanks @devGregA, I appreciate.

imohammedzen:
I second @maennel's opinion, benefits outweigh the risks for migrations. The most important for my perspective is ability to test the change. 

For example, with API v2 I didn't have way to test if it will break something, if I had access to migrations, I would be able to see the breaking changes.

devGregA:
All, closing for now. Regarding API V2, I believe the issue was due to poor travis tests. Having Migrations would not have affected the testing outcome as migrations only manage schema, not the actual data. A pre-populated previous version of Dojo in CI would have caught the issue.

PR #535, opened 2018-04-10
External db build args

t3h2mas:
Small `Dockerfile` change to allow devs to set env variables for an external db in the `docker build` command instead of modifying the `Dockerfile` using build args.

This makes it easier to use an external database, while keeping repos in sync with the upstream repository by avoiding merges.

The current build doesn't change because defaults are set (blank strings) so no errors get thrown if someone does not choose to use a build arg.

[dockerfile ARG reference](https://docs.docker.com/engine/reference/builder/#arg)
[docker build command build arg reference](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,23 @@ MAINTAINER Matt Tesauro <matt.tesauro@owasp.org>
 
 # # # Create a single Docker image running DefectDojo and all dependencies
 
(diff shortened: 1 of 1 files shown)


devGregA:
@aaronweaver please let me know if you have any concerns.

PR #554, opened 2018-04-25
Add API v2 using Django Rest Framework

secfigo:
This PR adds v2 of API using Django Rest Framework. 

1. The new code comes with 100% code coverage 
2. The new code is PEP8 compliant
3. Adds both Swagger UI and DRF browsable APIs.
4. Token Generation is also available via API.

To test it locally please follow the below steps

1. Setup Vagrant setup locally using the following Script from django-DefectDojo(obviously replace the existing Vagrantfile)

```bash
# -*- mode: ruby -*-
# vi: set ft=ruby :

# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
VAGRANTFILE_API_VERSION = "2"

Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
    config.vm.box = "bento/ubuntu-16.04"
    config.vm.hostname = "defect-dojo"

    config.vm.network :private_network, ip: "192.168.13.37"

    # Exposes port 8000 on host from guest.
    config.vm.network "forwarded_port", guest: 8000, host: 8000

    # Shared folder for local code base
    config.vm.synced_folder ".", "/opt/defect-dojo/dojo/app/"

    config.vm.provider "virtualbox" do |v|
      v.memory = 1024
      v.name = "Defect-Dojo"
      v.cpus = 1
    end
end
```

2. login into vagrant box and run setup.bash script
```bash
vagrant ssh
cd /opt/defect-dojo/dojo/app/
./setup.bash  # select mysql, set root password and type it again when asked about MySQL details.
python manage.py runserver 0.0.0.0:8000
```
3.  Visit http://localhost:8000/ on your host machine, login and browse to v2 of api below
[http://localhost:8000/api/v2/](http://localhost:8000/api/v2/)


aaronweaver:
@secfigo This looks great. I'll try this out and let you know if I have any issues.

dojo/models.py
```diff
@@ -457,7 +461,7 @@ class IPScan(models.Model):
 
 
 class Engagement_Type(models.Model):
-    name = models.CharField(max_length=200)
+    name = models.CharField(max_length=200, primary_key=True)
```
aaronweaver: How would you resolve an issue if there is a duplicate engagement type?

aaronweaver:
@secfigo Could you merge to the dev branch instead of master?

maennel:
I think this is great work, @secfigo ! A big thumbs up.
One question: is there a specific reason the code in dojo/unittests/test_rest_framework.py is replicated in tests/rest_framework_test.py? From what I could see, the two files differ only in one line. 
If not needed, I would suggest to remove the file at tests/rest_framework_test.py in favour of the other, automatically executed one, in order to have the same code not appearing twice in the same repo.

dojo/models.py
imohammedzen: I'm not aware if there will be such case. Can you please explain why/when that would happen ?

secfigo: I'm not aware, if there will be such a case. Can you please explain why/when that would happen ?

devGregA: Why change the primary key?

devGregA:
@secfigo really like this, but I don't know how to auth for testing 😂 I think we could really use some docs.

dojo/models.py
imohammedzen: @devGregA there is only one field in this model but will make it unique. DRF will throw an error if there is a duplicate engagement type but the bigger question is,  would there be any reason for engagement name to be duplicate  ?

secfigo:
@devGregA Added instructions in the PR and also API v2 key under user profile.

devGregA:
Awesome @secfigo! I'm good with merging. @aaronweaver good with you? @secfigo is there anything else you wanted to add before we merge? Or should I go ahead?

devGregA:
Spoke to @aaronweaver merging.

PR #568, opened 2018-05-08
replace risk_factor tag with severity attribute in nessus xml parser

no-sec-marko:
The nessus xml parser supposed the xml tag _risk_factor_ as severity classification. However, each ReportItem in the nessus xml has an attribute _severity_ which represents the "correct" tenable severity rating.


secfigo:
@no-sec-marko Travis Job failed because you have a blank line at the end of file and flake8 is complaining about it :)

PR #594, opened 2018-05-21
Add finding hotkeys

no-sec-marko:
This PR adds an edit and save hotkey to the process of finding creation.
If you view a finding on the _view_finding.html_, you can press "**e**" to edit the finding. If you are in promote, edit, add from template or add a add new finding editor, you can press "**ctrl+s**" to save the finding. This behaviour helps in the work with findings or in the review.


aaronweaver:
@no-sec-marko That's a nice little feature to have! @devGregA ok to merge?

PR #600, opened 2018-05-24
Centospostgresfix2

eik18:
Please submit your pull requests to the 'dev' branch.

When submitting a pull request, please make sure you have completed the following checklist:

- [ ] Your code is pep8 compliant (Dojo's code isn't currently pep8 compliant, but we're trying to correct that)
- [ ] If this is a new feature and not a bug fix, you've included the proper documentation under the /docs folder

PR #615, opened 2018-06-11
Revert "Centospostgresfix2"

devGregA:
Reverts DefectDojo/django-DefectDojo#600


devGregA:
@eik18 Reverting the reversion. We think it's an issue with Travis or the actual test.

PR #633, opened 2018-06-21
If the server does not have FQDN, create the endpoint with the IP

adanalvarez:
Hello,

I tried to upload a Qualys scan and I got an error. The error was produced because the parser looks for the DNS tag and this tag is not always in the XML.

With this change if DNS is not in the XML, the endpoint is going to be created using the IP.

Also, I changed the way of creating the endpoint when the FQDN is available because currently it creates only one endpoint without name.

Regards,


aaronweaver:
@adanalvarez Thanks for the contribution, much appreciated.

PR #667, opened 2018-07-27
Goast Scanner Was Renamed to Gosec Scanner

raghunath24:
Goast scanner was renamed by the authors of tool. Just updating the names and folders to avoid confusion.


aaronweaver:
Thanks for updating it so quickly.

PR #671, opened 2018-08-01
fix nessus xml parser if item is None or empty

no-sec-marko:
If an item of the nessus xml file is empty (e.g. `<plugin_output></plugin_output>`), the parser will crash when try to access the text of the element.


aaronweaver:
Thanks for the PR @no-sec-marko

PR #679, opened 2018-08-07
Added support for JSON import of NPM Audit scans

benno001:
The Node Package Manager has an 'Audit' package which checks npm dependencies.

Commits add support for the '--json' output of 'npm audit'.


benno001:
@aaronweaver Done - also added a little documentation

PR #681, opened 2018-08-07
Add unique constraints for some model fields

Nilix007:
Add unique constraints on the following fields:
- `Product.name`
- `Product_Type.name`
- `Tool_Type.name`
- `JIRA_Issue.jira_id`
- `Engagement.(name,target_start,product)`
- `Endpoint.(protocol,host,path,query,fragment,product)`

Now, all these objects can be uniquely identified by their names which is handy for synchronizing DefectDojo's state with external sources.

Some parts of the code already relied on the unique constraints, for example by using `Model.objects.get(name=<...>)`. (There are probably even more cases)

In turn, this requires `max_length <= 1024` for the corresponding char fields
and `innodb_large_prefix` must be ON which is the default for MySQL >= 5.7.7.

See also: https://stackoverflow.com/questions/45233362/django-says-that-mysql-does-not-allow-unique-charfields-to-have-a-max-length-2

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -276,7 +276,7 @@ class Contact(models.Model):
 
 
 class Product_Type(models.Model):
-    name = models.CharField(max_length=300)
+    name = models.CharField(max_length=300, unique=True)
     critical_product = models.BooleanField(default=False)
     key_product = models.BooleanField(default=False)
     updated = models.DateTimeField(auto_now=True, null=True)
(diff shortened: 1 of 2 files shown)


aaronweaver:
The views don't gracefully handle a duplicate entry so it would be better to add the check and display a friendly message. Secondly, since some users may have duplicate issues this should be noted in the upgrade notes.

Nilix007:
Looks good for me without any additional modifications:
screenshot_2018-08-08 new product defectdojo
screenshot_2018-08-08 edit product defectdojo

aaronweaver:
Perfect, thanks for checking!

PR #685, opened 2018-08-07
Make `Finding.age` a property and use it instead of custom SQL query

Nilix007:
With this patch, I can use SQLite as the database in my dev environment :)


aaronweaver:
@Nilix007 Let me check with @devGregA to see if there was some legacy reason it was done this way.

devGregA:
I think there was a reason why we did it the way we did it, but I'm not able to ascertain it now.

PR #694, opened 2018-08-08
APIv2:  Use IDs instead of URLs as identifier for objects

Nilix007:
This PR modifies the APIv2 so that the IDs of an object are used as identifier instead of the object's URL.

Additionally, this PR removes a lot of serializer field definitions as the generated default fields can be used now.


aaronweaver:
@Nilix007 would you mind resolving the merge conflict?

PR #699, opened 2018-08-09
Add support for JSON output of Clair scans

benno001:
Clair scans Docker images and can output findings in JSON. Added support for importing said findings.

### DIFF
--- a/dojo/fixtures/test_type.json
+++ b/dojo/fixtures/test_type.json
@@ -215,5 +215,12 @@
     },
     "model": "dojo.test_type",
     "pk": 31
+  },
+  {
+    "fields": {
+      "name": "Clair Scan"
+    },
+    "model": "dojo.test_type",
+    "pk": 32
   }
 ]
--- a/dojo/fixtures/test_type.json
+++ b/dojo/fixtures/test_type.json
@@ -146,6 +146,7 @@
     "model": "dojo.test_type",
     "pk": 20
   },
+
   {
     "fields": {
       "name": "Qualys Scan"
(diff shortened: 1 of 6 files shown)


aaronweaver:
@benno001 thanks for adding the Clair importer, that's a fantastic addition.

wurstbrot:
Hi @benno001,

which clair client are you using producing that format?

Cheers,
Timo

benno001:
Hi @wurstbrot, 
I used arminc's clair-scanner (https://github.com/arminc/clair-scanner)

wurstbrot:
@benno001 as there are a lot of clair clients, shouldn't we have a prefix like clair- and the actual name of the client,  e.g 
* clair-scanner
* clair-klar

For clair-scanner it could also be clair-clair-scanner. I am thinking of adding a clair-klar support.

benno001:
@wurstbrot clair-clair-scanner is probably technically correct, but I'd prefer your suggestion clair-scanner for clearness. That, and I don't know other clients specifically called clair-scanner. What do you think?

wurstbrot:
@benno001 yeah, I think so too.
I am writing a clair-klar scanner right now. Why have you choosen that "Unknown" is "Info"? I would assume the worst -> "Ciritical".

benno001:
@wurstbrot  I haven't seen anything with severity 'Unknown' yet. Assuming the worst for an unknown severity is a good way to make yourself very unpopular in lots of environments, so I'd advise against that. 
On the other hand I agree info might be too lax, so maybe we should consider something more in the middle (medium/warning). Now that I'm reviewing my commit, I see we might also need to add translation from Defcon1 severity to critical (see also: https://github.com/coreos/clair/blob/e907e4d263a743f7d4923c177e5db37f656a53b4/database/severity.go#L68).

wurstbrot:
> I haven't seen anything with severity 'Unknown' yet.
As you say it never showed up before, it is not such a big pain.

yes, we have to have a mapping for defcon1.

benno001:
One critical with all the bells and whistles going off may be all it takes to lose credibility. Of course I don't know about your organizational context, we might need some more input on this one.

Created issue #861.

PR #700, opened 2018-08-09
Fixup for #681: Reduce `max_length` on some `Endpoint` fields

Nilix007:
MySQL requires for unique constraints that the combined size of all
fields used in the index is less than 3072 bytes (or 1024 chars).  This
PR sets the `max_length` on the `Endpoint` fields so that the total
length is less than 1024 chars.

This PR is a follow-up fix for #681.

Initially, I thought that it is sufficient when each `CharField` field is at most 1024 chars long. The original PR somehow worked for me, but the migrations failed on our preview instance as I tried to deploy a prerelease version of v.1.5.3.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -792,19 +792,19 @@ class Endpoint_Params(models.Model):
 class Endpoint(models.Model):
     protocol = models.CharField(null=True, blank=True, max_length=10,
                                 help_text="The communication protocol such as 'http', 'ftp', etc.")
-    host = models.CharField(null=True, blank=True, max_length=500,
+    host = models.CharField(null=True, blank=True, max_length=200,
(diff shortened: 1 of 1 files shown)


aaronweaver:
@Nilix007 as this is truncating fields there is the potential that we might cause data truncation issues for existing users. The field most likely that would be truncated is `Path` field. Are there any other possible alternatives? /cc @devGregA

devGregA:
I agree with @aaronweaver that there is risk with truncating the fields. I do not think there is a good alternative solution at this time.

aaronweaver:
For now I'm going to close this PR as I don't want to possibly truncate the endpoint. I'll have a following PR to remove the unique_together for the time being.

PR #713, opened 2018-08-20
Bug fix/yum assume yes

code4innerpeace:

- [ ] Your code is flake8 compliant (Dojo's code isn't currently flake8 compliant, but we're trying to correct that)

----
Currently, for 'yum' commands '-y'(assume_yes) is not enabled, due to which we need to explicitly enter 'Yes'. So add '-y' parameter to 'Yum' commands.


devGregA:
@aaronweaver I think we have had issues with folks wiping out conflicts with -y which is why it was initially removed. I'm not opposed to having it, but maybe we should have a warning? Having the user interact may not be the worst thing.

aaronweaver:
Ah, hmmm, so what @code4innerpeace was going for is a 100% automated install. Is it worth adding an option for interactive installations?

devGregA:
Yeah, we had a complaint when we wiped out someone's dependencies / db or something to that effect. Adding a interactive vs not option would probably be the best way to go

devGregA:
We could go ahead and keep as accepted and just add an option to setup.bash before the next release?

PR #722, opened 2018-08-24
Do not update `Finding.hash_code` on every save

### LINKED ISSUE #715: Deduplication does not work well with edited findings
### Issue Type 
- [x] Bug
- [ ] Enhancement

### Description

Currently, findings are deduplicated if they have the same hash code which depends on the title, among other values. [0] This hash code is updated each time the finding is saved to the database. [1]

If a finding's title is edited, the next re-upload of this findings will not be deduplicated anymore. (And the following re-uploads will have the previous re-upload as original)

I had a chat yesterday with @xthk, and we believe that the hash code should be immutable so that re-uploads of findings will be correctly deduplicated even if the original finding has been edited.

@aaronweaver: What are your thoughts about this?

[0] https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/models.py#L1044
[1] https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/models.py#L1185

Nilix007:
Fixes #715.

### DIFF
--- a/dojo/management/commands/dedupe.py
+++ b/dojo/management/commands/dedupe.py
@@ -21,7 +21,7 @@ class Command(BaseCommand):
         findings = Finding.objects.all()
         print "######## Updating Hashcodes ########"
         for finding in findings:
-            finding.hash_code = finding.get_hash_code()
+            finding.hash_code = finding.compute_hash_code()
             finding.save()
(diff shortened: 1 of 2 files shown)


aaronweaver:
Looks good @Nilix007, I've asked @devGregA to take a look as well.

PR #724, opened 2018-08-24
APIv2:  Skip duplicate findings on import under certain conditions

Nilix007:
On import, skip duplicate findings for the same engagement if the scan
is automated.  A scan is assumed to be automatic when the following
conditions are satisfied:
    - Deduplication is activated
    - The test tool is static
    - The engagement type is CI/CD

@aaronweaver what are your thoughts about this idea?

TODO:
- Maybe, we should also add an opt-in flag in the system settings for this feature?
- Needs to be rebased if #722 is merged.


aaronweaver:
@Nilix007 I like the idea, how about making it an option on the API instead of config? That would allow the flexibility to enable per engagement.

PR #748, opened 2018-09-24
APIv2: Scan import options

Nilix007:
Hey @aaronweaver,

I was still working on #724 :laughing: 

This is the second part which adds another option `close_old_findings` that will close old findings which were created by the same scanner tool and are not present anymore in this scan. (Very useful for scanner that produce static findings like Dependency Check)

I've also got some tests for these options.


Nilix007:
Currently, the HTML in the description of the notification get escaped.

wurstbrot:
@aaronweaver what is the status of this PR? Will it be merged?

aaronweaver:
@wurstbrot Fixed the conflict and merged.

PR #755, opened 2018-09-26
An automated closed finding should be marked as mitigated

wurstbrot:
A finding needs to be mitigated to not show up under Metrics.
See also https://github.com/DefectDojo/django-DefectDojo/pull/748


Nilix007:
Hey Timo,

good catch :smile:

devGregA:
@aaronweaver  good with merging?

PR #758, opened 2018-09-28
APIv2: Also skip duplicates of duplicates on import

Nilix007:
I forget the case that someone manually marks a finding as duplicate. In that case, any duplicates of this duplicates should be skipped as well.


devGregA:
@aaronweaver good with merging?

PR #827, opened 2018-12-07
Fix max_length on primary key Product.name

wurstbrot:
See https://docs.djangoproject.com/en/2.1/ref/databases/#notes-on-specific-fields
Fixed error (mysql):

==> logs/dojo.log <==
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/commands/runserver.py", line 124, in inner_run
    self.check(display_num_errors=True)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 405, in check
    raise SystemCheckError(msg)
django.core.management.base.SystemCheckError: SystemCheckError: System check identified some issues:

ERRORS:
dojo.Product_Type.name: (mysql.E001) MySQL does not allow unique CharFields to have a max_length > 255.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -279,7 +279,7 @@ class Contact(models.Model):
 
 
 class Product_Type(models.Model):
-    name = models.CharField(max_length=300, unique=True)
+    name = models.CharField(max_length=255, unique=True)
     critical_product = models.BooleanField(default=False)
     key_product = models.BooleanField(default=False)
     updated = models.DateTimeField(auto_now=True, null=True)


devGregA:
@aaronweaver I'm reading the django notes, and it makes sense. However I'm worried it will mess up people's instances with longer names. I'm also confused why we haven't seen this issue. Also ping @mtesauro

aaronweaver:
That's b/c SILENCED_SYSTEM_CHECKS = ['mysql.E001'] is set in settings.py and that warning doesn't show. 

You can set innodb_large_prefix to on and that's a default in Mysql 5.7.7+. I also would be concerned about truncation for current users, not sure if there is a good way to set maxchar based on the backend db.

Issue #813, opened 2018-11-21
Nessus imported scan findings are not pushed to Jira


alekbc:
Please use the following template when submitting issues: (If this is a security issue, then review the [disclosure policy](https://github.com/DefectDojo/django-DefectDojo/blob/master/SECURITY.md) and submit via our security reporting program.

### Issue Type 
- [X] Bug
- [ ] Enhancement

### Description
Nessus imported Findings are not pushed to Jira. Manual findings are pushed successfully.

Celery Error: "Exception: JIRA_Issue matching query does not exist."

### Operating System 
Please include the version number (ex: Ubuntu 16.04)
Debian buster/sid
### Install

- [X] setup.bash
- [ ] docker

### DefectDojo Version
1.5.2

### Steps to Reproduce
Configure Jira and import a nessus scan.

### Console Logs (optional)
Nov 21 14:56:27 XXXXXXX celeryworker[26065]: [2018-11-21 14:56:27,921: ERROR/ForkPoolWorker-1] Task update_issue_task[c76396b1-f7ea-4946-bb5d-2e7cb66e8476] raised unexpected: DoesNotExist('JIRA_Issue matching query does not exist.',)
Nov 21 14:56:27 XXXXXXX celeryworker[26065]: Traceback (most recent call last):
Nov 21 14:56:27 XXXXXXX celeryworker[26065]:   File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 382, in trace_task
Nov 21 14:56:27 XXXXXXX celeryworker[26065]:     R = retval = fun(*args, **kwargs)
Nov 21 14:56:27 XXXXXXX celeryworker[26065]:   File "/usr/local/lib/python2.7/dist-packages/celery/app/trace.py", line 641, in __protected_call__
Nov 21 14:56:27 XXXXXXX celeryworker[26065]:     return self.run(*args, **kwargs)
Nov 21 14:56:27 XXXXXXX celeryworker[26065]:   File "/opt/dojo/django-DefectDojo/dojo/tasks.py", line 235, in update_issue_task
Nov 21 14:56:27 XXXXXXX celeryworker[26065]:     update_issue(find, old_status, push_to_jira)
Nov 21 14:56:27 XXXXXXX celeryworker[26065]:   File "/opt/dojo/django-DefectDojo/dojo/utils.py", line 1066, in update_issue
[...]

PR #833, opened 2018-12-20
fix issue #813

alekbc:

- [x ] Your code is flake8 compliant (Dojo's code isn't currently flake8 compliant, but we're trying to correct that)
- [x ] If this is a new feature and not a bug fix, you've included the proper documentation under the /docs folder

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ that should give you an idea of how to use DefectDojo for your own team.
 
 ![Slack](https://raw.githubusercontent.com/DefectDojo/Documentation/master/doc/img/slack_rgb.png)
 
-Realtime discussion is done in the OWASP Slack Channel, #defectdojo. [Get Access.](https://owasp.herokuapp.com/)   
+Realtime discussion is done in the OWASP Slack Channel, #defectdojo. [Get Access.](http://owaspslack.com/)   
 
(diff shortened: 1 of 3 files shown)


aaronweaver:
@devGregA What do you think about this one as well?

devGregA:
I haven't had a chance to test. I'm a little worried that we don't have CI
tests around JIRA.

This PR makes a lot of sense to me. A couple things I'm confused on:

Overwriting status as is done in the PR doesn't make sense to me
(views.py). Status is a method that is constructed based on data about the
finding (models.py).

Callable checks if an object is null. I'm not sure what is being caught in
the else that isn't caught by the previous elif.

OH maybe it's being overwriting because otherwise we're losing the
transition from active to non-active. Would be great if we could just get
some insight into the thinking.

Greg Anderson
DefectDojo Project Leader
greg.anderson@owasp.org

alekbc:
Hi Greg,
Thanks for taking a look into this.

Considering this part of the code belongs to the “closing” status part, I’ve just added the status that is set after you “close” the finding. There’s probably a better way to do this. However, I’m not familiarized with the code so I wasn’t sure how to do it.


product_tab = Product_Tab(finding.test.engagement.product.id, title="Close", tab="findings")
           # HERE - lines below added to update the finding status in jira (when it's closed)
finding2 = finding
finding2.status = list(('Mitigated','Inactive','Verified'))
update_issue(finding2, old_status, True)
           # END


Regards,
Alejandro

PR #841, opened 2019-01-09
Dedupe functionaltiy doesn't correctly recognize all duplicates and l…

bend18:
…ooks at few original findings as duplicates too. This is to recreate the hash_code to look at the Description where the Issue ID is added to create a different dedupe hash so that only true duplicates are flagged by dedupe.

- [x] Your code is flake8 compliant (Dojo's code isn't currently flake8 compliant, but we're trying to correct that)

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1060,7 +1060,7 @@ class Finding(models.Model):
         ordering = ('numerical_severity', '-date', 'title')
 
     def compute_hash_code(self):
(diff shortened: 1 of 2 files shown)


devGregA:
@aaronweaver  do you think including issue id would be problematic? I don't use veracode so I don't have much stake in this game.

devGregA:
I'm good with this change.

PR #867, opened 2019-02-03
prevent sb-admin-2 from 4.0.0 upgrade

h00die:
Posted screenshots in slack of the issue.

After ~2.5yrs, startbootstrap-sb-admin-2 updated from 3.3.7+1 to 4.0.0 [reference](https://github.com/BlackrockDigital/startbootstrap-sb-admin-2/releases).  This happened January 29.

The changes are significant and break the UI (no longer a `/dist` folder, and major css changes that modify the left icon bar's width to about the whole page, etc).  This change locks `startbootstrap-sb-admin-2` to the known good version.

PR #869, opened 2019-02-04
prevent sb-admin-2 from 4.0.0 upgrade

h00die:
See #867


h00die:
too late for #swag reward ?

aaronweaver:
@h00die never to late for that. Hit me up on slack with your address and we'll get some swag sent over, Thanks for helping to make DefectDojo better.

PR #874, opened 2019-02-06
Fixed migration issue

Maffooch:
- Removed all but __init__.py from DefectDojo/dojo/db_migration
 - Fixed the header issue where DefectDojo header took up top half of screen
 - Fixed add finding issue where an assignment procedure was deprecated

All testing was done in docker so I am pretty sure this thing is actually python3 now
Please reach out if there are any further issues/questions/complaints.


Maffooch:
I didn't know this was here. I overwrote the default comment from above. I apologize.

PR #892, opened 2019-02-12
Fix bulk edit in findings and nessus report parsing

HumanoidPhantom:
- Fix bulk status change in findings(verified, false positive and out of scope): there was a problem with several same id attributes on the same page.  
- Fix Nessus report parsing: if Nessus plugin output contains HTML, the finding page markup is broken. Plugin output is now wrapped in a code block.


aaronweaver:
Thanks for the PR!

PR #927, opened 2019-03-11
Make DefectDojo work on Kubernetes

hendrikhalkow:
* Introduced new docker images that runs exactly one process per container, but multiple containers
* Introduced Helm chart to allow installing DefectDojo on Kubernetes
* Docker compose file updated to use these new images

- [x] Your code is flake8 compliant (Dojo's code isn't currently flake8 compliant, but we're trying to correct that)
- [x] If this is a new feature and not a bug fix, you've included the proper documentation under the /docs folder


wurstbrot:
There is a problem with the images on hub.docker.com:

$ docker-compose up -d
Pulling initializer (defectdojo/defectdojo-initializer:latest)...
ERROR: manifest for defectdojo/defectdojo-initializer:latest not found

hendrikhalkow:
> There is a problem with the images on hub.docker.com:
> 
> $ docker-compose up -d
> Pulling initializer (defectdojo/defectdojo-initializer:latest)...
> ERROR: manifest for defectdojo/defectdojo-initializer:latest not found

You need to check out the branch where this pull request comes from. I Merged the uwsgi, celery and initializer image into one image, which is defectdojo/defectdojo-django.

aaronweaver:
@hendrikhalkow Nice work. A few items:

1. For the docker-compose it's better to have defaults with option to override. These should be put back in to allow the user to override some of the settings. `${DD_DATABASE_PASSWORD:-dojodbpwd}`, `published: ${DD_PORT:-8080}` etc.
2. For travis the flake8 tests need to be added back in and if the unittests fail, how does that failure log so that the issue can be reviewed in travis? 
3. On the Kubernetes side, must the user create a TLS certificate store? I'd like to see if that step can be simplified or if there are other options available.

hendrikhalkow:
Hi @aaronweaver thanks.

1. Unfortunately, Docker Compose doesn't support support variables in the compose file. We can do that for the database password, etc, but not for the published port. If you want to do that, we need to generate the Docker Compose file from a template. This is a whole other story. We can put the supported environment variables into the compose file which will allow overwriting them with `docker-compose run -e ...`, but this won't work for the published port. If somebody wants flexibility like this, Kubernetes is the way to go.

2. I am not a Travis expert and I could use some help with that. Let's chat on Slack in #defectdojo about this.

3. The ingress is optional. If you use an ingress, you need a TLS certificate. if you deploy the helm chart with `--set django.ingress.enable=false` you will get the information how to access DefectDojo without ingress. See NOTES.txt in helm/defectdojo/templates folder.

aaronweaver:
3.7 supports variables as I've tested it. How are you generating the compose file?

ports:
      - target: 8080
        published: ${DD_PORT:-8080}
        protocol: tcp
        mode: host

hendrikhalkow:
Thanks for mentioning that, @aaronweaver . I wasn't aware of that yet. I just put default values into the compose file as you asked for and updated the pull request.

PR #935, opened 2019-03-15
Simplify All Calls to urlparse and ensure ZAP parser works properly to Import Results via API.

ninp0:
Tested in our lab successfully.


ninp0:
Updated Arachni parser as well.

ninp0:
Bumped Django module from minor rev .19 to .20 since .19 doesn't appear to exist (even in the Travis build).

PR #967, opened 2019-03-22
Fix 964, first issue

Ravenons:
This avoids computing what is needed for the deletion confirmation view (GET) for the deletion itself (POST) too.


Ravenons:
Sorry, I used master instead of dev when developing.

Let me try to create a new one...

PR #981, opened 2019-03-27
Engagement Planner, Dynamic tool scan, renaming of portscan and minor fixes

JamesCullum:

- [ X] Your code is flake8 compliant (DefectDojo's code isn't currently flake8 compliant, but we're trying to correct that.): Coverage has been extended beyond own new code (since then, GitHub apparently detects it as 400+ commits behind)
- [x] If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR: Provided in https://github.com/DefectDojo/Documentation/pull/29
- [X] Add applicable tests to the unit tests: Added basic tests for tool UI and both planner & tool scan


JamesCullum:
This development has been sponsored by "Panasonic Information Systems Europe GmbH" to match it to internal business needs. According to sponsoring guidelines, this matches 1000 lines needed for non-financial sponsorship, which we would like to use.

**Engagement Planner**
- Allows to simply plan engagements for a set time
- Respects capacity, breaks and local changes
- Allows to automatically run a tool at the beginning of an engagement

**Tool Run**
- It is needed to have an easy way to run tools (e.g. vulnerability scanners) automatically from within DefectDojo, both internally and externally. The portscan was a good attempt, but is not dynamic enough.
- As the data for authentication was already included, so now it can use SSH and HTTP connections to run tools externally, optionally importing the result. It is also possible to run software locally without authorization, but this feature is disabled in the settings.py by default.
- Tools are configured per product and progress can be monitored from there (or re-run manually). This allows to locally overwrite parts as needed.

**Breaking changes**
- All portscan related classes and activities have been renamed to reduce confusion. The cronjob now needs to run `run_portscan` instead of `run_scan` to clarify the purpose.
- Tools can only run if at least one endpoint is tagged as `tool-export`. For each endpoint marked as such, a thread runs with this endpoint as parameter. It is supposed to be the entry point for e.g. a vulnerability scanner.

aaronweaver:
@JamesCullum would you mind cleaning up the merge conflicts. Thanks for this contribution, I think it's a great addition to DefectDojo!

JamesCullum:
@aaronweaver I'm not too sure what some of those changes were about, as there appears to be some historical changes (e.g. the deduplication field on engagements), I tried to resolve the conflicts to my best knowledge, so it would be appreciated if you could review it again.

Otherwise the conflicts should be gone now.

aaronweaver:
@JamesCullum What does your defectdojo.sh script look like that is referenced in the docs?

JamesCullum:
@aaronweaver It's just a simple "Hello World" shell script. It is used in the unit test to execute a local tool scan with that path and checking if the correct text is returned. I can imagine that the pathing has issues, even though I tried to keep it as dynamic as possible. Is that causing issues? Where can I see the log?

aaronweaver:
@JamesCullum I think an example like arachni would be great to have and also in the documentation it isn't clearly specified on how the parameters are passed to the script.

JamesCullum:
@aaronweaver Where and how would I provide this example for arachni? It is simply passed as parameter to the file, so if you specify ssh://example.com/file.sh, it will connect via SSH to example.com and run file.sh with the first parameter being the exported endpoint, eg `./file.sh http://example.com`.

Should I clarify this more in the documentation?

aaronweaver:
@JamesCullum I'll be submitting a few changes shortly, if the following could be modified then it can be merged.

**Engagement Planner**
1. Looks like there are a couple validation issues. I was able to trigger an error when I didn't select any products.
2. Date dropdown, switch to the date picker that is used in creating engagements (for example) rather than the dropdown.
3. I think some sort of indication should be given back to the user when engagements are planned. For example, how about a count: 15 Engagements were planned with the prefix of "test-" or something along those lines. Right now, when you first use it, the first question I had was, what exactly did it do?

**Tools**
1. Instead of using a tag on the endpoints, create a boolean value to indicate what the scanner should use as the target URL/IP.
2. Need an example and documentation on what is required for adding tools. (Passes in IP/URL and expects back the text report output of the scanner.) Create a folder in the root (for now at least) scanner-automation/arachni.sh.

JamesCullum:
@aaronweaver Thanks for the details!

**Engagement Planner**

1. It is intended that if you don't select any product you get an error, hinting you to select one. I assume you mean a different one? Can you show me how to reconstruct it?
2. Searched for one but only found one with month/year. Will take a look.
3. Sure, can be added.

**Tools**

1. It is unclear to me what you mean - can you clarify your suggestion? The idea of using a tag is that it conveniently allows to select as many endpoints as the user likes. One thread is ran per marked endpoint, which a boolean wouldn't be able to do (if I understand you right).
2. Sure, I can add the example. Where would I put documentation, though? Is it sufficient to add some comments into the shell script?

aaronweaver:
@JamesCullum For engagement planner I had a django error, at the moment I don't remember exactly what I clicked to trigger it but I had a stack trace. Will play around more with it.

For Tools on endpoint, For each endpoint let's add a boolean it would function the same as a tag but I'd like to keep the tag functionality as just a tag. (overloading the model). Example is fine, maybe link over to it from the docs?

aaronweaver:
@JamesCullum would you mind submitting this to a new branch, maybe something like sec-automation? That will make it a bit easier for us to work on and then merge. Thx!

JamesCullum:
@aaronweaver So you mean all changes from now into a different branch?
Will do that, sure!

PR #1000, opened 2019-04-03
fix deduplication_on_engagement regression caused by optimistic linting

### LINKED ISSUE #983: Deduplication doesn't work anymore when using deduplication_on_engagement: True
### Issue Type 
- [X] Bug

### Description
Findings from OWASP Dependency Checker XML are not deduplicated

### Operating System 
Debian 9, MySQL.

### Install

- [X] setup.bash / legacy-setup.bash

### DefectDojo Version
1.5.4 from develop, so 46692a0f7bf568f5739d093ab0e4d2ef6733faf6

### Steps to Reproduce
Make sure celery is running
Create CI/CD engagement with deduplication on engagement.

Generate an XML report with OWASP Dependency Checker and make sure it has some findings.

Upload it via API or browser.
Upload it via API or browser.
Wait couple of minutes.
Observe DD has recorded all findings twice and does not deduplicate.

Start deupe manually:

```python manage.py dedupe```

Observe DD has still recorded all findings twice and none of them are deduplicated.

It's been behaving like this this for 24-48 hours so it's not a timing or async issue, I think.

When uploading the same report twice as ad hoc import (so no deduplication on engagement) it works fine and DD see the issues as duplicate.

Also when disabling "deduplication_on_engagement" for the CI/CD engagment resolves the issue.

I disable that however as we're building a couple of branches within this product each in their own engagement and we don't want to deduplicate across branches/engagements

### Sample Scan File (if applicable)
Test file from DD unit test.
[owasp-dependency-check-dummy.xml.txt](https://github.com/DefectDojo/django-DefectDojo/files/3013499/owasp-dependency-check-dummy.xml.txt)

valentijnscholten:
Regression was caused by https://github.com/DefectDojo/django-DefectDojo/commit/e4cbdfab1a3d034777a38396d663efb2cee1c964#diff-63c97e945dc0af3507d89798a8841819

I also fixed a typo and renamed a method to make its purpose more clear.

This fixes #983

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -557,9 +557,13 @@ class ImportScanSerializer(TaggitSerializer, serializers.Serializer):
             for item in parser.items:
                 if skip_duplicates:
                     hash_code = item.compute_hash_code()
-                    if Finding.objects.filter(Q(active=True) | Q(false_p=True) | Q(duplicate=True),
+                    if ((test.engagement.deduplication_on_engagement and
+                        Finding.objects.filter(Q(active=True) | Q(false_p=True) | Q(duplicate=True),
+                                              test__engagement=test.engagement,
+                                              hash_code=hash_code).exists())
+                    or (Finding.objects.filter(Q(active=True) | Q(false_p=True) | Q(duplicate=True),
                                               test__engagement__product=test.engagement.product,
-                                              hash_code=hash_code).exists():
+                                              hash_code=hash_code).exists())):
                         skipped_hashcodes.append(hash_code)
                         continue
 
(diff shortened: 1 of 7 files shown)


aaronweaver:
Thanks @valentijnscholten looks good. I'll just have @devGregA double check and will merge in.

valentijnscholten:
> Wouldn't this cause issues to become duplicates solely on having all endpoints matching ?

The logic hasn't changed, the only thing I changed was the indentation that was broken by e4cbdfa#diff-63c97e945dc0af3507d89798a8841819

devGregA:
@valentijnscholten still reviewing. Would you please correct the flake8 errors: https://travis-ci.org/DefectDojo/django-DefectDojo/jobs/515287349

valentijnscholten:
I have no idea where this second set of commits is coming from, let me check what happened.

valentijnscholten:
Closing because it is pulling in to many commits, will open een new PR based on a branch with only the commits intended for this PR, sorry about the mess :)

aaronweaver:
@valentijnscholten no worries thanks for working on it!

Issue #1002, opened 2019-04-03
Deduplication_on_engagement not supported during import with skip_duplicates


valentijnscholten:
- [X] Enhancement

### Description
During #983 I found that "skip_duplicates" doesn't support deduplication_on_engagment. It always deduplicates across the whole product. I guess this is due to an incomplete implementation of #757 

This the code in serializers.py:

```
                if skip_duplicates:
                    hash_code = item.compute_hash_code()
                    if Finding.objects.filter(Q(active=True) | Q(false_p=True) | Q(duplicate=True),
                                              test__engagement__product=test.engagement.product,
                                              hash_code=hash_code).exists():
                        skipped_hashcodes.append(hash_code)
                        continue
```

It should get aligned with what's in utils.py in the sync_dedupe method.

valentijnscholten:
Fixed by #1006

PR #1007, opened 2019-04-03
make deduplication_on_engagement work with close_old_findings

valentijnscholten:
During #983 and #1002 I realized that deduplication_on_engagement is not yet support with close_old_findings during imports. I think this is needed because otherwise the import could close findings in other engagements and the whole idea of deduplication_on_engagement (I think) is to keep all "duplicate checks" inside the engagement instead of the product.


devGregA:
Hi @valentijnscholten seeing merge conflicts here.

devGregA:
Not sure what happened. Seemed to pop up once I merged your other PR for some reason.

valentijnscholten:
Probably because I had to fix the same flake8 warnings in all three PRs

valentijnscholten:
Conflicts are resolved now.

PR #1012, opened 2019-04-04
jira integration improvements including bulk push to jira

valentijnscholten:
- bulk push to jira
- tooltip on jira isuse link logo/image
- extra fields from engagement sent to jira description field


aaronweaver:
Looks good.

PR #1028, opened 2019-04-09
Fix docker for release (issue 1026)

ptrovatelli:
Starts to fix https://github.com/DefectDojo/django-DefectDojo/issues/1026
This separates the docker build between dev build and release build
- Dev build is the one that was introduced at commit https://github.com/DefectDojo/django-DefectDojo/commit/6d4cc0f4c244675e9e8a6cadf240b29990d35741 and then https://github.com/DefectDojo/django-DefectDojo/commit/e0091ef9442cfc5b315a55de6bedcf4e7049fe1c and tries to start a dockerised dev environement with code hot-loading 
It's still not working very well for me, but this PR isn't about that
- Release build is for running docker-compose based on locally built images (the way it was before commit 6d4cc0f)

I think the solution with the docker-compose.override.yml is not always working because when we try to use the "release" mode (using the override file, with "docker-compose up") docker-compose is not just replacing the uwsgi configuration it's rather doing a merge of both configurations (see https://docs.docker.com/compose/extends/ )
I had it working once I don't know how, most of the time I get this error: 
```
uwsgi_1         |   File "./dojo/wsgi.py", line 29, in <module>
uwsgi_1         |     application = get_wsgi_application()
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/wsgi.py", line 13, in get_wsgi_application
uwsgi_1         |     django.setup(set_prefix=False)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/__init__.py", line 22, in setup
uwsgi_1         |     configure_logging(settings.LOGGING_CONFIG, settings.LOGGING)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 56, in __getattr__
uwsgi_1         |     self._setup(name)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 41, in _setup
uwsgi_1         |     self._wrapped = Settings(settings_module)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 110, in __init__
uwsgi_1         |     mod = importlib.import_module(self.SETTINGS_MODULE)
uwsgi_1         |   File "/usr/local/lib/python2.7/importlib/__init__.py", line 37, in import_module
uwsgi_1         |     __import__(name)
uwsgi_1         | ImportError: No module named settings
uwsgi_1         | unable to load app 0 (mountpoint='') (callable not found or import error)
```

Tested on 
- docker desktop for windows 
```
>docker --version
Docker version 18.09.2, build 6247962

>docker-compose --version
docker-compose version 1.23.2, build 1110ad01

```
- docker on centos 7 VM
```
$docker --version
Docker version 18.06.1-ce, build e68fc7a
$docker-compose --version
docker-compose version 1.22.0, build f46880f
```

The result just seems unpredictible.

I have also removed the mysql volume: for some reason, with this volume the auto-generated admin password is never correct and I cannot login. 
To fix this situation I had to remove the volume configuration and "docker-compose down --volumes"


surewould:
@ptrovatelli see [comment](https://github.com/DefectDojo/django-DefectDojo/issues/1036#issuecomment-481739831) on my thread. TL;DR fix is 'cp dojo/settings/settings.dist.py dojo/settings/settings.py' from django-DefectDojo folder and rebuild. The problems seems to be in the Dockerfile.django file which causes the settings.dist.py file not to be copied over to settings.py for some reason.

aaronweaver:
@ptrovatelli Going to test this in the docker branch.

ptrovatelli:
Hi @aaronweaver ,
This article might be worth digging into : it describes how to set up dev vs release (aka prod) environnement with the override files
https://medium.com/softonic-eng/docker-compose-from-development-to-production-88000124a57c
Few changes would be needed based on what we have in dev branch :
- the dev conf should be in the override file, not the main one
- specific commands are used to run dev or production. 
dev: 
`docker-compose up -d`
production: 
`docker stack deploy my-stack --compose-file docker-compose.yml --with-registry-auth`

Not sure which solution is best at this point, we could try and implement that as well and play along with both versions to see which one is the more handy.

PR #1038, opened 2019-04-11
remove http timeout from the entrypoint as it's not required

ptrovatelli:
After more testing, this timeout is not required: the one set in the nginx.conf alone seems enough right now. 
Anyway this parameter was not active as I've lost a backslash when merging

### DIFF
--- a/docker/entrypoint-uwsgi.sh
+++ b/docker/entrypoint-uwsgi.sh
@@ -6,4 +6,3 @@ exec uwsgi \
   "--${DD_UWSGI_MODE}" "${DD_UWSGI_ENDPOINT}" \
   --protocol uwsgi \
   --wsgi dojo.wsgi:application
-  --http-timeout 1800


valentijnscholten:
I will have to leave this for the other reviewers as I don't have any experience with pyhon/unicorn/uwsgi/...

ptrovatelli:
hi @valentijnscholten 
this is just removing part of my previous commit https://github.com/DefectDojo/django-DefectDojo/commit/f68a4eec547faeab9028d80bc7baae584ebd349d 
that commit on  entrypoint-uwsgi.sh wasn't doing anything anyway because I missed a "\" when merging, we can see in the container using "ps" that the http-timeout is not there. 
I had done lots of testing with nging timeout only a couple of weeks ago, it didn't seem to be enough and I thought that the "http-timeout" had done the trick; Later when I realized that the "http-timeout" wasn't there due to the missing \, I did a couple more tests, it's fine now with the nginx tuning only.

Issue #1037, opened 2019-04-10
duplicate in same vulnerability found by two different scanners


NRGLine4Sec:
How to deduplicate the result of a scan with OpenVAS and Nexpose on the same target ?
When I import the OpenVAS report and the Nexpose report, the vulnerability found are detected as different vulnerability or most of them are the same.

valentijnscholten:
The logic is in utils.py#sync_dedupe method: https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/utils.py. That might help (if you can read Python :))

NRGLine4Sec:
Sorry, I'm not a developer so I understand a little bit Python and I haven't seen any line of code allows to check if a same vulnerability (like same CVE) is discovered by Nexpose and by OpenVAS for example.
And especially it doesn’t tell me how to do it in DefectDojo.
I would at least like to know if it is possible to import a multitude of reports from several scanners to the same target and to obtain a result of vulnerabilities without having duplicates in vulnerabilities discovered by all these scanners.

adracea:
Hey , I was thinking of re-writing the logic behind deduplication using signals as suggested by @aaronweaver , the current deduplication works based on a multitude of elements but they are all specific to the scanner , like xss as reported by burp will have all of these elements different from an xss reported by nessus , this is an option that needs to be explored though .

NRGLine4Sec:
So it is actually impossible to import report of OpenVAS and Nexpose or Nessus and to obtain a result of vulnerability without duplicates ? That is to say to obtain the correlated result of the vulnerabilities of several scanners without duplicates.

valentijnscholten:
There is a pull request (#973 ) coming which adds CVE field, I think that might also help to have better deduplication.

adracea:
I think that based on CWE and endpoint match we could have a decent enough dedupe process , the only issue is that currently most tools don't give a CWE field in the report or not one that is easy to parse (burp suite for example only sometimes gives a CWE-id or sometimes gives more than one) which would imply that we'd need to automatically populate the CWE-id field .
CVE field I find is a bit finicky for dedupes ?
@aaronweaver @devGregA  Any ideas ?

NRGLine4Sec:
Maybe some of these open source projects can help you
https://github.com/archerysec/archerysec
https://github.com/infobyte/faraday
https://github.com/dradis/
https://github.com/cornerpirate/ReportCompilerSource
https://github.com/denimgroup/threadfix

[...]

PR #1043, opened 2019-04-12
Fixes Deduplication Logic

adracea:
This fixes deduplication logic and probably a multitude of issues concerning it .

- [X] Your code is flake8 compliant (DefectDojo's code isn't currently flake8 compliant, but we're trying to correct that.)


adracea:
I think that deduplication in itself should be just that , find fully duplicate issues , mark them as such , any small difference that breaks the pattern should be investigated by someone . If that someone considers the vuln is indeed a duplicate , he should be able to link the issues as duplicate manually via the UI as discussed in #1037 .
Also apiv2 tests seem to be failing , I'll have to look into those .
@aaronweaver @devGregA any suggestions ?

ptrovatelli:
@adracea thanks for this work :). This fixes an issue I've found with dependency check reports. I'll attach it for reference it's for the webgoat app 
[dependency-check-report.zip](https://github.com/DefectDojo/django-DefectDojo/files/3085670/dependency-check-report.zip)

- On dev branch: importing this gives many duplicates although all descriptions are unique and hash_code differ
- with this PR: no duplicates are found when importing the file

i've found another issue though that's present on the dev branch and also on this PR: a nasty refresh problem for finding duplicates across engagments. Have you noticed this? I'm working with this checkmarx report which contains a single vuln: [Webgoat_reference.zip](https://github.com/DefectDojo/django-DefectDojo/files/3085728/Webgoat_reference.zip)
Steps to reproduce (see screenshots in [simple_test.docx](https://github.com/DefectDojo/django-DefectDojo/files/3085736/simple_test.docx) ): 
- system settings: activate dedup (do not delete duplicates)
- create webgoat product
- create engagement webgoat 1.0.0, do not tick "duplicate on engagement". What I understand on this option is that dedup will work across distinct engagements rather than only within the same engagement (although the readthedoc might need a little refresh as it describes a "Deduplication Level" config)
- import scan 
- create engagement webgoat 1.0.1, do not tick "duplicate on engagement". 
- import same scan again
At this point, no duplicate was found which is not correct: it should have found the same finding as a duplicate. 
In the database the hash_code for both findings is b773d31522cb904be49ff505389ad7b48fbf59f66ef1ab0958792775ec477bcc
I have also noted a second occurrence of this in the console output: 
`celeryworker_1  | [2019-04-16 16:04:02,686: INFO/MainProcess] async_dedupe[b679cedb-7d3a-4c5c-8920-6516b9c74ef4]: running deduplication
celeryworker_1  | sync_dedupe for: 2:Stored XSS (Users.java)
`

- Now we'll go to the 2nd finding and add a note
We can see again this message in the console (3rd occurrrence of "running deduplication")
```
celeryworker_1  | [2019-04-16 16:05:47,329: INFO/MainProcess] async_dedupe[52b43624-2fa9-4da9-86f7-08c144be16f7]: running deduplication
celeryworker_1  | sync_dedupe for: 2:Stored XSS (Users.java)
```

- when we go back to the list of engagements the second finding is now correctly detected as a duplicate! 
The hash_code didn't change in the database.
What do you think?

valentijnscholten:
Regarding the webgoat thing:
I think some debug logging for deduplication would be beneficial, currently it's not always clear why sometimes an issues is seen as duplicate of another. There can be multiple reasons, hashcode is not the only criteria. Alternatively the note that is added to a duplicate issue might be enriched with the reason for being seen as duplicate.
I'm not sure I understand your scenario, but what I have seen is that deduplication is happening asynchronously. So it can take a couple of minutes before it is completed / visible in the UI. I am not seeing that is the issue here, just checking to make sure you are aware that it is happening in the background.

Regarding the owasp dependency checker thing:
I have seen the same, it deduplicates because the filename + linenumber + cwe is the same. They are not really duplicates, on the other hand it combines all vulnerabilities in a single dependency into 1 single finding. That's not to bad because upgrading that dependency will most probably fix all the vulnerabilities. While writing this I think it might indeed be better to keep them separate as sometimes people only upgrade to the latest minor version possible and not do a major upgrade. In this case it might be that some vulnerabilities are resolved, and some have to remain open.

adracea:
I've tested this by forcing the deduplication via `manage.py dedupe` but I personally don't think I've missed any Super saves outside the necessary ones ? I'll look into it a bit more and see what's what , but , I'd really like to see more opinions on this . I'm also not sure why and I haven't been able to test , why the apiv2 tests failed .

ptrovatelli:
@valentijnscholten 
- for the checkmarx thing: i've seen thoses "async..." logs but it's just a single vuln report so I assumed it couldn't be that slow. i'll do more tests and wait longer :). The scenario is just importing the same report twice, into a different engagement (each report is seen as a release of the application having the same vuln that wasn't fixed between the first and second release); At first the duplicate is not found, after editing the second finding, all of sudden it's seen as duplicate. 

- for dependency check: I actually agree that it can be seen as better to "over-dedup" as we used to do because it makes it easier to analyse afterwards and we'll probably fix everything at once like you said. However if it's intended, it should be documented because it's a bit misleading. Also if we keep this, in the example  file i've come across a "high" and a "medium" vuln but the medium one was kept. If we go that way I think we should keep the highest severity one, or upgrade the rate of the one we keep with the highest severity if it's not too complicated

ptrovatelli:
> we should keep the highest severity one, or upgrade the rate of the one we keep with the highest severity if it's not too complicated

even better: how about, if two findings don't have the same severity they can't be duplicate?

ptrovatelli:
- regarding the checkmarx thing: calling this
`./manage.py dedupe`
will indeed make the GUI be updated and find the duplicates across engagements. Like adding a note to a finding will.  But if we don't do anything , just refresh the page or browse the GUI the deduplication will not happen. I've waited 10 mn :)

ptrovatelli:
by the way it's the same behavior if we're importing the reports inside of the same engagement. It's only after adding a note to a finding that the duplicate is found. 
See screenshots 
[simple_test_same_engagement.docx](https://github.com/DefectDojo/django-DefectDojo/files/3089230/simple_test_same_engagement.docx)

valentijnscholten:
Is that on dev branch or this PR or both?

ptrovatelli:
it's on both dev and this PR branches

ptrovatelli:
@valentijnscholten sorry i've just retested on dev branch: now this works on dev branch.

ptrovatelli:
@adracea do you reproduce? I thought I had reproduced on dev too but maybe i messed up with my repo, or there is something else that makes it happen or not.
This log is a bit intriguing but it's there so often that I don't even notice anymore; when starting up the app: 
```
rabbitmq_1      | 2019-04-17 11:39:11.343 [info] <0.559.0> accepting AMQP connection <0.559.0> (172.20.0.7:34318 -> 172.20.0.2:5672)
rabbitmq_1      | 2019-04-17 11:39:11.365 [info] <0.559.0> connection <0.559.0> (172.20.0.7:34318 -> 172.20.0.2:5672): user 'guest' authenticated and granted access to vhost '/'
celeryworker_1  | [2019-04-17 11:39:11,398: INFO/MainProcess] Received task: dojo.tasks.async_dupe_delete[5f8d7404-831f-4a45-9ca6-1b84754ae9ed]
celeryworker_1  | [2019-04-17 11:39:11,399: INFO/MainProcess] dojo.tasks.async_dupe_delete[5f8d7404-831f-4a45-9ca6-1b84754ae9ed]: delete excess duplicates
celeryworker_1  | [2019-04-17 11:39:11,462: ERROR/MainProcess] Task dojo.tasks.async_dupe_delete[5f8d7404-831f-4a45-9ca6-1b84754ae9ed] raised unexpected: DoesNotExist('System_Settings matching query does not exist.',)
celeryworker_1  | Traceback (most recent call last):
celeryworker_1  |   File "/usr/local/lib/python2.7/site-packages/celery/app/trace.py", line 382, in trace_task
celeryworker_1  |     R = retval = fun(*args, **kwargs)
celeryworker_1  |   File "/usr/local/lib/python2.7/site-packages/celery/app/trace.py", line 641, in __protected_call__
celeryworker_1  |     return self.run(*args, **kwargs)
celeryworker_1  |   File "/app/dojo/tasks.py", line 283, in async_dupe_delete
celeryworker_1  |     system_settings = System_Settings.objects.get()
celeryworker_1  |   File "/usr/local/lib/python2.7/site-packages/django/db/models/manager.py", line 85, in manager_method
celeryworker_1  |     return getattr(self.get_queryset(), name)(*args, **kwargs)
celeryworker_1  |   File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 380, in get
celeryworker_1  |     self.model._meta.object_name
celeryworker_1  | Exception: System_Settings matching query does not exist.
```

I've attach dev and pr-1043 logs. For each environment i did two tests: 
- import report twice, into different engagement
- import report twice, into same engagement

I did each time a clean checkout, rebuild images, "docker-compose down --volumes", use "release" mode
I did not edit findings in these logs. 

[duplicate_not_found_logs.zip](https://github.com/DefectDojo/django-DefectDojo/files/3089431/duplicate_not_found_logs.zip)

Thanks

adracea:
That is strange...this doesn't change *when* deduplication happens , it just changes *how* it happens . This might be one of the implications of re-calculating the `hash_code` outside of the new findings area . The only problem I have with that is that it feels like there's some data that might be missing at that time .
I think I should change that like to something like
```python
if not self.hash_code:
    self.hash_code = self.compute_hash_code()
```
So that it calculates it only once and not every time it calls back upon the save function .

ptrovatelli:
well done fixing travis, still the refresh issue is present :). Am I the only one to reproduce this? 
Also regarding the dependency check dedup, what happened to the idea of keeping the old system (dedup on filename + linenumber + cwe ) + add the severity in the dedup keys to make sure we don't hide a high behind a medium? Analysing dependency check reports may be very time consuming, I think the old system was good after all for this. Not sure about other types of reports though. Would that be complicated to add report-specific rules if this is problematic for other reports?

adracea:
@ptrovatelli for me , this only seems to be happening only when I import the scan with deduplication disabled and then turning it on , it won't immediately catch on to the fact that the issues are duplicates .

ptrovatelli:
@adracea have you tried with the provided checkmarx report? I've done tests with other reports: no problem with "zap_sample.xml" from DD "tests" folder but I have the issue with this every checkmarx report (single or multi-vuln reports). The issue is something about checkmarx reports.

PR #1044, opened 2019-04-12
Slimmer docker image

alles-klar:
Reduces the size of the docker image

---

### DIFF
--- a/Dockerfile.django
+++ b/Dockerfile.django
@@ -1,6 +1,7 @@
 
 # code: language=Dockerfile
-FROM python:2
+########## Stage: build ##########
+FROM python:2 as build
 WORKDIR /app
 RUN \
   apt-get -y update && \
@@ -13,7 +14,36 @@ RUN \
   rm -rf /var/lib/apt/lists && \
   true
 COPY requirements.txt ./
-RUN pip install -r ./requirements.txt
+RUN pip install \
+	--no-cache-dir \
+	-r ./requirements.txt
+RUN pip wheel --wheel-dir=/tmp/wheels -r ./requirements.txt
+
+########## Stage: release ##########
+FROM python:2-slim as release
+WORKDIR /app
+RUN \
+  apt-get -y update && \
+  apt-get -y install --no-install-recommends \
+	  # libopenjp2-7 libjpeg62 libtiff5 are required by the pillow package
+    libopenjp2-7 \
+		libjpeg62 \
+		libtiff5 \
+    dnsutils \
+		mysql-client \
+		libmariadbclient18 \
+    && \
+  apt-get clean && \
+  rm -rf /var/lib/apt/lists && \
+  true
+RUN pip install --no-cache-dir --upgrade pip
+COPY --from=build /tmp/wheels /tmp/wheels	
+COPY requirements.txt ./
+RUN pip install \
+	--no-cache-dir \
+  --no-index \
+  --find-links=/tmp/wheels \
+	-r ./requirements.txt
 COPY \
   docker/entrypoint-celery-beat.sh \
   docker/entrypoint-celery-worker.sh \


ptrovatelli:
@alles-klar @aaronweaver I think this broke the nginx image build. 
Can be seen in travis logs; We need to improve travis test to end KO in that case :)

https://travis-ci.org/DefectDojo/django-DefectDojo/jobs/519142559
unfold "Sending build context to Docker daemon  7.414MB"

```
Processing triggers for libc-bin (2.24-11+deb9u4) ...
/bin/sh: 1: curl: not found
E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation
The command '/bin/sh -c apt-get -y update &&   apt-get -y install apt-transport-https ca-certificates &&   curl -sSL https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - &&   echo "deb https://deb.nodesource.com/node_11.x stretch main" | tee /etc/apt/sources.list.d/nodesource.list &&   curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - &&   echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list &&   apt-get -y update &&   apt-get -y install nodejs &&   apt-get -y install --no-install-recommends yarn &&   apt-get clean &&   rm -rf /var/lib/apt/lists &&   true' returned a non-zero code: 255
```

aaronweaver:
I'll back it out. I don't know why the smoke test didn't pick that up. I'm wondering if the kubernetes containers aren't pulling locally even though it's specified to not pull from docker hub.

ppiotr3k:
> @alles-klar @aaronweaver I think this broke the nginx image build.

I agree the build is broken when pulling locally; looks like `python:2-slim` (`Dockerfile.django`) is lacking the following packages `curl` (or `wget`, smaller footprint), `gnupg`, and `git`, needed in `Dockerfile.nginx`.
Should they be added to `Dockerfile.nginx`? If so, I have code ready for PR fixing it that way.


> Can be seen in travis logs; We need to improve travis test to end KO in that case :)

Indeed. Any idea why it's not failing and how to fix it?

ptrovatelli:
about the travis tests, we build the images but don't check the return code. When the build fails we don't notice because the app will probably run with a previously built image, or with the dockerhub one. I'd say we should check the return code to make sure that this "docker build" ends with return code 0, which is most probably not the case when the build fails

ptrovatelli:
maybe it's not that commit the issue actually; i have the same problem in my PR branch https://github.com/DefectDojo/django-DefectDojo/pull/1045 although I don't have your changed code. I don't know where this comes from then, maybe the python:2 image changed and lost curl?

ptrovatelli:
See an example of travis change here to detect errors while building: 
https://github.com/ptrovatelli/django-DefectDojo/compare/a89370bebe8f63d5704a7642df97bb4b6da947a0...fix-docker-with-override
You can ignore the change in the building command though (this PR uses docker-compose to build images)
```
travis/script.sh:
  docker-compose -f docker-compose.yml build
  return_value=${?}
  if [ ${return_value} -ne 0 ]; then
    echo "ERROR: cannot build django or nginx image"
    exit ${return_value}
  fi
```

alles-klar:
> > @alles-klar @aaronweaver I think this broke the nginx image build.
> 
> I agree the build is broken when pulling locally; looks like `python:2-slim` (`Dockerfile.django`) is lacking the following packages `curl` (or `wget`, smaller footprint), `gnupg`, and `git`, needed in `Dockerfile.nginx`.
> Should they be added to `Dockerfile.nginx`? If so, I have code ready for PR fixing it that way.

I don't think it's a good idea to install all missing packages if the goal is to build a small image. This can become very complex. Maybe we should create a base image and use it to build the django and nginx image? This would remove the dependency between these two images.

ptrovatelli:
@alles-klar @aaronweaver  for the record: 
- nginx image: size is stable 191MB
- django image: slimmed from 1.31GB to 599MB (that is from https://github.com/DefectDojo/django-DefectDojo/pull/1105 , tested right now). This is nice :)

PR #1045, opened 2019-04-12
Fix docker with override

ptrovatelli:
Edit 15/04/2019: i'm putting a bit of context on this PR. 
This is for putting in place a docker-compose configuration that handles release and dev modes with minimal configuration duplication, using a docker-compose.override.yml file for dev mode with hot-reloading of python and static resources
Everything is explained in the DOCKER.md
The docker-compose setting is inspired by this article: https://medium.com/softonic-eng/docker-compose-from-development-to-production-88000124a57c
Currently on the dev branch, the system in place uses a set of 3 files: 
- release mode: docker-compose_base.yml + docker-compose_uwsgi-release.yml
- dev mode: docker-compose_base.yml + docker-compose_uwsgi-dev.yml

Advantages of this PR versus the current configuration on the dev branch: 
- with the override file, minimal duplication is done and it's easy to find out what changes between release and dev mode: it's what is in docker-compose.override.yml 
- with the 3 yml files some things are duplicated and to know what changes one need to diff the files
- for dev mode, i have handled the python hot-reloading with uwsgi which allows us to keep the nginx server in the architecture so we have an architecture that's identical in dev and release (easier to maintain, spot issues earlier). In the dev branch currently, nginx is not present in the dev configuration and I'm still not able to load static resources (even with DD_WHITENOISE=on) so the app is not pretty. 
- added the mysql port forwarding so that we can access the DB from outside the container int he dev setting
- updated the documentation with details about how docker-compose works and how to fix the permission issue on settings.py (thanks to @surewould)
- added environments variables to handle version tagging of images by docker-compose

You're welcome to try out this PR to make sure it works for everyone. To try this, stash or commit your work then: 
```
git fetch origin pull/1045/head:pr-1045
git checkout pr-1045

```
Thanks

Now back to the original discussion

@aaronweaver @ all  how about something like that? I still need to add some documentation but can you try this and see how it works? seems good to me but haven't done lots of testing. 

release/prod mode: 
```
# building
docker-compose -f docker-compose.yml build
# running
docker-compose -f docker-compose.yml up
```

dev mode: 
```
# we still need a base image to work on, so we need to build it at least once to make sure we have an image coherent with our code
docker-compose -f docker-compose.yml build
docker-compose up
```

- Hot-loading: tested on python file (example forms.py); still needs restarting: 
```
ctrl+C
docker-compose up
```

- Hot-loading for nginx static resources doesn't need restarting (example dojo/static/dojo/css/dojo.css). Just ctrl + F5

- Point of attention: when modifying an entrypoint script, we need to rebuild the image to have the modification in place

Possible improvments: we should update the docker-compose file in order to build the nginx image just like the uwsgi one. 
After that we can probably remove the instructions for building images with "docker build" as "docker-compose -f docker-compose.yml build" will build everything

by the way: I have removed the "dev entrypoint" that was added: this created a different architecture for dev (where we would access the uwsgi container directly on exposed port 8000) and the release/prod where we would pass through nginx. I think it will be easier to maintain and spot issues early if we have the same architecture for dev and release/prod. 
We could go back on that, but at least one thing was missing: nginx was handling some static resources, so we needed to move those to the uwsgi container (see Dockerfile.nginx: COPY --from=build /app/static/ /usr/share/nginx/html/static/ )


aaronweaver:
@ptrovatelli I prefer hot-reloading as that's a much easier workflow then restarting the container each time. We can have it go through nginx, but there would need to be a different config as it's using binary uwsgi. For the static files we could consider a multi-stage docker, I had one define before and there is a PR for multi-stage. I haven't looked closely at it.

ptrovatelli:
> @ptrovatelli I prefer hot-reloading as that's a much easier workflow then restarting the container each time.

the entrypoint-uwsgi-dev.sh did python hot reloading without restarting the app? with this PR, the python files are persisted to the uwsgi container through the volume but the application doesn't seem to be picking them up without restarting. I assumed it was just a python limitation

ptrovatelli:
@aaronweaver so that left us two solutions
- not using nginx; then we would need to configure django to serve the static files, something like that https://docs.djangoproject.com/en/2.1/howto/static-files/ 
- using nginx: i've found a solution to enable hot reloading and keep nginx. We can't use that "manage.py runserver" because it means that the uwsgi container is not really using uwsgi and then nginx can't find the uwsgi upstream. I was able to fix this issue with a simple argument to the uwsgi call: "--py-autoreload 1" will handle the hot reloading of python code by uwsgi!

There is a single issue remaining now: that settings.py file is not so easy to copy. For some reason, we cannot modify the volume mounted by docker-compose from inside the container even though we're root and we can't chown files either
```
I have no name!@78c15a1a4a4e:/app$ id -a
uid=1001 gid=0(root) groups=0(root)

I have no name!@b7917364d363:/app$ ls -la dojo/settings
total 36
drwxr-xr-x  2 1000 mail  4096 Apr 12 15:32 .
drwxr-xr-x 41 1000 mail  4096 Apr 12 15:33 ..
-rw-r--r--  1 1000 mail     0 Apr 12 15:32 __init__.py
-rw-r--r--  1 1000 mail 18702 Apr 12 15:32 settings.dist.py
-rw-r--r--  1 1000 mail  2078 Apr 12 15:32 template-env
-rw-r--r--  1 1000 mail   382 Apr 12 15:32 unittest.py
```

see chown in entrypoint-uwsgi-dev.sh, it gives: 
`uwsgi_1         | chown: changing ownership of '/app/dojo/settings/': Operation not permitted`

There is an easy workaround though: copy that file manuallly from the host, then we don't have any issue. If we do that, we need to discard my last commit as we won't need to write to the volume from inside the container
The "z" flag is supposed to fix that issue for SELinux. I'm not on SELinux though: 
```
$ getenforce
Disabled

```

aaronweaver:
Thanks, I wasn't aware of the reload in uwsgi. Is copying settings.py something to do with Debian and docker? I have no issues with it on mac. For serving files, all that needs to be done is set whitenoise, DD_WHITENOISE=on. 

I'm going to let this PR sit for a bit if that's ok and we can have others try it out. It's been very time consuming constantly updating everything docker, kubernetes, travis and readthedocs.

ptrovatelli:
sure i'll do some more testing on various environments too. the risk also is that the application tries to write to that volume at some point, I don't know if that can happen, I haven't done much testing yet apart from logging in.

ptrovatelli:
OK I got to the bottom of this permission issue. For volumes mounted by docker-compose to be writable from within the container, you're supposed to pick a user in the container with the same uid as the host user that runs "docker-compose up" and mounts the volume (at least for Debian distros and probably most linux ones)
The user that runs the django-defectdojo_uwsgi_1 container has uid 1001 (set in Dockerfile.django). Changing this to the host user's uid makes that settings.py copy possible (and any other writing into /app/)
I dug a little to see if we can automate this but it's really a mess. This article is an interesting read https://jtreminio.com/blog/running-docker-containers-as-current-host-user/ . For me it wasn't working though because my guid was already taken by another group in the python:2 image :). We could keep part of that article, only the uid not the gid, and use "www-data" to run everything but's I think it's better to leave it here for what we need: 
- either manually copy settings.py before running the app:
`cp dojo/settings/settings.dist.py dojo/settings/settings.py `
- or update the USER in Dockerfile.django and rebuild the image. Eg: 
`USER 1000`

madchap:
> For serving files, all that needs to be done is set whitenoise, DD_WHITENOISE=on.

@aaronweaver Should we expect to see all images, .js, .css etc... to then suddenly work? Because setting just this variable in my case is not working. There is only the `dojo` directory under `static` in my container and the site is of course barely useable (even though better than nothing).

Good work @ptrovatelli by the way :)

I had also the permission issue on Linux, copied from the host as you did. It is possible that the most valuable workaround is to have that documented and in the comment of the `.sh` file.

surewould:
Nice! @ptrovatelli. I was going crazy trying to figure out the "right way" to fix that permission issue instead of just copying the file over beforehand. That article was very useful. The part about why Windows and Mac users don't run into the issue explains why @aaronweaver wasn't seeing it either. ^_^

alles-klar:
@ptrovatelli the idea is very nice!

Instead of the setEnv script we also could use the docker-compose command with the -f option (https://docs.docker.com/compose/extends/). I would prefer this method because copying the compose files could be very error prune. What do you think?

For example:
`
docker-compose -f .\docker-compose.yml -f .\docker-compose.override.dev.yml up
`
instead of 
`
docker/setEnvDev.sh
`
`
docker-compose up
`

ppiotr3k:
>  I would prefer this method because copying the compose files could be very error prune. What do you think?
> 

I don't like copying either. What about a symbolic link?
It is an approach used by many applications. As an example, Apache `a2ensite` command creates symbolic links in `sites-enabled/` from `sites-available/` to enable a configuration.

From an end-user experience perspective, what bothers me the most it that:
- there is no default settings as of today
- multiple `-f` is counter-intuitive for most people who expect a simple `docker-compose up`
- each time I have a new shell environment, I must explicitly set aliases, either for _dev_ or _release_

These could be fixed by the symbolic link approach, and default settings, either _dev_ or _release_.
Neat, clean and user-friendly.

If we still must source aliases, and not use symlinks, could you please regroup the aliases in a single file, for example `docker/setenv.sh`, or whatever preferred other name?
The call therefore could be something like `. docker/setenv.sh dev` or `. docker/setenv.sh release`.

alles-klar:
> From an end-user experience perspective, what bothers me the most it that:
> 
> * there is no default settings as of today
> * multiple `-f` is counter-intuitive for most people who expect a simple `docker-compose up`
> * each time I have a new shell environment, I must explicitly set aliases, either for _dev_ or _release_

You can simply run docker-compose up and it will use the docker-compose.yml and docker-compose.override.yml  by default. Only for development or other purposes you have to use `-f`. You do not need to set aliases, if you want to use the default values.

ppiotr3k:
> You can simply run docker-compose up and it will use the docker-compose.yml and docker-compose.override.yml by default.

@alles-klar, there is no `docker-compose.yml` and `docker-compose.override.yml` files in the repository as of today, there are three files. And the documentation in [`DOCKER.md`, as of today](https://github.com/DefectDojo/django-DefectDojo/blob/dev/DOCKER.md#setup-via-docker-compose), points to either setting up aliases, or using multiple `-f`.

ptrovatelli:
@alles-klar @ppiotr3k thanks for your feedback :). On dev branch right now "docker-compose up" without -f is not working, we need those multiple "-f". On this PR, it works and use the default configuration (the release one). 
I agree that multiple -f is annoying; aliases make it OK I think but of course one need to think of loading the alias first.
It's true that copying is a little error prone especially if you start modifying the files; and you may not know at a given point what's your current configuration. 
I'd thought of symlinks too but the difficulty is to provide a working configuration right after git clone without duplicating files. I'm discovering now that git seems to support commiting symlinks so it might be doable

ppiotr3k:
@ptrovatelli, I think we might go without committing symlinks to Git.
Assuming `docker-compose.yml` contains everything needed for _release_, and `docker-compose.dev.yml` contains everything needed to create an override for _dev_, I wrote this script to switch from one environment to the other.
It is based on the previous idea of creating a symlink called `docker-compose.override.yml` pointing to `docker-compose.dev.yml` when activating the _dev_ configuration, and removing it when using the _release_ configuration.

```bash
#!/bin/bash

# docker-compose.*.yml files names and their position compared to this script.
# Here: in parent directory.
target_dir="${0%/*}/.."
override_link='docker-compose.override.yml'
override_file='docker-compose.dev.yml'

function show_current {
    if [ -L ${override_link} ]
    then
        echo "Using 'dev' configuration."
        echo "-> swicth to 'release' with '${0} release'"
    else
        echo "Using 'release' configuration."
        echo "-> swicth to 'dev' with '${0} dev'"
    fi
}


function set_release {
    if [ -L ${override_link} ]
    then
        rm ${override_link}
        echo "Now using 'release' configuration."
    else
        echo "Already using 'release' configuration."
    fi
}


function set_dev {
    if ! [ -L ${override_link} ]
    then
        ln -s ${override_file} ${override_link}
        echo "Now using 'dev' configuration."
    else
        echo "Already using 'dev' configuration."
    fi
}

# Change directory to allow working with relative paths.
cd ${target_dir}

if [ ${#} -eq 1 ] && [[ 'dev release' =~ "${1}" ]]
then
    set_"${1}"
else
    show_current
fi
```
What do you think?
I could use some _shell tricks_ to make it shorter and avoid duplication/similarities, but it would be less readable/accessible, so I proposed this version.

Tested on Linux. Should work on macOS.
Windows users will have to manage the symlink by hand.

ptrovatelli:
@ppiotr3k thanks for this but we also need an override for release see https://github.com/ptrovatelli/django-DefectDojo/blob/fix-docker-with-override/docker-compose.override.yml . It contains the mysql volume. We cannot putit in the main docker-compose.yml because it makes the unit test env fail (it not possible to remove a volume with an override). Which is why I think wee need to commit a symlink with the default override linked to the release one.

ppiotr3k:
Oh! Thanks @ptrovatelli for the explanation, helped me a lot! :pray:
-> as the proposed [docker-compose.yml](https://github.com/DefectDojo/django-DefectDojo/blob/f684104e1bc38b4829ef807889afdfb4d2a9b61d/docker-compose.yml) has no volume, and the volumes for MySQL in [docker-compose.override.yml](https://github.com/DefectDojo/django-DefectDojo/blob/f684104e1bc38b4829ef807889afdfb4d2a9b61d/docker-compose.override.yml) and [docker-compose.override.dev.yml](https://github.com/DefectDojo/django-DefectDojo/blob/f684104e1bc38b4829ef807889afdfb4d2a9b61d/docker-compose.override.dev.yml) are the same, I didn't get the issue is from a third environment, _unittests_, maybe because I haven't see it mentioned in the conversation before.

So currently, if I understand correctly, the proposed solution would work for _dev_ and _release_, but wouldn't work anymore since also introducing a _unittests_ environment. Is there anything else you see which would be an issue with _unittests_ besides the volume?

alles-klar:
@ppiotr3k can you describe your problem with the volume when you start unittests? I guess if you remove the volumes afterwards with `docker-compose down --volumes` it may solve your problem.

ptrovatelli:
@alles-klar if you use `docker-compose down --volumes` you'll loose all your data, I don't think it's a good idea. However, there's probably a way to make the unit test env work with the volume mounted but I didn't dig it as It works without the volumes and we don't really need the volume for unit tests.
The unit test uses a database called "test_defectdojo" and the other environements use a database "defectdojo" so technically they could coexist in the same volume. However for some reason, when starting the unit test env when the regular database is already existing, the test database creation fails. I don't have the stack trace here, I can post it later.

ptrovatelli:
@alles-klar here's what happens when we try to start the unit test env and another database already exists (the release or dev database from the volume) :

```
uwsgi_1         |     errors.extend(cls._check_fields(**kwargs))
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/models/base.py", line 1359, in _check_fields
uwsgi_1         |     errors.extend(field.check(**kwargs))
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/models/fields/__init__.py", line 913, in check
uwsgi_1         |     errors = super(AutoField, self).check(**kwargs)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/models/fields/__init__.py", line 219, in check
uwsgi_1         |     errors.extend(self._check_backend_specific_checks(**kwargs))
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/models/fields/__init__.py", line 322, in _check_backend_specific_checks
uwsgi_1         |     return connections[db].validation.check_field(self, **kwargs)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/mysql/validation.py", line 49, in check_field
uwsgi_1         |     field_type = field.db_type(self.connection)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/models/fields/__init__.py", line 644, in db_type
uwsgi_1         |     return connection.data_types[self.get_internal_type()] % data
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/utils/functional.py", line 35, in __get__
uwsgi_1         |     res = instance.__dict__[self.name] = self.func(instance)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/mysql/base.py", line 174, in data_types
uwsgi_1         |     if self.features.supports_microsecond_precision:
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/utils/functional.py", line 35, in __get__
uwsgi_1         |     res = instance.__dict__[self.name] = self.func(instance)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/mysql/features.py", line 53, in supports_microsecond_precision
uwsgi_1         |     return self.connection.mysql_version >= (5, 6, 4) and Database.version_info >= (1, 2, 5)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/utils/functional.py", line 35, in __get__
uwsgi_1         |     res = instance.__dict__[self.name] = self.func(instance)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/mysql/base.py", line 385, in mysql_version
uwsgi_1         |     with self.temporary_connection() as cursor:
uwsgi_1         |   File "/usr/local/lib/python2.7/contextlib.py", line 17, in __enter__
uwsgi_1         |     return self.gen.next()
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/base/base.py", line 591, in temporary_connection
uwsgi_1         |     cursor = self.cursor()
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/base/base.py", line 254, in cursor
uwsgi_1         |     return self._cursor()
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/base/base.py", line 229, in _cursor
uwsgi_1         |     self.ensure_connection()
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/base/base.py", line 213, in ensure_connection
uwsgi_1         |     self.connect()
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/utils.py", line 94, in __exit__
uwsgi_1         |     six.reraise(dj_exc_type, dj_exc_value, traceback)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/base/base.py", line 213, in ensure_connection
uwsgi_1         |     self.connect()
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/base/base.py", line 189, in connect
uwsgi_1         |     self.connection = self.get_new_connection(conn_params)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/mysql/base.py", line 274, in get_new_connection
uwsgi_1         |     conn = Database.connect(**conn_params)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/MySQLdb/__init__.py", line 86, in Connect
uwsgi_1         |     return Connection(*args, **kwargs)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/MySQLdb/connections.py", line 204, in __init__
uwsgi_1         |     super(Connection, self).__init__(*args, **kwargs2)
uwsgi_1         | django.db.utils.OperationalError: (1044, "Access denied for user 'defectdojo'@'%' to database 'test_defectdojo'")

```

alles-klar:
@ptrovatelli this makes sense because mysql uses the volume and does not start the init process. So the mysql database test_defectdojo is not created. You can fix this, if you specify another volume or creating the test database during the first run.

ptrovatelli:
@alles-klar good thinking! creating a volume with a different name on the same folder makes it possible to override the volume created in the main docker-compose.yml to another volume
@ppiotr3k I have used your script and made small changes to make it work for all three environments. What do you think?

ptrovatelli:
I still need to add the celeryworker code hot reloading for dev env (useful when working on deduplication)

ppiotr3k:
> @ppiotr3k I have used your script and made small changes to make it work for all three environments. What do you think?

I like it, well done :smiley: ! There is a simplification we can add, I'll get back to you shortly with a proposal.

ptrovatelli:
I've just added a volume for the celeryworker container so that changed files on the host are mapped inside the container and we don't need to recreate the docker image after each change! It's not doing hot-reloading though. A few links if someone feels like playing with this, I think I'll leave it as it is for now: 
- with watchdog or django:
https://avilpage.com/2017/05/how-to-auto-reload-celery-workers-in-development.html
- with docker
https://medium.com/aubergine-solutions/auto-reload-development-mode-for-celery-worker-using-docker-compose-and-django-management-2ba8e313eb37

For now, just restart the celeryworker container with: 
`docker restart django-defectdojo_celeryworker_1`

Docker.md is up to date

ptrovatelli:
@ppiotr3k what is this simplification you had in mind? the set_dev and set_unit_tests functions are similar and code could be factorised is that it? However set_release is slightly different and would required adding a specific test in order not to create the link, if factorised.
I think that branch is close to be mature enough to be merged; people keep trying to make the dev mode work with the aliases_dev.sh but I'm not sure that's ever worked...

devGregA:
Is this ready to merge @ptrovatelli @aaronweaver?

ptrovatelli:
@devGregA  as far as I know, the setEnv script could use a little refactoring, other than that I think we can merge. We can do that refactoring the next time we add an environment (it would be nice to have an environment with selenium for smoke tests)
I don't know if @ppiotr3k  had something else in mind.

PR #1054, opened 2019-04-15
Close/reopen an engagement from API

twsagarcia:

- [X] Your code is flake8 compliant (DefectDojo's code isn't currently flake8 compliant, but we're trying to correct that.).

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -1,11 +1,11 @@
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404
-from django.utils import timezone
 from rest_framework import viewsets, mixins
 from rest_framework.permissions import DjangoModelPermissions
 from rest_framework.decorators import detail_route
 from django_filters.rest_framework import DjangoFilterBackend
 
+from dojo.engagement.services import close_engagement, reopen_engagement
 from dojo.models import Product, Product_Type, Engagement, Test, Test_Type, Finding, \
     User, ScanSettings, Scan, Stub_Finding, Finding_Template, \
     JIRA_Issue, Tool_Product_Settings, Tool_Configuration, Tool_Type, \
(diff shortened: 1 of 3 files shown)


adracea:
@valentijnscholten I think you're more familiar with this area since I haven't used api_v2 as much , at first glance it looks good .

valentijnscholten:
I like the addition, but would like to see the logic moved out of the api file and maybe reuse the logic in engagement/views.py or have shared logic somewhere else. The existing logic also closed the epic in JIRA, so I think that should be performed as well when closing/opening via API.

twsagarcia:
@valentijnscholten that's a good idea. 

There are a lot of possible ways to do it, what does suit better for you guys?

To move out the common logic: 
- as methods at the Engagement model. e.g. def close(self) / def reopen(self).
- in a new module in engagements or in engagements.__init__.py. Any proposal for naming the module? The methods would be something like: def close(eng) / def reopen(eng).
- in new methods in engagements/views.py. Any proposal for naming these methods?
- somewhere else. Proposals?

jpescalona:
@twsagarcia i would suggest to create this logic on a common module on views. Model is not the correct place to handle logic. You should not overthink where to place the logic, it just a matter of the concept of [domain model](https://stackoverflow.com/questions/12578908/separation-of-business-logic-and-data-access-in-django/12579490#12579490), probably a helper module under engagements will be okay for this purpose.

twsagarcia:
Updated PR with @jpescalona and @valentijnscholten proposals. 
Please review.

valentijnscholten:
Great, indeed no need to overthink but very good that "logic" like this in one place.

jpescalona:
Great job @twsagarcia!

devGregA:
@aaronweaver would you please review & merge if it looks good to you.

PR #1059, opened 2019-04-17
UI Template Modularization and Restructuring

efficiosoft:
Hi,

While exploring the existing template structure and fixing usage of the aria API, I found the templates to include significant redundancy, which makes ensuring and maintaining accessibility quite hard. Of course, maintainability in general suffers a lot from this fact as well.

That's why I started thinking about making the templates more modular by creating reusable parts for various components. As a proof of concept, I did this for dropdown menus, panels and tag links in this PR. The ``view_finding.html`` template now uses these components.

I don't expect this PR to be merged as is. It's by no means complete. Rather take it as a starting point for a discussion about restructuring the UI parts in both an accessible and maintainable, future-proof way that could make using the whole system more consistent and easy.

Best regards
Robert


efficiosoft:
Hmm, is there a chance to mute the accessibility linting bot for this PR? I know there are issues, but for now it'll just disturb the conversation...

valentijnscholten:
Adding 8 labels will mute it ;-)

efficiosoft:
This is a proof of concept at the moment. And the labels mentioned are not necessary at all for accessibility.

efficiosoft:
In general, buttons having a text themselves, like ``<input type="submit" value="some text"/>`` or ``<button>some text</button>`` don't need an additional aria-label, however the bot doesn't recognize this for some reason.

valentijnscholten:
I am not familiar with accessibillity in web development, but it seems these aria fields are needed / helpful for screenreaders etc. Last time I checked @aaronweaver wants to implement the suggestions from the accesslinter as we have users depending on good accessibility. 

In general I like your idea to having reusable UI fragments, but I would like to check with @aaronweaver to see if there are other/conflicting refactorings planned for the future that might make this PR obsolete.

efficiosoft:
I'm a screenreader user myself and probably brought this topic up last week. That's why I can say - at least for buttons with a text label - no aria-label is required and would just add redundancy to be maintained later.

efficiosoft:
Originally, I just planned to make all parts of dojo's UI accessible, but during this work I noticed that it would be more reasonable to refactor the whole thing into reusable fragments straight away.

aaronweaver:
@efficiosoft Current state + fine tuning sounds good.

PR #1076, opened 2019-04-22
add checkmarx parser unit tests

ptrovatelli:
it's not covering the hot topics of the moment (deduplication, hash computing, database inserting) but it's a starter for checkmarx unit tests. 
To go further, would you advise adding some smoke tests or do you think it can be easily done with unitests? 
Thanks


devGregA:
Hi @ptrovatelli, would it be possible to put the sample scans here: https://github.com/DefectDojo/sample-scan-files to keep the core code base as lean as possible?

ptrovatelli:
@devGregA all the other unit test scan files are here with the unit tests. How can we access files from another git repo in the unit tests?

aaronweaver:
@devGregA you ok with merging? I agree we'd rather pull for the repo but we can do that with the GSOC candidates and in the meantime have coverage for checkmarx.

ptrovatelli:
OK, so what I can do is clone that repo in unit-tests.sh so that the scans are accessible for all tests?

devGregA:
@aaronweaver good with merging

PR #1079, opened 2019-04-23
Add optional title field to Test

efficiosoft:
Occasionally, it comes handy to name a test. Therefore, I added a new optional ``title`` field to the ``Test`` model which is represented alongside the test type everywhere.


devGregA:
@efficiosoft thank you for your PR! Some of the unit tests are failing https://travis-ci.org/DefectDojo/django-DefectDojo/jobs/523506506

efficiosoft:
@devGregA Thanks, hadn't seen the results yet.

valentijnscholten:
Isn't 1000 maxlength a bit steep for a title field? 255 or even 100 should suffice I think. It's not as if anything more fits on the screen in the places where it's being displayed :-D

efficiosoft:
@valentijnscholten You're so right with this objection. I just didn't want to add another inconsistency and did it the way it was done for engagements.

efficiosoft:
Hmm, you could reach 255 easily when including something like an URL, but IMHO doing so in a title should be avoided anyway. So 255 probably is a good value. Should I change it?

valentijnscholten:
Yes

efficiosoft:
Done.

efficiosoft:
I rebased this onto dev to include the fixed initial migration.

efficiosoft:
Rebased the migrations onto dev again...

efficiosoft:
There seems to be something wrong with networking between travis and github, that's where the builds have failed...

Hopefully they'll go through this time.

efficiosoft:
Thanks for merging!

PR #1080, opened 2019-04-23
Added A Password visibility toggle button To Login Page

propersam:
For good User experience sake, I find it stressful when I just copy and paste password into the login page or type blindly into the password field without knowing what was entered. Sometimes my keyboard even enters a key twice and then there's no way of cross checking what I typed in like logging in to my gmail account for example. So I decided to take my time to add the toggle password functionality to the login page for myself and made a pull request in case of someone else that might need it. :smiley:


propersam:
I did nt add that..dont even know what it's for...it came when i pulled updates before making a push.

But i have removed it now.

aaronweaver:
Like the change as well but agree with @adracea on  "eye svg"

propersam:
The form is auto generated by django. I could nt edit the input box for password to fix in the eye svg icon. But let me see if i cn change checkbox to d eye svg icon. In the same position where checkbox currently is.

propersam:
How abt now. @aaronweaver @adracea  How do you see it. I can't use button for the toggle because buttons automatically submits the form. And I can't put the SVG eye icon inside the password input box because it is auto generated by django form. That's why I wrapped the toggle icon in a clickable div element.

valentijnscholten:
Your PR still has this file removed, which I don't think is intended? `dojo/unittests/test_dependency_check_parser.py`

propersam:
The file test_dependency_check_parser.py was intentionally removed by @aaronweaver 

https://github.com/DefectDojo/django-DefectDojo/pull/1080/commits/38d351e6388b12a5d1a3da3299e1c359bf0ba52b

valentijnscholten:
Even if it was intentional I don't think it should be part of this PR.

propersam:
Oh. Well, i don't really know what to do now. I couldn't make a push until i pulled all updates from dev branch. And the above commit updates not made by me where pulled into my code before i was able to push my own update. 

Please, What do you suggest i do to make the PR the way it should be. Because i dont know if git has the functionality to control what updates come in nd which one's dont.

valentijnscholten:
Getting it right "the git way" may be a bit of challenge, can't see exactly what happened there. Can't find that commit anywhere else but in this PR.
Easiest fix would possibly be `git revert 38d351e6388b12a5d1a3da3299e1c359bf0ba52b` or recreate the file with contents from the dev branch: https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/unittests/test_dependency_check_parser.py

propersam:
I have restored the file. 
I hope @aaronweaver is okay with this for whatever reason it was deleted in the first place.

valentijnscholten:
Thanks, just needs a newline at the end of that file to make it flake8 compliant (and risk to the dev branch)

aaronweaver:
@valentijnscholten @adracea Are we ok with merging? Looks like all changes have been addressed.

PR #1082, opened 2019-04-24
fix product grading

### LINKED ISSUE #1034: Grading can never return to A
**Bug description**
The grading for a product does(can) never return to "A" once all findings are closed.

**Steps to reproduce**
1. Take a product with grading different from 'A'
2. Close all Findings
3. Grade stays stuck to latest calculate value

**Expected behavior**
If all findings are closed the grade for a product should be 'A'

- [ ] Kubernetes
- [X] Docker

**Environment information**
not relevant

I believe the issue is located in utils.py calculate_grade() . The query for "severity_values" will return nothing if (for example) all findings are in a closed(inactive) state which will therefore not trigger a recalculation of the grade (line 1606 in utils.py).

dr3dd589:
fix #1034 

I think there is no need for `if` statement on `severity_values`. because it returns `none` type when there are no active findings.

### DIFF
--- a/dojo/utils.py
+++ b/dojo/utils.py
@@ -1605,14 +1605,12 @@ def calculate_grade(product):
                 medium = severity_count['numerical_severity__count']
             elif severity_count['severity'] == "Low":
                 low = severity_count['numerical_severity__count']
-
-        if severity_values:
-            aeval = Interpreter()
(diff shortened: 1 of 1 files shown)


zfirered:
is there any fix fr this?
Because product grading doesn't show up anymore

PR #1088, opened 2019-04-25
Feature: view findings per engagement

valentijnscholten:
Based on the CICD example in https://github.com/aaronweaver/defectdojo_api and https://github.com/aaronweaver/AppSecPipeline/blob/master/tools/defectdojo/dojo_ci_cd.py I have setup our DD and Jenkins pipelines to use CICD engagements and upload a new scan every day.
Every branch has their own engagement. This works fine, except that DD by default shows findings per product. So you get findings for different engagements (branches) mixed through each other.
We have discovered we "need" to be able to view findings per engagement.

This PR is a start / work in progress to see if it can be achieved and if we can fit it into the UI in some more or less intuitive way. Curently the UI is based on products and the product tab provides menu entries to view all/open/critical/... findings per product. I would like to have something similar for engagements.

For now I have added a menu item to view all findings for an engagement to the view product page and the view engagment page.

Let me know what you think!

2019-04-25 17_26_21-Photos
2019-04-25 17_24_56-View CI_CD Engagement _ DefectDojo


aaronweaver:
@valentijnscholten Looks good and very useful, just a few flake8 items:

```
Found violations: dojo/engagement/urls.py
	W291 @ 42:73 - trailing whitespace
Found violations: dojo/engagement/views.py
	F821 @ 1017:27 - undefined name 'Engagment'
	F821 @ 1024:14 - undefined name 'ProductFindingFilter'
Found violations: dojo/finding/views.py
	E265 @ 69:5 - block comment should start with '# '
	E265 @ 579:13 - block comment should start with '# '
	E265 @ 1513:5 - block comment should start with '# '
Found violations: dojo/models.py
	E265 @ 808:5 - block comment should start with '# '
```

valentijnscholten:
Done. I also removed a view and url pattern that I previously added as I don't think it is used/needed.
I will remove the TODOs once I have done some more testing on monday.
I will probably also add a "View Open Findings" menu item below the "View All Findings" I already added.

valentijnscholten:
This is now ready for merging I believe.

adracea:
@valentijnscholten some TODO items seem to still be around , do you still need them or can they be removed ? :) Otherwise this looks great !

valentijnscholten:
I left them as help for anyone who might want to build more "per engagement" stuff. But maybe better if in remove them?

aaronweaver:
@valentijnscholten, agree with @adracea  I'd probably remove the todo comments. Looks good!

PR #1089, opened 2019-04-26
Slimmer docker images

alles-klar:
Adding a new build image to build the django and nginx image. This removes the dependency
between the django(release) image and the nginx image.

Also adding a condition which breaks the pipeline when the docker build fails.

Last discussion about this topic: #1044

PR #1096, opened 2019-04-27
Update documentation related to #1089

alles-klar:
Update the documentation on how docker images are build locally - Related to pull request #1089.

The docker-compose build does not work if someone wants to build all images locally, so I removed it.


alles-klar:
I will move the code from the build dockerfile into the two other images, because it turns out, that developing with the build image is not really comfortable. The pull request will follow soon. Documentation is not needed anymore.

Issue #1098, opened 2019-04-28
Unknown column 'dojo_finding.cve' in 'field list'


madchap:
**Bug description**
Checked out fresh copy, `dev` branch.

```
uwsgi_1         | [28/Apr/2019 12:19:30] "GET / HTTP/1.1" 302 0
uwsgi_1         | Internal Server Error: /dashboard
uwsgi_1         | Traceback (most recent call last):
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/exception.py", line 41, in inner
uwsgi_1         |     response = get_response(request)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
uwsgi_1         |     response = self._get_response(request)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 187, in _get_response
uwsgi_1         |     response = self.process_exception_by_middleware(e, request)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 185, in _get_response
uwsgi_1         |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapped_view
uwsgi_1         |     return view_func(request, *args, **kwargs)
uwsgi_1         |   File "/app/dojo/home/views.py", line 71, in dashboard
uwsgi_1         |     for finding in findings:
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 250, in __iter__
uwsgi_1         |     self._fetch_all()
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 1121, in _fetch_all
uwsgi_1         |     self._result_cache = list(self._iterable_class(self))
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 53, in __iter__
uwsgi_1         |     results = compiler.execute_sql(chunked_fetch=self.chunked_fetch)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 899, in execute_sql
uwsgi_1         |     raise original_exception
[...]

PR #1103, opened 2019-05-02
Test verifying there are no unmigrated model changes

efficiosoft:
This PR is related to #1098 and adds the test as discussed with @aaronweaver.

Testing explicitly if migrations apply cleanly is not required, as the Django test runner applies them automatically before running the tests and hence would reveal problems itself.

### DIFF
--- a/docker/entrypoint-unit-tests.sh
+++ b/docker/entrypoint-unit-tests.sh
@@ -2,6 +2,26 @@
 # Run available unittests with a simple setup
 
 cd /app
-python manage.py makemigrations dojo
-python manage.py migrate
-exec python manage.py test dojo.unittests --keepdb
+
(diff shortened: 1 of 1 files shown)


efficiosoft:
I'm running this now on travis to see if it works as intended, afterwards i'll revert the commit which added the `Product_Type.someweirdfield` field again.

efficiosoft:
@aaronweaver  Ok, ready for review now.

The travis build is going to fail, but that's because the `hash_code` field was changed some time ago and has no migration yet. This just shows the test is working.

aaronweaver:
Thanks @efficiosoft this is a much needed test. I've added the hash_code migration as well.

efficiosoft:
> Thanks @efficiosoft this is a much needed test. I've added the hash_code migration as well.

You're welcome. Thanks as well.

PR #1105, opened 2019-05-02
Removing Dockerfile.build

alles-klar:
Move content of Dockerfile.build into the two other images.

A separate build image makes development very complicated. Maybe it is better to have redundant code instead of a third image. Due to the caching mechanism of docker it makes no difference for the build process.

Related to: #1089

Activating tests for postgresql and redis.


alles-klar:
@ppiotr3k thanks for the feedback, I have added it.

ppiotr3k:
Thanks @alles-klar! I made some additional changes, for which I created a PR in your repo: https://github.com/alles-klar/django-DefectDojo/pull/1, will be easier to merge :smiley:

ppiotr3k:
@alles-klar, regarding the _ugly fix to install postgresql-client without errors_ you propose, I see no better way at the moment.

To document and explain this a bit more: without the _ugly fix_, the error is due to the way `update-alternatives` behaves in the `stretch:slim` image - which prunes manpages by default in `/etc/dpkg/dpkg.cfg.d/docker`, and on which `python:2-slim` is built; cf. https://github.com/debuerreotype/debuerreotype/issues/10/

However, the good news is that apparently this is [fixed in unstable](https://github.com/debuerreotype/debuerreotype/issues/10#issuecomment-333335162), meaning it can be removed in the future! :tada: I'll create an issue to track this down.

ppiotr3k:
@aaronweaver, thanks for the reminder! I've just tested the procedure in `DOCKER.md`, no changes. All lights look green so I'm merging the PR. Thanks @alles-klar!

For the record, to track and remove the _ugly fix_ in the future, which actually is more the currently recommended workaround, I created #1110.

PR #1112, opened 2019-05-07
Spelling mistake in "System Settings" form

Tibo-le-canard:
Hello,
I made a small change for a spelling mistake.

- [X] Unit tests are passing


aaronweaver:
Thanks for the code change however I believe this is causing an issue. The 0001_initial.py can't be changed that way. The migration will need to generate the alter statement. See the slack channel for the issue this is causing.

aaronweaver:
@efficiosoft What would be the best why to catch this in the migration test?

ppiotr3k:
You are totally right @aaronweaver, sorry about that... :disappointed: 

If I understand correctly, the PR changed the model, in which DB rows are named after class attributes automatically, therefore the ORM Mapper/Binder cannot associate the class attributes with rows defined in DB table schema.

I'm not as seasoned with Django ORM as with SQLAlchemy, but easiest quick fix for this case, if possible with Django ORM, would be to put an explicit DB row name for the renamed attribute; and in another PR migrate the table properly to correct the row name.

I see two options:

1. revert this PR; correct attribute name, have migration to correct row name in DB, and have a new PR.
1. create a new PR quick fixing this with an explicit row name, have another PR to migrate the table.

I prefer option 1.

ppiotr3k:
It is indeed possible to provide an explicit column name with `db_column`, cf. [Django documentation](https://docs.djangoproject.com/en/1.11/ref/models/fields/#db-column). However, I still prefer to revert this PR for the time being.

efficiosoft:
Hm, the migration test itself can't detect such changes because they don't create an invalid migration chain, they rather create a different one...

IMHO, the PR should definitely be reverted and a new one with proper migration be created. We somehow need to make it crystal clear that never ever one may change an existing migration file. `makemigrations` usually does the right things, if just it was used.

efficiosoft:
@aaronweaver I don't know, but maybe there's a check or similar offered by github that fails when an existing file in a directory was changed? This could then be used to lock existing migration files.

ppiotr3k:
> Hm, the migration test itself can't detect such changes because they don't create an invalid migration chain, they rather create a different one...
> 

Agreed. However, app setup/deployment should have failed, as the Django ORM Mapper/Binder is not able to associate class attributes to table columns. No?

aaronweaver:
I'm with option 1 as well for reverting.

efficiosoft:
> Agreed. However, app setup/deployment should have failed, as the Django ORM Mapper/Binder is not able to associate class attributes to table columns. No?

It is, it just now names the column differently. This causes no issue when the DB is set up from scratch, but existing DBs that - according to django's records - were already migrated up to 0004 and hence have the 0001 migration already applied don't get migrated.

efficiosoft:
The right way is to revert, make the model changes again, run

    ./manage.py makemigrations dojo

See if that correctly detects that the field has been renamed (not delete + create) and then commit the resulting migration file (0005_something.py) together with the changes. A

    ./manage.py migrate dojo

then correctly alters the schema for both new and existing DBs.

If, however, `makemigrations` doesn't detect the renaming correctly, do

    ./manage.py makemigrations --empty dojo

and add the `RenameField` operation manually (https://docs.djangoproject.com/en/1.11/ref/migration-operations/#renamefield).

efficiosoft:
@aaronweaver Should I add a `README.rst` to the `db_migrations` directory that tells everybody to not alter existing migrations? I could make it quite noisy and obvious... :)

ppiotr3k:
> 
> It is, it just now names the column differently. This causes no issue when the DB is set up from scratch, but existing DBs that - according to django's records - were already migrated up to 0004 and hence have the 0001 migration already applied don't get migrated.

For a fresh deployment indeed, wouln't be an issue.

My question wasn't precise enough: with data / test data, preexisting in DB, shouldn't the setup/deployment fail?

For me, yes, agreed, the migration wouldn't be applied because of current migration ID in the DB migration version table. However, the app should fail to load. At least this is what happens with SQLAlchemy and Alembic.

efficiosoft:
Hm, apparently it only crashes when the model is actively used, such as in the cleanup celery tasks. It doesn't check whether models and schema match on startup. Maybe because there's no real difference in terms of startup procedure between loading the WSGI app and running one of the management commands, such as `makemigrations`, which would actually need to be able to run even with a missmatching schema.

ppiotr3k:
So, without model preloading, looks like a good way to detect such issues would be be stimulating the app through more integration tests.

efficiosoft:
Yes, more tests are always good. But for this specific case, ensuring no migration files get changed at all seems to be the more direct approach. Although I've no idea how to acomplish that apart from writing a README and recognizing it when reviewing PRs. Shouldn't be too much of an issue after all.

ppiotr3k:
As we humans can always do mistakes, even with the best intentions, I prefer an automated check.
I have an idea for a Travis-CI check but let's work together on the specs :wink:
 
General idea: use Tracis-CI to list modified files in a given directory in current PR branch (`$TRAVIS_BRANCH`), and exit fail if any existing file has been modified.

Working prototype, fails if any file in `dojo/db_migrations/` is modified, OK if new ones are added:
```bash
if [ -z "$(git diff --name-only --diff-filter=M HEAD...$TRAVIS_BRANCH dojo/db_migrations/)" ]; then
    (>&2 echo "ERROR: Never change existing migrations, add new ones instead.")
    exit 1
fi
```

aaronweaver:
@ppiotr3k I like this approach. As the old saying goes let's have our belt and suspenders, or at least something like that.

efficiosoft:
> As we humans can always do mistakes, even with the best intentions, I prefer an automated check.
> I have an idea for a Travis-CI check but let's work together on the specs wink
> 
> General idea: use Tracis-CI to list modified files in a given directory in current PR branch (`$TRAVIS_BRANCH`), and exit fail if any existing file has been modified.

That's great. Thanks for working this out!

Additional idea: Maybe limit the check to `*.py` files.

PR #1115, opened 2019-05-08
Add PHP Symfony Security Check parser

valentijnscholten:
This PR adds a parser for the [PHP Symfony Security Checker](https://symfony.com/doc/current/security/security_checker.html).

The reason for this is that the PHP Scanner of the OWASP Dependency Check is in an experimental state and basically not working. The snyk scanner is a commercial scanner which is not available to everyone.

My initial tests show this works, there's one open question:

- How to the handle the fact the symfony security scanner doesn't provide a severity rating in its output?

A) Chose a default
This is what I have done for now, defaulted to 'Info'. The risk is that the vulnerabilities might be skipped if a minimum severity is specified for the import.

B) Leave empty
This would require changes to DefectDojo as the datase column is not allowed to be null and the import process checks the severity against a minimum severity.

C) Add a new severity value of 'Unknown' or 'TBD' to DefectDojo
This would require some more changes and might be difficult with metrics and maintaining a mapping to numbers to compare vulnerabilities. Not sure if we can expect more scanners to not specify a severity.

D) Allow unverified findings to have no severity rating
Even more work :)

Please let me know your thoughts.


valentijnscholten:
Also it looks this isn't working when the Finding is created. The CWE field remains empty:

`cwe=1035`

Not sure why as other parsers are doing the same thing.

valentijnscholten:
I've looked at some of the vulnerabilities in the database used by this scanner and it seems quite common to have vulnerabilities without a severity rating, some don't have a CVE as well, etc.
So ideally DefectDojo would support findings without a severity that will only get a severity after triage in DefectDojo.

aaronweaver:
@valentijnscholten Info is what I've done in the past. I think I prefer Info to blank though. Would you mind uploading a sample to [sample files](https://github.com/DefectDojo/sample-scan-files)?

valentijnscholten:
> Also it looks this isn't working when the Finding is created. The CWE field remains empty:
> 
> `cwe=1035`
> 
> Not sure why as other parsers are doing the same thing.

For the record, this works fine now.

PR #1177, opened 2019-05-24
Fix Repo field

dr3dd589:
changed Repo model to `URLField`  from `CharField`.
DB migration file added.  


valentijnscholten:
What if the existing data in this column contains data that doesn't pass the https://docs.djangoproject.com/en/2.2/ref/validators/#django.core.validators.URLValidator used by the URLField? Or are there other checks in DD which should have prevented this already?

dr3dd589:
hello, @valentijnscholten I have tested it locally and it does not effect on existing data that does not validate URLValidator rule.

Issue #1065, opened 2019-04-17
Fix Genereic CSV


devGregA:
the generic CSV format doesn't allow to upload Findings with Active=True and Verified=True: once the import is done, even by setting these columns to "True", the findings are always Inactive and Verified is unticked. Is that expected? In that case, what's the point to have these columns in the CSV format?

devGregA:
More info: https://github.com/DefectDojo/django-DefectDojo/issues/810

devGregA:
Eric, Sample file is here: https://github.com/DefectDojo/sample-scan-files/tree/master/generic

ptrovatelli:
@devGregA I think this broke the APIV1 importscan call (there is no test on this; only apiv2 import is tested). 
Found on python2 and python3 branch
python2 branch: 
```
Internal Server Error: /api/v1/importscan/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/tastypie/resources.py", line 221, in wrapper
    response = callback(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/tastypie/resources.py", line 461, in dispatch_list
    return self.dispatch('list', request, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/tastypie/resources.py", line 493, in dispatch
    response = method(request, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/tastypie/resources.py", line 1401, in post_list
    updated_bundle = self.obj_create(bundle, **self.remove_api_resource_names(kwargs))
  File "./dojo/api.py", line 1467, in obj_create
    parser = import_parser_factory(bundle.data['file'], t, bundle.data['scan_type'])
TypeError: import_parser_factory() takes at least 4 arguments (3 given)
```

python3 branch: 
```
Internal Server Error: /api/v1/importscan/
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/tastypie/resources.py", line 228, in wrapper
    response = callback(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/tastypie/resources.py", line 468, in dispatch_list
    return self.dispatch('list', request, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/tastypie/resources.py", line 500, in dispatch
    response = method(request, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/tastypie/resources.py", line 1409, in post_list
[...]

Issue #1171, opened 2019-05-23
Server Error (500) importing Fortify scan XML file


dougmcdorman:
When attempting to import a Fortify XML scan report the server gives Server Error (500) from the underlying "TypeError: cannot concatenate 'str' and 'int' objects" fault.

**Steps to reproduce**
1. Download  docker-compose_base.yml docker-compose_uwsgi-release.yml
2. docker-compose -f docker-compose_base.yml -f docker-compose_uwsgi-release.yml up
3. Attempt to upload a Fortify Scan XML file created from results.fpr using ReportGenerator 
4. 500 error (from webserver)

**Expected behavior**
Fortify scan would be imported.

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [x] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Windows 10 with Docker Toolbox and Virtualbox
 - Lastest on git release website, just downloaded the 2  docker-compose_base.yml docker-compose_uwsgi-release.yml files directly

**Sample scan files** (optional)
Will try to create redacted version.

**Console logs** (optional)
uwsgi_1         | Internal Server Error: /product/1/import_scan_results
uwsgi_1         | Traceback (most recent call last):
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/exception.py", line 41, in inner
uwsgi_1         |     response = get_response(request)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
uwsgi_1         |     response = self._get_response(request)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 187, in _get_response
uwsgi_1         |     response = self.process_exception_by_middleware(e, request)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 185, in _get_response
uwsgi_1         |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapped_view
uwsgi_1         |     return view_func(request, *args, **kwargs)
uwsgi_1         |   File "./dojo/product/views.py", line 382, in import_scan_results_prod
uwsgi_1         |     return import_scan_results(request, pid=pid)
[...]

PR #1191, opened 2019-05-29
Dev Commits #1065

devEricA:
These are my fixes for issue #1065 

Import Scan Results now reads the active and verified columns of a .csv file when the check boxes for active and verified are marked. This is done when Generic Findings Upload is selected.


devEricA:
These additional commits are for #1171.

Issue #1143, opened 2019-05-18
Inconsistent behavior to view "open findings"


madchap:
**Bug description**
When wanting to looking at open findings from within an engagement, the filter is set so that only "verified" and "active" findings are shown (which I personally find not very intuitive), e.g. the filter's default value set on "verified" is set to "Yes".

However, when directly selecting open findings from the main findings menu, it shows them all. E.g. the "verified" is set to "Unknown".

**Steps to reproduce**
1. Import a scan into an engagement, checking the findings imported to be "active" but not "verified" (I am guessing the usual?)
2. "View open findings", see 0 findings. This is only means there is no visual cue that the findings were actually imported.
3. On the other hand, from the main menu, select "Findings" > "Open Findings"
4. Notice that you see them.

**Expected behavior**
Consistent behavior is expected.

I would vote to have the "verified" filter to "unknown" and not "yes", so changing the engagement open finding behavior.

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [x] Docker
- [ ] setup.bash / legacy-setup.bash

**Additional context** (optional)
- dev branch

propersam:
this issue should be solved now.

check out my PR and point out if you detect any other open finding inconsistency in any other page

propersam:
The first PR was made to master branch by mistake.

Use the second #1207

madchap:
Thanks @propersam. I am guessing this fixes the inconsistent behavior.

Now I am guessing it is a question of semantic: do we classify an "open" finding as one that we know is a true-positive (e.g. verified -- I am guessing the original intent), or just as one that is to be checked and eventually marked as "verified" (e.g. only "active")?

From a workflow perspective, today, I know I have to reset the filter all the time to see the "open" as active only. Just so that I can get on working what I have just imported and then mark them as what they are really. Or import directly with both "active" and "verified" checked, which I am guessing would defeat the original intent.

@aaronweaver @devGregA what are your thoughts on this? Maybe just some wording changes, and/or an extra menu entry?

propersam:
Actually with the little experience I have gotten from researching/using this functionality, I discovered that When a Finding is created It uses the following work flow:
once created -> Active, verified: unkown (and this is why we had to fix the inconsistency)

when a review is requested on the finding -> Active,Under Review,  verified: no (actually in backend this is 'false')

[...]

PR #1207, opened 2019-06-03
Fix open finding filter inconsistency

propersam:
Fix issue #1143 

funny enough I only added a single word to fix this. as you can see in my commit.

If this is not what is expected do let me know.


propersam:
This task is yet to be assigned to me

PR #1209, opened 2019-06-03
Add Google/OKTA OAUTH2 Sign-in options

Maffooch:
Tutorial for establishing OAUTH clients can be found in the DefectDojo Documentation. Redirection to that page is default when attempting to login with either method without proper configuration.

- [X] Your code is flake8 compliant (DefectDojo's code isn't currently flake8 compliant, but we're trying to correct that.)


arkwrn:
I've already follow exactly the same as the instructions on the docs but instead of login using google it's redirect to https://defectdojo.readthedocs.io/en/latest/social-authentication.html#google

Maffooch:
Hi @arkwrn, I'm glad you found a solution to the issue. Do you happen to be running on the release build? Any changes made to the project can only be applied after either rebuilding everything or doing your solution. If this is not your case, please let me know.

arkwrn:
I run based on latest master (https://github.com/DefectDojo/django-DefectDojo.git)

Maffooch:
From my understanding, you are correct. One way to avoid that is to build your local repo with the modified settings.py, and then launch the images. This way, instead of pulling from the registry, it will pull from the newly created local builds. There is more on that here: [https://github.com/DefectDojo/django-DefectDojo/blob/master/DOCKER.md](https://github.com/DefectDojo/django-DefectDojo/blob/master/DOCKER.md)

Issue #1155, opened 2019-05-21
Simple search should display engagement info


madchap:
**Is your feature request related to a problem? Please describe**
When simple searching, the finding results table should include the engagement and associated test in which findings reside. It would make it easier to know where it is located without having to click on the title.

**Describe the solution you'd like**
Add an extra column(s) that display and link to finding's engagement and associated test.

devGregA:
@aaronweaver do you have any thoughts on this?

aaronweaver:
@devGregA yes, makes sense.

PR #1223, opened 2019-06-06
Enhancement #1155

devEricA:
This is my enhancement for #1155. Added columns for engagement and associated test whenever a finding is pulled up in the simple search. The engagement column lists the engagement that the finding is associated with. The associated test column displays the test that the finding is associated with. If the associated test is not titled, the associated scan is displayed instead.


madchap:
Looks good to me! Thank you.

Issue #1219, opened 2019-06-06
Making Accepting Risks easier to access for Findings


nwse-strm-suer:
**Is your feature request related to a problem? Please describe**
Vulnerability manager normally don't patch software by themself. If a finding is reported to the responsible person/team, a product owner may accept this risk of running vulnerable software. At the moment DefectDojo does not offer any solution for "risk acceptance". For a documentation and as a kind of audit trail, this feature would be helpful. The "false positive" status does not cover this correctly and other stati would generate new findings of the same vulnerability.

**Describe the solution you'd like**
An additional findings status like "accepted risk" or similar to display the correct status. This would express a not "active" status and offer a filter in the findings list. Also a state chart for the different stati and possible transitions would be helpful.

**Describe alternatives you've considered**
The current alternative is to add a note and maybe set the status to false positive to remove it from active findings. This is not really practical and does not reflect the reality.

aaronweaver:
Hi @nwse-strm-suer, did you take a look at test, then scroll down to 'Risk Acceptance'? 

![image](https://user-images.githubusercontent.com/220838/59036390-2fedb180-883d-11e9-850e-8d0481bdd0f8.png)

devGregA:
We should probably add risk acceptance off the finding menu bar, rather than just be available via the engagement interface.

devGregA:
@devEricA Please add accepting findings to the dropdown finding menu. Adding a screenshot momentarily.

devGregA:
![Screen Shot 2019-06-06 at 3 02 43 PM](https://user-images.githubusercontent.com/4741312/59062790-4bbc7c00-886c-11e9-9474-b588dc2dd477.png)
![Screen Shot 2019-06-06 at 3 03 01 PM](https://user-images.githubusercontent.com/4741312/59062791-4bbc7c00-886c-11e9-8fa5-a4515d3ea270.png)

devGregA:
Below "Close Finding" please.

valentijnscholten:
Where is the functionality currently located? I remember seeing it and testing it but cannot find it anymore anywhere currently on the dev branch.

madchap:
> Where is the functionality currently located? I remember seeing it and testing it but cannot find it anymore anywhere currently on the dev branch.

[...]

PR #1225, opened 2019-06-07
Solution for #1219

devEricA:
This is my solution for #1219. The option for accepting risks has been added under the close finding option in view finding, titled 'Add Risk Exception' (Titled in this manner as the headline on the page to accept a risk is headlined 'Risk Exception'). When 'Add Risk Exception' is clicked, the user will be directed to the page to accept the risk.


madchap:
"Risk exception" does not really sound right for me. I would rather have "Add risk acceptance", or simply maybe "Accept risk".

Issue #1075, opened 2019-04-22
IBM App Scan Integration


vivek1123:
I am currently performing DAST activities for my security team and using IBM App Scan Standard for the same. Recently I came across Defect Dojo which I think can be very effective for Defect Tracking as well as client integration. 

Bu the tool doesn't support the IBM App Scan output.

Can you help me with this new solution.

aaronweaver:
@vivek1123  There is app scan support. Is the importer not working? Please provide an error if there is one.

valentijnscholten:
Which tool should be selected as I can't see any IBM either in https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/factory.py

aaronweaver:
My bad I was thinking of AppSpider, instead of AppScan. @vivek1123 can you provide a sample export file and upload [here](https://github.com/DefectDojo/sample-scan-files)?

vivek1123:
@aaronweaver not able to upload files at ur link, requires additional push rights.

aaronweaver:
@vivek1123  you need to submit a PR.

aaronweaver:
@vivek1123 We can work on this, just attach the sample file directly to this issue.

dougmcdorman:
You might find the parsers from OWASP Benchmark useful to refer to
https://github.com/OWASP/Benchmark/tree/master/src/main/java/org/owasp/benchmark/score/parsers
AppScanDynamicReader.java
AppScanDynamicReader2.java

propersam:
I am waiting for a valid sample XML, CSV or JSON file for this issue before I can dig in.

aaronweaver:
@propersam See attached xml file.

[testfire.xml.zip](https://github.com/DefectDojo/django-DefectDojo/files/3249135/testfire.xml.zip)

propersam:
seen. You can assign this to me. and set to In Progress

propersam:
The threat class in this template are not categorized into mediun, info, low, high, etc..
..I think I need a list of all the threat class used by IBM APP Scan categorized according to severity.

propersam:
The threat I have seen sofar in this template are:

catPredicatableResourceLocation
catInformationLeakage
catcontentSpoofing
catUrlRedirectoryAbuse
catCrossSiteRequestForgery
etc..

They are not categorized according to severity level

propersam:
[...]

PR #1248, opened 2019-06-11
Import Scanner for IBM APP Scan

propersam:
Import Scanner for IBM APP scan file #1075 


tests/ibm_appscan_test.py
```diff
@@ -0,0 +1,93 @@
+from selenium import webdriver
```
Maffooch: A selenium is good, but a little slow compared to what could be done. Check out this [parser test](https://github.com/propersam/django-DefectDojo/blob/9280eb868d8eaadba4f6fd02676680393f73cf96/dojo/unittests/test_immuniweb_parser.py).

propersam: I wanted to use a parser test... but the sample file of IBM APP Scan is very large

Maffooch: Can it be reduced in size?

propersam: I don't know how possible that is...

IBMApp Scanner xml result is a really sophisticated xml file by design.

tests/ibm_appscan_test.py
```diff
@@ -0,0 +1,93 @@
...
+        Select(driver.find_element_by_id("id_scan_type")).select_by_visible_text("IBM AppScan DAST")
+        # Upload Scan result file
+        scanner_file = os.path.join(dir_path, "ibm_appscan_xml_file.xml")
+        driver.find_element_by_name("file").send_keys(scanner_file)
+        # click on upload button
+        driver.find_element_by_css_selector("input.btn.btn-primary").click()
+        # Query the site to determine if the finding has been added
+        productTxt = driver.find_element_by_tag_name("BODY").text
+        # Assert the query to determine status or failure
+        self.assertTrue(re.search(r'IBM App Scan processed, a total of 27 findings were processed', productTxt))
```
Maffooch: Need to update the name here.

tests/ibm_appscan_test.py
Maffooch: I've checked it out and you're right. There's not much that can be done about this. Great work making a parser from that mess of a file.

propersam: yea... thanks

PR #1284, opened 2019-06-20
Fixes #1249

### LINKED ISSUE #1249: Cannot batch update findings with status "mitigated"
**Bug description**
There is no way to mark findings mitigated in a batch way through the UI.

**Steps to reproduce**
1. Go to a finding report
2. Select several findings which are marked for example as "inactive", "out-of-scope"
3. Attempt to mark them as "mitigated"
4. See there is no way than to go one by one.... and cry.

**Expected behavior**
There is a UI option to allow to mark findings as "mitigated".

- [ ] Kubernetes
- [x] Docker

devEricA:
This is my fix for #1249. Added a mitigated checkbox to the bulk edit menu in the finding lists of Findings_list and view_test. When a bulk edit is submitted with the mitigated checkbox marked, the status 'Mitigated' will appear in the finding, and vise versa.


dojo/db_migrations/0009_auto_20190621_1445.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 2.2.1 on 2019-06-21 14:45
```
Maffooch: There is no need for two migrations that do the same things. Will most likely result in errors.

dojo/db_migrations/0008_finding_mitigation.py
```diff
@@ -0,0 +1,36 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='finding',
+            name='is_Mitigated',
+            field=models.BooleanField(default=False),
+        ),
+        migrations.AlterField(
+            model_name='child_rule',
```
Maffooch: All of these changes referencing rules were not changed in this PR. These changes have already been accommodated in the 0006 migration file as well. For some reason, they were not applied when that file ran. They cannot be in both places though.

PR #1296, opened 2019-06-26
Harcoding fullcalendar pkg version. Fixes #1011

### LINKED ISSUE #1011: HTTP 404 for static file fullcalendar.min.js, yarn dependency range not correct
**Bug description**
Frontend is getting HTTP 404 when trying to load _/static/fullcalendar/dist/fullcalendar.min.js_, basically because version in _package.json_ is "*", which installs recently released version 4 of the dependency, and that file is not there anymore. I solved it by specifying version "3" and installing the package again, which installs version "3.10.0" and solves the issue. I'm not submitting a PR for this, because I'm no expert on it and somebody may like to review dependencies for additional wildcards (I would say that's a bad practise)".

**Steps to reproduce**
1. Install DefectDojo
2. Open Network tab in Developer Tools in your favourite web browser
3. Open DefectDojo and observe HTTP 404 for the specified file

**Expected behavior**
HTTP 200 for the specified file

- [ ] Kubernetes
- [ ] Docker
- [X] setup.bash / **legacy-setup.bash**

**Environment information**
 - Operating System: Debian 9.8 (stretch)
 - DefectDojo version: cce0532: Merge pull request #856 from devGregA/master [2019-01-28 10:29:22 -0600]

dougmorato:
Since version 3.10.0, there are breaking changes in the fullcalendar
yarn package version, which breaks the calendar functionality on Dojo.

Jquery support has been removed on version 4.0 as seen on link:
https://fullcalendar.io/blog/2018/04/alpha-release-jquery-removal

Jquery version mismatches happen if we use version 3.10.0

Thus, hardcoding to version 3.9.0

### DIFF
--- a/components/package.json
+++ b/components/package.json
@@ -15,7 +15,7 @@
     "@yarn_components/flot.tooltip": "krzysu/flot.tooltip#~0.8.4",
     "@yarn_components/font-awesome": "FortAwesome/Font-Awesome#4.4.0",
     "@yarn_components/fontawesome": "FortAwesome/Font-Awesome#~4.5.0",
-    "@yarn_components/fullcalendar": "fullcalendar/fullcalendar#*",
+    "@yarn_components/fullcalendar": "fullcalendar/fullcalendar#3.9.0",
     "@yarn_components/google-code-prettify": "tcollard/google-code-prettify#~1.0.4",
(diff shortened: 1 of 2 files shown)


johnmckinght:
Guys, any plans to backport this to `1.5.3.1`?

PR #1328, opened 2019-06-28
Use print() function in both Python 2 and Python 3

cclauss:
**Note: DefectDojo is now on Python3 and Django 2.2. Please submit your pull requests to the 'python3_dev' branch as the 'dev' branch is only for bug fixes. Any PR's submitted to the 'dev' branch should also be converted to Python3 and submited to the 'python3_dev' branch. 

### DIFF
--- a/dojo/api.py
+++ b/dojo/api.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 # see tastypie documentation at http://django-tastypie.readthedocs.org/en
 from django.core.exceptions import ImproperlyConfigured, ValidationError
 from django.core.urlresolvers import resolve, get_script_prefix
@@ -1457,8 +1458,8 @@ class ImportScanResource(MultipartResource, Resource):
         try:
             t.full_clean()
         except ValidationError as e:
-            print "Error Validating Test Object"
-            print e
+            print("Error Validating Test Object")
+            print(e)
 
         t.save()
         t.tags = bundle.data['tags']
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 # #  engagements
 import logging
 import os
@@ -518,8 +519,8 @@ def import_scan_results(request, eid=None, pid=None):
 
             try:
                 for item in parser.items:
-                    print "item blowup"
-                    print item
+                    print("item blowup")
+                    print(item)
                     sev = item.severity
                     if sev == 'Information' or sev == 'Informational':
                         sev = 'Info'
--- a/dojo/management/commands/run_scan.py
+++ b/dojo/management/commands/run_scan.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import Queue
 from datetime import datetime
 import sys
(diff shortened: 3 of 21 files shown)


dojo/management/commands/run_scan.py
```diff
@@ -99,10 +100,10 @@ def runIPScan(host, service_dict):
             try:
                 nm = PortScannerAsync()
             except PortScannerError:
-                print('Nmap not found', sys.exc_info()[0])
+                print(('Nmap not found', sys.exc_info()[0]))
```
eamanu: I think that here is print a tuple

dojo/management/commands/run_scan.py
```diff
@@ -99,10 +100,10 @@ def runIPScan(host, service_dict):
...
                 sys.exit(0)
             except:
-                print("Unexpected error:", sys.exc_info()[0])
+                print(("Unexpected error:", sys.exc_info()[0]))
```
eamanu: here the same

tests/zap.py
```diff
@@ -100,11 +101,11 @@ class Main:
...
         summary.add_row(["Medium", medium])
         summary.add_row(["High", high])
-        print summary
+        print(summary)
 
         for url in sort_by_url:
-            print
-            print url
+            print()
+            print(url)
```
eamanu: what about ```print("\n" + url)``` and avoid the first  print?

dojo/management/commands/run_scan.py
cclauss: You are correct.  Lines 204 and 212 contain legacy __print__ statements so we must assume that this file has not been ported to Python 3.  In Python 2, this line (without the import print_function) would print-a-tuple so now when we add the import print_function, we must ensure that it prints-a-tuple on both Python 2 and on Python 3.

eamanu: Oh yes, I was confused. Sorry for the noise. 

IMO this print should be just a str not a tuple. But well, that is a core decision

dojo/api.py
```diff
@@ -1457,8 +1458,8 @@ def obj_create(self, bundle, **kwargs):
         try:
             t.full_clean()
         except ValidationError as e:
-            print "Error Validating Test Object"
-            print e
+            print("Error Validating Test Object")
```
eamanu: What about?

```suggestion
            print("Error Validating Test Object\n ".format(e))
```

dojo/engagement/views.py
```diff
@@ -518,8 +519,8 @@ def import_scan_results(request, eid=None, pid=None):
 
             try:
                 for item in parser.items:
-                    print "item blowup"
-                    print item
+                    print("item blowup")
```
eamanu: what about?

```suggestion
                    print("item blowup\n{}".format(item))
```

dojo/api.py
cclauss: That would halt the program.  The goal of this PR is not to optimize or improve the code.  We want to make safe changes to get it closer to Python 3 compatibility.

devGregA:
Hi @cclauss, thank you for your PR! We have already converted Dojo to python 3.7 in the dev branch. The current master release is the second to last Python 2.7 release that will exist for Dojo. We have another legacy branch that will be the last 2.7 release once testing is completed on it. We don't accept PRs to the master brach only to dev per our PR template. Hope that all makes sense!

PR #1350, opened 2019-07-05
API ImportScan Fix

devEricA:
This is a fix for the API ImportScan. Added True for the active and verified of the import parser factory calls. The two checks correspond to whether or not the user trusts the 'Active' and 'Verified' columns of a generic findings import file.

### DIFF
--- a/dojo/api.py
+++ b/dojo/api.py
@@ -1463,7 +1463,8 @@ class ImportScanResource(MultipartResource, Resource):
         t.tags = bundle.data['tags']
 
         try:
-            parser = import_parser_factory(bundle.data['file'], t, True, True, bundle.data['scan_type'])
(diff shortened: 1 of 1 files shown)


devGregA:
@devEricA I don't see any record of this bug. Would you please provide the issue number? Also won't this set it to true every time rather than taking user input?

devEricA:
It's supposed to be an extended fix for #1065. @ptrovatelli commented on that issue and noted that the import scan results was broken for the API. @aaronweaver requested that I investigate and check out what was going on, and I noticed that the signatures in the API were not properly updated. 

Good point on the things being set to true though, I will go ahead and make the modifications to allow user input.

ptrovatelli:
Thanks that fixed it in dev branch. shouldn't we port this to master and legacy-python2.7?
@devGregA

Issue #1149, opened 2019-05-20
Unify permissions and permission checks throughout dojo


efficiosoft:
**Is your feature request related to a problem? Please describe**
Concerns have already been brought up in https://github.com/DefectDojo/django-DefectDojo/issues/1048, where @valentijnscholten  summarizes the problems of the existing permission and role handling.

**Describe the solution you'd like**
As I'm currently developing an entirely new report builder component for dojo, I thought of starting to follow a different approach on how to handle permissions.

There exists a small and well-matured framework for object-level permission checks [rules](https://github.com/dfunckt/django-rules), which makes declaring permissions for objects of a model very straightforward and expressive. It integrates well with django through an authentication backend, which allows checking permissions via the regular ``User.has_perm()`` method and some decorators and view mixins for convenience.

Permissions will have to be declared **centrally** and only once per model using predicates, like so:

    import rules

    @rules.predicate
    def is_authorized_for_product(user, product):
        return product.authorized_users.all().exists(user)

    rules.add_perm("dojo.add_product", rules.is_staff)
    rules.add_perm("dojo.change_product", rules.is_staff | is_authorized_for_product)
    rules.add_perm("dojo.delete_product", rules.is_staff | is_authorized_for_product)
    rules.add_perm("dojo.filter_product", rules.is_authenticated)
    rules.add_perm("dojo.view_product", rules.is_staff | is_authorized_for_product)

It'll be crystal clear how will be allowed to perform a specific action.

Django has become smarter over the last years as well, hence I'm developing my new component with class-based views where appropriate. I implemented a view mixin which automatically requires a user to have permission to do the action the view it was mixed in performs and raise a 404 to hide the object otherwise. Views for filtering and deletion could then look as simple as:

    from django.views.generic.edit import DeleteView
    from django_filters.views import FilterView

    from dojo.permissions import DojoUserPermissionMixin
    from dojo.filters import ProductFilter
    from dojo.models import Product

    class ProductFilterView(DojoUserPermissionMixin, FilterView):
        filterset_class = ProductFilter

    class ProductDeleteView(DojoUserPermissionMixin, DeleteView):
        model = Product

Everything else is taken care of automatically. No more manual object retrieval, explicit permission checking, context construction and boiler-plate form declarations. While this isn't strictly related to reworking the permission checking, it shows how nice it plays with class-based views.

A ``@require_permission("dojo.delete_product")`` decorator for legacy function-based views is provided as well.

[...]

Issue #1048, opened 2019-04-12
User roles / permission unclarities / suggestions


valentijnscholten:
Hi,

I am struggling to understand the permission model / user types of Defect Dojo (running dev branch, using ldap).

I have seen some github issues that try to explain it, but a later github issue describing that the authorized user part was redone around March 2018.
Antoher issue refers to the documentation which should explain the roles, but for me they don't :-)
I also get conflicting report from other DD users, for them the superuser seems to work like my staff users and vice versa.

**Staff user**
```
Designates whether the user can log into this site.
```

This seems to mean that this role allows the user to login to the Django Admin site at '/admin'?
But it also seems that you need this permission to most of the admin tasks in DD itself?
Like adding products, product types, users, engagements, etc?

Questions:
- What was the intention for this role, to assign it to security engineers who can manage all products/engagements/findings? Or to DD administrators who can do all that, plus administer DD?

**Super user**
```
Designates that this user has all permissions without explicitly assigning them.
```
Questions:
- What does "all permissions" mean? 
	- Is it that the are supposed to get access to all products/engagements/findings? (Currently not working like that)
	- Is it that is means they have permissions to manage DD, so add products, tools, edit configuration?
	- Is there a reason why superusers can only see products for which they are authorized?

**Product based role: Authorized Users**

```
List of users who are allowed to view and interact with the product.
```

When I authorize a standard user for a product, that user can see the products/engagement/findings, but cannot change anything and cannot push findings to JIRA.

Questions:
- Is there a reason why authorized users can not see accepted findings or closed findings?

**Django permissions:**
This seems to be implemented in some places, but not in other places. There's a list of permissions in the /admin portal, and group management functionality, but I don't think these are used?
In the code I only see statements around is_staff_user and is_superuser.

**My usecase and suggestions**

I am a strong believer of having the development team or at least the technical leads manage their findings, push them to jira, mark them as false positives or accept the findings.
I would like to have the following two roles, but I am not sure if they can be mapped to one of the roles above or new roles shoudl be added:

- For me an "authorized user" would mean they can manage engagements/findings, push to jira, accept findings, edit findings (for authorized products).
- For me a "superviewer" role would allow compliance people to view findings, statistics, metrics of all products/engagements/findings. (like #882)
- For me a "superengineer" role would have "authorized user" style permissions for all products without explicitly assigning them.
[...]

PR #1362, opened 2019-07-10
New permission handling and ReportNG report builder

efficiosoft:
This is the PR corresponding to #1149. Please see the issue for a detailled explanation.

- [X] If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR. (https://github.com/DefectDojo/Documentation/pull/60)
- [X] Model changes must include the necessary migrations in the dojo/dd_migrations folder.


alles-klar:
I think the idea is very nice but we should verify if this approach also works if we integrate more roles, for example @valentijnscholten suggestions #1048. For me the few rolls are one of the main drawbacks in DefectDojo.

efficiosoft:
> I think the idea is very nice but we should verify if this approach also works if we integrate more roles, for example @valentijnscholten suggestions #1048. For me the few rolls are one of the main drawbacks in DefectDojo.

I don't see a reason it won't work for more roles. You can define custom predicates for permission checking (as done in `reportng.models.is_report_requester` and their evaluation is very cheap. Of course, we have to define what roles exist first. That's outside the scope of this implementation.

efficiosoft:
The extended `Q` object implementation and `QProvider` have now been generalized and I plan to [release it as a library](https://github.com/efficiosoft/QProvider), unless it's accepted by upstream Django of course. Anyway, this makes the additions to Dojo even smaller and the API simpler.

alles-klar:
Is this version able to run? When I start it with docker-compose I get some errors in the log and a server error 500 when I visit defectdojo in the browser.

For example:
```
initializer_1   |   Applying dojo.0010_auto_20190711_0940...Traceback (most recent call last):
...
initializer_1   |   File "/usr/local/lib/python3.7/site-packages/django/db/models/fields/related.py", line 604, in resolve_related_fields
initializer_1   |     raise ValueError('Related model %r cannot be resolved' % self.remote_field.model)
initializer_1   | ValueError: Related model 'dojo.User' cannot be resolved
```

and 

```
uwsgi_1         | Internal Server Error: /
uwsgi_1         |   File "/usr/local/lib/python3.7/site-packages/django/contrib/sessions/backends/base.py", line 189, in _get_session
uwsgi_1         |     return self._session_cache
uwsgi_1         | AttributeError: 'SessionStore' object has no attribute '_session_cache'
...
uwsgi_1         | MySQLdb._exceptions.ProgrammingError: (1146, "Table 'defectdojo.django_session' doesn't exist")
```

efficiosoft:
Ah, there were some changes for which migrations weren't right yet, but now it should run just fine.

efficiosoft:
This has now been rebased onto latest dev and thus also uses the newer base image.

efficiosoft:
I'm currently trying to get generic parts of the model permission registration and view mixins upstream into the rules library, which will move another ~100 lines out of Dojo.

dr3dd589:
hey, @efficiosoft great work. Here is one more point regarding permission. it would be great if you can add this or suggest to us how we can apply this:-
let say I'm a user with no staff permission (only active member). I am authorized for only one product.
--> currently matrix section implemented for same for all. it shows the same data to admin, staff and user. what I thinking is it should only show matrix for products that are allowed to me.
--> same as in `metrics/product/type` it should show the only information of findings which are allowed to me. Currently, it is showing all the findings.

efficiosoft:
Hi @dr3dd589 

Sure, that's going to be super convenient.

1. I already converted Product, Engagement, Test and Finding to the new `DojoModel` and `DojoModelManager`, which gives each of these models a `Model.objects.for_user(user)` method that builds a base queryset with only those objects the user has read access to. The goal is to move all models, or at least all models for which views exist to this pattern.
2. All class-based views having `DojoPermissionViewMixin` mixed in automatically use that per-user restricted base queryset and hence don't even see the objects to which the user has no read access.
3. When it's needed to get a user-restricted base queryset somewhere else, e.g. in a model-specific method which computes metrics, that method can simply use `Model.objects.for_user(user)` or `Model.objects.for_user.as_q(user)` to get such a restricted queryset.

dr3dd589:
thank @efficiosoft  Any idea when you are ok to merge this so that I can implement this as you suggested.

efficiosoft:
I want to see what happens to https://github.com/dfunckt/django-rules/pull/99 and https://github.com/dfunckt/django-rules/pull/100 first, because that will reduce the amount of code to be added to Dojo.

Besides that, the ReportNG component still requires some testing and I can't give a timeline for it yet. Expect some weeks please.

efficiosoft:
@dr3dd589 However, you can look at the views in `dojo.reportng.views` to get an idea of how to implement these kind of class-based views beforehand.

dr3dd589:
Sure, @efficiosoft  I'll see this. thanks!

efficiosoft:
Just want to give an update on the process of this PR.

1. django-flexquery, which will do the user-based queryset filtering for us, now has 100% test coverage.
2. I added tests for the generic permission checking parts that'll go upstream into django-rules soon.

The good news about this is that we don't have to test those generic parts ourselves. Once 2. happened I'll be able to rebase the PR and get it ready for merging.

efficiosoft:
Hi @devGregA, @valentijnscholten, @dr3dd589, @alles-klar and others who showed interest in this PR,

All additions to third-party libraries are now merged and released and I rebased the PR for merging finally.

Best regards
Robert

efficiosoft:
And, of course, just ask if you face any issue while migrating a view.

efficiosoft:
@Maffooch Thanks for the feedback.

As you have read in the docs, ReportNG is just the framework the TeX builder was built ontop of, so I didn't enable a builder by default. However, enabling it with the default configuration is quite simple with an environment variable like this:

    DD_REPORTNG_BUILDERS="['dojo.reportng.builders.tex.TeXReportBuilder']"

This is also denoted in the docs as an example.

Given that the TeX builder isn't of any use without at least one report template, and an admin has to read the docs anyway in order to write templates properly, I don't want to enable something which is going to require manual configuration by default.

Do you have concrete suggestions how to improve the current behavior?

Maffooch:
@efficiosoft Thats the largest issue I have with this. The user has two choices: Configure something basic based on what they find in the docs, or configure something that fits their needs exactly. One way or another, configuring is going to take place. Like you said, reporting is useless without at least one template. Maybe a possible solution would be to add an example page in the docs where a user could copy/paste without anything else? Another one would be to make a bare bones config by default. If it doesn't fit the users needs, they can spend the time reading the docs, understanding the TeX framework, and playing around with thier configs to get something they like. Not all users will want to take that time to do that. I don't want to see all your hard work wasted.

efficiosoft:
> @efficiosoft Thats the largest issue I have with this. The user has two choices: Configure something basic based on what they find in the docs, or configure something that fits their needs exactly.

The docs explain all available options, not just basic things.

> One way or another, configuring is going to take place. Like you said, reporting is useless without at least one template.

Doesn't exactly this speak against enabling the TeX builder by default?

> Maybe a possible solution would be to add an example page in the docs where a user could copy/paste without anything else?

Copy/paste what? The value to set for `DD_REPORTNG_BUILDERS`? This already exists.

> Another one would be to make a bare bones config by default. If it doesn't fit the users needs, they can spend the time reading the docs, understanding the TeX framework, and playing around with thier configs to get something they like. Not all users will want to take that time to do that. I don't want to see all your hard work wasted.

They get a barebone config with just a single line. But that doesn't free them from learning how to write a template, which is what the whole Tex builder docs are about.

Maffooch:
It seems I still don't understand how this is should be working then. I suppose I'm off to read the docs again and try to figure it out.

efficiosoft:
If all you want is provide a way to quickly create a report, no matter how that looks... then maybe just create a static, unconfigurable asciidoc builder for ReportNg and enable that by default? I don't see a use case for using the TeX builder without an own template.

Maffooch:
Just something for the uninitiated to start generating reports out of the gate. I mean, after thirty minutes in the docs yesterday, I am evidently still missing the point here. It has got to be dummy proof.

efficiosoft:
It's all about how you're going to use the created reports. If they should simply serve as a kind of printable dump of the findings, then maybe some kind of default template is enough for you and can be included with the docs. However, I'm afraid I can't create one that looks good.

If however, the reports are something you want to hand out to customers you tested some product for, then you surely want (or even are forced to use) a highly-customized template, which is where the power of TeX really pays out.

efficiosoft:
Ah, and even with a sample template, you still need to either install texlive together with dojo (quite large with a useful set of packages) or compile the report locally.

Maffooch:
I understand that. I doesn't have to be something you'd hand to your boss. Anything would work really. Just something that enables the report builder to be at least usable. I thought something installed incorrectly when there was no option to create a new report. Little things like that gave me the impression this was a WIP instead of a completed project.

efficiosoft:
Well, we can of course enable the TeX builder by default, since it's the only implementation at this time. But you'll still have to fetch the sample template from some source, let's say the docs, and copy it into place. But then there's the question: Which of the many, many fields of the models (product, engagement, test, finding) should be included in that template and how should they be displayed?

efficiosoft:
Such a thing can easily reach ~1000 lines.

Maffooch:
So something nice like that would take ~1000 lines? A user has to take the time to learn how to do this and then write a small novel to get it looking presentable? This is exactly why It think a dinky default is necessary. Maybe the default can be a simplified version of the docs? It would show a POC of the UI working and the other integrated features. It could be the selling point for whether the user is willing to create their custom template.

Maffooch:
If I am the only one with these concerns, then they can be put to bed. No point in implementing something that big to appease one person.

efficiosoft:
> So something nice like that would take ~1000 lines?

If you want to include all the different fields of tests and findings, generate conditional output for booleans, display endpoints properly etc... remember, this is TeX ;)

> Maybe the default can be a simplified version of the docs? It would show a POC of the UI working and the other integrated features. It could be the selling point for whether the user is willing to create their custom template.

I don't like duplicating content because of maintainability reasons, but a POC should be doable, although nobody will use it for something real.

> If I am the only one with these concerns, then they can be put to bed. No point in implementing something that big to appease one person.

I'm sure you're not, but then there is again the question: Who will use such a sample template?

A pretty short POC template is already included in the docs, however it doesn't generate meaningful content, just demonstrates how to output the different fields.

Maffooch:
I'll check out the POC template in the docs and get back to you. My gripes may have been in vain as I was/am too dense to realize they are already there!

efficiosoft:
No problem ;)

IIRC, that one just outputs finding titles for demonstration purposes... but it works as a starting point for something usable.

I could maybe add two more fields, such as description and mitigation to demonstrate how to convert from the markdown fields to TeX.

valentijnscholten:
I'm mainly here for the permission refactoring (could it be a separate PR?), but it sounds like the reporting stuff needs some (basic) default report otherwise nobody will use it. And reporting is, as you said, an important feature for a large set of usecases.

efficiosoft:
@valentijnscholten Separating the PR would be possible, but be quite involved because ReportNG uses the permission system and there are the template snippets and filterset which sit somewhere in between.

@Maffooch and @valentijnscholten I extended the sample template in the docs to demonstrate more of the concepts, but it really is a POC and nothing for real world usage for the reasons I already described above. It really depends on what you want to use it for and I expect people to write their custom ones for different occasions and report types.

Maybe we can link to one (or more) community-provided template in the future?

Maffooch:
@efficiosoft I have my dunce cap waiting on standby. I cannot for the life of me figure out how to get a template loaded. I made one based off the sample one you made in the docs. Its located in media/reporting/tex_templates. I also used the sample Reporting Builders config you provided as well that points to that location. What am I doing wrong here?

efficiosoft:
reporting -> reportng ;)

efficiosoft:
I'll change the template loader to try to create the directory if it doesn't exist, which will make it harder to run into these kind of slips.

efficiosoft:
And please use the latest revision, because I implemented a fix in there today.

efficiosoft:
@Maffooch TeX builder is now enabled by default and the template loader will create the configured template directories if they don't exist.

Maffooch:
Finally got a report to build after you set some defaults! Got a .tex report generated and then converted afterwards, Could not get the pdf to generate due to a missing lualatex module. This going to be super sweet in the right hands (I don't know latex enough to make something cool). Fabulous work.

efficiosoft:
Yeah, as described in the docs, the texlive suite isn't included in the docker image, because a full texlive has ~5Gb, which I think isn't justifiable to ship by default. That's why there is the option to only get the source code as well.

Maybe we can create a repo to which community members can contribute their report templates for public use. Seems more appropriate to me than assuming a fixed default for everybody.

Maffooch:
A report repo is an excellent idea! The default should stay as well as a tangible example of it working. It would also make it easier for users to find the correct place to put new templates if they don't read the docs, or do a poor job of doing so (as I did). Would be simple enough to just replace the default with something from the repo or a handmade template.

efficiosoft:
You still have to place the sample template into the tex_templates directory yourself. Just the empty directory now gets created for you.

efficiosoft:
I'm not going to rebase and redo the migrations for this PR again and again until it has been agreed to merge it, so please ignore the CI failures for now.

Maffooch:
Im cool with reporting now with a default. @valentijnscholten I think the new permission system would be your forte.

efficiosoft:
Just to demonstrate how easy it is, I added another builder implementation that can generate HTML reports in ~100 lines. If one wants to, he could port the existing HTML template of the old report builder over.

Finding image handling in reports has also been reworked to support the new image captions and more fields that might be added in the future.

Give it a try if you like.

efficiosoft:
... and docs for the HTML builder are now added as well, including the necessary markdown rendering filter.

efficiosoft:
Any new objections on this? If you would like to merge it anytime soon, I'd go and resolve the merge conflicts, but I don't want to take the effort of rebasing again and again.

Best regards
Robert

valentijnscholten:
The reason why I haven't looked at this in detail is because it's a massive PR with 76 files changed and seems to implement 2 things in one PR: permission changes and reporting features. Is there any chance it can be split up? I'm very interested in the permission changes, but don't have enough knowledge about the reporting part and Django UI stuff in general.

Maffooch:
It is huge but would likely take a ton of work to split into two. I had tested the reporting and made a fuss about it a few months ago. It's all good and ready to roll. The new permissions system is something I think you'd be more familiar with and could suggest some valuable feedback.

efficiosoft:
@valentijnscholten The ReportNG component depends on the new permission handling, and since I developed all this during an internship as part of my computer science program at university I couldn't wait to get the permission stuff merged before building ReportNG.

However, both parts of the PR can be distinguished solely by file/directory. The ReportNG work completely lies inside the `dojo/reportng` and `dojo/templates/dojo/reportng` directories. Everything else is the permission handling and acompanying template snippets to make use of it.

madchap:
So it seems that the permissions part would be the one requiring attention. It'd be sad to lose all this work :-(

If @efficiosoft were to rebase one more time, who volunteers to take this through a deep dive?

valentijnscholten:
I am planning on taking a look in february.

valentijnscholten:
I have looked at this, and:
- A large part is indeed the reporting part. I haven't looked at that
- The other part is around permissions/queries/simplification. It all looks OK to me, but to be honest I don't know enough about python/django/django-orm to easily approve this. There are also some things I don't fully understand, such as the Engagement class using the Product manager to return objects? Same for Test class using the Engagement manager, etc. 
I think it might be better to cherry-pick the things we think are useful and implement them in their own PR(s).
- I need at least one more person/mod to work with me on this.

efficiosoft:
Hi @valentijnscholten 

> I have looked at this, and:
> 
>     * A large part is indeed the reporting part. I haven't looked at that

Right, @Maffooch looked at the reporting part some months ago already and deemed it ok, I made no changes since then.

>     * The other part is around permissions/queries/simplification. It all looks OK to me, but to be honest I don't know enough about python/django/django-orm to easily approve this.

Let me try to explain it to you in more detail. The main problems we have so far with permission handling in dojo are massive redundancy and widely distributed enforcing of role permissions all over the different views. My PR addresses these two things:

1. You declare **once** who should be able to view, change, delete or add a specific object right in its model definition. This is handled by the rules framework (which I contributed the model declaration part to).

2. You tell which action a particular view wants to perform on what object. Now, the rules framework checks whether the current user has permission to do so, just by looking at the predicates set inside the model definition. In case you're using django's class-based CRUD view types, all of this checking is even done transparently (also contributed this to rules).

So after all models have been annotated with permission predicates and all views were converted to class-based ones (which can be a bit-by-bit process), the views no longer have to be aware of permissions and roles at all, which finally solves the two aforementioned problems.

> There are also some things I don't fully understand, such as the Engagement class using the Product manager to return objects? Same for Test class using the Engagement manager, etc.

As I said, permission declarations are baked into the model definitions directly. Now, permission checking using these declarations can reason whether a user may perform an action on **a specific** object. In django, however, a queryset containing all objects of a model is generated by a model manager, not the `Model` class itself (usually using the `.all()` or `.filter()` etc. methods). To fully isolate objects a user should not even know of, the new permission system needs a way to restrict the base queryset all operations of a user operate on. For that purpose, I extended model managers so that they can provide all views/filtersets/API viewsets etc. with such a user-specific base queryset. In case of `Engagement` and `Test`, for example, we see that the base queryset of engagements depends on the products that are in the same user's base queryset (aka the products a user should be able to see). The tests a user should have access to depend on the engagements he has access to, the findings depend on the tests and so on. This is what the custom managers do, delegate the checks. And to make all of this efficient with a single db query and no joins and sub-queries, I created and used the [django-flexquery library](https://github.com/efficiosoft/django-flexquery) for this purpose.

That way, the question "Has a user access to finding X?" can be answered by"Yes, if he has access to the test finding X belongs to." instead of hard-coding the check if the user is in a product's `authorized_users` list everywhere all over the place.

>       I think it might be better to cherry-pick the things we think are useful and implement them in their own PR(s).

The permission part is quite dense (it hasn't lots of lines anyway), and IMHO you won't be able to break some parts out of it without breaking the whole idea, too.

Just ask if you've got more questions about the implementation, I'll be happy to help.

Best regards
Robert

valentijnscholten:
Thanks for the info. One reason also for suggesting to cherrypick the permission/role part is the conflicts we have against current dev. We did a bunch of optimizations there around querysets. 
To get this PR merged I think we would need:
- me + 1 (or more) persons looking at the permission/role code
- 2 or more other persons looking at the report code
- you doing one more rebase on / merge with dev so we can take it from there.
Tonight we have a moderator meeting. It's not on the agenda, but let me see if I can recruit some volunteers.

efficiosoft:
> Thanks for the info. One reason also for suggesting to cherrypick the permission/role part is the conflicts we have against current dev. We did a bunch of optimizations there around querysets.

Actually, the conflicts all regard the permission handling part, ReportNG is mostly self-contained and doesn't cause conflicts. Even the existing conflicts can easily be resolved and shouldn't cause any interference with the recent changes I think.

> To get this PR merged I think we would need:
> 
>     * me + 1 (or more) persons looking at the permission/role code
> 
>     * 2 or more other persons looking at the report code
> 
>     * you doing one more rebase on / merge with dev so we can take it from there.
> 
> 
> Tonight we have a moderator meeting. It's not on the agenda, but let me see if I can recruit some volunteers.

I'm quite busy at the moment and won't have time for any rebasing before March 18th. But then, it should be doable.

efficiosoft:
Hi,

As Promised, here is the rebase onto latest dev.

Best regards
Robert

valentijnscholten:
Thanks, I tried to push a commit to fix the flake8 errors, but permission denied.

valentijnscholten:
Did a quicktest, but everything regarding product seems to raise an error (`/product` and `/product/1`)

```
File "/c/Data/django-DefectDojo/dojo/product/views.py" in product
  56.     prod_filter = ProductFilter(request.GET, queryset=prods, user=request.user)

File "/c/Data/django-DefectDojo/dojo/filters.py" in __init__
  341.         super(DojoFilter, self).__init__(*args, **kwargs)

Exception Type: TypeError at /product
Exception Value: __init__() got an unexpected keyword argument 'user'
```

valentijnscholten:
Also took a look at the reporting part. I like the idea, but it is extremely slow for some reason. Page load times sometimes over 60 seconds. Lots of database queries, I think prefetching is needed there.

efficiosoft:
Fixed the errors regarding `/product`, something got messed up there during rebasing.

> Also took a look at the reporting part. I like the idea, but it is extremely slow for some reason. Page load times sometimes over 60 seconds. Lots of database queries, I think prefetching is needed there.

For me a page load in the builder takes ~200ms, though admittedly I don't have a huge dataset. It get's significantly better in production mode, because then, templates only need to be compiled once and are then cached.

It'd be interesting to know whether the evaluation of queries needed for working out all the available filtering choices takes most time or those for rendering the result lists. I'll add a `select_related()` tomorrow and see how that goes.

> The 'select all' tickboxes do not seem to work, or maybe they are very slow because they trigger some request to the server.

They do work, but yes, they trigger a request. And that could be the bottleneck actually, because they count database objects filtered by a list of primary keys using the SQL IN operator and then send all the PKs back as hidden HTML form fields. I think it could become problematic when selecting around 10000+ findings at once to be included in a single report... Who does so btw? What magnitude of number of products, engagements, tests and findings do you have in your db and how many did you select?

> It also says I need to select a template, but the dropdown is empty.
> OK just read above about the templates. As I said I'm not really looking for reporting, but I think it could come in handy someday. But having a reporting feature without some template out of the box doesn't make sense to me.

The documentation tells you how to set up a template and provides a minimalist example for demonstration purposes. But yeah, I guess somebody will write and share a template that could be included by default some day. It depends on what you want to use the reports for after all.

efficiosoft:
> Thanks, I tried to push a commit to fix the flake8 errors, but permission denied.

Hmm, I checked "Allow edits from maintainers", don't know if there's anything else to activate.

valentijnscholten:
```
Found violations: dojo/filters.py
	F811 @ 575:1 - redefinition of unused 'EngagementFilter' from line 547
Found violations: dojo/models.py
	F811 @ 408:5 - redefinition of unused 'Meta' from line 394
Found violations: dojo/reportng/builders/html/__init__.py
	F401 @ 6:1 - '..base.ReportGenerationError' imported but unused
```

valentijnscholten:
still couple of flake8 error so can't merge it locally

valentijnscholten:
I have looked at the permission part and so far my understanding:
- API v2 automatically user the restricted queryset because of the mixins
- Our legacy (View based) codebase need to be updated to user the for_user() explicitly (intead of filter("test__engagement__product__authorizedusers_in", ...) kinda stuff

This all makes sense and we can use that. What I do not understand fully is how the permission system will restrict access in the same way. Or maybe it doesn't? For example how to I prevent users from deleting findings for products they are not authorized for? Or is the idea that if we use `for_user` everywhere, it will never return findings the user is not authorized for?

efficiosoft:
First, you need to switch `Finding` to inherit from `dojo.models_base.DojoModel` and declare appropriate predicates for object-level permissions in its `Meta` using the `rules_permissions` setting. These will describe whether a user will be allowed to add/change/delete a particular finding. `dojo.models_base.DojoModel.preprocess_rules_permissions()` sets some defaults. In the case of findings, you probably will set:

    rules_permissions = {"add": rules.is_authenticated, "change": rules.always_true}

This is safe since the base queryset will only contain findings of products the user is authorized for anyway, and the user should be able to change/delete all of these findings.

`dojo.reportng.models.ReportNG` shows how object-level permissions are declared. It's pretty simple and self-contained actually.

In legacy function-based views, you'll then need to use:

* `Finding.objects.for_user(request)` to get the queryset of all findings to be listed to the user. That's defined by the custom model manager attached to the `Finding` model and works straight away.
* Before deleting a finding, check the result of `request.user.has_perm(Finding.get_perm("delete"), concrete_finding_obj)` to check whether the user may delete the particular finding.

Note, the view functions don't need to know that there's a relation between a user's permission to delete a finding and whether he's in the product's autorized users list.

All these checks would be handled automatically when using django's native class-based views instead of plain functions, at least for CRUD actions. See `dojo/reportng/views.py` to get a sense of how clean that would look, especially `ReportNGDeleteView`, `ReportNGDownloadView`, `ReportNGDetailView` and `ReportNGFilterView`, all of which have automated permission checking via the mixin and don't even need to do db lookups themselves.

The API viewsets will need the `dojo.mixins.DojoPermissionAPIViewSetMixin` mixed in for the automatic permission checking and queryset restriction to work.

Honestly, would it help you any if we had a talk together? Writing all of this down turns out to be quite time-consuming.

efficiosoft:
> This all makes sense and we can use that. What I do not understand fully is how the permission system will restrict access in the same way. Or maybe it doesn't? For example how to I prevent users from deleting findings for products they are not authorized for? Or is the idea that if we use for_user everywhere, it will never return findings the user is not authorized for?

The idea is to use `for_user()` everywhere indeed, and for object-level perms on top of that see my previous comment.

> I also think that if a user tries to retrieve an object it has no access to it will generate a 404 instead of 401/403?

That's the consequence if the user should not even be able to see the object, aka not know of its existence at all. When just an object-level permission check fails, e.g. user tries to delete an object he only ahs read-only access to, he'll get a 403.

efficiosoft:
BTW, we could write a simple decorator for legacy function-based views that executes `model.objects.for_user(request).get(pk=pk)`, performs an appropriate object-level permission check for e.g. change or delete on the fetched object and, if all went well, passes the object on to the regular view function. That way, view functions that act on a single object don't need to do the boilerplate `get_object_or_404()` + permission checking stuff.

valentijnscholten:
Well you're welcome on the slack channel: https://owasp-slack.herokuapp.com/ I think I did sent an invite to one of your e-mail a while ago.

The problem with the PR is that it is doing multiple things in one PR modifying 76 files. And some of the queryset/permission parts are implemented some of them are not. I was under the impression this PR contained a fully implemented new way of handling permissions, but it turns out it is not and there is more implementation work to be done in various places. 
I really would have liked a separate PR for the queryset and permission part only. But when you stated that you couldn't make one I decided we will have to make our own containing the stuff we want/like to have. This will take more time from everyone, so yes time could have been saved along the way. 
One of the biggest problems with the UI today is that the authorized users thing doesn't work. You need to be superuser to be able to modify findings and I haven't seen yet if this PR can help with that.

For now I like the way we can more easily restrict the querysets for listing objects. I don't like the 404s for objects that do exist but the user has no access to. 

But may I don't know enough about django and permissions etc and somebody else will need to handle this PR.

efficiosoft:
> Well you're welcome on the slack channel: https://owasp-slack.herokuapp.com/ I think I did sent an invite to one of your e-mails a while ago.

Joined, thanks.

> The problem with the PR is that it is doing multiple things in one PR modifying 76 files.

85% of the changes are ReportNG + template snippets. Permission stuff is only these files IIRC:

* dojo/mixins.py
* dojo/models_base.py
* dojo/models.py
* rules and flexquery in requirements.txt

> And some of the queryset/permission parts are implemented some of them are not.

Not quite. The permission checking and queryset building is fully implemented.
The only thing you maybe need to tweak is the default object-level permissions in `dojo.models_base.DojoModel.preprocess_rules_permissions()`. A user could be allowed to write to all objects he has in his base queryset, for instance.

> I was under the impression this PR contained a fully implemented new way of handling permissions, but it turns out it is not and there is more implementation work to be done in various places.

The permission checking is ready to be used, however, your code base is massive and I told you from the beginning that I didn't touch the existing components. Due to the way the existing views were written, it has to be applied component by component. So, to put it all in a nutshell, these places are:

* The models have to be switched to `dojo.models_base.DojoModel`, which involves no db changes, just switching the base class.
* The models optionally could get a `rules_permissions` setting.
* The views need to use the user-specific base queryset, maybe even via a decorator which would make things easier for legacy view functions.

> I really would have liked a separate PR for the queryset and permission part only. But when you stated that you couldn't make one I decided we will have to make our own containing the stuff we want/like to have. This will take more time from everyone, so yes time could have been saved along the way.

I have been employed for developing ReportNG in the first place, and time was rare. On the other hand, hard-coding the permissions was no option. That's why I developed both hand in hand and ReportNG now demonstrates how to use the permission system, which, by itself, is very slim.

> One of the biggest problems with the UI today is that the authorized users thing doesn't work. You need to be superuser to be able to modify findings and I haven't seen yet if this PR can help with that.

I think I explained that it can.

> For now I like the way we can more easily restrict the querysets for listing objects. I don't like the 404s for objects that do exist but the user has no access to.

That lies in the nature of filtering the base queryset. Of course one could do another db query if an object wasn't found in the restricted queryset to see if it would have existed without queryset restrictions, but that seems like a bad idea to me, considering it'd need another query and the objects that cause a 404 aren't even listed to the user, so how should he get redirected to them at all?

> But may I don't know enough about django and permissions etc and somebody else will need to handle this PR.

I'll contact you on Slack and we'll see if we can sort things out more easily.

efficiosoft:
Right, you now have something concrete to work with. I built a decorator (`dojo.utils.ensure_permission()`) that makes using the new permission checking paradigm a breeze, even for legacy view functions. It's documented and I applied it to `delete_finding()` and `edit_finding()` for demonstration purposes. With that in place, an authorized user should be able to access the finding edit page without being staff, although all the other views need to be adjusted so that he can actually see the products/engs/tests/findings and navigate appropriately, but that's technically it.

And don't forget to switch the models to `DojoModel`, like I did with `Finding` already.

alles-klar:
@efficiosoft Today we had a moderator meeting and decided that we will create new branch for the next release soon. Afterwards we want to merge this PR to dev, so we have some time to get feedback and improve this new features.

valentijnscholten:
We discussed some PRs during our monthly call. Next week `dev` will be promoted as RC for the next release. Once that RC branch is there, we can merge this PR here into `dev` to further test & improve it.

Michael-Aguilera:
Hi, I'm hoping to find some help with this. I've installed this and have been following the Documents to find where to put the templates. I've found it says:
# List of directories to scan for templates.
["<MEDIA_ROOT>/reportng/html_templates"]

My installation directory is /opt/django-DefectDojo so i've tried putting the template in the root of that file so it would be: 
/opt/django-DefectDojo/reportng/html_templates
That didn't work so I tried it in 
/opt/django-DefectDojo/dojo/reportng/html_templates

I've tried looking around to see where in the files it points to ["<MEDIA_ROOT>/reportng/html_templates"] but can't for the life of me find it. I'm on the Slack under the name "Michael Aguilera".

efficiosoft:
@Michael-Aguilera The media root is the directory you configured for storing uploaded files, such as finding images. At the docker-based installs, it's by default a directory named `media`, don't know where you have it though.

valentijnscholten:
There is a PR that make findings/media work in at least the `dev` docker setup: #2056

Michael-Aguilera:
Okay I found the directory you're talking about! Me being me was looking at the files stored on my file directory rather than the docker vms.... The default user is "I have no name!", any idea what the root password is?

valentijnscholten:
@efficiosoft We are ready to merge this into dev and finalize it there. Would you be willing to help by:
- resolving conflicts one more time
I am thinking about marking the new reporting feature as "beta" or "preview" in the menu as during my testing I saw some issues that I think need to be resolved before it is production ready.

alles-klar:
We should care about the migrations. At the moment the current migration file " dojo/db_migrations/0015_findingimage_caption.py" is changed. Can we create a new migration file instead? Otherwise we will get in trouble when updating old DD installations.

valentijnscholten:
Looks like there is some pickling / serialization issue, possibly the rules/permission lru cache?

```
uwsgi_1_819b0a12f074 | Internal Server Error: /product/5/import_scan_results
uwsgi_1_819b0a12f074 | Traceback (most recent call last):
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/kombu/serialization.py", line 50, in _reraise_errors
uwsgi_1_819b0a12f074 |     yield
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/kombu/serialization.py", line 221, in dumps
uwsgi_1_819b0a12f074 |     payload = encoder(data)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/kombu/serialization.py", line 350, in pickle_dumps
uwsgi_1_819b0a12f074 |     return dumper(obj, protocol=pickle_protocol)
uwsgi_1_819b0a12f074 | _pickle.PicklingError: Can't pickle <functools._lru_cache_wrapper object at 0x7f966bd35198>: it's not the same object as django.contrib.auth.models.PermissionsMixin.has_perm
uwsgi_1_819b0a12f074 | 
uwsgi_1_819b0a12f074 | During handling of the above exception, another exception occurred:
uwsgi_1_819b0a12f074 | 
uwsgi_1_819b0a12f074 | Traceback (most recent call last):
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/exception.py", line 34, in inner
uwsgi_1_819b0a12f074 |     response = get_response(request)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 115, in _get_response
uwsgi_1_819b0a12f074 |     response = self.process_exception_by_middleware(e, request)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 113, in _get_response
uwsgi_1_819b0a12f074 |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
uwsgi_1_819b0a12f074 |     return view_func(request, *args, **kwargs)
uwsgi_1_819b0a12f074 |   File "./dojo/product/views.py", line 492, in import_scan_results_prod
uwsgi_1_819b0a12f074 |     return import_scan_results(request, pid=pid)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
uwsgi_1_819b0a12f074 |     return view_func(request, *args, **kwargs)
uwsgi_1_819b0a12f074 |   File "./dojo/engagement/views.py", line 695, in import_scan_results
uwsgi_1_819b0a12f074 |     url=reverse('view_test', args=(t.id, )))
uwsgi_1_819b0a12f074 |   File "./dojo/notifications/helper.py", line 20, in create_notification
uwsgi_1_819b0a12f074 |     process_notifications(event, system_notifications, *args, **kwargs)
uwsgi_1_819b0a12f074 |   File "./dojo/notifications/helper.py", line 121, in process_notifications
uwsgi_1_819b0a12f074 |     send_alert_notification_task.delay(event, notifications.user, *args, **kwargs)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/celery/app/task.py", line 425, in delay
uwsgi_1_819b0a12f074 |     return self.apply_async(args, kwargs)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/celery/app/task.py", line 568, in apply_async
uwsgi_1_819b0a12f074 |     **options
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/celery/app/base.py", line 775, in send_task
uwsgi_1_819b0a12f074 |     amqp.send_task_message(P, name, message, **options)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/celery/app/amqp.py", line 559, in send_task_message
uwsgi_1_819b0a12f074 |     **properties
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/kombu/messaging.py", line 169, in publish
uwsgi_1_819b0a12f074 |     compression, headers)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/kombu/messaging.py", line 252, in _prepare
uwsgi_1_819b0a12f074 |     body) = dumps(body, serializer=serializer)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/kombu/serialization.py", line 221, in dumps
uwsgi_1_819b0a12f074 |     payload = encoder(data)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/contextlib.py", line 77, in __exit__
uwsgi_1_819b0a12f074 |     self.gen.throw(type, value, traceback)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/kombu/serialization.py", line 54, in _reraise_errors
uwsgi_1_819b0a12f074 |     reraise(wrapper, wrapper(exc), sys.exc_info()[2])
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/vine/five.py", line 194, in reraise
uwsgi_1_819b0a12f074 |     raise value.with_traceback(tb)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/kombu/serialization.py", line 50, in _reraise_errors
uwsgi_1_819b0a12f074 |     yield
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/kombu/serialization.py", line 221, in dumps
uwsgi_1_819b0a12f074 |     payload = encoder(data)
uwsgi_1_819b0a12f074 |   File "/usr/local/lib/python3.5/site-packages/kombu/serialization.py", line 350, in pickle_dumps
uwsgi_1_819b0a12f074 |     return dumper(obj, protocol=pickle_protocol)
uwsgi_1_819b0a12f074 | kombu.exceptions.EncodeError: Can't pickle <functools._lru_cache_wrapper object at 0x7f966bd35198>: it's not the same object as django.contrib.auth.models.PermissionsMixin.has_perm
```

efficiosoft:
Too sad if this work would be lost.

After you got the permission stuff implemented, any thoughts on merging ReportNG?

PR #1385, opened 2019-07-15
Fix nikto parsing and importing. FIxes #1382

### LINKED ISSUE #1382: Nikto importing fails
**Bug description**
When attempting to import XML reports from the NIKTO nxmlversion="1.2", the import fails as the importer is not able to find the scan details

I do not have an old file to compare, but the current NIKTO export has 2 <niktoscan> tags in the XML.

dougmorato:
Nikto has 2 “NIKTOSCAN” tags now, updating the root to find the scan details and also improve description parsing/importing

Fix issue #1382

### DIFF
--- a/dojo/tools/nikto/parser.py
+++ b/dojo/tools/nikto/parser.py
@@ -21,6 +21,10 @@ class NiktoXMLParser(object):
         tree = ET.parse(filename)
         root = tree.getroot()
         scan = root.find('scandetails')
+        # New versions of Nikto have a new file type (nxvmlversion="1.2") which adds an additional niktoscan tag
+        # This find statement below is to support new file format while not breaking older Nikto scan files versions.
+        if scan is None:
+            scan = root.find('./niktoscan/scandetails')
 
         for item in scan.findall('item'):
             # Title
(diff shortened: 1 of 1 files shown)


dojo/tools/nikto/parser.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, filename, test):
 
         tree = ET.parse(filename)
         root = tree.getroot()
-        scan = root.find('scandetails')
+        scan = root.find('./niktoscan/scandetails')
```
Maffooch: Replacing the old with the new breaks older scans like [this one](https://github.com/DefectDojo/sample-scan-files/blob/master/nikto/nikto.xml). Maybe add this:
```
scan = root.find('scandetails') 
if scan is None:
    scan = root.find('./niktoscan/scandetails')
```

dougmorato: Thank you @Maffooch , that is a fantastic proposal to maintain backwards compatibility. 

I have amended my commit based on your suggestion and have pushed it. After is done building, it should be ready for your approval,

PR #1395, opened 2019-07-17
Refractor deduplication and implement with django signals

Maffooch:

Deduplication has been heavily tested and tweaked to fit the the description found in the documentation. Major difference here is now new findings will default to being of type dynamic. This change was motivated as a tweak to deduplication to standardize how a finding will be declared as a duplicate. Too many factors were left as a responsibility of a scan parser.

A new finding will default to dynamic. If mentioned finding has a line number and path, but no endpoints, it will be defined as static instead. If at any time a finding has a line number, file path, and any number of endpoints, it will be declared as both static and dynamic. All of these checks take place when a finding is saved so that they may be updated without explicit command.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -6,19 +6,15 @@ from dojo.models import Product, Engagement, Test, Finding, \
     Notes, DojoMeta
 from dojo.forms import ImportScanForm, SEVERITY_CHOICES
 from dojo.tools.factory import import_parser_factory
-from dojo.utils import create_notification
 from django.core.validators import URLValidator, validate_ipv46_address
 from django.conf import settings
-from django.urls import reverse
 from rest_framework import serializers
 from django.core.exceptions import ValidationError
 from django.utils import timezone
-from django.db.models import Q
 import datetime
 import six
 from django.utils.translation import ugettext_lazy as _
 import json
-from tagging.models import Tag
 
 
 class TagList(list):
@@ -561,21 +557,6 @@ class ImportScanSerializer(TaggitSerializer, serializers.Serializer):
         skipped_hashcodes = []
         try:
             for item in parser.items:
-                if skip_duplicates:
-                    hash_code = item.compute_hash_code()
-
-                    if ((test.engagement.deduplication_on_engagement and
-                             Finding.objects.filter(Q(active=True) | Q(false_p=True) | Q(duplicate=True),
-                             test__engagement=test.engagement,
-                             hash_code=hash_code).exists()) or
-                        (not test.engagement.deduplication_on_engagement and
-                            Finding.objects.filter(Q(active=True) | Q(false_p=True) | Q(duplicate=True),
-                                                    test__engagement__product=test.engagement.product,
-                                                    hash_code=hash_code).exists())):
-
-                        skipped_hashcodes.append(hash_code)
-                        continue
-
                 sev = item.severity
                 if sev == 'Information' or sev == 'Informational':
                     sev = 'Info'
@@ -593,7 +574,7 @@ class ImportScanSerializer(TaggitSerializer, serializers.Serializer):
                 item.last_reviewed_by = self.context['request'].user
                 item.active = data['active']
                 item.verified = data['verified']
-                item.save()
+                item.save(dedupe_option=False)
 
                 if (hasattr(item, 'unsaved_req_resp') and
                         len(item.unsaved_req_resp) > 0):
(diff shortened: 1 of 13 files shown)


dojo/api_v2/serializers.py
```diff
@@ -561,21 +557,6 @@ def save(self):
         skipped_hashcodes = []
         try:
             for item in parser.items:
-                if skip_duplicates:
```
devGregA: @Maffooch a little confused about this removal. These hash codes aren't related to dedupe. Would you please explain?

Maffooch: These hash codes were computed in APIv2 as a way to preform a different kind of deduplication than that of the GUI and APIv1. They were deleted so that all three methods of Dojo interaction could be consistent.

efficiosoft:
@Maffooch and @devGregA  This PR again changed an existing migration (b53afd930e296ff331876d1c2313b53a409f81bc). Didn't we agree that existing migration files, once merged to dev, shouldn't be touched anymore to not make migrations pointless at all?

```
commit b53afd930e296ff331876d1c2313b53a409f81bc
Author: Maffooch <codymaffucci@Codys-MacBook-Pro.local>
Date:   Wed Jul 17 11:22:36 2019 -0500

    Refractor deduplication and implement with django signals

diff --git a/dojo/db_migrations/0009_endpoint_remediation.py b/dojo/db_migrations/0009_endpoint_remediation.py
index 057d381c..1a2e298f 100644
--- a/dojo/db_migrations/0009_endpoint_remediation.py
+++ b/dojo/db_migrations/0009_endpoint_remediation.py
@@ -15,4 +15,9 @@ class Migration(migrations.Migration):
             name='remediated',
             field=models.BooleanField(blank=True, default=False),
         ),
+        migrations.AlterField(
+            name='dynamic_finding',
+            field=models.BooleanField(default=True),
+        ),
     ]


```

Or did I miss something here? Thanks for clarifying.

Maffooch:
@efficiosoft It seems I am the one missing something. This won't happen again from my end.

efficiosoft:
> @efficiosoft It seems I am the one missing something. This won't happen again from my end.

Brilliant, thanks. Just wanted to be sure it wasn't for some particular reason.

dojo/api_v2/serializers.py
valentijnscholten: I'm using this so not in favor of removing this.

Maffooch:
@valentijnscholten @ptrovatelli  This is going to a long one as there were a lot of tough choices to be made here.

**Modification to Hash Code Computation**
Hash codes were being computed too early. In most cases (especially in APIv2) vulnerabilities from imports would have hash codes computed before ever being initialized. Before modifications, if a finding were dynamic, the only info the hash would be title, description, and CWE. This is because endpoints are linked to a finding after they are initially saved. There could be multiple identical vulnerabilities across the test subject with all different endpoints. Only one of them would be considered "new" and the rest as duplicates. If Delete Duplicates is enabled, the user would not know they were ever imported. Therefore, accommodations needed to be made to resolve these issues.

**Removal of skip_duplicates**
Since duplicates cannot be determined before importing vulnerability into the database, they have to be imported to get an accurate duplicate status. Since it's in the DB now, this defeats the purpose of "skipping" the duplicates. Using the Delete Duplicates option in system settings is the equivalent of "skipping" the duplicate before it is imported. 

 **Removal of close_old_findings**
I totally misunderstood this one. This should not have been deleted. I was under the impression that this was happening at the same time as skip_duplicates was. This is not the case as it only opens after all imported findings have been initialized. Perhaps this should be done by default, but only engagements?

**Dynamic Finding By Default**
A reoccurring issue that kept appearing is whether a finding was dynamic or static. If a parser did not include this information, a finding will remain as neither static nor dynamic. In this case, the hash codes will again not be completely sufficient to determine duplicates. I believe it's a safe assumption to make findings dynamic by default. There are measures in place to correct this action in the presence of static qualities.

valentijnscholten:
Why can't you compute the hash code without saving it to db? I think we still need something to avoid keep importing the same findings over and over again only to mark them as duplicate.

Maffooch:
Endpoints need to be included in the hash code. They cannot be added to a finding until after the finding has been saved.

Something I did not add to my report:
Having both system wide deduplication and the APIv2 options is redundant. They both do the same thing in a different order. It would be silly to have them both in all three places. Maybe it was actually better to have APIv2 separate from the other two means on importing findings? It seemed to make everybody happy having different options.

valentijnscholten:
As a user I want defect dojo *not* to import the exact/identical same finding over and over again just to mark it as a duplicate. Regardless of how it is currently implemented. The hashcode calculation is pretty simple so can we adjust it to work with in memory objects instead of db entities?

Maffooch:
Enabling delete duplicates will remove the duplicates. You won't see them. I could not figure out a way to to work with in memory objects correctly. I implore you to figure it out though as I was falling short.

valentijnscholten:
I finally got around to upgrading my instance to the latest dev, including theses changes around configurable dedup per parser etc. I still think skip_duplicates is useful / desired. 
I a CI/CD scenario you get lots of scans with lots of duplicate findings.Currently these findings are created first and then deleted if dupe delete is enabled. In a way it works fine, but it has some undesired side effects. When a new finding gets created:

- it can trigger notification / alerts /emails
- it can trigger updates to external systems like sonarcube or JIRA
- the "push all findings" function for JIRA is currently broken, but when fixed it might push duplicate findings to JIRA.
- the importer doesn't know if it is a duplicate, so it cannot report back in the API result how many duplicates were in the scan

I have looked at the code and as far as I can see the only reason currently why it is hard to implement skip_duplicates seems to be that the endpoints are needed to calculate the hash (dynamic findings only). And the hash computation expects these to be in the database. It looks like it is not too hard to rewrite this a little so it can use in memory endpoints. In fact this mechanism is already more or less in place via the "unsaved_endpoints" field.

Let me know what you think or if I am underestimating things or missing some obstacles.

PR #1427, opened 2019-07-28
Fix missing "FixedBy"

wurstbrot:
As I found out, the "FixedBy" is not mandatory set by clair klar.


wurstbrot:
This one might be important for the clair scanner also, @benno001

Issue #810, opened 2018-11-15
DefectDojo SonarQube Integration


rjimgal:
First of all, thank you for this amazing tool, I'm just starting to use it.

I would like to use this recent Integration, however I am unable to generate HTML report in SonarQube.

Sample HTML file can be obtained at https://github.com/DefectDojo/sample-scan-files/blob/master/sonarqube/sonarqube_v6.7.5.html and it works like a charm.

I am also using SonarQube 6.7.5 version, on its Community Edition.

Having a look to HTML sample code, it contains some banner about @AdrienGuillerme employer, which leads me to think, that provided HTML report is an internal report generated by them. If so, would you mind to share how you generate it? Maybe some template / SonarQube API calls.

Thanks in advance.

AdrienGuillerme:
I will tak a look at it

aaronweaver:
Hi @AdrienGuillerme just wanted to checkin to see how the file is generated. Thanks!

rjimgal:
Hi @AdrienGuillerme, any update on this?

W1R3D-Code:
also interested, for what it's worth :)

agigleux:
Hello,

I'm working for SonarSource the company making SonarQube. We are looking at the options to provide an official integration with DefectDojo that will be supported by SonarSource.

Do you know if DefectDojo is supporting a generic import format? Or shall we define our own export format and provide the importer in the DefectDojo repo?

Thanks

aaronweaver:
@agigleux That's fantastic news, we do have a generic csv or if it's easier you can define a json export and we can create it. Thanks!

agigleux:
@aaronweaver I tried to follow what is documented [here](https://defectdojo.readthedocs.io/en/latest/integrations.html) and generate a file corresponding to it but no luck so far. Can you provide a sample CSV file so I compare with what I generated?

I used https://defectdojo.herokuapp.com/ and selected "Generic Findings Import". Do you confirm this instance allows upload of files, maybe that's the reason why I can't import it?

aaronweaver:
I'll take a look and get back to you shortly. Thanks!

aaronweaver:
Hi @agigleux, I updated the csv field [here](https://github.com/DefectDojo/sample-scan-files/blob/master/generic/generic_csv_defect_dojo.csv). If you need any additional fields let me know and I can add that to this importer. Thx

agigleux:
[...]

PR #1445, opened 2019-08-01
Reactivate sonar

ptrovatelli:
List of changes: 
- Reactivate sonarQube parser. (more info: see https://github.com/DefectDojo/django-DefectDojo/issues/810)
- Fix it for python3
- add file_path and line_number
- fix an issue with the logic for finding the vulnerabilities and the rules
  - before it was finding all the "tbody" and expect to find data always at the same nth tbody. 
  - However sometimes there are more "tbody" in the rules description, and when there are no vulnerabilities found, one of the tbody disappears so it didn't work. Fix by using xpath instead to locate the list of vuln and the rules. 
- force the findings as static findings (they were dynamic before)
- add unit tests

Beware that you'll need the last version of the sonar html export for this to work. it's here https://github.com/soprasteria/sonar-report but the npm registry is not up to date yet. 

See also
- a sample report here https://github.com/DefectDojo/sample-scan-files/pull/32
- updated documentation: https://github.com/DefectDojo/Documentation/pull/57

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -118,7 +118,7 @@ or:
 docker logs django-defectdojo_initializer_1
 ```
 
-Beware that when re-running the application several times, there may be several occurrences of "Admin password". In that case you should use the last occurrence.
+Make sure you write down the first password generated as you'll need it when re-starting the application.
 
 ### Disable the database initialization
 The initializer container can be disabled by exporting: `export DD_INITIALIZE=false`. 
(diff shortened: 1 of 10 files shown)


ptrovatelli:
I have an idea for a perf improvment: that get_rule_details function has a complexity of number of findings * (number of rules/2). 
It could be lowered to number of findings + number of rules  if we go through the rules just once and put them in a hashmap. 
@praveendvd also found a issue when importing non-vulnerability findings. i'll check this too

Maffooch:
Hey @ptrovatelli please reactivate [here](https://github.com/DefectDojo/django-DefectDojo/blob/20d2e11305d673719015ea8ce581181d7db6da77/dojo/templates/dojo/import_scan_results.html#L70) as well. Thanks.

ptrovatelli:
> Hey @ptrovatelli please reactivate [here](https://github.com/DefectDojo/django-DefectDojo/blob/20d2e11305d673719015ea8ce581181d7db6da77/dojo/templates/dojo/import_scan_results.html#L70) as well. Thanks.

Nice catch thanks! done.

Issue #897, opened 2019-02-22
Redirection in Bulk Edit


HumanoidPhantom:
### Issue Type 
- [ ] Bug
- [x] Enhancement

### Description
It would be more convenient, if after bulk change all filter settings and current page number won't be discarded.

### Solution
Possible workaround would be to use referrer HTTP header (and check hostname, just in case) in dojo/finding/views.py:
```
referrer = request.META.get('HTTP_REFERER', '/')
if urlparse(referrer).hostname in settings.ALLOWED_HOSTS:
    return HttpResponseRedirect(referrer)
else:
    ... { the default behavior}
```

PR #1460, opened 2019-08-06
Enhancement #897 and Minor Wording Change

devEricA:
This is my enhancement for #897. Bulk Edit no longer resets the filters. 

In addition, whenever a user wants to delete multiple findings, the user will be greeted with "Are you sure you want to delete these findings?" instead of "Are you sure you want to delete this finding?"


Maffooch:
Hey @devEricA changes made through bulk edit were not applied when I tested this. The filters were consistent though.

devEricA:
Crap, your right @Maffooch. Working on getting it resolved.

Issue #1036, opened 2019-04-10
UWSGI container appears to be missing settings.py file


surewould:
**Bug description**
uwsgi container appears to be missing the settings.py file and fails to start the application

**Steps to reproduce**
1. Follow docker quick start... docker-compose up
2. Scroll down to initial uwsgi_1 status

```
uwsgi_1         | Traceback (most recent call last):
uwsgi_1         |   File "./dojo/wsgi.py", line 29, in <module>
uwsgi_1         |     application = get_wsgi_application()
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/wsgi.py", line 13, in get_wsgi_application
uwsgi_1         |     django.setup(set_prefix=False)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/__init__.py", line 22, in setup
uwsgi_1         |     configure_logging(settings.LOGGING_CONFIG, settings.LOGGING)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 56, in __getattr__
uwsgi_1         |     self._setup(name)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 41, in _setup
uwsgi_1         |     self._wrapped = Settings(settings_module)
uwsgi_1         |   File "/usr/local/lib/python2.7/site-packages/django/conf/__init__.py", line 110, in __init__
uwsgi_1         |     mod = importlib.import_module(self.SETTINGS_MODULE)
uwsgi_1         |   File "/usr/local/lib/python2.7/importlib/__init__.py", line 37, in import_module
uwsgi_1         |     __import__(name)
uwsgi_1         | ImportError: No module named settings
```

settings.py appears to be missing

```
root@a5e158be0e3b:/app/dojo/settings# ls
__init__.py  settings.dist.py  template-env  unittest.py
root@a5e158be0e3b:/app/dojo/settings# 
```

**Expected behavior**
I expected the application to load as per the quick start.

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [X] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: Manjaro 18.0.4
 - DefectDojo version: d957356: Merge pull request #1022 from DefectDojo/dev [2019-04-08 15:09:21 -0400]

surewould:
I may have been premature. This looks like it is well explored below.

[...]

PR #1473, opened 2019-08-10
docker-compose not working

wurstbrot:
For dev env. it is given to copy the settings.py.

I could not make it work without the copy step in docker-compose test env., too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ credentials.
 ```sh
 git clone https://github.com/DefectDojo/django-DefectDojo
 cd django-DefectDojo
+# Quickfix
+cp dojo/settings/settings.dist.py dojo/settings/settings.py
 # building
 docker-compose build
 # running


ptrovatelli:
Hi @wurstbrot. Thanks for sharing. What do you mean by "docker-compose test env"? What environment are you using? can you check with: 
`docker/setEnv.sh`
or 
`ls -l docker-compose.override.yml`

You have added the comment in the quickstart instructions which is the "release" env by default.
Also you have commited this on master. Are you using the master branch? Usually we ask to create PRs on the dev branch alghough being just a doc push it can be discussed; still will need to be integrated to dev branch is confirmed. 

Thanks!

wurstbrot:
It is the same as https://github.com/DefectDojo/django-DefectDojo/pull/1474

ptrovatelli:
@wurstbrot I think you probably have a docker-compose.override.yml link remaining which is why you're seing this issue. 
It may be related to this: https://github.com/DefectDojo/django-DefectDojo/pull/1479
What's your OS? some OSes may checkout this file as a working symbolic link. 
For me  (linux) it was checked out with in contents the name of the destination (docker-compose.override.dev.yml)
If you remove the file, you shouln't have a problem.

ptrovatelli:
the missing "settings.py" issue is only present when mounting the docker volume (in dev setup) if the unix user inside of the container fails doing that copy. 
the easy fix is to do the copy manually before starting the app but this is only required when the volume is mounted. 
the "clean" fix is setting a coherent unix user id in Dockerfile.django but this depends on each host! 
it's here:
`USER 1001`
that uid needs to be the one running the application on the host (see https://github.com/DefectDojo/django-DefectDojo/pull/1045#issuecomment-482736718 )

this was discussed in 
https://github.com/DefectDojo/django-DefectDojo/issues/1036
https://github.com/DefectDojo/django-DefectDojo/issues/1026
https://github.com/DefectDojo/django-DefectDojo/pull/1028

wurstbrot:
you are right, for master it is working after changing fullcalendar version.

PR #1485, opened 2019-08-12
ptvsd debug option

madchap:

Coming back with this, I would really love to see this get over the line. A step-by-step option to debug on container is super useful, even with no hot reloading.

I would very likely require some help for 2 things, for which I have not had the time to look very hard into due to lack of time mostly. Some more expert eyes may see the issue right away.
- The static assets need to be copied manually right now via a `docker cp` operation onto the uwsgi container
- For some reason, when then serving over port 8000 directly from the uwsgi container, the page hangs when loading the resource, whereas they're there. A conflict with the then un-used nginx container or something?


madchap:
@ptrovatelli a wild shot in the dark, since I see you tinkering with docker options right now.. I'd need some help to get this through. Maybe you could take a look if you have time?

ptrovatelli:
i'll have a look asap :)

ptrovatelli:
hi @madchap . How about leaving that nginx container as defectDojo works so well with it :)
something like that: https://github.com/ptrovatelli/django-DefectDojo/tree/pr-1485
On that branch defectDojo works good but i wasn't able to make breakpointing work yet; I don't see why it wouldn't work though: you can work with nginx and do your magic on port 3000 can't you?

alles-klar:
@ptrovatelli I had a similar idea but I was not able to set a breakpoint too. I ended up in using the django runserver and generating the static files locally. But I have performance problems when using the option --nothreading with the django runserver. Do you have similar problems?

In my solution I put everything in a docker-compose file so we don't need a new entrypoint script and the modifications in settings.py.

ptrovatelli:
I had it working with nginx. 
- when using uwsgi, the debug server launch needs to be in wsgi.py not manage.py
- i have added a couple of flags in the entrypoint, based on some discussions in forums and such; they may not be all mandatory (this for eg: https://www.mail-archive.com/uwsgi@lists.unbit.it/msg05414.html) : 
`  --enable-threads --lazy-apps --honour-stdin`

I think --honour-stdin is more for pdb and might be removed. 

It works now with this code: https://github.com/ptrovatelli/django-DefectDojo/tree/pr-1485
just need to re-attach when saving a file (hot-reloading dettaches the debuger). 
debugger

I have also used a hardcoded localRoot in launch.json to troubleshoot "unverified breakpoint" error when using ${workspaceFolder}

Not sure about performances, I haven't used it a lot. with --enable-threads it should be fine. If not stable, should be replaced by --nothreading and will probably be slower. 
Hope that helps!

ptrovatelli:
We could maybe put this code in the "dev" environment to limit the number of environments? 
It would be nice to have this working for celery worker too (deduplication works in celery worker container). 
Celery worker is not using uwsgi but celery (see entrypoint-celery-worker.sh).
I have tried starting the debug server in celery.py but it doesn't seem to have any effect: nothing listening inside the container on port 3001:
```
sudo nsenter -t $(docker inspect -f '{{.State.Pid}}' django-defectdojo_celeryworker_1) -n netstat -an | grep LISTEN
tcp        0      0 127.0.0.11:35986        0.0.0.0:*               LISTEN     
```
draft is here: https://github.com/ptrovatelli/django-DefectDojo/tree/pr-1485-celeryworker

@alles-klar does your method work for celery worker? can you share the code?

madchap:
I will give that a shot, thanks!

> just need to re-attach when saving a file (hot-reloading dettaches the debuger).

Really? Everything I read was pointing to have it mono thread and no hot reload. But why not :)

> Not sure about performances, I haven't used it a lot. with --enable-threads it should be fine. If not stable, should be replaced by --nothreading and will probably be slower.
Well, when debugging, I don't care that much about performance. Unless it takes 10mn to ingest a single finding :-)

madchap:
Hey @ptrovatelli 

I brought the changes from your branch onto mine, and it seems to work well :+1: I did only a small test, but it sure enough is already a better experience! Thanks :) I will push over here in a moment.

It could be this can be integrated for `dev` as to not have yet another `ptvsd` environment for sure. Maybe @Maffooch can pitch in on this?

madchap:
As far as I am concerned, this is now satisfying the base use-case I guess.

There are a couple things that would still need to be decided on worked on:
* [Enhancement, could be done on another PR] Get this working for celery.
* [Decision] do we want to directly include this in `dev` rather than have a distinct `ptvsd` target for `setEnv.sh`

For the first point, I would say we could decouple it from here.
The second point would require some input from @Maffooch , @aaronweaver or some other maintainer I guess.

I would definitely like to see this included in `dev` as to avoid losing it another time.

ptrovatelli:
I guess we can leave it in a different entrypoint for now; this will simply make the "dev" env obsolete as it does nothing more than the ptvsd env. if we confirm that it's all good, we can just remove the dev env in a while.

Issue #1254, opened 2019-06-12
Complete ImportScanner Unit Test Requirements


Maffooch:
**Scanner Name**
[Located in python3_dev](https://github.com/DefectDojo/django-DefectDojo/blob/python3_dev/tests/Import_scanner_unit_test.py)

**Sample File**
[To-Do](https://docs.google.com/document/d/14ncrWjAPYQvsYOPg5nMkVW1CO9IJwFhaMRd2w9WpxJI/edit#)

propersam:
Unit test is my focus....

i'm working on this now

propersam:
it's probably gonna last all through next week... but i should be done before GSOC evaluation week...

propersam:
I have made a PR to update sample_test_file repo..

devGregA:
@Maffooch am I good to close this?

Maffooch:
Don't think he is done with it yet. @propersam What is your status on this one?

propersam:
No.. Don't close this yet...  We are still on it...

PR #1490, opened 2019-08-13
Wrap Up Import Scanner Unit Test

propersam:
#1254 


Maffooch:
@propersam So this is totally complete now? It can be ran without failing?

propersam:
@Maffooch i've fixed the code from this end. There are just some sample files yet to be added to the sample scan files repository which would be the only thing obstructing this test from running.

So I would focus there now as I get sample scan files to add to the repo.

These are the currently missing sample scan files in sample-scanfile-repo:
aws_scout2
mobsf
skf
spotbugs
vcg

Maffooch:
Got it. Whenever you get everything situated, could you make this test and the other unit tests in the tests folder run on travis?

propersam:
> @propersam So this is totally complete now? It can be ran without failing?

it will fail for the missing-sample scan test files. but others work fine.

Issue #1482, opened 2019-08-11
Trufflehog reports are not getting uploaded


praveendvd:
**Bug description**

Gets json should be str adn not bytes error on uploading trufflehog html.

**Steps to reproduce**

1. Go to engagements
2. Upload trufflehog json
3. Exception is thrown

**Expected behavior**

File should get uploaded

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [X] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: CentOS 7
 - DefectDojo version: c8a639a: Merge pull request #1467 from Maffooch/dev [2019-08-09 11:43:07 -0500]

**Sample scan files** (optional)

[t.json.zip](https://github.com/DefectDojo/django-DefectDojo/files/3492377/t.json.zip)

**Screenshots** (optional)

![image](https://user-images.githubusercontent.com/45095911/62839211-e2e3ec80-bc7e-11e9-9ad6-77cf9991dea7.png)

PR #1491, opened 2019-08-13
Fix Trufflehog and change dedupe test file path

Maffooch:

Trufflehog fix motivated from #1482 
Dedupe path change is for obvious reasons. 

### DIFF
--- a/dojo/tools/trufflehog/parser.py
+++ b/dojo/tools/trufflehog/parser.py
@@ -69,7 +69,10 @@ class TruffleHogJSONParser(object):
 
     def parse_json(self, json_output):
         try:
-            json_data = json.loads(json_output)
+            try:
+                json_data = json.loads(str(json_output, 'utf-8'))
+            except:
+                json_data = json.loads(json_output)
         except ValueError:
             raise Exception("Invalid format")
 
--- a/tests/dedupe_unit_test.py
+++ b/tests/dedupe_unit_test.py
@@ -29,6 +29,7 @@ class DedupeTest(unittest.TestCase):
         self.base_url = "http://localhost:8000/"
         self.verificationErrors = []
         self.accept_next_alert = True
+        self.relative_path = dir_path = os.path.dirname(os.path.realpath(__file__))
 
     def login_page(self):
         driver = self.driver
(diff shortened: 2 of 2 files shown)


dojo/tools/trufflehog/parser.py
```diff
@@ -69,7 +69,10 @@ def __init__(self, filename, test):
 
     def parse_json(self, json_output):
         try:
-            json_data = json.loads(json_output)
+            try:
+                json_data = json.loads(str(json_output, 'utf-8'))
```
ptrovatelli: What will happen for those bytes that are not utf-8?
If you have a doubt you're really reading utf-8, I believe the clean way is : 
`json_output.decode('utf-8')`
by defaut this will throw an exception if it's not utf-8, but you can also : 
`json_output.decode('utf-8', errors='replace')`
which will replace unknown characters by �
Or
`json_output.decode('utf-8', errors='ignore')`
which will strip out unknown characters

From that point on you're working with a <type 'bytes'> (json_output was a <type 'str'>)
when you're finished working with it and want a type str (to pass to the database for ex) then you'll need to 
`encode('utf8')`

I believe you need to encode before passing to the loads function as in version 3.5, `loads` doesn't support bytes yet: 
https://docs.python.org/3/library/json.html
```
json.loads(s, *, encoding=None, cls=None, object_hook=None, parse_float=None, parse_int=None, parse_constant=None, object_pairs_hook=None, **kw)
Deserialize s (a str, bytes or bytearray instance containing a JSON document) to a Python object using this conversion table.
(...)
Changed in version 3.6: s can now be of type bytes or bytearray. The input encoding should be UTF-8, UTF-16 or UTF-32.
```

so in the end this should be fine with for example: (don't trust me though please test :) I don't see unit tests and don't have samples with me...)
`json_data = json_output.decode('utf-8', errors='replace').encode('utf-8')`

(no exception handling needed)

Maffooch: @ptrovatelli Good point on the chars that are not utf-8. I had not thought about that. The reason for the exception handling is that the files are not always bytes or strings. For example, I tried a fix that handled imports correctly, but would fail the unit tests. They used the same function, so I was totally stumped. Unfortunately, there's about 25 other parsers that suffered from this issue. Here is the rest on #1467. I had forgotten Trufflehog in the cluster!

ptrovatelli: OK. it's just that json.load takes utf-8 strings in input so it seems strange to give a string that failed str(..., utf-8). However after a few tests it doesn't seem as easy as I expected, i'll need to dig it a little to see what's what.

Maffooch: Believe me, I know. It nearly took me three times to get it working! I've seen the import through GUI/API reads files as bytes. I suspect it's because of the temp file upload. The unit tests open the file to read so it auto converts to string. Thats why I did it in that order.

Issue #1492, opened 2019-08-13
Configure Travis to run all unit tests in the /tests folder


Maffooch:
These tests were written to be used by Travis.

ptrovatelli:
@propersam congrats on the first PR on your repo :)
https://github.com/propersam/django-DefectDojo/pull/1
i have fixed the "dedupe_unit_test" in that PR 

It seems that you have already fixed a number of tests, well done!
Do you know if those work with travis? If yes I think you should probably push those that work as soon as possible, and finish the others later. the more we wait, the more people may break tests that currently work, which means more time needed to fix them again.
I have setup an ubuntu VM to run the tests which is probably what you did too; I think we could probably create a new Dockerfile for this environment so that everyone can run the tests locally with docker without the need of ubuntu. I'll have a look at this.

propersam:
Yea..many of what i fixed work. Infact almost all the work are suppose to work. I am only encountering a challenge with user permission.

in the Dockerbuild file there is a command that restricts users.
i think it's `USER 1001`. This command limits some of the test and causes permission error. This is what i have been trying to find work around for these past few months.

if that command is command from the Dockerbuild file, all my test current having errors works perfectly but I guess the command is needed and should not be removed.

So I have been think of how a different docker build cn be configured with full permission just for unittest sake. If this can be achieve. my PR will be well ready for merge.

propersam:
I think for now..i'm going to have to comment out those tests that requires permission...So that the tests that work cn be merged and the travis unittest test is added to the CI pipeline...

ptrovatelli:
@propersam I think I know that issue. for permissions to work properly in the container, that USER id that you put in the Dockerfile should be the same as the host user id (uid) you're using when cloning the git repo because that uid in Dockerfile is the user id that you'll have inside the container, and the files belong to the user outside the container that did the git clone. 
typically it's 1000 instead of 1001 (type simple the command "id" to find out)
The problem is that depending on the host configuration it may vary but i think we can simply add this in the doc so that ppl put their own uid.
 By default we should leave the uid that works for travis, it might be 1001
There are already some words about that issue in the DOCKER.MD: 

*Notes about volume permissions*

*The manual copy of settings.py is sometimes required once after cloning the repository, on linux hosts when the host files cannot be modified from within the django container. In that case that copy in entrypoint-uwsgi-dev.sh fails.* 

[...]

PR #1501, opened 2019-08-16
Integrate Unit tests to Travis CI jobs

propersam:
#1492 Configured Travis to Run Unit Tests


devGregA:
@propersam please see https://travis-ci.org/DefectDojo/django-DefectDojo/builds/574357177?utm_source=github_status&utm_medium=notification

propersam:
I just closed this PR while I continue to work on it...I'll re-open once it's good to go

propersam:
#1492

propersam:
I've been able to squash all the numerous commits..

0001_initial.py
```diff
@@ -0,0 +1,141 @@
+# Generated by Django 2.2.1 on 2019-07-17 21:07
+
+from django.conf import settings
+from django.db import migrations, models
+import django.db.models.deletion
+import django_extensions.db.fields
+
+
+class Migration(migrations.Migration):
```
Maffooch: This migration file shouldn't be here. Everything else looks great. Wonderful job tackling this huge assignment!

propersam: thanks..it's great to finally get this up and running...
that new file was introduced when i was trying to squash the commits..please check to make sure all other migration files are okay too...cause i had to resolve some conflicts from the migration files...

travis/script.sh
```diff
@@ -200,6 +200,15 @@ echo "Running test ${TEST}"
       echo "Docker compose container status"
       docker-compose -f docker-compose.yml ps
       ;;
+    unittests)
+      echo "run unittests scripts"
+      # change user id withn Docker container to user id of travis user
+      sed -i -e "s/USER\ 1001/USER\ `id -u`/g" ./Dockerfile.django
+      cp ./dojo/settings/settings.dist.py ./dojo/settings/settings.py
+      source ./docker/setEnv.sh release
+      docker-compose build
```
ptrovatelli: we could use the `build_containers` function like it's done in the other functions

travis/script.sh
```diff
@@ -200,6 +200,15 @@ echo "Running test ${TEST}"
       echo "Docker compose container status"
       docker-compose -f docker-compose.yml ps
       ;;
+    unittests)
+      echo "run unittests scripts"
+      # change user id withn Docker container to user id of travis user
+      sed -i -e "s/USER\ 1001/USER\ `id -u`/g" ./Dockerfile.django
+      cp ./dojo/settings/settings.dist.py ./dojo/settings/settings.py
+      source ./docker/setEnv.sh release
```
ptrovatelli: this line is not required as it's the default mode

travis/script.sh
```diff
@@ -200,6 +200,15 @@ echo "Running test ${TEST}"
       echo "Docker compose container status"
       docker-compose -f docker-compose.yml ps
       ;;
+    unittests)
+      echo "run unittests scripts"
+      # change user id withn Docker container to user id of travis user
+      sed -i -e "s/USER\ 1001/USER\ `id -u`/g" ./Dockerfile.django
+      cp ./dojo/settings/settings.dist.py ./dojo/settings/settings.py
```
ptrovatelli: this line is not required (only required for dev env)

docker/entrypoint-unit-tests.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 # Run available unittests with a simple setup
+umask 0002
```
ptrovatelli: I don't think you need this as you use the release mode, so it's using entrypoint-uwsgi.sh

docker/entrypoint-unit-tests-devDocker.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 # Run available unittests with a simple setup
+umask 0002
```
ptrovatelli: I don't think you need this as you use the release mode, so it's using entrypoint-uwsgi.sh

.travis.yml
```diff
@@ -16,6 +16,7 @@ env:
     - TEST=flake8
     - TEST=snyk
     - TEST=docker
+    - TEST=unittests
```
ptrovatelli: as discussed, how about calling that "integration_tests" in order not to mix them with the unit tests? the dojo.unittests are run under the helm context in travis script.sh main section
```
helm/defectdojo/templates/tests/unit-tests.yaml
      command: ['/entrypoint-unit-tests.sh']
        entrypoint-unit-tests.sh
          python manage.py makemigrations dojo
          python manage.py migrate
          exec python manage.py test dojo.unittests --keepdb
```
the "unittest-script.sh" script name should also be changed

tests/Engagement_unit_test.py
```diff
@@ -66,22 +69,6 @@ def test_edit_created_new_engagement(self):
         EngagementTXT = driver.find_element_by_tag_name("BODY").text
         self.assertTrue(re.search(r'Engagement updated successfully.', EngagementTXT))
 
-    def test_engagement_import_scan_result(self):
```
ptrovatelli: what's wrong with that section? not working or not interesting? if just not working, could you leave it only commented out so that someone can work on it later it they find it's actually an interesting test? (or rather: leave those lines and comment out the call `suite.addTest(EngagementTest('test_engagement_import_scan_result'))`

travis/script.sh
propersam: using build containers i won't be able to run *docker-compose up -d* so that all the services can run and communicate together in the background and browsers will be able to access the service in *headless mode* within the travis runner

propersam: I won't also be able to fetch password that i cn now set in environment for the login pages...

travis/script.sh
propersam: line 206 helps me solve the challenge i hv been having for several months now...error from permission..

line 207 and 206 seems like it's not needed now..but when i was debugging error ..release mode does not tell me what's wrong..but i used dev mode then. With dev mode i was able to debug and trace my issue....It's when i fixed it..i now returned it back to release...
..some one may need to change it back to dev in future when they are trying to debug why their test fails....that's why i would suggest we let it be...

propersam: it won't really do anytn now since the container is already in release mode..but it would be cool to know that one can change the mode at anytime....someone may want to run the test that's available only to unittest mode..

docker/entrypoint-unit-tests.sh
propersam: yea...i can remove that...was added when i was really trying to find out how to fix the issue i had.

docker/entrypoint-unit-tests-devDocker.sh
propersam: removed..

docker/entrypoint-unit-tests.sh
propersam: removed..

tests/Engagement_unit_test.py
propersam: exactly the same test exist in finding_unit_test.py file ... **test_import_scan_result**

PR #1515, opened 2019-08-21
Added option to set a caption for finding images

efficiosoft:
- [X] Model changes must include the necessary migrations in the dojo/dd_migrations folder.
  -- Don't think this is needed for a plain `CharField` with all form handling performed implicitly by the Django machinery.


efficiosoft:
Think so as well. I randomized filename generation for uploaded images in the same run to prevent brute force attacks. What I'm still unsure about is whether to preserve the original filename as prefix or just use the random UUID to have them all the same length and not expose (potentially) private names. I'm slightly in favor of not keeping the original name. What do you think?

Maffooch:
Uniform length filenames would be best. Also removes the concern of going over on filename size. I don't believe the UUID can be determined without the actual file either can it? Seems to be a good direction.

efficiosoft:
Yes, max path length was also a concern of mine. Changed to drop the original basename.

efficiosoft:
BTW, the UUID is fully random, no hash, so it can't be determined at all without the database record :)

Maffooch:
Cool! learn something new every day.

efficiosoft:
Hmm, but there's another problem. Currently, the filename extension is preserved and thus could - in theory - be used to overflow the path length. But it can't be dropped because mimetype detection relies on it. I think we'll have to write a custom validator that only accepts a fixed set of extensions, e.g. `["jpg", "png"]`...

efficiosoft:
Right, Django's `ImageField` validates and only accepts image files, so we should be totally safe in that regard and the PR is ready for merging from my side at least.

Issue #1510, opened 2019-08-20
Ability to link the duplicate issue when marking a finding as duplicate


madchap:
**Is your feature request related to a problem? Please describe**
As far as I can tell, there is no way to point to the "original" finding when marking a duplicate.

**Describe the solution you'd like**
Probably a text field in the edit finding screen, that also does a check to make sure the linked finding is a valid one.

**Describe alternatives you've considered**
Dirty database query, such as `update dojo_finding set duplicate_finding_id = 11 where `id` = 22;`

PR #1516, opened 2019-08-21
Manually set parent of duplicate

Maffooch:

Motivated by #1510 

Findings that are manually marked as duplicate now provide a link to the original finding and will appear in the parent findings duplicate list. Duplicates are determined by title only. Information displayed on each finding is as follows:

id: title, product/engagement/test

The dropdown menu will only appear if the the duplicate box is selected and there are other findings that have a matching title.

Here is an example:
Screen Shot 2019-08-21 at 12 27 15 PM

### DIFF
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -560,6 +560,22 @@ def edit_finding(request, fid):
                 new_finding.false_p = False
                 new_finding.mitigated = None
                 new_finding.mitigated_by = None
+            if new_finding.duplicate:
+                new_finding.duplicate = True
+                new_finding.active = False
+                new_finding.verified = False
+                parent_find_string = request.POST.get('duplicate_choice', '')
+                if parent_find_string:
+                    parent_find = Finding.objects.get(id=int(parent_find_string.split(':')[0]))
+                    new_finding.duplicate_finding = parent_find
+                    parent_find.duplicate_list.add(new_finding)
+                    parent_find.found_by.add(new_finding.test.test_type)
+            if not new_finding.duplicate and new_finding.duplicate_finding:
+                parent_find = new_finding.duplicate_finding
+                if parent_find.found_by is not new_finding.found_by:
+                    parent_find.duplicate_list.remove(new_finding)
+                parent_find.found_by.remove(new_finding.test.test_type)
+                new_finding.duplicate_finding = None
 
             create_template = new_finding.is_template
             # always false now since this will be deprecated soon in favor of new Finding_Template model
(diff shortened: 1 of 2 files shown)


Maffooch:
@madchap Please check this out when convenient and provide feedback as necessary.

madchap:
Do you think it is possible and better eventually to have an autocomplete box instead? If thousands of findings, scrolling would definitely be a pain.

I saw this one: https://djangopackages.org/packages/p/django-autocomplete-light/ but I clearly have no experience or anything. Would it scale? be practical?

Should there be some kind of cascading approach eventually, to scope down to a product or engagement, prior to having the list of findings to choose from?

Maffooch:
The auto complete box would be cool. I worry about how complete it will get though. Im sure the motivation behind this enhancement was because central dedupe is very strict now. Thats the main reason I decided to only filter findings that had a matching title. 

Maybe a filter system could be implemented to allow users to decided which portions of dedupe they would like to use?

madchap:
Yea, that could be a good idea.

Your initial PR is not bad right. It could very well be a first iteration to get that feature out the door, until we get more feedback too. I'd already be happy with that :)

Maffooch:
Maybe to make it a bit more selective, descriptions would have to match as well. That would include all the non problem areas of dedupe. Almost positive there would not be thousands of findings listed that way ;)

madchap:
I am somewhat less sure about the description being a hard part of the filtering. To me, the chances the description gets an amend is higher, and would break the filtering. I would probably just then leave the title, which per your initial thought, is certainly enough.

By filter system, you mean making an option in the system settings to pick what makes up the filter right?

madchap:
Just tried the patch onto my local deployment, seems to work well.

Maffooch:
Glad to hear it! Think it would be best to call this a first iteration like you suggested and revisit later.

madchap:
> Glad to hear it! Think it would be best to call this a first iteration like you suggested and revisit later.

Sounds good to me. Do you want to keep the description in the filter though? Or remove it and keep only the title?

Also there seems to be a trailing slash showing up after the date for some reason.

Maffooch:
Decided to keep only title. 

The test the finding was associated was suppose to display, but the title is allowed to be Null. Decided to just scrap it. Something else can be added in its place in the next iteration.

madchap:
Using it now in semi-prod, I think we would benefit from adding the "test name" in the dropdown box, I.e. I have several matches, and the only differentiator is the finding ID.

Maffooch:
> The test the finding was associated was supposed to display, but the title is allowed to be Null. Decided to just scrap it. Something else can be added in its place in the next iteration.
  
Ill see what I can come up with.

madchap:
Working just peachy :+1:

madchap:
Should there be an accompanying change to the PATCH method of api v2?

madchap:
Sounds good.

Issue #1505, opened 2019-08-19
Image uploaded is 404


madchap:
**Bug description**
After uploading an image, it cannot be displayed because it is not found.

![image](https://user-images.githubusercontent.com/5468769/63258699-ee39a800-c27c-11e9-943e-5188ce95b5ee.png)

PR #1049, opened 2019-04-12
Fix markdown_render function

dr3dd589:
Fix markdown_render function that executing markdown javascript payload.
eg : `[xss](javascript:window.onerror=alert;throw%20document.cookie)`

- [x] Your code is flake8 compliant (DefectDojo's code isn't currently flake8 compliant, but we're trying to correct that.)
- [ ] If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR.
- [ ] Model changes should include the necessary migrations in the dojo/dd_migrations folder.
- [ ] Add applicable tests to the unit tests.

PR #1524, opened 2019-08-26
Expand default bleach whitelist

no-sec-marko:
The PR #1049 has broken the markdown rendering. After adding some multi-line code, the UI was not able to display the correct representation of the markdown input (see screenshots below).
  
markdown_bug_1
markdown_bug_2

This PR takes the default bleach_whitelist tags from [bleach_whitelist.py](https://github.com/yourcelf/bleach-whitelist/blob/master/bleach_whitelist/bleach_whitelist.py) and extends the "safe" html tags with the required tags to support all available markdown functions (see comments). Now, all dangerous html tags or attributes will be filtered/removed and the markdown render works as expected (see screenshots below).

markdown_bug_3
markdown_bug_4

If the additional html tags should not be used in defect dojo, it should be considered to remove the code highlighting or table feature.

### DIFF
--- a/dojo/templatetags/display_tags.py
+++ b/dojo/templatetags/display_tags.py
@@ -19,7 +19,6 @@ import datetime
 from ast import literal_eval
 from urllib.parse import urlparse
 import bleach
-from bleach_whitelist import markdown_tags, markdown_attrs
 import git
 
 register = template.Library()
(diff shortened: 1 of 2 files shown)


madchap:
Should fix https://github.com/DefectDojo/django-DefectDojo/issues/1505

PR #1530, opened 2019-08-27
add missing volumes in docker dev mode

ptrovatelli:
In docker dev mode, the volumes for the initializer and celerybeat containers are missing. 
- For the initializer it's important for db_migrations: without this volumes, we still needed to rebuild the django image after each change, even in dev mode (because the db_migrations are done in the initializer container, not uwsgi which already has this volume)
- add also celerybeat for completeness

'+ fixing a typo in Dockerfile.django


ptrovatelli:
@DefectDojo/defectdojo-moderators this one's a quicky. ok to merge?

Issue #1500, opened 2019-08-16
Make the deduplication configurable per parser


ptrovatelli:
As discussed a little bit in https://github.com/DefectDojo/django-DefectDojo/issues/1478, I believe it would be useful to have a configurable hash_code computation per parser, for the deduplication algorithm. 
In the past month, the deduplication was changed several times, with a mix of changing the way of computing the hash_code and (more often) specific rules addded in the deduplication algorithm itlsef, often breaking the deduplication for some parsers.
I'm hoping we could put in place a basic deduplication algorithm, based on hash_code only: 
- If the hash_code is the same, the vulnerability is a duplicated
- else, it's not

The complexity itself would reside in the hash_code computation. This would be done outside of the parser (as it is now).
In order to satisfy all use cases, it would be useful to make the set of fields used in the hash_code computation configurable, in settings.py for example. 
I'm thinking of 2 levels of configuration: 
- a configuration per family of parsers (+ for each parser a configuration to state in which family they are, or hardcoded?)
- an (optional) configuration for each parser, to allow specificities for each parser

The families of parsers that I know of at this point (please correct me / complete the list): 
- dynamic parsers (ex: ZAP)
  - hash_code should be computed based on title, endpoints, CWE
- static parser (ex: sonar, checkmarx)
  - hash_code should be computed based on  title, cwe, file_path, line number (or not), cwe
- dependency analysis (ex dependency check)
  - hash_code should be computed based on title, cwe, file_path (or CVE and Library, to be determined)

Another use case that this logic could handle is a tunable / better precision for static parsers.
Including the line number in the dedup fields for static parsers make tracking a unique vuln very weak: say we have a vuln at line 100, before the next scan we have added some code before line 100 so now our vuln is at line 101 -> it will not be detected as a duplicate but as a new vulnerability. 
Some people may want a more "aggressive" dedup system in order to attach information to a finding and making sure it will persist across scan (they will want to remove the line_number in the computation), other people may want a more exact count of findings. Thus the necessity of a configurable solution
Also some SAST tools have already advanced techniques to track unique vulnerability even after code change, using data flow graphs. Checkmarx for example can do this. I believe one of the fields `pathId` or `SimilarityId` can be used to track unique vulnerabilities. We could update the parser to extract those fields and put them in the dedup key in order to have a unique vuln tracking as clever as checkmarx has (when analysing checkmarx reports)
[...]

PR #1540, opened 2019-09-05
WIP Deduplication configuration per parser

ptrovatelli:
First batch regarding https://github.com/DefectDojo/django-DefectDojo/issues/1500
It's ready for review

It conforms mostly to what's described there except: 
- I haven't made a configuration per family of parser (I don't think it's required right now) but just per parser

I have some unit tests to add regarding deduplication, but as travis is not running them right now, I can look at this at a later time (see https://github.com/DefectDojo/django-DefectDojo/issues/1492  and https://github.com/DefectDojo/django-DefectDojo/pull/1501 )

Detailed changes: 
- checkmarx:
  - implemented a new "Checkmarx Scan detailed" scanner without aggregation so that the count of vuln is the same in defectDojo and in checkmarx so that we can use checkmarx technical id to track a vuln across different checkmarx scans.
  - added missing source and sink information in the model so that we know in defectDojo the difference between those previously aggregated findings. we can see those in the finding view (only for checkmarx non aggregated finding)
  - default "Checkmarx Scan" remains the lecacy scanner
- Sonar
  - added the import of "key" which is sonar technical id (use the latest version of sonar html import, here https://github.com/soprasteria/sonar-report/pull/8 (version 1.1.0) ) .
  - remove component name in filepath: 
    - before: `tomcat_180410:modules/.../DataSourceFactory.java`
    - now: `modules/.../DataSourceFactory.java`
- configuration
  in settings.py, for each parser you can choose a deduplication algorithm using the DEDUPLICATION_ALGORITHM_PER_PARSER list:
    - DEDUPE_ALGO_LEGACY : this is the default (will apply if no configuration is found). It's the current behavior
    - DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL : only for checkmarx and sonar for now. This makes sure that we will correcly track a given vuln even if the line number changes due to code change. This is a "same parser deduplication" algorithm
    - DEDUPE_ALGO_HASH_CODE : dedup on hash_code only. This is a "cross parsers deduplication" algorithm (can detect vulns added by different parsers)
    - DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE: this is the least tested thing as I haven't found a case when it's useful yet :) . It gives the possibility to deduplicate both on a technical id (same parser dedupe) and on functional information using the hash_code (different parser dedupe). This is done with a unique query in the DB using a python "Q". It could be useful for example if you track checkmarx vuln using "unique_id_from_tool" and also would like to deduplicate it with sonar. However, according to my tests, the hash_code always differ between sonar and checkmarx so dedup on unique_id_from_tool gives the same result right now


ptrovatelli:
took a deep dive in the checkmarx legacy scanner
I have found the below discrepancies that I fixed in commits 32a84e46a9bc96570f51c17f8995182c97f7c4b7 and ebc0734aa130a43091365828e2e6d0af08e950ee
- aggregation is on filename only, not filename + line number. See:
 `dupe_key = "{}{}{}{}".format(categories, cwe, name, result.get('FileName').encode('utf-8'))` 
(that i have renamed as  `aggregateKeys `)
So it actually didn't make a sense to keep a random line number from one of the finding (it was the line number from the first finding of the aggregate): most of the time, there are several line number, from all the aggregated vulns present in the given file with same cwe. I have removed the updation of line number in the finding for this scanner. 
- aggregation was on source filename but what we put in the finding is the sink filename. This could lead to incorrecly aggregated information -> I have made the aggregation on sink filename and keep also sink filename in the finding "file_path"
- the description was giving detailed data flow graph information for just one finding of the aggregate: same argument as for the line number: we must keep some information that is common to all the aggreated vunls, not just the first one. I have kept that detailed information for the "Checkmarx Scan detailed" but for the aggregated one I have instead put some information about all the source/sink filename, line number, object of the aggregate: 
```
            "Source filename: WebGoat/webgoat-lessons/missing-function-ac/src/main/java/org/owasp/webgoat/plugin/Users.java"
            "Source line number:  39"
            "Source object:  executeQuery"

            "Sink filename: WebGoat/webgoat-lessons/missing-function-ac/src/main/java/org/owasp/webgoat/plugin/Users.java"
            "Sink line number:  58"
            "Sink object:  allUsersMap",

-----------
(...)
```

I have also added a "nb_occurences" field in the finding so that there is a way to verify that we haven't missed any findings: adding all nb_occurences should give the total number of vulns in checkmarx. 

The good news is: without line number, the deduplication will work better: if the line number changes due to code change, identifying the finding is still possible. One would need to change the sink filename for the dedup to get lost. 

NB: I have set back the legacy scanner as default "Checkmarx Scan" in the last commit for better compatibility with existing systems

ptrovatelli:
I'm thinking we should go a little further and I would be as radical as stating that the aggregation keys should be the same as the deduplication keys, in order to have sane data. 
The issue when those differ is that: 
- say you aggregate on filename + title of the vulnerability
- say you deduplicate on filename + cwe
- say you have 2 times the same cwe in a file but with a different title

when importing the scan you'll import two unique vulnerabilities
when running the deduplication upon the next import, both vulnerabilities will appear as duplicate with one another. 
The solution to have consistent data is to enforce the same keys for aggregation and deduplication. 
So I'll try to align that hash_code computation with the aggregated keys, at least with the most sensible keys (i'm thinking filename + cwe or filename + CVE depending on the scanners) and for a couple of scanners
the problem also with finding title is that they tend to change when the tools version evolve, and for sure they don't work for cross-scanner deduplication.

ptrovatelli:
this is really coming into place :)
dedupe now works for dependency check. 
- before my last commit, that webgoat dependency check analysis imported twice found only 35 duplicates out of 75
- Now they are all correcly found
[dependency-check-report_webgoat.zip](https://github.com/DefectDojo/django-DefectDojo/files/3746010/dependency-check-report_webgoat.zip)

and as we use CVE it'll also work with other scanners or in case of dependency check evolution changing the title

ptrovatelli:
@DefectDojo/defectdojo-moderators I would really need a review on this in the next few days.
Note that most of the changes in that PR is deactivable by configuration. Keeping a previous settings.py will allow everyone to continue with the same hash_code / deduplication as before.
I have commited the advised configuration for a couple of scanners. The benefits from this PR can be applied to other parsers just by configuring them in settings.py

The only breaking change if I recall: 
- default "sonarQube scan" is now aggregated. the previous one is now "sonarqube scan detailed" with dedupe fixed thanks to the addition of sonarqube unique id

The default "checkmarx scan" was changed a little too only to fix a couple of discrepencies. 

I haven't been till benching yet, but another major benefit of deduplicating on a single field is that in the future we can simply add an index on that hash_code field if the database is too big to radically improve the performances.

I'll throw in some doc in the "readthedoc" after merging but basically everything is explained in the PR comments. 

I will enrich a bit more the "dedupe_unit_test.py" too but right now it's not run by travis anyway. 

Thanks!

ptrovatelli:
Note that for those who want to apply the new algorithm on previously imported findings, the hash_code of existing findings need to be recomputed. 
I just fixed the script that does it (it broke when we switched to django signals or to be exact, it still worked but displayed an error message due to the fact that we try to manually launch the deduplication but it's already happening upon finding save). 
This is the command to run this operation: 
```
docker exec -it django-defectdojo_uwsgi_1 bash
./manage.py dedupe
```

ptrovatelli:
@DefectDojo/defectdojo-moderators I have (again) merged dev to this branch; it's now conflictless; now would be a good time to have a look :)
btw this fixes https://github.com/DefectDojo/django-DefectDojo/issues/1654 and https://github.com/DefectDojo/django-DefectDojo/issues/1640, mostly by configuration, prooving that this PR is actually pretty helpful! 
thanks

dojo/models.py
```diff
@@ -1277,15 +1280,20 @@ class Finding(models.Model):
 
     line_number = models.CharField(null=True, blank=True, max_length=200,
                                    editable=False)  # Deprecated will be removed, use line
-    sourcefilepath = models.TextField(null=True, blank=True, editable=False)
+    sourcefilepath = models.TextField(null=True, blank=True, editable=False)  # Not used? to remove
```
madchap: Right, never seen it used either.

dojo/models.py
```diff
@@ -1294,6 +1302,16 @@ class Finding(models.Model):
...
     scanner_confidence = models.IntegerField(null=True, blank=True, default=None, editable=False, help_text="Confidence level of vulnerability which is supplied by the scannner.")
     sonarqube_issue = models.ForeignKey(Sonarqube_Issue, null=True, blank=True, help_text="SonarQube issue", on_delete=models.CASCADE)
+    unique_id_from_tool = models.CharField(null=True, blank=True, max_length=500, help_text="Vulnerability technical id from the source tool. Allows to track unique vulnerabilities")
+    sast_source_object = models.CharField(null=True, blank=True, max_length=500, help_text="Source object (variable, function...) of the attack vector")
+    sast_sink_object = models.CharField(null=True, blank=True, max_length=500, help_text="Sink object (variable, function...) of the attack vector")
+    sast_source_line = models.IntegerField(null=True, blank=True,
+                               verbose_name="Line number",
+                               help_text="Source line number of the attack vector")
+    sast_source_file_path = models.CharField(null=True, blank=True, max_length=4000, help_text="Source filepath of the attack vector")
```
madchap: I was wondering about the prefix `sast_` to these newly added fields. Something you expect only be used by SAST tools only? Or could it be made more generic?

ptrovatelli: it's filename and line number in code so for what other tools it could be used? maybe i forgot something but what is see is: 
- static analysis tools
- dynamic analysis
- dependency analysis (although it could be also seen as static)

dynamic and dependency analysis definitely won't have a line number and a source/sink filename

madchap:
Hi @ptrovatelli 

Looks like quite a bit of effort went in there.

I just spent some time just looking at the PR, but after fetching/merging latest `dev` and applying your PR on top, it failed with a bunch of errors?

Otherwise, this looks sane to me and can give the required flexibility when needed for special scanners, especially with the default behavior being the actual default to avoid bad surprises.

ptrovatelli:
Hi @madchap ; thanks for looking into it. 
this worked for me:
```
git checkout dev
git pull
# remove container and volume; beware that this will remove all data but sometimes required when going back an forth in django migrations
docker-compose down --volumes
git fetch origin pull/1540/head:pr-1540
git merge pr-1540
docker/setEnv.sh release
docker-compose build
docker-compose up
```

there are sometimes errors while the initializer is still running due to all the tables/data not being present yet, especially if you have windows open with the application (this is definitely a possible improvment in the application!) , but after the end of the initializer it should work. 

we can discuss your errors on slack if you're still having them.

devGregA:
@ptrovatelli let me see if I can get Aaron's eyes on this as well. I'm also reviewing.

devGregA:
I'm going to keep reviewing and make sure this is correct, but this appears to remove dedupe across a product, which would remove the original functionality. Since the hash_code method is being changed.

ptrovatelli:
@devGregA there's no reason it would prevent the dedupe product wide. For the scope of deduplication, I have kept the same logic as before; see utils.py. When using deduplication based on hash_code:
```
def deduplicate_hash_code(new_finding):
    if new_finding.test.engagement.deduplication_on_engagement:
        existing_findings = Finding.objects.filter(
            test__engagement=new_finding.test.engagement,
            hash_code=new_finding.hash_code).exclude(
                id=new_finding.id).exclude(
                    hash_code=None).exclude(
                        duplicate=True)
    else:
        existing_findings = Finding.objects.filter(
            test__engagement__product=new_finding.test.engagement.product,
            hash_code=new_finding.hash_code).exclude(
                id=new_finding.id).exclude(
                    hash_code=None).exclude(
                        duplicate=True)
```

In the `else` part is the dedupe product-wide, get all findings from the same product and check if they have the same hash_code.

ptrovatelli:
thanks for checking on this though :+1:

devGregA:
Thanks for the clarification. Will keep reviewing, back with you asap.

Greg Anderson
DefectDojo Project Leader
greg.anderson@owasp.org

devGregA:
@Maffooch please test with some example data and then let's touch base.

ptrovatelli:
just fixed the dedupe_unit_test.py which broke due to recent GUI change (and an error in one of the tests). 
- increased the window size: it was too small for the checkbox to delete all findings to appear
- get finding details based on xpath

ptrovatelli:
@madchap @Maffooch @devGregA where are we on this? 

just pushed a slight improvment in the dedup lifecycle: before the last commit, when creating a finding with state "duplicate", we didn't inherit the parent status, for example "false positive" status

It means that looking at the duplicate finding only, we don't know if it's a false positive or not, we need to check the parent status. 

When creating the finding, i have added inheritance of the parent attributes: 
- false_p
- out_of_scope
- under_defect_review

ptrovatelli:
created a new clean PR for all those sparse commits: 
https://github.com/DefectDojo/django-DefectDojo/pull/1665

PR #1444, opened 2019-08-01
Defect Dojo/SonarQube integration

twsagarcia:
**Note: DefectDojo is now on Python3 and Django 2.2.1 Please submit your pull requests to the 'dev' branch as the 'legacy-python2.7' branch is only for bug fixes. Any new features submitted to the legacy branch will be ignored and closed.**

When submitting a pull request, please make sure you have completed the following checklist:

- [x] Your code is flake8 compliant 
- [x] If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR.
- [x] Model changes must include the necessary migrations in the dojo/dd_migrations folder.
- [x] Add applicable tests to the unit tests.

PR #1552, opened 2019-09-12
Mark finding as inactive if specified by a parser

alles-klar:
If a finding is marked as active by a parser it overwrites the default value. This should also be possible if a finding is marked as deactivated or mitigated.

On behalf of DB Systel GmbH

---

- [] Add applicable tests to the unit tests.

### DIFF
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -544,8 +544,10 @@ def import_scan_results(request, eid=None, pid=None):
                     item.last_reviewed = timezone.now()
                     item.last_reviewed_by = request.user
                     if form.get_scan_type() != "Generic Findings Import":
-                        item.active = active
-                        item.verified = verified
+                        if not item.active:
+                            item.active = active
+                        if not item.verified:
+                            item.verified = verified
                     item.save(dedupe_option=False, false_history=True)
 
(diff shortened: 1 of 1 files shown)


alles-klar:
We have two default options the user can select: 
1. deactivate findings
2. activate  findings

If the user choose the second option and a parser marks a finding as activated, the finding will be activated.
But if the user choose the first option and a parser marks a finding as activated, the finding will be deactivated. For this option I aspect that the finding should be activated.

@Maffooch Makes that sense for you?

Maffooch:
Makes perfect sense. I was confused on why you were disabling the user selected active/non-active. That would mean if the scanner does not specify the active status, everything would remain inactive (default) unless explicitly made active by the user post import. Wouldn't it make more sense to disable that feature for a select few scanners that do provide that info instead of nuking it for all scanners?

alles-klar:
I think it would be more complex to add options to the scanners and the advantage is not very big. At the moment the scanner are more or less generic and it is very simple to implement one. 

For me it is okay, if the user choose the default option and the scanner can overwrite this.

Maffooch:
@alles-klar So I was looking at the conflict that came up and it looks like #1444 added a good solution to the issue here. It allows for specification of each parser that contains active/verified info. What do you think?

Issue #1560, opened 2019-09-18
Enhancement: Add OWASP Dependency Track Findings Importer


csansone-handy:
**Scanner Name**
OWASP Dependency Track

**Scanner Description**
Dependency Track is an intelligent Software Supply Chain Component Analysis platform that allows organizations to identify and reduce risk from the use of third-party and open source components. Dependency Track takes a unique and highly beneficial approach by leveraging the capabilities of Software Bill-of-Materials (SBoM) which provides capabilities that traditional Software Composition Analysis (SCA) solutions cannot achieve.

**Scanner Link**
https://dependencytrack.org/

**Sample File**
_Note: The information in this section has been copied from https://docs.dependencytrack.org/integrations/file-formats/ for purposes of documenting this enhancement in a complete manner and in case the Dependency Track documentation changes_

Dependency Track has an API-first design. API’s are simply at the heart of the platform. However, there are occasions when a file-based format is desired for integration with other systems, especially legacy systems or systems that are not API-aware or compatible.

Dependency Track has a native format that can be used to share findings with other systems. The findings contain identical information as presented while auditing, but also include information about the project and the system that created the file. The file type is called Finding Packaging Format (FPF).

FPF’s are json files and have the following sections:

| Name | Type | Description |
| ------- | ------ | ------------ |
| version | string | The Finding Packaging Format document version |
| meta | object | Describes the Dependency-Track instance that created the file |
| project | object | The project the findings are associated with |
| findings | array | An array of zero or more findings |

An example FPF file looks like this:
```
{
  "version": "1.0",
  "meta" : {
    "application": "Dependency-Track",
    "version": "3.4.0",
    "timestamp": "2018-11-18T23:31:42Z",
    "baseUrl": "http://dtrack.example.org"
  },
  "project" : {
    "uuid": "ca4f2da9-0fad-4a13-92d7-f627f3168a56",
    "name": "Acme Example",
    "version": "1.0",
    "description": "A sample application"
  },
  "findings" : [
  {
    "component": {
      "uuid": "b815b581-fec1-4374-a871-68862a8f8d52",
      "name": "timespan",
      "version": "2.3.0",
      "purl": "pkg:npm/timespan@2.3.0"
    },
    "vulnerability": {
      "uuid": "115b80bb-46c4-41d1-9f10-8a175d4abb46",
      "source": "NPM",
      "vulnId": "533",
      "title": "Regular Expression Denial of Service",
      "subtitle": "timespan",
[...]

PR #1561, opened 2019-09-18
Add ability to import OWASP Dependency Track Finding Packaging Format (FPF) Exports as a scan

csansone-handy:
This PR is for issue #1560 and adds the ability to import OWASP Dependency Track Finding Packaging Format (FPF) Exports as a scan.


csansone-handy:
Looking for some feedback from @stevespringett (core contributor/developer for Dependency Track) on this PR. Mainly need his thoughts on the `dojo/tools/dependency_track/parser.py` file as that is the core of where I am parsing a Dependency Track FPF file and making some (hopefully correct) assumptions on the format of each object in the findings array

dojo/tools/dependency_track/parser.py
```diff
@@ -0,0 +1,202 @@
...
+            return "Critical"
+        elif severity == "high":
+            return "High"
+        elif severity == "medium":
+            return "Medium"
+        elif severity == "low":
+            return "Low"
+        elif severity.startswith("info"):
+            return "Informational"
+        else:
```
stevespringett: Dependency-Track also supports 'UNASSIGNED' as a severity. 
https://github.com/DependencyTrack/dependency-track/blob/master/src/main/java/org/dependencytrack/model/Severity.java

This typically occurs when there's an incomplete CVE and OSS Index (or another source) picked it up. It's usually a confirmed vulnerability, but the severity is unknown as it was never formally rated.

csansone-handy: I believe DefectDojo only has [Critical, High, Medium, Low, Informational] so we would need to map to one of those. Here is where I default to a Critical severity if we were unable to map a Dependency Track severity to a DefectDojo severity: https://github.com/DefectDojo/django-DefectDojo/pull/1561/files#diff-0c6cbdfc684bc0d0f4300320b93fcadcR151

dojo/tools/dependency_track/parser.py
```diff
@@ -0,0 +1,202 @@
...
+            raise Exception("Missing 'vulnerability' node from finding!")
+        if 'vulnId' not in dependency_track_finding['vulnerability']:
+            raise Exception("Missing 'vulnId' node from vulnerability!")
+        vuln_id = dependency_track_finding['vulnerability']['vulnId']
+        if 'source' not in dependency_track_finding['vulnerability']:
+            raise Exception("Missing 'source' node from vulnerability!")
+        source = dependency_track_finding['vulnerability']['source']
+        if 'component' not in dependency_track_finding:
+            raise Exception("Missing 'component' node from finding!")
+        if 'purl' not in dependency_track_finding['component']:
```
stevespringett: Not all findings will have a purl. If a component is of type (operating system or hardware device), these components will not have a purl. Software components without a purl are also likely and may involve unmanaged dependencies, dependencies that were manually added, file components.

csansone-handy: Hmm interesting. Good to know! Any thoughts on how we should build the title/description of a finding in DefectDojo given the data from Dependency Track. Are there attributes of the component that are guaranteed to be there (e.g. name, version, etc.)?

stevespringett: Component name is the only field that is required. You can use the UUID of the vulnerability to make an additional query to retrieve the full description, recommendation, credits, etc. 

The vulnerability in a finding will always include uuid, source, and vulnid, and may optionally include a title and subtitle.

csansone-handy: The way DefectDojo seems to work is you upload a file (either manually in the UI or via the API) that contains the scan results. This would need to be self contained and parsed for all of the needed info.

Do you have some example FPF JSON payloads you can share that show some of these things:

Component:
- No purl
- Only component name (no uid/version)

Vulnerability:
- has a title and/or subtitle

csansone-handy: Dependency Track seems to indicate that the `version` of a component is required (as indicated by the asterisk *).

image

As such why would the version not come back in the FPF for a Component?

stevespringett: Components will also have a uuid as well. That's a requirement. But version is optional.

Also note that the findings API is just one way to get the results. The vulnerability api may also be used, but you won't get the component or analysis.

FYI, ThreadFix uses the vulnerability API, Kenna Security uses the Findings API that is then converted to KDI prior to publish.

stevespringett: Wait hold. On. The vulnerability description and recommendation WILL be populated in the findings results if those values are non-null.

stevespringett: if the case of CVE, this means that a vulnerabiliy will generally have a description. In the case of NPM Audit findings, the vuln will typically have a description and recommendation.

stevespringett: The 'version' is not a requirement in the data model. There might be some business logic enforcement to ensure version is there, but to be safe, I would account for instances where version is null

csansone-handy: Okay I made a few updates to accommodate for these optional fields from Dependency Track. Let me know what you think. Diff is here: https://github.com/DefectDojo/django-DefectDojo/pull/1561/commits/ff1fb01da3d86f3f4840e30c814a4b8d1999c98f

Maffooch:
Good wok here! The only thing I am not totally crazy about is the naming convention for each individual finding. Each one looks exactly the same at first glance and are super long.

Additionally, could you add the sample scan files [here](https://github.com/DefectDojo/sample-scan-files), and add it to list of scanners [here](https://github.com/DefectDojo/Documentation/blob/master/docs/integrations.rst). Thanks!

stevespringett:
Looks good. I didn't see anything obvious from a Dependency-Track perspective.

csansone-handy:
> Good wok here! The only thing I am not totally crazy about is the naming convention for each individual finding. Each one looks exactly the same at first glance and are super long.
> 
> Additionally, could you add the sample scan files [here](https://github.com/DefectDojo/sample-scan-files), and add it to list of scanners [here](https://github.com/DefectDojo/Documentation/blob/master/docs/integrations.rst). Thanks!

@Maffooch - thanks for the feedback.

Regarding the naming convention of the findings: If you look at the discussion I had with Steve above the `title` field of a vulnerability node is optional. We _could_ use the vulnerability title if it is present and make that the title of the DefectDojo finding....but then we would have cases where there is no title so we would need to fall back to what I am doing now which is making a generic title. I think I would rather have them all be consistently named.

Regarding adding the scan sample and list of scanners: will do! Thanks for pointing that out

csansone-handy:
Added the scan sample here: https://github.com/DefectDojo/sample-scan-files/pull/37
Added this to the list of scanners here: https://github.com/DefectDojo/Documentation/pull/64

dojo/fixtures/test_type.json
```diff
@@ -492,14 +492,20 @@
...
+	},
+	{
+		"fields": {
+			"name": "Dependency Track Finding Packaging Format (FPF) Export"
+		},
+		"model": "dojo.test_type",
+		"pk": 164
 
-	}
+	},
```
csansone-handy: @Maffooch thanks for fixing my botched attempt at resolving the merge conflict. Every time I tried to pull or push it said everything was up to date so I am not sure what was going on. There is a trailing comma here that might cause some issues though!

Maffooch: You're welcome! It can be a bit wonky sometimes. The trailing comma can be there or not in this case. May as well save the next person a keystroke :)

entrypoint_scripts/common/common-os.sh
```diff
@@ -89,7 +89,7 @@ bootstrap_install() {
 	   # Install user isn't root,  check for sudo privileges
 	   echo "  Checking for sudo access, you may be prompted for your password"
        sudo -v 2>/dev/null
-       SUDO_CHECK=`sudo -v | wc -l`
+       SUDO_CHECK=`sudo -v | wc -l | awk '{$1=$1; print}'`
```
devGregA: @csansone-handy was this change intended?

csansone-handy: Yes it was. I was trying to run something locally (macOS Mojave) and I was having issues. Some versions of `wc -l` will emit a tab in the front so you will get `<tab>0` instead of `0`. The  `| awk '{$1=$1; print}'` just trims the whitespace

csansone-handy: For example:

```
$ uname -v                                                                                                                                                                                            
Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64
$ cat ~/somefile | wc -l                                                                                                                                               
     559
```
vs.

```
$ uname -a
Linux 232842ec56f5 4.9.184-linuxkit #1 SMP Tue Jul 2 22:58:16 UTC 2019 x86_64 GNU/Linux
$ cat manage.py | wc -l
10
```

PR #1564, opened 2019-09-23
fix introduced regression to ingesting Xray scan

madchap:
Signed-off-by: Fred Blaise <fred.blaise@protonmail.com>

Found out some space removal introduced a month ago during some "cleanup", breaking XRay ingestion. Fixing these here.

This broke the intake with error `Unknown Test Type`.

### DIFF
--- a/dojo/fixtures/test_type.json
+++ b/dojo/fixtures/test_type.json
@@ -481,7 +481,7 @@
 	},
 	{
 	   "fields": {
-	      "name": "JFrogXray Scan"
+	      "name": "JFrog Xray Scan"
 		},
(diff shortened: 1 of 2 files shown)


valentijnscholten:
Looks like it was introduced in https://github.com/DefectDojo/django-DefectDojo/commit/8c4fe35865f5eff4741fadce0063038c41ad26d9 ?

madchap:
> Looks like it was introduced in [8c4fe35](https://github.com/DefectDojo/django-DefectDojo/commit/8c4fe35865f5eff4741fadce0063038c41ad26d9) ?

It was..

PR #1586, opened 2019-10-08
Introducing DataTables to Products and Findings

dougmorato:
**Introducing DataTables to Products and Findings**
Removing DataTables from node package as it is not working, but adding it as static files.
Added DataTables to Products and Findings HTML pages to support better sorting, filtering and exporting.

**Separating product tags into their own column:**
Product tags were added just to the right of the product name. this made very difficult to properly export as CSV or Excel in the upcoming DataTables effort, so splitting the tags which provide relevant information in their own column was the most efficient route to support DataTables export


Maffooch:
This is a fantastic feature! 

Couple of nitpicky requests if you're up for It:
1. Remove sorting abilities from the check boxes and the drop down menu.
2. Change the name of the checkboxes from 'Info low medium high critical all none' to something more intuitive and shorter.
3. Restrict the viewing window to non scrollable. After hiding/showing an item, word wrapping is not enabled and allows the scroll area to get very large.

dougmorato:
@Maffooch 
> 1. Remove sorting abilities from the checkboxes and the drop-down menu.
Addressed, no longer sorting on checkbox and action dropdown

> 2. Change the name of the checkboxes from 'Info low medium high critical all none' to something more intuitive and shorter.
I finally figured out how to NOT select those fields without having to list every single column. So, checkbox and action menu no longer exported

> 3. Restrict the viewing window to non-scrollable. After hiding/showing an item, word wrapping is not enabled and allows the scroll area to get very large.
Not sure if I understand you. When you click on any of the column visibility to "hide" or "appear" a column, as soon as you click away on any other place of the screen the datable re-adjusts to use the full width without any scrolling bars. If that's not the behavior you are seeing, a picture is worth a thousand words.... please send a screenshot on slack @dougmorato 

Commit addressing points 1 and 2 has been submitted for review

PR #1587, opened 2019-10-08
Adding the All Engagements View

dougmorato:
**Adding All Engagements**
Dojo menu had a "All Engagements" entry, but it opened Active Engagements only. This change still respects the Active Engagements but introduces the All Engagements view

**DataTables now available on All Engagements**
Introducing support for DataTables for all engagements view

### DIFF
--- a/dojo/engagement/urls.py
+++ b/dojo/engagement/urls.py
@@ -7,6 +7,7 @@ urlpatterns = [
     url(r'^calendar$', views.engagement_calendar, name='calendar'),
     url(r'^calendar/engagements$', views.engagement_calendar, name='engagement_calendar'),
     url(r'^engagement$', views.engagement, name='engagement'),
+    url(r'^engagements_all$', views.engagements_all, name='engagements_all'),
     url(r'^engagement/new$', views.new_engagement, name='new_eng'),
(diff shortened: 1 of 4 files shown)


dougmorato:
Please do not merge this PR until https://github.com/DefectDojo/django-DefectDojo/pull/1586 is merged. This change depends on #1586

Maffooch:
Similar feedback here as with #1586. The old All Engagements button bugged me as well.

dougmorato:
@Maffooch 
> 1. Remove sorting abilities from the checkboxes and the drop-down menu.

Addressed, no longer sorting on the "+add engagement" button

> 2. Change the name of the checkboxes from 'Info low medium high critical all none' to something more intuitive and shorter.

"+add engagement" column no longer being exported

> 3. Restrict the viewing window to non-scrollable. After hiding/showing an item, word wrapping is not enabled and allows the scroll area to get very large.

Same as #1586 . 
Not sure if I understand you. When you click on any of the column visibility to "hide" or "appear" a column, as soon as you click away on any other place of the screen the datable re-adjusts to use the full width without any scrolling bars. If that's not the behavior you are seeing, a picture is worth a thousand words.... please send a screenshot on slack @dougmorato 

Commit addressing points 1 and 2 has been submitted for review

Issue #1594, opened 2019-10-11
Add a model for CVEs and generic vulnerability databases


jvz:
**Is your feature request related to a problem? Please describe**

While working on a solution for https://github.com/DefectDojo/django-DefectDojo/issues/1119, I figured it would make sense to cache the CVE metadata in the database.

**Describe the solution you'd like**

As a user, I want my security scan uploads to be matched against an internal CVE database to ensure that findings contain the CWE as well as allowing for findings that reference the same CVE to be logically related. This internal CVE database should be configurable with a git repository which is imported and updated periodically. While I only need to support CVEs, this should be extensible to support other vulnerability databases that may be specific to individual security products.

**Describe alternatives you've considered**

Making calls to an API that serves CVEs could potentially be easier to use, but the ability to run DefectDojo offline prevents that.

**Additional context**

As of the date of filing, the git repository for https://github.com/CVEProject/cvelist is about 962 MB (including both the working directory and the .git/ directory). As such, I'd imagine the disk usage from importing to various databases to be comparable (i.e., less than a gigabyte). Due to this, this feature should likely be opt-in rather than enabled by default.

madchap:
For the offline issue, maybe another feature would be required to import a zip file or something. Then, for this feature, the defacto would be to just be online and perform queries?

Do we have any idea on how many installations of DD are online vs offline?

jvz:
Using an input source other than a git repository certainly helps inform how to structure the code, though that's not too big a deal to change in this project.

jvz:
So I noticed that there's a `CWE` model that isn't used much yet. I can likely link into that table for vulnerabilities with a CWE which provides a richer data model.

jvz:
Reopen?

Issue #1119, opened 2019-05-10
Automatically fill in CWE


madchap:
**Is your feature request related to a problem? Please describe**
Example: Scans import do not necessarily include the CWE related to the CVE reported, leaving the analyst looking it up and filling it manually (unless I missed something).

**Describe the solution you'd like**
If a finding has a CVE number, attempt to find the related CWE by querying a site like cvedetails.com and update automatically.

**Describe alternatives you've considered**
None so far.

**Additional context**
Add any other context, screenshots, sketch, code snippet, etc. about the feature request here.

valentijnscholten:
I've been thinking about the same when the severity is missing. There's a github repo containing all CVEs in json format that might be helpful: https://github.com/olbat/nvdcve

ptrovatelli:
could be nice but beware that defectDojo should still work in an isolated environment:
- add a configuration to allow offline working (should be offline by default to avoid problems)
- handle a default list of cwe that's commited in DD's github repo so that updating defectDojo will transparently update the known cwe. Ideally automate the updation of this list maybe with travis?

jvz:
In order to cache or use the CVE data offline, perhaps we can introduce a new model for it? This could be useful for linking metadata about a CVE to findings from https://github.com/CVEProject/cvelist directly. That seems like a lot of data to add to this repository directly, though if you're suggesting a cache of mappings between CVEs and CWEs, then that might be more feasible.

madchap:
Keeping it open.

PR #1602, opened 2019-10-17
Add initial support for syncing CVE lists

jvz:
This adds the initial support for synchronizing external CVE
and vulnerability lists for use with linking findings to
external vulnerabilities.

This addresses #1594 and eventually #1119.

Usage notes:

0. Enable feature in system settings
1. Run `python manage.py sync_vulnerability_mirror` to force an initial synchronization (default is hourly which seems to begin an hour after app startup).
2. When new findings are added or existing ones are updated, their mappings to vulnerabilities are updated. If a finding did not have a CWE but did have a CVE, the CWE is filled in from the vulnerability's CWE.

@reviewbybees @madchap 


dojo/tasks.py
```diff
@@ -310,3 +314,45 @@ def async_dupe_delete(*args, **kwargs):
...
+
+    # FIXME: this should check the updated time of the state to avoid re-processing in-progress data
+    last_processed_revision = VulnerabilityMirrorState.get_last_processed_revision(settings.VULNDB_URL)
+    mirror = VulnerabilityMirror(logger=logger)
+    checkpoint_revision = mirror.revision
+    header = chunk_task(process_new_vulnerability_files, mirror.added_files(last_processed_revision), 2000)
+    if last_processed_revision is not None and len(last_processed_revision) > 0:
+        header.extend(
+            chunk_task(process_updated_vulnerability_files, mirror.updated_files(last_processed_revision), 2000))
+    chord(header)(checkpoint_vulnerability_state.si(checkpoint_revision))
```
jvz: I'll note that while the checkpointing won't be done until after the files are processed, if this task is invoked again while the first invocation is still running, you'll end up with two parallel lists of the same files to be imported which will log a lot of uniqueness constraint errors. I'm thinking of using a timestamp in the state model here to ensure that only one task of this type is running at a time regardless if this top level task finished (this task doesn't wait for the remaining tasks to complete, so rate limiting doesn't help much here in practice).

dojo/models.py
```diff
@@ -1272,6 +1272,8 @@ class Finding(models.Model):
     jira_change = models.DateTimeField(editable=True, null=True)
     scanner_confidence = models.IntegerField(null=True, blank=True, default=None, editable=False, help_text="Confidence level of vulnerability which is supplied by the scannner.")
     sonarqube_issue = models.ForeignKey(Sonarqube_Issue, null=True, blank=True, help_text="SonarQube issue", on_delete=models.CASCADE)
+    vulnerabilities = models.ManyToManyField(to='Vulnerability', related_name='findings',
```
jvz: As I haven't linked this data in yet, some tests are failing due to a missing value for this field.

madchap:
@Maffooch @valentijnscholten @devGregA would you have some early feedback on this in order to make sure we'd have this going into the right direction? Thanks for the advices.

jvz:
I had a short lived idea to import a finding's title/description/URL/etc. into the `Vulnerability` model when the issue doesn't exist yet, though that would conflict with the `bulk_create` aspect of importing unprocessed CVEs upstream. Ideally, this vulnerability database would be more of a simple key/value store as the transaction guarantees required here are fairly weak (i.e., I generally only care about receiving the latest value for my key; I don't care about conflicts when I go to create/update a vulnerability since I'm saving the full row regardless), though it was far out of scope in this PR to introduce more infrastructure.

dojo/vulnerability/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+        git_dir = self._repository_root / '.git'
+        if not git_dir.exists():
+            self._logger.debug('Cloning vulnerability mirror from %s to %s', remote_url, repository_root)
+            self._repo = self.__clone()
+        else:
+            self._logger.debug('Reusing existing checkout at %s', repository_root)
+            self._repo = self.__update()
+
+    def __clone(self) -> git.Repo:
+        return git.Repo.clone_from(self._remote_url, self._repository_root, depth=1)
```
jvz: This significantly speeds up the initial clone, but now this can introduce a potential issue where the last processed commit is older than the clone on the worker.

madchap:
@aaronweaver what are your thoughts on this?

dojo/db_migrations/0021_vulnerability.py
```diff
@@ -0,0 +1,47 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='Vulnerability',
+            fields=[
+                ('vulnerability_id', models.CharField(help_text='Vulnerability id such as CVE-2000-0001.', max_length=200, primary_key=True, serialize=False)),
+                ('url', models.URLField(help_text='URL for more details about this vulnerability.')),
+                ('title', models.CharField(help_text='Title or summary of this vulnerability.', max_length=1000)),
+                ('description', models.TextField(blank=True, help_text='Further details of this vulnerability.')),
+                ('cwe', models.PositiveSmallIntegerField(help_text='Common weakness enumeration this vulnerability is categorized under.', null=True)),
```
madchap: Given that some tools now give propose CWE as an array, e.g. a vulnerability could link to several CWEs, would we want to extend that directly? (probably another PR though)

madchap: @ptrovatelli ^

madchap:
Hey @jvz 

Anything you can do to address @Maffooch 's concerns, do they make sense?

dojo/tasks.py
```diff
@@ -313,3 +317,40 @@ def async_dupe_delete(*args, **kwargs):
 @task(name='celery_status', ignore_result=False)
 def celery_status():
     return True
+
+
+@app.task(ignore_result=True)
+def synchronize_vulnerability_mirrors():
+    s = System_Settings.objects.get()
```
jvz: Can use `dojo.utils.get_system_setting('setting_name')`

jvz:
>     1. Whenever a finding is edited, the entire CVE database is shown in as selectable scroll menu. This results in a lot of wait time for the edit finding page to load. It can be removed by excluding it from the finding form.

Fixed that part in the findings forms.

jvz:
Does this need a `Serializer` class in `api_v2`?

Maffooch:
Thanks for removing the list from the form. Are you wanting to update the list via an API call? That would require a serializer.

dojo/models.py
```diff
@@ -2087,6 +2100,50 @@ def __str__(self):
         return self.product.name + " | " + self.project_key
 
 
+class Vulnerability(models.Model):
```
jvz: It may be useful to add the CVSS vector as suggested in #1887.

jvz:
> Thanks for removing the list from the form. Are you wanting to update the list via an API call? That would require a serializer.

No, that's not needed at this time.

valentijnscholten:
There are new migrations in `dev`, could you please update this PR accordingly?

NRGLine4Sec:
Any update about this PR ?

pmadhyasta:
Is this feature merged to any specific branch? I am unable to find any related feature to enable in settings in 1.7.0?

valentijnscholten:
It's not merged yet, I have removed the 1.7.0 label that was still attached incorrectly.

NRGLine4Sec:
@valentijnscholten Maybe you can re-open the PR ?
Because, I think this PR is the first step to get the feature I required in https://github.com/DefectDojo/django-DefectDojo/issues/2433.
But maybe not ?

PR #1607, opened 2019-10-18
Adding defectDojo_engagement_survey to Installed Apps

dougmorato:
After the merge of PR # 1601, the dev branch was throwing the error specified below. 

`RuntimeError: Model class defectDojo_engagement_survey.models.Question doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS. error 
`

This commits addresses the error by adding the engagement_survey  to installed apps.

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -436,6 +436,7 @@ INSTALLED_APPS = (
     'django.contrib.sites',
     'django.contrib.messages',
     'django.contrib.staticfiles',
+    'defectDojo_engagement_survey',
     'polymorphic',  # provides admin templates
     'overextends',
     'django.contrib.admin',


dougmorato:
@Maffooch Please check if I am doing this correctly.

Maffooch:
It’s already included on line 446. Not sure what the issue is. I’ll look into.

Maffooch:
I cant reproduce this issue. Don't know what the deal is. #1601 passed all the travis checks as well.

madchap:
I've the same issue after a merge from `dev`.

madchap:
I also ran into an issue today, running a `docker-compose build`, with the following message:

`django.core.exceptions.ImproperlyConfigured: Application labels aren't unique, duplicates: defectDojo_engagement_survey`

It seems it made it through another PR somehow... like @Maffooch 's 4 months ago? what..?

madchap:
So yea, something was/is weird. Had the issue in one env, but not the other.

Maffooch:
I have a theory here, but not sure if it's accurate. Pulling changes made to settings.py.dist do not update settings.py unless the entire file is copied over. If there was any sort of overwrite, the users personal settings in settings.py would be blown out. This explains why I couldn't reproduce the error with a fresh instance and why @dougmorato's fix was to include it as in app in settings.py. 

Whether this theory is correct or not, this pull request will not fix it because it's adding the same app twice a few lines apart.

PR #1624, opened 2019-10-27
Add nosniff header and other improvments related to installation

ptrovatelli:
…OWSER_XSS_FILTER in some files

- add header "X-Content-Type-Options: nosniff" with DD_SECURE_CONTENT_TYPE_NOSNIFF
- fix value of SECURE_BROWSER_XSS_FILTER : in some files it was "on" but should be "True" to be activated (sets header "X-XSS-Protection: 1; mode=block")

to test with docker, edit the docker-compose.yml (or an ovverride) for the uwsgi container, section environment) :
```
      DD_SECURE_BROWSER_XSS_FILTER: 'True'
      DD_SECURE_CONTENT_TYPE_NOSNIFF: 'True'
```


ptrovatelli:
I'v actually added the secure values as default. 
by the way I noticed that most parameters appear in configuration files as on/off whereas django actually expects True/False! 
see https://www.bedjango.com/blog/how-debug-django-useful-tips/
I've tried to fix everything; i can only guess that people have fixed it in their configuration files when they needed it but it's pretty error-prone
note that in python both "on" and "off" are truthies :)

ptrovatelli:
I have just pushed another change on the configuration: the parameters `DD_SECRET_KEY` and `DD_CREDENTIAL_AES_256_KEY` were hardcoded in the Dockerfile which means that they were fixed after image building (and not configurable at all other than changing the dockerfile) 
I have moved them to the `docker-compose.yml` so that users can configure them upon startup. 
It is advised to set them and keep them secret (not keeping the ones from github then)

more info: 
- SECRET_KEY
https://docs.djangoproject.com/en/2.2/ref/settings/
```
The secret key is used for:
All sessions if you are using any other session backend than django.contrib.sessions.backends.cache, or are using the default get_session_auth_hash().
All messages if you are using CookieStorage or FallbackStorage.
All PasswordResetView tokens.
Any usage of cryptographic signing, unless a different key is provided.

```

- DD_CREDENTIAL_AES_256_KEY
DD_CREDENTIAL_AES_256_KEY seems to be more specific to defectDojo cf readthedoc: 
```
DD_CREDENTIAL_AES_256_KEY
AES 256 key for encrypting sensitive data such as passwords in DefectDojo. Set to at least a 256-bit key and should be set to a unique, unpredictable value.
```

in both case they should be in the docker-compose so that we can generate them and put them in the environment when starting the application

devGregA:
@mtesauro would you review this please?

ptrovatelli:
@madchap variables in .env file will be picked up and used in the yml. 
this article explains the various ways to configure a docker-compose.yml: 
https://vsupalov.com/docker-arg-env-variable-guide/#the-dot-env-file-env

PR #1651, opened 2019-11-14
Repair Simple Metrics data displayed on /metrics/simple route

amnesik:
Hi !
Metrics on  /metrics/simple only return values from the day instead of the month (or the chosen month).
So i added in the djangoo orm filter the precision of a month instead of taking only one day.
I may recommend also to change the name of the variable "now" to "date_asked" or something like that, but it is your code :)
Very nice project !!!!


Maffooch:
Hello @amnesik and thank you for your contribution. Pull requests are only accepted to the `dev` branch, so please submit there instead.

PR #1664, opened 2019-11-25
drf_yasg for api documentation

alles-klar:
Using drf_yasg for the generation of swagger documentations allowed to use api generator like openapi-generator (https://github.com/OpenAPITools/openapi-generator). In first tests it worked very well and I am currently working on a api wrapper in python, generated by openapi-generator. 

Furthermore the api documentation contains much more details, e.g. return types. 

On behalf of DB Systel GmbH

---

### DIFF
--- a/dojo/urls.py
+++ b/dojo/urls.py
@@ -2,11 +2,13 @@ from django.conf import settings
 from django.conf.urls import include, url
 from django.conf.urls.static import static
 from django.contrib import admin
-from rest_framework_swagger.views import get_swagger_view
 from tastypie.api import Api
 from tastypie_swagger.views import SwaggerView, ResourcesView, SchemaView
 from rest_framework.routers import DefaultRouter
 from rest_framework.authtoken import views as tokenviews
+from rest_framework import permissions
+from drf_yasg.views import get_schema_view
+from drf_yasg import openapi
 from django.http import HttpResponse
 from defectDojo_engagement_survey.urls import urlpatterns as survey_urls
 
@@ -150,7 +152,17 @@ swagger_urls = [
     url(r'^schema/$', SchemaView.as_view(), name='schema'),
 ]
 
-schema_view = get_swagger_view(title='Defect Dojo API v2')
+schema_view = get_schema_view(
+    openapi.Info(
+        title="Defect Dojo API",
+        default_version='v2',
+        description="To use the API you need be authorized.",
+    ),
+    # if public=False, includes only endpoints the current user has access to
+    public=True,
+    # The API of a OpenSource project should be public accessible
+    permission_classes=(permissions.AllowAny),
+)
 
 urlpatterns = [
     #  tastypie api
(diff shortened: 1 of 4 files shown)


alles-klar:
A first Version of the api wrapper generated with openapi-generator can be found here: https://github.com/alles-klar/defectdojo-api-v2-client

dojo/urls.py
```diff
@@ -150,7 +152,17 @@
...
-schema_view = get_swagger_view(title='Defect Dojo API v2')
+schema_view = get_schema_view(
+    openapi.Info(
+        title="Defect Dojo API",
+        default_version='v2',
+        description="To use the API you need be authorized.",
+    ),
+    # if public=False, includes only endpoints the current user has access to
+    public=True,
+    # The API of a OpenSource project should be public accessible
```
propersam: Does this really apply with defect Dojo @aaronweaver the previous doc used did not reveal all endpoint to unauthenticated users...

alles-klar: If we want to restrict the api to authenticated users only, we can do this with the permission classes (line 164). The public flag configures only if a user with access can see the hole api documentation or only functions he can use. 

From my point of view the api access is already restricted and we do not need to restrict the access to the documentation.

Maffooch:
Hey @alles-klar I got the following error attempting to access the v2 docs at this endpoint: http://localhost:8080/api/v2/doc/

image

dojo/urls.py
Maffooch: I believe what @propersam was trying to say is in an instance of multiple products with varying users having access. If a user does not have access to a given product, they are not allowed to see any endpoints or findings. However, these endpoints could still be seen through API usage. Simply restricting by authenticated users would not completely satisfy this situation.

alles-klar: We have to distinguish between the authentication for the api and the authentication for the api documentation. This PR is only about the api documentation and does not change the authentication for the api.

Maffooch: I think I understand now. What you're talking about is the auth to access the api whereas Sam  and I were talking about the results from using the api. Is that correct?

alles-klar:
@Maffooch I can not reproduce your error. Did you rebuild your nginx container or the static files?

Maffooch:
Weird. I couldn't replicate either! Thanks for verification.

dojo/urls.py
propersam: I was actually referring to d authentication on d api docs... @alles-klar reasons for making the api docs open..seems like a valid reason..it does nt affect d user authentication....unauthenticated users won't be able to use any endpoint that requires authentication but they would be allowed to see it exists frm d doc

alles-klar: correct

Maffooch: V1 docs are already this way so now it's consistent I guess. Null auth error messages pop correctly too. I don't see a problem with this anymore. Sorry for the confusion!

PR #1665, opened 2019-11-25
Deduplication configuration per parser2

ptrovatelli:
New PR for https://github.com/DefectDojo/django-DefectDojo/pull/1540
I have squashed those sparsed commits and rebased them on the latest dev. 
This will give a much nicer history after merging. 
(Please do not squash commit when merging, those are already 3 big enough commits)

@madchap @Maffooch @devGregA can you check this out? 
thanks!


ptrovatelli:
if you had started looking at the other branch, you can verify that this branch has the same code as deduplication-configuration-per-parser with: 
```
 (deduplication-configuration-per-parser)
$ git difftool --dir-diff deduplication-configuration-per-parser2
```
-> nothing in output meaning it's the same code

piyarathnalakmali:
Hi @ptrovatelli 

I have 2 requirements related to deduplication.

1.  Attributes used to identify duplicates should be configurable. (Configurable duplicate identification attributes )
2. For each duplicate finding it has a link to it's original finding. I need to filter the most related finding ( Findings from the same engagement are related than findings from the same product ) and set it as the original of the duplicate finding.
    - Ex : For a newly added finding (F) , if two findings were identified as duplicates, one finding (F1) from the same engagement and one finding (F2) from the same product, I want finding F1 to be the original of new finding (F).

If I am developing on top of the changes of this pull request, as I understand I have to implement a new duplicate identification algorithm to fulfill my requirement 1. What do you think as the best approach?
As I want to meet both requirements, I will have to integrate the second one also to the same duplicate identification algorithm. My concern is if any other user who wants to use one from the above two requirements won't have a way to achieve that. Can you please give your suggestions.

ptrovatelli:
Hello @piyarathnalakmali,
- For requirement 1 if I understand correctly you have nothing to develop: it's only configuration if your work is based on that PR. 
It's all in settings.dist.py. 

for example for checkmarx :

this chooses which attributes are used to compute the hash_code
```
    # In checkmarx, same CWE may appear with different severities: example "sql injection" (high) and "blind sql injection" (low).
    # Including the severity in the hash_code keeps those findings not duplicate
    'Checkmarx Scan': ['cwe', 'severity', 'file_path'],
```

If the attribute you need isn't in this list: 
`HASHCODE_ALLOWED_FIELDS = ['title', 'cwe', 'cve', 'line', 'file_path', 'description', 'endpoints', 'unique_id_from_tool', 'severity']
`
You can add it. 
This list is just a help to provide attributes that are tested and known to be working but it can be extended if required. It the attribute is in the finding model it will work (code is generic). 

then this: 
```
HASHCODE_ALLOWS_NULL_CWE = {
    'Checkmarx Scan': False,
```
tells that if we find a null CWE, we switch back to legacy hash_code computation (because we do need a CWE to compute the hash_code when using HASHCODE_FIELDS_PER_SCANNER for checkmarx given our configuration)

and this tells that we dedup based on hash_code (so in our example: based on ['cwe', 'severity', 'file_path'])
```
    'Checkmarx Scan': DEDUPE_ALGO_HASH_CODE,
```

The other option in DEDUPLICATION_ALGORITHM_PER_PARSER is there if we want to use another field than hash_code to deduplicate; the usecase is typically a unique identifier used in the source tool. We can also dedup on both hash_code and that unique identifier using: 
    'myscanner': DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE,

- For requirement 2 i'm not sure I understand why you need it. You can use the "deduplicate on engagement" in order to restrain the deduplication to the same engagement but it has a very different meaning than deduplicate on product. 

  - I understand deduplicating on product to have a central point for each given vulnerability, so that you can enrich it and keep the information till the vuln is fixed when running again imports for differents builds/releases. currently deduplication or product also deduplicate on the same engagement at the same time
  -  I understand deduplicating on engagement so that you can see which vuln are duplicate in a given import
  - But I don't see the usecase of transforming a dedup found on product to a dedup on the engagement

I'd say ideally we should have two different systems that could be activated at the same time: 
- dedup on product that we could rename as same vulnerability tracking (wouldn't deduplicate on engagement)
- dedup on engagement which is really a deduplication

ptrovatelli:
Any feedback is welcome. 

You can try this PR with: 
```
git fetch origin pull/1665/head:pr-1665
git checkout pr-1665
docker-compose build
docker-compose up
```

It that's not working you may be on a different database branch; get a fresh environment with (that will remove all data!): 
```
docker-compose down --volumes
docker-compose up
```

piyarathnalakmali:
Thank you for the explanation @ptrovatelli 

@madchap @Maffooch I am planning on doing some improvements to the duplicate identification feature. If this PR is getting merged in the near future, I can implement them on top of this PR. If it is taking time, I am planning on doing those improvements on the existing functions. But, as this PR and my improvements will change the same set of files, there will be a lot of conflicts. so, Can you please let me know a rough date which this PR will be merged?

devGregA:
All, if you have any final concerns regarding this PR please let me know. @ptrovatelli has been waiting quite a while to get this PR through. I'm going to merge Friday December 6th unless there are other concerns brought up.

piyarathnalakmali:
@ptrovatelli 
In the current implementation, for each existing finding (F) which passes all the checks, new_finding is added as a duplicate to F's duplicate list when iterating in the for loop.

But, with your changes, [this line](https://github.com/DefectDojo/django-DefectDojo/blob/ab5014853213fd2a5ccad76b632ce52cb8ec1ba9/dojo/utils.py#L180) breaks the for loop after one successful iteration. Because of that, the new finding will be added as a duplicate, only to the first existing finding in the list. Is there any specific reason why you added a "break" statement to the end of each function? As I understand, a set of relationships which are created when using the existing function are missed with your changes.

Code line I mentioned is from the deduplication legacy algorithm. The same behavior is repeated in your new algorithms as well.

ptrovatelli:
@piyarathnalakmali it is an little optimization. It shouldn't change anything because the findings we are iterating over are non duplicates only, so there can be only one non-duplicate that matches, the other ones should be duplicates of this original finding:
See `exclude(duplicate=True)` in

```
            eng_findings_cwe = Finding.objects.filter(
                test__engagement=new_finding.test.engagement,
                cwe=new_finding.cwe).exclude(id=new_finding.id).exclude(cwe=0).exclude(duplicate=True)
            eng_findings_title = Finding.objects.filter(
                test__engagement=new_finding.test.engagement,
                title=new_finding.title).exclude(id=new_finding.id).exclude(duplicate=True)
(...)
total_findings = eng_findings_cwe | eng_findings_title
(...)
for find in total_findings:
```

PR #1668, opened 2019-11-25
Fix bug with deduplication in qualys findings

alles-klar:
Qualys findings contains information like dates or times found. When using deduplication the hash should not include this information. To archive this we using a separator.

Has someone has a better idea to solve this bug or is this solution fine?

On behalf of DB Systel GmbH

---

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1318,7 +1318,14 @@ class Finding(models.Model):
         return filtered.exclude(pk=self.pk)
 
     def compute_hash_code(self):
-        hash_string = self.title + str(self.cwe) + str(self.line) + str(self.file_path) + self.description
+        hash_string = self.title + str(self.cwe) + str(self.line) + str(self.file_path)
+        # additional check to exclude unwanted information from description for the duplicate detection, e.g. dates, times_found, ...
+        end_position = self.description.find("================")
+        if end_position > -1:
+            hash_string += self.description[:end_position]
+        else:
+            hash_string += self.description
+
         if self.dynamic_finding:
             endpoint_str = ''
             if len(self.unsaved_endpoints) > 0 and self.id is None:
--- a/dojo/tools/qualys/parser.py
+++ b/dojo/tools/qualys/parser.py
@@ -141,6 +141,7 @@ def issue_r(raw_row, vuln):
                                                    htmltext("QID: " + str(_gid)),
                                                    htmltext("Port: " + str(_port)),
                                                    htmltext("Result Evidence: " + _result),
+                                                   htmltext("================"),
                                                    htmltext("First Found: " + _first_found),
                                                    htmltext("Last Found: " + _last_found),
                                                    htmltext("Times Found: " + _times_found),


ptrovatelli:
this modification would impact all the scanners. 
when PR https://github.com/DefectDojo/django-DefectDojo/pull/1665 is merged, it will be possible to configure the hash_code fields for each parser

Can you have a look and maybe post an +1 in the PR if that works for you because this PR have been sitting there for a lot of time. Having some feedback would probably be helpful :)
Thanks!

alles-klar:
After PR #1665  is merged, I thought of 3 possibilities:
1. Use the existing workaround in this PR.
2. Remove the fields "Last Found" and "Times Found", because they are not so important. As a user I only want to know if a finding is there or not. "First Found" should be constant.
3. Remove the description field from deduplication. I am not sure if this causes any problems. 

I prefer solution two. @ptrovatelli What do you think?

ptrovatelli:
Hi @alles-klar . will have a look asap. i need to check what qualys reports look like but if we can fix it by configuration only it would be better. if not possible, maybe the scanner needs fixing :)
 in the meantime, I have added documentation about what is possible with that new deduplication configuration : please see https://github.com/DefectDojo/Documentation/pull/70

ptrovatelli:
@alles-klar your problem is that the legacy deduplication is using the description when computing the hash_code. with the new configurable deduplication system, you shouldn't need to change the code to get a proper deduplication because you can just configure how to compute the hash_code. 
The below configuration works OK but produces many duplicates (208 duplicates out of 299 in https://raw.githubusercontent.com/DefectDojo/sample-scan-files/master/qualys/Qualys_Sample_Report.xml ) due to a bug in the endpoint listing: the current code is not able to find the endpoints while computing the hash_code. I have picked the same code as the legacy algo for getting the endpoints but it's not working either in the legacy dedupe. 
There is something that I don't understand with the finding saving. it looks like it's triggered twice, the first time with self.id=None and then with the correct id. 
I haved added a couple of debug logs here, it will give an idea of what's wrong https://github.com/ptrovatelli/django-DefectDojo/tree/dedupe-dynamic-debug

ZAP scanner has the same issue, I think it's the same for all dynamic scanners.

anyway, the best I have right now: in settings.py: 

```
...
    'Qualys Scan': ['title', 'severity', 'endpoints']
}

HASHCODE_ALLOWS_NULL_CWE = {
...
    'Qualys Scan': True
}

...
    'Qualys Scan': DEDUPE_ALGO_HASH_CODE
}
```

NB: you'll find documentation about that deduplication system here (freshly merged) : https://defectdojo.readthedocs.io/en/latest/features.html#deduplication

ptrovatelli:
got it.      
Finding.save is called once from serializers.py with dedupe_option=False because the findings is not ready yet, for example the endpoints are not computed.
It is then called a second time with dedupe_option defaulted to true; now we can compute the hash_code and run the deduplication
So we only need to wait for the second call (dedupe_option=True) for computing the hash_code. 
I will push the fix very soon

ptrovatelli:
@alles-klar can you try out https://github.com/DefectDojo/django-DefectDojo/pull/1730 and tell me if that works for you? 
thanks!

alles-klar:
@ptrovatelli Thanks for your work. I think we don't need this PR anymore.

PR #1672, opened 2019-11-26
Add more product metrics, and base off open findings

madchap:
Trying to enhance the product metrics page a bit. Adding to the "overview":
- false-positive 
- out of scope 
- inactive 

Also, to follow-up on #1520 , detailed metrics now get their data from open findings (aka active) and not verified ones, hence actually showing something most of the time.

The motivation behind this is mainly for CI automatically imported reports, which establish a baseline, but sometimes do not get analysed when you have so many. Through re-importing scans on a regular basis, seeing the trend is good.

To that effect, I also think that having a dynamic way to select which "source" of findings to take to graph would be good; like a dropdown box to choose from "verified", "open" or even "inactive" findings instead of having that hardcoded. Some products findings may never really investigated (and so likely get imported as "inactive"), yet it's nice to have a trend as reports get re-imported over and over through CI.

It also adapts the product overview to use open (a.k.a active) findings and not verified findings.

image

- [] Your code is flake8 compliant 


madchap:
I am guessing this works well enough in my little local tests.

I wish we had an easy way to somehow replace this with something like plotly charts to make nice interactive diagrams, but 1. not really my cup of tea 2. does not seem so straightforward either.

madchap:
I can reproduce indeed. Does not happen when you mark it false-positive "individually" though, just through bulk.

There is also something else I need to track down. When there is an active (open) finding, the first graph classifies it as "accepted" and not "open".

madchap:
The bulk "false-positive" does not mark the finding as "mitigated", this is why it does not show.

madchap:
OK, this should be sorted.

PR #1673, opened 2019-11-26
adding multi-procs and threads by default

madchap:
Adding multi process and multi-thread by default to avoid blocking.
Otherwise, you may just be watching an import for 5 minutes and not do anything else.


ptrovatelli:
same as #1660 ?

madchap:
yea... no comment :-)

PR #1675, opened 2019-11-26
Checkmarx set false positive status

ptrovatelli:
Imports checkmarx vuln with false positive status when found as such in checkmarx report
In WIP because I have based this branch on https://github.com/DefectDojo/django-DefectDojo/pull/1665 where checkmarx parser was modified a lot. 
It's actually just a couple of lines of code in checkmarx parser.py for this PR

Note that ticking "active" / "verified" in the GUI may lead to status such as: 
active, verified, false positive

this is because active and verified when ticked in the GUI overwrites what is done by the parser.

### DIFF
--- a/dojo/tools/checkmarx/parser.py
+++ b/dojo/tools/checkmarx/parser.py
@@ -107,6 +107,7 @@ class CheckmarxXMLParser(object):
                            test=self.test,
                            active=False,
                            verified=False,
+                           false_p=result.get('FalsePositive') == "True",
                            # Concatenates the query information with this specific finding information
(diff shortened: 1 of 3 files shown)


ptrovatelli:
good for review. there is something off with the findings saving in general though regarding the active/verified status: 
- with api v2:
  - by default: active=true verified=true (even if false positive)
  - in API call if we force in the parameters active=false, verified=false =>  active=true, verified = false
- gui: 
  - if we tick active/verified: active/verified
  - else: inactive / not verified

Issue #1680, opened 2019-12-02
BlackDuck zip-file importer - always 0 findings


Apipia:
**Bug description**
BlackDuck import accepts a zip file with various csv's in them for upload. The code for this expects to find 2 files: files.csv and security.csv. However, these files are not named exactly that. As you can see below, the files contain a date. Also, in the particular version of blackduck that we use, the files.csv has been renamed to source.
![image](https://user-images.githubusercontent.com/17787069/70001653-b7795c00-1512-11ea-9ef5-734f26f2dd7e.png)

Code to be changed:
https://github.com/DefectDojo/django-DefectDojo/blob/4c5bf565ccce0df5c6e4fb3f23fd68f18864f2e4/dojo/tools/blackduck/importer.py#L58

**Steps to reproduce**
1. Create a blackduck report zip file
2. Upload the zipfile to a test in Defect Dojo
3. Note that no findings are found (since it doesn't find the files it's looking for)
4. (No error message)

**Expected behavior**
It should find the proper files and upload their results.

**Deployment method** *(select with an `X`)*
- [X] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: [e.g. Ubuntu 16.04]

This issue is quite similar to #1658, however that issue is more related to the CVE and database scheme, not the importer logic. I also noticed that amrityam's zip file is formatted similar to mine, meaning that we should be sharing this issue as well.

madchap:
Thanks for reporting this @Apipia !

As discussed on Slack, if you'd like to submit a small PR to fix this, I can assign this issue to you. Let me know :-)

Apipia:
Yeah that works! I'm testing my changes right now.

madchap:
Closing this, since your patch was merged. Cheers!

PR #1632, opened 2019-11-02
Fix cve regex 1553

blacklotos:
**Note: DefectDojo is now on Python3 and Django 2.2.1 Please submit your pull requests to the 'dev' branch as the 'legacy-python2.7' branch is only for bug fixes. Any new features submitted to the legacy branch will be ignored and closed.**

When submitting a pull request, please make sure you have completed the following checklist:

- [ ] Your code is flake8 compliant 
- [ ] Your code is python 3.5 compliant
- [ ] If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR.
- [ ] Model changes must include the necessary migrations in the dojo/dd_migrations folder.
- [ ] Add applicable tests to the unit tests.

PR #1683, opened 2019-12-03
Blackduck importer fixes

Apipia:
### Fixes

1. Parsing ZIP file now finds the correct files. (Solves always 0 findings)
2. Locations will cap the size of the various file paths at 4000 characters (so there is no sql error)
3. CVE's will be pulled out of the vulnerability id field (they will sometimes contain BDSA identifiers as well as CVE's) 
4. There is no "Product id" column in the csv's, that is replaced with "Component id"

Let me know if you have some different version of blackduck that has a Product id in the csv instead of Component id, I could make it backwards compatible.

This PR refers to issue #1680.
Have also updated the sample scan zip at DefectDojo/sample-scan-files and created a PR there.

My first contribution
@madchap, thanks for the help. Please take a look. :)


Apipia:
I would like to upgrade this parser farther and overhaul it. Blackduck is _less_ about the vulnerabilities in the included components for your projects, and _more_ about those packages being at vulnerable versions in the first place. 

For our team's use, I will be updating the parser and importer to support this way of presenting findings. An out-of-date and vulnerable component will be ONE finding, and it's various CVE's will be included in the description, the mitigation will simply be to update the vulnerable component to a secure version.
Another use case of Blackduck (the one that we care more about, honestly) is the License Risk, which isn't being considered with the current parser/importer logic. I will be updating it to create findings for License Risks as well. 

Would this be of value to any Blackduck users? If not I can just keep those future changes locally.

madchap:
Thanks @Apipia for working on this first contribution!

A few comments:
* I would indeed retain the backwards compatibility. We do not want to break anything for other users in general and be able to address the use-cases of many.
* Regarding your 2nd comment, you don't necessarily have to modify this very parser either. I feel it would then be overly complicated to maintain backward compatibility while trying to adapt to newer ways of presenting findings. You could also add a brand new Blackduck parser (e.g. different factory/class name) to handle the report in a novel way and eventually have the two options offered in parallel. Please use another PR for that.
* For the license risks, it will be interesting to see what changes you propose ;-) It is probably better for you to work on this in a separate PR to keep things simple.

If you would like people to help test your PR at some point, please upload some test files to help. You can of course anonymize them as needed. It would be good to have them in the other repo at https://github.com/DefectDojo/sample-scan-files.

dojo/tools/blackduck/importer.py
```diff
@@ -54,12 +55,13 @@ def _process_zipfile(self, report):
         security_issues = dict()
         try:
             with zipfile.ZipFile(str(report)) as zip:
-                for file_name in zip.namelist():
-                    if file_name.endswith('files.csv'):
-                        with io.TextIOWrapper(zip.open(file_name)) as f:
+                for full_file_name in zip.namelist():
+                    file_name = full_file_name.split("/")[-1]
+                    if 'source' in file_name:
```
madchap: Wasn't it `sources` you had mentioned in Slack and not `source`? Also of course, please think there about backward-compatibility.

dojo/tools/blackduck/importer.py
```diff
@@ -103,12 +118,16 @@ def _process_project_findings(self, project_ids, security_issues, files=None):
...
-                    ', '.join(locations)
+                    location
                 )
 
     def __partition_by_project_id(self, csv_file):
         records = csv.DictReader(csv_file)
         findings = defaultdict(set)
         for record in records:
-            findings[record.get('Project id')].add(frozenset(record.items()))
+            findings[record.get('Component id')].add(frozenset(record.items()))
```
madchap: Mind the backward compatibility here too.

dojo/tools/blackduck/importer.py
Apipia: Turns out the file is called "source".
Example: source_2019-11-25-223934.csv

Also copy that, I will consider backwards compatibility for these.

Apipia:
I have made the requested backwards compatibility changes.
Also I have a PR on the sample upload files here:
https://github.com/DefectDojo/sample-scan-files/pull/40
The data is completely random and will create 11 findings:
https://github.com/DefectDojo/sample-scan-files/pull/40

dojo/tools/blackduck/importer.py
```diff
@@ -103,12 +120,21 @@ def _process_project_findings(self, project_ids, security_issues, files=None):
...
+            key = "Project id"
+        else:
+            key = "Component id"
         for record in records:
-            findings[record.get('Project id')].add(frozenset(record.items()))
+            findings[record.get(key)].add(frozenset(record.items()))
         return findings
+
+    def get_cve(self, vuln_id):
+        cve_search = re.search("CVE-\d{4}-\d{4,7}", vuln_id, re.IGNORECASE)
```
adracea: I think this would need to be `re.search(r"CVE-\d{4}-\d{4,7}", vuln_id, re.IGNORECASE)` in order for it to be flake8 compliant .

dojo/tools/blackduck/importer.py
```diff
@@ -103,12 +120,21 @@ def _process_project_findings(self, project_ids, security_issues, files=None):
...
+            key = "Project id"
+        else:
+            key = "Component id"
         for record in records:
-            findings[record.get('Project id')].add(frozenset(record.items()))
+            findings[record.get(key)].add(frozenset(record.items()))
         return findings
+
+    def get_cve(self, vuln_id):
+        cve_search = re.search(r"CVE-\d{4}-\d{4,7}", vuln_id, re.IGNORECASE)
```
madchap: Actually, I am thinking here that CVE may not be the only pattern to match, if it comes from their own research/DB? I can't remember (because I am actually not using Blackduck), but it could be something like BD-1234567 or BLACKDUCK-1234567?

https://github.com/DefectDojo/django-DefectDojo/pull/1632 would have a better regex to apply eventually that would match a broader range of use-cases.

madchap:
I am guessing it should yes, even though the way we test today is a bit naive.

That being said, @Apipia, it would be great if you could add a test with regards to this new file format so that it gets automatically tested for both these cases.

dojo/tools/blackduck/importer.py
Apipia: The Vulnerability ID for Blackduck security findings might look like:
BDSA-2017-2248   (CVE-2017-13098)
OR
BDSA-2018-1190
The logic, as I have written it, will use the CVE if one is present (using that regex), or the BDSA if no CVE is present. The reason for this is that, in the first case, the vulnerability ID is too large as it is to be stored as a CVE in Defect Dojo, so I find and pull out the CVE field. However, there are many findings that do not have a CVE and only a BDSA. The return statement (line 140) will return the vulnerability ID as-is (BDSA) if no CVE is found.
`return cve_search.group(0) if cve_search else vuln_id`

Let me know if that works in general, but for our team this is ideal, also it avoids a SQL error with the cve field being too large. Or if this is impacted by #1632.

madchap: Maybe one last change here to make sure to address future extension of the number of digits by MITRE, maybe make it like `r"CVE-\d{4}-\d{4,9}"`. Also the regex is not strict, but I guess it does not matter in our case.

madchap:
Thank you @Apipia for the contribution!

PR #1689, opened 2019-12-07
Fixing endpoint pop-up in add_findings.html  

no-sec-marko:
The browser should create a new window / pop-up if a user wants to add an endpoint to a newly created finding. However, jQuery crashs with a Deferred exception and the browser doesn't create a new window:
```
jquery.min.js:2 jQuery.Deferred exception: Cannot read property 'document' of null TypeError: Cannot read property 'document' of null
    at HTMLDocument.<anonymous> (http://localhost:8080/endpoints/1/add?_popup:141:25)
    at l (http://localhost:8080/static/jquery/dist/jquery.min.js:2:29375)
    at c (http://localhost:8080/static/jquery/dist/jquery.min.js:2:29677) undefined
```
After adding the `RelatedObjectLookups.js`, the pop-up will be created.

### DIFF
--- a/dojo/templates/dojo/add_findings.html
+++ b/dojo/templates/dojo/add_findings.html
@@ -58,6 +58,7 @@
     <script type="application/javascript" src="{% static "chosen/chosen.jquery.min.js" %}"></script>
     <script type="application/javascript" src="{% static "simplemde/dist/simplemde.min.js" %}"></script>
     <script type="text/javascript" src="{% static "admin/js/jquery.init.js" %}"></script>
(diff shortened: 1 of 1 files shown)


madchap:
@no-sec-marko I could not reproduce the issue. 

I tried creating an endpoint alone, and then in edit mode in a finding add a new endpoint there. But could not see any js errors.

Could you please give a step-by-step to reproduce?

Cheers.

Maffooch:
Also list your browser please. I couldn't reproduce on safari nor chrome.

no-sec-marko:
Sorry, I miss to add the used browser in my description. The issue appears in the latest Firefox 68.3.0esr (64-Bit) (Linux & Windows).

Issue #1678, opened 2019-11-27
Dependabot should want to merge into dev instead of master


madchap:
Today, dependabot wants to merge to `master`. We would want it to merge to `dev` instead.

I am guessing this is the default behavior, a repo admin should change the configuration.

cc @aaronweaver @devGregA 

Cheers,
fred

Issue #1605, opened 2019-10-18
DefectDojo known security vulnerability in docker images w.r.t jquery


bilalk88:
Hello Team,

We are trying to on board defectdeojo in our corporate, but as per our policy we have to perform container security scanning. While scanning we came up accross with jquery version used by defectdojo is vulnerable.
DefectDojo uses jquery version 2.1.* and from our scanning it shows that it should be upgraded to latest version 3.0.

So could you please provide some suggestion on upgrading jquery or can we get latest update on via docker image itself.

ptrovatelli:
hi  @bilalk88 . on what branch are you? do you know where this dependency come from? from what I see in package.json on dev or master branch we're supposed to be using jquery 3: 
` "@yarn_components/jquery": "jquery/jquery-dist#~3.3.1",`

bilalk88:
Hi @ptrovatelli . Thanks for the response.
I'm in the master branch. I'm not sure from where this old version of jquery came from. But when I checked in the nginx container under  " /usr/share/nginx/html/static/ " there all are older version of jquery there.

ptrovatelli:
Those files are generated by (Dockerfile.nginx)
`python3 manage.py collectstatic --noinput`
they contain with other things, libraries pulled by yarn based on package.json.
this basically gets all static resources from django so that nginx can serve them. 

I'm actually even seeing what seems to be jquery 1.7 and jquery 1.8. Can you point us to the files you're seing that are jquery 2.1? 
```
docker exec -it django-defectdojo_nginx_1 bash
cd /usr/share/nginx/html/static
 find . -type f -name '*jquery*'
(...)
./font-awesome/src/3.2.1/assets/js/jquery-1.7.1.min.js
./fontawesome/src/3.2.1/assets/js/jquery-1.7.1.min.js
(...)
./jquery-ui/ui/jquery-1-7.js
(...)
./jquery.hotkeys/jquery-1.4.2.js
(...)
./rest_framework_swagger/lib/jquery-1.8.0.min.js
(...)
./tastypie_swagger/js/lib/jquery-1.8.0.min.js
./tastypie_swagger/js/lib/jquery.ba-bbq.min.js
./tastypie_swagger/js/lib/jquery.slideto.min.js
./tastypie_swagger/js/lib/jquery.wiggle.min.js
```

it's probably possible to improve the situation by tuning that package.json file.

bilalk88:
@ptrovatelli. Thanks for the response.

[...]

PR #1702, opened 2019-12-16
Dependabot config - target_branch

madchap:
Addresses #1678 

Dependabot should issue PR against `dev` and not `master`.

This config would normally work for python and yarn dependencies.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -12,4 +12,5 @@ update_configs:
 
   - package_manager: "docker"
     directory: "/"
+    target_branch: "dev"
     update_schedule: "weekly"
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,7 +1,5 @@
 version: 1
 update_configs:
-  # Keep package.json (& lockfiles) up to date as soon as
-  # new versions are published to the npm registry
   - package_manager: "python"
     directory: "/"
     target_branch: "dev"
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,13 @@
+version: 1
+update_configs:
+  # Keep package.json (& lockfiles) up to date as soon as
+  # new versions are published to the npm registry
+  - package_manager: "python"
+    directory: "/"
+    target_branch: "dev"
+    update_schedule: "live"
+
+  - package_manager: "javascript"
+    directory: "/"
+    target_branch: "dev"
+    update_schedule: "live"
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -9,3 +9,7 @@ update_configs:
     directory: "/"
     target_branch: "dev"
     update_schedule: "live"
+
+  - package_manager: "docker"
+    directory: "/"
+    update_schedule: "weekly"


alles-klar:
Can we also use dependabot to update the dockerfiles?

madchap:
@alles-klar we can, yet this would go against the recent decision to pin the version of the parent image with sha256 hashes.

alles-klar:
I think we pinned the version because we do not want to use a new version without testing or even know  it. But with dependabot we get a PR and can test the new version bevor merging. At least for security reasons we should periodically update the docker container.

Maffooch:
I like the idea of using Dependabot for Dockerfiles too. As with pip packages, somebody will check out the release notes and do a test drive to ensure it won't wreck the project. Packages we are intentionally keeping a bit dated, like python, can simply be declined for updates. I didn't see anything about Dependabot supporting Debian packages though.

madchap:
Alright then. Full disclosure, I have never used it for docker.

madchap:
We will need to change the Dockerfile to use regular tags again then to remove the sha256.

madchap:
Getting some tests done on my fork, I guess it's working good enough. docker OK, npm seems ok, python django not showing up, wondering why actually :thinking: 

image

madchap:
Also, note this change would have to be pushed to the default branch to take effect, which I am assuming to be `master` here.

ptrovatelli:
Hi guys, 
Regarding the version pinning in dockerfile with sha256: I put that in place because this was the advised way to have a reproducible build at a time where build failed due to an evolution in a docker image. found on an article edited by the renovate bot team. 
i have discussed this on the PR at the time. I wanted to add "renovate bot" who is able to automatically create PR to update those version but i think only the project owner can add renovate bot. 
I didn't notice at the time that we already have a bot that updates versions. I don't know if dependabot is able to update dockerfile version, i haven't see any PR from his doing that. 

this article explains why it's a good idea to pin dockerfile using sha256, due to the mu mutable nature of docker images: https://renovatebot.com/blog/docker-mutable-tags

more documentation: 
https://github.com/marketplace/renovate
https://renovatebot.com/docs/

I have started testing that bot and it seemed fine. i can do more test and tune up the configuration to help it's integration if dependabot isn't able to handle dockerfile dependencies

madchap:
Well, actually dependabot handles the sha256 just fine, so no need to change anything if we want to do that, updates will happen. See https://github.com/madchap/django-DefectDojo/pull/11 for an example.

The other deps updates came in just a bit afterwards. I believe we can move forward there. You can look at PRs done on my fork for full examples.

ptrovatelli:
great! so it's just a bit of configuration on dependapot files? 
regarding this python dependency though, there was another discussion :) as stated in the comments in dockerfile.django, we left it at 3.5.7 on purpose in order not to break infrastructures that wouldn't support python 3.6 and 3.7. 
However what we can do is upgrade to the latest 3.5.x hash. do you know if dependabot support this? 

also we need to be careful with the yarn dependencies; we have given a quick look at https://github.com/DefectDojo/django-DefectDojo/issues/1605 and when trying to update yarn dependencies it seemed to break the gui. hopefully selenium will pick that kind of issue but i'm not sure.

madchap:
> great! so it's just a bit of configuration on dependapot files?
> regarding this python dependency though, there was another discussion :) as stated in the comments in dockerfile.django, we left it at 3.5.7 on purpose in order not to break infrastructures that wouldn't support python 3.6 and 3.7.
> However what we can do is upgrade to the latest 3.5.x hash. do you know if dependabot support this?

Yes. We can tell it to ignore with a comment, and it will not bother us again unless we tell it to.

> also we need to be careful with the yarn dependencies; we have given a quick look at #1605 and when trying to update yarn dependencies it seemed to break the gui. hopefully selenium will pick that kind of issue but i'm not sure.

Indeed. We will have to do due diligence and not blindly accept all its PRs :)

madchap:
So, we want to try? Need one more approval.

madchap:
> @madchap Why do you choose update_schedule: weekly for docker and live for the other?

I guess just what I see in most files I look at.

PR #1704, opened 2019-12-16
Blackduck Component Risk (License and Security Risk)

Apipia:
# Changes
Because Blackduck is designed to scan an application for the components it's importing and check them for vulnerable versions and license risks, I felt that a change to the resulting Findings to more closely align with that model was necessary. 
1. Findings are either Security Risks or License Risks
2. Security risks contain details about the risks in the description, but the solutions/mitigation is to update to a secure version.
3. License risks are created based on the components csv and the description has details from the csv file about the license in use.
4. No longer require the source csv, instead the components csv is needed. 
5. Does not support uploading a csv file by itself, it needs the zip file that contains both security and components csv.

blackduck_v2_findings
Here is an example of how findings are displayed.

blackduck_v2_license
This is an example of a license risk finding and its description.

blackduck_v2_secdesc
This is an example of the description of a Security risk finding and its description.

# Looking for Feedback:
### models.py
Because these findings don't contain a CWE, CVE, or Line number; as well as the fact that for each package, the FilePath being use is the same, each finding will be considered a "Similar Finding" with every other finding. This Similar Finding's function is defined in `dojo/models.py`. I have updated it to include a unique_id_from_tool to avoid this issue. Looking for feedback on this.

### The name
Wasn't sure what else to call it besides V2. It ingests the same kind of report as the other Blackduck integration, but represents the findings much differently.

### License Risk Severity
Currently I have it hardcoded to High. Not sure the best way to handle this, I don't want to decree that all license risks should be considered High for everyone.

## To Test:
[blackduck_hub_v2.zip](https://github.com/DefectDojo/django-DefectDojo/files/3963018/blackduck_hub_v2.zip)
Zip file only contains necessary files and the ProjectPath is fictional.


madchap:
Will certainly take a deeper look next week, I should have more time. 

Glancing at it, looks OK to have a complete different v2 version until we could eventually deprecate the other one. And also to have someone like @Apipia who's apparently on the ball there for this :) (I don't use blackduck at all so).

Apipia:
Travis says chrome crashed during some selenium test.. how can I run travis again?

madchap:
Closed and re-opened to trigger Travis again.

madchap:
`forms.py` was not updated to have this new scanner show in the dropdown selection box. Was that intentional?

madchap:
Also missing an addition to fixture `test_type.json`.

Apipia:
> Also missing an addition to fixture `test_type.json`.

Should be good to go now. Lmk if you see anything else.

madchap:
So I am still unsure on whether bending the current model to risk is a good thing.

If I understand correctly, the latest `dev` "version 1" of the Blackduck scanner now does all the vulnerabilities correctly, including backward compatible format and new format and this v2 does the same.

For clarity's sake, do you think we should rather decouple the vulnerabilities and risks in their own parsers? That would be leave the v1 parser as is, and your v2 (to be renamed) here would solely be dedicated to managing the risk part. Therefore, people interested in this part would know exactly what to look for and there is no surprise. I would probably be more in favor of that.

I am sure some people will like/need the risk part that you bring up, but the current defectdojo model is not necessarily suited for that. Maybe there needs to be a deeper discussion around the model if we ever want to integrate these kind of findings in the future.

@Maffooch thoughts?

Apipia:
I see where you're coming from but let me just take one more crack at this before we decide.
Blackduck is not a tool for scanning code for vulnerabilities like a static code analysis tool. A team would not purchase Blackduck to find SQL injection or any other code vulnerabilities. A team purchases Blackduck to scan builds and looks at the **packages (components)** that are imported and used to make sure that they are not at a version that has vulnerabilities, and that they do not have any license risk associated with them. 
If Blackduck finds something, the mitigation is always to go and update the version (if its vulnerable) or find another component (if there is license risk). The "vulnerabilities" imported in version 1 of the blackduck importer are _not vulnerabilities in **your** code_, they're vulnerabilities in the packages/components that you are importing. The fact that Blackduck tells you what the vulnerabilities are in one of these vulnerable versions of a component is a kind of added bonus. Reporting those as findings, in my opinion, is a little misleading. It's not like you can fix those vulnerabilities (they exist in the code base of a component you're importing). 

So version 1 says: "Hey you've got a few CVE's in your code, you need to fix those CVE's". My version 2 says: "Hey you've got a vulnerable version of a component (foo version 2.1.0) that has a few CVE's in it, you need to update that component."

madchap:
A lot of tools will indeed focus on the OWASP A9, using components with known vulnerabilities.

> So version 1 says: "Hey you've got a few CVE's in your code, you need to fix those CVE's". My version 2 says: "Hey you've got a vulnerable version of a component (foo version 2.1.0) that has a few CVE's in it, you need to update that component."

I would still create a finding per CVE as to comply with the current model as it stands, instead of having it all in the description field, only be it from a "data model" point of view. It will also make it harder (impossible?) for dedupe and harder for search. You can still aggregate these individual findings later to a component level for reporting needs. But I guess what you've tried to achieve here is to just give it another "view" in the end? If so, maybe there would be some other solution? (like a new view to pivot on data?)

(I also tried to ingest the v1 sample files, but none of them make the new parser happy -- just a note)

Apipia:
> (I also tried to ingest the v1 sample files, but none of them make the new parser happy -- just a note)
Well I mentioned above that this new version does not work with only one file as a csv, and that I've coded it to take the zip file generated from blackduck. So the original csv files that worked with v1 don't work here.

It just seems to me that Defect Dojo is designed to aggregate the data/findings from all of the application security tools out there into one place, and the results here should line up with the tool's representation. 
In our use-case, Fortify may detect some vulnerable code in the source during static code analysis and generate a finding. Then DefectDojo should show exactly that, a line, a line number, etc. Blackduck detects that some dependency is at an old version and that version has some vulnerabilities, creating a finding in blackduck that lists the dependency, it's version, and any license risk. So, it's my (and my team's) opinion, that DefectDojo should display exactly that, the dependency, it's version, etc. We don't exactly care about the vulnerabilities inside the dependency, it's not actionable information. 
Creating a JIRA ticket for this item for the developers that own it would be silly, we can't do it. Although a JIRA ticket saying "hey you're importing this old vulnerable component you need to update it (btw it has these CVE's)" is actionable.

That being said, I'm perfectly fine just keeping it as a separate parser and naming it something like BlackDuck Hub Risk or whatever we decide. We're going to use this for our purposes as it more aligns with what we need in our reporting and metrics, i.e. Vulnerable version of a component = finding/vulnerability.

> Maybe there needs to be a deeper discussion around the model if we ever want to integrate these kind of findings in the future.
I definitely think that not every finding from the different application security tools that exist are going to fit a model like CVE, Line, file, line number, etc. I think DD's representation of findings needs to be flexible. Trying to take a representation of a finding from a tool and force it to have this specific representation is like trying to put a square peg through a round hole. 

(I'm sorry for this novel of a reply 😅)

Maffooch:
This looks good to me. I don't think it should be V2 parser unless both versions can be used reliably under a single parser. Seems like they are doing things slightly different anyways. The idea @madchap had for an extension of BlackDuck is probably the best way go.

The license issue is neat and definitely could be useful to those worried about that info. I don't see any huge differences in here that digress away from the finding model. Similar with the security risk finding, it be super cumbersome to have individual findings for each CVE found in a package. There is not a lot (if anything) to be done to fix those issues. Deduplication would not work here whatsoever though. I guess the tradeoff is either triage tons of CVE findings from the same package or not have deduplication.

I think scrapping deduplication would be the most practical since it would allow all new changes to be seen. Manually looking at all results is going to happen one way or other. Comes down to the lesser evil.

Apipia:
Alright, works for me. Only question I have then is what do we name this parser?
Blackduck Component and License Risk (bit of a mouthful)
Blackduck Component Risk?

Apipia:
I'll fix the merge conflicts once we know what it's called, since I'll have to change it all anyway.

Maffooch:
BlackDuck Competent Risk sounds good. @madchap is all this cool with you?

madchap:
I am cool with it ;-)

Maffooch:
Bet. @Apipia you know what to do! I'll approve afterwards.

Apipia:
Merged in dev and some stuff caught fire so this may take some time. Hopefully I'll have this ready by tomorrow

Issue #1657, opened 2019-11-16
All fortify issues are not reflecting in DefectDojo


amrityam:
**Bug description**
I am uploading the XML report exported from Fortify SCA 19.10 and 17.20 to DefectDojo. Instead of showing all the vulnerabilities from each category,  DefectDojo is showing one vulnerability from each category.

**Steps to reproduce**
1.For a product navigate to Findings tab and click on import scan results.
2.Fill up the required details and choose scan type as Fortify Scan and upload the xml file generated by fortify
3.After successful upload, one can see DefectDojo is showing one vulnerability from each fortify  scan category instead of showing all vulnerabilities from all category.

XML was generated by using below command.
ReportGenerator -format XML -template "DefaultReportDefinition.xml" -f "C:\Demo\SourceCode\WebGoat-Legacy\WebGoat_Fortify_DefaultReportDefinition.xml" -source  "C:\Demo\SourceCode\WebGoat-Legacy\webgoat6-cli.fpr"

Please note for other -template types like DeveloperWorkbook.xml or OWASP2013.xml , while uploading the xml files in DefectDojo, Server Error (500) is coming. 

**Expected behavior**
DefectDojo must show all the fortify reported vulnerabilites.

**Deployment method** *(select with an `X`)*
- [x] Docker

**Environment information**
 - Operating System: [e.g. Ubuntu 18.04]
[files.zip](https://github.com/DefectDojo/django-DefectDojo/files/3854428/files.zip)

**Sample scan files** (optional)
Attached

ptrovatelli:
Hello @amrityam . 
It seems that this parser was done for that "defaultReportDefinition" report which doesn't have all the details. 
The structure is quite strange actually: it gives counts for a given type of vulnerability and then gives a couple of random files concerned, but not all: 
```
                    <GroupingSection count="4">
                        <groupTitle>Command Injection</groupTitle>

(...)
                        <Issue iid="2390CF7C0D738E2BE2F968716E8C1539" ruleID="1C7A212C-7723-452B-97D5-4D94EC85829E">
                            <Category>Command Injection</Category>

(...)
                            <Primary>
                                <FileName>WSDLScanning.java</FileName>
                                <FilePath>src/main/java/org/owasp/webgoat/lessons/WSDLScanning.java</FilePath>
(...)
                            </Primary>
                            <Source>
                            <FileName>ParameterParser.java</FileName>
```

[...]

PR #1706, opened 2019-12-17
Fortify Parser Overhaul

Apipia:
# Changes
Fortify parser now contains the following information:
1. Abstract details (short definition)
2. Explanation (of finding)
3. Snippet
4. Source details (if available)
5. Mitigation with Recommendation and Tips (if available)

Fortify parser now also creates a single Finding from each issue in the XML (rather than trying to collect them all into buckets)

# Screenshots
findings_list
List of Findings

finding_details
Details of a finding

desc_explanation
Description containing an Explanation

mitigation
Mitigation information for a finding

# Testing
I used [this](https://github.com/JoyChou93/java-sec-code) code to create an FPR with Fortify, generated the XML and uploaded it for a total of 324 Findings. (334 were created, 10 were removed for deduplication, which I verified and am satisfied with). The XML is also added to the repo for unit tests, but I'll include it here:
[java-sec-code.txt](https://github.com/DefectDojo/django-DefectDojo/files/3975415/java-sec-code.txt)

# Notes
There are some details in the README located in the dojo/tools/fortify folder that provide information on how to generate an XML from an FPR with _all issues_. I have even included the XML for the ReportGenerator that one can use to with ReportGeneration to get all issues. Shout-out to @dougmorato for help with that.

### DIFF
--- a/dojo/tools/fortify/DefaultReportDefinitionAllIssues.xml
+++ b/dojo/tools/fortify/DefaultReportDefinitionAllIssues.xml
@@ -1,0 +1,184 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<ReportDefinition type="standard" ID="fortifySecurityReport">
+	<TemplateName>Fortify Security Report</TemplateName>
+	<TemplatePath></TemplatePath>
+	<LogoPath>/MF_logo.jpg</LogoPath>
+	<Footnote>Copyright 2018 Micro Focus or one of its affiliates.</Footnote>
+	<ReportSection enabled="true" optionalSubsections="false">
+		<Title>Executive Summary</Title>
+		<SubSection enabled="true">
+			<Title>Issues Overview</Title>
+			<Description>This section provides an overview of the issues uncovered during analysis. The report covers a summary of vulnerability categories discovered by the tool. The auditor should augment this section with higher-level conclusions derived from human review of the application (including architecture reviews,	black-box testing, compliance issues, etc.)</Description>
+			<Text>On $SCAN_DATE$, a source code review was performed over the $PROJECT_NAME$ code base. $SCAN_SUMMARY$ were scanned and reviewed for defects that could lead to potential security vulnerabilities. A total of $TOTAL_FINDINGS$ reviewed findings were uncovered during the analysis.</Text>
+		</SubSection>
+		<SubSection enabled="true">
+			<Title>Issue Summary by Fortify Priority Order</Title>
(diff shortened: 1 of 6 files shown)


madchap:
Dumb question probably, are the changes proposed here backward compatible? If not, should they?

Apipia:
> Dumb question probably, are the changes proposed here backward compatible? If not, should they?

Great question. Not they’re not _entirely_ backwards compatible. Since the original version creates one finding per “bucket” and (attempts to) put details of each individual instance (line no and file path) in the description. This change would make an individual finding for each issue. Which feels more like how defect dojo is intended to be used since each finding has a file path and line number. 

As far as “should it be backwards compatible”, Its not really possible. But I feel this is the most appropriate way to implement it. Definitely open to suggestions and feedback though.

Maffooch:
This is a huge improvement. It doesn't seem like backwards compatibility would be an issue here since the format of the fortify scans hasn't changed. Looks like this is just reorganization of everything in a more useful manner. Is that correct @Apipia?

Apipia:
> This is a huge improvement. It doesn't seem like backwards compatibility would be an issue here since the format of the fortify scans hasn't changed. Looks like this is just reorganization of everything in a more useful manner. Is that correct @Apipia?

Well if you had findings imported from fortify already, when this gets merged those will probably need to be deleted. If you had multiple files/line_no's of a vulnerability, say: "Password Management: Password in Configuration File Environment", the current fortify parser would create ONE finding called "Password Management: Password in Configuration File Environment" and place all line numbers, file paths, etc, into the description. (though it was slightly bugged).
Now it will create a separate finding for each file/line_no combination, each called: "Password Management: Password in Configuration File Environment - file.java: 23" with a file and line number.

So it's a grey area, but I believe it does a better job than the previous parser, especially given a concern I've seen already in bug #1657, as this WILL create a finding for each issue.

The question is simply this: Would we (the community) prefer a Finding be an individual issue (line number and file pair)? Or a type of vulnerability (Say SQLi) and the description contain all of the line numbers and file paths? Personally I prefer the former.

Maffooch:
Gotcha. The former is definitely preferred in my opinion. This would also allow for deduplication here as well. Im onboard!

devGregA:
This seems like an improvement. It won't effect previous findings dedupe just won't work on 'legacy' fortify findings. Happy to merge. @madchap @Maffooch anyone second?

ashishf6:
@Apipia Can you provide me the command to generate report in xml format from fpr in fortify.

PR #1709, opened 2019-12-18
Limit number of similar findings displayed to 10

jvz:
This addresses an issue from #1628 when a finding with no CVE, CWE, file
path, or line number would list all findings from the same product which
can potentially grow unbounded.

@madchap @jeffret-b

- [ ] ~If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR.~
- [ ] ~Model changes must include the necessary migrations in the dojo/dd_migrations folder.~
- [ ] ~Add applicable tests to the unit tests.~

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1335,7 +1335,7 @@ class Finding(models.Model):
             filtered = filtered.filter(file_path=self.file_path)
         if self.line:
             filtered = filtered.filter(line=self.line)
-        return filtered.exclude(pk=self.pk)
+        return filtered.exclude(pk=self.pk)[:10]
 
     def compute_hash_code(self):
         if hasattr(settings, 'HASHCODE_FIELDS_PER_SCANNER') and hasattr(settings, 'HASHCODE_ALLOWS_NULL_CWE') and hasattr(settings, 'HASHCODE_ALLOWED_FIELDS'):


jeffret-b:
Seems reasonable to limit to 10 findings. That's probably a good idea anyway. But, I'm unclear about the nature of the issue from reading the description. It sounds like the matcher is being overly permissive in what it declares to be a potential match. Should there be some sort of threshold that needs to be reached. It sounds like the finding has insufficient data from which to form a match so it's aggressively matching.

Maffooch:
@jeffret-b makes a great point about duplication status being a bit too lenient. This is exactly what brought on #1665. Is it possible to use the parser configured duplication rules to identify duplicates for this purpose?

jvz:
The matcher starts by looking for findings in the same product. Based on the presence of additional data in the finding, the similar findings are narrowed down to ones with the same attributes. In the case where a finding has no useful attributes (the four I mentioned in the ticket), then all findings within that product are shown. The options here would really be to either show nothing or limit the list for all situations. I thought of paginating it, but it's only a sub-page here, so I didn't want to get overly fancy with the UI.

Issue #1643, opened 2019-11-10
Engagement view: action shortcuts not working and test menu hardly visible


ptrovatelli:
**Bug description**
Two UI issues in engagement view, introduced at 1720c95bf831236b81c5713d6006b29e55dec33c I believe
- action buttons aren't working anymore: when pushing "i" or "a" or "e" it's not doing anything: 
` ProTip! Type e to edit this engagement. Type i to import scan results or a to add tests.`
- the test drop-down menu is hardly visible and we need to scroll to be able to see it (see screenshot)

**Steps to reproduce**
setup env:
```
git checkout 1720c95bf831236b81c5713d6006b29e55dec33c
# this will remove any data
docker-compose down --volumes
docker-compose build
docker-compose up
```

- create a product
- create an engagement
- import any scan (for exemple tests/zap_sample.xml)
- click on the engagement name to display the engagement window or navigate to http://localhost:8080/engagement/1
- push "i" key: nothing happens
- click "..." in the tests section

**Expected behavior**
- "i" should open the import scan window
- "..." should display a visible menu

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [ X] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
1720c95bf831236b81c5713d6006b29e55dec33c

**Screenshots** (optional)
![menu](https://user-images.githubusercontent.com/34663482/68552565-ac9e4080-0418-11ea-829e-71e4c333c65f.jpg)

**Additional context** (optional)
I have tried with the previous commit e49bbb5e4fb8cab0d8e2214f153744ed86a8d39a and everything works fine. 
It's hard to tell where the issue comes from. My guess is that it comes from the application imported in requirements that somehow changes the engagement view.
`git+https://github.com/DefectDojo/defectDojo-engagement-survey@master#egg=defectDojo-engagement-survey
`

I have saved the html page from e49bbb5e4fb8cab0d8e2214f153744ed86a8d39a and from 1720c95bf831236b81c5713d6006b29e55dec33c and there are some diff that I have a hard time understanding (don't seem to come from anything commited at 1720c95bf831236b81c5713d6006b29e55dec33c). the issue probably comes from one of theese diffs. 
See diff in attachment: 
[view_engagement_html_code.zip](https://github.com/DefectDojo/django-DefectDojo/files/3829165/view_engagement_html_code.zip)

@Maffooch any idea where this could be coming from?

[...]

PR #1713, opened 2019-12-19
Fix test dropdown menus

Maffooch:

Motivated by #1643 

The priority of visible objects was being warped from a modal from the engagement survey repo. The protips have now been moved the the engagement survey repo so that they appear at the bottom of the page.


ptrovatelli:
Hi @Maffooch ,
Is it fixing https://github.com/DefectDojo/django-DefectDojo/issues/1643 ?

Maffooch:
I just realized there is two parts to that issue...
The dropdown part is fixed. I'll make this a WIP and get the other one working too.

ptrovatelli:
great :D can't wait to try it out!

ptrovatelli:
Hello @Maffooch 
It fixes the dropdown menu but the shortcuts are still not working (tested on chromium and firefox)
By the way the pro-tip don't appear anymore. Why do you want to move those to the surveys repos? I think it was fine where they used to be and it's not respecting "separation of concerns" rule (it's not related to surveys so it shouln't be in the surveys repo)
thanks

Maffooch:
These issues stem from the survey repo as it has a `view_eng.html` template that extends the dojo `view_eng.html` template. All of the javascript from the survey repo overwrites that of the original file. The fixes are not working because changes were made to the survey repo. Either building from scratch or reinstalling that repo will load the changes. Regarding the protips, each of the messages are at the very bottom of the screen except for the engagement screen.  I'll move it back to due to the separation of concerns rule if you'd like.

To quickly reinstall through docker-compose:
`docker exec -it --user root <container name>_uwsgi_1 pip install --force-reinstall git+https://github.com/DefectDojo/defectDojo-engagement-survey@master#egg=defectDojo-engagement-survey`

ptrovatelli:
Hi @Maffooch ,
Nicely spotted, the docker exec command did the trick. I had already rebuilt the app but it wasn't enough; this is what I had done: 
```
git fetch origin pull/1713/head:pr-1713
git checkout pr-1713
docker-compose build
docker-compose down --volumes
docker-compose up
```

I think docker isn't able to see that this external repo has changed and uses a cached version. Do you think you could add this command somewhere in the DOCKER.MD saying that it's required to update the surveys code? 
Also I definitely think we should put the least code possible in that repo (just what concerns surveys) as it's pretty tricky to update, so keeping the pro tips in the main repo would be nice yes if that's not too much trouble for you :)
thanks!

ptrovatelli:
>Do you think you could add this command somewhere in the DOCKER.MD saying that it's required to update the surveys code?

on second thought: it doesn't feel right that the same requirements file produces very different applications depending on when the build is done and whether we have already built the application before or not. As for all libraries, I think that the surveys repo should be imported using a fixed version / tag name (instead of master). After each change in the surveys repo, it should be tagged and the django-defectDojo requirements file should be updated to fetch the new tag. What do you think? Ideally we should even pin the version with a hash to make sure that the tag wasn't overwritten, on purpose or not. The sha-1 commit hash could be used in that purpose (same way we pin the docker images with the sha-256)

By the way this also incudes a supply chain risk : if the surveys repo is compromised (or the django-tastypie-swagger or the django-custom-field.git) the django-defectdojo project will be compromised too. Getting the repos by tag name + hash would avoid the compromission because when reviewing django-defectDojo PR, we'll look at the dependent repo and will be able to detect the compromission. 
Hopefully github will replace sha1 by sha2 then we'll really be secure :) . we should be there soon, git itself is in the process of migration from sha1 to sha2: cf https://stackoverflow.com/questions/28159071/why-doesnt-git-use-more-modern-sha

Maffooch:
That's a great idea. I've been wanting to get django-custom-fields and tastypie-swagger to the DefectDojo repo for a while anyways. That as well as adding hashes should be a separate PR for record history. In the meantime, I'll add protips back to their original location, and add the hashing business to my list of things.

Maffooch:
So I changed the ownership real quick before I forgot about it. Requirements have been updated to reflect that for dev here and master in another PR. I should've waited in retrospect.

Maffooch:
@ptrovatelli Repos have been tagged with a version. The same changes on the master branch can be found in #1729.

ptrovatelli:
@Maffooch OK so it doesn't seem possible to get a tag + a commit hash like we do on docker images (https://pip.readthedocs.io/en/1.1/requirements.html#requirements-file-format ) 
so I think it's better to include the hash instead of the tag due to the mutable nature of tags. 
I think you need to leave the egg too according to the doc; example: 
`git://git.myproject.org/MyProject.git@da39a3ee5e6b4b0d3255bfef95601890afd80709#egg=MyProject`

you can still tag the project but it's not all that important if we use the commit only. 
thanks :)

PR #1726, opened 2019-12-27
Move image hash away from finding desc to references

madchap:
While setting up @ptrovatelli 's awesome dedup-per-parser configuration on Anchore, I noticed that dedupe didn't work as expected because of the description's inclusion of "imageDigest" (shows as "Image hash" in the UI). E.g., say you scan 10 docker images that share the same baseline, you can't dedup across them if you include `imageDigest` in the description and despite the fact the findings are indeed the same.

This PR simply moves that information to references.

Therefore, if setting up dedup-per-parser on Anchore, the `description` field can be used. If you want to keep the previous behavior for dedup, you could simply add `references` field to the dedup configuration fields.

### DIFF
--- a/dojo/tools/anchore_engine/parser.py
+++ b/dojo/tools/anchore_engine/parser.py
@@ -33,7 +33,6 @@ class AnchoreEngineScanParser(object):
                 cve = item['vuln']
 
             # Finding details information
-            findingdetail += 'Image hash: ' + data['imageDigest'] + '\n\n'
             findingdetail += 'Package: ' + item['package'] + '\n\n'
             findingdetail += 'Package path: ' + item['package_path'] + '\n\n'
             findingdetail += 'Package type: ' + item['package_type'] + '\n\n'
@@ -48,7 +47,8 @@ class AnchoreEngineScanParser(object):
             mitigation += "Upgrade to " + item['package_name'] + ' ' + item['fix'] + '\n'
             mitigation += "URL: " + item['url'] + '\n'
 
-            references = item['url']
+            references = 'Image hash: ' + data['imageDigest'] + '\n\n'
+            references += item['url']
 
             dupe_key = data['imageDigest'] + "|" + item['feed'] + "|" + item['feed_group'] + "|" + item['package'] + '|' + item['vuln']
 


ptrovatelli:
Hi @madchap :) ,
I don't think that keeping the description in the deduplication keys is the best option for this parser. that description is pretty long and specific (even without the image hash) and will change in many cases while the vuln in it's essence will still be the same. 
Looking at the report, i'd say ideally we should deduplicate on : 
- package_name
- cve
- severity (because i always fear that a less important vuln will be hidden by a more important one. In most case this one will probably be useless but it doesn't hurt)

that way for example you'll be able to track the vuln even if anchore version changes and the description changes a little bit, or when a library is updated but still has the same vulnerability. Also it will allow cross-parser dedupe because it's only standard data. 
Now, package_name is not recorded in defectDojo right now. 
A couple of options are: 
- add it for example in "sourcefile" field which is not really used right now (except by veracode parser). My first choice would be "filepath" but it's already used in this parser for something else
- if you're up to some model change mabe add a new field that will be used specifically for image names (or package name or something generic enough so that we can reuse for other similar things. in clair scanner for example that would be the "featurename" tag from the report)
- the most simple and pretty acceptable solution: deduplicate based on title + severity. you get rid of all issues related to dedupe based on description. you'll only fail detecting duplicates when the image changes version and still has the same vuln because the title includes not only the package name but also it's version)

madchap:
Indeed, your option 3 is what I am actually doing now short-term. 

Adding a specific field for the "artifact" name (i.e. docker image, war file fingerprint or else?) could be nice. What about `sourcefilepath`? As you noted in the source code through a comment, it does not seem to be used anywhere?

ptrovatelli:
> What about `sourcefilepath`? As you noted in the source code through a comment, it does not seem to be used anywhere?

well it's not really a sourcefilepath :) . technically it'll work but maybe we should add missing fields when we find them rather than re-using wrongly named fields

madchap:
You're right. I can take a look at that.

Regardless of that, do you think this PR can still get merged because it makes sense anyways? Or should I jut close it?

ptrovatelli:
i'd close this one :) . I'd say the image hash belongs more to description than references :) . what is usually stored in references is links to cve or cwe details isn't it?

madchap:
I have no strong opinion. I'll close it. Thanks for the feedback :)

madchap:
I am re-opening this, got bitten again today and even I know, this is just super counter intuitive.
I think it does make sense to move it to references actually.

If no one has a strong opinion on this, at least, it would avoid possible mis-interpretation and make my life easier :P

PR #1730, opened 2020-01-01
Fix dedupe for dynamic parsers: fix addition of endpoints in hash_cod…

ptrovatelli:
- Fix dedupe for dynamic parsers: fix addition of endpoints in hash_code computation.
- Add deduplication configuration for ZAP and Qualys
- Set dojo logger to INFO (instead of debug). 

I have noticed that the dynamic parsers such as ZAP and qualys weren't working correctly: they failed to include the endpoints in the hash_code because when the hash_code was computed, the endpoints weren't already built (both in legacy dedupe algo and using the new configurable dedupe)
I'm pretty sure this used to work some time ago but it probably broke at some point. 

By the way i found many calls to "save" when uploading reports. I'm no django expert but if we persist to the bdd each time a "save" is called, there is probably a major performance gain to be done here: 
- finding.save is called twice by serializer.py
- inside of finding.save, django.db.models.save is called twice too
so that's 4 calls to the db I guess for each finding. it may be why importing findings takes so much time. we may be talking x2 or x3 performance gain if we could save just once or twice instead of 4.

I have added a couple of debug logs there (switch back dojo logger to debug in order to see them)

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1382,7 +1382,9 @@ class Finding(models.Model):
                     deduplicationLogger.debug(hashcodeField + ' : ' + str(getattr(self, hashcodeField)))
                 else:
                     # For endpoints, need to compute the field
-                    fields_to_hash = fields_to_hash + self.get_endpoints()
+                    myEndpoints = self.get_endpoints()
+                    fields_to_hash = fields_to_hash + myEndpoints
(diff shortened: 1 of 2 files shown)


ptrovatelli:
@madchap nope all good thanks :)

PR #1750, opened 2020-01-09
Add basic parser for Outpost24 scan format

jvz:
@madchap @jeffret-b @Wadeck @daniel-beck

### DIFF
--- a/dojo/unittests/test_outpost24_parser.py
+++ b/dojo/unittests/test_outpost24_parser.py
@@ -9,6 +9,9 @@ class TestOutpost24Parser(TestCase):
         with open(filename) as file:
             parser = Outpost24Parser(file, Test())
         self.assertEquals(item_count, len(parser.items))
+        if item_count > 0:
+            for item in parser.items:
+                self.assertEquals(1, len(item.unsaved_endpoints), msg='Finding should have one endpoint')
 
     def test_parser_no_items(self):
         self.assert_file_has_n_items('dojo/unittests/scans/outpost24/none.xml', 0)
--- a/dojo/unittests/test_outpost24_parser.py
+++ b/dojo/unittests/test_outpost24_parser.py
@@ -5,7 +5,16 @@ from dojo.models import Test
 
 class TestOutpost24Parser(TestCase):
 
+    def assert_file_has_n_items(self, filename, item_count):
+        with open(filename) as file:
+            parser = Outpost24Parser(file, Test())
+        self.assertEquals(item_count, len(parser.items))
+
+    def test_parser_no_items(self):
+        self.assert_file_has_n_items('dojo/unittests/scans/outpost24/none.xml', 0)
+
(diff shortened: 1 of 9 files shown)


Maffooch:
Don't forget about adding the fixture :)

jvz:
Alright, I've addressed those issues. There's still a little bit of code to add for the endpoints part, though I'll follow up with that later today. Let's make sure this part passes CI at least.

jvz:
Seems to be a test failure still. I'll update later today.

jvz:
Figured it out. @jeffret-b feel free to review now!

dojo/unittests/test_outpost24_parser.py
```diff
@@ -0,0 +1,24 @@
+from django.test import TestCase
+from dojo.tools.outpost24.parser import Outpost24Parser
+from dojo.models import Test
+
+
+class TestOutpost24Parser(TestCase):
```
jeffret-b: Does it make sense to have at least one test that check the finding details?

jeffret-b:
> Remaining to be done before merge

@jvz is this comment in other original description still accurate? Is there still remaining work to be done?

dojo/unittests/test_outpost24_parser.py
jvz: You'd think so, but I didn't see any other tests doing that. 🤷‍♂

jeffret-b: I agree that it doesn't seem to be common in these tests, so maybe it isn't valuable. I did find this example, though, that is doing something along those lines: https://github.com/DefectDojo/django-DefectDojo/blob/ac0502498b467247ad237c5668a868f681d188ca/dojo/unittests/test_checkmarx_parser.py#L165

jvz:
When you merge this, please squash the history. Thanks!

madchap:
@jvz if you could upload that sample file as well in the sibling repo at https://github.com/DefectDojo/sample-scan-files, it would be great. Thanks!

ptrovatelli:
looks pretty neat. 
was this sample.xml file manually modified? there are a lot of duplicates for example `<id>1377090</id>` appearing twice with just `firstseen` changed
if this is a `regular` report, it might be good to do some aggregation upon import; it'll greatly improve the performances by inserted much less lines into the db and also will make the analysis easier.

dependency check amongst other scanners do this, using a dict. ideally update nb_occurences in that case. 

it might be nice to use that falsepositive flag too to fill in the according field in the db

jvz:
> was this sample.xml file manually modified? there are a lot of duplicates for example `<id>1377090</id>` appearing twice with just `firstseen` changed

Yes, I started from a real scan, removed several things, and didn't really update any summary statistics at the top as I wasn't sure how they're calculated by the tool itself. I'm not sure exactly what the `id` element corresponds to, but it's a tool-specific id.

PR #1752, opened 2020-01-12
Kubernetes 1.16 / helm3 improvements

ptrovatelli:
Various fixes and improvments for kubernetes: 
- Fixed yaml identations
- Convert to kubernetes 1.16 using "kubectl convert". This probably didn't change much but this is the first thing I tried when nothing was working for me :) . I have kept the changes which is basically a bit of identation and formating changes; it's more standard this way I guess because it's how kubernetes "natively" idents yaml. 
- add option "django.ingress.activateTLS" : before having this option, if you hadn't setup TLS, the application wasn't working (HTTP 404). Now we have a first step without TLS with a working application. (I haven't tested TLS yet)
- comment out nginx ingress controller annotation: due to this annotation, the application could only work with nginx controller (example: with traefik, HTTP 404 with error "backend not found" when debug is activated). For most users, this annotation is not required because it's only required when you have several ingress controllers. A possible improvment would be to make it configurable in values.yaml
- handle secrets outside of the dependent charts: I had issues with the rabbitmq chart, because when things are not working and you re-install the carts, due to (or thanks to) the statefulset, the first password generated is the correct one and after re-installing, the secret would get populated with a wrong password. By handling secrets outside of the dependent chart and by adding options to re-create or not the secrets, it's easier to handle the lifecycle of the secrets: we can keep the same secret when re-installing so that it still holds the correct value. 
- bump up dependencies (otherwise it wasn't working for kubernetes 1.16)
- bump up chart versions
- fix documentation for use with helm 3
- fix hot reloading in ptvsd env (not related to kubernetes by it had to be done)

### DIFF
--- a/KUBERNETES.md
+++ b/KUBERNETES.md
@@ -23,18 +23,60 @@ cd django-DefectDojo
 
 minikube start
 minikube addons enable ingress
+```
+Helm <= v2
+```zsh
 helm init
 helm repo update
+```
+
+Helm >= v3
+```zsh
+helm repo add stable https://kubernetes-charts.storage.googleapis.com/
+helm repo update
+```
+Then pull the dependent charts:
+```zsh
 helm dependency update ./helm/defectdojo
 ```
 
-Now, install the helm chart into minikube:
+Now, install the helm chart into minikube.
+
+If you have setup an ingress controller: 
+```zsh
+DJANGO_INGRESS_ENABLED=true
+```
+else: 
+```zsh
+DJANGO_INGRESS_ENABLED=false
+```
+
+If you have configured TLS: 
+```zsh
+DJANGO_INGRESS_ACTIVATE_TLS=true
+```
+else: 
+```zsh
+DJANGO_INGRESS_ACTIVATE_TLS=false
+```
+
+Helm <= v2:
 
 ```zsh
 helm install \
   ./helm/defectdojo \
   --name=defectdojo \
-  --set django.ingress.enabled=false
+  --set django.ingress.enabled=${DJANGO_INGRESS_ENABLED},django.ingress.activateTLS=${DJANGO_INGRESS_ACTIVATE_TLS}
+
+```
+
+Helm >= v3:
+
+```zsh
+helm install \
+  defectdojo \
+  ./helm/defectdojo \
+  --set django.ingress.enabled=${DJANGO_INGRESS_ENABLED},django.ingress.activateTLS=${DJANGO_INGRESS_ACTIVATE_TLS}
 ```
 
 It usually takes up to a minute for the services to startup and the
@@ -75,14 +117,22 @@ Log in with username admin and the password from the previous command.
 ### Minikube with locally built containers
 
 If testing containers locally, then set the imagePullPolicy to Never,
-which ensures containers are not pulled from Docker hub.
-
+which ensures containers are not pulled from Docker hub
+(helm 2) :
 ```zsh
 helm install \
   ./helm/defectdojo \
   --name=defectdojo \
-  --set django.ingress.enabled=false \
-  --set imagePullPolicy=Never
+  --set django.ingress.enabled=${DJANGO_INGRESS_ENABLED},django.ingress.activateTLS=${DJANGO_INGRESS_ACTIVATE_TLS},imagePullPolicy=Never
+```
+
+### Installing from a private registry
+If you have stored your images in a private registry, you can install defectdojo chart with (helm 3). 
+
+- First create a secret named "defectdojoregistrykey" based on the credentials that can pull from the registry: see https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
+- Then install the chart with:
+```zsh
+helm install  defectdojo ./helm/defectdojo/ --set repositoryPrefix=<myregistry.com/path>,imagePullSecrets=defectdojoregistrykey,django.ingress.enabled=${DJANGO_INGRESS_ENABLED},django.ingress.activateTLS=${DJANGO_INGRESS_ACTIVATE_TLS}
 ```
 
 ### Build Images Locally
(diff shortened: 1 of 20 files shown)


ptrovatelli:
fixed postgresql configuration: 
- as for mysql, I have left in comments the configuration using the secret from the chart in case someone likes it better (but the key name was erroneous)
- I have added an external secret so that we can keep it with the correct value when re-installing the chart (same as mysql, rabbitmq and redis)

alles-klar:
Nice work! I have some problems with the postgresql setup, related to the password. Did it work in your setup? At the moment I am not sure where the problem exactly is because postgesql and django container have the same password but it does not work.

ptrovatelli:
@alles-klar yes i haved tested with helm 3 and kubernetes 1.16 and 1.17 on a fresh install. are you doing a fresh install or trying to use an existing db? 

 i'm using the secret from outside of the chart so that it's not removed when the chart is un-installed: I install with: 
`helm install  defectdojo ./helm/defectdojo/ --set repositoryPrefixName=myregistry,imagePullSecrets=mytokentoregistry,django.ingress.activateTLS=false
`

The first time it'll create the database and generate a random password. If you need to re-install for any reason but want to keep the same database (including any issue during the first install), you need to use the below command to re-use the existing secrets (see § Re-install the chart in the readme): 

```
helm uninstall defectdojo
helm install  defectdojo ./helm/defectdojo/ --set createSecret=false,createRabbitMqSecret=false,createMysqlSecret=false,createPostgresqlSecret=false,createRedisSecret=false
```

If you manage to unsynchronize your secret with the real database password (which shouln't be easy I hope!) you can start over by deleting the secrets and the postgres pvc. 
```
helm uninstall defectdojo
kubectl delete secrets defectdojo defectdojo-redis-specific defectdojo-rabbitmq-specific defectdojo-postgresql-specific defectdojo-mysql-specific
kubectl delete pvc data-defectdojo-rabbitmq-0 data-defectdojo-postgresql-0
```

Then you can install with the first install command that creates the secrets

alles-klar:
With a fresh install and postgresql.persistence.enabled=false it is working now. So my error may come from permission problems in the file system. 

@ptrovatelli Did you run helm upgrade and helm test? With both commands there seem to be some problems with the correct database password. Not sure where the problem comes from. The environment variables seem to be correct.

ptrovatelli:
I think I did but i can try another test from scratch. it should be working with persistence enabled too

alles-klar:
The issues seems related to the tls config. I will create a separate PR to fix these issues.

ptrovatelli:
@alles-klar have to pulled this last commit: https://github.com/DefectDojo/django-DefectDojo/pull/1752/commits/87018efdcdf7695bb5e7d68a542e45505dd16dc9 it fixes and issue with TLS
I have tested TSL too, it works for me

dsever:
@ptrovatelli,

Regarding generating secrets  for configuration, I don't think it is good idea, each secrets need to be created out of helm (to enable automation and support external secret repositories like Vault). At the other hand if we keep keep, this part  for creating secrets e.g. #createPostgresqlSecret: true" is bad idea, in case of  False, it will generate new template by using Helm without this secrets, so kubernetes will drop this secret, and than in case if pod fails, it will not be able to recreate pod.

I would rather keep default values like: 
```postgresql-password : {{ .Values.postgresqlPassword | default "postgres" | b64enc | quote }}```

and change it by ```--set``` or ```--values```

**Comment by ptrovatelli** [comment]
@dsever could you clarify what you advise? 

what is a bad idea? `createPostgresqlSecret: true` or `createPostgresqlSecret: false` ?

with the newly added parameters everything is possible by configuration:
- do not create the secret and use a custom one (vault or anything)
```
createPostgresqlSecret: false
postgresql.existingSecret : any secret
```
- create the secret automatically: 
```
createPostgresqlSecret: true
existingSecret: defectdojo-postgresql-specific
```

you can also use the secret from the chart like it used to work (`postgresql.existingSecret` unset) but when you uninstall the chart you loose the secret and the database becomes unusable. which is why I added this outside secret.

**Comment by dsever** [comment]
@ptrovatelli Let me explain:

e.g.

Installations of secrets for test:
`
kubectl get secrets
NAME                              TYPE                                  DATA   AGE
default-token-k4nbt               kubernetes.io/service-account-token   3      5m31s
defectdojo-mysql-specific         Opaque                                2      3m45s
defectdojo-postgresql-specific    Opaque                                1      3m45s
defectdojo-rabbitmq-specific      Opaque                                2      3m45s
defectdojo-rabbitmq-token-h4szq   kubernetes.io/service-account-token   3      5m31s
defectdojo-redis-specific         Opaque                                1      3m45s
sh.helm.release.v1.test.v1        helm.sh/release.v1                    1      3m45s
test-defectdojo                   Opaque                                3      3m45s
test-rabbitmq-token-4mwgz         kubernetes.io/service-account-token   3      3m45s
`
e.g secrets:
helm install:
`
DD_ADMIN_PASSWORD: WnkyMnBPdm5wemJJcXdJS2d5YnBNTg==
  DD_CREDENTIAL_AES_256_KEY: ODJ0d2tvZHZGYkZYb2ZlQndmQWxUVmZZQ0JmZlhNYTI=
  DD_SECRET_KEY: bWlERjAxYXpnMUl1ZDl5TEp3UzB2azR3ZWJnbk9YUEo=
`
after  helm upgrade  (createSecret=false) 
data:
  DD_ADMIN_PASSWORD: M1ZOQ0FMbGo1MmxMazBpTEw1dTlwQQ==
  DD_CREDENTIAL_AES_256_KEY: aVBIR0s5R08yb0FYTHczSTZ4SnBGTnlvWUtzZWc2ZnA=
  DD_SECRET_KEY: Wk9vTXZyY1doNFVpUXZpcXkyZTVqdVZxWVhkZEtsZUM=

Completely new sets of secrets 

 Then run:
`helm upgrade --set "createSecret=false" test .
`
will actually remove secret:
`
kubectl get secret test-defectdojo -o yaml
Error from server (NotFound): secrets "test-defectdojo" not found
`
if you list secrets you fill find out, really it is not in list anymore:

`
env) dsever@PN-HR-NB33:~/Pan-Net/test-projects2020/django-DefectDojo/helm/defectdojo$ kubectl get secret
NAME                              TYPE                                  DATA   AGE
default-token-k4nbt               kubernetes.io/service-account-token   3      12m
defectdojo-mysql-specific         Opaque                                2      10m
defectdojo-postgresql-specific    Opaque                                1      10m
defectdojo-rabbitmq-specific      Opaque                                2      10m
defectdojo-rabbitmq-token-h4szq   kubernetes.io/service-account-token   3      12m
defectdojo-redis-specific         Opaque                                1      10m
sh.helm.release.v1.test.v1        helm.sh/release.v1                    1      10m
sh.helm.release.v1.test.v2        helm.sh/release.v1                    1      3m23s
sh.helm.release.v1.test.v3        helm.sh/release.v1                    1      91s

test-rabbitmq-token-4mwgz         kubernetes.io/service-account-token   3      10m
`
Actually you have in https://github.com/DefectDojo/django-DefectDojo/blob/dev/KUBERNETES.md#re-install-the-chart that for preserving secret during uninstall, that's fine it should work, but for upgrade will not.


NOTE: Even I post secret in base64 this are adhoc only to point out the problem.

So what I'm doing, I'm actually using yaml for secrets, but each secret I'm injecting from vault, during the deployment cycle. Ok I can use existingSecrets, but in that case i have to create completely new resource out of provided helm. 

I hope it is now clear, I will get slack approval in few days.

**Comment by ptrovatelli** [comment]
@dsever OK i replicate the issue. I think @alles-klar saw the same problem. 

Those new secrets have the `helm.sh/resource-policy: keep` annotation that tells helm to keep this resource upon uninstall but that has no effect on `helm upgrade`. The same problem exists for the `defectdojo` secret. 

I'm looking for a solution.

**Comment by dsever** [comment]
It is the same for all secrets, this was only example

**Comment by ptrovatelli** [comment]
yes but the defectdojo secret is pre-existing to this PR :)

**Comment by ptrovatelli** [comment]
this seems to work: https://github.com/helm/helm/pull/5290#issuecomment-462512418
```
  annotations:
    "helm.sh/hook": "pre-install"
    "helm.sh/hook-delete-policy": "before-hook-creation"
```

will create a new PR asap.

ptrovatelli:
it's here : https://github.com/DefectDojo/django-DefectDojo/pull/1924

NB: this also has some valuable corrections: https://github.com/DefectDojo/django-DefectDojo/pull/1881

PR #1756, opened 2020-01-16
Improve Qualys Parser

alles-klar:
- add date to Finding
- add qid as unique_id_from_tool
- parser handles active/verified status now

On behalf of DB Systel GmbH

---

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -570,7 +570,7 @@ HASHCODE_FIELDS_PER_SCANNER = {
     # possible improvment: in the scanner put the library name into file_path, then dedup on cve + file_path + severity
     'Whitesource Scan': ['title', 'severity', 'description'],
     'ZAP Scan': ['cwe', 'endpoints', 'severity'],
-    'Qualys Scan': ['title', 'endpoints', 'unique_id_from_tool', 'severity']
+    'Qualys Scan': ['title', 'endpoints', 'severity']
 }
 
 # This tells if we should accept cwe=0 when computing hash_code with a configurable list of fields from HASHCODE_FIELDS_PER_SCANNER (this setting doesn't apply to legacy algorithm)
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -570,7 +570,7 @@ HASHCODE_FIELDS_PER_SCANNER = {
     # possible improvment: in the scanner put the library name into file_path, then dedup on cve + file_path + severity
     'Whitesource Scan': ['title', 'severity', 'description'],
     'ZAP Scan': ['cwe', 'endpoints', 'severity'],
-    'Qualys Scan': ['title', 'endpoints', 'severity']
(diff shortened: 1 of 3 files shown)


dojo/settings/settings.dist.py
```diff
@@ -570,7 +570,7 @@ def generate_url(scheme, double_slashes, user, password, host, port, path):
     # possible improvment: in the scanner put the library name into file_path, then dedup on cve + file_path + severity
     'Whitesource Scan': ['title', 'severity', 'description'],
     'ZAP Scan': ['cwe', 'endpoints', 'severity'],
-    'Qualys Scan': ['title', 'endpoints', 'severity']
```
ptrovatelli: In order to tap into the full potential of the "unique_id_from_tool", I think the configuration could be changed a little: you're still deduplicating based on the hash_code here, but the hash_code is based on unique_id_from_tool + other fields. After qualys version change, the title may change even though it's the same vuln but you'll loose the value of unique_id_from_tool because the hash_code will change in this case. 
provided you have a unique_id_from_tool, the best options are: 
- like "Checkmarx Scan detailed" and "SonarQube Scan detailed", 
  - HASHCODE_FIELDS_PER_SCANNER : leave legacy (not configured) or any of your choice (you may even keep the unique_id_from_tool); that won't really matter
  - DEDUPLICATION_ALGORITHM_PER_PARSER :  DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL.
With this setting, the hash_code is not really important and the dedupe is fully based on the unique id

The ideal though, that I haven't configured yet for lack cross-parser data and time to add the applicable unit tests: 
- hash_code: configure here a set of fields that will enable a cross-parser dedupe. For example:  ['cwe', 'severity', 'file_path'] (nothing parser-specific like title or description)
- DEDUPLICATION_ALGORITHM_PER_PARSER : DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE
- HASHCODE_ALLOWS_NULL_CWE : false if using the cwe as one of the main identifiers

With this settings, you should have the advantages of both worlds: 
- when comparing to a finding of the same test type, you'll deduplicate based on the unique_id_from_tool
- when comparing to a finding of a different parser, the cross-parser compatible fields will be used (the ones that built the hash_code).

ptrovatelli: even though you may use, DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL I think it's best to start now using only cross-parser compatible fields in HASHCODE_FIELDS_PER_SCANNER so that we can switch later to  DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE painlessly
There is always an adaptation period when changing hash_code computation where previously added findings have the wrong hash_code (except if recomputing the full database after the configuration change which is possible with `./manage.py dedupe` btw)

alles-klar: I get your point and agree. One notice for further development: Qualys has an id per vulnerability. So it is only unique if the endpoint is included.

PR #1760, opened 2020-01-17
Added exception catching, error message, and redirect

Apipia:
Was getting pretty annoyed by the **500 Internal Server Error** message when there are exceptions in the parser logic. 

Not sure if this is the _most correct_ way to solve the problem but I gave it a shot.
Reloads the page with an error message that I pull straight from the exception. Up to the coder of the parser logic to make sure the error message is useful. (i.e. Key error message are pretty useless by themselves)

I just print the exception message in the error, could change this to have less information (just something like "there were errors importing scan results" without specifying the errors)

Following screenshot was generated by adding to the hackerone parser:
`raise Exception("Raising an exception in the parser for error testing.")`
image

### DIFF
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -555,7 +555,14 @@ def import_scan_results(request, eid=None, pid=None):
                     new_f.cred_id = cred_user.cred_id
                     new_f.save()
 
-            parser = import_parser_factory(file, t, active, verified)
+            try:
+                parser = import_parser_factory(file, t, active, verified)
+            except Exception as e:
+                messages.add_message(request,
+                                     messages.ERROR,
+                                     "Error importing scan results: {}".format(str(e)),
+                                     extra_tags='alert-danger')
+                return HttpResponseRedirect(reverse('import_scan_results', args=(eid,)))
 
             try:
                 for item in parser.items:
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -36,6 +36,7 @@ from dojo.tasks import update_epic_task, add_epic_task
 from functools import reduce
 
 logger = logging.getLogger(__name__)
+parse_logger = logging.getLogger('dojo')
 
 
 @user_passes_test(lambda u: u.is_staff)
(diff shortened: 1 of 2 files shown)


madchap:
I like the idea.

Do you think this could be made more generic, so that we would not have to touch each parser?

Cheers!

Apipia:
> I like the idea.
> 
> Do you think this could be made more generic, so that we would not have to touch each parser?
> 
> Cheers!

The way it's currently implemented it will display any exception thrown within any parser. So one does not need to touch every parser. It's quite generic. However, it will just display whatever message is in the exception. Meaning that if you get a key error on a dictionary with a key like "start_date", then the error message will say:
"Error importing scan result: start_date"
Which isn't _incredibly_ helpful.

I think, should this be implemented, we should require better error handling in parser logic moving forward. (Catch exceptions like Key Errors and write a better, more user-friendly, message). 

There's also a security concern with just taking whatever error message the code throws and displaying it to the user without any consideration.

But essentially, as is, this will stop the **500 Internal Server Error** page from being displayed on failed parsers, and give some indication as to why it may have failed.

Maffooch:
This is a great change. Maybe the message could be "An error occurred in the parser. Please see error logs for details.", or something along those lines. Super generic, minimal work going forward, and still private/secure.

Apipia:
> This is a great change. Maybe the message could be "An error occurred in the parser. Please see error logs for details.", or something along those lines. Super generic, minimal work going forward, and still private/secure.

Yeah I like that. Only problem I've discovered with this so far is that if you're making a new parser, since it exits gracefully, the stdout doesn't print your error/stack-trace. The only way to figure out what went wrong then would be to slap a break-point in there and debug it.
To get the error to actually appear somewhere (error logs, as you say) what might I add to this exception catch block?

Maffooch:
There is a logging suite found at the very bottom of settings.dist.py that would be perfect here. One similar to the deduplication would be ideal as it prints to stdout. The actual printing calls take place in dojo/utils.py in the sync_dedupe function.

Apipia:
> There is a logging suite found at the very bottom of settings.dist.py that would be perfect here. One similar to the deduplication would be ideal as it prints to stdout. The actual printing calls take place in dojo/utils.py in the sync_dedupe function.

Added a log-line to show the error in the console/dojo logger. 
image

Apipia:
Also I changed the message to be more generic, now it says exactly what you proposed: "An error occurred in the parser. Please see error logs for details"

PR #1767, opened 2020-01-18
Fix PHP Symfony Security Checker import and dedupe

valentijnscholten:
Importing PHP Symfony Security Checker seems to be broken for a long time as the scan_type became out of sync in factory.py in 86f2de355c2cc80ce44ad8b437aa87d8afce89d4 (at least in APIv2)

This small commit changes the id/key in factory.py to match with those in test_type.json etc.

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -571,6 +571,9 @@ HASHCODE_FIELDS_PER_SCANNER = {
     'Whitesource Scan': ['title', 'severity', 'description'],
     'ZAP Scan': ['cwe', 'endpoints', 'severity'],
     'Qualys Scan': ['title', 'endpoints', 'severity']
+    'PHP Symfony Security Check': ['title', 'cve'],
+    # for backwards compatibility because someone decided to rename this scanner:
+    'Symfony Security Check': ['title', 'cve'],
 }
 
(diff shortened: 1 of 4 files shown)


valentijnscholten:
Seems like travis failed for technical reasons, any way I can retrigger a travis build?

valentijnscholten:
Added the necessary dedupe config to settings.dist.py because without dedupe is broken using legacy dedupe.

ptrovatelli:
no need to tune HASHCODE_FIELDS_PER_SCANNER and HASHCODE_ALLOWS_NULL_CWE? the legacy hash_code algo includes a lot of fields which makes it weak at detecting the same vulns when the code changes a bit (due to the inclusion of line number for eg) and at cross parser detection (due to the inclusion of tool specific fields like title and description).

valentijnscholten:
> no need to tune HASHCODE_FIELDS_PER_SCANNER and HASHCODE_ALLOWS_NULL_CWE? the legacy hash_code algo includes a lot of fields which makes it weak at detecting the same vulns when the code changes a bit (due to the inclusion of line number for eg) and at cross parser detection (due to the inclusion of tool specific fields like title and description).

Good point. It seems to work fine with default fields, but I have no configured specific fields. There really are only two fields in the json that can be used.

PR #1784, opened 2020-01-19
dashboard: use aggregate to populate severity pie chart

valentijnscholten:
As I have over 10000 findings in my instance, I'm getting nice visibility of the bottlenecks in Defect Dojo.

The dashboard is showing a pie chart with all-time severity count. It populates this by retrieving ALL findings from the database and let some django code count the severities (no comment).

I've rewritten this to use an aggregate in the queryset.

There's more to be optimized in the dashboard statistics, but the weekend is almost over.


valentijnscholten:
Superseeded by #1796 which contains the same (and more)

PR #1787, opened 2020-01-19
Removing threading from ptvsd entrypoint

madchap:
Running ptvsd with threads and potentially multi-processes will mess with you soon or later.

If you're debugging something anyways, you probably do not need multiple threads, but rather will like to have 100% of your breakpoints work or hot-reload to be super fast.

fyi @ptrovatelli

### DIFF
--- a/docker/entrypoint-uwsgi-ptvsd.sh
+++ b/docker/entrypoint-uwsgi-ptvsd.sh
@@ -17,4 +17,4 @@ exec uwsgi \
   --protocol uwsgi \
   --wsgi dojo.wsgi:application \
   --py-autoreload 1 \
-  --lazy-apps --honour-stdin
+  --enable-threads --lazy-apps --honour-stdin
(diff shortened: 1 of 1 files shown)


ptrovatelli:
this used to be `--enable-threads --lazy-apps --honour-stdin` at 9cb619c4af7678668eade30bc9e20352a4dfc0c2

i haven't tested without `--enable-threads`, I guess it's the same 

have you tested a bit with this configuration?

madchap:
Hey @ptrovatelli,

Yes, it works. However, I read some documentation this morning after reading your answer, and it seems to be somehow "required" to have it for python background threads -- as far as uwsgi. In our case, it does work without it, but i guess it is better to go "by the book". 

Therefore, I pushed a commit to put it back in there. We know it works too, so no problem there.

Thanks.

ptrovatelli:
thanks :)

PR #1796, opened 2020-01-22
dashboard and punchcard performance improvements

valentijnscholten:
As I have over 10000 findings in my instance, I'm getting nice visibility of the bottlenecks in Defect Dojo.

Turns out that the dashboard itself and the underlying get_punchcard_data where big bottlenecks.

**Option 1**: Disable the punchcard. But there is no fun in that :-)
**Option 2**: Do some optimizations.

I went for Option 2. Took a bit of time, but great way to learn lots of things about Python, Django, Django ORM, Mysql and Defect Dojo.

**Changes made:**

- Pie chart:

	The dashboard is showing a pie chart with all-time severity count. 
	It populated this by retrieving ALL findings and associated data from the database and let some django code count the severities.

	I've rewritten this to use an aggregate in the queryset.

- Severity by month graph:

	Was only showing findings reported by the user, so missing a lot for superuser who usually see stuff about all findings.
	Als changed it to use database aggregate instead of counting in Python or making multiple queries.
	

- Punchcard:

	This was retrieving all findings again and iterate over them in Python about 25-30 times.
	I changed it to use aggreagetes in the queryset, which greatly improves the performance, especially for superusers/staff users.
	

I did notice the punchcard is used in two other places:

**Metrics:** Here it's working for me with my data (20000 findings)

I played it safe by adding some exception catching to prevent a possible small error in the dashboard/metric code from breaking the whole page.
	

Let me know what you think.

BTW this replaces #1784 which only contained the pie chart improvement

BTW The numbers shown in the punchcard is mostly wrong, but this was already the case with the old version before this PR. The day with the highest number of issues is correct, all others are wrong. It's a result of the graph javascript wanting the ratio of the radius of the circles of the values instead of the values themselves. And then in javascripts it tries to reconstruct the radius values into absolute issue counts, but there's something wrong there in the javascript and/or the python code.  I didn't youch that really, just changed the logic to get dat absolute numbers from the db.

### DIFF
--- a/dojo/urls.py
+++ b/dojo/urls.py
@@ -196,3 +196,6 @@ if hasattr(settings, 'DJANGO_ADMIN_ENABLED'):
 
 if settings.DEBUG:
     urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
+
+import debug_toolbar
+urlpatterns += [url(r"^__debug__/", include(debug_toolbar.urls))]
--- a/dojo/urls.py
+++ b/dojo/urls.py
@@ -196,6 +196,3 @@ if hasattr(settings, 'DJANGO_ADMIN_ENABLED'):
 
 if settings.DEBUG:
     urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
-
-import debug_toolbar
-urlpatterns += [url(r"^__debug__/", include(debug_toolbar.urls))]
--- a/dojo/home/views.py
+++ b/dojo/home/views.py
@@ -1,10 +1,8 @@
 # #  dojo home pages
 import logging
-from calendar import monthrange
-from datetime import datetime, timedelta
-from math import ceil
-
+from datetime import timedelta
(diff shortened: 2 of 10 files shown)


valentijnscholten:
Just found that the data generated is not correct, please hold.

valentijnscholten:
OK I fixed the code that generated the labels displaying the wrong counts. So should be easier to review now by testing it :-)

dojo/urls.py
```diff
@@ -196,3 +196,6 @@
 
 if settings.DEBUG:
     urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
+
+import debug_toolbar
```
madchap: This blows up on my dashboard loading, `no module named debug_toolbar`.

valentijnscholten:
Pushed some changes:
- Removed reference do debug_toolbar
- Fixed handling of weeks without findings
- Simplified the code slightly
I think you can safely test this, let me know if you find any corner cases.

PR #1800, opened 2020-01-23
A bit fancier release-drafter template

madchap:
This would go hand in hand with more attention to labels when some issues come in.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,4 +1,22 @@
+name-template: 'v$NEXT_PATCH_VERSION 🌈'
+tag-template: 'v$NEXT_PATCH_VERSION'
+categories:
+  - title: '🚀 New scanners'
+    labels:
+      - 'Import Scans'
+  - title: '🚀 Features and enhancements'
+    labels:
+      - 'feature'
+      - 'enhancement'
+  - title: '🐛 Bug Fixes'
+    labels:
+      - 'fix'
+      - 'bugfix'
+      - 'bug'
+  - title: '🧰 Maintenance'
+    label: 'dependencies'
+change-template: '- $TITLE @$AUTHOR (#$NUMBER)'
 template: |
-  ## What's Changed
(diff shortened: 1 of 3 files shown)


ptrovatelli:
will it work if merged on dev or should it be merged to master?

madchap:
I believe it to be a `master` thing, like dependabot.

ptrovatelli:
cause this PR is based on dev :)

madchap:
Well, I am obeying the rules :-D

ptrovatelli:
Is there a way we can see what it's doing?

ptrovatelli:
if not i'm fine with merge and see; can't be worse :)

madchap:
You'd have the sample example I took directly out of https://github.com/marketplace/actions/release-drafter.

release-drafter will generate a draft and add each PR to it. Then it's up to us to check it out, and eventually amend it then. So if it's not to our liking, we can always change it on the fly.

I'd really like for this to get merged during `-rc6`.

ptrovatelli:
are you planning to use actions or the deprecated github app? i suggest we go for actions which is not deprecated
that doc states that we need a workflow file too under .github/workflows/release-drafter.yml . 
there must be way to test release drafter on your own repository: if you add the proper workflow you should see the actions and releases appear on your repo here https://github.com/madchap/django-DefectDojo/actions/new

madchap:
> are you planning to use actions or the deprecated github app? i suggest we go for actions which is not deprecated

actions. We need to ask a leader to remove the app, as it was probably installed before to avoid any conflict.

> that doc states that we need a workflow file too under .github/workflows/release-drafter.yml .
> there must be way to test release drafter on your own repository: if you add the proper workflow you should see the actions and releases appear on your repo here https://github.com/madchap/django-DefectDojo/actions/new

Yes, it is working fine.

valentijnscholten:
let's hope this is a sign of a new release coming ;-)

PR #1802, opened 2020-01-24
cache alert_count within request scope

valentijnscholten:
Current dev branch doesn't cache alert_count when rendering pages. So each http requests results in multiple queries to get the alert count. Although not always very heavy, it makes no sense to do ~5 queries per page in each request.

I don't know everything about django, so couldn't find an easy way to cache this value without too many changes. This PR is to get some feedback.

I know most of you probably don't care about these small queries, but I'd like to see if they can be avoided. In large instances you can have lots of alerts so they can be 10ms each resulting in a 50ms delay for each page.

### DIFF
--- a/dojo/templatetags/navigation_tags.py
+++ b/dojo/templatetags/navigation_tags.py
@@ -9,7 +9,11 @@ register = template.Library()
 
 @register.simple_tag(takes_context=True)
 def alert_count(context):
-    count = Alerts.objects.filter(user_id=context['request'].user).count()
+    count = context.get('dojo_alert_count', None)
+    if count is None:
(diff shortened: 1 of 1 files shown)


madchap:
I don't know enough either, but I do care about performance improvements you're trying to make :-)

madchap:
A suggestion on PR if you're looking for feedback is to open it in "Draft" mode. You should have a little arrow next to the "Create pull request" button which let's you choose that option.

image

valentijnscholten:
Thanks, didn't yet see draft PRs :-)
I will close this one. Although I look this little optimization, I think in the end maybe some ORM Query caching is needed after all. Too many things that would need changing to make performance a lot better without such a cache.

PR #1807, opened 2020-01-25
Make Similar findings play nice with dedupe on engagement

valentijnscholten:
Currently similar findings doesn't respect the deduplication_on_engagement flag and allows users to mark findings from other engagements as duplicate. When you have dedeuplication removal turned on this will remove the duplicate issue which is probably not what the users want in this scenario.

This PR makes sure the similar finding view only returns findings within the same engagement iff deduplication_on_engagement is turned on.

And it prevents the marking as duplicate from happening alltogether by adding some checks to the lower level code.

### DIFF
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -2121,6 +2121,11 @@ def get_missing_mandatory_notetypes(finding):
 def mark_finding_duplicate(request, original_id, duplicate_id):
     original = get_object_or_404(Finding, id=original_id)
     duplicate = get_object_or_404(Finding, id=duplicate_id)
+
+    if original.test.engagement != duplicate.test.engagement:
+        if original.test.engagement.deduplication_on_engagement or duplicate.test.engagement.deduplication_on_engagement:
+            raise ValueError('Marking finding {} as duplicate of {} failed as they are not in the same engagement and deduplication_on_engagement is enabled for at least one of them')
+
     duplicate.duplicate = True
     duplicate.active = False
     duplicate.verified = False
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1326,7 +1326,13 @@ class Finding(models.Model):
 
     @property
     def similar_findings(self):
-        filtered = Finding.objects.filter(test__engagement__product=self.test.engagement.product)
+        filtered = Finding.objects.all()
+
+        if self.test.engagement.deduplication_on_engagement:
+            filtered = filtered.filter(test__engagement=self.test.engagement)
+        else:
+            filtered = filtered.filter(test__engagement__product=self.test.engagement.product)
+
         if self.cve:
             filtered = filtered.filter(cve=self.cve)
         if self.cwe:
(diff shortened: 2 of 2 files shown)


madchap:
fyi @jvz

dojo/utils.py
```diff
@@ -266,6 +266,11 @@ def deduplicate_uid_or_hash_code(new_finding):
 
 
 def set_duplicate(new_finding, existing_finding):
+    if existing_finding.test.engagement != new_finding.test.engagement:
```
ptrovatelli: i'm confused with this portion of code. the debug log added states that we're skipping dedupe but we're not skipping dedupe, we're just adding a log
moreover, the deduplication on engagement mismatch verification is already done before calling the set_duplicate function and set_duplicate is not called in case of mismatch

valentijnscholten: It was just intended as a safeguard as I thought there were maybe more present or future locations from where functionality could accidentally set mark a finding as a duplicate from another engagement. For example the similar finding functionality, but it turns out that it has its own logic in its views.py.

The code does stop the marking of the finding as a duplicate, so the log is correct? Maybe better wording needed if it's not clear. 

Or just remove the check altogether and trust future contributors to perform a check engagement matching before calling this method.

ptrovatelli: maybe i'm missing something but i'm only seeing two if statements that lead to a log. apart from this log this isn't doing anything in the set_duplicate function, is it?

ptrovatelli: no matter it those `if` are true or not, the rest of the function will be executed

valentijnscholten: ah, you're right there's missing a 'return' or 'else'. Will update tomorrow.

ptrovatelli: if right now this is correcly handled before calling the function, i don't think it's a good idea to add another verification in this function as it'll add noise and make the code harder to understand. as the function name states, this was intended to just mark as duplicate; not do some verifications to see if we should or not mark as duplicate :)

What can be done is add a comment about what this function is doing and that it should only be called when we're sure we want to mark as duplicate and there is no engagement mismatch. what do you think?

valentijnscholten: agreed, I was already in doubt when I was adding this check. I have removed it. I did add it to finding/views.py as it seems that views are the place in django where input validation can/should take place. let me know what you think.

valentijnscholten:
Made some changes and did a force push to avoid having all kinds of strange / changed / reverted commits.

valentijnscholten:
Probably it should, yes. For me this was just a quick fix to respect the dedupe on engagement flag and i'm not too fond changing more things here. It would be rare to have a line number but no filepath anyway.

madchap:
@valentijnscholten I'll let you merge whenever you feel you're 100% ready, just in case.

ptrovatelli:
> Probably it should, yes. For me this was just a quick fix to respect the dedupe on engagement flag and i'm not too fond changing more things here. It would be rare to have a line number but no filepath anyway.

yehp it's totally unrelated with this PR;  my point is not about having a line number without a filepath; the point is, comparing line number independently of the filename doesn't make sense

PR #1835, opened 2020-02-01
show local dates for tests target_start and target_end

### LINKED ISSUE #970: Date of test (scan) in UI off by 1
### Issue Type 
- [X] Bug

### Description
When importing a scan via UI or APIv2, the date that is displayed afterwards in the DD UI is off by one (1 day earlier)

### Operating System 
Debian 9

### Install

- [X] setup.bash

### DefectDojo Version
dev branch 22/3/2019 (46692a0f7bf568f5739d093ab0e4d2ef6733faf6)

### Steps to Reproduce
import a npm audit scan into an engagement using UI or APIv2. Select today as import date.
For example attached report.json. 
Observe after import the date column for the scan displays yesterdays date

valentijnscholten:
Currently the list of tests in an engagement and also the view tests page shows the target_start and target_end date without localizing them.

I'm in CET and this means tests will be 1 day off always. On import the datetime is set to UTC in the database. This works fine for datetimes, which are localized correctly by django.
But when shortening them to the date part only, it fails.

i.e. target_start = 'Feb 01, 2020' on import, results in 2020-01-31 23:00:00 in the database, which is shown on screen as 'Jan 31, 2020' (instead of 'Feb 01, 2020').

~~I don't know why this is and found lots and lots of (stack overflow) posts around the same issues.~~

~~In the end I just created a filter that helps us to localize the date.~~

Fixes #970

### DIFF
--- a/dojo/templates/dojo/view_eng.html
+++ b/dojo/templates/dojo/view_eng.html
@@ -293,7 +293,7 @@
                                        </li>
                                    </ul>
                                 </td>
-                                <td>{{ test.target_start|local_date }} - {{ test.target_end|local_date }}</td>
+                                <td>{{ test.target_start|date }} - {{ test.target_end|date }}</td>
                                 <td>
                                   {% if test.lead.get_full_name and test.lead.get_full_name.strip %}
                                       {{ test.lead.get_full_name }}
--- a/dojo/templates/dojo/view_eng.html
+++ b/dojo/templates/dojo/view_eng.html
@@ -293,7 +293,7 @@
                                        </li>
                                    </ul>
                                 </td>
-                                <td>{{ test.target_start.date }} - {{ test.target_end.date }}</td>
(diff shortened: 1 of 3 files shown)


valentijnscholten:
Found out what was happening. DD was using XXXX.date, where django's builtin filter works better, so using XXXX|date.
Probably lots of places where it should be fixed, let me know if you want me to go ahead and do that.

I saw that the engagement list is already doing this, but using a non-default format:

```
{{ eng.target_start|date:"jS F" }}
```

But in other places I see

```
 {{ eng.target_start|date:"SHORT_DATE_FORMAT" }}
```

Shall I just go ahead and change all places where there's `.date` begin used into `|date`? And choose a format that fits best with what's already there or on similar places?

Would be nice to get rid of these corner cases where the date is one day off confusing everyone, especalliy in the ci/cd world :-)

PR #1847, opened 2020-02-05
TLS Improvments

jtg-dbs:
Generate TLS Certificate

The Nginx generates a TLS Certificate on every start if there is the environment variable is not set. You can add your own cetrificate by mounting it to /etc/nginx/ssl/ and setting GENERATE_TLS_CERTIFICATE to false. You can activate the TLS configuration fo nginx with the environment variable USE_TLS.

From my point of view it is an security issue that the certificate is inlcuded during the docker build step. It should be added at the runtime not at buildtime.

Hardening on the TLS Configuration of nginx.

On behalf of DB Systel GmbH

----

- [ x] Your code is flake8 compliant 
- [ x] Your code is python 3.5 compliant
- [ x] If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR.
- [ x] Model changes must include the necessary migrations in the dojo/dd_migrations folder.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -198,18 +198,19 @@ docker-compose down --volumes
 ```
 
 # Run with docker using https
+## use your own  Credentials
 To secure the application by https, follow those steps
 *  Generate a private key without password
 *  Generate a CSR (Certificate Signing Request)
 *  Have the CSR signed by a certificate authority
 *  Place the private key and the certificate under the nginx folder
-*  Replace nginx/nginx.conf by nginx/nginx_TLS.conf
-*  In nginx.conf, update that part: 
+*  copy your secrets into: 
 ```
         server_name                 your.servername.com;
-        ssl_certificate             /yourCertificate.cer;
-        ssl_certificate_key         /yourPrivateKey.key;
+        ssl_certificate             /etc/nginx/ssl/nginx.crt
+        ssl_certificate_key        /etc/nginx/ssl/nginx.key;
 ```
+*set the GENERATE_TLS_CERTIFICATE != True in the docker-compose.override.https.yml
 * Protect your private key from other users: 
 ```
 chmod 400 nginx/*.key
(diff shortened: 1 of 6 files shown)


NRGLine4Sec:
Hello, Why you don't add TLS 1.3 ?

jtg-dbs:
added TLS1.3

ptrovatelli:
hello @jtg-dbs ; 

getting the certificate into the container at build time was just an easy way to deploy a custom certificate to the container with an already known command: that of building the nginx image + an override (see docker-compose.override.https.yml). I agree it may be confusing though when several instances must use the same image with a different certificate, altough it's always possible to change the certificate after build time. 
 Could you update DOCKER.MD please? that part: `Run with docker using https` with the commands that need to be run in order to place a custom certificate into the container and see what will happen to docker-compose.override.https.yml with your solution? for example your changes in docker-compose.yml are already there (just with a different port).
also, what list of ciphers is that? can you maybe drop a comment about it in the code? 
thanks!

jtg-dbs:
@ptrovatelli  
Thank you for the comment. i updated the MD and inserted the link to the mozilla ssl generator which i used.

Dockerfile.nginx
```diff
@@ -57,17 +57,23 @@ RUN \
...
-  if [ -f /*.key -o -f /*.cer ]; then chown 1001 /*.key /*.cer; fi && \
+  mkdir -p /etc/nginx/ssl && \
+  chown 1001 /etc/nginx/ssl/ &&\
   true
 ENV \
   DD_UWSGI_PASS="uwsgi_server" \
   DD_UWSGI_HOST="uwsgi" \
-  DD_UWSGI_PORT="3031"
+  DD_UWSGI_PORT="3031" \
+  GENERATE_TLS_CERTIFICATE="true" \
```
ptrovatelli: shouldn't that be "True" / "False"? (capital T/F)

DOCKER.md
```diff
@@ -218,6 +219,15 @@ chmod 400 nginx/*.key
 
 ```docker build  -t defectdojo/defectdojo-nginx -f Dockerfile.nginx .```
 
+* Run defectDojo with: 
+```
+rm -f docker-compose.override.yml
+ln -s docker-compose.override.https.yml docker-compose.override.yml
+docker-compose up
```
ptrovatelli: this appears twice. can you put it once only please? maybe create two sections with the pre-requisites for: 
- using a self-signed certificate
- using your own certificate (where to copy the certificate, set generate_certificate=False)

Then describe the common part to both (start with the https override)

ptrovatelli:
other than that it's fine; thanks and sorry for the delay :)

jtg-dbs:
thanks @ptrovatelli for the comments.

PR #1853, opened 2020-02-07
Fix import error for Crashtest Security importer

Phylu:
Due to a minor change in the output format, the importer crashes and produces a 500 server error.

A test file in the new format is provided here: https://github.com/DefectDojo/sample-scan-files/pull/45

### DIFF
--- a/dojo/tools/crashtest_security/parser.py
+++ b/dojo/tools/crashtest_security/parser.py
@@ -48,13 +48,21 @@ class CrashtestSecurityXmlParser(object):
             if failure is None:
                 continue
 
-            title = node.get('name').upper()
+            title = node.get('name')
+            # Remove enumeration from title
+            title = re.sub(r' \([0-9]*\)', '', title)
+
+            # Attache CVEs
             if "CVE" in title:
                 cve = re.findall(r'CVE-\d{4}-\d{4,7}', title)[0]
             else:
                 cve = None
             description = failure.get('message')
-            severity = failure.get('type')
+            severity = failure.get('type').capitalize()
+
+            # This denotes an error of the scanner and not a vulnerability
+            if severity == "Error":
+                continue
 
             find = Finding(title=title,
                            description=description,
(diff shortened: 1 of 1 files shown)


dojo/tools/crashtest_security/parser.py
```diff
@@ -48,13 +48,21 @@ def get_items(self, tree, test):
...
                 continue
 
-            title = node.get('name').upper()
+            title = node.get('name')
+            # Remove enumeration from title
+            title = re.sub(r' \([0-9]*\)', '', title)
+
+            # Attache CVEs
             if "CVE" in title:
                 cve = re.findall(r'CVE-\d{4}-\d{4,7}', title)[0]
```
madchap: I would probably seize the chance here to modify the regex, and replace the `7` by a `10` to accommodate MITRE's number extension.

Phylu: Thanks for the good hint. I will update this.

dojo/tools/crashtest_security/parser.py
```diff
@@ -48,13 +48,21 @@ def get_items(self, tree, test):
             if failure is None:
                 continue
 
-            title = node.get('name').upper()
+            title = node.get('name')
+            # Remove enumeration from title
+            title = re.sub(r' \([0-9]*\)', '', title)
```
madchap: hmm.. I would probably be stricter on the regex here. This means that any number or hyphen within the whole title would be stripped out.

madchap: Not exactly "anywhere" but well.. you I am sure get the point :) (not sure what crashtest titles look like, but assuming the worse is probably a good thing)

Phylu: It should not match the hyphen, but only numbers. Still, I will restrict it to the end of the string where this appears.

Phylu:
The flake8 test seems to have failed due to a network error on the travis server. Flake8 runs through without errors locally. Is there any way to restart that build?

PR #1854, opened 2020-02-07
Add artifact metadata to findings

jvz:
This adds two new finding fields, `artifact_name` and `artifact_version`, which correspond to the underlying artifact metadata for several scanners. I've updated a handful of the existing parsers to insert this data where possible, and I've updated both the findings list and findings detail pages to show this info when available.

@madchap 


madchap:
@ptrovatelli just in case, this is something we had talked about related to the anchore dedup thing (e.g. description not being well and the title having the version and such), if you remember.

ptrovatelli:
great! shouldn't we start on documenting which fields are used on what purpose for what kind of scanners? i totally see how ppl will wonder which should go in file_path or artifact_name given that existing scanner may still be using file_path when it's actually more an artifact. maybe in that CONTRIBUTING.md ?

jvz:
> great! shouldn't we start on documenting which fields are used on what purpose for what kind of scanners? i totally see how ppl will wonder which should go in file_path or artifact_name given that existing scanner may still be using file_path when it's actually more an artifact. maybe in that CONTRIBUTING.md ?

I think it'd be a great idea to have some sort of common guide for implementing tool parsers!

valentijnscholten:
Some thoughts:
- Dependency Check might be a parser that also could/should set the artifact name (on top of filepath)
- Do you have a screenshot of what the findings list looks like now? I think it is already quite dense with the recent addition of 2 or 3 JIRA related columns.

jvz:
Findings list:
Screen Shot 2020-02-10 at 10 08 28 AM

Hover on the file icon:
Screen Shot 2020-02-10 at 10 09 00 AM

Finding details:
Screen Shot 2020-02-10 at 10 11 17 AM

madchap:
Apparently, Travis is throwing a migration error `CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0028_finding_indices, 0028_finding_component in dojo).` Can you please take a look and fix?

Otherwise, lgtm.

I think the "doc" issue is bigger than documenting the finding model actually.

dojo/settings/settings.dist.py
```diff
@@ -592,7 +592,7 @@ def generate_url(scheme, double_slashes, user, password, host, port, path):
 # List of fields that are known to be usable in hash_code computation)
 # 'endpoints' is a pseudo field that uses the endpoints (for dynamic scanners)
 # 'unique_id_from_tool' is often not needed here as it can be used directly in the dedupe algorithm, but it's also possible to use it for hashing
-HASHCODE_ALLOWED_FIELDS = ['title', 'cwe', 'cve', 'line', 'file_path', 'description', 'endpoints', 'unique_id_from_tool', 'severity']
+HASHCODE_ALLOWED_FIELDS = ['title', 'cwe', 'cve', 'line', 'file_path', 'component_name', 'component_description', 'description', 'endpoints', 'unique_id_from_tool', 'severity']
```
madchap: Shouldn't it be `component_version` and not `component_description` ?

valentijnscholten:
There are new migrations in `dev`, could you please update this PR accordingly?

madchap:
@ptrovatelli @valentijnscholten rebase was done here, if we can get this through before the next round.

@jvz #1602 would have to be rebased again after this one gets merged.

valentijnscholten:
I think this can be helpful, just wondering if we want to also include some more parsers? Currently I am using 3 parsers and for none of them these new fields are populated by this PR :-)
- Dependency Check
- NPM Audit
- Symfony security Check
By the looks of it all of them have the info in some shape or form in their reports.

madchap:
> I think this can be helpful, just wondering if we want to also include some more parsers? Currently I am using 3 parsers and for none of them these new fields are populated by this PR :-)
> 
> * Dependency Check
> * NPM Audit
> * Symfony security Check
>   By the looks of it all of them have the info in some shape or form in their reports.
>   Not saying it's a blocker, just thinking it's maybe easy to add now.

I'd be grateful if it could be done in a 2nd step, so that we do not postpone the migration and it sinks in.

PR #1859, opened 2020-02-09
use local dates for tests and engagement start/end dates

valentijnscholten:
Following #1835 I found some more places where target_start and target_end for engagement/test where not localized before being displayed.

In theory I think we might have to apply the "|date" filter everywhere also where finding.date, note.date etc is being used. However for now I think the test/engagement dates are the most sensitive to a "1 day off issue", so I fixed those first and will leave the rest as-is for now.


madchap:
@valentijnscholten let's not forget the labels on PR ;-)

PR #1862, opened 2020-02-10
Finding:  Fix unbound variable error in `edit_template`

Nilix007:
This errors occurs in case of a save request with an invalid form:

    Internal Server Error: /template/XXX/edit
      [...]
      File "./dojo/finding/views.py", line 1242, in edit_template
        'count': count,
    UnboundLocalError: local variable 'count' referenced before assignment

### DIFF
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -1676,7 +1676,6 @@ def edit_template(request, tid):
     template = get_object_or_404(Finding_Template, id=tid)
     form = FindingTemplateForm(instance=template)
 
-    apply_message = ""
     if request.method == 'POST':
         form = FindingTemplateForm(request.POST, instance=template)
         if form.is_valid():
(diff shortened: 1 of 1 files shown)


Maffooch:
I could not reproduce with the info provided here. Is this still an issue on the current dev branch @Nilix007? If so, can you add a little extra info to reproduce?

Nilix007:
Hi Cody,

yep, it's still an issue on the dev branch.

The error occurs when somebody is editing a template, wants to save it and the validation fails (e.g. the data in the cve field does not match the regex).

Then, `request.method == 'POST'` on line 1717 is true and `form.is_valid()` on line 1719 is false.  In this case, `count` is never initialized before its usage on line 1750.

https://github.com/DefectDojo/django-DefectDojo/blob/945bd232576b843998d50e2253f7e823cfba5883/dojo/finding/views.py#L1717-L1753

HTH,
Felix

PR #1866, opened 2020-02-11
Bugfix dashboard do not reraise e

valentijnscholten:
There is a try..except around the dashboard punchcard code to help iron out corner cases.
The `raise(e)` in there was accidentally left from debugging, this PR removes it.


Maffooch:
I can't figure out why snyk would fail here.

madchap:
fyi @Apipia

PR #1881, opened 2020-02-14
kubernetes: fix node selectors; add limits

ptrovatelli:
- add a nodeselector in the initializer job
- fix nodeSelector in django deployment (was at the root, should be under django as in values.yaml)
- add limits

### DIFF
--- a/helm/defectdojo/templates/django-deployment.yaml
+++ b/helm/defectdojo/templates/django-deployment.yaml
@@ -108,7 +108,7 @@ spec:
           failureThreshold: 12
         resources:
           {{- toYaml .Values.django.nginx.resources | nindent 10 }}
-      {{- with .Values.nodeSelector }}
+      {{- with .Values.django.nodeSelector }}
       nodeSelector:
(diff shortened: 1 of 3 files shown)


madchap:
Don't forget the label... ;-)

ptrovatelli:
> Don't forget the label... ;-)

arf sorry :)

ptrovatelli:
adding limits; sorry for mixing different things but handling different branches is complicated
we used to have requests configured but not limits: requests will tell to which node the pod can be deployed but it won't prevent the pod to take up all RAM in the node. 
limits is here for that.

setting big limits for starters, can be tuned down if necessary.

ptrovatelli:
removed the celery nodeselector and limits now that this pod doesn't exist anymore :)

ptrovatelli:
(rebase + force-push to try and fix travis that fails without a reason)

ptrovatelli:
@madchap @alles-klar care to have a look at this one? thanks!

PR #1896, opened 2020-02-19
bugfix: include verified findings in product metrics on product detai…

### LINKED ISSUE #1875: Validate a finding exclude it from anything
**Bug description**
When checking the box "Verified" in a finding, this finding is excluded from "Open findings". 
The metrics are excluding this finding from the findings count.
The score grade is going crazy and switch from C grade to F grade on a testing product.

**Steps to reproduce**
1. Go to findings
2. Update the status of a finding to "Verified"
3. Go back to the product home
4. See the metrics and the score grade

**Expected behavior**
A verified finding should be not be counted as a closed finding !

**Environment information**
DefectDojo 1.5.4 RC6

EDIT: Update the expected behavior

valentijnscholten:
(quick)fixes #1875

There are more places where I think findings are only included in metrics when they are NOT verified. This sounds counter intuitive to me, but I would like to check with the other developers / owners what the idea of this verified field was before changing a lot of queries.
I do think it makes sense to change this one query here because the product details page is probably used a lot and currently never shows any findings, unless you mark them as NOT verified.

### DIFF
--- a/dojo/product/views.py
+++ b/dojo/product/views.py
@@ -94,7 +94,6 @@ def view_product(request, pid):
 
     open_findings = Finding.objects.filter(test__engagement__product=prod,
                                                 false_p=False,
-                                                verified=False,
                                                 active=True,
                                                 duplicate=False,
                                                 out_of_scope=False).order_by('numerical_severity').values('severity').annotate(count=Count('severity'))


madchap:
I think this makes sense. Removing the filter ought to do it.
Verified findings should be counted, except in the circumstances where only Open ones should, obviously. I have not checked the "blame", but I would guess some of these issues come from some of my previous PRs. Thanks for the eyes!

dougmorato:
@valentijnscholten Not to steal your thunder, I think there are 2 more locations where changes need to happen:

https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/product/views.py#L205 

https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/product/views.py#L240

In my mind (and what I've done locally on my prod instance), this line ( verified=.... ) needs to be removed alltogether, because my interpretation of an open finding is the following: 

- An active finding (active=True)
- that is verified OR not verified (findings that have yet to be confirmed) 
- AND is not out of scope (out_of_scope=false)
- AND is not duplicate (duplicate=False)
- AND is not mitigated 
- AND is not false positive

@madchap  Need your interpretation here as well

valentijnscholten:
Yeah let's create a new pr to define what is active and what is open and fix all queries. When accepted they are neither until acceptance expires. In my opinion.

madchap:
Did we have a new PR for that?

valentijnscholten:
No I posted some table in slack to start discussing it but no replies so far haha, maybe I should create an issue.

PR #1901, opened 2020-02-20
Fix duplicates and whitesource

MarianG:
We had an issue where we encountered circular dependencies for duplicates in findings. We assume it occured because of multi-threading and the asynchronous duplicate-detection. We had issues where Finding A had duplicates B, C, and D. B had the duplicate C, C had the duplicate D and D had B. Then the clean-up job that should delete some duplicates produced a recursion error, because it tries to delete duplicates of a finding before deleting the finding. 
I understand that in theory a duplicate should not have a duplicate on its own, but it happened.
We updated the *set_duplicate* method and included a way to identify and repair loops. 
We also found the duplicate_list field is unnecessary, since we have the information in the original_finding field (backwards from duplicate_finding Foreign Key)

The other change is for the whitesource parser. We fixed an issue where the parser fails when the CVE is not set. Additionally the updated title resulted in problems for the duplicate detection of old findings, since now the lib_name is used, which changed the title. We included a method to recreate all old titles and recalculate the hash to make duplication work again.

On behalf of DB Systel GmbH


dojo/tasks.py
```diff
@@ -291,20 +311,20 @@ def async_update_findings_from_source_issues(*args, **kwargs):
...
-            duplicate_list = finding.duplicate_list \
+            duplicate_list = finding.original_finding \
                     .filter(duplicate=True).order_by('date')
             dupe_count = len(duplicate_list) - dupe_max
             for finding in duplicate_list:
-                deduplicationLogger.debug('deleting finding {}:{} ({}))'.format(finding.id, finding.title, finding.hash_code))
+                logger.info("Debug-Dupe: Delete Finding with id: %d", finding.id)
                 finding.delete()
                 dupe_count = dupe_count - 1
                 if dupe_count == 0:
```
valentijnscholten: I think there is an unintended change from `deduplicationLogger` to `logger`?

valentijnscholten:
Would be good if @ptrovatelli takes a look (the king of dedupe).

dojo/tasks.py
MarianG: ah yes. I think this can be changed back

MarianG: should i revert this to deduplicationLogger?

madchap:
I don't know enough of the internal dedup, but I welcome the dedup tests for sure.

The whitesource change is minor and is fine. The renaming piece looks OK as well.

MarianG:
I am also not sure why the smoke test fails. If I run this locally it looks good.

madchap:
So, should this PR now be closed in favor of #2009, #2010 and #2011 ?

MarianG:
yes I think we can close. I just wanted to wait until the tests, for the other PR's #2009 #2010 #2011 are successful

Issue #1903, opened 2020-02-21
Easy single-container run for demo/dev purposes


AlexanderTyutin:
**Is your feature request related to a problem? Please describe**
When I was trying to install the defectDojo the first time it was quite complicated:
- some lack of real documentation (howtos and references)
- repeat a lot of steps to deploy the application
- 5 containers (in docker-compose version) to run a simple demo or a dev version

**Describe the solution you'd like**
What I did:
- automation of single-container environment creation with persistent storage
- transparent README

**Describe alternatives you've considered**
Only the alternative is the docker-compose version but it is more complicated than this version.

**Additional context**
Why it is useful:
- you need run only two commands for running full-functional environment (show your boss or interviewer)
- you can easily run new local environment to test something or begin work on a new branch

So I will be glad if you'll assign this task to me :)

AlexanderTyutin:
Created PR https://github.com/DefectDojo/django-DefectDojo/pull/1905

PR #1905, opened 2020-02-21
Easy single-container run scripts (enhancement #1903)

AlexanderTyutin:

- [ ] Your code is python 3.5 compliant


valentijnscholten:
I think it can be helpful to people to have this, it will require maintenance though.

AlexanderTyutin:
Also prepared documentation [here](https://github.com/AlexanderTyutin/Documentation/blob/master/docs/running-in-single-container.rst)

alles-klar:
I don't know if we really need another setup method. We already have bash, docker-compose and kubernetes/helm. For a demo we have the demosite. It will be very hard to maintain als these methods. I agree that we have so improve the documentation and maybe simplify the setup but I don't think another container solves this problem.

madchap:
I have never had much of an issue just doing `docker-compose up`, so I am not leaning towards having an extra method either.

However, improving documentation could be it.

We're also working on having more regular and frequent releases, so it should ease the process too. I believe quite some frustration stemmed from the fact people had to use `dev`, which not everyone wants to do.

AlexanderTyutin:
@alles-klar Maybe the demo app is cool but I see this caption for long time 
image

And this is the first reason of my tryings to get quick and ready environment to put first data and show it in my company.

The next reason is that we don't use docker-compose. And the third reason is that when we was tried to run defectDojo in local K8s we've got tons of errors in nginx and the question "why we need 5 containers if we do not have such huge amount of vulnerabilities and services" :)

But if these scripts are off-top and there is no need with them the maintainers may feel free to close the issue :) I'll put it in my github library because they are really helpful for me and my environment.

valentijnscholten:
I am also more leaning towards improving the existing docker(-compose) setup instead of a new single container variant. Especially since the single container scripts seem to involve multiple manual steps for initializing things.

alles-klar:
> @alles-klar Maybe the demo app is cool but I see this caption for long time

You are right, but there is a person working on it. Should be available soon. 
> 
> And this is the first reason of my tryings to get quick and ready environment to put first data and show it in my company.
> 
> The next reason is that we don't use docker-compose. And the third reason is that when we was tried to run defectDojo in local K8s we've got tons of errors in nginx and the question "why we need 5 containers if we do not have such huge amount of vulnerabilities and services" :)

Docker is designed to have one process per container. One reason for this is scalability. But I agree that it is may a bit of overhead for a simple demo. 

Did you deploy DefectDojo in K8s with the helm? Should be very straight forward. There is also a  	"KUBERNETES.md" which includes some documentation.

> 
> But if these scripts are off-top and there is no need with them the maintainers may feel free to close the issue :) I'll put it in my github library because they are really helpful for me and my environment.

Would be sad to waste your effort. Feel free to add your scripts here: https://github.com/DefectDojo/Community-Contribs

alles-klar:
https://github.com/DefectDojo/Community-Contribs/pull/10

PR #1922, opened 2020-02-24
Product/Product Type: Filter count for active findings

wichmannpas:
As all occurrences of the value of `findings_count` link to the list
of *open* findings, this commit changes the count to only include
active issues. Otherwise, the value has a discrepancy between the
shown number of findings and the actual length of the list when
following the link.

(Generally speaking, this should be changed to use `annotate` rather than retrieve the value from the database for every row of the table as it is currently done)

- [x] This is not a new feature
- [x] No model changes
- [x] This is not a new feature.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -376,6 +376,7 @@ class Product_Type(models.Model):
 
     def findings_count(self):
         return Finding.objects.filter(mitigated__isnull=True,
+                                      active=True,
                                       verified=True,
                                       false_p=False,
                                       duplicate=False,
@@ -577,6 +578,7 @@ class Product(models.Model):
     @property
     def findings_count(self):
         return Finding.objects.filter(mitigated__isnull=True,
+                                      active=True,
                                       verified=True,
                                       false_p=False,
                                       duplicate=False,


madchap:
Given some other modifications that were made -- and probably an oversight on my part --, I believe your change makes sense and we should remove the `verified=True`, to have both active findings and active,verified findings.

@valentijnscholten would you agree?

valentijnscholten:
Yes, but I would have to check what is in my PR #1899 as that one rewrites the queries into prefetch. Easiest maybe to remove it in that PR to avoid conflicts?

madchap:
> Yes, but I would have to check what is in my PR #1899 as that one rewrites the queries into prefetch. Easiest maybe to remove it in that PR to avoid conflicts?

It would make sense for you to include that change on your side then.

valentijnscholten:
Thank you @wichmannpas for the PR. In my #1899 PR this was already fixed for Product_Type. I just also added the same fix for Product. Could you let us know if that looks OK to you?
I do notice there is still a difference across Defect Dojo on wait "Open" or "Active" means, especially with regards to the "Verified" property. I guess we have to define/document how it is supposed/designed to work and then check all the queries (different issue/PR) :-)

wichmannpas:
At least for the case where the `findings_count` of `Product` is not annotated for some reason, the filter for `active` is not there: https://github.com/DefectDojo/django-DefectDojo/blob/d38207133cfcac3d6b1ea3c2af57b06247522159/dojo/models.py#L586.

Furthermore, I would rather remove the commented-out `findings_count` method definition in the `Product_Type` (if it should be required to restore it at some point, there is the Git history which knows all the history …)

Except for that, I guess your PR should be fine to replace this one.

valentijnscholten:
Sorry, forgot to push. It's now there https://github.com/DefectDojo/django-DefectDojo/blob/76bb0a649db9f12b0cca8402c5d62f9663d04e1c/dojo/models.py#L587

valentijnscholten:
Closing as it's incorporated in #1899

PR #1924, opened 2020-02-24
Fix kubernetes helm upgrade

ptrovatelli:
- helm upgrade was re-initializing the secrets that are stored inside of the defectdojo chart
- do not create those specific secrets by default

see comments at the end of https://github.com/DefectDojo/django-DefectDojo/pull/1752

NB: this also has some valuable corrections: https://github.com/DefectDojo/django-DefectDojo/pull/1881

### DIFF
--- a/KUBERNETES.md
+++ b/KUBERNETES.md
@@ -60,6 +60,8 @@ else:
 DJANGO_INGRESS_ACTIVATE_TLS=false
 ```
 
+Warning: Use the `createSecret*=true` flags only upon first install. For re-installs, see `§Re-install the chart`
+
 Helm <= v2:
 
 ```zsh
@@ -67,8 +69,12 @@ helm install \
   ./helm/defectdojo \
   --name=defectdojo \
   --set django.ingress.enabled=${DJANGO_INGRESS_ENABLED} \
-  --set django.ingress.activateTLS=${DJANGO_INGRESS_ACTIVATE_TLS}
-
+  --set django.ingress.activateTLS=${DJANGO_INGRESS_ACTIVATE_TLS} \
+  --set createSecret=true \
+  --set createRabbitMqSecret=true \
+  --set createRedisSecret=true \
+  --set createMysqlSecret=true \
+  --set createPostgresqlSecret=true
 ```
 
 Helm >= v3:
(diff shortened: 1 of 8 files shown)


KUBERNETES.md
```diff
@@ -236,7 +250,12 @@ helm install \
   --set database=postgresql \
   --set postgresql.enabled=true \
   --set postgresql.replication.enabled=true \
-  --set postgresql.replication.slaveReplicas=3
+  --set postgresql.replication.slaveReplicas=3 \
+  --set createSecret=true \
+  --set createRabbitMqSecret=true \
+  --set createRedisSecret=true \
+  --set createMysqlSecret=true \
```
alles-klar: We don't need the mysql secret of we use postgresql. Same for Redis.

ptrovatelli: yes I know but it seemed more simple to just give the command that always works.

ptrovatelli: I have added the info in the doc. hopefully that will make travis happy too, not sure what happened with travis

ptrovatelli:
force-pushed travis fix (needs the createsecret* flags)

ptrovatelli:
it was working fine when upgrading using the same image, but when changing image i've found an issue with the initializer job (that needs being removed) and now with the postgresql password for some containers (celery worker, initializer). 

working on it...

travis/script.sh
```diff
@@ -72,20 +72,22 @@ if [ -z "${TEST}" ]; then
...
-	  "
-	  ;;
+    rabbitmq)
+      HELM_BROKER_SETTINGS=" \
+        --set redis.enabled=false \
+        --set rabbitmq.enabled=true \
+        --set celery.broker=rabbitmq \
+        --set createRabbitMqSecret=true \
+    "
+    ;;
```
alles-klar: Maybe missing an indent here? The travis jobs looks different than before.

ptrovatelli: I have tried homogenizing the indents: was a mix of tabs and blanks, I have used only 2 blanks for indentation. bad idea? what looks different exactly?

alles-klar: I provided a PR, but it seems it has nothing todo with travis...

alles-klar:
flake8 fails, but this is already fixed here: https://github.com/DefectDojo/django-DefectDojo/pull/1992 and synk should be allowed to fail. So for me it is ok now. You may merge my PR https://github.com/ptrovatelli/django-DefectDojo/pull/3 before merging to dev.

ptrovatelli:
@DefectDojo/defectdojo-moderators we're gonna need another reviewer for helm/kubernetes :)

Issue #1926, opened 2020-02-25
Mitigated finding upon re-import keeps on commenting


madchap:
**Bug description**
When a finding is found to be mitigated through a scan re-import (e.g. fixed upstream), DD adds a comment and changes the finding's status to Inactive, Mitigated.

However, on further re-import iterations, it keeps on adding comments. So we can end up having multiple useless comments as re-imports occur.

![image](https://user-images.githubusercontent.com/5468769/75250466-b6d22800-57d8-11ea-9dee-0a3ea58d800d.png)

**Steps to reproduce**
1. Import a scan
2. Re-import over it, having removed findings from the report. This should tell DD the finding is mitigated, as not found anymore in the uploaded report.
3. Notice the finding status change and comment
4. Re-import the same report as in step 2.
5. Notice a comment is again being added.

**Expected behavior**

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [x] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
Latest `dev` as of today.

valentijnscholten:
yes I have seen this also and fixed in my branch: https://github.com/valentijnscholten/django-DefectDojo/tree/reupload_mitigate_only_once will make a PR soon.

madchap:
This was fixed by the related PR.

PR #1929, opened 2020-02-27
Cicd improvements: reupload mitigation, overdue

valentijnscholten:
Some CI/CD related improvements I implemented during due to due usage:

- (fix) #1926 Mitigate only once on reuploads
- show overdue also in list of active/all engagements
- ci/cd:
  - set target_end to to the date of the latest scan import
  - use 10 day grace period before ci/cd engagement becomes overdue

The idea is that for CI/CD engagement a target_end usually doesn't make sense. By updating the target_end whenever a scan is uploaded it becomes nicely apparent which engagements are scanned regularly. We have a policy that every project(or branch) should be scanned at least weakly, so a 10 day grace period fits nicely with that.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -7,7 +7,7 @@ from dojo.models import Product, Engagement, Test, Finding, \
 from dojo.forms import ImportScanForm, SEVERITY_CHOICES
 from dojo.tools import requires_file
 from dojo.tools.factory import import_parser_factory
-from dojo.utils import create_notification
+from dojo.utils import create_notification, max_safe
 from django.urls import reverse
 from tagging.models import Tag
 from django.core.validators import URLValidator, validate_ipv46_address
@@ -536,6 +536,7 @@ class ScanSerializer(serializers.ModelSerializer):
 
 class ImportScanSerializer(TaggitSerializer, serializers.Serializer):
     scan_date = serializers.DateField(default=datetime.date.today)
+
     minimum_severity = serializers.ChoiceField(
         choices=SEVERITY_CHOICES,
         default='Info')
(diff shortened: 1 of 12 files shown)


dojo/api_v2/views.py
```diff
@@ -25,6 +25,7 @@
 from dojo.api_v2 import serializers, permissions
 
 
+
```
madchap: I can approve that! :-D

madchap: Actually, I am surprised that flake8 is ok with 3 lines.

valentijnscholten: I think I even added it because flake in my visual studio code complained about it.

ycrumeyrolle:
> We have a policy that every project(or branch) should be scanned at least weakly, so a 10 day grace period fits nicely with that.

Can the 10 days grace period become a system settings ?

valentijnscholten:
> > We have a policy that every project(or branch) should be scanned at least weakly, so a 10 day grace period fits nicely with that.
> 
> Can the 10 days grace period become a system settings ?

Could be yes. Is it a blocker for you that it's not right now? I didn't implement it yet because I first wanted to see if it worked for me and others.

ycrumeyrolle:
No it is not a blocker. 
For the moment we are re-uploading the scan files only once a month, so I am wondering the impact of such enhancement.

valentijnscholten:
build errors should be fixed by #1937

valentijnscholten:
Noticed that `test/views.py` contains a similar piece of logic as `serializers.py`, so pushed a commit to apply most of the changes also there. Except for the CICD changes, or should that also happen when using the UI?

madchap:
@valentijnscholten it seems you have unresolved conflicts here.

madchap:
More conflicts :-/

PR #1933, opened 2020-02-27
Fixed typo 'crital'

doomguy:
Fixed typo 'crital'


madchap:
Hi @doomguy 

We do not change already existing migration files to avoid breaking existing installations.

Please submit a PR with a new migration file instead that would fix the issue.

Thank you!

Issue #1931, opened 2020-02-27
Too many levels of symbolic links


arkwrn:
**Bug description**
Cannot build nginx image based on guideline in https://github.com/DefectDojo/django-DefectDojo#quick-start and always showing error `Too many levels of symbolic links`

**Steps to reproduce**

1. - git clone https://github.com/DefectDojo/django-DefectDojo
2. - cd django-DefectDojo
3. - docker-compose build

**Expected behavior**
Build success.

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [X] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: CentOS Linux release 7.7.1908 (Core) and also happen in macOS Catalina 10.15.3
 - DefectDojo Commit Message: d411791: Merge pull request #1834 from ptrovatelli/v1.5.4rc6 [2020-02-24 16:13:31 -0600]

**Console logs** (optional)
```
 ---> Running in b9f7d039d07a
yarn install v1.22.0
warning package.json: No license field
info No lockfile found.
warning No license field
[1/5] Validating package.json...
warning No license field
[2/5] Resolving packages...
warning @yarn_components/mocha > jade@0.26.3: Jade has been renamed to pug, please install the latest version of pug instead of jade
warning @yarn_components/mocha > glob > minimatch@0.2.14: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
warning @yarn_components/mocha > glob > graceful-fs@2.0.3: please upgrade to graceful-fs 4 for compatibility with current and future versions of Node.js
[3/5] Fetching packages...
[4/5] Linking dependencies...
warning " > @yarn_components/metisMenu@3.0.5" has unmet peer dependency "jquery@^3.3.1".
[5/5] Building fresh packages...
success Saved lockfile.
Done in 73.17s.
yarn cache v1.22.0
warning package.json: No license field
success Cleared cache.
Done in 0.93s.
Traceback (most recent call last):
  File "manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.5/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.5/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
[...]

PR #1937, opened 2020-02-28
Upgrade NodeJS and Fixing issue #1931

arkwrn:
- Upgarde NodeJS version :
Upgrade from version 11.x to 12.x due to decprecation warning, since nodejs 11.x is no longer actively supported

```
================================================================================
================================================================================

                              DEPRECATION WARNING                            

  Node.js 11.x is no longer actively supported!

  You will not receive security or critical stability updates for this version.

  You should migrate to a supported version of Node.js as soon as possible.
  Use the installation script that corresponds to the version of Node.js you
  wish to install. e.g.

   * https://deb.nodesource.com/setup_10.x — Node.js 10 LTS "Dubnium" (recommended)
   * https://deb.nodesource.com/setup_12.x — Node.js 12 LTS "Erbium"

  Please see https://github.com/nodejs/Release for details about which
  version may be appropriate for you.

  The NodeSource Node.js distributions repository contains
  information both about supported versions of Node.js and supported Linux
  distributions. To learn more about usage, see the repository:
    https://github.com/nodesource/distributions

================================================================================
================================================================================
```

- Fixing issue when building nginx image (https://github.com/DefectDojo/django-DefectDojo/issues/1931)

### DIFF
--- a/Dockerfile.nginx
+++ b/Dockerfile.nginx
@@ -23,16 +23,22 @@ RUN pip3 wheel --wheel-dir=/tmp/wheels -r ./requirements.txt
 FROM build AS collectstatic
 
 USER root
+ENV \
+  # This will point yarn to whatever version of node you decide to use
+  # due to the use of nodejs instead of node name in some distros
+  node="nodejs"
 RUN \
   apt-get -y update && \
   apt-get -y install apt-transport-https ca-certificates && \
   curl -sSL https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - && \
-  echo "deb https://deb.nodesource.com/node_11.x stretch main" | tee /etc/apt/sources.list.d/nodesource.list && \
+  curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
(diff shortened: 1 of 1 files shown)


arkwrn:
> thanks that fixes it indeed!
> 
> a few thoughts though:
> 
> * yarn: why not stick to stable? wasn't working?
> * i think we should pin the versions: install a specified version of node and yarn instead of the latest one. that will hopefully prevent this kind of erratic behavior in the future. we have a dependency bot that can (hopefully) warn us when a new version is available for us to upgrade while keeping control on it.

I need a quick fix that's why I try to do a tracing problem and find the fastest way to do that is to use the nightly version, I totally agree with you to freeze the version, but I'm sorry at this time I still don't have much time to test the version which yarn is good to use, hopefully many can help to do the trial 😄

madchap:
We'll have to revisit this when yarn 1.23 makes it into stable.
But for now, it is needed to build the nginx so.. merging.

arkwrn:
Hi @madchap , @ptrovatelli I have tested for a yarn package manager stable version that can be used and I think it is enough to use, the version that I confirmed is the stable version 1.21.0.

PR #1938, opened 2020-02-28
add git tips

ptrovatelli:
a few tips on using git. that will give us nicer git history :)


madchap:
I believe a little guidance on git can be useful here.

However, I am wondering if this is actually a good idea to have "so much" about git in there. Maybe we should only have a couple one-liners about fetch dev and rebase and not try to replicate possible other thorough git guidance/howtos?

ptrovatelli:
I don't think it's useful to add too basic information about git because this is already known by most and easy to find out

the other stuff isn't that easy to find, although when put together it's not all that complicated

ptrovatelli:
i'd be glad to just point to an article that explains just that if it exists  but i've learnt that after reading a pretty long article.

PR #1939, opened 2020-02-28
Performance prefetching phase1 migrationfix

valentijnscholten:
Migration needs an update because another migration was merged just before #1899 was merged.
It also seems that 'active=True' reappeared for active finding count.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -584,6 +584,7 @@ class Product(models.Model):
         except AttributeError:
             # ideally it's always prefetched and we can remove this code in the future
             self.active_finding_count = Finding.objects.filter(mitigated__isnull=True,
+                                            active=True,
                                             false_p=False,
                                             duplicate=False,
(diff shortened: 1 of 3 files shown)


madchap:
Potentially a quick approve and merge @ptrovatelli ? Cheers.

dojo/models.py
```diff
@@ -585,7 +585,6 @@ def findings_count(self):
             # ideally it's always prefetched and we can remove this code in the future
             self.active_finding_count = Finding.objects.filter(mitigated__isnull=True,
                                             active=True,
-                                            verified=True,
```
ptrovatelli: is this wanted? why removing this filter?

madchap: an active (or open) finding does not have to be verified.

valentijnscholten: as per #1922

ptrovatelli: ok

PR #1940, opened 2020-02-28
dependabot: get prs only once a week

valentijnscholten:
dependabot: add default label and get prs once a week

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -4,16 +4,22 @@ update_configs:
     directory: "/"
     target_branch: "dev"
     update_schedule: "weekly"
+    default_labels:
+    - dependencies
+    - python    
 
   - package_manager: "javascript"
-    directory: "/"
+    directory: "/components"
     target_branch: "dev"
     update_schedule: "weekly"
+    default_labels:
+    - dependencies
+    - javascript    
 
   - package_manager: "docker"
     directory: "/"
     target_branch: "dev"
     update_schedule: "weekly"
-
     default_labels:
-      - "dependencies"
+    - dependencies
+    - docker
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -3,14 +3,17 @@ update_configs:
   - package_manager: "python"
     directory: "/"
     target_branch: "dev"
-    update_schedule: "live"
+    update_schedule: "weekly"
 
   - package_manager: "javascript"
     directory: "/"
     target_branch: "dev"
-    update_schedule: "live"
+    update_schedule: "weekly"
 
   - package_manager: "docker"
     directory: "/"
     target_branch: "dev"
     update_schedule: "weekly"
+
+    default_labels:
+      - "dependencies"
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -4,22 +4,13 @@ update_configs:
     directory: "/"
     target_branch: "dev"
     update_schedule: "weekly"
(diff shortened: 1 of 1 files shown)


madchap:
I don't think this will change much? Label is already applied by default btw.

valentijnscholten:
Well, I was about to set on a quest to configure the hell out of dependabot. But I found out there's not much to configure in this config file :-). Still thought these changes would be good. Wouldn't want to have new PRs on daily basis. Once a week is enough to worry about those.

One thing that's difficult currently is how to test the PRs, especially if you want to test them together. Not sure if it's possible, but maybe we could create a dependy update branch or something where they keep getting updated. Every once in a while we could test that branch en merge into dev if succesful.

madchap:
Should probably be closed in favor of #1943 ?

valentijnscholten:
I would prefer weekly schedule instead of live

madchap:
I don't have a strong opinion. Not sure if it'd help to switch to weekly.

Also, if in the meantime there is a newer version detected for a certain lib, dependabot is smart enough to detect there is a open PR for that lib, and updates that that PR instead of opening a new one.

Any other mods have a strong opinion here? I think @alles-klar was actually all about `live` updates. Not sure about the others.

valentijnscholten:
The weekly thing is more that we don't get new PRs (for dependencies without an existing PR) every day. They pollute the default view when you go to the PR tab. But as I said it's more important to find a view that it doesn't go updating dependencies from from 1.x to 9999.x

.dependabot/config.yml
```diff
@@ -3,14 +3,17 @@ update_configs:
   - package_manager: "python"
     directory: "/"
     target_branch: "dev"
-    update_schedule: "live"
+    update_schedule: "weekly"
 
   - package_manager: "javascript"
     directory: "/"
```
madchap: Can you please put `/components` here, as to match what is now in `master`?

.dependabot/config.yml
```diff
@@ -3,14 +3,17 @@ update_configs:
...
-    update_schedule: "live"
+    update_schedule: "weekly"
 
   - package_manager: "docker"
     directory: "/"
     target_branch: "dev"
     update_schedule: "weekly"
+
+    default_labels:
+      - "dependencies"
```
madchap: This may actually have a negative effect (not tested), as the specific labels such as `python`, `javascript`, `docker` may go away if this overwrites dependabot default behavior.

valentijnscholten:
Fixed both. But now I am just copying some parts from the dashboard to the config here. Might be better to copy all, or just forget about the yml file and only use the dashboard.

valentijnscholten:
On second thought, copying everything into the file makes no sense. The @ignore this @ignore that comments in the PRs nicely integrate with the dashboard (I presume). So maybe just set it to weekly in the file manage the rest in the dash?

madchap:
There was nothing to do on the dashboard. It is all out of the box.

I'd suggest we just move it to `weekly` if you like, and leave the rest as it was.

madchap:
If you want this effective ASAP, you'd also have to put a PR up against `master`, as the default active branch is the one that triggers the thing.

valentijnscholten:
If #2002 brings us in a stable/mostly up-to-date situation where dependabot doesn't create lots of PRs we might keep it at daily after all.

valentijnscholten:
Dependabot has quieted down now, so let's leave it as is - closing.

PR #1943, opened 2020-02-28
Remove root package.json

madchap:
Please squash this when merging, whoever you will be. Messed up some stuff.

Removing root's package.json file and adapting dependabot config to take the one in components directory. Dockerfiles use that one, and I could `docker-compose build` without issue without the root's file.


valentijnscholten:
Just type `reset upstream/master` and commit :-) (Assuming you have the dd repo added as remote called upstream)

madchap:
> Just type `reset upstream/master` and commit :-) (Assuming you have the dd repo added as remote called upstream)

Just learned something new ;-) I was always trying to fiddle with rebase, etc... I guess my commit message could have been a bit more to the point of this PR then!

PR #1953, opened 2020-02-29
performance: cache system_settings in views

valentijnscholten:
The codebase is using lots of `System_Settings.objects.get()` calls to read the system settings. This results in lots of database queries. Sometimes over a 100 queries just to render one page.

Alternatives considered:

1. Rewrite all occurrences into something like `get_system_settings()` to abstract away from the implementation of system settings as a model
2. Use threadlocals in some way
3. Use a full blown Redis ORM cache or similar

Although option 3 provides many benefits, it's also an extra step/hurdle/maintenance burden. Also having 100s of queries for a single will page will never lead to optimal results.

So I opted for option 4:

4. Sneaky caching of System_Settings model
Basically it's a fully supported features of Django ORM to use custom managers for your model. For example to all kinds of things with optimizing queries or whatever. Most ORM caching frameworks such as `cachalot` use these as well. Here we just use it to store the `System_Settings` model instance in the default django cache backend. And we clear the cache when changes are saved to `System_Settings`. The default cache backend is local within a process. So if you have multiple (uwsgi) processes, this will lead to stale data in the cache. Hence a 30s timeout. 
Each request will now lead to at most 1 database queries to retrieve `System_Settings`.

You can test this for example by enabling/disabling JIRA in system settings. It will show/hide the two extra JIRA columns on the open findings page (or any page showing a list of findings).

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -92,8 +92,10 @@ class Regulation(models.Model):
 class SystemSettingsManager(models.Manager):
     CACHE_KEY = 'defect_dojo_cache.system_settings'
 
-    def get(self, *args, **kwargs):
+    def get(self, no_cache=False, *args, **kwargs):
         # cache only 30s because django default cache backend is local per process
+        if no_cache:
+            return super(SystemSettingsManager, self).get(*args, **kwargs)
         return cache.get_or_set(self.CACHE_KEY, lambda: super(SystemSettingsManager, self).get(*args, **kwargs), timeout=30)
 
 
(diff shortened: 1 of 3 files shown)


valentijnscholten:
The test fails because travis running multi process python. Will push a fix later.

valentijnscholten:
I created another PR with prefetching in other places: #1955 
Tried to avoid conflicts, although there may be one because I also added a `context_processor` (thanks for the inspiration).

PR #1954, opened 2020-03-01
Update to fontawesome 5

madchap:
Breaking changes there.


madchap:
@Maffooch I think you have some experience with JS stuff? Any idea why icons and font would not be loading? There is no error in building the nginx image, nor when loading the pages, but things are just not showing. Cheers.

Maffooch:
Sorry It took so long to respond. I don't have a ton of experience with JS, but did take a look at the issue. Console logs say the file at the this url: http://localhost:8080/static/dojo/js/pdfmake.min.js.map
is not there. This file is related to the recent data tables inclusion found in the all findings view. Not sure what the deal is here.

Maffooch:
So it looks like the upgrade is not as simple as once thought. Font Awesome changed a lot of naming conventions and straight up changes the names of some of the icons we were using. I tried changing a few, but still got the empty rectangles. No idea where to go from here.

Link for reference:
https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4

valentijnscholten:
For now we should stay at v4. No reason to go for v5 which is a complete rewrite. Also there are lots of javascript issue so I think we need to stabilize first.

i.e. https://github.com/DefectDojo/django-DefectDojo/issues/1999

madchap:
> Console logs say the file at the this url: http://localhost:8080/static/dojo/js/pdfmake.min.js.map
> is not there. This file is related to the recent data tables inclusion found in the all findings view. Not sure what the deal is here.

I think it is unrelated to the move to font-awesome 5. @valentijnscholten brought this up a short while ago even with our current version.

> For now we should stay at v4. No reason to go for v5 which is a complete rewrite. Also there are lots of javascript issue so I think we need to stabilize first.

Agreed. I'll close this and inform dependabot to not bump major if it can listen.

PR #1970, opened 2020-03-03
Fix import error for Crashtest Security importer for skipped scans

Phylu:
I just noticed another crash during the import of a Crashtest Security Scan file, in the case that the scan contains any skipped test case.

This PR fixes prevents this crash by adding a check for skipped tests.


Phylu:
> Thanks. If you have time, could you upload a sample file with skipped tests to https://github.com/DefectDojo/sample-scan-files and/or add it as a unit test such as we have for most other parsers?

I have added the skipped test to the sample file here: https://github.com/DefectDojo/sample-scan-files/pull/45

PR #1972, opened 2020-03-03
Fix flake8

madchap:
#1967 merged a little too quickly with flake8 issue. This fixes them.

Thanks @alles-klar for the :eyes:  !


alles-klar:
@madchap Do we need to add a label before merging?

PR #2009, opened 2020-03-09
Fix duplication issue

MarianG:
This Pull Request is a subset of the bigger PR (  https://github.com/DefectDojo/django-DefectDojo/pull/1901) from a while ago. Since I mixed several issues (whitesource-parser, deduplication error, and fix for duplicate-loops) in one request, I thought creating different smaller requests might reduce the workload. The second part, which includes the loop-detection and correction is located in #2010.

Description from previous PR:
We had an issue where we encountered circular dependencies for duplicates in findings. We assume it occured because of multi-threading and the asynchronous duplicate-detection. We had issues where Finding A had duplicates B, C, and D. B had the duplicate C, C had the duplicate D and D had B. Then the clean-up job that should delete some duplicates produced a recursion error, because it tries to delete duplicates of a finding before deleting the finding.
I understand that in theory a duplicate should not have a duplicate on its own, but it happened.
We updated the set_duplicate method.
We also found the duplicate_list field is unnecessary, since we have the information in the original_finding field (backwards from duplicate_finding Foreign Key)

On behalf of DB Systel GmbH

- enhancement
- feature
- bugfix
- maintenance (a.k.a chores)
- dependencies


valentijnscholten:
Great idea to split them up. When trying to assess whether the `duplicate_list` relation could indeed be removed, I found two files that are referencing this field and don't seem to be updated yet by this PR:
- dojo/finding/views.py
- dojo/tasks.py

MarianG:
I changed the views and tasks according to your proposal. The tasks.py was only changed in #2010. But thanks for pointing out these references

dojo/models.py
```diff
@@ -1255,7 +1255,6 @@ class Finding(models.Model):
     duplicate_finding = models.ForeignKey('self', editable=False, null=True,
                                           related_name='original_finding',
                                           blank=True, on_delete=models.CASCADE)
-    duplicate_list = models.ManyToManyField("self", editable=False, blank=True)
```
madchap: @ptrovatelli unsure if this was used before at all?

dojo/db_migrations/0032_remove_finding_duplicate_list.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 2.2.9 on 2020-02-14 09:14
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dojo', '0031_finding_component'),
```
madchap: isn't the same as recently added `component_name` ?

dojo/models.py
ptrovatelli: i didn't use it specifically but I think it's displayed in the gui isn't it?

dojo/db_migrations/0032_remove_finding_duplicate_list.py
valentijnscholten: it's depending on that migration yes

dojo/models.py
valentijnscholten: it's displayed yes. but the list can already be retrieved via the duplicate_finding ForeignKey so there's no need for an additional ManyToMany relationship. At least that's the hypothesis of this PR. If 1.6RC is already branched of we work/test with this in `dev` for a bit until the next release. (with some support from the PR author)

MarianG: It was used in some views. 
Also in the models.py the function duplicate_finding_set used this duplicate_list before.
However you dont need it if you use the backwards relation

valentijnscholten: In the old situation you could have a finding being a duplicate of multiple findings (via duplicate_list):

```
A->B--|
|     v
->C---D
```

In the new situation this would always have to be a straight line:

```
A->B->C->D
or
A->C->B->D
```

alles-klar: We should avoid lists. If I am right, in the new situation it looks like this:
A->D
B->D
C->D

All list will be resolved. Not by the model, but during the deduplication.

valentijnscholten: Yeah that's the other part of this PR I think that a duplicate can never be the original of another issue.

valentijnscholten:
@MarianG are you happy to have this merged as-is?

madchap:
@alles-klar maybe you want the honor here?

MarianG:
Yes I think I would be happy to see it included as is

PR #2034, opened 2020-03-10
quickfix #1995 bulk edit in view_test

### LINKED ISSUE #1995: Can no longer bulk edit findings
**Bug description**

The checkboxes remain greyed out, unable to select new status.

image

Latest `dev`

valentijnscholten:
Fixes #1995 and adds an integration test to prevent future regressions

### DIFF
--- a/dojo/templates/dojo/findings_list.html
+++ b/dojo/templates/dojo/findings_list.html
@@ -46,22 +46,24 @@
                                     <option value="High">High</option>
                                     <option value="Critical">Critical</option>
                                 </select><br/><br/>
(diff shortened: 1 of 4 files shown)


valentijnscholten:
renamed javascript ids to make them unique on the page. this only affects javascript code as the python code looks at the `name` of the fields and not the `id`

Maffooch:
The conflicts are strange here. They don't quite match the diffs from this PR. Could you check it out @valentijnscholten?

valentijnscholten:
Should be fixed, also I see that one line was removed by #2015 by mistake, but is now added back in here (which is good)

Issue #2045, opened 2020-03-11
finding images don't work in docker deployment


valentijnscholten:
When running integration tests uploading of findings works fine, but when returning to the view finding page findings cannot be retrieved:

```
GET http://localhost:8080/static/dojo/img/zoom-in.cur 404 (Not Found)
5d242da1aa6aa4fd589e851648b9edf1.jpg:1 GET http://localhost:8080/media/CACHE/images/finding_images/e88bcf2b-5076-4078-b646-b0f2b4de4bbe/5d242da1aa6aa4fd589e851648b9edf1.jp
```
(errors from browser console)

`dev` at 3d3804589049ff538734b47d6147077caaea8998

ClintCrawford:
I solved it this way
https://github.com/DefectDojo/django-DefectDojo/issues/1930

valentijnscholten:
Thanks, I was hoping it was something easy to fix. Will take a look.

valentijnscholten:
The main reason it was not working is because the images cannot be retrieved via nginx. Persistent volumes are indeed needed if you want to keep them working across restarts, but I already ran into the problem right after uploading the image. Will probably have a PR soon, might try to fix risk acceptance in one go.

dayal18:
Any idea on how to fix this in setup.bash?
I have the same issue.

dayal18:
the interesting thing is that the image exists at that location when I click the magnifying symbol, I see my URL changes to:

**http://localhost:8000/media/finding_images/2d53a220-da1a-40c3-8178-f9632a8f9207.png**

I looked at the server for the image, and I find the image at this location:
Command run on the server:

root@ip-10-31-23-147:/home/ubuntu# updatedb
root@ip-10-31-23-147:/home/ubuntu# locate 2d53a220-da1a-40c3-8178-f9632a8f9207.png
**/opt/dojo/django-DefectDojo/media/finding_images/2d53a220-da1a-40c3-8178-f9632a8f9207.png**
root@ip-10-31-23-147:/home/ubuntu#

----

and still I see 404..

dayal18:
And in logs I see:
[09/Apr/2020 23:25:38] "**GET /media/CACHE/images/finding_images/2d53a220-da1a-40c3-8178-f9632a8f9207/eed88358fb41221c1db604dc4422efe4.jpg HTTP/1.1**" 404 22622

Any help please

valentijnscholten:
[...]

PR #2056, opened 2020-03-13
make finding images work in docker dev and ptvsd setup

valentijnscholten:
Partial fix for #2045, makes finding images work in docker development setup.

Making it work in release mode will have to wait until @devGregA has finalized the migration to AppVeyor.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -25,6 +25,9 @@ http {
     location /static {
       alias /usr/share/nginx/html/static;
     }
+    location /media {
+        alias /usr/share/nginx/html/media;
+    }
     location / {
       include /run/uwsgi_pass;
       include /etc/nginx/wsgi_params;
--- a/docker-compose.override.dev.yml
+++ b/docker-compose.override.dev.yml
@@ -5,6 +5,8 @@ services:
     entrypoint: ['/wait-for-it.sh', 'mysql:3306', '-t', '30', '--', '/entrypoint-uwsgi-dev.sh']
     volumes:
       - '.:/app:z'
+      - './media:/app/media/finding_images'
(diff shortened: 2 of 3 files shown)


nginx/nginx.conf
```diff
@@ -25,6 +25,9 @@ http {
     location /static {
       alias /usr/share/nginx/html/static;
     }
+    location /media {
+        alias /usr/share/nginx/html/media;
+    }
```
alles-klar: This nginx.conf is used in many setups. So I don't think it is a good idea to add it here and only provide a solution for docker-compose.

valentijnscholten: I couldn't find any other place it is being used

ptrovatelli: @alles-klar are you thinking of kubernetes? i don't have a strong opinion on this, but if the images are already not working, this is a start... :)

alles-klar: Sorry for the long delay. I think all setups are using the nginx config - also the productive ones. If we add the /media path there, all data that is stored there is free accessible without any authentication. If user add confidential pictures this is a security issue.

valentijnscholten: the production docker-compose setup is a container purely meant for defect dojo. if somebody would  put sensitive data in there I don't think they know what they are doing :-). Also the production (release) config doesn't map any local files to the nginx media folders.  `setup.bash` doesn't use this `nginx.conf`.

alles-klar: Using this setup just for debugging and development is fine but I am afraid that someone adopt this solution for other setups. Maybe we add a comment in the nginx config that it is currently for development only with docker-compose only and that this directory has to be empty in production?

PR #2066, opened 2020-03-15
add watson middleware for more efficient search index updates

valentijnscholten:
During (re-)import findings are updated multiple times, triggering multiple updates of the watson search index (via post_save) signals.

Turns out `django-watson` provides a middleware to help with this. All updates to model instances are grouped / deduplicated and executed at the end of the request.

This saves 3 database queries per (created/updated) finding in an import.


ptrovatelli:
seems awesome; have you been able to measure the gain or see any difference by tracking the sql requests issued?

valentijnscholten:
Yes, using django-debug-toolbar. If only the auditlog had the same feature 😩

PR #2080, opened 2020-03-18
fix removed system settings processor

valentijnscholten:
fix merge artifact from #1955 (this fixes all JIRA stuff not being displayed)


madchap:
Confirmed, and was hit in prod by that issue. Thanks.

PR #2084, opened 2020-03-18
docker initializer: do not give a new non-working passwd if admin present in DB

madchap:
Make sure to not give a new non-working admin passwd if present in DB, and print the command to create a new superuser in case the user cannot remember.

### DIFF
--- a/docker/entrypoint-initializer.sh
+++ b/docker/entrypoint-initializer.sh
@@ -9,8 +9,6 @@ then
 fi
 echo "Initializing."
 
-echo "Admin user: ${DD_ADMIN_USER}"
-
 echo -n "Waiting for database to be reachable "
 until echo "select 1;" | python3 manage.py dbshell > /dev/null
 do
@@ -22,6 +20,7 @@ echo
 python3 manage.py makemigrations dojo
 python3 manage.py migrate
 
+echo "Admin user: ${DD_ADMIN_USER}"
 ADMIN_EXISTS=$(echo "SELECT * from auth_user;" | python manage.py dbshell | grep admin)
 # Abort if the admin user already exists, instead of giving a new fake password that won't work
 if [ ! -z "$ADMIN_EXISTS" ]
--- a/docker/entrypoint-initializer.sh
+++ b/docker/entrypoint-initializer.sh
@@ -11,13 +11,6 @@ echo "Initializing."
 
 echo "Admin user: ${DD_ADMIN_USER}"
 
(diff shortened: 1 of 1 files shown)


docker/entrypoint-initializer.sh
```diff
@@ -29,7 +20,23 @@ echo
 python3 manage.py makemigrations dojo
 python3 manage.py migrate
 
+echo "Admin user: ${DD_ADMIN_USER}"
 ADMIN_EXISTS=$(echo "SELECT * from auth_user;" | python manage.py dbshell | grep admin)
```
alles-klar: "grep admin" -> shouldn't we better grep for DD_ADMIN_USER?

docker/entrypoint-initializer.sh
```diff
@@ -29,7 +20,23 @@ echo
...
+    echo "If you don't remember the password, you can create a new superuser with:"
+    echo "$ docker-compose exec uswgi /bin/bash -c 'python manage.py createsuperuser'"
+    exit
+fi
+
+if [ -z "${DD_ADMIN_PASSWORD}" ]
+then
+  export DD_ADMIN_PASSWORD="$(cat /dev/urandom | LC_ALL=C tr -dc a-zA-Z0-9 | \
+    head -c 22)"
+  echo "Admin password: ${DD_ADMIN_PASSWORD}"
```
alles-klar: If a admin exists, the admin password should not be generated or printed out.

alles-klar: oh I missed the exit command

docker/entrypoint-initializer.sh
madchap: Yes. Will update.

docker/entrypoint-initializer.sh
```diff
@@ -29,7 +20,23 @@ echo
 python3 manage.py makemigrations dojo
 python3 manage.py migrate
 
+echo "Admin user: ${DD_ADMIN_USER}"
 ADMIN_EXISTS=$(echo "SELECT * from auth_user;" | python manage.py dbshell | grep admin)
+# Abort if the admin user already exists, instead of giving a new fake password that won't work
+if [ ! -z "$ADMIN_EXISTS" ]
+then
+    echo "Initialization detected that the admin user ${DD_ADMIN_USER} already exists in your database."
```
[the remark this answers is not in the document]
madchap: I'd probably would change the doc instead. Will push something.

PR #2110, opened 2020-03-25
Checkmarx: set verified to True for urgent/confirmed findings

dshraddha23:
Checkmarx has following statuses for findings:
0: "To Verify"
1: "Not Exploitable"
2: "Confirmed"
3: "Urgent"
4: "Proposed Not Exploitable"
The following changes will add verified=true only for Confirmed and Urgent (Actionable items)

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -21,6 +21,7 @@ import six
 from django.utils.translation import ugettext_lazy as _
 import json
 
+
 class TagList(list):
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


ptrovatelli:
Hello,
Could you check if this can be done in the checkmarx scanner instead? if possible it would be cleaner because this is adding some parser specific logic in a generic class. 

also this code will apply on API v2 only, not on GUI import I think, does it?

madchap:
We should indeed keep the class generic. I don't believe this belongs here. Closing.

PR #2114, opened 2020-03-27
Dockerfile for integration tests

alles-klar:
A Dockerfile to simply run the integration tests.

On behalf of DB Systel GmbH


valentijnscholten:
There are now 3 different names used in this PR: `integrationtest`, `integration_tests` and `integration-tests`. Might be small change to align them to being the same?

madchap:
(I like underscores best :P)

alles-klar:
> There are now 3 different names used in this PR: `integrationtest`, `integration_tests` and `integration-tests`. Might be small change to align them to being the same?

I just oriented at the unit tests and named them in the same way. Not sure what is the best way to go because different technologies use different names and have different requirements. We have the same Problem for "False" in Django and "false" in helm/kubernetes/bash.

Maybe we better discuss it in a separate Issue/PR/Meeting?

valentijnscholten:
OK let's leave it then, it's not super imporant. Don't want to change existing things too much now.

PR #2115, opened 2020-03-27
Add TLS for Nginx Helm Chart

alles-klar:
* fixing nginx tls config (missing "/")
* add tls config for nginx in helm chart
* standardize boolean variables for all bash and helm variables

On behalf of DB Systel GmbH


ptrovatelli:
Nice feature. Could you drop a few words in the kubernetes.md to explain the difference between this (end to end encryption to the nginx pod) and the the pre-existing way to configure TLS at ingress level (which was fine too I guess provided the cluster itself is sufficiently hardened and has things like mutual tls between nodes)? 
Also could you document the command you would use to get your own certificate at the required location during `helm install` ?

alles-klar:
@ptrovatelli I added some documentation for the new part. But there is still room for improvements. Maybe in some later PRs :)

alles-klar:
> as commented, could you add a bit of documentation?
> thanks!

@ptrovatelli it semms I cant merge the PR in github until you approve that your requested change is fulfilled. Please take a look and merge if possible.

ptrovatelli:
OK OK go... please let all remember that self-signed certificate by itself is only a first step because it's only good against network sniffing. Protection against man in the middle requires proper CA validation.

madchap:
Alright.. merging...

PR #2127, opened 2020-03-30
Allow staff users to delete notes

madchap:

This PR grants superuser users the ability to delete notes from the API, and also make the `patch` endpoint work to remove a note from a finding.

When wanting to remove a note from a finding with the same user from the UI or from the APIv2, the APIv2 forbid me to delete if not the same user, something we don't have from the UI.

Also, the `patch` or `put` methods were actually forbidden (error 405), even though specified on the `remove_note` endpoint, it had to be specified also on the `notes` to work.

Since both seemed to achieve the same thing, I removed the `put` method.

- enhancement
- feature
- bugfix
- maintenance (a.k.a chores)
- dependencies

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -912,7 +912,7 @@ class AddNewNoteOptionSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Notes
-        fields = ['entry', 'private', 'note_type']
+        fields = ['entry', 'private']
 
 
 class FindingToFindingImagesSerializer(serializers.Serializer):
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -912,7 +912,7 @@ class AddNewNoteOptionSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Notes
-        fields = ['entry', 'private']
+        fields = ['entry', 'private', 'note_type']
 
 
 class FindingToFindingImagesSerializer(serializers.Serializer):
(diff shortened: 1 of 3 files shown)


valentijnscholten:
what about staff users? or is that the same? I always get confused because I think what django calls superuser is called staff user in defect dojo?

madchap:
> what about staff users? or is that the same? I always get confused because I think what django calls superuser is called staff user in defect dojo?

Superusers and staff are different permission sets, even though maybe today the difference is thin. My reference is what you look at a user, you have `active`, `staff` and `superuser`, which supposedly has more rights.

valentijnscholten:
Should staff users also be allowed to remove notes then?

madchap:
> Should staff users also be allowed to remove notes then?

I thought not, but somehow was totally arbitrary (e.g. like an "incremental" change, since everyone was denied before).

valentijnscholten:
Well, if you want it to be the same as the UI.... :-D

madchap:
Yea. What to say about this model. I can definitely change it.

madchap:
> I would also have approved if both superuser and staff were there :-D

Dude.

madchap:
More seriously, can an admin not be staff?

valentijnscholten:
I have no idea. I studied computer science for 5 years but I think a PhD is needed to understand superuser vs staff vs admin ;-). So my initial thought was to allow them both to have access, sorry if I was not clear.

madchap:
> I have no idea. I studied computer science for 5 years but I think a PhD is needed to understand superuser vs staff vs admin ;-). So my initial thought was to allow them both to have access, sorry if I was not clear.

No problem ;-)

Actually I think in our state-of-the-art case of permission modeling, being an admin without being a staff is probably something we would not see much -- even though technically possible.

PR #2131, opened 2020-03-31
Move similar finding below actual finding main info

madchap:

When viewing findings which have "similars", the similar takes a big amount of place.
And if the finding has "extra" info, such as components, or I imagine endpoint info and such, you have to scroll down or minimize the "similar finding" block, which is aggravating.

This moves the "similar findings" block below these.

image


madchap:
accesslint screams, but nothing changed.

Issue #1973, opened 2020-03-03
Push to JIRA option when Importing findings (UI and API)


Apipia:
**Is your feature request related to a problem? Please describe**
Would be nice to be able to import findings and check "Push these new findings to Jira" so one doesn't have to take an extra step to bulk edit findings and push them to Jira.

**Describe the solution you'd like**
A Push To Jira checkbox on the import/re-import findings UI, one that, when checked, will push all resulting Findings to whatever JIRA configuration (if one exists) for the product. Adding this functionality to the import/re-import API endpoints as well.

**Describe alternatives you've considered**
There is also the option of creating API endpoints for JIRA related actions as well. Possibly updating the PUT for Findings to include a PushToJira boolean value that will allow an API user to send a request to push the issue to JIRA. 
Or a separate endpoint all together that takes a list of Finding ID's and performs similar logic to the bulk edit that will push all findings to JIRA.
OR a various /jira/ endpoints that can exist on Products, engagements, tests, that will push all of those findings within each to JIRA. 
PUT jira/engagement/{engagement_id}
PUT jira/product/{product_id}
With empty bodies? 
Looking for feedback here.

**Additional context**
I wouldn't mind implementing these (or attempting to implement them) but I'm looking for some community feedback on the best approach.

valentijnscholten:
There is a checkbox on Product settings called 'Push all issues' which is supposed to do this, but it doesn't work. As far as I can see it's only partially implemented and some parts are commented out. Probably because somebody stopped halfway implementing it.
I can definately see a usecase for this, especially for companies which are based completely of JIRA for developer workloads (such as mine).

Apipia:
> There is a checkbox on Product settings called 'Push all issues' which is supposed to do this, but it doesn't work. As far as I can see it's only partially implemented and some parts are commented out. Probably because somebody stopped halfway implementing it.
> I can definately see a usecase for this, especially for companies which are based completely of JIRA for developer workloads (such as mine).

Yeah I've noticed this as well. Do you think it's supposed to enlist it in some celery-based task that constantly keeps the issues synced..? Or possibly any time a finding is edited in a product where that's been checked, it calls the push-to-jira action? That might not be a bad method, I could just try to see if that's an implementation I could finish.

valentijnscholten:
I think from what I saw it was only the initial push to JIRA that was (partly) implemented.
[...]

PR #2140, opened 2020-04-02
Feature/jira overhaul (Push All Issues)

Apipia:
# Jira Overhaul (First pass)
The changes here can be summarized into three bullets:

- Product level Push All Issues setting now does something (not sure it did before). And that something is to maintain parity of findings and JIRA issues automatically.
- Added **Push to JIRA** option to importing and re-importing scans. 
- Updated the API to contain the same logic/options.

## Push All Issues
The logic is pretty simple: If this is checked, anytime and issue that belongs to the product where this is enabled, it will push to JIRA while updating/saving this finding. The product **push all issues** help text has been updated:
image

In order to make this obvious to the user when editing/adding a finding, the checkbox for push to jira will look like this:
image
It is checked and disabled, informing the user that this will happen since the product has push all issues enabled. This was added in all of the locations I could find where one can edit a finding. Locations are:

- [x] Editing a finding
- [x] Promoting a finding
- [x] adding a finding
- [x] adding and ad-hoc finding
- [x] Bulk updating Findings (I couldn't figure out how to grey-out and check the box but it'll still push them to JIRA if you have push all issues enabled)
- [x] Importing and re-importing (Didn't use to exist here at all before)

## Import & Re-import
Form now contains push to jira checkbox
image
The rest of the logic I pretty much already covered, just wanted to show the updated form.

## API Changes
Added a push_to_jira option.
image
Endpoints changed:

- [x] Editing a finding (PUT)
- [x] Creating a new finding (POST)
- [x] Importing and re-importing
- [ ] Stub findings. (PUT/POST)

## Engagements
I noticed that there's a push to JIRA checkbox in engagements. I wasn't sure what this meant and after looking at the code it appears it's for engagement-epic mapping. I decided to make this a little more clear:
image
This will never be checked and greyed out like the others, since it technically does something entirely different.

### This is ready for review
Still need to add it stub findings and I'm still testing all of these to make sure they work appropriately. I'll need to test for a product that has push_all_issues enabled, one that doesn't, and one that doesn't have any JIRA settings enabled. 

Let me know if there are any concerns or required changes. 🤓
- Anthony

Reference: #1973


Apipia:
Can someone help me with this? I can't reproduce this db migration error that travis is getting:
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0032_push_all_issues_help_text, 0033_remove_finding_duplicate_list in dojo).

madchap:
> Can someone help me with this? I can't reproduce this db migration error that travis is getting:
> CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0032_push_all_issues_help_text, 0033_remove_finding_duplicate_list in dojo).

I had this just a couple days ago I think, my local `dev` branch was not exactly up-to-date, likely due to recently merged PRs.

I will close and re-open the PR, to re-trigger Travis. This thing is impossible these days... :-(

valentijnscholten:
@Apipia can you rebase and/or update your migration inheritance?

madchap:
So, just updated my local code with this latest PR. If I configure a product with some JIRA setup and save, it does not save the `Project key` nor the `JIRA configuration` for some reasons. It does save the checkboxes state (e.g. `Push all changes`).

Apipia:
@madchap @valentijnscholten 
I've discovered the issue but I'm having trouble resolving it. 
It seems that the new Github form has the same names (project_key, conf) as the JIRA form items, so the the values get clobbered in the POST request. I've tried changing the names in the GITHUB_PKey model and create and make the migration, but it only seemed to work partially.

mtesauro:
FWIW, this looks good to me.  Didn't just add myself as an approver as @valentijnscholten has way more background on this than me.

valentijnscholten:
I approved it, so it can go into `dev` where we iron out the last issues (if any).

PR #2147, opened 2020-04-04
Email notification improvements

valentijnscholten:
Based on https://github.com/DefectDojo/django-DefectDojo/pull/1929

Some work I'm doing on email notifications:
- add details about new/closed/reactivated findings
- also send results_added when (re)importing via api v2
- also send test created when importing via api v2


valentijnscholten:
this was supposed to be a draft looks like i can't make it a draft after i created it as "real"., closing...

PR #2157, opened 2020-04-06
Fix test notes not appearing in report + cosmetic improvement

Apipia:
Just a quick little typo fix.

Issue:
When generating an HTML report for a Test, test notes were not appearing.

Cause:
Typo - was looking at find.notes rather than test.notes


valentijnscholten:
Looks OK, maybe rename the PR to reflect that it's also improving the notes?

PR #2160, opened 2020-04-07
integration tests fixes and improvements

valentijnscholten:
Integration tests were failing randomly in Travis. Turns our this was not Travis' fault, but our own code.

This PR fixes two main problems with Defect Dojo / Integration tests.

- Dedupe tests failing. This was caused by a corner cases being affected by #1953. The cache there is local per worker, so if you have multiple workers, you can have stale system settings for at most 30s. This is usualy OK, but in integration tests it's an issue. This PR contains a workaround, proper fix to follow based on #2164

- `StaleElementReferenceException` by Selenium. This was caused by #1586. After the page is loaded, the datatables javascript applies itself to the DOM and changes it. It wraps itself around the original products table causing the element that selenium just found to go stale. This is a bit of a race condition, explaining why it sometimes works. Fixed by waiting for the datatables element being present in the DOM.

Improvements:

- Check if an import was successful. This makes sure the test fails when the import was not succesful. And also it makes sure the test waits until the import has finished before importing the next scan!

- Increase wait time for background dedupe. Could be that it's just very slow.

- Delete existing QA Product that could be left behind by previous failing tests. Shouldn't happen inside travis, but you never know and it also helps when running local integration tests on an existing database.

- Display logs if integration tests fail.

- Remove some old commented code

- Add some comments here and there to explain tests

Reverted changes from earlier version of this PR:

-  Make dedupe synchronous so it doesn't rely on celery in the background
might be useful to actually let celery perform dedupe.

### DIFF
--- a/tests/dedupe_unit_test.py
+++ b/tests/dedupe_unit_test.py
@@ -6,7 +6,7 @@ import unittest
 import re
 import sys
 import os
-from base_test_class import BaseTestCase
+from base_test_class import BaseTestCase, on_exception_html_source_logger
 from Product_unit_test import ProductTest
 import time
 
@@ -24,7 +24,7 @@ class DedupeTest(BaseTestCase):
     def check_nb_duplicates(self, expected_number_of_duplicates):
         print("checking duplicates...")
         retries = 0
-        for i in range(0, 3):
+        for i in range(0, 18):
             time.sleep(5)  # wait bit for celery dedupe task which can be slow on travis
             driver = self.login_page()
             driver.get(self.base_url + "finding")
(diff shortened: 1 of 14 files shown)


tests/dedupe_unit_test.py
```diff
@@ -24,8 +24,8 @@ def setUp(self):
     def check_nb_duplicates(self, expected_number_of_duplicates):
         print("checking duplicates...")
         retries = 0
-        for i in range(0, 3):
-            time.sleep(5)  # wait bit for celery dedupe task which can be slow on travis
+        for i in range(0, 6):
+            time.sleep(15)  # wait bit for celery dedupe task which can be slow on travis
```
alles-klar, line 28: Maybe it is better to add the sleep at the end of the loop, so we have to wait only if celery is to slow? Furthermore I think 10s should be enough and if not, its better to increase the counter.

valentijnscholten: I don't think it really matters, but I have changed it so now you can't delay approval ;-)

tests/dedupe_unit_test.py
```diff
@@ -24,7 +24,7 @@ def setUp(self):
     def check_nb_duplicates(self, expected_number_of_duplicates):
         print("checking duplicates...")
         retries = 0
-        for i in range(0, 3):
+        for i in range(0, 18:
```
madchap, line 27: Missing `)`?

valentijnscholten: crap, used the web ui. I knew I had better just left it as is.

valentijnscholten:
Looks like dedupe just fails sometimes, even when it is synchronous. So might be an actual bug in the code.

valentijnscholten:
Added some logging and I can see this being printed:
`uwsgi_1_f59058437a3a | skipping dedupe because it's disabled in system settings`

travis/integration_test-script.sh
```diff
@@ -34,109 +29,132 @@ echo "export DD_ADMIN_USER=admin" >> ~/.profile && \
...
 export DD_BASE_URL='http://localhost:8080/'
 
-echo "Running Product type integration tests"
+function fail() {
+    echo "Error: $1 test failed"
+    docker-compose logs --tail="all" uwsgi
+    exit 1    
+}
+
+function succes() {
```
madchap, line 38: It's be nice to correct the typo here :) `success`

valentijnscholten: ty done

tests/dedupe_unit_test.py
```diff
@@ -33,25 +33,50 @@ def check_nb_duplicates(self, expected_number_of_duplicates):
             trs = driver.find_elements_by_xpath('//*[@id="open_findings"]/tbody/tr')
             for row in trs:
                 concatRow = ' '.join([td.text for td in row.find_elements_by_xpath(".//td")])
+                # print(concatRow)
                 if '(DUPE)' and 'Duplicate' in concatRow:
                     dupe_count += 1
 
             if (dupe_count != expected_number_of_duplicates):
-                print("duplicate count mismatch, let's wait a bit for the celery dedupe task to finish and try again (5s)")
+                print("duplicate count mismatch, let's wait a bit for the celery dedupe task to finish and try again (15s)")
```
alles-klar, line 41: There is a small type, waiting for 5s now.

valentijnscholten: ty fixed

PR #2174, opened 2020-04-11
cache system_settings v2

### LINKED ISSUE #2164: Caching of system_settings breaks integration tests
See #2160 for details.

The cache that is used for system setting is local to a worker. So if you have multiple workers, the cache can be out of sync. But only for 30seconds, because that's the livetime of the cache. We (I) thought this would be OK, but in the particular case of having an integration test enabling dedupe and then relying on it, it will sometimes fail. So I guess it's my own fault then :-0. Now thinking about how to fix it. I really don't want to go back to having 50 database queries per request just to retrieve system settings.

The alternative that I considered earlier was to have a middleware that caches the settings per request.

valentijnscholten:
fixes #2164 and properly fixes the workaround from #2160 

This is an improved caching of system settings. The one in #1954 was using the django cache.
This works fine under normal use, but when using multiple workers even the limited ttl of 30s is too much.

This PR introduces a middleware to cache the system_settings during a single request.
It uses a threadlocal, which should be used with care. In this case it's reliable because the middleware 
always cleans up the threadlocal at the end of the request.
Anythin that is not a request (celery tasks) will not use the middleware, so are not affected and will always 
query the database.

Alternatives considered:
- Use an external query caching tool such as Redis. 
This can still be used, but I didn't want to introduces more 
components/containers to be installed/documented/supported.

- Use a dependency injection framework such as django-injector. 
This would require changes to almost any view class and also introduces a completely new concept.

- Use an existing library that does something with caching or storing the request.
I tried https://github.com/benrobster/django-threadlocals, but this doesn't clean up after the request and doesn't handle exceptions.
I tried https://github.com/t-evans/django-threadlocalrequest, but this is 7 years old.

- Use query cache based on django cache
I am using https://github.com/noripyt/django-cachalot and it works fine. But the cache is local to each worker, so would suffer from the same issue as #2160.

So I ended up writing a middleware myself, it's quite simple but feedback is welcome.
I left some debug logging as comments which can easily be enabled if needed.
I'm missing the TRACE log level ;-)

### DIFF
--- a/docker/entrypoint-initializer.sh
+++ b/docker/entrypoint-initializer.sh
@@ -52,11 +52,11 @@ User.objects.create_superuser(
 )
 EOD
 
+  python3 manage.py loaddata system_settings
   python3 manage.py loaddata initial_banner_conf
   python3 manage.py loaddata product_type
   python3 manage.py loaddata test_type
   python3 manage.py loaddata development_environment
-  python3 manage.py loaddata system_settings
   python3 manage.py loaddata benchmark_type
   python3 manage.py loaddata benchmark_category
   python3 manage.py loaddata benchmark_requirement
--- a/dojo/middleware.py
+++ b/dojo/middleware.py
@@ -2,7 +2,11 @@ from django.http import HttpResponseRedirect
 from django.conf import settings
 from django.utils.http import urlquote
 from re import compile
+import logging
+from threading import local
+from django.db import models
 
+logger = logging.getLogger(__name__)
 
 EXEMPT_URLS = [compile(settings.LOGIN_URL.lstrip('/'))]
 if hasattr(settings, 'LOGIN_EXEMPT_URLS'):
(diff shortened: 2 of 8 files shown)


valentijnscholten:
Additional comments from Slack were:
Valentijn Scholten  6 hours ago
You can use redis but you would need another querycaching library to make use of that. There are only a few and none of them seem really good and they can lead to stale data also. And since users already have trouble installing DD as is I don't want to make it harder. We could make it optional or something, but usually when I see a project where I have to install Redis and more just to get it going I quit.

Valentijn Scholten  6 hours ago
the main things is also that I think for most queries we do not want to cache them. We don't want outdated lists of findings, or outdated statuses, or outdated metrics. So system_settings is one of the few things we actually want to cache. Feels a bit over the top to include redis just to cache our system settings. Something that might have been better designed anyway, maybe it should be a config file.

Valentijn Scholten  5 hours ago
I am using https://github.com/noripyt/django-cachalot in my prod instance and it works well. Without redis. It is using the django cache so cache is local to workers so is not a solution for system_settings.

PR #2178, opened 2020-04-12
docker: update to python 3.6

valentijnscholten:
docker: update to python 3.6

### DIFF
--- a/Dockerfile.django
+++ b/Dockerfile.django
@@ -7,7 +7,7 @@
 # Using 3.5.7 to avoid compatibility issues that may be introduced by python 3.6 and 3.7.
 # Please upgrade before end-of-life in september 2020!
 # Ref: https://devguide.python.org/#branchstatus
-FROM python:3.6.10-buster@sha256:1baef6be00b82fbd77f1b60ab227a1dbede6f23825ce1b7f1e9c6f7d1469a45c as build
+FROM python:3.6.10-buster as build
 WORKDIR /app
 RUN \
   apt-get -y update && \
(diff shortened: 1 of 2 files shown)


valentijnscholten:
These were just some tests to see what the test suite in travis would do. It's on the agenda for next meeting.

valentijnscholten:
swtiched the latest to make sure it pulls 3.6. We can add a SHA later, but I am not a fan of it.

ptrovatelli:
yeah I thought of something like, in dev branch use the latest and pin on master, but I don't know how we could do that. It would mean to manually edit all the dockerfiles when merging to master, it won't be done and stuff won't be pinned

pinning is the only way to guarantee repoducibility; we have enough issues with travis right now to not add that kind of noise.

ptrovatelli:
@valentijnscholten I think you're good to go further with this. @madchap has already created a PR for the associated documentation https://github.com/DefectDojo/django-DefectDojo/pull/2274

can you check my comments? adding the sha will not change anything in most case. in rare case only the maintainers will overwrite the tag so we won't be "more up to date" if we pin to the minor and remove the sha. however it will always rule out a source of possible randomness that can generate different behavior from one build to another

if we'd want to be "more up to date" on dev, then we'd need to pull "python 3" for example, but that will generate randomness in the dev build and we'd need to set a procedure to pin that before merging to master.

valentijnscholten:
I think even if we pin to the sha, dependabot will alert is for new (minor) versions. So I think pinning to 3.6#sha would be ok.

valentijnscholten:
So this was merged before travis build was completed resulting in a broken dev branch. New PR raised to revert the changes.

PR #2189, opened 2020-04-13
travis: reorder tests to prioritize important tests

valentijnscholten:
In my experience the two most important Travis steps are the integration tests and flake8. Always good fun to let it build for 40 minutes only to see that you have left 1 space at the end of some line.

This PR reorders the travis build matrix to have these two tests on top so they execute first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,14 +9,14 @@ env:
     - HELM_VERSION=v2.13.0
     - CHANGE_MINIKUBE_NONE_USER=true
   matrix:
+    - TEST=flake8
+    - TEST=integration_tests
+    - TEST=docker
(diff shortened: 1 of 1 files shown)


mtesauro:
I am good with this - makes perfect sense to me to fail early on those kind of things.

valentijnscholten:
I was thinking if it should fail fast or not. Sometimes you want to have all the errors from all the build steps and not just the first one. Anyway, Travis doesn't really support fail fast, so it will always complete all steps. It does support `fast_finish` which will set the overall build status to 'failed' as soon as possible (but will continue running). I have added this now, let's see if it still buils ;-)

PR #2192, opened 2020-04-13
updating dev release-drafter

madchap:
Needed also to make sure `master` does not get overidden with the wrong settings.
`master` is already running with this configuration.


mtesauro:
I am OK with this PR

PR #2193, opened 2020-04-13
renaming plantuml extension, flow modifications

madchap:
renaming to match "official" plantuml extension (and action expectation)


madchap:
The GH action will not trigger without this change.

madchap:
> as discussed, can you have a look at the latest changes in https://github.com/madchap/django-DefectDojo/pull/94/files#diff-4dc853cafe879b0e93cf395267b3b6d1
> 
> * add release name in the diagram
> * change merge to cherry-pick when applicable

I am not sure we should go in that detail here. The "implementation detail" of using cherry-pick over merge is probably to be left in the hand of the maintainer, based on context and the actual thing to backport.

> * removed "on master" "on release" (the diagram should be enough) + there is also some text in the .MD
> 
> regarding the "anonymous" git tag/ dockerhub tag, i still have doubts. It's not good practice to overwrite a tag over and over and it will be complicated for people to understand why they have different results each day while they are still using the same tag (using the SHA256 isn't very intuitive)

I would still argue that for RC releases, it may not be that important. Yet, I fully understand your point. We can have a quick vote on the next meeting ;-)

> maybe we can leave as it is in the diagram, try it out and see how it goes, but if possible it will be better to increment that tag. maybe only after a couple of bug fixes are merged if it's manual? or systematically increment it for each merge if we can do it with github actions

Possible yes. As we don't have stringent timelines when a fix is backported, etc.. it's still a bit ad-hoc. So we've merged like 4 now, 3 remains open, and I just pushed new -rc docker images up (over the same tag for now).

ptrovatelli:
> I am not sure we should go in that detail here. The "implementation detail" of using cherry-pick over merge is probably to be left in the hand of the maintainer, based on context and the actual thing to backport.

sorry for splitting hair here :) but it can't be merge because if we merge, we'll get not only the fix but also all the other changes from the dev branch. I can't think of another way than cherry-pick and I don't think the context will have much impact. The other option is to manually re-do the changes, copy/paste style, which will occasionally end up in merging errors. 

I do like precise instructions because leaving people choices make them enter in a tiresome process of thinking and/or trial and errors, which you don't wan't when dealing with product release (especially the trial and error part). Also by noting "cherry-pick", if there is a better way the next guy will stand up and say that it's not right and it should be done differently. Giving precise instructions will ensure that we'll have a consistent release no matter who's doing it. 

and for the "add release name" part, it's just stuff like replace "Start release x.y.0" by "Create branch release/x.y.0". Again, precise instructions, consistent releases...

If your point is to keep a non-technical diagram why not, then maybe we could put the technical stuff in the .md. Although it may be simpler to just put everything in the diagram and remove the .md chit chat.

madchap:
I added some more precisions @ptrovatelli, if you could once more take a look. Thanks.

PR #2195, opened 2020-04-14
Add GitLab SAST parser

Juu:
Integration for GitLab SAST integrates into GitLab's Continuous Integration and leverages FLOSS scanning tools for the most common languages (<https://docs.gitlab.com/ee/user/application_security/sast/#supported-languages-and-frameworks>). It then reads the results of those tools and creates a report in a common JSON format.

Website: https://docs.gitlab.com/ee/user/application_security/sast/

PR for documentation: https://github.com/DefectDojo/Documentation/pull/84
PR for sample files: https://github.com/DefectDojo/sample-scan-files/pull/49

- N/A Model changes must include the necessary migrations in the dojo/db_migrations folder.
- [x] Add the proper label to categorize your PR

### DIFF
--- a/dojo/tools/gitlab_sast/parser.py
+++ b/dojo/tools/gitlab_sast/parser.py
@@ -1,0 +1,139 @@
+import json
+from dojo.models import Finding
+
+
+class GitlabSastReportParser(object):
+    def __init__(self, json_output, test):
+        self.items = []
+
+        if json_output is None:
+            return
+
+        tree = self.parse_json(json_output)
+        if tree:
+            self.items = [data for data in self.get_items(tree, test)]
+
+    def parse_json(self, json_output):
+        try:
+            data = json_output.read()
+            try:
+                tree = json.loads(str(data, 'utf-8'))
+            except:
+                tree = json.loads(data)
+        except:
+            raise Exception("Invalid format")
+
+        return tree
+
+    def get_items(self, tree, test):
+        items = {}
+
+        for node in tree['vulnerabilities']:
+            item = get_item(node, test)
+            if item:
+                items[item.unique_id_from_tool] = item
+
+        return list(items.values())
+
+
+def get_item(vuln, test):
+    if vuln['category'] != 'sast':
+        # For SAST reports, value must always be "sast"
+        return None
+
+    unique_id_from_tool = None
+    if 'id' in vuln:
+        unique_id_from_tool = vuln['id']
+    else:
+        # If the new unique id is not provided, fall back to deprecated "cve" fingerprint (old version)
+        unique_id_from_tool = vuln['cve']
+
+    title = ''
+    if 'name' in vuln:
+        title = vuln['name']
+    elif 'message' in vuln:
+        title = vuln['message']
+    elif 'description' in vuln:
+        title = vuln['description']
+    else:
+        # All other fields are optional, if none of them has a value, fall back on the unique id
+        title = unique_id_from_tool
+
+    description = 'Scanner: {}\n'.format(vuln['scanner']['name'])
+    if 'message' in vuln:
+        description += '{}\n'.format(vuln['message'])
+    if 'description' in vuln:
+        description += '{}\n'.format(vuln['description'])
+
+    location = vuln['location']
+    file_path = location['file'] if 'file' in location else None
+    sourcefile = location['file'] if 'file' in location else None
+
+    line = location['start_line'] if 'start_line' in location else None
+    if 'end_line' in location:
+        line = location['end_line']
+
+    sast_source_line = location['start_line'] if 'start_line' in location else None
+
+    sast_object = None
+    if 'class' in location and 'method' in location:
+        sast_object = '{}#{}'.format(location['class'], location['method'])
+    elif 'class' in location:
+        sast_object = location['class']
+    elif 'method' in location:
+        sast_object = location['method']
+
+    severity = vuln['severity']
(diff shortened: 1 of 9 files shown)


dojo/tools/gitlab_sast/parser.py
```diff
@@ -0,0 +1,139 @@
...
+    if vuln['category'] != 'sast':
+        # For SAST reports, value must always be "sast"
+        return None
+
+    unique_id_from_tool = None
+    if 'id' in vuln:
+        unique_id_from_tool = vuln['id']
+    else:
+        # If the new unique id is not provided, fall back to deprecated "cve" fingerprint (old version)
+        unique_id_from_tool = vuln['cve']
```
madchap, line 49: Are we sure this is what we want to do here? Would that then still correctly *uniquely* identify this finding? FYI, this field is not mandatory.

Juu: The new field "id" should be introduced soon, in GitLab 13.0, and is a hash combining all the fields of the vulnerability. So I think it's reliable.
This will replace the field "cve": https://gitlab.com/gitlab-org/gitlab/-/issues/209850
The field "cve" is, according to the doc: 
> A fingerprint string value that represents a concrete occurrence of the vulnerability. Is used to determine whether two vulnerability occurrences are same or different. May not be 100% accurate. 

Actually, this "cve" field is dependent from each underlying scanner, in some cases it's a proper hash, in some cases it is a concatenation of some fields and does not guarantee the vulnerability is unique.

Juu: Actually, if you check in the file _dojo/unittests/scans/gitlab_sast/gl-sast-report-many-vuln.json_, you can see that some vulnerabilities have the same "cve" value.

For example, the value `"dojo/static/dojo/js/datatables.min.js:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855:security/detect-object-injection"` has 218 occurences.

I've taken the side of "trusting" this field and I create only one finding in this case (putting the value unique_id_from_tool as a dictionary key in function _get_items_).
But this is only my interpretation

valentijnscholten: So the cve field does not contain a CVE value such as CVE-2020-1234? Weird approach :-)

Juu: No, a real CVE value can be in "vulnerabilities[].identifiers[]" when `vulnerabilities[].identifiers[].type` == 'cve', and in that case I put it in the "cve" field of the defectDojo finding (line 106 of _parser.py_).

But the "vulnerabilities[].cve" field is a fingerprint without any relation with a CVE (yes, that's weird)

valentijnscholten: and probably the reason they are changing it :-) I still stand behind my earlier approval :-) thanks for contributing

Juu: @madchap what do you think?

madchap: OK, thanks for the explanations.

madchap:
Relaunching a Travis before merge.

Maffooch:
> I would like to request a review of the use of `unique_id_from_tool`, I am not certain it is used as it should.

I agree here. Generally, deciding duplicates in these parsers involves comparing hash values of multiple values strung together such as title, awe/cve, and file path. Obviously, using the vulnerability id would be ideal, but the hash method would be a good second as opposed to just cve.

Juu:
> I agree here. Generally, deciding duplicates in these parsers involves comparing hash values of multiple values strung together such as title, awe/cve, and file path. Obviously, using the vulnerability id would be ideal, but the hash method would be a good second as opposed to just cve.

If you allow me to paraphrase myself in the answers to madchap (https://github.com/DefectDojo/django-DefectDojo/pull/2195#discussion_r408646158), the "vulnerabilities[].cve" attribute I use for `unique_id_from_tool` is dependent from each underlying scanner, in some cases it's a proper hash, in some cases it is a concatenation of some fields.

Indeed it does not guarantee the vulnerability is unique (and it will be replaced by a proper hash in upcoming version of GitLab), but has no link to a Common Vulnerabilities and Exposures number (a rea CVE number is stored somewhere else in the parsed file: https://github.com/DefectDojo/django-DefectDojo/pull/2195#discussion_r408662815)

valentijnscholten:
I think the explanation makes sense and the code is good to go.

valentijnscholten:
> I think the explanation makes sense and the code is good to go.

Except that after some merges there are conflicts now. Could you look at them @Juu ? Should be easy fixes.

valentijnscholten:
I think you need to rebase as the PR looks crazy now with 100+ files modified :-)

Juu:
> I think you need to rebase as the PR looks crazy now with 100+ files modified :-)

Well, sorry about that. I tried to rebase, but it seems I cannot go back.
Do I need to create a new PR?

valentijnscholten:
You should be able to rebase if you have an upstream remote definede, i.e. https://gist.github.com/ravibhure/a7e0918ff4937c9ea1c456698dcd58aa

madchap:
I think your `test_type.json` needs a little amendment, another scanner got merged before yours ;-)

Juu:
I think it's ok, the pk is unique.

valentijnscholten:
yeah, it only looks weird in the diff but the contents are fine. Let's merge @madchap  :-)

PR #2200, opened 2020-04-15
Travis reorder: remove finish_fast that is breaking builds

valentijnscholten:
Turns out the `finish_fast` flag breaks the travis builds, PR to remove it.


valentijnscholten:
Needs squashing before merge

PR #2207, opened 2020-04-17
docker-compose-dev: allow chosing of admin password

valentijnscholten:
This PR ensures that the environment variable DD_ADMIN_USER and DD_ADMIN_PASSWORD present at build time are passed on to the initializer for the docker-compose-dev override environment.

This allows users to choose a predicatble admin password (and username) in dev. 

When you are working with different branches / pull requests it's not uncommon to clear containers/images/volumes often, which results in a new random admin password everytime. With this PR it becomes predictable.

The release docker-compose-override still generates a random admin password on initialization.

### DIFF
--- a/docker-compose.override.dev.yml
+++ b/docker-compose.override.dev.yml
@@ -19,6 +19,7 @@ services:
     volumes:
       - '.:/app:z'
     environment:
+      DD_ADMIN_USER: ${DD_ADMIN_USER:-admin}
       DD_ADMIN_PASSWORD: ${DD_ADMIN_PASSWORD:-admin}
   nginx:
     volumes:


Maffooch:
Could you check out the conflicts? A similar PR may have been merged already.

valentijnscholten:
> Could you check out the conflicts? A similar PR may have been merged already.

Yeah looks like part of it was in my other PR. Still some changes needed to make it 100% I think, rebased it.

valentijnscholten:
So we ended up now having the ability to set the admin user/pass via environment variables AND it's now set by default to admin/admin in the dev profile. Is this OK? Or do we not want the defaults?

Maffooch:
I think defaults are fine for dev profile. The only issue I can think of Is if someone started in dev and then switched to release without changing credentials.

PR #2212, opened 2020-04-17
mitigated and accepted findings: strikethrough text decoration

### LINKED ISSUE #1782: Add subtle styling to inactive/closed/duplicate findings
Current situation:
Currently Active and Inactive (closed) findings appear the same in the lists of "All Findings".

image

Proposed situation
Make a subtle distinction, for example the way JIRA does it by making the closed/inactive issues striked through. This helps to immediately see which ones are active / duplicate.

image

valentijnscholten:
fixes #1782

~~closed (inactive)~~ mitigated and risk accepted findings are shown with a text strikethrough effect.
by default when viewing engagements and tests, all findings are shown, including ~~inactive/closed ~~ mitigated / risk accepted ones. adding a subtle styling will help focus on the active/open ones.

before:
image

this PR:
image

EDIT: After some feedback I changed the PR to apply the strikethrough to mitigated findings and risk accepted findings. Should work for the happy path where findings get mitigated or accepted. And for the crazy path where people import findings as inactive intitally ;-)


madchap:
If one imports all findings as non-active/non-verified though (I do now for CI imports for example, before there is a review and since there is still too much noise), the list view will show all findings as striked-through. This will look very odd.

valentijnscholten:
Yeah, don't do that :-). I didn't think of this scenario, but I'm not sure if that makes sense. Importing findings as inactive? I think the `verified` status is made for that?

madchap:
> Yeah, don't do that :-). I didn't think of this scenario, but I'm not sure if that makes sense. Importing findings as inactive? I think the `verified` status is made for that?

I would not think so. A CI import does import a lot of crap that really can't be qualified as "Verified". To me, it can't be even considered "Active" because of the low trust I have in automated scanners findings.

We can probably review what status mean too during our next meeting :)

valentijnscholten:
Could be interesting to hear other people's usecases, for me it seems that importing findings as inactive sounds like abusing that status. But it might be that there is not a good alternative. Maybe we should have a flag for _draft_ findings?
In my view the inactive status is designed to be something that is indeed done and can be forgotten about. I could make a simple settings flag that can disable the striketthrough.
Or add a "custom css" field in the system settings to override stuff. But including user defined css is kind of weird for a security tool :-)

madchap:
> Could be interesting to hear other people's usecases, for me it seems that importing findings as inactive sounds like abusing that status. But it might be that there is not a good alternative. Maybe we should have a flag for _draft_ findings?
> In my view the inactive status is designed to be something that is indeed done and can be forgotten about. I could make a simple settings flag that can disable the striketthrough.
> Or add a "custom css" field in the system settings to override stuff. But including user defined css is kind of weird for a security tool :-)

Yea, I am not disputing your reasoning, I was just explaining what I have right now :) (and I would agree it's most likely abusing that status... however, until we publish something that puts everyone on the same baseline... ;-))

valentijnscholten:
After some feedback I changed the PR to apply the strikethrough to mitigated findings and risk accepted findings. Should work for the happy path where findings get mitigated or accepted. And for the crazy path where people import findings as inactive intitally ;-)

Should we strikethrough also false positive and out of scope? or only mitigated?

madchap:
> Should we strikethrough also false positive and out of scope? or only mitigated?

Sounds good to me.

I was wondering if we could make "style" stuff a flag in the system settings?

valentijnscholten:
You could yes ;-) IT's possible, but the stylesheets are static files so it would involve some messy if statements, or just a "free form" text field in the system settings where admins could include custom css. Not really looking forward to that.

alles-klar:
I agree with @madchap. The strikethrough makes it hard to read the findings and sometimes I also have to investigate closed findings (i.e. when they are automatically closed). If I work with open findings, I use the extra view for it or use filtering. 
If we find a more moderate style it would be ok, strikethrough isn't the right way for me. Maybe just make open/active findings bold?

madchap:
So what about the background color change instead of the strike-through for a more "moderate" style change?

madchap:
@valentijnscholten such a shame to see this PR sitting here. What can we do to bring colors to DD? :-D

madchap:
@valentijnscholten I'd love to see this come through some time. I think we're not so far to have a more appealing experience :-)

valentijnscholten:
changing backgroud is no so easy because the table has odd/even row background colouring. I was thinking about lighter grey font colour or making font italic. but got too many conflicts so will try again later.

damiencarol:
@valentijnscholten any news about this one? Do you want to rebase it?

valentijnscholten:
not as long as we don't know what we want to do with it instead of strikethrough

valentijnscholten:
To be honest I don't really miss the proposed changes....

PR #2218, opened 2020-04-18
implement redirects to open products/engagements by name and by metadata etc

valentijnscholten:
Sometimes products are logically linked to products/applications/projectes/repositories in other system, i.e. gitlab. If the setup is such that their names are equal in both systems, it would be nice if a link to defect dojo could work by product name, i.e.

`https://dojo/s/product/name/valentijn`

In other situations people can use the custom fields in Defect Dojo to add a name+value to logically link the DD product to some other system, i.e. instance=valentijn. In that case it would be nice if a link to defect dojo could work by this custom field, i.e.:

`https://dojo/s/product/meta/instance/valentijn`

It's called meta because that's how DD stores the data.

This PR implements both. There may be more stuff coming around this as I am contemplating a badge system that could use similar urls.

I have put all the redirect code in a separate folder / module.

EDIT: added also some redirects for engagement by name, branch_tag.

### DIFF
--- a/dojo/product/urls.py
+++ b/dojo/product/urls.py
@@ -1,5 +1,4 @@
 from django.conf.urls import url
-from django.urls import path
 
 from dojo.product import views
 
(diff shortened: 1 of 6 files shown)


valentijnscholten:
fyi @christophe226

PR #2219, opened 2020-04-19
fix broken html due to invalid truncate

valentijnscholten:
in `view_eng.html` we use the django built-in filter `truncatechars`. However currently on `dev` this outputs something like this:

```
<tr>
              <td><strong>Commit Hash</strong></td>
              <td><em class="t…</td>
            </tr>
            <tr>
              <td><strong>Branch/Tag</strong></td>
              <td>develop</td>
            </tr>
            <tr>
              <td><strong>Orchestration</strong></td>
              <td>
                    <em class=" text-muted"="">Not Specified</em>
                  
              </td>
            </tr>
````
which breaks the html because it opens a `<em>` tag and never closes it.

using django built-in filter `truncatchars_html" must be used

### DIFF
--- a/dojo/templates/dojo/all_product_findings.html
+++ b/dojo/templates/dojo/all_product_findings.html
@@ -43,7 +43,7 @@
                         {% for finding in findings %}
                             <tr>
                                 <td><a title="{{ finding.title }}"
-                                       href="{% url 'view_finding' finding.id %}">{{ finding.title|truncatechars:60 }}</a>
+                                       href="{% url 'view_finding' finding.id %}">{{ finding.title|truncatechars_html:60 }}</a>
                                        <sup>
                                            {% for tag in finding.tags %}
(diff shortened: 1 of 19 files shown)


madchap:
I see you've put the milestone there, something you wanna see merged into the RC?

valentijnscholten:
Yes, but didn't want to cause more work as it's not high prio. How would this work. You (or me) just cherry picking this commit into the RC branch? Or create another PR? Don't like PR handling, so might as well wait for 1.6.1 or 1.7.0.

madchap:
> Yes, but didn't want to cause more work as it's not high prio. How would this work. You (or me) just cherry picking this commit into the RC branch? Or create another PR? Don't like PR handling, so might as well wait for 1.6.1 or 1.7.0.

I was gonna go for cherry-picking. That's what we documented as well, also introducing a new tag for now I guess to make sure the PR is frozen (`ready-for-backport`).

PR #2224, opened 2020-04-20
Introduce new deduplication algorithm 

piyarathnalakmali:

- [x] Add the proper label to categorize your PR 

**By this PR I have introduced a new deduplication algorithm which identifies duplicate findings based on a set of configured attributes**. 

A finding has an attribute named _line_ which records the code line where the vulnerability is present. When the code gets updated, that vulnerable line can be moved up or down a little. When _line_ is considered in the deduplication algorithm, it fails to identify findings having a slight difference in _line_ as duplicates. So, I have introduced a new attribute named _offset_ which can identify duplicate findings having a difference of _line_ within a range of +/- 100.


valentijnscholten:
The existing dedupe algorithm allows you to specify which fields to consider for deduplication. For static findings the line number can be used, for dynamic findings line numbers should not be used.
Does that not work for you?
Currently it looks like you are creating an entriely new dedupe algorithm with the only difference that it checks the line number for an offset? Also it would break anything that is using the hash_code for deduplication or unique identification of findings. Which we might use more in the future. Coulld you elaborate more, because I might be misunderstanding.

ptrovatelli:
hello all,
I'm not convinced by the arbitrary 100 lines offset; some files will be less than 100 lines and it will be pretty much the same as not looking at the line at all (all vuln of the same type in the file will be duplicate). other big files will sometimes shift by more than 100 lines when introducing big portions of code so it won't find duplicates. 

For what parser do you need this? 

For me the solution to this line number shift issue was to not include the line number is the deduplication configurations :) . the presence of the line number in the legacy algorithm is one of the things that made it flawed. 

The two solutions i've found are implemented in the checkmarx scanners: 
- use the source tool intelligence to tell what is the correct unique identifier of a vulnerability: this is implemented in the scanner "checkmarx detailled scanner" : the field "unique_id_from_tool" holds an identifieir that makes it possible to identify an unique vulnerability, even if the line number changes. checkmarx can mess it up sometimes, but not too often. 
- use a parser that aggregates per filename and count the occurences. this is implemented in the "checkmarx scanner" (not detailed). with this parser there is no line number, problem solved. it makes imports much faster too and the db smaller

ptrovatelli:
same solution is there for sonar parser

ptrovatelli:
@piyarathnalakmali with that info in mind, do you still think that this PR is required? thanks

PR #2230, opened 2020-04-21
merge findings: allow setting to disable it

valentijnscholten:
I have seen some weird stuff happening when users are merging findings and then the next (re)import comes from jenkins or any other ci tooling.
Sometimes existing findings get closed and the new ones get created, but the new ones are then removed by dedupe. And then when the user had chosen merge and delete initially the findings are gone.
Could be some bugs in the importer, but don't have the time to look into it.
Just added a quick setting to allow sysadmins to disable merging alltogether.
It's in settings.py as I don't want to spend too much time maintaining a database migration for this.

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -684,6 +684,10 @@ DEDUPLICATION_ALGORITHM_PER_PARSER = {
     'DSOP Scan': DEDUPE_ALGO_HASH_CODE,
 }
 
+# merging findings doesn't always work well with dedupe and reimport etc.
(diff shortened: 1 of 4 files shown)


mtesauro:
Generally OK with this change.  Do you think adding a DD_DISABLE_FINDING_MERGE is worth adding the the ENV variables?  IMHO, that would be in a separate PR but wondering if people will want to set this at install time.  I'm inclined to say is should be added.
Thoughts?

valentijnscholten:
added that and resolved conflicts.

PR #2235, opened 2020-04-21
similar findings: fix bug and extend with identical issues

valentijnscholten:
Bugfix:
- The field `duplicate_finding_id` was recently removed in the dedupe fields overhaul, braking the showing of buttons for similar findings to mark them as duplicate (or ofiginal).

Enhancements:
- Not only show similar findings, but also show identical findings based on hash_code within the same product. This shows you on which engagement (branch) a finding is also seen. 
- With dedupe on engagement enabled, the buttons for marking as duplicate in this case won't work as there is a server side check to prevent this from completing.
- Hiding the buttons from the template is too much hassle as you need to expose certain functions from the views as template tags. Or just copy and paste code into the html template.
- Also improved the layout a bit to make it more readable.

I'm dogfooding this.

### DIFF
--- a/dojo/templates/dojo/view_finding.html
+++ b/dojo/templates/dojo/view_finding.html
@@ -500,16 +500,16 @@
                                 </td>
                                 <td>{{ similar.status }}{% if similar.duplicate_finding_id %}, <a href="{% url 'view_finding' similar.duplicate_finding_id %}">Original</a> {% endif %}</td>
                                 <td>
-                                    <a target="#" data-toggle="tooltip" data-placement="bottom" title="{{ similar.title }}" href="{% url 'view_finding' similar.id %}">{{ similar.date }}</a>
+                                    {{ similar.date|date }}</br>
                                     {% with similar.notes.count as note_count %}
                                         ({{ note_count }} note{{ note_count|pluralize }})
                                     {% endwith %}
                                 </td>
                                 <td><a href="{% url 'view_test' similar.test_id %}">{{ similar.test.test_type.name }}</a></td>
                                 <td><a href="{% url 'view_engagement' similar.test.engagement_id %}">{{ similar.test.engagement.name }}</a>
-                                {% if similar.test.engagement.version%}
+                                {% if similar.test.engagement.version %}
                                     <sup>
-                                          <a target="_blank" class="tag-label tag-version" data-toggle="tooltip" data-placement="bottom" title="Product Version: {{ eng.version }}">
+                                          <a target="_blank" class="tag-label tag-version" data-toggle="tooltip" data-placement="bottom" title="Product Version: {{ similar.test.engagement.version }}">
                                           {{ similar.test.engagement.version }}</a>
                                     </sup>
                                 {% endif %}
--- a/dojo/templates/dojo/view_finding.html
+++ b/dojo/templates/dojo/view_finding.html
@@ -473,12 +473,14 @@
                 <div id="similar_findings" class="panel-body collapse in">
                     <table class="table table-striped table-hover centered">
                         <tr>
-                            <th>Finding</th>
+                            <th>Title</th>
                             <th>Status</th>
-                            <th>Tags</th>
+                            <th>Date</th>
                             <th>Test</th>
                             <th>Engagement</th>
-                            <th>Version</th>
+                            {% if finding.cwe > 0 %}
+                            <th>CWE</th>
+                            {% endif %}
                             <th>CVE</th>
                             <th>File</th>
                             <th>Deduplicate?</th>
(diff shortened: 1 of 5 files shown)


dojo/templates/dojo/view_finding.html
```diff
@@ -473,48 +473,64 @@ <h4>Similar Findings<span class="pull-right"><a data-toggle="collapse" href="#si
...
+                                        <i class="fa fa-external-link"></i> {{ finding.cwe }}
+                                    </a>
                                 </td>
-                                <td><a href="{% url 'view_test' similar.test_id %}">{{ similar.test.test_type.name }}</a></td>
-                                <td><a href="{% url 'view_engagement' similar.test.engagement_id %}">{{ similar.test.engagement.name }}</a></td>
-                                <td>{{ similar.test.engagement.version }}</td>
+                                {% endif %}
                                 <td>{{ similar.cve }}</td>
-                                <td>{{ similar.file_path }}{% if similar.line > 0 %} (Line {{ similar.line }}){% endif %}</td>
+                                <td title="{{similar.file_path}}">{{ similar.file_path|truncatechars_html:20 }}{% if similar.line > 0 %} (Line {{ similar.line }}){% endif %}</td>
```
madchap, line 525: I would still want to be able to copy the full file path here. The hover alone does not allow me to do that.

dojo/templates/dojo/view_finding.html
```diff
@@ -473,48 +473,64 @@ <h4>Similar Findings<span class="pull-right"><a data-toggle="collapse" href="#si
...
                                     {% with similar.notes.count as note_count %}
                                         ({{ note_count }} note{{ note_count|pluralize }})
                                     {% endwith %}
                                 </td>
-                                <td>{{ similar.status }}{% if similar.duplicate_finding_id %}, <a href="{% url 'view_finding' similar.duplicate_finding_id %}">Original</a> {% endif %}</td>
+                                <td><a href="{% url 'view_test' similar.test_id %}">{{ similar.test.test_type.name }}</a></td>
+                                <td><a href="{% url 'view_engagement' similar.test.engagement_id %}">{{ similar.test.engagement.name }}</a>
+                                {% if similar.test.engagement.version%}
+                                    <sup>
+                                          <a target="_blank" class="tag-label tag-version" data-toggle="tooltip" data-placement="bottom" title="Product Version: {{ eng.version }}">
```
madchap, line 512: The trailing `:` is probably not needed here.

dojo/templates/dojo/view_finding.html
```diff
@@ -473,48 +473,64 @@ <h4>Similar Findings<span class="pull-right"><a data-toggle="collapse" href="#si
...
+                                        {% for tag in similar.tags %}
+                                            <a title="Search {{ tag }}" class="tag-label tag-color"
+                                               href="{% url 'simple_search' %}?query={{ tag }}">{{ tag }}</a>
+                                        {% endfor %}
+                                    </small>
+                                    {% endif %}
+                                </td>
+                                <td>{{ similar.status }}{% if similar.duplicate_finding_id %}, <a href="{% url 'view_finding' similar.duplicate_finding_id %}">Original</a> {% endif %}</td>
                                 <td>
                                     <a target="#" data-toggle="tooltip" data-placement="bottom" title="{{ similar.title }}" href="{% url 'view_finding' similar.id %}">{{ similar.date }}</a>
```
madchap, line 503: Can probably be removed, since we have a whole column to bring about that data now?

dojo/templates/dojo/view_finding.html
valentijnscholten: I don't see it :-)

dojo/templates/dojo/view_finding.html
valentijnscholten: hmm ok, let me see. the problem with these long file paths is that they don't wrap very well.

dojo/templates/dojo/view_finding.html
valentijnscholten: Not sure I understand, what part do you think can be removed?

madchap: The whole hover part. What is displayed is essentially what's now in the new column `Title`

dojo/templates/dojo/view_finding.html
madchap: Ah yes.. well, that's bc in my test it shows `Product Version:` and the `eng.version` is empty.. so it looks weird.

dojo/templates/dojo/view_finding.html
madchap: I have another tool that resolved this by putting a small "copy" button on the left of the string. Would that be possible maybe? Therefore, no need to display the whole line.. we can just copy it and see it elsewhere in full.

dojo/templates/dojo/view_finding.html
valentijnscholten: ah, ok nice catch. And quite special that all lines are made green by github, except the one you are referring to haha

dojo/templates/dojo/view_finding.html
valentijnscholten: I think that's maybe because if the %} without a space earlier. The whole title thing shouldn't be displayed if there's no version.

valentijnscholten: ah eng.version was wrong, fixed in new commit

dojo/templates/dojo/view_finding.html
valentijnscholten: was thinking about that as well, just need to find out what the proper way to copy to clipboard is in 2020.

dojo/templates/dojo/view_finding.html
valentijnscholten: should be fixed in next commit

dojo/templates/dojo/view_finding.html
valentijnscholten: ok, used a library make it "just work"

valentijnscholten:
Great feedback @madchap , should be better now.

valentijnscholten:
@madchap are you happy with the changes?

PR #2273, opened 2020-04-26
Fix dojo version

madchap:
That appears in footer

### DIFF
--- a/dojo/__init__.py
+++ b/dojo/__init__.py
@@ -4,6 +4,6 @@
 # Django starts so that shared_task will use this app.
 from .celery import app as celery_app  # noqa
 
-__version__ = '1.6.0'
+__version__ = '1.6.0-dev'
 __url__ = 'https://github.com/DefectDojo/django-DefectDojo'
 __docs__ = 'http://defectdojo.readthedocs.io/'
--- a/dojo/__init__.py
+++ b/dojo/__init__.py
@@ -4,6 +4,6 @@
 # Django starts so that shared_task will use this app.
 from .celery import app as celery_app  # noqa
 
-__version__ = '1.6.0-dev'
+__version__ = '1.7.0-dev'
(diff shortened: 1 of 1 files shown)


valentijnscholten:
I think it should only become 1.6.0 when merging into master. I think current status should be
master: 1.5.4
release/1.6.0: 1.6.0-rc (or .rc or RC or something)
dev: 1.7.0-dev (as dev will become 1.7.0 if I understand correctly)

madchap:
OK, so I don't like your proposal :-P 

What would eventually be a more painless alternative yet agnostic enough, to the way the version is currently retrieved?

valentijnscholten:
I am open to be convinced, but why would it make sense to have dev branch the same version number as master and rc branch?

alles-klar:
For me the dev branch is for developing the next version, so at the moment we could use 1.7.0 there. The master is the latest version - at the moment 1.5.x and the current release branch has version 1.6.0. 

Otherwise we need to automate this process because we have to update the version very often.

valentijnscholten:
I think we should stick what everybody else is doing and not invent something different. a dev branch is not a version or anything, so it should just be called dev or something like 1.7.0-dev to make clear it's a dev version and not a stable pinpointed version number. Same for RC it should be 1.6.0-rc in this case. Master should be 1.5.4 and when we release it should become 1.6.0. And then when a hotfix branch is there it should be 1.6.1-hotfix and master becomes 1.6.1 when released.
Changing the version number is the least of our problems and one of the reasons the person doing the releases should have write access to all branches.

alles-klar:
For me it isn't super important how we do it, but we need a concept and should document it. The version from @valentijnscholten is also great for me.

madchap:
OK, this is fine. I modified the file. We will need a way to "think" about it, or not.

Issue #2284, opened 2020-04-28
ZAP Scan Parser: do not resolve hostname to IP address during endpoint creation


AlexanderTyutin:
Today we have millions of web applications behind Cloudflare ot another reverse-proxy. So IP value is not so important for scan (if we are not scanning the IP address). Really it will be useful to see the webapp address in the endpoints.

![image](https://user-images.githubusercontent.com/49285004/80485858-d3b6e300-8962-11ea-9f82-c65f1e3b4157.png)

I'm preparing the PR for it.

PR #2286, opened 2020-04-28
ZAP parser: do not resolve hostname to IP address during endpoint creation (#2284)

AlexanderTyutin:
#2284 

ZAP Parser:
* disabled fqdn to ip resolving;
* changed main endpoint visualization from IPport to scheme://fqdn:port

### DIFF
--- a/dojo/tools/zap/parser.py
+++ b/dojo/tools/zap/parser.py
@@ -58,7 +58,7 @@ class ZapXmlParser(object):
         items = list()
         for node in tree.findall('site'):
             site = Site(node)
-            main_host = Endpoint(host=site.ip + site.port if site.port is not None else "")
+            main_host = Endpoint(host=site.name + (":" + site.port) if site.port is not None else "")
             for item in site.items:
                 severity = item.riskdesc.split(' ', 1)[0]
                 references = ''
@@ -129,8 +129,8 @@ def get_attrib_from_subnode(xml_node, subnode_xpath_expr, attrib_name):
 class Site(object):
     def __init__(self, item_node):
         self.node = item_node
+        self.name = self.node.get('name')
         self.host = self.node.get('host')
-        self.ip = self.resolve(self.host)
         self.port = self.node.get('port')
         self.items = []
         for alert in self.node.findall('alerts/alertitem'):


valentijnscholten:
I think it's a good improvement, but I am wondering if we should make it optional to use the old behaviour? Someone has experience with ZAP scans/imports? Changing endpoint value may also impact dedupe.

AlexanderTyutin:
@valentijnscholten Any way except creating new scanner entry like "ZAP Scan with Hostnames"?
We can think about adding a parameter "use_hostnames" with default value "false" if not set, for example. But it will take more efforts I think.
Also really this PR doesn't change the behavior if there was an IP address used as host address.

So, it your target was https://10.0.0.1:8443 you will have exactly the same endpoint. I just removed enforcing of converting from hostname to ip address. I.e. the endpoint is just a real from the report (now endpoint is changing from original by importing tool).

I think this option was useful for the company who wrote this import in 2013 (see the copyright note). But it is obsoleted now.

AlexanderTyutin:
Just look: with this fix we have clear value of the endpoint while without it we have something like XX.XXX.XXX.XXXYYYY (X - IP address, Y - port number)

image

valentijnscholten:
I think it would have to be a setting in settings.py. But you may be right that it is not needed. I have no experience with endpoints at all, so I'll have to leave it to others to decide.

AlexanderTyutin:
A new challenge to learn DefectDojo closer :-D :+1:

mtesauro:
I've only used endpoints a bit - typically they're not as important as the product name + test environment especially for AppSec work.

That said, I think it's a good idea to default to this behaviour (not resolving hostnames to IPs) and to put an boolean in settings.py for those that really want the previous style of behaviour.  Heck, you can control how this works by using the IP or hostname in Zap so it's upstream of DD's import.

TBH, when I've needed/wanted to change the behaviour of a scanner's import, I've just written either a pre or post importing tools that either changes the file before importing or updated findings using the API - so those options will still exist after this PR is merged for those that have special reasons to save scanner data stored in a certain way.

AlexanderTyutin:
Thanks, @madchap. Will try :)

valentijnscholten:
@mtesauro @devGregA I suggest we go with this PR and remove the DNS resolution stuff as it seems out of place anyway. No other scanner importer does this. So no settings flag for the old behaviour, just remove it. WDYT?

Maffooch (approved):
Agreed that removing this the host resolving is a good move. Since the user can choose either the host or IP at the ZAP level then there is no need to force it here.

valentijnscholten (approved):

valentijnscholten:
2 approvals

Issue #2276, opened 2020-04-27
Detail on Acunetix Import


zfirered:
Hey team thanx for Fixing Acunetix Import! :D

can you make mode detail  for Acunetix?
Example when Acunetix found HTML CSRF has 3 finding, but now acunetix import count as 1 HTML CSRF

What i want is Dojo Import the detail too, like
1. HTML CSRF at www.example.com/bla/bla
2. HTML CSRF at www.example.com/asd/
3. HTML CSRF at www.example.com/jkl/

steeve85:
@zfirered can you share an example of Acunetix file? I do not have access to Acunetix to export scan results.

zfirered:
here the example files
[XML_http_example_co_id_.zip](https://github.com/DefectDojo/django-DefectDojo/files/4544179/XML_http_example_co_id_.zip)

PR #2289, opened 2020-04-29
Acunetix parser: Import all affected items + technical details

steeve85:
Fixing the dedupe issue mentioned in #2276 and adding more details from the Acunetix file

- enhancement
- bugfix


valentijnscholten:
Thanks again. I took the libery to make the title more descriptive. Are you sure it fixes said issue? I think the expectation there would be to have 3 findings instead of 1. This PR seems to still have 1 finding, but more details in the description? Not sure how acunetix groups findings, it might be better to keep them grouped.

steeve85:
TL;DR: I will update the deduplication algorithm of the parser

The current version of Acunetix parser do a deduplication on the Finding name. So if the website has 10 XSS, only one will show and the impacted URL will not be displayed in DDJ.

What I did, is that I added the affected path in the title to make sure we have XSS for all impacted URLs. If there is 2 different XSS on the same URL, this might be missing. 

I will update the scanner to add the findings details into the deduplication algorithm. At first, I choose to not do it to remove some noise / False Positive, but I realized that this should be handled manually and not automatically by DDJ because it could remove legit findings.

steeve85:
I updated the parser based on my previous comment.
That will be great to have Acunetix XML format documentation to implement everything properly :)

zfirered:
hi! thanks for fixing!

i really sorry, but can you change () instead of "at"?
so it will look more tidy
example
HTML CSRF (/bla/bla)

steeve85:
@zfirered I updated the title format.

dojo/tools/acunetix/parser.py
```diff
@@ -21,38 +22,30 @@ def __init__(self, xml_output, test):
...
                             references=acunetix_defectdojo_finding.references,
                             impact=acunetix_defectdojo_finding.impact,
-                            false_p=defectdojo_falsep,
+                            false_p=get_false_positive(acunetix_defectdojo_finding.false_p),
                             dynamic_finding=acunetix_defectdojo_finding.dynamic_finding
                 )
-                defectdojo_findings.append(finding)
             else:
-                print(("Duplicate finding : {defectdojo_title}".format(defectdojo_title=defectdojo_title)))
+                print(("Duplicate finding : {defectdojo_title}".format(defectdojo_title=acunetix_defectdojo_finding.title)))
```
valentijnscholten, line 46: can this be turned into a debug log statement, or maybe even removed?

dojo/tools/acunetix/parser_helper.py
```diff
@@ -175,12 +177,20 @@ def get_defectdojo_findings(filename):
     for report_item in acunetix_scan_report.ReportItems:
         defectdojo_finding = dict()
 
-        defectdojo_finding['title'] = report_item['Name']
+        if "Affects" in report_item:
+            affects = (" ({})".format(report_item['Affects']))
+        else:
+            affects = ""
+        defectdojo_finding['title'] = "{}{}".format(report_item['Name'], affects)
```
valentijnscholten, line 184: Changing the title format will break deduplication as the acuentix parser does not yet use the new DEDUPLICATION_ALGORITHM_PER_PARSER config in settings.py.

dojo/tools/acunetix/parser.py
steeve85: I can change that

dojo/tools/acunetix/parser_helper.py
steeve85: What do you recommend to do in that case?
- Should I not update the title? Doing this will remove similar vulnerabilities on different endpoints
- Modify the deduplication algorithm in settings.py?
- Something else?

valentijnscholten:
I don't know the details are about the request, but is it now that in the old situation Acunetix was reporting 1 finding and in the new situation it is reporting the same finding again, but additionally more findings if there are affected items? In that case I would try to keep the title of that first finding the same. The new ones can have the affects part in their title.

Additionally it would be good or even better to configure the dedupe algo. If the title it not needed for that, it's less of an issue if it changes.

zfirered:
> I don't know the details are about the request, but is it now that in the old situation Acunetix was reporting 1 finding and in the new situation it is reporting the same finding again, but additionally more findings if there are affected items? In that case I would try to keep the title of that first finding the same. The new ones can have the affects part in their title.
> 
> Additionally it would be good or even better to configure the dedupe algo. If the title it not needed for that, it's less of an issue if it changes.

if the new situation it is reporting the same finding again, just skip it, and just update the time and date the finding

example, the first one 
HTML CSRF > this was found on date 13/05/2020
and when the next one Acunetix has the same finding just skip it and just update the date so in the list will be like this
HTML CSRF > 15/05/2020

as for the title i think we really need to show where the affectted items is, or you can make more tidy for that?

steeve85:
@valentijnscholten I replaced the `print` statement by a logging message.
Regarding the deduplication process, I think the change is worth it. Currently, lot of findings are missing because if Acunetix identify an XSS on pageA and pageB, only one of them will be imported into DefectDojo. So an Acunetix scan with hundreds of XSS (or other type of vulnerability), only one will be reported, which is not very useful.

With this PR update, all vulnerabilities (example: "XSS on pageA" and "XSS on pageB") will be imported into DefectDojo.

The only issue I see is for current users of Acunetix parser, they will have to manually process the deduplication. But I think that is better than missing lot of findings (identified on different pages).

@valentijnscholten if my explanation is not clear, we can discuss about it on Slack, it might be easier

valentijnscholten:
I also took a look at the sample file in 2273 and it looks like these affects values are (relative) urls? This makes me think these should map to endpoints?

steeve85:
@valentijnscholten I updated the debug log.
I did not update the endpoint because the sample files we have don't have a good variety of url. All vulnerabilities are reported at the root. I can work on this later when we get better Acunetix files for testing

mtesauro:
FYI: I've merged the sample files:
https://github.com/DefectDojo/sample-scan-files/pull/53

madchap:
Relaunched the travis job.

dojo/tools/acunetix/parser_helper.py
valentijnscholten: Best would be to modify/set the dedupe algorithm in settings.dist.py

steeve85: I made an update. Hope it's all good

madchap:
@steeve85 I resolved a couple conflicts, and a couple flake8 is failing on a couple points.

valentijnscholten (dismissed):

valentijnscholten (approved):
looks good to me, but I have experience in importing Acunetix reports so hopefully someones else can confirm the logic.

devGregA (approved):

mtesauro (approved):
Approved

Maffooch (approved):

PR #2342, opened 2020-05-09
Adding GH action on dev branch

madchap:
Explicitly adding workflow on dev too.
GH doc seems to say it has to be on the branch, but I think/thought it's only on the default branch. Would not make sense otherwise.
But for the sake of troubleshooting I guess..

### DIFF
--- a/.github/workflows/plantuml.yml
+++ b/.github/workflows/plantuml.yml
@@ -1,0 +1,39 @@
+name: Generate PlantUML Diagrams
+on:
+  push:
+    paths:
+      - '**.puml'
+    branches:
+      - master
+      - dev
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    env:
+      UML_FILES: ".puml"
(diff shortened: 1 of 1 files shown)


madchap:
I guess not in the end. It works on other test repos, not sure what's wrong here.

madchap:
Correction -- so re-opening.

madchap:
Somewhat unrelated, but while performing tests, our "branching" diagram fails to produce png/svg, because the github action uses a too old version of plantuml and there is something too "new" in there. 

So a linked PR https://github.com/cloudbees/plantuml-github-action/pull/4 there to bump the version.

Otherwise, the workflow file needs to be present in branches where you want it to trigger, and not only the default branch.

mtesauro:
Are you wanting a review on this cause it LGTM? Let me know.

madchap:
No, it's all good.. but still needs to approvals to merge on dev ;-)

PR #2343, opened 2020-05-09
finding: add index to hash_code, line and unique_id_from_tool

valentijnscholten:
@madchap  found out the hard way that there is no index on `hash_code` field (for Finding).
This PR adds an index. The field also turned out to be a TextField which cannot have an index, so we change it into a CharField with max_length=64 for sha256.

Potentially a migration that is slow, but on my instance with around 3000 findings it was done in 1 second.

v2: the (async) dedupe code shows we need also indices on `line` and `unique_id_from_tool` fields.


madchap:
I will try this out on a local instance loaded up with all my 100Ks of findings... sh** is really hitting the fan hard on my end. Thanks for looking into this!

valentijnscholten:
@madchap did you get a chance to test? I'm about to rebase this and update the migration, but need to be confident it gets merged soon.

PR #2348, opened 2020-05-12
zap parser: add title to dedupe fields + unittests

valentijnscholten:
The current dedupe config for ZAP scans is "not unique enough':

`ZAP Scan': ['cwe', 'endpoints', 'severity']`

even our own zap sample contains multiple different findings with the same `cwe` + `endpoints` + `severity`, so we need to include either the `name` or `pluginid` or `alert` fields.

https://github.com/DefectDojo/django-DefectDojo/blob/dev/tests/zap_sample.xml

Only the `name` field is available in defect dojo and it is stored in the `title` field, so unfortunately we have to include the `title` field`

Also add unittests for the zap parser as they were missing (just basic tests as any other parser, not dedupe or anything).


valentijnscholten:
What do you think @ptrovatelli

ptrovatelli:
OK what you're pointing out is problematic indeed. I guess we can use the title until we find something better;

Issue #2346, opened 2020-05-12
Limit on report size that can be imported


madchap:
**Is your feature request related to a problem? Please describe**
Right now, anyone can import any size a report, which can lead to a DoS -- as I can experiment now. For example, git repositories scans can generate a ton of findings (esp. if you don't pay the right attention to false-positives). 
In my case, well over 100K across all branch caused a DoS on my instance.

We should have an upper limit to mitigate DoS.

**Describe the solution you'd like**
The idea is to limit -- both at UI and API levels -- the size of the report that can be uploaded. It could be number of rows or file size.

valentijnscholten:
isn't there a limit by default? we get questions about how to raise the limit on a weekly basis in the clack channel. Or is that only for the standalone / setup.bash installs?

madchap:
Maybe through the UI, but not the API somehow.

madchap:
https://github.com/DefectDojo/django-DefectDojo/blob/dev/nginx/nginx.conf#L14

800m.. we're asking to be DoS'ed it seems.

Maffooch:
Seems to be fixed

PR #2350, opened 2020-05-12
Reduce nginx client_max_body_size

madchap:
Let's avoid getting DoS'ed by supersized reports.

Addressed #2346


mtesauro:
There's a way to configure a one-off Nginx config already if you're using docker-compose.  I just committed to the Community-Contribs repo this deployment example that includes a mechanism to have a custom nginx for just your install without any changes to the default defectojo-nginx container.  Its at https://github.com/DefectDojo/Community-Contribs/blob/master/CentOS-7-deploy/dojo/docker-compose.yml#L19
It would probably also work for k8 but I've not tried it there.

madchap:
Superseded by https://github.com/DefectDojo/django-DefectDojo/pull/2528

PR #2359, opened 2020-05-13
fix alerts for product_added, and more

valentijnscholten:
There were some issues with notifications / alerts after the revamp earlier:
- alerts were not created for staff/admin users
- ~~alerts were failing for product_added notifications~~ already fixed in #2412

### DIFF
--- a/dojo/notifications/helper.py
+++ b/dojo/notifications/helper.py
@@ -17,7 +17,12 @@ def create_notification(event=None, *args, **kwargs):
         system_notifications = Notifications()
 
     logger.debug('creating system notifications')
-    process_notifications(event, system_notifications, *args, **kwargs)
+
+    # send system notifications to all admin users
+    admin_users = Dojo_User.objects.filter(is_staff=True)
+    for admin_user in admin_users:
+        system_notifications.user = admin_user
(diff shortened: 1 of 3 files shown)


dojo/notifications/helper.py
```diff
@@ -17,7 +17,12 @@ def create_notification(event=None, *args, **kwargs):
         system_notifications = Notifications()
 
     logger.debug('creating system notifications')
-    process_notifications(event, system_notifications, *args, **kwargs)
+
+    # send system notifications to all admin users
+    admin_users = Dojo_User.objects.filter(is_staff=True)
```
madchap, line 22: I really dislike this `is_staff` anymore. So, since pretty much all would be staff, everyone would get notifications then :-/
Do we want to start using `is_superuser` in some places, which could translate as "I am really an admin of DefectDojo"? Note this is not mega important, until we tackle that other PR... Just thought I'd throw this comment in anyways.

valentijnscholten: yeah, it 's not ideal. But I think we want to decide on how we want to progress with roles/permission and then adjust notifications on top of that. the above is more to restore the old functionality which I broke with my previous PR.

Issue #1754, opened 2020-01-14
Safety Import Scan Error 500


Dorfus1988:
**Bug description**
When i am trying to import a safety scan result (JSON), it fails with status code 500

**Steps to reproduce**
git clone --branche master https://github.com/DefectDojo/django-DefectDojo && cd django-DefectDojo
docker-compose build
docker-compose up
Import a Safety Scan in an engagement 

**Expected behavior**
The scan result should be processed

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [x] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: [e.g. Windows 10]
 - DefectDojo version: Master Branch DefectDojo 
Commit Message: d8cd7607: Merge pull request #1729 from DefectDojo/Maffooch-patch-1 [2020-01-03 18:41:53 -0600]

**Console logs** (optional)
uwsgi_1         | Internal Server Error: /engagement/4/import_scan_results
uwsgi_1         | Traceback (most recent call last):
uwsgi_1         |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/exception.py", line 34, in inner
uwsgi_1         |     response = get_response(request)
uwsgi_1         |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 115, in _get_response
uwsgi_1         |     response = self.process_exception_by_middleware(e, request)
uwsgi_1         |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 113, in _get_response
uwsgi_1         |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
uwsgi_1         |   File "/usr/local/lib/python3.5/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
uwsgi_1         |     return view_func(request, *args, **kwargs)
uwsgi_1         |   File "./dojo/engagement/views.py", line 523, in import_scan_results
uwsgi_1         |     parser = import_parser_factory(file, t, active, verified)
uwsgi_1         |   File "./dojo/tools/factory.py", line 158, in import_parser_factory
uwsgi_1         |     parser = SafetyParser(file, test)
uwsgi_1         |   File "./dojo/tools/safety/parser.py", line 10, in __init__
uwsgi_1         |     response = urllib.request.urlopen(url)
uwsgi_1         | NameError: name 'urllib' is not defined

Dorfus1988:
[...]

PR #2360, opened 2020-05-13
Fix safety parser (Issue #1754)

k-nguyen:
Bug fixes for safety parser:
- fixed issue #1754 when cve for finding is `None`
- fixed issue with importing empty file
- set cve to finding's object for safety


valentijnscholten:
Thanks!

PR #2365, opened 2020-05-15
Datatables for engagement list and test view

madchap:
For some use-cases, such as scanning hundreds or more github repositories for secrets under a product, having the power of the datatables is a must have.
That dynamic search box does wonders.

Since I don't really know what I am doing here, this is still a draft :)

I also re-aligned the ellipses to be on the left side, like for products.

### DIFF
--- a/dojo/templates/dojo/product.html
+++ b/dojo/templates/dojo/product.html
@@ -365,18 +365,18 @@
                         columns: ':not(.noVis)'
                     },
                     {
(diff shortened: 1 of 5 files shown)


madchap:
Current "status":

image

madchap:
Seems to perform as expected now, no console error. Let's see what Travis says. Converting to regular PR.

PR #2366, opened 2020-05-15
restore pdf export button and update datatable to 1.0.21

valentijnscholten:
the latest update of datatables was missing the pdf plugin so the pdf export button was lost.
this pr updates datatable to 1.0.21 and now includes the pdf export plugin.

added later:
- export options were lost at some point during the last few months, this PR restores them which mainly means useless columns are excluded from the exports
- simplified the initialization logic as datatables now by default strips html for exports, so less custom export data extractors are needed.


Maffooch:
I'm still not seeing a pdf button. Fresh build.
image

valentijnscholten:
oh wow, indeed. It works on my machine, but turns out the `dojo/static` folder was on the `.gitignore` list for some reason? Corrected that now!

Maffooch:
Still not appearing :/ Tried removing every single image to rule out older cached versions. Not sure what the deal is!

JoseRoman:
Maybe a cross-browser compatibility issue?

PR #2373, opened 2020-05-16
make notifications work with users created without usercontactinfo

valentijnscholten:
Add a check to avoid exceptions for users created without usercontactinfo (i.e. via /admin, or manage.py createsuper).


valentijnscholten:
Turns out it can be simplified a little bit since django 1.6, which I did.

PR #2378, opened 2020-05-17
Merge master into release 1.6.0

valentijnscholten:
In the new (git flow based) branching/release model, our branches should never be behind master. At least not for a long time or for many commits.

image

Currently `dev` and also `release/1.6.0` are 62 commits behind. This PR resolves that for the `release/1.6.0` branch. After the release, the release branch will be merged to `master`, but also to `dev` as per git flow and the plantuml diagram by @madchap 

The step in this PR is only needed one-time because in the past we didn't have a clear branching/releasing model.

Does this make sense?

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -6,7 +6,7 @@ update_configs:
     update_schedule: "live"
 
   - package_manager: "javascript"
-    directory: "/"
+    directory: "/components"
     target_branch: "dev"
     update_schedule: "live"
 
--- a/.github/workflows/plantuml.yml
+++ b/.github/workflows/plantuml.yml
@@ -1,0 +1,39 @@
+name: Generate PlantUML Diagrams
+on:
+  push:
+    paths:
+      - '**.puml'
+    branches:
+      - master
+      - dev
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    env:
+      UML_FILES: ".puml"
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v1
+      - name: Get changed UML files
+        id: getfile
+        run: |
+          echo "::set-output name=files::$(git diff-tree -r --no-commit-id --name-only ${{ github.sha }} | grep ${{ env.UML_FILES }} | xargs)"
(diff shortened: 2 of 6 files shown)


mtesauro:
I'm for anything that gets us sync with the new branching model.  Thanks for doing this work.

madchap:
IIUC correctly, RC is not merging back to `dev`, since anything that comes to RC comes from `dev` in the first place. This is also for the diagram depicts at https://github.com/DefectDojo/django-DefectDojo/blob/dev/BRANCHING-MODEL.md .. the dotted arrow line back to `dev` is not a "merge" line, just a "come back" line as per its comment ("release done").

madchap:
Also, if going straight to https://github.com/DefectDojo/django-DefectDojo/compare/release/1.6.0...master?expand=1 it does seem to give the same merge results as this PR.

14 files changed for 61 commits. 

I will draft the PR directly from the 2 branches, it is probably better to do it directly across GH "local branches" than through an intermediary PR.

valentijnscholten:
Why are you trying to recreated the PR I already did? And where I fixed merged conflicts, and fixed the yarn install, etc?

valentijnscholten:
> IIUC correctly, RC is not merging back to `dev`, since anything that goes to RC comes from `dev` in the first place. This is what for the diagram depicts at https://github.com/DefectDojo/django-DefectDojo/blob/dev/BRANCHING-MODEL.md .. the dotted arrow line back to `dev` is not a "merge" line, just a "come back" line as per its comment ("release done").

Yeah, that was why I said I think that's not ideal. Recently we have merged some bugfixes for 1.6.0 into dev. How do we know which ones have been merged into 1.6.0? We should follow git flow and do release bugfixes on the release branch. Then merge release branch into dev.

madchap:
> How do we know which ones have been merged into 1.6.0? We should follow git flow and do release bugfixes on the release branch. Then merge release branch into dev.

I look at the commits on a given branch to see what have been cherry-picked and relate it to what appears on the related milestone. But I think this may not be sustainable.

PR #2390, opened 2020-05-18
Add risk acceptance flag and simple bulk risk acceptance

valentijnscholten:
This builds upon (and includes) #2379, so merge that first.

This PR: 
- add risk acceptance boolen status flag
- add bull risk acceptance (and unaccceptance) 
- bulk (un)acceptance works globally, so acrross engagements, across products.

Screenshots:
image

image

image

image

Under the hood this still uses the same `Risk_Acceptance` model, and it will also show up in the view engagement page. You can add/remove findings from that as usual, but only *per engagement* (that's how it was setup originally and I didn't want to refactor everything at once).

It's called "Simple" Risk Acceptance, because later on we may add bulk risk acceptance with custom expiration dates, custom names, custom owners, etc.

This should allow at least for people to start using the risk acceptance feature similar to having the status flags.

There are some more improvements that could be made, but those might be for a next PR:
- Add simple risk acceptance to API?
- Add tests

I'd prefer to merge this PR now because we can start using it and gain experience and feedback. Nothing existing is changed/removed, just adding a simple variant of risk acceptance.


mtesauro:
+ 1 on:
There are some more improvements that could be made, but those might be for a next PR:

    Add simple risk acceptance to API?
    Add tests

And I think a follow-up PR makes perfect sense.

PR #2394, opened 2020-05-19
Merge master to 1.6.0

madchap:
PR created while looking at #2378, which only shows 6 modified files.

This PR is created directly on DD's upstream branches, without any intermediary code pull.


madchap:
Unsure why `entrypoint_scripts/common/dojo-shared-resources.sh` re-appears here, since both branches do have the proper changes in :thinking:

madchap:
Resolving the conflicts here would merge to `master`, so we probably should hold off until all pending PR to be applied to 1.6.0 are done and we're ready to release.

madchap:
Talked with @valentijnscholten, moving with sibling PR.

PR #2426, opened 2020-05-25
return age and sla days remaining for findings apiv2

### LINKED ISSUE #2423: Exposing Finding.sla to API
Hey team :)

I would like to use Defectdojo API in my CI/CD, in the way I can config jobs to failed or not based on SLAs and severity.

Premisses: I am sending reports to defectdojo:
Eg: I have 1 repo that is been scanned by static analysis code. If the SAST tool finds 1 vuln with a medium severity, I will have 30 days to fix it. In case the next job finds the same vuln, the job should fail if SLA is overdue.
This logic could be possible using "finding" property "sla" in api.

What do you think? Does make sense to expose this property?

valentijnscholten:
fixes #2423


edersonbrilhante:
Can someone review it?

PR #2434, opened 2020-05-28
Upgrade docker containers to python 3.6

madchap:
With additional care this time around.

### DIFF
--- a/Dockerfile.nginx
+++ b/Dockerfile.nginx
@@ -3,13 +3,15 @@
 # The code for the build image should be idendical with the code in
 # Dockerfile.django to use the caching mechanism of Docker.
 
-FROM python:3.5.9-buster@sha256:1baef6be00b82fbd77f1b60ab227a1dbede6f23825ce1b7f1e9c6f7d1469a45c as build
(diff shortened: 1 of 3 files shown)


madchap:
We updated the PR template to now accept 3.6 contributions. Should this then be part of 1.6.1?

valentijnscholten:
> We updated the PR template to now accept 3.6 contributions. Should this then be part of 1.6.1?

don't think so as anything that gets contributed will go into dev/1.7.0 anyway which will be 3.6 then?

Dockerfile.nginx
```diff
@@ -29,7 +31,7 @@ ENV \
   node="nodejs"
 RUN \
   apt-get -y update && \
-  apt-get -y install apt-transport-https ca-certificates && \
+  apt-get -y install apt-transport-https ca-certificates curl wget && \
```
alles-klar, line 34: do we really need to install curl and wget? Doesn't seem related to the switch from python 3.5 to 3.6?

valentijnscholten: probablty needed because they are no longer included by default?

madchap: Indeed.

ptrovatelli:
these need an update  , in dockerfile.django:
```
# Using 3.5.7 to avoid compatibility issues that may be introduced by python 3.6 and 3.7.
# Please upgrade before end-of-life in september 2020!
```

PR #2436, opened 2020-05-28
Added endpoint to re_open a finding via the API

Apipia:
re-open a finding via the API. 
Stole most of the logic from the finding/views.py open endpoint logic.

pretty straight forward, but it will allow us to update a finding and remove the "Mitigated" status via the API which currently isn't possible.

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -192,6 +192,19 @@ class FindingViewSet(mixins.ListModelMixin,
         else:
             return serializers.FindingSerializer
 
+    @action(detail=True, methods=['get'])
+    def re_open(self, request, pk=None):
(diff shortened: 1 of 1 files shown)


Maffooch:
Oops didn't see the conflicts

dojo/api_v2/views.py
```diff
@@ -192,6 +192,19 @@ def get_serializer_class(self):
         else:
             return serializers.FindingSerializer
 
+    @action(detail=True, methods=['get'])
```
madchap, line 195: The `GET` method that actually induce a change seems weird to me. In my mind, `GET` should trigger no change.

Would it better to eventually call `PUT` or `PATCH` instead of `GET`?

Issue #2121, opened 2020-03-27
Server 500 error when viewing finding if duplicate is True and duplicate_finding is None


keenan-v1:
**Bug description**
When a finding has `duplicate` as `True` and yet `duplicate_finding` is `None` the server will give a 500 response when viewing that finding. This state can be achieved via the API or via Edit Finding and marking the finding as a duplicate.

**Steps to reproduce**
1. Edit a finding
2. Ensure Active and Verified are not checked, then check Duplicate
3. Save the finding.
 
**Expected behavior**
I expect the server to return the finding page even if `duplicate_finding` is `None` as you cannot assign a `duplicate_finding` due to it being not editable.

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [X] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: Debian Stretch
 - DefectDojo Commit Message: 64d5f348: Merge pull request #2032 from DefectDojo/update-contributing-master [2020-03-12 22:41:42 +0100]

**Console logs** (optional)
https://pastebin.com/QCRgKp2F

**Additional context** (optional)
The cause of the 500 error is accessing `id` on a NoneType, which in this case is `duplicate_finding`. To fix the symptom, simply add `and self.duplicate_finding` to the if statement on line 1496. However I feel there's a greater issue that you can modify `duplicate` to `True` via the API and the Edit Finding page, yet you cannot indicate what original finding it is a duplicate of due to `duplicate_finding` being read-only.

valentijnscholten:
Lots has changed regarding this on `dev` branch. Could you confirm it's still present there?

keenan-v1:
> Lots has changed regarding this on `dev` branch. Could you confirm it's still present there?

Ah, I posted the commit message from the wrong branch. I was working out of dev. I'll do a pull and re-test though and confirm if it's still an issue.

keenan-v1:
Still able to be reproduced. Here's the commit I'm at:

d88b957d: Add component name and version for JFrog scans (#1979) [2020-03-23 16:28:01 +0100]

As I said, the cheap fix is just checking for `duplicate_finding` as that's where the exception is happening. The bigger fix is allowing us to link the actual duplicate from the Edit screen and via the API.

valentijnscholten:
That's not latest dev.

keenan-v1:
> That's not latest dev.

I edited the comment.

Apipia:
[...]

PR #2445, opened 2020-05-29
Setting a duplicate without a parent_finding

Apipia:
Will keep from receiving a 500 error when viewing the Finding.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1557,8 +1557,12 @@ class Finding(models.Model):
                 r.delete()
 
     def duplicate_finding_set(self):
-        if self.duplicate and self.duplicate_finding is not None:
-            return Finding.objects.get(id=self.duplicate_finding.id).original_finding.all().order_by('title')
+        if self.duplicate:
+            if self.duplicate_finding is not None:
+                return Finding.objects.get(
+                    id=self.duplicate_finding.id).original_finding.all().order_by('title')
+            else:
+                return []
         else:
             return self.original_finding.all().order_by('title')
 
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1557,7 +1557,7 @@ class Finding(models.Model):
                 r.delete()
 
     def duplicate_finding_set(self):
-        if self.duplicate:
+        if self.duplicate and self.duplicate_finding is not None:
             return Finding.objects.get(id=self.duplicate_finding.id).original_finding.all().order_by('title')
         else:
             return self.original_finding.all().order_by('title')


Apipia:
Will resolve Issue #2121

valentijnscholten:
I find the naming of this relation very confusing:

```
    duplicate_finding = models.ForeignKey('self', editable=False, null=True,
                                          related_name='original_finding',
                                          blank=True, on_delete=models.CASCADE)
```

If I understand the code correctly the field is called duplicate_finding, but it stores actually the original finding that is being duplicated here? So we have the names duplicate_finding for the field, but when accessing the relation you need to use original_finding? 

From what I can see this PR makes the if statement wrong. If a finding is marked as duplicate, but there are actually no duplicates, it will switch to returning original findings? As I said I don't understand how it works because of the namings, but shouldn't it be something like:

```
    def duplicate_finding_set(self):
        if self.duplicate:
			if self.duplicate_finding is not None:
				return Finding.objects.get(id=self.duplicate_finding.id).original_finding.all().order_by('title')
			else:
				return []
        else:
            return self.original_finding.all().order_by('title')
```

Apipia:
@valentijnscholten To be honest, I am very much confused about the model and that code. I've made the change that you suggested and there seems to be no difference in the UI about how information is displayed one way or the other. 

Before the logic to me seemed like 
```
IF it is a duplicate and we can actually get those Findings,
    List those duplicate findings
ELSE
    ? List the original finding ?
```
And the change is now:
```
IF it is a duplicate
    If we can get those duplicate findings:
        List them
    ELSE
        Empty list
ELSE
    ? List original finding ?
```

I'm not really sure what ` return self.original_finding.all().order_by('title')` actually does.
This function is called in the template:
```
{% if finding.duplicate_finding_set %}
    <td>
        <div class="align-top">
        <div class="dropdown">
            Duplicate List <a href="#" data-toggle="dropdown">&nbsp;<i class="fa fa-ellipsis-v"></i>&nbsp;</a>
             <ul class="dropdown-menu" role="menu" aria-labelledby="dropdownMenu1">
              {% for duplicate in finding.duplicate_finding_set %}
                    <li>
                     <a class="" href="{% url 'view_finding' duplicate.id %}">
                     <i class="fa fa-bug"></i> {{ duplicate.title }}, <i>{{ duplicate.date }}</i>
                     </a>
                      </li>
               {%  endfor %}
               </ul>
          </div>
       </div>
    </td>
```
You would think an empty list would still show the **Duplicate List** text somewhere on the page but I couldn't find it. Maybe I'm just not sure I understand how this logic works.

Apipia:
Maybe `{% if finding.duplicate_finding_set}` resolves to False if it returns an empty list?

valentijnscholten:
Yeah, empty lists are {{False}} according to Python. Looking at it again, I think the else part of this if will result in an error. The original_finding field will return 1 single finding, which doesn't have an all() function.
With the original version of this PR I think it would get None as original_finding so would return another error. I think this is a left over from the PR that changed some fundamentals regarding duplicates.
I think this PR is now as good as it can get :-)

valentijnscholten:
@Apipia Can I suggest to put this into 1.6.1? Otherwise we risk to have it open for a long time, get conflicts maybe, etc.

valentijnscholten:
just to be sure: you need to rebase and change the base branch of the PR.

Issue #2460, opened 2020-06-02
Hiding "Bug" Icon for the Products without JIRA config


dayal18:
**Is your feature request related to a problem? Please describe**
Hiding the "Bug" / JIRA icon, when the product is not configured for JIRA would help to have a quick view on what applications are enabled with JIRA and what is not. Currently, you can only get this information when you hover over your mouse on that icon.

**Describe the solution you'd like**
If the Bug icon is hidden it would help to quickly analyze or sort the products without JIRA configurations.

![image](https://user-images.githubusercontent.com/43857391/83567720-a3c1a900-a4e7-11ea-924d-7004dd6256e8.png)

madchap:
I don't see the bug icon when a product has no JIRA configuration. Can you please detail? Thanks.

Could you check if what happens in the following scenario:
- Configure JIRA for a product and save --> the icon is there
- Remove the JIRA configuration --> the icon is there
- Create a new product, but never touch its JIRA configuration --> confirm no icon is present.

I think the DB may have an empty record, which then gets shown.

dayal18:
Yes, its Bug...

Use case 1: Create a new Product --> No JIRA --> Save... there is NO ICON
Use case 2: Create a new product --> No JIRA --> Save --> Edit again --> no changes --> Save --> See icon

madchap:
Yes, got it. Fixed in a local PR of mine, will push later.

dayal18:
Thank you..

dayal18:
@madchap Thanks that fix works :)

Issue #2458, opened 2020-06-02
JIRA WebHook and  other test use cases including bulk + product level push all (Master 1.6 - 654801f9)


dayal18:
**Bug description**
Tested few uses cases with JIRA 'Push all findings + Notes + Individual push of findings'
Here are my results.

**use case 1: (Worked)
In this use case, I have disabled push to JIRA at the Product level, as well as disabled Notes**

**Product level - JIRA disabled, Notes disabled**

Create new finding in a new engagement --
1) In new finding - status = Active + verified and select 'push to JIRA' - Works
2) In a finding - status = Active, then edit the finding - status = Active + Verified and select 'push to JIRA' - Works

**use case 2: (Half worked, error handling should be enabled)
In this use case, I have enabled push to JIRA at the Product level, and disabled Notes**

**Product level - JIRA enabled, Notes disabled**

Create new finding in a new engagement --
1) In new finding - status = Active + verified and auto-select 'push to JIRA' - Works
2) In a finding - status = Active then save (though auto-select JIRA is enabled it will save), then edit the finding - status = Active + Verified and auto-select 'push to JIRA' - Doesn't work, the JIRA Age and JIRA change appear - but with no JIRA ticket info or actual JIRA ticket creation. (I repeat this step of editing again, no luck) -- see step 4 when I used 'bulk edit'
3) In a finding - status = Under Review (anything other than Active or Active + verified), then I cannot save finding, as JIRA requires Active + Verified to be enabled by default.
4) I tried bulk edit option and select only 1 or all the findings and 'Push to JIRA' -- This step worked and fixed the issue of Step 2.

**use case 3: (Half worked, error handling should be enabled)
In this use case I have enabled push to JIRA at the Product level, and also enabled Notes (because I need bi-directional JIRA conversation)**

Results the same as to use case 2...

**use case 4: (Worked, but some cosmetic changes as the info about JIRA age and JIRA change doesn't populate)
In this use case, I have disabled JIRA and enabled NOTES at product level..**

**Product level - JIRA disabled, Notes enabled**

Create a new finding in a new engagement ---
1) In new finding - status = Active + verified and select 'push to JIRA' - Works
2) In a finding - status = Active, then edit the finding - status = Active + Verified and select 'push to JIRA' - Works
3) Create all findings status = Active + Verified, then finally use the Bulk Edit options and select 'Push to JIRA' - Works but no JIRA Age or JIRA Change info update on DD.

[...]

PR #2402, opened 2020-05-20
JIRA Config Bulk update fix

Apipia:
bulk updating jira alert logic fixed to not alert on issues that have no jira config

PR #2462, opened 2020-06-02
JIRA fixes

madchap:
Some jira fixes for unreferences issues and:
- #2460
- Most of #2458 I guess, but this will be left as an exercise to @dayal18 ;-)

### DIFF
--- a/dojo/github_issue_link/urls.py
+++ b/dojo/github_issue_link/urls.py
@@ -2,7 +2,7 @@ from django.conf.urls import url
 from . import views
 
 urlpatterns = [
-    url(r'^webhook', views.webhook, name='web_hook'),
+    url(r'^github-webhook', views.webhook, name='github_web_hook'),
     url(r'^github/add', views.new_github, name='add_github'),
     url(r'^github/(?P<tid>\d+)/delete$', views.delete_github,
         name='delete_github'),
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1729,7 +1729,6 @@ class Finding(models.Model):
             return self.test.engagement.product.jira_pkey_set.all()[0].conf
         except:
             return None
-            pass
 
     def long_desc(self):
         long_desc = ''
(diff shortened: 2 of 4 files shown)


madchap:
FYI @mestrade: https://github.com/DefectDojo/django-DefectDojo/pull/2462/commits/f844036295a070151dd1157fbe42dd8dff0df1c0

Unsure how the github webhook (probably not used by anyone right now) is used further, but I *think* it may have broken the JIRA webhook. So, just FYIm the url's changed.

dayal18:
Thanks @madchap I'll try to change it locally on my machine and check this.

dayal18:
@madchap 2 issues:

1) When Product level - JIRA is enabled -- Using "Bulk Edit" is creating JIRA ticket, but the JIRA Age and JIRA Change is blank.

image


2) The Webhook didn't work.. :(  JIRA Webhook is critical functionality for our process, any other suggestions, please.

dayal18:
Rest all looks good, the bug icon, + other issues which I mentioned, those are fixed. only the above 2 are still open.

dayal18:
ok another issue.. I'm started getting those JIRA alerts again. something it didn't like..

image


I have Product level JIRA push + Push all Notes..

dayal18:
and it doesn't stop.. as I have only 1 note the speed is less.. unlike last time.. where I have 10 notes.. and 10 alerts were bumping every 10 sec

valentijnscholten:
The alert thing is fixed by #2402 in 1.6.1. The alerts are there for a reason, to give the user feedback if something wrong. So they should stay as not every user has access to the server log files.
The reason there are so many alert now is because the JIRA integration is more less completely broken by the push all issues PR in 1.6.0. So if we fix thare, the alerts will be useful again.

madchap:
> The alert thing is fixed by #2402 in 1.6.1.

Will be fixed, when out ;-)

dayal18:
> The alert thing is fixed by #2402 in 1.6.1. The alerts are there for a reason, to give the user feedback if something wrong. So they should stay as not every user has access to the server log files.
> The reason there are so many alert now is because the JIRA integration is more less completely broken by the push all issues PR in 1.6.0. So if we fix thare, the alerts will be useful again.

Ant idea if this also fixes the JIRA WebHooks? which is not working now?

madchap:
> Any idea if this also fixes the JIRA WebHooks? which is not working now?

Best to read the PR description or the few lines changed to know. It is likely that a PR fixing the webhook will somehow mention it :-)

PR #2463, opened 2020-06-02
maintainers update

madchap:


madchap:
@DefectDojo/defectdojo-moderators please have a look.

People not showing in this list normally have crossed the 6 months threshold and should have their rights removed.

There might be some not in there but somehow responsible for plugins, and we should possibly list them as such.

PR #2470, opened 2020-06-03
add known issue about finding images not working in release mode

valentijnscholten:
add known issue about finding images not working in release mode


JamesCullum:
It would make sense to add what exactly needs to be changed to get this running.

PR #2471, opened 2020-06-03
Verified Findings Check History

ssridhartibco:
This change in the utils.py file checks the history to see if the findings are already verified, in order to make sure that the history is not checked again.

### DIFF
--- a/dojo/utils.py
+++ b/dojo/utils.py
@@ -75,7 +75,7 @@ def sync_false_history(new_finding, *args, **kwargs):
     if total_findings.count() > 0:
         new_finding.false_p = True
         new_finding.active = False
-        new_finding.verified = False
+        new_finding.verified = True
         super(Finding, new_finding).save(*args, **kwargs)
 
 


valentijnscholten:
I don't understand why this is needed or what problem it is trying to solve. If it is a bugfix and should go into 1.6.1, the PR must be based against the `release/1.6.1` branch as we are using git flow from now on.

Maffooch:
The idea is if a finding is imported into DD that has been marked as false positive before, the new one will be auto set to verified to remove the need to triage it again.

valentijnscholten:
I would expect that if it is marked as false positive already the verified flag wouldn't influence anything, but I can understand it might not be working like that yet as we found out the statuses aren't 100% consisten everywhere. Do you want to have this in 1.6.1?

Maffooch:
Yes 1.6.1 would be ideal. We will get it rebased to abide by the git flow. Sorry about that!

madchap:
@ssridhartibco Changing the branch in this PR is not enough. You probably want to `git rebase --onto` on your side and force-push or something. Alternatively, close this PR and re-open against dev after rebasing. Thanks.

valentijnscholten:
Just spent 30 minutes to find out what was happening and why rebasing resulted in an empty PR. Turns out a new PR #2483 was created and already merged....

PR #2481, opened 2020-06-04
merge master into dev after release 1.6.0

valentijnscholten:
I thought we already did this, but it seems `dev` is still/again 80+ commits behind `master`. This shouldn't happen with Git Flow, but maybe we need one more merge like this to get a clean start from now on.


valentijnscholten:
doens't like right, closing

Issue #2485, opened 2020-06-04
Duplicate JIRA issues created when an ad-hoc finding is created.


kareem-DA:
**Bug description**
DDJ creates duplicate issue in JIRA when an ad-hoc finding is created.

**Steps to reproduce**
1. Setup DDJ to report issues to JIRA
2. Create a new finding significant enough to create a ticket in JIRA
4. Multiple tickets will get created in JIRA.  

**Expected behavior**
Only one ticket gets created

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [X] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: ubuntu 18.04
 - DefectDojo Commit Message: [use `git show -s --format="%h: %s [%ci]"`]

Sample log output has been added 

https://gist.github.com/kareem-DA/808799245be125df3948cfc7b4ecf4e0

I modified the code right before the add_issue_task.delay command to output a traceback. 

The add_issue_task call gets called 2 from product/views.py. Once from line 938 and again from line 964

valentijnscholten:
Which branch/commit is this on? There are some fixes in `release/1.6.1` and some PRs pending to go in there as well. There was talk about duplicates and it might have been resolved in one of those PRs.

kareem-DA:
This was on 1.6.0.   I haven't looked at the 1.6.1 release yet.

valentijnscholten:
I just checked and I think this one is not yet fixed by 1.6.1 nor any PRs. @madchap @Apipia

Maffooch:
Associated PR has been merged

PR #2489, opened 2020-06-05
Fix duplicate issues create in JIRA when creating an ad-hoc finding

kareem-DA:
This will fix issue #2485

When creating an ad-hoc finding against a project, the product/view.py calls the model.py Finding.save() twice.  The findings save then creates multiple issues in JIRA.  I got this fix from Apipia and tested it in my environment.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1766,12 +1766,6 @@ class Finding(models.Model):
             self.jira_change = timezone.now()
             if not jira_issue_exists:
                 self.jira_creation = timezone.now()
-        # If the product has "Push_all_issues" enabled,
-        # then we're pushing this to JIRA no matter what
-        if not push_to_jira:
-            # only if there is a JIRA configuration
-            push_to_jira = self.jira_conf_new() and \
-                           self.jira_conf_new().jira_pkey_set.first().push_all_issues
 
         if self.pk is None:
             # We enter here during the first call from serializers.py


valentijnscholten:
Although I can see it fixing things, I can also see potential for it to break things with all the spread out checks for push_all boolean. If you/aipipia want it go through, it should be based of `release/1.6.1`

madchap:
This should be rebased against 1.6.2 in any case. Thanks.

valentijnscholten:
needs rebasing to 1.6.5

madchap:
I can't reproduce this on my `fix-jira-bulk-issues` branch for which a WIP PR is at https://github.com/DefectDojo/django-DefectDojo/pull/2462. I only get one finding created, and no stracktrace pointing to a duplication. That branch is ahead of `1.6.0` for sure.

Could you please try it?

madchap:
So I was wondering:
* https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/product/views.py#L932 calls `save` before setting endpoints
* https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/test/views.py#L326 calls `save` before setting endpoints
* https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/finding/views.py#L659 does NOT call `save` before setting endpoints.

I vaguely remember something around that.. can someone refresh my memory? Why we would save before setting endpoints? (I rarely use them, more on SAST these days...)

Otherwise, https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/models.py#L1761 would also have to be cleaned up if this the way we want to go. Looking quickly at other `views.py` it seems ok.

I would also advise to make sure variable names like `enabled` are never used again, but get more specific such as `jira_enabled` or `jira_push_all_issues_enabled` to make the code a bit clearer.

madchap:
@kareem-DA please rebase onto 1.6.5.

Locally, you could probably run something like:

`git rebase --onto release/1.6.5 release/1.6.0 hotfix-jiradups` and `git push --force`

Then you would come on this PR, click on Edit at the top as if to change the title of your PR and modify its base.

valentijnscholten:
I think the finding needs to be in the database before endpoints can be added, so save() is needed. In the edit_finding this is not needed because the finding already exists.
But this should/could all be handled differently using something like unsaved_endpoints I have seen around. But a bit out of scope for this ticket or at least this release.

PR #2495, opened 2020-06-06
docker-compose: add REMOTE_USER header authentication support

ptrovatelli:
This makes it possible to log in using the django "remote_user" feature. Useful for some single sign on systems

Ref: https://docs.djangoproject.com/fr/2.2/howto/auth-remote-user/

By default it's disabled.

To log in, the user needs to set a configurable header with the wanted username in order to be logged in the application with that user. 
Nginx will pass-on this configurable header to uwsgi, translating it to the appropriate header that uwsgi needs. 
Obviously this should be activated only in a controlled network environment, with a reverse proxy controlling this header based on any other authentication system handled outside of defectdojo (ldap for example).
Example: 
- the reverse proxy sets x-my-header=admin when calling defectdojo's nginx
- nginx sets REMOTE_USER=admin when calling uwsgi
- uwsgi automatically authenticate the user as "admin" (no password needed). Works with any user existing in the database

Usage: 
- docker-compose.yml (or .env orverrides): 
  - set `NGINX_ACTIVATE_REMOTE_USER_AUTHENTICATION=True` so that nginx does the header translation
  - Configure `NGINX_REMOTE_USER_HEADER` to the name of the header that you want to use when entering into nginx
  - set `DD_ACTIVATE_REMOTE_USER_AUTHENTICATION=True` so that uwsgi accepts remote_user authentication

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,10 @@ services:
       - uwsgi
     environment: 
       NGINX_METRICS_ENABLED: ${NGINX_METRICS_ENABLED:-false}
+      NGINX_ACTIVATE_REMOTE_USER_AUTHENTICATION: 'False'
+      # header name in the request arriving to nginx server containing the remote user name.
+      # Low-case only and "-" replaced by "_". Starting with "x_" by convention. For "x-my-header" for example:
+      NGINX_REMOTE_USER_HEADER: 'x_my_header'
     ports:
       - target: ${DD_PORT:-8080}
         published: ${DD_PORT:-8080}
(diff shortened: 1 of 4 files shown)


valentijnscholten:
Not sure if we want to make it this easy to use REMOTE_USER headers without making sure that our nginx is only accessible from the reverse proxy in front of it. If it can be access from elsewhere, people could spoof the REMOTE_USER header. Might be better to put it in contribs or just explain in the docs how to do it?

ptrovatelli:
it's not ready for k8s; this is only for docker-compose for now (renaming the PR title)
@valentijnscholten I understand your concerns, it will make it much less usable if it's in the contributions repository though. 

how about letting this PR sit here for a while. if no-one's interested, OK to move it to the contribs repos. 

If someone needs this, please +1 here :)

PR #2496, opened 2020-06-06
JIRA fixes

madchap:
Some jira fixes for unreferences issues and: - #2460 - Most of #2458 I guess, but this will be left as an exercise to @dayal18 ;-) - Eventually fixes issue where jira webhook would not work anymore (bc of recent github integration, which might have stepped on its toes).


valentijnscholten:
duplicate PR, not needed.

PR #2506, opened 2020-06-08
fixed duplicate_finding_set function to not throw error

Apipia:
cut from release/1.6.5


valentijnscholten:
Clones from #2445, see conversation there.

Issue #2508, opened 2020-06-08
Setting a finding status to Duplicate results in errors


Apipia:
**Bug description**
I believe there should be some clarification on the intent of this possibility. It seems pretty broken right now and we should probably make some decisions about how it should work. 
Currently you can set the status of a Finding to "Duplicate" without necessarily choosing a Finding as the parent duplicate. This is done easily but editing the Finding and checking Duplicate. Once you've done this, you will create many issues, as all over the code-base and in various places, we see check like:
```
if finding.duplicate:
    finding_id = finding.duplicate_finding.id
```
Where we make an assumption that because Duplicate is True, there is a duplicate_finding (Parent finding) object associate with this finding. This is a broken assumption, since it's quite possible to set a finding as Duplicate without setting a duplicate_finding.

Currently this issue exists in dojo/templatetags/display_tags line 771. 

```
if finding.duplicate:
        url = reverse('view_finding', args=(finding.duplicate_finding.id, ))
```

A quick fix would be to change the if statement to: 
`if finding.duplicate and finding.duplicate_finding is not None:`

But I wonder if that's the appropriate solution, and we should ask ourselves: 
1. Should a user be able to set a finding to Duplicate without a Parent Finding?
2. Should a user be able to set any Finding as a parent finding? (If so, this is not possible, and how would we update the UI to allow for that?)

**Steps to reproduce**
1. Edit a finding that doesn't have any Duplicates proposed by the platform.
2. Uncheck active and verified and check Duplicate (Note that no dropdown appears to select a parent finding)
3. Save the finding
4. Attempt the view the Test page and notice the error:
```
  File "./dojo/templatetags/display_tags.py", line 772, in finding_display_status
    url = reverse('view_finding', args=(finding.duplicate_finding.id, ))
```

Apipia:
@valentijnscholten I think the `finding_display_status` function was added by you quite recently. Maybe you know a clean way to fix this? Currently giving me 500 errors on tests that I visit.

valentijnscholten:
I think the display_tag is just old code in a new place. If you replace the final part with the below, it will work better I think:

```
    if finding.duplicate:
        url = None
        name = 'unknown'
        if finding.duplicate_finding:
            url = reverse('view_finding', args=(finding.duplicate_finding.id, ))
            name = finding.duplicate_finding.title + ', ' + finding.duplicate_finding.created.strftime('%b %d, %Y, %H:%M:%S')

[...]

PR #2509, opened 2020-06-08
Fix 500 rendering error for duplicate without original_finding

Apipia:
To resolve issue #2508


valentijnscholten:
getting confused again, didn't we already fix this in 1.6.5? seems like (part of) is in #2506

madchap:
Yea, at least part of it.

@Apipia the first part will come about when we release and merge. The 2nd part is not part of #2506 at all.

PR #2514, opened 2020-06-09
k8s: Add hint about preventing gateway timeouts

philband:
This PR adds a hint to the helm `values.yaml` about how to fix Gateway Timeouts that frequently happened in our setup during dependency-scan uploads with a default nginx-ingress configuration.

Labels:
enhancement

### DIFF
--- a/helm/defectdojo/values.yaml
+++ b/helm/defectdojo/values.yaml
@@ -69,6 +69,9 @@ django:
     annotations:
       # Restricts the type of ingress controller that can interact with our chart (nginx, traefik, ...)
       #kubernetes.io/ingress.class: nginx
+      # Depending on the size and complexity of your scans, you might want to increase the default ingress timeouts if you see repeated 504 Gateway Timeouts
+      #nginx.ingress.kubernetes.io/proxy-read-timeout: "1800"
+      #nginx.ingress.kubernetes.io/proxy-send-timeout: "1800"
   nginx:
     tls:
       enabled: false


helm/defectdojo/values.yaml
```diff
@@ -69,6 +69,10 @@ django:
     annotations:
       # Restricts the type of ingress controller that can interact with our chart (nginx, traefik, ...)
       #kubernetes.io/ingress.class: nginx
+      # Depending on the size and complexity of your scans, you might want to increase the 
+      # default ingress timeouts if you see repeated 504 Gateway Timeouts
+      #nginx.ingress.kubernetes.io/proxy-read-timeout: "1800"
+      #nginx.ingress.kubernetes.io/proxy-send-timeout: "1800"
```
Maffooch, line 75: Please remove one of the lines.

philband: Done

Maffooch:
Does not look like the new commit got posted to this branch?

philband:
> Does not look like the new commit got posted to this branch?

I squashed and force-pushed it. Maybe you meant for me to make a different change than the one I made? I removed the wrap in the comment I added, reducing the comment to one line.

Maffooch:
Ahh yes that what it is! I guess I read too quickly because I read the two following lines as identical.
```
#nginx.ingress.kubernetes.io/proxy-read-timeout: "1800"
#nginx.ingress.kubernetes.io/proxy-send-timeout: "1800"
```

Approving

Issue #2525, opened 2020-06-11
Reopening of findings marked as false positive


wurstbrot:
**Bug description**
Defectdojo creates new alerts when it detects a duplicate which is marked as false positve

**Steps to reproduce**
* Turn on deduplication
* Upload Dependency Check report with deduplicaiton on engagement
* Mark finding as false positive, inactive, not verified
* Upload Dependency Check report with deduplicaiton on engagement
-> A new finding without deduplication is created.

In the old finding there is a message "This finding has been automatically re-opend as it was found in recent scans".

**Expected behavior**
The finding should stay inactive.

**Deployment method** *(select with an `X`)*
- [x] Kubernetes
- [x] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - DefectDojo 1.6.2
This bug exists since the update from 1.5.4.1 to 1.6.2

**Sample scan files** (optional)
If applicable, add sample scan files to help reproduce your problem.

**Screenshots** (optional)

**Console logs** (optional)
If applicable, add console logs to help explain your problem.

**Additional context** (optional)

wurstbrot:
I assume the corresponding line is https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/utils.py#L299

By marking it as FP, "mitigated" becomes 
defectdojo=> select mitigated from dojo_finding where id=452933;
           mitigated           
-------------------------------
 2020-06-11 06:22:07.058072+00
(1 row)

I assume the statements needs to be:
if (existing_finding.is_Mitigated or existing_finding.mitigated) and not existing_finding.false_p and new_finding.active and not new_finding.is_Mitigated:

wurstbrot:
Here is a video showing that a finding will get active again (even with false positive history enabled):
https://drive.google.com/file/d/1ezpA7Qk8xiAhh_sROeiG1EKWkxa4pVYA/view?usp=sharing

PR #2526, opened 2020-06-11
hotfix: do not reopen false positives and out_of_scope

wurstbrot:
see https://github.com/DefectDojo/django-DefectDojo/issues/2525

I have tested the fix locally.

From my point of view, this needs to be a hotfix and 1.6.3. should be released. I will wait one day and in case there is not release, I will build and deploy my own dd version with this fix (this is something I would not expect for a bug like this).

### DIFF
--- a/dojo/utils.py
+++ b/dojo/utils.py
@@ -291,7 +291,7 @@ def set_duplicate(new_finding, existing_finding):
     if existing_finding.id == new_finding.id:
         raise Exception("Can not add duplicate to itself")
     deduplicationLogger.debug('New finding ' + str(new_finding.id) + ' is a duplicate of existing finding ' + str(existing_finding.id))
-    if (existing_finding.is_Mitigated or existing_finding.mitigated) and new_finding.active and not new_finding.is_Mitigated:
+    if (existing_finding.is_Mitigated or existing_finding.mitigated) and not existing_finding.out_of_scope and not existing_finding.false_p and new_finding.active and not new_finding.is_Mitigated:
         existing_finding.mitigated = new_finding.mitigated
         existing_finding.is_Mitigated = new_finding.is_Mitigated
         existing_finding.active = new_finding.active


valentijnscholten:
This is a good idea, but what about findings that are `out_of_scope`? or `duplicate`?

wurstbrot:
@valentijnscholten what do you mean by _duplicate_?

valentijnscholten:
An issue can be marked as duplicate of another issue if DD recognizes it is the same issue based on a hash_code field, which is calculated over a subset of fields.

wurstbrot:
We are in the context of a duplicate, already (method set_duplicate).
Therefore, I do not see the need to explicitly add duplicate handling.

valentijnscholten:
I am just thinking out loud. With the current PR it can still happen that if someone imports a report with a finding that was previously marked as duplicate, it will reopen it (active becomes true).
Hmm I now see that you are not editing the reimport code but the set_duplicate code. I think if we make a change it has to be consistent everywhere, so also in the reimport code in serializers.py and test/views.py.
There is a feature called "False Positive History" that I think is supposed to handle this even before it arrives in the deduplication code, but it looks like it's not working correctly.
I now some users are having defect dojo instances with 100k+ findings and are relying on deduplication heavily, so I am hesitant to make changes deep in these core methods of deduplication too quickly in a bugfix release. We really need to have testcases to move forward with deduplication logic changes/fixes.

wurstbrot:
I have never used "false positive history" before. I will try it out. Maybe we just need to document an upgrade path from 1.5 to 1.6 which highlights that you should tick that box in case deduplication is used and it is expected.

wurstbrot:
That box is ticked, already :(
I will try it out locally again (there I did not tick it). But the likelihood that it is not an issue is very low.

wurstbrot:
The finding is still reopened as active. I alert on active findings, so my developers are getting an information on each scan, even when they marked it.

I created a video and linked in https://github.com/DefectDojo/django-DefectDojo/issues/2525 .

alles-klar:
Makes sense for me. But we should create a test case for this scenario.

wurstbrot:
Maybe this one is the root cause for opening it as active, verified: https://github.com/DefectDojo/django-DefectDojo/pull/2483

madchap:
This now would need to be rebased against `1.7.0`

valentijnscholten:
> This now would need to be rebased against `1.7.0`

Against `dev` that is

madchap:
> > This now would need to be rebased against `1.7.0`
> 
> Against `dev` that is

Right, thought 1.7.0 was already created.

wurstbrot:
Wanted to create tests, but dev is not working (for me)
https://github.com/DefectDojo/django-DefectDojo/issues/2590

wurstbrot:
see https://github.com/DefectDojo/django-DefectDojo/pull/2599

PR #2528, opened 2020-06-11
 Check the size of the scan report when is imported through API

### LINKED ISSUE #2346: Limit on report size that can be imported

ccojocar:
Add a check in the scan report import and re-import APIs which verifies iif the scan report size is below a pre-defined maximum size when the data is validated by the serializer.

fixes #2346 

cc @madchap 

Signed-off-by: Cosmin Cojocar <cosmin.cojocar@gmx.ch>

- enhancement
- performance
- feature
- bugfix
- maintenance (a.k.a chores)
- dependencies
- New Migration

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -7,7 +7,7 @@ from dojo.models import Product, Engagement, Test, Finding, \
 from dojo.forms import ImportScanForm, SEVERITY_CHOICES
 from dojo.tools import requires_file
 from dojo.tools.factory import import_parser_factory
-from dojo.utils import max_safe
+from dojo.utils import max_safe, is_scan_file_too_large
 from dojo.notifications.helper import create_notification
 from django.urls import reverse
 from tagging.models import Tag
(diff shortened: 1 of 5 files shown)


madchap:
Relates to https://github.com/DefectDojo/django-DefectDojo/pull/2350

ccojocar:
@valentijnscholten @madchap I pushed a change which checks also the size of the scan file when is uploaded through UI.

ccojocar:
> Actually, this needs to be added at the Test level on the UI side. It would need to occur on re-import as well. Sorry for the premature approval!

@Maffooch I added the check also in the tests where the report gets re-imporetd. Is it complete now? Thanks

ptrovatelli:
i'm late for the party, but would be very nice to have a "DD_SCAN_FILE_MAX_SIZE" env variable too; 
just need to 
- add it at the begining of the settings.dist.py, putting the default value there
```
env = environ.Env(
    ...
    DD_SCAN_FILE_MAX_SIZE=(int, 100),
...
```

- and then create SCAN_FILE_MAX_SIZE based on DD_SCAN_FILE_MAX_SIZE like it's done for most of the other variables

- update docker-compose config. 
- for kubernetes someone added a way to use any extra variable recently so nothing more is needed

- for setup.bash it's EOLing, if someone needs it they can always add it. 

otherwise ppl may get completely stuck because of this.

ccojocar:
It makes sense. I would create a follow up issue for this. What do you think?

valentijnscholten:
yes, but you could use the same branch (I think).

ibcoleman:
Hey @ptrovatelli 

> otherwise ppl may get completely stuck because of this.

Can confirm: I am completely stuck because of this.

Issue #2531, opened 2020-06-11
Re-Import Scan Result Feature for SonarQube API Scan fails due to missing file


zorianix:
Re-Import Scan feature request for a file upload for Sonarqube API Scan types this cause the request to fail with a 500 error as there is no file uploaded . A validation is required to check whether the type of Scan Import was Sonarqube API Scan , in that case it should allow for re-generation of api call to sonarqube server

1. Go to Import Scan Results
2. Select Scan Type to: Sonarqube Api Import and add complete Import Process
3. Select the Test again and choose Re-Import Scan Result, now if the user blindly clicks on the import this result in a Server 500 error as no file is uploaded in the Re-Import process

Expected behavior
An API request should be re-generated for the re-import request

**Deployment method** *(select with an `X`)*
- [X] Kubernetes
- [X] Docker
- [X] setup.bash / legacy-setup.bash

Maffooch:
Fixed in #2606

PR #2532, opened 2020-06-11
Fix for Re-Import for SonarQube Api Result Failure

zorianix:
The suggested fix for #2531 . Added code to validate if the scan type for reimport is sonarqube api, if yes there will be no file present hence  passing scan_type to the function call

### DIFF
--- a/dojo/test/views.py
+++ b/dojo/test/views.py
@@ -628,20 +628,34 @@ def re_import_scan_results(request, tid):
             tags = request.POST.getlist('tags')
             ts = ", ".join(tags)
             test.tags = ts
-            try:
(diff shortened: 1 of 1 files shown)


Maffooch:
There are some indenting problems here.

Maffooch:
There are some Flake8 violations there as well. 

Also please submit this fix the release/1.6.5 branch as it is a bug fix people would likely want quickly. This PR will have to be closed and remade with the original base set to 1.6.5. Otherwise, the commit history gets super dirty, and we will not merge it. 

Ask for help on Slack if you run into any issues. Thanks.

PR #2547, opened 2020-06-12
travis: update components

valentijnscholten:
travis, update:
- k8s
- helm
- minikube

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ services:
   - docker
 env:
   global:
-    - K8S_VERSION=v1.13.4
-    - MINIKUBE_VERSION=v0.35.0
-    - HELM_VERSION=v2.13.0
+    - K8S_VERSION=v1.18.0
+    - MINIKUBE_VERSION=v1.11.0
+    - HELM_VERSION=v2.16.8
     - CHANGE_MINIKUBE_NONE_USER=true
   matrix:
     - TEST=flake8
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,9 +4,9 @@ services:
   - docker
 env:
   global:
-    - K8S_VERSION=v1.13.4
-    - MINIKUBE_VERSION=v0.35.0
-    - HELM_VERSION=v2.13.0
+    - K8S_VERSION=v1.18.0
```
madchap, line 7: Maybe try with a least aggressive update? FYI, Google GKE stable is on 1.14'ish. Maybe 1.16.11?

.travis.yml
```diff
@@ -4,9 +4,9 @@ services:
...
+    - MINIKUBE_VERSION=v1.11.0
```
madchap, line 8: Same here.. there is sooo many versions in between, it's scary :fearful: 
1.0 was the next version after 0.35.0.

I can see in their changelog that 1.4.0 was the first release to support k8s 1.16.0 and that maybe v1.5.2 could be a good pick? Who knows. Wanna try?

valentijnscholten:
I thought this was just a quick update, but apparently not. Maybe someone who actually knows k8s can take a look, or we just leave it as is.

.travis.yml
valentijnscholten: I think we need to be mooaaaaarrrr aggressive ;-)

.travis.yml
valentijnscholten: Let's go 1.12!!!!!

valentijnscholten:
Superseeded by #2826

PR #2550, opened 2020-06-13
add required middleware to set actor_id in auditlog entries

valentijnscholten:
I thought we already did this one, but looks this middleware is not included in `settings.dist.py`
It's needed to populate the `author` field for auditlog entries. Currently they're all empty (None).


Maffooch:
Shows no files changed?

valentijnscholten:
haha, that show it was in fact done at some point! No idea, must have been fat fingering somewhere or comparing to an old commit on dev.

Issue #1661, opened 2019-11-20
Integrate SAML2 into DefectDojo.


piyarathnalakmali:
**Is your feature request related to a problem? Please describe**

WSO2 uses DefectDojo as the Vulnerability Management System and we need to do the authentication using a custom Identity provider. Our IDP supports both SAML and JWT s. Between those two options SAML is the easiest to implement in DefectDojo. I found [this PR](https://github.com/DefectDojo/django-DefectDojo/pull/1363) which includes **SAML integration for DefectDojo**. By applying the proposed changes manually, I was able to successfully integrate SAML into DefectDojo. But, when we want to migrate the VMS to new DefectDojo versions we will have to repeat this configuration again and again. The same problem will be applicable to other co-operate DefectDojo users as well.

**Describe the solution you'd like**
It would be grate if SAML is integrated into DefectDojo by default.

dougmcdorman:
At first glance it looks like the social auth https://python-social-auth.readthedocs.io/en/latest/backends/saml.html?highlight=saml is a general SAML implementation and so could probably be used by your corporate SAML implementation with DefectDojo. Social auth is mentioned in the referenced PR.
I'm not clear yet if DD just needs configuration to do social auth or actual code changes.

alles-klar:
@piyarathnalakmali I am using the implementation from https://github.com/DefectDojo/django-DefectDojo/pull/1363 and it works very well for me. I had some problems to configure saml with social auth, so I decided to stick to the version from the PR because it was already running.

PR #1363, opened 2019-07-10
SAML integration

alles-klar:
Integrate support for SAML

These PR adds all required dependencies to integrate SAML. There is already a example configuration in the settings.dist.py. I do not provide additional documentation, because the djangosaml2 documentation describes already the relevant parameters. Is this sufficient?

In my setup it works fine but I don't know how to provide good tests for this SAML integration. If someone has a good idea feel free to leave a comment.

Related to PR #1141 which was closed during the python3 shift.

---

**Note: DefectDojo is now on Python3 and Django 2.2.1 Please submit your pull requests to the 'dev' branch as the 'legacy-python2.7' branch is only for bug fixes. Any new features submitted to the legacy branch will be ignored and closed.**

When submitting a pull request, please make sure you have completed the following checklist:

- [ ] Your code is flake8 compliant 
- [ ] If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR.
- [ ] Model changes must include the necessary migrations in the dojo/dd_migrations folder.
- [ ] Add applicable tests to the unit tests.

PR #2573, opened 2020-06-17
Add SAML2 support to DefectDojo

cody-m-tibco:
Adds a plugin that makes SAML2 authentication setup very similar to the OAuth suite currently in place. 

[Documentation](https://github.com/DefectDojo/Documentation/pull/102)


alles-klar:
Seem to be the same discussion as in this PR https://github.com/DefectDojo/django-DefectDojo/pull/1363 and this issue https://github.com/DefectDojo/django-DefectDojo/issues/1661. 

From my point of view saml is very useful and I am using the implementation from https://github.com/DefectDojo/django-DefectDojo/pull/1363 now over one year in production.

cody-m-tibco:
Oh I remember that PR now. Yes, SAML is super useful from an org perspective since it is a little more secure than OAuth. Without any bias, I think this package makes it a little more simple and user friendly at the expense of customization. What do you think?

alles-klar:
@cody-m-tibco did you took a look if saml would work with [social auth](https://python-social-auth.readthedocs.io/en/latest/backends/saml.html?highlight=saml)? To be honest, I didn't because I already had a solution which worked for me.

cody-m-tibco:
I did not! I was under the impression social-auth was primarily OAuth. Probably best to keep SSO configs consistent. Thanks for steering me in the right direction!

cody-m-tibco:
So after trying off and on for a bit, its going to take 3 extra debian packages to get the social-auth SAML backend working. Seems like the one from this PR would be the lightest option.

PR #2576, opened 2020-06-18
Bugfix for burp test upload

JamesCullum:
The Burp upload has been broken in two places:

1. Base64 can't be displayed because the base64 package is not imported
2. If there are multiple collaboratorEvent, usually DNS+HTTP, HTTP will be detected as DNS and it crashes

This development is sponsored by Panasonic Information Systems Company Europe.


valentijnscholten:
Thanks for the PR. Do you have 1 (or a couple) of example files that were failing and are now working again with this PR? If we update the unittests to include those example(s) it will help make the parser stable, prevent regressions, etc.

JamesCullum:
Sure - attached you find a zip containing the XML that I anonymized a bit.

[report.zip](https://github.com/DefectDojo/django-DefectDojo/files/4797938/report.zip)

valentijnscholten:
Thanks, what I meant is update the unittests to use that XML. Turns out we don't have a unit test yet for the burp parser, would you be willing to create something along the lines of https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/unittests/test_npm_audit_scan_parser.py

JamesCullum:
Sure, let me give it a shot.

valentijnscholten:
Thanks. Travis found some issues: https://api.travis-ci.org/v3/job/700000328/log.txt and some flake findings https://travis-ci.org/github/DefectDojo/django-DefectDojo/jobs/700000326

JamesCullum:
@valentijnscholten Works now, sorry for the increments - had to test on Travis as DD doesn't work on Windows

valentijnscholten:
np, we're messing up stuff all the time ;-)
I am running DD on WSL, so that might be an option for you. MySQL runs on Windows.

PR #2581, opened 2020-06-18
Notes improvements (UI and Jira integration)

steeve85:
This PR improves few things related to notes:
- Private notes are not exported to Jira (change discussed on Slack)
- Jira configuration name is added to dropdown in product settings. This adds more context to the Jira configuration selection when a single Jira user is used across several Jira boards/configurations
- Jira notifications in DD are updated to use the Jira configuration name instead of the Jira URL. Using the Jira URL is confusing if we have several configuration using the same Jira host.

- enhancement


valentijnscholten:
When reviewing this I noticed the existing code allows for up 2000 characters as configuration name, username or password. very efficient ;-)

PR #2587, opened 2020-06-19
bugfix: remove false positiv history

alles-klar:
Remove code about false positiv history.

Currently there is a problematic false positive function: If there is a finding flagged as fp in the engangement, all other findings with the same cwe or title will also marked as fp. Since different findings can have the same cwe (i.e. [CWE-1035](https://cwe.mitre.org/data/definitions/1035.html)), it is problematic. 

I suggest, to delete the code about false positive history because it functionality is covered by deduplication.

On behalf of DB Systel GmbH


valentijnscholten:
Not everyone is using deduplication, I believe this false positive history is in fact an alternative for deduplication. I am not saying it is perfect, but removal might be too drastic before we know if someone is using. If you are using deduplication, you should turn off false positive history in system settings.

madchap:
Maybe the help-text bubble could be made more informative on the system settings page at least, as the "workaround" is to basically not enable FP?

The same for deduplication actually, especially after the advent of the per-parser algo.

alles-klar:
I wasn't aware that it can be disabled in the setting. Thanks for that hint, for me that is ok as a workaround but maybe we should think about an end of life of this feature? It is very similar to the deduplication and has the same problems.

madchap:
> I wasn't aware that it can be disabled in the setting. Thanks for that hint. For me that is ok as a workaround but maybe we should think about an end of life of this feature? It is very similar to the deduplication and has the same problems.

Yea, I like the EOL idea in the spirit of not surprising users that could be using it.

With this in mind, should we close this PR and have another that adds the bubble hint + doc update?

PR #2595, opened 2020-06-22
doc: add docker-compose build

wurstbrot:
add docker-compose build to docker readme


Maffooch:
I don't think this is really necessary due to the 'Building images' section that is a little above the 'Run with Docker compose in development mode with hot-reloading' section.

madchap:
> I don't think this is really necessary due to the 'Building images' section that is a little above the 'Run with Docker compose in development mode with hot-reloading' section.

Maybe.. but I'd rather see one time too much than not enough :-)

PR #2603, opened 2020-06-22
Add SAML2 support

cody-m-tibco:
Copy of #2573 due to hastily deletion of old branch.

Adds a plugin that makes SAML2 authentication setup very similar to the OAuth suite currently in place.

[Documentation](https://github.com/DefectDojo/Documentation/pull/102)


dsever:
@cody-m-tibco, for k8s it is bad that we need to change setting.py, can we have it in the form of environment variables like is there for other type of  authentications?

cody-m-tibco:
It's not ideal from the plugin perspective, but it can be done.

dsever:
@cody-m-tibco manage to use your code and make it work, also i added config param over te env, please take a look:

https://github.com/dsever/django-DefectDojo/blob/defect_dojo_saml_integrated/dojo/settings/settings.dist.py#L350

cody-m-tibco:
Looks good. I arrived at a similar place.

Issue #1975, opened 2020-03-04
How to work with multiple URL's in Same Issue (scanner import)


gautammayekar:
So, I have a question here with regards to multiple URL's reported in scanners for a same issue. 

For Eg. An IBM Appscan report says XSS vulnerabilities found and it has around 10 URL's out of which 4 are false positives. How to tackle this problem. 

I only get an option to tick the whole XSS issue as False positive. I would like to have few URL's / Parameters flagged as false positives and few as not.

gautammayekar:
Hi, 

Any update on this?

madchap:
This is currently not supported, and would probably require more than just a parser update.

gautammayekar:
Hi,

Thanks for the response. Any plans to support this feature in future?

gautammayekar:
Hello, any progress and discussions on this? 
Kindly let me know,

Thanks.

cody-m-tibco:
Hello @gautammayekar, there is a notion of remediated endpoints that allow a user to mark a specific endpoint as remediated or still vulnerable. After remediation, it will be moved to another dropdown box in the finding view. Though it does not mark the endpoint as any of the status' a finding could have, it may be of good use to you. Currently, the remediation only applies to the host, but it should apply to the path instead. I can fix that for you. What do you think?

devGregA:
@cody-m-tibco please reference this issue in your PR when ready.

Maffooch:
Fixed

PR #2632, opened 2020-07-01
Partial endpoint remediation

cody-m-tibco:
Add the ability to partially remediate endpoints. Before, an endpoint remediated on one finding would be remediated for every single finding the endpoint was attached to. Now, each endpoint now has its own status and activity logging so that each endpoint on a given finding can be manipulated, but not carried over to other vulnerabilities. An endpoint status is created when an endpoint is attached to a finding in all avenues (API, UI, Import, re-import). 

Motivated by #1975 

- [] If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR.


dojo/templates/dojo/custom_pdf_report_endpoint_list.html
```diff
@@ -112,17 +112,23 @@ <h6>Vulnerable Endpoints / Systems</h6>
...
-                                                {% for row in grouped_items %}
-                                                    <tr>
-                                                        {% for item in row %}
-                                                            <td>{% if item %}<i class="fa fa-circle"></i>
-                                                                {{ item }}{% endif %}
-                                                            </td>
-                                                        {% endfor %}
-                                                    </tr>
-                                                {% endfor %}
+                                            <table id="vuln_endpoints" class="table-striped table ">
```
valentijnscholten, line 115: is this a typo?

valentijnscholten: don't know what happened, but I selected a different line/section. There's an occurrence of 'ndpoints' in there somewhere.

cody-m-tibco: I thought you meant the whole block for second! Yes I found it. Good eyes. Holding off on committing in case you find something else.

valentijnscholten:
Nice work, good to see some improvements towards better endpoint management. I would like to also make the dedupe a little better around them if anybody ever has a need for that.
Some things popping up in my head while looking at the code:
- should we stick to 'mitigated' instead of 'remediated', same as for findings? Or maybe change mitigated status for findings to remediated if that sounds better? :-)
- did you consider using a m2m relationship with a "through" table? I think that's how Django ORM envisions these m2m relationships with extra properties
- should we allow authorized users in the UI to change the status of endpoints now that we have our quickwin authorized user permission pr merged? (2615)
- i see a bit repetitive html snippets around display a table of endpoints + statuses. are those suited to be put into a "snippet" html template that is included?
- did you consider any impact on dedupe? it looks like there is none?

cody-m-tibco:
I will update the marked boxes:

- [x] Mitigation probably makes a bit more sense from a definition perspective. Easy fix there.
- [ ] Have not heard of the the "through" attribute. Looks to accomplish what I did but in a much cleaner and simpler way. May be a lot of work converting due to templatetag/html.
- [x] Yes, these should be permissioned the same way bulk findings are. Will update.
- [x] It is a bit repetitive and not very maintenance friendly. A snippet would be ideal. Great suggestion!
- [ ] No impact as it was intentionally avoided. Considered the impact of dedupe, adjusting to the "new way of doing things", and converting from old system to the new. Surely there are angles I was not considering, so I decided to wait for the mod meeting for discussion.

cody-m-tibco:
- AccessLint passed, but requests changes on things I did earlier? Resolving.
- DB changes to use 'through' tag turned out to be super complicated when switching on an established DB. It either requires multiple migrations or data loss. Keeping as is. 
- Made some changes to permissions suite. Noticed some funky stuff that slipped past me when I merged earlier.

valentijnscholten:
Another thing I thought of:
- When you run re_import or normal import with close_old_findings=True, DD will close (mitigated) any findings that are no longer present in the scan report being imported. Should something similar happen for endpoints now?

dojo/templates/dojo/snippets/endpoints.html
```diff
@@ -0,0 +1,240 @@
...
+                                        class="glyphicon glyphicon-chevron-up"></i></a></span>
+                            </h4>
+                        </div>
+                       
+                        <table id="vuln_endpoints" class="table-striped table table-hover">
+                            <thead>
+                                {% if user|is_authorized_for_change:finding %}
+                                    <th class="hidden-sm" title="Select all vulnerable endpoints." style="width: 10%;">
+                                        <form class="inline-form" action="#">
+                                            <input type="checkbox" label="select_all_vulnerable" name="select_all_vulnerable" id="select_all_vulnerable"/>
```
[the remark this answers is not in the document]
cody-m-tibco: This is just not true

PR #2633, opened 2020-07-01
fix: rename HuskyCi to HuskyCI Report in test_type

edersonbrilhante:
This PR will fix test_type of HuskyCI (rename HuskyCi to HuskyCI Report)

Label: bugfix


edersonbrilhante:
Hi @Maffooch Thank you for your comment. I got it :), but I felt I should fix my mistake :P

PR #2641, opened 2020-07-03
Fix generic parser

Nilix007:
Previously, the parser looked for the active and verified columns twice:
First, by looking for such column names and, secondly, by checking
column number 9 and 10 (which are fixed).  This PR removes the
latter part.


valentijnscholten:
Thanks for the PR @Nilix007 . Are you in the position to add some small unit tests like the other parsers have? Would make it easier for us to trust the PR also ;-)

Nilix007:
Uh, lost track of this PR.  Rebased to dev 👍

PR #2643, opened 2020-07-04
JIRA/GitHub: Use templates to allow customizing issue description / body

### LINKED ISSUE #2642: jira config: don't ask password when editing some fields
Currently when you want to edit a JIRA config, you have to enter the password for the JIRA instance again. This is typically stored in some vault somewhere. Might be better not to make it mandatory? If empty -> don't make any changes to the password, only to the other fields.

### LINKED ISSUE #2655: view finding: JIRA/Github link is only shown for static findings
In #2630 we simplified the display of the link to JIRA/github on the view_finding page. Turns out it is now only shows for static findings and not for dynamic findings.

valentijnscholten:
based of 2650.
- Fix jira/github issue displaying for dynamic findings (fixes #2655)
- Improves the description that is sent to JIRA/GitHub issues;
- Allow customization of that description via django templates;
- Improve JIRA config list and editing (fixes #2642)
~~This requires a migration to switch reporter from `User` model to `Dojo_User` model~~
~~I decided to change this everywhere to be done without with 1 migration~~
- Fix incorrect logging in travis integration tests script


valentijnscholten:
Had to remove the migration part as it's conflicting with how django internally works with request.user. Will need a bigger effort if we want to swap from `User` to `Dojo_User` everywhere.

madchap:
Got an error upon pushing to jira, from a fresh instance

```
Traceback:

File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py" in inner
  34.             response = get_response(request)

File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  115.                 response = self.process_exception_by_middleware(e, request)

File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  113.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "./dojo/user/helper.py" in _wrapped
  79.         return view_func(request, lookup_value, *args, **kwargs)

File "./dojo/finding/views.py" in edit_finding
  690.             new_finding.save(push_to_jira=push_to_jira)

File "./dojo/models.py" in save
  1927.                 if self.reporter.usercontactinfo.block_execution:

File "/usr/local/lib/python3.6/site-packages/django/db/models/fields/related_descriptors.py" in __get__
  415.                     self.related.get_accessor_name()

Exception Type: RelatedObjectDoesNotExist at /finding/1/edit
Exception Value: User has no usercontactinfo.
```

valentijnscholten:
I think you will also be getting this on `dev` then? I don't understand how you keep ending up with users without usercontactinfo. Should be created automatically by the initializer since a couple of months.
The problem is we can't add any methods to `request.user` to make it better. But I guess if we keep running into these issues, I just have to make some utility method that always works.

PR #2630, opened 2020-07-01
JIRA/github improvements and fixes July 2020

valentijnscholten:
bugfixes:
- bulk edit with push to github: create github issue instead of JIRA issue

improvements:
- set `duedate` field based on SLA targets when creating JIRA issue and if SLA is enabled in DD
- display jira issue + url in  edit_finding
- display github issue + url in  edit_finding (fixes #2607)
- simplify displaying of jira_issue and github_issue in view_finding
- send full url to DD finding to JIRA / GitHub instead of only finding id
- send labels to JIRA in the same call as creating/updating an issue

maintenance
- if a user has set `block_execution` set to True in their profile, the JIRA/GitHub actions will now run in uwsgi and not in celery. (similar to what happens with dedupe, notifications etc)
- simplify code around accessing jira/github issue and configuration in view/edit finding.

image
image

PR #2650, opened 2020-07-06
jira: only set duedate when field is available in jira

valentijnscholten:
Setting duedate field based on SLA in #2630 breaks JIRA integration if the `duedate` field is not available for that JIRA project + issuetype.
This PR retrieves metadata from JIRA and only sets `duedate` field when possible.

Now that we have the metadata anyway, we also set the environment field to contain the list of endpoints for the finding if possible.

### DIFF
--- a/dojo/utils.py
+++ b/dojo/utils.py
@@ -1394,14 +1394,12 @@ def add_issue(find, push_to_jira):
                     # meta = jira.createmeta(projectKeys=jpkey.project_key, expand="fields")
                     if not meta:
                         meta = jira_meta(jira, jpkey)
-                    # print('VALENTIJN:', meta['projects'][0]['issuetypes'][0]['fields']['summary'])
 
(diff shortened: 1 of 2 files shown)


dojo/utils.py
```diff
@@ -1388,13 +1390,28 @@ def add_issue(find, push_to_jira):
...
+                        # jira wants YYYY-MM-DD
+                        duedate = find.sla_deadline().strftime('%Y-%m-%d')
+                        fields['duedate'] = duedate
+
+                if len(find.endpoints.all()) > 0:
+                    if not meta:
+                        meta = jira_meta(jira, jpkey)
+
+                    if 'environment' in meta['projects'][0]['issuetypes'][0]['fields']:
+
```
madchap, line 1410: extra blank line :P

valentijnscholten: free for you

PR #2662, opened 2020-07-09
uwsgi: add variable to configure the buffer-size

xens:
This commit adds a variable DD_UWSGI_BUFFER_SIZE to set the USWGI buffer-size, so it'll then be possible to set a custom buffer-size as the default one is very small (4096 Bytes) and can cause issue with large cookies.

### DIFF
--- a/docker/entrypoint-uwsgi-dev.sh
+++ b/docker/entrypoint-uwsgi-dev.sh
@@ -16,4 +16,5 @@ exec uwsgi \
   --enable-threads \
   --processes 2 \
   --threads 2 \
-  --py-autoreload 1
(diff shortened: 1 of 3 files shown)


madchap:
Maybe we'd want it to be in the other uwsgi entrypoint files to remain consistent?

xens:
> Maybe we'd want it to be in the other uwsgi entrypoint files to remain consistent?

Absolutely ! It's now done for all the uwsgi entrypoints.

valentijnscholten:
@xens could you recreate the PR against `dev` as per instructions in the PR template?

PR #2667, opened 2020-07-10
Default jira severity to Low instead of None

### LINKED ISSUE #2629: jira minimum severity not working when left default
On a fresh instance the jira minimum severity is not set in system settings. This seems to break JIRA integration alltogether as all I'm seeing in the celery worker logs  is 
```
Finding 153 is below the minimum jira severity threshold
The JIRA issue will NOT be created.
```
And that finding is Medium severity. Settings threshold to Info in system settings fixes it.

madchap:
Fixes #2629


valentijnscholten:
If we want to fix existing installs, we could add a data migration as well.

dojo/db_migrations/0047_jira_minimum_severity_default.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 2.2.14 on 2020-07-10 08:53
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    def set_default_to_low(apps, schema_editor):
+        system_settings = apps.get_model('dojo', 'System_Settings')
```
valentijnscholten, line 9: does this work? I would expect something like 
```
System_Settings = apps.get_model('dojo', 'System_Settings')
ss = System_Settings.objects.get(0)
...
```

But maybe the thing in the PR also works?

madchap:
Alright, this seems to all work as expected now.

dojo/db_migrations/0047_jira_minimum_severity_default.py
```diff
@@ -0,0 +1,33 @@
...
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    def set_default_to_low(apps, schema_editor):
+        system_settings = apps.get_model('dojo', 'system_settings')
+        try:
+            ss = system_settings.objects.get(id=1)
```
valentijnscholten, line 11: I think it should be just `System_Settings.objects.get()` like everywhere else, or does this not work?

dojo/db_migrations/0047_jira_minimum_severity_default.py
```diff
@@ -0,0 +1,33 @@
...
+            jira_sev_value = ss.jira_minimum_severity
+            if jira_sev_value is None:
+                ss.jira_minimum_severity = 'Low'
+                ss.save()
+        except Exception as e:
+            # probably a test run such as running unittest, no values in table
+            print("No data migration because of exception caught: {}".format(e))
```
valentijnscholten, line 18: I think changing existing installs is a best effort anyway so suggest to not log anything in this case as it may confuse users?

dojo/db_migrations/0047_jira_minimum_severity_default.py
valentijnscholten: I don't think it's guaranteed to have id=1 for system settings.

madchap: In what instance could there be a 2nd row for the global settings?

dojo/db_migrations/0047_jira_minimum_severity_default.py
madchap: Makes sense. Will amend.

dojo/db_migrations/0047_jira_minimum_severity_default.py
madchap: just `.get()` didn't work IIRC, hence the `id=1`. I know it's used elsewhere, but it didn't like it in that case.

valentijnscholten: there shouldn't be more then 1 row indeed, but you won't know for sure the id is going to be 1. someone might have changed settings along the way, ran manual queries. maybe has some migrations of his/her own.

.get() should work. Or .objects.all().first() or something.

PR #2668, opened 2020-07-10
SLA notifications

madchap:
This PR adds functionality to receive SLA pre-breach/breach notifications, based on the SLA settings defined.

2 modes:
- A celery default `crontab` is added (arbitrary)
- A manage.py command is introduced

The CLI invocation looks like:
```
$ docker exec -ti madchap_ddojo_uwsgi_1 bash -c 'python3 manage.py sla_notifications'
```

The regular notification system is used, with the addition of a new "type" -- `sla_breach`.

image

You can see it at the Product notifications level:

image

Or in the System/Personal notifications screen:

image

There is a caveat whenever a finding is linked to a JIRA issue. A comment can be posted to the JIRA issue then, and this is opt-in in the JIRA configuration settings, at the bottom:

image

results in something like:

image

Some settings changed to drive the notification behavior:
```
# Setting SLA_NOTIFY_ACTIVE and SLA_NOTIFY_ACTIVE_VERIFIED to False will disable the feature
SLA_NOTIFY_ACTIVE = False  # this will include 'verified' findings as well.
SLA_NOTIFY_ACTIVE_VERIFIED_ONLY = True                         
SLA_NOTIFY_WITH_JIRA_ONLY = False  # will only process findings that have a JIRA associated
SLA_NOTIFY_PRE_BREACH = 3  # in days, notification until date of breach. 
```

Enhancements can certainly be made, but getting some basic SLA notifications in place seems like the right thing to do.


dojo/settings/settings.dist.py
```diff
@@ -375,6 +376,13 @@ def generate_url(scheme, double_slashes, user, password, host, port, path):
 AUTHORIZED_USERS_ALLOW_CHANGE = env('DD_AUTHORIZED_USERS_ALLOW_CHANGE')
 AUTHORIZED_USERS_ALLOW_DELETE = env('DD_AUTHORIZED_USERS_ALLOW_DELETE')
 
+# Setting SLA_NOTIFY_ACTIVE and SLA_NOTIFY_ACTIVE_VERIFIED to False will disable the feature
+SLA_NOTIFY_ACTIVE = False
+SLA_NOTIFY_ACTIVE_VERIFIED = True
```
ccojocar, line 381: It would be great to add some comments to these settings keys.

dojo/settings/settings.dist.py
```diff
@@ -625,6 +633,10 @@ def generate_url(scheme, double_slashes, user, password, host, port, path):
     'update-findings-from-source-issues': {
         'task': 'dojo.tasks.async_update_findings_from_source_issues',
         'schedule': timedelta(hours=3),
+    },
+    'compute-sla-age-and-notify': {
+        'task': 'dojo.tasks.async_sla_compute_and_notify',
+        'schedule': crontab(hour=7, minute=30),
```
ccojocar, line 638: Why 7.5 hours and not once per day?

dojo/utils.py
```diff
@@ -2108,3 +2121,100 @@ def is_scan_file_too_large(scan_file):
...
+                query = Q(active=True, verified=True, is_Mitigated=False, duplicate=False)
+            logger.debug("My query: {}".format(query))
+
+            no_jira_findings = {}
+            if settings.SLA_NOTIFY_WITH_JIRA_ONLY:
+                logger.debug("Ignoring findings that are not linked to a JIRA issue")
+                no_jira_findings = Finding.objects.exclude(jira_issue__isnull=False)
+            findings = Finding.objects \
+                .filter(
+                    query
```
ccojocar, line 2179: Is it this query limited per product, or is for all active findings?

dojo/settings/settings.dist.py
madchap: Yep.. they were there, and then not (i.e. they were in the actual PR desc).

dojo/utils.py
madchap: Unless I am mistaken, it is every day at 7:30AM (TZ being defined by Django).

dojo/settings/settings.dist.py
madchap: Unless I am mistaken, it is every day at 7:30AM (TZ being defined by Django).

dojo/utils.py
madchap: All findings indeed. I think I have a TODO to eventually scope it at the Product level. But wanted to get it out first before the week-end.

dojo/settings/settings.dist.py
madchap: https://docs.celeryproject.org/en/stable/userguide/periodic-tasks.html#crontab-schedules

madchap:
@valentijnscholten any potential feedback?

dojo/db_migrations/0047_sla_notifications.py
```diff
@@ -0,0 +1,24 @@
...
+
+    dependencies = [
+        ('dojo', '0046_endpoint_status'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='jira_conf',
+            name='sla_notification',
+            field=models.BooleanField(default=True, verbose_name='Send SLA notifications as comment?'),
```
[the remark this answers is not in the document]
madchap: I was thinking this could somehow be "overridden" through the notification checkbox on the product view page. If unchecked, then the notifications would not be sent. Could be checked by default, something like that.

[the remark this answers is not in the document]
madchap: So actually, if nobody set notifications anywhere, no one would get them, which is what was intended I guess.
However, it will still go through all the findings -- potentially for nothing.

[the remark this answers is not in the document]
madchap: But probably makes more sense.. will try something for this then:

image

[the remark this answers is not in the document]
valentijnscholten: I think that's the better level to have it at.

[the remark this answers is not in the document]
madchap: Done. Product level notification will override the global JIRA setting.

dojo/utils.py
madchap: My TODO was not for that, but a JIRA scope override (JIRA SLA comment), which is now in.

The assumption here is that if you enable SLA at the System Settings level, then you implicitly want to know when you breach them -- it was kind of a missing feature IMO. So basically, just aligning with what's already in there. No plan to change that at the moment, unless one day SLA goes to the Product level instead of System Settings level.

madchap:
Seems my crontab task is not getting registered and does not get triggered. Checking..

madchap:
There was a template exception. Now OK.

Also added documentation in the companion PR https://github.com/DefectDojo/Documentation/pull/112

dojo/utils.py
```diff
@@ -2108,3 +2121,102 @@ def is_scan_file_too_large(scan_file):
...
+                ).exclude(id__in=no_jira_findings)
+
+            for finding in findings:
+                # A finding with 'Info' severity will not be considered for SLA notifications (not in model)
+                if finding.severity == 'Info':
```
ccojocar, line 2182: nit: you can add this filter directly into the query

dojo/tasks.py
```diff
@@ -382,3 +382,9 @@ def send_hipchat_notification_task(*args, **kwargs):
 def send_alert_notification_task(*args, **kwargs):
     logger.debug("send_alert_notification")
     send_alert_notification(*args, **kwargs)
+
+
+@app.task(name='dojo.tasks.async_sla_compute_and_notify')
+def async_sla_compute_and_notify_task(*args, **kwargs):
+    logger.debug("Computing SLAs and notifying as needed")
+    sla_compute_and_notify(*args, **kwargs)
```
ccojocar, line 390: nit: maybe you should catch and handle all the exception which might be eventually thrown by this function in order to avoid the cron job from crashing

dojo/utils.py
```diff
@@ -2108,3 +2121,102 @@ def is_scan_file_too_large(scan_file):
...
+            logger.debug("My query: {}".format(query))
+
+            no_jira_findings = {}
+            if settings.SLA_NOTIFY_WITH_JIRA_ONLY:
+                logger.debug("Ignoring findings that are not linked to a JIRA issue")
+                no_jira_findings = Finding.objects.exclude(jira_issue__isnull=False)
+            findings = Finding.objects \
+                .filter(
+                    query
+                ).exclude(id__in=no_jira_findings)
```
valentijnscholten: if you add the filter directly in the query it will be a lot faster I think.

the no_jira_findings query to me looks it will be all findings WITH a jira_issue, so inverted?

dojo/utils.py
```diff
@@ -2108,3 +2121,102 @@ def is_scan_file_too_large(scan_file):
...
+                if sla_age is None:
+                    sla_age = 0
+
+                do_jira_sla_comment = False
+                if finding.has_jira_issue():
+                    jira_config = finding.jira_conf_new()
+                    if jira_config is not None:
+                        logger.debug("JIRA config for finding is {}".format(jira_config))
+                        # global config or product config set, product level takes precedence
+                        product_jira_sla_comment_enabled = finding.test.engagement.product.jira_pkey_set.first().product_jira_sla_notification
```
valentijnscholten, line 2194: this could crash if there is no jira_jpkey for a product (corner case).
in my JIRA PR I have introduced a `jira_pkey` property for product that will work with prefetching.

dojo/utils.py
valentijnscholten: I think you also want to do some prefetching because otherwise the part below will be very slow and hog your celery if you have 500k findings like I know you have :-)

for example something like this:

```
        prefetched_findings = prefetched_findings.select_related('reporter')
        prefetched_findings = prefetched_findings.select_related('jira_issue')
        prefetched_findings = prefetched_findings.prefetch_related('test__test_type')
        prefetched_findings = prefetched_findings.prefetch_related('test__engagement__product__jira_pkey_set__conf')
        prefetched_findings = prefetched_findings.prefetch_related('found_by')
        prefetched_findings = prefetched_findings.prefetch_related('risk_acceptance_set')
        # we could try to prefetch only the latest note with SubQuery and OuterRef, but I'm getting that MySql doesn't support limits in subqueries.
        prefetched_findings = prefetched_findings.prefetch_related('notes')
        prefetched_findings = prefetched_findings.prefetch_related('tagged_items__tag')
        prefetched_findings = prefetched_findings.prefetch_related('endpoints')
        prefetched_findings = prefetched_findings.prefetch_related('test__engagement__product__authorized_users')
```

dojo/tasks.py
madchap: Added try/catch, thanks ;-)

dojo/utils.py
madchap: Done, cheers.

dojo/utils.py
```diff
@@ -2108,3 +2121,102 @@ def is_scan_file_too_large(scan_file):
...
+    Notifications are managed the usual way, so you'd have to opt-in.
+    Exception is for JIRA issues, which would get a comment anyways.
+    """
+    def _notify(finding, title):
+        create_notification(
+            event='sla_breach',
+            title=title,
+            finding=finding,
+            sla_age=sla_age
+        )
```
valentijnscholten, line 2140: just thinking out loud: if a user once was an authorized user for a product and has configured notifications, the user will still receive those even after it would have been removed as authorized user from the product. probably not wanted/needed to solve here, so #2700

dojo/utils.py
madchap: Most of my stuff are "inactive", so it would not go through all of them, just a couple thousands. It takes about 5 minutes then on a 5yr old workstation.

madchap: prefetching didn't work. Do you think it's a blocker for this PR? We could always look afterwards. I've exhausted my time on this PR :-(

madchap:
From a performance point of view, I have 600K findings loaded, and only 1500 are in-scope of the SLA notification either being active or active-verified, based on the settings. It takes roughly 5 minutes to process them on my workstation (5yr old cpu).

madchap:
Thanks for the approvals :)

I will correct the conflicts and merge when I am 110% good to go, and after discussing some more with @valentijnscholten on this one as he requested changes ;-)

PR #2669, opened 2020-07-10
fix: prevent saving empty cve

edersonbrilhante:
## Description:
The fields CVE in fiends allows null value, however, when the finding is updated empty is overwritten this.
I notified this issue when I was using an openapi client consuming the API:
``` raise ValueError("Invalid value for `cve`, length must be greater than or equal to `1`")```

label: bugfix


Maffooch:
Was there no migration generated with the change to models.py?

madchap (dismissed):
Missing the migration file definitely.

edersonbrilhante:
Hello @madchap and @Maffooch  :)
Thank you for the review.

Django is not generating the migrations.

https://docs.djangoproject.com/en/2.2/topics/db/models/#field-options
Django document says:

> null
> If True, Django will store empty values as NULL in the database. Default is False.
> 
> blank
> If True, the field is allowed to be blank. Default is False.
> 
> Note that this is different than null. null is purely database-related, whereas blank is validation-related. If a field has blank=True, form validation will allow entry of an empty value. If a field has blank=False, the field will be required.

Just changes in `null` are needed migrations

madchap:
Turns out that's true! Sorry about that, and thanks for double-checking @edersonbrilhante !

valentijnscholten:
Have you tested it to be sure that nothing breaks by seeing a `None` value instead of ```` or `0`?

valentijnscholten:
Are these screenshots from the API? What I meant was under the hood the cve field is used for certain things like deduplication and I was wondering if that all still works. What if existing findings have cve as ''  or as 0 and new findings get 'None' for cve. Will the dedupe still work? Will the hash be the same? Also there are some `if cve` kind of constructs in templates, which should be checked but I think those are fine because for 0 or '' or None python always returns False I think

edersonbrilhante:
Hi @valentijnscholten, I was busy with another project, so I couldn't check this earlier :(.
Yes, these screenshots are from API. 
I will check your questions

edersonbrilhante:
Just remembering, I want to fix the issue that overwrites the cve field from None to '', when I am using the interface.
The hash is not changed when the finding is updated. So parsers with `cve = None` are ok.

Parsers with `cve = ''`
- dojo/tools/anchore_engine/parser.py
- dojo/tools/ibm_app/parser.py
- dojo/tools/snyk/parser.py
- dojo/tools/sonatype/parser.py
I can change the parser to set None as default value, because I see that as a bug. What do you think?

madchap:
I think they should be updated, yes. If you can take care of them, it's be great :)

edersonbrilhante:
> Just remembering, I want to fix the issue that overwrites the cve field from None to '', when I am using the interface.
> The hash is not changed when the finding is updated. So parsers with `cve = None` are ok.
> 
> Parsers with `cve = ''`
> 
> * dojo/tools/anchore_engine/parser.py
> * dojo/tools/ibm_app/parser.py
> * dojo/tools/snyk/parser.py
> * dojo/tools/sonatype/parser.py
>   I can change the parser to set None as default value, because I see that as a bug. What do you think?

Hi folks. 
First, sorry for fixing this so late. I changed the files:
* dojo/tools/anchore_engine/parser.py
* dojo/tools/ibm_app/parser.py
* dojo/tools/snyk/parser.py
* dojo/tools/sonatype/parser.py

PR #2679, opened 2020-07-15
Improve the helm chart and add support for cloudsql-proxy

ccojocar:
The following improvements are made to the helm chart:
- create a dedicated service account for all pods created by defectdojo and allows custom annotations (e.g. this is useful for workload identity)
- add support for cloudsql-proxy which is deployed as a side car container in all Pods which needs to connect to DB
- extend the mysql and postgresql secrets to allow to inject a user password via a helm value when the in-cluster DB is not used (e.g. useful for an external cloud hosted DB)
- change the deployments to use the same secret for DB password instead of expecting secrets to be created outside of the helm chart
- fix the celery host in the configmap
- increase the chart version to 1.1.0

Signed-off-by: Cosmin Cojocar <cosmin.cojocar@gmx.ch>

### DIFF
--- a/helm/defectdojo/templates/celery-beat-deployment.yaml
+++ b/helm/defectdojo/templates/celery-beat-deployment.yaml
@@ -23,11 +23,26 @@ spec:
         app.kubernetes.io/name: {{ include "defectdojo.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
     spec:
+      serviceAccountName: defectdojo
       {{- if .Values.imagePullSecrets }}
       imagePullSecrets:
       - name: {{ .Values.imagePullSecrets }}
       {{- end }}
       containers:
+      {{- if .Values.cloudsql.enabled  }}
+      - name: cloudsql-proxy
+        image: {{ .Values.cloudsql.image.repository }}:{{ .Values.cloudsql.image.tag }}
+        imagePullPolicy: {{ .Values.cloudsql.image.pullPolicy }}
+        securityContext:
+          runAsNonRoot: true
+        command: ["/cloud_sql_proxy"]
+        args:
+        {{- if eq .Values.database "postgresql" }}
+        - "-instances={{ .Values.cloudsql.instance }}=tcp:{{ .Values.postgresql.service.port }}"
+        {{- else if eq .Values.database "mysql" }}
+        - "-instances={{ .Values.cloudsql.instance }}=tcp:{{ .Values.mysql.service.port }}"
+        {{- end }}
+      {{- end }}
       - command:
         - /entrypoint-celery-beat.sh
         name: celery
@@ -40,25 +55,16 @@ spec:
         - name: DD_CELERY_BROKER_PASSWORD
           valueFrom:
             secretKeyRef:
-              # Use broker chart secret
-              # name: {{ $fullName }}-{{ .Values.celery.broker }}
-              # Use secret handled outside of the chart
               name: defectdojo-{{ .Values.celery.broker }}-specific
               key: {{ .Values.celery.broker }}-password
         - name: DD_DATABASE_PASSWORD
           valueFrom:
             secretKeyRef:
               {{- if eq .Values.database "postgresql" }}
-              # use postgresql chart secret
-              # name: {{- if .Values.postgresql.enabled }} {{ $fullName }}-{{ .Values.database }} {{- else }} {{ .Values.postgresql.existingSecret }} {{- end }}
-              # Use secret handled outside of the chart
-              name: {{- if .Values.postgresql.enabled }} defectdojo-postgresql-specific {{- else }} {{ .Values.postgresql.existingSecret }} {{- end }}
+              name: {{ .Values.postgresql.existingSecret }}
               key: postgresql-password
               {{- else if eq .Values.database "mysql" }}
-              # Use mysql chart secret
-              # name: {{- if .Values.mysql.enabled }} {{ $fullName }}-{{ .Values.database }} {{- else }} {{ .Values.mysql.mysqlPasswordSecret }} {{- end }}
-              # Use secret handled outside of the chart
-              name: {{- if .Values.mysql.enabled }} defectdojo-mysql-specific {{- else }} {{ .Values.mysql.mysqlPasswordSecret }} {{- end }}
+              name: {{ .Values.mysql.existingSecret }}
               key: mysql-password
               {{- end }}
         - name: DD_SECRET_KEY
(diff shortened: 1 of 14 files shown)


alles-klar:
@ccojocar can you check the mysql setup? The tests are failing.

ccojocar:
@alles-klar The unit tests should be fixed now. The build is green.

helm/defectdojo/templates/sa.yaml
```diff
@@ -0,0 +1,16 @@
+kind: ServiceAccount
+apiVersion: v1
+metadata:
+  name: defectdojo
```
alles-klar, line 4: Can you rename it, that it is consistent with the other templates (i.e. {{ $fullName }}-service-account). Otherwise we may get problem in environments with multiple deployments.

helm/defectdojo/templates/celery-beat-deployment.yaml
```diff
@@ -23,11 +23,26 @@ spec:
         app.kubernetes.io/name: {{ include "defectdojo.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
     spec:
+      serviceAccountName: defectdojo
       {{- if .Values.imagePullSecrets }}
       imagePullSecrets:
       - name: {{ .Values.imagePullSecrets }}
       {{- end }}
       containers:
+      {{- if .Values.cloudsql.enabled  }}
```
alles-klar, line 32: Is it possible to define this code only once? Maybe including it or define a separate service. Otherwise we have a lot of duplicate code or do I miss some benefits of this solution?

helm/defectdojo/templates/secret-postgresql.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: defectdojo-postgresql-specific
+  name: {{ .Values.postgresql.existingSecret }}
```
alles-klar, line 5: Is "existingSecret" not a bit misleading? If I understood it correct the secret is created always, but in one case the password is given and not generated.

helm/defectdojo/templates/configmap.yaml
```diff
@@ -16,7 +16,7 @@ data:
   DD_ALLOWED_HOSTS: {{ .Values.host }}
   DD_CELERY_BROKER_SCHEME: {{ if eq .Values.celery.broker "rabbitmq" }}amqp{{ end }}{{ if eq .Values.celery.broker "redis" }}redis{{ end }}
   DD_CELERY_BROKER_USER: '{{ if eq .Values.celery.broker "rabbitmq" }}user{{ end }}'
-  DD_CELERY_BROKER_HOST: {{ $fullName }}-{{ if eq .Values.celery.broker "rabbitmq" }}rabbitmq{{ end }}{{ if eq .Values.celery.broker "redis" }}redis-master{{ end }}
+  DD_CELERY_BROKER_HOST: {{ .Release.Name }}-{{ if eq .Values.celery.broker "rabbitmq" }}rabbitmq{{ end }}{{ if eq .Values.celery.broker "redis" }}redis-master{{ end }}
```
alles-klar, line 19: Was it your intent to change the name here? If yes, can you please shortly explain why?

alles-klar:
@ccojocar Thanks for your contribution. I provided some comments, but generally it is great work and helps to improve DefectDojo :)

helm/defectdojo/templates/configmap.yaml
ccojocar: The $fullname does appear to be the correct name for the broker host but rather the `.Releese.Name`.

helm/defectdojo/templates/secret-postgresql.yaml
ccojocar: That's right, the previous implementation was expecting to have the secret created upfront, outside of the helm chart, which is not quite self-contained for the case when you want to install it for instance with helmfile in one go, and it uses an external database.

helm/defectdojo/templates/sa.yaml
ccojocar: fixed

helm/defectdojo/templates/celery-beat-deployment.yaml
ccojocar: I'll check if is possible to have this snippet into an helper and include when is needed.

ccojocar: I tried to play with the template to include this snippet but I haven't come to a good solution. Are you fine if we merge this in this form and try to follow up later with a refactoring?

ccojocar:
@alles-klar Thanks for your review! I've tried o address your comments. Please let me know if it is anything required in order to merge this pull request.

ccojocar:
@alles-klar It seems that the random values used for all credentials values are regenerated every time by the helm update regardless of the hooks present in the annotations. This is the case especially with helmfile.  See https://github.com/helm/helm/issues/3053

I pushed this commit https://github.com/DefectDojo/django-DefectDojo/pull/2679/commits/da4be89c9bcb19c9fc4f2d57a04876cf2d9f60c6 which allows to inject all credentials via helm values, when no value is provided, a  random value is generated. This allows us to control which credentials values are used, and keep them stable on chart updates.

helm/defectdojo/templates/celery-beat-deployment.yaml
alles-klar: If there is no quick solution it is fine for me.

helm/defectdojo/templates/secret-postgresql.yaml
alles-klar: Should we rename the variable to "secret" ?

alles-klar:
> @alles-klar It seems that the random values used for all credentials values are regenerated every time by the helm update regardless of the hooks present in the annotations. This is the case especially with helmfile. See [helm/helm#3053](https://github.com/helm/helm/issues/3053)
> 
> I pushed this commit [da4be89](https://github.com/DefectDojo/django-DefectDojo/commit/da4be89c9bcb19c9fc4f2d57a04876cf2d9f60c6) which allows to inject all credentials via helm values, when no value is provided, a random value is generated. This allows us to control which credentials values are used, and keep them stable on chart updates.

Great point, I think this was the reason why we decided to place the secret outside of the helm chart. Otherwise is would not be possible to update the helm chart and providing secrets via value-files is not the best practice. I really don't want to miss this functionality.

ccojocar:
> Otherwise is would not be possible to update the helm chart and providing secrets via value-files is not the best practice. I really don't want to miss this functionality.

The old behavior is still in place if the values are empty.

helm/defectdojo/templates/secret-postgresql.yaml
ccojocar: I can rename them.

ccojocar: It should be renamed now.

ccojocar:
The build failure seems to me related to the current state of dev branch, because it started to fail after rebase.

PR #2682, opened 2020-07-15
Revert "uwsgi: add variable to configure the buffer-size (#2662)"

Maffooch:
This reverts commit 9c7d7fea0012dccc94655b8d6e791ee163e71318.

Didn't notice that #2662 was based on master until after merge :/ 
Not sure if reverting is the right move or not. Just don't want master to have commits that dev would not have. 
@madchap @valentijnscholten

### DIFF
--- a/docker/entrypoint-uwsgi-dev.sh
+++ b/docker/entrypoint-uwsgi-dev.sh
@@ -16,4 +16,5 @@ exec uwsgi \
   --enable-threads \
   --processes 2 \
   --threads 2 \
-  --py-autoreload 1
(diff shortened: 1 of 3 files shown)


valentijnscholten:
@Maffooch  Reverting would have worked. But I just did a `git reset --hard xxxxx` to the previous commit and force pushed the `master` branch. I had to temporarily allow force push on the master branch, but it now looks clean as it was before. I've asked xens to recreate 2662 against dev.

Maffooch:
You rock @valentijnscholten thank you!

PR #2686, opened 2020-07-16
K8s check dd status 

valentijnscholten:
In travis build https://travis-ci.org/github/DefectDojo/django-DefectDojo/builds/708492749 we see that defect dojo is not working correctly, the login page doesn't work. The integrationtests step in travis detects this and lets the test fail correctly.
The kubernest steps in travis are still looking green and happy. This PR is an attempt to add the same check there to avoid false negatives.

### DIFF
--- a/travis/script.sh
+++ b/travis/script.sh
@@ -163,6 +163,11 @@ if [ -z "${TEST}" ]; then
     return_value=1
   fi
 
+  echo
+  echo "UWSGI logs"
+  sudo kubectl logs --selector=defectdojo.org/component=django -c uwsgi
+  echo
+
   echo "Testing DefectDojo Service"
   # sudo kubectl port-forward --namespace=default service/defectdojo-django 8080:80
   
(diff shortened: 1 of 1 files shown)


valentijnscholten:
ok, I was trying the url specified by minikube, but looks like more steps are required to be able to access it:

```
To be able to access it, set up an ingress or access the service directly by
running the following command:

    kubectl port-forward --namespace=default \
      service/defectdojo-django 8080:80

As you set your host value to defectdojo.default.minikube.local, make sure that it resolves to
the localhost IP address, e.g. by adding the following two lines to /etc/hosts:

    ::1       defectdojo.default.minikube.local
    127.0.0.1 defectdojo.default.minikube.local

To access DefectDojo, go to <http://defectdojo.default.minikube.local:8080>.
```

valentijnscholten:
@dsever any idea how we could fix this?

dsever:
I've just spent few minutes on it, but for me it looks like race condition, I will try to work with that today

valentijnscholten:
superseeded by #3104

PR #2705, opened 2020-07-21
merge master into dev after 1.7.0 release

Maffooch:
- [x] Give a meaninful name to your PR, as it may end up being used in the release notes.

@madchap @valentijnscholten 
There is some commits from the last release that are in here and I'm not sure if that's correct.


valentijnscholten:
The first commit in here from June is a bit weird, but maybe we missed one back then. Ideally we shouldn't make any other changes in this PR such as changing version numbers, but in practice it won't break anything I think.

PR #2708, opened 2020-07-21
Add support for another version of Aqua scanner

### LINKED ISSUE #2389: Aqua scanner format no longer working
**Scanner Name**
Aqua.

**Sample File**
@madchap I have a sample file internally from a real report. The gist here is that the old aqua file format uses a top level JSON key called "resources" where each scanned item has a list of vulnerabilities. In the new format (or at least the alternative format?), the top level JSON key is called "cves" and has flattened the structure to be more similar to how the Anchore file format works.

jvz:
This fixes #2389.

@jeffret-b @madchap FYI

### DIFF
--- a/dojo/tools/aqua/parser.py
+++ b/dojo/tools/aqua/parser.py
@@ -38,6 +38,10 @@ class AquaJSONParser(object):
                     item = get_item(resource, vuln, test)
                     unique_key = resource.get('cpe') + vuln.get('name', 'None')
                     items[unique_key] = item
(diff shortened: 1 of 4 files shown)


madchap:
Thanks @jvz :)

Could you please put these to https://github.com/DefectDojo/sample-scan-files/tree/master/aqua as well?

Cheers.

jvz:
Recreated at https://github.com/DefectDojo/django-DefectDojo/pull/2711

PR #2711, opened 2020-07-22
Add support for another Aqua format

### LINKED ISSUE #2389: Aqua scanner format no longer working

jvz:
@madchap 

Supersedes #2708.

This fixes #2389

### DIFF
--- a/dojo/tools/aqua/parser.py
+++ b/dojo/tools/aqua/parser.py
@@ -38,6 +38,10 @@ class AquaJSONParser(object):
                     item = get_item(resource, vuln, test)
                     unique_key = resource.get('cpe') + vuln.get('name', 'None')
                     items[unique_key] = item
+        elif 'cves' in tree:
+            for cve in tree['cves']:
+                unique_key = cve.get('file') + cve.get('name')
+                items[unique_key] = get_item_v2(cve, test)
 
         return list(items.values())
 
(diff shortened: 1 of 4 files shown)


madchap:
@jvz this one is still trying to merge to `master` and not `dev`?

madchap:
I could change it manually here without trouble, since everything is so fresh due to the latest release and given the scope of the PR. I relaunched a Travis, since it failed somehow on generating the nginx image.

dojo/tools/aqua/parser.py
```diff
@@ -100,4 +95,41 @@ def get_item(resource, vuln, test):
...
+    solution = item.get('solution')
+    fix_version = item.get('fix_version')
+    if solution:
+        mitigation = solution
+    elif fix_version:
+        mitigation = f'Upgrade to {fix_version}'
+    else:
+        mitigation = 'No known mitigation'
+
+    return Finding(title=f'{cve}: {file_path}',
```
madchap, line 114: It seems that `name` would contain what would be considered the `file_path` or `component_name` in DefectDojo, e.g. for java components.

valentijnscholten:
Turns out this PR introduced some flake8 errors, fixed in #2726 2726

PR #2717, opened 2020-07-23
Rework apiv2 findings filters

xens:
This PR is the first in a series to improve DDJ APIv2. As of today the `/findings` api endpoint does not support:
* multiple values for the same key
* ordering
* partial text-string search

With this PR we will be able to query most of the fields of the `findings` endpoint, except a few ones that are not relevant, and we'll be able to do sorting on the most relevant ones. We'll also be able ot do test-string search on strings fieds.

A couple of notes:

1. unified query-style: I'm not at all a REST expert but for this PR I tried to stick to one logic when supporting multiple values per keys. For all the fields that supports it, the query syntax is the following: `/findings?key=value1,value2`. I did it this way because only the MutipleChoiceFilter is supporting the `/findings?key=value1&key=value2` syntax so I would not have been able to have unified query-style across all the fields with this style.

2. for long strings fields like `description`, `title`, `step_to_reproduce`, it's not possible to do a multiple value/key query, it's too messy to manage when there's `,` inside the string. I guess that's acceptable.

3. This change introduce **breaking** change for some fields like:
  * `test__engagement__risk_acceptance`
  * `test__engagement__product`
  * `test__test__type`

4. As mentioned in the intro I'll probably rework also some other endpoints (like products) as most (if not all) have the same limitations, so we need to properly review it and understand the pros and cons of this proposal before pushing that logic further.


valentijnscholten:
In the past we had some issues with filters like this. In this case the filter is initiliazed with a list of CWE values to choose from. Which may make sense in the UI. But if the API request tries to filter by a CWE value in that is not in the list, I'm not sure what happens. In the past I've seen cases where the filter is ignored and findings are returned regardless of CWE. So that is one thing to check. Second thing is I am still not sure if we want to have the fitler used in both places. Might be better to have a little bit of code duplication but complete flexibility. I think the API needs to be able to filter by all fields for example.

xens:
@valentijnscholten this is exactly why I created a dedicated filter for the API, there's too much things that are UI-specific to have one filter that handles both the API and the UI.

Regarding the filtering; Yes we should be able to filter on all the available fields on the `findings` endpoint, that's already the case with my PR.

Regarding the behavior of the API in general; I think that the API should support multiple values for the same keys for all the fields except for boolean and dates (for example; cwe=xxx,yyy,zzz, or ids=111,222,333) 

Regarding the behavior with CWE; With my current proposal it'll work this way

1. If you don't provide any value to a key, for example `api/v2/findings/?cwe=` -> the filter will not be active and it'll return all the findings (I assume that it's the expected behavior ?)
2. If there's one or two existing CWEs  `api/v2/findings/?cwe=111,222` -> it'll return the findings for both CWE 111 and 222
3. If the filter contains a number for existing CWEs and non existing CWEs only the findings for existing CWEs will be returned.

Let me know if that makes sense, if it does I'll continue to work into that direction.

valentijnscholten:
Sounds good to me

xens:
One thing that I'm not sure is how do we want to accept list of parameter as input, do we want to repeat the key (`/?id=101404&id=7267261`) or pass a comma-separated list of values ? (`/?id=101402,7267261`)

xens:
I answer my question myself in the mean time, I'll go with `/?id=101402,7267261` as most of the filters only support this style. @valentijnscholten I've removed the WIP on my side I think I'm good.

dmut:
Would be really useful to get this merged. 
I need filtering by date, and was just about to open a PR with a more simplistic approach like

```
filter_fields = {
        'id': ['exact'],
        'title': ['exact', 'icontains'],
        'date': ['exact', 'gt', 'lt', 'gte', 'lte'],
        'severity': ['exact', 'in'],
        'description': ['exact', 'icontains'],
        ...
```

But this is much nicer and extensive

dojo/filters.py
```diff
@@ -27,10 +28,39 @@
...
+
+def filter_by_test__engagement__product(queryset, name, value):
+    values = value.split(',')
+    return queryset.filter(test__engagement__product__in=values)
+
+
+def filter_by_test__test__type(queryset, name, value):
+    values = value.split(',')
+    return queryset.filter(test__test__type__in=values)
+
```
valentijnscholten, line 54: Could this be refactored to have less code duplication? The `test__test__type__in=values` could maybe be turned into a Q  object, or a function that returns a Q object?

valentijnscholten:
Still feels that the cve=1&cve=2&cve3 is the more natural choice, but when looking around I see the csv based approach also being used sometimes. It's even supported by django in by user `getlist()` on request objects. But I can't find any native support for either way in the Django Rest Framework.  So might be almost good to go.

valentijnscholten:
> Would be really useful to get this merged.
You speed up the review/merge process by merging locally and testing it :-)

xens:
> Still feels that the cve=1&cve=2&cve3 is the more natural choice, but when looking around I see the csv based approach also being used sometimes. It's even supported by django in by user `getlist()` on request objects. But I can't find any native support for either way in the Django Rest Framework. So might be almost good to go.

Yes I agree, that would have been my preferred choice if the framework supported it properly.

dojo/filters.py
xens: `filter_by_test__engagement__product` and `filter_by_test__test__type` were left-overs thanks for pointing that out, I removed them.
I also made the code a bit DRYer using a Q filter (the doc is a pain seriously it took me 3 hours at least :sweat_smile: )

valentijnscholten: Yeah, that `Q(**{..}` is always fun, but looks nice and tidy now!

valentijnscholten:
Thanks so far. I tempted to ask for a small addition to `test_rest_framework.py` to touch at least once upon this new filter type?

xens:
> Thanks so far. I tempted to ask for a small addition to `test_rest_framework.py` to touch at least once upon this new filter type?

Of course !

xens:
@valentijnscholten I quickly checked and there's already some tests for the ```FindingViewSet```  https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/unittests/test_rest_framework.py#L182 that I modified part of this PR, so they should already validate this new filter.

I'm not super familiar with this test framework, is there anything in particular that you want me to add ? Or a new test targeting something specific that I've added ? Any hints are welcome :)

xens:
@valentijnscholten any news regarding my last comment ?

madchap (approved):

madchap:
@xens you have a conflict to resolve to clear this up.

I re-requested @valentijnscholten 's quick review on this too, so we can move forward.

madchap:
The conflict was easy to fix, so I went ahead and did it. Let's wait for the travis checks to (successfully) complete.

xens:
> The conflict was easy to fix, so I went ahead and did it. Let's wait for the travis checks to (successfully) complete.

Awesome thanks for fixing that, I was waiting for this PR to be merged before doing the same for the `/products/` endpoint

valentijnscholten:
Does the swagger output also look OK with this new filter in place? We recenlty made an effort to make the swagger spec 100% so users can easily generate api wrapper from it via openapi for example. Would be good to check that.

xens:
I think it does, if you want to cross check I'm adding the /findings/ GET Swagger json output here [findings_get.zip](https://github.com/DefectDojo/django-DefectDojo/files/5212587/findings_get.zip).

I also cross-checked the serialization format on the [Swagger documentation](https://swagger.io/docs/specification/serialization/) and the syntax `/findings?key=value1,value2`  is something valid (see Query Parameters section).

valentijnscholten:
I'm not that into swagger/openapi json, but I see also in current dev the swagger docs for findings_list have almost no specs on filtering. At some point we may want to look at that because a good api spec can generate good wrappers so we don't need to maintain our own wrappers. approving now so you can continue with the other PRs as well.

valentijnscholten (approved):

xens:
I need someone to click on merge :pray: ;)

madchap:
I guess I just hit a breaking change with `engagement__test`.. more on https://owasp.slack.com/archives/C014H3ZV9U6/p1602624904093500

xens:
> I guess I just hit a breaking change with `engagement__test`.. more on https://owasp.slack.com/archives/C014H3ZV9U6/p1602624904093500

Fixed by https://github.com/DefectDojo/django-DefectDojo/pull/3000

PR #2719, opened 2020-07-23
Add endpoint_status creation script

cody-m-tibco:
After the upgrade to 1.7.0, endpoints on a finding were no longer displayed in the UI. This is because an Endpoint_Status object had not been created for the endpoint/finding pair. This manage.py command will only need to be ran one time when upgrading to 1.7.0 as it creates new status objects for every endpoint/finding pair. 

A [PR](https://github.com/DefectDojo/Documentation/pull/114) in the documentation repo coming soon. 

- [x] Give a meaninful name to your PR, as it may end up being used in the release notes.


valentijnscholten:
What happens if someone in the mean time created some new endpoints that already have an endpoint_status record?
What happens if 1 creation fails? Might be best to catch the error and continue?
Could / should it be added as migration to facilitate the process?

cody-m-tibco:
* I cannot think of a way to add a new endpoint hat already has a status record. A status record is unique as it has ForiegnKeys to the single endpoint and finding it's attached to. 
* Yes error catching is a good move. Will update.

valentijnscholten:
If someone didn't read the upgrade notes and didn't run the script and then imported a bunch of scans he/she would end up with some findings with endpoint_status records created by 1.7.x. The script in this PR would then try to create another endpoint_status record because it just goes over all findings and all endpoints. But maybe it's pythonic to do that and just catch and ignore any errors :-)

cody-m-tibco:
The error checking here:
* If there is already a status, it will retrieve the object and move onto the next one. 
* Endpoints retrieved are only attached to the current finding, so adhoc endpoints would not apply here.

Let's see what travis thinks :)

valentijnscholten:
Great, this is what I was thinking. Still wondering if a try catch would be beneficial? what if there's one weird finding half way through crashing the migration?

cody-m-tibco:
Yeah we can do that. I don't like using them unless as a last resort.

valentijnscholten:
approved

valentijnscholten:
Did a sepearte PR to quickfix the aqua parser flake8 stuff in #2726.

cody-m-tibco:
Making a new PR baed on `master` to remove flake8 fixes on aqua and have a cleaner commit history.

PR #2727, opened 2020-07-25
fix popovers in view_test and improve js overall

### LINKED ISSUE #2631: Component tooltip not displayed on first row
When in a test page, with the list of findings, the table entries contain a black file icon when the filepath to the vulnerable component is known. 

That icon generates a tooltip when we move the mouse over it. Except for the first row of the table (even after changing page).

tooltip_not_displayed

valentijnscholten:
- popovers were broken on `view_test` page (fixes #2631)
- extracted `tooltip` and `popover` javascript calls into `base.html`
- moved the `notes` icon from the JIRA column to the title column
- improved styling of notes icon and number of notes displayment
- some small js / whitespace changes here and there

### DIFF
--- a/dojo/templates/dojo/edit_rule.html
+++ b/dojo/templates/dojo/edit_rule.html
@@ -16,7 +16,6 @@
     $(document).ready(function() {
         $('#id_applied_field).empty();
         $('#id_match_field).empty();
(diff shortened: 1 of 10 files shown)


valentijnscholten:
@ccojocar thanks for taking over review duty from @madchap  ;-)

PR #2728, opened 2020-07-26
improve deduplication on engagement wording

### LINKED ISSUE #2172: dedupe on engagement naming confuses users
A couple of times the `dedupe on engagement` checkbox has been confusing for users. They tend to think it will enable dedupe for their engagements, but don't realize it will actually narrow the scope of dedupe to be performed only within that engagement.

Anyone got a suggestion for a better name for that setting?

'dedupe only inside engagement'?
'dedupe only within engagement'?

Add a helptext as well, or maybe only a helptext is sufficient.

valentijnscholten:
fixes #2172 
use more clear wording to explain that deduplication_on_engagement means that deduplication happens insidie/within that engagement as opposed to the default product level based deduplication.


valentijnscholten:
waiting for #2719 which may become a hotfix and has a migration as well.

Maffooch:
Migration from #2719 is now in dev!

valentijnscholten:
ready to be (reviewed and) merged!

Issue #2734, opened 2020-07-28
Eslint parsing error


SPoint42:
When ruleId or message is not a string, the parser generate a 500 error. 

whith this patch you avoid this. 

```

===================================================================
--- dojo/tools/eslint/parser.py	(date 1595949920131)
+++ dojo/tools/eslint/parser.py	(date 1595949920131)
@@ -28,7 +28,7 @@
                 continue
 
             for message in item["messages"]:
-                title = message["message"] + " Test ID: " + message["ruleId"]
+                title = str(message["message"]) + " Test ID: " + str(message["ruleId"])
 
                 #  ##### Finding details information ######
                 findingdetail += "Filename: " + item["filePath"] + "\n"

```

valentijnscholten:
Thanks, are you in the position to create a PR (against dev)?

SPoint42:
Done in PR  #2736

PR #2736, opened 2020-07-28
ESLint parser title error checking

SPoint42:
when ruleId=null or message=null, there is a 500 error. This avoid it

### DIFF
--- a/dojo/tools/eslint/parser.py
+++ b/dojo/tools/eslint/parser.py
@@ -29,7 +29,7 @@ class ESLintParser(object):
 
             for message in item["messages"]:
 
-                if message["message"] is  None:
+                if message["message"] is None:
                     title = str("Finding Not defined")
                 else:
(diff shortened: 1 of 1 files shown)


SPoint42:
refer to #2734

dojo/tools/eslint/parser.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, filename, test):
                 continue
 
             for message in item["messages"]:
-                title = message["message"] + " Test ID: " + message["ruleId"]
+                title = str(message["message"]) + " Test ID: " + str(message["ruleId"])
```
Maffooch, line 31: Can we do some kind of default title here? In the case of message=null and ruleId=null, the title of the finding would be "None Test ID: None".

SPoint42: Agreed. 
Will send a updated PR

PR #2649, opened 2020-07-06
jira: allow unlinking/linking existing JIRA issues

valentijnscholten:
* allows to push to jira from the view_finding page using nice ajax buttons. decided to NOT add this functionality to bulk edit menu as it's a corner case and we don't want to overcrowd the bulk edit menu

* allows to unlink a JIRA issue from a defect dojo finding. For example if the JIRA issue was deleted, or you want to create fresh one, or split JIRA issues, etc.

* allows to link an existing JIRA issue to DD finding. This turned out to be quite the effort. as evertyhing around jira and forms seems to coupled together. forms are being abused on different places and fields overwritten under the hood. So this PR also cleans up some stuff, and fixes a bunch of bugs:
     * fixes #2663 
     * fixes #2665
     * fixes #2693 
     * fixes #2690 
     * fixes #2694
     * fixes #2695 

* I wanted to do another cleanup of all the JIRA code, but I am running our of time.

* the jira_issue field that can be populated with an existing jira does two things:
     * validate against regex
     * strip full url down to just the issue key when you paste a full jira url into the field

* I did quite a lot of testing, at least the following:
```
jira system settings disabled:
- add finding
- update finding

jira system settings enabled:
	product jira disabled:
		- add finding
		- edit finding

	product jira enabled:
		- add finding:
			- with existing jira
			- without existing jira
		- edit finding:
			- add existing jira issue >
			- remove linked jira issue
				- without push_to_jira >
				- with push_to_jira >
			- modify linked jira issue
				- without push_to_jira >
				- with push_to_jira >
			- create new jira issue >
	
	
	add finding from template
	- with push_to_jira
	- without push_to_jira
	
	add stub finding and promote to finding
	- existing jira issue
	- no jira issue
	
	run imports
```

* I will clean up the prin/logging once reviewed

* a similar approache could be implemented for the github integration, but my sponsor is out of budget so that's left for a future PR.

PR #2737, opened 2020-07-28
JIRA: small bugfixes, improvements and cleanup

valentijnscholten:
- bugfix on `edit_engagement`
- set note author to be `jira` after pushing/linking to JIRA issue instead of `finding.reporter`. This is not perfect, but in celery we don't have the request object so we don't know who is actually pushing it.
- rename `add_issue` and `update_issue` in utils to more explicit `add_jira_issue` and `update_jira_issue`
- cleanup left over print/log statements from #2649 
- cleanup other print/log statements because they are not my favourite

### DIFF
--- a/dojo/tools/whitesource/parser.py
+++ b/dojo/tools/whitesource/parser.py
@@ -67,7 +67,7 @@ class WhitesourceJSONParser(object):
                             topfix_node.get('fixResolution')
                         )
                 except Exception as e:
-                    ("Error handling topFix node. {}").format(e)
+                    print("Error handling topFix node. {}").format(e)
 
             filepaths = []
             if 'sourceFiles' in node:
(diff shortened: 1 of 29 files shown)


dojo/tools/whitesource/parser.py
```diff
@@ -67,7 +67,7 @@ def _build_common_output(node, lib_name=None):
                             topfix_node.get('fixResolution')
                         )
                 except Exception as e:
-                    print("Error handling topFix node. {}").format(e)
+                    ("Error handling topFix node. {}").format(e)
```
Maffooch, line 70: Is a logger missing here?

valentijnscholten: wow, nice spot. am struggling lately with visual studio code is it's so damn slow with python code that even copy and pasting is laggy

valentijnscholten: I reverted it back to what it was.... which is invalid python code...

valentijnscholten:
Merged this into dev as the other mods are on holidays and this is a lightweight PR but has a bugfix for dev (bug is not present in 1.7.x).

PR #2740, opened 2020-07-29
Added user-agent in sonarqube api requests

felipemgurian:
While I was configuring the DefectDojo integration with SonarQube API, I came across the following problem, Sonarqube was returning StatusCode 403, after researching a bit, I found that error occurred because sonarqube was behind a WAF that did not accept requests without user-agent, I made this small change that solved my problem!


valentijnscholten:
Stupid WAF :-) We should put `1337 h4x0r` as user-agent.

PR #2748, opened 2020-07-31
view finding: duplicate & similar finding improvements and fixes

valentijnscholten:
A while ago the deduplication logic was changed to prevent circular dependencies and transitive duplicates. The similar findings feature wasn't completely updated and was broken because of that. 

Fixes:
- handle new setup with always 1 duplicate_finding field pointing to the original finding
- clearer list of possible actions and explanation for those actions
- hide findings for which user is not authorized
- display more columns for similar findings and duplicates
- more intuitive interface
- better overview of all the duplicates in a "cluster" of duplicates

image
image

All this was needed to make it work, but also to prepare for the actual enhancement I was looking to work on:

- allow actions on duplicates to reorganize hierarchy (duplicate vs original)
- better filtering for similar findings

Currently the similar findings are matched in a hardcoded way, so multiple fields have to match
- CVE
- CWE
- Filepath
- Linenumber
- unique_id_from_tool
- product or engagement

In a somewhat larger setup, there is a need for better filtering. For example only match on CVE but not on filepath because the filepath can be slightly different. Or the CWE can be absent in some findings.
Also it might be beneficial to be able to see similar findings across engagement / product / productype boundaries as some issues might already have been resolved elsewhere or can be an inspiration on how to solve them.

This is now also included:

image

Throughout the implementation I ran into several limitations (features) of the Django (template) framework. So I had to change my approach a couple of times. It could be that some things can still be improved, but I don't have any time left to go for those. I think the PR as is good enough and will allow us to gain experience with this new feature.

Future improvements:
- Add filtering on hash_code to allow listing of identical findings
- Improve performance of view_finding page currently the jira config takes lots of queries
- Improve field ordering in filter form
- Add paging instead of limiting to 10 findings
- Allow clearing of filter fields, currently they default to values from the finding being viewed
- Add toggles to include/exclude filtering on a certain field to allow for quickly playing around with the filers
- Add possibility to "link" to other findings that are not or cannot be marked as duplicate.


madchap:
fyi @Wadeck, @jvz

madchap (approved):
Looks awesome :)

Wadeck:
<3 not having to fight against the duplicates will be fantastic!

valentijnscholten:
I am ignoring all the accesslint issues as it keeps tripping over hidden fields. Maybe we should add labels anyway to silence accesslint :-)

PR #2420, opened 2020-05-24
Questionnaire Renaming and Fixes

JoseRoman:
- Updated messages that still said "Survey"
- Updated breadcrumbs 
- Updated views.py functions
- Updated HTML file names

Fixes #2418

PR #2774, opened 2020-08-13
questionnaire 500 internal server error

dayal18:
Changed files and names on the UI fixing the internal server 500 error.


dayal18:
I think there is #2420 which is covering most of these changes @JoseRoman can you change the PR to complete so that can be merged in next release.

Issue #2475, opened 2020-06-04
JIRA WebHook - Duplicate messages


dayal18:
**Bug description**
The JIRA webhook finally worked with 1.6.. for some reason the trailing "/" from the "https://<your DefectDojo url>/webhook/" URL in the JIRA config of our server was missing.

Anyways, as its working - the next issue I see is with duplicates updating JIRA webhook readings.

Add a new product with "Push Notes" enabled.

**Use case 1:** DefectDojo creates a finding to JIRA ---> Developer update comment in JIRA --> DefectDojo finding get the comments back in the Notes.. (good, works fine)

**Use case 2:** DefectDojo replies (notes get added) and ---> JIRA receives comments (Great) ----> But wait, as the JIRA got updated, so the comment comes back and updates it again in DefectDojo..

Master branch 1.6

dayal18:
This issue still exist in 1.7.1

![image](https://user-images.githubusercontent.com/43857391/90079042-65f1b580-dccc-11ea-87eb-406ceca16759.png)

dayal18:
I added notes to a finding in DefectDojo.. the notes add a comment in JIRA ticket, but also the comment comes back from JIRA back to DefectDojo

dayal18:
Hi guys, I'm close to fixing this, just want to understand how can I get the username value from the JIRA config:

get_system_setting. ???

![image](https://user-images.githubusercontent.com/43857391/90411609-a5742500-e071-11ea-881a-4f9aa39ab838.png)

@cody-m-tibco @Maffooch @mtesauro

Maffooch:
There's a Jira Config model that has that information. It is generally attached to a product, or can be queried directly if you have one of the fields in the config. [Here](https://github.com/DefectDojo/django-DefectDojo/blob/d8fb1cd38e981a2e24465dcc579409607274e137/dojo/finding/views.py#L364) is an example of such.

dayal18:
Thanks @Maffooch that worked.. working on PR now.. Thanks again !!

dayal18:
#2786 - Fixed

PR #2786, opened 2020-08-17
#2475 and enhancement - JIRA webhook notification

manav-nagla:
Check for JIRA username to prevent duplicate message from JIRA to DefectDojo.
Added notification when JIRA is updated via JIRA Webhook.


dayal18:
@Maffooch Noted !! Thanks :)

SPoint42 (approved):

SPoint42 (approved):

dojo/jira_link/views.py
```diff
@@ -95,6 +95,11 @@ def webhook(request):
             commentor = parsed['comment']['updateAuthor']['displayName']
             jid = parsed['comment']['self'].split('/')[7]
             jissue = JIRA_Issue.objects.get(jira_id=jid)
+            jira = JIRA_Conf.objects.values_list('username', flat=True)
+            for jira_userid in jira:
+                if jira_userid.lower() in commentor.lower():
```
xens, line 100: This condition will not work in all cases as the `displayName` inside Jira webhooks is not necessarily equal to the user name configured inside DDJ. The `displayName` in my integration is equal to the user + the name (john doe) and inside DDJ is the email address (john.doe@domain).

xens: proposal for a more robust mechanism here -> https://github.com/DefectDojo/django-DefectDojo/pull/2815

PR #2791, opened 2020-08-19
Django saml2 plugin dedicated cert

dsever:
Few improvements in terms of SAML2 integration:

* SAML2_AUTH will look for env variable DD_SAML2_METADATA_AUTO_CONF_URL, if not configured  will proceed with DD_SAML2_METADATA_LOCAL_FILE_PATH, this is needed because in k8s environment commenting part of configuration is not an option
* when IDP uses private signed certs (not included in certifi), we need to deliver cert into docker container to enable trust, so in Dockerfile.django is appended to copy content of project/certs/* to container root during the build phase, so later during application launching  pub cert can be referenced by  REQUESTS_CA_BUNDLE (https://github.com/DefectDojo/Documentation/pull/119)


dojo/settings/settings.dist.py
```diff
@@ -373,6 +370,13 @@ def generate_url(scheme, double_slashes, user, password, host, port, path):
     'ATTRIBUTES_MAP': env('DD_SAML2_ATTRIBUTES_MAP'),
 }
 
+# Metadata is required, choose either remote url or local file path
+if 'DD_SAML2_METADATA_AUTO_CONF_URL' in os.environ:
```
Maffooch, line 374: Change to the following so this isn't breaking for users that rely on the environ env vars within settings.py.

`if 'DD_SAML2_METADATA_AUTO_CONF_URL' in os.environ or len(env('DD_SAML2_METADATA_AUTO_CONF_URL')) > 0:`

Tested on my instance.

dsever: Done

madchap (approved):

PR #2810, opened 2020-08-28
api_v2: fix push notes to Jira

xens:
This commit fixes an issue where notes were not pushed to Jira when they were created from the API_V2.


valentijnscholten (approved):
ideally comment creation logic should be more centralized so pushing to jira is managed in one place, but for now this is the practical solution

PR #2815, opened 2020-08-28
jira: improve comments / webhook robustness

xens:
Goal of this commit is to make the DDJ<->Jira comments mechanism more robust.

Some notable changes that have been introduced with this commit:

1. Tracking of Jira comments IDs inside DDJ notes:
   To better handle loops a new field "jira_id" was added to DDJ notes
   so every time a new note is created, either from the UI or the API,
   the jira comment reference will be stored inside the DDJ note.

2. Better handling of comments loops:
   With the help of point 1 we can now properly track webhooks received
   from Jira.

3. Better control of comments creation date:
   We're now using the Jira comment creation time-stamp when storing a note
   coming from Jira inside a DDJ note, this way we avoid out-of-order messages
   inside DDJ notes. Also in case of issues between Jira and the DDJ webhook endpoint
   (network outage, ...) Jira will always retry later to resend comments that have
   not been posted, this way we ensure that the note creation time-stamp is correct.

4. Better error handling:
   A couple of try/except were added to catch common errors, previously DDJ was
   stack-tracing on some function for example if some optional fields were not
   present in the POST of an API_V2 call.


valentijnscholten:
Did you find out why we need to track the jira comment id. It should be sufficient to use Dave with push_to_jira set to false.
I don't think we want to track jus comment id, GitHub comment id, sonarcube comment id, gitlab comment id, etc.
If the timestamp is detailed enough it could be used to check and the content could also be checked.

xens:
> Did you find out why we need to track the jira comment id. It should be sufficient to use Dave with push_to_jira set to false.
> I don't think we want to track jira comment id, GitHub comment id, sonarcube comment id, gitlab comment id, etc.
> If the timestamp is detailed enough it could be used to check and the content could also be checked.

Well after investigation I think that tracking the jira comment id is the safest way to avoid comment loops (DDJ comment -> Jira comment -> Jira webhook -> DDJ comment -> ...). I carefully analysed data coming from Jira webhooks (when posting comments from Jira) and haven't found a better approach. Yes we could analyse the body of the comment + the timestamp but it's probably be less robust and if we want to also support comment deletion (Jira comment -> DDJ) later that's probably our best option.

xens:
I fixed the remaining issues and remove the WIP status.

madchap:
> 1. Tracking of Jira comments IDs inside DDJ notes:
>    To better handle loops a new field "jira_id" was added to DDJ notes
>    so every time a new note is created, either from the UI or the API,
>    the jira comment reference will be stored inside the DDJ note.

I believe it would be better long-term to have a dedicated structure to store meta information and not use notes there, to also set the tone for the future.

Could this be something using your proposal at https://owasp.slack.com/archives/C014H3ZV9U6/p1598967744017100 ?

xens:
Well IMHO "[meta information](https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/models.py#L484)" on a model should be here to store user-specific-use-cases only, like storing a custom key-value attribute that would not make a lot of sense for the upstream-project. So we should extend that logic to not only `products` and `endpoints` but to all the models to give more flexibility to users.

When it comes to this PR and tracking the ticketing-system-ID, it's a bit the opposite; it's should be part of the model as it'll help DDJ better handle external notes (Jira in that case). This filed will be helpful to:

* better control webhooks (avoid loops)
* support external notes deletion
* support external notes edition
* avoid notes duplicates

Maybe something we could do is rename the field `jira_id` to `external_note_id` or something similar, this way we can apply the same logic for other ticketing systems without duplicating the number of fields inside the notes model ?

dojo/jira_link/views.py
```diff
@@ -91,24 +91,30 @@ def webhook(request):
...
+                new_note.entry = '(%s): %s' % (commentor, comment_text)
+                new_note.author, created = User.objects.get_or_create(username='JIRA')
+                new_note.date = jira_note_creation_date
+                new_note.save()
+                finding.notes.add(new_note)
+                finding.jira_change = timezone.now()
+                finding.save()
+                create_notification(event='other', title='JIRA Update - %s' % (jissue.finding), url=reverse("view_finding", args=(jissue.id,)), icon='check')
+            except Exception as e:
+                print("catch all")
```
madchap, line 116: Using `logger` would be better than `print`.

madchap:
> Maybe something we could do is rename the field jira_id to external_note_id or something similar, this way we can apply the same logic for other ticketing systems without duplicating the number of fields inside the notes model ?

Definitely sounds right if we're going with this approach.

madchap:
> Did you find out why we need to track the jira comment id. It should be sufficient to use Dave with push_to_jira set to false.
> I don't think we want to track jira comment id, GitHub comment id, sonarcube comment id, gitlab comment id, etc.
> If the timestamp is detailed enough it could be used to check and the content could also be checked.

At any point in time, there would be only a single ID from either of these (external) systems I assume that could post back a note to DefectDojo.

What's the actual downside of storing an external reference ID in the Notes model, especially if it allows for a more deterministic handling?

valentijnscholten:
Using one external id won't work if someone has 2 tools connected is jira and GitHub.
I think it's just over engineering to start storing all kinds of ids just to be able to post a comment.
I think the jira integration should be lightweight so it can be maintained and be stable. Currently it's complicated and unstable and this will complicate it even more.

xens:
As I understand today it's not possible to have 2 tools connected at the same time from a note perspective as you can't select where to route the note (or maybe I'm missing something) so a single ID should work. But anyway I understand your concern about code complexity and the current (sad) state of the Jira integration. I'll close this PR for now, feel free to re-open it if you've find a better idea on how to implement that and I'll be happy to work on it.

xens:
:zombie: re-opening this PR after discussing with @valentijnscholten over Slack. I'll cherry-pick things that still make sense and rework it a bit (error handling, comments date/time, ...).

I also thought of another approach to handle comment loops [while reading the Jira webhook docuemntation](https://developer.atlassian.com/cloud/jira/platform/webhooks/) without having to maintain the note_id locally;

We could extend what has been done in https://github.com/DefectDojo/django-DefectDojo/pull/2786 by also leveraging the **updateAuthor.emailAddress** field as this field is also available in the webhook payload if the bot/user-profile is configured properly VS forcing users to have an exact match between the Jira username (user for the integration) and the Jira DisplayName which are usually different by default.

Let me know what you think.

valentijnscholten:
Not sure if email address will be reliable as most of the time defect dojo will use some kind of service account and we don't know if people will configure the correct emails address on both sides. A quick look at the JIRA docs shows they are using an accountId identifier for users. We could store that in the jira config (when the credentials are provided, the accountId can be retrieved from JIRA) on defect dojo and then compare on the incoming webhook.

madchap:
So it seems we have real-world issues: https://owasp.slack.com/archives/C2P5BA8MN/p1601478823038900

valentijnscholten:
I found out that the existing logic to detect that an incoming comment from JIRA is actually triggered by Defect Dojo was flawed. Is now fixed in #3606.

valentijnscholten:
> based on the discussion on slack it looks like the change in #2801 is causing the push_to_jira become true when it shouldn't
> 
> ```
>         push_to_jira = getattr(self, 'push_to_jira', push_to_jira)
> ```

This was also fixed in one of my JIRA revamp PRs

damiencarol:
@xens This branch have conflicts now, could you rebase your pull request?

madchap:
Do we actually still need it based on the discussion that aimed to reduce this PR's scope, as https://github.com/DefectDojo/django-DefectDojo/pull/3606 was merged?

damiencarol:
Good question

valentijnscholten:
Closing it for now. The most important problem, the looping, was resolved. Most current requests or problem reports around JIRA are mainly about syncing of statuses or having no sync at all without clear errors message.

PR #2823, opened 2020-09-01
Search fix watson references to deleted findings

valentijnscholten:
I found out that if you delete findings from Defect Dojo, they remain behind in the watson search index. 
It might work if findings are deleted via DD because it will use pre delete signal. But sometimes people cleanup via database queries and run `buildwatson` to rebuild the index. however even after running `buildwatson` to rebuild the index, the search entries remain. So we have to check if objects connected to searchentries still exist before trying to render them to prevent the template from crashing with a 500 error.

I've asked django-watson for info: https://github.com/etianen/django-watson/issues/276

The PR basically adds some ifs here and there, but also some indentation changes it seems.

### DIFF
--- a/dojo/templates/dojo/simple_search.html
+++ b/dojo/templates/dojo/simple_search.html
@@ -83,9 +83,17 @@
                         </thead>
                         <tbody>
                         {% for product in products %}
+                            {% comment %} sometimes the index contains references to findings that no longer exist. even after rebuilding the index. {% endcomment %}
+                            {% if product.object.id %}                        
                             <tr>
-                                <td><a class="search-finding"
-                                       href="{% url 'view_product' product.object.id %}">{{ product.object.name }}</a>
+                                <td>
+                                    {% comment %} it appears sometimes the finding id is missing from the search index, even after rebuilding the watson index {% endcomment %}
+                                    {% if product.object.id %}
+                                        <a class="search-finding"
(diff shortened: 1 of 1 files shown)


dojo/templates/dojo/simple_search.html
```diff
@@ -83,9 +83,17 @@ <h2> {{ name }}</h2>
...
                             <tr>
-                                <td><a class="search-finding"
-                                       href="{% url 'view_product' product.object.id %}">{{ product.object.name }}</a>
+                                <td>
+                                    {% comment %} it appears sometimes the finding id is missing from the search index, even after rebuilding the watson index {% endcomment %}
+                                    {% if product.object.id %}
+                                        <a class="search-finding"
+                                            href="{% url 'view_product' product.object.id %}">{{ product.object.name }}</a>
+                                    {% else %}
+                                        VALENTIJN{{ product.object.name }}
```
Maffooch, line 95: Is this an easter egg? :)

valentijnscholten: haha nice find. looks like my cherry-picking and squashing skills are still on holiday

valentijnscholten:
issue has now been confirmed by watson but no quick bugfix expected, so this PR is still needed.

madchap (approved):

Issue #2713, opened 2020-07-23
Problems repeated notifications in slack


undelete:
**Bug description**
At the moment when an engagement is created or a finding is deleted, notifications are repeated. I have observed this bug since I updated to version 1.7.0, I do not know if it also affects email notifications

**Steps to reproduce**
1. Create engagement or finding delete

**Expected behavior**
A one notification.

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [ x] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: [e.g. Ubuntu 18.04]
 - DefectDojo Commit Message: [use `git show -s --format="[%ci] %h: %s [%d]"`]

**Screenshots** (optional)

<img width="797" alt="Captura de Pantalla 2020-07-22 a la(s) 22 16 15" src="https://user-images.githubusercontent.com/25057634/88247313-505c1380-cc6b-11ea-81f2-46cea7ec3d14.png">

<img width="787" alt="Captura de Pantalla 2020-07-22 a la(s) 22 18 34" src="https://user-images.githubusercontent.com/25057634/88247333-5fdb5c80-cc6b-11ea-8225-3e295ce2a09d.png">

dayal18:
I have seen this issue only when I have notification enabled for both Global and Personal.
Later I have put everything in Global leaving personal as blank and then I get only 1 notification. I enabled email notification.

PR #2834, opened 2020-09-03
Fix notifications spam (issue #2713)

H4ckd4ddy:
Hi,

As described in issue #2713 created by another user, we receive repeated slack notifications.

When a system notification is created, DefectDojo try to send it to all administrators via slack. But if a user is not found in slack and a main slack channel is defined, the notification is redirected to this channel.

So I suggest that you send notifications to admins only if a dedicated channel is not defined.

That's all.

Thanks by advance

### DIFF
--- a/dojo/notifications/helper.py
+++ b/dojo/notifications/helper.py
@@ -10,6 +10,7 @@ logger = logging.getLogger(__name__)
 
 
 def create_notification(event=None, *args, **kwargs):
+    from dojo.utils import get_system_setting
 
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

madchap (approved):
Thanks @H4ckd4ddy :)

valentijnscholten:
This PR has broken any non slack-notifications if slack is configured. Admins are no longer receiving any UI alert for example.

Issue #1906, opened 2020-02-21
FileNotFoundError at /engagement/46/threatmodel/upload


bcncarlesc:
**Bug description**
I've created an engagement and I want add a Threat Modeling document. When I tried it a receive this error

FileNotFoundError at /engagement/46/threatmodel/upload

**Steps to reproduce**
1. Go to 'Product'-->'Engagement-->Upload Threat Modeling
2. Upload any doc 

**Expected behavior*

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [X] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: [e.g. Ubuntu 18.04]
uname -a
Linux ubuntu 5.3.0-28-generic #30~18.04.1-Ubuntu SMP Fri Jan 17 06:14:09 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

 - DefectDojo Commit Message: [use `git show -s --format="%h: %s [%ci]"`]
0d64c07d: Merge pull request #1698 from madchap/fix-1653 [2020-01-14 16:15:14 -0600]

**Sample scan files** (optional)
If applicable, add sample scan files to help reproduce your problem.

**Screenshots** (optional)
![image](https://user-images.githubusercontent.com/51196641/75067178-545bed80-54a1-11ea-9852-df76f8bfc4e9.png)

**Console logs** (optional)
If applicable, add console logs to help explain your problem.
![image](https://user-images.githubusercontent.com/51196641/75067272-80776e80-54a1-11ea-845e-5f59527fd0ca.png)

**Additional context** (optional)
I would like debug with vscode but I don't know how I need configure de projects to do  it. If you have any guide to config it I will try to debug imself

bcncarles:
I  found what is the problem 

I'm working with docker version and when I tried tu upload file the file is upload into docker image django-defectdojo_uwsgi_1 that it doesn't have permition to copy

![imagen](https://user-images.githubusercontent.com/6205426/76230745-746f0900-6224-11ea-839c-13d65b021e72.png)

valentijnscholten:
This is the same issue as with the finding images not working in docker-compose (release) mode.

valentijnscholten:
we need to at least update the docs for 1.7.0

madchap:
Actually, I don't think it is.

If you exec into uwsgi and do a `mkdir /app/media/threat`, then you can upload your model.

[...]

PR #2843, opened 2020-09-05
edit finding: fix parent/original duplicate + small fixes

valentijnscholten:
The parent duplicate status + dropdown on the edit finding page has two (breaking) issues:
image

1) When the page is loaded, the first findingin the dropdown is selected. So when quickly editing a finding to change a field and pressing save leads to the parent/original finding being changed silently

2) On the view finding page duplication can be managed and the finding can be marked as a finding that will not occur on the dropdown above. Basically meaning the finding can no longer be edited without breaking the duplicate relationship.

Solutions:
A) Add the similar finding filtering stuff also to the edit finding page. Could be done but would require either a long dropdown, or some ajaxy filtering and tables.

Feel free to add that, but for now that's too much work. Especially because all the duplication relationships can now be managed quite well from the view_finding page since #1906 

B) Make the parent duplicate (original finding) readonly on the edit_finding page.

This PR implements option B:

for duplicates:
image
image

for non duplicates:
image

PR also includes some small fixes

### DIFF
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -2242,7 +2242,7 @@ def calculate_possible_related_actions_for_similar_finding(request, finding, sim
     else:
         if similar_finding.duplicate_finding:
             # reset duplicate status is always possible
-            actions.append({'action': 'reset_finding_duplicate_status', 'reason': 'This will remove the finding from the cluster, effectively marking it no longer as duplicate'})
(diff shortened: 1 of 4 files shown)


madchap (approved):

Maffooch (approved):
Good catch and solution!

PR #2845, opened 2020-09-07
Set component name/version for php symfony & dependency check parsers

valentijnscholten:
me: "Let me quickly add some code to set `component_name` and `component_version` for some scanners".

10 hours later:

This PR:

- set component name + version for php symfony checker
- set component name + version for owasp dependency checker
- set component name + version for existing findings on reimport scan

update:
This week v6.0.0 from owasp dependency check was released which changed some things. They are backwards compatible, although some optional fields are now only available in a newer format / differently named xml tags.
I have updated the PR to reflect this, i.e. it will handle identifiers like:

```
                <identifiers>
                    <identifier type="cpe" confidence="HIGHEST">
                        <name>cpe:/a:apache:xalan-java:2.7.1</name>
                        <url>https://web.nvd.nist.gov/view/vuln/search-results?adv_search=true&amp;cves=on&amp;cpe_version=cpe%3A%2Fa%3Aapache%3Axalan-java%3A2.7.1</url>
                    </identifier>
                    <identifier type="maven" confidence="HIGHEST">
                        <name>xalan:serializer:2.7.1</name>
                        <url>https://search.maven.org/remotecontent?filepath=xalan/serializer/2.7.1/serializer-2.7.1.jar</url>
                    </identifier>
                </identifiers>

                newly found in v6.0.0
                <identifiers>
                    <package confidence="HIGH">
                        <id>pkg:maven/nl.isaac.client.offerservice/client-offer-service-codegen@1.0-SNAPSHOT</id>
                        <url>https://ossindex.sonatype.org/component/pkg:maven/nl.isaac.client.offerservice/client-offer-service-codegen@1.0-SNAPSHOT</url>
                    </package>
                </identifiers>
```

This PR also now processes all `relatedDependenciess` for a "main" dependencies. These were ignored until now. Sometimes this is correct as dependency check tries to group related dependencies. But most of the time this grouping makes no sense and you are missing vulnerable libraries. So I have chosen to add these also as findings.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -862,12 +862,17 @@ class ReImportScanSerializer(TaggitSerializer, serializers.Serializer):
 
             for item in items:
                 sev = item.severity
+
                 if sev == 'Information' or sev == 'Informational':
                     sev = 'Info'
 
                 if (Finding.SEVERITIES[sev] >
                         Finding.SEVERITIES[min_sev]):
                     continue
+
+                # existing findings may be from before we had component_name/version fields
+                component_name = item.component_name if hasattr(item, 'component_name') else None
+                component_version = item.component_version if hasattr(item, 'component_version') else None
 
                 from titlecase import titlecase
                 item.title = titlecase(item.title)
@@ -894,6 +899,11 @@ class ReImportScanSerializer(TaggitSerializer, serializers.Serializer):
                         finding.mitigated_by = None
                         finding.active = True
                         finding.verified = verified
+
+                        # existing findings may be from before we had component_name/version fields
+                        finding.component_name = finding.component_name if finding.component_name else component_name
+                        finding.component_version = finding.component_version if finding.component_version else component_version
+
                         finding.save()
                         note = Notes(
                             entry="Re-activated by %s re-upload." % scan_type,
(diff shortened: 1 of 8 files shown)


valentijnscholten:
OWASP dependency check v6.0.0 was released this week and it turns out there are some changes in the report format. I'll try to update the parser and include that in this PR.

valentijnscholten:
Also unittests were updated and are running succesfully locally here, but in travis they fail:
```
======================================================================
FAIL: test_parse_finding (dojo.unittests.test_dependency_check_parser.TestDependencyCheckParser)
  File "/app/dojo/unittests/test_dependency_check_parser.py", line 407, in test_parse_finding
    self.assertEqual(expected_references, finding.references)
AssertionError: 'name: Reference Name\\nsource: Reference1\nurl: http://localho[140 chars]\n\n' != 'name: Reference Name\nsource: Reference1\nurl: http://localhos[138 chars]\n\n'
- name: Reference Name\nsource: Reference1
+ name: Reference Name
+ source: Reference1
  url: http://localhost/badvulnerability.htm
  
- name: Reference for a bad vulnerability\nsource: MISC
?                                        --------------
+ name: Reference for a bad vulnerability
+ source: MISC
  url: http://localhost2/reference_for_badvulnerability.pdf
```

Maffooch (approved):

mtesauro (approved):
Approved.

I've been thinking about a more generic way to handle issues like the components stuff you're doing here.  I'll bring that up at the next maintainers meeting.  It would be a post 1.8.0 thing and this PR won't block that effort in future should we decide to go down that path.

PR #2849, opened 2020-09-07
bugfix for risk acceptance migration

alles-klar:
With existing risk acceptances in the database I got some error when running the migrations: psycopg2.errors.NotNullViolation: column "name" contains null values

Setting the default value to a dummy string instead of None fixes this error. Should not have any impact on current installations because the field is renamed by the following instruction anyway.

### DIFF
--- a/dojo/db_migrations/0042_risk_acceptance_improvements.py
+++ b/dojo/db_migrations/0042_risk_acceptance_improvements.py
@@ -22,7 +22,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='risk_acceptance',
             name='name',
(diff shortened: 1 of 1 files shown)


madchap:
Hey @alles-klar,

I think it's better to have a new migration, even for this small change.

alles-klar:
A new migration can't solve the bug.

valentijnscholten (approved):
I think changing the existing migration is the only way it can solve the problem, so let's go with that. For now we plan to release 1.8.0 end of the month.

valentijnscholten:
merging this as it is too important to have lingering around waiting for review.

PR #2851, opened 2020-09-07
[master] Remove heroku info

madchap:


Maffooch:
Thinking we should wait for the next release to touch master branch.

madchap:
Sure. I guess we're not so much in a hurry anyways.

PR #70, opened 2016-03-18
changed xml to defusedxml

devGregA:
would someone please test with a sample file?

PR #2855, opened 2020-09-10
Quantify endpoint status objects when viewing findings

cody-m-tibco:
Add visual representation of how many endpoint status objects are attached to each finding

### DIFF
--- a/dojo/templatetags/get_endpoint_status.py
+++ b/dojo/templatetags/get_endpoint_status.py
@@ -16,6 +16,16 @@ def get_mitigated_endpoints(finding):
 
 
 @register.filter
+def vulnerable_endpoint_count(finding):
+    return finding.endpoint_status.all().filter(mitigated=False).count()
+
+
+@register.filter
+def mitigated_endpoint_count(finding):
+    return finding.endpoint_status.all().filter(mitigated=True).count()
+
+
+@register.filter
 def endpoint_display_status(endpoint, finding):
     status = Endpoint_Status.objects.get(endpoint=endpoint, finding=finding)
     if status.false_positive:
@@ -51,3 +61,9 @@ def endpoint_mitigator(endpoint, finding):
 def endpoint_mitigated_time(endpoint, finding):
     status = Endpoint_Status.objects.get(endpoint=endpoint, finding=finding)
     return status.mitigated_time
+
+
+@register.filter
+def endpoint_check_active(endpoint, finding):
+    status = Endpoint_Status.objects.get(endpoint=endpoint, finding=finding)
+    return not status.mitigated
--- a/dojo/templatetags/get_endpoint_status.py
+++ b/dojo/templatetags/get_endpoint_status.py
@@ -16,16 +16,6 @@ def get_mitigated_endpoints(finding):
 
 
 @register.filter
-def vulnerable_endpoint_count(finding):
-    return finding.endpoint_status.all().filter(mitigated=False).count()
-
-
-@register.filter
-def mitigated_endpoint_count(finding):
-    return finding.endpoint_status.all().filter(mitigated=True).count()
-
-
-@register.filter
 def endpoint_display_status(endpoint, finding):
     status = Endpoint_Status.objects.get(endpoint=endpoint, finding=finding)
     if status.false_positive:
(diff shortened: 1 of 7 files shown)


valentijnscholten:
I like the idea, but the implementation currently performs 2 extra queries for each finding shown in the which can make it very slow. The `prefetch_for_finding` method shows how to prefetch extra data with at most 1 extra query for the whole page. The `prefetch_for_product` shows` how to prefetch counts similar to the ones in this PR. Could you try to prefetch these counts in this PR in a similar way?

cody-m-tibco:
Sure, I'll see what I can do!

cody-m-tibco:
Prefetching is fine and all, but attempting to annotate a count field results in the following:

`Expression #70 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'defectdojo.dojo_jira_issue.id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by`

I suspect this is because the `jira_issue` field is not actually a column in the db, but rather a `finding` class variable. The amount of `endpoint_status` count related queries is still reduced from 2 to 1.

valentijnscholten:
What count prefetch did you try

cody-m-tibco:
Pretty much the same thing from the `prefetch_for_product` function:

`prefetched_findings = prefetched_findings.annotate(active_endpoint_count=Count(endpoint_status__id', filter=Q(endpoint_status__mitigated=False)))`

valentijnscholten:
I think you need to use Prefetch(...) in this case.

cody-m-tibco:
I've got `prefetch('endpoint_status')` in there already, so the counting should be quicker than before.

valentijnscholten:
It won't be because you are querying Endpoint directly in the tag. It could be if you use finding.endpoint_status.all() or something like that but you would have to count in python to avoid queries.

dojo/templatetags/get_endpoint_status.py
```diff
@@ -15,6 +15,16 @@ def get_mitigated_endpoints(finding):
...
+@register.filter
+def vulnerable_endpoint_count(finding):
+    return finding.endpoint_status.all().filter(mitigated=False).count()
+
+
+@register.filter
+def mitigated_endpoint_count(finding):
+    return finding.endpoint_status.all().filter(mitigated=True).count()
+
+
```
cody-m-tibco, line 27: The tags do use `finding.endpoint_status.all()`. Are you saying it would be quicker converting a list and getting the length than one query to the db by using `.count()`?

Unfortunately checking the status of the the `endpoint_status` object requires both the endpoint and the finding, so I'm not sure how to prefetch with that one. Both `endpoints` and `endpoint_status` objects are prefetched, so maybe it works?

valentijnscholten: the filter() will trigger a new query so you would have to filter in python. and also count in python (with len or so) because count would trigger a query even without filter(). This is exactly what happens if you use prefetching, but then for all rows at once and transparantly by django orm.

I just tried it in here, and the prefetch just works for me:

> prefetched_findings = prefetched_findings.annotate(active_endpoint_count=Count(endpoint_status__id', filter=Q(endpoint_status__mitigated=False)))

But then again I don't use endpoitn usually. I added some via the UI but no matter what I do, there are never any entries in the `dojo_finding_endpoint_status`  table, only in the `dojo_endpoint_status` table so the count is always zero.

Maffooch: I wonder if it doesn’t quite work on the API then. Sounds like creation is just fine, but establishing the link between endpoints and findings isn’t quite working.

cody-m-tibco:
Couple things with this last commit
 * Prefetch actually works now. Turns out the the issue I was having is related to `select_related` in comparison with `prefetch_related`. Changing the `jira_issue` to `prefetch_related` worked because it removed it from the GROUP BY part of the query. 
* Created a custom create function for `Endpoint_status` objects that attach them to the supplied `Finding` and `Endpoint` objects.

devGregA (approved):

valentijnscholten (approved):
not sure about the prefetch but let's see

PR #2859, opened 2020-09-11
Move charts to bitnami's repo

madchap:
The stable postgresql chart is deprecated. Moving to bitnami, which is the referred chart.
Bumped up the version of other dependent charts.


madchap:
Seems like the bitnami postgres chart must be behavior a bit different than the @stable chart.

madchap:
So locally (docker minikube), I actually had to delete the `pvc` for the password to get reset and for all to work with the bitnami postgresql... which seems to be yet another error in travis?

ccojocar (approved):
LGTM

ccojocar:
@madchap I think this pull request needs a rebase.

ptrovatelli:
@madchap could you check the conflicts and ping me when ready for review? thanks :)

madchap:
> @madchap could you check the conflicts and ping me when ready for review? thanks :)

Yea, I looked at them and oddly enough, I don't have issues when running locally.
The main issue seems to be switching to the now official/maintained bitnami postgresql.

This is not urgent, I think the current charts were updated to whatever was in `@stable`. I will mark this as draft for now.

madchap:
image

yep, coming from `master` :thinking:

dsever:
@madchap do you want me to add it into  GH Actions?

madchap:
> @madchap do you want me to add it into GH Actions?

Not yet. I have some odd behaviors locally anyways, want to try to iron that out first.

madchap:
So, I've added the `helm repo add` to master, dev and my PR, but somehow, it's still missing it.
Stopping being frustrated for now.

madchap:
@ptrovatelli this is now normally working (had to add a postgres secret to satisfy bitnami's requirement), if you want to take a look.

valentijnscholten:
wdyt @dsever

dsever:
I would like to make to check it more deeply, this can break upgrade.

madchap:
Exchanging with @dsever on this. I will also have more time tomorrow to test a bit further. 
At least now, we know how to make it work.

madchap:
From https://github.com/helm/charts/tree/master/stable/postgresql
image

So, as we override the `postgresqlUsername` with `defectdojo`, the new parameter is also used for the `defectdojo` user.
The `postgres` user still has no password set, as I could see:

```
2020-11-14 14:59:20.723 GMT [465] FATAL:  password authentication failed for user "postgres"
2020-11-14 14:59:20.723 GMT [465] DETAIL:  User "postgres" has no password assigned.
```

Having only `postgresql-postgres-password` does not suffice. It does still need to other key, too. Otherwise

```
Error: couldn't find key postgresql-password in Secret default/defectdojo-postgresql-specific
```

So these passwords actually serve the same purpose. It seems that moving over the bitnami repo, we must set these 2. To the same value then. 

From an upgrade perspective, we would actually have both keys in the postgresql secret, so it should be just fine.

dsever:
@madchap  looks good.

alles-klar:
Looks good but we need to add some documentation, how to add the new secret to existing installations. 

Furthermore redis and rabbitmq seem to be broken. I will create another PR to move them to bitnami too.

alles-klar:
> Looks good but we need to add some documentation, how to add the new secret to existing installations.
> 
> Furthermore redis and rabbitmq seem to be broken. I will create another PR to move them to bitnami too.

Creating another PR is not so easy, because it would have conflict with this one. 

@madchap Can you replace these lines in requirement.yaml and update the .lock file? My own tests doesn't cause any problems.
```
  - name: rabbitmq
    version: 8.0.2
    repository: "@bitnami"
    condition: rabbitmq.enabled
  - name: redis
    version: 12.0.0
    repository: "@bitnami"
    condition: redis.enabled
```

madchap:
Can do, and will rename the PR too.

> Looks good but we need to add some documentation, how to add the new secret to existing installations.

There is nothing to change there, it is the same as before. If there's already a secret, it's taken for both and if not, the same random password is taken for both.

alles-klar:
@madchap If you use the chart with the option createPostgresqlSecret: false because you already have one generated, you get an error that postgresql-postgres-password is missing.

madchap:
> @madchap If you use the chart with the option createPostgresqlSecret: false because you already have one generated, you get an error that postgresql-postgres-password is missing.

Ah right, good catch. So I guess a bit of documentation is indeed needed.

alles-klar (approved):

valentijnscholten:
according to @madchap @alles-klar  we don't need documenation, just a note in the release notes. merging.

valentijnscholten (approved):

Issue #2854, opened 2020-09-10
api_v2: support filtering on metadata for products, endpoints and findings


xens:
**Description**
DefectDojo offers the possibility to create custom key/value entries called metadata in the `products` and `endpoints` models.

```     
"product_meta": [
        {
          "name": "customfield1",
          "value": "value"
        },
        {
          "name": "customfield2",
          "value": "value"
        },
]
``` 
it's a great way to extend the existing models for user-specific use cases, but as-of-today DDJ lacks the ability do filtering on those fields which is quite limiting when dealing with its API.

**Describe the solution you'd like**

As a user I should be able to query the different API endpoints that support these metadata and do filtering based on their key or on their key **and** value, so a query like this

`api/v2/products/?metadata=customfield1` should return only the products containing this key.

Filtering on the key + the value should also be supported

`api/v2/products/?metadata=customfield1:value`

**Limitations**

custom metadata are a great way to add flexibility to an existing model but there's limitations like the lack of a proper type (all fields are strings) so it's not possible to easily do sorting on queries as we can't predict the kind of data the user will store on them.

**Proposal**

If the above description works for you I should be able to start to work on its implementation soon. I'll also open a separate PR to implement the metadata fields on the `/findings/` endpoint.

valentijnscholten:
I think this is useful. Is that ':' construct a common thing, or just something you thought would work well?

xens:
I've spent some time this afternoon reading things on that subject and that's what I've ended up with basically.

There's also this alternative that is quite popular `?metadata[customfield1]=value` and part of the OpenAPIv3 spec (the more I read on the subject the more I'd tend to use that syntax).

refs:
* https://swagger.io/docs/specification/serialization/
* https://github.com/OAI/OpenAPI-Specification/issues/1706
* https://www.moesif.com/blog/technical/api-design/REST-API-Design-Filtering-Sorting-and-Pagination/

ptrovatelli:
seems nice to me. the array quotation seems more standard and intuitive indeed `?metadata[customfield1]=value`

mtesauro:
Seems like a very useful API addition as I've used custom metadata for many different DefectDojo installs.

I also like the `?metadata[customfield1]=value` syntax

valentijnscholten:
Thanks for looking into some syntax that is common / standard / spec-compliant, I think that's the preferred option.

xens:
[...]

PR #2862, opened 2020-09-12
findings: add support for metadata

xens:
This commit adds support for metadata on findings.

This is the first PR of a series linked to https://github.com/DefectDojo/django-DefectDojo/issues/2854


Maffooch:
This looks good! Are you planning a new migration for each PR in the future?

xens:
To be honest I don't know. If there's some best-practices regarding migrations I'd be happy to follow them.

Maffooch:
Not really a best practice, but I would rather have one migration instead of a handful of them. That would require all model changes happen in the same PR though.

xens:
Yeah makes sense, that's also what I did for this PR packing all migrations together.

Maffooch:
hi @xens a PR with the same number migration was merged yesterday. Can you update this one to accommodate?

valentijnscholten (approved):

xens:
> hi @xens a PR with the same number migration was merged yesterday. Can you update this one to accommodate?

@Maffooch rebased and edited ! (looks like that the pipelines are timing out don't know if you can restart them ?)

Maffooch (approved):
Looks like it was just cached as still running, but they are all passing now. Happens to me all the time!

PR #2873, opened 2020-09-15
Bugfixing when the title of a vulnerability is more than 511 characters in SonarQube API

dvelardez:
The length of it is limited. Prevent MySQLdb._exceptions.DataError: (1406, "Data too long for column 'title' at row 1").


Maffooch (approved):

valentijnscholten (approved):

valentijnscholten:
Looks OK. Just for fun, there's a built in function to truncate like this: 
image

Issue #2738, opened 2020-07-28
Notifications related to products should only be sent to authorized (or staff) users


kbonato:
When I create a product I associate a user with it, so that user only has access to that product, but he receives notifications for other products.

valentijnscholten:
what notification settings has the user set, both in settings and product. what kind of user is it, staff user?

kbonato:
The user is only as an active user, and I associated only one product with him, the notifications are only active for him, but this is standard.

kbonato:
![image](https://user-images.githubusercontent.com/26029330/88704020-dfda4a00-d0e3-11ea-83b7-ae649014294b.png)
![image](https://user-images.githubusercontent.com/26029330/88704031-e9fc4880-d0e3-11ea-975d-55286441a07d.png)
the user is xpto1 and is associated with project 1, but receives notifications from 2

valentijnscholten:
What does it show on http://host/notifications/personal for that user? What notifications is he getting? alerts? emails? slack?
What type of event? product_added? scan_added? results_added? Please provide a way to reproduce this as currently we're just guessing what you are seeing/doing.

valentijnscholten:
In the screenshot I see only `product_added`. Is that what the user is seeing? That is expected as product_added applies globally. Only the settings visible on the product details page itself are per product. Maybe this needs to be made more clear or adjusted.

kbonato:
![image](https://user-images.githubusercontent.com/26029330/88793876-eb755180-d173-11ea-95f5-5e11272d917d.png)
![image](https://user-images.githubusercontent.com/26029330/88793883-efa16f00-d173-11ea-9358-52641f65c23b.png)

The problem is that the user is ONLY subscribed to the xpto1 product, and even then receives information from xpto2.
regardless of whether the notifications are active or not, he should only receive the product he is associated with, correct?

valentijnscholten:
It should, but this was not implemented yet in that way. Thanks for reporting.

kbonato:
I thought it was a bug lol, thanks so, I hope it helped!

kbonato:
Good afternoon @valentijnscholten , how are you?

do you have a forecast for when this feature will be implemented?

dayal18:
PR #2876

PR #2876, opened 2020-09-17
Notification Enhancement - Only Authorized users will see notification for their product

manav-nagla:
#2738 (Notifications related to products should only be sent to authorized (or staff) users)
Added verification checks on the notification, so only authorized user or staff would see the notifications.


manav-nagla:
Merging again.. Looks like I messed up again with Git..

dayal18:
Ok this time I got it right #2877

PR #2879, opened 2020-09-17
Veracode import: fixed false-positives and added SCA findings

reinier-vegter:
The veracode parser didn't fully work: false positives didn't get detected as such.
Since we're just using the 'detailed XML report's from veracode, it seems their structure was changed a bit (or it never actually worked).

In any case, I decided to refactor the whole parser, split some code, got rid of some deeply nested loops (using xpath) etc.
On the fly, some more details (loc, sast-source) were added to the report.

In addition, as veracode now produces SCA findings in that same report, I added parsing for that as well:
Screen Shot 2020-09-17 at 18 26 45 PM

As for a unit test: there wasn't one, and I didn't add one, since I have already been spending quite some of my clients time on this.

Manually tested cases:

-   import a fresh report with both SAST and SCA findings from veracode: all findings imported with correct data
-   import a second report (same project, same build), but now with a finding marked `false positive` and `mitigation accepted`: finding marked as mitigated and FP in dojo
-   import a third report, now with a build that fixed a finding: new finding was recorded as inactive/mitigated

Please note that the dupe_key for SAST findings is still the same to have that backwards compatible.

### DIFF
--- a/dojo/tools/veracode/parser.py
+++ b/dojo/tools/veracode/parser.py
@@ -5,6 +5,10 @@ from dojo.models import Finding
 from hashlib import sha256
 
 
+'''
+This parser is written for Veracode Detailed XML reports, version 1.5.
+Version is annotated in the report, `detailedreport/@report_format_version`.
+'''
 class VeracodeXMLParser(object):
     ns = {'x': 'https://www.veracode.com/schema/reports/export/1.0'}
     vc_severity_mapping = {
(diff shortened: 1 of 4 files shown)


valentijnscholten:
Thanks for the PR. For us it's hard to review/approve without any tests. The unittests for parsers are pretty basic, could you look at wrapping one up similar to https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/unittests/test_acunetix_parser.py

reinier-vegter:
Hi @valentijnscholten, just added one. Hope it's enough to merge this.
Thanks for reviewing.

valentijnscholten (approved):
Looks good, thanks!

Maffooch (approved):

PR #2883, opened 2020-09-18
Helm run changes

madchap:
Related to how docker-compose has to run the celery beat container as root, do the same for helm. 

Otherwise, the pod exists with a `permission denied` when trying to drop the app privileges.

I have not found any alternative to that. Either run the pod as root and drop the app privileges, or run the pod as non-root and run the app as root.


dsever:
Can this be added in dev Asap, without it, k8s helm is useless

PR #2887, opened 2020-09-20
Adding a parser for Checkov scanner

StefanFl:


StefanFl:
I couldn't find how to set a label when issuing the pull request. It should be "Import Scans (for new scanners/importers)"

valentijnscholten:
Any data available to set `component_name`? maybe use `resource` for that?

StefanFl:
Good point. After a second look in some actual reports and their sources, I removed source and sink completely. Only location, line number and component name are used. The vulnerability is about a whole definition in a terraform or cloudformation file. The end of a definition is easily visible in the source, so mention it in the finding doesn't add any value. Now the finding is less crowded.

madchap (approved):
Thanks for the parser :) There a couple conflicts to fix now, due to other merges. Cheers.

StefanFl:
Conflicts are resolved

valentijnscholten (approved):

valentijnscholten:
https://github.com/DefectDojo/Documentation/pull/127

schosterbarak:
This is awesome! thanks @StefanFl  (p.s. if you have any feedback on checkov feel free to send them my way :) )

Issue #2700, opened 2020-07-20
notifications get sent to users that are no longer authorized for a product


valentijnscholten:
**Bug description**
if a user once was an authorized user for a product and has configured notifications, the user will still receive those even after the user would have been removed as authorized user from the product. 

I haven't tested this, but as I wrote the code I think this can happen :-)

kbonato:
I am having a similar bug, a user who is not in other products he receives notifications normally, but cannot access.

valentijnscholten:
is he a staff user? staff users have another set of notifications in global settings.

kbonato:
No, active user normal

dayal18:
I have fixed this in #2877

PR #2877, opened 2020-09-17
Notification Enhancements - Only Authorized users will see notification for Product / Engagement / Test - #2738

manav-nagla:
#2738 Added checks to look for authorized user in notifications.

PR #2889, opened 2020-09-20
(jira) alerts: improve error handling and more

### LINKED ISSUE #2702: JIRA creation issue: no "priority" field
**Bug description**

When the target jira project does not have the priority field in the screen:

```
celeryworker_1  | [2020-07-21 10:20:25,542: WARNING/ForkPoolWorker-38] jira.exceptions.JIRAError: JiraError HTTP 400 url: https://cloudbees.atlassian.net/rest/api/2/issue
celeryworker_1  | 	text: Field 'priority' cannot be set. It is not on the appropriate screen, or unknown.
```

### LINKED ISSUE #2738: Notifications related to products should only be sent to authorized (or staff) users

### LINKED ISSUE #2888: non-slack notifications/alerts no longer sent to admins
In the past all (system) alerts/notifications were sent to all admin/staff users so they could keep an eye on what was going on, i.e. errors in the JIRA integration happening in the celery background process.

In #2834 a change was implemented to prevent duplicate notifications being sent to slack channels/users. A unintended side effect of this is that if a DD instance has a global slack channel configured, all non-slack notifications are no longer sent to admin users.

I am not saying sending the notifications to all admins is the best solution, but it used to be like this until that PR so this change may confuse some (admin) users.

Might be better not to send the notification to all admins, but only create 1 alerts *without* a user_id. These alerts can be seen as system alerts that can be viewed/cleared by all admins. Not sure how that would work for email notifications, etc. Might need to use a central system email recipient for that? Similar to the system/global slack channel?

### LINKED ISSUE #2890: Users not seeing JIRA errors in UI alerts
JIRA interactions are done in the celery background process. If an error occurs, a dojo alert is created, which is shown in the UI. This isn't always working as expected for admins (see 2888). But also for regular users this is not working.

### LINKED ISSUE #2894: block_execution settings used from reporter instead of user operating defect dojo
Each user has a setting in their profile named `block_execution`. By default this is False and most expensive actions triggered by the user are handle in the background by celery (i.e. dedupe, push to jira, apply rules, ...).

When saving a finding, the current user is not nown in the `finding.save()` method. As a workaround the `block_execution` setting from the `reporter` is used. This can lead to confusing situations where actions are sometimes async and sometimes synchronous. This can lead to users missing errors in the jira integration, missing alerts, etc.

Possible solutions:
A) pass on the `request` or `request.user` object everywhere in the (kw)args.
B) start layering like a (Java Enterprise) boss with all kinds of DAO objects
C) use a middleware like `django-crum`: https://pythonhosted.org/django-crum/

valentijnscholten:
Various fixes and improvements around alerts. Mainly to make sure all JIRA errors are generating the correct alerts in the UI:
- make sure dojo alerts are generated also for non admin users fixes #2890 
- make sure admins get non-slack alerts again fixes #2888 
- make sure only admins or authorized users get notifications for their products fixes #2738 and #2700 
- only set priority and label fields in jira if they exist in jira fixes #2702 
- unify the way we check for `block_execution` fixes #2894
- add title to alerts popup in UI so hovering over it shows details of the alert
- remove 'migrate alerts' function as all urls are relative since a long time
- turns out all the functions in `tasks.py` that just wrap another function to make them async are not needed. The celery decorator can be used on the functions performing the logic themselves.
- log all jira errors also to logfile to aid in troubleshooting


Maffooch (approved):
Did not deploy, but everything looks good. Looks like it will remove the need for #2877, but I would rather this one because of the overhaul instead of the bandage. Great work here.

madchap (approved):
lgtm. There is a small conflict to fix before merging, will let you do that @valentijnscholten :)

PR #2895, opened 2020-09-21
Revert "build(deps): bump python from `09bb817` to `e525911`"

Maffooch:
Reverts DefectDojo/django-DefectDojo#2891

Cannot build because of unsigned signatures.

image


Maffooch:
Merging like a psycho

valentijnscholten:
may have been a temporarity things as it seems to build OK here and in travis?

PR #2909, opened 2020-09-24
Veracode: use unique_id_from_tool (fixed deduplication issues)

reinier-vegter:
I added the use of `unique_id_from_tool` during import, so that deduplication can be much more effective.
We had problems since the description-field was part of the hashing, but it changed sometimes, for example when a component has the version in it's name, highly dependent on how teams implement their CI/CD.

For SAST findings, we rely now on the app_id and issue_id from the report.
For SCA, it's vendor, library and version.

edit: Also moved the variable stuff (that component name) and veracode issue id to the references section, which should keep the description field a bit more static.

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -768,7 +768,7 @@ DEDUPLICATION_ALGORITHM_PER_PARSER = {
     'PHP Symfony Security Check': DEDUPE_ALGO_HASH_CODE,
     'Clair Scan': DEDUPE_ALGO_HASH_CODE,
     'Clair Klar Scan': DEDUPE_ALGO_HASH_CODE,
-    'Veracode Scan': DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL,
+    'Veracode Scan': DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE,
     # for backwards compatibility because someone decided to rename this scanner:
     'Symfony Security Check': DEDUPE_ALGO_HASH_CODE,
     'DSOP Scan': DEDUPE_ALGO_HASH_CODE,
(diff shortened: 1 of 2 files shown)


valentijnscholten:
The DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL algorithm basically prevents any deduplication from occurring across different scanner (veracode, checkmark, sonar, ....). So not sure if you want, or that it should be the default.
I haven't used it yet, but DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE might be a better option to have as default because it will allow deduplication across different scanners @ptrovatelli ?

reinier-vegter:
Hey, I agree, but then again, in the default settings DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE doesn't seem to be used by others, so figured this was the way to go.

BTW, does a change to the default settings impact existing installations?

reinier-vegter:
kudos BTW that you guys respond so swiftly 👍

ptrovatelli:
using the UNIQUE_ID_FROM_TOOL for dedupe will indeed make it impossible to dedupe across parsers but in most cases this is already not working. If it is working currently, yes it's better to use DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE  instead. it's not used yet only for lack of taking time to test which parsers are "dedupe compatible" ie generating the exact same  hash_code values. 

Given that we're currently using the default legacy algo, the cross-parser dedupe is probably not working (it would need standard elements like cwe or cve when building hash_code instead of title to have a chance at working)

Note that changing the dedupe algo will not update previously imported findings. to be able to dedupe new findings against previously imported findings in that case, one would need to update the full database unique_id_from_tool fields for that parser. there a piece of code to update the hash_codes and recompute duplication but it's not updating the unique_id_from_tool; that would require a specific processing. 

FYI for that hash_code / dedup full db dedupe recompute:
```
docker exec -it django-defectdojo_uwsgi_1 bash
./manage.py dedupe
```

valentijnscholten:
Could you try DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE and see how it goes? 
Changing the default in settings.dist.py may affect existing installs if users are not creating their own copy of settings.dist.py as settings.py

reinier-vegter:
done

valentijnscholten:
Is it working OK? In that case we can approve

reinier-vegter:
Deduplication works for us with this setting

valentijnscholten (approved):

Maffooch (approved):

PR #2917, opened 2020-09-26
add flake8 check as github action

valentijnscholten:
I wanted to do some testing with GitHub Actions, as they are now free for public repositories.

The idea: "Let me quickly add a flake8 action/check"

Reality: There are N different actions all doing something with flake8 on PRs. None of them did what I wanted, or were failing in some way with errors.

Solution: As any real developer, I created the N+1-th flake8 GitHub action :-D

Honestly I just forked an existing one, but made it work better. Some of the others were posting all the flake8 violations as comments on the PR. But that was cluttering too much, so this one annotates the source code in the "Files Changed" tab.
So you can see exactly where the violations are. The annotations will just dissappear when they are fixed. See example image below.

I didn't make this GitHub action a required check before a PR can be merged. But if we like it enough we could do so, maybe even replacing the flake8 check in travis.

image

### DIFF
--- a/.github/workflows/flake8-your-pr.yml
+++ b/.github/workflows/flake8-your-pr.yml
@@ -1,0 +1,14 @@
+name: Flake8 your PR
+on: [pull_request]
+jobs:
+  flake8-your-pr:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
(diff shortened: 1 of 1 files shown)


valentijnscholten:
force merging as it will only be triggered by github if the workflow is present on the base branch which in this case is dev

valentijnscholten:
example run: https://github.com/DefectDojo/django-DefectDojo/pull/2915/checks?check_run_id=1170782555

PR #2918, opened 2020-09-26
add GitHub action for integration tests

valentijnscholten:
Add Integration Tests as GitHub action as an experiment. GitHub actions integrate nicely with GitHub of course and are supposed to support caching of docker layers. This caching is currently not working, but still interesting to see how we like this as a Git Hub action.

I haven't made it a required check for merging PRs, but we could in the future.


valentijnscholten:
It won't run in this PR itself because the action runs in the context of the base branch which is dev. And dev doesn't have this workflow file yet. So just approve and merge please ;-)

PR #2946, opened 2020-10-01
Don't copy inexistent certs folder in Dockerfile.django

uncycler:
When building Dockerfile.django with DOCKER_BUILDKIT=1 option in docker, the build fails with:

failed to solve with frontend dockerfile.v0: failed to build LLB: lstat /var/lib/docker/tmp/buildkit-mount548069851/certs: no such file or directory

- [ ] Your code is python 3.6 compliant (specific python >3.6 syntax is currently not accepted).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,7 +116,8 @@ Pipfile*
 
 
 #ignore locally added certs
-certs/
+docker/certs/*
+!docker/certs/readme.txt
 
 # Helm dependencies
 helm/defectdojo/charts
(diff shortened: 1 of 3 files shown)


dsever:
@uncycler I added this option intentionally, see https://github.com/DefectDojo/Documentation/blob/master/docs/social-authentication.rst#saml-20 

Idea is to enable generic way of inserting of private CA serts that are not included inside of certifi

uncycler:
@dsever Maybe just adding a certs directory with a readme.txt. We could put it inside /docker so it don't clobber the root. What do you think?

dsever:
You mean just to have some content, not to break the build? If yes than I can add readme.txt and some instructions to it.

uncycler:
I update the PR to add a placeholder.

madchap (approved):
lgtm.

valentijnscholten (approved):
ah great, I was thinking about a placeholder also but didn't see the PR was already updated with that. approving..

PR #2949, opened 2020-10-01
Add related fields on findings

xens:
Goal of this commit is to add some relevant info inside findings that are external to the `Findings` model. For now I identified: `test`, `engagement`, `product` and `environment` but more can be added later. Some external fields in the current model (foreign keys) can also later be moved to follow this logic (jira_id, notes, ...). To display these related fields a query-param must be specified (?related_fields). This logic can also further be extended to select which field the API must return (?related_fields=a,b,c).

This feature has been discussed there https://owasp.slack.com/archives/C014H3ZV9U6/p1601305649026400


valentijnscholten:
Thanks for the PR, I may actually have a need for this feature :-)
Any reason why you are lowercasing all the request parameters before checking?
I think you may want to override `get_query_set` to prefetch these relatedfields as otherwise the performance will suffer when retrieving lists of findings (search codebase for prefatch_related for examples)

xens:
ok make sense let me check how to achieve this. Regarding the case it wanted to make the parameters case insensitive as it's kinda best-practice, but I don't know if we have a different strategy in DDJ ?

valentijnscholten:
I don't think we're doing it anywhere. It's ok if it would be effecient, but now all parameters are converted into lowercase just to check this one parameter.

xens:
@valentijnscholten not sure how I can test that the prefetch is working as expected but I wanted to do seomthing like this 
 https://github.com/DefectDojo/django-DefectDojo/blob/c4d78aaa751653720cb7a20382b9add08cf5679a/dojo/api_v2/views.py#L262

```diff
- return Finding.objects.all()
+ return Finding.objects.all().prefetch_related('test')
```

If I'm not in the right direction any pointers are welcome I'm sweating a bit on this prefetch thing :)

valentijnscholten:
yeah, the prefetch stuff is quite complicated, took me a couple of nights to understand it. but without DD can become quite slow quite easily if you have 1000s of findings.
you could try to just to `return prefetch_related_for_findings(Findings.objects.all())` by importing that prefetch function from the `finding/views.py`. 
Plan B would be to add the prefetches in the api view like you said, but include all the models referenced by your related fields, so 
.prefetch_related('test', 'test__engagement', 'test__enagement__product', 'test_engagement__product_prod_type').

I think product type might be a good addition to the related fields as it completes the chain all the way from finding to product type.

xens:
thanks for the pointers that's helpful, and agree for the product_type, I've added it. 
For the prefetch I used your suggestion B, but I guess that I'll need to do it also on the other part of the condition where the user is not staff. Let me know what you think, maybe it's DRYer to do it solution A.

valentijnscholten:
yeah instead of returning directly Finding...... you can assign it to a variable findings = Finding.objects.....
and then outside the if do return findings.prefetch_related(.......)

xens:
> yeah instead of returning directly Finding...... you can assign it to a variable findings = Finding.objects.....
> and then outside the if do return findings.prefetch_related(.......)

Yeah that's very obvious looks (tiring week.... ;))

dojo/api_v2/views.py
```diff
@@ -256,10 +256,17 @@ def perform_update(self, serializer):
 
     def get_queryset(self):
         if not self.request.user.is_staff:
-            return Finding.objects.filter(
+            finding = Finding.objects.filter(
                 reporter_id__in=[self.request.user])
         else:
-            return Finding.objects.all()
+            finding = Finding.objects.all()
+
```
valentijnscholten, line 263: sorry, but let's call it `findings` as it is usually multiple findings :-)

valentijnscholten (changes requested):

valentijnscholten (approved):

Maffooch (approved):

PR #2951, opened 2020-10-02
jira: fix looping when pushing from api

### LINKED ISSUE #2948: Transitions in JIRA will result in an endless loop of updates
Bug description
An update to a finding that will result in a transition to a JIRA issue also ends up in an endless loop

**Steps to reproduce**
1. Create a finding in a product with push_to_JIRA set.
2. From the API send 
{"push_to_jira": "true",
  "active": "false",
  "verified": "true",
  "is_Mitigated": "true"}
To patch that finding. 
3. DDJ will loop in updates to JIRA to try to update a ticket that has already been done

**Expected behavior**
DDJ will update the JIRA ticket once only

- [ ] Kubernetes
- [X] Docker

**Environment information**
DDJ version 1.8.0. Installed with docker compose.

**Screenshots** (optional)

**Console logs** (optional)

O.K.  I started this thread on https://owasp.slack.com/archives/C2P5BA8MN/p1601478823038900, but the discussion really belongs here.  So, I am continuing here.  When I started this, I really didn’t expect to be digging into code.  After a lot of debugging, here is where I am and a couple of questions that I have.
First, the setup:
I have a finding that I am trying to close with the API.  So I am pushing to the API.
{ “push_to_jira”: “true”,
  “active”: “false”,
  “verified”: “true”,
  “is_Mitigated”: “true”}
When it saves it gets processed by finding.save() and then gets sent to function update_jira_issue.  the update_jira_issue function calls find.save several times.  At least twice when it there is a state transition for JIRA. Once around line 1610 and another time after the transition is complete.  When find.save() is called, it doesn’t specifically state that push_to_jira is true, but the object still has push_to_jira set, so .save() will overwrite the default value of push_to_jira (~ 6 lines into the function).  This happens 2 per call to update_jira_issue.  Now the code is running in a loop.
Questions:
If push_to_jira is specifically called out when the .save() functions is called, should the value in the object be overwriting it?
Should the value of find.push_to_jira be set to false by update_jira_issue before calling save again? (edited) 

See Slack Chats.
https://owasp.slack.com/archives/C014H3ZV9U6/p1601570100039600?thread_ts=1601563511.038000&cid=C014H3ZV9U6
https://owasp.slack.com/archives/C2P5BA8MN/p1601478823038900

valentijnscholten:
Fixes #2948 

When adding jira_push_all and push_to_jira to the api, there was a bug introduced by passing on the push_to_jira as a model field instead of a parameter. This had no downside yet for us, until in #2801 code was added that was trying to use this field to fix another bug where push_to_jira wasn't passed for PATCH (update) requests. That caused loops to occur:

findings.save -> update_jira_issue -> finding.save -> update_jira_issue -> finding.save -> ... (#2948)

This PR fixes this by:
- reverting the changes in #2801 
- fixing the original bug in the serializer.update()

According to the django docs the current of handling push_to_jira in the api is not how it was envisioned, and a get_serializer_context should be used. But that had other side effects when I tried it. So I decided to leave as much as possible in place and just fix the bug(s).

Here and there I changed some code / names to make it more readable for the future.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -23,7 +23,10 @@ import datetime
 import six
 from django.utils.translation import ugettext_lazy as _
 import json
+import logging
 
+
+logger = logging.getLogger(__name__)
 
 class TagList(list):
     def __init__(self, *args, **kwargs):
@@ -636,6 +639,26 @@ class FindingSerializer(TaggitSerializer, serializers.ModelSerializer):
         model = Finding
         fields = '__all__'
 
+    # Overriding this to push add Push to JIRA functionality
+    def update(self, instance, validated_data):
+        # remove tags from validated data and store them seperately
+        to_be_tagged, validated_data = self._pop_tags(validated_data)
+
+        # pop push_to_jira so it won't get send to the model as a field
+        push_to_jira = validated_data.pop('push_to_jira') or instance.get_push_all_to_jira()
+
+        instance = super(TaggitSerializer, self).update(instance, validated_data)
+        # No need to save the finding twice if we're not pushing to JIRA
+        if push_to_jira:
+            instance.save(push_to_jira=push_to_jira)
+        else:
+            instance.save()
+        
+        # not sure why we are returning a tag_object, but don't want to change too much now as we're just fixing a bug        #         
+        tag_object = self._save_tags(instance, to_be_tagged)
+        return tag_object
+        pass
+
     def validate(self, data):
         if self.context['request'].method == 'PATCH':
             is_active = data.get('active', self.instance.active)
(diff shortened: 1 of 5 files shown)


madchap:
fyi @xens

dojo/api_v2/serializers.py
```diff
@@ -636,6 +640,26 @@ class Meta:
...
+
+        # If we need to push to JIRA, an extra save call is needed.
+        # TODO try to combine create and save, but for now I'm just fixing a bug and don't want to change to much
+        if push_to_jira:
+            instance.save(push_to_jira=push_to_jira)
+
+        # not sure why we are returning a tag_object, but don't want to change too much now as we're just fixing a bug
+        tag_object = self._save_tags(instance, to_be_tagged)
+        return tag_object
+        pass
```
madchap, line 661: pass not needed

valentijnscholten: it's old code, didn't want to change too much

dojo/api_v2/serializers.py
```diff
@@ -698,21 +722,25 @@ class Meta:
...
         if push_to_jira:
-            # Saving again with push_to_jira context
-            tag_object.save(push_to_jira=push_to_jira)
-        return self._save_tags(tag_object, to_be_tagged)
+            new_finding.save(push_to_jira=push_to_jira)
+
+        # not sure why we are returning a tag_object, but don't want to change too much now as we're just fixing a bug
+        tag_object = self._save_tags(new_finding, to_be_tagged)
+        return tag_object
         pass
```
madchap, line 744: same here, no need for pass (and I've seen this in many other places)

valentijnscholten:
@kareem-DA could you look at this or maybe even test if this fixes the loop you encountered?

kareem-DA:
> @kareem-DA could you look at this or maybe even test if this fixes the loop you encountered?

I made the changes to my DEV system tonight.  Looks like it is working as expected.  Thanks for the fast fix.

xens:
@valentijnscholten thanks for digging into this, as I understand by fixing the original bug in the serializer.update() my patch in #2801 is not needed anymore thus PATCH method should still work.

valentijnscholten:
1 approval fro mods + 1 confirmed succesfull test + 1 code review by the original raiser of the issue -> merging.

PR #2952, opened 2020-10-02
fix/optimize search and autocomplete

valentijnscholten:
Search / autocomplete related:
- optimized queries for autocomplete to run without ordering
- add index on `component_name` field
- limit max results for autocomplete to 20000 to avoid 5MB javascript snippets (DD_MAX_AUTOCOMPLETE_WORDS, MAX_AUTOCOMPLETE_WORDS in settings.py)

PR #2955, opened 2020-10-03
Fix cb search

valentijnscholten:
New PR because the previous one (#2952) was messed up by the merge/push problem we had on the dev branch.

Search / autocomplete related:
- optimized queries for autocomplete to run without ordering
- add index on `component_name` field
- limit max results for autocomplete to 20000 to avoid 5MB javascript snippets (DD_MAX_AUTOCOMPLETE_WORDS, MAX_AUTOCOMPLETE_WORDS in settings.py)

I think with the index it still performs OK even with 750k findings, but probably not usefull in that case to send 5MB of javascript to the client. Might need a better autocomplete in the future, ajax based.


madchap (approved):

dojo/settings/settings.dist.py
```diff
@@ -135,7 +135,8 @@
     DD_SLA_NOTIFY_POST_BREACH=(int, 7),
 
     # maximum number of result in search as search can be an expensive operation
-    DD_SEARCH_MAX_RESULTS=(int, 100)
+    DD_SEARCH_MAX_RESULTS=(int, 100),
+    DD_MAX_AUTOCOMPLETE_WORDS=(int, 20000)
```
madchap, line 139: That still seems like a lot? 5000 would still look reasonable?

valentijnscholten: Difference in performance is almost zero between 5kand 20k. 20k seemed to me like a nice seeing that will cover 90% of installs (except yours 🤭)

madchap:
Just pulled it locally over a 750K finding instance, works much better, thanks :-)

Maffooch (approved):

PR #2959, opened 2020-10-04
more tests for import and deduplication

valentijnscholten:
Deduplication in defect dojo is not always clear in how it works. Especially with the mix between the legacy algorithm and the newer "per parser" configurations. The legacy algorithm also has some interesting dynamics around endpoints and static versus dynamic findings. 
These unittests will help us to capture the current behaviour and understand things better. Ideally someone turns these into documentation (hint hint).

Another "issue" is that the import is calling `finding.save()` two times and inside the `finding.save()` method there's another call to `self.save()`. So in total 4 calls to save 1 finding. Which is isn't efficient, especially if you count in that this means 4x audit log (6-8 queries), 4 times calculate product rating, 4 times async history, etc.
These unittests allow use to change / improve this without affecting behaviour / breaking functionality.
If we change functionality, we are forced to update the tests to make us and everyone else aware of the impact of the changes.

- add synthetic tests for `finding.save()`
- add apiv2 tests for import->reimport
- add apiv2 tests for import->import (these are already covered by the integration test that uses the UI, so good enough for now)

This was a massive undertaking, I'd like to thank @madchap for his support ;-)

I also threw in some bonus tests:
- false positive history

I don't fully understand this feature and its usefulness, but we now have some tests to see what it does. I think there are some "issues" that need fixing, but that's a different discussion.


madchap (approved):
Tremendous work.

Maffooch (approved):

PR #2961, opened 2020-10-04
finding: document all the fields in the model

xens:
Goal of this commit is to add a proper description for all the fields in the `finding` model. Swagger will then be able to generate the proper documentation automatically.


xens:
I also updated the documentation at the same time -> https://github.com/DefectDojo/Documentation/pull/132/files. I still miss information for fields `param` and `thread_id` but I have no idea what it is,  maybe we can merge without them and see later. There's also some fields that are unused or deprecated that we should plan to remove somehow (`line_number`, ...) ?

valentijnscholten:
Changing fields will require a migration (another concept to get familiar with ;-)).
Also not sure, but I think changing the verbose name will affect what is displayed so something to check. Otherwise everything will look different in lots of places :-)

xens:
Are you sure that it'll require a migration to set `verbose_name` and `help_text` ? (maybe I have a way to check it locally ?)

Swagger-side the `Title` and the `Description` are now properly populated:

image

valentijnscholten:
yes, it does. you'll see if you run the unittest docker compose env. or the output of travis.

xens:
Ok good to know, and I guess that it's quite logical as these fields metadata are probably also stored on the DB... I guess that'll need to check for all these fields what the impact is on the frontend, see if there's changes ? If there's something else I need to check/do let me know ! (apart from adding the migration file).

xens:
Ok here's what it looks like on the frontend, the fields name on the left took the `verbose_name` and the fields with the `?` now have the text coming from `help_text`:
image

Previously it looked like this:

image

When it comes to terminology the documentation (ReadTheDocs and the model) sometimes differed a bit (flaw VS finding). I tried to do my best but we might want to make sure that it's used in the appropriate context for accuracy / consistency.

xens:
And what it looks like in the findings search view:

image

dojo/models.py
```diff
@@ -1413,101 +1413,288 @@ class Finding(models.Model):
...
+    under_review = models.BooleanField(default=False,
+                                       verbose_name="Under Review",
+                                       help_text="Denotes is this flaw is currently being reviewed.")
+    review_requested_by = models.ForeignKey(Dojo_User,
+                                            null=True,
+                                            blank=True,
+                                            related_name='review_requested_by',
+                                            on_delete=models.CASCADE,
+                                            verbose_name="Review Requested By",
+                                            help_text="Document who requested a review for this finding.")
```
madchap, line 1519: `Documents` if you want to remain aligned with the other sentences ;-)

dojo/models.py
```diff
@@ -1413,101 +1413,288 @@ class Finding(models.Model):
...
+    reviewers = models.ManyToManyField(User,
+                                       blank=True,
+                                       verbose_name="Reviewers",
+                                       help_text="Document who reviewed the flaw.")
```
madchap, line 1523: Same here.

dojo/models.py
```diff
@@ -1413,101 +1413,288 @@ class Finding(models.Model):
...
+    under_defect_review = models.BooleanField(default=False,
+                                              verbose_name="Under Defect Review",
+                                              help_text="Denotes if this finding is under defect review.")
+    defect_review_requested_by = models.ForeignKey(Dojo_User,
+                                                   null=True,
+                                                   blank=True,
+                                                   related_name='defect_review_requested_by',
+                                                   on_delete=models.CASCADE,
+                                                   verbose_name="Defect Review Requested By",
+                                                   help_text="Document who requested a defect review for this flaw.")
```
madchap, line 1535: Same here.

dojo/models.py
```diff
@@ -1413,101 +1413,288 @@ class Finding(models.Model):
...
+                                     blank=True,
+                                     verbose_name="Mitigated",
+                                     help_text="Denotes if this flaw has been fixed, by storing the date it was fixed.")
+    mitigated_by = models.ForeignKey(User,
+                                     null=True,
+                                     editable=False,
+                                     related_name="mitigated_by",
+                                     on_delete=models.CASCADE,
+                                     verbose_name="Mitigated By",
+                                     help_text="Documents who has deemed this flaw as fixed.")
```
madchap, line 1553: `marked` instead of `deemed` seems a better choice of word.

dojo/models.py
```diff
@@ -1413,101 +1413,288 @@ class Finding(models.Model):
...
+                                       verbose_name="Is Mitigated",
+                                       help_text="Denotes if this flaw has been fixed.")
+    thread_id = models.IntegerField(default=0,
+                                    editable=False,
+                                    verbose_name="Thread ID")
+    mitigated = models.DateTimeField(editable=False,
+                                     null=True,
+                                     blank=True,
+                                     verbose_name="Mitigated",
+                                     help_text="Denotes if this flaw has been fixed, by storing the date it was fixed.")
```
madchap, line 1546: Remove comma after `fixed`.

madchap:
Thanks for taking the time to go through this. In-app documentation is important -- as no sane engineer would read the "real" documentation anyways :smile:

xens:
Thanks for the review @madchap. I've maked the PR as DRAFT for now and I'll generate the db migration when all the conversation will be resolved.

dojo/models.py
```diff
@@ -1413,101 +1413,288 @@ class Finding(models.Model):
...
+                                 help_text="A hash over a configureable set of fields that is used for findings deduplication.")
+    line = models.IntegerField(null=True,
+                               blank=True,
                                verbose_name="Line number",
-                               help_text="Source line number of the attack vector")
-    sast_source_file_path = models.CharField(null=True, blank=True, max_length=4000, help_text="Source filepath of the attack vector")
-    nb_occurences = models.IntegerField(null=True, blank=True,
-                               verbose_name="Number of occurences",
-                               help_text="Number of occurences in the source tool when several vulnerabilites were found and aggregated by the scanner")
+                               help_text="Line number. For SAST, when source (start of the attack vector) and sink (end of the attack vector) information are available, put sink information here.")
```
xens, line 1620: This one is a bit weird

dojo/models.py
```diff
@@ -1413,101 +1413,288 @@ class Finding(models.Model):
...
+    file_path = models.CharField(null=True,
+                                 blank=True,
+                                 max_length=4000,
+                                 verbose_name="File path",
+                                 help_text="File name with path. For SAST, when source (start of the attack vector) and sink (end of the attack vector) information are available, put sink information here.")
```
xens, line 1625: This one as well

dojo/models.py
madchap: I thought the original one was ok? (Note, I/tools I use never really used that field)

dojo/models.py
madchap: A real like example with Anchore would be `/usr/share/webapp/webapp.war:WEB-INF/plugins/some-plugins.hpi:WEB-INF/lib/jackson-databind-2.7.3.jar`

In the UI, the header is "Location".

Maybe something simpler like "Identified file(s) containing the vulnerability"?

valentijnscholten:
Just to be sure: You can create draft pull request "officially": https://github.blog/2019-02-14-introducing-draft-pull-requests/

dojo/models.py
Maffooch: I think the original is fine here too.

dojo/models.py
Maffooch: +1 on the "identified files" idea

valentijnscholten:
Heads up, you need to update your migration as we merged another PR with a migration into dev. Quick update to the `dependencies` value.

xens:
> Just to be sure: You can create draft pull request "officially": https://github.blog/2019-02-14-introducing-draft-pull-requests/

Thanks didn't knew that ! That's my bad habits from gitlab ;)

dojo/models.py
xens: Agree, it's a bad copy / paste from the documentation.

dojo/models.py
xens: Good idea @madchap, I've put flaw instead of vulnerability as some issues are not necessarily vulnerabilities and it's the word used in many places in the doc.

Maffooch (approved):

xens:
I think I'm good with this one, documentation PR is also ready on the documentation repo. https://github.com/DefectDojo/Documentation/pull/139

Maffooch:
Looks like one more migration got pushed ahead of yours. Looks good to go after that fix.

xens:
> Looks like one more migration got pushed ahead of yours. Looks good to go after that fix.

one-more-time ! :smile_cat:

valentijnscholten (approved):

PR #2975, opened 2020-10-08
Remove HipChat integration, add support for Microsoft Teams instead

### LINKED ISSUE #2916: Remove HipChat integration, add support for Microsoft Teams instead
Atlassian has discontinued Hipchat Data Center more than 3 years ago and HipChat on premise 1.5 years ago. I can't imagine organisations are still using HipChat. So I think it's time to discontinue sending notifications from DefectDojo to HipChat as well.

Microsoft Teams is used by a lot of organisations nowadays and should be supported for notifications.

If the DefectDojo maintainers support this feature request, I am happy to implement it, because my company is using MS Teams.

StefanFl:
Resolves #2916 

Two remarks:
- I chose the name "Microsoft Teams" in the user interface, because Teams is a quite generic word. What do you think, let it stay like this, remove Microsoft or use "MS Teams"?
- Please have a look at line 75 in dojo/notifications/helper.py: In my tests I had the feeling the user isn't set properly and included "notifications_set.user = user". Does that make sense?


dojo/db_migrations/0056_ms_teams.py
```diff
@@ -0,0 +1,117 @@
+from django.db import migrations, models
+import multiselectfield.db.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dojo', '0055_notifications_jira_update_verbose_name'),
```
valentijnscholten, line 8: needs an update due to new migrations in dev

dojo/notifications/helper.py
```diff
@@ -72,7 +72,7 @@ def create_notification(event=None, *args, **kwargs):
                 applicable_notifications.append(system_notifications)
 
             notifications_set = Notifications.merge_notifications_list(applicable_notifications)
-
+            notifications_set.user = user
```
valentijnscholten, line 75: I don't think the notifications_set.user can ever be None, but it doesn't hurt too bad to set it again

valentijnscholten:
Nice PR, I was hoping someone would do this :-)

dojo/db_migrations/0056_ms_teams.py
StefanFl: Done

madchap (approved):
lgtm! Not easily testable on my side, but anyways, this will be much better than having something that just no longer exists :-) Thanks a lot for your contribution :)

valentijnscholten (approved):

PR #3000, opened 2020-10-14
apiv2: add misssing fields for filtering and remove UI related stuff

xens:
This PR add / fixes some issues with two fields in the `findings` filter of the APIv2 (`test__test_type` and `test__engagement`). It also removes some unused code (related to the UI) that was initially wrongly copy/pasted when creating filters that are dedicated to the API.

For ref: https://owasp.slack.com/archives/C014H3ZV9U6/p1602624904093500

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -413,18 +413,6 @@ class ApiProductFilter(DojoFilter):
         )
     )
 
-    def __init__(self, *args, **kwargs):
-        self.user = None
-        if 'user' in kwargs:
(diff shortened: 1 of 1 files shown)


madchap (approved):
Tested my initial use-case, work good now. I have not tested any other use-cases.

xens:
Test filtering on the three external fields

- test__test_type
- test__engagement
- test__engagement__product

and it's working as expected (multiple values also work).

valentijnscholten (approved):

PR #3002, opened 2020-10-14
Environment list unique name

dsever:
This PR includes migration  and unit test that ensures Environment can be only unique.

- enhancement


valentijnscholten:
Thanks.... but the migration will fail if some actually has non-unique values in that table. So it would need a data migration applied before the schema migration. Something similar to https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/db_migrations/0047_jira_minimum_severity_default.py (but ordered swapped).

dsever:
> Thanks.... but the migration will fail if some actually has non-unique values in that table. So it would need a data migration applied before the schema migration. Something similar to https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/db_migrations/0047_jira_minimum_severity_default.py (but ordered swapped).

@valentijnscholten maybe is better than to use catch/except block, and skip migration on error? If products are in the env with the same name, it would be problem to track it.

valentijnscholten:
You can't skip a migration (from code) as the database will get out of sync with the model in django. I think the chances are small that many people will have non-unique names. So a pragmatic solution is to put somthine very clear in the upgrade notes in the documentation that they need to make sure there are no duplicate entries before upgrading.

valentijnscholten:
the alternative is to write a data migration. go over all the environments and if there is one name with multiple entries, they need to be merged. so connect all tests from entry 2 to entry 1 and then remove entry 2.

valentijnscholten:
migration needs rename/updated parent as we just merged a pr with a migration.

PR #3004, opened 2020-10-14
Specific variable for https port

madchap:
Specifying a port through `DD_PORT` results in an error, as that variable is used for both the http and https endpoint.
This PR simply specifies a new variable for the https endpoint and adapts the nginx config file accordingly.
Some extra passing of variables needed.

```
$ docker-compose up -d                                             
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.nginx.ports value [{'target': 8081, 'published': 8081, 'protocol': 'tcp', 'mode': 'host'}, {'target': 8081, 'published': 8081, 'protocol': 'tcp', 'mode': 'host'}] has non-unique elements
```

Grepping through the codebase, this is apparently the only place that has it.

### DIFF
--- a/docker-compose.override.https.yml
+++ b/docker-compose.override.https.yml
@@ -6,8 +6,8 @@ services:
       USE_TLS: 'true'
       GENERATE_TLS_CERTIFICATE: 'true'
     ports:
-      - target: ${DD_PORT:-8443}
-        published: ${DD_PORT:-8443}
+      - target: ${DD_TLS_PORT:-8443}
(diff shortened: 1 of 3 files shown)


Maffooch (approved):

valentijnscholten:
What is the goal of these variables? I would expect that the goal is to specify the _external_ port you want to use for http / https?
Can't we configure this with the `publish` and `target` configuration? It just feels wrong to be modifying an nginx.conf file in a container setup.

madchap:
Superseded by #3041.

PR #3007, opened 2020-10-15
Add authorized users at the product type level

Maffooch:
Add authorized users to the product type level. The same `change`, `delete`, and `staff` settings options still apply to most of the functions that are used with authorized users where applicable.


valentijnscholten:
I haven't checked the queries, but merging querysets and applying distinct can break the ordering and can be slow. It's better and cleaner to use `filter(Q(condition1) | Q (condition2))`

dsever:
@Maffooch there is typo https://github.com/DefectDojo/django-DefectDojo/pull/3007/files#diff-949b6855a89b62c4501ed494c692f63e7cbfb2514e31ca56c0e0fc92d72f6ad8R11 so migration fails.

devGregA:
@valentijnscholten Unless things have changed. querysets are not evaluated until execution where they are combined to one query. Doing `object.filter(x=x).filter(y=y)` is the same as `object.filter(x=x,y=y)` Can you like the concerning code?

Maffooch:
The double filter and the Q method are identical in functionality, but the Q method is definitely cleaner in code. Will change to those

The OR operators return duplicate entries so using distinct is a must for queries that return authorized objects like findings. So if a user is authorized in a product and then also in the product type the product is in, they will see findings from that product twice.

valentijnscholten:
> @valentijnscholten Unless things have changed. querysets are not evaluated until execution where they are combined to one query. Doing `object.filter(x=x).filter(y=y)` is the same as `object.filter(x=x,y=y)` Can you like the concerning code?

in this case two querysets are combined like query1 | query 2 which will result in 2 queries.

> The double filter and the Q method are identical in functionality, but the Q method is definitely cleaner in code. Will change to those
> 
> The OR operators return duplicate entries so using distinct is a must for queries that return authorized objects like findings. So if a user is authorized in a product and then also in the product type the product is in, they will see findings from that product twice.

If it returns duplicate entries then it's not identical to Q(cond1) | Q(cond2) :-) 
If you use the Q method, you don't need distinct. It will generate the exact SQL query you want.

Maffooch:
It's a single query per the docs:
https://docs.djangoproject.com/en/3.0/ref/models/querysets/#or

valentijnscholten:
I've seen unions happen in these cases and the fact that you need distinct seemed to suggest that. 
But if the docs are right, you wouldn't need distinct.

dojo/notifications/helper.py
```diff
@@ -61,7 +61,8 @@ def create_notification(event=None, *args, **kwargs):
 
         # only send to authorized users or admin/superusers
         if product:
-            users = users.filter(Q(id__in=product.authorized_users.all()) | Q(is_superuser=True) | Q(is_staff=True))
+            qs = (product.authorized_users.all() | product.prod_type.authorized_users.all()).distinct()
+            users = users.filter(Q(id__in=qs) | Q(is_superuser=True) | Q(is_staff=True))
```
valentijnscholten, line 65: now I'm curious because I would expect this would work just as well:
`users = users.filter(Q(id__in=product.authorized_users.all()) | Q(id__in=productprod_type.authorized_users.all()) | Q(is_superuser=True) | Q(is_staff=True))`

Maffooch:
Im not totally sure if SQL OR will remove duplicates or not. Logically speaking it should not remove duplicates since it would count both queries individually. Guess it's time to do a test!

valentijnscholten:
to be sure, I was talking about these constructs:
```
        qs = Finding.objects.filter(
            test__engagement__product__authorized_users__in=[request.user]
        )
        qs = qs | Finding.objects.filter(
            test__engagement__product__prod_type__authorized_users__in=[request.user]
        )
        findings = qs.distinct()
```

Maffooch:
Right. I guess we can ty getting rid of the distinct and see what happens.

dojo/notifications/helper.py
```diff
@@ -61,7 +61,8 @@ def create_notification(event=None, *args, **kwargs):
 
         # only send to authorized users or admin/superusers
         if product:
-            users = users.filter(Q(id__in=product.authorized_users.all()) | Q(is_superuser=True) | Q(is_staff=True))
+            qs = (product.authorized_users.all() | product.prod_type.authorized_users.all())
+            users = users.filter(Q(id__in=qs) | Q(is_superuser=True) | Q(is_staff=True))
```
valentijnscholten, line 65: why not keep the same construct as on the left?
```
users = users.filter(Q(id__in=product.authorized_users.all()) | Q(id__in=productprod_type.authorized_users.all()) | Q(is_superuser=True) | Q(is_staff=True))
```

valentijnscholten (changes requested):

dojo/notifications/helper.py
Maffooch: A grab at brevity?

valentijnscholten: hmm the reason I ask is because when reading the code it makes me think that there is something special going on, some kind of construct to make it work. but after reading it a couple of times I realize it's just filtering the same as the rest.
so from that perspective I'd think it's better to have it the same so readers don't have to think extra.

valentijnscholten (approved):

madchap (approved):

madchap:
Whenever you're ready @Maffooch

PR #3011, opened 2020-10-15
APIv2 findings performance improvements

alles-klar:
Adding prefetch_related fields.

On behalf of DB Systel GmbH

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -236,8 +236,6 @@ class FindingViewSet(mixins.ListModelMixin,
                      ra_api.AcceptedFindingsMixin,
                      viewsets.GenericViewSet):
     serializer_class = serializers.FindingSerializer
-    queryset = Finding.objects.all().prefetch_related('endpoints', 'reviewers', 'images', 
-                                                      'found_by', 'notes', 'risk_acceptance_set')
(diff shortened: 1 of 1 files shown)


dojo/api_v2/views.py
```diff
@@ -257,9 +258,11 @@ def perform_update(self, serializer):
...
-                test__engagement__product__authorized_users__in=[self.request.user])
+                test__engagement__product__authorized_users__in=[self.request.user]).prefetch_related(
+                    'endpoints', 'reviewers', 'images', 'found_by', 'notes', 'risk_acceptance_set')
         else:
-            findings = Finding.objects.all()
+            findings = Finding.objects.all().prefetch_related('endpoints', 'reviewers', 'images', 
+                                                              'found_by', 'notes', 'risk_acceptance_set')
         return findings.prefetch_related('test',
                                         'test__test_type',
                                         'test__engagement',
```
valentijnscholten, line 268: could it be better to add them here (and only once)?

alles-klar: I followed your advise. I also removed queryset because if I am correct it is overwritten by get_queryset anyway?

valentijnscholten (approved):

devGregA (approved):

PR #3012, opened 2020-10-15
Add annotations for pods

madchap:
Allows to put annotations for django pod, celery pods.

Useful to add monitoring annotation, for example. Tested with datadog here, works well.

Also bumped up chart version as well as app version we didn't do last release.

### DIFF
--- a/helm/defectdojo/Chart.yaml
+++ b/helm/defectdojo/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: "1.7.1"
+appVersion: "1.8.0"
 description: A Helm chart for Kubernetes to installs DefectDojo
 name: defectdojo
-version: 1.3.1
+version: 1.3.2
 icon: https://www.defectdojo.org/img/favicon.ico
 dependencies:
   - name: mysql
--- a/helm/defectdojo/Chart.yaml
+++ b/helm/defectdojo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.8.0"
 description: A Helm chart for Kubernetes to installs DefectDojo
 name: defectdojo
-version: 1.3.2
+version: 1.4.0
 icon: https://www.defectdojo.org/img/favicon.ico
 dependencies:
   - name: mysql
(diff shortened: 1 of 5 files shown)


dsever:
@madchap I don't think this is good way to build helm. Helm should not contain additional tools as default.

madchap:
@valentijnscholten is there something to understand in the "flake8-you-pr" here? It hurt my eyes, so I went away ;-)

madchap:
fyi @dsever

madchap:
The GH flake8 failure here is ... whatever.

valentijnscholten:
I think you need to merge dev or rebase to make it use the fixed workflow definition as it uses the one from the latest dev commit on your branch.

helm/defectdojo/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
-appVersion: "1.7.1"
+appVersion: "1.8.0"
 description: A Helm chart for Kubernetes to installs DefectDojo
 name: defectdojo
```
ptrovatelli, line 4: Sorry for splitting hair but I would do 1.4.0 (feature added, not bugfix)

valentijnscholten (approved):

helm/defectdojo/templates/django-deployment.yaml
```diff
@@ -22,6 +22,10 @@ spec:
         defectdojo.org/component: django
         app.kubernetes.io/name: {{ include "defectdojo.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
+      annotations:
+      {{- with .Values.django.annotations }}
```
ptrovatelli, line 26: Could you update the values.yml with empty annotations for all three changes  ?  It's usually done as a helper so that looking at the values.yml only ppl know what they can configure.

ptrovatelli (changes requested):
Only minor stuff thanks 😀

helm/defectdojo/Chart.yaml
madchap: The line you pointed out was not the Chart version, but I understood it as such. Pushed a new commit.

ptrovatelli (approved):

PR #3014, opened 2020-10-15
Add delete methods for products and environments in api v2

RomainJufer:
In this PR we add the following to the API V2 :
- New endpoint to delete a product.
- New endpoint to delete an environment.

We have tested that in both cases the deletion cascade correctly. 
These changes have been made in collaboration with @xens

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -561,6 +561,7 @@ class DojoMetaViewSet(mixins.ListModelMixin,
 class ProductViewSet(mixins.ListModelMixin,
                      mixins.RetrieveModelMixin,
                      mixins.CreateModelMixin,
+                     mixins.DestroyModelMixin,
                      mixins.UpdateModelMixin,
                      viewsets.GenericViewSet):
     serializer_class = serializers.ProductSerializer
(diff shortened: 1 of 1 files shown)


xens:
Approving these changes as I tested them myself locally.

valentijnscholten:
Just thinking out loud. One of reasons we don't have DELETE currently is because there are no permission checks on the API?

xens:
When you say no permission checks it's like no check at all ? Like anybody with basic privileges (active?) is able to do operations on the API ? If that's the case what'd you recommend ? Allow only superusers to delete products first and then implement proper RBAC per product ?

valentijnscholten:
I think someone on slack told me there where no permission checks on the apiv2, but when I look at the code there are some checks there. maybe not all of them, but there's a bunch.

xens:
Ok I'll give it a try tomorrow.

RomainJufer:
We tried with the user type "Guest" and "Staff" and in both cases, the action is forbidden.

xens (approved):
Approving this PR as the DELETE method on both endpoints requires a user with `SuperUser` role we should be safe permission-wise.

valentijnscholten:
So a user who is authorized for the product cannot delete findings/products via the api?

RomainJufer:
We've just tested that and even if it is authorized, a user cannot delete the product. It seems that this is enforced via the DjangoModelPermission at the level of the model, preventing standard users from performing the operation. This is the same behavior as the one in the current UI.

valentijnscholten:
yeah, I remember that the apiv2 uses the granular model permissions configured in the django admin. the UI doesn't use those.

valentijnscholten (approved):

Maffooch (approved):

PR #3015, opened 2020-10-15
Import JSON reports of SSLyze scanner

### LINKED ISSUE #2987: SSLyze doesn't support XML output anymore
SSLyze has removed XML output with version 3.0. The SSLyze importer needs to be changed to read the JSON output of SSLyze.

I am already on it and will deliver the pull request soon 😃

StefanFl:
Resolves #2987


valentijnscholten:
I'm not fully familiar with the tool, but is it expected that everyone will upgrade to v3.0 and we can remove the v2 xml format? I see that v3 requires python 3.7+ which might not be available to everyone running ssllyze?
i.e. should we keep the xml format parser also in place for a while?

valentijnscholten:
BTWI think the `Test_Type.json` must be kept in sync with the list of scanners.

StefanFl:
Good point @valentijnscholten, I will have a look into it and amend the pull request or create a new one.

dojo/forms.py
```diff
@@ -380,7 +380,8 @@ class ImportScanForm(forms.Form):
                          ("Wpscan", "Wpscan"),
                          ("Sslscan", "Sslscan"),
                          ("JFrog Xray Scan", "JFrog Xray Scan"),
-                         ("Sslyze Scan", "Sslyze Scan"),
+                         ("SSLyze 2 Scan (XML)", "SSLyze 2 Scan (XML)"),
```
valentijnscholten, line 383: changing scanner id/names will break existing api scripts. it might work if you just changed the 2nd part of the tuple, but haven't tried that
```
("Sslyze Scan", "Sslyze Scan v2 (xml)"),
```

valentijnscholten (changes requested):

dojo/forms.py
StefanFl: I like to be explicit, so it's clear for users what is expected and what get's delivered. Couldn't we mention the breaking change in the release notes? So users of SSLyze will be made aware it's time to switch to a newer version which is more reliable and delivers better results.

If that it's not an option for you, I of course can change it and make it backwards compatible.

valentijnscholten: I don't like breaking things, having to deal with github issues, questions on slack, release notes etc.

StefanFl: Alright, you might have more experience with this in open source projects. I will change it making it backwards compatible.

valentijnscholten (approved):

dojo/forms.py
valentijnscholten: Merci :-)

StefanFl:
Sorry, just found out how to make a proper link for the reference

Maffooch (approved):

PR #3017, opened 2020-10-16
findings api: added display_status and branch to related_fields

ricardomeulendijks:
**Endpoint: api/v2/findings?related_fields=true**

Added fields at related_fields
- Added field: display_status
- Added field: engagement/branch

related_fields_pr


valentijnscholten:
Did you test what happens if `branch_tag` is empty or None?
Ik think for the `display_status` you should use `finding.status()`

valentijnscholten (approved):

ricardomeulendijks:
> Did you test what happens if `branch_tag` is empty or None?
> Ik think for the `display_status` you should use `finding.status()`

If branch tag is empty the value is null. I changed the display_status to finding.status()

Maffooch (approved):

PR #3021, opened 2020-10-16
bump Chart to v2

madchap:
This was overlooked and needed to generate a Chart.lock instead of the requirements.lock. Without this, things are not so peachy.


madchap:
Closing as having dependencies in charts breaks the `conditions`, see https://github.com/helm/helm/issues/5780

PR #3025, opened 2020-10-17
swagger ui: use sessions, more compact UI

valentijnscholten:
Small update to config of django-rest-swagger:
- Use django session. So if you're already logged in, you don't need to mess with the token. It will show the username of the logged in user top righ, see screenshot. If you need to test a specific api token from a different user maybe, open a new (incognito) browser tab/window.
- This also allows you to just enter some api urls in the browser to see the json, i.e. `http://dojo/api/v2/findings/1`
- Show a little less expaned view by default so it's easier to scroll to the api endpoint you want.
- Enable json editor. Not seen what it does yet, but it's supposed to help editing complex json bodies. Can't hurt I guess.

image


madchap (approved):
lgtm

Maffooch (approved):

Issue #3001, opened 2020-10-14
Add mitigated column in engagement view


madchap:
To add clarity at a glance, add a mitigated column which shows the number of mitigated findings (whichever sort).
Currently, open and duplicate are shown. But maybe out of these 100 findings, all are mitigated and you need to actually open the engagement to find out.

valentijnscholten:
I think this again boils down to definition of `open`. I think open should mean something like  "`active` and not `mitigated`"

valentijnscholten:
fixed by #3029 @madchap ?

madchap:
Yes

PR #3029, opened 2020-10-17
Add more findings info to engagement/test views

madchap:
Addresses https://github.com/DefectDojo/django-DefectDojo/issues/3001

I always find myself wondering how many mitigated findings or active/verified are in the engagements when looking at a glance.
With so many CI/CD imports, at a glance look become important.

This PR adds a couple columns when viewing engagements,

image

and drilling to tests

image

### DIFF
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -9,7 +9,7 @@ from django.contrib import messages
 from django.contrib.auth.decorators import user_passes_test
 from django.core.exceptions import PermissionDenied
 from django.urls import reverse
-from django.db.models import Q
+from django.db.models import Q, Count
 from django.http import HttpResponseRedirect, StreamingHttpResponse, Http404, HttpResponse, FileResponse
 from django.shortcuts import render, get_object_or_404
 from django.views.decorators.cache import cache_page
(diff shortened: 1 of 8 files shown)


madchap:
I can't reproduce any JS error locally :thinking: 

`integrationtest_1  | {'level': 'SEVERE', 'message': "http://nginx:8080/static/jquery/dist/jquery.js 4054:2 Uncaught TypeError: Cannot read property 'mData' of undefined", 'source': 'javascript', 'timestamp': 1602955753210}`

valentijnscholten:
we should really prefetch those counts

Maffooch:
+1 on the prefetching. Annotate those counts to test query set and reference them directly in the template.

devGregA:
I always like more info. It seems like the only one not included at this point is false positive count?

madchap:
> I always like more info. It seems like the only one not included at this point is false positive count?

False-positives are also marked as mitigated, so will fall within that count. I think not putting every single status in there matters just as much as putting enough (like "mitigated" overall regardless of the specific status). Wdyt?

valentijnscholten (approved):

Maffooch (approved):

PR #3030, opened 2020-10-17
fix jira column on similar findings rows

valentijnscholten:
fix jira column on similar findings rows


madchap (approved):

madchap:
Nice branch name :laughing:

PR #3031, opened 2020-10-17
title: suppress superfluous log lines

valentijnscholten:
Do you also NOT like this log line appearing everywhere? 
```
No config file found at /home/john_doe/.titlecase.txt
```

This PR suppresses that.


madchap (approved):
:rocket:

Maffooch (approved):

PR #3032, opened 2020-10-17
Set finding is_Mitigated on finding update

### LINKED ISSUE #3026: Un-marking findings active shows them as "Mitigated", but do not set "is_Mitigated"
**Bug description**
When one unmarks an Active finding, it goes to "Inactive, Mitigated". 
However, the `is_Mitigated` is not set to `True`.

However, in bulk edit for example, marking findings as "Mitigated" is possible and sets that field.

This results in discrepancies.

madchap:
fix #3026

**Note: DefectDojo is now on Python3.6 and Django 2.2.x.**


madchap:
For some reasons, I think I may be missing something obvious. Definitely works for my use-case through the UI, have to look elsewhere too. @valentijnscholten ?

valentijnscholten:
Bulk edit and API come to mind

madchap:
> Bulk edit and API come to mind

They seem work well when testing.

valentijnscholten:
When searching for `.active = False` I see lots of results and I think 40% of them do not set `is_Mitigated`.
image

madchap:
I think this one already fixes a couple issues known so far. I'd be in favor of merging, and see if we experience any more side effects of it not getting set?

valentijnscholten:
I don't remember all the details from the status discssion some months ago. But if "not active" always means "is_Mitigated", then we wouldn't need 2 booleans? I think we decided back then we need 2 booleans because "not active" is not always the same as "is_Mitigated", and vice versa. If that is the case, I think we should have them seperate everywhere and always let the user decide the value for both independently. 

The good thing about this PR is that it only adds 'is_Mitigated = False' in those places where we were already settings `mitigated = False`. So PR is good IMHO.

valentijnscholten (approved):

devGregA:
I'd like to ask a potentially silly question. Why is the 'M' capitalized?

devGregA:
Not related to PR on previous quesiton, approving.

madchap:
> I'd like to ask a potentially silly question. Why is the 'M' capitalized?

Aha! yup... I didn't git blame, because I don't want to know.

devGregA (approved):

madchap:
> The good thing about this PR is that it only adds 'is_Mitigated = False' in those places where we were already settings `mitigated = False`. So PR is good IMHO.

`mitigated` is actually a timestamp. Otherwise yes, no risk.

valentijnscholten:
I think we could do a migration that renames the field (in a future PR).

PR #3041, opened 2020-10-19
Fix nginx port definition in docker-compose.yml

rookies:
If `DD_PORT` is set, it overwrites not only the published port but also the target port, which is wrong because nginx always listens on port 8080 inside the container. The same is true for port 8443 with TLS enabled.

This pull request also adds the variable `DD_TLS_PORT` (by default 8443) that allows setting the port for HTTPS to a different value than the port for HTTP.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,12 +11,12 @@ services:
     environment:
       NGINX_METRICS_ENABLED: ${NGINX_METRICS_ENABLED:-false}
     ports:
-      - target: ${DD_PORT:-8080}
+      - target: 8080
         published: ${DD_PORT:-8080}
         protocol: tcp
         mode: host
-      - target: ${DD_PORT:-8443}
-        published: ${DD_PORT:-8443}
+      - target: 8443
+        published: ${DD_TLS_PORT:-8443}
         protocol: tcp
         mode: host
   uwsgi:


valentijnscholten (approved):

valentijnscholten:
probably replaces #3004

madchap:
Duplicates #3004 somewhat, the latter going the extra step to keep the flexibilty.

madchap:
Or maybe #3004 is wrong and this one is simpler and would also make it work. Re-opened and will close #3004.

madchap (approved):
Thanks and good :eyes: :+1:

rookies:
I forgot to update `docker-compose.override.https.yml` as well, I think changing

```yaml
    ports:
      - target: ${DD_PORT:-8443}
        published: ${DD_PORT:-8443}
        protocol: tcp
        mode: host
```
to
```yaml
    ports:
      - target: 8443
        published: ${DD_TLS_PORT:-8443}
        protocol: tcp
        mode: host
```
should do the job.

What's the usual approach in such a case, edit this pull request or create a new one?

valentijnscholten:
you can't edit this one :-)

rookies:
Okay, I never tried and was confused by the “Revert” button, but now I saw that it just creates a new pull request. Then a new one it is. :)

PR #3043, opened 2020-10-19
Kubernetes liveness and readiness improvement

uncycler:
The current liveness and readiness checks are done on the nginx container which proxy to uswgi. If uwsgi hangs or is overloaded, only the nginx container is restarted.

This PR correct this by doing the following:
* nginx: Check the liveness of nginx by using a dummy endpoint. This will avoid nginx getting restarted because of uwsgi failure.
* nginx: Check the readiness by checking the availability of uwsgi. In case of uwsgi failure, nginx will be removed from the loadbalancer but won't be restarted.
* uwsgi: Enabled the http server and check the availability of uwsgi. In case of failure, uwsgi will be restarted. There is no point to do readiness on this container since there is no service for it.

### DIFF
--- a/docker/entrypoint-uwsgi.sh
+++ b/docker/entrypoint-uwsgi.sh
@@ -10,4 +10,5 @@ exec uwsgi \
   --threads ${DD_UWSGI_NUM_OF_THREADS:-2} \
   --wsgi dojo.wsgi:application \
   --buffer-size="${DD_UWSGI_BUFFER_SIZE:-4096}" \
+  # HTTP endpoint is enabled for Kubernetes liveness checks. It should not be exposed as a serivce.
   --http 0.0.0.0:8081 --http-to ${DD_UWSGI_ENDPOINT}
--- a/docker/entrypoint-uwsgi.sh
+++ b/docker/entrypoint-uwsgi.sh
@@ -9,4 +9,5 @@ exec uwsgi \
   --processes ${DD_UWSGI_NUM_OF_PROCESSES:-2} \
   --threads ${DD_UWSGI_NUM_OF_THREADS:-2} \
   --wsgi dojo.wsgi:application \
-  --buffer-size="${DD_UWSGI_BUFFER_SIZE:-4096}"
+  --buffer-size="${DD_UWSGI_BUFFER_SIZE:-4096}" \
+  --http 0.0.0.0:8081 --http-to ${DD_UWSGI_ENDPOINT}
--- a/helm/defectdojo/templates/django-deployment.yaml
+++ b/helm/defectdojo/templates/django-deployment.yaml
@@ -92,19 +92,20 @@ spec:
           value: {{- if or .Values.django.ingress.activateTLS .Values.django.nginx.tls.enabled }} "True" {{- else }} "False" {{- end }}
         - name: DD_CSRF_COOKIE_SECURE
           value: {{- if or .Values.django.ingress.activateTLS .Values.django.nginx.tls.enabled }} "True" {{- else }} "False" {{- end }}
+        {{- if .Values.django.uwsgi.livenessProbe.enabled }}
         livenessProbe:
           httpGet:
             path: /login?next=/
             port: http
-            {{- if .Values.django.nginx.tls.enabled }}
-            scheme: HTTPS
-            {{- end }}
             httpHeaders:
             - name: Host
               value: {{ .Values.host }}
-          initialDelaySeconds: 120
-          periodSeconds: 10
-          failureThreshold: 6
+          failureThreshold: {{ .Values.django.uwsgi.livenessProbe.failureThreshold | default 6 }}
+          initialDelaySeconds: {{ .Values.django.uwsgi.livenessProbe.initialDelaySeconds | default 120 }}
+          periodSeconds: {{ .Values.django.uwsgi.livenessProbe.periodSeconds | default 10 }}
+          successThreshold: {{ .Values.django.uwsgi.livenessProbe.successThreshold | default 1 }}
+          timeoutSeconds: {{ .Values.django.uwsgi.livenessProbe.timeoutSeconds | default 5 }}
+        {{- end }}
         resources:
           {{- toYaml .Values.django.uwsgi.resources | nindent 10 }}
       - name: nginx
@@ -143,6 +144,7 @@ spec:
           initialDelaySeconds: 10
           periodSeconds: 10
           failureThreshold: 6
+        {{- if .Values.django.uwsgi.livenessProbe.enabled }}
         readinessProbe:
           httpGet:
             path: /uwsgi_health
@@ -153,9 +155,12 @@ spec:
             httpHeaders:
             - name: Host
               value: {{ .Values.host }}
-          initialDelaySeconds: 120
-          periodSeconds: 10
-          failureThreshold: 12
+          failureThreshold: {{ .Values.django.uwsgi.livenessProbe.failureThreshold | default 6 }}
+          initialDelaySeconds: {{ .Values.django.uwsgi.livenessProbe.initialDelaySeconds | default 120 }}
+          periodSeconds: {{ .Values.django.uwsgi.livenessProbe.periodSeconds | default 10 }}
+          successThreshold: {{ .Values.django.uwsgi.livenessProbe.successThreshold | default 1 }}
+          timeoutSeconds: {{ .Values.django.uwsgi.livenessProbe.timeoutSeconds | default 5 }}
+        {{- end }}
         resources:
           {{- toYaml .Values.django.nginx.resources | nindent 10 }}
       {{- with .Values.django.nodeSelector }}
--- a/helm/defectdojo/templates/django-deployment.yaml
+++ b/helm/defectdojo/templates/django-deployment.yaml
@@ -52,6 +52,10 @@ spec:
         volumeMounts:
         - name: run
           mountPath: /run/defectdojo
+        ports:
+        - name: http
+          protocol: TCP
+          containerPort: 8081
         envFrom:
         - configMapRef:
             name: {{ $fullName }}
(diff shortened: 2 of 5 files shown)


helm/defectdojo/templates/django-deployment.yaml
```diff
@@ -88,6 +92,19 @@ spec:
           value: {{- if or .Values.django.ingress.activateTLS .Values.django.nginx.tls.enabled }} "True" {{- else }} "False" {{- end }}
         - name: DD_CSRF_COOKIE_SECURE
           value: {{- if or .Values.django.ingress.activateTLS .Values.django.nginx.tls.enabled }} "True" {{- else }} "False" {{- end }}
+        livenessProbe:
+          httpGet:
+            path: /login?next=/
+            port: http
+            {{- if .Values.django.nginx.tls.enabled }}
```
ptrovatelli, line 99: I thought this probe was adressing uwsgi without going through nginx? why care about `django.nginx.tls.enabled` then?

helm/defectdojo/templates/django-deployment.yaml
```diff
@@ -88,6 +92,19 @@ spec:
...
+            scheme: HTTPS
+            {{- end }}
+            httpHeaders:
+            - name: Host
+              value: {{ .Values.host }}
+          initialDelaySeconds: 120
```
ptrovatelli, line 105: I found that 120 is a bit low in some contexts (db needs to be ready too, with migrations etc); could you increase and/or better make it configurable in values.yml? i have recently switched to 600 which is a big much and could be set finer probably. 

 to be as fine as possible, we could set up a startup probe too (see https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes )

nginx/nginx_TLS.conf
```diff
@@ -113,6 +113,17 @@ http {
...
+    # Used by Kubernetes liveness and readiness checks
+    location = /nginx_health {
+      return 200 "Born to be alive!\n";
+      access_log off;
+    }
+    location = /uwsgi_health {
+      rewrite /.+ /login?next=/ break;
+      include /run/defectdojo/uwsgi_pass;
+      include /etc/nginx/wsgi_params;
+      access_log off;
```
ptrovatelli, line 125: can the application be normally accessed from /uwsgi_health endpoint, with access_log off? doesn't that have security implications like lack of logging in case of password brute-force using this endpoint?

docker/entrypoint-uwsgi.sh
```diff
@@ -9,4 +9,5 @@ exec uwsgi \
   --processes ${DD_UWSGI_NUM_OF_PROCESSES:-2} \
   --threads ${DD_UWSGI_NUM_OF_THREADS:-2} \
   --wsgi dojo.wsgi:application \
-  --buffer-size="${DD_UWSGI_BUFFER_SIZE:-4096}"
+  --buffer-size="${DD_UWSGI_BUFFER_SIZE:-4096}" \
+  --http 0.0.0.0:8081 --http-to ${DD_UWSGI_ENDPOINT}
```
ptrovatelli, line 13: is that opening the possibility to acess uwsgi directly without going through nginx? doesn't that have security implications? it's my understanding that nginx works as a kind of firewall in front of uwsgi, and for example won't let any headers going through but the ones that are allowed as per wsgi_params file.

ptrovatelli (commented):
these changes make a lot of sense! can you check my remarks though? thanks!

helm/defectdojo/templates/django-deployment.yaml
ptrovatelli: on second thoughts, not sure about that startup probe: defectdojo doesn't really like getting requested before the application is up and running: that tends to generate nasty errors and stack traces. probably better to just use a big enough initialDelay

nginx/nginx_TLS.conf
ptrovatelli: we could maybe use  some dummy django endpoint, other than login, that leads nowhere

docker/entrypoint-uwsgi.sh
ptrovatelli: could this one go through nginx too without opening anything new on uwsgi?

helm/defectdojo/templates/django-deployment.yaml
uncycler: You're right, i'll remove it

helm/defectdojo/templates/django-deployment.yaml
uncycler: startupProbe is alpha since 1.16 and beta in 1.18. For compatibility, it's better not to use it. Maybe, in the future, add some condition in the helm chart.

uncycler: I'll make initialDelaySeconds configurable.

nginx/nginx_TLS.conf
uncycler: The best would be to have a real django readiness and liveness endpoint.

Since location use an = sign, it will only respond to /uwsgi_health, not /uwsgi_health/somestuff. And the rewrite will always redirect to /login?next=/. I'll had a condition to only forward if the method is GET. What do you think?

docker/entrypoint-uwsgi.sh
uncycler: Unless you don't expose the port in docker-compose or Kubernetes, it should not be an issue. I don't know if there is a better way to do it. I found some tools like this one: https://github.com/andreif/uwsgi-tools but I'm not sure it's a good idea to integrate that in image.

nginx/nginx_TLS.conf
ptrovatelli: is the `/login?next=/` logged in access log upon `/uwsgi_health` access ? If yes, no problem you can leave it, I just wasn't sure of how the rewrite would be handled.

ptrovatelli: adding that GET limitation might be a good idea though just in case :) .

docker/entrypoint-uwsgi.sh
ptrovatelli: actually going through nginx would be a bad idea because if nginx fails we would restart uwsgi for no reason. 

you're right it's not a big risk as the port is not exposed, and this script is not used outside of docker-compose/ kubernetes context. 

you're right too, it's not trivial to check a uwsgi server state; the socket endpoint is listening on port 3031 but we can't do much without additional tooling. just tried curl for example isn't really helpful
```
curl localhost:3031
curl: (52) Empty reply from server
```

 you probably came across this article :) https://stackoverflow.com/questions/57122352/how-to-check-uwsgi-with-gevent-at-socket-file-in-k8s-readinessprobe-and-liveness

you're all good for this comment then; could you just add a comment in the entrypoint script that explains that this `--http ` option is required for the k8s uwsgi livenessProbe configured in django-deployment.yaml ? 

thanks!

nginx/nginx_TLS.conf
ptrovatelli: just tried: it is currently possible to login through the /uwsgi_health uri, and we are missing some nginx access logs in this case: the ones for `GET /login?next=/` and `POST /login?next=/`

so it may be a way to bypass some IDS based on access logs. 

I think allowing only GET like you said should be enough

thanks!

uncycler: It's done!

docker/entrypoint-uwsgi.sh
uncycler: Done as well!

helm/defectdojo/templates/django-deployment.yaml
uncycler: It's now possible to configure livenessProbe of uswgi (which configure readiness on nginx container as well). Is that ok the way I did it?

docker/entrypoint-uwsgi.sh
ptrovatelli: thanks :)

nginx/nginx_TLS.conf
ptrovatelli: great!

helm/defectdojo/templates/django-deployment.yaml
ptrovatelli: it's great that you added the timeout configuration too and that you increased the default 1s which was a bit low I think! 

for the configuration, I'd have configured all three probes with different values so that looking at the values.yml one immediately understands that there are 3 probes and how to configure them. however if you prefer like this I can live with it :)

ptrovatelli: you could also put the default values directly in the values.yml and no default in the django-deployment.yaml because now it seems there are 2 default values (one of which is commented) but sometimes they differ (like uwsgi.livenessProbe.timeoutSeconds)

uncycler: > for the configuration, I'd have configured all three probes with different values so that looking at the values.yml one immediately understands that there are 3 probes and how to configure them. however if you prefer like this I can live with it :)

Well, nginx is very fast to start and current values should enough even on raspberry! So I don't see the point to configure it. And since readiness on nginx and liveness on uwsgi are probing the same thing, I thought it would be simpler to have only one configuration for both.

ptrovatelli: ok fine!

ptrovatelli (approved):

ptrovatelli:
@alles-klar good to go? (after travis ends)

alles-klar:
Hi @uncycler,

thanks for this great improvement! 

I saw you already discussed the liveness probe for uwsgi. I personally would prefer using https://github.com/andreif/uwsgi-tools instead of introducing a new http endpoint. @valentijnscholten what do you think about this?

ptrovatelli:
> Hi @uncycler,
> 
> thanks for this great improvement!
> 
> I saw you already discussed the liveness probe for uwsgi. I personally would prefer using https://github.com/andreif/uwsgi-tools instead of introducing a new http endpoint. @valentijnscholten what do you think about this?

I did have mixed feelings too, but I think introducing a new dependency is less safe than opening a port that won't be reachable from the outside world. Who will asses the content of uwsgi-tools? who will update uwsgi-tools when uwsgi evolves? ...

uncycler:
Would it be possible to use manage.py or some lightweight python script instead?

devGregA:
@mtesauro please weigh in since this touches docker. We have had issues with Kube improvements causing issues for docker-compose and docker users.

mtesauro:
FWIW, I'd prefer keeping this as simple as possible and not add a new dependency unless there's a compelling reason that I'm missing.

If I don't care about liveliness checks and/or running DefetDojo with docker-compose, I can comment out some lines in nginx (nginx.conf or nginx_TLS.conf) and not expose port 8081 outside the uwsgi container (entrypoint-uwsgi.sh) to be back where I was before this PR.

Worst case, those running in docker-compose have nginx listening on a port that isn't exposed and that's not much of any resource overhead for those use cases.

valentijnscholten:
The PR feels like like overkill, but I guess there are no simple solutions with kubernetes ;-)

valentijnscholten (approved):

mtesauro (approved):
Approved

Issue #3018, opened 2020-10-16
Product : Retrieving related object with single request


RomainJufer:
This proposal has been made conjointly with @xens 

We would like to reduce the number of queries required to fetch all the data related to a `Product`. More specifically, we would like to provide a way to retrieve in a single query the data associated with fields defining a relation.

To do that we came up with essentially three proposals to achieve this :

1. Directly modifying the ProductSerializer to force it to follows these relations and get the data. The django framework support this feature natively and thus it is really easy to implement. However, the downside is that it would mean forcing a complete retrieval even if it is not needed.

2. Modifying the endpoint to support an additional parameter specifying if we want standard (as currently implemented) or full retrieval. Something along the line of `/products/?full`.

3. Modifying the endpoints to be even more parametric and be able to specify for which field we want to follow the relation.

As an example, we have implemented the retrieval for `product_manager` yielding

```
{
      "id": 3,
      "findings_count": 0,
      "findings_list": [],
      "product_manager": {
        "id": 1,
        "username": "admin",
        "first_name": "Admin",
        "last_name": "User",
        "email": "admin@defectdojo.local",
        "last_login": "2020-10-16T07:31:49.047121Z",
        "is_active": true,
        "is_staff": true
      },
      "tags": [],
      "product_meta": [],
      ...
}
```

instead of 

```
{
      "id": 3,
      "findings_count": 0,
      "findings_list": [],
      "product_manager": 5,
      "tags": [],
      "product_meta": [],
      ...
}
```

valentijnscholten:
How does option 1 work? Could it be used but only when a parameter is specified (option 2). If possible, you don't want to write all the code to get all the relations (and their relations etc). How many levels deep will it go?

RomainJufer:
For option 1 we simply modify the `ProductSerializer` using the serializers corresponding to the fields for which we want to follow the relation and load the data. For example, for `product_manager` we use the `UserSerializer`. In our case we only need to follow the relation for the fields : 

- product_manager
- technical_manager
- team_manager
- prod_type
- regulations

We would like to avoid doing it for `findings` as the list can be relatively big. This is why we also propose option 3 that offers more granularity and allows us to generalize the approach. While option 2 is also possible we think that it would not scale well if, for example, we include also `findings` which might be required to be semantically coherent. For the depths of the retrieval, we only need 1 level in our case. I dont know if this is something worth adding as a parameter and also how easy it would be to implement it.

[...]

PR #3056, opened 2020-10-21
 Product : Retrieving related object with single request

RomainJufer:
This feature is related to the discussion at : https://github.com/DefectDojo/django-DefectDojo/issues/3018

Add a new query parameter to the endpoints for product retrieval (list/retrieve) that allow the automatic expansion of the relational fields. More specifically, the endpoints can now be called with :

- `/products/?expand={array_of_fields_to_expand}`
- `/products/{id}/?expand={array_of_fields_to_expand}`

The fields to expand are provided as a comma-separated array of strings. Upon receiving the parameters, the `ProductSerializers` will try to follow the relations of the fields specified. If a field is not a relation or not existent in the model, it will simply be ignored without error to preserve the original behavior of the endpoint.

As an example, a `Product` is related to a set of regulations and a user for the technical contact. Both of these fields can be expanded by calling the endpoint

`/products/?expand=regulations,technical_contact`

This new parameter has been documented to be included in the swagger documentation.

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -571,19 +571,21 @@ class DojoMetaViewSet(mixins.ListModelMixin,
     filter_backends = (DjangoFilterBackend,)
     filter_fields = ('id', 'product', 'endpoint', 'name', 'finding')
 
+
 def _product_expand_decorator():
     return swagger_auto_schema(
-    responses={status.HTTP_200_OK: serializers.ProductSerializer},
-    manual_parameters=[
-        openapi.Parameter(
-            name="expand", 
-            in_=openapi.IN_QUERY, 
-            description="Fields to expand based on their relation", 
-            type=openapi.TYPE_ARRAY, 
-            items=openapi.Items(
-                type=openapi.TYPE_STRING,
-                enum=list(serializers.ProductSerializer._expansion_serializers.keys())))
-    ])
+        responses={status.HTTP_200_OK: serializers.ProductSerializer},
+        manual_parameters=[
+            openapi.Parameter(
+                name="expand",
+                in_=openapi.IN_QUERY,
+                description="Fields to expand based on their relation",
+                type=openapi.TYPE_ARRAY,
+                items=openapi.Items(
+                    type=openapi.TYPE_STRING,
+                    enum=list(serializers.ProductSerializer._expansion_serializers.keys())))
+        ])
+
 
 @method_decorator(name="list", decorator=_product_expand_decorator())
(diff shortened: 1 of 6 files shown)


valentijnscholten:
Looks great! Could you add / modify some test cases to make sure the test suite touches these expansion filters? Ideally a test with and without and verify the response on presence/absence.

dojo/api_v2/views.py
```diff
@@ -557,7 +559,20 @@ class DojoMetaViewSet(mixins.ListModelMixin,
...
 
-
+def _product_expand_decorator():
+    return swagger_auto_schema(
+    responses={status.HTTP_200_OK: serializers.ProductSerializer},
+    manual_parameters=[
+        openapi.Parameter(
+            name="expand", 
+            in_=openapi.IN_QUERY, 
+            description="Fields to expand based on their relation",
```
xens, line 569: Really cool to be able to document it this way, nice finding ! We could give more info on the `description` as it might be hard to understand without context but I'm unable to find something better ^^.

RomainJufer:
Yes of course, however, I've got some troubles finding where to put those tests. I was thinking adding them in `tests/Product_unit_test.py` however it seems that these tests go through the UI. In our case, I think it would make more sense to directly test the endpoint since the UI does not use the newly added field expansion mechanism. Therefore, do I put them inside that same file anyway or do I put them somewhere else ? Unfortunately, I didn't manage to find a test suite that tests the API endpoints, do you have one ?

valentijnscholten:
`test_rest_framework.py`

Maffooch:
Please rebase to the dev branch.

valentijnscholten:
I changed the base branch for the PR which seems to work more or less, but there are conflucts so a merge of dev or rebase onto dev is needed.

RomainJufer:
I've rebased and added unit tests for the feature. I also improved a little bit the swagger doc by restricting the range of the value taken by the new parameter with an `enum`. My only concern is that I've done that dynamically and by doing so it breaks the encapsulation principle as `_expansion_serializers`should not be accessed from outside. However, since it is for documentation purposes I think it might be fine. What do you think ?

Also, it looks like the flake8 for unit tests is not going through and one of the reasons is
`'dojo.api_v2.views.SonarqubeProductViewSet' imported but unused (F401)`

But since it is not something that I've introduced (I think) I don't know if I should remove it or not ?

valentijnscholten:
yeah remove that unused import. and could you resolve the conflicts (and other flake8 items).

PR #3065, opened 2020-10-21
master->dev: github action workflow sync

valentijnscholten:
commits directly to master have to be in dev


valentijnscholten:
github can't handle us

PR #3069, opened 2020-10-22
Allow to load extra fixtures in docker deployment

uncycler:
This PR allows user to add their own fixtures to docker image that are automatically loaded by the initializer.

- [ ] Your code is python 3.6 compliant (specific python >3.6 syntax is currently not accepted).

### DIFF
--- a/Dockerfile.django
+++ b/Dockerfile.django
@@ -80,6 +80,8 @@ RUN \
   cp dojo/settings/settings.dist.py dojo/settings/settings.py
 COPY tests/ ./tests/
 RUN \
+  rm -f /readme.txt && \
+  rm -f dojo/fixtures/readme.txt && \
   mkdir -p dojo/migrations && \
   chmod g=u dojo/migrations && \
   true
--- a/Dockerfile.django
+++ b/Dockerfile.django
@@ -71,12 +71,17 @@ COPY \
   /
 COPY wsgi.py manage.py tests/unit-tests.sh ./
 COPY dojo/ ./dojo/
+
+# Add extra fixtures to docker image which are loaded by the initializer
+COPY docker/extra_fixtures/* /app/dojo/fixtures/
+
 # Legacy installs need the modified settings.py, do not remove!
 RUN \
   cp dojo/settings/settings.dist.py dojo/settings/settings.py
 COPY tests/ ./tests/
 RUN \
   rm -f /readme.txt && \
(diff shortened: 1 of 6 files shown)


Dockerfile.django
```diff
@@ -71,12 +71,17 @@ COPY \
...
+# Add extra fixtures to docker image which are loaded by the initializer
+COPY docker/extra_fixtures/* /app/dojo/fixtures/
+
 # Legacy installs need the modified settings.py, do not remove!
 RUN \
   cp dojo/settings/settings.dist.py dojo/settings/settings.py
 COPY tests/ ./tests/
 RUN \
   rm -f /readme.txt && \
+  rm -f dojo/fixtures/readme.txt && \
```
alles-klar, line 84: What do you think of adding this file to .dockerignore instead of deleting it afterwards?

docker/entrypoint-initializer.sh
```diff
@@ -65,6 +65,13 @@ EOD
   python3 manage.py loaddata regulation
   python3 manage.py import_surveys
   python3 manage.py loaddata initial_surveys
+
+  # If there is extra fixtures, load them
+  for i in $(ls dojo/fixtures/extra_*.json 2>/dev/null) ; do
+    echo "Loading $i"
+    python3 manage.py loaddata ${i%.*}
+  done
+
```
valentijnscholten, line 74: should there be some kind of predictable ordering (filename?) so people can have dependencies between fixtures?

Dockerfile.django
uncycler: Thanks, forgot about this one!

I'll ignore /readme.txt coming from docker/certs/readme.txt as well.

docker/entrypoint-initializer.sh
uncycler: Good idea! Adding | sort -n

valentijnscholten (approved):

Dockerfile.django
uncycler: @alles-klar I revert the change. If I add readme.txt in dockerignore, COPY will fail because the directory is empty.

valentijnscholten: can we put a comment in there so everybody knows why this is happening|?

uncycler: Yep, it's done

alles-klar (approved):

Issue #2795, opened 2020-08-21
External Dependency in Safety Parser Requires Internet Connection


alles-klar:
**Bug description**
Until release 1.7. it was possible to run defectdojo in environments without internet connection. The new safety parser requires a connection to https://raw.githubusercontent.com/pyupio/safety-db/master/data/insecure_full.json. In setups without a internet connections the unit tests and also using the parser will fail.

dojo/tools/safety/parser.py", line 11

**Steps to reproduce**
1. Run the unit tests of defectdojo in a environment without internet connection.

**Expected behavior**
Handling the Error when the url is not reachable.

**Deployment method** *(select with an `X`)*
- [x] Kubernetes
- [x] Docker
- [ ] setup.bash / legacy-setup.bash

**Console logs** (optional)

 ERROR: test_no_cve (dojo.unittests.test_safety_parser.TestSafetyParser)

 Traceback (most recent call last):
   File "/usr/local/lib/python3.6/urllib/request.py", line 1349, in do_open
     encode_chunked=req.has_header('Transfer-encoding'))
   File "/usr/local/lib/python3.6/http/client.py", line 1272, in request
     self._send_request(method, url, body, headers, encode_chunked)
   File "/usr/local/lib/python3.6/http/client.py", line 1318, in _send_request
     self.endheaders(body, encode_chunked=encode_chunked)
   File "/usr/local/lib/python3.6/http/client.py", line 1267, in endheaders
     self._send_output(message_body, encode_chunked=encode_chunked)
   File "/usr/local/lib/python3.6/http/client.py", line 1038, in _send_output
     self.send(msg)
   File "/usr/local/lib/python3.6/http/client.py", line 976, in send
     self.connect()
   File "/usr/local/lib/python3.6/http/client.py", line 1425, in connect
     super().connect()
   File "/usr/local/lib/python3.6/http/client.py", line 948, in connect
     (self.host,self.port), self.timeout, self.source_address)
   File "/usr/local/lib/python3.6/socket.py", line 724, in create_connection
     raise err
   File "/usr/local/lib/python3.6/socket.py", line 713, in create_connection
     sock.connect(sa)
 TimeoutError: [Errno 110] Connection timed out
 During handling of the above exception, another exception occurred:
 Traceback (most recent call last):
   File "/app/dojo/unittests/test_safety_parser.py", line 19, in test_no_cve
     parser = SafetyParser(f, Test())
[...]

PR #3081, opened 2020-10-23
enable safety parser to work offline

alles-klar:
Fixes https://github.com/DefectDojo/django-DefectDojo/issues/2795

On behalf of DB Systel GmbH

### DIFF
--- a/Dockerfile.django
+++ b/Dockerfile.django
@@ -45,6 +45,7 @@ RUN \
     git \
     # only required for the dbshell (used by the initializer job)
     postgresql-client \
+    curl \
     && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists && \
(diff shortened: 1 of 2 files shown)


ptrovatelli (approved):

valentijnscholten:
Nice solution, but in my opinion defect dojo is not really designed to enricht scan reports with data from external sources. So ideally the report would be made complete before it is imported into DD, similar to npm audit parser, owasp dependency check parser, php symfony check parser, whitesource parser, etc.

alles-klar:
> Nice solution, but in my opinion defect dojo is not really designed to enricht scan reports with data from external sources. So ideally the report would be made complete before it is imported into DD, similar to npm audit parser, owasp dependency check parser, php symfony check parser, whitesource parser, etc.

I fully agree with you, but the parser is already there and it is not created by me. What do you suggest to do? For me it is important do run defect dojo and the unit test in a cooperated environment without internet connection. So we could remove the parser or accept a workaround like this one.

valentijnscholten (approved):

valentijnscholten:
yeah we have no real choice now but to merge it.

PR #3082, opened 2020-10-23
Allow requiring a secret for incoming JIRA webhook requests

valentijnscholten:
The current JIRA webhook can be used just be sending POST data to `/webhook`. 
This PR changes this into `/jira/webhook/<secret>` where secret is a slug to be configured in the system settings.
For backwards compatibility reasons we keep the `/webhook/` url working as well.

Note:
- For existing installs the JIRA Webhook secret will be disabled to keep them working
- For new installs the webhook is disabled anyway, but the initializer generates a random (uuid) secret and enabled secret by default;

System Settings page now looks like this:

image

The disable option is mainly to keep existing installs from breaking on upgrade.

Docs: https://github.com/DefectDojo/Documentation/pull/145


mtesauro (approved):
Approved

madchap (changes requested):
I am going to nitpick... but I'd love to have a small "copy" icon next to that link :)

dojo/jira_link/views.py
```diff
@@ -26,12 +26,23 @@
 
 
 @csrf_exempt
-def webhook(request):
-    # Webhook shouldn't be active if jira isn't enabled
+def webhook(request, secret=None):
```
madchap, line 29: `@require_http_methods(["POST"])` maybe? What do we want to allow through that webhook?

valentijnscholten: add. good idea, although it makes testing it harder ;-)

madchap (approved):

valentijnscholten:
copy button would be nice, but takes some time and can be added by anyone who loves javascript ;-)

PR #3087, opened 2020-10-25
travis: remove non-k8s tests, reduce matrix for k8s

valentijnscholten:
As discussed / proposed on Slack:
- Travis tets are extremely slow
- The non-k8s tests are running fine in GitHub Actions

The plan was to let both Travis and GitHub actions run alongside eachother for a couple of weeks. But waiting times for Travis are currently up to 3-4 hours sometimes, which breaks the feedback loop we need.

This PR:
- disables/removes all non-k8s tests / steps from the travis build
- reduces the build matrix for k8s tests to alleviate the strain on travis and reduce our waiting times.

Proposed for after the release of 1.9.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,22 +9,22 @@ env:
     - HELM_VERSION=v3.3.2
     - CHANGE_MINIKUBE_NONE_USER=true
   matrix:
-    - TEST=flake8
-    - TEST=docker_integration_tests
-    - BROKER=rabbitmq DATABASE=mysql
+    # - TEST=flake8
+    # - TEST=docker_integration_tests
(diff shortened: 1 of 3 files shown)


valentijnscholten:
image

madchap:
As the k8s GHA is come through, do we want to push this PR even further?

valentijnscholten:
ideally yes, but we could merge this one today. and let the k8s tests run for a week or so to be sure they're all ok. and they fail if something is wrong in unit tests or migrations or...

valentijnscholten:
merged as we no longer have choices as travis already hangs for 20+ hours and travis-ci.org is being phased out.

PR #3091, opened 2020-10-25
Centralize logic to decide on async or foreground tasks

valentijnscholten:
Currently the code is "scattered" with logic to decide if a task should be handled asyncrhonously by celery, or in the foreground:

```
if Dojo_User.wants_block_execution(request.user):
    update_jira_issue(finding, True)	
else:	
    update_jira_issue.delay(finding, True)
```

There were 18 ocurrences. and this PR brings that down to 2 occurrences. This done by implementing a decorator `dojo_async_task` that decides if the underlying task (function) should be executed asynchronously or in the foreground.

This PR also removes all wrapping / glue code from tasks.py as that is not needed.

```
@app.task(name='applying rules')	
def async_rules(new_finding, *args, **kwargs):	
    logger.info("applying rules")	
    sync_rules(new_finding, *args, **kwargs)
```

This PR also removes the dedupe_signal. It was called manually from finding.save() which makes it no different from a normal method/function call. 

A file `test_celery_decorators.py` is added for further testing in the future and to capture some initial results regarding decorators inside and outside celery tasks which may become helpful in the future.

### DIFF
--- a/dojo/decorators.py
+++ b/dojo/decorators.py
@@ -1,0 +1,21 @@
+from functools import wraps
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+# Defect Dojo performs all tasks asynchrnonously using celery
+# *unless* the user initiating the task has set block_execution to True in their usercontactinfo profile
+def dojo_async_task(func):
+    @wraps(func)
+    def __wrapper__(*args, **kwargs):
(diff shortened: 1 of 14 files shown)


Maffooch (approved):
Thanks for the cleanup here!

mtesauro (approved):
Approved

PR #3092, opened 2020-10-25
Celery: only send model ids and not full instances

valentijnscholten:
based on 3091, so merge that first. there's also a quickfix (but not complete fix) for #3099 

Currently Defect Dojo is sending the full model instances to celery. These can be fairly large, especially if there are fields stored such as `unsaved_request` and `unsaved_response`. These can be large base64 encoded "blobs" putting straing on celery / rabbitmq. The current approach only works because we are using the "insecure" `pickle` serializer, which at some point we have to get rid of (#3093)

This PR adds a setting / environment variable (DD_)CELERY_PASS_MODEL_BY_ID to convert all models to ids before calling celery. Inside the celery tasks the models will be loaded from the database based on this id.
This is done using 2 decorators that can be transparently used on celery tasks to make sure only the id is passed to celery.
And on the celery side the 2nd decorator will load the model instance from the database from that id and transparently injects it into the function parameters again.

So no code changes needed, apart from applying these two decorators on celery tasks. The alternative was to use ids everywhere instead of model instances. This would make for ugly code and probably people will forget it etc.

The DD_CELERY_PASS_MODEL_BY_ID is *disabled*  by default. So people must make a conscious decision to start using this feature, i.e. it's a feature flag. For now I think it's good to do some testing, but at some point I think we should make it default to True. Passing ids instead of model instances is a best practices and helps with stability of celery/rabbitmq.

image

Old:
```
@dojo_async_task
@task
def test_valentijn(new_finding):
    logger.debug('test_valentijn:')
    logger.debug(new_finding)
```

New:
```
@dojo_model_to_id
@dojo_async_task
@task
@dojo_model_from_id
def test_valentijn(new_finding):
    logger.debug('test_valentijn:')
    logger.debug(new_finding)
```

Celery tasks almost always have a single finding as parameter, so those are the defaults. But you can tweak 2 settings:

Will convert parameter 1 (product) to id before sending to celery and convert back into a Product before execution in celery starts.
```
@dojo_model_to_id
@dojo_async_task
@task
@dojo_model_from_id(model=Product, parameter=1)
def test_valentijn(dummy_parameter_not_convert, product):
    logger.debug('test_valentijn:')
    logger.debug(product)
```

This approach works will because almost always we are only sending 1 parameter: the finding on which some logic should occur. If we were to have lots of different parameters all the time, it might be better to force ourselves to pass ids instead of instances. 
Or to make a more generic decorator that converts all parameters that are model instances into ids (and the reverse inside celery)

### DIFF
--- a/dojo/decorators.py
+++ b/dojo/decorators.py
@@ -1,4 +1,6 @@
 from functools import wraps
+from dojo.models import Finding
+from django.db import models
 import logging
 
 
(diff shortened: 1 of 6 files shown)


madchap (approved):
The feature flag helps :+1: I've just retested celery dedupe on 800K across 3 workers, holds well, but celery is having a hard time relinquishing memory and such. Getting this through is certainly the thing to do. Cheers.

valentijnscholten:
Going to merge this as it is an optional feature behind feature flag (which is disabled by default). Want to get this in >7 days before the next release so we can have some people running dev testing it, at least for not having any regressions.

PR #3105, opened 2020-10-27
Optional env variable for custom footer version

russellballestrini:
This Pull request introduces a new optional env variable to allow end users to configure a custom version in the footer of `base.html` 

	modified:   dojo/settings/settings.dist.py
	modified:   dojo/templatetags/display_tags.py

Let me know if I should also submit a sibling pull request to this file: https://github.com/DefectDojo/Documentation/blob/master/docs/settings-docs.rst

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -6,6 +6,7 @@ from celery.schedules import crontab
 import environ
 root = environ.Path(__file__) - 3  # Three folders back
 
+# reference: https://pypi.org/project/django-environ/
 env = environ.Env(
     # Set casting and default values
     DD_SITE_URL=(str, 'http://localhost:8080'),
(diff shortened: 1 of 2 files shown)


madchap (approved):
Looks simple enough to me, why not. Out of curiosity, what do you use it for?

russellballestrini:
@madchap internally we have GitLab building a fork of this repo on every commit and shipping it to our stage defectdojo cluster running on ECS/Fargate.

This PR lets me use an environment variable to pass the commit SHA so that I know what version I'm on.
In prod we use only stable releases using git tags.

Our fork is in a stable situation but I would like any new development to go upstream (here). I'm hoping that I can one day remove the internal fork and simply pin to whatever is stable in this repo.

valentijnscholten (approved):

valentijnscholten:
@russellballestrini 
> Let me know if I should also submit a sibling pull request to this file: https://github.com/DefectDojo/Documentation/blob/master/docs/settings-docs.rst

Yes please

PR #3114, opened 2020-10-28
Fix Placeholder Text in New Product Type Form

J12934:
Noticed a small typo in the placeholder text of the "new product type" form.
This was newly added in 1.9 via #3007


valentijnscholten (approved):

Maffooch (approved):
Thanks :)

PR #3131, opened 2020-10-30
Raise ImproperlyConfigured exception if no DD_SECRET_KEY is set

rookies:
Django raises an ImproperlyConfigured exception if no `SECRET_KEY` is configured, but this is prevented in DefectDojo because `DD_SECRET_KEY` has a default value of `"."`. This might be a security problem because `"."` is then used as the `SECRET_KEY`.

This pull request changes the default value to `""` (empty string), leading to an ImproperlyConfigured exception if no `DD_SECRET_KEY` is passed.

It might be a good idea to add the same change for `DD_CREDENTIAL_AES_256_KEY` but then we would need to add extra code that checks for an empty value because this is not a setting Django cares about.


rookies:
Hm, now the tests fail because there is no key set... Is there a better way to check for a missing key only at runtime? Or can we just set a key for the integration tests?

valentijnscholten:
you can set the env variable for integration tests in docker-compose override.
maybe a unit test can be made to test the case where there is not KEY set

rookies:
Thanks, now the tests pass :+1:

Maffooch (approved):

valentijnscholten (approved):

PR #3132, opened 2020-10-30
allow TEMPLATE_DEBUG to be set in settings

valentijnscholten:
Djanog has a TEMPLATE_DEBUG config setting that provides extra information when an error occurs in a template AND DEBUG is set to True.

This PR allows us to configure this using DD_ env variable, and sets default to False.


cody-m-tibco (approved):
Wish I would have known about this earlier. Good find!

Maffooch (approved):

PR #3133, opened 2020-10-30
add datetime to logformat (use verbose)

valentijnscholten:
By default the log output from DD is not displaying any date or time. Personally I prefer to have the default to use the `verbose` logformat which basically only adds a datetimestamp at the beginning of the line.

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -841,6 +841,7 @@ LOGGING = {
         },
         'console': {
             'class': 'logging.StreamHandler',
+            'formatter': 'verbose'
         },
         'json_console': {
             'class': 'logging.StreamHandler',


Maffooch (approved):

madchap (approved):
Useful for bash deployments at least ;-)

madchap (changes requested):
Are you sure you want to merge to `master` or not `dev`?

valentijnscholten:
> Are you sure you want to merge to `master` or not `dev`?

good point. I think this explains why my other PR few days ago was still pointing to dev. must have had them mixed up. not a surprise with these helpful branchnames...

Maffooch:
Merging because Fred approved when it was on `master`

PR #3136, opened 2020-10-30
settings: support add settings.py and support local_settings.py

valentijnscholten:
Currently we have some copy commands here and there to copy the settings.dist.py to settings.py. Sometimes this works, sometimes it fails (silently) with permission issues.
When people need to make override or additions to settings, they have to edit settings.dist.py or maintain their own copy of settings.py. With the regular changes in settings.dist.py, this leads to painful upgrades.

This PR:
- retains the `settings.dist.py`
- adds a `settings.py` file to the git repo
- that `settings.py` file loads settings.dist.py and optional after that loads `local_settings.py`
- this allows for overriding / extending the settings, without having to maintain a full copy of `settings.dist.py`
- using `.env.prod` or DD_ENV_PATH is still supported (but only allows for environment variable overrides, not changing loggers, middlewares, etc)
- an example `template-local_settings` is added

The local_settings file is used in the dev, ptvsd, unit_tests, integration_tests docker compose override scenario's.
In release mode the `local_settings.py` file (and any `.env.xxx` files) are NOT copied into the container.
For this scenario, users can put a `local_settings.py` in `/docker/extra_settings`. This file will be copied up starting uwsgi as the `/docker/extra_settings` folder is mounted into the container by docker-compose.


devGregA:
All, I just wanted to note. This is an excellent PR. In the early days, we did try to separate settings, but it created scoping issues. @cneill do you remember what library we used?

Maffooch:
Looks good to me! I'm wondering how the upgrade process will be for those with existing`settings.py` files. Would users be able to just rename their `settings.py` file to `settings.dist.py`?

valentijnscholten:
I think they will get a conflict when doing git pull or git merge. They could rename their file to `local_settings.py`.  Or `settings.dist.py` but then they wouldn't benefit from this PR.

Maffooch:
There should probably be a well documented entry in the documentation explaining how to cut over then.

valentijnscholten:
yeah, I was sort of waiting to get some feedback before updating, but most of the docs are already there so I'll add an "upgrading note".

valentijnscholten:
https://github.com/DefectDojo/Documentation/pull/148

mtesauro:
Looks good to me. Tested successfully with a couple of different docker-compose setups. Thanks @valentijnscholten for adding in my feedback on the container paths provided via Slack. This PR is much more elegant than my PR #3045

mtesauro (approved):
Approved

Maffooch (approved):

Issue #3100, opened 2020-10-27
get_related_fields in FindingSerializer fails when environment is unknown


RomainJufer:
@xens 

When the environment is not set in a product the method get_related_fields fails. This happens when the endpoint `/api/v2/findings/?related_fields=true` is called

```
uwsgi_1         | Internal Server Error: /api/v2/findings/
uwsgi_1         | Traceback (most recent call last):
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
uwsgi_1         |     response = get_response(request)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
uwsgi_1         |     response = self.process_exception_by_middleware(e, request)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
uwsgi_1         |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
uwsgi_1         |     return view_func(*args, **kwargs)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/rest_framework/viewsets.py", line 114, in view
uwsgi_1         |     return self.dispatch(request, *args, **kwargs)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 505, in dispatch
uwsgi_1         |     response = self.handle_exception(exc)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 465, in handle_exception
uwsgi_1         |     self.raise_uncaught_exception(exc)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 476, in raise_uncaught_exception
uwsgi_1         |     raise exc
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 502, in dispatch
uwsgi_1         |     response = handler(request, *args, **kwargs)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/django/utils/decorators.py", line 45, in _wrapper
uwsgi_1         |     return bound_method(*args, **kwargs)
[...]

PR #3141, opened 2020-11-02
APIv2: Use serializers instead of direct access to fields of model instance

RomainJufer:
Fix the issue raised in #3100. 

Introduce serializers for the related fields instead of accessing the fields directly from the model. While this follows the design principles of django it also changes a bit the structure of the returned object. More specifically, we don't have all the referred fields as direct children of the parent JSON object `related_fields`. Instead, we have an object with the same structure as defined in the models.

This fix also adds the swagger documentation for the response as it was missing from the current implementation.

Note : I kept the name `related_fields`, however, we might want to rename it ?

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -674,15 +674,6 @@ class FindingTestSerializer(serializers.ModelSerializer):
         fields = ["id", "title", "test_type", "engagement", "environment"]
 
 
-class FindingRelatedFieldsSerializer(serializers.Serializer):
-    test = FindingTestSerializer(required=False)
-    status = serializers.SerializerMethodField(required=False)
-
-    @swagger_serializer_method(serializers.ListField(serializers.CharField()))
-    def get_status(self, obj):
-        return obj.status()
-
-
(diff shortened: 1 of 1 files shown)


valentijnscholten (changes requested):
Thanks. Two observations:
- This PR removes branch and display_status
- This PR removes the need to provide a value for the related_fields parameter (true)

RomainJufer:
Nice catch, thanks ! Sorry. I've corrected both points, should be fine now

valentijnscholten:
We named the status field `display_status` as it is a field meant for displaying and not really a real status field that can be used to make comparisons or implement logic in it. Could name it as that?

What we did wrong in the previous PR is add this field to `related_fields`. I think it should be just a field on the finding? Maybe you could move it into the right spot now that the serializer is already there?

RomainJufer:
It is now named `display_status` and I it isn't related to the `related_fields` parameters. Is it fine like that ?

valentijnscholten (approved):

valentijnscholten:
yes, thanks.

madchap (approved):
Neat, thanks.

PR #3145, opened 2020-11-02
npm audit: refuse npm7 or higher as it misses most important fields

valentijnscholten:
Node 15 has been released, which includes npm7. This completely changes the output of `npm audit --json`.
I was about to update the parser to handle the new format, but found out most of the fields we're using are missing.

I raised a report at: https://github.com/npm/npm-audit-report/issues/45

For now I have implemented a blunt check in the parser to prevent npm7 reports from being uploaded. Without that check, the parser thinks the report is empty and it will close all open findings in defect dojo (if `close_old_findings=true` is specified on `reimport_scan`).

### DIFF
--- a/docker/entrypoint-unit-tests-devDocker.sh
+++ b/docker/entrypoint-unit-tests-devDocker.sh
@@ -13,7 +13,13 @@ unset DD_DATABASE_URL
 python3 manage.py makemigrations dojo
 python3 manage.py migrate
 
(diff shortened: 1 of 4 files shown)


Maffooch (approved):

mtesauro:
Sad that the new version has less info.  Although this is listed as a bug PR, it's more like an external bug.

Thanks for submitting the bug upstream.

mtesauro (approved):
Approved

PR #3150, opened 2020-11-02
Improve Qualys WAS parser.

iwalton3:
This PR provides several improvements to the reporting quality from the Qualys WAS reports.

Changes:
 - Change invalid severity Informational to Info.
 - Truncate fields too long for database model.
 - Map the finding date.
     - For combined findings, the most recent date is used.
 - Map the QID to `unique_id_from_tool`.
 - Only decode base64 when the tag indicates it is used.
 - Map information gathered items into findings.
     - Include data in findings.
     - Cap Scan Diagnostic severity to Info.
 - Handle empty request/response list.
 - Handle null values instead of converting "None" to a string.
 - Allow importing both application and scan reports.

Caveats:
 - There may be some disagreements on how the dates are handled.
 - The QID may not be the most suitable field for `unique_id_from_tool`.

For reference, we use this configuration in settings, which may be something
that should be added to the PR:

```python3
    'Qualys Webapp Scan': ['title', 'unique_id_from_tool', 'endpoints'],
}

    'Qualys Webapp Scan': DEDUPE_ALGO_HASH_CODE,
}
```


iwalton3:
The tests are still failing because I have added informational findings, which were previously not included.

valentijnscholten:
Sounds all good to me, but I don't have experience with Qualys reports so would like some others to look at this one.

madchap:
I am not doing Qualys either.

Can you please check if the files you're working with are the same format as the ones found under https://github.com/DefectDojo/sample-scan-files/blob/master/qualys/README.md? Otherwise, could you please submit a PR on that other repo too? It would help in testing.

Cheers.

iwalton3:
I found the [following scans](https://github.com/DefectDojo/sample-scan-files/tree/master/qualys_webapp) in the repository linked. It seems as if [the unit tests](https://github.com/DefectDojo/django-DefectDojo/tree/dev/dojo/unittests/scans/qualys_webapp) are only using the three scans from 2020, which do import into the parser.

The older scan from 2017 does not have a date in some of the findings, so it fails to import. We have run a large number of Qualys scans through the import and have yet to see one without a date, but this can be added for robustness if desired. The scan importer supports both application-level and scan-level reports, but currently the unit tests only cover scan-level reports.

Maffooch:
Yeah parsing qualys WAS reports is tricky because there are like. 4 different. ways to generate them based on where you're at in their portal. This is a very welcome change! I've been meaning to do something along these lines but have been a bit short on time. I will take a closer look at this soon.

dojo/tools/qualys_webapp/parser.py
```diff
@@ -28,11 +35,17 @@ def get_cwe(cwe):
...
 # them to fit their respective model structures and the adds them to a
 # newly generated Finding
-def attach_extras(endpoints, requests, responses, finding):
+def attach_extras(endpoints, requests, responses, finding, date, qid):
     if finding is None:
         finding = Finding()
         finding.unsaved_req_resp = list()
         finding.unsaved_endpoints = list()
+        finding.date = date
+        finding.unique_id_from_tool = str(qid)
```
alles-klar, line 44: The qid is not unique in qualys because in can be found on multiple targets. This is not the idea of unique_id_from_tool. But you can add it in the description.

iwalton3: Would you consider it reasonable to add another database column for something like "vuln_id_from_tool"? We'd really like this information to be available for de-duplication and reporting purposes.

alles-klar: I don't know if many tools have there own ids? If yes, it is a great idea.

iwalton3: I know at least Veracode does.

Maffooch:
This looks wonderful! Now that the parser now supports both scan and app reports, the unit tests will need to be tweaked a bit. They are at `dojo/unittests/test_qualys_webapp_parser.py`

iwalton3:
Ok. The two issues with the tests would be:
 - The number of findings has changes, since informational findings are now included.
 - The application reports are supported but untested.

I could likely transform one of the existing reports into an application report format. That would likely be easier than releasing an actual test app report, unless you are aware of some that are already floating around.

Maffooch:
It's the counts. It definitely wouldn't hurt to have a app scan report in there too though. The only difference between app and scan reports is the extra `WEB_APPLICATION` tag around the vulnerability list block
Also, you can see why things are failing if you look at the 'details' of the test that failed.

valentijnscholten:
Is #3153 something to look into?
Is this new field really needed? If so, can you document it in the API and docs repo?

iwalton3:
It looks like the PR fixes the issue described, although I have not confirmed why or if all of the findings are listed.

The new database field allows more accurate deduplication of issues and additionally assists with reporting the findings in a machine-readable format. We historically have implemented this using unique_id_from_tool, but as described earlier, this is not appropriate. Using the description would mean tooling may have to parse the description to obtain vulnerability ids, and changing the title would break past results people have imported.

alles-klar (approved):

alles-klar:
Looks good for me now, but please take a look a the unit tests.

jagusztinl:
Please merge this, waiting for the solution.. :-)

Maffooch (approved):

Maffooch (approved):

Maffooch:
Double approved for whatever reason 🤪

valentijnscholten:
@iwalton3 any reason why this fields need to 500 characters long? Sounds way too long for an id field.

madchap:
Getting hit by this. Could this relate?

```
Running migrations:
  Applying dojo.0062_add_vuln_id_from_tool...Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 84, in _execute
  File "/usr/local/lib/python3.6/site-packages/django/db/backends/mysql/base.py", line 71, in execute
    return self.cursor.execute(query, args)
  File "/usr/local/lib/python3.6/site-packages/MySQLdb/cursors.py", line 206, in execute
    res = self._query(query)
  File "/usr/local/lib/python3.6/site-packages/MySQLdb/cursors.py", line 319, in _query
    db.query(q)
  File "/usr/local/lib/python3.6/site-packages/MySQLdb/connections.py", line 259, in query
    _mysql.connection.query(self, query)
MySQLdb._exceptions.IntegrityError: (1364, "Field 'id' doesn't have a default value")
```

dojo/db_migrations/0062_add_vuln_id_from_tool.py
```diff
@@ -0,0 +1,33 @@
...
+
+    dependencies = [
+        ('dojo', '0061_jira_webhook_secret'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='finding',
+            name='vuln_id_from_tool',
+            field=models.CharField(blank=True, help_text='Non-unique technical id from the source tool associated with the vulnerability type.', max_length=500, null=True, verbose_name='Vulnerability ID from tool'),
```
madchap, line 16: Should be TextField eventually, to allow for more than 255 chars.

valentijnscholten: no please. it's just an id, maybe 64 or 128.

valentijnscholten:
> Getting hit by this. Could this relate?
> 
> ```
> Running migrations:
>   Applying dojo.0062_add_vuln_id_from_tool...Traceback (most recent call last):
>   File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 84, in _execute
>     return self.cursor.execute(sql, params)
>   File "/usr/local/lib/python3.6/site-packages/django/db/backends/mysql/base.py", line 71, in execute
>     return self.cursor.execute(query, args)
>   File "/usr/local/lib/python3.6/site-packages/MySQLdb/cursors.py", line 206, in execute
>     res = self._query(query)
>   File "/usr/local/lib/python3.6/site-packages/MySQLdb/cursors.py", line 319, in _query
>     db.query(q)
>   File "/usr/local/lib/python3.6/site-packages/MySQLdb/connections.py", line 259, in query
>     _mysql.connection.query(self, query)
> MySQLdb._exceptions.IntegrityError: (1364, "Field 'id' doesn't have a default value")
> ```

don't think so as the new field isn't indexed.

valentijnscholten:
Sounds more like somehow a finding without an `id` is being inserted or is already in place?

madchap:
Probably something in that local environment, worked in other places. Weird.

iwalton3:
> still not in favour of these extremely long fields for just an id value :-)

I copied the `unique_id_from_tool` field, but I do agree that there are some questionably long fields in the database. For instance, the [engagement status field is 2000 characters long](https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/models.py#L984) when it is an enum of values not exceeding 20 characters.

valentijnscholten:
Aha, the technical debt devil popping its head. Well I guess we need a bigger cleanup at some point, maybe for a v2.0

madchap:
Posting stuff here -- unsure if @iwalton3 is on Slack -- I can consistently reproduce my `id` issue as posted above.

I can workaround it by modifying the mySQL database with `SET GLOBAL sql_mode='';` and then the migration works fine.

I could not put my finger on it yet, but without the above workaround, I have pretty much everything that breaks. It seems postgreSQL is not impacted.

madchap:
db backend debug turned on.. does not yield that obvious a clue... https://gist.github.com/madchap/4d16d90b96f99050abbaf4949607d44c

iwalton3:
I know that the migration worked when I ran it in Docker, but I haven't tested it anywhere else.

iwalton3:
What exact MySQL database was this connecting to? I just tested and the migrations work fine from a clean database within docker. I deleted all of the tables in the local Docker MySQL and ran `python3 manage.py migrate`.

valentijnscholten:
For me it also ran fine on two big datasets. But this was mysql 8.0.

devGregA:
All, this PR has a lot of great improvements, but seems to have broken the core functionality and is causing not all findings to be imported. I'm still investigating. Will update shortly. I'm hoping is just the 'False' exclusions on line 123, but will update shortly.

devGregA:
Please disregard my previous (deleted) comment.

ZsanettMagyar:
Does anyone know in what version will be merged/released to production?
I can't find any PR to master which contains this fix.

valentijnscholten:
Normaal releases are five from dev branch to master. This pr is already in 1.10.x

PR #3119, opened 2020-10-28
Issue and PR template update

valentijnscholten:
Add more explicit note about required information and the necessitiy to specify the exact error message being observed.

PR #3152, opened 2020-11-02
Sync Master 1.9.1 with Dev

Maffooch:
Align the commits of master with dev so they will be equal for a minute


Maffooch:
@valentijnscholten Merge conflicts on the versions, but fixing would require to change them in dev on this pr, and then changing them right back to what they were immediately afterwards. Is this really worth it since to changes in master are already in dev?

valentijnscholten:
We need the commits to be in sync. You can resolve the conflicts and set version to 1.10.0-dev and then merge without squashing. It will be an empty PR I think but it will prevent conflicts on the next release. The only reason this is happening is because the PRs where merged into dev instead of the hotfix branch.

valentijnscholten:
This one is now also included in this PR, which is fine: https://github.com/DefectDojo/django-DefectDojo/pull/3119

PR #3154, opened 2020-11-03
APIv2: Support for data prefetching on endpoints

RomainJufer:
This PR is related to the following discussion #3018 

We introduce the following new packages:
- `prefetch` : contains the code to perform prefetching based on a set of serializers.
- `schema` : introduce a new swagger schema generator to support manual edit of an existing schema.

With the above, prefetching can be added to any viewset for the endpoint(s) `list` and/or `retrieve`. This is done by inheriting the new prefetch mixins `PrefetchRetrieveMixin` and `PrefetchListMixin`. In this PR we have only introduce prefetching for the endpoints related to `product` and `findings`. In addition, to generate the correct documentation associated with this operation, the viewset needs to be decorated with the `prefetch_decorator` otherwise the doc swagger will not have the correct query parameters nor the correct response model.

Prefetching is completly automatic in the sense that it automatically retrieve the fields that could be prefetched and try to match them against a given set of serializers. No need for further parametrization then inheriting the mixin. Note that we only support depth 1 relation expansion.

The swagger documentation is also automatically generated for the given viewset based on the fields that has been deduced as prefetchable.

This is a first version and it remains open to changes in the design/architecture or any other suggestion. More information can be found in the discussion mentionned in the beginning. I did not included it here to avoid redundancy.

What do you think ?

### DIFF
--- a/dojo/api_v2/prefetch/decorator.py
+++ b/dojo/api_v2/prefetch/decorator.py
@@ -1,0 +1,49 @@
+from drf_yasg2 import openapi, utils
+from .prefetcher import _Prefetcher
+from .utils import _get_prefetchable_fields
+from ..schema.extra_schema import ExtraSchema
+from ..schema.utils import LazySchemaRef
+
+
+def prefetch_decorator(methods, serializer):
+    """Decorator to generate swagger doc for a view set implementing prefetching.
+
+    Args:
+        methods (list[string]): list of the method for which prefetching is enabled. List must be a sublist of ["read", "list"].
+        serializer (Serializer): the serializer associated to the model represented by the viewset
+
+    Returns:
+        func(class)->class: The decorator to apply
+    """
+    _supported_methods = ["read", "list"]
+
+    def _decorator(cclass):
+        prefetcher = _Prefetcher()
+        fields = _get_prefetchable_fields(serializer())
+
+        field_to_serializer = dict([(name, prefetcher._find_serializer(field)) for name, field in fields])
+        fields_to_refname = dict([(name, utils.get_serializer_ref_name(serializer())) for name, serializer in field_to_serializer.items()])
+        fields_name = [name for name, _ in fields]
+
+        # New openapi parameter corresponding to the prefetchable fields
+        prefetch_params = openapi.Parameter("prefetch", in_=openapi.IN_QUERY, required=False, type=openapi.TYPE_ARRAY, items=openapi.Items(type=openapi.TYPE_STRING, enum=fields_name))
+
+        # New response field corresponding to the prefetch field
+        additional_props = dict([(name, openapi.Schema(type=openapi.TYPE_OBJECT, read_only=True, additional_properties=LazySchemaRef(fields_to_refname[name], True))) for name in fields_name])
+        prefetch_response = openapi.Schema(type=openapi.TYPE_OBJECT, properties=additional_props)
+
+        # Create the object representing the information to add to the swagger doc
+        extra_params = dict()
+        for method in methods:
+            if method in _supported_methods:
+                extra_params[method] = {"parameters": prefetch_params, "responses": {"200": {"prefetch": prefetch_response}}}
+
+        # Generate the swagger schema generator
+        generator = ExtraSchema.create(extra_params)
+
+        # Set the generator as the generator of the given class provided that one has not already been set
+        assert not hasattr(cclass, "swagger_schema"), "swagger_schema applied twice to class"
+        setattr(cclass, "swagger_schema", generator)
+        return cclass
+
+    return _decorator
--- a/dojo/api_v2/prefetch/mixins.py
+++ b/dojo/api_v2/prefetch/mixins.py
@@ -1,0 +1,42 @@
+from rest_framework.response import Response
+from rest_framework.mixins import ListModelMixin, RetrieveModelMixin
+from .prefetcher import _Prefetcher
+
+
+class PrefetchListMixin(ListModelMixin):
+    def list(self, request, *args, **kwargs):
+        prefetch_params = request.GET.get("prefetch", "").split(",")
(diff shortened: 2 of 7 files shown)


valentijnscholten:
Could you explain with an example request/response? I think this is a way to provide a parameter to trigger DD to return extra fields in the response? The term prefetch seems to point do Django ORM prefetch, but I don't think that's what it does?

RomainJufer:
If we consider for example `Product`, we have the following fields that are relationnal fields
  -> authorized_users, regulations, technical_contact, product_manager, prod_type, team_manager

By adding the mentioned mixins and decorator to the `ProductViewSet`, we can now call `/products/` and `/products/{id}/` with the new parameter `prefetch` which is a list of one or more of the above fields. Without loss of generality, if we consider only `/products/{id}/` and we call with `?prefetch=regulations,prod_type` (supposing that they are not null or empty), we will get, for example, the following response.

```
{
  "id": 1,
  "findings_count": 1,
  "findings_list": [
    1
  ],
  "tags": [],
  "product_meta": [],
  "name": "dasdas",
  "description": "dsadas",
  "created": null,
  "prod_numeric_grade": null,
  "business_criticality": null,
  "platform": null,
  "lifecycle": null,
  "origin": null,
  "user_records": null,
  "revenue": null,
  "external_audience": false,
  "internet_accessible": false,
  "product_manager": null,
  "technical_contact": null,
  "team_manager": null,
  "prod_type": 1,
  "authorized_users": [],
  "regulations": [
    9
  ],
  "prefetch": {
    "prod_type": {
      "1": {
        "id": 1,
        "name": "Research and Development",
        "critical_product": false,
        "key_product": false,
        "updated": null,
        "created": null,
        "authorized_users": []
      }
    },
    "regulations": {
      "9": {
        "id": 9,
        "name": "California Security Breach Information Act",
        "acronym": "CA SB-1386",
        "category": "privacy",
        "jurisdiction": "United States, California",
        "description": "In the United States, the California Security Breach Information Act (SB-1386) is a California state law requiring organizations that maintain personal information about individuals to inform those individuals if the security of their information is compromised. The Act stipulates that if there's a security breach of a database containing personal data, the responsible organization must notify each individual for whom it maintained information. The Act, which went into effect July 1, 2003, was created to help stem the increasing incidence of identity theft.",
        "reference": "http://en.wikipedia.org/wiki/California_S.B._1386"
      }
    }
  }
}
```

The goal of this mechanism is related to what can be done with django internally but now the result of prefetching can be forwarded to the caller or the endpoint. The goal is to avoid having to do multiple queries to retrieve data specified by relationnal fields.
I don't know if it is clear ?

valentijnscholten:
Thank you. It's nice to have something plug-and-play, but it's still quite a bit of code that we'll need to maintain. I was initially confused by the use of the term 'prefetch'. In Django this is a query optimization technique and I would to suggest a different name here, for example `include_related`.

Another question, I don't have time to try it, but what would happen with `authorized_users` for example if `prod_type` is included? Will it always be empty because max depth = 1? Or will it contain username (from `str()`)?

RomainJufer:
Indeed this is quite a bit of code to maintain unfortunately and I don't think I can really reduce it without losing some of the interesting features. Do you think we should add that feature to defectdojo ?

About what you said with the `prefetch` of django this is exactly what I thought in fact. The idea of this parameter is to avoid subsequent queries due to having to fetch the data related to some ids. Hence, by adding this new parameter, we can ask defectdojo to already include these data in this kind of cache that is stored under this new field `prefetch` in the response. Then, on the client who called the endpoint, instead of performing the usual subsequent queries to retrieve the data from the ids, it can instead go through that cache and match the ids he got with the data already retrieved, hence the name `prefetch`. But I've got no problem with renaming it `include_related` if you prefer :)

Finally, I'm not sure I completly understand your question but the code retrieves the related objects for each field appearing as value of the `prefetch` parameter independently. It then stores these objects under the tuple `(field_name, id)` where `field_name` is the name of the original field, for example `authorized_users` and `id` is the id of the object. This is to avoid duplicate. For example, in the case of the `list` operation, we do that for each product retrieved and the `prefetch` field is not added to each product separately but at the top-level of the JSON object returned. Let me know if I did not answered your question.

valentijnscholten:
OK, just noticed that the prefetch block is on the top level. So the client still has to do some logic to bind the ids to the data in the prefetch block. I get the idea, let me think about it if it can be improved.

What I meant with my question is, `authorized_users` is empty in the example above:
```
    "prod_type": {
      "1": {
        "id": 1,
        "name": "Research and Development",
        "critical_product": false,
        "key_product": false,
        "updated": null,
        "created": null,
        "authorized_users": []
      }
```
Will it always be empty? Or will it contain ìd`s it there are authorized users.

RomainJufer:
Indeed, it will contain the ids and the prefetch will contain the user object. I just generated the above example with a product that didn't have any authorized users. Essentially this will yield the same result as for the `regulations` but for `user` instances.

RomainJufer:
It will contain ids but the `prefetch` will not retrieve the corresponding object due to the depth = 1 as you said, yes.

valentijnscholten:
@RomainJufer did you see that this PR is waiting for conflict resolution?

RomainJufer:
Indeed, sorry, I thought that we were not proceeding further with that idea of "prefetching" since it would add too much code that would need to be maintained. We kept it in our local fork and we improved it as well so this PR is not up to date anymore so we can close it if you agree ? I can reopen a new PR with our current version if you would like to have this feature as well otherwise we will keep it on our version only.

valentijnscholten:
I was hesitant yes. But it looks like there is no easy way to avoid all this code and during the presentation you did at the moderator meeting I decided we should go all in on the API and am actually looking forward to the stuff you are doing/adding.

valentijnscholten:
superseded

PR #3155, opened 2020-11-03
GitHub Action: k8s helm dockerhub login

dsever:
Setting k8s in actions to use:
-  imagePullSecrets
-  kubectl to fetch django web
- documented way of secret creation in  helm values

### DIFF
--- a/.github/workflows/k8s-testing.yml
+++ b/.github/workflows/k8s-testing.yml
@@ -108,11 +108,11 @@ jobs:
         databases: [pgsql,mysql]
         brokers: [redis,rabbit]
     steps:
-      - name: Login to DockerHub
-        uses: docker/login-action@v1
-        with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
+#      - name: Login to DockerHub
+#        uses: docker/login-action@v1
+#        with:
+#          username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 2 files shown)


valentijnscholten (approved):

valentijnscholten:
Looks great, thanks for looking into this!

mtesauro (approved):
Approved

valentijnscholten:
just waiting for @dsever as it would be better to have this go straight into master as github will only look there for workflow definitions.

madchap:
> just waiting for @dsever as it would be better to have this go straight into master as github will only look there for workflow definitions.

Would only look there when action is launched with `master` as ref branch.

madchap (approved):

valentijnscholten:
> > just waiting for @dsever as it would be better to have this go straight into master as github will only look there for workflow definitions.
> 
> Would only look there when action is launched with `master` as ref branch.
My experience is that most of the time it will look in the master branch for the workflow definition. But it can be an older version if the PR was branched from dev which was branched from an earlier commit on master. Another reason for dev to be in sync with master!
So this PR has to be merged into master and then we have to merge master into dev. This behaviour of GitHub actions is undocumented, but as long as you make sure the workflow definition is on master and dev it should be OK.
But the PR has to have the commit with the latest workflow definition. So any older PRs have to be rebased, or they have to merge dev into themselves, before the new workflow definition is used.

valentijnscholten:
master updated in #3169 and merged back into dev in #3161

PR #3156, opened 2020-11-03
Improve default date handling.

iwalton3:
This PR is more of a workaround than an actual fix to this issue, but at the very least it should start some discussion about the best way to handle this problem.

The DefectDojo Finding model defaults to the current date for the date that a finding was discovered. This is a reasonable default, but the problem comes later when the model is processed. Depending on which re-import endpoint you call, something different can happen:

 - Web UI: If the finding was discovered on the current date, then the test start date is used.
 - API v1: The test start date is always used.
 - API v2: The scan date is always used.

This PR will make it so that regardless of how a scan is re-imported, the finding date will always be used as long as the scan importer provided it. If the date is the default, then it uses the start date from the test. It does so by creating a DefaultDate subclass of the datetime object, which we can check for but behaves identically to a normal datetime object in all other ways.

Caveats:
 - Does not fix API v1 Scan Import (Only Re-Import).
 - Does not fix Web UI Scan Import (Only Re-Import).
 - This, in the current form, completely breaks the database migration generator.

These caveats can be addressed if this solution is one we would like to implement.


valentijnscholten:
Thanks for the PR... But.... if we need this kind of complexity for a default date, I think we (so not you) are doing something wrong :-) I think we should just take "today" if there is no date available in the scan report nor something specified in the requests coming in to the APIv2 or to the backend for the WebUI. Is that possible?

iwalton3:
> I think we should just take "today" if there is no date available in the scan report

The finding defaults to the current date, so removing the code in the import handlers that sets the date to the target_start date (or other variants) would solve this easily with no special DefaultDate class.

valentijnscholten:
wdyt @mtesauro @devGregA

Maffooch:
There definitely needs to be some sort of homogeneity with how dates are handled, but I think adding a new class to handle this is a little overkill. Just make the default date today and call it a day.

valentijnscholten:
@iwalton3 can you adjust the PR to do just that? seems like you are in the know now of where and when :-)

valentijnscholten (approved):

valentijnscholten:
@iwalton3 if you merg upstream dev into this branch, it will pick up the new workflow definition and stop tests from failing

Maffooch (approved):

PR #3160, opened 2020-11-03
APIv2: Allow updating users and set is_superuser

iwalton3:
This allows updating user profiles via the API and additionally allows setting the user's superuser status.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -276,7 +276,7 @@ class UserSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = User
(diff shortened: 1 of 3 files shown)


madchap:
I think you won the privilege to fix 2 flake8 issues spotted by the GH action. Thanks in advance :)

valentijnscholten:
Thanks! If you really want to add value to the test, you'll have to check the result after the update call :-)

iwalton3:
@valentijnscholten It looks like the class the test inherits from (RESTEndpointTest) verifies the change.

valentijnscholten (approved):
ah yes, forgot about that! approving.

Maffooch (approved):

PR #3171, opened 2020-11-04
logging: improve variable naming for HANDLERs

valentijnscholten:
Recently somebody (from Switzerland ;-)) introduced a DD_LOGGING_FORMAT variable.
The variable is used to configure *handlers*  and not formatters, so this PR renames it to DD_LOGGING_HANDLER.
I considered making it backwards compatible, but couldn't find a straigforward way. The variable is new in 1.9.0 and I don't think may people are using it.
With the new `local_settings.py` setup people will have to review their settings anyway when upgrading.


madchap (approved):
The guy from Switzerland says OK ;-)

Maffooch (approved):

valentijnscholten:
label+=switzerland can't make up good variable names :-)

PR #3175, opened 2020-11-04
Fix reports: Add missing Q import

Maffooch:
dojo/reports/views.py was also missing a Q import


madchap:
What was the actual issue/problem occuring?

madchap (approved):

valentijnscholten:
We decided to include this as a bugfix in 1.9.3, which has been merged into dev as well. So cloing here.

PR #3183, opened 2020-11-06
Moving away from kubernetes to charts repo 

dsever:
kubernetes helm repo is not available anymore, so we are going to  use charts.helm.sh/stable


valentijnscholten:
this needs to go into master to fix broken GHA actions. will sync to dev later today with some other workflow changes.

madchap:
Should fix https://github.com/DefectDojo/django-DefectDojo/issues/3186

Issue #2226, opened 2020-04-20
display_tags.py throws exception when urls larger than 50 chars are specified as Endpoints


bhepp-sec:
I found an issue in /dojo/templatetags/**display_tags.py** and was hoping you could make the update without a pull-request. It's an fairly obvious line-liner. 

![image](https://user-images.githubusercontent.com/64037862/79789323-fe2bef00-8317-11ea-976c-54d3f376757a.png)

When the input **value** is a class, like Endpoint, and the path length is one, and the input **value** length is greater than 50, it throws with _django-DefectDojo/dojo/templatetags/display_tags.py", line 81, in url_shortner
    return_value = "..." + return_value[50:]
TypeError: 'Endpoint' object is not subscriptable_

Line 79 needs a 'str' added to **value**, like this:

![image](https://user-images.githubusercontent.com/64037862/79789769-b5286a80-8318-11ea-957a-78f97eafc78c.png)

namloc2001:
Hi, did anything progress with this?

madchap:
I'd say it would be good for you to open a PR for the one liner :-)

namloc2001:
> 
> 
> I'd say it would be good for you to open a PR for the one liner :-)

Done.

PR #3118, opened 2020-10-28
Update url_shortner to fix bugs and show more meaningful part of the endpoint

johnlanda:
Fixes #3117

PR #3184, opened 2020-11-06
Update display_tags.py

namloc2001:
As per #2226 .

When the input value is a class, like Endpoint, and the path length is one, and the input value length is greater than 50, it throws with django-DefectDojo/dojo/templatetags/display_tags.py", line 81, in url_shortner
return_value = "..." + return_value[50:]
TypeError: 'Endpoint' object is not subscriptable

Line needs a 'str' added to value.

### DIFF
--- a/dojo/templatetags/display_tags.py
+++ b/dojo/templatetags/display_tags.py
@@ -90,7 +90,7 @@ def url_shortner(value):
     if url.path:
         return_value = url.path
         if len(return_value) == 1:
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Thanks, but this was already fixed in dev in #3118

Maffooch:
Closing due to the issue already being fixed and this PR being based on `master` anyways.

PR #3195, opened 2020-11-07
Reduce default loglevel values to INFO

madchap:
Unsure why `DEBUG` was the default, `INFO` seems like a better start.


valentijnscholten:
I know everybody hates DEBUG, but I'd rather have DEBUG logs for every github issue that is opened ;-)

valentijnscholten (approved):

Maffooch (approved):

PR #3212, opened 2020-11-08
Check if Finding SLA enabled to run SLA notifications

madchap:
The async task would still run even if the Finding SLA enabled was actually disabled.


mtesauro (approved):
Approved

valentijnscholten (approved):

PR #3213, opened 2020-11-08
system setting: set random secret when jira secret is disabled and empty

### LINKED ISSUE #3211: Cannot save system settings if jira webhook secret is disabled
image

When this is the case, you cannot save the form.

valentijnscholten:
fixes #3211 

error happened when the jira secret was empty and the jira secret disable checkbox was checked.

It's because the field is mandatory but empty. The django javascript tries to focus the field and display the error. Not sure how to fix it. Workaround is to put some value in there.


madchap:
I am guessing no collateral damage of doing that. If people enable it afterwards, they would get the value pre-filled I assume and would not even have to click on the refresh icon. UX gain!

madchap (approved):

Maffooch (approved):

PR #3227, opened 2020-11-10
allow bulk delete in view_test

### LINKED ISSUE #3225: view_test: no bulk delete possible
in view_test there is no delete button when selecting multiple findings. (when viewing global open findings and other places it is possible to bulk delete selected findings)

valentijnscholten:
fixes #3225 

bulk delete was not available on `view_test` due to outdated permission check.

### DIFF
--- a/dojo/templates/dojo/view_test.html
+++ b/dojo/templates/dojo/view_test.html
@@ -215,7 +215,7 @@
                         </a>
                     </button>
                 {% endif %}
-                {% if user|is_authorized_for_delete:finding %}                    
+                {% if user.is_staff or 'AUTHORIZED_USERS_ALLOW_DELETE'|setting_enabled or 'AUTHORIZED_USERS_ALLOW_STAFF'|setting_enabled %}
                     <button type="button" class="btn btn-info btn-sm  btn-primary" data-toggle="tooltip" data-placement="bottom" title="Delete Findings">
                         <a class="white-color delete-bulk" href="#" alt="Delete Findings">
                             <i class="fa fa-trash"></i>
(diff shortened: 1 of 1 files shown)


Maffooch:
The old way uses the `user_is_athorized` function in `dojo/user/helper.py` and there are many other instances of `user|is_authorized_for_delete:finding`. I think the better way of fixing this is finding out why the that helper function denied access.

valentijnscholten:
good point. I just copied over a working version over the non-working version as I needed to use it often yesterday.

but the nonworking version is referring to a finding, but there is no finding in this case. this delete button is outside the list of findings and applies to the whole list. so that's why we need a more generic condition that looks at the users permissions in general.

Maffooch:
Give it the test then.

Maffooch (approved):

valentijnscholten:
what do you want to do with the other occurrences of the "old" construct?

Maffooch:
I should clarify. Using the `user_is_athorized` function is the "newer" way of doing things, but I described it as old because this PR was changing it to something new. 

A couple months ago, I went through and changed most of the `if user.is_staff` type of checks to use this decorator that you wrote some months back. This one looks to be one of my signature typos from my PRs.

Maffooch:
Also look on line 203 of this file to see the same check with the correct value.

valentijnscholten:
I know, but I was talking about these: 
image
But there you don't always have a / one product.

Maffooch:
Yeah those are instances where there is just no object to check the relationships.

madchap (approved):

PR #3228, opened 2020-11-10
Add finding > engagement and engagement > status filters.

iwalton3:
This adds the following new filters:
 - Allow viewing findings from a particular engagement or set of engagements.
 - Allow filtering the list of engagements by status.


valentijnscholten (changes requested):
Can you make sure ENGAGEMENT_STATUS_CHOICES  and/or the values in that are only in one place in the codebase?

valentijnscholten (approved):

Maffooch (approved):

PR #3229, opened 2020-11-10
Fix CWE URL on the finding page.

iwalton3:
The CWE link on the finding page is missing a `"`, which causes the link to go to the wrong location and also malforms the HTML on the page.


valentijnscholten (approved):
nice find thanks

Maffooch (approved):

PR #3238, opened 2020-11-10
Add bitnami to helm repo list

madchap:
to support testing and eventually passing of https://github.com/DefectDojo/django-DefectDojo/pull/2859

This add should not change anything, since the actual chart taken into consideration is in requirements.yaml


madchap:
cc @dsever

madchap:
Used powers to merge, with no review :crossed_fingers:

PR #3244, opened 2020-11-11
Apply changes to actions k8s

dsever:
Applies changes against last version in master


valentijnscholten:
looks like master already has the right version of the workflow and it's just that github doesn't always pick it up in PRs even when merged into dev etc.

Issue #3245, opened 2020-11-11
View finding: CWE link contains " at the end


Seppl2202:
**Bug description**
When viewing a finding and following the CWE link, this link includes a " at the end: https://cwe.mitre.org/data/definitions/200.html%22

**Steps to reproduce**
1. Go to 'View finding'
2. Click on 'CWE link'
4. See error 'Not found' due to the appended "

**Expected behavior**
The page should be found and displayed

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [X] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: [Ubuntu]
 - DefectDojo Commit Message: [use `git show -s --format="[%ci] %h: %s [%d]"`]

**Additional context** (optional)
The problem is templates/dojo/view_finding.html: 329:

` <a target="_blank" href={{ finding.cwe|cwe_url }}">`
As you can see, there is a quote in the `href` which gets appended.

Maffooch:
Fixed in `dev` already. Thanks for reporting!

PR #3246, opened 2020-11-11
removed quotes from CWE link

Seppl2202:
This is the "fix" for #3245


Maffooch:
This was fixed in `dev` already. In the future, please base your PRs against `dev` instead of `master`

valentijnscholten:
fixed by #3229

PR #3247, opened 2020-11-11
Add BlackDuck import functionality for License Risks that should be reviewed

WheelsVT:
Currently, only "In Violation" license issues are imported by the BlackDuck importer, and with an auto-severity of High. Unfortunately, I believe 3rd party BlackDuck scans are often missing policies that would make a violation determination, which leads to all components being marked "Not In Violation", and subsequently NO license risks being imported in DefectDojo.

This PR adds functionality to import potential license risks for review with a title of "Review License Risk: XXX". This occurs when the import encounters  "Not In Violation" licenses with License Risk values other than "OK", and enters these new review findings into DefectDojo at their respective severity.

Also added origin and match type info to the license risk description field to assist remediation efforts.

Added parsing of BlackDuck sources.csv and dictionary for matching path and archive context to license finding descriptions.


Maffooch:
Hi @WheelsVT in the future please start and base your PRs against the `dev` branch. I will try to rebase this one to the `dev` but not totally sure if the commits will play nicely. Worst case, you'll have to make another PR from the `dev` branch instead of `master`.

Maffooch:
Looks like it worked out all right this time :)

valentijnscholten:
Thanks! One of the most common cause for 500 errors is parsers that access properties / dictionary entries that aren't actually present in the report. Could you double check and if anywhere in the parser it would be better to replace `component['blabla']` with something like `component.get('blabla')` ?

WheelsVT:
Thanks. That should cover the updates to address the requests, and I added in some additional parsing of the sources.csv file so that location information can be included in the finding description (otherwise it can be difficult to identify where the license finding is originating).

devGregA:
I'm wondering if v3 is backwards compatible with v2... Reviewing this week.

madchap:
@devGregA did you get a chance to review?

devGregA:
@madchap sorry, will do it tomorrow.

Maffooch (approved):

devGregA:
So, it looks like our sample isn't actually the right format. Will have another update today.

devGregA:
Okay so our current integration is with BlackDuck Hub? Is this also Hub or Codescan or?

devGregA:
@WheelsVT please see previous comment. Just trying to decide if we need to make a distinction on blackduck scanners.

WheelsVT:
@devGregA this is what I used for the Component Risk scan. I actually don't have the BlackDuck platform myself, just a lot of output to process.

devGregA (approved):

devGregA:
I went ahead and approved this PR to not keep you waiting. I believe that there are different versions of black duck that we may need distinctions around. A sample scan would be helpful to keep the tooling working, if you are able to provide one here: https://github.com/DefectDojo/sample-scan-files Thank you for your contribution.

PR #3250, opened 2020-11-12
Reintroduce HTML report builder

Maffooch:
The report builder used to support configurable PDF reports though the Wkhtmltopdf package. Since modern browsers can now print to PDF, it made sense to just create an HTML version of the report builder. So here it is.


valentijnscholten:
Sounds like a good idea. In the light of recent errors around reporting, could you add some sort of tests to the integration tests?

valentijnscholten (approved):

devGregA (approved):

PR #3254, opened 2020-11-12
API_V2 : Add metadata operation on findings endpoints

RomainJufer:
The goal of this PR is to support metadata operations directly on the findings endpoints to avoid having to perform double queries. More specifically, by introducing `/findings/{id}/metadata` we can directly interact with the metadata of a given findings without having to first retrieve its id and then perform the operation. This is quite similar to what is already done with the notes and the tags.

Concretely, we introduce the methods `POST`, `GET`, `PUT`, `DELETE` on `/findings/{id}/metadata`. We decided to not support `PATCH` as a metadata, for a given finding, is only a tuple (name, value) and the name is required to identify the metadata on which we want to perform the operation. Hence, we don't think that there is any advantage is supporting `PATCH` at the moment.

This PR also add the correct swagger doc for the new endpoints as well as a new test case in `test_rest_framework.py`.

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -580,7 +580,7 @@ class FindingViewSet(mixins.ListModelMixin,
         responses={status.HTTP_200_OK: ""},
         methods=['delete'],
         manual_parameters=[openapi.Parameter(
-            name="name", in_=openapi.IN_QUERY, required=True, type=openapi.TYPE_STRING, 
+            name="name", in_=openapi.IN_QUERY, required=True, type=openapi.TYPE_STRING,
             description="name of the metadata to retrieve. If name is empty, return all the \
                             metadata associated with the finding")]
     )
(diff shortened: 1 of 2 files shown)


valentijnscholten:
to make sure I understand: what can this do more than the `/metadata/` endpoint where you can also specify a `finding` as a parameter?

RomainJufer:
In the metadata endpoint you have to provide the metadata id and the finding id. Here, the finding id is implicitly provided by the endpoint query parameter and you provide the name instead of the metadata id. This is valid since a metadata is uniquely defined by the tuple (finding_id, name). That way you avoid having to first get the metadata id and then perform the operation. Let me know if this is still not clear

valentijnscholten:
As dicsussed yesterday I really like the API being improved and the swagger spec being fixed. It does feel like a lot of code just to expose some meta data on a finding, but maybe that's just the way "REST API life" is. If in anyway possible it would be nice to have something generic so that if we decide to expose something else, maybe risk acceptance for a finding it would be able to reuse some parts in a parametrized way and not having to copy-and-paste 100 lines of code and just changing a few model names. But it might not possible, and something you already considered.

valentijnscholten (approved):

devGregA (approved):

Maffooch (approved):

devGregA:
This is an excellent PR! Thank you @RomainJufer!

I agree with Valentijn that a larger refactor would be helpful to reduce code reuse, but that is an effort the group can take up later :)

devGregA:
Forgot to squash :( sorry about that.

Issue #3223, opened 2020-11-10
Error when using search filter with CVE


Wadeck:
While inside the search page, I tried to search with the advanced filter feature. When providing the CVE value in the CVE field, I got an error.

<details>

<summary>Screenshots</summary>

### Regular search page
![Screenshot_2020-11-10_114634_001](https://user-images.githubusercontent.com/2662497/98664820-267a4a00-234b-11eb-9e25-d8567a8fadca.png)

### After filter opened
![Screenshot_2020-11-10_114645_001](https://user-images.githubusercontent.com/2662497/98664829-27ab7700-234b-11eb-9eca-82e192ad4f00.png)

### After click on "Apply Filters"
![Screenshot_2020-11-10_114651_001](https://user-images.githubusercontent.com/2662497/98664857-32fea280-234b-11eb-8bac-617fddf4d095.png)

(url: https://<defect-dojo-url>/simple_search?finding-title=&finding-date=&finding-cve=CVE-20XX-XX51&finding-cvssv3=&finding-steps_to_reproduce=&finding-severity_justification=&finding-active=unknown&finding-verified=unknown&finding-false_p=unknown&finding-duplicate=&finding-out_of_scope=unknown&finding-under_review=unknown&finding-review_requested_by=&finding-under_defect_review=unknown&finding-defect_review_requested_by=&finding-is_Mitigated=unknown&finding-last_reviewed_by=&finding-sourcefilepath=&finding-param=&finding-payload=&finding-file_path=&finding-component_name=&finding-component_version=&finding-static_finding=unknown&finding-dynamic_finding=unknown&finding-sonarqube_issue=&finding-unique_id_from_tool=&finding-sast_source_object=&finding-sast_sink_object=&finding-sast_source_line=&finding-sast_source_file_path=&finding-nb_occurences=&finding-last_reviewed=&finding-test__engagement__risk_acceptance=&finding-has_jira_issue=unknown&finding-jira_issue__jira_key=&finding-has_notes=unknown&finding-o=&query=CVE-20XX-XX51)

</details>

**Expected behavior**
A regular search result page with the filter being applied.

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [ ] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: [e.g. Ubuntu 18.04]
 - DefectDojo Commit Message: [use `git show -s --format="[%ci] %h: %s [%d]"`]

[...]

PR #3269, opened 2020-11-16
uwsgi: increase default buffer-size

### LINKED ISSUE #3241: Cannot submit comment after having used filter for "Similar Findings"
Browse a finding with Similar Findings. Add a comment at the bottom of the page, it's working fine. Now, open the "Similar Findings" panel, change a bit the filters and click on "Apply Filters". You will see the URL changes with all the fields being inside (that could be the root cause).

Now, within that page with long URL, try adding a comment. I got a NGinx error.

**Expected behavior**
I do not expect any effect of the filters used on the ability to submit comments.

- [ ] Kubernetes
- [x] Docker

**Environment information**

**Screenshots** (optional)

**Console logs** (optional)

cc @valentijnscholten @madchap

valentijnscholten:
Some pages with complex filter sets get result in requests being over 4096 bytes which trips uwsgi into blocking the request due to the default 4096 buffer-size introduced recently.

fixes #3241 #3223


mtesauro:
100% agree with this - upped the defaults to that on my install after noticing some issues and hadn't gotten a chance to do a PR like this.  Nice job.

mtesauro (approved):
Approved

valentijnscholten:
if only k8s would build with it in GHA...

valentijnscholten:
> if only k8s would build with it in GHA...

it does now, saw some weird docker errors in GHA for about an hour.

Maffooch (approved):

PR #3275, opened 2020-11-17
Set flag for auto refresh of alert/counts

Maffooch:
Add the option to disable the notification recount through the `alerts/count` endpoint. Also add a manage.py command to remove alerts from the database with the options of removing the entire table, just system alerts, or alerts belonging to specific users. They tend to add up quickly with many users!

### DIFF
--- a/dojo/management/commands/clear_alerts.py
+++ b/dojo/management/commands/clear_alerts.py
@@ -1,0 +1,39 @@
+from django.core.management.base import BaseCommand
+from dojo.models import Alerts, Dojo_User
+
+"""
+Author: Cody Maffucci
+This script will remove all alerts in a few different ways
+all: Remove all alerts from the database
+user: Clear alerts for a given user
+system: Clear system alert
+"""
+
+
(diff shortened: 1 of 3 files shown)


valentijnscholten:
I thought about this as well, what is your reason to disable it? I get constant logging messages, but that could be resolved otherwise.

Maffooch:
It ties up a thread. When multiple users are logged it once, the `alerts/count` endpoint is hit every 10 seconds for each user logged in. It adds up quickly.

valentijnscholten:
maybe at some point we could look at websockets: https://medium.com/@ksarthak4ever/django-websockets-and-channels-85b7d5e59dda

valentijnscholten:
I think it should be on by default though, especially when using JIRA any errors will occur there. if people start F5-ing all the time your server load might be even higher?

Maffooch:
Oh darn. I intended to default to true but forgot to change it back after testing. Good catch.

valentijnscholten (approved):

PR #3279, opened 2020-11-17
CI/CD engagements auto-deletion

madchap:
In today's world, people (like me) have DefectDojo ingest a ton of reports through CI pipelines and do not use the re-import feature in order to keep track overtime of findings as they relate to various docker images tags for example, ending up with some 1 million findings in the database, where most of them are plain duplicates with no added value. This has adverse effects on performance, but also usability.

Say your pipeline runs some Anchore scans multiple times a day. Your pipeline tags the images different of course, and everytime you build an image, you scan it and results get uploaded to defectdojo. Chances are that most if not all scans bear no difference with a scan that got imported the previous day. 100% duplicate. You do not want to have your engineering teams do custom things in the pipelines too, as it add increased maintenance burden overtime. So you swallow them all.

This PR is about keeping this in check.

It introduces system settings such as and to delete "useless" engagements. It is **disabled by default**.

image

The conditions (to be documented) are hard-coded for some part in the utils.py file. Only to be considered for auto-deletion, engagements that:
- Have been created before today - x days (field to be renamed to make more sense in screenshot above)
- Are CI/CD engagements
- Must only contain duplicate findings
- Must not contain any notes (such as a human could have put a note there)
- Does not match the "lock tag" in system settings

It's got its own notification type, such as

image

and at product level

image


dojo/utils.py
```diff
@@ -1834,3 +1834,60 @@ def create_bleached_link(url, title):
...
+            event='auto_delete_engagement',
+            title=engagement_title,
+            id=engagement_id,
+        )
+
+    system_settings = System_Settings.objects.get()
+    if system_settings.engagement_auto_delete_enable:
+        # how to not exclude the tag when not empty? If empty, then query results are unexpected.
+        # setting arbitrary string for now, which is unlikely to be a used tag.
+        lock_tag = system_settings.engagement_auto_delete_lock_tag or 'qAEH2HL6Qd9ofZYLCGykN2WQ'
```
madchap, line 1862: @valentijnscholten aside your regular feedback, a quick tip on that?

valentijnscholten: what's the question

alles-klar:
Great idea, I have the same problem. But I am using the delete duplicate function, so I have a lot of engagements with zero findings. 

I would prefer a solution that is similar to duplicated findings: Find "duplicated engagements" - for example only the date differs - and that an option to delete some of them.

valentijnscholten:
The existing cleanup job could be extended with some system settings options:

- only delete duplicates (existing option)
- only delete aside xxx  duplicates (existing option)
- only delete findings untouched for XXX days
- only delete findings without notes
- also delete empty tests
- also delete empty engagements
- only delete inactive mitigated findings (so not only focus on duplicates)

Maybe add a 'protected' flag to engagements (or all defect dojo models) that prevents automatic (or accidental) deletion.

mtesauro:
Maybe change "Engagement Auto-Deletion Days" to "Engagement Auto-Deletion TTL" aka Time To Live.

Also, the PR says "CI/CD engagements auto-deletion" but if I were to see the included settings UI, how would I know that these only apply to CI/CD engagements vs engagements in general?

madchap:
> Maybe change "Engagement Auto-Deletion Days" to "Engagement Auto-Deletion TTL" aka Time To Live.

Why not. Just wanted to align on the auto-close engagements verbiage.

 
> Also, the PR says "CI/CD engagements auto-deletion" but if I were to see the included settings UI, how would I know that these only apply to CI/CD engagements vs engagements in general?

Currently, this is indicated in the `?` hover.

madchap:
I am going to close this, and let's then think on a better way then.

PR #3200, opened 2020-11-07
JIRA: Allow config per engagement, incl big JIRA refactor

valentijnscholten:
To allow users to have some flexibility in their JIRA setup  / mapping to Defect Dojo, we'd decided to allow users to the configure a JIRA project on the engagement level. This is an optional config that will override the config set on the product above it.

While implementing it I came across some logic related to this that I wanted to centralize and make it more DRY. While doing it I realized I might as well put all the JIRA related stuff in one `helper` file. And clean up some of the logic, improve connection handling, error handling, exception handling, avoid swallowing exceptions.

As I had to test "all things JIRA" I also fixed a couple of things regarding the "push all issues" stuff that has been haunting us after it was merged. Some other things found along the way and fixed:

#3205
#3206
#2664 
#3224
#3240
#3261

I also added a check when configuring a JIRA project to make sure Defect Dojo can retrieve the metadata for that JIRA project. We see quit a bit of issues on github that relate to empty metadata due to JIRA permission errors throwing a stacktrace. That is now fixed and handled more gently with error messages in the UI.

So basically this is a big refactor of the JIRA stuff, so I did extensive manual testing:

	add finding with jira enabled, without jira enabled, with push_all, without push_all
	add ad hoc finding with jira enabled, without jira enabled, with push_all, without push_all
	edit finding with jira enabled, without jira enabled, with push_all, without push_all
	upload report with jira enabled, without jira enabled, with push_all, without push_all
	reimport report with jira enabled, without jira enabled, with push_all, without push_all
	api upload report with jira enabled, without jira enabled, with push_all, without push_all
	api reupload report with jira enabled, without jira enabled, with push_all, without push_all
	api add finding with jira enabled, without jira enabled, with push_all, without push_all
	api add finding with push to jira with jira enabled, without jira enabled, with push_all, without push_all
	api add finding with push_all with jira enabled, without jira enabled, with push_all, without push_all
	push epic from engagements
	create /edit epic without jira enabled
	api patch finding with push_all with jira enabled, without jira enabled, with push_all, without push_all
	check api response for findings
        add /edit jira config for product
        add / edit jira config for engagement
        check used jira config if overridden by engagement
        webhook (added automated tests case)

There is a somewhat complex migration for which it would be really helpful if people could test that. Apart from that, it's not a PR that you can just easily approve by looking at the code for 5 minutes. Sorry for that ;-)

[...]

PR #3284, opened 2020-11-17
jira: split url handling for issues and projects

valentijnscholten:
in #3200 there was a `jira_url` filter that was (supposed to be) "smart". Turns out there are many variations of findings with a jira issue or findings wanting to display a jira project url if they have no issue, but only on the edit screen, but not in the findings list, etc. etc. 

so this PR splits out the `jira_url` filter into two seperate filters so it's clear what it does and will or will not display.

thanks @madchap for reporting.

### DIFF
--- a/dojo/forms.py
+++ b/dojo/forms.py
@@ -2087,7 +2087,7 @@ class JIRAProjectForm(forms.ModelForm):
         self.product = kwargs.pop('product', None)
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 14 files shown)


madchap (approved):

mtesauro:
I think this is good for now.  Agreed that it could probably be improved.  Added to my getting longer "I should look at this more" list.

mtesauro (approved):
Approved

PR #3285, opened 2020-11-17
apiv2: set mitigated date if applicable

keenan-v1:
- Sets or clears mitigated date based on the value of is_Mitigated
- Sets or clears mitigated_by based on the value of is_Mitigated and current user context

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -789,9 +789,16 @@ class FindingCreateSerializer(TaggitSerializer, serializers.ModelSerializer):
 
         push_to_jira = push_to_jira or new_finding.get_push_all_to_jira()
 
+        # Allow setting the mitigation date based upon the state of is_Mitigated.
+        if new_finding.is_Mitigated:
+            new_finding.mitigated = datetime.datetime.now()
+            new_finding.mitigated_by = self.context['request'].user
+            if settings.USE_TZ:
+                new_finding.mitigated = timezone.make_aware(new_finding.mitigated, timezone.get_default_timezone())
+
         # If we need to push to JIRA, an extra save call is needed.
         # TODO try to combine create and save, but for now I'm just fixing a bug and don't want to change to much
-        if push_to_jira:
+        if push_to_jira or new_finding.is_Mitigated:
             new_finding.save(push_to_jira=push_to_jira)
 
         # not sure why we are returning a tag_object, but don't want to change too much now as we're just fixing a bug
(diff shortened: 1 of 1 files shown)


valentijnscholten (changes requested):
thanks. i was confused a bit, but the logic looks ok. could you make it so we have 2 booleans and call save if one of them is True. so 'push_to_jira' and 'set_mitigated'? now if you quickyl read it could make readers think in some cases we will not save the instance at all due to the naming of the boolean.

valentijnscholten:
should the same logic be in perform_create? if somebody creates a finding that is already marked as mitigated?

dojo/api_v2/serializers.py
```diff
@@ -693,9 +693,25 @@ def update(self, instance, validated_data):
...
+        save_instance = push_to_jira
+
+        # Allow setting or clearing the mitigation date based upon the state of is_Mitigated.
+        if instance.is_Mitigated and instance.mitigated is None:
+            save_instance = True
+            instance.mitigated = datetime.datetime.now()
+            instance.mitigated_by = self.context['request'].user
+            if settings.USE_TZ:
+                instance.mitigated = timezone.make_aware(instance.mitigated, timezone.get_default_timezone())
+        elif not instance.is_Mitigated and instance.mitigation is not None:
```
valentijnscholten, line 706: should this check for `instance.mitigated` instead of `instance.mitigation` ?

keenan-v1: Good catch. My testing didn't catch this. Will fix.

keenan-v1:
I've added the requested changes and fixed the bug in variable naming. Tested the functionality and it is working as intended.

valentijnscholten (approved):

valentijnscholten:
thanks for the PR

madchap:
As a push to jira will actually overwrite the jira description field, this update may come as a surprise to some users. On the UI, there is the `?` that warns about it, but in this case it will just do it.

keenan-v1:
> 
> 
> As a push to jira will actually overwrite the jira description field, this update may come as a surprise to some users. On the UI, there is the `?` that warns about it, but in this case it will just do it.

Isn't that more a problem with the push-to-jira patch and less of a problem with this one? I'm not entirely sure how the Jira stuff works as I don't use it yet. I simply worked around the Jira code here as it was overriding the same areas as my original fix for the mitigated date did.

madchap (approved):

PR #3286, opened 2020-11-17
Fix finding export for non-staff users.

iwalton3:
The finding export currently behaves differently depending on user privileges because the column headers and export options were not kept in sync. This version fixes it so that the reporting format is consistent and does not include garbage columns in the export.

This fixes the issue for the two exports which vary based on user privileges. All the other ones I found did not appear to have this issue.


madchap (approved):

valentijnscholten (approved):
only just understood what this was about :-) good PR.

PR #3288, opened 2020-11-17
reimport: set component_name&version on existing findings

valentijnscholten:
in the past findings didn't have a `component_name` and `component_version` fields. when reimporting a scan these existing findings are now updated with these fields to get these populated.


Maffooch (approved):
Sorry, no enhancements allowed the week before a release 😉

madchap (approved):

Issue #3290, opened 2020-11-17
Stacktrace when receiving jira webhook


madchap:
When receiving a webhook from JIRA, I get the following stacktrace for issues that do not have a match on the defectdojo side.
This causes errors 500 to be produced in the logs as well as just quite a bit of log pollution.

The functionality does not seem to be impacted.

```
ERROR:django.request:Internal Server Error: /jira/webhook/xxxxxxxxx-fbe6-xxxx-9585-50460xxxxxxxx
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.6/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "./dojo/jira_link/views.py", line 111, in webhook
    jissue = JIRA_Issue.objects.get(jira_id=jid)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py", line 408, in get
    self.model._meta.object_name
dojo.models.JIRA_Issue.DoesNotExist: JIRA_Issue matching query does not exist.
```

500, and not 404 errors.

PR #3292, opened 2020-11-17
Add try-catch for jira webhook processing

madchap:
Fixes https://github.com/DefectDojo/django-DefectDojo/issues/3290

### DIFF
--- a/dojo/jira_link/views.py
+++ b/dojo/jira_link/views.py
@@ -53,99 +53,106 @@ def webhook(request, secret=None):
     if request.method == 'POST':
         # logger.debug('jira_webhook_body:')
         # logger.debug(request.body.decode('utf-8'))
-        parsed = json.loads(request.body.decode('utf-8'))
-        if parsed.get('webhookEvent') == 'jira:issue_updated':
-            jid = parsed['issue']['id']
-            jissue = get_object_or_404(JIRA_Issue, jira_id=jid)
-            if jissue.finding is not None:
-                finding = jissue.finding
-                jira_instance = jira_helper.get_jira_instance(finding)
-                resolved = True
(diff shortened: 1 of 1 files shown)


valentijnscholten (changes requested):
Good to improve the webhook, but I don't like the idea of swallowing exceptions and just return a 200. There may be a temporary problem for which we would like JIRA to retry the webhook, so better to return a 500 when there is an exception.

I also would rather see new method on the jira helper class to find a JIRA_Issue instance in a safe way instead of this huge try-catch thing.

madchap:
> Good to improve the webhook, but I don't like the idea of swallowing exceptions and just return a 200. There may be a temporary problem for which we would like JIRA to retry the webhook, so better to return a 500 when there is an exception.

I am not sure how you'd tell JIRA how to retry a webhook. It would actually have to be supported by JIRA, which I don't think it is looking at the configuration there.

But we can return something other than 200, I agree.
 
> I also would rather see new method on the jira helper class to find a JIRA_Issue instance in a safe way instead of this huge try-catch thing.

Yes, I will see what I can do.

valentijnscholten:
the "official" way to log exceptions is `logger.exception()`

valentijnscholten:
I would also look at what is causing the exception. There is a JIRA_Issue but no finding attached. That means:

- it is an epic attached to an engagement
- somebody removed findings manually without removing JIRA_Issue instances

The first one is a valid case and could use some proper handling (just skip it without errors or 404).
The second one is corner case. If the exception is catched only to log it, it is no different than just letting the exception happen?

madchap:
Superseded by https://github.com/DefectDojo/django-DefectDojo/pull/3321

PR #3295, opened 2020-11-17
Allow re-import scan to function without JIRA

Maffooch:
Attempting to re-import a scan does not work when there is no valid Jira config. I also made a slight performance change to the jira_helper stuff too, but nothing major.

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -77,7 +77,7 @@ def get_jira_project(obj, use_inheritance=True):
 
     if isinstance(obj, Finding):
         finding = obj
-        return get_jira_project(finding.test.engagement)
+        return get_jira_project(finding.test)
 
     if isinstance(obj, Test):
         test = obj
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -77,7 +77,7 @@ def get_jira_project(obj, use_inheritance=True):
 
     if isinstance(obj, Finding):
         finding = obj
-        return get_jira_project(finding.test)
+        return get_jira_project(finding.test.engagement)
 
     if isinstance(obj, Test):
         test = obj
--- a/dojo/test/views.py
+++ b/dojo/test/views.py
@@ -1,44 +1,40 @@
-#  engagements
+# #  tests
+
 import logging
-from datetime import datetime
 import operator
+import json
+import httplib2
 import base64
-from django.contrib.auth.models import User
+from datetime import datetime
+import googleapiclient.discovery
+from google.oauth2 import service_account
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.decorators import user_passes_test
-from django.core.exceptions import PermissionDenied
+from django.core.exceptions import ValidationError
 from django.urls import reverse
-from django.db.models import Q, Count
-from django.http import HttpResponseRedirect, StreamingHttpResponse, Http404, HttpResponse, FileResponse
+from django.db.models import Q, QuerySet, Count
+from django.http import HttpResponseRedirect, Http404, HttpResponse
 from django.shortcuts import render, get_object_or_404
 from django.views.decorators.cache import cache_page
 from django.utils import timezone
-from time import strftime
 from django.contrib.admin.utils import NestedObjects
 from django.db import DEFAULT_DB_ALIAS
+from tagging.models import Tag
 
-from dojo.engagement.services import close_engagement, reopen_engagement
-from dojo.filters import EngagementFilter
-from dojo.forms import CheckForm, \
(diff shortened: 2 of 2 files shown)


devGregA (approved):

mtesauro (approved):
Approved

dojo/jira_link/helper.py
```diff
@@ -77,7 +77,7 @@ def get_jira_project(obj, use_inheritance=True):
 
     if isinstance(obj, Finding):
         finding = obj
-        return get_jira_project(finding.test)
+        return get_jira_project(finding.test.engagement)
```
valentijnscholten, line 80: my idea was to have the cascading of jira_project up through the hiearchie in one place only so I decided to go from finding to test and not straight to engagement. if we ever get jira config per test, we again will have to modify multiple places.

dojo/test/views.py
```diff
@@ -633,19 +633,19 @@ def re_import_scan_results(request, tid):
     form = ReImportScanForm()
     jform = None
     jira_project = jira_helper.get_jira_project(test)
-    push_all_jira_issues = jira_helper.is_push_all_issues(test)
+    push_all_jira_issues = jira_project.push_all_issues if jira_project else False
```
valentijnscholten, line 636: the idea was to have the logic around push_all_jira_issues in one place or at least in the helper, so I don't think this change is needed.

dojo/test/views.py
```diff
@@ -633,19 +633,19 @@ def re_import_scan_results(request, tid):
...
 
     form.initial['tags'] = [tag.name for tag in test.tags]
     if request.method == "POST":
         form = ReImportScanForm(request.POST, request.FILES)
-        if jira_helper.get_jira_project(test):
+        if jira_project:
             jform = JIRAImportScanForm(request.POST, push_all=push_all_jira_issues, prefix='jiraform')
 
-        if form.is_valid() and jform.is_valid():
+        if form.is_valid():
```
valentijnscholten, line 648: i think it's better to do the same as in import_scan_results, so
```
        if form.is_valid() and (jform is None or jform.is_valid()):
```

dojo/test/views.py
```diff
@@ -692,7 +692,9 @@ def re_import_scan_results(request, tid):
 
                 # can't use helper as when push_all_jira_issues is True, the checkbox gets disabled and is always false
                 # push_to_jira = jira_helper.is_push_to_jira(new_finding, jform.cleaned_data.get('push_to_jira'))
-                push_to_jira = push_all_jira_issues or (jform and jform.cleaned_data.get('push_to_jira'))
+                push_to_jira = False
+                if jform and jform.is_valid():
+                    push_to_jira = push_all_jira_issues or (jform and jform.cleaned_data.get('push_to_jira'))
```
valentijnscholten, line 697: changes is not needed

valentijnscholten (changes requested):
Thanks Cody, this must have slipped through the cracks. Show again we should try to avoid logic in multiple places. This code is almost the same as the normal import and that again is almost the same as the import in the apiv2 and reimport in apiv2. 

I modified the PR a bit to make the reimport_scan here align as much as possible with import_scan so there differences are minimal.

Hope you don't mind. I have set it to changes requests to make it clear I'd like to run it past you first before somebody merges it.

dojo/jira_link/helper.py
Maffooch: I can understand why this was done this way with the possible intro of jira for tests, but otherwise its just multiple calls to the same function where one could do the trick. But we can keep it the way it is I suppose.

dojo/test/views.py
Maffooch: It's redundant. `is_push_all_issue` just calls `get_jira_project` to check the db field `push_all_issues`. The object is already here so why not use it?

valentijnscholten:
if the internals of `push_all_issues` changes, we would have to update all the places where it is used.

dojo/test/views.py
valentijnscholten: if the internals of is_push_all_issues change, we would have to update all the places where it is used.

Maffooch: I would hope the db fields don't change often enough for that to be an issue 😉 perhaps it can be used as a deterrent to leave them alone. But if you insist it's to be done this way then let's merge.

valentijnscholten: well at some point it was considered to make the `push_all_issues` on product level always overrule that on the engagement level. another thing could be the api where there could be logic around the push_to_jira coming in from the api request needs to be combined with the push_all_issue flag in the db. Does the incoming request overrule that, or not? Logic like that I'd like to have it in only 1 place

Maffooch: Alright fair enough.

valentijnscholten (approved):

Issue #3268, opened 2020-11-16
everytime i import nikto scan i get Server Error (500) same for trufflehog


AliElKhatteb:
**Bug description**
A clear and concise description of what the bug is. For errors include at least the exact error message you are seeing (including traceback).

**Steps to reproduce**
1. Go to 'finding'
2. Click on 'import scan results'
3. scan type nikto scan
4. import

**Expected behavior**
its should be import but its not
**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [ X] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: [e.g. Ubuntu 20.04]

**Sample scan files** (optional)
?xml version="1.0" ?>
<!DOCTYPE niktoscan SYSTEM "/usr/share/doc/nikto/nikto.dtd">
<niktoscan hoststest="0" options="-h 127.0.0.1 -p 8070 -output nikto-output.xml" version="2.1.5" scanstart="Mon Nov 16 03:45:06 2020" scanend="Wed Dec 31 19:00:00 1969" scanelapsed=" seconds" nxmlversion="1.2">

<scandetails targetip="127.0.0.1" targethostname="localhost" targetport="8070" targetbanner="" starttime="2020-11-16 03:45:07" sitename="http://localhost:8070" siteip="http://127.0.0.1:8070" hostheader="localhost" errors="0" checks="6544">

<item id="999976" osvdbid="0" osvdblink="http://osvdb.org/0" method="GET">
<description><![CDATA[The anti-clickjacking X-Frame-Options header is not present.]]></description>
<uri><![CDATA[/]]></uri>
<namelink><![CDATA[http://localhost:8070/]]></namelink>
<iplink><![CDATA[http://127.0.0.1:8070/]]></iplink>
</item>

<item id="999984" osvdbid="0" osvdblink="http://osvdb.org/0" method="GET">
<description><![CDATA[Server leaks inodes via ETags, header found with file /favicon.ico, fields: 0xW/21630 0x1602011686000 ]]></description>
<uri><![CDATA[/favicon.ico]]></uri>
<namelink><![CDATA[http://localhost:8070/favicon.ico]]></namelink>
<iplink><![CDATA[http://127.0.0.1:8070/favicon.ico]]></iplink>
</item>

<item id="500008" osvdbid="39272" osvdblink="http://osvdb.org/39272" method="GET">
[...]

PR #3318, opened 2020-11-19
Nikto quick fix

madchap:
Fixes nikto issue reported in https://github.com/DefectDojo/django-DefectDojo/issues/3268

The regex probably needs rework. This PR is to get a working nikto parser before the next release.

### DIFF
--- a/dojo/tools/nikto/parser.py
+++ b/dojo/tools/nikto/parser.py
@@ -4,9 +4,11 @@ import re
 from defusedxml import ElementTree as ET
 import hashlib
 from urllib.parse import urlparse
+import logging
 
 from dojo.models import Finding, Endpoint
 
+logger = logging.getLogger(__name__)
 
 class NiktoXMLParser(object):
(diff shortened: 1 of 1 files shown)


madchap:
The flake8 may/will actually fail because of the regex (which sometimes work).

valentijnscholten:
probably caused by this value in the problematic report:
`<iplink><![CDATA[http://:/favicon.ico]]></iplink>`
maybe include that in the unit tests xml file so if we refactor it in the future, we'll see if it still works with this (stupid) url?

madchap:
I am actually not seeing any unit tests for nikto.

mtesauro (approved):
Approved

valentijnscholten:
hmm no unit tests, then we should merge. Usually I try to but the data causing the bug in the unit test which then proves that it is fixed. and stays fixed.

valentijnscholten (approved):

PR #3324, opened 2020-11-20
tags: add testcases

valentijnscholten:
in the near future I am going to work on replacing the `django-taggit` library with `django-tagulous`.

to capture the current behavior, I have added some more test cases around tags. this will help with regression testing after the move to `django-tagulous`


mtesauro (approved):
Approved

PR #3325, opened 2020-11-20
Add duplicate finding support to API v2.

iwalton3:
This exposes the duplicate finding functionality through the REST v2 API, as currently there is no way to (that I know of) reassign duplicate findings while preserving the duplicate cluster through the API.

The following endpoints are added:

 - `GET ​/findings​/{id}​/duplicate​/` - Returns the contents of `duplicate_cluster`.
 - `POST ​/findings​/{id}​/duplicate​/reset​/` - Resets the duplicate status of a finding.
 - `POST ​/findings​/{id}​/original​/{new_fid}​/` - Allows reassignment of duplicate findings.

The three endpoints use logic from the already existing `dojo/finding/views.py` file instead of creating duplicate code.

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -23,6 +23,8 @@ from dojo.models import Product, Product_Type, Engagement, Test, Test_Type, Find
 
 from dojo.endpoint.views import get_endpoint_ids
 from dojo.reports.views import report_url_resolver
+from dojo.finding.views import set_finding_as_original_internal, reset_finding_duplicate_status_internal, \
+    duplicate_cluster
 from dojo.filters import ReportFindingFilter, ReportAuthedFindingFilter, ApiFindingFilter, ApiProductFilter
 from dojo.risk_acceptance import api as ra_api
 from dateutil.relativedelta import relativedelta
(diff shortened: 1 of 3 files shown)


valentijnscholten:
Looks like a good addition to the API. Did you split off the `messages.add` stuff because it was causing issues with request via the API? 
Could you look at "adding something" to `test_rest_framework.py` to trigger this logic / endpoints?

iwalton3:
I removed the messages from the API requests because [they don't really make sense in an API](https://docs.djangoproject.com/en/3.1/ref/contrib/messages/).

valentijnscholten:
Yeah, I was just wondering if they were causing issues. I left them in place in other places and it worked out fine. We are looking at some ways to provide more (error/warning/details) feedback in API response and messages might be one option.
Also splitting it into two parts/methods make for some ugly '_interal' methods ;-)

iwalton3:
I did split the methods since I'm not sure what those decorators in the request do. But there could be a better pattern for shared code, but I didn't see it anywhere obvious. (It looks like most of the code between the API and web UI is duplicated.)

valentijnscholten:
Yeah, needs to be deduplicated. But we are first working on more test coverage so we minimize chance of bugs during refactoring.

valentijnscholten:
Hence my other question 😊


> Could you look at "adding something" to `test_rest_framework.py` to trigger this logic / endpoints?

iwalton3:
@valentijnscholten I have added unit tests, which exercise the three added endpoints and are passing in CI.

valentijnscholten (approved):

Maffooch (approved):

PR #3335, opened 2020-11-24
jira: fix add/edit engagement if no jira config used

valentijnscholten:
engagements without JIRA config would not get updated on form submit due to a similar bug as in #3295 

add/edit product is not suffering from this.


madchap (approved):

mtesauro (approved):
Approved

PR #3339, opened 2020-11-24
Correct filter for findings for non-staff users

StefanFl:
When a non-staff user opens the list of finding in one of the products, he sees not only the findings of this particular product but all findings for all products he is allowed to see. This is due to a bug, where the queryset for the findings doesn't get a new filter to limit the findings to his products but a new filter is created, which disregards the initial selection.

This bug fix uses the previously defined filter and uses it for the authorization.


valentijnscholten (approved):
good find, recent bug introduced in authorized users for product types.

mtesauro (approved):
Approved

PR #3343, opened 2020-11-24
securityContext related updates to Helm chart

namloc2001:
Also added in ability to generate an OpenShift SCC for nonroot if required.

OpenShift.enabled flag can be used at install time as follows: `helm install <release_name> --set OpenShift.enabled=true`


madchap (approved):
Pulled the PR locally and ran a basic helm deployment, works fine. The openshift part can be welcome for many users I assume, and the securityContext is a good explicit addition. Thanks for the PR.

namloc2001:
Changed to lowercase 'o' to respect camelCase.

madchap:
@namloc2001 sorry I forgot to ask, could you also bump the version to 1.4.1 in Chart.yml? Thanks.

namloc2001:
@madchap Chart.yaml version has now been updated.

alles-klar:
Happy to see someone else running DefectDojo on Openshift!

Adding the security context for each container is great. If it is required in your setup you may also want to add them for redis/rabbitmq and the database.

A problem with the security context runAsUser is, that the current implementation only work with the uid 1001. So if we make this value configurable, we should ensure that it also work with other values.

I am not very happy with the scc in this chart because it is more infrastructure related that application related. In my setup for example I am not allowed to create a scc because this is set by the cluster administrator. I would suggest to remove this part. 

Are there any advantages when using a different (new) service account for the test cases or can we just reuse the existing one from defectdojo?

namloc2001:
@alles-klar:

1. Yes the runAsUser must be 1001 at the moment because the containers specify this user and therefore associated file permissions are aligned to work with this user. I added the comments to the values.yaml to highlight the reasoning for `1001` being the specified value. If you wish for the containers to work with the restricted SCC in OpenShift, then the containers will need to be configured with chgrp and chmod to root gid and g=u permissions to enable them to function with the arbitrarily assigned UID (see "Support Arbitrary User IDs" section here: https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html).

2. I included the SCC because the containers have not been configured in such as a way as to be compatible with OpenShift deploying them using the restricted SCC. I think one or the other should be included. If you'd like the SCC inclusion to be dropped, then as per ("Support Arbitrary User IDs" section here: https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html), the containers should be configured with the appropriate chgrp & chmod.

The basic guidance is:
```
RUN chgrp -R 0 /some/directory && \
    chmod -R g+rwX /some/directory
```
And looking at the [Dockerfile.django](https://github.com/DefectDojo/django-DefectDojo/blob/master/Dockerfile.django) I can see:
```
RUN \
  mkdir -p dojo/migrations && \
  chmod g=u dojo/migrations && \
```
Which I believe is being performed as root anyway, so this effectively handles the chgrp & chmod requirements. And:
```
RUN \
    adduser --system --no-create-home --disabled-password --gecos '' \
        --uid ${uid} ${appuser} && \
    chown -R ${appuser} /app && \
    chmod 0700 /app && \
    chmod 0750 -R /app/* && \
    mkdir /var/run/${appuser} && \
    chown ${appuser} /var/run/${appuser}
```
So this requires a little rework to give the appropriate access to "arbitrary user" who is not appuser/1001, but is assigned the root group.

The container also shouldn't specify `USER <name>`, it should always be `USER <UID>` so that K8s and OpenShift can validate against `mustRunAsNonRoot` rules etc.

Looking at [Dockerfile.nginx](https://github.com/DefectDojo/django-DefectDojo/blob/master/Dockerfile.nginx) I can see:
```
RUN \
  apk add --no-cache openssl && \
  chmod -R g=u /var/cache/nginx && \
  mkdir /var/run/defectdojo && \
  chmod -R g=u /var/run/defectdojo && \
  mkdir -p /etc/nginx/ssl && \
  chmod -R g=u /etc/nginx && \
  true
```
I think this might be ok as the `RUN` command is performed by root, would probably just want to check the `/var/cache/nginx` as it's not created in this step.

Thoughts?

3. The test case serviceAccount does not have to be separate. The existing can be reused if desired. We normally just split these out to facilitate more granular controls if required.

alles-klar:
@namloc2001 thanks for the detailed statement. I would prefer to configure the container with the appropriate chgrp & chmod. Sound like a much more universal solution. Can you implement it? Otherwise I would suggest to create another PR for this.

3: I agree with you, but until there is not difference between the service accounts I would like to keep it simple. Otherwise we should be consistent and create service accounts for the other deployments too.

namloc2001:
I'll take a look

namloc2001:
@alles-klar I've made the changes. Are you confident the automated tests will highlight if my changes introduce any bugs in the images due to permission changes?

helm/defectdojo/values.yaml
```diff
@@ -23,6 +23,23 @@ repositoryPrefix: defectdojo
...
 
+securityContext:
+  djangoSecurityContext:
+    # django dockerfile sets USER=1001
+    runAsUser: 1001
+  nginxSecurityContext:
+    # nginx dockerfile sets USER=1001
+    runAsUser: 1001
+
+test:
```
alles-klar, line 34: Can we rename it to:
``` 
tests:
    unit-tests:
```
So we are able to maybe add an integration test or something similar.

alles-klar:
@namloc2001 Thank you!

Tests should find this issues but you never know. I did a quick test on my own instance and it seems still working. 

Two litte things, before we can merge:
- rename test in values, as described above
- add the security context to the unit-test template

helm/defectdojo/values.yaml
namloc2001: Done, please take a look

alles-klar (approved):
Many thank for your work.

PR #3347, opened 2020-11-25
Fix helm rabbitMQ not grabbing secret after move to bitnami

madchap:
**Helm deployment option broken until this is merged.**

As we moved to Bitnami repo, the rabbitMQ chart got bumped 2 major version.

A key changed from `rabbitmq` to `auth`.

_Support: There could also be some "support" strain, where a statefulset cannot be upgraded due to k8s immutable specification around volume claims, so deleting the statefulset would be required for the upgrade_

### DIFF
--- a/helm/defectdojo/templates/secret-rabbitmq.yaml
+++ b/helm/defectdojo/templates/secret-rabbitmq.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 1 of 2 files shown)


alles-klar (approved):

dsever:
k8s actions works well with proposed  patch: 

https://github.com/dsever/django-DefectDojo/runs/1459044395?check_suite_focus=true

ptrovatelli (approved):

PR #3350, opened 2020-11-25
Helm posgreSQL, use existing secret even if postgres not enabled

madchap:
If you use an external postgreSQL, you'd still have a secret yet have it disabled within the helm chart.

### DIFF
--- a/helm/defectdojo/templates/secret-postgresql.yaml
+++ b/helm/defectdojo/templates/secret-postgresql.yaml
@@ -14,7 +14,6 @@ metadata:
     helm.sh/hook-delete-policy: "before-hook-creation"
 type: Opaque
 data:
-{{- if .Values.postgresql.enabled }}
 {{- if .Values.postgresql.postgresqlPassword }}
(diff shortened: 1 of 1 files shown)


madchap:
fyi @yilmi

helm/defectdojo/templates/secret-postgresql.yaml
```diff
@@ -23,11 +23,17 @@ data:
   postgresql-postgres-password: {{ $postgresRandomPassword }}
   {{ .Values.postgresql.secretKey }}: {{ $postgresRandomPassword }}
 {{- end }}
+{{- else }}
```
alles-klar, line 26: Can't we just remove the if query, code seems to be identical in both cases?

madchap: That's what I was thinking too.

madchap:
Some notes for the next release put in https://github.com/DefectDojo/Documentation/pull/156

ptrovatelli (approved):

PR #3359, opened 2020-11-26
Aqua parser: correct missing values

Wadeck:
When a score is 4 or 7, it becomes "Critical" as those values were not caught.

@valentijnscholten @madchap


madchap (approved):

valentijnscholten:
should this be `release/1.10.1` or `dev` instead of master?

madchap:
> should this be `release/1.10.1` or `dev` instead of master?

Changed to `dev`, I guess it's good enough? We could also plan it for the 1.10.1 if you want.

valentijnscholten (approved):

PR #3362, opened 2020-11-26
FIX APIv2 : Add exception when parser fail to avoid error 500 without error info

RomainJufer:
The method `save` in the serializers for import scan and re-import scan can fail if the file parsed is not of the type given by scan type. As a result, the client gets an error 500 without any error message. Therefore, this PR aim at solving that issue by raising an exception which is catched in the viewset and transformed into a ParseError that yields an HTTP error 400 with an error message.

Currently, I wrote a generic message since there is not error handling in the `import_parser_factory` that would yield a more specific error message. Let me know if you think that something else would be more appropriate.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1188,6 +1188,8 @@ class ReImportScanSerializer(TaggitSerializer, serializers.Serializer):
                                            data['scan_type'],)
         except ValueError:
             raise Exception("Parser ValueError")
+        except:
+            raise Exception('Error while parsing the report, did you specify the correct scan type ?')
 
         try:
             items = parser.items
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1002,6 +1002,8 @@ class ImportScanSerializer(TaggitSerializer, serializers.Serializer):
                                            data['scan_type'],)
         except ValueError:
             raise Exception('FileParser ValueError')
+        except:
+            raise Exception('Error while parsing the report, did you specify the correct scan type ?')
 
         new_findings = []
         skipped_hashcodes = []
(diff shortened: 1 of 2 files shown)


madchap (approved):

madchap:
@RomainJufer we're planning to release 1.10.1 soonish, would you mind rebasing off of `release/1.10.1` branch please? That way, we can include the fix there too. Thanks.

Maffooch (approved):
It's a little larger of a rebase than I was hoping. Will merge after rebase is complete.

dojo/api_v2/serializers.py
```diff
@@ -1002,6 +1002,8 @@ def save(self, push_to_jira=False):
                                            data['scan_type'],)
         except ValueError:
             raise Exception('FileParser ValueError')
+        except:
+            raise Exception('Error while parsing the report, did you specify the correct scan type ?')
```
valentijnscholten, line 1006: I think this hides the original exception? should it be `raise Exception('blabla', e)` or something?

valentijnscholten (changes requested):

madchap:
> It's a little larger of a rebase than I was hoping. Will merge after rebase is complete.

I am sure @RomainJufer will come up with some `git rebase --onto` magic and get this by like a breeze ;-)

dojo/api_v2/serializers.py
valentijnscholten: Hmm just read that it was on purpose to catch it in the viewset. This was on our list for a long time to make error handling better, so thanks for doing this.

valentijnscholten (approved):

valentijnscholten:
Or just leave it at dev. The hotfix shouldn't be too large and yours6 or more of an enhancement.

PR #3379, opened 2020-11-30
pip: use old legacy resolver

valentijnscholten:
switch to old legacy resolver due to pip now using a new more strict resolver that we cannot use yet


madchap (approved):

mtesauro (approved):
Approved

PR #3386, opened 2020-12-01
Include local_settings in celery containers if present

madchap:
If you want to add a new CELERY_BEAT_SCHEDULE or override deduplication algorythms by way of HASHCODE_FIELDS_PER_SCANNER and friends, the configuration needs to also be propagated to the celery containers and not only uWSGI.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -52,6 +52,10 @@ services:
       DD_CELERY_BROKER_PASSWORD: ${DD_CELERY_BROKER_USER:-guest}
       DD_SECRET_KEY: ${DD_SECRET_KEY:-hhZCp@D28z!n@NED*yB!ROMt+WzsY*iq}
       DD_CREDENTIAL_AES_256_KEY: ${DD_CREDENTIAL_AES_256_KEY:-&91a*agLqesc*0DJ+2*bAbsUZfR*4nLw}
+    volumes:
+        - type: bind
+          source: ./docker/extra_settings
+          target: /app/docker/extra_settings
   celeryworker:
     image: defectdojo/defectdojo-django:latest
     depends_on:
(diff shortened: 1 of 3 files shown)


valentijnscholten:
Until now we have not configured these bind mounts in the docker-compose release setup. As these will override any local_settings.py file that someone might have created in dev mode inside dojo/settings/ folder.
So updating the entrypoint scripts make sense, but the bind mounts may conflict with what the intention was for dev mode.

I am not sure what the best option is, but I am thinking we should completely seperate "dev" mode from "release" mode.
Have 1 monster docker-compose file with everything needed for dev, local source mounted into the container, etc.
And a seperate docker-compose setup for production hosting.

madchap:
> Until now we have not configured these bind mounts in the docker-compose release setup. As these will override any local_settings.py file that someone might have created in dev mode inside dojo/settings/ folder.

Wouldn't https://github.com/DefectDojo/django-DefectDojo/blob/master/docker-compose.yml#L40-L42 make it also available in release mode?

Seems it does for me.

image

valentijnscholten:
wow, forgot about that I added that. I guess it all works as expected until you add a local_settings.py in the extra_settings folder. that would overwrite the one in dojo/settings. maybe just merge it and see what happens :-D

madchap:
My use-case is that I guess test in `dev` or `ptvsd`, and then I know that putting that same file over to extra_settings would give me the same thing in release. I guess it adds up in my head :man_shrugging: but I may be missing something -- would not surprise me, especially these days.

madchap:
But we should at least be consistent! So I guess we should merge this anyways.

valentijnscholten (approved):

Maffooch (approved):

Issue #3252, opened 2020-11-12
Engagement report should list tests titles and not just types


madchap:
The engagement has 4 distinct tests. However, the report loops over the first one over and over again.

![image](https://user-images.githubusercontent.com/5468769/98948037-35513000-24f6-11eb-9570-1ae2a665c5e6.png)

PR #3402, opened 2020-12-03
Fix reports: print test names instead of test types - #3252

yilmi:
Hi,

Could you please check this fix for Issue #3252 ?

Thanks,
Yassine


madchap (changes requested):
Thanks for the PR.

I pulled it locally, and I think there just needs to be a minor adjustment in case there is no test name specified by the user (it's an optional field).

The 3rd test here had no "name". But now, we don't even see what type it was. We should still see the type of the test.

image

valentijnscholten:
Yeah the title field on tests was added later, I think I have seen logic in other places where it is checked.

valentijnscholten:
It's in the "to string" of the Test model:
image

So it might work to just put {{ test }} or {{ t }} in this case.

yilmi:
Thanks for the comments, I was able to change and test most of the reports with this change except those using celery for reports generation.
I also fixed an issue in the `dojo/templates/dojo/product_type_pdf_report.html` template, the if statements weren't closed properly, closed them where the error was reported and tested it.

valentijnscholten:
Thanks. The misaligned if statements were already fixed in #3448 so now you have a small conflict to resolve :-)

valentijnscholten (approved):

madchap (approved):
lgtm :)

madchap:
I cannot test, because it seems the latest tagulous work is breaking report.

image

madchap (approved):
The code must work, but opened a bug which breaks the report anyways. But nothing to do with this PR. Thanks @yilmi for the contribution :)

madchap:
Should be fixed by https://github.com/DefectDojo/django-DefectDojo/pull/3486

madchap:
Or not...

```
Traceback:

File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py" in inner
  34.             response = get_response(request)

File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  115.                 response = self.process_exception_by_middleware(e, request)

File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  113.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/usr/local/lib/python3.6/site-packages/django/contrib/auth/decorators.py" in _wrapped_view
  21.                 return view_func(request, *args, **kwargs)

File "./dojo/reports/views.py" in engagement_report
  368.     return generate_report(request, engagement)

File "./dojo/reports/views.py" in generate_report
  688.         ids = set(finding.id for finding in findings.qs)

File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py" in __iter__
  274.         self._fetch_all()

File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py" in _fetch_all
  1244.             self._prefetch_related_objects()

File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py" in _prefetch_related_objects
  771.         prefetch_related_objects(self._result_cache, *self._prefetch_related_lookups)

File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py" in prefetch_related_objects
  1614.                                      (through_attr, first_obj.__class__.__name__, lookup.prefetch_through))

Exception Type: AttributeError at /engagement/2/report
Exception Value: Cannot find 'tagsnotes' on Finding object, 'tagsnotes' is an invalid parameter to prefetch_related()
```

valentijnscholten:
hmm don't understand why the integration test doesn't fail and and why it is working for me. but looks like a comma is missing and python then thinks well then it must be one big string hahaha let's concatenate it into something that doesn't work.

valentijnscholten:
@madchap what steps do you take to reach this error? The test suite is supposed to generate an engagement report, but apparantly not in a way it reaches this code.

madchap (approved):
Tested with latest fixes, all good! Thanks.

PR #3408, opened 2020-12-04
[APIv2] Update put semantic and doc for endpoint /finding/{id}/metadata

RomainJufer:
This PR is an improvement of the (sub)endpoint recently added, `/finding/{id}/metadata`. In the current implementation, the semantic of PUT is only an update semantic, meaning that if the given metadata does not exist, it will not create it. However, since we are in the particular case where a metadata can be uniquely identified by a pair `(finding_id, metadata_name)` we think it makes sense to change the semantic to that of update or create. Note that this does not contradict the original semantic intended for the PUT operation in the context of REST API since we are not identifying the metadata with an ID.

The main advantage of this improvement is to reduce the number of queries from two to one in the case where we want to do an update or create. We also kept the POST operation to still support a create only that could fail instead of updating the metadata if it already exists.

We also took advantage of this improvement to enhance the documentation of that endpoint, more specifically the responses that the endpoint could return.

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -4,6 +4,7 @@ from django.utils import timezone
 from django.core.exceptions import ValidationError
 from rest_framework import viewsets, mixins, status
 from rest_framework.response import Response
+from django.db import IntegrityError
 from rest_framework.permissions import DjangoModelPermissions
 from rest_framework.decorators import action
 from rest_framework.parsers import MultiPartParser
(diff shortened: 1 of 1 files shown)


madchap (approved):

valentijnscholten:
>...because the name of the metadata already exists

Is this the only reason these exceptions can occur, or should we add a 'probably' in the message to cater for other (corner) cases?

RomainJufer:
You're right, I'm not entirely sure that the unique check will only fail if the name already exists so I added a bit of uncertainty in the message as suggested :)

valentijnscholten (approved):

PR #3416, opened 2020-12-06
release workflow: simplify matrix

valentijnscholten:
the current workflow builds all images twice. once to push version `x.y.z` and once to push `latest`
this workflow simplifies this, which also results in the sha's for both tags being the same (which is good).


madchap (approved):
Well done.

valentijnscholten:
time to merge :-)

PR #3423, opened 2020-12-07
[feat/login-form] Allowing login forms to be hidden

natebwangsut:
Aesthetic enhancement, when integrating with OAuth services, usually the application would redirect to external page for authentication purpose and username/password input form created a login process confusion.

- Allowing input form to be hidden (showing by default)

### DIFF
--- a/dojo/context_processors.py
+++ b/dojo/context_processors.py
@@ -4,7 +4,8 @@ from django.conf import settings
 
 def globalize_oauth_vars(request):
     # return the value you want as a dictionnary. you may add multiple values in there.
-    return {'CLASSIC_AUTH_ENABLED': settings.CLASSIC_AUTH_ENABLED,
+    return {'SHOW_LOGIN_FORM': settings.SHOW_LOGIN_FORM,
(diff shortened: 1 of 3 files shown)


valentijnscholten:
First thought: should there be only one setting either this one or the existing CLASSIC_AUTH_ENABLED? The latter hides the login button which should be done also if the login form is hidden?

natebwangsut:
I had the same thought originally but say if the form can still be useful (e.g. external authentication service where it can be use with API call while a username need to be submitted?) it can be rare case but I prefer the flexibility of configuration.

For the login button with `CLASSIC_AUTH_ENABLED`, do you reckon that it should be hidden as well if the form is hidden ?

valentijnscholten:
I was thinking the same about passing the username to IdP. Currently not happening, but maybe in the future. 
Hiding login button without form would be best.

natebwangsut:
I've added a commit which would hide a login button if the form is disabled

valentijnscholten (approved):

madchap (approved):
Cool, thanks for the PR :-)

PR #3427, opened 2020-12-07
Allow marking Qualys WAS security weaknesses as findings.

iwalton3:
This is a compromise between #3348 and #3150. Some users may want to include the "Security Weakness" findings from Qualys as findings, while others may want to make them informational. (Note that the minimum finding level is left at Low. That was not changed from #3348, but may be a subject of contention between users.)

This PR improves the logic for determining this and adds a `QUALYS_WEAKNESS_IS_VULN` boolean toggle to settings to allow you to opt-in to having "Security Weakness" have a severity.


valentijnscholten:
Can you also add a `DD_QUALYS_WEAKNESS_IS_VULN`? Or maybe even make the name more explicit to indicate it affects the web parser, but not the infra parser?

iwalton3:
I have fixed the naming and added an environment variable.

devGregA:
@cody-m-tibco please review.

valentijnscholten:
@iwalton3 could you resolve the conflicts please?

valentijnscholten (approved):

madchap (approved):
lgtm

valentijnscholten:
thanks @iwalton3 
@madchap @damiencarol @Maffooch still happy to merge this?

damiencarol (changes requested):
We should have a unit test that shows the parser modification.
As it modifies how the parser works, could you add a unit test that shows the weakness as finding when we activate this option?

iwalton3:
> We should have a unit test that shows the parser modification.
> As it modifies how the parser works, could you add a unit test that shows the weakness as finding when we activate this option?

I added a unit test for this.

damiencarol (approved):

PR #3428, opened 2020-12-08
Nikto parser for scan of multiple hosts

StefanFl:
Nikto can produce results for multiple hosts in one scan, eg. by reading nmap output to supply a list of hosts and ports to scan. The current Nikto parser reads only the result of the first host. With this pull request it will iterate over the results for all scanned hosts. This will only work for the new file type (nxvmlversion="1.2").


valentijnscholten (approved):

madchap (approved):
:+1:

PR #3433, opened 2020-12-09
add note for not working AWS redis

wurstbrot:
as discussed in slack, aws redis is not working (2020-12-08). 
Let us prevent other from being frustrated.

In case you know a better place to high-lite this, please let me know.

### DIFF
--- a/KUBERNETES.md
+++ b/KUBERNETES.md
@@ -101,6 +101,8 @@ status of the containers can be viewed by starting up ```minikube dashboard```.
 Note: If the containers are not cached locally the services will start once the
 containers have been pulled locally.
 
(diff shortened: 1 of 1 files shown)


valentijnscholten:
For others it's working fine out of the box: https://owasp.slack.com/archives/C2P5BA8MN/p1607496169370700?thread_ts=1607441990.368500&cid=C2P5BA8MN

Besides that, we cannot maintain a list of "things defect dojo doesn't work with".

We will look at making / keeping redis working out of the box. If people use any specific serverless solution, it's out of our control.

wurstbrot:
thanks

PR #3435, opened 2020-12-09
Allow Info findings to be pushed to JIRA without SLA

Maffooch:
Findings with an Info severity do not get an SLA. When creating new issues to push to JIRA, a findings SLA is used to set the due date on the JIRA issue. Since Info findings don't have an SLA, trying to add one results in an exception and the issue is not pushed. This added check just ignores that field if it is an Info finding.

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -453,7 +453,7 @@ def add_jira_issue(find):
 
             if System_Settings.objects.get().enable_finding_sla:
 
-                if 'duedate' in meta['projects'][0]['issuetypes'][0]['fields']:
+                if 'duedate' in meta['projects'][0]['issuetypes'][0]['fields'] and Finding.get_number_severity(find.severity):
                     # jira wants YYYY-MM-DD
                     duedate = find.sla_deadline().strftime('%Y-%m-%d')
                     fields['duedate'] = duedate
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -453,10 +453,11 @@ def add_jira_issue(find):
 
             if System_Settings.objects.get().enable_finding_sla:
 
-                if 'duedate' in meta['projects'][0]['issuetypes'][0]['fields'] and Finding.get_number_severity(find.severity):
+                if 'duedate' in meta['projects'][0]['issuetypes'][0]['fields']:
                     # jira wants YYYY-MM-DD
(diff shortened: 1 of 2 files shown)


valentijnscholten:
I would expect the find.sla_deadline() to be None and we should check on that to be None and skip duedate setting in JIRA?
We should try to minimize internal logic in the JIRA code.

mtesauro:
Not exactly sure how the internals of the Jira integration work exactly but I can tell you that I was getting a Python exception on info findings being pushed to Jira.  I mentioned this on a chat with Maffooch yesterday and I guess he listened :-)

mtesauro (approved):
Approved

valentijnscholten:
The fact that it works, doesn't mean it's solved in the best place. To me it feels weird that the JIRA integration code needs to know about that Info findings don't have a sla deadline. Second weird is that the jira code and anyone who reads it needs to know that the get_number_severity method returns False or None (or zero?) for Info findings. Or does it return that for all findings without an sla deadline? 
Apart from that if other integrations would also want to use a duedate / deadline field, thy would need to implement the same logic and we would have the same logic in multiple places (github integration, sla notifications, future bugtrackers, ...)

So what I am saying is that finding.sla_deadline() should return None if there is no deadline. The JIRA then only needs to do a simple ìf` wihtout knowing anything about the relationship between SLA's, severities, numerical severities, etc.

valentijnscholten (changes requested):
The fact that it works, doesn't mean it's solved in the best place. To me it feels weird that the JIRA integration code needs to know about that Info findings don't have a sla deadline. Second weird is that the jira code and anyone who reads it needs to know that the get_number_severity method returns False or None (or zero?) for Info findings. Or does it return that for all findings without an sla deadline? 
Apart from that if other integrations would also want to use a duedate / deadline field, thy would need to implement the same logic and we would have the same logic in multiple places (github integration, sla notifications, future bugtrackers, ...)

So what I am saying is that finding.sla_deadline() should return None if there is no deadline. The JIRA then only needs to do a simple ìf` wihtout knowing anything about the relationship between SLA's, severities, numerical severities, etc.

devGregA (approved):

devGregA:
I believe this complies with Valtenjin's suggestion. Can adjust if that is not the case.

valentijnscholten:
Better yes, thanks!

Issue #3207, opened 2020-11-08
product metrics very slow & > 500 queries


valentijnscholten:
The product metrics tab is very slow, sometimes performing over 500 queries.
Seems like `test_type` is not prefetched causing lots of extra queries.

madchap:
I can confirm, even getting 504 gateway timeout on my big database.

PR #3438, opened 2020-12-09
WIP: Refactor for performance improvements on metrics page

danielnaab:
This branch is intended to address #3207. It refactors database queries on the metrics page into a separate module, which has test coverage to assert on the number of queries the ORM triggers. In addition to aiding in optimizing the query count, the intent is also to improve test coverage over the reporting metrics.

The move of functions from `views.py` into `queries.py` is in a separate commit from actual code changes. Hopefully reading this branch commit-by-commit, rather than looking at the entire diff at once, will make it clear what was actually changed.

To assist in the refactor, I used pylint. In the process, I fixed a couple minor issues like a reference to an undefined object and removal of unused variables.

This PR is a work-in-progress - at the moment, it doesn't decrease the query count, but just makes assertions to use to aid in confidently making follow-up changes.


danielnaab:
Closing due to #3549, and moving additional changes into #3730.

PR #3439, opened 2020-12-09
logging: add DD_LOG_LEVEL setting

valentijnscholten:
Whenever people report bugs / issues, it's good to have DEBUG logs available. This PR allows to easily turn on DEBUG logging without having to rebuild containers or modifying settings files.

- adjust log level based on DEBUG value. DEBUG=True -> log level DEBUG, otherwise log level INFO
- add DD_LOG_LEVEL setting to allow finetuning / overriding
- update docs

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -12,6 +12,7 @@ env = environ.Env(
     DD_SITE_URL=(str, 'http://localhost:8080'),
     DD_DEBUG=(bool, False),
     DD_TEMPLATE_DEBUG=(bool, False),
+    DD_LOG_LEVEL=(str, ''),
     DD_DJANGO_METRICS_ENABLED=(bool, False),
     DD_LOGIN_REDIRECT_URL=(str, '/'),
     DD_DJANGO_ADMIN_ENABLED=(bool, False),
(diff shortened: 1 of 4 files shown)


dojo/settings/settings.dist.py
```diff
@@ -841,6 +842,11 @@ def generate_url(scheme, double_slashes, user, password, host, port, path):
 # See http://docs.djangoproject.com/en/dev/topics/logging for
 # more details on how to customize your logging configuration.
 LOGGING_HANDLER = env('DD_LOGGING_HANDLER')
+
+LOG_LEVEL = env('DD_LOG_LEVEL')
+if not LOG_LEVEL:
+    LOG_LEVEL = 'DEBUG' if DEBUG else 'INFO'
```
madchap, line 848: Why not pass `INFO` as default value to `DD_LOG_LEVEL`?

valentijnscholten: Because I want it to be aligned with DD_DEBUG by default so we don't have to add DD_LOG_LEVEL everywhere

mtesauro (approved):
Approved.

Great idea - kinda surprised we didn't do this a while ago.

madchap (approved):

PR #3450, opened 2020-12-10
Allow to specify the `Environment` when importing data from the APIv2 and the UI

xens:
### What

This commit add the ability to specify the `Environment` (prod, dev, ...) when importing scan data from both the APIv2 and the UI.

The current behavior of DDJ is to put everything inside the `Development` environment (APIv2+UI) when importing data without being able to specify it. I modified this logic and it'll put everything inside the `Default` environment if not specified when importing data from the APIv2, and when uploading data from the UI the user will need to specify it manually.

### Notes

* I haven't added the ability to edit the `Environment` when using the `reimport_scan` endpoint, I didn't think that it was relevant and it's still possible to edit the environment afterwards by editing the associated `test`. Let me know what you think.
* I wasn't sure if it was best to specify the `environment_name` or the `environment_id` when uploading a scan from the APIv2, I went with the first as I thought that it was more convenient (as we do for specifying the scanner name).

### DIFF
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -557,9 +557,11 @@ def import_scan_results(request, eid=None, pid=None):
                 return HttpResponseRedirect(reverse('import_scan_results', args=(engagement,)))
 
             tt, t_created = Test_Type.objects.get_or_create(name=scan_type)
-            # will save in development environment
-            environment, env_created = Development_Environment.objects.get_or_create(
-                name="Development")
+
+            # Will save in the provided environment or in the `Development` one if absent
+            environment_id = request.POST.get('environment', 'Development')
+            environment = Development_Environment.objects.get(id=environment_id)
+
             t = Test(
                 engagement=engagement,
                 test_type=tt,
--- a/dojo/forms.py
+++ b/dojo/forms.py
@@ -433,6 +433,8 @@ class ImportScanForm(forms.Form):
     active = forms.BooleanField(help_text="Select if these findings are currently active.", required=False)
     verified = forms.BooleanField(help_text="Select if these findings have been verified.", required=False)
     scan_type = forms.ChoiceField(required=True, choices=SORTED_SCAN_TYPE_CHOICES)
+    environment = forms.ModelChoiceField(
+        queryset=Development_Environment.objects.all().order_by('name'))
     endpoints = forms.ModelMultipleChoiceField(Endpoint.objects, required=False, label='Systems / Endpoints',
                                                widget=MultipleSelectWithPopPlusMinus(attrs={'size': '5'}))
     tags = forms.CharField(widget=forms.SelectMultiple(choices=[]),
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -947,6 +947,7 @@ class ImportScanSerializer(TaggitSerializer, serializers.Serializer):
     tags = TagListSerializerField(required=False)
     close_old_findings = serializers.BooleanField(required=False, default=False)
     push_to_jira = serializers.BooleanField(default=False)
+    environment = serializers.CharField(required=False)
 
     def save(self, push_to_jira=False):
         data = self.validated_data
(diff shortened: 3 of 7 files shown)


dojo/engagement/views.py
```diff
@@ -558,8 +558,11 @@ def import_scan_results(request, eid=None, pid=None):
 
             tt, t_created = Test_Type.objects.get_or_create(name=scan_type)
             # will save in development environment
-            environment, env_created = Development_Environment.objects.get_or_create(
-                name="Development")
+
+            environment_id = request.POST.get('environment', 'Default')
+            environment = Development_Environment.objects.get(id=environment_id)
+            environment, created = Development_Environment.objects.get_or_create(id=environment_id)
```
valentijnscholten, line 564: the first one should be removed I guess?

dojo/forms.py
```diff
@@ -433,6 +433,8 @@ class ImportScanForm(forms.Form):
     active = forms.BooleanField(help_text="Select if these findings are currently active.", required=False)
     verified = forms.BooleanField(help_text="Select if these findings have been verified.", required=False)
     scan_type = forms.ChoiceField(required=True, choices=SORTED_SCAN_TYPE_CHOICES)
+    environment = forms.ModelChoiceField(
+        queryset=Development_Environment.objects.all().order_by('name'))
```
valentijnscholten, line 437: the import scan is broken in integration test, not sure, but could be caused by this?

valentijnscholten (changes requested):

dojo/engagement/views.py
xens: I did it intentionally to make the import fail if the `Environment` that was specified during the import does not exist, to be consistent with the API. I agree that it could be removed here as the selection is made from a pre-filled list in the UI, let me know what you think.

dojo/forms.py
xens: It is, I didn't know we had integration tests for the UI that's neat !

dojo/engagement/views.py
valentijnscholten: I don't understand. If the environment doesn't exist, the first line will raise an exception. If it does exist, the 2nd statement won't do anything. So why have two statements?

xens: You're right that doesn't make any sense thanks for pointing that out. What I wanted to do initially is to make sure that the `Default` environment gets created if it doesn't exists to avoid breaking existing deployments (where the new `Default` environment wouldn't be provisioned by fixtures). I've modified a bit the logic, let me know what you think !

valentijnscholten: I would do:
- UI: User selects from dropdown so we have to assume the environment exists under normal circumstances
- API: Environment can be freely supplied and not guaranteed to exist. I suggest to create the env if needed.

xens: * UI: Agree, make sense, I adapted the code
* API: I still think that we shouldn't automatically create the environment if it doesn't exist and let the user provision it separately. If we allow auto-create, users will end up having 2 or 3 time their environment with different cases or with typos `Development`, `development`, `dev`, and as there's no way to merge these environments it'll be nightmare-ish for companies with many teams to have something consistent over time.

valentijnscholten: so then inboth cases this will suffice:
`environment = Development_Environment.objects.get(name=environment_name)`

xens: Yes you're right but if we do so for the API then it'll break existing integrations as people will be now forced to specify the `environment` in their POST. If we want to simply the code without breaking the existing integrations we could do:

```sh
environment, created = Development_Environment.objects.get_or_create(name=environment_name)`
```

valentijnscholten: just leave it optional and leave the default at "Development" and we don't need the extra code. In the long term we better make the environment optional.

xens: Ok done !
I left the `Default` environment in the fixtures and in the unit-tests, this way we can maybe switch in a future release to this environment that make more sense than the `Development` one.

valentijnscholten (approved):

madchap (approved):
lgtm, merry xmas!

PR #3453, opened 2020-12-11
[FIX] Remove call to old method and add call to new method for jira integration

RomainJufer:
This PR is pretty straight forward, part of the code was still using the old method `add_comment_task` that has been replaced by `add_comment`. This resulted in errors when adding notes to findings. Therefore, we replace the old method by the new method which solve the problem.

Let me know if there is more to it as I'm not completely familiar with the new integration :).


valentijnscholten:
PR looks good. Not so good we missed this bug....

valentijnscholten (approved):

madchap (approved):

Issue #3445, opened 2020-12-10
Support SARIF (OASIS Static Analysis Results Interchange Format)


damiencarol:
**SARIF parser**
SARIF represents a leap forward in the usability of static analysis tools. Many organizations in the safety and security communities use several competing tools on their code. SARIF will allow them to combine and compare the results more easily to gain a sharper picture of the issues in their code that need to be addressed. Engineering teams will be able to easily access a broad range of potential defects and vulnerabilities in compliance with privacy and accessibility standards. SARIF will support the development of products whose code spans languages and operating systems.

[OASIS TC](https://www.oasis-open.org/committees/tc_home.php?wg_abbrev=sarif)
[Official Spec](https://github.com/oasis-tcs/sarif-spec)

**Sample File**
From spec github:
[https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Documents/ScratchSamples/179-311.sarif](https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Documents/ScratchSamples/179-311.sarif)
[https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Documents/ScratchSamples/314-338.sarif](https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Documents/ScratchSamples/314-338.sarif)

tohch4:
I'd be super excited to help with this @damiencarol. See DefectDojo/django-DefectDojo#2720 for previous discussion, but I thought there little or not interest. If there are two of us, I would like to move this forward!

valentijnscholten:
it sounds interesting, don't think anybody noticed there was discussion around this in another PR.

tohch4:
I am very interested in this space, I can look into on the weekend and revive previous draft efforts, unless @damiencarol wants to work it with us or for us, I will try and move ahead.

My main question: this is not tied to an import type anymore, and is more generalized. `semgrep` obviously supported, but a variety of static analysis tools have rapidly ramped up support after Github and Gitlab had announced integration in their APIs.

So how do we annotate the import correctly to allow SARIF to be used in abstract but concretely explain the tool that did the analysis and serialized it, @valentijnscholten?

damiencarol:
@tohch4 @valentijnscholten Thanks a lot for your feedback. I'm pushing a branch with a minimum viable code and unit tests.
Take a look at the big SARIF report. It's a scan of DefectDojo itself... 😆 

[...]

PR #3464, opened 2020-12-11
Importer: Add OASIS SARIF format #3445

damiencarol:
## OASIS SARIF format parser
SARIF represents a leap forward in the usability of static analysis tools. Many organizations in the safety and security communities use several competing tools on their code. SARIF will allow them to combine and compare the results more easily to gain a sharper picture of the issues in their code that need to be addressed. Engineering teams will be able to easily access a broad range of potential defects and vulnerabilities in compliance with privacy and accessibility standards. SARIF will support the development of products whose code spans languages and operating systems.

[OASIS TC](https://www.oasis-open.org/committees/tc_home.php?wg_abbrev=sarif)
[Official Spec](https://github.com/oasis-tcs/sarif-spec)

### Sample Files
4 example files come from the specification itself
1 report comes from [LGTM.com](LGTM.com) web site scans of master branch

#### A. SARIF reports could contain more than one "test".
But to keep this pull request simple, all findings are merged in one test

#### B. The current code supports a simple version for CWE but no CVE
SARIF support for CVE / CWE is limited/not well defined.
SARIF specification doesn't describe exactly in detail how it could work. So the current code assumes these requirements:
1) the CWE is one of the `tags` properties is the format `foo/bar/cwe-999` (case insensitive)
You can see an example report in a unit test 
2) We need more data/reports to improve this part (more reports from different tools


valentijnscholten:
Thanks. Is there any way we can store/show what scanner is being imported? Maybe set `test.title`?

damiencarol:
@valentijnscholten will work on that. I have few ideas. Basicaly SARIF carry a lot of details about the system that generated the report.
What do you wanna see in this title ?

damiencarol:
@valentijnscholten I added the Test title data.
Could you assign this PR to me and review it?
Also, the `Integration tests` fail is the last commit. But it's a chrome related issue, no link with my modifications.

valentijnscholten:
This format is interesting. I didn't know about it. Still thinking about how to best integrate it in defect dojo. Maybe we just merge this as a first iteration and gain some experience. In the end we may want to map the tool name / title to a specific test type in defect dojo so all tests of the same type can be grouped / deduplicated etc.

damiencarol:
@valentijnscholten interresting point of view. Honestly from an implementation perspective I managed it like just another format to support. so My patch is ok :D 
What you describe is more linked to the fact that more than one tool use the same format. So there should be a concept of "Tool" + "version" and some features around that.
Just to be clear, the dudplication works with this integration as I followed and implemented "file_path"+"line" + "title" + "cwe"

madchap:
I see no problem integrating a first iteration as it stands today. The parser could possibly also support a zip file upload, which would then be extracted and imported if several sarif files are to be processed?

It could be the first parser to eventually be a sort of meta-parser.

madchap (approved):

valentijnscholten (approved):

PR #2983, opened 2020-10-09
Fix APIv2 Endpoint Status bug

Maffooch:
Create endpoint status objects from APIv2. This feature was missing from 1.7.0 and was just discovered in 1.8.0 due to some updated displays surrounding endpoint status objects.

- [x] Give a meaningful name to your PR, as it may end up being used in the release notes.
- [x] Your code is flake8 compliant.
- [x] Your code is python 3.6 compliant (specific python >3.6 syntax is currently not accepted).
- [ ] If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR.
- [ ] Model changes must include the necessary migrations in the dojo/db_migrations folder.
- [ ] Add applicable tests to the unit tests.
- [x] Add the proper label to categorize your PR.

PR #3468, opened 2020-12-14
apiv2: fix endpoint status creation during scan import

alles-klar:
https://github.com/DefectDojo/django-DefectDojo/pull/2983 introduced bug (maybe a typo?) which caused a server error when importing a scan with an additional endpoint.

On behalf of DB Systel GmbH


valentijnscholten:
thanks. there's another occurrence a little further down in `ReImportScanSerializer`

alles-klar:
But there it is correct?

valentijnscholten:
> But there it is correct?

Ah, yes, I see.

valentijnscholten (approved):

madchap (approved):

PR #3469, opened 2020-12-14
fix(risk_acceptance): remove hard coded user_id

alles-klar:
Code will cause an error if user with id does not exist.

@valentijnscholten  I have not found code where owner_id is used and the Risk_Acceptance object does not contain such a attribute. Can we remove the hole line?

On behalf of DB Systel


valentijnscholten:
Don't know, previously it was called reporter and was hardcoded to id 1 in this place. But with the introduction of `get_current_user` it's now easier to get rid of those hardcoded ids and use a reference to the user executing the request.

valentijnscholten (approved):

valentijnscholten (changes requested):
not sure, but it could be that there is no user if the task is executed from celery. so maybe add a fallback to 1 or the id of the user `admin`?

alles-klar:
It should not be possible to create a risk acceptance without a user. Otherwise we have a big permission problem. 

But looking at the risk_acceptance it should be named owner instead of onwer_id.

valentijnscholten:
yeah, that's why it defaults to id 1. So my proposal is to use the id from the current user. But if there is no current user, use id 1 or the id of the 'admin' user.
If you run a management command, or a task is executed in celery or unit test, there is no current user and the code will fail because owner cannot be None/Null.

alles-klar:
In my case I have deleted the user with id 1 because I don't want to have the admin. Any other ideas because I want to avoid a buggy bugfix. Maybe throw an exception?

madchap:
> In my case I have deleted the user with id 1 because I don't want to have the admin. Any other ideas because I want to avoid a buggy bugfix. Maybe throw an exception?

You could eventually have just ripped out admin from all its rights except active maybe, with no django perms either? Did you actually delete it?

valentijnscholten (approved):

valentijnscholten:
approved so it can go in the release. will make it even more bullet proof in the upcoming risk acceptance enhancement PR.

Issue #3455, opened 2020-12-11
DSoP parser no longer working: Excel xlsx file; not supported


valentijnscholten:
As of today the unit tests seem to be failing with the below error. This means also the DSoP parser probably no longer works when somebody builds the django container today or later.

```
uwsgi_1         | ======================================================================
uwsgi_1         | ERROR: test_zero_findings (dojo.unittests.test_dsop_parser.TestDsopParser)
uwsgi_1         | ----------------------------------------------------------------------
uwsgi_1         | Traceback (most recent call last):
uwsgi_1         |   File "/app/dojo/unittests/test_dsop_parser.py", line 11, in test_zero_findings
uwsgi_1         |     parser = DsopParser(file, Test())
uwsgi_1         |   File "/app/dojo/tools/dsop/parser.py", line 14, in __init__
uwsgi_1         |     f = pd.ExcelFile(file)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/pandas/io/excel/_base.py", line 867, in __init__
uwsgi_1         |     self._reader = self._engines[engine](self._io)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/pandas/io/excel/_xlrd.py", line 22, in __init__
uwsgi_1         |     super().__init__(filepath_or_buffer)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/pandas/io/excel/_base.py", line 351, in __init__
uwsgi_1         |     self.book = self.load_workbook(filepath_or_buffer)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/pandas/io/excel/_xlrd.py", line 35, in load_workbook
uwsgi_1         |     return open_workbook(file_contents=data)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/xlrd/__init__.py", line 170, in open_workbook
uwsgi_1         |     raise XLRDError(FILE_FORMAT_DESCRIPTIONS[file_format]+'; not supported')
uwsgi_1         | xlrd.biffh.XLRDError: Excel xlsx file; not supported
```

valentijnscholten:
caused by `xlrd` 2.0.0 being released today and not pinned in `requirements.txt`

valentijnscholten:
to fix this, apply the change from https://github.com/DefectDojo/django-DefectDojo/pull/3457 to your local install and rebuild.
or wait for that PR to be release into 1.11.0 or 1.10.5

PR #3473, opened 2020-12-15
Add redis transit encryption

KarstenSiemer:
enables the use of a tls secured connection to a redis cluster, elasticache for example


KarstenSiemer:
I have no clue why the unit test failed, can someone help with it?

valentijnscholten:
> I have no clue why the unit test failed, can someone help with it?

It's because of #3455. Can you rebase against dev, or merge dev into your branch? It's fixed in dev.

KarstenSiemer:
@valentijnscholten ah great! Looks good now 👍

valentijnscholten (dismissed):
I believe the term typically used for this is `TransportEncryption`. Could you rename your variables to reflect that?

madchap (approved):
Looks ok to me. I am fine with the `transitEncryption` term as it's what I see when quickly googling around. Not sure if you had something else in mind @valentijnscholten ?

valentijnscholten:
either 'transport encryption' or 'in transit encryption'

madchap (changes requested):
Thinking a bit more into it, the change in the helm template helper probably warrant a minor bump in the Chart.yaml as well (e.g. to 1.4.3) to notify of the change.

KarstenSiemer:
I also fought with myself when naming it transitEncryption. But this is how it is referred to everywhere in aws.
I am fine however with both, since I think everybody will understand what it is about.
We could also simply name it "tls". Whatever is floats your boat

madchap (approved):

valentijnscholten (approved):
Didn't realize transit it's the term used by aws. And didn't turn up on my google bubble.

KarstenSiemer:
This PR was sponsored by SDA SE Open Industry Solutions

PR #3474, opened 2020-12-15
Typo fix in models.py

SPoint42:
Typo fix for Crital => Critical


valentijnscholten:
any change in model definitions needs a new migration file.

SPoint42:
> any change in model definitions needs a new migration file.

Even for a minor Typo fix ?

valentijnscholten:
yes, that's how Django works.

madchap:
It would also have to be rebased on dev not master.

valentijnscholten:
@SPoint42 in this case it might work if you also update the 0001_initia.py migration.

valentijnscholten:
@SPoint42 I think in this case it might actually be sufficient if you change `0001_initial.py` to include the same typo fix.

madchap:
Hi @SPoint42 

> @SPoint42 I think in this case it might actually be sufficient if you change `0001_initial.py` to include the same typo fix.

Do you want to do that? Cheers.

SPoint42:
close by https://github.com/DefectDojo/django-DefectDojo/pull/4024

PR #3477, opened 2020-12-15
Fix popup message on SLA displays

Maffooch:
Display tags did not correctly create the popup required to display the status of a findings SLA. Also cleaned up the logic when determining a findings SLA by removing the ugly "is Info" check.


devGregA (approved):

mtesauro (approved):
Approved

PR #3480, opened 2020-12-16
Fix exception during excess duplicate deletion tasks

### LINKED ISSUE #3368: dedupe delete job: ValueError('Cannot use None as a query value',)
Seen this in some logs submitted in other issues, let's investigate.
```
Received task: dojo.tasks.async_dupe_delete[b9ddfdba-d0c1-452e-8b32-24dc4acde1eb]
delete excess duplicates
delete excess duplicates
Task dojo.tasks.async_dupe_delete[b9ddfdba-d0c1-452e-8b32-24dc4acde1eb] raised unexpected: ValueError('Cannot use None as a query value',)
  File "/usr/local/lib/python3.6/site-packages/celery/app/trace.py", line 412, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/celery/app/trace.py", line 704, in __protected_call__
    return self.run(*args, **kwargs)
  File "/app/dojo/tasks.py", line 333, in async_dupe_delete
```

valentijnscholten:
fixes #3368 , exception during excess duplicate removal task on max_dupes being None/Null.

it also makes the job more efficient and provides an upper bound on the number of findings deleted per run.

On a 800K instance the first query to find excess duplicates takes ~6s. That query runs every minute, even if you do not have any excess duplicates. So basically 10% of the time the database is busy with this query only. If there excess duplicates found, it will do more queries. If you turn on the dedupe task on an existing instance, it could initially try to deletes 1000s of findings. This can result in overlapping rung of the dedupe tasks, each performing the same heavy queries. That's why we now have a (configurable) maximum of 200 findings per run that will be deleted. The job runs every minute so even with a backlog of 10000 excess duplicates to be deleted it will finish in a couple of minutes.


madchap (approved):
lgtm.

devGregA (approved):

PR #3489, opened 2020-12-17
Fix(helm-unittests): add secret key and credential key 

alles-klar:
* fix(helm-tests): add secret key and credential key because they are not allowed to be empty
* create the name in the same manner than the other templates

On behalf of DB Systel GmbH


valentijnscholten:
@alles-klar could you resolve the conflict?

madchap (approved):

PR #3500, opened 2020-12-19
Test suite and scripts cleanup

valentijnscholten:
- remove superfluous tests that are no longer valid or already covered elsewhere
- move (disabled) apply template test to unittests for faster execution when re-enables it
- re-enable commented out tests whenever possible
- reduce number of logins per test to save time
- rename files (remove `unit` as these are not unit tests)
- fix timing is on delete engagement test
- remove everything related to travis
- remove unused files in setup and test and docker folders


valentijnscholten:
@aaronweaver I cleaned up some stuff around test scripts, travis, etc. Afterwards I realized there were 2 or 3 files related to building  docker images for `appsecpipeline`. My impression is these scripts were used before dojo had the "official" docker images we have now. But I might be wrong. What do you think?

madchap (approved):
I am guessing the integration tests speak for themselves if they don't break :) Thanks for doing this.

damiencarol:
It seems a lot of dead code was removed. well done @valentijnscholten 👍

valentijnscholten:
So we have a small amount of extra tests, but the duration of the test suite has dropped by ~4minutes / 16% because we only login once per test case instead of for every test method.

PR #3508, opened 2020-12-20
[fix/helm-rabbitMQ]: Fix incorrect YAML key for RabbitMQ chart

natebwangsut:
RabbitMQ Chart
https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/values.yaml#L308


madchap (approved):
Thanks for catching it.

alles-klar (approved):

PR #3516, opened 2020-12-21
[APIv2] Prefetch Mixins + Composable Swagger Schema

RomainJufer:
**Note**
Follow-up to #3154. I decided to open a new PR as mentioned since we changed a lot since then and it was easier (avoiding rebasing, etc.). Feel free to close the old PR when you think that it's not useful anymore.

**Summary**
This PR add two new components to the APIv2

1. Prefetching Mixins : We introduce two new mixins that can be used in place of their standard counterparts found in the django API. Both mixins add to the endpoints they target (read and list) the support for prefetching fields at relational depth 1. The fields that can be prefetched are those that are represented by One-to-One and One-to-Many relationships. Invalid field are simply ignored. 

    The prefetched data is returned in the response under the top-level attribute `prefetch`. This attribute acts as a form a cache where the corresponding data is stored under two level of indirection, more specifically the name of the field referencing it and its primary key. This two level indirection allow us to easily reconstruct the object on the client-side while avoiding duplicated data.

     Here is a example to illustrate prefetching on `/findings/{id}`. Here we queried the endpoint with the query parameter `?prefetch=defect_review_requested_by,found_by`
```
{
  "id": 2,
  "severity": "Critical",
  "description": "It's possible to by-pass the authentication by sending xyz payload par of a GET request.",
  "mitigation": "Implement proper input sanitization.",
  "impact": "Critical, it's possible to access the application without credentials.",
  "defect_review_requested_by": 2,
  "found_by": [
    3
  ],
  ...,
  "prefetch": {
    "defect_review_requested_by": {
      "2": {
        "id": 2,
        "username": "techmanager",
        "first_name": "Technical",
        "last_name": "Manager",
        "email": "tech.manager@company1.com",
        "last_login": null,
        "is_active": true,
        "is_staff": false,
        "is_superuser": true
      }
    },
    "found_by": {
      "3": {
        "id": 3,
        "name": "Pen Test",
        "static_tool": false,
        "dynamic_tool": false
      }
    }
  }
}

```

2. ComposableSchema : drf_yasl2 offers support for swagger schema but it is not flexible enough for our application. In our case, we would like to easily add documentation for prefetching and other fields (related_fields, etc.). The main problem is that, a ViewSet expects a single swagger schema generator and we cannot compose multiple ones that would add different parameters or fields in the response. Therefore, we introduce wrappers around the structure defined in the drf_yasl2 library that allow easy schema composition and also the ability to add new query parameters and new fields in the response of an endpoint. Currently we have the following structure

     - ComposableSchema : base class for composable schema, each composable structure must inherit from it
     - IdentitySchema : An empty schema whose only purpose is to easily chain schema composition operation
     - ExtraParameters : schema that adds new parameters inside the query structure
     - ExtraResponseField : schema that adds new fields inside the specified responses
    
     With these classes we add automatic documentation for endpoints supporting prefetching.

As a final note, this PR also add prefetching to the finding and product endpoints as they are the main reason why this mechanism is introduced.

### DIFF
--- a/dojo/api_v2/prefetch/__init__.py
+++ b/dojo/api_v2/prefetch/__init__.py
@@ -1,2 +1,4 @@
 from .mixins import PrefetchListMixin, PrefetchRetrieveMixin
 from .schema import get_prefetch_schema
+
+__all__ = ['PrefetchListMixin', 'PrefetchRetrieveMixin', 'get_prefetch_schema']
--- a/dojo/api_v2/prefetch/__init__.py
+++ b/dojo/api_v2/prefetch/__init__.py
@@ -1,0 +1,2 @@
+from .mixins import PrefetchListMixin, PrefetchRetrieveMixin
+from .schema import get_prefetch_schema
--- a/dojo/api_v2/prefetch/mixins.py
+++ b/dojo/api_v2/prefetch/mixins.py
@@ -1,0 +1,42 @@
+from rest_framework.response import Response
+from rest_framework.mixins import ListModelMixin, RetrieveModelMixin
+from .prefetcher import _Prefetcher
+
+
+class PrefetchListMixin(ListModelMixin):
+    def list(self, request, *args, **kwargs):
+        prefetch_params = request.GET.get("prefetch", "").split(",")
+        prefetcher = _Prefetcher()
+
+        # Apply the same operations as the standard list method defined in the django rest framework
+        queryset = self.filter_queryset(self.get_queryset())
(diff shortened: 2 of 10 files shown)


valentijnscholten:
Looks good, can you add some test cases to trigger this new feature?

RomainJufer:
I added some tests, they should be generic enough to run on any `viewset` that inherits from the prefetch mixin.

valentijnscholten (approved):

madchap (approved):

PR #3521, opened 2020-12-22
Various Bug Fixes. Make 'Active' Default On Scan Import Forms

devGregA:
BugFix 1. I **believe** from tagulous, a notion of a ‘CastTaggedQuerySet' is introduced. When Googling, I only saw this type with reference to tagulous. Our report generator does not know how to handle this. QuerySets are only referenced in two locations in the generate report method and only come from a few sources. There maybe a more eloquent fix, but to keep consistent with other if else logic.  I simply added an 'or' in the same style which seems to have corrected the issues. I believe we want to treat CastTaggedQuerySet in the same fashion as QuerySets.

BugFix 2. Function not updated to reflect removal of args parameter. Line 21 helper/notifications still has args even though it appears all other functions were rewritten to use kwargs. create_notification causes closing engagements / tests / findings / to throw an error. I've removed the args to make the methods consistent, which seems to resolve my issue.

Enhancement 1. Setting default for scan import status to active. This seems to be consistent with our current usage. 2. Not having by default makes it easy to create many inactive / mitigated findings from reupload usage.

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -1358,6 +1358,10 @@ class MetricsFindingFilter(FilterSet):
             'numerical_severity'
         ).values_list('severity', 'severity').distinct()
         if get_current_user() is not None and not get_current_user().is_staff:
+            logger.debug('*******************')
+            logger.debug('FORM FIELDS ARE')
+            logger.debug(self.form.fields)
+            logger.debug('*******************')
             self.form.fields[
                 'test__engagement__product__prod_type'].queryset = Product_Type.objects.filter(
                 authorized_users__in=[get_current_user()])
@@ -1366,7 +1370,7 @@ class MetricsFindingFilter(FilterSet):
                 Q(engagement__product__authorized_users__in=[get_current_user()]) |
                 Q(engagement__product__prod_type__authorized_users__in=[get_current_user()]))
             self.form.fields[
-                'duplicate_finding'].queryset = Finding.objects.filter(
+                'duplicate'].queryset = Finding.objects.filter(
                 Q(test__engagement__product__authorized_users__in=[get_current_user()]) |
                 Q(test__engagement__product__prod_type__authorized_users__in=[get_current_user()]))
 
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -1365,10 +1365,6 @@ class MetricsFindingFilter(FilterSet):
                 'test'].queryset = Test.objects.filter(
                 Q(engagement__product__authorized_users__in=[get_current_user()]) |
                 Q(engagement__product__prod_type__authorized_users__in=[get_current_user()]))
-            self.form.fields[
-                'duplicate'].queryset = Finding.objects.filter(
-                Q(test__engagement__product__authorized_users__in=[get_current_user()]) |
-                Q(test__engagement__product__prod_type__authorized_users__in=[get_current_user()]))
 
     class Meta:
         model = Finding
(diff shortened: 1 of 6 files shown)


mtesauro (approved):
Approved

valentijnscholten (approved):
Some nice fixes. I think the `type(obj).__name__ == "QuerySet"` could be made more safe by using `isinstance` as that wil take inheritance into account. Up to you if that's this PR or the next :-)

devGregA:
Two additional bug fixes. 
1. Certain metrics pages are broken for non-staff users. A key error was created from: https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/filters.py#L1085. Upon examining the form fields, the field is actually 'duplicate' on that specific form.
2. Reupload was broken and caused all findings to always be mitigated on reupload. Status is decided by diffing sets here: https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/test/views.py#L841                 to_mitigate = set(original_items) - set(reactivated_items) - set(unchanged_items)
The issue is that original_items is not of the same type and thus the intersection can't be calculated. Currently, the type of 'original_items' is 'CastTaggedQuerySet' which surprised me. I understand transitional cast classes, but this queryset isn't even related to tags, and I believe taguluous is the only addition using such cast classes. Even before tagulous, I'm not sure how going to flat object created a successful comparison, or if it was bugged previously. My suspicion is that because ID is our unique id, it somehow functioned as intended. I think the larger problem I'm not addressing may be the CastTaggedQuerySet class. I'm investigating how we could re-cast or type in a better fashion, but there isn't much documentation so it will require more digging on my part to come up with a more complete solution. I'd like to go ahead and address the issues while I continue to research the casting, as this may create or have created other undiscovered bugs.

devGregA:
After further review on 2. the loop expects finding ids. This could be addressed by going to objects or sticking with ids. Flattening the other sets may solve the issue. My solution will be to flatten all lists as I'm assuming there is a reason for working with id's rather than objects.

devGregA:
In my local testing I am not able to replicate when a finding_id is a finding. Trying some debug lines.

devGregA:
Removing debugging. Unsure why it is working now.

valentijnscholten:
I think there are unit tests for that reimport so I'm interested what is not working with opening/closing of findings on reimport.
I also think the same logic is in place in the API, so we should keep those aligned.

dojo/filters.py
```diff
@@ -1366,7 +1366,7 @@ def __init__(self, *args, **kwargs):
                 Q(engagement__product__authorized_users__in=[get_current_user()]) |
                 Q(engagement__product__prod_type__authorized_users__in=[get_current_user()]))
             self.form.fields[
-                'duplicate_finding'].queryset = Finding.objects.filter(
+                'duplicate'].queryset = Finding.objects.filter(
```
valentijnscholten, line 1369: the problem is that `duplicate_finding` is excluded from the field list. so this whole statement with assigning a queryset can be removed. the `duplicate` field is a boolean so I think assigning a queryset that to a boolean field will either do nothing or raise an error.

valentijnscholten (changes requested):

dojo/filters.py
devGregA: OH makes sense. Updating.

valentijnscholten (approved):

valentijnscholten:
look good now, i'll leave the merge up to you.

valentijnscholten (approved):

Issue #3522, opened 2020-12-22
Saving an inactive finding should not mark it automatically as mitigated


madchap:
**Bug description**
Upon importing scans as "Inactive" (debating of actually status semantics or use-case is certainly another effort), editing the finding and then saving it moves the finding in "Inactive, Mitigated" status, which is the wrong thing to do.

fyi @Wadeck

PR #3523, opened 2020-12-22
Only mitigate finding if previously active

madchap:
Fixes https://github.com/DefectDojo/django-DefectDojo/issues/3522


valentijnscholten:
Should the same check be applied in the API (haven't checked what's there now).
Also it looks like finding is set as mitigated when marked as false positive? (existing code)

madchap:
> Should the same check be applied in the API (haven't checked what's there now).
> Also it looks like finding is set as mitigated when marked as false positive? (existing code)

I'll take a look in the API.

Right, when FP, it's currently mitigated :man_shrugging: I didn't want to change it at first, but maybe this can be something to do now.

valentijnscholten:
not sure. it might have some benefits to mark the finding as mitigated when marked as false_p. SLA counter stops etc.
but in that case it should be the same for out_of_scope

madchap:
Tested the API, it seems it never had that behavior actually. It's not trying to assume anything.

valentijnscholten:
There is something around mitigated but it's about keeping the is_Mitigated in sync with mitigated. Maybe we should add the same logic there for false poistive and out of scope.

dojo/finding/helper.py
```diff
@@ -0,0 +1,47 @@
...
+        new_state_finding.mitigated = datetime.datetime.now()
+        new_state_finding.mitigated_by = request_user
+        if settings.USE_TZ:
+            new_state_finding.mitigated = timezone.make_aware(new_state_finding.mitigated, timezone.get_default_timezone())
+    elif not new_state_finding.is_Mitigated and new_state_finding.mitigated is not None:
+        finding_status_changed = True
+        new_state_finding.mitigated = None
+        new_state_finding.mitigated_by = None
+
+    logger.info("Finding status after update: {}".format(new_state_finding.status()))
```
valentijnscholten, line 45: lower log level or remove?

valentijnscholten (changes requested):
there are some other places that need to use the new logic: 
image

dojo/finding/helper.py
madchap: It was supposed to be debug definitely.

valentijnscholten (approved):

Issue #3524, opened 2020-12-22
Losing SLA information when mitigated within SLA range 


madchap:
**Bug description**
(When the SLA feature is activated of course)

When a finding gets mitigated in time, its SLA counter gets set to `0` and in blue background (red is for findings that are not mitigated in time, green for those that still have time remaining). When mitigated out of time, also set to `0` and background is orange.

This means we're losing the positive fact that teams have actually resolved an issue in time and met the SLA, or how long after it got resolved despite breaching the SLA.

**Expected behavior**
The blue and orange backgrounds indicating SLA being met is here to stay. However, instead of `0`, the number indicated should simply be frozen to whatever number of days was remaining to breach or past breach.

This makes it easy when reporting on SLA and get accurate metrics.

PR #3525, opened 2020-12-22
Retain SLA days for mitigated findings

madchap:
Addresses https://github.com/DefectDojo/django-DefectDojo/issues/3524

image

Essentially, only the blue SLA would show the number instead of 0. The orange one (mitigated past SLA) already was working.


ccojocar (approved):
LGMT

valentijnscholten (approved):

valentijnscholten:
@madchap  the word 'Freeze' keeps making me thinking something crazy is going on in this PR :-) Would it be better to call it "Retain" instead?

Issue #3342, opened 2020-11-24
APIv2 : Swagger Doc Test Results


RomainJufer:
I post here a first draft of the errors found by the test for the swagger schema. The test works by matching the swagger schema against the result of a query on a given endpoint. It tests the methods : list, read, create, update, partial_update provided that they are supported. Note that it does not check the request schema and currently does not check all the "sub-endpoints" (for example, `/findings/{id}/tags`).

**Results**
- **url :** /engagements/{id}/notes/
**method :** GET, POST,  PATCH
Result model is Engagement while it should be Note
- **url :** /engagements/{id}/accept_risks/
**method :** POST
According to schema it returns a response 201 but in reality it returns a response 200
- **url :** /findings/
**method :** GET
According to schema it returns a Finding but it should returned paginated Finding array
- **url :** /findings/{id}
**method :** GET, PUT, PATCH
request_response field is string according to schema but the value returned is a more complex structure
- **url :** /jira_configurations/
**method :** GET
password is still required but if endpoint queried without being superuser it will not be returned
- **url :** /jira_configurations/{id}
**method :** PATCH
configuration_name has default value blank according to model but does not allow blank value
info_mapping_severity is null but does not allow null value
- **url :** /jira_configurations/{id}
**method :** POST, PUT
same errors as PATCH and GET(above)
- **url :** /jira_configurations/{id}
**method :** GET
same errors as GET
- **url :** /products/
**method :** GET
findings_count is a string according to schema but is an integer
findings_list is a string according to schema but is a list of findings
- **url :** /products/{id}
**method :** GET
same errors as above

In addition, it seems that the endpoint for system settings expect an ID while not really expecting one as the description suggests to always use id=1 ? Furthermore, there is a lot of errors that I did not report here due to fields that are not nullable but for which the response contained a null value. However, upon inspection of the schema they are always for fields that are "read_only" as defined by the openapi standard. So I'm not sure if "read_only" means also nullable or not. According to `https://swagger.io/specification/`, "read_only" means that the property has no guarantee of being sent but it does not say anything about it being null. Does anyone know more about that ?

I will open a PR with the test later in the week as I did not find a way of getting the result of the test without it failing due to assertions.

JorisVanEijden:
> findings_list is a string according to schema but is a list of findings
It's actually an array of integers (finding ids)

PR #3528, opened 2020-12-22
Add swagger schema test with tagged test cases

RomainJufer:
This PR adds a first version of the test suite for the swagger schema. This is the test used to report #3342. I have divided the test cases into two categories. The test that are currently succeeding and the tests that are failing. The later have been tagged as `broken`. By doing so, we can run the first category by calling
`python manage.py test dojo.unittests.test_swagger_schema.py --exclude-tag broken`
and the second category by calling
`python manage.py test dojo.unittests.test_swagger_schema.py --tag broken`

Ideally, this test would be implemented in the current pipeline first, by adding `--exclude-tag broken` to the command called by the job in charge of running the unit tests then by adding a new job that would run only that test but with `--tag broken` and that job would be allowed to fail. By doing so, we wouldn't have a pipeline that fails until we correct all the errors in the schema but we would still ensure that new features are not breaking the schemas that are valid.

As we correct the tests that are broken, we only need to remove the tags so that the corresponding test case will be executed by the standard pipeline.

I don't know if my explanation is clear enough, let me know what you think of this approach :)

**Note** Do not merge already as the pipeline needs to be modified accordingly

### DIFF
--- a/docker/entrypoint-unit-tests-devDocker.sh
+++ b/docker/entrypoint-unit-tests-devDocker.sh
@@ -13,7 +13,13 @@ unset DD_DATABASE_URL
 python3 manage.py makemigrations dojo
 python3 manage.py migrate
 
-python3 manage.py test dojo.unittests --keepdb -v 3 
+echo "Swagger Schema Tests - Broken"
+echo "------------------------------------------------------------"
+python3 manage.py test dojo.unittests -v 3 --no-input --tag broken && true
+
+echo "Unit Tests"
+echo "------------------------------------------------------------"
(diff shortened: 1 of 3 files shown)


valentijnscholten:
Great stuff. Sounds good, so you could adjust https://github.com/DefectDojo/django-DefectDojo/blob/dev/docker/entrypoint-unit-tests-devDocker.sh and https://github.com/DefectDojo/django-DefectDojo/blob/dev/docker/entrypoint-unit-tests.sh to reflect the broken / non-broken stuff.

valentijnscholten (approved):

madchap (approved):

Issue #3527, opened 2020-12-22
Error importing Prisma Cloud / Twistlock CSV file from "Deployed Images" in Prisma/Twistlock


macedogm:
When importing a Twistlock image scan CSV file generated in Prisma Cloud / Twistlock interface in Compute > Monitor > Vulnerabilities > Images > Deployed Images or also from their API [/images](https://cdn.twistlock.com/docs/api/twistlock_api.html#images), DefectDojo will fail with an error when parsing the CSV file.

```
uwsgi_1         | ERROR:django.request:Internal Server Error: /api/v2/import-scan/
uwsgi_1         | Traceback (most recent call last):
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
uwsgi_1         |     response = get_response(request)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
uwsgi_1         |     response = self.process_exception_by_middleware(e, request)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
uwsgi_1         |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
uwsgi_1         |     return view_func(*args, **kwargs)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/rest_framework/viewsets.py", line 125, in view
uwsgi_1         |     return self.dispatch(request, *args, **kwargs)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 509, in dispatch
uwsgi_1         |     response = self.handle_exception(exc)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 469, in handle_exception
uwsgi_1         |     self.raise_uncaught_exception(exc)
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
uwsgi_1         |     raise exc
uwsgi_1         |   File "/usr/local/lib/python3.6/site-packages/rest_framework/views.py", line 506, in dispatch
uwsgi_1         |     response = handler(request, *args, **kwargs)
[...]

PR #2626, opened 2020-07-01
Twistlock CSV Parser Feature

bvcelari:
This template is for your information. Please clear everything when submitting your pull request.

**Note: DefectDojo is now on Python3.5 and Django 2.2.x Please submit your pull requests to the 'dev' branch as the 'legacy-python2.7' branch is only for bug fixes. Any new features submitted to the legacy branch will be ignored and closed.**

When submitting a pull request, please make sure you have completed the following checklist:

- [x] Your code is flake8 compliant.
- [x] Your code is python 3.5 compliant (specific python >=3.6 syntax is currently not accepted).
- [x] If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR.
- [x] Model changes must include the necessary migrations in the dojo/db_migrations folder.
- [x] Add applicable tests to the unit tests.
- [x] Add the proper label to categorize your PR.

Current accepted labels for PRs:
- Import Scans (for new scanners/importers)
- enhancement
- performance
- feature
- bugfix
- maintenance (a.k.a chores)
- dependencies
- New Migration

PR #3532, opened 2020-12-23
Fix issue 3527 while importing some Twistlock scans

macedogm:
Fix issue #3527.

The following improvements were made:
1. Added two global variables to control the column displacement when importing the different type of Twistlock image scan.
2. Wrap finding's `title` and `component_name`, otherwise when they are too long the finding will fail to be imported in the database.
3. Properly validate if the finding severity is `moderate` or `Moderate`.
4. Cast `moderate` severity to `Medium` and `important` to `High`.

Please let me know if you agree with the proposed solution. Best regards.


valentijnscholten:
Thanks! Are you able to provide a sample scan of that other format as part of the unit test in `test_twistlock_parser.py`?

valentijnscholten:
and can you base it against `dev` please.

macedogm:
Yes, I will rebase it against `dev` and provide a sample scan. Will do it today.

macedogm:
@valentijnscholten Can you please have a look now?
I've added the unit tests. The integration tests are still failing, but I think that it's due to another reason than this PR.

valentijnscholten:
can you rebase or merge `dev` into the PR?

macedogm:
Rebased and all checks passed 👍🏻 
Thanks for the help @valentijnscholten !

madchap (approved):
lgtm, thanks :+1:

damiencarol:
@macedogm @madchap @valentijnscholten could it be better to use `csv` module? Like do this:
```
records = csv.DictReader(csv_file)
```
Because the files have headers. Could be better to rely on header column name

macedogm:
@damiencarol Good idea, makes total sense. I'll submit a PR using `csv.DictReader`. It should make the code smaller and works for both types of reports.

macedogm:
Following @damiencarol suggestion, I modified the Twistlock CSV parser to use use `csv.DictReader()`, so the same parser can work without modifications for both types of Twistlock reports.

One important thing to mention is that I removed the [`escape()`](https://github.com/DefectDojo/django-DefectDojo/pull/3532/files#diff-3176294e4014378ed7fcff6167cc0a0e5e077c2dd816ebed034e2c5c23e7e754L19) of the contents, since the the escaping doesn't seem to be done in all report parsers. Wouldn't be the case of moving it to the parts of the DefectDojo responsible for saving the findings in the database and handling the views? (Referencing [this](https://github.com/DefectDojo/django-DefectDojo/pull/2626#pullrequestreview-442591229) comment made by @valentijnscholten on another PR.

Can you all please review the new improvement and see if it makes sense? And if we must add an `escape()` on each field?

Best regards.

dojo/tools/twistlock/parser.py
```diff
@@ -3,55 +3,39 @@
...
-        data_cvss = self.get_field_from_row_or_default(row, cvss_column, '')
-        data_description = self.get_field_from_row_or_default(row, description_column, '')
+        data_vulnerability_id = row['CVE ID']
+        data_package_version = row['Package Version']
+        data_fix_status = row['Fix Status']
+        data_package_name = row['Packages']
+        data_id = row['Id']
+        data_severity = row['Severity'].capitalize()
+        data_cvss = row['CVSS']
+        data_description = description_column = row['Description']
```
valentijnscholten, line 27: what happens if any of these fields are not present. does the dictreader return some kind of default, or is a check needed?

dojo/api_v2/serializers.py
```diff
@@ -1055,6 +1055,10 @@ def save(self, push_to_jira=False):
                 sev = item.severity
                 if sev == 'Information' or sev == 'Informational':
                     sev = 'Info'
+                elif sev.lower() == 'moderate':
+                    sev = 'Medium'
+                elif sev.lower() == 'important':
+                    sev = 'High'
```
valentijnscholten, line 1061: might be best to handle all severities in the same way, i.e. convert to lowercase before comparison?

dojo/tools/twistlock/parser.py
```diff
@@ -132,7 +112,7 @@ def get_item(vulnerability, test):
         # If we're dealing with a license finding, there will be no cvssScore
         if vulnerability['severity'] == 'important':
             severity = "High"
-        elif vulnerability['severity'] == 'moderate':
+        elif vulnerability['severity'].lower() == 'moderate':
             severity = "Medium"
         else:
             severity = vulnerability['severity'].title()
```
valentijnscholten, line 118: not sure what values title can have but the parser needs to make sure only to insert severities known to dojo, otherwise there can be unpredictable results.

valentijnscholten (changes requested):

valentijnscholten:
Thanks, using some standard components is always good. I don't think we want/need to escape in the parser. Defect Dojo should not trust (but escape) any user provided data  in the database anyway upon rendering pages.

dojo/api_v2/serializers.py
damiencarol: 1) I agree with @valentijnscholten comment. If you really want to handle `moderate` and `important` values in API v2 code, you should follow the same logic that handles `Information`/`Informational` values.
2) I think this change is not good. it will impact every parser. We should rely on the parsers to set the severity to the good values of the `dojo/model.py` file:
```Python
    SEVERITIES = {'Info': 4, 'Low': 3, 'Medium': 2,
                  'High': 1, 'Critical': 0}
```
Each tool have his own logic, the contract is that the parser set the value to one of `Info`, `Low`, `Medium`, `High` or `Critical`.

dojo/tools/twistlock/parser.py
damiencarol: `.title()` do an uppercase for the first letter and a lower case for the rest of the string. It is a good function for severity in parsers code as for now, we have `Info`, `High`, etc... as values.
My advice here is just to encapsulate the logic of converting severity values from the format to Dojo values in a separate function. And just call it when creating the finding, like this:
```Python
severity = self.convert_severity(vulnerability['severity'])
```

dojo/tools/twistlock/parser.py
damiencarol: @valentijnscholten `csv.DictReader(csv_file)` return dictionnaries for items. So we could use `row.get('column', default_value)` if we want something more stable if one column is missing.

dojo/tools/twistlock/parser.py
```diff
@@ -3,55 +3,39 @@
...
-        data_id = self.get_field_from_row_or_default(row, id_column, '')
-        data_severity = self.get_field_from_row_or_default(row, severity_column, 'Info').capitalize()
-        data_cvss = self.get_field_from_row_or_default(row, cvss_column, '')
-        data_description = self.get_field_from_row_or_default(row, description_column, '')
+        data_vulnerability_id = row['CVE ID']
+        data_package_version = row['Package Version']
+        data_fix_status = row['Fix Status']
+        data_package_name = row['Packages']
+        data_id = row['Id']
+        data_severity = row['Severity'].capitalize()
```
damiencarol, line 25: `capitalize` is useless IMHO. See my comment on creating a function to manage severity values below.

damiencarol (changes requested):
The pull request is good and adds a lot of new unit tests.
The switch to use `textwrap` and `csv` module are good for me. I made some comments to improve the quality of the parsing logic for severities.
Tagging as `Request changes` as I think the modifications in `dojo/api_v2/serializers.py` are not aligned with the rest of the code.

macedogm:
Thanks for all the feedback. I'll work on the changes today and update the PR.

dojo/tools/twistlock/parser.py
macedogm: Validation for missing columns improved as suggested - `row.get('column', default_value)`.

dojo/tools/twistlock/parser.py
macedogm: Created a [function](https://github.com/DefectDojo/django-DefectDojo/pull/3532/files#diff-3176294e4014378ed7fcff6167cc0a0e5e077c2dd816ebed034e2c5c23e7e754R148) to properly parse the severity. Modified both CSV and JSON parser to use the function.

dojo/tools/twistlock/parser.py
macedogm: Fixed in the conversation above.

macedogm:
Reverted the modifications in `dojo/api_v2/serializers.py` in favour of using the function inside the parser to properly parse the severity.
Also added a validation for the [finding's title](https://github.com/DefectDojo/django-DefectDojo/pull/3532/files#diff-3176294e4014378ed7fcff6167cc0a0e5e077c2dd816ebed034e2c5c23e7e754R27-R32), since in some cases, mostly when the finding is related to a compliance check from Twistlock, it has no CVE and no package, so it's displayed as `:-` (image 1). If this case happens again, it will now display the description of the finding instead (image 2).

Image 1
image

Image 2
image

Please let me know if this makes sense or not, or if you would suggest a more pragmatic approach.

Sorry for the back and forth in the PR, Python is not my main language. Appreciated all the recommendations. 👍🏻

dojo/api_v2/serializers.py
macedogm: Created a [function](https://github.com/DefectDojo/django-DefectDojo/pull/3532/files#diff-3176294e4014378ed7fcff6167cc0a0e5e077c2dd816ebed034e2c5c23e7e754R148) to properly parse the severity. Modified both CSV and JSON parsers to use the function.

damiencarol (approved):
LGTM

valentijnscholten (approved):

valentijnscholten:
thanks for contributing and sticking with us through the recommendations :-)

PR #3536, opened 2020-12-23
Add PyJWT to requirements.txt

squ1rr3lly:
PyJWT 2.0.0 renamed some of the exceptions, which breaks social_core 3.4.0 that is used by social-auth-app-django 4.0.0. This causes new installations or container builds to throw a 500 error on attempting to log in.

From PyJWT 2.0.0 CHANGELOG:
`Removed `ExpiredSignature`, `InvalidAudience`, and `InvalidIssuer`. Use `ExpiredSignatureError`, `InvalidAudienceError`, and `InvalidIssuerError` instead.`

Requiring pyjwt==1.7.1 resolves the issue.


requirements.txt
```diff
@@ -69,3 +69,4 @@ django-debug-toolbar==3.1.1
 django-debug-toolbar-request-history==0.1.3
 vcrpy==4.1.1
 vcrpy-unittest==0.1.7
+pyjwt==1.7.1
```
damiencarol, line 72: The real name of the package is `PyJWT` could we keep the same case?

squ1rr3lly: Verified it isn't an issue to keep the case. Committed the change.

madchap:
Let's take this one instead of https://github.com/DefectDojo/django-DefectDojo/pull/3535

Unsure why the DSOP and xlsx error, however.

madchap:
Rebased on `dev` (was `master`).

madchap:
Thanks @squ1rr3lly !

PR #3538, opened 2020-12-23
Use full absolute url in notifications

### LINKED ISSUE #3537: Broken Slack notifications when username is not set
**Bug description**
When sending a Slack notification without a username set in the profile, the application throws an error and notifications are not sent.

**Steps to reproduce**
1. Set the Slack app token, user e-mail and check the 'add_product' box for slack notifications. Do not fill "slack username"field.
2. Add a new product
3. An error similar to the following is thrown and no notifications are received.

Screen Shot 1399-10-03 at 16 46 26

**Expected behavior**
The application should fetch slack user id from the e-mail by querying Slack api.

- [ ] Kubernetes
- [X] Docker

marcosValle:
I just realized this error is due to missing imports and not to a problem in the code logic.


madchap:
Painful experience than to scroll down the unittests github actions logs. Can you run the unittests locally @marcosValle and tell us where it fails? (I still could not reach the bottom on GHA). Thanks.

marcosValle:
> Painful experience than to scroll down the unittests github actions logs. Can you run the unittests locally @marcosValle and tell us where it fails? (I still could not reach the bottom on GHA). Thanks.
> 
> Something like `docker-compose exec uwsgi /bin/bash -c 'python manage.py test dojo.unittests -v2'`

It seems like `others.tpl` was not loading the necessary filters. All unit tests should be ok now.

valentijnscholten:
There is also a 'download raw log' on the top right somewhere to view a raw txt file in your browser with the full logs.

valentijnscholten (approved):

madchap (approved):

PR #3542, opened 2020-12-24
Importer: Burp REST API (Fix #3447)

### LINKED ISSUE #3447: Burp suite json format
**Scanner Name**
Burp Suite **Pro**

**Sample File**
[example.json](https://github.com/DefectDojo/django-DefectDojo/files/5673899/example.txt)

**Where is a problem with existing reporters?**
They don't support json and burp can not export it's results in xml via REST API. 
It makes integration very difficult.

**Why is it important?**
It can be integrated in CI/CD in docker (will update this ticket with an example).
But it is required to import it in DefectDojo to make the process simple and support this tool as well as another tools.

PS. It is also required for building metrics
https://www.youtube.com/watch?v=6FGV4OcrIB8&t=2s

damiencarol:
this PR fixes #3447


madchap:
You seem to be missing quite a few files to have this done properly.

You can take a look at some parsers at https://github.com/DefectDojo/django-DefectDojo/labels/Import%20Scans and see what's left to do.

madchap:
Also, you did mean Burp Pro and not Enterprise, correct? Such as in here: https://portswigger.net/blog/burps-new-rest-api

If so, it may be good to take a look if this new ingestion could be merged with the existing Burp Pro importer to avoid having many "Burp Pro". E.g. by making some assumption on the type of file being uploaded (XML Vs JSON).

valentijnscholten:
Thanks. Everytime a dictionary is accessed like this `issue_event["issue"]["type_index"]` there's a chance of a KeyError, one of the most common reasons why we see imports failing after some time. Can you make all these cases "None-safe" by using something like `issue_event.get('issue', None)` etc.?

damiencarol:
@valentijnscholten for the "Non-safe" pb, it is covered by the `if`:
```
if "issue_found" == issue_event.get("type"):
```
If the event is `issue_found`, the attribute `issue` is here. I can move some to `issue_event["issue"].get("type_index", None)` but some attribute are mandatory. It doesn't make sense to not have them.

valentijnscholten:
that's what everybody thinks when they write a parser, and then after a year we get 500 errors because a field was added/removed/renamed

damiencarol:
@madchap I added the missing pieces. My pull request was not finished when you reviewed it 😅 . Regarding how to "class" this new parser. We can't use the other ones. This one is dedicated to the data of the REST API. It doesn't depend on Pro/Community/Business version and we can't rely on the file name as it is JSON response from an API... so no file name.

@valentijnscholten As you requested, I added more `dict.get('attr', 'def')`

dojo/tools/factory.py
```diff
@@ -108,6 +109,8 @@ def import_parser_factory(file, test, active, verified, scan_type=None):
         parser = BurpXmlParser(file, test)
     elif scan_type == "Burp Enterprise Scan":
         parser = BurpEnterpriseHtmlParser(file, test)
+    elif scan_type == "Burp REST API":
```
valentijnscholten, line 112: Suggest to name it "Burp RESP API report" or similar. We also have other imports such as SonarQube API which actually goes out to the sonarqube API to retrieve data.

valentijnscholten: nvm I see that there's already some info on the actual dropdown further up in the PR.

valentijnscholten (approved):

dojo/tools/factory.py
damiencarol: Sonarqube parser do the same.
```
SCAN_SONARQUBE_API = 'SonarQube API Import'
```

valentijnscholten: yes, but that scan is actually import findings via the sonarcube api. The burp rest api parser processes a file that is uploaded to defect dojo.

damiencarol: ok

dojo/templates/dojo/import_scan_results.html
```diff
@@ -62,6 +62,7 @@ <h4> JIRA </h4>
 	<li><b>Burp XML</b> - When the Burp report is generated, the recommended option is Base64 encoding both the request and
 	    response fields. These fields will be processed and made available in the 'Finding View' page.</li>
 	<li><b>Burp Enterprise Scam</b> - Import Burp Enterprise Edition findings in HTML format</li>
```
madchap, line 64: @damiencarol could you please take the chance to correct the typo here?

madchap (approved):

damiencarol:
@madchap fixed the typo you found. Good catch!

Issue #3548, opened 2020-12-24
metrics: extremely slow loading (product) metrics


valentijnscholten:
since #2974 and #3410 the metrics have become very slow. I'm seeing 45s loading times on a 15k instance and >3 minutes loading time on a 700k instance.

Issue #3547, opened 2020-12-24
metrics: no findings in product metrics due to swallowed exception


valentijnscholten:
on the product metrics there are no findings shown. there is a try-catch that swallows the exception, but the exception is: 

```
Traceback (most recent call last):
  File "/home/valentijn/dd/dojo/product/views.py", line 237, in finding_querys
    start_date = findings_qs.earliest('date').date
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/models/query.py", line 646, in earliest
    return self._earliest(*fields, field_name=field_name)
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/models/query.py", line 643, in _earliest
    return obj.get()
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/models/query.py", line 406, in get
    raise self.model.DoesNotExist(
dojo.models.Finding.DoesNotExist: Finding matching query does not exist.
```

valentijnscholten:
Seems to be caused by multiple levels of 'get earliest finding' and 'get earliest date' and daterangefilter in the productmetricsfindingfilter.

valentijnscholten:
![image](https://user-images.githubusercontent.com/4426050/103103502-c94c0680-4621-11eb-8811-c49a7840c150.png)

valentijnscholten:
happens when a product has no new findings in the last 30 days. it looks like the code is trying to prevent this from happening, but is failing.

PR #3549, opened 2020-12-24
(product) metrics: fixes and speedup

### LINKED ISSUE #3546: metrics: exception when browsing to product metrics
when browsing directly to a product metrics page, there is no referrer and this generates an exception:

```
Internal Server Error: /product/156/metrics
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 113, in _get_response
  File "/home/valentijn/dd/dojo/user/helper.py", line 79, in _wrapped
    return view_func(request, lookup_value, *args, **kwargs)
  File "/home/valentijn/dd/dojo/product/views.py", line 408, in view_product_metrics
    view = identify_view(request)
  File "/home/valentijn/dd/dojo/product/views.py", line 206, in identify_view
    if referer.find('type=Endpoint') > -1:
AttributeError: 'NoneType' object has no attribute 'find'
```

valentijnscholten:
fixes #3546 #3547 #3548 #3207 #3554 #3555

the DateRangeFilter on product metrics is now disabled because it hides all findings if you have no findings in the last 30 days.
we have seen reports on this on slack and I am seeing it in my prod instance as well.

also removed old legacy code retrieving `ra_count` with raw sql.


valentijnscholten:
@danielnaab fyi

danielnaab:
@valentijnscholten that all looks great. I also have a few more changes we've got on an internal fork I haven't pushed up to #3438, but won't get the chance to test that until after the holidays. One or maybe two are in the same areas you hit here. We also removed a couple filters from the view and wanted to bring up if that made sense or not.

Happy holidays! We'll try your improvements and catch up with you in 2021.

madchap (approved):

PR #3559, opened 2020-12-28
Add missing modifications for SARIF format

damiencarol:
Add missing data in fixtures and some UI elements to import a SARIF report from the UI.


valentijnscholten (approved):

madchap (approved):

madchap:
Classifying as "bugfix" for lack of better labels, considering an importer should always be available through the UI.

PR #3562, opened 2020-12-28
system settings through apiv2

manuel-sommer:
This PR extends the API capabilities of DefectDojo. As of yet, not all System Configurations are accessible through the API. Thus, this PR advances DefectDojo's API capabilities to e.g. configure DefectDojo in an automatic way through IaC.


valentijnscholten:
Thanks. Would it be better to just use the built in "ModelSerializer" so we don't have to have all the custom code / listing / mapping of fields?

manuel-sommer:
@valentijnscholten Thanks for the comment. Is it fine like this?

valentijnscholten:
Thanks. Does it work as expect with POST updating all fields and PATCH allowing update of a subset of fields? I think evertyhing related to tags can be removed from this serializer.

manuel-sommer:
@valentijnscholten I tested POST and PATCH. It worked fine.

valentijnscholten (approved):

madchap (approved):
Looks simple :) Cheers.

PR #3566, opened 2020-12-30
Add Arbitrary File Uploads

Maffooch:
The Manage Finding Images page only allows for images to be uploaded, but no other file formats. This PR allows for all file types to be upload, viewed, and downloaded, from the Engagement, Test, and Finding levels. 

There is now a message on the manage finding images page explaining that the feature will be removed on 06/31/2021


dojo/db_migrations/0068_file_upload.py
```diff
@@ -0,0 +1,37 @@
...
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('title', models.CharField(max_length=100, unique=True)),
+                ('file', models.FileField(upload_to=dojo.models.UniqueUploadNameProvider('uploaded_files'))),
+            ],
+        ),
+        migrations.AddField(
+            model_name='engagement',
+            name='files',
+            field=models.ManyToManyField(blank=True, editable=False, to='dojo.FileUpload'),
```
valentijnscholten, line 25: Does this have to be a many-to-many? I think each file can only ever belong to at most 1 engagement/finding/test?

valentijnscholten:
Two things the come to mind:
- Do we need to keep both features in parallel? It might be better to just remove the finding images feature and add some RunPython to the migration to migrate existing images to the new data structure?
- This PR allows uploading to Finding, Test and Engagement. What if we want to upload to Product or Product Type? Or to an endpoint? Do we (maybe later?) want a more generic solution where you can upload files to any "ContentType" (model) in defect dojo? Or at least a construct where we could add more models to upload to without adding fields to the database schema?

dojo/db_migrations/0068_file_upload.py
Maffooch: That would be the hope, but could be nasty to debug that if a lot of files are present and you forgot where you put it the first time.

Maffooch:
- I was thinking about that, but it could break a lot of workflows for users that are uploading images via the api. Once we yank it, nobody can be upset since the warning message had been present for 6 months. 
- Hmm I didn't know that was possible. Is there an example of that somewhere in dojo? Not totally sure how we could add more models without touching the database schema.

valentijnscholten:
Good point about the API. In theory we could leave the existing finding image endpoints / function in place in the API, but change the logic to store the images in the new generic upload way. But maybe that's too much effort. The main thing I would prefer is that when we remove the finding image functionality, we have a django migration that migrates existing data. If we leave the finding images for now, the migration also will have to wait.

I think we have a bunch off stuff being deprecated in the coming months. So we could at some point start a 2.0 branch where we remove all that stuff and then we have 1 release with all the things removed and a big shiny change log mentioning all that.

The generic relationship thing I was referring to is in the content_type framework of Django: https://docs.djangoproject.com/en/3.1/ref/contrib/contenttypes/

Maffooch:
I think the generic relationship thing would require a rewrite.

mtesauro (approved):
Approved

devGregA (approved):

PR #3534, opened 2020-12-23
Add support for GitLab Dependency Scanning reports

macedogm:
Fix #3533.
Basically the GitLab Dependency Scanning reports are quite similar to the SAST reports, except for:
1. `vuln['category']` is `dependency_scanning`.
2. The analyzers used by GitLab don't provide the `scanner_confidence` property.

It seemed to make sense to create a new importer, despite the fact that the changes were only 2 compared to the current SAST importer. The changes might increase in the future.

Please let me know if this make sense or not. Best regards.

PR #3567, opened 2020-12-30
Parameter confidence is optional and CWE's are not always numerical in GitLab SAST

StefanFl:
GitLab SAST doesn't always set the parameter ``confidence`` for a vulnerability, leading to an exception when importing the vulnerabilities.

### DIFF
--- a/dojo/tools/gitlab_sast/parser.py
+++ b/dojo/tools/gitlab_sast/parser.py
@@ -90,9 +90,7 @@ def get_item(vuln, test):
         title = '[{} severity] {}'.format(severity, title)
         severity = 'Info'
     numerical_severity = Finding.get_numerical_severity(severity)
-    scanner_confidence = None
-    if 'confidence' in vuln:
-        scanner_confidence = Finding.get_number_severity(vuln['confidence'])
+    scanner_confidence = get_confidence_numeric(vuln.get('confidence', 'Unkown'))
 
     mitigation = ''
     if 'solution' in vuln:
@@ -139,3 +137,14 @@ def get_item(vuln, test):
                       dynamic_finding=False)
 
     return finding
+
+
+def get_confidence_numeric(argument):
+    switcher = {
+        'Confirmed': 1,    # Certain
+        'High': 3,         # Firm
+        'Medium': 4,       # Firm
+        'Low': 6,          # Tentative
+        'Experimental': 7  # Tentative
+    }
+    return switcher.get(argument, None)
--- a/dojo/tools/gitlab_sast/parser.py
+++ b/dojo/tools/gitlab_sast/parser.py
@@ -90,7 +90,9 @@ def get_item(vuln, test):
         title = '[{} severity] {}'.format(severity, title)
         severity = 'Info'
     numerical_severity = Finding.get_numerical_severity(severity)
-    scanner_confidence = Finding.get_number_severity(vuln['confidence'])
+    scanner_confidence = None
+    if 'confidence' in vuln:
+        scanner_confidence = Finding.get_number_severity(vuln['confidence'])
 
     mitigation = ''
     if 'solution' in vuln:
--- a/dojo/tools/gitlab_sast/parser.py
+++ b/dojo/tools/gitlab_sast/parser.py
@@ -102,7 +102,10 @@ def get_item(vuln, test):
     if 'identifiers' in vuln:
         for identifier in vuln['identifiers']:
             if identifier['type'].lower() == 'cwe':
-                cwe = identifier['value']
+                if isinstance(identifier['value'], int):
+                    cwe = identifier['value']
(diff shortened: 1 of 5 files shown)


damiencarol:
I just reviewed your modification and I have 2 comments here:
1) as the file is loaded with `json.load()` you have dictionaries for the items.
It means that you should use something like that to manage optional attributes:

```Python
scanner_confidence = vuln.get('confidence'], 'Unknown')
```

It seems that the default value is `Unknown`
2) the current code doesn't make any sense.

```Python
scanner_confidence = Finding.get_number_severity(vuln['confidence'])
```
[see here](https://github.com/DefectDojo/django-DefectDojo/pull/3567/files#diff-b50bc2a09a1d04194c4525846f3635f10820bca1ef2ad5acb0ddbd3e430e2248L93)
the function `get_number_severity` has no link to the confidence. **For me it is a bug.**

damiencarol:
Also, @StefanFl could you add another report and keep the old one? Could be good to have the report that makes you exceptions.
If you are ok to provide your report, keep in mind that it's an open-source project so remove the personal details in the report.
Thanks in advance.

dojo/tools/gitlab_sast/parser.py
```diff
@@ -90,7 +90,9 @@ def get_item(vuln, test):
         title = '[{} severity] {}'.format(severity, title)
         severity = 'Info'
     numerical_severity = Finding.get_numerical_severity(severity)
-    scanner_confidence = Finding.get_number_severity(vuln['confidence'])
+    scanner_confidence = None
+    if 'confidence' in vuln:
+        scanner_confidence = Finding.get_number_severity(vuln['confidence'])
```
damiencarol, line 95:
```suggestion
        scanner_confidence = vuln.get('confidence', 'Unknown')
```

damiencarol (changes requested):
See my previous comments. Some modifications could be different.

damiencarol:
@StefanFl does this issue is linked to #3534?

StefanFl:
@damiencarol GitLab has 2 different scanners. GitLab SAST is static code analysis (the parser of this PR), while the PR you have referenced parses the results of the GitLab Dependency Scanner. File format for these 2 scanners seems to be pretty similar.

Thank you for the review, will change the files according to your suggestions.

StefanFl:
Interesting one, it was definitely a bug. GitLab SAST can produce a confidence, but the values (search for confidence in https://docs.gitlab.com/ce/development/integrations/secure.html) need to be mapped to the values understood by DefectDojo (see ``Finding.get_scanner_confidence_text()``).

The new report is an amended excerpt from findings for the OWASP Juice Shop, I think this should be allowed.

StefanFl:
Please wait with the review, I think there is another problem with these reports

damiencarol:
Take your time and happy new year!
For the record:
[https://docs.gitlab.com/ce/development/integrations/secure.html#severity-and-confidence](https://docs.gitlab.com/ce/development/integrations/secure.html#severity-and-confidence)

> The confidence ranges from Low to Confirmed, but it can also be Unknown, Experimental or even Ignore if the vulnerability is to be ignored. Valid values are: Ignore, Unknown, Experimental, Low, Medium, High, or Confirmed
>
> Unknown values means that data is unavailable to determine it’s actual value. Therefore, it may be high, medium, or low, and needs to be investigated. We have provided a chart of the available SAST Analyzers and what data is currently available.

StefanFl:
All values are mapped apart from `

-  ``Unkown`` where the confidence is set to ``None`` 
- ``Ignore`` where the confidence is set to ``None`` as well, because there is no respective value in DefectDojo

For the others see the method ``get_confidence_numeric(argument)`` in the parser

StefanFl:
A further test showed, that GitLab sometimes writes non-numerical CWE's which can not be stored in the database. There is a fix for it as well know including a new test case. 

Please re-review the whole PR, thank you.

damiencarol (approved):
LGTM, high quality modification, this will make Gitlab SAST parser more robust!

One minor feedback, you could have used the feature of subtests to check the items in the unit tests: [https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests)

damiencarol:
@valentijnscholten @madchap it's ready for review.

valentijnscholten (approved):

damiencarol:
@StefanFl why did you closed this PR ?

StefanFl:
Oops, that wasn't intended, thanks @damiencarol for notifying me.

valentijnscholten (approved):

damiencarol (approved):

madchap (approved):
good job @StefanFl and thanks for the tight review @damiencarol :)

PR #3572, opened 2020-12-31
Add the close_old_findings parameter also to reimport API

ccojocar:
Add the `close_old_findings` parameter to `reimport` API similarly like for `import` API. This param will control if the old findings are going to be mitigated or not. The change maintains the current behaviour by mitigating the old findings if no parameter is set in the reimport API request. 

The reason for this change, it is triggered by one of our use cases where some git repositories are continuously scanned for secrets. After the initial scan, when the entire commits history is checked, the scanner goes over the commits only from the last 2 days. This generates just a diff of findings which when they get reimported leads to the mitigation of all other findings. Obviously those findings are not yet mitigated. That being said, we want to control when the old findings are mitigated by the reimport command.

cc @madchap @yilmi


madchap (approved):

madchap:
lgtm, especially as this does not change the current behavior and allows for more flexibility.

valentijnscholten:
Please keep UI and API logic in sync.

ccojocar:
> Please keep UI and API logic in sync.
@valentijnscholten what exactly do you mean by this? Can you provide some pointer to the UI part which is out of sync?

madchap:
> > Please keep UI and API logic in sync.
> 
> @valentijnscholten what exactly do you mean by this? Can you provide some pointer to the UI part which is out of sync?

Please have a look at https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/test/views.py#L618

ccojocar:
@valentijnscholten @madchap I added it the `close_old_findings` as a form param in the ReImport viewer in order to keep the functionality in sync with the API.

valentijnscholten:
Thanks. Sorry if my remarks was not clear, but unfortunately currently the reimport logic is duplicated in two places (and slighly different). We want to refactor this at some point, which makes it even more important to make sure the behaviour is covered by the unit tests or integration tests. Could you update the existing tests around import/reimport in `test_import_reimport` to reflect the new parameter?

Also, to have the new boolean appear on the form, it would need to be added to ReImportScanForm in `forms.py`, where a help_text can also be supplied to explain the behaviour.

valentijnscholten (changes requested):
see previous comment

ccojocar:
@valentijnscholten I added the field into the form and also added a test to validate the reimport of old findings without closing them.

valentijnscholten:
The number of notes also changes when closing or not closing findings.
Also, could you make the sample files in the same form as the existing one regarding filenames and contents? (Comment out the findings instead of removing them). So it's easy to see the differences when maintaining/updating tests in the future? 
I spent quite some time on making the tests readable / understandable / maintainable.

Also you probably can use the existing sample files as there is already 1 or more testcases testing the reimport with some findings no longer in the report.

ccojocar:
@valentijnscholten I reworked the unit test following your comments and it is now green. Please have a look if you are happy with the current implementation. Thanks!

valentijnscholten (approved):
thanks

madchap (approved):

PR #3578, opened 2021-01-04
Component enhancements

ricardomeulendijks:
Added and changed functionality based on finding components as listed below: 

- Added second level menu item for components
- Added autocomplete for component names in filters
- Added default ordering based on total components amount descending
- Added filter to filter on findings with empty components
- The component filter has two filters with the same name (Product type). Changed one to product name 
- Component versions on one line to get a better overview of amount of findings per component . See the screenshot below (tested on MySQL and PostgreSQL)

defectdojo_oneline_versions


valentijnscholten (approved):

valentijnscholten:
Thanks. I like that now there is 1 row per component name so that's it's more clear which dependency is causing the most findings. Unit tests still seem to work, so approved.

madchap (approved):
lgtm :-)

At some point, I have been thinking on how much of a stretch it'd be to integrate some DependencyTracker functionality straight into DefectDojo...

PR #3580, opened 2021-01-05
Update sample data fixture file

Maffooch:


madchap:
Re-ordering? What's the PR about?

Maffooch:
The sample data file would not get loaded with the way it was ordered before. The tags were being referenced before definition. This resolves that issue so that the file is actually worth something again.

valentijnscholten:
should we add a quick unit test that uses this fixture just to make it loads well?

valentijnscholten:
nice, now I will know if I break it again :-)

valentijnscholten (approved):

dojo/fixtures/defect_dojo_sample_data.json
```diff
@@ -40633,26 +40653,6 @@
         "meta_encoded": "{}"
     }
 },
-{
```
damiencarol, line 40636: Why reordering that @madchap ?

valentijnscholten:
I think you can just do 'fixture = blabla.json' and django will load it for you. But explicit load also works

madchap (approved):

PR #3585, opened 2021-01-05
Fix JFrog Xray JSON parser for CWE and CVSS v3 (#3597 fix)

damiencarol:
### Work

- [X] add more unit tests (new file)
- [X] add more tests on Finding attributes
- [X] found that CWE are not well managed (bug) :D 
- [X] found also that CVSS v3 are not well managed also :D (no data)
- [X] reduce test file size
=> fix issue #3597


dojo/tools/jfrogxray/parser.py
```diff
@@ -59,28 +58,25 @@ def get_item(vulnerability, test):
...
-            # some can be "NVD-CWE-noinfo"
-            if not type(cwe) is int:
-                cwe = 0
-
-        if 'cvss_v3' in vulnerability['component_versions']['more_details']['cves'][0]:
-            cvss_v3 = vulnerability['component_versions']['more_details']['cves'][0]['cvss_v3']
-
-        if 'cvss_v2' in vulnerability['component_versions']['more_details']['cves'][0]:
-            cvss_v2 = vulnerability['component_versions']['more_details']['cves'][0]['cvss_v2']
+    cves = vulnerability['component_versions']['more_details'].get('cves', [])
```
madchap, line 67: This will fail if `['component_versions']['more_details']` does not exist. Maybe preserving the previous `if` statement would be good?

dojo/unittests/test_jfrogxray_parser.py
```diff
@@ -12,9 +16,22 @@ def test_parse_file_with_one_vuln(self):
...
         testfile = open("dojo/unittests/scans/jfrogxray/many_vulns.json")
         parser = XrayJSONParser(testfile, Test())
         testfile.close()
         self.assertEqual(3, len(parser.items))
+
+    def test_parse_file_with_many_vulns2(self):
+        testfile = open("dojo/unittests/scans/jfrogxray/many_vulns2.json")
+        parser = XrayJSONParser(testfile, Test())
+        testfile.close()
+        self.assertEqual(357, len(parser.items))
```
madchap, line 32: That's a lot of items, this will slow down the tests. Could we make it a smaller sample?

dojo/tools/jfrogxray/parser.py
```diff
@@ -59,28 +58,25 @@ def get_item(vulnerability, test):
 
     cve = "No CVE on file"
     cwe = 0
+    cvssv3 = None
```
madchap, line 61: possibly make the variable name more explicit, with regards to what it will truly hold?

madchap (changes requested):
Left a few comments, thanks!

dojo/tools/jfrogxray/parser.py
damiencarol: It will hold the value of the atribute `cvssv3` of the `Finding`. I can rename it to `finding_cvss3` if it make things more clear.

dojo/tools/jfrogxray/parser.py
damiencarol: The previous version fail `['component_versions']['more_details']` does not exist. I can improve that.

dojo/tools/jfrogxray/parser.py
madchap: I had to go read what the cvss package was doing and the specific function too. So I guess I meant maybe rename to something like `cvss3_clean_vector` ?

dojo/tools/jfrogxray/parser.py
madchap: Right, I read too quickly. Cheers.

dojo/tools/jfrogxray/parser.py
madchap: Forget me, that's the one part of the model :facepalm:

dojo/unittests/test_jfrogxray_parser.py
damiencarol: Done (2 findings now)

valentijnscholten (approved):

damiencarol:
@madchap could you take another look?

dr3dd589 (approved):

madchap (approved):

Issue #3586, opened 2021-01-05
Test list in engagement view show active findings count and not active-verified count


madchap:
The count here is wrong. Latest dev.

![image](https://user-images.githubusercontent.com/5468769/103671967-eb2b7e80-4f7b-11eb-9db8-3ef42c8e21f3.png)

PR #3590, opened 2021-01-05
Engagement view list active and verified counts

madchap:
fixes https://github.com/DefectDojo/django-DefectDojo/issues/3586 and provides a slightly improved listing.


valentijnscholten:
As discussed would be good to have one way of display for Test / Engagement / Product / Product Type. So everywhere like this Active (Verified). Or everywhere just Active.

aaronweaver:
I agree with @valentijnscholten it should be consistent but this is a good step as I like the ability to see verified as well. I'm not sure if we could create a template that all the views could use so that we aren't continually editing templates in multiple areas.

madchap:
I believe this is now all aligned.

dojo/models.py
```diff
@@ -688,12 +688,29 @@ def findings_count(self):
             # ideally it's always prefetched and we can remove this code in the future
             self.active_finding_count = Finding.objects.filter(mitigated__isnull=True,
                                             active=True,
+                                            verified=False,
```
valentijnscholten, line 691: I think we should stick to the plan we had and active just means active, regardless of verified. i.e. the definitions in https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/finding/views.py

valentijnscholten (changes requested):

dojo/product_type/views.py
```diff
@@ -48,13 +48,22 @@ def prefetch_for_product_type(prod_types):
 
     if isinstance(prefetch_prod_types, QuerySet):  # old code can arrive here with prods being a list because the query was already executed
         active_findings_query = Q(prod_type__engagement__test__finding__active=True,
+                                prod_type__engagement__test__finding__mitigated__isnull=True,
+                                prod_type__engagement__test__finding__verified=False,
+                                prod_type__engagement__test__finding__false_p=False,
+                                prod_type__engagement__test__finding__duplicate=False,
+                                prod_type__engagement__test__finding__out_of_scope=False)
```
valentijnscholten, line 55: all these should go, we should stick to the definitions made earlier and used in https://github.com/DefectDojo/django-DefectDojo/blob/c67c3feb86fe778aac115fcfb93034f1721add2e/dojo/finding/views.py

dojo/models.py
```diff
@@ -688,12 +688,29 @@ def findings_count(self):
...
+            # if prefetched, it's already there
+            return self.active_verified_finding_count
+        except AttributeError:
+            # ideally it's always prefetched and we can remove this code in the future
+            self.active_verified_finding_count = Finding.objects.filter(mitigated__isnull=True,
+                                            active=True,
+                                            verified=True,
+                                            false_p=False,
+                                            duplicate=False,
+                                            out_of_scope=False,
```
valentijnscholten, line 710: same

dojo/engagement/views.py
```diff
@@ -269,7 +269,8 @@ def view_engagement(request, eid):
         Test.objects.filter(engagement=eng)
         .prefetch_related('tags', 'test_type')
         .annotate(count_findings_test_all=Count('finding__id'))
-        .annotate(count_findings_test_active_verified=Count('finding__id', filter=Q(finding__active=True)))
+        .annotate(count_findings_test_active=Count('finding__id', filter=Q(finding__active=True)))
```
valentijnscholten, line 272: now this one is actually correct :-D

dojo/models.py
madchap: No, not that one. That one is actually to filter specifically `verified`.

dojo/engagement/views.py
madchap: It's not, it's actually an oversight. If so, then there is no longer a difference with verified at all.

madchap: Typed too quickly, disregard.

valentijnscholten (approved):
tears in my eyes when I woke up this morning to this revised PR

damiencarol (approved):

PR #3592, opened 2021-01-06
Send status changes from re-import to jira

Maffooch:
When re-uploading scan files, changes in status such as mitigated and reactive findings were not being updated in jira despite push all issues being set or just pushing individual imports.


valentijnscholten:
Good find / fix. Add a quick test to `test_jira_import_and_pushing_api.py` to cover this case?

Maffooch:
Sure. Coming soon.

madchap (approved):

valentijnscholten (approved):

PR #3593, opened 2021-01-06
UI: Make Endpoint status reflect Finding status after close/reopen

### LINKED ISSUE #3530: Endpoint Statuses are not updated when a finding is closed.
**Slack us first!**
The easiest and fastest way to help you is via Slack. There's a free and easy signup to join our #defectdojo channel in the OWASP Slack workspace: [Get Access.](https://owasp-slack.herokuapp.com/)
If you're confident you've found a bug, or are allergic to Slack, you can submit an issue anyway.

**Be informative**
Please enter as much information as possible, otherwise we can't provide support. If possible upgrade to the latest release or dev version and try again.

**Bug description**
Endpoint statuses are not updated when a finding is closed.

**Steps to reproduce**
1. Upload sample scan with endpoints.
2. Close finding.
3. Endpoint statuses remain active.

**Expected behavior**
Endpoint statuses should follow finding statuses.

- [ ] Kubernetes
- [ X ] Docker

**Environment information**

**Screenshots** (optional)

**Console logs** (optional)

Maffooch:
Fixes #3530 

Endpoint Status objects were not being updated when a finding was closed or reopened.


valentijnscholten:
I saw this behaviour when writing the unit tests a while back: https://github.com/DefectDojo/django-DefectDojo/blob/70c649cd7799ae9269a8d71e992ebc91eaac41cd/dojo/unittests/test_import_reimport.py#L379

Can we update the unit tests to capture / check on the new behaviour?

Maffooch:
Sure. Coming soon.

Maffooch:
Since the functionality for closing and reaping findings is specific to the UI then selenium tests needed to be made. The modification to the endpoint view was to remove a javascript error. The old way of doing tags was still used in that template and it simple looked better switching it over to the one all endpoints used.

valentijnscholten (changes requested):
Shouldnt'we also update serializers.py? Now there is a difference between UI import/reimport and API import/reimport.

valentijnscholten:
Hmm just noticed that this PR doesn't affect import/reimport at all, but only manual actions in the UI. The import/reimport already look good, except for that possible bug I linked to above. Wdyt about that? Could be for another PR.

In this PR it should be considered what to do with the (bulk)editing of findings. Both `edit_finding` and `finding_bulk_update_all` can also be used to "close" a finding (making active=False). Should in these case also the endpoints be marked at mitigated? The existing code does not yet set the mitigated / is_Mitigated in these cases on the finding itself.

Maffooch:
This is PR is only scoped for the close and reopen aspect of the UI. The other two things you mentioned will be in future PR's. 
I'm still trying to figure out why the integration tests won't pass anymore. They passed on previous commits, then nothing was changed, and now it won't pass. Rerunning the test makes it fail at different places for totally unrelated issues.

valentijnscholten:
weird failures yes. but they run fine on all other PRs so it's not something globally wrong with the integration tests I think.

valentijnscholten (approved):

devGregA (approved):

PR #3594, opened 2021-01-06
Reopen or mitigate existing finding in case of duplicate import with …

reinier-vegter:
We had an issue with imports that contain issues already in defectdojo, so that the new findings are marked as duplicate and closed.

Makes sense, but in the event of a new (duplicate) finding being mitigated by the import parser (for example when trigated in the source system), the original finding in defectdojo does not get marked mitigated.
Code to reopen was already in place, so I added the mitigation part (the other way around) to it, and changed  function names accordingly.

I hope it helps others too.

As I don't have a fully operational python dev setup, I was not able to run tests. Apologies for that.


valentijnscholten:
Is there a way to import findings as mitigated?
Either way, could you update the testcases related to imports / deduplication?
Not the best of tests, but at least we capture the behaviour: https://github.com/DefectDojo/django-DefectDojo/blob/329b047bda4c5d7ebc49f1a96ae9f3064092dae1/dojo/unittests/test_utils_deduplication_reopen.py

reinier-vegter:
`set_duplicate` calls `if is_duplicate_reopen(new_finding, existing_finding)`,
which returns `True` if (among other conditions) `not existing_finding.false_p`.

In case of existing finding being a false positive, it will get a `False`, which _prevents_ the existing finding from taking properties from the new finding (in `set_duplicate_reopen`).
In other words, it doesn't reopen.


Now, the test `test_false_positive_reopen` calls
`set_duplicate(self.finding_b, self.finding_a)`,  where `a` is a FP.

Later on it does 
```
self.assertTrue(self.finding_a.false_p)
self.assertTrue(self.finding_a.is_Mitigated)
self.assertFalse(self.finding_a.active)
self.assertFalse(self.finding_a.verified)
```
In other words, test `test_false_positive_reopen` is actually testing/asserting that a FP issue will _not_ be reopened while it should :p
Correct me if I'm wrong, but it seems that it passes because both the test and the feature are broken.

---
I'll try to improve it a bit, but as said, lacking a proper running dev environment (sadly this is a side thing for me), let's see where it ends.

BTW, if you have a pointer to some simple way (docker?) to run these tests without a lot of setting up, that'd be beneficial.

valentijnscholten:
Why would you want to reopen a false positive? Especially in ci/cd scenario's and automated tests it's common to have the same finding reported every time, so you want to mark those as false positives and leave them like that.

reinier-vegter:
> Why would you want to reopen a false positive? Especially in ci/cd scenario's and automated tests it's common to have the same finding reported every time, so you want to mark those as false positives and leave them like that.

I don't, I was just confused for a minute by the name of the test. See edited comment: https://github.com/DefectDojo/django-DefectDojo/pull/3594#issuecomment-756146139 .
Also had a small question in there.

I renamed the current tests and added another now.

reinier-vegter:
> BTW, if you have a pointer to some simple way (docker?) to run these tests without a lot of setting up, that'd be beneficial.

Ah, just found `.github/workflows/unit-tests.yml`. Should do next time.

valentijnscholten:
quickest way to run unit tests might be the one in https://github.com/DefectDojo/django-DefectDojo/blob/dev/doc/guide_to_parser_writing.md

Could you look at resolving the flake8 violations?

reinier-vegter:
Hi @valentijnscholten ; I think this merge still makes sense, so I resolved the conflicts.

Also discovered an issue with the recursive resetting of duplicates which caused recursive saving of the existing finding.
That didn't cause issues earlier (except waste of I/O I guess) since it didn't close findings, only reopened.
My change makes it more flexible (adopts mitigated/closed status if report/parser says so), which had the side-effect of the original finding re-opening and closing again, since it adopts status from multiple old duplicates (instead of only the new one).
Fixed that by adding a default flag so save the existing finding; turned `False` for the recursive re-linking of duplicates.

valentijnscholten:
I have reread all the comments and I think the main goal of this PR is to "Close original findings on import of mitigated duplicate"?

reinier-vegter:
That's correct; instead of only reopening it just adopts the status from the import (as long as an importer supports that of course). This also implies becoming inactive if the imported findings are not marked active; perhaps that something you want to consider if it's desirable or not.

Added a last commit to ensure inactive status; The veracode importer is my example case. The importer marks findings correctly not active (in case of mitigated), but I think somewhere else it's turned into active, probably since the import-call through the API has that flag (so on the import level instead of the finding level) (`true` in my usecase).

valentijnscholten:
That latest commits feels very dirty, why can't we rely on the field in the new finding? Also I think `is_Mitigated` is never set on new findings, except for the veracode parser. I think if we want to allow parsers to set this field, we should handle it on a higher lever, for example in the import/reimport. By putting it deep down here it can lead to surprising results.

reinier-vegter:
I agree it's better to fix it on the right level.
Tried to track where this is happening; seems like `ImportScanResource.()` > `item.active = bundle.data['active']`.
Is that correct, and is it the only place ?

valentijnscholten:
I think you need to look at importer.py and reimporter.py. api.py is deprecated.

valentijnscholten (changes requested):
as per comment

reinier-vegter:
Hi @valentijnscholten , just had some time to have a look at this.
I moved the reopening/mitigation logic to the importers as you asked.

Then again, in the original code, the reopening logic is also present in `utils.py` (`set_duplicate()`), which I just extended earlier.

So, given that, perhaps it's an idea to just drop the reopening (and the additional mitigation) of existing findings from `set_duplicate()` altogether ? I'm lacking too many detailed knowledge of the code on this, but perhaps you can shine your light on it and see if the functionality is now fully provided in the (re)importer code.

Kind regards

PR #3598, opened 2021-01-08
feat(alerts): automated cleanup of alerts per user

alles-klar:
Some users receives a lot of alerts but never reads or deletes them. This feature deletes the oldest x (default 999) alerts per user. The feature can be disabled by set DD_MAX_ALERTS_PER_USER to -1.

On behalf of DB Systel GmbH

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -154,6 +154,8 @@ env = environ.Env(
     DD_LOGGING_HANDLER=(str, 'console'),
     DD_ALERT_REFRESH=(bool, True),
     DD_DISABLE_ALERT_COUNTER=(bool, False),
+    # to disable deleting alerts per user set value to -1
+    DD_MAX_ALERTS_PER_USER=(int, 999),
     DD_TAG_PREFETCHING=(bool, True),
 
     # when enabled in sytem settings,  every minute a job run to delete excess duplicates
@@ -235,6 +237,7 @@ TEST_RUNNER = env('DD_TEST_RUNNER')
 
 ALERT_REFRESH = env('DD_ALERT_REFRESH')
 DISABLE_ALERT_COUNTER = env("DD_DISABLE_ALERT_COUNTER")
+MAX_ALERTS_PER_USER = env("DD_MAX_ALERTS_PER_USER")
 
 TAG_PREFETCHING = env('DD_TAG_PREFETCHING')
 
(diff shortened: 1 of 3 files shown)


dojo/settings/settings.dist.py
```diff
@@ -698,6 +701,10 @@ def generate_url(scheme, double_slashes, user, password, host, port, path, param
         'schedule': timedelta(hours=1),
         'args': [timedelta(hours=1)]
     },
+    'cleanup-alerts': {
+        'task': 'dojo.tasks.async_dupe_delete',
```
valentijnscholten, line 705: wrong task :-)

valentijnscholten:
Thanks, useful. I do think future cleanup tasks should not be seperate tasks, some in settings.py, some in system settings ui. I think we should create a "cleanup" feature and ui where people can check/uncheck some things to cleanup and configure some thresholds.

dojo/settings/settings.dist.py
alles-klar: good catch!

damiencarol (approved):
LGTM

valentijnscholten (approved):

dojo/tasks.py
```diff
@@ -75,6 +75,24 @@ def add_alerts(self, runinterval):
...
+        max_alerts_per_user = settings.MAX_ALERTS_PER_USER
+    except System_Settings.DoesNotExist:
+        max_alerts_per_user = -1
+
+    if max_alerts_per_user > -1:
+        total_deleted_count = 0
+        logger.info('start deleting oldest alerts if a user has more than %s alerts', max_alerts_per_user)
+        users = User.objects.all()
+        for user in users:
+            alerts_to_delete = Alerts.objects.filter(user_id=user.id).order_by('-created')[max_alerts_per_user:].values_list("id", flat=True)
```
madchap, line 90: Would not this actually return the top most recent `x` alerts, due to the `-` in front of `created` in the `order_by` sorting descending?

valentijnscholten: Yes, but it start deleting only after the first bunch due to the [:]

madchap (approved):

dojo/tasks.py
damiencarol: [:] syntax is tricky sometime :D but the variable name is very clear. (and the end of the instruction `.values_list("id", flat=True)` clearly show what is returned)

PR #3600, opened 2021-01-08
feat(clair parser): parse component name and version

alles-klar:
On behalf of DB Systel GmbH

### DIFF
--- a/dojo/tools/clair/parser.py
+++ b/dojo/tools/clair/parser.py
@@ -54,6 +54,8 @@ def get_item(item_node, test):
                       str(item_node['vulnerability']),
                       mitigation=item_node['fixedby'],
                       references=item_node['link'],
+                      component_name=item_node['featurename'],
+                      component_version=item_node['featureversion'],
                       cve=item_node['vulnerability'],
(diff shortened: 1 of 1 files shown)


damiencarol (approved):
LGTM

valentijnscholten:
are these fields always present, or do they need a safer way to read them from the dict?

damiencarol:
@valentijnscholten @madchap As I understand how Clair works, it's how the tool works. First it detect the component+version, second, it check with vuln data to produce a finding. You can't have a finding without a component.
For me this pull request is a must have. Users of Clair will see as normal to see component in the UI.

madchap:
Yea, seems it's ok.

https://github.com/arminc/clair-scanner/blob/master/clair.go#L18-L20 and https://github.com/arminc/clair-scanner/blob/master/clair.go#L93

madchap (approved):

valentijnscholten (approved):
Thanks. Would be nice if we could update more parsers to set these fields.

damiencarol:
I will take a look @valentijnscholten

PR #3601, opened 2021-01-08
fix(qualys-parser): map qid to vuln_id_from_tool

alles-klar:
The qid in Qualys is not unique.

On behalf of DB Systel GmbH


damiencarol (approved):
Efficient patch!

damiencarol (approved):

madchap (approved):
I was looking around https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/settings/settings.dist.py#L789 and I see nothing regarding specific use of that field for Qualys, so good. I am guessing users who may have overrode dedupe settings should pay attention.

valentijnscholten (approved):

PR #3605, opened 2021-01-09
Findings list: Display jira key instead of jira bug icon

madchap:
When wanting to look at a glance or eventually take a screenshot, seeing the jira bug does not help much.

This PR reverts to the old display we used to have.

image


valentijnscholten (changes requested):
Good idea. `view_test.html` needs the same change.

damiencarol:
@valentijnscholten @madchap is it possible to "test" this UI changes? Test the template or something?

valentijnscholten (approved):

damiencarol (approved):

damiencarol:
We have selenium in some integration tests. it is possible to check if some HTML element are in the page generated.

PR #3606, opened 2021-01-10
jira webhook: fix incoming author name check

valentijnscholten:
On an incoming comment from JIRA, the webhook checks for the author of the comment. If it matches a username used in one of the JIRA instances, it assumes that the incoming comment is actually a comment that was entered as note in defect dojo and pushed to jira. Which triggers a webhook back to defect dojo. The webhook needs to be ignored, because the comment is already in defect dojo. 

This intention is correct.

However, the check was done on JIRA `displayName` instead of the JIRA `key` field. So the check only succeeded if the displayname in JIRA was the same as the username used in defect dojo. In my setup it was:

username: defect.dojo
displayname: Defect Dojo

So it never matches. The check was also doing a `if jira_username_in_dojo in comment_author_from_jira` which is not always correct.

This PR changes the matching logic to use the JIRA username field instead of the displayname, and uses an exact match. And adds a unit test.

```
[10/Jan/2021 10:47:52] DEBUG [dojo.jira_link.views:184] skipping incoming JIRA comment as the user id of the comment in JIRA (defect.dojo) matches the JIRA username in Defect Dojo (defect.dojo)
```


dojo/jira_link/views.py
```diff
@@ -358,684 +404,644 @@ def delete_jira(request, tid):
...
-#          },
-#          "displayName":"Valentijn Scholten",
-#          "active":true,
-#          "timeZone":"Europe/Amsterdam"
-#       },
-#       "created":"2020-11-11T18:55:21.425+0100",
-#       "updated":"2020-11-11T18:55:21.425+0100"
-#    }
-# }
+# example incoming requests from JIRA Server 8.13.0
```
madchap, line 408: Do we want maybe to move these example payloads away in specific files in the repos?

dojo/jira_link/views.py
```diff
@@ -125,23 +126,68 @@ def webhook(request, secret=None):
...
             if jissue.finding:
-                logger.debug('finding: %s', vars(jissue.finding))
+                # logger.debug('finding: %s', vars(jissue.finding))
                 jira_usernames = JIRA_Instance.objects.values_list('username', flat=True)
                 for jira_userid in jira_usernames:
-                    if jira_userid.lower() in commentor.lower():
-                        logger.info('skipping incoming JIRA comment as the user id of the comment mathces the JIRA user in Defect Dojo')
+                    # logger.debug('incoming username: %s jira config username: %s', commentor.lower(), jira_userid.lower())
+                    if jira_userid.lower() == commentor.lower():
+                        logger.debug('skipping incoming JIRA comment as the user id of the comment in JIRA (%s) matches the JIRA username in Defect Dojo (%s)', commentor.lower(), jira_userid.lower())
```
madchap, line 185: I am thinking maybe at times, there could be corner cases where the jira userid and the commentor name could match, and possibly, a legit comment would not make it to defectdojo?

How about making this an `logger.info`?

madchap (approved):
Good find.

dojo/jira_link/views.py
valentijnscholten: It's almost impossible to match by accident. In that case the defect.dojo user in JIRA is used to place a comment directly into JIRA?
This log line will happen *for every* note posted to dojo where a JIRA issue is attached to the finding, so it should be not logged at all or only at debug level.
If we worry about comments being missed we should improve the check and look for matching timestamps of the note in dojo and jira etc.

dojo/jira_link/views.py
mtesauro: That's an interesting point.  I was looking for some Jira examples and went to the tests.  I don't think it would have occurred to me to look in the views file TBH.

damiencarol: I agree. Should be better to move these exemple as unit tests data.
Also a link to the current JIRA documentation could be usefull https://developer.atlassian.com/server/jira/platform/webhooks/

damiencarol (changes requested):
I think big object dumps of JIRA hook should be removed from code

dojo/jira_link/views.py
valentijnscholten: if we create files we are not using it becomes confusing so that's why I didn't do that. but this PR doesn't change anything around examples, just the logic of the author name check. and some moving around of the example.

valentijnscholten: but there are already examples in the unit tests code, so I will put a link to that then

damiencarol (approved):

PR #3608, opened 2021-01-10
APIv1: disable by default

valentijnscholten:
Now that APIv1 is deprecated, it might be best to disable it by default. It can be enabled by setting `DD_LEGACY_API_V1_ENABLE` environment variable to `True`. Or by setting `LEGACY_API_V1_ENABLE` to `True` in a `local_settings.py` file.

By default all apiv1 request return a 400 message with an error:

```
BadRequest at /api/v1/findings/1234/
{"error": "{'code': 666, 'message': \"APIv1 is deprecated and may contain vulnerabilities. It's disabled by default. At your own risk it can be enabled by setting DD_LEGACY_API_V1_ENABLE to True, or by setting LEGACY_API_V1_ENABLE to True in  settings(.dist).py or local_settings.py\"}"}
```

When APIv1 is enabled again, it will work, but will also return some warnings:
```
Response Headers
{
  ...
  "deprecation": "true",
  "warning": "666 APIv1 APIv1 is deprecated and may contain vulnerabilities. It's disabled by default. At your own or your admins risk it has been enabled.",
  ...
}
```

With unit test! And existing unit tests still work using a override_settings decorator.

### DIFF
--- a/dojo/api.py
+++ b/dojo/api.py
@@ -1,8 +1,11 @@
 # see tastypie documentation at http://django-tastypie.readthedocs.org/en
 from django.core.exceptions import ImproperlyConfigured, ValidationError
+from django.http import HttpResponse
 from django.urls import resolve, get_script_prefix
+from django.utils.html import escape
 import base64
 from tastypie import fields
+from tastypie import http
 from tastypie.fields import RelatedField
 from tastypie.authentication import ApiKeyAuthentication
 from tastypie.authorization import Authorization
@@ -13,10 +16,13 @@ from tastypie.http import HttpCreated
 from tastypie.resources import ModelResource, Resource
 from tastypie.serializers import Serializer
 from tastypie.validation import FormValidation, Validation
+from tastypie.exceptions import BadRequest
 from django.urls.exceptions import Resolver404
 from django.utils import timezone
 from django.db.models import Count, Q
-
+from django.conf import settings
+from django.utils.cache import patch_cache_control, patch_vary_headers
+from django.views.decorators.csrf import csrf_exempt
 from dojo.models import Product, Engagement, Test, Finding, \
     User, ScanSettings, IPScan, Scan, Stub_Finding, Risk_Acceptance, \
     Finding_Template, Test_Type, Development_Environment, \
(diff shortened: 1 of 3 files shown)


dojo/api.py
```diff
@@ -89,7 +95,85 @@ def form_args(self, bundle):
...
 class BaseModelResource(ModelResource):
+
+    def wrap_view(self, view):
+        """
+        Wraps views to check if APIv1 is enabled
+        """
+        @csrf_exempt
+        def wrapper(request, *args, **kwargs):
+            try:
+                api_v1_deprecation_warning = 'APIv1 is deprecated and may contain vulnerabilities. It\'s disabled by default. '
```
mtesauro, line 113: If we change the message to:
APIv1 is deprecated and may contain vulnerabilities. **It is** disabled by default.

Doesn't that remove the need for lines 98 to 101 above?  I'd rather remove conjunctions (don't, can't, ...) from our messages + remove that character as accepted vs have conjunctions and use the sanitize function.  Keeping things simple :wink:   

And yes, I see the irony of me using multiple conjunctions in this comment. :shrug:

dojo/api.py
```diff
@@ -89,7 +95,85 @@ def form_args(self, bundle):
...
+                if not settings.LEGACY_API_V1_ENABLE:
+                    raise BadRequest({'code': 666, 'message': api_v1_deprecation_warning + 'At your own risk it can be enabled by setting DD_LEGACY_API_V1_ENABLE to True, or by setting LEGACY_API_V1_ENABLE to True in  settings(.dist).py or local_settings.py'})
```
madchap, line 115: :-D

dojo/settings/settings.dist.py
```diff
@@ -161,7 +161,9 @@
     # when enabled in sytem settings,  every minute a job run to delete excess duplicates
     # we limit the amount of duplicates that can be deleted in a single run of that job
     # to prevent overlapping runs of that job from occurrring
-    DD_DUPE_DELETE_MAX_PER_RUN=(int, 200)
+    DD_DUPE_DELETE_MAX_PER_RUN=(int, 200),
+    # APIv1 is depreacted and will be removed 30/06/2021
```
madchap, line 165: typo, and `will be removed on 2021-06-30` to avoid date format wars?

dojo/api.py
```diff
@@ -89,7 +95,85 @@ def form_args(self, bundle):
         return kwargs
 
 
+def sanitize(text):
```
madchap, line 98: shouldn't this be in a helper of sort?

valentijnscholten: It's just for this apiv1 thing that is going to be thrown away in june

dojo/settings/settings.dist.py
valentijnscholten: Will do

dojo/api.py
madchap: I would agree and I always think that writing with full words and no contraction is certainly better. But we are bound to have some. Maybe move it to a general helper instead though? (and put `it is` ;-) )

dojo/api.py
madchap: I would reword as `It can be enabled at your own risk by setting...`, or put a `,` after risk as it stands. Now, I didn't intend to put so many spelling comment in the review of that PR! ;-)

dojo/api.py
valentijnscholten: The sanitize is there from tasty pie itself, didn't want to change it. And there can be other errors from normal api operations. I think in general escaping is not needed for json api, but easiest is to stick what tasty pie does (the code is copied from there)
Would be bad usually. But this will be thrown away in june.

valentijnscholten: Can look at avoiding conjunctions.

madchap (approved):
Nicely done! I'd still love to see some of my comments addressed, yet this is dealing nicely with the deprecation we've had in the documentation for some time now.

dojo/api.py
madchap: ok

dojo/settings/settings.dist.py
mtesauro: I prefer the YYYY-MM-DD format myself - though it's not what most of my US neighbours use. +1 for ISO date formats

valentijnscholten: > typo, and `will be removed on 2021-06-30` to avoid date format wars?

OK. Although I'm pretty sure nobody would have read it as "day 6 of the 30th month of the year 2021" :-)

valentijnscholten:
Please note that this was meant to be a quick sunday afternoon PR. Turns out that it's not so easy to override requests in tastypie because the framework itself is wrapping it's view method in a `wrap_view` wrapper. So we can't simply call `super().wrap_view`  or `super.callback()` or `super.view()`. 
Every example that I found (stackoverflow FTW), is duplicating some code from the tastypie package itself.
Usually this would be a red flag. But in this case apiv1 is deprecated and will be removed in the next half year, so I'm good with it.

Same for those sanitize/escape things in there. They are used by the code from tastypie so wouldn't want to remove them and then see someone posting something on H1 :-D

dojo/api.py
valentijnscholten: @mtesauro let me know if it is acceptable now :-)

valentijnscholten: sanitize removed and import from tastypie

mtesauro (approved):
Approved.

PR #3629, opened 2021-01-14
fix anchore reimport, sync reimport logic API<->UI, add unit tests

valentijnscholten:
```
some parsers generate 1 finding for each vulnerable file for each vulnerability
i.e
#: title                     : sev : file_path
1: CVE-2020-1234 jquery      : 1   : /file1.jar
2: CVE-2020-1234 jquery      : 1   : /file2.jar

if we don't filter on file_path, we would find 2 existing findings
and the logic below will get confused and map all incoming findings
from the reimport on the first finding
```
for Anchore we fix this here, we may need a broader fix (and testcases)
or we may need to change the matching logic here to use the same logic
as the deduplication logic (hashcode fields)

```
     if scan_type == 'Anchore Engine Scan':
           if item.file_path:
                findings = findings.filter(file_path=item.file_path)
```

I have refactored the existing 9 testscases to be run against the APIv2 and UI.

And added an extra testcase for anchore. In the first unit test run I did not yet include the fix described above and you can see the test failing. This test is using a real world anchore report: https://github.com/DefectDojo/django-DefectDojo/pull/3629/checks?check_run_id=1707458436#step:9:40351

I fixed some other things a long the way:
- There was a diffrence in logic between the API reimport and UI import when the matching found multiple existing findins for a finding in the reimported report. This is now made in sync, UI now follows API logic
- don't run dedupe on findings that we are closing
- remove some push_to_jira=False as that is the default
- use the same way to calculate differences between new items / reactivated items / closed items in API and UI
- add some DEBUG logging to help future troubles

Question for a future PR: 
- Should we add this fix for all scanners? 
- Should we use the deduplication config in reimport instead of hardcoded field matching? 
- Should we raise an error if we detect scenario's where the reimport can't handle it correctly?
- Should we add instructions for people to recalculate their hashes if they change dedupe config?

### DIFF
--- a/dojo/test/views.py
+++ b/dojo/test/views.py
@@ -711,14 +711,14 @@ def re_import_scan_results(request, tid):
                     from titlecase import titlecase
                     item.title = titlecase(item.title)
                     if scan_type == 'Veracode Scan' or scan_type == 'Arachni Scan':
-                        finding = Finding.objects.filter(title=item.title,
+                        findings = Finding.objects.filter(title=item.title,
                                                         test__id=test.id,
                                                         severity=sev,
                                                         numerical_severity=Finding.get_numerical_severity(sev),
                                                         description=item.description)
 
                     else:
-                        finding = Finding.objects.filter(title=item.title,
+                        findings = Finding.objects.filter(title=item.title,
                                                       test__id=test.id,
                                                       severity=sev,
                                                       numerical_severity=Finding.get_numerical_severity(sev))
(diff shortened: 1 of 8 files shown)


madchap:
fyi @yilmi

dojo/test/views.py
```diff
@@ -748,17 +770,18 @@ def re_import_scan_results(request, tid):
                                 status.mitigated_time = None
                                 status.mitigated = False
                                 status.last_modified = timezone.now()
-                                status.save(push_to_jira=push_to_jira)
+                                status.save()
```
madchap, line 773: Always false anyways, so good to remove :+1:

madchap (approved):
Reviewed live with @valentijnscholten, looks good to me. Thanks.

damiencarol (approved):

PR #3632, opened 2021-01-15
Upgrade to Django 3.1

valentijnscholten:
2.0.0 is a nice moment to upgrade to Django 3.x.
3.2 was just released but has some bugs, so let's start with 3.1.7.

Modifications done (thanks @damiencarol for the help):
- [X] Update some package to version that support Django 3.x
- [x] update templates to replace `{% load static from staticfiles %}` with `{% load static %}` (fixed in #3637 )
- [x] `filters.py` reference a vendored version of `six` (fixed in #3649 )
- [x] windowname_to_id javascript function removed in django 3.1, fixed here
- [x] fields=[] on DeleteXXX forms no longer works in Django 3.x, fixed here
- [x] JSONField backport no longer needed, included in Django 3.1, fixed here

Some interesting items from the 3.x release notes:
image
image
image
image
image
image
image
image
image
image
image
image
image


damiencarol:
@valentijnscholten seems tastypie first blocker :) 
`ImportError: cannot import name 'python_2_unicode_compatible'`

damiencarol:
Starting with Django 3.0.4,
```Python
from six import python_2_unicode_compatible
```
instead of
```
from django.utils.six import python_2_unicode_compatible
```
It seems that our tastypie fork will need some work to upgrade to Django > 3.x

valentijnscholten:
don't waste time on it as tastypie will be removed soon.

damiencarol:
@valentijnscholten I think I was wrong. the only change for now is `six` that is not in Django anymore but a separate package.
So it could be only updating some package to versions compatible with Django 3.x branch.
At least for me it is worth a try.

valentijnscholten:
yes. but my proposal will be to do the upgrade as part of defect dojo 2.0 which will have some stuff removed like apiv1, setup.bash install method, some old django-tagging code, old permission system. 
so if those are causing troubles it's a waste of time to fix those.
but of course you are welcome to try . on tuesday we'll discuss 2.0 in the meeting

damiencarol:
Deal. I will continue this test to have some insight for the 2.0 discussion. 👍

damiencarol:
rebased on last `upstream/dev`

damiencarol (approved):
The final version is neat. I'm glad that we are finally able to get ride of the back-port :)

madchap (approved):

PR #3634, opened 2021-01-16
Add more unit tests for AppSpider report parser

damiencarol:
## Work
- [X] add unit test file with one vulnerability
- [X] add unit test for CWE
- [X] add unit test for null

### DIFF
--- a/dojo/tools/appspider/parser.py
+++ b/dojo/tools/appspider/parser.py
@@ -3,10 +3,10 @@ from defusedxml import ElementTree
 from dojo.models import Endpoint, Finding
 import html2text
 import urllib.parse
-import logging
 
 
 class AppSpiderXMLParser(object):
+    """Parser for Rapid7 AppSpider reports"""
     def __init__(self, filename, test):
         self.items = []
 
--- a/dojo/tools/appspider/parser.py
+++ b/dojo/tools/appspider/parser.py
@@ -1,15 +1,8 @@
-
-
-from datetime import datetime
 from xml.dom import NamespaceErr
-
 from defusedxml import ElementTree
-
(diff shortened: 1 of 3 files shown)


dojo/tools/appspider/parser.py
```diff
@@ -1,23 +1,21 @@
...
     def __init__(self, filename, test):
+        self.items = []
+
+        if filename is None:
+            return
 
         if "VulnerabilitiesSummary.xml" not in str(filename):
-            raise NamespaceErr('Please ensure that you are uploading AppSpider\'s VulnerabilitiesSummary.xml file.'
-                               'At this time it is the only file that is consumable by DefectDojo.')
+            logging.warning('Please ensure that you are uploading AppSpider\'s VulnerabilitiesSummary.xml file.'
```
valentijnscholten, line 17: are you sure this is intended behaviour? the warning will not be seen by the end user and the import will continue anyway instead of raising an exception.

damiencarol: this check is useless. it force the code to have a real file with a fixed name. it is not what we want.
we want to load data evene without a real file on the disk

valentijnscholten: ah, now I see. I thought it was checking for some XML node to be present. But checking the filename makes no sense, so the whole ìf` block can be removed I think.

damiencarol: Don't ask me twice to remove useless code 😄 !
Made the change requested.

valentijnscholten (approved):
approved assuming you'll fix the flake8 violation :-)

madchap (approved):

PR #3639, opened 2021-01-17
Bump django-auditlog to 1.0a1

damiencarol:
Fix dependency hell by upgrading only django-auditlog.

`django-auditlog` fixed `dateutil` to a specific version which has introduced a "dependency hell".

A more recent version (1.0a1) fixes that (only require `dateutil` >=2.6.0). [https://github.com/jazzband/django-auditlog/commit/33fa2490714954b189e24963e4586f86ce4f99c2](https://github.com/jazzband/django-auditlog/commit/33fa2490714954b189e24963e4586f86ce4f99c2)

```
ERROR: Cannot install -r ./requirements.txt (line 8) and python-dateutil==2.8.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested python-dateutil==2.8.1
    django-auditlog 0.4.7 depends on python-dateutil==2.6.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```


madchap (approved):

valentijnscholten:
that is an alpha version so let's not use that.

valentijnscholten:
I've been watching this lib for their imminent release that would allow us to use the new pip builder. If they don't release in the next few weeks, we'll have to make our own fork that just updates that dependency thing.
The audit log is already a performance hog so I wouldn't want to go with an alpha version.

damiencarol:
Converting to a draft, waiting for fixed version

damiencarol:
@valentijnscholten I made some research on this package before pushing this PR. This package is now managed by a collective. It's not a one-person project anymore. I saw some quality work pushed since version `0.4.7` (the version we use) but I agree I will wait before pushing again this one.

valentijnscholten:
That collective is usually used for packages that otherwise would get abandoned. So I wouldn't trust that it's a good sign, usually they are just saved from abandonment. I do see some commits around ci/cd, but nothing much around the actual audit log functionality.
Having said that, that might be a good sign as 1.0a1 is almost identical to 0.4.7 apart from dropping python2 and some small changes: https://github.com/jazzband/django-auditlog/compare/ee8a700b1b97e405638af5f9f9369651e8601583..565239180e2cc386156e80c98566df2c77f55463
I suggest we merge after releasing 1.12.0 so can get some mileage on dev before 1.13.0.

damiencarol:
@valentijnscholten that's sound like a plan

valentijnscholten (approved):

PR #3649, opened 2021-01-18
Remove usage of django.utils.six vendored library version of six

damiencarol:
`django.utils.six` is a vendored version of the package `six` that we already have as a dependency.

As it is only used 1 time in `filters.py`, this import is completely useless.

(!) This simple reference break compatibility with Django 3.x as Django project cleaned there vendoring by removing it of 3.x branch.

Reference [https://docs.djangoproject.com/en/3.0/releases/3.0/#removed-private-python-2-compatibility-apis](https://docs.djangoproject.com/en/3.0/releases/3.0/#removed-private-python-2-compatibility-apis)

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -5,7 +5,7 @@ from datetime import timedelta, datetime
 from django.apps import apps
 from auditlog.models import LogEntry
 from django.contrib.auth.models import User
-from django.utils import six
+import six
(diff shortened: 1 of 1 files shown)


valentijnscholten (approved):
nice!

madchap (approved):

PR #3657, opened 2021-01-19
Fix impact bug on Burp REST API parser and add more unit tests

damiencarol:
fix "impact can not be none" error


valentijnscholten:
I think this shows that unit tests not just need to parse the report, but also need to store the findings to make sure they are actually valid and can be stored.

damiencarol:
I totally agree. I can improve this one (I have code that do that) but I want to fix it fast.

madchap (approved):

valentijnscholten (approved):

PR #3667, opened 2021-01-20
dependency track parser: fix dedupe, set hash code fields

### LINKED ISSUE #3647: Dependency Track deduplication not working
**Bug description**
While uploading the same scan result multiple times from ZAP, SonarQube, Dependency Track etc to a CICD engagement in DefectDojo with Deduplication enabled, Dependency Track findings are not seen as duplicates.

**Steps to reproduce**
Enable Deduplication feature
Upload the same dependency track scan finding to CI/CD engagement multiple times. 

**Expected behavior**
dependency track findings should register as duplicate.

- [x] Kubernetes
- [ ] Docker

**Environment information**
 DefectDojo v. 1.9.3 ( release mode ) 

**Sample scan files**
JSON files below
[deptrack1.txt](https://github.com/DefectDojo/django-DefectDojo/files/5831290/deptrack1.txt)
[deptrack2.txt](https://github.com/DefectDojo/django-DefectDojo/files/5831291/deptrack2.txt)

**Screenshots**
deduplication

valentijnscholten:
fixes #3647

- add dedupe hash code config
- add file_path to make legacy dedupe happy
- add vuln_id_from_tool field
- make title format/value more useful
- allow .txt file uploads
- extend unit tests

### DIFF
--- a/dojo/forms.py
+++ b/dojo/forms.py
@@ -438,7 +438,7 @@ class ImportScanForm(forms.Form):
     tags = TagField(required=False, help_text="Add tags that help describe this scan.  "
                     "Choose from the list or add new tags. Press Enter key to add.")
     file = forms.FileField(widget=forms.widgets.FileInput(
-        attrs={"accept": ".xml, .csv, .nessus, .json, .html, .js, .zip, .xlsx"}),
+        attrs={"accept": ".xml, .csv, .nessus, .json, .html, .js, .zip, .xlsx, .txt"}),
         label="Choose report file",
         required=False)
 
@@ -482,7 +482,7 @@ class ReImportScanForm(forms.Form):
     tags = TagField(required=False, help_text="Modify existing tags that help describe this scan.  "
                     "Choose from the list or add new tags. Press Enter key to add.")
     file = forms.FileField(widget=forms.widgets.FileInput(
-        attrs={"accept": ".xml, .csv, .nessus, .json, .html, .js, .zip, .xlsx"}),
+        attrs={"accept": ".xml, .csv, .nessus, .json, .html, .js, .zip, .xlsx, .txt"}),
         label="Choose report file",
         required=False)
     close_old_findings = forms.BooleanField(help_text="Select if old findings get mitigated when importing.",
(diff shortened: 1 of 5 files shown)


madchap (approved):
Would this come with a note to eventually run the manage.py `dedupe` at all?

valentijnscholten:
> Would this come with a note to eventually run the manage.py `dedupe` at all?

I was thinking about a migration to do that. But might be to much for big instances. In general we need to document that a dedupe run (or hash code compute run) is needed.

dojo/forms.py
```diff
@@ -438,7 +438,7 @@ class ImportScanForm(forms.Form):
     tags = TagField(required=False, help_text="Add tags that help describe this scan.  "
                     "Choose from the list or add new tags. Press Enter key to add.")
     file = forms.FileField(widget=forms.widgets.FileInput(
-        attrs={"accept": ".xml, .csv, .nessus, .json, .html, .js, .zip, .xlsx"}),
+        attrs={"accept": ".xml, .csv, .nessus, .json, .html, .js, .zip, .xlsx, .txt"}),
```
damiencarol, line 441: could you add ".sarif" extention? 
```suggestion
        attrs={"accept": ".xml, .csv, .nessus, .json, .html, .js, .zip, .xlsx, .txt, .sarif"}),
```

dojo/forms.py
```diff
@@ -482,7 +482,7 @@ class ReImportScanForm(forms.Form):
     tags = TagField(required=False, help_text="Modify existing tags that help describe this scan.  "
                     "Choose from the list or add new tags. Press Enter key to add.")
     file = forms.FileField(widget=forms.widgets.FileInput(
-        attrs={"accept": ".xml, .csv, .nessus, .json, .html, .js, .zip, .xlsx"}),
+        attrs={"accept": ".xml, .csv, .nessus, .json, .html, .js, .zip, .xlsx, .txt"}),
```
damiencarol, line 485: could you add ".sarif" extention?
```suggestion
        attrs={"accept": ".xml, .csv, .nessus, .json, .html, .js, .zip, .xlsx, .txt, .sarif"}),
```

damiencarol (approved):
approved with one suggestion

dojo/forms.py
valentijnscholten: Not really related but ok

ptrovatelli:
I think a migration is a bit much as many probably don't use this scanner and re-computing the full db can take time. we probably should update the next release note and tell ppl that use the scanner and that use dedupe to run the full db deduplication. how about adding a label in the PRs that change the dedupe configuration so that we can flag those easily from the release drafter? does the " Requires settings change or database migration" family in the release drafter automatically find the PR tagged as "new migration"?

PR #3669, opened 2021-01-20
fix engagement styling bug

valentijnscholten:
view engagment layout div structure was broken after #3566 . not sure if this is perfect fix but at least it looks back to normal now.


madchap (approved):
The best diff I've ever seen!

Maffooch (approved):

PR #3673, opened 2021-01-21
Fix bug in redirect function call

damiencarol:
According to `dojo/utils.py`, the function `redirect` has 2 parameters.
https://github.com/DefectDojo/django-DefectDojo/blob/279debf699ed2512f0af53696df5b95bcf02597f/dojo/utils.py#L1670-L1674 
This means that these calls are bugs:
https://github.com/DefectDojo/django-DefectDojo/blob/279debf699ed2512f0af53696df5b95bcf02597f/dojo/endpoint/views.py#L445
https://github.com/DefectDojo/django-DefectDojo/blob/279debf699ed2512f0af53696df5b95bcf02597f/dojo/reports/views.py#L321

### DIFF
--- a/dojo/endpoint/views.py
+++ b/dojo/endpoint/views.py
@@ -1,28 +1,29 @@
-# # endpoints
-
 import logging
 from datetime import datetime
+
 from dateutil.relativedelta import relativedelta
 from django.contrib import messages
+from django.contrib.admin.utils import NestedObjects
 from django.contrib.auth.decorators import user_passes_test
 from django.core.exceptions import PermissionDenied
-from django.urls import reverse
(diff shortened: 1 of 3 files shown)


valentijnscholten (changes requested):
good find. but can you make PR to only change those 2 lines and not reformat all the import lines.

damiencarol:
I used `isort`. all the imports in all files are a mess. can we start to clean them?

valentijnscholten:
yes, but in a dedicated PR.

damiencarol:
@valentijnscholten fixed

valentijnscholten (approved):

Maffooch (approved):

Issue #3677, opened 2021-01-21
PicklingError: Can't pickle <class 'tagulous.models.descriptors.TagRelatedManager'>


valentijnscholten:
The `django-tagulous` library has an [issue](https://github.com/radiac/django-tagulous/issues/109) where models with a `tags` attribute can't be serialized using the `pickle` algorithm used by python/celery. 

For dedupe and other tasks we solve this by converting all models to an id before passing the value on to celery. This works OK.

For notifications we are sending more complex constructs such as lists of models, and list of models with attached manytomany models etc.

As a work around we implemented a `model_to_dict` and `list_of_models_to_dict`. These worked out fine and all testcases are passing. However there are still cases where models are not converted into dicts and the `tags` field on those model causes the PicklingError (even when the model has no actual tags, the error is caused by the attribute itself being there). This seems to happen in nested relationships such as finding->duplicate_finding or finding->original_set or other nested constructs.

As always a workaround that looks splendid at first, comes back to haunt you later. So for now we'll have to make the notifications synchronous again (don't use celery). This will make them a bit slower, especially if you have lots of `staff` users (new permission model anyone?).

This PR removes / comments out the `@dojo_async_taks` decorators on the notifications (and removes the conversion of model to dicts becaues it's not needed now).

valentijnscholten:
was fixed in 1.12.0 but we should look at making them async again, now they are slowing down the UI.

PR #3678, opened 2021-01-21
notifications: make them synchronous

valentijnscholten:
make notifications synchronous due to bug in django-tagulous, see #3677


madchap (approved):
I am actually impacted by that, finding out exactly what that was was nearly impossible for me! Thanks!

valentijnscholten:
the reason why it was so damn hard to find the cause is due to https://github.com/DefectDojo/django-DefectDojo/issues/3679

damiencarol (approved):

PR #3682, opened 2021-01-22
Increase alert field size to 250

madchap:
When engagement name get a little too long, the alert title size goes over 100 characters, which provokes an error upon import such as

```
fail to import scan into engagement "xxxxx/ph-blablabla-value-beep-test:1.272.1", http status: 400, resp: {"detail":"{'title': ['Ensure this value has at most 100 characters (it has 111).']}"}]
```


valentijnscholten:
I already increased it locally in the past and found out 250 is about a good value. I also added a [:250] where the title for the alerts is being set to that it just truncates and doesn't prevent the import from completing.

valentijnscholten:
where did you get that error from? because in dev it is already truncating to 100:

image

that should be adjusted then to 250. 

and if this was causing an error where the title was over 100 there might another place where Alert objects are created?

valentijnscholten (changes requested):
^^

valentijnscholten:
I see where you got the error, there is a second place where Alerts are being created and it truncates the `url` fields instead of `title`.

image

valentijnscholten:
Can you remove the incorrect truncation of url?

valentijnscholten (approved):

dojo/notifications/helper.py
```diff
@@ -331,8 +331,8 @@ def log_alert(e, notification_type=None, *args, **kwargs):
     for user in users:
         alert = Alerts(
             user_id=user,
-            url=kwargs.get('url', reverse('alerts'))[:100],
-            title=kwargs.get('title', 'Notification issue'),
+            url=kwargs.get('url', reverse('alerts')),
```
damiencarol, line 334: this is a good catch!

damiencarol (approved):

PR #3684, opened 2021-01-22
Fix tag migration for legacy products

### LINKED ISSUE #3683: upgrade failing to 1.11.0 or higher when prod_type=0
In 1.11.0 we replaced our tagging library (#3364 

The migration for that fails when people have products for which the `prod_type` field is 0 (`prod_type_id`). These are probably old products from when the prod_type field wasn't mandatory yet.

Example report from Slack: https://owasp.slack.com/archives/C2P5BA8MN/p1610644710030600

```
  File "/usr/local/lib/python3.6/site-packages/django/db/backends/mysql/operations.py", line 181, in validate_autopk_value
    raise ValueError('The database backend does not accept 0 as a '
ValueError: The database backend does not accept 0 as a value for AutoField.
```

Product data
```
{
  "count": 1,
  "next": null,
  "previous": null,
  "results": [
    {
      "id": 113,
      "findings_count": 0,
      "findings_list": [],
      "tags": [
        "not supported by ft techn",
        "none"
      ],
      "product_meta": [],
      "name": "abc",
      "description": "Circulation Auditing.\r\n\r\nWe deliver industry-agreed standards for media brand measurement across print, digital and events. We also verify data, processes and good practice to industry-agreed standards.",
      "created": null,
      "prod_numeric_grade": null,
      "business_criticality": null,
      "platform": null,
      "lifecycle": "production",
      "origin": null,
      "user_records": null,
      "revenue": null,
      "external_audience": false,
      "internet_accessible": false,
      "product_manager": null,
      "technical_contact": null,
      "team_manager": null,
      "prod_type": 0,
      "authorized_users": [],
      "regulations": []
    }
  ]
}
```

I have reproduced this by trying to save a product with prod_type 0 and I am getting the same error.

valentijnscholten:
make tagulous migration work with (legacy) products without a product type. (fixes #3683)

if a product is found without a product type (prod_type==0), we create a `_tag migration lost and found` product type and attach the product to it.

this a rare case where updating an existing migration is harmless as we're not changing the database schema, just the data migration. this will only affect people for which the migration was failing.

for those people, you need to comment out the following part in `dojo/db_migrations/0066_django_tagulous.py`:

``` 
    #     migrations.AddField(
    #         model_name='app_analysis',
    #         name='tags_from_django_tagging',
    #         field=models.TextField(blank=True, editable=False, help_text='Temporary archive with tags from the previous tagging library we used'),
    #     ),
    #     migrations.AddField(
    #         model_name='endpoint',
    #         name='tags_from_django_tagging',
    #         field=models.TextField(blank=True, editable=False, help_text='Temporary archive with tags from the previous tagging library we used'),
    #     ),
    #     migrations.AddField(
    #         model_name='engagement',
    #         name='tags_from_django_tagging',
    #         field=models.TextField(blank=True, editable=False, help_text='Temporary archive with tags from the previous tagging library we used'),
    #     ),
    #     migrations.AddField(
    #         model_name='finding',
    #         name='tags_from_django_tagging',
    #         field=models.TextField(blank=True, editable=False, help_text='Temporary archive with tags from the previous tagging library we used'),
    #     ),
    #     migrations.AddField(
    #         model_name='objects',
    #         name='tags_from_django_tagging',
    #         field=models.TextField(blank=True, editable=False, help_text='Temporary archive with tags from the previous tagging library we used'),
    #     ),
    #     migrations.AddField(
    #         model_name='test',
    #         name='tags_from_django_tagging',
    #         field=models.TextField(blank=True, editable=False, help_text='Temporary archive with tags from the previous tagging library we used'),
    #     ),
    #     migrations.AddField(
    #         model_name='product',
    #         name='tags_from_django_tagging',
    #         field=models.TextField(blank=True, editable=False, help_text='Temporary archive with tags from the previous tagging library we used'),
    #     ),
    #     migrations.AddField(
    #         model_name='finding_template',
    #         name='tags_from_django_tagging',
    #         field=models.TextField(blank=True, editable=False, help_text='Temporary archive with tags from the previous tagging library we used'),
    #     ),

        migrations.RunPython(copy_existing_tags_to_tags_from_django_tagging_field, migrations.RunPython.noop),

```

So basically everything before the first RunPython command.

### DIFF
--- a/dojo/db_migrations/0066_django_tagulous.py
+++ b/dojo/db_migrations/0066_django_tagulous.py
@@ -36,9 +36,18 @@ class Migration(migrations.Migration):
                     #     obj.save(dedupe_option=False, rules_option=False, issue_updater_option=False, push_to_jira=False)
                     # else:
                     try:
+                        if hasattr(obj, 'prod_type_id') and obj.prod_type_id == 0:
+                            logger.warning('product found without product type (prod_type==0), changing to: "_tag migration lost and found" product type')
+                            prod_type_lost_and_found, created = Product_Type.objects.get_or_create(name='_tag migration lost and found')
+                            obj.prod_type = prod_type_lost_and_found
+                            obj.save()
+                            logger.warning('product type succesfully changed to %i', prod_type_lost_and_found.id)
+
                         obj.save()
                     except Exception as e:
                         logger.error('Error saving old existing django-tagging tags to new string field')
+                        logger.error('Known errors are products with prod_type equal to 0. ')
+                        logger.error('Run UPDATE dojo_product set prod_type=<valid_id> to fix this problem')
                         logger.error('Details of object:')
                         logger.error(vars(obj))
                         logger.error('Model to dict:')
--- a/dojo/db_migrations/0066_django_tagulous.py
+++ b/dojo/db_migrations/0066_django_tagulous.py
@@ -3,7 +3,7 @@
 from django.db import migrations, models
 # import django.db.models.deletion
 from tagging.registry import register as tag_register
-
+from django.forms.models import model_to_dict
 import tagulous.models.fields
 import tagulous.models.models
 import logging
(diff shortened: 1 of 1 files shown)


madchap:
Do you want to start an "Upgrade notes 1.12.0" in https://github.com/DefectDojo/Documentation ? Seems like this could be a good candidate to start things off.

valentijnscholten:
I'd rather create an UPGRADING.md

madchap (approved):
Good for me, just need to think about the proper upgrade notes.

valentijnscholten:
This PR is more or less meant to have no need for upgrade notes as it will fix a problem with the migration for a corner case.

damiencarol (approved):

PR #3687, opened 2021-01-22
Helm: Chart improvements for running on GKE

jalseth:
This PR adds some improvements for those running on GKE, such as the ability to use the GKE ingress controller which requires the `type` of the `Service` to be set to `NodePort`.


madchap:
fyi @ccojocar

madchap (dismissed):
Thanks for the PR, looks ok to me. The proper default values ensure there should be no bad surprise.

Could you please bump up the Chart version along with this PR?

jalseth:
@madchap done

alles-klar (approved):

valentijnscholten (approved):

PR #3689, opened 2021-01-23
Dynamic parser infrastructure (part 1)

damiencarol:
This PR is the first step to a more generic parser infrastructure.

Goals:
* remove duplicated lists with parser key
* add dynamically the parsers by following Factory pattern
* clean a lot of things (unit tests, how parser save data, move some constant in one file => factory.py)

The final goal is to have a factory that load automatically a parser defined in a sub module and register it in the factory.

This first PR focus on cleaning factory pattern and make future modifications easier


madchap (approved):
Pulled it down for a manual test drive, all looks OK so far :+1: Well done and thanks for tackling these things!

damiencarol:
Made a rebase on last upstream/dev.
@valentijnscholten if you can take a look, it the right moment :D

Dockerfile.nginx
```diff
@@ -54,15 +54,14 @@ RUN pip3 install \
 	-r ./requirements.txt
 
 COPY components/ ./components/
+RUN \
+  cd components && \
+  yarn
+
 COPY manage.py ./
```
valentijnscholten, line 61: was there a reason to move this part up? not sure, but it might actually help with better caching of docker layers

valentijnscholten (approved):
good improvement, let's merge. we can always iterate if we see some things later on. git doesn't seem to be able to track all renamed/moved files, but for testcases that's not too bad.

Dockerfile.nginx
damiencarol: Yes. it help a lot. because our npm dependencies doesn't change at all. So to have them before the copy of Dojo files let the cache use this layer. It take few minutes to build it.

valentijnscholten: thanks. I had this on my list for Dockerfile optimizations.

PR #3690, opened 2021-01-23
files upload: remove from filters

valentijnscholten:
the recent generic file upload fields were not excluded from filters, so every page with a filter was loading metadata for all uploaded files in the database

image


mtesauro (approved):
Approved

Maffooch (approved):

Issue #3653, opened 2021-01-19
Creating a project without jira info show warning


madchap:
**Bug description**
While creating a project on an instance that has system level JIRA enabled, creating the new product fails on the first try.

![image](https://user-images.githubusercontent.com/5468769/105006234-a94c0f00-5a36-11eb-92d9-618bd67c6987.png)

Something has to be specified for the project to be created.

However, as you're back on the edit product page, just saving it works.

Trying to reproduce it, subsequent attempts were not successful in reproducing this bug...

**Steps to reproduce**

This error is not always reproducible :man_shrugging: 

1. Enable JIRA system-wide
2. Go create a product
3. Do not enter anything for JIRA

**Expected behavior**
A product can be created without having to specify a JIRA setup.

**Deployment method** *(select with an `X`)*
- [x] Kubernetes
- [ ] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
Latest `dev`.

madchap:
Actually, creating really a new project triggers the issue. No server side error, however.

valentijnscholten:
I can't reproduce it. It's also covered by an integration tests, which succeeds.

valentijnscholten:
what do the logs say?
![image](https://user-images.githubusercontent.com/4426050/105079702-1213a580-5a90-11eb-9d0a-e50fff2e08c1.png)

madchap:
Not much yea

```
uwsgi_1         | [23/Jan/2021 09:22:59] DEBUG [dojo.forms:2113] validating jira project form
uwsgi_1         | [23/Jan/2021 09:22:59] DEBUG [dojo.jira_link.helper:1045] <ul class="errorlist"><li>__all__<ul class="errorlist nonfield"><li>Cannot save JIRA_Project without JIRA_Instance</li></ul></li></ul>
```

madchap:
I rebuilt on latest dev, the first product to be added worked, then I tried again, and it failed with the above.

valentijnscholten:
can you provide some more logs? and info on what you are putting in the fields on the form? this is really fragmented that doesn't provide a complete picture. It could be related to the "has_changed" function of the form, what do the logs say about this:
![image](https://user-images.githubusercontent.com/4426050/105575602-1b8d5e00-5d6d-11eb-9cdd-6d3190dcea71.png)

valentijnscholten:
[...]

PR #3692, opened 2021-01-23
add product via prod_type: use normal add product logic

valentijnscholten:
while investigating #3653 I noticed that adding a product via the product type view is using different logic to initialize the forms 
no jira, no github, no checks, etc.
we should reuse the add product functionality for adding a new product from the side menu.

also adds some logging to help with #3653


madchap (approved):
Just pulled it down, error is gone :-) good job, as usual. Tests looks good too.

damiencarol (approved):
it's a good patch that reduce code. 👍

PR #3696, opened 2021-01-23
Fix date format on Fortify parser

Maffooch:
The date parser on the Fortify parser was not totally accurate because of the punctuation used within the report. The `dateutil` parser would receive an input like `Oct31,2020,` and it would record the date in dojo as October 21, 2021. Instead, the parser will get an input of `Oct.31.2020` and the date will be recorded on the finding correctly.

### DIFF
--- a/dojo/tools/fortify/parser.py
+++ b/dojo/tools/fortify/parser.py
@@ -21,9 +21,9 @@ class FortifyXMLParser(object):
         # Get Date
         date_string = root.getchildren()[5].getchildren()[1].getchildren()[2].text
         date_list = date_string.split()[1:4]
-        date_act = "".join(date_list)
+        date_list = [item.replace(',', '') for item in date_list]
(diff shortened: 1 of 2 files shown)


valentijnscholten:
any chance for a `self.assertEqual(bla bla)` in the unit test?

damiencarol:
Could you provide a unit tests that show the bug ?

Maffooch:
@damiencarol I don't understand what you're asking. Like have a test that shows it's the date is not being recorded correctly? That would involve not actually fixing anything.

madchap (approved):

damiencarol:
@Maffooch sometimes it's usefull to create a unit tests in the first commits just to show the bug (like I made for this one #3585) before adding more commits to fix it.
But I saw you added unit tests so 👍

damiencarol (approved):

valentijnscholten (approved):

PR #3698, opened 2021-01-23
Remove unused setup.py

madchap:
As discussed in the last maintainer meeting, this file seems to be superfluous. We're not building any `defectdojo`  package either.


damiencarol (approved):

valentijnscholten:
`setup.py` is still used by the setup.bash install method, I thought we decided that should all be removed together in one ago?

damiencarol:
I think you're right @valentijnscholten , this script is still used for legacy-setup.bash

valentijnscholten:
For this usecase we may even want to keep maintaining a `setup.py`: https://owasp.slack.com/archives/C014H3ZV9U6/p1613934621008200 (somebody installing defect dojo from git via pip to reuse our code in some way)

damiencarol:
@valentijnscholten to make it work, we need that our setup.py read requirements.txt add add dependencies based on it.

madchap:
Reviving this old PR discussion. setup.bash is now EOL. Do we feel it's something to flag for 2.0.0?

mtesauro:
FWIW, godojo doesn't use setup.py - https://github.com/DefectDojo/godojo/search?q=setup.py

damiencarol:
@madchap this PR need a rebase.

damiencarol (approved):

valentijnscholten (approved):

Issue #3613, opened 2021-01-11
proper fix needed for duplicate endpoints / multipleobjects returned


valentijnscholten:
As described in #3612 multiple endpoint objects can be returned during re_import. This was somewhat fixed in #3531 But only for UI reimports and not for the API. And it would be better to find out / fix the issue at the core, i.e. more db constraints at the EndPoint model and/or better create/insert logic.

kiblik:
This was fixed in #4473. But please double-check.

kiblik:
Can we close this issue?

Issue #3612, opened 2021-01-11
reimport: Error in parser: get() returned more than one Endpoint -- it returned 2!


BhaskiBuzz:
**Bug description**
While trying to upload webinspect xml report, I am getting error "Error in parser: get() returned more than one Endpoint -- it returned 2!"
**Steps to reproduce**
1. Create new test case with test type as Microfocus Webinspect Scan
2. upload the xml report. 

**Deployment method** *(select with an `X`)*
- [ ] Docker

**Environment information**
linux 64 bit
**Console logs** (optional)
Traceback (most recent call last):
  File "./dojo/test/views.py", line 587, in re_import_scan_results
    parser = import_parser_factory(file, test, active, verified)
  File "./dojo/tools/factory.py", line 232, in import_parser_factory
    parser = MicrofocusWebinspectXMLParser(file, test)
  File "./dojo/tools/microfocus_webinspect/parser.py", line 107, in __init__
    self.process_endpoints(finding, host)
  File "./dojo/tools/microfocus_webinspect/parser.py", line 136, in process_endpoints
    path=path
  File "/usr/local/lib/python3.7/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 412, in get
    (self.model._meta.object_name, num)
dojo.models.Endpoint.MultipleObjectsReturned: get() returned more than one Endpoint -- it returned 2!
Error in parser: get() returned more than one Endpoint -- it returned 2!
[pid: 1|app: 0|req: 29/270] 127.0.0.1 () {58 vars in 1466 bytes} [Mon Jan 11 06:58:05 2021] POST /test/216/re_import_scan_results => generated 0 bytes in 7124 msecs (HTTP/1.1 302) 8 headers in 467 bytes (17 switches on core 0)

damiencarol:
Hello @BhaskiBuzz , thanks for reporting this one. Are you able to provide the XML (or an anonymized version) that produced this is the error? it helps a lot to debug and/or understand what's going on.

damiencarol:
@BhaskiBuzz do you have access to the DB? could you check that there is no 2 endpoint with the same protocol+host+query+fragment+path ? it could come from the code not managing well when we have 2 endpoints with the same protocol+host+query+fragment+path

valentijnscholten:
[...]

PR #3700, opened 2021-01-23
Multiple Endpoint object query fix

Maffooch:
This quick fix is an extension of #3612 such that a workaround is applied to import and re-import in both the UI and the API. Though this is not a "proper" fix as desired by #3613, it will get the job done in the meantime.


mtesauro (approved):
Approved

valentijnscholten:
Do we have some example data dumps maybe of instances where we have duplicate entries? Ideally we do the proper fix at some point :-)

Maffooch:
No actually. I've never been able to reproduce. From what I've gathered so far, this only happens in the the event of a race condition like situation such as when two operations use the `get_or_create` method at the exact same time and both end up creating the object. Once the third time comes around, it will then find two objects.

valentijnscholten:
Either way eventually there should be some kind of database constraint to prevent this from happening.

image


https://stackoverflow.com/questions/6586552/django-how-to-do-get-or-create-in-a-threadsafe-way

Maffooch:
Eventually yes.

devGregA (approved):

Issue #3655, opened 2021-01-19
Nessus CSV parser 500 error - , line 84 endpoint.Host += ":" + var


harishgowdahk:
Whevener CSV file is uploaded I am getting 500 error in the browser  and below is the log file 
ches on core 0)
Internal Server Error: /engagement/3/import_scan_results
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.5/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "./dojo/engagement/views.py", line 558, in import_scan_results
    parser = import_parser_factory(file, t, active, verified)
  File "./dojo/tools/factory.py", line 87, in import_parser_factory
    parser = NessusCSVParser(file, test)
  File "./dojo/tools/nessus/parser.py", line 84, in __init__
    endpoint.host += ":" + var
AttributeError: 'NoneType' object has no attribute 'Host'
[sample.zip](https://github.com/DefectDojo/django-DefectDojo/files/5836290/sample.zip)

harishgowdahk:
We can fix above error by changing elif heading[i] == "Host": to elif heading[i] == "IP Address":

But after fixing also results are coming empty without any error . Not sure why

madchap:
Can you please provide a sample file to reproduce the problem? Thanks.

damiencarol:
@harishgowdahk  I'm working on your issue I can reproduce it in a unit test. https://github.com/DefectDojo/django-DefectDojo/pull/3710

damiencarol:
funny... the tool doesn't generate valid CVSS v3 vectors...
```
ERROR: test_parse_some_findings_csv (dojo.unittests.test_nessus_parser.TestNessusParser)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/damien/dd/dojo/unittests/test_nessus_parser.py", line 35, in test_parse_some_findings_csv
[...]

PR #3710, opened 2021-01-25
Fix IP address/host decoding in Nessus CSV parser (#3655)

damiencarol:
Work:
 - [x] add more unit tests to show bug #3655 
 - [x] add test that check findings attributes
 - [x] add more findings attributes (CVSSv3, component name+version)
 - [x] git ride of pandas 

Many unit tests where added with reports generated from Nessus Pro 8.13.1 (257)


valentijnscholten (approved):
Thanks. Some suggestions:
- Update the title so it reflects that the parser itself also got updated/fixed
- Update the description to mention you generated some real reports and used it for the tests and parser updates.

damiencarol:
Added the Haso Code deduplication algorythm as discussed with @valentijnscholten IRL (and tested it also :D )

valentijnscholten:
if the endpoints are needed for the hash codes, might be better to have dedicated PR for that. I think switching to the hash code algorithm might impact hoe deduplication works with findings with similar endpoints or overlapping endpoints.
also I think if you have endpoints on the findings, you don't need the file_path, but not sure.

damiencarol:
@valentijnscholten tested it. I removed the file_path attributes because as you said rightfully, as I added endpoints and configured hash code algo we don't need it.
For now I configured these attributes: `'Nessus Scan': ['title', 'severity', 'cve', 'cwe', 'endpoints']`

valentijnscholten:
yes, but it changes behaviour. so better do that in a separate PR.

damiencarol:
Ok I'm splitting it.

madchap:
Good stuff - need to fix the conflict though :-)

madchap (approved):

damiencarol:
Rebased and removed pandas.

PR #3711, opened 2021-01-25
initializer: import fixtures all at once, skip watson build

valentijnscholten:
The initializer takes about 1m55s to initialize the database. Not too bad if you do this once in your lifetime. 
But it adds up on every test run on GitHub actions. And locally when you want to test a PR, it keeps you waiting.

Turns our there are two quick wins here:
- Load all fixtures in a single `loaddata` command, turns out to be a lots faster (~40s saved)
- Don't run `buildwatson`, there is no need to reindex an empty database (~4s saved)

### DIFF
--- a/docker/entrypoint-initializer.sh
+++ b/docker/entrypoint-initializer.sh
@@ -44,7 +44,9 @@ do
 done
 echo
 
+echo "Making migrations"
 python3 manage.py makemigrations dojo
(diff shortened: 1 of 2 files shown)


madchap (approved):
Good find.

damiencarol (approved):

PR #3715, opened 2021-01-26
JIRAError doesn't include message. .text will provide the content.

joebasirico:
Based on the JIRA docs there's no .message attribute on the JIRAError Exception. This was causing an error if the request failed. Changing this to .text will provide the content of the error.

https://jira.readthedocs.io/en/master/_modules/jira/exceptions.html


valentijnscholten:
Can you rebase against dev please.

joebasirico:
yep, sorry.

PR #3717, opened 2021-01-26
JIRAError doesn't include message. .text will provide the content.

joebasirico:
(resubmitting to dev) 

Based on the JIRA docs there's no .message attribute on the JIRAError Exception. This was causing an error if the request failed. Changing this to .text will provide the content of the error.

https://jira.readthedocs.io/en/master/_modules/jira/exceptions.html

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -272,7 +272,7 @@ def get_jira_connection_raw(jira_server, jira_username, jira_password):
         else:
             log_jira_generic_alert('Unknown JIRA Connection Error', e)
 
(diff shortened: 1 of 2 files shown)


valentijnscholten:
thanks, you seem to have included some semgrep changes by accident :-)
I remember we did some changes around this JIRAError recently to make it work. Not sure if we went from .text to .message or the other way around. Or that we messed up a merge and lost our earlier fix.

joebasirico:
Yea, that's the only change in this MR. The Semgrep rules also have an issue when the CWE isn't supplied in the json output, I'm looking at a nice fix for that too, but I'll just close this MR so you don't have to pull in the other mistaken changes.

PR #2241, opened 2020-04-22
Add a new field version to the test model

ccojocar:
This field can be used to store the version of the artifcat for which the test was performed (e.g. a container image tag)

Signed-off-by: Cosmin Cojocar <cosmin.cojocar@gmx.ch>

**Note: DefectDojo is now on Python3.5 and Django 2.2.x Please submit your pull requests to the 'dev' branch as the 'legacy-python2.7' branch is only for bug fixes. Any new features submitted to the legacy branch will be ignored and closed.**

When submitting a pull request, please make sure you have completed the following checklist:

- [x] Your code is flake8 compliant 
- [x] Your code is python 3.5 compliant (specific python >=3.6 syntax is currently not accepted)
- [ ] If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR.
- [x] Model changes must include the necessary migrations in the dojo/db_migrations folder.
- [ ] Add applicable tests to the unit tests.
- [x] Add the proper label to categorize your PR 

Current accepted labels for PRs:
- Import Scans (for new scanners/importers)
- enhancement
- feature
- bugfix
- maintenance (a.k.a chores)
- dependencies

cc @madchap

PR #3726, opened 2021-01-27
test.version: add missing places in UI, filters and importers

valentijnscholten:
Continuing on #2241
The version field for Test objects was not shown in similar ways as for engagements. And it wasn't possible to set it on import/reimports via the UI and also not via reimport in the API v2.


damiencarol (approved):

madchap (approved):
lgtm.

Issue #3725, opened 2021-01-27
Sonarqube import - "Error while parsing the report, did you specify the correct scan type?"


lucaverderame:
Dear all,
I'm experiencing a problem with the import of a **Sonarqube** report generate using [https://github.com/soprasteria/sonar-report.git](https://github.com/soprasteria/sonar-report.git). 

The report has been generated using the following command:

`sonar-report \
  			--sonarurl="http://172.17.0.1:9000" \
  			--sonarcomponent="DevSecOps" \
  			--project="DevSecOps Project" \
  			--sonartoken="<api_key>" \
  			--sinceleakperiod="false" \
  			--fixMissingRule="true" \
  			--noSecurityHotspot="true" \
  			--allbugs="true" > sonar.html
`
The corresponding sonar.html file has been successfully generated ![Screenshot_2021-01-27 SonarQube Vulnerability Report](https://user-images.githubusercontent.com/16080554/106012805-13f6ed80-60bc-11eb-855e-90745de26c29.png))

Now I'm stuck in the importing phase using the DefectDojo API. Using the command:

`curl -X POST "http://172.17.0.1:8081/api/v2/import-scan/" -H  "accept: application/json" -H "Authorization: Token 6fb3abcb97c8b970c840b9685ef2e41508008bc4" -H  "Content-Type: multipart/form-data" -H  "X-CSRFToken: B9oWt7gJxUNBA7H4y4UYnBeqQPs6ra5luEEDFxjpN6hlcxpDSgoe651sikKOCOnA" -F "scan_date=2021-01-27" -F "minimum_severity=Info" -F "active=true" -F "verified=true" -F "scan_type=SonarQube Scan detailed" -F "file=@sonar.html;type=text/html" -F "engagement=4" -F "close_old_findings=true" -F "push_to_jira=false"
`
I'm receiving the following error: `{"detail":"Error while parsing the report, did you specify the correct scan type ?"}`
I've also tried to 

1. change the scanner type (i.e., "SonarQube")
2. use the API gateway

Still I'm getting the same error. Can you please help me?

**Configurations**
SonarQube Version 8.6.0.39681
DefectDojo v. 1.11.1 ( release mode )

valentijnscholten:
do you see any other exceptions in the logs? might be that they are swallowed as there are some problems with some recent changes to the error handling.

valentijnscholten:
[...]

PR #3727, opened 2021-01-27
api v2: remove try-catch that swallows all exceptions

### LINKED ISSUE #3679: errors in APIv2 are swallowed and turned into 400 errors
Since #3362 when an exception occurs during import, it gets swallowed. It's not logged anywhere and only a summary is returned in the API response:

```
{
    "detail": "Can't pickle <class 'tagulous.models.descriptors.TagRelatedManager'>: attribute lookup TagRelatedManager on tagulous.models.descriptors failed"
}
```

nothing in the logs so no traceback to find the root cause of the problem. 

it's caused by this construct in the api viewset: 

```
        try:
            serializer.save(push_to_jira=push_to_jira)
        except Exception as e:
            raise ParseError(detail=e)
```

valentijnscholten:
fixes #3679

a recent PR changed the error handling of the api v2 to catch everything and wrap it in a parseerror.
this leads to all exceptions being turned into a 400 error which means:
- the exception is not logged
- the exception is not returned in the api response
- the response code is 400 instead of 500 leading to wrong conclusions by users / log parsers

I tried to contact the original author of the reason for this try-catch, but I am not getting any reponse.

This PR reverses it, which will help us and others, for example in #3725

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -8,7 +8,6 @@ from django.db import IntegrityError
 from rest_framework.permissions import DjangoModelPermissions
 from rest_framework.decorators import action
 from rest_framework.parsers import MultiPartParser
-from rest_framework.exceptions import ParseError
 from django_filters.rest_framework import DjangoFilterBackend
 from drf_yasg2 import openapi
 from drf_yasg2.utils import swagger_auto_schema, no_body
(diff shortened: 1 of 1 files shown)


valentijnscholten:
fyi @xens

damiencarol (approved):

madchap (approved):
I think the 400 may be correct sometimes, but most problematic was indeed not having the stacktrace. I am ok with this change.

valentijnscholten:
400s will stay 400s even after this change

PR #3730, opened 2021-01-28
Metrics query optimizations

danielnaab:
This PR builds on the unit tests merged in #3743 (which assert on the number of ORM queries made in the metrics endpoint), and updates the implementation to decrease the count of queries. This is aided by splitting the queries in to separate functions from the main view function, which might be a technique to use on the additional metrics views.

(This builds on original PR #3438. Since several of the changes in #3438 were addressed in #3549, this separates out the remaining changes into a new clean PR.)

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -1412,6 +1412,10 @@ class MetricsFindingFilter(FilterSet):
     start_date = DateFilter(field_name='date', label='Start Date', lookup_expr=('gt'))
     end_date = DateFilter(field_name='date', label='End Date', lookup_expr=('lt'))
     date = MetricsDateRangeFilter()
+    test__test_type = ModelMultipleChoiceFilter(
+        queryset=Test_Type.objects.all().order_by('name'),
+        label="Test Type"
+    )
     test__engagement__product__prod_type = ModelMultipleChoiceFilter(
         queryset=Product_Type.objects.all().order_by('name'),
         label="Product Type")
--- a/dojo/utils.py
+++ b/dojo/utils.py
@@ -35,7 +35,6 @@ import crum
 from celery.decorators import task
 from dojo.decorators import dojo_async_task, dojo_model_from_id, dojo_model_to_id
 
-
 logger = logging.getLogger(__name__)
 deduplicationLogger = logging.getLogger("dojo.specific-loggers.deduplication")
 
--- a/dojo/utils.py
+++ b/dojo/utils.py
@@ -645,7 +645,7 @@ def get_punchcard_data(objs, start_date, weeks, view='Finding'):
                                         .annotate(count=Count('id')) \
                                         .order_by('date')
         # return empty stuff if no findings to be statted
-        if severities_by_day.count() <= 0:
+        if len(severities_by_day) == 0:
             return None, None
 
         # day of the week numbers:
--- a/dojo/utils.py
+++ b/dojo/utils.py
@@ -35,6 +35,7 @@ import crum
 from celery.decorators import task
 from dojo.decorators import dojo_async_task, dojo_model_from_id, dojo_model_to_id
 
+
 logger = logging.getLogger(__name__)
 deduplicationLogger = logging.getLogger("dojo.specific-loggers.deduplication")
 
@@ -644,7 +645,7 @@ def get_punchcard_data(objs, start_date, weeks, view='Finding'):
                                         .annotate(count=Count('id')) \
                                         .order_by('date')
         # return empty stuff if no findings to be statted
-        if len(severities_by_day) == 0:
+        if severities_by_day.count() <= 0:
             return None, None
 
         # day of the week numbers:
--- a/dojo/metrics/views.py
+++ b/dojo/metrics/views.py
@@ -437,6 +437,34 @@ def get_accepted_in_period_details(findings):
     return accepted_in_period_details
 
 
+def get_closed_in_period_details(findings):
+    closed_in_period_counts = {"Critical": 0, "High": 0, "Medium": 0,
+                               "Low": 0, "Info": 0, "Total": 0}
+    closed_in_period_details = {}
+
+    for obj in findings:
+        closed_in_period_counts[obj.severity] += 1
+        closed_in_period_counts['Total'] += 1
+
+        if obj.test.engagement.product.name not in closed_in_period_details:
+            closed_in_period_details[obj.test.engagement.product.name] = {
+                'path': reverse('closed_findings') + '?test__engagement__product=' + str(
+                    obj.test.engagement.product.id),
+                'Critical': 0,
+                'High': 0,
+                'Medium': 0,
+                'Low': 0,
+                'Info': 0,
+                'Total': 0
+            }
+        closed_in_period_details[
+            obj.test.engagement.product.name
+        ][obj.severity] += 1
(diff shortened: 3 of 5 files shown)


dojo/filters.py
```diff
@@ -1354,12 +1358,6 @@ def __init__(self, *args, **kwargs):
             self.form.fields[
                 'test__engagement__product__prod_type'].queryset = Product_Type.objects.filter(
                 authorized_users__in=[get_current_user()])
-            self.form.fields[
-                'test'].queryset = Test.objects.filter(
-                Q(engagement__product__authorized_users__in=[get_current_user()]) |
-                Q(engagement__product__prod_type__authorized_users__in=[get_current_user()]))
-        # str() uses test_type
-        self.form.fields['test'].queryset = self.form.fields['test'].queryset.prefetch_related('test_type')
```
valentijnscholten, line 1362: I think this should stay

dojo/filters.py
```diff
@@ -1385,7 +1383,8 @@ class Meta:
                    'jira_creation',
                    'jira_change',
                    'tags_from_django_tagging',
-                   'files'
+                   'files',
+                   'test',
```
valentijnscholten, line 1387: unless you hide it haha

dojo/utils.py
```diff
@@ -328,7 +330,7 @@ def set_duplicate(new_finding, existing_finding):
         raise Exception("Can not add duplicate to itself")
     deduplicationLogger.debug('Setting new finding ' + str(new_finding.id) + ' as a duplicate of existing finding ' + str(existing_finding.id))
     if is_duplicate_reopen(new_finding, existing_finding):
-        set_duplicate_reopen_(new_finding, existing_finding)
+        set_duplicate_reopen(new_finding, existing_finding)
     new_finding.duplicate = True
     new_finding.active = False
     new_finding.verified = False
```
valentijnscholten, line 336: prefer not to change this here as it's unrelated and there's another PR open touching this code probably leading to conflicts.

dojo/utils.py
```diff
@@ -1616,7 +1608,7 @@ def set_default_notifications(sender, instance, created, **kwargs):
 
 
 @receiver(post_save, sender=Engagement)
-def engagement_post_Save(sender, instance, created, **kwargs):
+def engagement_post_Save(sender, instance, created, **_kwargs):
```
valentijnscholten, line 1611: why using _kwargs instead of kwargs what is used everywhere else?

valentijnscholten:
Thanks! I did leave some comments as there are many unrelated (cosmetic) changes that there are some conflicts with other open PRs. I don't think we have any way of verifying this is all correct without doing lots of manual tests or checking the test cases in detail. An approach we sometimes follow for refactoring is that we first create a PR with the tests and then the refactoring PR after that to prove that it didn't break anything.
Is there any possibility the testscase can be submitted in their own PR first?

dojo/utils.py
danielnaab: Since it's unused, the underscore tells the linter to ignore it.

dojo/filters.py
danielnaab: IIRC, each entry in the drop-down was triggering an extra DB hit, but that might have been addressed by one of the `select_related` or `prefetch_related` additions.

I'll revisit. We had a large number of tests, and this drop-down was originally problematic... but we reduced the number of tests in our DB.

dojo/utils.py
danielnaab: Just to clarify, I didn't rename this.... `set_duplicate_reopen_` is undefined, so this code path is currently failing.

dojo/filters.py
valentijnscholten: I think removing it might not be bad idea, but I beleive the performance hit was solved. I remember because it was a pain to find out where the root cause was :-)

danielnaab:
> Thanks! I did leave some comments as there are many unrelated (cosmetic) changes that there are some conflicts with other open PRs. I don't think we have any way of verifying this is all correct without doing lots of manual tests or checking the test cases in detail. An approach we sometimes follow for refactoring is that we first create a PR with the tests and then the refactoring PR after that to prove that it didn't break anything.
> Is there any possibility the testscase can be submitted in their own PR first?

Yes I can do that. The tests were added at the beginning of the branch, so for now you can check this for reference, comparing the original test to the HEAD on the branch:

https://github.com/DefectDojo/django-DefectDojo/compare/88c00b7535905cf030a1f8958f4c93b435d5b571..2d6dfd03541117ed87ca02adb6eaaf4473b067f7#diff-81d6a68c14530c1f33a14502d72099894bf5f813509ef610344809d389919769

valentijnscholten:
thanks, but most of that test seems to be a big piece of expected html that was actually changed by the PR? Sorry but there's too much going on in the PR. Might be that most of it is linting but it I wouldn't know until going over each and every change. Can you move the tests that you created into the same folder as the existing unit tests so we can start by looking at what tests are added?

danielnaab:
> thanks, but most of that test seems to be a big piece of expected html that was actually changed by the PR? Sorry but there's too much going on in the PR. Might be that most of it is linting but it I wouldn't know until going over each and every change. Can you move the tests that you created into the same folder as the existing unit tests so we can start by looking at what tests are added?

How would you like the new tests organized? I saw the unittests directory doesn't have UI-level tests, so there wasn't 

The two new tests are simply confirming that this PR doesn't change behavior. The more important one is `test_queries` - https://github.com/DefectDojo/django-DefectDojo/pull/3730/files#diff-81d6a68c14530c1f33a14502d72099894bf5f813509ef610344809d389919769 We can just removed the HTML assertions. It was useful during development to make sure I didn't inadvertently changed the HTML but I don't see value in it now.

danielnaab:
@valentijnscholten see #3743 for tests isolated in separate PR

dojo/filters.py
danielnaab: The branch just pushed to this PR has this added back in.

dojo/utils.py
danielnaab: linting removed from PR, I'll offer something separate for that

dojo/utils.py
```diff
@@ -344,10 +346,13 @@ def set_duplicate(new_finding, existing_finding):
...
-        return True
-    else:
-        return False
+    return (
+        (existing_finding.is_Mitigated or existing_finding.mitigated) and
+        not existing_finding.out_of_scope and
+        not existing_finding.false_p and
+        new_finding.active and
+        not new_finding.is_Mitigated
+    )
```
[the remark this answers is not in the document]
damiencarol: @danielnaab try to switch to flake8 linter as the CI pipeline check syntax iwth flake8
@valentijnscholten in VS code you can select the linter. I'm using VS code also with flake8 configured for the project and it works well (same warning in VS code as in the CI pipelines)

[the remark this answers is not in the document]
valentijnscholten: I also use flake8 in vscode. And the new pylance, it's an improvement!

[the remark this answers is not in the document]
danielnaab: It might in part be a matter of the `.flake8` config in the codebase, but flake8 is missing a lot of structural issues - undefined variables, etc. pylint found several such bugs missed by flake8.

valentijnscholten (approved):

valentijnscholten:
Thanks, much easier to review now.

damiencarol (approved):

Issue #3731, opened 2021-01-28
jira webhook: key missing on comment update


madchap:
**Bug description**
```
Traceback (most recent call last):
  File "./dojo/jira_link/views.py", line 184, in webhook
    commentor = parsed['comment']['updateAuthor']['name']
KeyError: 'name'
```

In some instances - like mine - there is no `name` but `emailAddress`, most likely due to SSO logins in JIRA.

**Steps to reproduce**

**Expected behavior**

**Deployment method** *(select with an `X`)*
- [x] Kubernetes
- [ ] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
1.12.0

PR #3732, opened 2021-01-28
Webhook: Fix JIRA key error for name

madchap:
Addresses #3731


valentijnscholten:
Not sure if you should split the email address. Probably the jira username Used by defect dojo is also an email address in that case so need to compare against that. Or against both maybe.

madchap:
You're right, just made the test, email is in full. Will amend.

madchap:
I was looking at the fixture we use for test and I could not see easily which jira configuration would be used for whatever jira_id is specified in the unittest. Seems no product actually has it? Maybe I missed it, will look again tomorrow.

valentijnscholten:
what we are seeing is that some instances send only name some instance only emailaddress. So two separate cases, needing 2 separate code paths / unit tests? You can use finding5 for bothI think, but your new testcase shouldn't have the name field in the json I think, only email.

dojo/fixtures/dojo_testdata.json
```diff
@@ -1686,6 +1686,24 @@
...
+      "url": "http://www.testjira.com",
+      "medium_mapping_severity": "test severity",
+      "open_status_key": 333,
+      "close_status_key": 334,
+      "critical_mapping_severity": "test severity",
+      "high_mapping_severity": "test severity",
+      "password": "user69",
+      "epic_name_id": 666
+    }
+  },
```
valentijnscholten, line 1706: do we need this?

valentijnscholten: ah yes to make one with an email address.

madchap: yea, just have to fiddle with extra things now in the fixture...

valentijnscholten: easier just to change the username on jira_instance with id 2 in python code and do .save()

valentijnscholten: fixtures are a pain and should onyl be used for a very minimal set of data to get the instance going.

madchap:
Should be OK now, except for that unrelated report timeout selenium test failing intermittently.

valentijnscholten (approved):

valentijnscholten:
thanks for fixing this!

damiencarol (approved):

PR #3737, opened 2021-01-29
Contrast parser: maintenance

damiencarol:
**Work:**
- [x] Remove pandas (we switch all parser to only use csv or json packages)
- [x] add request details
- [x] add unit tests (of course)


valentijnscholten (approved):
thanks. using a full data science library to parse some json is indeed a bit over the top. and pandas doesn't provide any updates anymore for python 3.6 and lower.

madchap (approved):

PR #3743, opened 2021-01-29
Add metrics queries tests

danielnaab:
Unit tests and bugfix extracted from #3730


dojo/metrics/views.py
```diff
@@ -321,7 +321,7 @@ def endpoint_querys(prod_type, request):
                 Q(endpoint__product__prod_type__authorized_users__in=[request.user]))
         accepted_endpoints_counts = severity_count(accepted_endpoints_counts, 'aggregate', 'finding__severity')
     else:
-        endpoints_closed = Endpoint_Status.objects.filter(mitigated__date__range=[start_date, end_date]).prefetch_related(
+        endpoints_closed = Endpoint_Status.objects.filter(date__range=[start_date, end_date]).prefetch_related(
```
valentijnscholten, line 324: should this be `mitigated_time__range=...` ?

valentijnscholten:
Thanks, this will help with reviewing "the other PR" :-)

dojo/metrics/views.py
danielnaab: Makes sense, updated

valentijnscholten (approved):
thanks. I am wondering what else we would need to test for the other parts of the metrics, i.e. some funky python logic going on in `product/views.py#view_product_metrics`. But not all are easy to capture in unit tests. So maybe gather some evidence to make it easy for us later on to see the results are the same.

Maffooch (approved):

Issue #3744, opened 2021-01-29
Google Sheet integration broken with KeyError and Data too long error


madchap:
**Bug description**
It seems the Google Sheet integration is currently broken.

![image](https://user-images.githubusercontent.com/5468769/106308714-645b8000-6261-11eb-9a2a-32b340d6b8d5.png)

Found out linked to https://github.com/DefectDojo/django-DefectDojo/issues/3742

madchap:
fyi @piyarathnalakmali  - I don't know if you'd still be interested in keeping the integration up-to-date? Thanks.

madchap:
Also found on fresh local instance the following upon configuring it

```
Traceback:

File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py" in _execute
  84.                 return self.cursor.execute(sql, params)

File "/usr/local/lib/python3.6/site-packages/django/db/backends/mysql/base.py" in execute
  71.             return self.cursor.execute(query, args)

File "/usr/local/lib/python3.6/site-packages/MySQLdb/cursors.py" in execute
  206.         res = self._query(query)

File "/usr/local/lib/python3.6/site-packages/MySQLdb/cursors.py" in _query
  319.         db.query(q)

File "/usr/local/lib/python3.6/site-packages/MySQLdb/connections.py" in query
  259.         _mysql.connection.query(self, query)

The above exception ((1406, "Data too long for column 'column_widths' at row 1")) was the direct cause of the following exception:

File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py" in inner
  34.             response = get_response(request)

File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  115.                 response = self.process_exception_by_middleware(e, request)

File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  113.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/usr/local/lib/python3.6/site-packages/django/contrib/auth/decorators.py" in _wrapped_view
  21.                 return view_func(request, *args, **kwargs)

File "./dojo/google_sheet/views.py" in configure_google_sheets
  99.                     system_settings.save()

File "/usr/local/lib/python3.6/site-packages/django/db/models/base.py" in save
  744.                        force_update=force_update, update_fields=update_fields)

File "/usr/local/lib/python3.6/site-packages/django/db/models/base.py" in save_base
[...]

Issue #3742, opened 2021-01-29
view test: Arbitrary file uploads conflicts with Google Sheet integration, causing 500 errors


madchap:
**Bug description**
Unsure if it's again me or if that's also others. At any rate, viewing test page results in 500 with

```
Template error:
In template /app/dojo/templates/base.html, error at line 703
   expected str, bytes or os.PathLike object, not NoneType
   693 :         }
   694 : 
   695 :         function get_alerts() {
   696 :             $('.dropdown-alerts').html('<div class="text-center"><i class="fa fa-spin fa-spinner"></i></div>');
   697 :             $.get("{% url 'alerts_json' %}?limit=12", function (data) {
   698 :                 $('.dropdown-alerts').empty();
   699 :                 $.each(data, function(i, elem) {
   700 :                   var titleField = elem.fields.title ? elem.fields.title : elem.fields.description;
   701 :                   var display_title = titleField.length < 50 ? titleField : jQuery.trim(titleField).substring(0,50).trim(this) + '...';
   702 :                   var description = elem.fields.description
   703 :                   v ar display_description = description  ? description.length < 100 ? description : jQuery.trim(description).substring(0,100).trim(this) + '...': ''
   704 :                   var dropdown_alert = "alert-list" + i;
   705 :                   $('.dropdown-alerts').append('<li><div style="padding: 3px 20px;"><i class="fa fa-' + htmlEscape(elem.fields.icon) +
   706 :                                                 ' fa-fw"></i><a href="' + htmlEscape(elem.fields.url) + '" id=' + dropdown_alert +
   707 :                                                 ' title="' + htmlEscape(display_description) + '"></a><span class="pull-right text-muted small">' +
   708 :                                                htmlEscape(elem.fields.source) + '</span></div></li><li class="divider"></li>');
   709 :                   $("#" + dropdown_alert).text(display_title);
   710 :                 });
   711 :                 $('.dropdown-alerts').append('<li><a class="text-center" href="{% url "alerts" %}"><strong>See All Alerts </strong>' +
   712 :                                              '<i class="fa fa-angle-right"></i></a></li>');
   713 :                 $('.dropdown-alerts').append('<li><a class="text-center" href="{% url "delete_alerts" %}"><strong>Clear All Alerts </strong>' +


Traceback:

File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py" in inner
  34.             response = get_response(request)

[...]

PR #3747, opened 2021-01-30
Fix  google sheets + tests

madchap:
Addresses https://github.com/DefectDojo/django-DefectDojo/issues/3742 and https://github.com/DefectDojo/django-DefectDojo/issues/3744


dojo/db_migrations/0072_sheets_textfields.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.17 on 2021-01-30 07:23
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dojo', '0071_product_type_enhancement'),
```
damiencarol, line 9: we already passed this number.
```suggestion
        ('dojo', '0073_product_type_enhancement'),
```
Also rename the file

damiencarol (approved):

damiencarol:
It seems that there is a file missing in the PR:
```
uwsgi_1         | ======================================================================
uwsgi_1         | ERROR: test_config_google_sheets (dojo.unittests.test_google_sheets_configuration.GoogleSheetsConfigTestApi)
uwsgi_1         | ----------------------------------------------------------------------
uwsgi_1         |   File "/app/dojo/unittests/test_google_sheets_configuration.py", line 54, in test_config_google_sheets
uwsgi_1         |     with open('tests/defectdojo-sheets-localdev.json', 'rb') as f:
uwsgi_1         | FileNotFoundError: [Errno 2] No such file or directory: 'tests/defectdojo-sheets-localdev.json'
```

madchap:
> It seems that there is a file missing in the PR:
> 
> ```
> uwsgi_1         | ======================================================================
> uwsgi_1         | ERROR: test_config_google_sheets (dojo.unittests.test_google_sheets_configuration.GoogleSheetsConfigTestApi)
> uwsgi_1         | ----------------------------------------------------------------------
> uwsgi_1         | Traceback (most recent call last):
> uwsgi_1         |   File "/app/dojo/unittests/test_google_sheets_configuration.py", line 54, in test_config_google_sheets
> uwsgi_1         |     with open('tests/defectdojo-sheets-localdev.json', 'rb') as f:
> uwsgi_1         | FileNotFoundError: [Errno 2] No such file or directory: 'tests/defectdojo-sheets-localdev.json'
> ```

Right, I was fighting the VCR, hence why the PR was still in Draft status when you reviewed it. Should be OK now, but waiting for tests to complete.

madchap:
I could not make the VCR test work replay the stream with success if the json key file drifts too much. So that's the former key I used to test with, which obviously is now gone, account disabled, etc.. :man_shrugging: 

Wondering if we'll get an email saying we've pushed private creds..

If anyone has a better idea?

valentijnscholten:
Find out which method performs the validation and mock it to always return True.

valentijnscholten:
I tried to mock the validation method, but that skips some of the http requests breaking the test using the recorded/replayed request. So quickest solution is just to merge as-is. It looks weird, but the private key is not a real existing key, so no harm done.

Issue #3631, opened 2021-01-15
Re-import information should be displayed (permanently) in the UI


madchap:
Currently, we can have an email for that but it'd be real good to see it in the UI.

fyi @valentijnscholten per our discussion.

Issue #3375, opened 2020-11-30
Search inside findings tab does not find excerpted value


Wadeck:
**Bug description**
The Search box inside the Findings table (acting as a filter) does take into account only the visible values. It means that when you have a long "name", the value is excerpted to be displayed with some ... at the end. But the filter should take into account the real value, not the displayed one.

You can also check by looking just for "...", which does not appear in any of my data, but due to the excerpt, show lots of findings.

**Steps to reproduce**
1. Go to a test
1. Inside the Findings table
1. Look for a name that is excerpted (the "..." at the end)
1. Try filtering (using the Search box inside the table) with a part that is not displayed
1. It will not appear

**Expected behavior**
The filter should use the real data and not the visible ones.

**Screenshots** (optional)

### The desired finding
![Screenshot_2020-11-30_105501_001](https://user-images.githubusercontent.com/2662497/100595158-d6175c00-32fa-11eb-86d0-43679fe55d32.png)

### Regular filter with its name
![Screenshot_2020-11-30_104639_001](https://user-images.githubusercontent.com/2662497/100594537-1a562c80-32fa-11eb-989f-dc8b5743a89a.png)

### Filtering with "..." for the example
![Screenshot_2020-11-30_105309_001](https://user-images.githubusercontent.com/2662497/100595188-e0395a80-32fa-11eb-8e2d-d5e31ba8c3c7.png)

### Buggy filtering with its CVE due to excerpt
![Screenshot_2020-11-30_105511_001](https://user-images.githubusercontent.com/2662497/100595235-f1826700-32fa-11eb-8866-0bb280ce89af.png)

valentijnscholten:
Please bear in mind this is just a client side filter, so it can only filter on data that is present there. Easiest fix for this particular usecase is to add the CVE column which is present everywhere else where we show lists of findings.

valentijnscholten:
CVE field was added here as fix for the immediate issue at hand.

PR #3748, opened 2021-01-30
feature flag: Track Import history per Test

### LINKED ISSUE #3746: allow different pagesizes when having multiple paging lists on the same page
some pages have multiple lists which are paged. the page number is already niceley separated so you can have list 1 on page 5 and list 2 on page 1. but the page size is linked. so if you choose pagesize 100 for list 100, it will also set pagesize 100 for list 2.

valentijnscholten:
The ReImport feature is very powerful as it recognizes existing findings and will not create duplicate if not needed. It will also close findings that are no longer present in reports. And it will re-open findings that have re-appeared or have been marked as mitigated incorrectly.

The biggest reason I have heard for people _not_ using ReImport is that it doesn't retain its history. It does put a Note on a finding when it closes or reactivates a finding. But there's no clear overview of what happened. 

Now that we have better support for versions on tests via #3726, we add a "ReImport History" feature in this PR. It is a very basic implementation only for ReImport and not for the normal Import yet. We'd like to get some experience & feedback from the community to see which direction it needs to be improved. My personal goal is to make ReImport so much better that people need much less duplicates. Currently people are importing every scan, every version, every branch seperately. Creating 1000s of tests, 10s of thousands of findings, etc. 

The ReImport History is protected by a feature flag, which is disabled by default: `TRACK_IMPORT_HISTORY (DD_TRACK_IMPORT_HISTORY)`

Features:
- view_test: provide (collapsed by default) list of all reimports and the changes made by that reimport  (+initial import) (closed x findings, created y findings, reactivated z findings);
- view_test: this also shows the version and endpoint provided for each import/reimport;
- view_finding and lists of findings: provide a history icon next to the status to see history of the status (limited to status changes by import/reimport only);
- view_engagement: show number of reimports for each test
- unit tests: annotate the existing `test_import_reimport` tests with assertions on the correct number of models expected to be created. Although it's very basic to assert only number of objects created, I did find a nice clean way to have only 1 line of code to test all created/closed/reactivated/etc types of objects created.
- other fixes #3746, #3375 , #3756
- Also addresses https://github.com/DefectDojo/django-DefectDojo/issues/3631

Screenshots:
image
import settings:
image
status history per finding
image
clicking on reimport 7 with 3 created findings:
image
view engagement:
image


madchap:
fyi @Wadeck @jvz

madchap (approved):
Seems to work like a charm. Really nice addition to see what changes in between :+1:  good job.

Maffooch (approved):
This is super cool! The last test is accessLint but its usually bugged so approving

PR #3757, opened 2021-01-31
AuthZv2.0: Object-based authorization (1st pull request)

StefanFl:
This pull request is the start of an implementation for a new object based authorization system. 
* https://docs.google.com/document/d/1XcbNtc_66rFIYQ4ANw2TcLn8ZhC6ftSXkfGtVvcfFEw/edit?usp=sharing is the documentation for it, 
* https://drive.google.com/file/d/1-Ly67lTDRkWFZEGBF3wOSQIQw7RULf8s/view?usp=sharing is a short video, showing how it will work with the product types as an example. 

The implementation will touch many areas of DefectDojo and will need some time. Therefore I plan with a series of pull request and using a feature flag. Nothing will change for the users until it is finished, but you can change the feature flag to get a preview of the new authorization.

This first pull request contains most of the base functionality, management of permissions for product types and using permissions for product types:

* **dojo/authorization/*.py** and **dojo/templatetags/authorization_tags.py :** This is the base functionality, including the definition of check permissions and roles and functions to check if a users has a permission
* **dojo/models.py** and **dojo/db_migrations/0072_permissions.py:** Changes in the model and new data base tables to manage memberships of users to products and product types
* **dojo/product/queries.py** and **dojo/product_type/queries.py:** Centralized queries to return products and product types a user is allowed to see according to his memberships
* **dojo/product_type/urls.py** and **dojo/product_type/views.py:** Management of memberships for product types and permission checks based on memberships
* **dojo/forms.py**,  **dojo/templates/delete_product_type_member.html**, **dojo/templates/dojo/edit_product_type_member.html** and **dojo/templates/dojo/new_product_type_member.html**: Dialogues to manage membership for product types
* **dojo/templates/dojo/view_product_type.html** and **dojo/templates/dojo/product_type.html:** List of members with menus to manage them and permission checks for all actions
* **dojo/api_v2/*:** Check of new authorization for product types and serialization of product type members
* **dojo/settings/settings.dist.py:** The feature flag
* **dojo/unittests/authorization/test_authorization.py:** The start for unit tests

This is planned to be the largest pull request, the next ones will be substantially smaller.


valentijnscholten:
Great contribution. First thoughts:
- the API also needs some updates I think (APIv2)?
- is it possible to test the decorator(s) itself/themselves? i.e. call a decorated (but patched) function and see if the decorator grants/denies access whenever applicable? for old and new style auth?
- in the old style we had some `if is_staff` and somtimes `is is_superuser`. Both are different types of users. Does the "new style" handle this properly?
- could we think about making the templates more feature flag independant so that if we in the future _remove_ the (current) old way of authorization, we don't have to change the templates to get rid of everything named "feature_new_authorization" or similar?
 i.e. change `{% if request.user|feature_new_authorization_or_user_is_staff %}` into `{% if request.user|show_context_menu %}` and have the logic for that in the tag filter implementation?
similar maybe for `{% if feature_new_authorization_flag %}` into `{% if show_authorized_users_for_product_type %}`

Not sure if it's possible everywhere but my feeling is some place can be made a bit more cleaner in the template.

StefanFl:
Hi @valentijnscholten, thanks for the feedback.

* the API also needs some updates I think (APIv2)?
  --> Yes, this would be something I want to look at it soon. Would it help for this PR, if I include the authorization for product types to see how it works? Didn't do it so far to make this PR not too large.
* is it possible to test the decorator(s) itself/themselves? i.e. call a decorated (but patched) function and see if the decorator grants/denies access whenever applicable? for old and new style auth?
  --> Definitely, didn't want to invest too much in tests before this implementation of a new authorization is approved by the maintainers, but then all functions in the view shall be tested if they are properly authorized.
* in the old style we had some if is_staff and somtimes is is_superuser. Both are different types of users. Does the "new style" handle this properly?
  --> The majority of cases where ''is_superuser'' is used are not related to objects but is staff like managing users or the configuraion. These will not be changed. There are some in the dashboard and the metrics I will have to investigate.
* could we think about making the templates more feature flag independant so that if we in the future remove the (current) old way of authorization, we don't have to change the templates to get rid of everything named "feature_new_authorization" or similar?
    i.e. change {% if request.user|feature_new_authorization_or_user_is_staff %} into {% if request.user|show_context_menu %} and have the logic for that in the tag filter implementation?
    similar maybe for {% if feature_new_authorization_flag %} into {% if show_authorized_users_for_product_type %}
  --> I did it intentionally this way. All the ''{% if feature_new_authorization_flag %}'' shall be removed completely, after the implementation of the new autorization is finished. This will make the templates much tidier after the transition.

damiencarol:
Awesome work @StefanFl ! I will help to test it!

StefanFl:
@valentijnscholten suggested to make the model field for the role in product_member and product_type_member a string to be more descriptive. I like that idea but it would disrupt the testing because of the database change. Any objections?

StefanFl:
After some consideration I would like to keep the Roles an integer field. 

- The role is used in subqueries to find authorized objects. Integer fields might be faster there than strings.
- While there is no hierarchy in Permissions, there is actually one in Roles. When you are editing members, you cannot assign another user a higher role than you have yourself. Otherwise a Maintainer could assign another user to be an Owner.

StefanFl:
@valentijnscholten With the commit https://github.com/DefectDojo/django-DefectDojo/pull/3757/commits/9d431fca5dad7f6ed31f3e645cd99a03ad61cd34 I removed the last remains of an ordering of roles. But another reason to let the roles be an integer in the database is the possibility of having configurable role. With the latest change it will be possible to make roles completely flexible and have the configuration in the database, if we want to do it in a later stage. Change would be in 1 or 2 files in the dojo/authorization folder.

I am really eager to get more feedback to see if we need more changes to it, did you talk in the group of maintainers about it?

valentijnscholten:
Did a quick review, although I think it's easier to have this merged and do some testing in real life.
Some thoughts, some of which may already be planned or known issues: 
- UI for products: Would be welcome, currently it is still showing authorized users. Not sure how that is now interacting with the legacy vs new authorization? Looking at the code I think it's ignored so might be best to hide this field.
- Caching of permissions: I think there multiple queries and joins done for permission checks or at least when getting authorized models, so maybe cache the permissions in the user serssion
- Tests: as we've seen recently we almost have no other choice but to put all API endpoints/methods and all UI API/endpoints methods in testcases to make sure we haven't forgotten or misplaced any annotation. Or we should find a way to make all those things available only to staff users, unless a decorator is present to specify otherwise
- global roles: I think we found out in the past we may need some other system wide roles as well, i.e. SuperReader that can view all products/findings, but not modify them
- technical user: I think the technical user may need more permissions, for example to create engagements or reimport scans, edit engagements, edit tests. 
- loss of functionality: in the legacy model people can go into django admin and specify some permissions for a user and user that in some ci/cd system for example. currently we have a fixed TechnicalUser role for that, so it's less flexible. For me it's not to big of an issue, but maybe for others it is. I think when using CICD you almost need staff permissions because most of the time you want to autocreate product types, products, engagements, tests, etc.
- staff vs superuser: in the legacy system there is no distinction between staff users and superusers. So if you want to allow someone to manage all findings and products, they also get to manage system settings, jira integration, notifications, etc. If I understand the Django idea correctly the staff user is conceptually meant for the system management tasks. Either via the django admin portal or otherwise. The superuser is meant to have access to all models, which is confusing because that includes system settings. Anyway, I think I would like to see a seperate role/permission that's needed for management system settings.

StefanFl:
Thanks a lot @valentijnscholten for your review, here are my thoughts about your remarks:

- UI for products: Would be welcome, currently it is still showing authorized users. Not sure how that is now interacting with the legacy vs new authorization? Looking at the code I think it's ignored so might be best to hide this field.
_--> I haven't implemented it yet to make the first PR not too large. Can do it quite quickly if it helps. But my plan was that the new authorization is inconsistent until it's completely finished, until then it's only a preview and the feature flag shouldn't be set to true in production_
- Caching of permissions: I think there multiple queries and joins done for permission checks or at least when getting authorized models, so maybe cache the permissions in the user serssion
_--> The new authorization is more heavy on the database, so a cache might help. The most complex query you find in https://github.com/StefanFl/django-DefectDojo/blob/permissions/dojo/product/queries.py, we will have similar queries for other classes as well. What is the right way to cache these? I found something like https://github.com/noripyt/django-cachalot_
- Tests: as we've seen recently we almost have no other choice but to put all API endpoints/methods and all UI API/endpoints methods in testcases to make sure we haven't forgotten or misplaced any annotation. Or we should find a way to make all those things available only to staff users, unless a decorator is present to specify otherwise
_--> Yes, we will need more testcases, I am afraid every endpoint need 2 testcases, one authorized and one unauthorized._ 
- global roles: I think we found out in the past we may need some other system wide roles as well, i.e. SuperReader that can view all products/findings, but not modify them
_--> That's a good idea and will be quite easy to implement because the authorization checks have a central logic. But I don't think we need it from the start._ 
- technical user: I think the technical user may need more permissions, for example to create engagements or reimport scans, edit engagements, edit tests.
_--> Easy to implement, maybe not so easy to decide which permissions the technical user shall have_
- loss of functionality: in the legacy model people can go into django admin and specify some permissions for a user and user that in some ci/cd system for example. currently we have a fixed TechnicalUser role for that, so it's less flexible. For me it's not to big of an issue, but maybe for others it is. I think when using CICD you almost need staff permissions because most of the time you want to autocreate product types, products, engagements, tests, etc.
_--> Good point. I could amend the `user_has_permission` method and look for the django permissions. But these would give a user access for all objects, so it would have to be used with caution._
- staff vs superuser: in the legacy system there is no distinction between staff users and superusers. So if you want to allow someone to manage all findings and products, they also get to manage system settings, jira integration, notifications, etc. If I understand the Django idea correctly the staff user is conceptually meant for the system management tasks. Either via the django admin portal or otherwise. The superuser is meant to have access to all models, which is confusing because that includes system settings. Anyway, I think I would like to see a seperate role/permission that's needed for management system settings.
--> I have the feeling it's a bit different. A staff user can work with all data but cannot manage the system settings, while the superuser can do everything. Concerning system settings my approach wouldn't change that, the system settings can still only be managed by the superuser.

valentijnscholten (approved):
Let's merge as-is, and improve from there. Maybe add a comment to the feature flag that it is in alpha / incomplete state at the moment.

StefanFl:
Thanks @valentijnscholten, changed the comment.

madchap (approved):

PR #3759, opened 2021-01-31
Allow Findings filter by tags of all object levels

Maffooch:
Add the ability to filter tags of all objects instead of just findings. For example, when generating an engagement report, some tests might be tagged to be reported while others are tagged to be excluded. Only one test tag needs to be specified instead of tagging each finding in the test to be reported. This policy also applies to products and engagements.

image

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -746,23 +746,24 @@ class OpenFindingFilter(DojoFilter):
     test__tags = ModelMultipleChoiceFilter(
         field_name='test__tags__name',
         to_field_name='name',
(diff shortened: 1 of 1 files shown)


valentijnscholten:
I thought this was already in place, but then I noticed what was happening. I added these filter fields in the tagulous PR or the one thereafter. But it uses the Finding tags to populate all of the multiselects. So this PR is actually a bugfix? :-) (and addition to closedinfindings filter).

valentijnscholten (approved):

mtesauro (approved):
Approved

Issue #3766, opened 2021-02-01
[GHA] Add k8s template verification


madchap:
**Is your feature request related to a problem? Please describe**
Need to do basic checks on k8s stuff.

**Describe the solution you'd like**
Start with `helm template --debug .` or something alike, on basic as-is + with some annotations.

PR #3767, opened 2021-02-01
Fix helm chart

madchap:
I got some 

```  unknown field "kubernetes.io/ingress.class" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta, ValidationError(Ingress.metadata): unknown field "kubernetes.io/tls-acme" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta, ValidationError(Ingress.metadata): unknown field "nginx.ingress.kubernetes.io/proxy-body-size" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta, ValidationError(Ingress.metadata): unknown field "nginx.ingress.kubernetes.io/proxy-read-timeout" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta, ValidationError(Ingress.metadata): unknown field "nginx.ingress.kubernetes.io/ssl-redirect" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta]```

When updating the chart. 

The `annotations` is already with 2 indents, so below needs to be `4` and the leading spaces throw things off, e.g. relative indents do not seem to work.

Thanks @ccojocar for the hand :+1:

### DIFF
--- a/helm/defectdojo/templates/django-ingress.yaml
+++ b/helm/defectdojo/templates/django-ingress.yaml
@@ -12,9 +12,9 @@ metadata:
     helm.sh/chart: {{ include "defectdojo.chart" . }}
 {{- if or .Values.django.ingress.annotations .Values.gke.useGKEIngress }}
   annotations:
-  {{- with .Values.django.ingress.annotations }}
-  {{ toYaml . | indent 2 }}
-  {{- end }}
+{{- with .Values.django.ingress.annotations }}
+{{ toYaml . | indent 4 }}
+{{- end }}
   {{- if .Values.gke.useGKEIngress }}
     kubernetes.io/ingress.class: gce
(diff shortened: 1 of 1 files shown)


madchap:
fyi @jalseth -- I have not checked things further, but at least it fixes the issues on my end.

madchap:
Diff now works, but my deploy still fails. Putting to draft.

ccojocar (approved):

madchap:
Had an unrelated error, seems good :+1:

damiencarol (approved):

valentijnscholten:
is there anyway we could have prevented this, or only if we had a GKE to test it on?

valentijnscholten:
ah, just noticed #3766

alles-klar (approved):

PR #3768, opened 2021-02-01
risk acceptance: fix notes bugs

valentijnscholten:
two bugs fixed:
- when adding a risk acceptance with a note, a many2many relationship related error occurred
- the 'e' keydown event (to edit the risk acceptance) conflicted with the notes textarea (works fine on every other page we use the same construct)

### DIFF
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -855,6 +855,17 @@ def add_risk_acceptance(request, eid, fid=None):
     if request.method == 'POST':
         form = RiskAcceptanceForm(request.POST, request.FILES)
         if form.is_valid():
+            # first capture notes param as it cannot be saved directly as m2m
+            notes = None
+            if form.cleaned_data['notes']:
+                notes = Notes(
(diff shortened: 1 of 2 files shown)


madchap (approved):
lgtm, thanks :)

Maffooch (approved):

wurstbrot:
Note for others:
In case the finding is not correctly stored (my developers say that the error is sometimes not shown but maybe in the background), it might lead to think the finding got accepted (because it is shown) and it is true. The process of setting the status to inactive is not applied and therefore the finding will be shown again.

Attached "Correct application of acceptance" and "Bad application of acceptance":
Screenshot from 2021-03-16 09-24-00
Screenshot from 2021-03-16 09-24-26

valentijnscholten:
@wurstbrot Is this a new bug you are seeing or are you just adding info to the old/fixed bug?

wurstbrot:
Just adding info to the old/fixed bug.
I haven't tried the new version with this bug fix, but I belive that my described behaivor is fixed by this PR.

PR #3769, opened 2021-02-01
jira webhook: fix risk acceptance handling

valentijnscholten:
webhook had some copy and paste madness causing issues when un-accepting a finding.


madchap:
Do we even want to do anything to risk acceptance when incoming jira come in? I would just leave them be.

madchap (approved):
I am still not entirely convinced, but since it's known behavior and it's currently broken, let's go ahead.

Maffooch (approved):

Issue #3771, opened 2021-02-01
The Debug toolbar does not load when enabled and DefectDojo runs in Debug mode


nlandais:
If you add the debug_tools to the list of installed in your local settings, you'll notice that the login page does not load and the debug information provided points to the fact that the URL cannot be found.
Suggested fix, include the debug_toolbar urls to the list of url patterns

valentijnscholten:
We don't want to include/enable the debug toolbar by default. To enable it locally is documented, for example here: https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/settings/template-local_settings

PR #3772, opened 2021-02-01
Adding debug_toolbar to urlpatterns

nlandais:
Fix for issue #3771, add debug_toolbar urls to the list of of url patterns in order to get some insights into the system when running in Debug mode


valentijnscholten:
We don't want to include/enable the debug toolbar by default. To enable it locally is documented, for example here: https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/settings/template-local_settings

PR #3780, opened 2021-02-02
product: prefetch verified count, add missing indexes 🏎️

valentijnscholten:
the verified count for products was not prefetched causing 25 extra queries on the product pages 

by using annotate on the main query, the counts are now retrieved in 1 query for all products at once.

but still the product overview page took over 3s to load, same for product types overview

Products: (>3s)
image

Product Types: (>3s)
image

All relevant columns were indexed, but on more complex queries with lots of joins, MySQL is unable to use them. It needs a composite index combining `test` and `active` and `verified`. So this PR adds that.

Both overview pages are now fast again, even on a 700k finding instance:

Products: (<1s)
image

Product Types: (<1s)
image

This index will probably help in more places where we do joins and look at active/verified at the same time.

I'll probably do some more digging with regards to our queries and some other indices that might be needed. This one here for now is a big improvement, so we should merge it :-)

PR also does some formatting improvement and correct link on list of product types active/verified finding counts

EDIT:
- Went over the codebase and queries and there are some more places where we have the same pattern (list of engagements, list of active engagements, list of endpoints, list of vulnerable endpoints). So I added some more needed composite indexes.
- Added some prefetches for the list of (vulnerable) endpoints so it performs a bit better (but that needs more work in differents PRs).


madchap (approved):
I owe you another beer.

devGregA (approved):

PR #3781, opened 2021-02-02
Format code with Black and add pre-commit hook

danielnaab:
This PR runs Black and adds configuration for work with it.

1. Formatted entire repository:

```bash
black --fast .
```

2. Added `.git-blame-ignore-revs` with the git commit from step 1 above. This can be used to omit the formatting when doing `git blame`: `git blame important.py --ignore-revs-file .git-blame-ignore-revs`

3. Added `.pre-commit-config.yaml` with Black configuration, as outlined here: https://black.readthedocs.io/en/stable/version_control_integration.html#version-control-integration


valentijnscholten:
We may need to discuss which formatter/linter we want to use. My first impression is that black generates some very ugly looking code. Might be a reason for it, but not yet convinced.

devGregA:
Closing, as this appears to have gone stale, happy to discuss linting in slack via GH discussions.

PR #3782, opened 2021-02-02
Re-enable Jira Epic Mapping

Maffooch:
Enable Jira epic mapping again and add some unit tests. Unit tests involve adding an epic and ensuring issues are linked to the epic upon pushing findings to jira.

All other jira test cassettes were updated as well to guarantee the tests till work.


dojo/jira_link/helper.py
```diff
@@ -874,6 +889,16 @@ def jira_get_issue(jira_project, issue_key):
...
+# Return a list of jira issue in json format.
+def get_epic_issues(engagement):
+    instance = get_jira_instance(engagement)
+    jira = get_jira_connection(instance)
+    epic_id = get_jira_issue_key(engagement)
+    url = instance.url.strip('/') + '/rest/agile/1.0/epic/' + epic_id + '/issue'
+    response = jira._session.get(url).json()
+    return response.get('issues', [])
+
+
```
valentijnscholten, line 901: is this only used in the tests? because in that case it would be better placed in dojo_test_case.

dojo/jira_link/helper.py
```diff
@@ -807,6 +814,14 @@ def update_epic(engagement):
...
 def add_epic(engagement):
+    add_epic_sync(engagement)
+
+
+# This code was not being tested, but calling it
+# as a task was not working. Using block execution
+# would not either here because this method is called
+# explicitly instead of through a save function
+def add_epic_sync(engagement):
     logger.info('trying to create a new jira EPIC for %d:%s', engagement.id, engagement.name)
```
valentijnscholten, line 825: might be better ad debug level?

dojo/jira_link/helper.py
```diff
@@ -807,6 +814,14 @@ def update_epic(engagement):
...
 @dojo_model_from_id(model=Engagement)
 def add_epic(engagement):
+    add_epic_sync(engagement)
+
+
+# This code was not being tested, but calling it
+# as a task was not working. Using block execution
+# would not either here because this method is called
+# explicitly instead of through a save function
+def add_epic_sync(engagement):
```
valentijnscholten, line 824: what was the error seen when doing async? I think we should keep everything related to JIRA in  celery otherwise it gets confusing.

valentijnscholten:
left comments. In general I would prefer the new logging statements to de at debug as they are all just low level details normally not needed. maybe epic creation is interesting, but others are more for debug level.

dojo/unittests/test_jira_import_and_pushing_api.py
```diff
@@ -223,6 +223,15 @@ def test_import_twice_push_to_jira_push_all_issues(self):
...
 
+    def test_engagement_epic_mapping(self):
+        eng = self.get_engagement(3)
+        self.create_engagement_epic(eng)
+        import0 = self.import_scan_with_params(self.zap_sample5_filename, push_to_jira=True, engagement=3)
+        test_id = import0['test']
+        self.assert_jira_issue_count_in_test(test_id, 2)
+        self.assert_epic_issue_count(eng, 2)
+        self.assert_cassette_played()
+
```
valentijnscholten, line 234: I think they main problem we saw in the past that if engagement mapping was enabled, the epic was not created and then pushing failed. So I think the tests should test the full flow including creation of the epic. And also the case where findings are pushed and epic mapping is enabled but the epic is not present, it should create a new epic maybe or if not, we should test for the epic not being created in that case.

dojo/jira_link/helper.py
Maffooch: Sure can do.

dojo/jira_link/helper.py
Maffooch: No I don't think so. The Jira integration is very error prone to set up the first time. Not knowing if one of the functions even ran is extremely annoying. Most people look to the logs to figure out what happened and then will get nothing out of them.  Logging is not set to debug for most prod instances and knowing dojo well enough to know that enabling debugging in the setting is necessary is not common either.

dojo/jira_link/helper.py
Maffooch: Celeryworkers exit before unit tests run, so most of the tests are ran with a user in block execution mode. The jira_import_and_pusing_api tests have a user that user uses block execution, but the dojo_api_test_case does not. The functions I am using are in the dojo_api_test_case so the add_epic function will run async. With celery not active, the function never runs. Maybe I could try putting these functions in the same file as the unit test and see what happens.

dojo/unittests/test_jira_import_and_pushing_api.py
Maffooch: The full flow of the epic being created is present here. Specifically on line 228. I don't like the case of auto creating epics for mapping if the epic did not already exist. This means it would have to be explicitly decided to make it map by using the "creat epic" button on the new/edit engagement page. There is error handling present for this exact use case, but I guess a test couldn't hurt. 

So just to be clear. You want a test to show epic mapping enabled, epic is not created, findings are still pushed.

dojo/jira_link/helper.py
Maffooch: So all the tests here are run through the API under a user with a token. This function is ran unauthenticated though. So there is no way to specify it be ran under block execution. Having a sync and async version of the function is the only way. This is not confusing in the slightest though. The function is ran through celery in every single instance except this unit test. It's no different than the old way of doing things with functions set up like utils/tasks.

valentijnscholten: That's whay I meant by that the engagement should not be created from the unittest. The unit test should call the rest api or the ui api with a post to create the engagement. It should assert that the epic is created if the epic mapping is enabled. Another test should assert an epic is not created if the epic mapping is not enabled.

dojo/unittests/test_jira_import_and_pushing_api.py
valentijnscholten: In theory if the epic mapping is enabled, the epic is created. But sometimes this fails, for example if the jira instance requires some extra fields which dojo doesn't know/provide. So for the scenario we should verify that the findings can still be pushed. And if we see it as a requirement that dojo should not try to create the epic, we should assert that as well.

dojo/jira_link/helper.py
valentijnscholten: I think this indicates that we should do a better job somehow in the UI. I think we already do by posting alerts with the error messages.

dojo/unittests/test_jira_import_and_pushing_api.py
Maffooch: No that's not how that works. That would be fine if the jira is created on the engagement level,  but what if it is inherited from the product? Should every engagement have an epic or just the ones you explicitly ask for?

dojo/jira_link/helper.py
Maffooch: Okay well I'll set it to debug  on this PR because this is getting out of scope.

dojo/jira_link/helper.py
Maffooch: Epic mapping and epic creation are not a package deal.

Maffooch:
Alright so here's what I did tonight:
- Ad unit tests for all permutations of epic mapping and epic creation. This consists of four tests.
  - Epic mapping enabled and epic is created (findings are pushed to epic)
  - Epic mapping enabled and epic is NOT created (findings are pushed to jira, but not to the epic since it does not exist)
  - Epic mapping disabled and epic is created (findings are pushed to jira, but not to the epic)
  - Epic mapping disableed and epic is NOT created (findings are pushed to jira, but not to the epic since it does not exist)
- Move some unit test helper functions to correct places
- Make relevant logging debugging instead of info
- Fix the express jira creation. It used a similar helper function I needed for the other unit tests so I thought why not. There is no test for this as I don't know how to simulate correct credentials there. That will be for another PR anyways.

mtesauro (approved):
Approved

valentijnscholten:
Thanks, test really help to ensure that it works and keeps working. One thing I think is important is that now the epic is created from inside the unit test. So outside dojo. We should test the scenario where dojo creates the epic using the normal way, so the  UI or the API. Easiest to do this is via the API, but I just noticed the epic mapping is not implemented in the API. you could look at the tests in test_import_reimport,py to see how the UI can be used from the unit tests, but it would be better to also implement the epic mapping in the API when engagements are added / modified.

I also noticed that if a JIRA issue was pushed from dojo to JIRA before the epic mapping was enabled, it is not added to the epic. Is this expected behaviour?

In the webhook there is commented out behaviour on the handling of epics. Should that also be reinstated, or removed?

I think users may have the same questions, so we should also document the epic/engagement mapping describing these cases.

dojo/jira_link/helper.py
```diff
@@ -807,7 +814,15 @@ def update_epic(engagement):
...
 def add_epic(engagement):
-    logger.info('trying to create a new jira EPIC for %d:%s', engagement.id, engagement.name)
+    add_epic_sync(engagement)
+
+
+# This code was not being tested, but calling it
+# as a task was not working. Using block execution
+# would not either here because this method is called
+# explicitly instead of through a save function
+def add_epic_sync(engagement):
```
valentijnscholten, line 824: this method is only here to make the unit test work which is a sign we should change that. let me know if you need help, I can explain what I mean.

valentijnscholten (changes requested):

madchap:
I support to have the jira epic mapping in the API to make the JIRA integration more robust. I believe that many people rely on that JIRA integration and after the recent effort done to make it better and more stable, it makes sense to probably put in that extra effort :+1:

Maffooch:
>  One thing I think is important is that now the epic is created from inside the unit test. So outside dojo. We should test the scenario where dojo creates the epic using the normal way, so the UI or the API. 

The epic is created from within the unit test yes, but it's using the exact same function the UI would use. So I believe it using the normal way it is created.

> it would be better to also implement the epic mapping in the API when engagements are added / modified.

This is a massive undertaking to an already large PR.  This should definitely be broken into multiple PR's. In the event I break something (which is generally a guarantee) it would be best to have this effort modularized so that tracking down bugs or reverting the PR would be much less invasive. 

> also noticed that if a JIRA issue was pushed from dojo to JIRA before the epic mapping was enabled, it is not added to the epic. Is this expected behaviour?

Yes this is expected behavior. Epic mapping only occurs if an epic is created and associated to an engagement. This does not need to happen at engagement creation as it can happen at any point in time (similar to pushing an issue to jira). If a user decides they want to create an epic mapping after issues have already been pushed from Dojo to jira, then they will need to enable epic mapping, create an epic on the desired engagement, and simply repush the findings they want to be mapped to that engagement. This functionality is the original design and not something new. 

> In the webhook there is commented out behaviour on the handling of epics. Should that also be reinstated, or removed?

I did not see that part. It will need to be reinstated yes, but should be a separate PR for the same reasoning as the API implementation above. 

> I think users may have the same questions, so we should also document the epic/engagement mapping describing these cases.

Agreed. Will make a PR there too.

> I support to have the jira epic mapping in the API to make the JIRA integration more robust. I believe that many people rely on that JIRA integration and after the recent effort done to make it better and more stable, it makes sense to probably put in that extra effort 👍

Agreed that would be nice, but in a future PR.

valentijnscholten:
> > One thing I think is important is that now the epic is created from inside the unit test. So outside dojo. We should test the scenario where dojo creates the epic using the normal way, so the UI or the API.
> 
> The epic is created from within the unit test yes, but it's using the exact same function the UI would use. So I believe it using the normal way it is created.
In the UI flow there is an extra method with 140 lines of code to handle the form post and parameters. This is usually where things go wrong. Or went wrong before we added unit tests for the JIRA config.

> 
> > it would be better to also implement the epic mapping in the API when engagements are added / modified.
> 
> This is a massive undertaking to an already large PR. This should definitely be broken into multiple PR's. In the event I break something (which is generally a guarantee) it would be best to have this effort modularized so that tracking down bugs or reverting the PR would be much less invasive.
It's just one form post and it is already implemented in https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/unittests/test_jira_config_engagement.py. The only thing needed is to uncomment the lines around epic mapping.

If you insist on doing it from the unit test, there's an example to impersonate a user so the extra 'sync' method is not needed: https://github.com/DefectDojo/django-DefectDojo/blob/1238d9b9ea20d17d8dedc482dbeb7a6572abb11d/dojo/unittests/test_deduplication_logic.py#L135


> 
> > also noticed that if a JIRA issue was pushed from dojo to JIRA before the epic mapping was enabled, it is not added to the epic. Is this expected behaviour?
> 
> Yes this is expected behavior. Epic mapping only occurs if an epic is created and associated to an engagement. This does not need to happen at engagement creation as it can happen at any point in time (similar to pushing an issue to jira). If a user decides they want to create an epic mapping after issues have already been pushed from Dojo to jira, then they will need to enable epic mapping, create an epic on the desired engagement, and simply repush the findings they want to be mapped to that engagement. This functionality is the original design and not something new.
From what I can see repushing the findings will not add them to the epic. They have to be manually added to the epic in JIRA.

Maffooch:
> In the UI flow there is an extra method with 140 lines of code to handle the form post and parameters. This is usually where things go wrong. Or went wrong before we added unit tests for the JIRA config.

Interesting. I didn't see that happening.  The jira engagement config tests can run that code if I could get one to work. 

>It's just one form post which is already implemented in https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/unittests/test_jira_config_engagement.py. The only thing needed is to uncomment the lines around epic mapping. And 1 or 2 asserts to check that the epic was created.
>If you insist on doing it from the unit test, there's an example to impersonate a user so the extra 'sync' method is not needed:

So I tried the test and keep failing both the product and engagement tests, but I cant figure out why. Jira projects cant be found for the engagement part, and the add_epic function does not even run on the product part. Spent about 90 minutes tracing everything and trying to figure out where the issue lay, but still coming up blank. Here are some logs with all the notification junk stripped out of it.
[test_logs.txt](https://github.com/DefectDojo/django-DefectDojo/files/5930788/test_logs.txt)

>From what I can see repushing the findings will not add them to the epic. They have to be manually added to the epic in JIRA.

Not sure how that did not make it into the commits?? Added now fore sure. Thanks for catching that.

valentijnscholten:
Ok, thanks for trying the suggestions. I'll see if I can add those tests later.

valentijnscholten:
I modified the tests a little bit so the epic creation is now tested fully in the unit tests. Turns out there was a bug in the UI part I mentioned. So sorry for pushing you to write tests, but my feeling got confirmed with this (small) bug.
So with existing findings now being added to the epic on repushing all items are resolved. If we accept that the API for this is for a next PR.

BTW I couldn't push to your branch so I created a PR on your repo:https://github.com/Maffooch/django-DefectDojo/pull/2

valentijnscholten (approved):

PR #3784, opened 2021-02-03
DSOP parser: remove pandas and fix twistlock CVE

damiencarol:
Refactor DSOP parser to not use pandas.

Fix one bug that don't detect CVE for Twistlock CVE data.

Need https://github.com/DefectDojo/django-DefectDojo/pull/3763 first

### DIFF
--- a/dojo/tools/dsop/parser.py
+++ b/dojo/tools/dsop/parser.py
@@ -1,141 +1,180 @@
 import re
-
-import pandas as pd
+from openpyxl import load_workbook
 
 from dojo.models import Finding
-
-__author__ = 'Matt Sicker'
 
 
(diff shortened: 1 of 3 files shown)


madchap:
fyi @yilmi

madchap (approved):
:+1: I'll give it whirl, but as I mentioned on Slack, I may have been the sole user of this parser at some point, and we may just want to remove altogether. But since you've put in the effort, I will also put some on my end as soon as I have 10mn :-)

damiencarol:
@madchap should we ask the question to the maintainers? (to keep it or not)

valentijnscholten (approved):

PR #3786, opened 2021-02-03
Improve GitLab dependency scanning parser

macedogm:
1. Improve GitLab dependency scanning parser, adding component name and version for the findings.
2. Removed SAST components, since they are not present in this kind of report.

Current findings from GitLab dependency scanning have this structure:
image

Now they will appear as:
image

### DIFF
--- a/dojo/unittests/tools/test_gitlab_dep_scan_parser.py
+++ b/dojo/unittests/tools/test_gitlab_dep_scan_parser.py
@@ -23,14 +23,17 @@ class TestGitlabDepScanReportParser(TestCase):
 
     def test_parse_file_with_one_vuln_missing_component_has_one_finding(self):
         testfile = open(
-            "dojo/unittests/scans/gitlab_dep_scan/gl-dependency-scanning-report-1-vuln-missing-component.json"
+            "dojo/unittests/scans/gitlab_dep_scan/gl-dependency-scanning-report-2-vuln-missing-component.json"
         )
         parser = GitlabDepScanReportParser()
         findings = parser.get_findings(testfile, Test())
-        self.assertEqual(1, len(findings))
+        self.assertEqual(2, len(findings))
         finding = findings[0]
         self.assertEqual(None, finding.component_name)
         self.assertEqual(None, finding.component_version)
+        finding = findings[1]
+        self.assertEqual("golang.org/x/crypto", finding.component_name)
+        self.assertEqual("v0.0.0-20190308221718-c2843e01d9a2", finding.component_version)
 
     def test_parse_file_with_multiple_vuln_has_multiple_findings(self):
         testfile = open(
--- a/dojo/unittests/tools/test_gitlab_dep_scan_parser.py
+++ b/dojo/unittests/tools/test_gitlab_dep_scan_parser.py
@@ -21,6 +21,14 @@ class TestGitlabDepScanReportParser(TestCase):
         findings = parser.get_findings(testfile, Test())
         self.assertEqual(1, len(findings))
 
+    def test_parse_file_with_one_vuln_missing_component_has_one_finding(self):
+        testfile = open(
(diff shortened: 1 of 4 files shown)


valentijnscholten (changes requested):
thanks. can you add some checks to make the parser not raise an expcetion if these fields aren't present in the report?

macedogm:
Thanks for the quick feedback @valentijnscholten .
That's an interesting point that I was thinking about. According to GitLab's Dependency Scanning [documentation](https://docs.gitlab.com/ee/development/integrations/secure.html#dependency-scanning), those fields must be present, otherwise the finding would not make much sense.
Do you still find prudent to add the checks anyway?

damiencarol:
also as you are adding more data to findings, could you add more checks in the unit tests?

valentijnscholten:
> Thanks for the quick feedback @valentijnscholten .
> That's an interesting point that I was thinking about. According to GitLab's Dependency Scanning [documentation](https://docs.gitlab.com/ee/development/integrations/secure.html#dependency-scanning), those fields must be present, otherwise the finding would not make much sense.
> Do you still find prudent to add the checks anyway?

Yes. Everytime I agree not to add any checks, the parser has failed 3 weeks later and the PR author was gone and we had to deal with the slack support, github issues, bugfix PRs etc. The parsers should really be made fool proof otherwise we can't keep up.

macedogm:
@valentijnscholten Can you please reopen the PR?
I was about to add the checks and test data. Just didn't had enough time to update before you closed it.

valentijnscholten:
I didn't mean to close it, sorry. Must have misclicked or something. Sorry!

macedogm:
No worries @valentijnscholten and thanks for reopening the PR.
I've added the checks and also a unit test. Please let me know if everything is as expected or not.

Additionally, could you also review the PRs https://github.com/DefectDojo/Documentation/pull/173 and https://github.com/DefectDojo/sample-scan-files/pull/69 to update the documentation regarding the new integration and to add the sample scan files, respectively, please?

valentijnscholten (approved):

valentijnscholten:
thanks for the PR, just one more approval needed....

dojo/unittests/tools/test_gitlab_dep_scan_parser.py
```diff
@@ -21,6 +21,14 @@ def test_parse_file_with_one_vuln_has_one_finding(self):
...
         self.assertEqual(1, len(findings))
 
+    def test_parse_file_with_one_vuln_missing_component_has_one_finding(self):
+        testfile = open(
+            "dojo/unittests/scans/gitlab_dep_scan/gl-dependency-scanning-report-1-vuln-missing-component.json"
+        )
+        parser = GitlabDepScanReportParser()
+        findings = parser.get_findings(testfile, Test())
+        self.assertEqual(1, len(findings))
+
```
damiencarol, line 31: Could you have something like that?
```suggestion
        finding = findings[0]
        self.assertEqual("...", finding.component_name)
        self.assertEqual("...", finding.component_version)
        <... mode tests? ...>
```

macedogm: Can you please review @damiencarol ?
Hope that I understood what you asked for. Otherwise, let me know if it needs improvement (I'm not very proficient in Python 😬).

damiencarol: it's good but you only check 1 finding without component data.

damiencarol: add another one with component data plz

macedogm: Done! One finding has no components and the other has both name and version. PTAL.

damiencarol (approved):

PR #3790, opened 2021-02-03
Security fixes for reports/notes/enpoints via APIv2

StefanFl:
### Issue
[CWE-284](https://cwe.mitre.org/data/definitions/284.html)

The API v1 and v2 are lacking appropriate access controls.

### Impact
Both APIs v1 and v2 allowed to retrieve findings, endpoints and other data for which a user was not authorized. The main endpoints affected are the ones to generate report, retrieval of notes, retrieval of endpoints as well as nested objects inside endpoints.

### Patches
We have added additional authorization checks to fix this in APIv2. APIv1 being deprecated and disabled by default since 1.12.0, it shoud no longer be used. APIv1 has not been and will not be fixed. We recommend to all users to switch to APIv2 for continued support and security fixes, and to report bugs or missing features to the project.

### Workarounds
* Disable untrusted users.
* Disable APIs. APIv1 is disabled by default since 1.12.0. APIv2 can be disabled by manually editing the configuration (not supported).

### For more information
If you have any questions or comments about this advisory:
* Visit our #defectdojo Slack channel via: https://owasp-slack.herokuapp.com/
* For sensitive issues, follow the security advisory process https://github.com/DefectDojo/django-DefectDojo/security/policy (report on hacker one)

### DIFF
--- a/dojo/api_v2/permissions.py
+++ b/dojo/api_v2/permissions.py
@@ -1,40 +1,6 @@
 from rest_framework import permissions
 
 
-class UserHasProductPermission(permissions.BasePermission):
-    """
-    @brief      To ensure that one user can only access authorized project
-    """
-    def has_object_permission(self, request, view, obj):
-        return request.user in \
-            (obj.authorized_users.all() | obj.prod_type.authorized_users.all()) or \
-            request.user.is_staff
-
-
-class UserHasReportGeneratePermission(permissions.BasePermission):
-    """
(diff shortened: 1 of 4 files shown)


valentijnscholten (approved):
did some tests after latest revisions and works ok. also tested succesfully using a non staf non superuser with limited model persmissions in django admin (and authorized on some products) and import/reimport works in that scenario.

madchap (approved):

Issue #3795, opened 2021-02-04
Burp XML parser: blank responses break the parser


damiencarol:
If you import a Burp report in XML with blank response, it crash the import when the finding is saved.

damiencarol:
Part of `1.13.0` release

PR #3796, opened 2021-02-04
Fix Burp blank response bug #3795

damiencarol:
When you import a Burp XML report with blank response it crash the save of the findings during the save of Request/response part.

Here:
https://github.com/DefectDojo/django-DefectDojo/blob/e90562e75d334d8c5fddb23bce15398e51c34615/dojo/engagement/views.py#L642

This pull request fix that and improve the Burp parser.
- [x] support blank request/response
- [x] support CWE
- [x] support `unique_id_from_tool` and `vuln_id_from_tool` attributes
- [x] improve the way we manage endpoints (no more DB access)
- [x] add unit tests to check some attributes


valentijnscholten (approved):
I like the use of unsaved endpoints so dojo core can decide the best way to create/attach them.

damiencarol:
yeah. I'm refactoring every parser to use the same workflow. And doing that it **reduce the number of DB call** in the parser code.

madchap (approved):
Refactoring machine.

PR #3802, opened 2021-02-05
anchore parser: Fix image_digest/imageDigest error

damiencarol:
Some recent version of anchore had renamed one attribute `imageDigest` to `image_digest`

This break the current parser.

### DIFF
--- a/dojo/tools/anchore_engine/parser.py
+++ b/dojo/tools/anchore_engine/parser.py
@@ -1,59 +1,45 @@
-__author__ = 'jaguasch'
-
 import json
-from datetime import datetime
 
 from dojo.models import Finding
 
 
 class AnchoreEngineScanParser(object):
     def get_findings(self, filename, test):
-        tree = filename.read()
-        try:
-            data = json.loads(str(tree, 'utf-8'))
-        except:
-            data = json.loads(tree)
+        data = json.load(filename)
         dupes = dict()
-        find_date = datetime.now()
-
         for item in data['vulnerabilities']:
-            categories = ''
-            language = ''
-            mitigation = ''
-            impact = ''
-            references = ''
-            findingdetail = ''
-            title = ''
-            group = ''
-            status = ''
-            cve = None
+            cve = item.get('vuln')
 
             title = item['vuln'] + ' - ' + item['package'] + '(' + item['package_type'] + ')'
 
-            if item['vuln']:
-                cve = item['vuln']
-
             # Finding details information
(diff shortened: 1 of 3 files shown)


dojo/tools/anchore_engine/parser.py
```diff
@@ -1,59 +1,45 @@
...
-            impact = ''
-            references = ''
-            findingdetail = ''
-            title = ''
-            group = ''
-            status = ''
-            cve = None
+            cve = item.get('vuln')
 
             title = item['vuln'] + ' - ' + item['package'] + '(' + item['package_type'] + ')'
```
madchap, line 13: This could then also read `cve` and follow the changes you made to use `item.get()`?

dojo/tools/anchore_engine/parser.py
```diff
@@ -1,59 +1,45 @@
...
-            language = ''
-            mitigation = ''
-            impact = ''
-            references = ''
-            findingdetail = ''
-            title = ''
-            group = ''
-            status = ''
-            cve = None
+            cve = item.get('vuln')
```
madchap, line 11: I think sometimes, there is `vuln` actually. Not sure anymore, but I would ensure this does not bomb.

madchap (changes requested):
See comments. Also, Anchore Enterprise 3.0 just came out, we may have to revisit the parser soon, so thanks for the cleanup!

dojo/tools/anchore_engine/parser.py
valentijnscholten: no bombing please

damiencarol: No. can't happen. Vuln is the "id" of the data. Anchore use this field to link it to a CVE.
This specific line don't crash if vuln is not here compared to previous code with [] everywhere.

dojo/tools/anchore_engine/parser.py
damiencarol: I don't want to rewrite entirely the parser.
I made a lot of fixes and code cleaning with this PR, I think this version is better.

damiencarol:
I made a lot of fixes and code cleaning with this PR, I think this version is better and support both version of Anchore (prior to 2.x and > 2.x).
I don't want to rewrite entirely the parser.
When Anchore is out, I could revisit and add a report for this version but for now, we have a breaking bug that this PR show and fix.
I prefer to merge this out and revisit later if Anchore 3.x really break it.

madchap (approved):

valentijnscholten (approved):

PR #3803, opened 2021-02-05
Add Close Engagement Notification

Maffooch:
Currently, the only way to be notified if an engagement has been closed is to enable notifications for all findings to be closed as well since they both fell under the "other" notification type. This PR provides a solitary notifications for when an engagement closes.


valentijnscholten:
first reaction would be that it might be interesting for the personal notifications on the view_product_defails page to allow people to only get these notifications for select products?

Maffooch:
Not a bad idea. Will look into it.

Maffooch:
Not totally sure what happened on this one..

madchap (approved):
lgtm - but you need to modify the db migration now, another PR went in before yours.

mtesauro (approved):
Approved

PR #3804, opened 2021-02-05
node/yarn: update to v14 and 1.22.10

valentijnscholten:
node v14 is the current/active LTS version, so we should start using that.
also bumps yarn to the latest version
includes a small optimization to only update the newly added APT repositories and not update ALL apt repositories again

### DIFF
--- a/Dockerfile.integration-tests
+++ b/Dockerfile.integration-tests
@@ -19,7 +19,6 @@ RUN \
 # Installing Google Chrome browser
 RUN \
   curl -sS -o - https://dl.google.com/linux/linux_signing_key.pub | apt-key add && \
-  # only update chrome repo, not all other repo's again
   apt-get -y update && \
   apt-get -y install \
     google-chrome-stable \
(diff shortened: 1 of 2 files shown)


damiencarol:
@valentijnscholten I think we should switch to a builder image based on `node` Docker image.
Installing an maitaining these tools by hand is not ok.

valentijnscholten:
I am not married to the current "hacky way" of installing node and yarn, but also see no harm in updating to the active LTS. But if you want to refactor the Dockerfile now instead of this PR, also fine for me.

madchap:
I am fine whatever way. Probably @damiencarol is right, but might need more testing and time not at hand? Up to you.

valentijnscholten:
what do you want to test? the only we thing we use npm for is for generating the javascript files which are tested by the test suite.

there's no need to not update node even if we want something different in the future. we also keep upgrading bootstrap, jquery, etc even though at some point we want a single page application with React or something else.

damiencarol (approved):
let's merge this one and refactor as Docker build image later.

madchap (approved):

PR #3809, opened 2021-02-06
feat(doc): integrate documentation in main dojo repo

alles-klar:
Moves documentation from https://github.com/DefectDojo/Documentation to this repository. Furthermore hugo will be uses to build the website and it will be hosted with github pages.


valentijnscholten:
I think we should just let de action run on every push to dev / master. Is there really a need for another label?
Not sure how GiHub pages / hugo works with versions, but would be nice to also generate docs for dev.

valentijnscholten (approved):

damiencarol (approved):

madchap (approved):
Let's do this and go through the small things the automated migrations could not handle. Thanks @alles-klar for putting this together :)

Issue #3800, opened 2021-02-05
Datatables have paging and search limitations


madchap:
**Bug description**
On the product list page (with a bit less than 50 products):

`Showing 1 to 25 of 25 entries`

plus a link to page 2. Clicking it I go to /product?page=2 which says

`Showing 1 to 21 of 21 entries`

Beyond this, the filter also only works for the displayed findings only, and not the whole set.

**Expected behavior**
Showing 1 to 25 of 46 entries
Showing 26 to 46 of 46 entries

**Deployment method** *(select with an `X`)*
- [x] Kubernetes
- [ ] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
Latest `dev`

madchap:
I have not looked at alternatives, but maybe there is better. Who's our best JS guy out there? ;-)

valentijnscholten:
I saw the same, been like this for a long time I think. The numbers being wrong is not related to datatables. We are doing paging and outside database and searching/filtering both inside and outside. datatable can handle AJAX calls to retrieve data and let server side do the filtering/search. not sure if it can handle paging correctly in that case.

bit dramatic title, no? :-)

valentijnscholten:
I have fixed the "showing entries X to Y of Z" in an upcoming PR.

PR #3812, opened 2021-02-06
add paging + filter to list of tests, engagements. similar findings

### LINKED ISSUE #3765: Dropboxes should always have the values listed in consistent order
Currently in full risk acceptance (might be elsewhere too, didn't pay attention), the "Owner" dropdown box lists users in whatever order. It's super hard to find the name you're looking for.

The list should be ordered in alphabetical order to facilitate finding the value.

Also, it should not list users who no longer have any rights (e.g. basically disabled users).

valentijnscholten:
This PR:
- adds paging and filtering for engagements inside the product view
- adds paging and filtering for tests inside the engagement view inside the product view
- adds paging and filtering for similar findings and show total number of similar findings if > 25
- use select2 as alternate jquery widget for multiselects in similar findings (experiment)
- better clear/restart for similar findings filter
- add (?) help text for similar findings
- add paging to search results (findings only and still limited by watson performance to 100 results max)
- add more help text to search function

Cleanups along the way:
- always use Dojo_User in filters to get fullname
- always sort Dojo_User (by default) by firstname (fixes #3765)
- removed some old code in view_engagement around finding, start_date, etc.
- add some prefetching in places where it was needed
- fix green/red misaligned engagement icon in product list
- fix "showing entries X to Y of total Z" as reported in #3800 
- render the filter div straight from html instead of adding attirbutes later via javascript
- open filter div only when related parameters are in the request instead of always opening it even when pagesize was changed, but no filtering happened
- support multiple filtering divs on the same page
- remove spinners / up+down arrows on number fields on filter/forms

Note:
- the PR modifies an existing migration. This is not harmful as it just adds an ordering clause which only affects python and has no effect on the database schema.

Screenshots:
image
image


dojo/api_v2/serializers.py
```diff
@@ -1312,9 +1359,7 @@ def save(self, push_to_jira=False):
             scan_date_time = timezone.make_aware(scan_date_time, timezone.get_default_timezone())
         verified = data['verified']
         active = data['active']
-        version = None
-        if 'version' in data:
-            version = data['version']
+        version = data.get('version', None)
```
damiencarol, line 1362: None is the default value when the key is not present so:
```suggestion
        version = data.get('version')
```

valentijnscholten: not very useful to comment on draft PRs with old code that is already changed elsewhere.

madchap (approved):
Pulled down on my prod-like staging env, looks good so far.

damiencarol (approved):

PR #3813, opened 2021-02-06
Make mitigated date editable

### LINKED ISSUE #3584: Make mitigated date editable
We are using the mitigated date field of a finding to track time to resolve of an issue. We currently can not rely on this measurement because we are either not diligent enough to mark finding mitigated  immediately when it is mitigated, or we are importing past data. I would note that we mostly use DefectDojo manually, without automation.

Make the field editable via user interface and API.

Currently we parse this from Jira where we can and track this externally, but it is only applicable where Jira issue for issue mitigation exists. 

**Additional context**
/

greginvm:
Hi,

this one adds the ability to change mitigated date and mitigated by. Only available when `EDITABLE_MITIGATED_DATA` is `True` and user is superuser. Closes #3584 

I have used browsers native date and time pickers. "Mitigated" and "Mitigated by" fields auto-hide/appear depending whether finding is marked as _active_ and whether setting + permission are set accordingly. "Mitigated" and "Mitigated by" behaves as it behaved before - it is set automatically if it is not set by a user, additionally any modifications are accepted. For example, if time is empty but "mitigated by" is set, time will be automatically set to current one. "Mitigated by" is required and by default current user is selected.

Screenshot from 2021-02-07 21-35-39

Screenshot from 2021-02-07 21-26-24


damiencarol:
@greginvm reviewed some part of your pull request and fixed conflicts. It's hard for me to validate fully because I'm not a Django expert. Will reach the other mods to have feedbacks/review.

valentijnscholten:
Thanks for the PR. I thought we were using a bootstrap datepicker and was hoping for something similar for this case as said before. Turns out we're using the django datepicker and the django way of turning it into a datetimepicker requires glue code and looks like ssssssssss with those things separated. On the otherhand it's hidden behind a feature flag for corner cases anyway, so might be OK to merge as is.

greginvm:
Hey, yes - it turned out that another library would be needed, separate, and would make things different from current approach anyway. So to get a MVP I have opted to built-ins. I am willing to extend this feature with other UI libs if need be in a separate pr.

valentijnscholten:
I think we can leave it as is for now as it's behind a feature flag, disabled by default.

valentijnscholten (approved):

madchap (approved):

PR #3814, opened 2021-02-07
Add parser for Anchore Grype scan

damiencarol:
This pull request add a new parser for Anchore Grype, the scanner for Anchore.

https://github.com/anchore/grype

Grype is a vulnerability scanner for container images and filesystems. Easily install the binary to try it out.


valentijnscholten:
so you don't like the unit tests for empty reports anymore? :-D
I think they were created to make sure that if empty reports are coming in, the parser doesn't crash. Similar to that we want to make sure we don't crash on missing nodes / elements / attributes.

damiencarol:
I can try to generate an empty one. Let me add it.

madchap:
So I have not tried or looked hard, but per your previous parser refactoring, that's all that's needed now to have parser? We probably need to update the parser doc.

damiencarol:
yes. I tested it with ptvsd configuration and some real reports of the tool (version `0.7.0`).
I'm working on the second part of the dynamic infrastructure. after this part is done, the factory will not need modifications when adding a new parser.

valentijnscholten (approved):

Maffooch:
There's a couple changes that need to be made that were missed. See #2287 for details.

madchap:
Actually, i think grype will/is leveraged at the decentralized way to gather findings too in their enterprise v3?

Good job anyways :)

madchap (approved):

madchap:
@Maffooch the whole parser stuff has been completely overhauled in https://github.com/DefectDojo/django-DefectDojo/pull/3689

Also need to update the doc. @damiencarol could you amend eventually?

valentijnscholten:
Are the docs and sample repo updated?

damiencarol:
@valentijnscholten for the samples https://github.com/DefectDojo/sample-scan-files/pull/71 . Doc is comming

PR #3819, opened 2021-02-08
Show boolean fields in product types as icons

StefanFl:
This PR makes the UI for product types nicer. The attributes for critical product and key product have been shown as True or False in the list of product types and when viewing a single product type. With this PR they are shown as a red cross or a green tick, similar to the boolean fields in the list of users.

### DIFF
--- a/dojo/templates/dojo/product_type.html
+++ b/dojo/templates/dojo/product_type.html
@@ -1,5 +1,7 @@
 {% extends "base.html" %}
 {% load navigation_tags %}
+{% load display_tags %}
+
 {% block content %}
     <div class="row">
         <div class="col-md-12">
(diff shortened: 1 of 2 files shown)


valentijnscholten:
you have screenshot? saves me from building etc :-)

StefanFl:
Here they are:

2021-02-08 18_28_47-Product Type List _ DefectDojo
2021-02-08 18_29_03-View Product Type _ DefectDojo

valentijnscholten:
thanks. maybe I'm just nitpicking but the red cross might induce the feeling "something is wrong", but not sure if there's a better icon :-)

StefanFl:
These seem to be the standard icons. I googled for "icons true false" and you get the green tick and the red cross. And it's consistent with the list of users.

valentijnscholten (approved):
approved by the international red cross!

Maffooch (approved):

PR #3822, opened 2021-02-08
Fix deduplication hashcode fields for Dependency Track scanner

svdm:
The Dependency Track scanner was configured to use `'component'` in its dedupe hashcode, but that field does not exist and is not in `HASHCODE_ALLOWED_FIELDS`. As a result deduplication falls back to default fields, with suboptimal results. Most likely `component` was a typo of `component_name` here (could also add the component version like other scanners, but [as the parser documents](https://github.com/DefectDojo/django-DefectDojo/blob/08bc2f906b34cd6f76f5abc4e5e6d801fa556bbd/dojo/tools/dependency_track/parser.py#L130), DT does not always supply a version).


valentijnscholten (approved):
thanks, added a label as this PR means people will need to recalculate their hashcodes using `dedupe` management command.

Maffooch (approved):

Issue #3824, opened 2021-02-08
Display correct engagement count for staff users on home dashboard


michaelgibson:
Currently the total number of active engagements rendered on the home page dashboard is restricted to either superuser or engagement_lead.
My understanding is that "staff" level access should return the total count of engagements similar to superuser.
Something similar to this seems to correct the issue for me. May or may not be missing other updates.
Diff not rendering.
Submitting PR:
https://github.com/DefectDojo/django-DefectDojo/pull/3826

valentijnscholten:
Can you make a PR?

michaelgibson:
Closing. resolved'ish in #3826 :)

PR #3826, opened 2021-02-08
fixing #3824

michaelgibson:


valentijnscholten:
is there a reason why you have a staff user that is not a superuser? the staff user is mainly meant for accessing the '/admin' portal. I think most checks / queries use the superuser property to distinguish between regular users and "admins".

michaelgibson:
Here is what we see for is_staff users when browsing to /admin
image

We use is_staff for our greater Security team be able to access all projects and vulns for remediation process but not have Application Admin access(adding/removing users).

Looks like the other finding views on dashboard are affected by this as well(for is_staff only user)
This PR currently only addresses the engagements stats:

"Finding Details Last Seven Days": 0 (should be 28)
"Finding Details Closed In Last Seven Days": 0 (should be 2)
"Finding Details Risk Accepted In Last Seven Days": Throws 500 error

- Looks like '&test__engagement__risk_acceptance__reporter=4800' gets appended when not superuser. Removing this and page loads correctly

stacktrace from 500:
<pre><code>
ERROR:django.request:Internal Server Error: /finding/accepted
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
  File "/usr/local/lib/python3.6/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "./dojo/finding/views.py", line 102, in accepted_findings
    django_filter=accepted_findings_filter)
  File "./dojo/finding/views.py", line 157, in findings
    title_words = get_words_for_field(findings_filter.qs, 'title')
  File "/usr/local/lib/python3.6/site-packages/django_filters/filterset.py", line 243, in qs
    qs = self.filter_queryset(qs)
  File "/usr/local/lib/python3.6/site-packages/django_filters/filterset.py", line 230, in filter_queryset
    queryset = self.filters[name].filter(queryset, value)
  File "/usr/local/lib/python3.6/site-packages/django_filters/filters.py", line 248, in filter
    qs = self.get_method(qs)(q)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py", line 892, in filter
    return self._filter_or_exclude(False, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/query.py", line 910, in _filter_or_exclude
    clone.query.add_q(Q(*args, **kwargs))
  File "/usr/local/lib/python3.6/site-packages/django/db/models/sql/query.py", line 1290, in add_q
    clause, _ = self._add_q(q_object, self.used_aliases)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/sql/query.py", line 1312, in _add_q
    current_negated, allow_joins, split_subq, simple_col)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/sql/query.py", line 1318, in _add_q
    split_subq=split_subq, simple_col=simple_col,
  File "/usr/local/lib/python3.6/site-packages/django/db/models/sql/query.py", line 1224, in build_filter
    self.check_related_objects(join_info.final_field, value, join_info.opts)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/sql/query.py", line 1084, in check_related_objects
    self.check_query_object_type(value, opts, field)
  File "/usr/local/lib/python3.6/site-packages/django/db/models/sql/query.py", line 1067, in check_query_object_type
    (value, opts.object_name))
ValueError: Cannot query "Test Local (test-local)": Must be "Risk_Acceptance" instance.
</pre></code>

valentijnscholten:
We are working on a new permission system, but still not sure if your usecase is clear. Superuser = users with access to all products. Staff = access to admin portal. But access to admin portal without superuser doesn't make much sense because you can't see anything.

madchap:
> Superuser = users with access to all products. Staff = access to admin portal.

No, that's the other way around.

valentijnscholten:
> > Superuser = users with access to all products. Staff = access to admin portal.
> 
> No, that's the other way around.

No it's not:

image
image

I always mix them up because they make no sense, but this is the django definition.

madchap:
Then try, I just did :) The labels are wrong.

madchap:
So as it turns out, you need both `staff` AND `superuser` to access admin. The labels are misleading.

madchap (approved):
I am fine with the PR and the use-case. Let's see what our friend Stefan comes up with and the new permission system anyways :-)

As staff-only and having access to all products, seeing these info on the dashboard makes sense.

valentijnscholten:
yes, but everything defect dojo is not designed like this so only fixing the dashboard makes it more confusing I think

devGregA (approved):

valentijnscholten:
I am changing the title as that is what appears on the release notes.

valentijnscholten:
I think everybody including me is/was confused what Django meant by this is_staff flag as I now see that in defect dojo the superuser is allow to modify the system setting. But in Django terms I would expect that permission to be for staff users.

image

So if we implement a new permission system, we should stay close to what people are used to with Django is my opinion.

PR #3827, opened 2021-02-08
Dynamic parser infrastructure part 2

damiencarol:
This part cover auto loading only.

As it modifies the contract between the parsers and the factory, all the parsers are modified.
But the modification is the same everywhere:
```Python

    def get_scan_types(self):
        return ["Wapiti Scan"]  # list of scan_type that this parser cover (in the dynamic infrastructure, parsers can support more than 1 type) 

    def get_label_for_scan_types(self, scan_type):
        return "Wapiti Scan"  # depending on the scan_type we could show a nice label

    def get_description_for_scan_types(self, scan_type):
        return "Import XML report"  # description that is pushed in UI
```

On the side of parsers we now have only few requirements:
* code of the parser **must** to be in a sub-folder of `dojo/tools/` (ex: `/dojo/tools/burp`) for the factory to take the parser into account
* there **must** be a module called `parser.py` in this folder
* in this module there must be a class with a name start with the module name and ends with `Parser` (ex: `BurpParser`)
* this class **must** implements these methods:
  - `get_scan_types` which return a list of scan_type
  - `get_label_for_scan_types` which return a string if given scan_type took from the previous method
  - `get_description_for_scan_types` which return a string if given scan_type took from the previous method
  - `get_findings` which return data when given a file (this one is the same old method from previous versions)

## Big example with Foo security tool

Our example is a new security tool called `IBM Foo scanner`

So we can choose an identifier like `ibm_foo` or directly `foo` (get ride of scanner/security/... words)
Let say we choose `foo`

1) create these files:
 - `dojo`
   - `tools`
     - `foo` 
       - `__init__.py` (you could let this one empty)
       - `parser.py`

2) create a class `FooParser` in file `dojo/tools/foo/parser.py`

```Python

import json
import logging

from dojo.models import Finding


class FooParser(object):

    def get_scan_types(self):
        return ["foo"]

    def get_label_for_scan_types(self, scan_type):
        return "`IBM Foo scanner"

    def get_description_for_scan_types(self, scan_type):
        return "`IBM Foo scanner export in JSON report format (option -j)"

        data = json.load(filename)

        dupes = dict()
        if "generated_at" in data:
            find_date = datetime.strptime(data["generated_at"], '%Y-%m-%dT%H:%M:%SZ')

        # in this example the findings are in node 'result'
        for item in data["results"]:  # in this example the findings are in node 'result'
        
            dupe_key = title + item["filename"] + str(item["line_number"])

            if dupe_key in dupes:
                find = dupes[dupe_key]
            else:
                dupes[dupe_key] = True

                find = Finding(title=title,
                               test=test,
                               active=False,
                               verified=False,
                               description=findingdetail,
                               severity=sev.title(),
                               numerical_severity=Finding.get_numerical_severity(sev),
                               mitigation=mitigation,
                               impact=impact,
                               references=references,
                               file_path=item["filename"],
                               line=item["line_number"],
                               url='N/A',
                               date=find_date,
                               static_finding=True)
                logging.debug(f"Bandit parser {find}")
                dupes[dupe_key] = find
                findingdetail = ''

        return list(dupes.values())  # return list of Finding object
```


valentijnscholten (approved):

madchap (approved):
Massive work, thanks for the refactoring!

PR #3839, opened 2021-02-10
finding.save() + add logging of the JIRA issue key to jira webhook

madchap:
Adding a few lines to help in debugging without full debug turned on.

As it turns out, the findings were not saved? I could reproduce that with `finding.save()` calls, the finding status did reflect the jira status, but not without.

### DIFF
--- a/dojo/jira_link/views.py
+++ b/dojo/jira_link/views.py
@@ -78,7 +78,7 @@ def webhook(request, secret=None):
 
                     if resolution is None:
                         resolved = False
-                    if finding.active == resolved:
+                    if finding.active is resolved:
                         if finding.active:
                             if jira_instance and resolution['name'] in jira_instance.accepted_resolutions:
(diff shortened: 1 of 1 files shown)


damiencarol (approved):

valentijnscholten:
In general I think some more logging is good for the webhook. Ideally on DEBUG level, but if we have some stability issues, INFO level can help everyone now.

The root cause seems to be that these 3 lines at 

https://github.com/DefectDojo/django-DefectDojo/blob/f567f53aeeb33404762fd44a28a7d18362cb9d68/dojo/jira_link/views.py#L126

were indented 1 step to the right, causing them to be no longer executed for findings that were active and are being closed in the webhook.

Although the extra saves are helping, it might be better to unindent these 3 lines again as they also set some timestamps correctly.

madchap:
Tested your suggestion, that was it. Good :eyes:  :+1:

valentijnscholten (approved):

PR #3843, opened 2021-02-11
Improvement: Product add/edit SonarQube configuration  visibility 

dsever:
Little improvement for adding and editing product, to not show SonarQube setting by default.

Collapsed: 

image

image

### DIFF
--- a/dojo/templates/dojo/edit_product.html
+++ b/dojo/templates/dojo/edit_product.html
@@ -30,9 +30,22 @@
             {% include "dojo/form_fields.html" with form=gform %}
         {% endif %}
         {% if sonarqube_form %}
-            <h4> SonarQube API </h4>
-            <hr>
-            {% include "dojo/form_fields.html" with form=sonarqube_form %}
+         <div class="panel-group" id="accordion" role="tablist" aria-multiselectable="true">
+          <div class="panel panel-default">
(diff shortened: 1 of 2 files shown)


valentijnscholten:
Thanks, but I think we need to think about this a bit more. If we decide sections should be collapsable, I think we should to so for all sections, so also JIRA and Gitlab.
The blue panels are typically used in defect dojo on the "view product/test/finding"  pages. The edit modes are all black(grey) and white. We also use a glyphicon there to open/close the collapsed panel.

Maffooch:
I think instead of the collapsable field here it should disappear completely without the existence of a "SonarQube" Tool Type. A user has to explicitly add that tool type anyways so that's sorta like enabling the sonarqube integration. That way it will be consistent with the Jira and Github integrations since they are not present unless enabled.

madchap:
Agreed with @Maffooch, let's not show it if there is no need - and also with @valentijnscholten to keep it consistent.

ptrovatelli (changes requested):
@dsever do you agree to apply the requested changes? (hide the panel disappear if sonar is not configured)
thanks!

dsever:
@valentijnscholten @madchap @ptrovatelli That sounds good as well, the problem I'm idiot when it comes to frontend dev (CSS and HTML). So I'm not sure that I can do it.

PR #3849, opened 2021-02-12
Publish the Helm chart tarball with each release

### LINKED ISSUE #3778: Publish the Helm chart tarball with each release
Before using the Helm chart we need to clone the git repository.

Instead, I would like Helm to download the Chart by itself.
This could be either from a chart repository or as a tarball published on the release page on Github.
Maybe using [chart-releaser](https://github.com/helm/chart-releaser)

bgoareguer:
Fixes #3778

Pre-requisites:
- The gh-pages branch needs to be created by hand
- Github Pages need to be activated for the django-DefectDojo repository with `gh-pages` as source branch and `/ (root)` as the folder

The Helm chart tarball is published as a release asset.
The `gh-pages` branch is used to store the Helm repository index. Each version of the chart points to the asset added to the corresponding release.

The resulting Helm repository should be added with the following command (not sure about the capital letters): 
```
helm repo add defectdojo https://defectdojo.github.io/django-defectdojo
```


madchap:
I took a look tonight, I'd rather we try to use the action I suggested in the other PR.

That being said, your PR would not be working either because you also need to add the other helm repos for the build to work at least.

Cheers and thanks for the feature idea :)

madchap (changes requested):
Needs at least helm repo add.

.github/workflows/new-release-tag-docker.yml
```diff
@@ -35,6 +35,17 @@ jobs:
         run: |
           git tag -a ${{ github.event.inputs.release_number }} -m "[bot] release ${{ github.event.inputs.release_number }}"
           git push origin ${{ github.event.inputs.release_number }}
+      - name: Install Helm
```
madchap, line 38: Could you please replace that stanza with the helm action, such as

```
      - name: Install Helm
        uses: azure/setup-helm@v1
        with:
          version: v3.4.0
```

for now?

.github/workflows/new-release-tag-docker.yml
```diff
@@ -35,6 +35,17 @@ jobs:
...
+        run: |
+          curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
+          chmod 700 get_helm.sh
+          ./get_helm.sh
+      - name: Package Helm chart
```
madchap, line 43: Before this step, you'd need to add the repos, such as:

```
- name: Configure Helm repos
        run: |-
             helm repo add stable https://charts.helm.sh/stable
             helm repo add bitnami https://charts.bitnami.com/bitnami
             helm dependency list ./helm/defectdojo
             helm dependency update ./helm/defectdojo
```

or the build will fail.

madchap (changes requested):

madchap:
I think this is more flexible and can offer what we'd want, compared to my testing done through https://github.com/DefectDojo/django-DefectDojo/pull/3923/files

If you could please make the suggested changes, I could test it further.

Cheers!

.github/workflows/new-release-tag-docker.yml
```diff
@@ -47,6 +63,34 @@ jobs:
...
+        id: upload-release-asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: ./build/defectdojo-${{ steps.package-helm-chart.outputs.chart_version }}.tgz
+          asset_name: defectdojo-${{ steps.package-helm-chart.outputs.chart_version }}.tgz
+          asset_content_type: application/octet-stream
+      - name: Update Helm repository index
```
madchap, line 76: I just learned about https://github.com/actions/upload-release-asset, so it might actually be better to also use it to upload it. The example workflow should nearly be a copy-paste to replace this step, and maybe a slight adjustment to the asset content type.

```      
      - name: Upload Release Asset
        id: upload-release-asset 
        uses: actions/upload-release-asset@v1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          upload_url: ${{ steps.create_release.outputs.upload_url }}
          asset_path: ./build/defectdojo-${{ steps.package-helm-chart.outputs.chart_version }}.tgz
          asset_name: defectdojo-${{ steps.package-helm-chart.outputs.chart_version }}.tgz
          asset_content_type: application/tar+gzip
```

madchap (changes requested):
I would think the last easy change :) Cheers.

.github/workflows/new-release-tag-docker.yml
bgoareguer: Sorry, I don't understand the other change you are requesting. `actions/upload-release-asset@v1` is already used to upload the chart tarball.

madchap: Nevermind, I got mixed up, sorry about that!

.github/workflows/new-release-tag-docker.yml
```diff
@@ -47,6 +63,34 @@ jobs:
...
+        id: update-helm-repository-index
+        run: |
+          git config --global user.name "${{ env.GIT_USERNAME }}"
+          git config --global user.email "${{ env.GIT_EMAIL }}"
+          git remote update
+          git fetch --all
+          git checkout gh-pages
+          git pull
+          if [ ! -f ./index.yaml ]; then
+            helm repo index ./build --url "https://github.com/DefectDojo/django-DefectDojo/releases/download/${{ github.event.inputs.release_number }}/"
```
madchap, line 86: Do you think there is a eventually better way than hard-coding the URL string?

bgoareguer: I've just pushed a modification to try to use environment variables but I can't check if the resulting URL is correct (I don't see the corresponding workflow in the Github interface...).

madchap (approved):
lgtm now, we'll see how it rolls and tune if needed :) Thank you!

.github/workflows/new-release-tag-docker.yml
madchap: Seems to be good to me.

alles-klar:
Hi @bgoareguer, 
a great Idea. I have two generell questions to your implementation: 

We are already using the gh-pages branch to publish our defectdojo documentation. I am not sure if it is a good idea to use this branch for publishing helm charts too. Can we use any other branch or is it not a problem?

There is already a predefined github action for publishing helm charts: https://github.com/stefanprodan/helm-gh-pages. Is there a reason for maintaining our own action?

bgoareguer:
Hi @alles-klar,

I didn't know about https://github.com/stefanprodan/helm-gh-pages. I think it can be reused here.

Github seems to allow only a single Github Pages site per Github project, so it is not possible to publish the Helm chart to a different branch. 
Another option would be to create a dedicated Github project to host the Helm chart. This would require to store (inside the django-defectdojo project settings) a Github token that can push code to the project hosting the Helm chart.
Here is an example: https://github.com/aquasecurity/trivy/pull/888

madchap:
> Hi @alles-klar,
> 
> I didn't know about https://github.com/stefanprodan/helm-gh-pages. I think it can be reused here.

I didn't know either, good find @alles-klar 

> 
> Github seems to allow only a single Github Pages site per Github project, so it is not possible to publish the Helm chart to a different branch.

With the above action, it seems you can. Per their README:

`branch The branch to publish charts, defaults to gh-pages`

bgoareguer:
> With the above action, it seems you can. Per their README:
> 
> `branch The branch to publish charts, defaults to gh-pages`

Sorry if I was unclear. Yes, the above action lets you publish the chart to any branch, but the restriction is on Github side: you can only have a single Github Pages site per project. You already configured Github Pages to serve the DefectDojo documentation from the `gh-pages` branch, so no other branch will be served by Github Pages.

bgoareguer:
I think that hosting the Helm chart in a dedicated project would be the cleanest solution.
What do you think?

madchap:
I just did a test without gh-pages (after reading an article), seems to work.

Checked out a [new branch](https://github.com/DefectDojo/django-DefectDojo/compare/dev...madchap:test-helm-repo-not-ghpages?expand=1) in my fork, packaged the last from `dev` today

```
$ helm package helm/defectdojo
$ helm repo index .
```

committed and pushed the resulting tgz and index.yaml file.

```$ helm repo add test-wo-ghpages 'https://raw.githubusercontent.com/madchap/django-DefectDojo/test-helm-repo-not-ghpages'
"test-wo-ghpages" has been added to your repositories

$ helm repo update
```

Search for defectdojo helm chart
```
$ helm search repo defectdojo
NAME                      	CHART VERSION	APP VERSION	DESCRIPTION                                      
test-wo-ghpages/defectdojo	1.5.1        	1.14.0-dev 	A Helm chart for Kubernetes to install DefectDojo
```

Could that be a viable option?

madchap:
Idea of actual name for the official branch: `defectdojo-charts`

alles-klar:
I like the idea @madchap! 

Naming: Do we want to push more then the helm package to this branch? If not I suggest to name the branch "helm-charts" because the name doesn't have to contain "defectdojo" - it is already in the name of this repository.

Reinvent the wheel: We can start with the solution from @bgoareguer but in the feature I prefer to the predefined action https://github.com/stefanprodan/helm-gh-pages. Less code to maintain and easier to read.

madchap:
> Reinvent the wheel: We can start with the solution from @bgoareguer but in the feature I prefer to the predefined action https://github.com/stefanprodan/helm-gh-pages. Less code to maintain and easier to read.

But incompatible with our documentation.

Let's start with that and see how it behaves for 1.14.0. I will make the small mods to the PR now.

alles-klar (approved):

alles-klar:
Ok, lets go. 

Anyhow, a short paragraph in the documentation would be nice!

madchap:
> Ok, lets go.
> 
> Anyhow, a short paragraph in the documentation would be nice!

Yep, pinged you on the companion PR ;-)

PR #3854, opened 2021-02-12
RENAMED: Finding.is_Mitigated to is_mitigated

valentijnscholten:
came across this annoying typo and decided to fix it. 
let me know what you think, might be nitpicking, might be good for readability!


damiencarol (approved):

madchap:
This would probably need a heads-up in the release notes. I - for one - am going to be impacted with regards to scripts outside of DefectDojo.

madchap:
Adding the `Breaking Changes` label. It's bending the label's arm a bit, dunno. Better suggestion?

madchap (approved):

valentijnscholten:
hmm thought might not be too big an issue for the API, but maybe we should do this in v2.0 where some other breaking changes might take place?

madchap:
Removing my approval for now to avoid a too quick merge. Let's take it on slack?

madchap:
Sounds like this one will need to make it in sooner to avoid potential future headaches...

Issue #3848, opened 2021-02-12
False positive history is not working with reimport-scan API


macedogm:
This is potentially a duplicate of https://github.com/DefectDojo/django-DefectDojo/issues/1584 .

The false positive history setting doesn't work as expected.

**Steps to reproduce**

1. In settings, check `False positive history` and uncheck `Deduplicate findings`.
2. Import a new report using the `Import Scan Results` page.
3. Flag a finding as `False Positive` and uncheck `Active` and `Verified`. See image 1 below.
4. Reimport the same or a new scan through the `/reimport-scan` API. Flags `active` and `verified` must be set to `true` while using the API. This scan must contain the same finding from item 3.
5. We would expect that the finding manually marked as `False positive` would remain inactive and as a false positive.
6. After the reimport is completed, the finding will be reactivated and have all three flags enabled - `Active`, `Verified` and `False Positive`.  See console logs below. See image 2 below.

**Console logs**
For some reason, the false positive history is considered as disabled -> `skipping false positive history because it's disabled in system settings or false_history param is False`.
```
uwsgi_1         | [12/Feb/2021 08:25:57] DEBUG [dojo.api_v2.serializers:1388] starting reimport of 45 items.
uwsgi_1         | [12/Feb/2021 08:25:57] DEBUG [dojo.api_v2.serializers:1442] 0: reactivating: 1:CVE-2019-0210: TEST Out-of-Bounds Read:None:None
uwsgi_1         | [12/Feb/2021 08:25:57] DEBUG [dojo.models:2056] finding.save() getting current user: admin
uwsgi_1         | [12/Feb/2021 08:25:57] DEBUG [dojo.models:2068] Saving finding of id 1 dedupe_option:False (self.pk is not None)
uwsgi_1         | [12/Feb/2021 08:25:57] DEBUG [dojo.decorators:129] model_or_id: CVE-2019-0210: TEST Out-of-Bounds Read
uwsgi_1         | [12/Feb/2021 08:25:57] DEBUG [dojo.decorators:56] converting model_or_id to id: CVE-2019-0210: TEST Out-of-Bounds Read
celeryworker_1  | [12/Feb/2021 08:25:57] INFO [celery.worker.strategy:157] Received task: dojo.utils.do_apply_rules[7686ceb1-c84b-4c24-8c76-2c0ad1d15e74]  
[...]

PR #3858, opened 2021-02-13
Reimport: keep false positive and out of scope history #3848

macedogm:
Fix issue https://github.com/DefectDojo/django-DefectDojo/issues/3848 regarding the reimport of reports where previous findings marked as false positive and/or out of scope are re-activated instead of ignored. Now the history of false positive and out of scope findings is properly preserved.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1438,8 +1438,8 @@ class ReImportScanSerializer(TaggitSerializer, serializers.Serializer):
                 if findings:
                     # existing finding found
                     finding = findings[0]
-                    if finding.false_p or finding.out_of_scope:
-                        logger.debug('%i: skipping existing finding (it is marked as false positive:%s and/or out of scope:%s): %i:%s:%s:%s', i, finding.false_p, finding.out_of_scope, finding.id, finding, finding.component_name, finding.component_version)
+                    if finding.false_p or finding.out_of_scope or finding.risk_accepted:
+                        logger.debug('%i: skipping existing finding (it is marked as false positive:%s and/or out of scope:%s or is a risk accepted:%s): %i:%s:%s:%s', i, finding.false_p, finding.out_of_scope, finding.risk_accepted, finding.id, finding, finding.component_name, finding.component_version)
                     elif finding.mitigated or finding.is_Mitigated:
                         logger.debug('%i: reactivating: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
                         finding.mitigated = None
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1438,7 +1438,9 @@ class ReImportScanSerializer(TaggitSerializer, serializers.Serializer):
                 if findings:
                     # existing finding found
                     finding = findings[0]
-                    if finding.mitigated or finding.is_Mitigated:
+                    if finding.false_p or finding.out_of_scope:
+                        logger.debug('%i: skipping existing finding (it is marked as false positive:%s and/or out of scope:%s): %i:%s:%s:%s', i, finding.false_p, finding.out_of_scope, finding.id, finding, finding.component_name, finding.component_version)
+                    elif finding.mitigated or finding.is_Mitigated:
                         logger.debug('%i: reactivating: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
                         finding.mitigated = None
                         finding.is_Mitigated = False
--- a/dojo/test/views.py
+++ b/dojo/test/views.py
@@ -776,8 +776,8 @@ def re_import_scan_results(request, tid):
 
                     if findings:
                         finding = findings[0]
-                        if finding.false_p or finding.out_of_scope:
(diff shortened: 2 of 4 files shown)


valentijnscholten:
Thanks. Please note that almost the same code is also located in `test/views.py`. Could you also add a few lines to the description of what problem is being solved?

macedogm:
Thanks @valentijnscholten . I'll review your comment. Right now I'm trying to fix the unit test, it's failing when it's executed simulating the UI (`FAIL: test_import_reimport_keep_false_positive_and_out_of_scope (dojo.unittests.test_import_reimport.ImportReimportTestUI)`). I'm trying to understand the logic behind this, since in the UI there is no way to re-import a report to the same `test_id`. Every import in the UI will generate a new test, with a new `test_id`. Right? Or am I missing something?

valentijnscholten:
You can reimport into the same test. Look for "re-upload" as it's called that in some places.

macedogm:
@valentijnscholten Thanks for pointing me to the right direction. It took me a while to follow through the code and understand the flow that is happening in `ImportReimportTestUI`, `test/views.py` and why the test was failing. Please let me know what do you think now.

macedogm:
ℹ️  Might make sense to merge https://github.com/DefectDojo/django-DefectDojo/pull/3854 before and then I update this MR accordingly.

valentijnscholten:
Thanks, this is a good improvement. Weird how nobody ever ran into this, might be that not many people are using reimport? We are improving reimport big time as I believe it is very powerful and useful.
Looking again at this I realize `risk_accepted`  should be added to this same if clause, agree?
For `duplicate` I am in doubt, might be that it needs to reactivated when needed.
And ideally we also test the opposite of the if, so that if a finding is not false_p/out_of_scope/risk_accepted, reactivation will be performed? In a way I think that is already tests by the pre existing test cases.

macedogm:
> We are improving reimport big time as I believe it is very powerful and useful.

For our team the reimport feature is super useful and important. We ran dozens of daily scans and we only noticed the issue when we marked some findings as false positives and they started to appear again. So I took the time to investigate and tried to fix this.
We were also hit by an issue related to the comment below during a reimport of a GitLab dependency scanning report (I submitted this parser a few weeks ago): https://github.com/DefectDojo/django-DefectDojo/blob/33fc63962b8e901bdf2b31bad735277e48951b1c/dojo/api_v2/serializers.py#L1420-L1432 . I also plan to submit a PR with a proposal to improve this during the week.


> Looking again at this I realize risk_accepted should be added to this same if clause, agree?

Yes, I agree. Missed that one. Will update the PR.

> For duplicate I am in doubt, might be that it needs to reactivated when needed.

In theory a `duplicate` can be reactivated and the `dedup` algorithm will take care of deduplicating it later. Right?

> And ideally we also test the opposite of the if, so that if a finding is not false_p/out_of_scope/risk_accepted, reactivation will be performed? In a way I think that is already tests by the pre existing test cases.

That's a good point. I can add a test for it inside test case. What do you think?

Please let me know if you agree with the above comments and I'll work on those improvements tomorrow or Tuesday.

valentijnscholten:
> We were also hit by an issue related to the comment below during a reimport of a GitLab dependency scanning report (I submitted this parser a few weeks ago):
> 
> https://github.com/DefectDojo/django-DefectDojo/blob/33fc63962b8e901bdf2b31bad735277e48951b1c/dojo/api_v2/serializers.py#L1420-L1432


I would wait maybe for #3753 as that will change the matching logic for reimport.

macedogm:
> Looking again at this I realize risk_accepted should be added to this same if clause, agree?

Added validation to skip the reimport of findings that are `false_p` or `out_of_scope` or `risk_accepted`.

> And ideally we also test the opposite of the if, so that if a finding is not false_p/out_of_scope/risk_accepted, reactivation will be performed? In a way I think that is already tests by the pre existing test cases.

Added inside the test an assertion to make sure that one finding that was deleted was properly reimported. Do you agree with this (please let me know if the test got too big)?
I had to add some support functions in `unittests/dojo_test_case.py` in order to be able to:
1. Delete a finding.
2. Get the `engagement_id` from a `test_id` through API request.
3. Get the `product_id` from an `engagement_id` also through API request.
4. Update a product configuration through the `product_id` to allow findings to be marked as `risk_accepted` through `enable_simple_risk_acceptance` flag.

> For duplicate I am in doubt, might be that it needs to reactivated when needed.

I'm also in doubt about this, because if the finding was marked as `duplicate`, then its duplicated finding will:
1. Be reopened through the reimport, in case it was deleted; or
2. Be kept inactive, in case it's a `false_p` or `out_of_scope` or `risk_accepted`

So I feel that this behaviour is already tested. What to you think?

damiencarol (approved):

valentijnscholten (approved):

Gby56:
This seems to have bugs with some parsers, the Veracode parser definitely has the problem because "Mitigated by Design" doesn't map to false positive, out of scope, or risk_accepted, so they always get reopened...
Also, if we're completely skipping a finding for these justifications, we might be missing some updated justifications on a reimport ? I would expect a date comparison to see if the scanner data has changed in the meantime

PR #3860, opened 2021-02-14
Fix editing of Engagement checklist

StefanFl:
When trying to edit a previously completed checklist, it was not loaded and instead the user had to start with an empty checklist again. With this fix, the checklist will be loaded and the user can continue where she stopped before.


valentijnscholten (approved):
Thanks for fixing, although the longtime brokeness of the feature does raise the question if anybody is using it.

StefanFl:
Good point @valentijnscholten, I am not sure how useful it is. But when it's there it should at least work properly :-)

Maffooch (approved):

Issue #3823, opened 2021-02-08
Move up the "cvss" field when creating/editing finding


madchap:
**Is your feature request related to a problem? Please describe**
The cvss3 field is way at the bottom. Too bad, it's got a nice JS to actually fill out the cvss right within dojo!

**Describe the solution you'd like**
Put it right under "severity" maybe?

valentijnscholten:
I noticed it also lower on the page on the view finding page using a whole panel for itself. Might be more logical to put it at the top, maybe in the same row as location / component name+version?

damiencarol:
It's a good idea

PR #3861, opened 2021-02-14
Rearrange for cvssv3

madchap:
Addresses #3823 - The useful cvssv3 javascript pop-up is certainly underutilized as it's buried down the field. Bringing it up next to severity so one is actually tempted to give findings a vector string.

Edit:

image

View - pop-over severity if exists and remove the panel.

image


damiencarol:
could we switch to label `CVSS v3` exactly ?

madchap:
Or something like that to avoid the new row?

image

madchap:
> could we switch to label `CVSS v3` exactly ?

Sure.

Maffooch:
Well that was not when I tried to do :/ got the migrations in order but the unit tests wouldn't recognize it so I tried to merge dev into this branch and got them switched.. May need to revert if that's the best idea.

valentijnscholten:
Maybe just make a new one that fixes it

PR #3867, opened 2021-02-15
Remove an unused script that validate Acunetix files

damiencarol:
This pull request remove a script that should be in the contrib repository

The pb with this script is that it depend on lxml which is vulnerable and we are moving all XML modules to defusexml.


valentijnscholten:
If it should be in the contrib repository, should it be added there? It also could be check to have in the parser?
To be honest, I think it's old code and can be removed.

valentijnscholten (approved):

Maffooch (approved):

PR #3873, opened 2021-02-16
Improve GitLab Dependency Scanning hash code configuration

macedogm:
This PR improves the hashcode configuration fields for GitLab Dependency Scanning Report, now using also `file_path`, `component_name` and `component_version` in the calculation.

GitLab dependency scanner can flag the same vulnerability multiples times if the scanned project has multiples versions (`component_version`) of the same the same dependency (`component_name`). Previously a reimport of a report with this scenario would close the "repeated" findings and keep only one open. This would cause valid findings to be closed.

In order for this to work, I had to use the same logic in `api_v2/serializers.py` that is currently being used for `Veracode Scan`, `Arachni Scan` and `Anchore Engine Scan` reports. This is just a hack and can be removed when PR https://github.com/DefectDojo/django-DefectDojo/pull/3753 is merged.


ptrovatelli:
hello. Hopefully i'm close to the end on https://github.com/DefectDojo/django-DefectDojo/pull/3753 :) 
I was trying to fix some issue with the dynamic parsers at the same time : identifying the findings without endpoints and mitigating the endpoints that are no longer present

macedogm:
That’s nice to know @ptrovatelli and thanks for the update.
Please let me know if I can be of any help 👍🏻

madchap:
Hey @macedogm

your PR now has conflict, could you check that and we can proceed with reviewing? Thanks!

macedogm:
@madchap I fixed the conflicts. Can you please review the PR?

madchap (approved):

dojo/settings/settings.dist.py
```diff
@@ -801,8 +801,8 @@ def generate_url(scheme, double_slashes, user, password, host, port, path, param
     'Acunetix Scan': ['title', 'description'],
     'Trivy Scan': ['title', 'severity', 'cve', 'cwe'],
     'Snyk Scan': ['vuln_id_from_tool', 'file_path', 'component_name', 'component_version'],
+    'GitLab Dependency Scanning Report': ['title', 'cwe', 'cve', 'file_path', 'component_name', 'component_version', 'severity'],
```
valentijnscholten, line 804: do we really need severity to be part of the hashcode calculation? always has felt weird to me. if there are two issues with all fields the same, except severity, aren't they just the same vulnerability but just the severity was changed?

valentijnscholten (changes requested):

dojo/settings/settings.dist.py
macedogm: Agree, your thought makes sense.
I also removed CWE, since it's not actually filled by the GitLab dependency scanner.

macedogm:
@valentijnscholten can you please review the PR now? I made the proposed fix.

dojo/settings/settings.dist.py
```diff
@@ -823,6 +823,7 @@ def generate_url(scheme, double_slashes, user, password, host, port, path, param
     'DSOP Scan': True,
     'Acunetix Scan': True,
     'Trivy Scan': True,
+    'GitLab Dependency Scanning Report': True,
```
valentijnscholten, line 826: i think this one can now be removed if cwe is not part of the hash calculation.

valentijnscholten (changes requested):
1 more comment :-)

dojo/settings/settings.dist.py
macedogm: done! thanks for spotting that 👍🏻

macedogm:
Should be good now.

valentijnscholten (approved):

Issue #3879, opened 2021-02-17
new findings with Info severity cannot be pushed to JIRA


valentijnscholten:
In the system settings there's a configuration option to specify the minimum severity for findings to be allowed to be pushed to JIRA. We can set this to 'Info' or leave it empty to allow all findings to be pushed to JIRA.

The `test/views.py#add_finding` logic doesn't respect this setting and has a hardcoded check to not push any Info findings to JIRA.

PR #3883, opened 2021-02-17
improve jira validation and error reporting

### LINKED ISSUE #3369: no error message when trying to create JIRA on existing finding
**Bug description**
If you **edit** an existing finding, and attempt to **push to jira without having "active, verified" set**, then **you will have no error message,** despite the fact it is needed. Instead, you'd see a "successful" message.

This behavior _does not happen when it's a brand new finding_. If you create a new finding and try to submit it directly to JIRA (unchecking "active" for example), then the error messages are shown, both at the top and next to the missed checkbox at the bottom.
 
**Expected behavior**
The error messages appear whether it's a new or existing finding.

Errors expected to be shown

image

image

- [x] Kubernetes
- [ ] Docker

**Environment information**
latest dev

valentijnscholten:
When Defect Dojo pushes a finding to JIRA, this happens in the background/celery process. So any errors will only show up later in the alerts on the top right. For technical errors this might be OK as they are rare and we don't have any other choice if we want to keep the background processing in place.
Validation errors are more common and we keep getting reports from users they are confused why their JIRA issue was not created.
Validations in place:
- finding must be active and verified
- finding must be above minimum severity from system settings

Currently the validation is most of the time also performed in the background process, leasing to confusion. This PR performs all validations in the foreground before handing of to the background process. This can be done without a big performance hit because it's "just validation" and no network traffic is happening.

Cases covered:
- add findings
- edit finding
- promote finding
- add ad hoc finding
- ajax add issue
- bulk edit push to jira

Also fixes 
- #3369 no error message when trying to create JIRA on existing finding
- #3879 new findings with Info severity cannot be pushed to JIRA

Screenshots:

2021-02-18 19_31_09-View Finding _ DefectDojo

2021-02-18 19_32_56-View Finding _ DefectDojo

2021-02-18 19_31_34-Edit Finding _ DefectDojo

2021-02-18 19_31_39-Edit Finding _ DefectDojo

2021-02-18 19_32_34-Edit Finding _ DefectDojo

2021-02-18 19_32_39-Edit Finding _ DefectDojo

2021-02-18 19_44_08-Test _ DefectDojo


Maffooch (approved):
Big fan of this

madchap (approved):
locally test, lgtm. I'll let you merge at your convenience. Cheers! :+1:

PR #3885, opened 2021-02-18
aftercare: don't show paging snippet if no results, fix tags field layout

valentijnscholten:
Aftercare for #3812 

After my latest paging PR it looked kinda ugly when there were no results:

image

This PR fixes that by hiding the paging snippet rows when there are no results

Also fixes multiple tags fields appearing in the bulk edit form.


Maffooch (approved):
Good change

madchap (approved):

PR #3888, opened 2021-02-19
Auth V2: three fixes for the new autorization system

StefanFl:
Further tests on the merged new authorization system showed three areas, where the behaviour was not exactly the same as before when old authorization is still active:

- dojo/authorization/authorization_decorators.py: Call legacy user authorization, otherwise non-staff users can't view product types where they are an authorized user
- dojo/product/views.py: Staff users are allowed to add new products
- dojo/product_type/views.py: Don't save product type members


valentijnscholten (approved):

valentijnscholten:
thanks. this backwards compatibility when the flag is turned off is really important.

Maffooch (approved):

PR #3893, opened 2021-02-20
API V2 Documentation should use URL Prefix (if set)

JoseRoman:
When hosting DefectDojo under a sub-directory, the URL for API V2 documentation is broken since it doesn't use the URL_PREFIX value. 

If `DD_URL_PREFIX` is set to 'defectdojo/', then the URL should be https://example.com/defectdojo/api/v2/doc/


Maffooch (approved):
Good find!

valentijnscholten:
I haven't looked at the uwsgi/django docs yet, but it feels like we're doing something wrong if we have to add this in so many places....?

valentijnscholten (approved):

PR #3894, opened 2021-02-20
Revert "Rearrange for cvssv3"

Maffooch:
Reverts DefectDojo/django-DefectDojo#3861


madchap:
Why reverting the whole thing? (have not looked deeply, but I find it odd)

Maffooch:
I don't know what else to do :/ I was gonna approve after fixing the mgiration but merged instead. I don't think it's really a problem to keep it merged.

madchap:
I think it's fine, I was done with it anyways. Was that your only concern? If so, we can close this one, no need to revert and let's move on.

PR #3899, opened 2021-02-21
fix typo in setup.py for jsonlines

kmcquade:
Related slack thread: https://owasp.slack.com/archives/C014H3ZV9U6/p1613934621008200

[jsonlines](https://pypi.org/project/jsonlines/) - it looks like this is the package that was intended. That is the one referenced in requirements.txt

[json-lines](https://pypi.org/project/json-lines/) - this is the one that it currently points to in the master branch.

This PR updates to the proper package and most recent version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,7 @@ setup(
         'whitenoise==4.1.4',
         'django-environ==0.4.5',
         'titlecase',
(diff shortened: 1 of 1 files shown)


valentijnscholten:
oops, didn't see it was based against master. should have been dev. 
anyway, will be corrected as part of the release on tuesday.

PR #3676, opened 2021-01-21
Update Qualys_Infrascan_Webgui parser.py

tvtongerloo:
Catching "No registered hostname" as non-valid value for fqdn, falling back to IP address to be used for the endpoint; otherwise you get an endpoint called "No registered hostname" and findings are aggregated incorrectly.

Raw XML report extract example (actual public IP has been replaced with 127.0.0.1):

`<IP value="127.0.0.1" name="No registered hostname">`
`<OS><![CDATA[Linux 2.6]]></OS>`
`<INFOS>`
`<CAT value="Information gathering">`
`<INFO number="6" severity="1">`
`<TITLE><![CDATA[DNS Host Name]]></TITLE>`
`<LAST_UPDATE><![CDATA[2018-01-04T17:39:37Z]]></LAST_UPDATE>`
`<PCI_FLAG>0</PCI_FLAG>`
`<DIAGNOSIS><![CDATA[The fully qualified domain name of this host, if it was obtained from a DNS server, is displayed in the RESULT section.]]></DIAGNOSIS>`
`<CONSEQUENCE><![CDATA[N/A]]></CONSEQUENCE>`
`<SOLUTION><![CDATA[N/A]]></SOLUTION>`
`<RESULT format="table"><![CDATA[IP address Host name 127.0.0.1 No registered hostname]]></RESULT>`
`</INFO>`

PR #3907, opened 2021-02-22
Qualys Infrastructure Scan parser: maintenance and fix date of finding

damiencarol:
Work done:

- [x] Remove dependency to lxml (switch to defusedxml)
- [x] add more unit tests on finding attributes
- [x] remove useless code

Supersede https://github.com/DefectDojo/django-DefectDojo/pull/3676


devGregA:
@Maffooch  please look at this one with the qualys pain from previous.

damiencarol:
@devGregA do you want to add more unit tests? What was the pbs in the past with this parser?

Maffooch:
Before PR:
image


After PR:
image


Things to note:
- Change in severity
- Loss "found" times

damiencarol:
@Maffooch @devGregA : some feedback about the changes you saw.

**1) Change in severity**
As you can this in my patch I changed the parser to just get the severity as-is from the report:
https://github.com/DefectDojo/django-DefectDojo/blob/93669545eb1674851a5543d72206acfabb6edf4c/dojo/tools/qualys_infrascan_webgui/parser.py#L82
The function `qualys_convert_severity` translate the severity directly. No specific severity management. Like this:

| Value | Severity in parser |
| ------ | ------------------ |
| 1 | Info |
| 2 | Low |
| 3 | Medium |
| 4 | High |
| 5 | Critical |
| `None` | Info |

So... not sure severity are different now. Could this be a bug in the the old version of the parser manage it?

**2) Loss "found" times**

For now this data was just *false* , always 1. Not sure if something wrong is useful.

https://github.com/DefectDojo/django-DefectDojo/blob/0dc1c498410edd212b8dae05659672c7a9e814f4/dojo/tools/qualys_infrascan_webgui/parser.py#L82-L84

I can modify it to be correct, I mean taking care of number of duplicate the parser saw in the report and using this number.
**same comment apply for First Found/Last Found**

ptrovatelli (changes requested):
changes were requested in the discusssion
(if aggregating stuff in the parser, maybe use the nb_occurences in finding model)

madchap:
@Maffooch some follow-up to @damiencarol 's response? Cheers.

damiencarol:
@ptrovatelli there is no aggregating stuff in the parser, so no need to update nb_occurences. Maybe in another pull request I can add more aggregation

Maffooch:
Apologies for the delayed response

1. Does not seem like there is much we can do 
2. Ahh it looked like maybe the parser could have aggregated stuff automatically with the use of that `times_found` var but @ptrovatelli is right, if the field does have something other than one, then it should be handled by the`nb_occurences` field.

damiencarol:
@Maffooch @ptrovatelli I double checked the parser file before and after my PR and I can confirm there is no aggregation in the parser. So:
1. I can update the nb_occurence to be always 1
2. let it be like that (removing _times_found wich make feel that it could have aggregation in the parser **but** it doesn't exists)

Maffooch (approved):

dojo/tools/qualys_infrascan_webgui/parser.py
```diff
@@ -66,146 +40,78 @@ def issue_r(raw_row, vuln, scan_date):
...
+                                                    htmltext("**Port:** " + str(_port)),
+                                                    htmltext("**Result Evidence:** " + _result),
+                                                   ])
+            # Impact description
+            _temp['IMPACT'] = htmltext(str(vuln_details.findtext('CONSEQUENCE')))
+
+            # CVE and LINKS
+            _cl = []
+            _temp_cve_details = vuln_details.iterfind('CVE_ID_LIST/CVE_ID')
+            if _temp_cve_details:
```
ptrovatelli, line 75: @damiencarol isn't this aggregating several CVE into a single defectdojo finding? I don't know these reports but maybe it makes sense to count the CVE and put that number into nb_occurences. 

If not, I believe it's better to leave nb_occurences to null rather than always hardcoded to 1

ptrovatelli:
@damiencarol fine by me apart from the above comment. thanks :)

dojo/tools/qualys_infrascan_webgui/parser.py
damiencarol: interesting comment.
FYI all the parsers implements `nb_occurences` this way:
1) always set it to 1 when creating the finding **or** setting to the same value if the source aggregate the things
2) if there is de-duplication in the parser itself, update `nb_occurences` 
To be honest I was wondering to work on updating every parsers to follow these guidelines.
So when you create a single finding and set `nb_occurences` to 1, it's not *hard-coded*, it's the correct value.

I think you're right to point to the pb of this parser that it aggregates all CVE in one finding **but** it's not how the data model works right now.
The parser should create one finding by CVE and copy all other data, as for now `cve` is an attribute of the finding.
I can change that in a future pull request but I don't want this pull request to become a monster PR that change everything in this parser.
@ptrovatelli could we post-pone this big change in another pull request and merge this one as-is?

damiencarol: Also if we follow your way, we will have one finding with one CVE that have `nb_occurences` > 1. it could make think the user that the CVE appear more than one time. Which is not the case.

ptrovatelli: initially nb_occurences was meant for parsers that aggregate several occurences of the same vulnerability from the source into a single finding. for example same cve at multiple line number in the same file. The goal of nb_occurences is that when looking at the source tool, ppl can make sure that defectdojo didn't loose finding during the process: i have 10 vuln in my source tool, only one in defectdojo, but as nb_occurences=10 i'm good. that's the case for checkmarx parser for example which tends to count many vulns that are aggretated to a single finding in defectdojo. 

it seems that qualys aggregates cve1, cve2, cve3 into the same finding, so it seems reasonnable to use nb_occurences too, even though it's a little bit different than previous use case. 

anyways if the parser is always setting 1 i think it's better to leave empty. same discussion as "no mitgation found" vs null :)

ptrovatelli (approved):

dojo/tools/qualys_infrascan_webgui/parser.py
damiencarol: completely agree with your comment.
I think I will work on a PR that dissociate the many CVE into multiple findings to stick to this model.
Could be good also to explain this details in the documentation that help implementing parsers.

PR #3913, opened 2021-02-22
Add: dependencies in helm3 style

dsever:
Moving away from helm v1 to helm v2

Would be nice to setup minimal k8s supported version:

kubeVersion: ">= 1.16.0"


madchap:
Was not that what we tried before and we had issues with the `condition` statements?

dsever:
Yes I remember it, but I tested locally and would say it work properly

madchap:
You'd need to rebase @dsever, not sure what happened to the PR here.

dsever:
@madchap  I will open new PR

PR #3914, opened 2021-02-23
Add configurable disclaimer to notifications and reports

Maffooch:
Add a disclaimer to reports and notifications. This disclaimer is configurable through the system settings. There is an option to add it to all generated reports (HTML/AsciiDoc) while notifications will have the disclaimer if it exists in system settings.

System Settings Field:
image

### Notifications
Alerts:
image

Email:
image

MSTeams:
image

### Reports
Options:
image

Report:
image


mtesauro (approved):
Approved

valentijnscholten:
Do we want the disclaimer in the alerts inside defect dojo?
I would think it's not needed there and also not in Teams / Slack messages?

Maffooch:
Could remove it from alerts, but not teams/slack. Think of contractors that would need to see that reminder from time to time.

devGregA (approved):

Maffooch:
@valentijnscholten you good with this?

valentijnscholten (approved):

PR #3921, opened 2021-02-24
Bugfix at close_old_findings

adiffpirate:
Fixed a bug where the 'is_Mitigated' field is not updated when findings are automatically closed, which leads to them not been considered closed internally by DefectDojo (although on the UI it appears "Mitigated"). This causes these findings not to appear on metrics nor at the 'View Closed Findings' option.


adiffpirate:
Closing because I wrongfully requested to the master branch. Gonna open another PR later pointing to dev 😉

PR #3923, opened 2021-02-24
Helm release actions

madchap:
Attempt to use https://github.com/helm/chart-releaser-action to satisfy for https://github.com/DefectDojo/django-DefectDojo/issues/3778

Converting to draft right now, some odd behavior testing on my fork, probably not far to succeed I guess.

Not sure why I am getting

image

fyi @bgoareguer


madchap:
Moving.. only specifying `helm` for the directory :man_shrugging: 

image

"Problem" there is that it creates the release directly, so need to probably amend our GHA anyways.

madchap:
Attempt to use the chart-release option leads to an error, kind of as expected.

image

Also on that same run, something seems not right as it does not seem to notice nothing in the chart changed, and attempts to push over again, which results in an error as the tgz already exists.

image

madchap:
Seems that when running in upstream, things happen

image

madchap:
Closing in favor of https://github.com/DefectDojo/django-DefectDojo/pull/3849

PR #3929, opened 2021-02-25
set mitigation related fields correctly at close_old_findings

adiffpirate:
Fixed a bug where the 'is_Mitigated' field is not updated when findings are automatically closed, which leads to them not been considered closed internally by DefectDojo (although on the UI it appears "Mitigated"). This causes these findings not to appear on metrics nor at the 'View Closed Findings' option.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1262,6 +1262,7 @@ class ImportScanSerializer(serializers.Serializer):
 
             for old_finding in old_findings:
                 old_finding.active = False
+                old_finding.is_Mitigated = True
                 old_finding.mitigated = datetime.datetime.combine(
                     test.target_start,
                     timezone.now().time())
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1263,14 +1263,7 @@ class ImportScanSerializer(serializers.Serializer):
             for old_finding in old_findings:
                 old_finding.active = False
                 old_finding.is_Mitigated = True
-                old_finding.mitigated = datetime.datetime.combine(
-                    test.target_start,
-                    timezone.now().time())
-                if settings.USE_TZ:
-                    old_finding.mitigated = timezone.make_aware(
-                        old_finding.mitigated,
-                        timezone.get_default_timezone())
-                old_finding.mitigated_by = self.context['request'].user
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Thanks, you might as well set `mitigated_by`, or even better, use `dojo/finding/helper.py#update_finding_status`

adiffpirate:
Done. I didn't know `dojo/finding/helper.py#update_finding_status` existed, thanks

adiffpirate:
Is there an easy way we can get `new_finding` from `old_finding`? If so, we could also pass `new_finding` as a parameter to `finding_helper.update_finding_status` and avoid [these lines](https://github.com/DefectDojo/django-DefectDojo/pull/3929/commits/9d551e535c9604dd5d17bd544f20e205eee76bd8#diff-401b1bd65c9f7a0024ca5d7bc72958d3f941282a26b008feea5ae71bd5cfe2a1R1264-R1265)

valentijnscholten:
It's a bit of a weird helper function, but I am working on a PR that changes most of this anyway so I would leave it as is now. Instead could you maybe look at adding a few asserts in `test_import_reimport.py`? That will help us when refactoring to prove we didn't break anything.

adiffpirate:
Sure, I'll look into that 😉

valentijnscholten:
Probably your usecase is now covered by #3947

adiffpirate:
I'm still working on making this use case a unit test, but it may take some time (I'm still learning about unittests in Python and the Django framework). In the meantime, I thought it would be nice to explain better the bug context.

So, this bug happens to me when I'm importing a new test with `close_old_findings = true` via APIv2. 
The expected behaviour is that all old findings from the imported test_type in that product are closed (I use product wide deduplication). 
What actually happens is that those old findings are being set to inactive and mitigated but they don't appear at the 'View Closed Findings' field nor the 'Metrics'.

valentijnscholten:
Set is_Mitigated to True is the official way and should fix that part.

adiffpirate:
So, what should we do here? Close this PR and work on a fix at #3947 or merge and refactor it later on?

valentijnscholten (approved):

valentijnscholten:
I think it's best to merge your PR, as mine will take a little longer to review possibly.

madchap (approved):

PR #3935, opened 2021-02-26
TruffleHog3 parser support and refresh version 2

damiencarol:
What this pull request do:
- [x] add unit tests for TruffleHog3
- [x] add unit tests for TruffleHog (no unit tests currently!)
- [x] implement a new parser for TruffleHog3 (https://github.com/feeltheajf/truffleHog3) which is not the same software as TruffleHog but a fork
- [x] implements attributes `nb_occurences` for both parsers
- [x] Improve description part (see screenshot)
- [x] add attribute `line` which enable de-duplication algorithm as-is

image


valentijnscholten:
Is it reliable to have both formats handled by a single parser? Also with regards to deduplication? Could be that v2 gets deduplicated against v3. Might be a good thing. Or not.  :-)

damiencarol:
I think it's not a good idea to have both formats handled by a single parser. in fact TruffleHog and TruffleHog3  are not the same project. TruffleHog3 is a fork.

valentijnscholten:
So not a good idea, but we have to keep it for now as is?

damiencarol:
No I created q specific parser for TruffleHog3 and made some clean for TruffleHog. As it is maintained as 2 different projects and produce different data, I'm against merging them.

valentijnscholten:
Oops, I didn't notice they were already split into two parsers 😆 Sorry for the confusion, will approve right away!

valentijnscholten (approved):

madchap (approved):

PR #3936, opened 2021-02-26
GHA: Test multiple k8s versions

dsever:
Added into minikube config to test k8s v.1.16.0 and  v.1.20.0

### DIFF
--- a/.github/workflows/k8s-testing.yml
+++ b/.github/workflows/k8s-testing.yml
@@ -108,6 +108,7 @@ jobs:
       matrix:
         databases: [pgsql, mysql]
         brokers: [redis, rabbit]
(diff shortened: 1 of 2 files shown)


madchap:
Supersedes #3934

valentijnscholten:
I think it also fixes a bug, not just added tests?

valentijnscholten (approved):
we need to merge this before GHA will run the new workflow definition.

damiencarol (approved):

madchap:
> we need to merge this before GHA will run the new workflow definition.

Actually, I think we should merge https://github.com/DefectDojo/django-DefectDojo/pull/3948 before, and so that we see this is safe.

Issue #3931, opened 2021-02-25
Can SAST details be included when pushing issues to JIRA?


mperrotti-fdbl:
I'm using DefectDojo as middleware between Veracode and JIRA.

When I import SAST findings into DefectDojo, I see a lot of useful information, such as Source Filepath, Source Line Number, Sink Filepath, Sink Line number.

However, when creating JIRA tickets based off of these findings, these details are missing.

The JIRA ticket does contain links back to DefectDojo, but that's an extra step for the developers.

The JIRA ticket also contains the Veracode finding ID, but that's an extra step for the developers also.

Would it be possible for those source and sink details to be pushed to JIRA when populated in a SAST finding?

I am currently updating every ticket I push manually with that level of detail so the developers understand where the problem is, without requiring them to go to another system unless they need additional details on the problem.

Thanks in advance!

valentijnscholten:
You can edit `issue-tracker/jira-description.tpl`, might be even worthwhile to make a PR from it.

mperrotti-fdbl:
I will look into it, thank you.

mperrotti-fdbl:
I was able to do this no problem.

{% if finding.component_name %}
*Vulnerable Component*: {{finding.component_name }}
*Source Object*: {{ finding.sast_source_object }}
*Source File*: {{ finding.sast_source_file_path }}
*Source Line*: {{ finding.sast_source_line }}
*Sink Object*: {{ finding.sast_sink_object }}
{% endif %}

I would issue a PR but other folks might want to see it differently.

valentijnscholten:
Added those fields in #3938

PR #3938, opened 2021-02-26
JIRA: allow different templates for jira description rendering

valentijnscholten:
Depending on your JIRA instance or JIRA project, you may want to use different templates for rendering the issue description. Sometimes you want to provide all the details and want them to be visible in JIRA.
Sometimes you want to provide only a link back into defect dojo and not too much details.
This PR allows you to chose between these two templates, or add even more of your own templates.
Any templates added to (or modified in) `dojo/templates/issue-trackers` automatically become options in the jira config dropdowns.

The PR is always falling back to the default full description template already in use. So when people to do nothing after upgrading, it just keeps work as it was. It's all optional.

The PR also adds the `component_name`, `component_version` and all the `SAST_xxxx` fields to the "full/default" jira template, fixing #3931 

image
image

BTW The first attempt at this PR was using a new model `JIRAIssueTemplate` that could be edited via the Django Admin portal at `/admin`. But this opened up risks that people could use the template language to retrieve all kinds of sensitive details. In the current world I think a lot of setups still have "everyone" as staff/superuser, so there was too much risk.
With a better permission model we could still switch to using a model instead of having to use templates on the filesystem and having to rebuild the images (or use complex mounting nobody understands ;-))

BTW2 In v2 of this PR I used the `FilePathField` on the model to let django populate the dropdown for us. But this resulted in the full path being stored in the issue_template field, where the template loader expects a relative path. So this required adding/removing of the prefix in some places, which broke the model validation, etc. 

BTW3 So then I moved to just a CharField with choices. This worked, but if a user adds a template on disk, the choices change and django expects you to make a migration. Our initializer automatically makes those resulting in instances getting our of sync and in conflict with upstream releases.

So I fell back to just using a CharField and populating the dropdown on the forms.


madchap:
fyi @yilmi

madchap (approved):
Locally tested in staging environment, works peachy :-)

This will allow us to address some teams wanting to restrict data access for confidentiality purposes, while others want all details in JIRA directly without having to navigate to Dojo.

damiencarol (approved):
Gave a try. it's a good feature. Maybe some users will want to add there own template in the future

Issue #3881, opened 2021-02-17
bulk edit, reimport, update_from_source don't set mitigated_by field


valentijnscholten:
There is a helper class `findings/helper.py` which has a method to help setting mitigated / is_Mitigated / mitigated_by always correctly. This should be called whenever the status of a finding changes. This call is missing from bulk_update.

Same for reimport that can close or reopen findings.

Same for update_from_source (used by sonarqube integration).

I think it's time for a more transparent solution to set these fields always in finding.save() or using signals :-)

PR #3947, opened 2021-02-27
add last_status_update field and make status updates consistent

valentijnscholten:
I wanted to make a management command to reconciliate findings status between defect dojo and JIRA. Because sometimes the webhook throws an error. Or JIRA is down. Or the internet is down. Etc.

To be able to reliably do this, we need to know on both sides when the latest status changed happened, because the latest status change will "win" any difference in status.

In JIRA we can obtain the latest status change via the REST api, but in Defect Dojo we don't have this field. Closes is the `last_reviewed` field. But it's only set in some cases and not always. And it's also set when no fields are changed, or only fields not related to statuses.

So this PR introduces a generic mechanism to set a new field `last_status_update` on the Finding model. We had some code in place to set other fields consistently, for example set `mitigated` when `is_Mitigated` is (being set to) True.
But this involved calling this helper method manually everywhere it was needed, and we missed a few spots, see #3881.

So this PR introduces the `django-field-signals` package which allows us to "listen" to changes in certain fields.

I'm usually not a fan of Django signals, but in this case I think it's a nice way to reuse some code and have a plug-and-play 'what fields are being changed` mechanism.

This generic approach also opens up the possibility to extend the "finding status history" that we are already tracking for reimports.

The migration is split in 2 parts to ease any troubleshooting if the data migration encounters an issue.

The PR has a breaking changes label. Although it's just making the behaviour around the above mentioned fields consistent, there will be 1 or 2 places where that wasn't the case. In those places now the mitigated field gets set correctly, which might be surprising for people relying on it being incorrectly empty.

UPDATE:
- rewrote the migration as for 750.000 findings the original migration was taking 1,5hours due to django loading all findings in memory. now it uses cursors and takes about 1minute.


adiffpirate:
I did some manual tests and [this bug](#3929) is still happening here. The problem is within [this loop](https://github.com/valentijnscholten/django-DefectDojo/blob/status-auditlog/dojo/api_v2/serializers.py#L1253-L1275), but I'm not sure how to fix it since I'm not very familiar with this new approach of using Django signals.

valentijnscholten:
Ah, the code in that loop is just plain wrong. Probably because of the confusion between mitigated and is_Mitigated. Maybe we should check everywhere where these fields are set and fix.

valentijnscholten:
> Ah, the code in that loop is just plain wrong. Probably because of the confusion between mitigated and is_Mitigated. Maybe we should check everywhere where these fields are set and fix.

That part is fixed in #3929, which is merged in dev, which is now merged into this PR.

madchap (approved):
lgtm. I actually like signals ;-)

dojo/finding/helper.py
```diff
@@ -1,59 +1,105 @@
...
-    if not new_state_finding.is_Mitigated and new_state_finding.mitigated_by is not None:
-        finding_status_changed = True
-        new_state_finding.mitigated_by = None
-
-    return finding_status_changed
-
-
-def can_edit_mitigated_data(request_user):
-    return settings.EDITABLE_MITIGATED_DATA and request_user.is_superuser
+    print('setting last status update: ' + str(now))
```
damiencarol, line 100:
```suggestion
```

dojo/models.py
```diff
@@ -2110,8 +2122,9 @@ def save(self, dedupe_option=True, false_history=False, rules_option=True,
             from dojo.utils import do_apply_rules
             do_apply_rules(self, *args, **kwargs)
 
-        from dojo.utils import calculate_grade
-        calculate_grade(self.test.engagement.product)
+        if product_grading_option:
+            from dojo.utils import calculate_grade
```
damiencarol, line 2126: could we keep the imports at the start of the file?

damiencarol (changes requested):

dojo/models.py
valentijnscholten: no due to circular dependencies

damiencarol: circular dependencies could shows a bad code organization/splitting pb but ok.

damiencarol (approved):

valentijnscholten:
Looks like the migration refactor caused the postgres matrix of the kubernetes builds in GHA to fail, will investigate.

valentijnscholten:
Turns out that the `is_Mitigated` column is causing problems due to the capital M in there. Postgres converts this back into lowercase, and then tells you the column doesn't exist. To solve it, the column name needs double quotes. But then the query won't work in MySQL. I fixed it by implementing an if statement based on the database vendor. This worked with  750k findings in MySQL and 990k on postgres.

But I decided to revert back to the slightly slower ORM based migration as we don't want any risks with migrations, they can be a pain to debug. 

So end result is 750k findings in 5 minutes on MySQL, 990k findings in 6 minutes on Postgres.

madchap:
Yea that's certainly good enough. I had that issue with pgloader when migrating from mysql to pgsql indeed.

PR #3952, opened 2021-02-27
[GHA] Remove useless docker login

madchap:


valentijnscholten:
the login is there to reduce the chance of getting rate limited. but we removed it on the actions triggered by pull requests to reduce the chance of people stealing our credentials.

damiencarol (approved):

madchap:
> the login is there to reduce the chance of getting rate limited. but we removed it on the actions triggered by pull requests to reduce the chance of people stealing our credentials.

OK. So? :)

valentijnscholten:
I think we can remove it until we hit any rate limits, which we haven't so far.

PR #3957, opened 2021-02-28
JIRA status reconciliation management command

valentijnscholten:
based on #3947, so merge that one first.

Basically what I added to the docs, explains this PR:

**Status reconciliation**

Sometimes JIRA is down, or Defect Dojo is down, or there was bug in a webhook. In this case
JIRA can become out of sync with Defect Dojo. If this is the case for lots of issues, manual reconciliation
might not be feasible. For this scenario there is the management command 'jira_status_reconciliation'.

```
usage: manage.py jira_status_reconciliation [-h] [--mode MODE] [--product PRODUCT] [--engagement ENGAGEMENT] [--dryrun] [--version] [-v {0,1,2,3}]

Reconcile finding status with JIRA issue status, stdout will contain semicolon seperated CSV results. Risk Accepted findings are skipped.

optional arguments:
  -h, --help            show this help message and exit
  --mode MODE           - reconcile: (default)reconcile any differences in status between Defect Dojo and JIRA, will look at the latest status change
                        timestamp in both systems to determine which one is the correct status
                        - push_status_to_jira: update JIRA status for all JIRA issues
                        connected to a Defect Dojo finding (will not push summary/description, only status)
                        - import_status_from_jira: update Defect Dojo
                        finding status from JIRA
  --product PRODUCT     Only process findings in this product (name)
  --engagement ENGAGEMENT
                        Only process findings in this product (name)
  --dryrun              Only print actions to be performed, but make no modifications.
  -v {0,1,2,3}, --verbosity {0,1,2,3}
                        Verbosity level; 0=minimal output, 1=normal output, 2=verbose output, 3=very verbose output
```

This can be executed from the uwsgi docker container using:

``` {.sourceCode .bash}
$ docker-compose exec uwsgi /bin/bash -c 'python manage.py jira_status_reconciliation'
```

DEBUG output can be obtained via `-v 3`

``` {.sourceCode .bash}
$ docker-compose exec uwsgi /bin/bash -c 'python manage.py jira_status_reconciliation -v 3'
```

At the end of the command a semicolon separated CSV summary will be printed. This can be captured by redirecting stdout to a file:

``` {.sourceCode .bash}
$ docker-compose exec uwsgi /bin/bash -c 'python manage.py jira_status_reconciliation > jira_reconciliation.csv'
```

image

To achieve this, I had to make some changes inside Defect Dojo:
- extract webhook code to update finding status, move this to jira helper
- extract code to update the status in JIRA into its own method
- improve verbose_name and help_text for open_status_key en close_status_key. this doesn't affect the SQL schema, so it's safe to modify the existing migrations.
- use `updated` timestamp from JIRA when handling incoming webhook

Also includes some bugfixes:
- #3965


Maffooch:
Been dreaming of this one! Everything looks good to me except for the migrations. 

- Why the decision to edit older ones? 
- Is this to avoid a new migrations? 
- Do these changes take affect for existing installs? Does Django even notice a change?
- Are these changes even needed for existing installs?

valentijnscholten:
What do you mean by editing older ones? Do you mean why I am modifying existing findings and setting the `last_status_update` field there?
Or do you mean why I am editing existing migrations? I have noticed that existing migrations can be editing safely as long as they don't affect the SQL schema. So it helps to avoid having lots of small migrations only changing display texts. It really helps me if you have multiple PRs and have to keep all the migrations in sync/rebased/renamed.

What django does is that it builds a model in memory from applying all the migrations from 0001 to the end. And then it compares that to what is in `models.py`. So these needs to be in sync. And all this has to be in sync with the database. But django is only able to compare things with the database that are actually there, so new columns, renamed columns, changed columns, new indexes etc they need a new migration because they need to be applied to existing instances.
But things like `verbose_name` and `help_text` or even `choices` are not in the sql schema. So there is no migration needed, and also django won't be able to compare anything with the database so there is nothing that can be out of sync.
For those 'display' fields the only thing that matter is that the end result of all the migrations (in memory) is identical to what is in `models.py`.

madchap (approved):

madchap:
@Maffooch would you be willing to test this? I'd be keen on getting this to `dev` so I can push to my prod :-D

Maffooch:
This is super useful! Works like a charm too. Very glad to see that pulling changes from JIRA and reflecting them on dojo works without a webhoook. The only thing I noticed was the `jira change` column was not being updated consistently when viewing a finding. When findings are reconciled and changes in dojo were different than JIRA, the `jira change` value was not updated. The opposite occurred when changes in JIRA were different than in dojo as the `jira change` column _was_ updated.
Besides that, this is a solid PR and ready for an approval from me.

dojo/management/commands/jira_status_reconciliation.py
```diff
@@ -0,0 +1,221 @@
...
+from django.utils import timezone
+from django.utils.dateparse import parse_datetime
+
+from dojo.models import Engagement, Finding, Product
+import dojo.jira_link.helper as jira_helper
+import logging
+logger = logging.getLogger(__name__)
+
+
+"""
```
damiencarol, line 13: make it the comment of the class plz.
Also I don't think it's necessary to put name of author everywhere. we have a VCS that track that, better to let git save this kind of data.

damiencarol (approved):
Made a comment

dojo/management/commands/jira_status_reconciliation.py
valentijnscholten: fixed

valentijnscholten:
> This is super useful! Works like a charm too. Very glad to see that pulling changes from JIRA and reflecting them on dojo works without a webhoook. The only thing I noticed was the `jira change` column was not being updated consistently when viewing a finding. When findings are reconciled and changes in dojo were different than JIRA, the `jira change` value was not updated. The opposite occurred when changes in JIRA were different than in dojo as the `jira change` column _was_ updated.
Great find! Fixed now.

Issue #3971, opened 2021-03-02
error while importing nessus scan results - 'NessusXMLParser' object has no attribute 'parse'


asiyani:
While uploading nessus scan results, getting following error. I have tried both `csv` and `nessus` file format both of them giving same errror message.

Error message on application
`An error has occurred in the parser, please see error log for details.`

**Steps to reproduce**
1. Go to Product -> findings -> Import Scan Results
2. Select Scan type -> Nessus Scan
3. Choose File to upload
4. Click Import
5. See error

**Expected behavior**

**Deployment method** *(select with an `X`)*
- [x] Kubernetes
- [ ] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Docker image: `defectdojo/defectdojo-django:1.13.1`

**Console logs** (optional)
```
[02/Mar/2021 16:15:17] ERROR [dojo:590] 'NessusXMLParser' object has no attribute 'parse'
Traceback (most recent call last):
  File "./dojo/engagement/views.py", line 583, in import_scan_results
    parser_findings = parser.get_findings(file, t)
  File "./dojo/tools/nessus/parser.py", line 256, in get_findings
    return list(NessusXMLParser().parse(filename, test).values())
AttributeError: 'NessusXMLParser' object has no attribute 'parse'
[02/Mar/2021 16:15:17] ERROR [dojo:591] Error in parser: 'NessusXMLParser' object has no attribute 'parse'
```
we get simmilar message for csv file `NessusCSVParser object has no attribute 'parse'`

damiencarol:
@asiyani are you able to share the report? if not maybe an anonymized version?

damiencarol:
Ok I'm able to reproduce the bug. I'm pushing a pull request with a fix

damiencarol:
@asiyani  https://github.com/DefectDojo/django-DefectDojo/pull/3975

asiyani:
@damiencarol Thank you very much for quick reply & fix!

damiencarol:
@asiyani it was fixed in `1.13.2`. thanks for your feedback

PR #3975, opened 2021-03-02
Fix nessus parsers

damiencarol:
Fixes https://github.com/DefectDojo/django-DefectDojo/issues/3971


damiencarol:
Already part of `1.13.2` so merged in master already.

Issue #3820, opened 2021-02-08
Cleanup: Automatic nmap scans need to be fixed or should be removed


StefanFl:
**Bug description**
DefectDojo provides a functionality to run regular nmap scans and send the results via email. It is hidden in the UI, there are no menus pointing to it. It can be configured when a user knows the URL's or via the API. 

The scans don't run in a docker or kubernetes environment, because nmap must be installed.

**Steps to reproduce**
1. Add a scan with https://dd.example.com/product/1/scan/add
2. Open the scan with https://dd.example.com/product/1/scan/1/settings
3. Choose `Launch Scan Now` in the menu of the Scan Settings
4. Check logs, you will get the error message `('Nmap not found', <class 'nmap.nmap.PortScannerError'>)`
5. The UI doesn't report any failure but shows a completed scan

**Expected behavior**
I see 2 options:
- Fix the functionality by installing nmap in the docker image, add the management of scans to the UI and create findings instead of sending an email
- Remove the functionality, because it seems to have been added in 2015 and has only been touched for maintenance of other stuff

**Deployment method** 
- Kubernetes (impacted)
- Docker (impacted)

**Context**
This feature use a very old lib with last relase from 2016/07/29 (v0.6.1)

If the decision is to remove the functionality, I am happy to do it.

StefanFl:
**Supplement:** The library `python-nmap`, which is used for the scan, has been updated the last time in 2016 and should be updated to the much more recent `python3-nmap`, in case the functionality shall be fixed.

damiencarol:
@StefanFl I see this feature as out of scope for the project, we discussed to remove it in 2.0.0

StefanFl:
Duplicate of #3974

valentijnscholten:
We have decided to remove it during the team meeting last night. Feel free to go ahead and remove it.

PR #3993, opened 2021-03-03
Removal of automatic nmap scans

StefanFl:
DefectDojo provides a functionality to run regular nmap scans and send the results via email. It is hidden in the UI, there are no menus pointing to it. It can only be configured when a user knows the URL's or via the API. The scans don't run in a docker or kubernetes environment, because nmap must be installed.

This functionality is not needed anymore and will be removed with this PR, resolving #3820.


damiencarol (approved):

StefanFl:
I have updated the new documentation in /docs. Is it still necessary to update the documentation for readthedocs as well?

damiencarol:
@StefanFl not critical. If you have time you can push something but as we will remove it soon, it's ok to let be.

valentijnscholten (approved):

valentijnscholten:
thanks for the cleanup, didn't realize it was so much code!

PR #3994, opened 2021-03-04
Updates to Bulk Edit for Findings

blakeaowens:
Bulk Edit for findings allowed for contradictory selections (e.g. "Active" and "Mitigated" simultaneously). Added functionality similar to Bulk Edit for Endpoints to eliminate this.


Maffooch (approved):

valentijnscholten (approved):

valentijnscholten:
@blakeaowens Just noticed after merging that `view_test.html` contains the same logic and should be updated as well. Could you make another PR maybe?
At some point we need to deduplicate these 2 ways of displaying a list of findings, but currently we have to maintain both pages.

PR #3998, opened 2021-03-04
Make the Django session duration configurable

### LINKED ISSUE #3970: Make the Django session duration configurable
Currently the Django session cookie is valid for 14 days (this is the default value).
I would like to be able to lower this value.

The Django session duration can be changed by setting the `SESSION_COOKIE_AGE` variable.
It would be nice if this value could be set as an environment variable.

An interesting discussion about this: https://stackoverflow.com/questions/14830669/how-to-expire-django-session-in-5minutes/14830874

bgoareguer:
Fixes #3970


valentijnscholten (approved):

madchap (approved):
Thanks :)

PR #4008, opened 2021-03-05
Rename db migration for ipscan removal

StefanFl:
The order of db migrations was not correct anymore after 3 parallel changes


valentijnscholten:
ah, there was a migration hidden in those 42 files of #3993 :-)
I merged to soon, but in my defense, there was no 'New Migration' label ;-)
I think there's a GHA we could use to automatically label our PRs based on files changed.

PR #4013, opened 2021-03-05
Anchore Grype parser: add deduplication

damiencarol:
Work done:
- [x] activate de-duplication algorithm
- [x] add more details on findings like paths that match for component detection
- [x] add more unit tests


valentijnscholten:
unable to approve, getting github error.

Maffooch (approved):

valentijnscholten (approved):

valentijnscholten:
added hash_code_update_needed label, maybe we should already add something to the upgrading docs now to say that `dedupe` command must be run for the anchore grype parser? that will be possible with #3753

damiencarol:
@valentijnscholten not sure it is mandatory as this parser is configured to use uniq id from tool (DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL)

opperpanter:
Goog point, forget my comment :-)

valentijnscholten:
Good point, forget my comment, shouldn't review PRs before coffee :-)

PR #4014, opened 2021-03-05
import-scan not preserving active and verified statuses from findings

michaelgibson:
When using import-scan through API, the active and verified statuses are getting statically assigned here:
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/api_v2/serializers.py#L1176-L1177
which gets defaulted to True here:
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/api_v2/serializers.py#L1065-L1066

The result is that all findings get created with active=True and verified=True regardless of what the parser is setting for those.
This PR seems to workaround the issue for ImportScanSerializer.

ReImportScanSerializer may or may not also be affected by this but I have not verified. If so there may be another PR necessary for that as well.


Maffooch (changes requested):
Hi @michaelgibson please base PR's onto the `dev` branch from now on. This time wasn't too bad, but you'll need to remove the extra stuff that is not yours. Unit tests will likely pass afterwards too!

The change seems fine though. Will approve after all tests pass!

michaelgibson:
> Hi @michaelgibson please base PR's onto the `dev` branch from now on. This time wasn't too bad, but you'll need to remove the extra stuff that is not yours. Unit tests will likely pass afterwards too!
> 
> The change seems fine though. Will approve after all tests pass!

Merged dev. Unit tests still failing. Strange as this change should not impact unit tests.

valentijnscholten:
You're changes are affecting the way `verified` and `active` are handled so I'm pretty sure that's what's causing the unit tests failures.

michaelgibson:
Tests are failing in the import/re-import checks(and other places)
https://github.com/DefectDojo/django-DefectDojo/runs/2057616771?check_suite_focus=true#step:8:50492

I think the tests are failing now that the serializer is respecting the value of active/verified.
Ex. this test is searching Verified=True
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/unittests/test_import_reimport.py#L160
This would have always returned results before since it would have been forced True in serializer.

If we do not force Verified to True in the serializer, The test cases that get returned by:
https://github.com/DefectDojo/django-DefectDojo/blob/6a2b0e5939155554a284f90f58e6287632436ce8/dojo/unittests/dojo_test_case.py#L436
show `verified: False` by default and return 0 results for that test.

I was going to suggest modifying the test cases but I do not know where/how these are stored or where the False value comes from.
Not sure how to approach this.

michaelgibson:
Another side-effect I am noticing is that this change "breaks" quite a few of the other parsers.
Because most(all?) parsers were developed against a serializer that was marking everything active=True/verified=True, they may not be expecting it to no longer "correct" them.

The Nessus parser for example:
It is marking every finding active=False/verified=False 
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/nessus/parser.py#L207-L208
This currently has no effect since the serializer is overwriting those. If this PR is merged, then all Nessus findings will start getting imported as inactive.

There are possibly/probably other consequences to this change. I think this probably needs further review.

valentijnscholten:
I'm not even sure if I fully understand the problem you're trying to fix. Is it:

- active/verified set by parsers get overwritten?
- active/verified set by api client gets overwritten?

I think in general the default for active/verified everywhere is True. So it might be that any item arriving at the serializer will always have a value and is never None, which would lead to ignoring the values provided in the request to the API?
How does this work in the UI import code which is almost identical?

michaelgibson:
> I'm not even sure if I fully understand the problem you're trying to fix. Is it:
> 
>     * active/verified set by parsers get overwritten?
Yes
Active is getting overwritten to `data['active']` here.
https://github.com/DefectDojo/django-DefectDojo/blob/24bc536217fdbeac282c6b6019d9f1cbddcd39c5/dojo/api_v2/serializers.py#L883

`data = self.validated_data`
https://github.com/DefectDojo/django-DefectDojo/blob/24bc536217fdbeac282c6b6019d9f1cbddcd39c5/dojo/api_v2/serializers.py#L881
self.validated_data does not inherit what the parser sets. It appears to always be True by default.

When the parser gets called:
It is passing the value of active(still True at this point)
https://github.com/DefectDojo/django-DefectDojo/blob/24bc536217fdbeac282c6b6019d9f1cbddcd39c5/dojo/api_v2/serializers.py#L927

The parser "could" set the value of active to something else...
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/nessus/parser.py#L207-L208

But this has no effect since later in the serializer it continues setting active to data['active'] again overwriting whatever is returned from parsers.
https://github.com/DefectDojo/django-DefectDojo/blob/24bc536217fdbeac282c6b6019d9f1cbddcd39c5/dojo/api_v2/serializers.py#L954

>     * active/verified set by api client gets overwritten?
No
> 
> I think in general the default for active/verified everywhere is True. So it might be that any item arriving at the serializer will always have a value and is never None, which would lead to ignoring the values provided in the request to the API?

The default is True for data['active'] but not necessarily item.active which is what gets returned from the parsers.

https://github.com/DefectDojo/django-DefectDojo/blob/24bc536217fdbeac282c6b6019d9f1cbddcd39c5/dojo/api_v2/serializers.py#L954

It may be True, False or None. But it doesn't really matter at the moment since it is overwritten.

item.active is what is being checked for None in this PR.

Many parsers were developed against this behavior. "Fixing" it may cause undesired effects for them.
The Nessus parser is defaulting everything to False for some reason.
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/nessus/parser.py#L207-L208

This results in all Nessus imports defaulting to False.


> How does this work in the UI import code which is almost identical?
Almost but not quite...
The UI does not appear to use this serializer class:
https://github.com/DefectDojo/django-DefectDojo/blob/24bc536217fdbeac282c6b6019d9f1cbddcd39c5/dojo/api_v2/serializers.py#L855
It uses the `import_scan_results` method from engagement views.py:
https://github.com/DefectDojo/django-DefectDojo/blob/24bc536217fdbeac282c6b6019d9f1cbddcd39c5/dojo/api_v2/serializers.py#L855

Here is where it loads the parser.
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/engagement/views.py#L582
If you continue below there into where it loads the values from the parser items:
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/engagement/views.py#L603

You can see there is no place where active or verified is overwritten. meaning they are preserved from the parser.

valentijnscholten:
> item.active is what is being checked for None in this PR.

Well, it breaks lots of testcases so something is going wrong there. You either have to fix the code or the tests. Or create a test showing the problem so we can better see how to fix it. To me it sounds more like we need to remove the defaults from the serializer, but haven't looked at the order in which these fields are set.

michaelgibson:
Removing the defaults would work too but still result in the same side-effects.
I agree, The testcases need to be fixed/updated to align with this change. I mentioned previously I do not know where/how the testcases get their active/verified statuses set. Can you point me in a direction? I can work on that part.

Assuming the unit tests pass, I was additionally pointing out that this change will likely cause other issues with most if not all parsers.

valentijnscholten:
For the failing tests the filenames / test method names can be seen in the logs: https://github.com/DefectDojo/django-DefectDojo/runs/2057616771?check_suite_focus=true

So it's mainly `test_import_reimport.py` and `test_import_and_pushing_api`. Lots of failures, but in the end it might just be one small thing to fix to get the desired behaviour.

michaelgibson:
I'll use this one as an example:
https://github.com/DefectDojo/django-DefectDojo/runs/2057616771?check_suite_focus=true#step:8:50489

In this case, the dojo.unittests.tools.test_veracode_parser.TestVeracodeScannerParser is returning tests as active/verified "False" since that is what it is(and always has been) set to do: 
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/veracode/parser.py#L205-L206
The serializer is no longer forcing True for these results.

Do we "Fix" the test to verified=False for this test to match the parser? Or do we "Fix" all the parsers to default to active/verified=True if not specified?

valentijnscholten:
In this case the test is sending active/verified as True, so it should overwrite anything set by the parser.

So I would create one or more new testcases that handle the scenario when the request to the API doesn't contain active/verified, in which case it should retain the value from the parsers.

valentijnscholten:
But if we want to mimic that in the UI, we would indeed need to have a third option in the UI to retain the default from the parser. 

I think the whole system of active/verified and their defaults is weird currently. Either way, we should have the option to "leave them as default" next to True or False which we already have. It's the first time we've had any reports about this, so not sure how often people see the current behavior as a problem.

michaelgibson:
> In this case the test is sending active/verified as True, so it should overwrite anything set by the parser.
This does not overwrite what is set by the parser. That is the whole problem I am trying to solve for with this PR.

The test findings get imported at this point:
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/unittests/test_import_reimport.py#L139

Which eventually calls this setting `active=True, verified=True, scan_type='ZAP Scan'`:
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/unittests/test_import_reimport.py#L685
All good so far.

if we go to the zap scan parser you can see it is hard coded to set all findings to False/False:
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/zap/parser.py#L82-L83
Completely ignoring whatever was just passed into it.
In the past(and currently) the Serializer would normally then re-overwrite this to be True, True
But with this change, the parsers settings are preserved.
This results in findings in the testdb as active: False, verified: False

So it only appears as though things are testing correctly in the past when in fact the parser was actually setting values opposite the test.

> But if we want to mimic that in the UI, we would indeed need to have a third option in the UI to retain the default from the parser.

Personally, I am only concerned with `import-scan` from the API. Not trying to mimic anything in the UI.

> I think the whole system of active/verified and their defaults is weird currently. Either way, we should have the option to "leave them as default" next to True or False which we already have. It's the first time we've had any reports about this, so not sure how often people see the current behavior as a problem.

This is affecting ALL parsers when using `import-scan`. It is weird that no one has brought it up before but perhaps they are happy with everything importing as active/verified True. 
This is not the case for us. veracode and gitlab findings have statuses we want to preserve for example. this issue is preventing us from using the server side parsers and forces us to parse client-side, then post individual findings which is extremely time-consuming.

michaelgibson:
Attaching simple test case CSV
both Active and Verified are set to False for finding.
Expectation: 
Finding will have status inactive when using "Generic Findings Import"
Result: 
Finding is Active, Verified

submitting through UI, finding remains "inactive"

[4014.csv.txt](https://github.com/DefectDojo/django-DefectDojo/files/6118943/4014.csv.txt)

michaelgibson:
I've removed the defaults for active/verified from the parsers leaving it to the serializer to set that. (True if None)
This seems to satisfy the existing unit tests except for one.
The sonarqube parser which was explicitly testing the parser's active/verified status.
I've changed the test for that to expect True now that the serializer is setting it.

valentijnscholten:
Very good improvement to remove active and verified being set by the parsers. And wow, I just only now noticed the active and verified flags are sent to the parser factory, but they are not pass on to the parser. That also explains the confusion. So now I don't know what the previous authors ideas was on how this should be handled :-)

Anyway, at the point in the serializer where you check for 'item.active is None', item.active will always have a value.
Either False set by the parser, or True set by the default on the model. So I don't think it's working as intended.

Please note the UI counterpart would need to have the same logic as the serializer

UI Import:
Importing a scan via the UI. Should probably get the same change
 https://github.com/DefectDojo/django-DefectDojo/blob/6a2b0e5939155554a284f90f58e6287632436ce8/dojo/engagement/views.py#L534




I would maybe expect maybe three modes, of which the first 2 actively used by people:
1) I want to chose the values for active/verified in the UI or API request
2) I want to use the defaults from the parser (which gets the default from the model).
3) I want to use the defaults from the parser, and if there is no default from the parser (None), use the parameter from the API request / UI.

I think this PR now implements (or tries to implement) only option 3?

Should we have 3 choices for both active / verified in the API / UI? 
None -> use default from parser
True -> always set to True
False -> always set to False

Or should we only implement / support method 3?

michaelgibson:
> Anyway, at the point in the serializer where you check for 'item.active is None', item.active will always have a value.
> Either False set by the parser, or True set by the default on the model. So I don't think it's working as intended.

Not necessarily. 
item.active is what may or may not get set in the parser for each individual finding(item) from the file getting posted/parsed.
There is nothing requiring that field to exist at the individual finding level.
And since I just removed all the parsers' defaults, this may now indeed end up being None.
This is not the same "active" that gets set in the POST "data" block for setting ALL findings to active(which is also an optional argument).
That "active" is what data['active'] returns at the serializer level.

> Please note the UI counterpart would need to have the same logic as the serializer
> 
> UI Import:
> Importing a scan via the UI. Should probably get the same change
> 
> https://github.com/DefectDojo/django-DefectDojo/blob/6a2b0e5939155554a284f90f58e6287632436ce8/dojo/engagement/views.py#L534

The UI does not seem to be affected by this change since it is using the same ImpotScanForm the serializer uses. 
Meaning data['active'] will still default to True or accept whatever is checked in UI

> I would maybe expect maybe three modes, of which the first 2 actively used by people:
> 
>     1. I want to chose the values for active/verified in the UI or API request
> 
>     2. I want to use the defaults from the parser (which gets the default from the model).
> 
>     3. I want to use the defaults from the parser, and if there is no default from the parser (None), use the parameter from the API request / UI.
>
> 
> I think this PR now implements (or tries to implement) only option 3?
> 
> Should we have 3 choices for both active / verified in the API / UI?
> None -> use default from parser
> True -> always set to True
> False -> always set to False
> 
> Or should we only implement / support method 3?

I'm actually trying to implement option 4. Set active/verified from field values of individual findings within the file/scan being posted.

I'll try not to go too far off topic here but this PR is a dependency for my next "potential" PR...A bulk findings parser.
For some tools, we prefer(or have no choice) to parse scans client side. 
Currently the /findings endpoint only accepts individual entries. 
I was going to look into adding a bulk findings endpoint to the API but utilizing import-scan/parsers method seemed simpler.
I've created a "JSONParser" that accepts a list of finding dicts and loops through them.
Each of these have their own active/verified fields that I would like to persist but are currently overwritten by the serializer.

valentijnscholten:
> item.active is what may or may not get set in the parser for each individual finding(item) from the file getting posted/parsed.
> There is nothing requiring that field to exist at the individual finding level.
> And since I just removed all the parsers' defaults, this may now indeed end up being None

I don't think it works that way. The parser generates a list of Finding models. If the parser doesn't modify the `item.active` or `item.verified` flags, they will by `True` by default. I have tested this today, `item.active` will never be `None` in the serializer.

For example when I set `active=False`, it will end up as `active=True` in the database. I added a testcase that shows this: https://github.com/DefectDojo/django-DefectDojo/pull/4014/commits/ba7723933b72b70feadf2e6c50e0b6a78ec96771
This test is currently failing because `item.active` is never None in the serializer.

> I'm actually trying to implement option 4. Set active/verified from field values of individual findings within the file/scan being posted.

I think this the same is my option 3, although I phrased it not clear enough.

"Use the value set by the parser. If the parser doesn't set any value, use the value from `data['active']`".

I think this might be a good default as these are really corner cases for just a couple of parsers. So I think we should implement only your option 4 (which equals my option 3 :-)).

michaelgibson:
> I have tested this today, `item.active` will never be `None` in the serializer.

Unless...you are explicitly setting `"active": None` in your data...Which I was doing for some reason.
So bad testing on my part and now I see the default in the Model.

> "Use the value set by the parser. If the parser doesn't set any value, use the value from data['active']".

So for this to work I will have to figure out how to determine if the parser is setting it or if it's just using the default.

damiencarol:
@michaelgibson @valentijnscholten I'm following the discussion on this PR since few days and I strongly support option 3.
I also think your commit to remove all the `active=False` in every parsers is the best thing to do here.
Thanks a lot for pushing this.

> Very good improvement to remove active and verified being set by the parsers. And wow, I just only now noticed the active and verified flags are sent to the parser factory, but they are not pass on to the parser. That also explains the confusion. So now I don't know what the previous authors ideas was on how this should be handled :-)

Saw that few weeks ago. Yes the code around parsers doesn't make sense for some part. I think we will have more PR like this one in the future. And it's a good thing.

I'm waiting for green to approve this one :D

Also it could be good to describe option 3 somewhere in the doc to make it explicit to everyone.

dojo/api_v2/serializers.py
```diff
@@ -1121,8 +1121,12 @@ def save(self, push_to_jira=False):
...
                 item.last_reviewed = timezone.now()
                 item.last_reviewed_by = self.context['request'].user
-                item.active = data['active']
-                item.verified = data['verified']
+
+                if item.active:
+                    item.active = data['active']
+                if item.verified:
+                    item.verified = data['verified']
+
```
valentijnscholten, line 1129: I think this will work, maybe we should put a small comment in here to explain "Only set active / verified flags if they were not changed by the parser from their default value (True)"?

Also we need to add the same logic to the UI import.

valentijnscholten (changes requested):
see comment, and, could we have a test case covering this situation? so a test that imports a scan which has a finding marked as inactive and then check after the import check if it is still inactive?

valentijnscholten:
@michaelgibson are you able to resolve the conflicts? I think it's a good PR to merge and I am about to make some bigger changes to the import/reimport code, so would be good to merge yours before that to avoid more conflicts :-)

michaelgibson:
> @michaelgibson are you able to resolve the conflicts? I think it's a good PR to merge and I am about to make some bigger changes to the import/reimport code, so would be good to merge yours before that to avoid more conflicts :-)

Yup, Will do!

valentijnscholten:
Just noticed this line: https://github.com/DefectDojo/django-DefectDojo/blob/96f2f34814b46d83fec85c42c88d01e291ea64ec/dojo/engagement/views.py#L638. Looks like in the past somebody else started on a different solution. But I think this current PR is better, just make sure to remove also that helper method if it's not used anywhere anymore.

michaelgibson:
> Just noticed this line:
> 
> https://github.com/DefectDojo/django-DefectDojo/blob/96f2f34814b46d83fec85c42c88d01e291ea64ec/dojo/engagement/views.py#L638
> . Looks like in the past somebody else started on a different solution. But I think this current PR is better, just make sure to remove also that helper method if it's not used anywhere anymore.

Yep, should already be part of PR.

valentijnscholten:
@michaelgibson thanks, just pinging you to let you know you have some flake8 violations and failing unit tests.

michaelgibson:
> @michaelgibson thanks, just pinging you to let you know you have some flake8 violations and failing unit tests.

What I get for using the UI to resolve conflicts...
Looks all green now.

valentijnscholten (approved):

Maffooch (approved):

PR #4015, opened 2021-03-06
add magical illusionary roadmap

valentijnscholten:
adds some high level topics we are working on, without promising anything. still good to share some info.


Maffooch (approved):
Can we get a time estimate on these potential action items 😉

opperpanter:
Actually I almost added something like: "Anyone who asks about timelines will be banned from GitHub" ;-)

valentijnscholten:
Actually I almost added something like: "Anyone who asks about timelines will be banned from GitHub" ;-)

damiencarol (approved):

PR #4016, opened 2021-03-07
Add checkmarx OSA parser and cvss3 score. Make some fields optional

ptrovatelli:
Add a parser for checkmarx OSA "Open Source Analysis". 

Tool by checkmarx that analyzes dependencies (front and back). 
(the other "checkmarx" parsers are SAST parsers = analysis of source code)

db migrations: 
- add a field for cvssv3 score.
  - when the vector is given but not the score, the score is automatically computed upon finding save
  - it's possible to give the score without the vector for parsers that have only the score
- add a field for the vulnerability publish date. helps answer the question "Did I add this vulnerability or was it discovered recently?"

other:
- makes impact and mitigation optional
- makes  reference and url really optionnal (when empty, a hardcoded text was added)


dojo/tools/checkmarx_osa/parser.py
```diff
@@ -0,0 +1,95 @@
...
+    # 100% = Certain
+    # 70% = Firm
+    def checkmarx_confidence_to_defectdojo_confidence(self, checkmarx_confidence):
+        return round((100 - checkmarx_confidence) / 10) + 1
+
+    def safe_use_str(self, item, key):
+        return item[key] if key in item else ""
+
+    # Default to the max score so that we notice if it ever happens that we don't have a score
+    def safe_use_score(self, item, key, cve):
```
damiencarol, line 90: I think it's better to make your new column null-able and get ride of this function

dojo/tools/checkmarx_osa/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'SPoint42'
```
damiencarol, line 1:
```suggestion
```
Meta attribute __author__ are useless as we are tracking authorship with our VCS.

dojo/tools/checkmarx_osa/parser.py
```diff
@@ -0,0 +1,95 @@
...
+        items = []
+        for item in vulnerabilities:
+            library = libraries_dict[item['libraryId']]
+            # Possible status as per checkmarx 9.2: TO_VERIFY, NOT_EXPLOITABLE, CONFIRMED, URGENT, PROPOSED_NOT_EXPLOITABLE
+            status = item['state']['name']
+            finding_item = Finding(
+                title='{0} {1} | {2}'.format(library['name'], library['version'], item['cveName']),
+                severity=item['severity']['name'],
+                description=item['description'],
+                unique_id_from_tool=item['id'],
```
damiencarol, line 36: Could we add 'libraryId' as vuln_type_id_from_tool?

dojo/settings/settings.dist.py
```diff
@@ -844,6 +846,7 @@ def generate_url(scheme, double_slashes, user, password, host, port, path, param
     'Anchore Engine Scan': DEDUPE_ALGO_HASH_CODE,
     'Checkmarx Scan detailed': DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL,
     'Checkmarx Scan': DEDUPE_ALGO_HASH_CODE,
+    'Checkmarx OSA': DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE,
```
damiencarol, line 849: As you always set uniq id, you should switch to DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL
```suggestion
    'Checkmarx OSA': DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL,
```

dojo/fixtures/test_type.json
```diff
@@ -719,6 +719,13 @@
         },
         "model": "dojo.test_type",
         "pk": 194
+    },
```
damiencarol, line 722: this is useless

damiencarol (changes requested):

dojo/tools/checkmarx_osa/parser.py
```diff
@@ -0,0 +1,95 @@
...
+    #   3->5 = Firm
+    #   >=6 : Tentative 
+    #   -> defectdojo has a quotation from 1->(say)11 with 1 the highest confidence
+
+    # 100% = Certain
+    # 70% = Firm
+    def checkmarx_confidence_to_defectdojo_confidence(self, checkmarx_confidence):
+        return round((100 - checkmarx_confidence) / 10) + 1
+
+    def safe_use_str(self, item, key):
```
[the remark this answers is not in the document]
ptrovatelli: thanks! it'll be lighter indeed.

dojo/tools/checkmarx_osa/parser.py
```diff
@@ -0,0 +1,95 @@
...
+            library = libraries_dict[item['libraryId']]
+            # Possible status as per checkmarx 9.2: TO_VERIFY, NOT_EXPLOITABLE, CONFIRMED, URGENT, PROPOSED_NOT_EXPLOITABLE
+            status = item['state']['name']
+            finding_item = Finding(
+                title='{0} {1} | {2}'.format(library['name'], library['version'], item['cveName']),
+                severity=item['severity']['name'],
+                description=item['description'],
+                unique_id_from_tool=item['id'],
+                references=self.safe_use_str(item, 'url'),
+                mitigation=self.safe_use_str(item, 'recommendations'),
```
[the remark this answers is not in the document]
ptrovatelli: you're right it's not nullable; it's a long string though, how about something like "NC"? (not communicated)

dojo/tools/checkmarx_osa/parser.py
ptrovatelli: yeah I wasn't too happy about this part either. actually it's already nullable I just wasn't sure how to handle it without too much code. i'll try to create the finding with None see if that works.

dojo/tools/checkmarx_osa/parser.py
ptrovatelli: you're suggesting added a field to track the technical id from checkmarx OSA that identifies that specific library ? to what purpose? the component_name and component_version should be enough for us I believe

dojo/settings/settings.dist.py
ptrovatelli: although the unique_id will always be there, the hash_code can be used for cross-parser deduplication. this is still experimental and will probably require a bit of standardization, still this configuration is the first step :)

dojo/fixtures/test_type.json
ptrovatelli: should we remove this file then?

dojo/tools/checkmarx_osa/parser.py
ptrovatelli: works like a charm :)

dojo/tools/checkmarx_osa/parser.py
```diff
@@ -0,0 +1,95 @@
...
+        for item in vulnerabilities:
+            library = libraries_dict[item['libraryId']]
+            # Possible status as per checkmarx 9.2: TO_VERIFY, NOT_EXPLOITABLE, CONFIRMED, URGENT, PROPOSED_NOT_EXPLOITABLE
+            status = item['state']['name']
+            finding_item = Finding(
+                title='{0} {1} | {2}'.format(library['name'], library['version'], item['cveName']),
+                severity=item['severity']['name'],
+                description=item['description'],
+                unique_id_from_tool=item['id'],
+                references=self.safe_use_str(item, 'url'),
```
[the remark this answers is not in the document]
ptrovatelli: i'll go for `item.get('url', None)` which will insert "null" in the database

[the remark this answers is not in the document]
damiencarol: if your default value is None you can use the simple form `item.get('url')`. It use None as default value without raising an exception.

dojo/tools/checkmarx_osa/parser.py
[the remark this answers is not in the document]
damiencarol: I suggest this exact string as **a lot** of parsers use this exact string for exactly the same use case.

dojo/tools/checkmarx_osa/parser.py
damiencarol: Maybe I didn't understand well what 'libraryId' is. Forget my comment.

dojo/settings/settings.dist.py
damiencarol: Regarding this specific topic, as I know more well parsers, I really think that relaying on the hash code for cross-parser de-duplication will be really hard. Most of the parsers use too much specific data for the hash code like titles that are very different between parsers.
The only reliable columns that could help us to do that will be **CVE**.

dojo/fixtures/test_type.json
damiencarol: It break some unit tests/integration tests. You could push some modifications to do that.

dojo/tools/checkmarx_osa/parser.py
```diff
@@ -0,0 +1,99 @@
...
+            # Possible status as per checkmarx 9.2: TO_VERIFY, NOT_EXPLOITABLE, CONFIRMED, URGENT, PROPOSED_NOT_EXPLOITABLE
+            status = item['state']['name']
+            cve = item.get('cveName', 'NC')
+            finding_item = Finding(
+                title='{0} {1} | {2}'.format(library['name'], library['version'], cve),
+                severity=item['severity']['name'],
+                description=item.get('description', 'NC'),
+                unique_id_from_tool=item.get('id', None),
+                references=item.get('url', None),
+                mitigation=item.get('recommendations', 'NC'),
```
damiencarol, line 48:
```suggestion
                mitigation=item.get('recommendations', 'No mitigation provided'),
```

dojo/settings/settings.dist.py
ptrovatelli: some parsers have cve, others have cwe. you also need to add the component name, or file path to identify the vulnerability because CVE only will match many different findings. For checkmarx OSA i have choose the component name and version, which should be more or less standard and could be the same as other parsers (I haven't checked).  All this information that makes it possible to identify a vulnerability should  go into the hash_code so that cross-parser deduplication can maybe work one day. I agree that many parsers have a poorly configured hash_code but that can always be changed later. 

 I think CVE only cannot work so I see only two solutions: 
1) use the hash_code
2) add another hash_code or something similar. specifically for cross-parser dedupe. that might make sense, we could discuss this but we'll need to change the deduplication to use that

damiencarol: It's very interesting to understand your point of view! Thanks for sharing!
2 things I saw when I deep dive into parsers code:
1. component_name is very new and most of parser imp were wrong. (sometime mixing vendor/name/version in this attribute). I fixed few but for now every parser use it in different ways. Less reliable than CVE :/
2. parsers implement things in very different ways. for ex title are implemented differently even with the same attributes! So different implementation of fields will lead to completely different and incomparable hash codes. we will need a document that define what the "perfect" hash code should be, and what the attributes should contains and fix every parsers to follow this spec to have something that follow this spec. near impossible goal IMO.

Regarding component name and version, the work of fixing them is not done.

dojo/tools/checkmarx_osa/parser.py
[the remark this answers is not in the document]
ptrovatelli: ok ok... how about making the field nullable :) . the fact that many parsers use this string proves that the field should be nullable. null takes less room in db than any other string, and has the universal meaning of : "I don't have that information right now". 
for checkmarx OSA it doesn't really matter as it seems it won't be missing often, but there are probably some parsers that will always fill in "No mitigation provided" string. after 1M rows the space taken will start to matter...

dojo/models.py
```diff
@@ -1687,6 +1689,17 @@ class Finding(models.Model):
                                         verbose_name="Number of occurences",
                                         help_text="Number of occurences in the source tool when several vulnerabilites were found and aggregated by the scanner.")
 
+    # this is useful for vulnerabilities on dependencies : helps answer the question "Did I add this vulnerability or was it discovered recently?"
+    publish_date = models.DateTimeField(null=True,
+                                         editable=False,
```
valentijnscholten, line 1694: should this be editable? Many parsers don't set this field and people may want to set it during triage, inspecting reference urls, etc.?

dojo/tools/checkmarx_osa/parser.py
```diff
@@ -0,0 +1,99 @@
...
+                impact='No impact provided',
+                component_name=library['name'],
+                component_version=library['version'],
+                cve=cve,
+                # 1035 is "Using Components with Known Vulnerabilities"
+                # Possible improvment: get the CWE from the CVE using some database?
+                # nvd.nist.gov has the info; see for eg https://nvd.nist.gov/vuln/detail/CVE-2020-25649 "Weakness Enumeration"
+                cwe=1035,
+                cvssv3_score=item.get('score', None),
+                publish_date=datetime.strptime(item['publishDate'], '%Y-%m-%dT%H:%M:%S'),
```
valentijnscholten, line 58: are you sure this field is always there?

valentijnscholten (changes requested):

dojo/models.py
ptrovatelli: good point, I hadn't noticed this field

dojo/tools/checkmarx_osa/parser.py
ptrovatelli: nope; adding a verification...

dojo/models.py
ptrovatelli: done. made the score editable too. However i'm thinking now that we should do something with the vector and the score: if the vector is given by the parser or the user, the score should be updated automatically. 
(If the score only is given, the vector may remain empty because the score only is not enough to know the full vector).

dojo/models.py
```diff
@@ -2086,6 +2087,13 @@ def save(self, dedupe_option=True, false_history=False, rules_option=True, produ
 
         # Assign the numerical severity for correct sorting order
         self.numerical_severity = Finding.get_numerical_severity(self.severity)
+
+        # Synchronize cvssv3 score using cvssv3 vector
+        if self.cvssv3:
+            cvss_object = CVSS3(self.cvssv3)
+            # use the environmental score, which is the most refined score
+            self.cvssv3_score = cvss_object.scores()[2]
```
damiencarol, line 2095: why not the base score ?

ptrovatelli: it goes like: 
- base score
- if temporal info is given, temporal score is computed from base score and temporal info (else temporal score = base score)
- if environmental info is given, environmental score is computed based on temporal score and environmental info (else environmental score = temporal score)

so the environmental score is the most precise, or at worst equal to base score.

damiencarol: good to know. seems the best way to manage it!

damiencarol (approved):

damiencarol:
@ptrovatelli validated but you will need to fix/check the unit tests.

ptrovatelli:
@valentijnscholten all good for me on this one :)

valentijnscholten:
I feel that sneaking in switching some fields from required to optional might be a surprise for some people. Would it be possible to create a small companion PR for that? I think there are or at least have been people relying on those fields to be mandatory for their processes. Having it in its own PR will show it on the release notes and allows people to give their opinion/review.

ptrovatelli:
@valentijnscholten that's gonna be a lot of work because it all over the place: in the migrations, in the integration tests, in specific code that generates dummy defaults... I have already raised the topic on slack. I can create another conversation about this to make sure everyone notices? 

note that switching from mandatory to optional shouldn't be as big a problem as the opposite would be

ptrovatelli:
added a comment on the main channel; we can still mention the change in the release note

madchap (commented):
Let's have a special note in the release notes about the fields getting optional.

madchap (approved):

valentijnscholten:
The release notes are automatically generated. I suggest we create a new PR that only updates the docs, but as title has "Make mitigation, impact, url fields optional".

damiencarol:
@ptrovatelli @madchap @valentijnscholten should we split this PR in two part? one that modify these fields to make them optionals and the rest based on that?

ptrovatelli:
@damiencarol we agreed on creating a doc PR. just done now : https://github.com/DefectDojo/django-DefectDojo/pull/4122

I'll try to separate things from the start next time, but when it comes to migrations it is a bit of work to maintain several PR with migrations.

valentijnscholten (approved):

PR #4017, opened 2021-03-07
Allow grouping of Findings (and push to JIRA)

valentijnscholten:
This PR introduces the concept of groups of findings (Finding Group). The main reason to introduce this concept is to be able to group findings together before pushing them to JIRA as one JIRA ticket. For example when you have multiple vulnerabilities in a dependency. You still want to see all separate dependencies in Defect Dojo with their CVE, CWE, severity, etc. And also be able to use deduplication and reimport as usual. 

At first I tried using the merge functionality in Defect Dojo, but it doesn't always work well with deduplication nor with Reimport. So I chose to introduce this concept. It can be used for other stuff as well in the future, for now the main reason is to group vulnerabilities together into 1 JIRA ticket.

Some automatic grouping could be implemented in the future:
- automatically group findings by `component_name`
- additionally push these groups to JIRA automatically (or manually)
- group findings that are in the same advisory

To use the finding groups you need to enable a feature flag: `FEATURE_FINDING_GROUPS` or `DD_FEATURE_FINDING_GROUPS`

To make it work with the webhook, notes, jira pushing, findings list, etc. I touched quite a bit of code. The good thing is that all our unit tests and integration tests still work. The bad thing is that it will require quite a chunk of time to add tests for the finding groups feature. So that's why I have hidden it behind a feature flag, and propose to do some dogfooding in `dev` before we release it. So I suggest target release 1.15.0 end of April.
I also like to get some feedback from others before investing more time. I need to know this is something we want to keep and expand.

This PR also introduces a second django template, which is used to generate the description of the JIRA issue for the finding group: `issue-trackers/jira_full/jira-finding-group-description.tpl`. I have moved the 'full' and 'limited' templates in their own subfolders so we still have only one selection to make in the JIRA Project config of a Product/Engagement. This is a breaking change for people who have configured non-default JIRA templates in their instance. I expect this affects not many people.

I added some API methods / fields to be able to retrieve the finding_group for findings and filter on them.

Bugfixes along the way:
- #4118


madchap:
Locally tested push and also the webhook coming back, seems to work fine. Remains a couple of minor things to finalize that first iteration, but looks good so far :+1:  Thanks!

madchap:
Also, I would not show column "Group" is all findings are to the `N` value.

Maffooch (commented):
Looks great!

Maffooch (approved):

damiencarol (approved):

PR #4022, opened 2021-03-08
README: add wishlist

valentijnscholten:


madchap:
I am not convinced. I think we should possibly give users the option to thumb things up or down and such, even without talking about promising anything.

valentijnscholten:
It's all nice all those ideas for voting etc, but we will never have time for those. This wishlist is not meant as a way for users to submit their wishes, they can use github issues for that. It's meant as a way for the team to show what we think / wish for to be implemented. But I'm open to alternatives, as long as it's not something we will never be able to maintain.

Maffooch (commented):
I think this a good idea! It might be a good idea to include something like "these items are nothing more than thoughts and wishes. However, we encourage contributors to propose design plans for current wishes or new ones"

Something like that but more elegant

valentijnscholten:
Elegant is not compatible with me, so feel free to commit something to this PR.

devGregA:
I think this is a good edition, but I'm not sure the readme is the right place. Maybe we consider publishing a rough roadmap?

valentijnscholten:
there is a rough roadmap also in the README since a few weeks.

madchap (approved):
Let's push it for now, does not hurt.

PR #4024, opened 2021-03-08
Typo fix for Crital => Critical

SPoint42:
Typo fix for Crital => Critical

PR #4026, opened 2021-03-08
Fix gitleaks import when no findings found

SPoint42:


valentijnscholten:
This PR seems to contains also the changes from #4024 , so could you fix that first so we can merge that one?

SPoint42:
> This PR seems to contains also the changes from #4024 , so could you fix that first so we can merge that one?

✅

damiencarol (approved):

PR #4034, opened 2021-03-09
Authorization V2: Rename of feature flag

StefanFl:
As discussed on Slack. Thanks @dsever and @valentijnscholten for the suggestions, it is a better name now.


StefanFl:
That looks much better for the test cases, good to know the frameworks 😄

PR #4045, opened 2021-03-11
Netsparker: Fix crashing parser at method get_findings()

AsierRF:
The parser is not working (TypeError: object of type 'NoneType' has no len()), which triggers an HTTP 500 response. 
This is because the get_findings(self, filename, test) method defined for the Netsparker parser (line 30) does not return the list of findings as expected in "https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/engagement/views.py#L583" (line 583) "parser_findings = parser.get_findings(file, t)"
Just adding a return statement solves the import issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,8 +76,6 @@ A magical, illusionary, non-existent, YMMV, wannabe, no guarantees list of thing
 - New permission model
 - Push groups of findings to a single JIRA ticket
 - Reimport matching improvements
-- New UI (SPA on top of API)
-- More parsers (CycloneDX, OVAL)
 
 
 ## Support, Bug Reports and Getting Involved
--- a/README.md
+++ b/README.md
@@ -71,6 +71,15 @@ See [Wrappers](WRAPPERS.md)
 See [Release and branch model](BRANCHING-MODEL.md)
 
 
+## Roadmap
+A magical, illusionary, non-existent, YMMV, wannabe, no guarantees list of thing we may or may not be working on:
+- New permission model
+- Push groups of findings to a single JIRA ticket
+- Reimport matching improvements
+- New UI (SPA on top of API)
+- More parsers (CycloneDX, OVAL)
+
+
 ## Support, Bug Reports and Getting Involved
 Please come to our Slack channel first, where we can try to help you or point you in the right direction:
 
--- a/dojo/tools/netsparker/parser.py
+++ b/dojo/tools/netsparker/parser.py
@@ -79,5 +79,4 @@ class NetsparkerParser(object):
                 dupes[dupe_key] = find
                 findingdetail = ''
 
-        self.items = list(dupes.values())
-        return self.items
+        return list(dupes.values())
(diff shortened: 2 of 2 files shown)


damiencarol (approved):

valentijnscholten:
Thanks for the fix. There are 3 locations in the code where this currently results in an exception: 

`            logger.debug('starting reimport of %i items.', len(items))`

but if we fix those, it would probably crash later on:

`            for item in items:`

Should we add a check in the next pieces to make sure we throw a clear error when the `parser_findings == None`?

```
        try:
            parser = import_parser_factory(data.get('file', None),
                                           test,
                                           active,
                                           verified,
                                           data['scan_type'])
            parser_findings = parser.get_findings(data.get('file', None), test)
        except ValueError:
            raise Exception("Parser ValueError")
        except:
            raise Exception('Error while parsing the report, did you specify the correct scan type ?')
```

And while we're at it, make sure that we include the root cause when raising `Exception("Parser ValueError")`?

damiencarol:
@valentijnscholten I think parsers should all the time return a list. Never `None`. At least it's what I made in the many refactoring I made on parsers.
For your comment on the exception handling, I think we should just remove the complete `try/catch` block and don't swallow the exception.

valentijnscholten:
The problem is if you throw a ValueError, DRF will turn it into a 400 and hide the exception / traceback.
Yes parser should never return None, but they do. So it's better to have a clear error like 'Parser returned None` or something like that instead of an error related to printing a debug statement which crashes because it performs `len(None)`.

I see the base branch is `master`. This would be correct for bugfix release. Or do we want to include it in the regular 1.14.0 release?

damiencarol:
I don't think a lot of people use this parser so it could wait `1.14.0` IMO.
For the exception management, maybe we can add the root cause when raising the exception.
Let's merge this one and I will work on the part that call the parser methods.

valentijnscholten:
@AsierRF Could you rebase against `dev`? I tried to do at, but seem to have no permission to push to your PR.

AsierRF:
> 
> 
> @AsierRF Could you rebase against `dev`? I tried to do at, but seem to have no permission to push to your PR.

@valentijnscholten  I did the rebase into Dev. Can you verify everything is fine? Apparently there were 3 more commits being applied to Dev.

AsierRF:
@valentijnscholten Just out of curiosity, am I entitled to the Swag Rewards listed in the main page?

It would be cool to get the t-shirt :)

valentijnscholten (approved):
there's one commit too many, but it's OK it will end up in dev later anyway.

valentijnscholten:
> @valentijnscholten Just out of curiosity, am I entitled to the Swag Rewards listed in the main page?
> 
> It would be cool to get the t-shirt :)

At the moment we don't have any swag unfortunately :-( We did order a couple of t-shirts as a test, working on something later this year.

PR #4046, opened 2021-03-11
Update Aqua-scanner import for CVSS3 finding field

SPoint42:
Add CVSS3 score if exists.
Add unittests for it


damiencarol (approved):

valentijnscholten (approved):
thanks!

PR #4048, opened 2021-03-12
dedupe: select oldest (lowest id) finding as original

valentijnscholten:
The current deduplication logic uses the default ordering when looking identical existing findings: 

`        ordering = ('numerical_severity', '-date', 'title')`

Because we are performing deduplication asynchronously, this can lead to situation where finding_100 can be marked as a duplicate of finding_101 depending on the order chosen by the celeryworkers. This can be confusing especially if you have different engagements where a finding in the firstly imported engagement is a duplicate of one in the later imported engagement. (sometime multiple engagements are processed in parallel, or somebody runs the dedupe script, or just bulk edits some findings).

Apart from confusion, I think it can also lead to some timing isssues where for example finding_100 is being marked as duplicate of finding_101, but finding_101 is processed in parallel and is made a duplicate of finding_99. Depending on the timing again, this can lead to 'transitive duplicates': finding_100 -> finding_101 -> finding_99.

Real world example:

```
dojo=# select f.id, test_id, engagement_id, duplicate, duplicate_finding_id from dojo_finding f join dojo_test t on t
.id=f.test_id where duplicate_finding_id=1186900 or f.id=1186900
;
id    | test_id | engagement_id | duplicate | duplicate_finding_id
---------+---------+---------------+-----------+----------------------
1183386 |    7480 |          6232 | t         |              1186900
1186205 |    7494 |          6236 | t         |              1186900
1186900 |    7499 |          6237 | t         |              1069084
1187286 |    7500 |          6238 | t         |              1186900
```

My theory is that by always chosing the identical finding with the lowest id we get a more predictable result and reduce the chance of 'transitive duplicates'.

If anybody has experience with these scenario's, please let me know. I think this PR is an improvement in itself, I updated the testcases to check that the lowest id is chosen (instead of the finding with the highest (severity, date, title) combination).


dojo/management/commands/fix_loop_duplicates.py
```diff
@@ -21,6 +21,7 @@ def handle(self, *args, **options):
 def fix_loop_duplicates():
     candidates = Finding.objects.filter(duplicate_finding__isnull=False, original_finding__isnull=False).all().order_by("-id")
     deduplicationLogger.info("Identified %d Findings with Loops" % len(candidates))
+    # raise ValueError('b')
```
damiencarol, line 24:
```suggestion
```

valentijnscholten: oops, thanks!

ptrovatelli:
thanks  @valentijnscholten that seems like a very sensible change! shouldn't we need to apply this (the sorting) to the re-import too? the risk is lower as it's smaller sets of findings but it couldn't hurt. i think it's all in dojo/reimport_utils.py

valentijnscholten:
> thanks @valentijnscholten that seems like a very sensible change! shouldn't we need to apply this (the sorting) to the re-import too? the risk is lower as it's smaller sets of findings but it couldn't hurt. i think it's all in dojo/reimport_utils.py

Good point. In setups which are configured correctly, in reimport we should only get 1 finding as match. But there could be multiple, so I added the ordering to make it more predictable.

valentijnscholten:
BTW I think this PR doesn't full prevent the "older gets marked as duplicate of newer", it can still happen:
```
...
...
finding_100
finding_101
...
...
```
If during import finding_100 gets deduped and is identical to only finding_101, it will be marked as a duplicate of finding_101 (a newer finding, which could even be in a different engagement). It will be hard to prevent this. The more I look at dedupe, the more I think it should be made synchronous. It could still be in a celery task, but in serial order.

madchap (approved):
lgtm. Even if not perfect in the current state, reducing the uncertainty is always nice. As tested in my own scenarios, running the `fix_loop_duplicates` management command on a regular basis helps in fixing the transitive relationships that arise from the issues you're pointing out.

ptrovatelli (approved):

ptrovatelli:
> The more I look at dedupe, the more I think it should be made synchronous. It could still be in a celery task, but in serial order.

see this discussion about it: https://stackoverflow.com/questions/12003221/celery-task-schedule-ensuring-a-task-is-only-executed-one-at-a-time/12003293 
there seems to be two solutions
- easy one is using single threaded celery worker
- the other solution is setting up a lock: see https://docs.celeryproject.org/en/latest/tutorials/task-cookbook.html#ensuring-a-task-is-only-executed-one-at-a-time

valentijnscholten:
We could also do database locks, but that has deadlock risks etc. Using that celery lock makes the 8 workers done people have pretty useless 😀

madchap:
> We could also do database locks, but that has deadlock risks etc. Using that celery lock makes the 8 workers some people have pretty useless 😀

No. idea who you're talking about here... :-D

PR #4049, opened 2021-03-12
Engagement page performance

danielnaab:
This PR proposes some changes on the `/engagement` view, with the intention to improve performance. It does this by:

- basing the db queries on the Engagement table rather than the Product table
- optimizing joins and prefetches

Additionally, the HTML table was revamped to accommodate the db changes, with some layout changes.


valentijnscholten:
Thanks, long overdue change :-)
It looks like this page is also printing the count, which would result in lots of queries without prefetching. Could you look at applying the count from #4033 ?

danielnaab:
Hey @valentijnscholten - I had actually intended to add something about that to the ticket - we can reduce the query count by replicating that prefetch, but it adds quite a bit of time to the total page load count. Again, that's with our very large test table. But I can give it another go and collect some stats on it.

valentijnscholten:
I think some paging would be in order too, that will really help performance.

dojo/engagement/views.py
```diff
@@ -85,36 +85,62 @@ def engagement_calendar(request):
...
+    )
+    engs = get_page_items(request, filtered_engagements.qs, 25)
+    product_name_words = sorted(set(
+        engagement.product.name
+        for engagement in engs.object_list
+    ))
+    engagement_name_words = sorted(set(
+        engagement.name
+        for engagement in engs.object_list
+    ))
```
valentijnscholten, line 130: not sure, but I think the name words are now limited to the engagements on the current page only?

valentijnscholten:
Interesting approach with the separate count, never understood why django is not doing it that way. Probably because of the joins it can't know we just want the count for all engagements/tests.

dojo/engagement/views.py
danielnaab: Oh yeah you're right, I got overzealous removing queries. Will fix.

valentijnscholten (changes requested):
as per comment

danielnaab:
@valentijnscholten sorry, thought I did that already - added name queries back.

valentijnscholten (approved):

madchap (approved):

PR #4051, opened 2021-03-13
JIRA: fix webhook mitigated by

valentijnscholten:
Currently in dev there is an error when a finding is closed in JIRA and the webhook is closing the finding in defect dojo.
The new status helper tries to set `mitigated_by` to `get_current_user()`, but the webhook is anonymous so this fails.

Two fixes:
- make status helper robust by only using `get_current_user()` if the user is authenticated
- set `mitigated_by` to the JIRA user in the webhook.

### DIFF
--- a/dojo/jira_link/views.py
+++ b/dojo/jira_link/views.py
@@ -108,7 +108,7 @@ def webhook(request, secret=None):
                                 finding.active = False
                                 finding.mitigated = now
                                 finding.is_Mitigated = True
-                                finding.mitigated_by = User.objects.get_or_create(username='JIRA')
+                                finding.mitigated_by, created = User.objects.get_or_create(username='JIRA')
                                 finding.endpoints.clear()
                                 finding.false_p = False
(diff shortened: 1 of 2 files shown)


dojo/jira_link/views.py
```diff
@@ -108,6 +108,7 @@ def webhook(request, secret=None):
                                 finding.active = False
                                 finding.mitigated = now
                                 finding.is_Mitigated = True
+                                finding.mitigated_by, created = User.objects.get_or_create(username='JIRA')
```
damiencarol, line 111: if you don't use `created`make it explicit with `_`, no?
```suggestion
                                finding.mitigated_by, _ = User.objects.get_or_create(username='JIRA')
```

damiencarol (approved):
Check at my comment

dojo/jira_link/views.py
valentijnscholten: good point

damiencarol (approved):

madchap (approved):
Good catch.

PR #4054, opened 2021-03-13
add missing ordering for api filters

valentijnscholten:
some api endpoints didn't allow ordering the results, some endpoints could use some extra ordering options.


madchap (approved):
:+1: @jvz will like.

Maffooch (approved):

PR #4056, opened 2021-03-14
Fix SSLyze scanner: Remove unnecessary read

StefanFl:
There was an unnecessary read of the file with the scan results. Because of this read, the following read didn't get data anymore and therefore the scanner produced a server error.

### DIFF
--- a/dojo/tools/sslyze/parser.py
+++ b/dojo/tools/sslyze/parser.py
@@ -21,8 +21,6 @@ class SslyzeParser(object):
         if filename is None:
             return list()
 
-        content = filename.read()
-
         if filename.name.lower().endswith('.xml'):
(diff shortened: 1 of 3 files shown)


damiencarol (approved):

valentijnscholten:
When did the problem occur? Was it failing always or only in certain cases? Looking at the code it is a generic problem causing problems with every scan upload of type sslyze?

StefanFl:
It did fail always. The problem seem to be the 2 read's on the same file. The second read in the actual JSON or XML parser modul delivered no data anymore.  After removing the unnecessary read it works reliably fine.

valentijnscholten:
ok, pretty big bug. updated the title to make it more clear.

valentijnscholten (approved):

PR #4068, opened 2021-03-16
SonarQube api filters

dsever:
Sonar API imports only filter VULNERABILITY

This PR is extending configuration for the scanner to be able define extras to add additional filters like:

BUG,VULNERABILITY,CODE_SMELL


Maffooch (approved):
Good move to make things easier for the future integrations

dojo/models.py
```diff
@@ -838,6 +838,8 @@ class Tool_Configuration(models.Model):
                                                 'Username/Password'),
                                                ('SSH', 'SSH')),
                                            null=True, blank=True)
+    extras = models.CharField(max_length=50, null=True, blank=True, help_text="Additional definitions that will be "
```
valentijnscholten, line 841: just a thought, but might 50 be too small for some future tools? maybe use 255?

dsever: Done

damiencarol (approved):

PR #4070, opened 2021-03-16
Fix deduplication for Dependency Track scanner

SPoint42:
It's better to dedup on component-name + component-version + CVE + CWE for a dependency.


valentijnscholten (approved):

valentijnscholten:
agreed, vuln_id_from_tool is not unique

damiencarol (approved):

PR #4074, opened 2021-03-16
fix counts returned to non-staff users in apiv2

### LINKED ISSUE #4069: Fetching findings with staff or regular user does not yield the same results
**Bug description**
If I query the /findings endpoints for a `test`, I get different `count` depending on whether I am a staff/superuser or just a regular user authorized to the product. The actual responses are nearly identical, but even with http://www.jsondiff.com it's hard to tell. For sure, there is not 10380 findings count, there is only a few lines difference which I guess could be explained by potential issues in accessing certain data?

This issue does not arise in the UI. A regular user will see the proper count of findings.

Instance runs with

```
        DD_AUTHORIZED_USERS_ALLOW_CHANGE: "True"
        DD_AUTHORIZED_USERS_ALLOW_DELETE: "False"
        DD_AUTHORIZED_USERS_ALLOW_STAFF: "True"
```

As a regular user:
image

As an admin user:
image

**Steps to reproduce**

**Expected behavior**

- [x] Kubernetes
- [ ] Docker

**Environment information**
Latest `dev`.

valentijnscholten:
fix #4069 

For non-staff users, we have queries where we filter objects based on the `authorized_users` property of the product or product type:

```
    def get_queryset(self):
        if not self.request.user.is_staff:
            return self.queryset.filter(
                Q(test__engagement__product__authorized_users__in=[self.request.user]) |
                Q(test__engagement__product__prod_type__authorized_users__in=[self.request.user])
            )
        else:
            return self.queryset
```  

This works OK, but will join a couple of tables and could lead to objects being in the result multiple times. This PR adds `.distinct()` to resolve this.

There's also one place where we had 2 different queries for the list of findings and the count of that list.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -386,7 +386,7 @@ class ProductSerializer(TaggitSerializer, serializers.ModelSerializer):
         }
 
     def get_findings_count(self, obj):
-        return obj.findings_count
+        return obj.findings_count()
 
     def get_findings_list(self, obj):
         return obj.open_findings_list()
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -386,10 +386,10 @@ class ProductSerializer(TaggitSerializer, serializers.ModelSerializer):
         }
 
     def get_findings_count(self, obj):
-        return obj.findings_count()
+        return obj.findings_count
 
     def get_findings_list(self, obj):
-        return obj.open_findings_list()
+        return obj.open_findings_list
 
 
 class ProductTypeMemberSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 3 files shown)


valentijnscholten:
@StefanFl I didn't change the authorization v2 code, but there we might have a chance of duplicate items in the result as well.

StefanFl:
If it's save to use distinct() without other side effects then it would make sense to put it in there.

valentijnscholten:
> If it's save to use distinct() without other side effects then it would make sense to put it in there.

it's safe but needed probably because a join creates all possible combinations. side effect is that distinct() might make it slightly slower.

madchap (approved):
Thanks. Let's try that out.

damiencarol (approved):

damiencarol:
good catch!

PR #4075, opened 2021-03-17
Add an extra javascript block to base.html

Maffooch:


devGregA (approved):

valentijnscholten:
Won't harm anything, but is it needed for something? tbh I haven't really used these blocks and their possibility to overwrite blocks on dependant templates etc.

Maffooch:
Using it for a plugin at work to avoid having to maintain changes not in repo

valentijnscholten (approved):

PR #4079, opened 2021-03-18
Nexpose parser: Process all hosts + add Endpoint (+ add new reference links)

kiblik:
Used [for-loop](https://github.com/DefectDojo/django-DefectDojo/blob/6a2b0e5939155554a284f90f58e6287632436ce8/dojo/tools/nexpose/parser.py#L228) processed only last "discovered" host in XML file. I move it to the upper loop.

There was missing also adding Endpoint - so it wasn't possible to identify where vulnerability is.

I also added a couple of new references used in Nexpose reports.


kiblik:
Replaced by #4082

PR #4090, opened 2021-03-19
Fix title max length and add more tests

damiencarol:
This pull request fix few potential bugs on the size of title in findings for 3 parsers:
- AWS Prowler
- AWS Scout
- Scoutsuite

Also this pull request add unit tests for scout2 parser which doesn't exists (!)


valentijnscholten (approved):

valentijnscholten:
Thanks, maybe we should implement something generic in models.py Finding.save()

damiencarol:
@valentijnscholten you're right, it's better to do it in model. or after the call to `get_findings()`. Will make some modifications this way.

madchap (approved):

PR #4097, opened 2021-03-21
Nikto parser: add support for JSON format

damiencarol:
Work done:
- [x] Support new JSON format (required by project SecureCodeBox https://docs.securecodebox.io/docs/hooks/defectdojo)
- [x] remove management of endpoints in the parser (bad, make query to DB every finding)
- [x] remove some useless attributes (url)
- [x] add more test data and unit tests

### DIFF
--- a/docs/content/integrations/import.md
+++ b/docs/content/integrations/import.md
@@ -332,7 +332,24 @@ Use the full XML export template from Nexpose.
 Nikto
 -----
 
-XML output
+Nikto web server scanner - https://cirt.net/Nikto2
+
+XML output (old and new nxvmlversion=\"1.2\" type) or JSON output
+See: https://github.com/sullo/nikto
+
+To launch a simple run if you have Docker installed:
+
+```shell
+git clone https://github.com/sullo/nikto.git
+cd nikto
+docker build -t sullo/nikto .
+# Call it without arguments to display the full help
+docker run --rm sullo/nikto
+# Basic usage
+docker run --rm sullo/nikto -h http://www.example.com
+# To save the report in a specific format, mount /tmp as a volume:
+docker run --rm -v $(pwd):/tmp sullo/nikto -h http://www.example.com -o /tmp/out.json
+```
 
 Nmap
 ----
--- a/docs/content/integrations/import.md
+++ b/docs/content/integrations/import.md
@@ -334,22 +334,12 @@ Nikto
 
 Nikto web server scanner - https://cirt.net/Nikto2
 
-XML output (old and new nxvmlversion=\"1.2\" type) or JSON output
+The current parser support 3 sources:
+ - XML output (old)
+ - new XML output (with nxvmlversion=\"1.2\" type)
+ - JSON output
+
 See: https://github.com/sullo/nikto
-
-To launch a simple run if you have Docker installed:
-
-```shell
-git clone https://github.com/sullo/nikto.git
-cd nikto
-docker build -t sullo/nikto .
-# Call it without arguments to display the full help
-docker run --rm sullo/nikto
-# Basic usage
-docker run --rm sullo/nikto -h http://www.example.com
-# To save the report in a specific format, mount /tmp as a volume:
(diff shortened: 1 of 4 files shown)


docs/content/integrations/import.md
```diff
@@ -332,7 +332,24 @@ Use the full XML export template from Nexpose.
 Nikto
 -----
 
-XML output
+Nikto web server scanner - https://cirt.net/Nikto2
+
+XML output (old and new nxvmlversion=\"1.2\" type) or JSON output
+See: https://github.com/sullo/nikto
+
+To launch a simple run if you have Docker installed:
```
madchap, line 340: Would we want maybe to re-direct to the official tool's pages, so that we don't have to maintain anything here?

dojo/tools/nikto/parser.py
```diff
@@ -13,6 +13,26 @@
 
 
 class NiktoParser(object):
+    """Nikto web server scanner - https://cirt.net/Nikto2
+
+    XML output (old and new nxvmlversion=\"1.2\" type) or JSON output
+
+    See: https://github.com/sullo/nikto
+
+    To launch a simple run if you have Docker installed:
```
madchap, line 22: Same comment here.

dojo/tools/nikto/parser.py
```diff
@@ -62,86 +139,42 @@ def process_scandetail(self, scan, test, dupes):
...
-                                    ("Description: " + item.find("description").text),
-                                    ("HTTP Method: " + item.attrib["method"]),
-                                    ))
-            mitigation = "N/A"
-            impact = "N/A"
-            references = "N/A"
-
-            dupe_key = hashlib.md5(description.encode("utf-8")).hexdigest()
+            description = "\n".join([
+                    f"**Host:** `{ip}`",
```
madchap, line 143: I was about to say we need to be compatible with 3.6... but these were introduced in 3.6, good thing then! :-)

dojo/tools/nikto/parser.py
```diff
@@ -62,86 +139,42 @@ def process_scandetail(self, scan, test, dupes):
...
-                                  url='N/A',
-                                  dynamic_finding=True)
+                                    test=test,
+                                    active=False,
+                                    verified=False,
+                                    description=description,
+                                    severity=severity,
+                                    numerical_severity=Finding.get_numerical_severity(
+                                        severity),
+                                    references="N/A",
```
madchap, line 174: Why keeping this one N/A, whereas impact and and mitigation were removed?

madchap (approved):

docs/content/integrations/import.md
damiencarol: Honestly, Nikto seems to not be very well (only one maintainer with few commits since 6 months), I'm afraid that this doc will disappear.
I can remove this part if you want.

dojo/tools/nikto/parser.py
damiencarol: impact and mitigation are now optional. not sure about references.

madchap: Optional too.

docs/content/integrations/import.md
madchap: I just don't want for us to set the wrong expectations that we will do that for every parser, because we likely won't :-)

damiencarol: ok. you convinced me. let's simplify that.

dojo/tools/nikto/parser.py
damiencarol: same comment here :D

dojo/tools/nikto/parser.py
damiencarol: ok let's simplify it.

damiencarol: done. it's wonderful! Parsers will be more clean without every "No data here" everywhere...

devGregA (approved):

PR #4110, opened 2021-03-22
allow jira reconciliation to run through celerybeat

valentijnscholten:


madchap (approved):
Locally tested (lightly), seems to not crash and triggers ;-) Cheers.

Maffooch (approved):

PR #4111, opened 2021-03-23
SARIF: support date of findings

damiencarol:
## Work
- [x] support date of findings form the report
ex: https://github.com/DefectDojo/django-DefectDojo/blob/ec47390ef2f8893b2a35c2bf8cbf4dd36584b5fe/dojo/unittests/scans/sarif/appendix_k.sarif#L150
- [x] add more tests (output of a new tool njsscan https://github.com/ajinabraham/njsscan which showed some bugs)

### DIFF
--- a/dojo/tools/sarif/parser.py
+++ b/dojo/tools/sarif/parser.py
@@ -2,8 +2,7 @@ import json
 import logging
 import re
 import textwrap
-from datetime import datetime
-
+import dateutil.parser
 from dojo.models import Finding
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


valentijnscholten:
Is the past around PASSED, lastobservedat and mitigated findings no longer needed? It looks like it was copied from another parser (aws?)

madchap (approved):

damiencarol:
@valentijnscholten yes you're right. bad code. didn't saw that because everything was good in UI and unit tests

valentijnscholten:
> @valentijnscholten yes you're right. bad code. didn't saw that because everything was good in UI and unit tests

so it is correct that it is removed? or should it be put back?

damiencarol:
It's useless code. there is no mitigation in SARIF format. I tested 6 different sources. So we need to remove it.

valentijnscholten (approved):

Issue #3890, opened 2021-02-19
cleanup 2.0: replace django-saml2-auth with django-saml2-pro-auth or djangosaml2 


valentijnscholten:
[django-saml2-auth ](https://github.com/fangli/django-saml2-auth)is no longer maintained making it a security risk and incompatible with some newer libraries.
For example, it requires `PyJWT<2`, where other libraries have switched to `PyJWT>=2.0.0` ([issue](https://github.com/fangli/django-saml2-auth/issues/155))
This prevents us from:
A) doing a clean install from requirements needing the legacy resolver which will be removed later this year (Q1?)
B) locking out dependencies with pipenv
C) upgrading to django 3+ or even 2.2+ (already at risk using django 2.2)
D) using python 3.6 or higher (we are already at risk using 3.6)

The replacement seems to be [django-saml2-pro-auth](https://github.com/zibasec/django-saml2-pro-auth)

Or we could use [djangosaml2](https://pypi.org/project/djangosaml2/) which is mentioned in our [README.md](https://github.com/DefectDojo/django-DefectDojo/blob/master/README.md)

peppelinux:
I suggest djangosaml2 :smile:

quaffapint:
Could I ask whether this will still be part of 2.0.0 and what the rough planned release date is for that? We are unable to get our old ADFS system working with the current SAML plugin and we're hoping djangosaml2 will work for us.

damiencarol:
@quaffapint we pinned this modification for `2.0.0`which will be release at the end of the month.

PR #4117, opened 2021-03-24
feat: replace django-saml2-auth with djangosaml2

alles-klar:
Issue: https://github.com/DefectDojo/django-DefectDojo/issues/3890

Documentation is part of the commit.

----

On behalf of DB Systel GmbH


devGregA:
@alles-klar  haven't forgotten about this, feedback coming shortly.

dojo/settings/settings.dist.py
```diff
@@ -703,6 +672,148 @@ def generate_url(scheme, double_slashes, user, password, host, port, path, param
...
 
+# ------------------------------------------------------------------------------
+# SAML
+# ------------------------------------------------------------------------------
+# For more configuration and customization options, see djangosaml2 documentation
+# https://djangosaml2.readthedocs.io/contents/setup.html#configuration
+# To override not configurable settings, you can use local_settings.py
+SAML2_ENABLED = env('DD_SAML2_ENABLED')
+SAML2_LOGOUT_URL = env('DD_SAML2_LOGOUT_URL')
+if env('DD_SAML2_ENABLED'):
```
Maffooch, line 683: Probably best to `SAML2_ENABLED` to accommodate users not using env vars. Also the env var is already stored there

Maffooch (commented):
This package is a little more complicated than the old one, but way more configurable. Definitely a good move from a maintainability perspective.

I migrated my dev instance from the old to the new, and it wasn't too painful. Lots of reading the docs and debugging to get the gist of things, but that's kinda unavoidable with anything saml related.

Will approve on comment resolution.

dojo/settings/settings.dist.py
```diff
@@ -120,7 +120,7 @@
     DD_SAML2_ATTRIBUTES_MAP=(dict, {
         # Change Email/UserName/FirstName/LastName to corresponding SAML2 userprofile attributes.
         'Email': ('email', ),
-        # 'UserName': ('username', ),
+        'UserName': ('username', ),
```
alles-klar, line 123: @Maffooch Does this work in your setup? In the past I had problems when mapping username here and also setting SAML_DJANGO_USER_MAIN_ATTRIBUTE = 'username'.

Maffooch: My org does not have a username attribute to provide, so I put the email attribute in both the email and username field on the user
```
SAML_ATTRIBUTE_MAPPING =  {
        # Change Email/UserName/FirstName/LastName to corresponding SAML2 userprofile attributes.
        'email': ('email', 'username'),
        'firstName': ('first_name', ),
        'lastName': ('last_name', ),
    }
```

alles-klar: well, users have to adopt there configuration anyway

valentijnscholten (approved):

Maffooch (approved):

PR #4121, opened 2021-03-25
Qualys parser: fixes TypeError: init() got an unexpected keyword argument 'resolve_entities'

alles-klar:
Current parser is broken and throws Server Error 500 / "TypeError: init() got an unexpected keyword argument 'resolve_entities'" when uploading a scan.

This PR fixes parser and adds unit-tests.

----
On behalf of DB Systel GmbH


dojo/tools/qualys/parser.py
```diff
@@ -3,7 +3,7 @@
 import logging
 import html2text
 from . import utfdictcsv
-from defusedxml import ElementTree as etree
+from lxml import etree
```
damiencarol, line 6: we are moving out of `lxml` and replacing it by `defusedxml` to address some security concerns around XML well-known attack. Why do you want to revert this work?

damiencarol (changes requested):

dojo/tools/qualys/parser.py
alles-klar: Because it is not working. Currently the parser fails with "TypeError: __init__() got an unexpected keyword argument 'resolve_entities'". If we have a working alternative I am open for the replacement.

valentijnscholten: that keyword param is not needed with defused: https://pypi.org/project/defusedxml/#external-entity-expansion-remote

damiencarol: made some changes, @alles-klar take a look.

damiencarol:
@alles-klar as you are working on this parser, could you get rid of `utfdictcsv` module and `report_writer` function?

damiencarol (approved):

valentijnscholten (approved):

PR #4124, opened 2021-03-25
Product engagements page performance

danielnaab:
On databases with large test counts per engagement, the `/product/<id>/engagements` page is sluggish. This PR proposes a few changes to improve performance:

- Remove the context menu next to the test count that has direct links to each test
- Tweak prefetches, conditionally prefetch JIRA tables
- Reduce query count by annotating test count


dojo/product/views.py
```diff
@@ -617,27 +617,47 @@ def view_engagements(request, pid, engagement_type="Interactive"):
...
+        count_findings_duplicate=Count('test__finding__id', filter=Q(test__finding__duplicate=True)),
+        count_findings_accepted=Count('test__finding__id', filter=Q(test__finding__risk_accepted=True)),
+    )
+
+    if System_Settings.objects.get().enable_jira:
+        base_engagements = base_engagements.prefetch_related(
+            'jira_project__jira_instance',
+            'product__jira_project_set__jira_instance',
+        )
+
```
valentijnscholten, line 645: the idea is to have prefetching in its own method so that we can reuse it elsewhere if needed.

dojo/templates/dojo/snippets/engagement_list.html
```diff
@@ -202,21 +202,12 @@ <h4> {% if status == "open" %}Active{% elif status == "paused" %}Paused {% else
...
-                        <a class="" href="{% url 'view_engagement' eng.id %}">{{ eng.test_set.all|length }}</a>
-                        <a href="#" class="dropdown-toggle" data-toggle="dropdown">&nbsp;<i class="fa fa-ellipsis-v"></i>&nbsp;</a>
-                        <ul class="dropdown-menu">
-                          {% for test in eng.test_set.all %}
-                               <li>
-                                 <a class="" href="{% url 'view_test' test.id %}">
-                                   <i class="fa fa-list-alt"></i> {{test}}, {{test.updated}}
-                               </li>
-                            {% endfor %}
-                        </ul>
```
valentijnscholten, line 216: not sure removing things to improve performance is the way to go

dojo/product/views.py
danielnaab: I assumed it previously was used in more than one place, but no longer is - but I can certainly put back in a separate function.

dojo/templates/dojo/snippets/engagement_list.html
danielnaab: I hadn't considered making it a "most recent test" menu - What do you think a reasonable limit is, n=10?

valentijnscholten: maybe 25, but depends on if that still performs OK

danielnaab:
@valentijnscholten I had trouble limiting to a fixed-count number of tests. For example, a limit with an `OuterRef` and `Subquery` like this:

```python
    engagements = engagements.prefetch_related(
        Prefetch('test_set', queryset=Test.objects.filter(
            id__in=Subquery(
                Test.objects.filter(
                    engagement_id=OuterRef('engagement_id')
                ).values_list('id', flat=True)[:25]
            ))
        ),
    )
```

... is unfortunately not supported with MySQL:

```python
NotSupportedError at /product/8/engagements
(1235, "This version of MySQL doesn't yet support 'LIMIT & IN/ALL/ANY/SOME subquery'")
```

Instead, limiting to the last X-days works with MySQL. General idea:

```python
    engagements = engagements.select_related(
        'lead'
    ).prefetch_related(
        Prefetch('test_set', queryset=Test.objects.filter(
            id__in=Subquery(
                Test.objects.filter(
                    engagement_id=OuterRef('engagement_id'),
                    updated__gte=timezone.now() - timedelta(days=recent_test_day_count)
                ).values_list('id', flat=True)
            ))
        ),
        'test_set__test_type',
    )
```

This is best option I can come up with balancing query speed with usefulness. Not sure if there are other options, but this seems like a decent compromise. Thoughts?

valentijnscholten (approved):

madchap (approved):
lgtm @danielnaab thanks for the PR :)

PR #4134, opened 2021-03-26
finding.save and dedupe streamlining

### LINKED ISSUE #3944: Jira tickets get created for duplicate findings
It looks like the Deduplication and Jira celery tasks are committed to the queue asynchronously.
The deduplication task marks a dupe finding duplicate, but because the jira task has finding data that preceeds that it still sees  'duplicate' field is always False.
Additionally, when it goes to save the finding, it overwrites the duplicate field as false even if the dedup task marked it as True.
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/jira_link/helper.py#L550

I've worked around this by telling the jira helper code to do a double check on the live status of the finding before proceeding.
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/jira_link/helper.py#L469

`  
finding = Finding.objects.get(id=find.id)

  if finding.duplicate:
    logger.debug('Finding has been marked duplicate. Skipping Jira ticket creation.')
    return
`

### LINKED ISSUE #4135: Bulk risk acceptance doesn't update product grading
Using the bulk update in the UI to risk accept/unaccept findings doesn't trigger an update of the product grade.

### LINKED ISSUE #4144: deduplication is case-sensitive
When a string field like `title`, `component_name` etc is used in the hashcode calculation there can be mismatches if different users/tools/parsers/version of parser are using different formatting with regards to casing. Also the title case we're using can cause mismatches if not correctly applied before calculation the hashcode.

valentijnscholten:
As we have seen in #3944 and #4048 the current `finding.save()` logic triggers multiple processes in parallel, leading to possible race conditions:

- deduplication
- false positive history
- pushing to sonarqube
- push to jira
- applying rules
- product grade calculation

My opinion is that at least we should make sure the logic that possibly changes a findings status should be predictable, i.e. ordered. Same for the hash code calculation which currently is case-sensitive.

This PR wraps all post processing in a single celery task. So we still get a performance benefit by executing these tasks in the background. But inside this celery task there are some sequential steps that possibly change the findings status:

- deduplication
- false positive history

After that some async celery tasks are kicked off for readonly tasks:

- tool updater (sonar)
- product grading
- push to JIRA

Fixes along the way:
- Titlecase was not always applied correctly, leading to deduplication mismatches
- Rules are now applied before deduplication. Some rules might be created to correct data on findings with the goal to improve the dedupe process, so we want to apply them before deduplication.

The product grading happens for each finding, which is really inefficient. If you import 1000 finding, the grade of the product gets recalculated a 1000 times. There currently is no reliable way to fix this as there still are background processes that can change a findings status, i.e. deduplication and false positive history.
Also the ordering of deduplication is still unpredictable as depending on the timing finding 100 can be marked as a duplicate of 101 or the other way around.

As I have said before, I think these should be run in the foreground to avoid all these complicated situations, or find a better solution. I will discuss on Slack for a next PR.

fixes #3944 
fixes #4135
fixes #4144

### DIFF
--- a/dojo/api.py
+++ b/dojo/api.py
@@ -1584,8 +1584,6 @@ class ReImportScanResource(MultipartResource, Resource):
                 if Finding.SEVERITIES[sev] > Finding.SEVERITIES[min_sev]:
                     continue
 
-                from titlecase import titlecase
-                item.title = titlecase(item.title)
                 if scan_type == 'Veracode Scan':
                     find = Finding.objects.filter(title=item.title,
                                                   test__id=test.id,
--- a/dojo/api.py
+++ b/dojo/api.py
@@ -1584,6 +1584,8 @@ class ReImportScanResource(MultipartResource, Resource):
                 if Finding.SEVERITIES[sev] > Finding.SEVERITIES[min_sev]:
                     continue
 
+                from titlecase import titlecase
+                item.title = titlecase(item.title)
                 if scan_type == 'Veracode Scan':
                     find = Finding.objects.filter(title=item.title,
                                                   test__id=test.id,
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1366,9 +1366,6 @@ class ReImportScanSerializer(TaggitSerializer, serializers.Serializer):
                 component_name = item.component_name if hasattr(item, 'component_name') else None
                 component_version = item.component_version if hasattr(item, 'component_version') else None
 
-                from titlecase import titlecase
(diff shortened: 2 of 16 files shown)


damiencarol:
@valentijnscholten honestly we should remove titlecase completly and let parsers manage title case.

valentijnscholten:
That's not for this PR.

damiencarol:
ok

damiencarol (approved):

madchap (approved):

PR #4142, opened 2021-03-27
don't delete all duplicates when deleting an original, fix product grade on deletion.

### LINKED ISSUE #3057: Don't delete duplicates when deleting an original
**Bug description**
In a single engagement, I had 3 tests, with 1, 3 and 12 findings respectively. The 1 and 3 findings were contained in the 12 findings (just sub set). When I deleted the first two tests, I saw my third test reduced from 12 to 9 findings.

**Steps to reproduce**
As the situation happened just before a correction of an unrelated issue, I did not try to reproduce the problem. I imagine at this point that we can see exactly the same behavior if we import a test two times, with the first one being only a sub set and then the second time the full report. If we delete the first (sub set), the second one will be impacted.

**Expected behavior**
If there is a duplicate, when I delete the test containing the original, I expected the duplicate finding in another test to be not deleted, or at least to have an option to unlink such duplicate to not loose information.

- [ ] Kubernetes
- [ ] Docker

**Environment information**

**Screenshots** (optional)

**Console logs** (optional)

### LINKED ISSUE #4140: Deleting findings/tests/engagements doesn't update product grade
Products grades don't get updated when findings/tests/engagements are deleted

valentijnscholten:
fixes #3057 

Current behavior:
Assume you have a finding A with two duplicates, so finding B is a duplicate of A and C is a duplicate of A:

B->A and C->A

When deleting finding A, also B and C get deleted. This is confusing as described in #3057 

New behavior:

When A gets deleted, the duplicates B and C remain and they get a new original finding assigned. The oldest of B and C becomes the new original, so C->B and B is the new original and no longer marked as duplicate.

There is a flag `DD_DUPLICATE_CLUSTER_CASCADE_DELETE` / `DUPLICATE_CLUSTER_CASCADE_DELETE` that can be set to `True` to revert to the old behavior. The default is `False`.

To achieve this, we had to change the `on_delete` parameter of the `duplicate_finding` field. This can be done without a new migration as it doesn't affect the SQL schema.

There is some discussion on Slack where some are arguing for more complicated scenario's where the data of the deleted original should be migrated to the next/new original. Sounds way to complicated for me. If the original contains valuable information, don't delete it! https://owasp.slack.com/archives/C2P5BA8MN/p1616780956093700

bug fixes along the way:
- fixes #4140

### DIFF
--- a/dojo/db_migrations/0058_document_finding_model.py
+++ b/dojo/db_migrations/0058_document_finding_model.py
@@ -72,7 +72,7 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='finding',
             name='duplicate_finding',
-            field=models.ForeignKey(blank=True, editable=False, help_text='Link to the original finding if this finding is a duplicate.', null=True, on_delete=django.db.models.deletion.CASCADE, related_name='original_finding', to='dojo.Finding', verbose_name='Duplicate Finding'),
+            field=models.ForeignKey(blank=True, editable=False, help_text='Link to the original finding if this finding is a duplicate.', null=True, on_delete=django.db.models.deletion.DO_NOTHING, related_name='original_finding', to='dojo.Finding', verbose_name='Duplicate Finding'),
         ),
         migrations.AlterField(
             model_name='finding',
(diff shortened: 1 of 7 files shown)


madchap (approved):

ptrovatelli (approved):

ptrovatelli:
for me the old behavior was fine and I would even prefer having DUPLICATE_CLUSTER_CASCADE_DELETE = True by default but I can understand how some people may be confused by it

PR #4145, opened 2021-03-27
Optimize (pdf) report in main menu and small fix for product type list

StefanFl:
This is an update to #4065. The Reports entry in the main menu still led to the report list. Now there is no submenu anymore but the report builder is immediately shown when clicking on Reports. Plus a small fix for the product type list for non-staff users and a small refactoring.

### DIFF
--- a/dojo/templates/base.html
+++ b/dojo/templates/base.html
@@ -276,18 +276,23 @@
                                 </li>
                             {% endif %}
                             <li>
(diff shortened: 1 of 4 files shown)


valentijnscholten (approved):

damiencarol (approved):

damiencarol:
@StefanFl is it finished?

PR #4148, opened 2021-03-27
allow non-staff users to perform bulk edit

### LINKED ISSUE #4136: non-staff users should be able to perform bulk edits
currently only staff users can perform bulk edits, but there are usecases where normal 'authorized' users should be able to perform bulk edits (but only on findings for which they are authorized).

valentijnscholten:
fixes #4136 

Most code was already in place, but there was a decorator on the bulk edit method that allowed only staff users to perform bulk edits. Removed decorator, updated permission checks and tested all scenarios around True/False for:

```
AUTHORIZED_USERS_ALLOW_CHANGE
AUTHORIZED_USERS_ALLOW_DELETE
AUTHORIZED_USERS_ALLOW_STAFF
```

Moved some code around to make `view_test` align more with `findings_list_snippet.html`

### DIFF
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -1797,15 +1797,16 @@ def merge_finding_product(request, pid):
 
 
 # bulk update and delete are combined, so we can't have the nice user_must_be_authorized decorator (yet)
-@user_passes_test(lambda u: u.is_staff)
-# @user_must_be_authorized(Product, 'staff', 'pid')
 def finding_bulk_update_all(request, pid=None):
     form = FindingBulkUpdateForm(request.POST)
     now = timezone.now()
     if request.method == "POST":
         finding_to_update = request.POST.getlist('finding_to_update')
+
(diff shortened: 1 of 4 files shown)


valentijnscholten:
@StefanFl not sure if this means anything for Authorizations V2.

StefanFl:
@valentijnscholten That helps a lot actually. I was already thinking how to deal with the bulk operations. With your changes it will be easy to implement the functionality for authorization V2.

valentijnscholten:
I implemented some logic to show the user how many findings were skipped due to not being authorized. But I think with the old authorization, the user only sees findings for which he/she is authorized so skipped count will always be zero. Unless someone messes with the client-side UI, in which case they don't deserve nice error messages ;-)
But maybe it's helpful still with AuthZ v2 as there the permissions can differ per product. For some products the user has view permissions for others writer, etc.

madchap (approved):
lgtm

madchap (approved):

damiencarol (approved):

PR #4149, opened 2021-03-27
alert count: don't get them on every dropdown click

### LINKED ISSUE #4147: a click on a dropdown always gets alerts from server
In `base.html` there's jQuery that attaches the `get_alerts()` function to every element with class `dropdown-toggle`.
The intention is to get the new alerts whenever somebody clicks on the alerts dropdown. But as a side effect it is also attached to every other dropdown on screen, i.e. opening the filters, opening similar findings, selecting findings for bulk edit, etc.

valentijnscholten:
fixes #4147 

In base.html there's jQuery that attaches the get_alerts() function to every element with class dropdown-toggle.
The intention is to get the new alerts whenever somebody clicks on the alerts dropdown. But as a side effect it is also attached to every other dropdown on screen, i.e. opening the filters, opening similar findings, selecting findings for bulk edit, etc.

### DIFF
--- a/dojo/templates/base.html
+++ b/dojo/templates/base.html
@@ -128,7 +128,7 @@
                         <!-- /input-group -->
                     </li>
                     <li class="dropdown">
-                        <a class="dropdown-toggle dropdown-toggle-h{{ alert_count }}" data-toggle="dropdown" href="#" aria-expanded="false" aria-label="{{ alert_count }} Alerts">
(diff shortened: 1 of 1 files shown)


damiencarol (approved):

valentijnscholten:
Later I thought maybe it was on purpose? To get updated alerts every time the user interacts with the UI? In that case we should change it into `update_alert_count()` because `get_alerts` doesn't update anything visible in the UI. wdyt @devGregA @mtesauro @Maffooch

madchap:
Still would seem overkill to me.

madchap (approved):

Issue #4161, opened 2021-03-29
ORT parser error when importing oss-review-toolkit file


devsecopsale:
**Slack us first!**
I get an error when I try to upload an ORT report file that contains violation rules items with no 'license' or 'license_source' fields. I actually fixed this by checking the values do exist before handling the python keys. I'm submitting a pull request

**Be informative**
This bug prevents from uploading ORT scans files. This will typically occurs when a custom license is defined and it's unresolved but declared.

**Bug description**
This bug prevents from uploading ORT scans files.

**Steps to reproduce**
1. Go to a Test type 'import scan results' from ORT type
2. Click on 'choose file'
3. Select the file that contains violation rules with no license or license_source fields.
4. See error and see debugging errors in the logs. Those ones are caused by the ORT parser.py file

**Expected behavior**
an error will be seen.

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [x ] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: [Redhat]
 - DefectDojo Commit Message: [use `git show -s --format="[%ci] %h: %s [%d]"`]

**Sample scan files** (optional)
If applicable, add sample scan files to help reproduce your problem.

**Screenshots** (optional)

**Console logs** (optional)
If applicable, add console logs to help explain your problem.

**Additional context** (optional)

damiencarol:
fixed in `dev` / `1.15.0`

PR #4162, opened 2021-03-29
Fix ORT parser error when importing oss-review-toolkit file

devsecopsale:
This is to fix issue #4161   - Some ORT output files will not contain license or license_source in its violation rules. If this is not being verified, the file import will fail.


valentijnscholten:
Thanks, could you base it against `dev` please? cc @damiencarol

devsecopsale (commented):
I removed the other files. Only 'dojo/tools/ort/parser.py' needs to be modified for this patch to work. Thanks

damiencarol (changes requested):
some important files are removed in this PR like README or unit tests pipeline.
it's a no-go for me

PR #4171, opened 2021-03-30
Publish the Helm chart tarball with each release (try 2)

valentijnscholten:
Reverts DefectDojo/django-DefectDojo#4170

so basically reimplement #3849 and #4143


madchap:
ah, how you will now take credit for some k8s work!

PR #4175, opened 2021-03-30
Product list page: prefetch prod_type

danielnaab:
For performance: on the product list page, prefetch `prod_type` rather than use `select_related`. On our database this leads to a significant speed-up.


valentijnscholten:
@danielnaab Do you agree this is no longer necessary since #4177?

PR #4183, opened 2021-03-31
Contrast parser: switch endpoint to unsaved_endpoints

damiencarol:
## Work
- [x] switch endpoint to unsaved_endpoints
- [x] enhance de-duplication
- [x] add more unit tests


dojo/tools/contrast/parser.py
```diff
@@ -36,45 +37,47 @@ def get_findings(self, filename, test):
...
+                description=description,
+                severity=severity,
+                numerical_severity=Finding.get_numerical_severity(severity),
+                dynamic_finding=True,
+                static_finding=False,
+                vuln_id_from_tool=row.get('Rule Name'),
+                unique_id_from_tool=row.get('Vulnerability ID'),
+                nb_occurences=1,
+            )
+            endpoint = Endpoint(host="0.0.0.0", protocol="http")
```
valentijnscholten, line 55: is this something we want to keep, http://0.0.0.0 as default?

aaronweaver: Probably not.

damiencarol: it was the value in the current code. so, ... what do we do??? :D

valentijnscholten: bring it inline with what we usually do, but there is this other big PR around endpoints coming anyway.

damiencarol: @aaronweaver @valentijnscholten I don't have a better option than keep the same behavior. Also this PR  fix de-duplication not only the endpoints. I already discussed with the author of the other big PR, it's not a pb to resolve conflits on his side.

valentijnscholten: maybe just don't add an endpoint if 'Request URI' is not there?

damiencarol: made this modification and updated unit tests

valentijnscholten: Thanks, just looked again and I thought the URI contained the hosts but there is no host so they use 0.0.0.0.
This is an interesting case as ideally the uploader should supply an endpoint in the upload and that should be used to get the hostname? Not for this PR maybe, but something to think about when improving/changing the parser contract.

mtesauro (approved):
Approved

valentijnscholten (approved):

PR #4188, opened 2021-04-01
Add full support for using ports and userinfo in endpoints

kiblik:
Hello.

based on [my analysis](https://owasp.slack.com/archives/C2P5BA8MN/p1616088028019100), I prepared redesign of using endpoints - it fully allows the use of ports and userinfos.

Changes:
- Every time, when an endpoint is copied, we copy also `userinfo`, `fqdn` and `port`
- Fix and use `Endpoint.from_uri()` - I remove many different implementations of parsing URLs; now, we are using only [hyperlink](https://pypi.org/project/hyperlink/), based on @damiencarol's [recommendation](https://owasp.slack.com/archives/C2P5BA8MN/p1616182871040200?thread_ts=1616088028.019200&cid=C2P5BA8MN)
- API
  - for creation (POST), require only `host` and `product`; others are optional
  - for update (PATCH), do not ask for `product` (if you don't want to update it)
- Add `userinfo` as `Endpoint` field - following [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.2.1)
- Validate `Endpoint` fields in the creation process
- `Endpoint.__str__()` create full URL (if is it possible)
- Rewrite (and fix) all parsers which use endpoints

TODO:
- [x] Write tests
   - [x] migration
   - [x] model
   - [x] api
- [x] Split views (and reports) to "hosts" (e.g. all vulnerabilities on specific host [ fqdn / ip ] = from all endpoints on host) and "endpoints" (e.g. all vulnerabilities on specific endpoint [ url ])
- [x] Double-check all forms and views
   - [x] ascii
   - [x] html
   - [x] pdf
- [x] we should forbid the update of product from API - Finding in product1 and Endpoint in product2? No way
- [x] redesign all `Endpoint.objects.get_or_create` and `Endpoint.objects.filter`
- [x] Form validations


kiblik:
🤔

- [ ] we should forbid the update of product from API - Finding in product1 and Endpoint in product2? No way

valentijnscholten:
amazing PR, thanks!

dojo/models.py
```diff
@@ -1085,10 +1078,12 @@ class Meta:
...
 class Endpoint(models.Model):
     protocol = models.CharField(null=True, blank=True, max_length=10,
-                                help_text="The communication protocol such as 'http', 'ftp', etc.")
+                                 help_text="The communication protocol/scheme such as 'http', 'ftp', 'dns', etc.")
+    userinfo = models.CharField(null=True, blank=True, max_length=500,
+                              help_text="User info as 'alice', 'bob', etc.")
     host = models.CharField(null=True, blank=True, max_length=500,
-                            help_text="The host name or IP address, you can also include the port number. For example"
-                                      "'127.0.0.1', '127.0.0.1:8080', 'localhost', 'yourdomain.com'.")
+                            help_text="The host name or IP address. It can not include the port number. For example"
```
damiencarol, line 1085:
```suggestion
                            help_text="The host name or IP address. It must not include the port number. For example"
```

kiblik: https://github.com/DefectDojo/django-DefectDojo/pull/4188/commits/7735463a2ee823a0aa7ba213b848e75f21002eab

damiencarol:
Awesome work. I'm following the final version of it.
Some feedback:
 - sometimes you replace `endpoint.host` to just `endpoint` in some views, I don't know if it is good. it will change the view and lost the users as `endpoint` is the full URI
 - you added changes in the parsers but I have few PR that are in conflict with these changes
 - you should avoid setting the product when creating the Endpoint. the point to use `unsaved_endpoints` is to keep the implementation of parser simple and not rely on nested attributes of Test/Engagement/Product/... (like `product=finding.test.engagement.product`)
 - you modified some parsers to add `finding.url` it's not a good move as we are pushing to remove `url` attribute which is useless

kiblik:
> Awesome work! I'm waiting with interest the final version of it.
> 
> Some feedback:
> 
>     * sometimes you replace `endpoint.host` to just `endpoint` in some views, I don't know if it is good. it will change the view and lost the users as `endpoint` is the full URI
As I wrote in [the first comment](https://github.com/DefectDojo/django-DefectDojo/pull/4188#issue-607395237), I need to check all views and templates. I know, I missed sth. In my option, we should split all views to "host view" (merged all findings with the same host - we don't care there are more services, on multiple ports, we need to see all vulns on the host) and "endpoint view" (findings detected only on the specific endpoint). 
> 
>     * you added changes in the parsers but I have few PR that are in conflict with these changes
Do not care. I will keep my 👀 on `dev` branch and I will fix all conflicts.
> 
>     * you should avoid setting the product when creating the Endpoint. the point to use `unsaved_endpoints` is to keep the implementation of parser simple and not rely on nested attributes of Test/Engagement/Product/... (like `product=finding.test.engagement.product`)
Definitely, I used it only in case, it was in origin parser like [this](https://github.com/DefectDojo/django-DefectDojo/pull/4188/files#diff-65ac6bd9a70f11f65ccf475570968d37b017a0c332ac7499fc197c9230847160L91), [this](https://github.com/DefectDojo/django-DefectDojo/pull/4188/files#diff-39912ff177e1d5cd784cb9d64ce628ee13b4d12cb15a15db24cec683467d0d44L107), [this](https://github.com/DefectDojo/django-DefectDojo/pull/4188/files#diff-25683e9e454613de07dfd1e4261af9f5cf8b20ab133b1a25a8070a3f7425e9b9L123), [this](https://github.com/DefectDojo/django-DefectDojo/pull/4188/files#diff-25683e9e454613de07dfd1e4261af9f5cf8b20ab133b1a25a8070a3f7425e9b9L157), [this](https://github.com/DefectDojo/django-DefectDojo/pull/4188/files#diff-d73d11ae11c335f0d99b97145e39ba647b7d856628b0e483cf799ed2e60c9bddL136) or [this](https://github.com/DefectDojo/django-DefectDojo/pull/4188/files#diff-19be89b02a721952809fc80e2c4034af174bf6f6ab157af16abec8cdb353de9cL99). I did it [once](https://github.com/DefectDojo/django-DefectDojo/pull/4188/files#diff-0d48f3120b242765a5755e58dd2164521b74e3cc1546d5c5e28a59971afb2e65R33) by mistake. It is a good idea, I will remove it: https://github.com/DefectDojo/django-DefectDojo/pull/4188/commits/ccb353ed33db13eebfb73b063b05a26136415f26
> 
>     * you modified some parsers to add `finding.url` it's not a good move as we are pushing to remove `url` attribute which is useless
Same as `product=finding.test.engagement.product`. It was there also before. Fix: https://github.com/DefectDojo/django-DefectDojo/pull/4188/commits/5405eb46c6347cd9915ef6b36310fc835a409b0e

damiencarol:
1. totally agree with you for the split host/endpoint views.
2. so, let's start the merge/rebase madness :) , you have some unit tests that fail and are fixed in `dev`, could you rebase your PR on last `dev` branch?
3. took a look at the unit tests, some fail are normal as the PR remove port in host attribute and some checks rely on `host = 'blabla:80'` thing

kiblik:
Btw, we will have to redesign also all `Endpoint.objects.get_or_create` and `Endpoint.objects.filter` because:
```python
$ docker-compose exec uwsgi python manage.py shell
[02/Apr/2021 14:35:04] INFO [dojo.models:3375] enabling audit logging
Python 3.6.12 (default, Sep 10 2020, 18:03:29)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from dojo.models import Endpoint
>>> Endpoint(host='FOO.BAR') == Endpoint(host='foo.bar')
True
```

valentijnscholten:
I just thought about existing Endpoint 9bjects. It's there anything we can or should do as part of this PR to update existing endpoints to match al the new / improved parsing and logic in this PR?

kiblik:
I prepared 2 db migrations:
- https://github.com/DefectDojo/django-DefectDojo/pull/4188/files#diff-0b6d092a24eaa20e1a11eca021b9304957bf331c7ed8003f5251c8f9580f0966R1
- https://github.com/DefectDojo/django-DefectDojo/pull/4188/files#diff-5168555147d14f3b7f573a9175293d4f8c1d8ed60b35a2c433dbaf1adf1ee05dR1

And I wrote also unittest for this migration https://github.com/DefectDojo/django-DefectDojo/pull/4188/files#diff-374ed8d6f1b1c884b4143f49361fc7f89bfa1816f16ee5d713445af44013f419R1

dojo/tools/nexpose/parser.py
```diff
@@ -229,7 +229,11 @@ def get_items(self, tree, vulns, test):
 
                     find = self.findings(dupe_key, dupes, test, vuln)
 
-                    endpoint = Endpoint(host=host['name'], port=service['port'], protocol=service['name'].lower())
+                    endpoint = Endpoint(
+                        protocol=service['name'].lower(),
```
damiencarol, line 233: Could you switch this to `Endpoint.from_uri()` ?

dojo/tools/openvas_csv/parser.py
```diff
@@ -77,92 +74,11 @@ def __init__(self):
...
-            except:
-                dupe_endpoint = None
-
-            if not dupe_endpoint:
-                endpoints = [Endpoint(host=url, product=finding.test.engagement.product)]
-            else:
-                endpoints = [dupe_endpoint]
-
-            finding.unsaved_endpoints = endpoints
+        endpoint = Endpoint(host=url)
```
damiencarol, line 79: Could you switch this to `Endpoint.from_uri()` ?

kiblik: Openvas store IP in this field (there is not real URL). Check [mapping](https://github.com/DefectDojo/django-DefectDojo/pull/4188/files/ad2e920f24609ef3686e51aa61adc3769f29667d#diff-d73d11ae11c335f0d99b97145e39ba647b7d856628b0e483cf799ed2e60c9bddR74) and [example scan](https://github.com/DefectDojo/django-DefectDojo/blob/ad2e920f24609ef3686e51aa61adc3769f29667d/dojo/unittests/scans/openvas/report-e2759495-f26d-4089-9c56-12a10dc36c9c.csv#L2)
We can change name of variable from `url` to `ip` or `host`.

dojo/tools/nexpose/parser.py
kiblik: Nexpose store IP/hostname, port, service in ["raw" format](https://github.com/DefectDojo/django-DefectDojo/blob/580d014a26cb6d32346646f52016668a2e6aff46/dojo/unittests/scans/nexpose/many_vulns.xml#L41-L43), not in URL.
I suppose it is better to use these value than create URL and parse it again

kiblik:
News:

- Product menu shows "number of host" / "number of endpoints" image
- Endpoints menu (in Product menu) Site menu and allow to show list by Endpoints or by Hosts image image
- List by Endpoints and by Hosts image image

dojo/tools/nexpose/parser.py
damiencarol: ok, let's keep it like that for now.

kiblik:
Changelog:

- "Vulnerable Services" and "Endpoint Service" have been changed to "Endpoints" and they are only in "Host view"
- Each Endpoint has Tick (non-vulnerable) or Cross (vulnerable)
- Tags are only in "Endpoint view"
- Actions:
    -  Edit Endpoint, Add Metadata, View History, Delete Endpoint are only in "Endpoint view"
    -  Report redirects to "Endpoint report" or "Host report (needs to be finished)"
  
image
image
image
image

kiblik:
Question for @damiencarol and/or @valentijnscholten:
Are 2 Endpoints, with same URI (like `str(ep1)==str(ep2)`) in different products same? They are not in my opinion. So I created [this commit](https://github.com/DefectDojo/django-DefectDojo/pull/4188/commits/0e4cd089bf0f25f83f19bbc1233af628dc9d2067).
But it automatically breake [tests like this](https://github.com/DefectDojo/django-DefectDojo/blob/ea0fbb63b9254115aabb2147dda5ca5e33e3c5e8/dojo/unittests/test_import_reimport.py#L223) because [`update_endpoint_status`](https://github.com/DefectDojo/django-DefectDojo/blob/ea0fbb63b9254115aabb2147dda5ca5e33e3c5e8/dojo/importers/reimporter/utils.py#L68) identifies them as different endpoints.
How we should handle this case?

valentijnscholten:
@kiblik Can you let us know when this is ready? I've seen quite some updates and only want to start the review process when you feel it is ready. Also I propose to merge this after the release on Tuesday 27th, as it's a very big PR.

damiencarol:
I agree better to wait.

kiblik:
Hi @valentijnscholten and @damiencarol.
I finished my part. You can review.
It is up to you when you will do it.

valentijnscholten:
Again thanks for the PR, looks like a good improvement and lots of work. Since there are so many changes in so many places, there is always a slight chance some breaks or maybe changes that affects people. So I have added the `breaking changes` label so people can clearly see in the release notes.

damiencarol:
reviewing it and testing it on my local dev env

valentijnscholten:
We will skip 1.16.0 and the next release will be 2.0.0 end of June. If you fix the PR, we can merge soon now as we'll have 2 months to test it :-)

damiencarol:
I agree, let merge this one and test it for 2 month before the `2.0.0`

damiencarol:
@kiblik fixed some conflicts and flake8 error, please take a look at failing tests (I'm doing the same).
I want to merge this one soon to start the cleaning on parsers.

kiblik:
@damiencarol, I just realised that one important part is missing in this PR.
Right now,`Endpoint(protocol='http',host='foo.bar') == Endpoint(protocol='http',host='FOO.BAR',port=80)`. It works when new unsaved Endpoint from test should be identified (and associated) as existing Endpoint.

In db_migration, we are checking only for the [right format](https://github.com/DefectDojo/django-DefectDojo/pull/4188/files#diff-7035ccbc8eb6c33d16bc71bffa7fd9fa31bebfa2ec4cacf512b916d23e97909a) of the existing Endpoint, however, we are not checking for duplicity.
- We should "merge" these Endpoints to the same one. 
- What if they have different Endpoint_Status?

damiencarol:
I think we should should "merge" these Endpoints to the same one.
If the status need to be changed, could we use the same logic that we have in the importer?

dinvlad:
Just to confirm, this also addresses `path` in endpoints? I.e. that endpoints with separate `path` will also be shown as separate endpoints in the UI?

kiblik:
@dinvlad, View in UI was splitted in to 2 sections: by host (all endpoints with the same host aggregated in to one) by endpoint (different path = different view, different query = different view)
Check https://github.com/DefectDojo/django-DefectDojo/pull/4188#issuecomment-816314610

damiencarol:
@kiblik we are ready to merge this one if you can complete the "merge" in the DB migration. I know it is new work but after discussing with other contributor it's easier to integrate it in the pull request instead of doing it later.
I can help/provide support for tests and/or managing conflicts. just ping me.

kiblik:
@damiencarol, I fixed merging endpoints: https://github.com/DefectDojo/django-DefectDojo/pull/4188/commits/e6e6d66915b4011febe844eb4555505cde8a3bab
I suppose we are ready.

dojo/db_migrations/0093_endpoint_host_migration.py
```diff
@@ -99,6 +103,44 @@ def clean_hosts(apps, schema_editor):
...
+                    epss = Endpoint_Status_model.objects\
+                        .filter(endpoint=ep_ids[0])\
+                        .values('finding')\
+                        .annotate(total=Count('id'))\
+                        .filter(total__gt=1)
+                    for eps in epss:
+                        esm = Endpoint_Status_model.objects\
+                            .filter(finding=eps['finding'])\
+                            .order_by('-date')
+                        esm.exclude(id=esm[0].pk).delete()
```
valentijnscholten: I think your approach is to keep the eps that was created latest? Might it be slightly better to keep the eps that was _updated_ the latest, i.e. order_by('-ast_modified') ?

valentijnscholten (commented):
asdfasdf

valentijnscholten:
should we add some unique constraints on the EndPoint model to avoid future duplicate entries / MultiplObjectsReturned problems?

dojo/db_migrations/0093_endpoint_host_migration.py
kiblik: You are right, I pushed update

kiblik:
Final rebase? I regularly merged content of the `dev` branch. Isn't it enough?

MultiplObjectsReturned: In my opinion, it shouldn't happen right now. But change is really complex and I hope I didn't miss anything.

valentijnscholten:
I meant renaming / updating migration

dojo/db_migrations/0093_endpoint_host_migration.py
```diff
@@ -99,6 +103,44 @@ def clean_hosts(apps, schema_editor):
         raise FieldError('It is not possible to migrate database because there is/are {} broken endpoint(s). '
                          'Please check logs.'.format(len(broken_endpoints)))
 
+    to_be_deleted = set()
+    for product in Product_model.objects.all().distinct():
+        for endpoint in Endpoint_model.objects.all():
```
valentijnscholten: Is it need to iterate over all products times all endpoints, or would it be enough to just to all endpoints and use endpoint.product for the filter?

valentijnscholten:
I also notice there can be situations where the data migration will result in an error due to invalid endpoints being found. The error message states the user should remove or update these endpoints. However at that point the user will no longer have a working Defect Dojo instance because the migration failed and successive migration may not have been applied yet.

Would it be possible to have a management command that checks all endpoints and outputs the invalid/broken ones so that the user can fix those?

kiblik:
Do you have some example for this "management command"?
It is a good idea but I don't know how it should look like.

valentijnscholten:
example is https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/management/commands/dedupe.py and other files in that folder. They can be executed with `./manage.py dedupe`, so we could add a similar command and instruction to the upgrade notes.

dojo/db_migrations/0093_endpoint_host_migration.py
valentijnscholten: fixed by commit

damiencarol:
@valentijnscholten with this new need of the management command the PR is growing again. I hope we can close and merge after this one.

valentijnscholten:
I was about to merge it, until I saw that the migration could break instances. So we have no choice but to make some more changes.

valentijnscholten:
Just fixed a bug in the simple search around Endpoint.product being null/None sometimes. Just a heads up here as maybe this PR here should also take that scenario into account?

kiblik:
I added a command which tests if migration will be successful https://github.com/DefectDojo/django-DefectDojo/pull/4188/files#diff-d57a70293407c160e2a8361c2247372d70bc0b5c1178442490d5ed1fcb1d6ae1
It is possible to run the check for the 2nd part of migration (remove Endpoints and merge EndpointStatus) but there can be a different result because the first part change Endpoints - it is the reason why this command doesn't check the 2nd part.
I'm not sure if the content of the command is what you expected. If not, feel free to change.

valentijnscholten:
Thanks, it's good except one error:

`django.db.utils.OperationalError: (1054, "Unknown column 'dojo_endpoint.userinfo' in 'field list'")`

People running 1.15.0 won't have this new column yet, so maybe to do some `hasattr` check in there to work around that?

valentijnscholten:
hmm, might not be so easy to fix. traceback: 
```
  File "./manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/home/valentijn/dd/dojo/management/commands/endpoint_pre-migration_check.py", line 17, in handle
    clean_hosts_run(apps=apps, change=False)
  File "/home/valentijn/dd/dojo/endpoint/utils.py", line 92, in clean_hosts_run
    for endpoint in Endpoint_model.objects.all():
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/models/query.py", line 274, in __iter__
    self._fetch_all()
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/models/query.py", line 1242, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/models/query.py", line 55, in __iter__
    results = compiler.execute_sql(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size)
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/models/sql/compiler.py", line 1142, in execute_sql
    cursor.execute(sql, params)
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 99, in execute
    return super().execute(sql, params)
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 67, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 76, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/utils.py", line 89, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
  File "/home/valentijn/venv/lib/python3.8/site-packages/django/db/backends/mysql/base.py", line 71, in execute
    return self.cursor.execute(query, args)
  File "/home/valentijn/venv/lib/python3.8/site-packages/MySQLdb/cursors.py", line 206, in execute
    res = self._query(query)
  File "/home/valentijn/venv/lib/python3.8/site-packages/MySQLdb/cursors.py", line 319, in _query
    db.query(q)
  File "/home/valentijn/venv/lib/python3.8/site-packages/MySQLdb/connections.py", line 259, in query
    _mysql.connection.query(self, query)
django.db.utils.OperationalError: (1054, "Unknown column 'dojo_endpoint.userinfo' in 'field list'")
```

kiblik:
@valentijnscholten, @damiencarol, I suppose we should drop field [`fqdn`](https://github.com/DefectDojo/django-DefectDojo/blob/cd404af90ff4d9407ce9c16edd3ecf5786660d25/dojo/models.py#L1069). It is used only in [`nmap`](https://github.com/DefectDojo/django-DefectDojo/blob/cd404af90ff4d9407ce9c16edd3ecf5786660d25/dojo/tools/nmap/parser.py#L56) and [`ssl_labs`](https://github.com/DefectDojo/django-DefectDojo/blob/cd404af90ff4d9407ce9c16edd3ecf5786660d25/dojo/tools/ssl_labs/parser.py#L199) parsers (only because they coexist with IP) and in [testdata](https://github.com/DefectDojo/django-DefectDojo/blob/cd404af90ff4d9407ce9c16edd3ecf5786660d25/dojo/fixtures/dojo_testdata.json#L582).
FQDN is not displayed anywhere in DD (until you do not click to "edit endpoint"). I do not see any reason to store this information.

damiencarol:
@kiblik  I'm ok with that. honestly even the pull request was limited to only manage "port" I would be happy :D

damiencarol (approved):
Awesome work @kiblik and thanks a lot for your patience and your dedication to take our feedback.

madchap (approved):
Great work there, lgtm :) Kudos @kiblik

valentijnscholten:
@kiblik We have removed the merge from `dev` as it's not fully ready yet as discussed per DM. Could you recreate?

I think the way forward is to not let the migration crash on corrupt endpoints, but warn the user. And offer a management command to resolve any corrupt endpoints.

madchap:
> Sorry about the confusion, I should have put a "changes requested" on it.

Apologies to all, I certainly jumped the gun a bit too quickly as well.

Thanks @kiblik to re-create the PR.

kiblik:
Recreated in #4473

PR #4189, opened 2021-04-01
fix ORT parser error when importing oss-review-toolkit file

devsecopsale:
this will fix bug #4161. It's verifying that license and license_source fields are defined. In case they are not, it set them as 'unset'.


devsecopsale:
For what I can see, the unit test seems to be failing due to an unrelated issue (some jira feature)

valentijnscholten (approved):

valentijnscholten:
failing unit tests is indeed different issue: #4191

Maffooch (approved):
Not sure what to do about unit test here, but PR looks good

valentijnscholten:
always better to rebase / merge dev and make sure the unit tests are passing.

PR #4196, opened 2021-04-02
Authorization V2: Findings and Endpoints

StefanFl:
Another large PR I am afraid. Good news: With the new authorization for Findings and Endpoints in this PR, the implementation is nearly feature-complete. After this I will concentrate on performance and some smaller additions.

Please have a look at https://github.com/DefectDojo/django-DefectDojo/compare/dev...StefanFl:authorization-5?expand=1#diff-a68b570dd86735b668cc0d1f375b85e37218f9a80a666a433d62721ee2a729c0. Product_Members and Product_Type_Members need to be read often to have only the options in the UI, that the users is allowed to see or do. Using a lru_cache seems to be a good way to mitigate the calls to the database. What do you think about it?


valentijnscholten:
Any caching like that will not work if you have multiple uwsgi processes or even multiple django containers/pods. We could settle for a cache with a relatively short timeout, but I am not a fan of that. Would it help if it was cached only during a single request? Or can we clear the cache when somebody logs on? The latter probably not with the default long sessions we have.

StefanFl:
The cache actually helps a lot for single requests. E.g. when you have a product with a lot of findings. The current implementation checks the authorization for each finding multiple times, to build the context menus. The product member role will be in the cache after the first read and then taken from the cache for the further checks in the same request.

So there should be no problem, when different pods have different caches. And because it's a least recently used cache, with a default max size of 128 elements, it will never grow too large and unused entries are removed after a while.

valentijnscholten:
It will be a problem if somebody changes the permissions/members on a product. Only one pod/process will be aware, the others will cache old data.

StefanFl:
Oh, good point, that's bad. I will look for a different solution.

valentijnscholten:
You could look at what we did for system settings possibly

StefanFl:
I have implemented a cache with a configurable time to live now, see https://github.com/DefectDojo/django-DefectDojo/pull/4196/files#diff-fd86831b9463dc44553fd136bbe8ed7ac51b85bfcc60fced475a5ad8e0438fef. With the defaults it is emptied every 10 seconds. That should be a good balance of caching request while rendering the page and a maximum timespan until a change in Product_Members or Product_Type_Members gets active.

dojo/api_v2/views.py
```diff
@@ -1194,35 +1167,30 @@ class TestImportViewSet(prefetch.PrefetchListMixin,
...
+                                        'findings_affected__jira_issue',
+                                        'findings_affected__reviewers',
+                                        'findings_affected__notes',
+                                        'findings_affected__notes__author',
+                                        'findings_affected__notes__history',
+                                        'findings_affected__files',
+                                        'findings_affected__images',
+                                        'findings_affected__found_by',
+                                        'findings_affected__tags',
+                                        'findings_affected__risk_acceptance_set',
```
valentijnscholten, line 1193: On the github diff it's not really clear what happened here, did you rename a field on the Test_Import model?

StefanFl: The field on the Test_Import model is `findings_affected = models.ManyToManyField('Finding', through='Test_Import_Finding_Action')`, I haven't changed that. But when testing the API for Test_Import I got an error message that I fixed here.

valentijnscholten: ah, so that field was actually renamed by me in the original PR but I forgot to update the prefetch in api. raised https://github.com/DefectDojo/django-DefectDojo/pull/4204 to fix that.

StefanFl:
The request cache library had 2 issues but doesn't seem to be maintained anymore. Since it has only 34 LoC plus comments, I copied the code into our repository and attributed it.

From my point of view this PR is now complete and ready for review.

valentijnscholten (approved):

valentijnscholten:
I like the per request cache and agree it's better to include it here as the original package requires another packages to keep track of the request and that adds some not needed ORM fields and middleware and installed app.
Still have to see how heavy this query is for each request, but for now I think we want to get this PR merged so we come closer to feature complete and doing some real life testing.
Thankgs again, approving.

madchap (approved):

PR #4199, opened 2021-04-03
optimize dedupe command, fix product grade

### LINKED ISSUE #4200: dedupe command doesn't update product grade
the dedupe command is running the deduplication on findings, but the product grade is not updated accordingly.

valentijnscholten:
The `./manage.py dedupe` command is used to recalculate the `hash_code` for findings and/or to start the deduplication process for findings. This iterates over all findings, which is slow by default. It could be sped up a little bit by prefetching endpoints and other fields, but for larger instances (>100k) this prefetching becomes very slow as it loads almost the whole database in memory.

On larger instances this means the dedupe command would run for 8+ hours or maybe even more

This PR switches to a paging based approach where the findings are processed and prefetched in batches of 1000 findings. This method is based on what we used in https://github.com/defectdojo/django-DefectDojo/blob/dev/dojo/db_migrations/0082_last_status_update_populate.py

Speed of hash_code generation:

before: ~1000-1500 findings per minute
after: ~90000 findings per minute

The speed of deduplication is not really improved by this PR as most of the work is the deduplication itself and not retrieving the findings. However since #4134 and #4048 the dedupe process has become much more predictable, so the default for the `dedupe` management command is not to do the deduplication in the background using a celery task.
There is a new parameter for the command for those who want it to run in the foreground: `--dedupe_sync`.

Speed of dedupe:
~7500 findings per minute (foreground, so single core)

Added some docs as well, which we can point to from future release notes if a release mandates a run of the `dedupe` command.

Also fixes #4200


dojo/management/commands/dedupe.py
```diff
@@ -1,23 +1,27 @@
...
 
 logger = logging.getLogger(__name__)
 deduplicationLogger = logging.getLogger("dojo.specific-loggers.deduplication")
 
-"""
-Author: Aaron Weaver
-This script will update the hashcode and dedupe findings in DefectDojo:
-"""
+
+def generate_hash_code(f):
```
damiencarol, line 14: Should we make it a method of object `Finding`?

damiencarol: Also as we use this function only when defining a lambda, it's better to move the code of this one line function in the lambda definition directly

damiencarol (approved):

madchap (approved):

PR #4207, opened 2021-04-05
jira: improve UI around inheritance for engagement JIRA config

valentijnscholten:
On an engagement, a user can chose to inherit the JIRA Project Config from the product, or to override it.

image

We have seen some confusion as it's not clear when inheritance will kick in, and when it is considered an override. For example if the user checks or unchecks one of the checkboxes, it is considered an override. In this case the JIRA Instance and Project Key fields become mandatory. If the user doesn't want an override after all, it's unclear which boxes to check/uncheck to revert to inheritance mode.
It's also not clear which settings are inherited.

This PR:
- Adds an explicit "Inherit from product" checkbox
- Show which settings are inherited
- Makes all checkboex unchecked by default to avoid any confusion

image

Additionally:
- It protects the on_delete action for `JIRA_Project` and `JIRA_Instance` to prevent any stored (links to) JIRA Issues from being deleted. In general once a Product or Engagement has a JIRA Project Config, it cannot be deleted. This was already the case in the UI, but is now also protected in the model. This situation is a bit weird, because users can change the project key or jira instance, which could also break existing jira issues. This PR doesn't change that as I'm not yet sure what the best solution would be.


Maffooch (approved):
Great improvement here

madchap (approved):
lgtm, thanks for that.

PR #4209, opened 2021-04-05
import history improvements

valentijnscholten:
- By default show all reimports, not only those that affected findings
- Add filter panel to filter on version, findings affected
- Add branch_tag, commit_hash, build_id to test / import / reimport
- Add branch_tag, commit_hash, build_id to import history filter
- Update our "official" api wrapper to be able to use these fields

In ci/cd situations it makes sense to use reimport to reimport a specifiek branch on every commit or on every day. So we want to have the branch_tag, commit_hash and build_id fields also on the test model. This allows for people to either use an engagement for every (manual) test import or a Test with reimport for every test import.

image

image

PS Typo's in screenshots are fixed


Maffooch (approved):

damiencarol (approved):

damiencarol:
@valentijnscholten `87` is already in dev branch for db migration scripts, we need to rename some file in this pull request I guess.

valentijnscholten:
There are a couple of other small things I need to add, have set it to WIP for now.

PR #4212, opened 2021-04-05
Added skip_duplicates flag when importing scans via APIv2

adiffpirate:
**Added:**

- `skip_duplicates` flag to `api/v2/import-scan/`
- Option to return the original finding when calling the deduplication method instead of saving the duplicate, this allows us to reuse the deduplication logic in other places, reducing the amount of duplicated code and making it easier to add features that require/use deduplication.
- A few docstrings <3
  
**Changed:**

- Now the legacy deduplication method also filters findings by hash_code instead of filtering only by title/cwe and looping through each one comparing hash_codes. This greatly improves performance when working with a lot of findings with similar title/cwe but different hash_codes.

**TODO:**

- [ ] Add unittests
- [ ] Refactor `return_original_without_save`

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1027,6 +1027,7 @@ class ImportScanSerializer(serializers.Serializer):
         queryset=User.objects.all())
     tags = TagListSerializerField(required=False)
     close_old_findings = serializers.BooleanField(required=False, default=False)
+    skip_duplicates = serializers.BooleanField(required=False, default=False)
     push_to_jira = serializers.BooleanField(default=False)
     environment = serializers.CharField(required=False)
     version = serializers.CharField(required=False)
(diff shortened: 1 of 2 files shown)


adiffpirate:
**TODO:**

- [ ] Add unittests

adiffpirate:
I'm thinking about implementing the same `skip_duplicates` logic to reimport, i.e. use the new `do_dedupe_finding` method instead of the current `match_new_finding_to_existing_finding`. That way we make import and reimport logic for deduplication consistent, avoiding the duplicated code at `dojo/reimport_utils.py`.

Although it's important to notice that the current legacy deduplication algorithm implemented at `dojo/reimport_utils.py` is very different from the legacy deduplication algorithm at `dojo/utils.py`, so this could be a major change for some projects. What you think @valentijnscholten ?

valentijnscholten:
I agree that both sets of logic should be identical, or there should be only one piece of logic. Thanks for looking into this, I think it can be helpful, but I think the solution with `return_original_without_save` is very ugly. It's better to have a matching method that is used on all places, including inside `dedupe_finding`

Maffooch:
Hi @adiffpirate how are things going on this effort?

adiffpirate:
@Maffooch I'm thinking about closing this one.

At the time I wrote this to fit a "one engagement per test" workflow (importing each test). But after talking to some people on Slack I felt like the consensus was to create one engagement per branch and reimport the tests, which has way better performance than import, so I changed my workflow.

Since then the reimport feature has improved a lot, so I'm not sure this PR makes sense anymore.

adiffpirate:
@valentijnscholten No man, you addressed all issues I had with reimport in the past. I don't see any problems with it so far (if something comes to mind later I'll hit you up on Slack, don't worry lol). Thanks for the great work.

Maffooch:
Glad the community could help you! Closing

Issue #4222, opened 2021-04-08
Add Path information for ORT security tool parser


devsecopsale:
**Is your feature request related to a problem? Please describe**
ORT file imports don't add the path of the file where the license (issue) was found.
I actually modified the parser to include this feature , which I'm sharing in a pull request shortly.

**Describe the solution you'd like**
I need the 'path' to be included so it's easier to find out what's the component that is affected by the issue. This will help identifying where the problem relies on or if it's a false positive.

**Describe alternatives you've considered**
The path needs to be extracted from the input file.

**Additional context**
NA

PR #4223, opened 2021-04-08
ORT security tool parser: add license path in finding description

devsecopsale:
This will add feature #4222

 - [x] A new value is added to the model in order to allow the finding path to be found: license id number.
 - [x] A new function is added in order to find the path of a license id number in a given findings list.
 - [x] A condition was added in order to capture the cases where findings are present and the path value is applicable (i.e. can be found in the file).

### DIFF
--- a/dojo/tools/ort/parser.py
+++ b/dojo/tools/ort/parser.py
@@ -116,11 +116,11 @@ def get_rule_violation_model(rule_violation_unresolved, packages, licenses, depe
     for id in project_ids:
         project_names.append(get_name_id_for_package(packages, id))
     package = find_package_by_id(packages, rule_violation_unresolved['pkg'])
-    if 'license' in rule_violation_unresolved: 
+    if 'license' in rule_violation_unresolved:
         license_tmp = rule_violation_unresolved['license']
     else:
         license_tmp = 'unset'
-    if 'license_source' not in rule_violation_unresolved: 
+    if 'license_source' not in rule_violation_unresolved:
         rule_violation_unresolved['license_source'] = 'unset'
     license_id = find_license_id(licenses, license_tmp)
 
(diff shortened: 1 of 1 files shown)


damiencarol:
@devsecopsale could you add a check in one unit tests that test the description with the new data?

valentijnscholten (changes requested):
as per damiens request

devsecopsale:
> @devsecopsale could you add a check in one unit tests that test the description with the new data?

Hi. sorry the late. I've been on leave. I'll try to take care of this next week.

StefanFl:
@devsecopsale Are you still considering to add a check for the new functionality in the unit tests?

Maffooch:
> Are you still considering to add a check for the new functionality in the unit tests?

@devsecopsale

devsecopsale:
> > Are you still considering to add a check for the new functionality in the unit tests?
> 
> @devsecopsale

Hi, sorry, I'm not sure which kind of unit test I should add to it. Isn't the standard unit test for the parser enough to cover it?

valentijnscholten:
> > > Are you still considering to add a check for the new functionality in the unit tests?
> > 
> > 
> > @devsecopsale
> 
> Hi, sorry, I'm not sure which kind of unit test I should add to it. Isn't the standard unit test for the parser enough to cover it?

nothing is checked yet at https://github.com/valentijnscholten/django-DefectDojo/blob/e515c0e4030efec056e0ab5536ffe7b6fcf715db/dojo/unittests/tools/test_ort_parser.py#L19, so ideally we add some checks for file path and other fields.

damiencarol:
> @devsecopsale could you add a check in one unit tests that test the description with the new data?

@devsecopsale do you need help to complete this PR?
A simple thing could be to test if the license path is in the description. I can help on this one, you can contact me in Slack.

PR #4227, opened 2021-04-08
Snyk parser: fix CVSS parsing

rmoldesc:
This PR addresses one bug:

- **CVSS3** string field for Snyk Findings was being added without the "CVSS:3.x/" prefix, causing failure in the cvssScore computation: https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/models.py#L2124

In the current implementation, the removal of the "CVSS:3.x/" prefix when parsing the CVSS string was based on the Finding "cvss3" field validation: https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/models.py#L1420 . I am not sure if this validation is correct taking into account the cvssScore computation logic.

Regards

### DIFF
--- a/dojo/tools/snyk/parser.py
+++ b/dojo/tools/snyk/parser.py
@@ -134,7 +134,7 @@ def get_item(vulnerability, test):
         "CVSS score of: **" + str(vulnerability['cvssScore']) + "**",
         cwe=cwe,
         cve=cve,
-        cvssv3=vulnerability['CVSSv3'][9:],
+        cvssv3=vulnerability['CVSSv3'],
         description="## Component Details\n - **Vulnerable Package**: " +
         vulnerability['packageName'] + "\n- **Current Version**: " + str(
             vulnerability['version']) + "\n- **Vulnerable Version(s)**: " +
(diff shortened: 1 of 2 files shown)


damiencarol (approved):

madchap:
> In the current implementation, the removal of the "CVSS:3.x/" prefix when parsing the CVSS string was based on the Finding "cvss3" field validation: https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/models.py#L1420 . I am not sure if this validation is correct taking into account the cvssScore computation logic.

Yea, the regex seems to accept both string, it should probably be made more stringent and actually expect the prefix for the computation.

madchap:
There could be a change in the regex along with a small migration script to loop over findings and prepend the missing prefix if needed.

mtesauro:
This change looks good to me but was @madchap wanting the Regex change and migration script as part of this - it's not clear to me.

damiencarol:
@mtesauro @madchap the whole situation for CVSSv3 vectors is a mess. The regex is wrong. CVSS v3 have prefix like `CVSS:3.0/...` but some parsers cut it and the regex exclude it. Also CVSS v3 vectors could have 'extentions'. 
Plus some part of the UI works only if the prefix is here...
In conclusion, could be good to unify this to vector + prefix everywhere and as @mtesauro said add a data migration.

damiencarol:
Superseded by https://github.com/DefectDojo/django-DefectDojo/pull/4284

PR #4240, opened 2021-04-09
login banner: allow clickable urls

valentijnscholten:
The login banner currently doesn't allow any html or clickable urls. That could be helpful to add a link to login/sso instructions or terms of use or just a link to fred.com:

image

This PR allows the basic html tags allowed by the bleach library and a couple of styling tags, i.e. to give a nice blue color to the url and to let it open in a new tab.

Also tried to simplify the setup and do something straightforward like {% if get_banner.banner_enable %} but apparently the django templates require this weird construct with passing in strings into filters/tags.


damiencarol (approved):

madchap (approved):
Thanks ;-) More traffic to fred.com.

PR #4246, opened 2021-04-09
Harmonize import and reimport logic

valentijnscholten:
This PR is something I wanted to do for over a year now. At first there were not enough unit / integration tests to do this big refactoring. After creating these tests and merging them, I got sidetracked on other PRs. Then there were some other big PRs going on, so I wanted to wait a bit to avoid too many conflicts. But now I need this PR because I am going to do some changes in a next PR around the import/reimport. 

This PR:
- moves all import and reimport code into one place, a new folder `importers`;
- so no more 95% identical copy-and-pasted code between ui and api (re)import;
- extracts some common logic between import and reimport into an utils module;
- standardizes messages and notifications;
- support `close_old_findings` setting on UI Import;

The test cases worked well, as I had to fix some small things found by them.


dojo/api_v2/serializers.py
```diff
@@ -1046,247 +1041,55 @@ class ImportScanSerializer(serializers.Serializer):
...
+                                                                             endpoints_to_add=endpoints_to_add,
+                                                                             scan_date=scan_date, version=version,
+                                                                             branch_tag=branch_tag, build_id=build_id,
+                                                                             commit_hash=commit_hash,
+                                                                             push_to_jira=push_to_jira,
+                                                                             close_old_findings=close_old_findings)
+        # convert to exception otherwise django rest framework will swallow them as 400 error
+        # exceptions are already logged in the importer
+        except SyntaxError as se:
+            raise Exception(se)
```
damiencarol, line 1086: could we use ValidationError or another one instead of untyped exception?

dojo/engagement/views.py
```diff
@@ -569,45 +586,32 @@ def import_scan_results(request, eid=None, pid=None):
...
-                tags=tags)
-            t.lead = user
-            t.full_clean()
-            t.save()
+                add_success_message_to_response(message)
+
+            except Exception as e:
+                # exceptions are already logged by the importer
+                add_error_message_to_response('An exception error occurred during the report import:%s' % str(e))
+                error = True
```
damiencarol, line 614: the way exception are managed is weird. catch everything is never the best solution. :/ also if we really want to log the exception, I prefer this form:

```Python
except XxxxxError as ex:
    logger.exception("something bad happened during import bla bla bla")
```

dojo/importers/importer/importer.py
```diff
@@ -0,0 +1,276 @@
...
+            target_end=scan_date if scan_date else now.date(),
+            environment=environment,
+            percent_complete=100,
+            version=version,
+            branch_tag=branch_tag,
+            build_id=build_id,
+            commit_hash=commit_hash,
+            tags=tags)
+        try:
+            # TODO What is going on here?
```
damiencarol, line 48: `full_clean()` validate attributes regarding the model.
this could lead to ValidationError. but... the try/catch structure completely swallow it. something is wrong here and need to change.
1) remove the bloc
or
2) log something

dojo/importers/importer/importer.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def process_parsed_findings(self, test, parsed_findings, scan_type, user, active, verified, minimum_severity=None,
+                                endpoints_to_add=None, push_to_jira=None, now=timezone.now()):
+        logger.debug('endpoints_to_add: %s', endpoints_to_add)
+        new_findings = []
+        items = parsed_findings
+        logger.debug('starting import of %i items.', len(items) if items else 0)
+        i = 0
+        for item in items:
+            sev = item.severity
```
damiencarol, line 64: please remove the complete bloc.
I want to enforce by unit tests that what's coming from parsers are ok better than relying on hack like this one in the upper levels

dojo/importers/importer/importer.py
```diff
@@ -0,0 +1,276 @@
...
+            if sev == 'Information' or sev == 'Informational':
+                sev = 'Info'
+
+            item.severity = sev
```
damiencarol, line 68: please remove the complete bloc.
I want to enforce by unit tests that what's coming from parsers are ok better than relying on hack like this one in the upper levels

damiencarol:
it's an awesome pull request! need!
Just one thing, is it possible to add unit tests for importer/re-importer ?

damiencarol (approved):

damiencarol:
Made some comments to improve some parts but ti's good as is for me as it's improving a lot the part arounf import/reimport

dojo/importers/importer/importer.py
valentijnscholten: this is old code, might need to leave it around until all the parsers are updated/tested?

dojo/api_v2/serializers.py
valentijnscholten: those will be swallowed by the rest framework and turned into 400 errors without any info.

dojo/engagement/views.py
valentijnscholten: that is exactly what happens one level deeper. but here we catch the exception to show a nice message in the UI. In api we expose the exception, that's why it all looks weird. but it's all existing code just moved somewhere else. people are free to improve further in future prs.

dojo/importers/importer/importer.py
valentijnscholten: I just moved this code, not really looking into trying out what happens if we change anything

valentijnscholten:
> Just one thing, is it possible to add unit tests for importer/re-importer ?

There are quite some unit tests that allowed us to have this PR, i.e. `test_import_reimport.py`, `test_jira_import_and_pushing_api.py`, `dedupe_test.py`. Including import -> reimport -> reimport with changed reports.

dojo/engagement/views.py
damiencarol: ok. as discussed in Slack, if error management is part of "normal" workflow, we should return a boolean with the result that make explicit if the import was a success or not instead of this exception programming. but as you said it's all existing code just moved. we can improve it in further PRs

dojo/importers/importer/importer.py
damiencarol: as you said it's all existing code just moved. we can improve it in further PRs

madchap (approved):
Only glanced at it, not tested. I will get to dogfood this some time later, but no time right now. In our tests we trust.

PR #4247, opened 2021-04-09
Harmonize managing of users with the rest of the UI

StefanFl:
The dialogues to manage users are quite different from the rest of DefectDojo's UI, with buttons to edit and add users and no dialogue to view user data. This PR brings managing of users in line with the rest of the UI:

List of users no has a three dot menu to view, edit or delete users:

2021-04-09 16_55_10-All Users _ DefectDojo

New dialogue to view users (current authorization):

2021-04-09 16_55_39-View User _ DefectDojo

Same dialogue for authorization v2 (including managing of memberships for the user):

2021-04-09 16_58_15-View User _ DefectDojo


mtesauro (approved):
I quite like the new views - nicely done.  I guess I just got used to how it looked but it's a good thing to bring the users UI into harmony.

Approved

valentijnscholten (approved):

PR #4249, opened 2021-04-10
Minor improvements to SpotBugs and Dependency Check parsers

adiffpirate:
**SpotBugs**

The report contains some HTML code that was not being properly parsed.
Now, using BeautifulSoup, we can work with tags and deliver a more precise report.
Enhancements:
- A more clear mitigation.
- References moved to where they belong with links that work.

dd_spotbugs_compare

**Dependency Check**
- Added mitigation


valentijnscholten:
Thanks. Are there any unit tests / samples you could add that "prove" the new html parsing part is working correctly?

dojo/tools/spotbugs/parser.py
```diff
@@ -4,9 +4,8 @@
 __status__ = "Development"
 
 import re
-
+from bs4 import BeautifulSoup
```
damiencarol, line 7: We use `html2text` for this part of the project. Please switch to the package `html2text`

damiencarol (changes requested):
Can't introduce BS instead of `html2text` that we use in a lot of parsers

adiffpirate:
I'm not sure if these unittests for the HTML parsing are good enough. Let me know what you guys think ;)
_(Also I'm sorry for taking so long to update this PR, it's been a crazy couple of months)_

setup.py
```diff
@@ -0,0 +1,75 @@
...
+        'jsonlines==2.0.0',  # requred by yarn audit parser
+    ],
+
+    extras_require={'mysql': ['mysqlclient==2.0.3']},
+
+    dependency_links=[
+        "https://github.com/grendel513/python-pdfkit/tarball/master#egg=pdfkit-0.5.0"
+    ],
+    url='https://github.com/DefectDojo/django-DefectDojo'
+)
```
valentijnscholten, line 75: this file was removed

dojo/tools/dependency_check/parser.py
```diff
@@ -238,6 +238,10 @@ def get_finding_from_vulnerability(self, dependency, related_dependency, vulnera
                                      'source: {1}\n' \
                                      'url: {2}\n\n'.format(name, source, url)
 
+        component_name, component_version = self.get_component_name_and_version_from_dependency(dependency, related_dependency, namespace)
+
```
valentijnscholten, line 242: don't think this is needed/ intended?

valentijnscholten (changes requested):

dojo/tools/spotbugs/parser.py
adiffpirate: Done :)

dojo/tools/dependency_check/parser.py
adiffpirate: Yup, it was not intended, thanks for catching that

setup.py
adiffpirate: Oops, sorry for that

valentijnscholten (approved):

damiencarol (approved):

PR #4256, opened 2021-04-12
add support django i18n

epochfairy:
modify: settings.dist.py
templeates: base.html，
add: locale
image


valentijnscholten:
How does this work? When I look at the Djano docs, they suggest using the `translate` tags in template to implement i8n. But this PR doesn't change any templates. Is it using some kind of generic "match and replace" logic to replace certain strings in http responses or in the output of template rendering? How does it know it needs to change the word "Login" on the login button, but not the word "Login" in a (english) description of a finding for example?

valentijnscholten:
BTW any non hotfix PRs should be based against `dev` as instructed in the PR template.

epochfairy:
Or you can realize internationalization

PR #4260, opened 2021-04-12
Added ability to use business days or calendar days

Hijerboa:
At my company, and I'd imagine in others as well, remediation SLA's are written in business days, not calendar days. 

The current implementation of SLA's exclusively uses calendar days. This update uses numpy's busday_count() function to calculate findings age, and in turn SLA in terms of business days, rather than calendar days.

Whether to calculate using calendar days or business days is controlled by a new setting, USE_BUSDAY_COUNT, located in the settings.dist.py file. When set to False, calendar days are used. When set to True, business days are used.


valentijnscholten:
Thank you for the PR. i don't think attackers will count in business days, but I get the idea ;-)
Could you rename the variable to something like SLA_BUSINESS_DAYS and locate it close to the other SLA related settings? And add a DD_SLA_BUSINESS_DAYS as well? And a 1 line comment explaining the purpose of the variable?

valentijnscholten (changes requested):
see above

Hijerboa:
Hi @valentijnscholten thanks for the feedback! I've updated the settings file as requested.

valentijnscholten (approved):

Maffooch (approved):

PR #4182, opened 2021-03-31
[anchore parser] add cvssv3_score when available

madchap:
Implement CVSS v3 score for Anchoer Engine parser

PR #4264, opened 2021-04-12
Grype - Align name and change default dedupe algo

madchap:
The name was off compared to all other scanners, showing `anchore_grype` instead of `Anchore Grype`.

Also, the dedupe was not working at all. The cve is not a unique ID for any tool. Changed the default dedupe algo to be closer to the ones from anchore-cli.


madchap:
OK, I am good to revoke my key..

damiencarol:
@madchap you said
> Also, the dedupe was not working at all. The cve is not a unique ID for any tool. Changed the default dedupe algo to be closer to the ones from anchore-cli.

1) But for this scanner it is. Take a look at the data:

```JSON
{
      "vulnerability": {
        "id": "CVE-2020-36221",
```

Also in the pull request you are remove implementation of `unique_id_from_tool`.

2) CVE is a unique ID in our current model. You can only have one and only one CVE for a Finding.
> The cve is not a unique ID for any tool. 

Strongly disagree with this assertion. CVE *are* uniq IDs. We should be able to de-duplicate between tools with CVEs. 

3) I need to understand better your pb. Did you saw some Ids from the tool that where not CVE ? If it's the pb, you're right there is some modifications needed.

damiencarol (changes requested):
Putting on hold as there is some discussions around the de-duplication

valentijnscholten (approved):

valentijnscholten (changes requested):
I think this will cause a new Test_Type to be created and creating a disconnect between existing tests and new tests for Anchore Grype. i.e. a data migration might be needed.

valentijnscholten:
In https://github.com/DefectDojo/django-DefectDojo/pull/4182/files#diff-7fb4443091c77732d3a550092087171effe27b6eafaafadaf97110db0789c812R88 the CVE was set as unique_id_from tool for the main anchore scanner

madchap:
> I think this will cause a new Test_Type to be created and creating a disconnect between existing tests and new tests for Anchore Grype. i.e. a data migration might be needed.

I thought that with @damiencarol 's refactor, this was a thing of the past? Like right now, there is no mention of grype in `test_type.json`.

madchap:
Adding `vuln_id_from_tool` per discussion in https://owasp.slack.com/archives/C014H3ZV9U6/p1619635577100800

damiencarol:
@madchap test type is still saved in DB by upper layers. so we need a DB migration for this one.

valentijnscholten (approved):
first() is fine because name is unique for Test_Type.

docs/content/en/integrations/import.md
```diff
@@ -23,7 +23,8 @@ like `anchore-cli --json image vuln <image:tag> all`
 
 JSON report format.
 
-### Anchore Grype (anchore_grype)
+Anchore Grype
+-------------
```
damiencarol, line 27:
```suggestion
### Anchore Grype
```

damiencarol (approved):

damiencarol:
@madchap approved but the doc have a syntax difference

PR #4267, opened 2021-04-13
Enable disable scanners

dsever:
This PR enables isStuff role to enable/disable loaded scanners.

First fixtures are required  and migration. 

After migration Stuff can disable/enable selected scanner, delete it or add new one.  Only enabled scanner will appear in dropdown menus  and as api supported.

After this I will drop PR that enables scanner dedicated configuration. 

Todo:

- [ ] test for visibility in importform  
- [ ] tests for API


damiencarol:
converting to draft as we are working on it

PR #4270, opened 2021-04-13
Refactor numerical_severity in importer/reimporter

damiencarol:
Move `item.numerical_severity = Finding.get_numerical_severity(sev)` in upper layers (importer/reimporter) to avoid having this code in *every* parser.

Found a bug in the process (fixed):
https://github.com/DefectDojo/django-DefectDojo/blob/96f2f34814b46d83fec85c42c88d01e291ea64ec/dojo/tools/outpost24/parser.py#L58-L61
numerical_severity is **not** cvss_score


mtesauro (approved):
Approved

damiencarol:
@mtesauro fixing 3 flake formatting pbs and merging after another review

valentijnscholten:
nice cleanup. should we add a couple of `asserts` to test_import_reimport or elsewhere to check that it's set correctly or at least not empty?

damiencarol:
I can but as we will remove it soon :/

valentijnscholten:
> @valentijnscholten Added a new test class with a method to test the attribute https://github.com/DefectDojo/django-DefectDojo/blob/f6fa0bea9223144b7a63ac51fd33525bf03fecb5/dojo/unittests/test_importers_importer.py

Thanks, approving and let you merge when the tests are green :-)

valentijnscholten (approved):

valentijnscholten:
>> nice cleanup. should we add a couple of asserts to test_import_reimport or elsewhere to check that it's set correctly or at least not empty?

>  I can but as we will remove it soon :/

I thought about that, but it might that we still need a numerical severity field for sorting, but as an int and not a varchar. Anyway, good to go now.

Issue #3908, opened 2021-02-22
tests: add from template should be tested


valentijnscholten:
Just fixed two bugs in add finding from template, we should add this feature to the integration tests.

PR #4273, opened 2021-04-14
fix Nonetype when adding from template

valentijnscholten:
fixes 2 exceptions in 1.14.1 when adding a finding from a template:

```
  File "/home/valentijn/dd/dojo/forms.py", line 2283, in __init__
    if self.instance.has_jira_group_issue:
AttributeError: 'NoneType' object has no attribute 'has_jira_group_issue'
```

and

```
  File "/home/valentijn/dd/dojo/finding/helper.py", line 58, in update_finding_status
    if 'is_Mitigated' in changed_fields or is_new_finding:
TypeError: argument of type 'NoneType' is not iterable
```

### DIFF
--- a/dojo/finding/helper.py
+++ b/dojo/finding/helper.py
@@ -45,7 +45,9 @@ pre_save_changed.connect(pre_save_finding_status_change, sender=Finding, fields=
 def update_finding_status(new_state_finding, user, changed_fields=None):
     now = timezone.now()
 
-    is_new_finding = changed_fields and len(changed_fields) == 1 and 'id' in changed_fields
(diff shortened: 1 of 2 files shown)


valentijnscholten:
if only we had added that test :-) https://github.com/DefectDojo/django-DefectDojo/issues/3908

damiencarol:
@valentijnscholten  is it too much work to add those unit tests?

damiencarol (approved):

JamesCullum:
Just experienced this breaking bug as well. Do we know which version introduced the bug exactly? This way we can downgrade in the meantime.

madchap (approved):

PR #4282, opened 2021-04-15
fix __str__ for GitHub_Issue model

valentijnscholten:
the __str__ was referencing invalid field names, causing an exception when trying to delete a product with github integration enabled.


damiencarol (approved):

damiencarol:
@valentijnscholten could you add a unit test for this one? should be simple.

valentijnscholten:
the github integration is completely broken anyway, so not really worth the time to create unit tests for a __str__ method

alles-klar (approved):

PR #4283, opened 2021-04-15
deletion: dupe-cascade refactor and fixes

### LINKED ISSUE #4304: ForeignKey violation when deleting engagements with duplicates
When deleting engagements that contain duplicates, from time to time there are ForeigKeyErrors. This is mainly when we have duplicates with the original inside the same engagement, or, with duplicates of duplicates (transitive duplicates).
The workaround is that everyone runs `fix_loop_duplicates()` but that's not ideal.

valentijnscholten:
Fixes #4304

In #4142 we introduced a new default to not delete all duplicates when deleting an original.
This works fine in most cases, but in complex cases it seems to still throw some errors from time to time on the found_by foreign key violation or endpoint_status foreign key violation. I have not been able to pinpoint the root cause, but it could be because of django trying to do everything related to the delete and its cascading actions inside a single transaction. Causing later steps in the delete to see old data that was modified by earlier steps or the other way around. Or something completely different.

I have now refactored the code to try and preprocess any duplicates of originals that are going to be deleted, and the complex deletes that were failing are now working successfully. It's quite a complex synergy (or crapiness) between `.delete()` not being called for bulk or cascaded deletes. But `pre_delete` signal is being called, but AFTER cascaded deletes have been performed. So we have to use `.delete()` to prepare duplicates for deletion of the original.

Again Django seems to have created a mechanism with which it's almost impossible to write clean/working code.

I think it's now in a better shape with this PR. Worst case users can set `DD_DUPLICATE_CLUSTER_CASCADE_DELETE` to True  to revert to the old behavior.

I added some unit tests around deletion of engagements and products. I tested both modes locally with a 1M finding instance with around 500k duplicates, and possibly some complex engagement with duplicates inside the same test and some transitive duplicates from before we fixed the logic around that. So the deletion/cascading seems to be a lot more robust with this PR.

The delete logic now calls `fix_loop_duplicates()` to resolve any transitive / circular duplicate relations that might have come to be in the past.


madchap (approved):
Looks good, thanks :)

devGregA (approved):

devGregA:
Approved @valentijnscholten, please let me know if you want me to go ahead squash / merge.

PR #4290, opened 2021-04-16
Authorization V2: Feature-complete, ready for testing

StefanFl:
This PR, together with PR #4247 makes the implementation of the new authorization feature-complete, so that it can be end-2-end tested. I worked on the permissions based on being a Product_Member or Product_Type_Member for:

- API for Import and Reimport
- API for Product_Members and Product_Type_Members
- Components
- Finding_Groups
- Selection of engagement lead and test lead
- Notifications
- GitLab
- Bulk edit and delete

About notifications: I had to fix a lot of create_notification() calls, where the product, engagement, test or finding weren't set. Because of that, notifications where send to all users, not only authorized users. This has been fixed for the current authorization as well.

About GitLab: I couldn't test this implementation, because I can't create an OAuth2 connection to a GitLab instance.

Question: What to do with Tags? Currently all tags are visible for every user. I wouldn't change that, tags shouldn't contain sensible information.

There are still some things missing before declaring general availability, which will come with following PRs:

- Integration tests
- Documentation
- Migration procedure from authorized users to Product_Members/Product_Type_Members

### DIFF
--- a/dojo/api_v2/permissions.py
+++ b/dojo/api_v2/permissions.py
@@ -104,6 +104,11 @@ class UserHasFindingPermission(permissions.BasePermission):
         return check_object_permission(request, obj, Permissions.Finding_View, Permissions.Finding_Edit, Permissions.Finding_Delete)
 
 
+class UserHasImportPermission(permissions.BasePermission):
+    def has_permission(self, request, view):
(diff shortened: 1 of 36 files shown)


valentijnscholten (approved):
awesome work again!

madchap (approved):
:tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada:

PR #4291, opened 2021-04-16
delete jira issues when product/engagement is deleted

valentijnscholten:
Recently we introduced a different `on_delete` strategy for a JIRA Project configuration. turns out this blocks us from deleting products or engagements which have JIRA Issues attached to their findings. This PR switches back to the previous setting to cascade deletes from products/engagements JIRA Config into the JIRA Issues. Otherwise we would never be able to delete the products/engagements as there is is no easy way for users to remove all the JIRA Issues from the project.
Maybe in the future we need a more advanced delete UI making more clear what is being deleted, opt out of certain cascading, etc.

We still have the "PROTECT" strategy on JIRA Instance as there it makes sense. We don't want someone to "delete an old jira instance" from DD and mistakenly delete all JIRA Issues created against that instance.


madchap:
> We still have the "PROTECT" strategy on JIRA Instance as there it makes sense. We don't want someone to "delete an old jira instance" from DD and mistakenly delete all JIRA Issues created against that instance

Where is the fun then?

dojo/models.py
```diff
@@ -2828,7 +2828,7 @@ class JIRA_Conf_Admin(admin.ModelAdmin):
 
 
 class JIRA_Issue(models.Model):
-    jira_project = models.ForeignKey(JIRA_Project, on_delete=models.PROTECT, null=True)  # just to be sure we don't delete JIRA_Issue if a jira_project is deleted
+    jira_project = models.ForeignKey(JIRA_Project, on_delete=models.CASCADE, null=True)  # just to be sure we don't delete JIRA_Issue if a jira_project is deleted
```
madchap, line 2831: Maybe remove the comment then?

madchap:
Can we maybe have a specific warning message if this occurs? I am guessing this would be listed in the big delete screen, but may be drown in it. I think it could take some people by surprise.

valentijnscholten:
yes, this PR just reverts to the old behavior and we may need some larger work to improve the deletion confirmation screen.

madchap (approved):

dojo/models.py
valentijnscholten: done

alles-klar (approved):

PR #4292, opened 2021-04-16
Spotbugs: fix error on missing start line value

SPoint42:
Some time, start is not defined and cause an 500 error.
This patch cover the error

### DIFF
--- a/dojo/tools/spotbugs/parser.py
+++ b/dojo/tools/spotbugs/parser.py
@@ -60,7 +60,8 @@ class SpotbugsParser(object):
             source_extract = bug.find('SourceLine')
             if source_extract is not None:
                 source_file = source_extract.get("sourcepath")
-                source_line = int(source_extract.get("start") if source_extract.get("start") is not None else "0")
+                if 'start' in source_extract and source_extract["start"].isdigit():
+                    source_line = int(source_extract["start"])
 
             if dupe_key in dupes:
                 finding = dupes[dupe_key]
--- a/dojo/tools/spotbugs/parser.py
+++ b/dojo/tools/spotbugs/parser.py
@@ -60,7 +60,7 @@ class SpotbugsParser(object):
             source_extract = bug.find('SourceLine')
             if source_extract is not None:
                 source_file = source_extract.get("sourcepath")
-                source_line = int(source_extract.get("start"))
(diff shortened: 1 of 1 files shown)


valentijnscholten:
i thinkt the desired solution is to set the line fields to `None` if the linue is not known.

SPoint42:
> i thinkt the desired solution is to set the line fields to `None` if the linue is not known.

But sast_source_line and source_line is an integer in the model....

`sast_source_line = models.IntegerField(` and ` line = models.IntegerField(`

so ?

valentijnscholten:
Just don't set the field and leave it None.

valentijnscholten:
Just don't set the field if there is no value.

damiencarol:
I agree. Better use this structure:

```Python
if 'line' in source_extract and source_extract["line"].isdigit():
  finding.line = int(source_extract["line"])
```

damiencarol:
@SPoint42 any news on this one?
Also could you add a report that break in the PR as unit test ?

dojo/tools/spotbugs/parser.py
```diff
@@ -60,7 +60,7 @@ def get_findings(self, filename, test):
             source_extract = bug.find('SourceLine')
             if source_extract is not None:
                 source_file = source_extract.get("sourcepath")
-                source_line = int(source_extract.get("start"))
+                source_line = int(source_extract.get("start") if source_extract.get("start") is not None else "0")
```
damiencarol, line 63:
```suggestion
                if 'start' in source_extract and source_extract["start"].isdigit():
                    source_line = int(source_extract["start"])
```

SPoint42:
> @SPoint42 any news on this one?
Sorry, I was busy on other subject...

> Also could you add a report that break in the PR as unit test ?
will see what I can do with an anonymize report and a unit test.

dojo/tools/spotbugs/parser.py
SPoint42: Thanks for the quick commit :)

damiencarol:
@SPoint42 just a gentle reminder that we will ship the version `2.0.0`next week. So if by any chance you can finish this PR, it will be shipped.

Maffooch:
@SPoint42 Is this still an active effort?

damiencarol:
Superseded by #5068

PR #4295, opened 2021-04-16
reimporter: make close_old_findings True by default everywhere

valentijnscholten:
Both the API and UI set close_old_findings to True for reimport, so all works well.
But in the reimporter.py kwarg the default is False, which is confusing and could lead to future bugs.


madchap (approved):

Maffooch (approved):

damiencarol:
Good PR !

PR #4296, opened 2021-04-16
[Nexpose parser] Fix error with Python 3.9

damiencarol:
Python 3.9 removed support for `getchildren` according to https://docs.python.org/3/whatsnew/3.9.html#removed

> * Methods getchildren() and getiterator() of classes ElementTree and Element in the ElementTree module have been removed. They were deprecated in Python 3.2. Use iter(x) or list(x) instead of x.getchildren() and x.iter() or list(x.iter()) instead of x.getiterator(). (Contributed by Serhiy Storchaka in bpo-36543.)

https://bugs.python.org/issue36543

This could lead to error: 
```
AttributeError: 'xml.etree.ElementTree.Element' object has no attribute 'getchildren'
```


valentijnscholten (approved):
approved assuming the tests go green

madchap (approved):

PR #4299, opened 2021-04-16
Add Microsoft CredScan parser

Maffooch:
Add CredScan parser

https://secdevtools.azurewebsites.net/helpcredscan.html

### DIFF
--- a/dojo/tools/cred_scan/parser.py
+++ b/dojo/tools/cred_scan/parser.py
@@ -1,0 +1,67 @@
+import csv
+import io
+from datetime import datetime
+
+
+from dojo.models import Finding
+
+
+class CredScanParser(object):
+
+    def get_scan_types(self):
+        return ["CredScan Scan"]
+
+    def get_label_for_scan_types(self, scan_type):
+        return "CredScan Scan"
+
+    def get_description_for_scan_types(self, scan_type):
+        return "Import CSV output of CredScan scan report."
+
+    def get_findings(self, filename, test):
+        content = filename.read()
+        if type(content) is bytes:
+            content = content.decode('utf-8-sig')
+        reader = csv.DictReader(io.StringIO(content), delimiter=',', quotechar='"')
+
+        dupes = dict()
+        for row in reader:
+            # Severity is not provided in this scanner, so marking all as info
+            severity = 'Info'
+
+            # Create the description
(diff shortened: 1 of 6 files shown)


valentijnscholten:
can you update the docs and PR desc with some info i.e. pointer to what scanner this is / their homepage?

dojo/tools/cred_scan/parser.py
```diff
@@ -0,0 +1,67 @@
...
+    def get_label_for_scan_types(self, scan_type):
+        return "CredScan Scan"
+
+    def get_description_for_scan_types(self, scan_type):
+        return "Import CSV output of CredScan scan report."
+
+    def get_findings(self, filename, test):
+        content = filename.read()
+        if type(content) is bytes:
+            content = content.decode('utf-8-sig')
```
damiencarol, line 23: Any reason to use the BOM encoding? I thnik when files come from the API part, the encoding will be 'utf8'

Maffooch: For some reason, the first field (TimeOfDiscovery) in the csv file came in with `u'\ufeff'` at the beginning. I was using the API to test, and was not having any issues.

valentijnscholten: I think `utf-8-sig` will handle both cases correctly, with and without BOM.

devGregA (approved):

mtesauro (approved):
Approved

damiencarol:
@Maffooch I know it's already merged but could you take a look at these comments?

dojo/tools/cred_scan/parser.py
```diff
@@ -0,0 +1,67 @@
...
+        reader = csv.DictReader(io.StringIO(content), delimiter=',', quotechar='"')
+
+        dupes = dict()
+        for row in reader:
+            # Severity is not provided in this scanner, so marking all as info
+            severity = 'Info'
+
+            # Create the description
+            description = row.get('Description', 'Description not provided')
+            is_suppressed = row.get('IsSuppressed', None)
```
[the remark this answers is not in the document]
Maffooch: Nope. Not sure what the field is referring to. Figured it would be useful in the description to those that use the tool and they can clarify later if it does not make sense.

PR #4302, opened 2021-04-16
Helm chart linting - pick target branch

### LINKED ISSUE #4176: Helm chart lint: bump version on release or change GHA
Whenever we do a release, the helm chart linting GHA fails because it expects us to increase the version of the helm chart.

Solution options:
- Update version of helm chart automatically or manually when merging a release back into dev, or, 
- Only run linting on PRs that actually change the helm chart similar to flake8.

madchap:
Should fix #4176

On master to take effect.

### DIFF
--- a/.github/workflows/test-helm-chart.yml
+++ b/.github/workflows/test-helm-chart.yml
@@ -40,13 +40,20 @@ jobs:
       - name: Run chart-testing (list-changed)
         id: list-changed
         run: |
-          changed=$(ct list-changed --config ct.yaml)
+          if [ ! -z ${GITHUB_BASE_REF} ]; then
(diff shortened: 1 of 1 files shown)


valentijnscholten:
What is the idea now with this change? Will it ignore changes made in other branches? I think in general it made sense that the linting complained about the versions as we bumped the defect dojo version.

valentijnscholten:
Read the code again, is it now only raising an error if the chart was actually changed in a pr?

madchap:
> Read the code again, is it now only raising an error if the chart was actually changed in a pr?

Right.

An actual pull request made from a release branch to master for example triggers a push event hence the else part.

madchap:
By default, target_branch was always there, but implicitly always against master.

valentijnscholten (approved):

alles-klar (approved):

PR #4305, opened 2021-04-18
view_finding: show hash_code in title of ID field

valentijnscholten:
sometimes it's useful to be able to see the hash_code values for findings when debugging some deduplication behaviour.
also add output to the `dedupe.py` command to show when a hash_code is actually changed.

image


damiencarol (approved):

ptrovatelli (approved):
nice, I didn't know we could add a tooltip with the title attribute :)

PR #4315, opened 2021-04-19
[Nexpose parser] add identification of L4 protocol tcp/udp for DNS service

kiblik:
Nexpose parser: DNS service - add identification of L4 protocol

In the case of DNS, some vulnerabilities are relevant only for UDP and some of them are relevant for TCP+UDP. It is important to identify UDP and TCP services as different endpoints.

### DIFF
--- a/dojo/tools/nexpose/parser.py
+++ b/dojo/tools/nexpose/parser.py
@@ -232,7 +232,7 @@ class NexposeParser(object):
                     endpoint = Endpoint(
                         host=host['name'],
                         port=service['port'],
-                        protocol=service['name'].lower() if service['name'] != "<unknown>" else None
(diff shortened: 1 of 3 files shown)


valentijnscholten:
can you add a PR description / motivation?

kiblik:
In the case of DNS, some vulnerabilities are relevant only for UDP and some of them are relevant for TCP+UDP. It is important to identify UDP and TCP services as different endpoints.

I added unittest which helps to understand the situation.

valentijnscholten:
I know, but I meant adding it to the description so that people can read it from the release notes.

Maffooch (approved):

valentijnscholten (approved):
Thanks!

PR #4322, opened 2021-04-20
Swap ptvsd for debugpy, ptvsd deprecated #4115

### LINKED ISSUE #4115: Swap ptvsd for debugpy, ptvsd deprecated
As mentioned by @ptrovatelli  at https://owasp.slack.com/archives/C014H3ZV9U6/p1616534386011200

> so it seems that ptvsd is deprecated, replaced by debugpy (see https://github.com/microsoft/ptvsd ). ptvsd is what we use for debugging with docker.

yilmi:
Trying to move to [debugpy](https://github.com/microsoft/debugpy) in replacement of ptvsd. fixes #4115
Changed a few things:

- Removed the ptvsd profile (setEnv.sh, compose file, entrypoint script and other references)
- Added support for two new environment variables:
     - `DD_DEBUG_PORT` to provide a custom port number for the debugging server
     - `DD_DEBUG_WAIT_FOR_CLIENT` pauses the execution until the debugging client is connected (doesn't work in multiprocess or multithreaded with uwsgi)

Tested debugging, managed to consistently hit breakpoints, inspect variables, etc...


docker-compose.override.dev.yml
```diff
@@ -8,7 +8,12 @@ services:
     environment:
       DD_DEBUG: 'True'
       DD_ADMIN_USER: ${DD_ADMIN_USER:-admin}
-      DD_ADMIN_PASSWORD: ${DD_ADMIN_PASSWORD:-admin}      
+      DD_ADMIN_PASSWORD: ${DD_ADMIN_PASSWORD:-admin}
+    ports:
+      - target: 3000
+        published: 3000
```
madchap, line 14: Could we eventually make this `${DD_DEBUG_PORT:-3000}`

madchap:
From my previous config, I simply:
- Rebuilt the image to get the new dep
- Relinked from `ptvsd` et `dev` 
--> and I could attach to debugpy and breakpoint with no further tweaks :+1: 

My VScode setup remained identical.

Thanks!

docker-compose.override.dev.yml
yilmi: Changed, thanks :+1:

madchap (approved):

docker/entrypoint-uwsgi-dev.sh
```diff
@@ -9,14 +9,19 @@ cd /app
...
+# Full list of uwsgi options: https://uwsgi-docs.readthedocs.io/en/latest/Options.html
+# --lazy-apps required for debugging --> https://uwsgi-docs.readthedocs.io/en/latest/articles/TheArtOfGracefulReloading.html?highlight=lazy-apps#preforking-vs-lazy-apps-vs-lazy
+
 exec uwsgi \
   "--${DD_UWSGI_MODE}" "${DD_UWSGI_ENDPOINT}" \
   --protocol uwsgi \
   --wsgi dojo.wsgi:application \
   --enable-threads \
-  --processes ${DD_UWSGI_NUM_OF_PROCESSES:-2} \
-  --threads ${DD_UWSGI_NUM_OF_THREADS:-2} \
```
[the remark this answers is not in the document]
valentijnscholten: not visible here, but the PR seems to hardcode it to 1 for both params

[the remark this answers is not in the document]
yilmi: I had issues when using uwsgi, debugpy works better with it when there's only one process and one thread. 

I did hardcode `1` and remove the variable substitution for `DD_UWSGI_NUM_OF_PROCESSES` and `DD_UWSGI_NUM_OF_THREADS`, because these variables were always set in the Dockerfile and the helm chart.

Otherwise, the debugging the container would have the 2 processes/2 threads that are an issue with uwsgi.

valentijnscholten:
I think users should be able to disable debugging if they don't want to use it for whatever reason (performance, overhead, ...)
And as such I think we should still support running more than 1 process/thread for those who want to test things related to multi threading.

madchap:
Yea, I think this was the reason we kept the 2 separate for a long time.

Might need something fancier to handle variables and keep it flexible.

damiencarol (approved):

helm/defectdojo/templates/django-deployment.yaml
```diff
@@ -93,8 +93,8 @@ spec:
         - name: http
           protocol: TCP
           containerPort: 8081
-        {{- if .Values.django.uwsgi.enable_ptvsd }}
-        - name: ptvsd
+        {{- if .Values.django.uwsgi.enable_debugpy }}
+        - name: debugpy
           protocol: TCP
           containerPort: 3000
```
madchap, line 99: Needs to be a variable now too, per the changes in the Dockerfile.

madchap (changes requested):
As @valentijnscholten mentioned, we'd need a way to control whether or not we want debugpy active or not and based on that, be able to run multi-process and multi-thread (which indeed is troublesome when wanting to attach a debugger).

helm/defectdojo/templates/django-deployment.yaml
yilmi: Added in lines 113 to 116

yilmi:
> I think users should be able to disable debugging if they don't want to use it for whatever reason (performance, overhead, ...)
> And as such I think we should still support running more than 1 process/thread for those who want to test things related to multi threading.

I changed the dev entrypoint to override the default if DD_DEBUG is set. I also added back the debug profile (with its docker-compose override)

valentijnscholten (approved):

madchap:
I'll give this a spin in the following days. Looks good at first sight though.

requirements.txt
```diff
@@ -49,7 +49,7 @@ social-auth-app-django==4.0.0
 social-auth-core==4.1.0
 Python-jose==3.2.0
 gitpython==3.1.14
-ptvsd==4.3.2
+debugpy>=1.2.1
```
madchap, line 52: Please make it a `==` (for dependabot)

madchap (changes requested):
Please bump the version with a minor in Chart.yml to pass the linting.
To be on the safe side, quickly rebase too.

Cheers :)

requirements.txt
damiencarol: it's not only for dependabot :)

helm/defectdojo/values.yaml
```diff
@@ -149,7 +149,7 @@ django:
     app_settings:
       processes: 2
       threads: 2
-    enable_ptvsd: false  # this also requires DD_DEBUG to be set to True
+    enable_debugpy: false  # this also requires DD_DEBUG to be set to True
```
damiencarol, line 152: Could you name it just `enable_debug` or something like that so we don't have to rename everything every time we switch from one lib to another?

helm/defectdojo/templates/django-deployment.yaml
```diff
@@ -110,6 +110,10 @@ spec:
             secretKeyRef:
               name: defectdojo-{{ .Values.celery.broker }}-specific
               key: {{ .Values.celery.broker }}-password
+        {{- if .Values.django.uwsgi.enable_debugpy }}
```
damiencarol, line 113:
```suggestion
        {{- if .Values.django.uwsgi.enable_debug }}
```

requirements.txt
damiencarol, line 52:
```suggestion
debugpy==1.2.1
```

damiencarol (changes requested):
Made some minor rename comment and also pin the version for debugpy

helm/defectdojo/values.yaml
yilmi: Let me change that quickly

requirements.txt
yilmi: You're right, accepted changes

damiencarol (approved):

valentijnscholten:
@madchap it's better to only rerun the failed tests instead of closing/reopening and running all tests again.

PR #4323, opened 2021-04-20
SpotBugs: Use hash_code for deduplication and minor enhancements

StefanFl:
With the legacy deduplication algorithm, the reimport of a SpotBugs report closes a majority of the findings. When using the hash_code it produces the expected results.

[EDIT] hash_code uses now the line number as well. There can be multiple instances for one CWE in a file, eg. when you have a controller providing some REST-Endpoints.

Furthermore there are 2 minor enhancements of the parser:

- Newlines for the description, otherwise it is just a long unreadable string
- Removal of 'N/A' for impact and references because they can be NULL now

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -843,7 +843,7 @@ HASHCODE_ALLOWS_NULL_CWE = {
     'DSOP Scan': True,
     'Acunetix Scan': True,
     'Trivy Scan': True,
-    'SpotBugs Scan ': False,
+    'SpotBugs Scan': False,
 }
 
 # List of fields that are known to be usable in hash_code computation)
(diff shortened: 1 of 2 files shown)


madchap (approved):

StefanFl:
Sorry, need to check if we need changes for the hash_code

valentijnscholten:
converted to draft as per latest comment from @StefanFl

StefanFl:
I have no idea, why the tests are failing. In my local environment everything works.

valentijnscholten:
image

dojo/settings/settings.dist.py
```diff
@@ -843,7 +843,7 @@ def generate_url(scheme, double_slashes, user, password, host, port, path, param
     'DSOP Scan': True,
     'Acunetix Scan': True,
     'Trivy Scan': True,
-    'SpotBugs Scan ': False,
+    'SpotBugs Scan': False,
```
damiencarol, line 846: good catch!

damiencarol (approved):

PR #4328, opened 2021-04-21
always log the admin password

madchap:
Even when a hardcoded default is set, show the admin password in the logs.


dsever:
@madchap from sec perspective this could open  risk, can you raise message that it has to be changed immediately. Even Logs will be rotated on docker level, still it will be preserved within Log Management Systems.

madchap:
> @madchap from sec perspective this could open risk, can you raise message that it has to be changed immediately. Even Logs will be rotated on docker level, still it will be preserved within Log Management Systems.

On first use, not any more than right now, since it is logged if it's not the default `admin` in most cases.

I guess you're right if you've specified a default. Forget it then.

PR #4329, opened 2021-04-21
Adjust jira risk acceptance to use helper

madchap:
When a risk acceptance was coming through a jira webhook, the result would be different than when using the UI and doing it manually.

This small changes adjusts it so that it shows up the same way.


valentijnscholten (approved):

madchap:
Superseded by #4332

PR #4331, opened 2021-04-21
Import History: fix Endpoint JSON Serialization issue

alles-klar:
A server error occurs when importing a scan in the apiv2 with the parameter endpoint_to_add and TRACK_IMPORT_HISTORY=True. 

This PR adds a test which (currently) breaks.

---
On behalf of DB Systel GmbH

### DIFF
--- a/dojo/importers/utils.py
+++ b/dojo/importers/utils.py
@@ -69,10 +69,15 @@ def update_import_history(type, active, verified, tags, minimum_severity, endpoi
 
     # tags=tags TODO no tags field in api for reimport it seems
     if endpoints_to_add:
-        import_settings['endpoints'] = str(endpoints_to_add)
+        import_settings['endpoints'] = [str(endpoint) for endpoint in endpoints_to_add]
+
+    print('import_settings')
+    print(import_settings)
 
     test_import = Test_Import(test=test, import_settings=import_settings, version=version, branch_tag=branch_tag, build_id=build_id, commit_hash=commit_hash, type=type)
     test_import.save()
+
+    print('saved!')
 
     test_import_finding_action_list = []
     for finding in closed_findings:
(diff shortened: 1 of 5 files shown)


valentijnscholten:
@alles-klar thanks for reporting, I think I fixed it, including the UI variant of your new unit test ;-)

dojo/importers/utils.py
```diff
@@ -69,7 +69,7 @@ def update_import_history(type, active, verified, tags, minimum_severity, endpoi
 
     # tags=tags TODO no tags field in api for reimport it seems
     if endpoints_to_add:
-        import_settings['endpoints'] = endpoints_to_add
+        import_settings['endpoints'] = [str(endpoint) for endpoint in endpoints_to_add]
```
alles-klar, line 72: @valentijnscholten do we really what to save the string representation of the endpoint or it is better to save the id of the endpoint? 
I prefer the id.

alles-klar:
> @alles-klar thanks for reporting, I think I fixed it, including the UI variant of your new unit test ;-) I will let you "undraft" it if you agree.

@valentijnscholten thanks for the fix!

dojo/importers/utils.py
valentijnscholten: maybe both? To  me the id says nothing and I'd like to see the actual URI

alles-klar: Maybe we save the id in the database but display the string representation in the ui?

valentijnscholten: If we add ids in the json, we have to lots of database lookups to display the URI in the UI. Maybe we should move it out of the settings and as its own field / relation on `Test_Import`.
For now maybe merge only the bugfix so it can be part of the release on Tuesday? I fixed the flake8 violation.

valentijnscholten (approved):

damiencarol (approved):

valentijnscholten:
@alles-klar can you double check? I fixed another bug

damiencarol (approved):
good catch

Issue #4330, opened 2021-04-21
Jira resolution not applied correctly if set as a 2nd step


madchap:
**Bug description**
When using JIRA resolution to set the finding's status as FP or RA, I guess you have 2 choices in JIRA:
1. You set a transition screen where you can set the resolution immediately before it actually transitions to "Done"
2. You do it afterwards, meaning the JIRA will be "Done", resolution set to "Done" (default) and THEN only set the appropriate resolution such as FP or RA.

In that 2nd scenario, the first webhook will have come in already and the set the DD finding to inactive. So whenever you set the RA or FP, it actually re-opens the finding as we're checking on `finding.active` being `True` :-D

PR #4332, opened 2021-04-21
JIRA webhooks to properly handle JIRA resolution when set as 2nd step

madchap:
Addresses #4330

Going in if the JIRA resolution is set, but then make no assumption on whether the DD finding is active or not.
Based on the JIRA resolution, it will simply go back in and re-set attributes.
Only if the resolution is removed, then the finding would be forced to re-open.


madchap:
The happy path works quite well at least. What am I missing @valentijnscholten ?

dojo/jira_link/helper.py
```diff
@@ -1341,47 +1341,38 @@ def process_resolution_from_jira(finding, resolution_id, resolution_name, assign
...
+        if jira_instance and resolution_name in jira_instance.accepted_resolutions:
+            logger.debug("Marking related finding of {} as accepted. Creating risk acceptance.".format(jira_issue.jira_key))
+            finding.active = False
+            finding.mitigated = None
+            finding.is_Mitigated = False
+            finding.false_p = False
+            ra = Risk_Acceptance.objects.create(
+                accepted_by=assignee_name,
+                owner=finding.reporter,
+            )
```
valentijnscholten, line 1354: This will create a new Risk Acceptance every time the issue gets updated in JIRA and the resolution is one that is in the list of accepted resultions. I think this code is only needed to be executed if the finding is not yet risk accepted.

Also, the risk acceptance should be tied to the engagement (pre existing problem), i.e. `finding.test.engagement.risk_acceptance.add(ra)`

dojo/jira_link/helper.py
```diff
@@ -1341,47 +1341,38 @@ def process_resolution_from_jira(finding, resolution_id, resolution_name, assign
...
+            status_changed = True
+        elif jira_instance and resolution_name in jira_instance.false_positive_resolutions:
+            logger.debug("Marking related finding of {} as false-positive".format(jira_issue.jira_key))
+            finding.active = False
+            finding.verified = False
             finding.mitigated = None
             finding.is_Mitigated = False
+            finding.false_p = True
+            ra_helper.risk_unaccept(finding)
+            status_changed = True
```
valentijnscholten, line 1365: this will execute every time a finding is updated and qualifies as false positive. generating db traffic and "empty" entries in the audit log. Also the setting of status_changed to True will lead to the reconciliation job outputting the the status of the finding was changed, while in fact it might very well be that it wasn't changed.

All this should only be executed if the finding is not yet a false positive.

dojo/jira_link/helper.py
```diff
@@ -1341,47 +1341,38 @@ def process_resolution_from_jira(finding, resolution_id, resolution_name, assign
...
+            logger.debug("Marking related finding of {} as mitigated (default)".format(jira_issue.jira_key))
+            finding.active = False
+            finding.mitigated = jira_now
+            finding.is_Mitigated = True
+            finding.mitigated_by, created = User.objects.get_or_create(username='JIRA')
+            finding.endpoints.clear()
             finding.false_p = False
-            ra_helper.remove_from_any_risk_acceptance(finding)
+            ra_helper.risk_unaccept(finding)
             status_changed = True
```
valentijnscholten, line 1376: basically same as above, only execute if needed.

valentijnscholten (changes requested):

madchap:
Thanks for the feedback. I need to think of this audit log I use to have disabled, going to get that database manager up from now on ;-)

valentijnscholten (approved):

damiencarol (approved):

PR #4342, opened 2021-04-22
No risk acceptance on duplicates

madchap:


valentijnscholten:
Not sure if I agree with the approach, shouldn't we just hide the risk acceptance options (both) in the UI for duplicates? Now the add_risk_acceptance method is "abused" to redirect back to the finding view.

madchap:
Right - I was not sure I was agreeing myself on my approach. I'll work on an alternative. Thanks for the feedback.

valentijnscholten (approved):

mtesauro:
Roasting some coffee and looking at PRs and I'm guessing this is going to be revamped or closed, right @madchap ?

StefanFl:
@madchap: This PR is open now for a long time. Does it still make sense to review it or shall it be revamped or closed as @mtesauro suggested?

PR #4346, opened 2021-04-23
AuthV2: Change wording (admin set user perms)

kiblik:
admin add "prodct (type)", no "member"


StefanFl:
Hi @kiblik, if we name it this way, it would be consistent to change the corresponding menus in view_product_details.html and view_product_type.html to "Add user". Would you mind changing this as well?

valentijnscholten (approved):

Maffooch (approved):

PR #4349, opened 2021-04-23
Re-import fix "Finding has not test" exception

Maffooch:
When re-importing, the hashcode of a finding is computed before it is compared with other findings. The test type is used in this hashcode. Some parsers, such as Qualys WebApp, do not supply the test to each finding object it creates. This results in the following exception:

`dojo.models.Finding.test.RelatedObjectDoesNotExist: Finding has no test.`

to combat this, check for a test object before computing the hash code, and if it does not exist, add the test to the finding. This step was initially done if a new finding was actually created if another finding did not match the new one, but moving that step a bit further up does not change anything.

### DIFF
--- a/dojo/importers/reimporter/reimporter.py
+++ b/dojo/importers/reimporter/reimporter.py
@@ -58,6 +58,9 @@ class DojoDefaultReImporter(object):
             component_name = item.component_name if hasattr(item, 'component_name') else None
             component_version = item.component_version if hasattr(item, 'component_version') else None
 
+            if not hasattr(item, 'test'):
+                item.test = test
+
             item.hash_code = item.compute_hash_code()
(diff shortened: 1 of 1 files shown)


valentijnscholten:
I was thinking about this last week, why don't we always set the test in the importer? Good find!

valentijnscholten (approved):

valentijnscholten:
Maybe add a little comment to remind future "clean code" fanatics to not remove this assignment?

devGregA:
Yeah, wondering if it is more performant to just set it and not do the if check.

devGregA:
Actually being nit picky, it's probably more performant as written.

devGregA (approved):

PR #4366, opened 2021-04-26
jira: fix silent epic error during unit tests

valentijnscholten:
during unit tests every push to JIRA resulted in an error because dojo was trying to add the issue to an epic, which didn't exist.
this error is swallowed silently by dojo so it was not failing the tests, but the logs were insinuating something was wrong.

most of the files changed are just re-recordings of the jira traffic

### DIFF
--- a/dojo/unittests/test_jira_import_and_pushing_api.py
+++ b/dojo/unittests/test_jira_import_and_pushing_api.py
@@ -32,7 +32,7 @@ logger = logging.getLogger(__name__)
 # please check the recorded files on sensitive data before committing to git
 
 
-class JIRAConfigAndPushTestApi(DojoVCRAPITestCase):
+class JIRAImportAndPushTestApi(DojoVCRAPITestCase):
     fixtures = ['dojo_testdata.json']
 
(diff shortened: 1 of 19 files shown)


madchap (approved):

dojo/unittests/test_jira_import_and_pushing_api.py
```diff
@@ -44,8 +44,8 @@ def assert_cassette_played(self):
             self.assertTrue(self.cassette.all_played)
 
     def _get_vcr(self, **kwargs):
-        my_vcr = super(JIRAConfigAndPushTestApi, self)._get_vcr(**kwargs)
-        my_vcr.record_mode = 'once'
+        my_vcr = super(JIRAImportAndPushTestApi, self)._get_vcr(**kwargs)
+        my_vcr.record_mode = 'none'
```
Maffooch, line 48: Was this change intended?

valentijnscholten: I think none is the right one for running existing unit tests

Maffooch (approved):

Issue #4319, opened 2021-04-19
Qualys Scan report import  fails to decode CVSS field


kareem-DA:
**Bug description**

I am using docker-compose to deploy.  I currently have the running on DDJ 1.8.0, but I have been testing on version 1.14.1 and It has the same issue.  

I am using the "Qualys Scan" import when I see this.  

I am using the API with a python script to to do the import.

so, I am   
       # Prior section, I get the report from the Qulays API store it into a byteIO object named "storage"
       # Get the project ID
        project_id = return_product_id(dd, dojo_project)
        # Get an engagement_id
        engagement_id = return_engagement(dd, project_id, args.dojoUser)

        active = True
        verified = True
        close_old_findings = True
        skip_duplicates = True
        scan_date = datetime.now().strftime("%Y-%m-%d")
        build = None
        scan_type = "Qualys Scan"
        minimum_severity = "High"
        push_to_jira = False
        tags = None

        data = {
            "file": storage.getvalue(),
            "engagement": ("", engagement_id.data["id"]),
            "scan_type": ("", scan_type),
            "active": ("", active),
            "verified": ("", verified),
            "close_old_findings": ("", close_old_findings),
            "skip_duplicates": ("", skip_duplicates),
            "scan_date": ("", scan_date),
            "tags": ("", tags),
            "build_id": ("", build),
            "minimum_severity": ("", minimum_severity),
            "push_to_jira": ("", push_to_jira),
        }
        response = dd._request("POST", "import-scan/", files=data)
        if response.success is False:
            _logger.error(
                "Posted Qualys results to DDJ failed. Response message: {}".format(
                    response.message
                )
            )
        dd.close_engagement(engagement_id.data["id"])

When I import a qualys scan, I am getting an error.  

```log
uwsgi_1         | ERROR:django.request:Internal Server Error: /api/v2/import-scan/
uwsgi_1         | Traceback (most recent call last):
uwsgi_1         |   File "./dojo/api_v2/serializers.py", line 1097, in save
uwsgi_1         |     parser_findings = parser.get_findings(data.get('file', None), test)
uwsgi_1         |   File "./dojo/tools/qualys/parser.py", line 233, in get_findings
uwsgi_1         |     return qualys_parser(file)
uwsgi_1         |   File "./dojo/tools/qualys/parser.py", line 216, in qualys_parser
uwsgi_1         |     master_list += issue_r(issue, d)
[...]

PR #4373, opened 2021-04-27
[Qualys parser] fix parsing issues when cvss field contains a vector

alles-klar:
Qualys has added a cvss vector to some of the cvss field. The parser is now able to handle this. Solves issue #4319.

----

On behalf of DB Systel GmbH

### DIFF
--- a/dojo/tools/qualys/parser.py
+++ b/dojo/tools/qualys/parser.py
@@ -2,6 +2,7 @@ import datetime
 import logging
 import html2text
 from defusedxml import ElementTree as etree
+from cvss import CVSS3
 
 from dojo.models import Endpoint, Finding
 
@@ -59,7 +60,7 @@ def split_cvss(value, _temp):
         split = value.split(" (")
         _temp['CVSS_value'] = float(split[0])
         # remove ")" at the end
-        _temp['CVSS_vector'] = split[1][:-1]
+        _temp['CVSS_vector'] = CVSS3("CVSS:3.0/" + split[1][:-1]).clean_vector()
     else:
         _temp['CVSS_value'] = float(value)
 
--- a/dojo/tools/qualys/parser.py
+++ b/dojo/tools/qualys/parser.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+import re
 import html2text
 from defusedxml import ElementTree as etree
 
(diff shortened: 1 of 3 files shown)


dojo/unittests/tools/test_qualys_parser.py
```diff
@@ -31,3 +31,17 @@ def test_parse_file_with_multiple_vuln_has_multiple_findings(self):
...
+        self.assertEqual(
+            finding.unsaved_endpoints[0].host, "demo14.s02.sjc01.qualys.com"
+        )
+        self.assertEqual(
+            # CVSS_FINAL is defined without a cvssv3 vector
+            finding.cvssv3, None
+        )
+        self.assertEqual(
+            finding.severity, "High"
+        )
```
damiencarol, line 47: Could you add a check that use the new way to gather the vector?

dojo/tools/qualys/parser.py
```diff
@@ -185,6 +220,8 @@ def parse_finding(host, tree):
         finding.mitigated = _temp['mitigation_date']
         finding.is_Mitigated = _temp['mitigated']
         finding.active = _temp['active']
+        if _temp.get('CVSS_vector') is not None:
+            finding.cvssv3 = _temp.get('CVSS_vector')
```
damiencarol, line 224: We are mixing vector v2 and v3 here :/ 
Sadly for now the data model doesn't support v2 so I'm not a fan of putting wrong data here

damiencarol (changes requested):

dojo/tools/qualys/parser.py
alles-klar: v2 would not match the regex and therefore not used here

dojo/tools/qualys/parser.py
```diff
@@ -51,6 +52,23 @@ def htmltext(blob):
...
+        _temp['CVSS_value'] = float(split[0])
+        # remove ")" at the end and validate
+        cvssv3_regex = regex = re.compile('^AV:[NALP]|AC:[LH]|PR:[UNLH]|UI:[NR]|S:[UC]|[CIA]:[NLH]', re.I)
+        if cvssv3_regex.match(split[1][:-1]):
+            _temp['CVSS_vector'] = split[1][:-1]
+        else:
+            logger.warn("%s does not match cvssv3 regex", split[1][:-1])
+    else:
+        _temp['CVSS_value'] = float(value)
+
```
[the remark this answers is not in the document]
alles-klar: I removed the regex check because it will be checked in the model.

dojo/unittests/tools/test_qualys_parser.py
alles-klar: good point, found a bug with the new test

Maffooch (approved):

dojo/tools/qualys/parser.py
alles-klar: if the vector comes from v2 is set to None now

mtesauro (approved):
Approved

dojo/tools/qualys/parser.py
damiencarol: ok

damiencarol (approved):

damiencarol:
Approved after some discussions with @alles-klar . more PR will come to improve some part.

Issue #4359, opened 2021-04-26
Python: Update to 3.7 or 3.8


valentijnscholten:
We always aimed to not be too strict on the minimum version of python required to run Defect Dojo. With the arrival of our Docker installs, it becomes less relevant. Currently we're on 3.6 which was release FIVE years ago. In #2978 we see that there's an urgent need now to update to 3.7 at least.

valentijnscholten:
We may want to discuss which python version to use. Sometimes you see that packages are quite aggressive with deprecating older versions, leaving us with old and possibly buggy/vulnerable versions of those libraries. i.e. I prefer to go to 3.8 or even 3.9.

damiencarol:
Could be good to start a draft to see if we discover incompatibilities with Python > 3.6.x (like https://github.com/DefectDojo/django-DefectDojo/pull/4296 )

valentijnscholten:
3.7 is part of #2978

madchap:
I think many run with 3.7+ already nowadays. Ubuntu LTS comes with 3.8 by default now, but RHEL 8 is still stuck at 3.6. RHEL 9 is around the corner I believe.

Python 3.6 is EOL end of 2021 with only security fixes until then.

I'd say it all depends on the support we want to provide to RHEL 8 people out of the box, for example.

mtesauro:
I am fine with moving Python to 3.8.  From my testing, godojo use 3.8 when it does an install (tested on Ubuntu 20.04 LTS and 20.10) so at the very least, DefectDojo can be installed and run with 3.8 without any blatant issues.

CentOS/RHEL are always a problem and have historically shipped with less than up-to-date versions of many of their packages.  Lots of places are still using CentOS/RHEL 7 but those who do still have the option of using docker/docker-compose where the OS provided Python version doesn't matter.  That may be the best alternative for OSes that aren't using more recent versions of Python.  At the very least, they have options beyond a native-OS install (godojo).  Also, it is possible to get newer versions of Python if you use epel and other means to bring newer packages to the RHEL/CentOS family of distros but that's not something I think we should directly support since there are other options.

valentijnscholten:
Looks like 3.8 would be better for RHEL:
![image](https://user-images.githubusercontent.com/4426050/116195546-a5fda900-a732-11eb-874e-fff047ca9ac3.png)

valentijnscholten:
But Debian Buster has 3.7 and BullsEye will have 3.9.

[...]

PR #4378, opened 2021-04-28
Bump python to 3.7

valentijnscholten:
It's time to update Python to at least 3.7, see more discussions in #4359.


valentijnscholten:
The kubernetes deployment keeps failing on Python 3.7, but works fine consitently on 3.8 so I propose we go with 3.8.

PR #2978, opened 2020-10-08
pip-2020: test with new dependency resolver

valentijnscholten:
https://discuss.python.org/t/announcement-pip-20-2-release/4863

testing the soon-to-become-new-default-dependency-resolver in pip using `pip  --use-feature=2020-resolver`

image

PR #4379, opened 2021-04-28
Bump python to 3.8

### LINKED ISSUE #4359: Python: Update to 3.7 or 3.8

valentijnscholten:
The kubernetes deployment in GHA keeps failing on Python 3.7, but works fine consitently on 3.8 so I propose we go with 3.8.
Fixes #4359

Python 3.8 also uses pip 21.1 which, by default, uses the new dependency resolver algorithm that caused a lot of issues in the past, but is now finally resolved as we can see in the tests in this PR and in #2978

### DIFF
--- a/Dockerfile.django
+++ b/Dockerfile.django
@@ -5,7 +5,7 @@
 # Dockerfile.nginx to use the caching mechanism of Docker.
 
 # Ref: https://devguide.python.org/#branchstatus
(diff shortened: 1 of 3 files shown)


damiencarol:
seems good :D

damiencarol (approved):

madchap (approved):

PR #4380, opened 2021-04-28
k8s: increase initializer memory

valentijnscholten:
Thanks to @dsever we finally found out why the GHA k8s workflow was failing randomly. This PR enables logging and that showed that the initilializer sometimes went out of memory. So we increase the limits a little bit and now it seems to run more consistently successful.

### DIFF
--- a/.github/workflows/k8s-testing.yml
+++ b/.github/workflows/k8s-testing.yml
@@ -181,13 +181,15 @@ jobs:
       - name: Check Application
         run: |-
              to_complete () {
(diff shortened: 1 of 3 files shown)


valentijnscholten:
The kubernetes deployment keeps failing on Python 3.7, but works fine consitently on 3.8 so I propose we go with 3.8.

alles-klar (approved):

madchap (approved):
OMG, @valentijnscholten doing some k8s :tada:

PR #4383, opened 2021-04-28
Anchore parser: add optional custom vulnerability description

ccojocar:
Add an optional field to add a custom vulnerability description in the Anchore parser. This will allow to enrich the finding description with further details which describes the vulnerability. 

For example, one could augment the details by grabbing informations from Anchore at https://anchore-api.mydomain.com/v1/query/vulnerabilities?id=CVE-ID URL and populate DefectDojo with more useful data.

### DIFF
--- a/dojo/tools/anchore_engine/parser.py
+++ b/dojo/tools/anchore_engine/parser.py
@@ -30,7 +30,9 @@ class AnchoreEngineParser(object):
             findingdetail += '**Feed**: ' + item['feed'] + '/' + item['feed_group'] + '\n\n'
             findingdetail += '**CVE**: ' + cve + '\n\n'
             findingdetail += '**CPE**: ' + item['package_cpe'] + '\n\n'
-            findingdetail += '**Description**: ' + item.get('description', '') + '\n\n'
+            vulndescription = item.get('description', '')
+            if vulndescription != '':
+                findingdetail += '**Description**: ' + vulndescription + '\n\n'
 
             sev = item['severity']
             if sev == "Negligible" or sev == "Unknown":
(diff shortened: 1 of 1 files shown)


dojo/tools/anchore_engine/parser.py
```diff
@@ -30,6 +30,7 @@ def get_findings(self, filename, test):
             findingdetail += '**Feed**: ' + item['feed'] + '/' + item['feed_group'] + '\n\n'
             findingdetail += '**CVE**: ' + cve + '\n\n'
             findingdetail += '**CPE**: ' + item['package_cpe'] + '\n\n'
+            findingdetail += '**Description**: ' + item.get('description', '') + '\n\n'
```
madchap, line 33: As this field is not in the default anchore's json, could we actually fully hide if under an `if` if there is nothing to be seen instead of having an empty description?

madchap (changes requested):

dojo/tools/anchore_engine/parser.py
ccojocar: Changed to not append an empty description.

madchap (approved):

dojo/tools/anchore_engine/parser.py
```diff
@@ -30,6 +30,9 @@ def get_findings(self, filename, test):
             findingdetail += '**Feed**: ' + item['feed'] + '/' + item['feed_group'] + '\n\n'
             findingdetail += '**CVE**: ' + cve + '\n\n'
             findingdetail += '**CPE**: ' + item['package_cpe'] + '\n\n'
+            vulndescription = item.get('description', '')
+            if vulndescription != '':
+                findingdetail += '**Description**: ' + vulndescription + '\n\n'
```
damiencarol, line 35:
```suggestion
            findingdetail += '**Description**: ' + item.get('description', '<None>') + '\n\n'
```

damiencarol (approved):

PR #4388, opened 2021-04-28
Allow alternate service types without enabling gke

joesteffee:
Deploying in EKS with an ALB ingress requires either type=LoadBalancer or type=NodePort. The only way to enable NodePort today is by enabling GKE which enables some other stuff we don't want. This allows the user to set a custom service type

Enable use of custom paths for ingress (alb ingress controller requires /* instead of /)


madchap:
@joesteffee can you please fix the linting? Cheers.

madchap:
Also, please rebase onto `dev`, not `master`, as indicated in the PR template. Cheers.

madchap (changes requested):
See above comments.

PR #4395, opened 2021-04-29
Anchore parser, use vuln_id_from_tool instead of unique_id_from_tool

madchap:
As discussed in https://owasp.slack.com/archives/C014H3ZV9U6/p1619635577100800

No real impact out of the box, unless for people that may have used a different dedupe algo than the default for this parser.


damiencarol (approved):

valentijnscholten:
what about existing findings?

madchap:
> what about existing findings?

You just love migrations :trollface:

valentijnscholten (approved):

PR #4397, opened 2021-04-30
Recognize .nessus file names

pburkholder:
As a new DefectDojo user, I thought I had a corrupt .nessus file, until I renamed it with .xml.

This change recognizes XML exports with the `.nessus` extension, matching the documentation at line #245.

Also clarifies the error when trying to import a Nessus file.

### DIFF
--- a/dojo/tools/nessus/parser.py
+++ b/dojo/tools/nessus/parser.py
@@ -246,9 +246,9 @@ class NessusParser(object):
 
     def get_findings(self, filename, test):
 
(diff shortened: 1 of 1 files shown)


damiencarol (approved):
Thanks for contributing. Very efficient patch!

damiencarol (approved):

madchap (approved):

madchap:
Oupsie. Clicked too fast. This was made again `master` and not `dev`. I reverted.

PR #4410, opened 2021-04-30
 Allow alternate service types without enabling gke #4388 

joesteffee:
Deploying in EKS with an ALB ingress requires either type=LoadBalancer or type=NodePort. The only way to enable NodePort today is by enabling GKE which enables some other stuff we don't want. This allows the user to set a custom service type

Enable use of custom paths for ingress (alb ingress controller requires /* instead of /)


dojo/fixtures/benchmark_requirement.json
```diff
@@ -7492,7 +7492,7 @@
     "pk": 418,
     "fields": {
       "category": 27,
-      "objective_number": "1.6.1",
+      "objective_number": "1.6.2",
```
madchap, line 7495: This file is the wrong file, I guess you were looking for Chart.yaml instead to bump the chart number.
It seems the linting didn't catch it, so I'll have to look again. However, you'd want to bump the Chart version to 1.6.3.

madchap (changes requested):

dojo/fixtures/benchmark_requirement.json
joesteffee: fixed

madchap (approved):

mtesauro (approved):
Approved

mtesauro:
@madchap Not sure why those k8s tests are getting cancelled TBH

PR #4415, opened 2021-05-01
REMOVED: S0/1/2/3/4/5 severity display option

valentijnscholten:
Removed an old display option to display severities in the form of S0, S1, S2... instead of Blocker, Critical, High, etc.

Recreated from #3754 due to lack of response and having no access to update that PR branch.


damiencarol:
@valentijnscholten I can keep track of this removal in the parsers after this PR

dojo/models.py
```diff
@@ -2209,15 +2202,8 @@ def clean(self):
...
-        try:
-            system_settings = System_Settings.objects.get()
-            if system_settings.s_finding_severity_naming:
-                return self.numerical_severity
-            else:
-                return self.severity
-
-        except:
-            return self.severity
+        system_settings = System_Settings.objects.get()
```
damiencarol, line 2205:
```suggestion
```

damiencarol (approved):

dojo/models.py
damiencarol: minor comment which remove a DB query.

madchap (approved):

PR #4422, opened 2021-05-02
search: don't fail on endpoints without product

valentijnscholten:
Simple search view would raise an exception if an `EndPoint` without `product` was in the search results.

### DIFF
--- a/dojo/templates/dojo/simple_search.html
+++ b/dojo/templates/dojo/simple_search.html
@@ -246,14 +246,18 @@
                                       </sup>
                                   {% endif %}
                                 </td>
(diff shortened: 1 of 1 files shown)


kiblik:
How `Endpoint` without `Product` can exist?

damiencarol:
@kiblik the foreign key column is null-able:
https://github.com/DefectDojo/django-DefectDojo/blob/bea0f43a27a65a9a1906bdcbc5f9c242172327b5/dojo/models.py#L1112
So it's possible in our current model

mtesauro (approved):
Approved

damiencarol (approved):

PR #4423, opened 2021-05-02
Tags: allow filtering on NOT having tags, add API filters for transitive tags

valentijnscholten:
I have spoken a few people who are using tags extensively and all of them arrive at some point where want to find entities that do NOT have a specific or set of tags. This PR implements "negative tag search" or whatever it should be called.

I also noticed that the `test-tags`, `engagement-tags` and `product-tags` filters where not yet implemented in the API, so this PR adds that as well.


madchap:
fyi @Wadeck

mtesauro (approved):
Approved

madchap (approved):

Wadeck (commented):
Seems fine, looking forward to test it "live" 🎉

Issue #4426, opened 2021-05-02
dedupe management fails


madchap:
**Bug description**
A clear and concise description of what the bug is. For errors include at least the exact error message you are seeing (including traceback).

While testing https://github.com/DefectDojo/django-DefectDojo/pull/4353 (but certainly unrelated, had a glitch in my auto test setup).

```
$ docker-compose exec uwsgi /bin/bash -c 'python manage.py dedupe'
[02/May/2021 21:48:03] INFO [dojo.models:3514] enabling audit logging
[02/May/2021 21:48:06] INFO [dojo.management.commands.dedupe:53] ######## Will process the full database with 392 findings ########
[02/May/2021 21:48:06] INFO [dojo.management.commands.dedupe:57] ######## Start Updating Hashcodes (foreground) ########
[02/May/2021 21:48:06] INFO [dojo.utils:1968] hash_code computation 0 out of 392 models processed ...
[02/May/2021 21:48:06] INFO [dojo.utils:1997] hash_code computation 392 out of 392 models processed ...
[02/May/2021 21:48:06] INFO [dojo.management.commands.dedupe:63] ######## Done Updating Hashcodes########
[02/May/2021 21:48:06] INFO [dojo.management.commands.dedupe:68] ######## Start deduplicating (background) ########
Traceback (most recent call last):
  File "manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/app/dojo/management/commands/dedupe.py", line 73, in handle
    mass_model_updater(Finding, findings.only('id'), lambda f: do_dedupe_finding_task(f.id), order='desc', log_prefix='deduplicating ')
TypeError: mass_model_updater() missing 1 required positional argument: 'fields'
```

PR #4427, opened 2021-05-02
Fix exception when running dedupe management command

madchap:
Fixes https://github.com/DefectDojo/django-DefectDojo/issues/4426

### DIFF
--- a/dojo/management/commands/dedupe.py
+++ b/dojo/management/commands/dedupe.py
@@ -67,10 +67,10 @@ class Command(BaseCommand):
             if get_system_setting('enable_deduplication'):
                 logger.info("######## Start deduplicating (%s) ########", ('foreground' if dedupe_sync else 'background'))
                 if dedupe_sync:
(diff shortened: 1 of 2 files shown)


valentijnscholten:
What I had as a bug a while ago is that I forgot to pass in the `fields` parameter somewhere, causing no fields to be updated. So I made it mandatory, but forgot dedupe.py it seems. I prefer to make the parameter mandatory, as having no fields is a non-default case and it would be good to force people to pass in `fields=None` if they don't want any fields to be updated.
i.e. change `dedupe.py` instead of `utils.py` :-)

valentijnscholten (changes requested):
adf

madchap:
That's what I thought first, but then I got lazy.

valentijnscholten:
I can fix it as it was my bug

madchap:
It's ok, you fix enough of mine :)

valentijnscholten (approved):

damiencarol (approved):

PR #4431, opened 2021-05-03
Add SecureCodeBox parser

JohannesZahn:
Implemented a [secureCodeBox](https://github.com/secureCodeBox/secureCodeBox) Findings Import. 

It is used in a secureCodeBox [Hook](https://docs.securecodebox.io/docs/hooks/defectdojo) to automatically import scans into DefectDojo (for scans that have no dedicated parser yet).

The Findings format is described [here](https://docs.securecodebox.io/docs/api/finding)

See this [issue](https://github.com/secureCodeBox/secureCodeBox/issues/332) for more information.


dojo/fixtures/test_type.json
```diff
@@ -1,723 +1,730 @@
 [
```
damiencarol, line 1: This file will be removed soon.
It's better to let it as-is. It is not required anymore to make modifications in it.

dojo/tools/securecodebox/parser.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def get_label_for_scan_types(self, scan_type):
+        return "SecureCodeBox Findings Import"
+
+    def get_description_for_scan_types(self, scan_type):
+        return "SecureCodeBox Findings file in JSON format can be imported."
+
+    def get_findings(self, file, test):
+        if file is None:
+            return []
```
damiencarol, line 23:
```suggestion
```

dojo/tools/securecodebox/parser.py
```diff
@@ -0,0 +1,89 @@
...
+        try:
+            tree = json.load(file)
+        except json.JSONDecodeError:
+            raise ValueError("file is not valid json")
```
damiencarol, line 27: You're swallowing the exception. it's better for use to have the root cause as it help us to help users.
```suggestion
        tree = json.load(file)
```

dojo/tools/securecodebox/parser.py
```diff
@@ -0,0 +1,89 @@
...
+        else:
+            dupes = dict()
+            for idx, content in enumerate(tree):
+                finding = Finding(
+                    static_finding=True, dynamic_finding=False, active=False, verified=False)
+                node = tree[idx]
```
damiencarol, line 33: Could you use this form instead? Easier to read.

```Python
for node in tree:
```

dojo/tools/securecodebox/parser.py
```diff
@@ -0,0 +1,89 @@
...
+            endpoint.protocol = attributes.get('protocol', endpoint.protocol)
+        finding.unsaved_endpoints = [endpoint]
+
+    def set_finding_base_info(self, finding, node, test):
+        na = 'N/A'
+        # using node.get('name',na) does not work because the value can still be None
+        finding.title = node.get('name') or na
+        finding.description = node.get('description') or na
+        finding.mitigation = na
+        finding.impact = na
```
damiencarol, line 75:
```suggestion
```
These 2 fields are optional

dojo/tools/securecodebox/parser.py
```diff
@@ -0,0 +1,89 @@
...
+        finding.severity = self.get_severity(node.get('severity'))
+        finding.numerical_severity = Finding.get_numerical_severity(
+            finding.severity)
```
damiencarol, line 78:
```suggestion
```
Not required anymore, upper layers take care of this one.

dojo/unittests/tools/test_securecodebox_parser.py
```diff
@@ -0,0 +1,82 @@
...
+        testfile.close()
+        self.assertEqual(5, len(findings))
+        self.assertEqual(findings[0].severity, "Info")
+        self.assertEqual(findings[0].title, "ssh")
+        self.assertEqual(findings[0].description,
+                         "Port 22 is open using tcp protocol.")
+        self.assertEqual(
+            findings[0].unsaved_endpoints[0].host, "scanme.nmap.org")
+        self.assertEqual(findings[0].unsaved_endpoints[0].protocol, "tcp")
+        self.assertEqual(findings[0].unsaved_endpoints[0].path, "")
```
damiencarol, line 47: Maybe not the best option here. Better to have `path` at `None` as this attribute is null-able.

dojo/unittests/scans/securecodebox/scb_one_vul.json
```diff
@@ -0,0 +1,25 @@
...
+      "protocol": "tcp",
+      "hostname": "scanme.nmap.org",
+      "method": "table",
+      "operating_system": null,
+      "service": "ssh",
+      "serviceProduct": null,
+      "serviceVersion": null,
+      "scripts": null
+    },
+    "id": "473454f7-14b4-4014-8c2b-d0325b55234e"
```
damiencarol, line 23: it seems that this data is not read by the parser and is a good candidate for the attribute `unique_id_from_tool`

dojo/tools/securecodebox/parser.py
```diff
@@ -0,0 +1,89 @@
+import hashlib
+import json
+from urllib.parse import urlparse
```
damiencarol, line 3: We want every parser to use `hyperlink` package. We need to rewrite the part that manage endpoints to leverage `hyperlink.parse()` instead

damiencarol (changes requested):
Made a lot of comments, one is blocking (urlparse/hyperlink)

dojo/fixtures/test_type.json
JohannesZahn: Done

dojo/tools/securecodebox/parser.py
JohannesZahn: Done

dojo/unittests/scans/securecodebox/scb_one_vul.json
madchap: is that `id` idempotent across runs and unique across whatever scans could be submitted by SCB?

JohannesZahn: Done

dojo/unittests/tools/test_securecodebox_parser.py
JohannesZahn: path, query and fragment are no longer some default values but None if they are empty

dojo/tools/securecodebox/parser.py
JohannesZahn: Done

dojo/fixtures/test_type.json
JohannesZahn: How is this going to be handled in the future?
Will the test types be generated automatically?

damiencarol: yes.
this part is dynamic since few release now. better to let it die

JohannesZahn: I could not spot a SecureCodeBox Test Type when using HTTP GET in the swagger docs. Any advice?

dojo/tools/securecodebox/parser.py
```diff
@@ -0,0 +1,77 @@
...
+            finding = Finding(
+                static_finding=True, dynamic_finding=False, active=False, verified=False)
+            self.set_finding_base_info(finding, node, test)
+            finding_unique_string = str(node['location'] +
+                                        finding.description + finding.title)
+            dupe_key = hashlib.md5(
+                finding_unique_string.encode('utf-8')).hexdigest()
+            if dupe_key not in dupes:
+                self.set_finding_endpoint(finding, node)
+                dupes[dupe_key] = finding
```
damiencarol, line 34: I thing you don't need this part. The SCB format have an ID. we should use this ID directly as the key.

dojo/fixtures/test_type.json
JohannesZahn: @damiencarol

JohannesZahn: I assumed to find it here. 

image

dojo/unittests/tools/test_securecodebox_parser.py
```diff
@@ -47,6 +47,9 @@ def test_scb_parser_with_many_vuln_has_many_findings(self):
         self.assertEqual(findings[0].unsaved_endpoints[0].query, None)
         self.assertEqual(findings[0].unsaved_endpoints[0].fragment, None)
         self.assertEqual(findings[0].unsaved_endpoints[0].port, 22)
+        self.assertEqual(findings[5].unsaved_endpoints[0].path, "/robots.txt")
+        self.assertEqual(findings[5].unsaved_endpoints[0].fragment, "test")
+        self.assertEqual(findings[5].unsaved_endpoints[0].query, None)
```
damiencarol, line 52: FYI: this form is also possible
```suggestion
        self.assertIsNone(findings[5].unsaved_endpoints[0].query)
```

JohannesZahn:
I am putting this on hold since we are currently experimenting with your new generic json parser

JohannesZahn:
We will probably not need this as we are now using the new generic JSON parser. So I will close the PR for now.

PR #4434, opened 2021-05-03
[CyloneDX parser] remove unused attributes and switch unique id to vuln id

damiencarol:
## Work
- [x] improve description for vulnerabilities (better management when duplicates, use text when existing)
- [x] remove useless setting of `numerical_severity` and `cvssv3_score` (calculated in upper layers)
- [x] switch `unique_id_from_tool` to `vuln_id_from_tool` (per discussions with the team)
- [x] add more test data (last version of Anchore Grype)
- [x] refactor the internal de-duplication to remove code duplicates
- [x] switch parser id from `cyclonedx` to `CycloneDX Scan`
- [x] migration DB for the switch of some attributes

image


mtesauro (approved):
Approved

damiencarol:
@mtesauro added the DB migration part and rebased. Ready for review.

valentijnscholten:
Was there a reason to rename the `name` in `Test_Type` other than cosmetic reasons? I think in the UI/API the label is used?

damiencarol:
Yes it's cosmetic reasons. to be aligned with other parsers (we made the same move for few others).
And yes this label is used in UI/Api

valentijnscholten:
Ideally the `name` is or should not be used in the UI/API to avoid all these migrations when we rename them.

valentijnscholten (approved):

damiencarol:
I agree, it's the final plan. I mean the UI/API don't rely on what's in the DB but only what the Python code load with the dynamic loading.
Do you want to implement your idea of a static list of ID<->parser ?

PR #4435, opened 2021-05-03
Add cvss score to sonatype parser

scirusvulgaris:
Sonatype is not providing detailed cvss v3 report, this cause cvssv3_score to be null, at the same time a pre calculate score is present in the vulnerability report. To remediate cvssv3_score have to be added in the finding object.

### DIFF
--- a/dojo/tools/sonatype/parser.py
+++ b/dojo/tools/sonatype/parser.py
@@ -109,7 +109,7 @@ def get_item(vulnerability, test):
         finding = Finding(
             title=finding_title,
             cve=cve,
-            cvssv3_score = score,
+            cvssv3_score=score,
             test=test,
             severity=severity,
             description=finding_description,
--- a/dojo/tools/sonatype/parser.py
+++ b/dojo/tools/sonatype/parser.py
@@ -109,6 +109,7 @@ def get_item(vulnerability, test):
         finding = Finding(
             title=finding_title,
             cve=cve,
+            cvssv3_score = score,
             test=test,
             severity=severity,
             description=finding_description,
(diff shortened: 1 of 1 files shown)


valentijnscholten (changes requested):
there seems to be too many changes/commits in the PR, can you rebase?

Also if I read it correctly, if there is no severity, the cvss3_score field will become `No CVSS score yet.`? I don't think that's desired behaviour.

damiencarol (changes requested):
Could you add unit tests that show the new data is here? Also you need to remove the modifications on `.github/workflows/test-helm-chart.yml`

scirusvulgaris:
> there seems to be too many changes/commits in the PR, can you rebase?
> 
> Also if I read it correctly, if there is no severity, the cvss3_score field will become `No CVSS score yet.`? I don't think that's desired behavior.

I'm not sure if we have same output, I can send you a Sonatype report if you would like to test it on your side.
Without cvssv3_score there's no score at all, either sonatype changes their vulnerability report, by adding cvssv3 either we remediate.

madchap:
> > there seems to be too many changes/commits in the PR, can you rebase?

Still needs to be done.

> > Also if I read it correctly, if there is no severity, the cvss3_score field will become `No CVSS score yet.`? I don't think that's desired behavior.
> 
> I'm not sure if we have same output, I can send you a Sonatype report if you would like to test it on your side.
> Without cvssv3_score there's no score at all, either sonatype changes their vulnerability report, by adding cvssv3 either we remediate.

Per https://github.com/DefectDojo/django-DefectDojo/blob/10e83ecf44e37e0c565880748abaa61645acfe9e/dojo/tools/sonatype/parser.py#L102 it would try to assign "No CVSS score yet" to https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/models.py#L1453 which is a float. You're probably not seeing anything, because there is a type mismatch.

valentijnscholten (changes requested):
@scirusvulgaris can you update the unit tests so it checks for the cvss to be populated correctly (or set to 0.0)?

PR #4437, opened 2021-05-03
Clean test_type fixture

damiencarol:


mtesauro:
My assumption is that this fixture data is no longer needed due to the dynamic parser loading changes you did recently, correct?

mtesauro (approved):
Approving assuming my comment ^ is accurate.  @damiencarol  let me know if that's wrong when you get  a chance.

mtesauro:
Interesting, looks like this is breaking some integration tests that are looking for specific test_types to be present  e.g.

> selenium.common.exceptions.NoSuchElementException: Message: Could not locate element with visible text: Bandit Scan

damiencarol:
@mtesauro yep. there is another PR that try to do the same thing by moving the load/save earlier in the loading and remove the same data as in this PR. we will see which one is finished first.

PR #4446, opened 2021-05-04
Fix edit finding view to unset duplicate status of a finding

### LINKED ISSUE #4338: Editing a finding to remove duplicate status has no effect
**Bug description**
Cannot manually remove duplicate status from a finding.

**Steps to reproduce**
On a finding marked as "Duplicate", edit it and uncheck the duplicate box. 
Save.
The status remains as Duplicate.

**Expected behavior**
The status changes.

- [ ] Kubernetes
- [x] Docker

**Environment information**
 dev

madchap:
Fixes #4338

### DIFF
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -807,8 +807,7 @@ def edit_finding(request, fid):
             # if we're removing the "duplicate" in the edit finding screen
             # do not relaunch deduplication, otherwise, it's never taken into account
             if old_finding.duplicate and not new_finding.duplicate:
-                reset_finding_duplicate_status_internal(request.user, new_finding.id)
-                new_finding.refresh_from_db()
+                new_finding.duplicate_finding = None
                 new_finding.save(push_to_jira=push_to_jira, dedupe_option=False)
             else:
                 new_finding.save(push_to_jira=push_to_jira)
@@ -2154,7 +2153,6 @@ def reset_finding_duplicate_status_internal(user, duplicate_id):
     if duplicate.duplicate_finding:
         # duplicate.duplicate_finding.original_finding.remove(duplicate)  # shouldn't be needed
         duplicate.duplicate_finding = None
-        duplicate.duplicate_finding_id = None
     duplicate.last_reviewed = timezone.now()
     duplicate.last_reviewed_by = user
     duplicate.save(dedupe_option=False)
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -804,6 +804,8 @@ def edit_finding(request, fid):
             # any existing finding should be updated
             push_to_jira = push_to_jira and not push_group_to_jira and not new_finding.has_jira_issue
 
+            # if we're removing the "duplicate" in the edit finding screen
+            # do not relaunch deduplication, otherwise, it's never taken into account
             if old_finding.duplicate and not new_finding.duplicate:
                 reset_finding_duplicate_status_internal(request.user, new_finding.id)
                 new_finding.refresh_from_db()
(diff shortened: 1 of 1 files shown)


dojo/finding/views.py
```diff
@@ -804,6 +804,10 @@ def edit_finding(request, fid):
             # any existing finding should be updated
             push_to_jira = push_to_jira and not push_group_to_jira and not new_finding.has_jira_issue
 
+            if old_finding.duplicate and not new_finding.duplicate:
+                reset_finding_duplicate_status_internal(request.user, new_finding.id)
+                new_finding.refresh_from_db()
+
             new_finding.save(push_to_jira=push_to_jira)
```
valentijnscholten, line 811: The behaviour I think is caused by the dedupe running on save. So if you do NOT want dedupe to run on save, the only change needed should be to supply `dedupe_option=False` to save() here. But maybe that's not what everyone wants, so  there could be a "Run deduplication" checkbox on screen, similar to what we have for "push to JIRA".

If we don't want that, and see this "mark no longer as duplicate" as a special case we should handle, I think we could suffice with adding `dedupe_option=False` for that scenario only?

madchap: Yes, specifying the `dedupe_option=False` makes it work as expected.

For these manual overrides, we should have a way to lock them, so that future occurrences of deduplication (e.g. running `manage.py dedupe`) do not go in a revert the state to whatever it was.

dojo/finding/views.py
```diff
@@ -2147,6 +2154,7 @@ def reset_finding_duplicate_status_internal(user, duplicate_id):
     if duplicate.duplicate_finding:
         # duplicate.duplicate_finding.original_finding.remove(duplicate)  # shouldn't be needed
         duplicate.duplicate_finding = None
+        duplicate.duplicate_finding_id = None
```
valentijnscholten, line 2157: not needed

dojo/finding/views.py
```diff
@@ -804,7 +804,14 @@ def edit_finding(request, fid):
             # any existing finding should be updated
             push_to_jira = push_to_jira and not push_group_to_jira and not new_finding.has_jira_issue
 
-            new_finding.save(push_to_jira=push_to_jira)
+            # if we're removing the "duplicate" in the edit finding screen
+            # do not relaunch deduplication, otherwise, it's never taken into account
+            if old_finding.duplicate and not new_finding.duplicate:
+                reset_finding_duplicate_status_internal(request.user, new_finding.id)
```
valentijnscholten, line 810: I wouldn't use this function here as it leads to weird code with refresh from db. Better just set duplicate=False, duplicat_finding=None, active=True and save without dedupe

dojo/finding/views.py
madchap: Still kinda dirty when it pops up in ptvsd mode.

dojo/finding/views.py
madchap: OK.

I don't think I want to even force it to `Active`, but I don't think we have anything better right now. We should somehow be able to tell the last status prior to dedupe (e.g. what status it was imported in) and reset to that.

valentijnscholten:
setting active to True is what was happening reset_finding_duplicate_status, but in hindsight, the user is editing the finding and all status checkboxes are on screen, so we should just use whatever the user has chosen on screen.

madchap:
> setting active to True is what was happening reset_finding_duplicate_status, but in hindsight, the user is editing the finding and all status checkboxes are on screen, so we should just use whatever the user has chosen on screen.

Works well like that, and is the obvious choice.

valentijnscholten (approved):

mtesauro (approved):
Approved

PR #4468, opened 2021-05-07
Documentation Update

StefanFl:
With this PR the documentation gets a much better flow. You can see how it will look like at https://stefanfl.github.io/django-DefectDojo/.

Apart from some reorganisation and lots of details there are some noteable changes:
- [About DefectDojo](https://defectdojo.github.io/django-DefectDojo/basics/about/) and [Models](https://defectdojo.github.io/django-DefectDojo/basics/models/) have been merged. They were very similar. The new page is [Core Data Classes](https://stefanfl.github.io/django-DefectDojo/usage/models/)
- All the attribute descriptions in [Features](https://defectdojo.github.io/django-DefectDojo/basics/features/) have been removed. I think the field names are self-explaining.
- I didn't see the value of [Usage examples](https://defectdojo.github.io/django-DefectDojo/basics/start-using/), so it has been removed. The section about Risk Acceptance has been moved to the Features.
- Please review the new [Architecture](https://stefanfl.github.io/django-DefectDojo/getting_started/architecture/) page.

I am not completely happy about the [Features](https://stefanfl.github.io/django-DefectDojo/usage/features/) page. DefectDojo has a lot more features, that are not mentioned here. And is everything still up-to-date? Maybe we find volunteers for that.


docs/content/en/getting_started/architecture.md
```diff
@@ -0,0 +1,47 @@
...
+## uWSGI
+
+[uWSGI](https://uwsgi-docs.readthedocs.io/en/latest/) is the application server
+that runs the DefectDojo application, written in Python/Django, to serve all
+dynamic content.
+
+## RabbitMQ
+
+The application server sends tasks to a [RabbitMQ](https://www.rabbitmq.com/)
+message queue for asynchronous execution.
```
damiencarol, line 24: @StefanFl the architecture page is not accurate. DefectDojo is **not strongly linked** to RabbitMQ.
DefectDojo rely on Celery. Celery can leverage RabbitMQ for message passing but this can be replaced by Redis or another system.
I would prefer to have something more high level in this page. If you really want to talk about implementation, you can redirect users to https://docs.celeryproject.org/en/stable/getting-started/brokers/index.html#broker-overview

damiencarol (changes requested):
I'm not 100% in all the modifications. many changes are good but removing features could be discussed

damiencarol (approved):

damiencarol:
@StefanFl made some comment but accepting as documentation can be improved/modified all the time.

docs/content/en/getting_started/architecture.md
madchap: I agree with @damiencarol, we should have the term `broker` maybe instead, and list rabbitMQ as being possibly the default for us.

madchap (approved):
Made some small comments, compared side-by-side, lgtm :) Thanks for taking care of this @StefanFl - let us know when you're ready to merge, and maybe other mods will want to look as well.

docs/content/en/getting_started/architecture.md
StefanFl: Yes, having a Message Broker in the diagram, with a reference to RabbitMQ and Redis in the text is much better.

And can do some double-edge arrows as well.

docs/content/en/usage/models.md
```diff
@@ -0,0 +1,77 @@
+---
+title: "Core data classes"
+description: "DefectDojo is based on a model that allows the ultimate flexibility in your test tracking needs."
```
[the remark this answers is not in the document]
StefanFl: Copied it from somewhere but maybe it's really a bit much. Will reword it.

damiencarol (approved):

StefanFl:
The last commit fixed the problem with the wrong link, @valentijnscholten pointed it out on Slack. Unfortunately we now get some links on the top right for editing the page and submitting issues. See https://stefanfl.github.io/django-DefectDojo/integrations/import/ for an example. The other option would have been no links to GitHub but a last modified date of 1.1.0000 or so.

For me this PR would be ready to merge

valentijnscholten:
Thanks again.

PR #4472, opened 2021-05-08
Authorization V2: Fixes for role name in API and permissions for role API_Importer

### LINKED ISSUE #4384: Extend reader rights to API importer role

The new `api_importer` role allows an account, scoped within a product or product type, to import/re-import scans. According to the Object Based Authorization Model, by design the API importer is not able to see (read) the engagement(s), test(s) or product(s) to which it is associated. Its purpose is strictly to import or re-import a scan.

This presents a problem, however. When I assign the `api_importer` role to an account to interact with DefectDojo via API in an automated way, it is not possible to determine whether it should import or re-import a scan (e.g., by evaluating if a test ID already exists for a particular engagement and what is the scan date of the original scan). It is quite limiting.

This situation has been briefly  discussed [here](https://github.com/DefectDojo/django-DefectDojo/issues/4347).

As a DevSecOps engineer, I want to use the `api_importer` role to programatically import/re-import scans into DefectDojo via API without having to grant unnecessary privileges, such as create/delete products etc., but that still leaves room for more sophisticated programmatic access to scan related data, such as scan dates, engagement and test IDs etc.

This can be achieved by extending the `reader` rights to the `api_importer` role - thus we'd be able to programatically evaluate and apply control flows for scripted interactions with the DefectDojo API.

The alternative solution is to use other roles such as `writer`, `owner` or `maintainer`, which grant more privileges than required.

While this approach works, I wouldn't want to distribute tokens for automated interactions with the API which grant more privileges than needed to import/re-import scans.

### LINKED ISSUE #4453: AuthV2: Missing Role names in API
API supports to see Member:Product and Member:ProductType relations including Role ID. However, it is not possible to show the list of Role names.

Bonus:
Prefetch roles

API call `GET /roles/` which will return something like:
```json
{
...
  "results": [
    {
      "id": 0,
      "name": "Reader"
    },
...
  ],
...
}
```

As I know, there is not alternative

StefanFl:
- [x] Role name is shown in API for product_member and product_type_member, fixes #4453
- [x] Role API_Importer has read permission for objects, fixes #4384

### DIFF
--- a/docs/content/en/usage/permissions.md
+++ b/docs/content/en/usage/permissions.md
@@ -24,7 +24,7 @@ Users can be assigned as members to Products and Product Types, giving them one 
 |                             | Reader | Writer | Maintainer | Owner | API Importer |
 |-----------------------------|:------:|:------:|:----------:|:-----:|:------------:|
 | Add Product Type <sup>1)</sup> |     |        |            |       |              |
-| View Product Type           | x      | x      | x          | x     |              |
+| View Product Type           | x      | x      | x          | x     | x            |
 | Remove yourself as a member | x      | x      | x          | x     |              |
 | Manage Product Type members |        |        | x          | x     |              |
 | Edit Product Type           |        |        | x          | x     |              |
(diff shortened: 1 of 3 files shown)


damiencarol (approved):

valentijnscholten:
Still thinking if we need even more permissions for api importer, for example I am editing the engagement after import to add some info / description. Same for test. Or maybe people should just use the writer role in that case.

StefanFl:
@valentijnscholten Yes, I think then it is the Writer role. Maybe we need configurable roles soon.

Thanks for merging.

PR #4473, opened 2021-05-08
Add full support for using ports and userinfo in endpoints v2

kiblik:
Recreate  #4188 because of https://github.com/DefectDojo/django-DefectDojo/pull/4188#issuecomment-834279255

Hello.

based on [my analysis](https://owasp.slack.com/archives/C2P5BA8MN/p1616088028019100), I prepared redesign of using endpoints - it fully allows the use of ports and userinfos.

Changes:
- Every time, when an endpoint is copied, we copy also `userinfo`, `fqdn` and `port`
- Fix and use `Endpoint.from_uri()` - I remove many different implementations of parsing URLs; now, we are using only [hyperlink](https://pypi.org/project/hyperlink/), based on @damiencarol's [recommendation](https://owasp.slack.com/archives/C2P5BA8MN/p1616182871040200?thread_ts=1616088028.019200&cid=C2P5BA8MN)
- API
  - for creation (POST), require only `host` and `product`; others are optional
  - for update (PATCH), do not ask for `product` (if you don't want to update it)
- Add `userinfo` as `Endpoint` field - following [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.2.1)
- Validate `Endpoint` fields in the creation process
- `Endpoint.__str__()` create full URL (if is it possible)
- Rewrite (and fix) all parsers which use endpoints

TODO:
- [x] Write tests
   - [x] migration
   - [x] model
   - [x] api
- [x] Split views (and reports) to "hosts" (e.g. all vulnerabilities on specific host [ fqdn / ip ] = from all endpoints on host) and "endpoints" (e.g. all vulnerabilities on specific endpoint [ url ])
- [x] Double-check all forms and views
   - [x] ascii
   - [x] html
   - [x] pdf
- [x] we should forbid the update of product from API - Finding in product1 and Endpoint in product2? No way
- [x] redesign all `Endpoint.objects.get_or_create` and `Endpoint.objects.filter`
- [x] Form validations


dojo/endpoint/views.py
```diff
@@ -27,78 +27,56 @@
...
-        endpoints = EndpointFilter(request.GET, queryset=endpoints.filter(id__in=ids), user=request.user)
-        paged_endpoints = get_page_items(request, endpoints.qs, 25)
-    add_breadcrumb(title="All Endpoints", top_level=not len(request.GET), request=request)
-
-    product_tab = None
-    view_name = "All Endpoints"
-    if product:
-        view_name = "Endpoints"
-        product_tab = Product_Tab(product.id, "Endpoints", tab="endpoints")
+                else:
```
StefanFl, line 69: Was the change of indentation of lines 69 and 70 on purpose? The else in line 69 must relate to if not settings.FEATURE_AUTHORIZATION_V2: Otherwise the permission check for AuthV2 doesn't work. Please revert the change in line 69 and 70.

damiencarol: some conflict resolution that went wrong?

kiblik: @StefanFl, you are right. It wasn't the purpose

dojo/endpoint/utils.py
```diff
@@ -127,89 +138,101 @@ def clean_hosts_run(apps, change):
...
-
-                    if ep.count() > 1:
-                        ep_ids = [x.id for x in ep]
-                        logger.info("Merging Endpoints {} into '{}'".format(
+                        message = '"{}" uses invalid format of host'.format(endpoint.host)
+                        html_log.append({**endpoint_html_log, **{'message': message}})
+                        logger.error('Endpoint (id={}) {}. {}'.format(endpoint.pk, message, error_suffix))
+                        broken_endpoints.add(endpoint.pk)
+
+    if broken_endpoints != set():
```
valentijnscholten, line 205: is this reliable? I think it could just be `if broken_endpoints:`

dojo/models.py
```diff
@@ -1086,46 +1084,73 @@ class Endpoint(models.Model):
...
 
     class Meta:
-        ordering = ['product', 'protocol', 'host', 'path', 'query', 'fragment']
+        ordering = ['product', 'host', 'protocol', 'port', 'userinfo', 'path', 'query', 'fragment']
         indexes = [
             models.Index(fields=['product', 'mitigated']),
         ]
 
+    def __init__(self, *args, **kwargs):
+        kwargs = self.verify_and_clean(**kwargs)
```
valentijnscholten, line 1093: when is __init__ executed, only for new model instances? Or also when an instance is loaded from the DB?

dojo/endpoint/utils.py
kiblik: Python says, `if broken_endpoints:` should be enought.
```python
>>> bool(set())
False
```

dojo/models.py
kiblik: It is used when an instance is imported (`python manage.py loaddata ...`), it is created (`e = Endpoint()`) but I don't know how does it work in case of loading from DB.

kiblik: But DD didn't fail if DB store a "broken" endpoint. So it looks, `verify_and_clean` is not called during loading from DB

dojo/endpoint/utils.py
damiencarol: I agree it works but it's not the best code to read. :/ should we use the form `len(broken_endpoints)` ? better for readability.

valentijnscholten:
```
>>> empty_valentijn = set()
>>> if empty_valentijn:
...     print('not empty')
... else:
...     print('empty')
...
empty
```

```
>>> valentijn = set()
>>> valentijn.add('test')
>>> if valentijn:
...     print('not empty')
...
not empty
```

valentijnscholten:
updated docs, but it has conflicts and I cannot push from CLI due to github permissions.

dojo/models.py
valentijnscholten: not sure about this one. pretty sure some people will run into problems loading fixtures with broken endpoints, but maybe it's OK to force them to fix those in the fixture?

dojo/endpoint/utils.py
```diff
@@ -0,0 +1,264 @@
...
+                        if change and (endpoint.pk not in broken_endpoints):  # do not save broken endpoints
+                            endpoint.save()
+
+                    except ValidationError:
+                        message = '"{}" uses invalid format of host'.format(endpoint.host)
+                        html_log.append({**endpoint_html_log, **{'message': message}})
+                        logger.error('Endpoint (id={}) {}. {}'.format(endpoint.pk, message, error_suffix))
+                        broken_endpoints.add(endpoint.pk)
+
+    if broken_endpoints != set():
```
valentijnscholten, line 205:
```suggestion
    if broken_endpoints:
```

kiblik:
based on discussion with @valentijnscholten, this PR will be rewritten: no validations during creation, edit, import, reimport, migration - we will be able to store also "broken" endpoints

valentijnscholten:
This is what I said on (private channel) Slack:

>There's a bit of a contradiction with this PR I think. We do lots of validation, but then we allow invalid endpoints to remain. So even with invalid endpoints, nothing breaks? Users can keep using Defect Dojo. So what if, for now in this PR, we mark the endpoint as invalid, but continue the import. People have the endpoint validation pages to check / fix the endpoints.

>After that we need to think in general how to handle errors during parsing. Currently any error breaks the import, leaving an empty test, or partially imported test. Maybe we should have a flag skip_errors or something that could dictate if the import should fail alltogether, or if it should just skip findings with errors.

>It would mean that I think we should validate all data before creating any findings so that we don't have partially / half imported scans

>Also we would need a construct to feedback warning/errors to the UI or the API

valentijnscholten:
So I suggest to focus on making these changes and merge this PR so that at least we have the biggest changes (77 files) available in dev. We can do a follow up PR once we know what to do with validation / error / warning handling

dojo/endpoint/utils.py
```diff
@@ -0,0 +1,255 @@
...
+        logger.error('Endpoint (id={}) {}. {}'.format(endpoint.pk, message, error_suffix))
+        broken_endpoints.add(endpoint.pk)
+    html_log = []
+    broken_endpoints = set()
+    Endpoint_model = apps.get_model('dojo', 'Endpoint')
+    Endpoint_Status_model = apps.get_model('dojo', 'Endpoint_Status')
+    Product_model = apps.get_model('dojo', 'Product')
+    # when run on older versions, we haven't had the migration applied which has userinfo field added
+    # so defer that field and order by id (default ordering references userinfo field)
+    for endpoint in Endpoint_model.objects.defer('userinfo').order_by('id'):
```
valentijnscholten, line 101: the defer userinfo is no longer needed if we are not backporting the script to 1.15.x

valentijnscholten (changes requested):

dojo/endpoint/utils.py
kiblik: done

valentijnscholten:
Will the validation (`clean`) also be triggered when someone creates an endpoint via `/api/v2/endpoints` ?

dojo/unittests/tools/test_acunetix_parser.py
```diff
@@ -27,10 +27,11 @@ def test_parse_file_with_one_finding(self):
             # check endpoints
             self.assertEqual(1, len(finding.unsaved_endpoints))
             endpoint = finding.unsaved_endpoints[0]
+            endpoint.clean()
```
damiencarol, line 30:
```suggestion
            self.assertFalse(endpoint.is_broken())
```

damiencarol: Could you make this modification in all unit tests?

kiblik:
@valentijnscholten 
> Will the validation (`clean`) also be triggered when someone creates an endpoint via `/api/v2/endpoints` ?

Yes, as you can see [here](https://github.com/DefectDojo/django-DefectDojo/pull/4473/files#diff-c38b7dcb401e57c318d54f7b0717414e8231cb6fbb540109e8f1cbf1965ee4cdR24-R28)

dojo/unittests/tools/test_acunetix_parser.py
kiblik: I'm not sure about removing `.clean()` because tests checking format of endpoints which will be stored (and they will be stored after calling [`.clean()`](https://github.com/DefectDojo/django-DefectDojo/pull/4473/files#diff-fe3c6e77f360d3f1156c8e78b208ac1608ccf54484123b42316731eb9122cecdR117))

But definitely I can add checks for `.is_broken()`

kiblik: We have a problem - `.is_broken` also checks if `product` is defined (because of [this](https://owasp.slack.com/archives/C014H3ZV9U6/p1621955497009900)) and endpoints coming from parser don't have defined products.

However, [`.is_broken == .clean() + check if product is defined`](https://github.com/DefectDojo/django-DefectDojo/pull/4473/files#diff-a797c37dd5bcb8a76fcf2ce34eb254ab52bbcbdea329b6f94080f96eeb6727feR1240), so calling `.clean()` without fail should be enough

I will revert last commits

valentijnscholten (approved):

damiencarol (approved):

PR #4475, opened 2021-05-09
Checkboxes to disable checklists, questionnaires and credentials

StefanFl:
Checklists, questionnaires and credentials can be disabled with checkboxes in the System Settings. This makes the user interface less crowded, if these features are not used. Default is still enabled, so nothing changes for the users, until the settings are actively changed.

I sneaked in a change of a documentation URL for product members and product type members, to avoid a separate small PR.


valentijnscholten:
My first reaction would be  "shouldn't we just call them enable checklists, enable questionaires, etc.?" Enabled by default.

StefanFl:
You are right @valentijnscholten, would be much better. I will change later in the afternoon.

StefanFl:
All three checkboxes are named "Enable ..." now and are checked by default.

valentijnscholten (approved):

valentijnscholten:
merging as is to avoid having to maintain the migration while everyone is on holiday

PR #4479, opened 2021-05-10
New parser Security code scan (DotNet Security Scanner) 

SPoint42:
Security Code Scan - static code analyzer for .NET 

https://security-code-scan.github.io/


dojo/tools/security_code_scan/parser.py
```diff
@@ -0,0 +1,23 @@
+import json
+
+from dojo.models import Finding
+
+# Import Result of CLI from SecurityCode Scan
+# https://github.com/security-code-scan/security-code-scan
+# We need to pass --cwe to the scanner for the data to feed CWE in the report
```
damiencarol, line 7: Please switch that to a docstring.

SPoint42:
After some Exchanges with @damiencarol and tests on m’y codebase, it seems this parler is not useful as we can have SARIF export. So I close this PR and we can use SARIF format to import SecurityCodeScan Findings

Thanks @damiencarol  for the HEADS Up !

PR #4494, opened 2021-05-12
Fix gitleaks parser report with redacted feature enabled

ansidorov:
If you import a JSON gitleaks report into DefectDojo that was run with the `--redacted` flag, only one finding will be created in DefectDojo.

In that case, the gitleaks report will have this information:
```
 {
  "line": "            \ "unique_id\": \ "REDACTEDS\",
  "lineNumber": 1089,
  "offender": "REDACTED",
  "commit": "fd35c04adafe3ae8ff7c78951440ca95f14b0d48",
  "repo": "."
  "repoURL": "",
  "leakURL": "",
  "rule": "AWS Access Key",
  "commitMessage": "up stage env\n",
  "author": "Some User",
  "email": "you@example.com",
  "file": "terraform/terraform.tfstate.backup",
  "date": "2020-08-27T14:13:43+05:00",
  "tags": "key, AWS".
 }, 
```

In this Pull Request I have fixed this problem and added an additional test.

Using this flag can be useful when we don't want to store sensetive data in DefectDojo and show it to all users who have access, we just want to tell developers where the problem is.


valentijnscholten:
Thanks, could you base your branch and PR against `dev`?

ansidorov:
> Thanks, could you base your branch and PR against `dev`?

Done

valentijnscholten (approved):

devGregA (approved):

damiencarol (approved):

PR #4501, opened 2021-05-13
Authorization V2: Switch to Dojo_User

StefanFl:
Using `Dojo_User` instead of `User` is the better option to make use of additional methods like get_full_name().

Thanks @valentijnscholten for making me aware of this.


valentijnscholten:
merging straightaway to avoid having to maintain the migration

PR #4502, opened 2021-05-13
Authorization V2: Management command to migrate users and set Auth V2 as the default

StefanFl:
The management command `migration_authorization_v2` creates Product_Members and Product_Type_Members:

- Staff users currently have all permissions for all product types and products --> Every staff users gets the Owner role for all product types.
- Authorized users for product types or products types currently are non-staff users. They get a role for the respective product type or product based on the settings parameter `AUTHORIZED_USERS_ALLOW_STAFF`:
    - If that parameter is `False`, authorized users can only read the product type / product and they will get the role Reader.
    - If that parameter is `True`, they have all permissions and will get the role Owner.

The settings parameters `AUTHORIZED_USERS_ALLOW_CHANGE` and `AUTHORIZED_USERS_ALLOW_DELETE` give authorized users currently permissions that are quite different from the defined roles. So these parameters are not used to define roles.

### DIFF
--- a/dojo/management/commands/migrate_authorization_v2.py
+++ b/dojo/management/commands/migrate_authorization_v2.py
@@ -1,7 +1,7 @@
 import logging
 from django.conf import settings
 from django.core.management.base import BaseCommand
-from dojo.models import Dojo_User, Product, Product_Member, Product_Type, Product_Type_Member
+from dojo.models import Dojo_User, Global_Role, Role, Product, Product_Member, Product_Type, Product_Type_Member
 from dojo.authorization.roles_permissions import Roles
 
 
@@ -16,32 +16,36 @@ class Command(BaseCommand):
     """
     help = 'Usage: manage.py migration_authorization_v2'
 
-    def handle(self, *args, **options):
+    def __init__(self, **kwargs):
+        self.reader_role = Role.objects.get(id=Roles.Reader)
+        self.writer_role = Role.objects.get(id=Roles.Writer)
+        self.owner_role = Role.objects.get(id=Roles.Owner)
 
+    def handle(self, *args, **options):
         logger.info('Started migrating users for authorization v2 ...')
 
         authorized_user_exists = False
 
-        product_types = Product_Type.objects.all().prefetch_related('authorized_users')
+        # Staff users have had all permissions for all product types and products,
+        # so they will be get a global role as Owner.
+        # Superusers will have all permissions anyway, so they must not be set as members.
         staff_users = Dojo_User.objects.filter(is_staff=True, is_superuser=False)
-        for product_type in product_types:
-            # Staff users have had all permissions for all product types and products,
-            # so they will be set as owners for all product types.
-            # Superusers will have all permissions anyway, so they must not be set as members.
-            for staff_user in staff_users:
-                # If the product type member already exists, it won't be changed
-                if Product_Type_Member.objects.filter(product_type=product_type, user=staff_user).count() == 0:
-                    product_type_member = Product_Type_Member()
-                    product_type_member.product_type = product_type
-                    product_type_member.user = staff_user
-                    product_type_member.role = 4  # Owner
-                    product_type_member.save()
-                    logger.info('Product_Type_Member added: {} / {} / {}'.format(product_type.name, staff_user.username, Roles(product_type_member.role).name))
-                else:
-                    logger.info('Product_Type_Member already exists: {} / {}'.format(product_type.name, staff_user.username))
+        for staff_user in staff_users:
+            global_role = staff_user.global_role if hasattr(staff_user, 'global_role') else None
+            if global_role is None:
+                global_role = Global_Role()
+                global_role.user = staff_user
+            if global_role.role is None:
+                global_role.role = self.owner_role
+                global_role.save()
+                logger.info('Global_Role Owner added for staff user {}'.format(staff_user))
+            else:
+                logger.info('Staff user {} already has Global_Role {}'.format(staff_user, global_role.role))
 
-            # Authorized users for product types will be converted to product type members
-            # with a role according to the settings
+        # Authorized users for product types will be converted to product type members
+        # with a role according to the settings
+        product_types = Product_Type.objects.all().prefetch_related('authorized_users')
+        for product_type in product_types:
             for authorized_user in product_type.authorized_users.all():
                 # If the product type member already exists, it won't be changed
                 if Product_Type_Member.objects.filter(product_type=product_type, user=authorized_user).count() == 0:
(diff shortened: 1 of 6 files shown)


valentijnscholten:
> The management command `migration_authorization_v2` creates Product_Members and Product_Type_Members:
> 
> * Staff users currently have all permissions for all product types and products --> Every staff users gets the Owner role for all product types.

I think staff or superusers should always have owner permissions, even if not assigned explicitly. Or there should be a global Owner rule so users can choose.

> * Authorized users for product types or products types currently are non-staff users. They get a role for the respective product type or product based on the settings parameter `AUTHORIZED_USERS_ALLOW_STAFF`:
>   
>   * If that parameter is `False`, authorized users can only read the product type / product and they will get the role Reader.
>   * If that parameter is `True`, they have all permissions and will get the role Owner.
> 
> The settings parameters `AUTHORIZED_USERS_ALLOW_CHANGE` and `AUTHORIZED_USERS_ALLOW_DELETE` give authorized users currently permissions that are quite different from the defined roles. So these parameters are not used to define roles.

I think we need to map these users, otherwise lots of mappings will get lost. So either pick a default to map both to Writer. Or have some mandatory parameters to force the user to choose.

StefanFl:
**Staff and superusers -** That's already implemented in Authorization V2: 
- Superusers have Owner permissions implicitly, they can do everything on every object. So they don't need to be assigned to Products or Product_Types
- Staff users have no permissions per default and must be assigned as members. But there is the settings parameter `AUTHORIZATION_STAFF_OVERRIDE`, when it is set to `True`, staff members are treated like superusers and have Owner permissions on all objects implicitly as well.

**`AUTHORIZED_USERS_ALLOW_CHANGE`** and **`AUTHORIZED_USERS_ALLOW_DELETE`** - These parameters don't seem to make any sense anyway. I checked with release 1.12 before the changes for Authorization V2:
- `AUTHORIZED_USERS_ALLOW_CHANGE`: If `True`, authorized users see a menu entry to change the products itself, but when you click on it you get a 403/no permission. They have no permissions to change Engagements, Findings, Tests or start imports.
- `AUTHORIZED_USERS_ALLOW_DELETE`: Basically the same, authorized users have a menu entry "Delete" but get a 403 as well, no other permissions.

As far as I can see, these 2 parameters have no practical effect on permissions, so they can be ignored and authorized users get the Read role.

valentijnscholten:
`AUTHORIZED_USERS_ALLOW_CHANGE` and `AUTHORIZED_USERS_ALLOW_DELETE` give permission to edit/delete findings. And maybe some more, but the main focus was on findings. It was not implemented for Product or Product_Type I believe.
Are you saying this is no longer working? I think you put efforts in to keep the legacy authz working?

StefanFl:
Oh, I didnt't dig that deep. Editing and deleting findings still work with these 2 parameters. Still not completely sure how to deal with it. Giving these users the Writer role would give theses users many more permissions than they currently have. Maybe we need 2 or 3 further roles, to cover these cases.

valentijnscholten:
I think we should map them to Writer. More roles would make it more complex. These fine grained existing permission where just some shortcuts people introduced to quickly allow users to modify findings. With Writer I think they get only permissions to modify stuff inside the product, not the product itself?

StefanFl:
That's true, _Writers_ can only change things within the Product like Engagements, Test, Findings, ..., but not the Product itself. So a bit of a permissions creep when `AUTHORIZED_USERS_ALLOW_CHANGE` is set seems acceptable to me. 

For ` AUTHORIZED_USERS_ALLOW_DELETE` it's a bit weird though, as the permissions are completely different than before, because a _Writer_ cannot delete Findings.

Shall we make authorized users to _Writers_ when `AUTHORIZED_USERS_ALLOW_CHANGE` is set and _Readers_ when only ` AUTHORIZED_USERS_ALLOW_DELETE` is set?

valentijnscholten:
`AUTHORIZED_USERS_ALLOW_DELETE` is only used together with `AUTHORIZED_USERS_ALLOW_CHANGE` to give some extra permissions. Even though with the `Writer` role doesn't allow deletion, I think it's most logical role to map to.

If you want to make it perfect, just add some parameters where uses can chose the mapping :-)

StefanFl:
I think you're right, `AUTHORIZED_USERS_ALLOW_DELETE` on its own might be really rare. Now both parameters result in setting the _Writer_ role.

dojo/management/commands/migrate_authorization_v2.py
```diff
@@ -0,0 +1,88 @@
...
+                # If the product type member already exists, it won't be changed
+                if Product_Type_Member.objects.filter(product_type=product_type, user=staff_user).count() == 0:
+                    product_type_member = Product_Type_Member()
+                    product_type_member.product_type = product_type
+                    product_type_member.user = staff_user
+                    product_type_member.role = 4  # Owner
+                    product_type_member.save()
+                    logger.info('Product_Type_Member added: {} / {} / {}'.format(product_type.name, staff_user.username, Roles(product_type_member.role).name))
+                else:
+                    logger.info('Product_Type_Member already exists: {} / {}'.format(product_type.name, staff_user.username))
```
valentijnscholten, line 41: I don't think we should do this. I think we should have superuser as the global type of users that have all permissions. Or possibly a global owner role talked about. Staff users I think should be purely for sysadmin kinda users, or solely used to grant access to /admin as Django seems to have intended this flag.
Adding staff users as owners to existing product types, but not to new ones might be confusing.

valentijnscholten (changes requested):
as per my comment

dojo/management/commands/migrate_authorization_v2.py
StefanFl: Staff users can open the admin interface, but they don't see anything there and they can't do any configuration in DefectDojo.

2021-05-15 12_54_11-Site administration _ Django site admin

For me staff users are employee of my company and non-staff users are external users who need access to DefectDojo.

Superusers can do administration and have full access to the data. This doesn't change with the new authorization.

For staff users there is a change of behaviour with the new authorization. While they had full access to data until now, they will need to be authorized per Product_Type, Product or Group by default in the future. To get the old behaviour, the system parameter `AUTHORIZATION_STAFF_OVERRIDE` can be set to `True`, then all staff users will have full access to all data again. Or a staff user is given a global role (will come soon) and then she will have access to all data according to the role.

So what the behaviour for staff users is with the new authorization is highly configurable but outside of this PR. Here we give staff users the same permissions for the new authorization as they had before.

valentijnscholten: > Staff users can open the admin interface, but they don't see anything there and they can't do any configuration in DefectDojo.
> 
> 2021-05-15 12_54_11-Site administration _ Django site admin
> 
> For me staff users are employee of my company and non-staff users are external users who need access to DefectDojo.

I have never heard that one, don't think that's how it's supposed to work or designed.
 
> Superusers can do administration and have full access to the data. This doesn't change with the new authorization.
> 
> For staff users there is a change of behaviour with the new authorization. While they had full access to data until now, they will need to be authorized per Product_Type, Product or Group by default in the future. To get the old behaviour, the system parameter `AUTHORIZATION_STAFF_OVERRIDE` can be set to `True`, then all staff users will have full access to all data again. Or a staff user is given a global role (will come soon) and then she will have access to all data according to the role.

So with global roles this is no longer needed. To me it makes no sense to add staff users to each product type.

StefanFl: Yes, using a global role makes sense. So I convert the PR to draft status until the global roles are implemented.

StefanFl:
I am pretty sure, the failing unit tests are a side effect of #4649. With the new authorization we get the member classes in the API, which only have one-to-one relationships.

dojo/settings/settings.dist.py
```diff
@@ -175,7 +175,7 @@
 
     # Feature toggle for new authorization, which is incomplete at the moment.
     # Don't set it to True for productive environments!
-    DD_FEATURE_AUTHORIZATION_V2=(bool, False),
+    DD_FEATURE_AUTHORIZATION_V2=(bool, True),
```
valentijnscholten, line 178: prefer to set this to True with the PR that adds some API testcases :-D

StefanFl: I am currently working on API test cases, they will come this weekend. I would like to keep this PR here as it is and you merge it after the API tests. Otherwise the migration and the documentation would need to be changed as well.

valentijnscholten (approved):

PR #4511, opened 2021-05-14
Add: super admin to create/fetch tokens for all users

dsever:
With this PR, superadmin can create/fetch token for all DD users. This PR solves problem when users are actually machines consumed in CICD manner, and process of user creation needs to be automated by using API. 

Superadmin can specify optional  target_user attribute  that represents user, for whom it is fetching API key.

Added tests together with negative.


dsever:
I see currently, there is not option to rotate API keys at all, and it is sort of sec issue, maybe good approach would be to use create_or_update instead of  get_or_create. So each time when API key is requested, ti will create new one.

devGregA:
@dsever  Thank you for your contribution, but I don't think this is the right direction. For CI, I would recommend creating a service account. Being able to issue tokens for different users creates a number of audit issues. I can't support this change.

dsever:
@devGregA in DD there is no thing as ServiceAccount, and possibility to fetch api key for other users than itself, when socAuth is in use

mtesauro:
Maybe I'm in the same place as @devGregA but I also see this as an issue.  Here's a typical usage of DefectDojo + the API I've seen at several companies:

Install is DefectDojo + SSO with some identity store having a group whose members can log into the web UI

- Every human that logs in is provided their own API key associated wit their username in DefectDojo
- For API/programmatic interactions with DefectDojo, a local (not SSO aka local only to DefectDojo) user is added  and the API token for that non-human user provided to that integration and stored in Vault or similar.  I won't speak for devGregA but I'd consider this a type of "ServiceAccount" aka something for non-humans to auth-n with.
- Humans via their associated API key can automate their work via the API as well as other integrations/programs/non-humans via the token associated with their local DefectDojo (non-SSO) user

If you had the situation above ^, plus the change suggested in the PR, how do I know that an actual user did something vs an integration based on the ID associated with the API token?   If as you say above:

> superadmin can create/fetch token for all DD users

and 

> Superadmin can specify optional target_user attribute that represents user, for whom it is fetching API key

Than any super user could obtain any user's API token  and appear to be making calls against the API as that user.  Personally, for any DefectDojo install I had, I would want to be able to provably deny that I had obtained an API token of any user.

Let's say grumpy DefectDojo superuser Eve has a some sort of issue with Alice (personal grudge, worried they will take Eve's job, ...) so they use the proposed API call to obtain Alice's API token and proceeds to delete 1/2 of the product_types in their DefectDojo install.  Because of cascading deletes, 1/2 of the data in DefectDojo is now gone and it looks like (if you dig around in logs and such) that the API token for Alice was the one that called DELETE on product_type.

Alice has no way to defend herself in this situation.

This is just one negative case I can come up with if superusers can see other's API tokens - this is why root users in Linux/Unix cannot see passwords in /etc/shadow (they are encrypted).  Why would we want to take the API tokens of DefectDojo back to some like the bad old days of having plain-text passwords in /etc/passwd?

So, I'm not a fan of this PR.  At the very least, it should be behind a feature flag that defaults to having that behavior turned off so people have to knowingly turn on a potential foot cannon.  Additionally, I would strongly suggest that a super user can only **change/update** a token for a user so that user's cannot claim their API token was read (and used) by a superuser (plausible deniability).

I am 100% sympathetic to the goal to have everything API-ified - most of the work I do with DefectDojo is via the API.

However, I think a small one-time cost of manually setting up a non-human user for programmatic API access is a pretty low cost vs loosing the ability to have API tokens only available to the individual users.

Let me know if I'm not understanding your use case.

dsever:
Hi Matt, thanks for sharing with me your view


> Install is DefectDojo + SSO with some identity store having a group whose members can log into the web UI
> 
> * Every human that logs in is provided their own API key associated wit their username in DefectDojo

This is correct 

> * For API/programmatic interactions with DefectDojo, a local (not SSO aka local only to DefectDojo) user is added  and the API token for that non-human user provided to that integration and stored in Vault or similar.  I won't speak for devGregA but I'd consider this a type of "ServiceAccount" aka something for non-humans to auth-n with.

But here is the catch, we can do this only using GUI (in current implementation of DD),  and we want to use completely automated way for local user creation, and fetching the API key for given user.  Complete provisioning in our case is supported by additional portal (within the company we are planing to have DD as a service), where user will specify "MachineAccounts" and after  portal will create MachineAccounts it will return to user API keys for Machineaccounts (in encrypted form)

> 
> If you had the situation above ^, plus the change suggested in the PR, how do I know that an actual user did something vs an integration based on the ID associated with the API token? If as you say above:
> 
> > superadmin can create/fetch token for all DD users
> 
> and
> 
> > Superadmin can specify optional target_user attribute that represents user, for whom it is fetching API key
> 
> Than any super user could obtain any user's API token and appear to be making calls against the API as that user. Personally, for any DefectDojo install I had, I would want to be able to provably deny that I had obtained an API token of any user.

For the audit purposes this could be logged, for the beginning, 

> Let's say grumpy DefectDojo superuser Eve has a some sort of issue with Alice (personal grudge, worried they will take Eve's job, ...) so they use the proposed API call to obtain Alice's API token and proceeds to delete 1/2 of the product_types in their DefectDojo install. Because of cascading deletes, 1/2 of the data in DefectDojo is now gone and it looks like (if you dig around in logs and such) that the API token for Alice was the one that called DELETE on product_type.
> 

@mtesauro I hope you are agree, that usage of super user should b strongly  avoided, not to have more than 1 (even staff has to much rights)   and AUTH2 should be enforced. To have legacy authentication model, instead of AUTH2 is huge gap (sorry for not bringing this up, my PR makes sense only if AUTH2 is in use, and superadmin is here only to manage core of applications).

but @mtesauro if you have some bad super user remember mu PR is not breaking anything, just take look, superuser can fetch all tokens in GUI anyway:

```http://localhost:8080/admin/authtoken/tokenproxy/```


> Alice has no way to defend herself in this situation.

>
> So, I'm not a fan of this PR. At the very least, it should be behind a feature flag that defaults to having that behavior turned off so people have to knowingly turn on a potential foot cannon. Additionally, I would strongly suggest that a super user can only **change/update** a token for a user so that user's cannot claim their API token was read (and used) by a superuser (plausible deniability).
> 

Actually I see it at the same way, to have it as an option, and  implement is as create/update.


> I am 100% sympathetic to the goal to have everything API-ified - most of the work I do with DefectDojo is via the API.
> 
> However, I think a small one-time cost of manually setting up a non-human user for programmatic API access is a pretty low cost vs loosing the ability to have API tokens only available to the individual users.
> 
> Let me know if I'm not understanding your use case.

At the end, what can is to add new API  endpoint where admin can setup password for new created user.

dsever:
Even this doesn't open new risks in current implementation (because keys are already exposed to superusers using admin interlace, so risk is already there), after discussion with  @valentijnscholten maybe than would be cleaner approach to request token from superadmin during account creation, like in JSON specify token=True?

valentijnscholten:
I didn't see the comments by Matt and Greg yet, but also informed Dubravko that it's not OK to expose all users API tokens via the API. Even though this is now possible via `/admin` already, that doesn't mean we should extend that. Quite the opposite, the new authz framework by stefan will help us reduce the number of superusers needed. 
To me it's not 100% clear what workflow / steps you are envisioning here. Could you list a simple list of steps that your automation is trying to do? i.e. 1) create user in DD 2) set permissions for the  user 3) store credentials in cicd $) configure cicd to upload reports to DD. Something like that?
I could imaging that maybe you want to be able to create users automatically and upon creating you get an API token for that newly created user. Maybe that could be something we could support, but only for extremely privileged users and with some form of audit logging probably.

dsever:
@valentijnscholten  we can discuss it in PM, I don't want disclose what we are doing externally.

dsever:
> So about this:

> I've already done the equivalent using automation - granted I automated the web UI for a portion of it but I have have a fully automated means get a non-human's (MachineAccount) token that works just fine. No humans need to be involved - 100% programmatically done. I'm not sure why this isn't a viable route to meet your needs. Yes, using Selenium or some other headless 'browser driver' isn't very exciting or sexy but it has worked very reliably for me.
> 

It will work, but you have to admit, introducing Selenium into the project to simulate user activities is not the most convenient  way. I would rather enable API endpoint to setup init password for the local user. 


> > if you have some bad super user remember mu PR is not breaking anything, just take look, superuser can fetch all tokens in GUI anyway:
> 
> The Django admin interface has been a sore point for me for quite some time and I'd like it removed during the 2.0 release so that DefectDojo ships with the Django admin not installed by default. That's another example of a great way for users to accidentally shoot themselves in the foot if they don't fully understand the inner workings of DefectDojo. The only remaining use for the admin interface is password sets/resets which is a PR away from being removed.
> 

Fully agree. 

> I do agree that Auth2 does help with the damage potential but that assumes the DefectDojo users use it correctly (aka have only 1 or very limited super users). While that helps, you could still automate the small portion of the web UI needed to meet your goal without an API change.
> 

Yes, agree but remember user is always weakness point. If there is no prudent man, most of controls are useless. From my PoV,  assign multiple users as superuser or executing daily tasks as superuser is big NO. 

> I'm with @valentijnscholten - I don't fully understand your use-case so it's hard to see why this potentially security-reducing change adds more value than it takes away.

As I said we can discuss it in private channel.

Thanks

aaronweaver:
I also agree with the sentiments shared by Matt, Greg and Valentijn. Standard security practices are to not have this available to any super user. AWS has this pattern and many the security products use a similar approach.

Maffooch:
I think this would be a good addition to the [Community Contributions](https://github.com/DefectDojo/Community-Contribs) repo. Fits the bill for the few looking for this exact workflow, but won't potentially penalize those that want it away from the main product.

dsever:
Guys, I don't have problem to move away from this approach. But please can we find optimal way to engage API, to automate behavior. 

As I proposed, to enable endpoint, to set init password using API for created user?

https://github.com/DefectDojo/django-DefectDojo/pull/4511#issuecomment-843940437

kiblik:
Everybody has meaningful arguments but I agree with @dsever. 
- if something is possible to do in GUI there should not be a problem to allow it also in API (same permissions, same actions)
- API keys are already accessible for every super-user
- in my opinion, selenium is a powerful tool for no-API applications and for UI unit tests. But when an application already supports API and there is no problem to replicate some functionality from UI (with no-breaking existing authz model) there is no reason to forbid this implementation
- it can be an opt-in function (disabled by default) if there are people who agree and also people who disagree (`DD_API_KEY_FETCH_BY_SUPERADMIN`)

valentijnscholten:
Still not seeing any steps / usecase description....

How is everyone currently obtaining the API token for machine users? I was about the remove the tokens from the `/admin` portal in a PR. What I do is:

As admin / superuser:
1) create user in DD
2) set correct permissions
3) set password

As "the new user":
4) Login with said password;
5) Obtain API token;
6) Use API token in Jenkins;
7) Profit

An admin can set passwords for users, I think that is a feature that will probably remain. So that could be used to perform steps  1-3. Isn't that enough to support this scenario in an automated way?

Currently a user or admin cannot set a password via the API, so the question is if we would be OK to add that feature? Probably doesn't feel good without audit logging, but this part is already possible via the UI and will probably remain as its part of Django by default. Would have been a lot better if Django would force users to change their password on first use.

There's also the `/manage.py changepassword` command that can be used to set a password for a user.

So @dsever I think you should describe your usecase more, and find a solution that doesn't cause easy exposure of (all) users API tokens. I haven't researched how AWS, GCP or Azure do this. I think they have special permissions needed to allow creation of new users and what permissions are allowed to be given to those users. We can't implement all that, but we can learn from the principles they have applied.

valentijnscholten:
https://github.com/encode/django-rest-framework/issues/8000

dsever:
Hi @valentijnscholten 

> Still not seeing any steps / usecase description....

I asked you for the call to describe it. 

> How is everyone currently obtaining the API token for machine users? I was about the remove the tokens from the `/admin` portal in a PR. What I do is:
> 
> As admin / superuser:
> 
> 1. create user in DD
> 2. set correct permissions
> 3. set password

Not possible in automated way, only by gui.

> 
> As "the new user":
> 4) Login with said password;
> 5) Obtain API token;
> 6) Use API token in Jenkins;
> 7) Profit
> 
> An admin can set passwords for users, I think that is a feature that will probably remain. So that could be used to perform steps 1-3. Isn't that enough to support this scenario in an automated way?

No, GUI needs to be in use, to set password, we are targeting zero touch. 

> 
> Currently a user or admin cannot set a password via the API, so the question is if we would be OK to add that feature when a new user is created? Probably doesn't feel good without audit logging, but this part is already possible via the UI and will probably remain as its part of Django by default. Would have been a lot better if Django would force users to change their password on first use.
> 

In any case, IAM accounts actions needs to recorded and archived, in nonrepudiation way. 

> There's also the `/manage.py changepassword` command that can be used to set a password for a user.

This is bad idea. 
 
> So @dsever I think you should describe your usecase more, and find a solution that doesn't cause easy exposure of (all) users API tokens. I haven't researched how AWS, GCP or Azure do this. I think they have special permissions needed to allow creation of new users and what permissions are allowed to be given to those users. We can't implement all that, but we can learn from the principles they have applied.

In AWS implementation  user with  enough rights, can fetch keys behalf of the others, same is in GCP very similar (they have Service Accounts).  

image

it is very similar what we proposed. 

I will repeat this PR should not be accepted, but still there should be option to setup init password than for the created user using API call.

Dubravko

kiblik:
Selenium is a useful tool for non-API application and there should be some way how to perform full user management using API (including setting a password and/or fetching API key).

This offered solution is opt-in (`DD_FEATURE_SUPERADMIN_TOKEN_FETCH` is by default `False`) and so it can be used only if you accept this approach.

I support this PR

PR #4516, opened 2021-05-14
Auto-redirect when only one Social Auth is defined

kiblik:
If `DD_SOCIAL_AUTH_SHOW_LOGIN_FORM` is set to `False` (local login is not allowed) and there is exactly one Social Auth definition, the `/login` page automatically redirect to the Social login form.
There is no reason to show only one Social button.


valentijnscholten:
I don't agree this should be the default behavior. Maybe optional via a config parameter, but not by default.

valentijnscholten (changes requested):
see comment

kiblik:
@valentijnscholten, fixed in https://github.com/DefectDojo/django-DefectDojo/pull/4516/commits/c5a25a96a0bef49f5a94efc1cc28ff6238d8ae95

valentijnscholten (approved):

valentijnscholten:
Thanks, can you update the description as well?

damiencarol (approved):

damiencarol (approved):

PR #4520, opened 2021-05-17
Authorization V2: Global roles

StefanFl:
Users can be assigned a global role in the _Edit User_ dialog. A global role gives a user access to all Product Types and Products, including the underlying data, with permissions according to the respective role.

A use case for a global role could be the Chief Information Security Officer of a company who needs an overview of all systems. If he gets the global role `Reader`, he can see the findings for all products and also all metrics.


StefanFl:
When we have global roles for users, we should have global roles for the newly invented groups as well.

StefanFl:
The global role is now in a model class on its own, to be more flexible for future enhancements

StefanFl:
... and there is a nice API now as well. No more changes from my side unless there are further change requests from reviewers.

StefanFl:
This PR has to wait for #4603, because global roles need to be tied to Role model class.

StefanFl:
All changes needed after moving roles into a model have been implemented now.

madchap (approved):

valentijnscholten (approved):

PR #4527, opened 2021-05-19
add new parser - Detect-secrets

ansidorov:
detect-secrets is an aptly named module for (surprise, surprise) detecting secrets within a code base.

It could be found at: https://github.com/Yelp/detect-secrets

### DIFF
--- a/docs/content/en/integrations/import.md
+++ b/docs/content/en/integrations/import.md
@@ -210,6 +210,10 @@ Import of JSON report from
 Import JSON container image linter reports
 <https://github.com/goodwithtech/dockle>
 
+### Detect-secrets
+
+Import of JSON report from <https://github.com/Yelp/detect-secrets>
+
 ### ESLint
 
 ESLint Json report format (-f json)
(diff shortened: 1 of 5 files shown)


valentijnscholten (approved):

ansidorov:

If I understand correctly, we need to use the default for detect-secrets parser.

devGregA (approved):

docs/content/en/integrations/import.md
```diff
@@ -668,6 +672,7 @@ The CSV output from the UI is now also accepted.
 
 ### Visual Code Grepper (VCG)
 
+
```
damiencarol, line 675:
```suggestion

```

damiencarol: Why this line.

dojo/unittests/scans/detect_secrets/no_findings.json
```diff
@@ -0,0 +1,94 @@
...
+    },
+    {
+      "path": "detect_secrets.filters.heuristic.is_sequential_string"
+    },
+    {
+      "path": "detect_secrets.filters.heuristic.is_templated_secret"
+    }
+  ],
+  "results": {},
+  "generated_at": "2021-05-19T10:45:24Z"
```
damiencarol, line 93: Could you use this attribute to set the `date` attribute of findings?

ansidorov: Done

damiencarol (approved):

PR #4528, opened 2021-05-19
add new parser - terrascan

ansidorov:
Terrascan is a static code analyzer for Infrastructure as Code.

It could be found at: https://github.com/accurics/terrascan

### DIFF
--- a/dojo/tools/terrascan/parser.py
+++ b/dojo/tools/terrascan/parser.py
@@ -29,6 +29,8 @@ class TerrascanParser(object):
         dupes = {}
         if 'results' not in data and 'violations' not in data.get('results'):
             raise ValueError("missing mandatory attribute 'results'")
+        if data.get('results') is None:
+            return list()
         for item in data.get('results').get('violations'):
             rule_name = item.get('rule_name')
             description = item.get('description')
--- a/dojo/tools/terrascan/parser.py
+++ b/dojo/tools/terrascan/parser.py
@@ -27,8 +27,9 @@ class TerrascanParser(object):
     def get_findings(self, filename, test):
         data = json.load(filename)
         dupes = {}
-        if data.get('results').get('violations'):
-            for item in data.get('results').get('violations'):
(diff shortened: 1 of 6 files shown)


valentijnscholten (approved):

ansidorov:

If I understand correctly, we need to use the default for terrascan parser.

devGregA (approved):

ansidorov:

I thought again and adjusted the deduplication algorithm.

dojo/tools/terrascan/parser.py
```diff
@@ -0,0 +1,65 @@
...
+        return "Terrascan Scan"
+
+    def get_description_for_scan_types(self, scan_type):
+        return "Import JSON output for Terrascan scan report."
+
+    def get_findings(self, filename, test):
+        data = json.load(filename)
+        dupes = {}
+        if data.get('results').get('violations'):
+            for item in data.get('results').get('violations'):
```
damiencarol, line 31: This conditions can fail if `results` attribute is not here.
```suggestion
        if 'results' not in data and 'violations' not in data.get('results'):
            raise ValueError("missing mandatory attribute 'results'")
        for item in data.get('results').get('violations'):
```

dojo/tools/terrascan/parser.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def get_description_for_scan_types(self, scan_type):
+        return "Import JSON output for Terrascan scan report."
+
+    def get_findings(self, filename, test):
+        data = json.load(filename)
+        dupes = {}
+        if 'results' not in data and 'violations' not in data.get('results'):
+            raise ValueError("missing mandatory attribute 'results'")
+        for item in data.get('results').get('violations'):
```
damiencarol, line 32:
```suggestion
        if data.get('results') is None:
            return list()
        for item in data.get('results').get('violations'):
```

damiencarol (approved):

PR #4534, opened 2021-05-20
add new parser - Nuclei

### LINKED ISSUE #3898: Nuclei importer
**Scanner Name**
Nuclei (JSON-per-line): https://github.com/projectdiscovery/nuclei (e.g.: https://github.com/projectdiscovery/nuclei/blob/31e9c04d5ec05674bad557ede44f22cecd15c0eb/v2/pkg/executer/output_http.go#L25 , https://github.com/projectdiscovery/nuclei/blob/31e9c04d5ec05674bad557ede44f22cecd15c0eb/v2/pkg/executer/output_dns.go#L16)

**Sample File**
Varies based on the template, / workflow, a simple example that doesn't use any extractor (e.g. WAF detection `nuclei -l target_urls.txt -t ./technologies/waf-detect.yaml -json`)

```json
{"name":"WAF Detection","author":"dwisiswant0","severity":"info","matcher_name":"perimx","matched":"https://example.com/","template":"waf-detect","type":"http","host":"https://example.com/"}
{"severity":"info","matcher_name":"ats","matched":"https://example.com/","template":"waf-detect","type":"http","host":"https://example.com/","name":"WAF Detection","author":"dwisiswant0"}
...
```

ansidorov:
Nuclei is used to send requests across targets based on a template leading to zero false positives and providing fast scanning on large number of hosts.

It could be found at: https://github.com/projectdiscovery/nuclei

Resolve #3898


valentijnscholten (approved):

valentijnscholten:

ansidorov:

If I understand correctly, we need to use the default for nuclei parser.

dojo/unittests/tools/test_nuclei_parser.py
```diff
@@ -0,0 +1,122 @@
...
+
+        with self.subTest(i=8):
+            finding = findings[8]
+            self.assertEqual("MySQL DB with enabled native password: nuclei-example.com:3306", finding.title)
+            self.assertEqual("Info", finding.severity)
+            self.assertEqual(1, finding.nb_occurences)
+            self.assertIsNotNone(finding.description)
+            self.assertIsNone(finding.references)
+            self.assertEqual("network,mysql,bruteforce,db", finding.unsaved_tags)
+            self.assertEqual([], finding.unsaved_endpoints)
```
damiencarol, line 122: Why this finding doesn't have a host?

damiencarol (changes requested):
This PR need some re-work since we merged #4473 + see my comments

valentijnscholten:
@ansidorov

dojo/tools/nuclei/parser.py
```diff
@@ -0,0 +1,84 @@
...
+                dupe_key = hashlib.sha256(
+                    (template_id + type + host + ip + matched).encode('utf-8')
+                ).hexdigest()
+
+                if dupe_key in dupes:
+                    finding = dupes[dupe_key]
+                    finding.nb_occurences += 1
+                else:
+                    finding = Finding(
+                        title=f"{name}: {matched}",
```
kiblik, line 60: I suppose, URL shouldn't be part of the title. URL should be added as an endpoint.
Same for `dupe_key`

kiblik (changes requested):
URL (matched) shouldn't be part of `dupe_key` and `Finding.title`

dojo/unittests/tools/test_nuclei_parser.py
```diff
@@ -0,0 +1,122 @@
...
+        testfile = open("dojo/unittests/scans/nuclei/no_findings.json")
+        parser = NucleiParser()
+        findings = parser.get_findings(testfile, Test())
+        self.assertEqual(0, len(findings))
+
+    def test_parse_many_findings(self):
+        testfile = open("dojo/unittests/scans/nuclei/many_findings.json")
+        parser = NucleiParser()
+        findings = parser.get_findings(testfile, Test())
+        self.assertEqual(9, len(findings))
```
kiblik, line 18: Please add also cleaning: 

```python
for finding in findings:
    for endpoint in finding.unsaved_endpoints:
        endpoint.clean()
```

https://github.com/DefectDojo/cookiecutter-scanner-parser/blob/main/temp-%7B%7B%20cookiecutter.tool_directory_name%20%7D%7D/unittests/tools/test_%7B%7B%20cookiecutter.tool_directory_name%20%7D%7D_parser.py#L20-L22

dojo/tools/nuclei/parser.py
ansidorov: I think it would be nice to leave the URL in these places, this allows you to visually and for deduplication determine different triggers, since without this defectdojo will assume that triggers for:
```
example.com/test1
example.com/test2
```
This is the same trigger.

dojo/unittests/tools/test_nuclei_parser.py
ansidorov: fixed

dojo/tools/nuclei/parser.py
```diff
@@ -0,0 +1,84 @@
...
+import hashlib
+import hyperlink
+from dojo.models import Finding, Endpoint
+from django.core.validators import URLValidator
+from django.core.exceptions import ValidationError
+
+
+class NucleiParser(object):
+    """
+    A class that can be used to parse the nuclei JSON report file
```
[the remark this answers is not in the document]
ansidorov: added

dojo/unittests/tools/test_nuclei_parser.py
ansidorov: added

kiblik: it is good to add it before all `self.assertEqual("...", finding.unsaved_endpoints[0]....)` because `.clean()` can change content of fields in `Endpoint`

ansidorov: Done

dojo/tools/nuclei/parser.py
kiblik: As I know, all findings from the dynamic analysis are every time created as one finding (if it is based on the same CVE, uses the same Title, or ...). The `Endpoints` are used to identify which on places is possible to find this vulnerability.

kiblik: Also `host` and `ip` shouldn't be in `dupe_key` (same reason as `matched`)

dojo/tools/nuclei/parser.py
```diff
@@ -0,0 +1,75 @@
...
+                        nb_occurences=1,
+                    )
+                    if info.get('description'):
+                        finding.description = info.get('description')
+                    if info.get('tags'):
+                        finding.unsaved_tags = info.get('tags')
+                    if info.get('reference'):
+                        finding.references = info.get('reference')
+                    finding.unsaved_endpoints = []
+                    if '://' in matched:
```
kiblik, line 69: Endpoints shouldn't be added only in "first match" but also when dupe is identified. I suppose move L69-73 should be enough and move L74 after L68

kiblik: It is good to write also unit tests for this - how many endpoints does finding have

dojo/tools/nuclei/parser.py
damiencarol: I think the same. The UI already take care of endpoints. It's better to let the right data structure support this data (endpoints/URL) instead of keeping it artificially in the title.

dojo/tools/nuclei/parser.py
damiencarol: Yes, you're right. @ansidorov could you add unit tests for endpoints?

dojo/unittests/tools/test_nuclei_parser.py
```diff
@@ -0,0 +1,130 @@
...
+
+        with self.subTest(i=0):
+            finding = findings[0]
+            self.assertEqual("OpenSSH 5.3 Detection", finding.title)
+            self.assertEqual("Low", finding.severity)
+            self.assertEqual(1, finding.nb_occurences)
+            self.assertIsNotNone(finding.description)
+            self.assertEqual("network,openssh", finding.unsaved_tags)
+            self.assertIsNotNone(finding.references)
+            self.assertEqual("nuclei-example.com", finding.unsaved_endpoints[0].host)
```
damiencarol, line 33: Add checks on the number of endpoints please.

damiencarol (changes requested):

dojo/tools/nuclei/parser.py
ansidorov: Fixed

dojo/tools/nuclei/parser.py
ansidorov: Fixed

dojo/unittests/tools/test_nuclei_parser.py
ansidorov: Added

dojo/tools/nuclei/parser.py
```diff
@@ -0,0 +1,75 @@
...
+                        nb_occurences=1,
+                    )
+                    if info.get('description'):
+                        finding.description = info.get('description')
+                    if info.get('tags'):
+                        finding.unsaved_tags = info.get('tags')
+                    if info.get('reference'):
+                        finding.references = info.get('reference')
+                    if endpoint not in finding.unsaved_endpoints:
+                        finding.unsaved_endpoints.append(endpoint)
```
damiencarol, line 73: Could you move this part before the "dedup" part?

dojo/tools/nuclei/parser.py
```diff
@@ -0,0 +1,75 @@
...
+    A class that can be used to parse the nuclei (https://github.com/projectdiscovery/nuclei) JSON report file
+    """
+
+    # table to match nuclei severity to DefectDojo severity
+    SEVERITY = {
+        "critical": "Critical",
+        "high": "High",
+        "medium": "Medium",
+        "low": "Low",
+        "info": "Info",
```
damiencarol, line 17: It seems that defect dojo severity = (Nuclei severity).title()
I have the feeling that this mapping is not needed

dojo/unittests/tools/test_nuclei_parser.py
```diff
@@ -0,0 +1,112 @@
...
+            self.assertEqual(443, finding.unsaved_endpoints[0].port)
+
+        with self.subTest(i=6):
+            finding = findings[6]
+            self.assertEqual("MySQL DB with enabled native password", finding.title)
+            self.assertEqual("Info", finding.severity)
+            self.assertEqual(1, finding.nb_occurences)
+            self.assertIsNotNone(finding.description)
+            self.assertIsNone(finding.references)
+            self.assertEqual("network,mysql,bruteforce,db", finding.unsaved_tags)
```
damiencarol, line 109: unsaved tags should be an array.
Also better to have checks like that:
```python
    self.assertIn("network", finding.unsaved_tags)
    self.assertIn("mysql", finding.unsaved_tags)
```

damiencarol (changes requested):

dojo/tools/nuclei/parser.py
ansidorov: I don't really understand how this can be done.

dojo/unittests/tools/test_nuclei_parser.py
ansidorov: Done

dojo/tools/nuclei/parser.py
damiencarol: You build the object `finding` before line 42

dojo/tools/nuclei/parser.py
```diff
@@ -0,0 +1,63 @@
...
+                    if endpoint not in finding.unsaved_endpoints:
+                        finding.unsaved_endpoints.append(endpoint)
+                    finding.nb_occurences += 1
+                else:
+                    finding = Finding(
+                        title=f"{name}",
+                        test=test,
+                        severity=severity,
+                        nb_occurences=1,
+                    )
```
damiencarol, line 53: It seems that `TemplateID` data is our `vuln_id_from_tool` attribute.
Could you add it here?
```suggestion
                        vuln_id_from_tool=item.get('templateID'),
                    )
```

damiencarol (changes requested):

dojo/tools/nuclei/parser.py
ansidorov: But I need to get this Findings class in dedupe and compare the presence of endpoints there. Right?

damiencarol: yes, you can keep the internal dedup part. just move the finding = Finding() outside this if/else block.

ansidorov: Done

damiencarol (approved):

damiencarol:
@kiblik thanks for your help in the review.

PR #4535, opened 2021-05-20
Create test types on parser registration instead of first scan import

JohannesZahn:
**Description**

Automatically generate Test Types for all Scan Types upon Parser Registration.
Currently Test Types are only generated upon using the import scan API.

This introduces problems for us (SecureCodeBox) as we want to create Tests in advance and then use the reimport scan API.

More info can be found here: https://owasp.slack.com/archives/C014H3ZV9U6/p1620734079161800

**Test results**

manually tested if the test types are created automatically or not. Works as expected

**Documentation**

Did not update

**Checklist**

- [x] Features/Changes/Bugfixes should be submitted against the `dev` branch by default.
- [x] Hotfixes should be submitted against master (urgent bugfixes requiring a hotfix release).


JohannesZahn:
Are the changes in the initial surveys json necessary or can I undo that commit?

valentijnscholten:
> Are the changes in the initial surveys json necessary or can I undo that commit?

those should be ignored :-)

JohannesZahn:
> > Are the changes in the initial surveys json necessary or can I undo that commit?
> 
> those should be ignored :-)

Done :)

valentijnscholten:
Looks like the parsers are imported before models are registered / migrations have been run.

valentijnscholten:
Seems I can't push to your PR? This commit could/should fix the timing issue of the initialization: https://github.com/valentijnscholten/django-DefectDojo/commit/ab191f9396a84fa6da8feb5f10791d768bbf92ad

wdyt @damiencarol

JohannesZahn:
On first time docker-compose up after docker-compose build your fix gives me errors because the table test_type does not yet exist

damiencarol:
Took a look at the failed unit tests.
It seems that you will need to remove the fixture that manage Test_Type completely to make it work.
```
uwsgi_1         | 
uwsgi_1         | During handling of the above exception, another exception occurred:
uwsgi_1         | 
uwsgi_1         |   File "/app/dojo/unittests/test_sample_data.py", line 10, in test_loaddata
uwsgi_1         |     self.assertEqual(False, True, e)
uwsgi_1         | AssertionError: False != True : Problem installing fixture '/app/dojo/fixtures/defect_dojo_sample_data.json': Could not load dojo.Test_Type(pk=4): (1062, "Duplicate entry 'Nessus Scan' for key 'name'")
uwsgi_1         | 
```
The system try to load them

valentijnscholten:
@JohannesZahn can you try to merge (or just mimic) the changes from https://github.com/DefectDojo/django-DefectDojo/pull/4437 ?

JohannesZahn:
@valentijnscholten I merged the changes from #4437 as requested :) I will get back to this next tuesday.

valentijnscholten:
Looks like `dojo_testdata.json` and `dojo_sampledata.json` also need to be cleaned up. But this leads to the problem that the fixture cannot load existing findings against know test_types.

damiencarol:
@valentijnscholten I agree, our reference should be the id of the parser aka the `name`, or  `scan_type` as it is named and managed by the factory.

This identifier need to be consistent across instances and is linked to the code of the parser (the parser return it and the factory use it as a key to store/manage them)
What doesn't make sense for me and could be changed is the fixtures. We should completely get ride of fixtures because:
1. the factory insert in DB as it discover new parsers on-the-fly
2. adding and/or removing parsers is dynamic so we can't force the data in DB to be fixed
3. if we need an up-to-date list of these references for API or the UI, the code should query the factory to have this list

valentijnscholten:
Downside would be that we would get varchar foreign keys. An alternative could be to degine an integer parser_id inside each parser. We could reserve 1-100000 for Defect Dojo, and instruct users to use 999999->downward for their custom non upstream parsers?

JohannesZahn:
We will probably not need this as we are now using the new generic JSON parser. So I will close the PR for now.

damiencarol:
@valentijnscholten 
> Downside would be that we would get varchar foreign keys. An alternative could be to degine an integer parser_id inside each parser. We could reserve 1-100000 for Defect Dojo, and instruct users to use 999999->downward for their custom non upstream parsers?

this key is only in `Test` and not in `Finding` so I guess the pb is smaller. Also you proposal is to switch to number instead of id in letter. I need to analyze that. Could be hard to change every parser and assign a number now that we maintain more than 120 parsers :D

PR #4548, opened 2021-05-21
Added Nessus WAS Scanner

blakeaowens:
Added Nessus WAS scanner which accepts .csv and .nessus/.xml reports.


dojo/tools/nessus_was/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'Blake Owens'
```
damiencarol, line 1:
```suggestion
```
We rely on the VCS to track authorship.

damiencarol (changes requested):

dojo/tools/nessus_was/parser.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+class NessusWASXMLParser(object):
+    def get_findings(self, file, test):
+        nscan = ElementTree.parse(file)
+        root = nscan.getroot()
+
+        if 'NessusClientData_v2' not in root.tag:
+            raise NamespaceErr('This version of Nessus report is not supported. Please make sure the export is '
+                               'formatted using the NessusClientData_v2 schema.')
```
damiencarol, line 106:
```suggestion
            raise ValueError('This version of Nessus report is not supported. Please make sure the export is '
                               'formatted using the NessusClientData_v2 schema.')
```

devGregA (approved):

damiencarol (approved):

PR #4551, opened 2021-05-24
Adding importer for Jfrog Xray Unified

LithiumBloom:
Add importer for Xray Unified. Jfrog's Xray tool is an add-on to their Artifactory repository that does Software Composition Analysis, see https://www.jfrog.com/confluence/display/JFROG/JFrog+Xray for more information. "Xray Unified" refers to Xray Version 3.0 and later.


dojo/tools/jfrog_xray_unified/parser.py
```diff
@@ -0,0 +1,118 @@
...
+    # TODO: Needs UNKNOWN new status in the model.
+    else:
+        severity = "Info"
+
+    cveIndex = highestCvssV3Index
+
+    cve = "No CVE on file"
+    cvss_v3 = "No CVSS v3 score."
+    cvss_v2 = "No CVSS v2 score."
+    mitigation = "N/A"
```
damiencarol, line 66: `cve`, and `cvssv3` attributes are validated by django validator.
Better to not set them when there is no value. 
Also `mitigation` is null-able.

dojo/unittests/tools/test_jfrog_xray_unified_parser.py
```diff
@@ -0,0 +1,303 @@
...
+
+        self.assertEquals("XRAY-139239 - This affects the package", item.title[:38])
+        self.assertEquals(" memory.", item.title[-8:])
+        self.assertEquals("CVE-2020-28493", item.cve)
+        self.assertEquals("Medium", item.severity)
+        self.assertEquals("This affects the package", item.description[:24])
+        self.assertEquals(" memory.", item.description[-8:])
+        self.assertIsNotNone(item.mitigation)
+        self.assertGreater(len(item.mitigation), 0)
+        self.assertEquals("pypi://Jinja2", item.component_name)
```
damiencarol, line 31: Should it be `Jinja2`?

dojo/tools/jfrog_xray_unified/parser.py
```diff
@@ -0,0 +1,118 @@
...
+        extra_desc = vulnerability['external_advisory_source'] + ": " + vulnerability['external_advisory_severity']
+
+    if vulnerability['issue_id']:
+        title = vulnerability['issue_id'] + " - " + vulnerability['summary']
+    else:
+        title = vulnerability['summary']
+
+    references = ""
+    for reference in vulnerability['references']:
+        references += reference + "\n"
```
damiencarol, line 93:
```suggestion
    references = "\n".join(vulnerability['references'])
```

dojo/tools/jfrog_xray_unified/parser.py
```diff
@@ -0,0 +1,118 @@
...
+        if 'cve' in cves[cveIndex]:
+            cve = worstCve['cve']
+        if 'cvss_v3_vector' in worstCve:
+            cvss_v3 = worstCve['cvss_v3_vector']
+        if 'cvss_v2_vector' in worstCve:
+            cvss_v2 = worstCve['cvss_v2_vector']
+
+    if 'fixed_versions' in vulnerability and len(vulnerability['fixed_versions']) > 0:
+        mitigation = "Versions containing a fix:\n"
+        mitigation = mitigation + "</br>".join(vulnerability['fixed_versions'])
```
damiencarol, line 81:
```suggestion
        mitigation = "Versions containing a fix:\n"
        mitigation = mitigation + "</br>".join(vulnerability['fixed_versions'])
```
The syntax should be Markdown here not HTML.

dojo/tools/jfrog_xray_unified/parser.py
```diff
@@ -0,0 +1,118 @@
...
+        mitigation=mitigation,
+        component_name=component_name,
+        component_version=component_version,
+        file_path=vulnerability['path'],
+        severity_justification="CVSS v3 base score: {}\nCVSS v2 base score: {}".format(cvss_v3, cvss_v2),
+        static_finding=True,
+        dynamic_finding=False,
+        references=references,
+        impact=severity,
+        cvssv3=cvss_v3)
```
damiencarol, line 116: I think we can add more attribute.
like the `date` which seems to be the attribute `artifact_scan_time` in the report.
Also `"issue_id": "XRAY-139239"` should be the attribute `unique_id_from_tool` or `vuln_id_from_tool`

damiencarol (changes requested):
Made some comments, the parser is good.
There is some flake8 errors and more attributes could be used from data source.

dojo/unittests/tools/test_jfrog_xray_unified_parser.py
LithiumBloom: Jinga2 seems right? and the test is passing... I'm not familiar with it since this is just a scan of a public repo, but looks like it is this: https://jinja2docs.readthedocs.io/en/stable/

dojo/tools/jfrog_xray_unified/parser.py
LithiumBloom: Ok I'll change it to use \n instead of </br> then? I'm assuming the suggested change is a mistake since i can't see a difference in the code?

LithiumBloom:
@damiencarol thanks for the feedback =) a few of these comments also apply to the Xray V2 parser that I was basing this one off, should I update the original too? I don't have access to an old version of Xray to test with, but can test with the files in the unit tests if that is sufficient

dojo/tools/jfrog_xray_unified/parser.py
LithiumBloom: fixed in https://github.com/DefectDojo/django-DefectDojo/pull/4551/commits/0f106188aecc6723c912301d1b6f01afa5c30903

dojo/tools/jfrog_xray_unified/parser.py
LithiumBloom: fixed in https://github.com/DefectDojo/django-DefectDojo/pull/4551/commits/33ca0146564f1eccd28c362ff818b24c985d008a

dojo/tools/jfrog_xray_unified/parser.py
LithiumBloom: fixed in https://github.com/DefectDojo/django-DefectDojo/pull/4551/commits/0f9f471c870e6ed15da7a17451c3ca986bc077a1

dojo/unittests/tools/test_jfrog_xray_unified_parser.py
damiencarol: I mean, the value that we should have it here should be `Jinja2` (the name of the lib Jinja). I think we should split the component name part from the rest of the URI here. `pypi://` is not part of the name.

damiencarol:
@LithiumBloom I'ts up to you to improve the other one but please do it in another PR for the sake of reviewing code/understand the goal.
I saw that you used `vuln_id_from_tool` for `issue_id` are you sure that it's not an unique id for the whole instance of Xray?
If yes it's better to set is as `unique_id_from_tool` take a look at our documentation to choose the right one => https://defectdojo.github.io/django-DefectDojo/usage/features/#deduplication-algorithms

dojo/unittests/tools/test_jfrog_xray_unified_parser.py
LithiumBloom: Oh right, thanks for the clarification. I've updated this in commit https://github.com/DefectDojo/django-DefectDojo/pull/4551/commits/c7f0c3a637f4e672abe0731d8870d042b34660d5

I think it's important to keep the package type and I couldn't see any standard fields it would make sense to capture this in, so I've added it as a tag - is that a good way to do it?

LithiumBloom:
> I saw that you used `vuln_id_from_tool` for `issue_id` are you sure that it's not an unique id for the whole instance of Xray?
> If yes it's better to set is as `unique_id_from_tool` take a look at our documentation to choose the right one => https://defectdojo.github.io/django-DefectDojo/usage/features/#deduplication-algorithms

thanks for the link, I had originally copied the deduplication code from the previous xray parser but have now updated it to hopefully be consistent with the general approach for deduplication... https://github.com/DefectDojo/django-DefectDojo/pull/4551/commits/76df49fc420e60d778849d59446d8455cf8f706b

dojo/unittests/tools/test_jfrog_xray_unified_parser.py
```diff
@@ -283,18 +288,19 @@ def test_parse_file_with_very_many_vulns(self):
...
         self.assertEquals("CVE-2019-19645", item.cve)
         self.assertEquals("XRAY-106044 - CVE-2019-19645 sqlite: infinite", item.title[:45])
         self.assertEquals("TABLE statements", item.title[-16:])
         self.assertEquals("Medium", item.severity)
         self.assertEquals("alter.c in SQLite", item.description[:17])
         self.assertEquals("TABLE statements.\n\nRed Hat Severity: Moderate", item.description[-45:])
         self.assertIsNone(item.mitigation)
-        self.assertEquals("rpm://7:sqlite:0", item.component_name)
+        self.assertEquals("7:sqlite:0", item.component_name)
+        self.assertEquals('"packagetype_rpm"', item.tags)
```
damiencarol, line 303: there is double quote `"` in the data. is it something we want?

LithiumBloom: For the tags? This was the syntax I found somewhere for asserting equal for lists (or arrays? whatever tags are). If more tags are added then the test can be edited to e.g.
`self.assertEquals('"packagetype_rpm", "someOtherTag"', item.tags)`

The actual data does not have double quotes in it. From parser.py:
`tags = ["packagetype_" + vulnerability['package_type']]`

Happy to change the tests if there's a clearer way to write this that you know of?

dojo/unittests/tools/test_jfrog_xray_unified_parser.py
damiencarol: yeah, you've done the right thing. You're right there is nothing to store the package type.
I made a proposal to the team to use PURL to store component data.

dojo/unittests/tools/test_jfrog_xray_unified_parser.py
damiencarol: This form is good:
```suggestion
        self.assertIn('packagetype_rpm', item.tags)
```

damiencarol (approved):

valentijnscholten (approved):

PR #4559, opened 2021-05-26
Set a configurable min_access_level for Gitlab projects import. Fix #4486

### LINKED ISSUE #4486: Set a configurable min_access_level for Gitlab projects import
As a user, when I use "Login with Gitlab" button and GITLAB_PROJECT_AUTO_IMPORT is set to True, all the Gitlab projects I am member of are imported as products, regardless of my access level on the project (Guest, Reporter, Developer...)

As DefectDojo admin, I want to set a global min_access_level for Gitlab projects import, so that my platform users do not see too many products. Only projects where the user has an access_level equal or higher than _the value of my choice_, defined in settings.

christophe226:
The current PR fixes issue #4486.
We add an new environment variable in settings: DD_SOCIAL_AUTH_GITLAB_PROJECT_MIN_ACCESS_LEVEL
In pipeline.py, we list only Gitlab projects which for which the user has at least this access level.


damiencarol (approved):

damiencarol:
@christophe226 could you add a description to the PR?

valentijnscholten (approved):

PR #4564, opened 2021-05-27
Nexpose parser: Add "Host Up" + "Open port" and handling non-standard names of services

kiblik:
- Inspired by NMAP parser
- add non-standard service names to nexpose reports
- cleaning service names and tags

### DIFF
--- a/dojo/tools/nexpose/parser.py
+++ b/dojo/tools/nexpose/parser.py
@@ -1,4 +1,4 @@
-import html2text
+import html2text, re
 from defusedxml import ElementTree
 
 from dojo.models import Finding, Endpoint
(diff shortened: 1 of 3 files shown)


dojo/tools/nexpose/parser.py
```diff
@@ -184,6 +185,14 @@ def get_items(self, tree, vulns, test):
                 host['services'] = list()
                 host['vulns'] = self.parse_tests_type(node, vulns)
 
+                host['vulns'].append({
+                    'name': 'Host Up',
+                    'desc': 'Host is up because it replied on ICMP request or some TCP/UDP port is up',
+                    'severity': 'Info',
+                    'resolution': 'N/A',
+                    'vector': 'No impact provided',
```
damiencarol, line 193: Not fan of 'N/A' and 'No xxx provided'. We have null-able attributes so, we can find a way to not set them if there is no data.

kiblik: Fixed https://github.com/DefectDojo/django-DefectDojo/pull/4564/commits/506033e219c51e6e3fa15ae64ce2124fc59f1a34

dojo/unittests/tools/test_nexpose_parser.py
```diff
@@ -109,8 +186,8 @@ def test_nexpose_parser_dns(self):
         # self.assertEqual('dns://192.168.1.1#tcp', str(finding.unsaved_endpoints[0]))
         # self.assertEqual('dns://192.168.1.1#udp', str(finding.unsaved_endpoints[1]))
```
damiencarol, line 187: I think you can implement these two

damiencarol (approved):
It's a good addon

dojo/unittests/tools/test_nexpose_parser.py
kiblik: You are right :) I forgot

valentijnscholten (approved):

PR #4565, opened 2021-05-27
Ingress pathtype fix

dsever:
Fix of broken Ingress specification

### DIFF
--- a/.github/workflows/k8s-testing.yml
+++ b/.github/workflows/k8s-testing.yml
@@ -2,11 +2,11 @@ name: k8s deployment
 on:
   pull_request:
   push:
-    branches:
-      - master
-      - dev
-      - release/**
-      - hotfix/**
+    # branches:
+    #   - master
+    #   - dev
+    #   - release/**
+    #   - hotfix/**
 
 env:
   DD_DOCKER_REPO: defectdojo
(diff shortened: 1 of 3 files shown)


valentijnscholten:
Is there anway we could have detected this error? I think we do linting and we do deployment to minikube, but still all tests were green?

dsever:
teset 

> Is there anway we could have detected this error? I think we do linting and we do deployment to minikube, but still all tests were green?

It was not green https://github.com/DefectDojo/django-DefectDojo/blob/dev/helm/defectdojo/templates/django-ingress.yaml 

As far I can see, test was canceled during the test execution, but still pushed to the dev.

valentijnscholten:
I think this was merged in https://github.com/DefectDojo/django-DefectDojo/pull/4410 with green tests?

valentijnscholten:
old comment, check the results, all green.

dsever:
> old comment, check the results, all green.

In any case, it should fail:

I assume problem is here in command:

--set django.ingress.enabled=false \

madchap (approved):
Thanks @dsever - indeed the proper thing to do, checked against https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types too.

valentijnscholten (approved):

PR #4566, opened 2021-05-27
Upgrade celery to 5.1.0

Homopatrol:
Updating Celery to the latest release of 5.1.0.

This release Includes the fix https://github.com/celery/celery/pull/6600. which solves https://github.com/celery/celery/issues/6579.

Using OpenShift with a restricted SCC - uid=1000640000(1000640000) & gid=0(root) causes the Celery container to terminiate when testing for root.
```
oc logs defect-dojo-defectdojo-celery-worker-78dbb86b48-2rxn2
uid=1000640000(1000640000) gid=0(root) groups=0(root),1000640000
Waiting for database to be reachable [06/Jan/2021 13:40:58] INFO [dojo.models:3474] enabling audit logging
[06/Jan/2021 13:41:00] INFO [dojo.models:3474] enabling audit logging
Running a worker with superuser privileges when the
worker accepts messages serialized with pickle is a very bad idea!
If you really want to continue then you have to set the C_FORCE_ROOT
environment variable (but please think about this before you do).
User information: uid=1000640000 euid=1000640000 gid=0 egid=0
```
v5.1.0 of celery should ensure gid=0 (used by OpenShift restricted SCC) is not deemed a security risk that results in container termination.


valentijnscholten:
please base against `dev` as per contribution guidelines and PR template.

valentijnscholten:
Earlier we decided to postpone the major upgrade to v5, see: https://github.com/DefectDojo/django-DefectDojo/pull/2913 Of course we are happy to do a major upgrade, but we have to make sure everything keeps working. Currently we're seeing some failures in the GitHub actions / test suite.

valentijnscholten:
please rebase to avoid the strange commits coming in

valentijnscholten:
For some reason in GHA all tasks are run in the foreground during the integration tests:
```
 dojo_async_task <function post_process_finding_save at 0x7fb067ce08b0>: running task in the foreground as block_execution is set to True for admin
```
But when I run the integration tests locally, they are executed in the background/celery.

valentijnscholten:
merging dev with the celery fixes for the integration tests

valentijnscholten (approved):

damiencarol (approved):

PR #4573, opened 2021-05-28
Update import.md to add Semgrep parser

damiencarol:


valentijnscholten:
Should we mention that SARIF format can be imported, but with sarif parser?

damiencarol:
I was thinking the same but I prefer to write it in the SARIF section of the page.
Like saying all the tools that produce SARIF output. Will make it in another PR.

valentijnscholten (approved):

Maffooch (approved):

PR #4574, opened 2021-05-29
Endpoint: fix migration steps

kiblik:
move "RunPython(move_fqdn_to_host)"


valentijnscholten:
unit tests = ok, means migrations = ok -> merging.

PR #4582, opened 2021-05-31
Gitlab projects tag_list and web_url reflected in products. Fix #4489

### LINKED ISSUE #4489: Gitlab projects tag_list and web_url reflected in products
As a user, sometimes I want to filter DefectDojo products on specific tags which were in Gitlab projects, but it is not possible:  
when my Gitlab projects are imported as DefectDojo products, only the Gitlab project name is used to create the product, tag_list information is not used.
I also would like to retrieve Gitlab project URL in DefectDojo product, in order to access it easily. For now, I don't have this URL in DefectDojo product.

As a user, I want that my DefectDojo products have automatically the same tag list as the corresponding Gitlab projects. The tag list would be synchronized when I use "Login with Gitlab" button. In the same step, I want to store Gitlab project URL in product description field.

**Screenshots**
Gitlab project:
Gitlab project

DefectDojo product:
DefectDojo product

christophe226:
The current PR fixes issue DefectDojo#4489. We add 2 new settings variables GITLAB_PROJECT_IMPORT_TAGS and GITLAB_PROJECT_IMPORT_URL. If set to True, we import tags and/or web URL.

Here is the original description of the issue:

As a user, sometimes I want to filter DefectDojo products on specific tags which were in Gitlab projects, but it is not possible:
when my Gitlab projects are imported as DefectDojo products, only the Gitlab project name is used to create the product, tag_list information is not used.
I also would like to retrieve Gitlab project URL in DefectDojo product, in order to access it easily. For now, I don't have this URL in DefectDojo product.

As a user, I want that my DefectDojo products have automatically the same tag list as the corresponding Gitlab projects. The tag list would be synchronized when I use "Login with Gitlab" button. In the same step, I want to store Gitlab project URL in product description field.


dojo/pipeline.py
```diff
@@ -86,6 +86,16 @@ def update_product_access(backend, uid, user=None, social=None, *args, **kwargs)
                 product, created = Product.objects.get_or_create(name=project_name, prod_type=product_type)
                 if not settings.FEATURE_AUTHORIZATION_V2:
                     product.authorized_users.add(user)
+                    # Add tags if needed
+                    if settings.GITLAB_PROJECT_IMPORT_TAGS:
+                        tags = [",".join(project.tag_list) for project in projects if project.path_with_namespace == project_name]
```
Maffooch, line 91: Does a project always have something in the tag_list field? I'm wondering if the default is at least `[]` so that the join function does not break.

christophe226: Not sure that there is always a tag_list field, so I will check. And `tag_list` seems to be deprecated, replaced by `topics`, so I will first look for topics in priority.

valentijnscholten (approved):

dojo/pipeline.py
```diff
@@ -86,6 +86,15 @@ def update_product_access(backend, uid, user=None, social=None, *args, **kwargs)
                 product, created = Product.objects.get_or_create(name=project_name, prod_type=product_type)
                 if not settings.FEATURE_AUTHORIZATION_V2:
                     product.authorized_users.add(user)
+                    for project in projects:
```
StefanFl, line 89: This logic will only be called when the new authorization is not active. Since the new authorization is the default now, this logic would not be called anymore. It needs to be moved directly behind creating the product.

dojo/settings/settings.dist.py
```diff
@@ -100,6 +100,8 @@
     DD_SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_RESOURCE=(str, 'https://graph.microsoft.com/'),
     DD_SOCIAL_AUTH_GITLAB_OAUTH2_ENABLED=(bool, False),
     DD_SOCIAL_AUTH_GITLAB_PROJECT_AUTO_IMPORT=(bool, False),
+    DD_SOCIAL_AUTH_GITLAB_PROJECT_IMPORT_TAGS=(bool, False),
```
StefanFl, line 103: Both new settings are not used in the logic to import tags and url.

StefanFl (changes requested):

Maffooch (approved):

dojo/pipeline.py
```diff
@@ -93,7 +93,20 @@ def update_product_access(backend, uid, user=None, social=None, *args, **kwargs)
                         # Make product member an Owner of the product
                         product_member.role = 4
                         product_member.save()
-
+                # Import tags and/or URL if necessary
+                for project in projects:
```
StefanFl, line 97: Here we have a loop over GitLab projects within a loop over GitLab project names. Took me a few seconds to understand why. I think you have to look for the project with the current name. It would be much better to refactor the method slightly and convert line 83 to be `for project in project`. Then we could get rid of that inner loop.

StefanFl (changes requested):

StefanFl (approved):

PR #4591, opened 2021-06-01
Stop deduplication across products

Maffooch:
When using a parser that has a dedupe algorithm set to `DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE`, such as Veracode, deduplication would occur between products. This can be seen by importing the same report into two different products


devGregA (approved):

valentijnscholten (approved):

valentijnscholten:
nice find, could be something to add to the unit tests as well, i can take a look.

PR #4595, opened 2021-06-02
Revert "Bump urllib3 from 1.26.4 to 1.26.5"

valentijnscholten:
Reverts DefectDojo/django-DefectDojo#4590


valentijnscholten:
reverted as it has to go into dev

PR #4599, opened 2021-06-02
Added ability to set user password when adding a new user

blakeaowens:
The ability to change password already exists, however this removes the previously used "generate unusable password" feature when creating a new user.


dojo/user/views.py
```diff
@@ -284,7 +284,7 @@ def add_user(request):
         contact_form = UserContactInfoForm(request.POST)
         if form.is_valid() and contact_form.is_valid():
             user = form.save(commit=False)
-            user.set_unusable_password()
+            user.set_password(request.POST['password'])
```
valentijnscholten, line 287: not sure if this is what we want / users expect. unusable password can be useful. maybe if left empty, set unusable? or have checkbox/radiobutton? or settings option?

valentijnscholten (changes requested):
see above

valentijnscholten:
Thanks for the PR, there was some talk about this to have this added to the API. Would you be willing to look at adding it there as well?

dojo/user/views.py
blakeaowens: The goal of this PR was to get rid of the admin page. From my understanding, "set_unusable_password()" was used initially as a temporary solution when creating users.

blakeaowens: Also, maybe adding this to the API should be done in a different PR.

valentijnscholten: Could you describe what problem you are trying to solve?

blakeaowens: The problem I am solving is getting rid of the admin page. Unusable passwords can only be reset through the admin page. With that page removed, there needs to be a way to set an initial password for a new user that can be changed through the UI, which cannot be done with unusable passwords.

valentijnscholten (changes requested):
As said above I think the operator should be able to choose what to do. I am using LDAP and when creating users I do actually want to have the unusable password as I don't want users to start resetting their passwords. 

The Django Admin is currently used to assign model permissions to users, for example when creating CI/CD users with least privilege. So we can't disable the admin yet as long as we are not completely migrated to the new Authorization V2.

Maffooch:
I did not think the django model permissions have ever worked with DefectDojo. This was because when dojo was originally written, the Django permissions did not sufficiently cover the authorized user permission scheme currently in place.

Also looks like the admin page is not removed, but just disabled by default. Before, since the admin page enabled variable was set as an environment variable through the docker file, there was no way to disable it without a rebuild. Now it looks like the admin page can be enabled/disabled through settings as it was intended.

valentijnscholten:
The permissions are working, @madchap is using them and we've been instructing others on slack to use them.

Maffooch:
Oh interesting! I wish I would've none that years ago. Either way, the admin page is still available. Just needs to be enabled in settings first.

valentijnscholten:
Yeah, easiest would be to leave it enabled until we have the new authorization V2. But if other insist on disabling it by default, we should add some docs to re-enable it easily.

dojo/user/views.py
blakeaowens: With my latest commit the Django Admin page is enabled in settings (can easily be disabled), and unusable password functionality is still available. When creating a new user, an unusable password will be set if the password field is left blank (added help text to this field on the add user page for clarification). I also tagged on a fix for a problem I discovered with the change password logic, which previously allowed the user to successfully leave both password fields blank.

Maffooch (approved):
Looks ready to go now

valentijnscholten (approved):

valentijnscholten:
Good solution, thanks.

PR #4623, opened 2021-06-07
Add hook to initializer

tmetodie:
Add a hook to the Helm template to ensure that the initializer does not get ran every time as it breaks the database schema/data.


valentijnscholten:
The initializer job has to run every time Defect Dojo is updated or new code is deployed.

dsever:
@tmetodie you are able to block execution of initializer https://github.com/DefectDojo/django-DefectDojo/blob/master/helm/defectdojo/values.yaml#L162  if you have spacial case, but yes, init job is mandatory for each upgrade of DD.

madchap:
You could also set `DD_INITIALIZE` to `False` and keep control manually.

StefanFl:
There are 2 ways to disable the initializer, so I close this PR

PR #4634, opened 2021-06-09
Added Meterian Scanner

n-insaidoo:
Added Meterian scanner which accepts JSON reports.

### DIFF
--- a/dojo/tools/meterian/parser.py
+++ b/dojo/tools/meterian/parser.py
@@ -1,0 +1,139 @@
+import json
+from urllib.parse import urlparse
+
+from dojo.models import Finding
+
+
+class MeterianParser(object):
+
+    def get_scan_types(self):
+        return ["Meterian Scan"]
+
+    def get_label_for_scan_types(self, scan_type):
+        return scan_type
+
+    def get_description_for_scan_types(self, scan_type):
+        return "Meterian JSON report output file can be imported."
+
+    def get_findings(self, report, test):
+        findings = []
+        report_json = self.parse_json(report)
+
+        security_reports = report_json["reports"]["security"]["reports"]
+        for single_security_report in security_reports:
+            findings += self.do_get_findings(single_security_report, test)
+
+        return findings
+
+    def do_get_findings(self, single_security_report, test):
+        findings = []
+        language = single_security_report["language"]
+        for dependency_report in single_security_report["reports"]:
+
+            lib_name = dependency_report["dependency"]["name"]
+            lib_ver = dependency_report["dependency"]["version"]
+            finding_title = lib_name + ":" + lib_ver
+            for advisory in dependency_report["advices"]:
+
+                severity = self.get_severity(advisory)
+                finding = Finding(
+                    title=finding_title,
+                    test=test,
+                    severity=severity,
+                    severity_justification="Issue severity of: **" + severity + "** from a base " +
+                    "CVSS score of: **" + str(advisory.get('cvss')) + "**",
+                    description=advisory['description'],
+                    component_name=lib_name,
+                    component_version=lib_ver,
+                    false_p=False,
+                    duplicate=False,
+                    out_of_scope=False,
+                    impact=severity,
+                    static_finding=True,
+                    dynamic_finding=False,
+                    file_path="Manifest file",
+                    unique_id_from_tool=advisory['id'],
+                    tags=[language]
+                )
+
+                if 'cve' in advisory:
+                    if "N/A" != advisory["cve"]:
+                        finding.cve = advisory["cve"]
+
+                if "cwe" in advisory:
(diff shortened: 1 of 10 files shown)


n-insaidoo:
@valentijnscholten @damiencarol

dojo/tools/meterian/parser.py
```diff
@@ -0,0 +1,139 @@
...
+                severity = advisory["severity"].title()
+
+        return severity
+
+    def uri_validator(self, url):
+        try:
+            result = urlparse(url)
+            return all([result.scheme, result.netloc, result.path])
+        except:
+            return False
```
damiencarol, line 114: We use `hyperlink` module for all ULR related code.

https://defectdojo.github.io/django-DefectDojo/contributing/how-to-write-a-parser/#do-not-parse-urls-by-hand

dojo/tools/meterian/parser.py
```diff
@@ -0,0 +1,139 @@
...
+        try:
+            data = json_input.read()
+            try:
+                json_element = json.loads(str(data, 'utf-8'))
+            except:
+                json_element = json.loads(data)
+        except:
+            raise Exception("Invalid format")
+
+        return json_element
```
damiencarol, line 139: Please use only `json.load()` in your code. Don't create some specific function to parse.

dojo/tools/meterian/parser.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def get_label_for_scan_types(self, scan_type):
+        return scan_type
+
+    def get_description_for_scan_types(self, scan_type):
+        return "Meterian JSON report output file can be imported."
+
+    def get_findings(self, report, test):
+        findings = []
+        report_json = self.parse_json(report)
```
damiencarol, line 20: See my other comment
```suggestion
        report_json = json.load(report)
```

dojo/tools/meterian/parser.py
```diff
@@ -0,0 +1,139 @@
...
+
+        security_reports = report_json["reports"]["security"]["reports"]
```
damiencarol, line 22: This could break. We saw a lot of issues due to keys not present in the files.
Could you add some checks that raise `ValueError("...")` if the file is invalid (with unit tests to check that)?

dojo/unittests/tools/test_meterian_parser.py
```diff
@@ -0,0 +1,95 @@
...
+        self.assertTrue(self.findings_contains_tag(findings, "ruby"))
+
+    def findings_contains_tag(self, findings, tag):
+        found = False
+        for finding in findings:
+            if finding.tags == [tag]:
+                found = True
+                break
+
+        return found
```
damiencarol, line 95:
```suggestion
        self.assertIn("nodejs", finding[0].tags)
        self.assertIn("ruby", finding[1].tags)
```

dojo/tools/meterian/parser.py
```diff
@@ -0,0 +1,139 @@
...
+        for dependency_report in single_security_report["reports"]:
+
+            lib_name = dependency_report["dependency"]["name"]
+            lib_ver = dependency_report["dependency"]["version"]
+            finding_title = lib_name + ":" + lib_ver
+            for advisory in dependency_report["advices"]:
+
+                severity = self.get_severity(advisory)
+                finding = Finding(
+                    title=finding_title,
```
damiencarol, line 40: There is a date in the report.
Could you consider adding the `date` attribute to the findings?

damiencarol (changes requested):
It's a good first shot.
Thanks for contributing, I made some request for change as we are focusing on QA in parsers

dojo/tools/meterian/parser.py
n-insaidoo: Right, I will use that module.

dojo/tools/meterian/parser.py
n-insaidoo: Will do so.

dojo/tools/meterian/parser.py
n-insaidoo: Changed as per suggestion.

dojo/tools/meterian/parser.py
n-insaidoo: Now checking whether the report file is not malformed raising the a `ValueError` exception as suggested.

dojo/tools/meterian/parser.py
n-insaidoo: Sure, I will add that.

dojo/unittests/tools/test_meterian_parser.py
n-insaidoo: Now using built-in assertion methods.

damiencarol (approved):

valentijnscholten (approved):

PR #4638, opened 2021-06-09
Leverage template inheritance more effectively

Maffooch:
Currently, almost none of the templates here use template inheritance effectively as initially setup in `base.html`. 

**What is the goal of this PR:**
When extending a template block (in most cases here it's the content block in `base.html`) the html from the child templates content block will **overwrite** everything in the parent templates content block. The additions made to each of the template blocks int this PR is the `block.super` tag. This tag allows for the child templates content block to be **appended** to the parent templates content block.

**Why is this important:**
When developing features for internal use, extending pages is essential to maintaining an untouched dojo instance, while also accomplishing the goal of external features. 

**Who does this change affect:**
This change will go virtually unnoticed as none of the templates in defectdojo extends more than `base.html`

References: [Template Inheritance](https://docs.djangoproject.com/en/3.2/ref/templates/language/#template-inheritance)

### DIFF
--- a/dojo/settings/settings.py
+++ b/dojo/settings/settings.py
@@ -22,5 +22,4 @@ from split_settings.tools import optional, include
 include(
     'settings.dist.py',
     optional('local_settings.py'),
-    optional('dark_mode_settings.py')
 )
--- a/dojo/settings/settings.py
+++ b/dojo/settings/settings.py
@@ -21,5 +21,6 @@ from split_settings.tools import optional, include
 
 include(
     'settings.dist.py',
-    optional('local_settings.py')
+    optional('local_settings.py'),
+    optional('dark_mode_settings.py')
 )
--- a/dojo/settings/settings.py
+++ b/dojo/settings/settings.py
@@ -21,5 +21,5 @@ from split_settings.tools import optional, include
 
 include(
     'settings.dist.py',
-    optional('local_settings.py'),
+    optional('local_settings.py')
 )
--- a/dojo/templates/403.html
+++ b/dojo/templates/403.html
@@ -1,5 +1,6 @@
 {% extends "base.html" %}
 {% block content %}
+    {{ block.super }}
     <h1>403</h1>
     <hr/>
     <h2>
--- a/dojo/templates/404.html
+++ b/dojo/templates/404.html
@@ -1,5 +1,6 @@
 {% extends "base.html" %}
 {% block content %}
+    {{ block.super }}
     <h1>404</h1>
     <hr/>
     <h1>
--- a/dojo/templates/base.html
+++ b/dojo/templates/base.html
@@ -78,7 +78,9 @@
     <link rel="shortcut icon" href="{% static "dojo/img/favicon.ico" %}"/>
     <link rel="stylesheet" href="{% static "chosen-bootstrap/chosen.bootstrap.min.css" %}">
     <link rel="stylesheet" href="{% static "fullcalendar/dist/fullcalendar.min.css" %}">
-    <link rel="stylesheet" href="{% static "dojo/css/dojo.css" %}">
+    {% block dojo_css %}
+        <link rel="stylesheet" href="{% static "dojo/css/dojo.css" %}">
+    {% endblock %}
     <link rel="stylesheet" href="{% static "datatables.net-dt/css/jquery.dataTables.min.css" %}">
     {% comment %} <link rel="stylesheet" href="{% static "datatables.net-buttons-dt/css/buttons.dataTables.min.css" %}"> {% endcomment %}
     <link rel="stylesheet" href="{% static "datatables.net-bs/css/dataTables.bootstrap.min.css" %}">
--- a/dojo/templates/base.html
+++ b/dojo/templates/base.html
@@ -161,7 +161,6 @@
                             <li><a href="https://defectdojo.github.io/django-DefectDojo" target="_blank">
                                 <i class="fa fa-question-circle fa-fw"></i> Documentation</a>
                             </li>
-                            {{ block.super }}
                             {% if user.is_staff and system_settings.enable_questionnaires %}
                                 <li><a href="{% url 'questionnaire' %}"> <i class="fa fa-hospital-o fa-fw"> </i> Questionnaires </a></li>
                             {% endif %}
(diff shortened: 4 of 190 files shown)


valentijnscholten:
Can you give an example? I may not understand this fully, but if I want to override a template can't I just add {{ block.super }} myself in that new template?

Maffooch:
So let's say there is template **alpha** and template **beta**
Alpha is the template that displays findings (`findings_list_snippet.html`)
Beta is a template to add a banner to be seen on every single page (this has to happen in `base.html`)

This is the **current** behavior **without** `{{block.super}}`
Beta extends `base.html` and adds the banner to the content block inside `base.html`
Alpha extends `base.html` to render `findings_list_snippet.html`
Because alpha does not use the `{{block.super}}` tag, Alpha will **overwrite** all the changes in the content block inside `bast.html`

This is the **desired** behavior **with** `{{block.super}}`
Beta extends `base.html` and adds the banner to the content block inside `base.html`
Alpha extends `base.html` to render `findings_list_snippet.html`
Now that Alpha has the `{{block.super}}` tag, Alpha will **append** all the changes in the content block in `findings_list_snippet.html` to the content block inside `bast.html`

valentijnscholten:
I don't see how the banner from beta.html will end up in alpha.html.
In the example above alpha extends only base.html so it will never see any content from beta regardless of whether block.super is used anywhere? And how does one override finding_list_snippet with another template since the template filename is hardcoded inside defect dojo? Or is the idea that a "extends alpha.html" is added to finding_list_snippet?

Maffooch:
**A clarification: `finding_list_snippet.html` does not have a content block, but `finding_list.html` does have a content block and includes `finding_list_snippet.html` within the content block of `finding_list.html`.**

The code to create the banner does not physically end up in alpha nor `base.html`. Both templates are putting code in the [content block of base.html](https://github.com/DefectDojo/django-DefectDojo/blob/36a12fa4eb53d62ba3b339814c44ce649633a339/dojo/templates/base.html#L618)


Here is the order of templates rendered **WITHOUT** `{{block.super}}`
1. `base.html`
2. Beta -> `custom_banner.html` extends `base.html`
3. Alpha -> `finding_list.html` extends `base.html`

Explaining each step:
1. `base.html` is rendered. Nothing special here
2. Beta extends `base.html`. Beta also has a content block of its own, so the code to make the banner in that block will **REPLACE** the code inside the [content block of `base.html`](https://github.com/DefectDojo/django-DefectDojo/blob/36a12fa4eb53d62ba3b339814c44ce649633a339/dojo/templates/base.html#L618)  
3. Alpha extends `base.html`. Alpha has a [content block of its own in `finding_list.html`](https://github.com/DefectDojo/django-DefectDojo/blob/36a12fa4eb53d62ba3b339814c44ce649633a339/dojo/templates/dojo/findings_list.html#L5), so the code to include `finding_list_snippet.html` inside that content block will **REPLACE** the code inside the [content block of `base.html`](https://github.com/DefectDojo/django-DefectDojo/blob/36a12fa4eb53d62ba3b339814c44ce649633a339/dojo/templates/base.html#L618)

So where did the banner go? It was overwritten by the [content block in `finding_list.html`](https://github.com/DefectDojo/django-DefectDojo/blob/36a12fa4eb53d62ba3b339814c44ce649633a339/dojo/templates/dojo/findings_list.html#L5)

How can the banner not be overwritten? Add `{{block.super}}`!


Here is the order of templates rendered **WITH** `{{block.super}}`
1. `base.html`
2. Beta -> `custom_banner.html` extends `base.html`
3. Alpha -> `finding_list.html` extends `base.html`

Explaining each step:
1. `base.html` is rendered. Nothing special here
2. Beta extends `base.html`. Beta also has a content block of its own, but with `{{block.super}}` , so the code to make the banner in that block will **APPEND** the code inside the [content block of `base.html`](https://github.com/DefectDojo/django-DefectDojo/blob/36a12fa4eb53d62ba3b339814c44ce649633a339/dojo/templates/base.html#L618)  
3. Alpha extends `base.html`. Alpha has a [content block of its own in `finding_list.html`](https://github.com/DefectDojo/django-DefectDojo/blob/36a12fa4eb53d62ba3b339814c44ce649633a339/dojo/templates/dojo/findings_list.html#L5), but with `{{block.super}}`,  so the code to include `finding_list_snippet.html` inside that content block will **APPEND** the code inside the [content block of `base.html`](https://github.com/DefectDojo/django-DefectDojo/blob/36a12fa4eb53d62ba3b339814c44ce649633a339/dojo/templates/base.html#L618)

Now both the banner and the finding list are rendered inside the [content block of `base.html`](https://github.com/DefectDojo/django-DefectDojo/blob/36a12fa4eb53d62ba3b339814c44ce649633a339/dojo/templates/base.html#L618)

mtesauro:
Nice troll with the PR title @Maffooch - got me to read this one while I was eating lunch. ;-)

Read through the comments and looked at the Django docs.  I especially found [this bit](https://docs.djangoproject.com/en/3.2/howto/overriding-templates/#extending-an-overridden-template) helpful and, from my reading, I really like the idea of {{block.super}} - especially when I read this bit in the docs:

> Combined with {{ block.super }}, this can be a powerful way to make small customizations.

Sounds perfect for DefectDojo since we're usually pulled in multiple directions based on differing user-base needs.

Over the years, we've had template issues and it seems like this may also help a bit in that as well.

valentijnscholten:
The docs are clear as as said before. A discussed on Slack, the example above discussed 2 separate templates and the contents of banners.html will never end up in finding_list.html because there is no relationship between those templates.

The example from the docs pointed to by Matt are about adding block.super to a new template that overrides an existing template. That is already possible without this PR, no need to add block.super to 185 files.

I can see maybe some scenario's where templates are included in other templates that are extending base.html that some things could be overwritten. But maybe that sometimes is exactly the intention of the template author. I don't think there's an arguable benefit of adding block.super everywhere. I don't think it's common either. I think we should let de template authors decide if they want to overwrite or append depending on their needs. 

My impression is that a banner would go into its own block "banner" making it explicit and easy to overwrite / append / remove in subtemplates.

The fact that it seems hard to provide an example of what problem this PR makes me doubt the benefits.

But you're probbly going to find 2 approvers and I am not going to "fight" this PR as it won't really hurt my usecase and there lots of other things to work on to get 2.0.0 finished and tested on time.

mtesauro (approved):
Approved

devGregA (approved):

PR #4643, opened 2021-06-10
Fix generic parser import endpoint without protocol

EndPositive:
The generic importer uses Endpoint.from_uri() see https://github.com/DefectDojo/django-DefectDojo/blob/64701f1dac416b4baa08f9f1d4609ab0438481ef/dojo/tools/generic/parser.py#L49-L53, which relies on hyperlink.parse() introduced in #4473.

This library, however, does not support url's without protocol. Example: hyperlink.parse("www.google.com") creates an object with `host: None` and `path: "/www.google.com"`.

There exists a workaround for this in some places in the code like:

https://github.com/DefectDojo/django-DefectDojo/blob/64701f1dac416b4baa08f9f1d4609ab0438481ef/dojo/forms.py#L1365-L1370

This PR, brings the same capabilities to the generic importer.

### DIFF
--- a/dojo/tools/generic/parser.py
+++ b/dojo/tools/generic/parser.py
@@ -116,7 +116,9 @@ class GenericParser(object):
 
             # manage endpoints
             if 'Url' in row:
-                finding.unsaved_endpoints = [Endpoint.from_uri(row['Url'])]
+                finding.unsaved_endpoints = [Endpoint.from_uri(row['Url'])
+                                             if '://' in row['Url'] else
+                                             Endpoint.from_uri("//" + row['Url'])]
 
             # manage internal de-duplication
             key = hashlib.sha256("|".join([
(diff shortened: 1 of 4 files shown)


valentijnscholten:
@kiblik  wdyt?

kiblik:
The implementation looks fine to me. But unit tests should be added for all 3 cases.

EndPositive:
While adding the tests, I noticed that  `get_findings_csv()` also required the url fix, so I updated that one as well: ea87468fd927838d1a965a04e6f0531009dc5b65

kiblik:
> While adding the tests, I noticed that `get_findings_csv()` also required the url fix, so I updated that one as well: [ea87468](https://github.com/DefectDojo/django-DefectDojo/commit/ea87468fd927838d1a965a04e6f0531009dc5b65)

Old good unit tests :)

damiencarol (approved):

valentijnscholten (approved):

PR #4649, opened 2021-06-12
fix missing one-to-one relationships from prefetch API specs

valentijnscholten:
Due to a indentation bug the one-to-one relationships were missing from the API specs since #4541


valentijnscholten:
@StefanFl

StefanFl:
This should do it, thank you @valentijnscholten

Maffooch (approved):

PR #4656, opened 2021-06-14
Allow null-able CharFields to be blank

kiblik:
Avoid 
```json
{
  "xyz": [
    "This field may not be blank."
  ]
}
```
for `PATCH` and `PUT` - update field `xyz` to `""`


damiencarol:
For `Finding.cvssv3` , I don't think it's a good idea.

kiblik:
> For `Finding.cvssv3` , I don't think it's a good idea.

You are right. I reverted also some others

Maffooch (approved):

valentijnscholten (approved):

PR #4659, opened 2021-06-15
Set Product name non unique

dsever:
This pull request removed  requirement for Product name to be unique, and moves requirement to (product_type, name).

Why: there are situations in companies where  multiple product are named the same, but in different units, so we do believe that  is more natural to preserve  constraint on (name, prod_type)


valentijnscholten:
This will break everything in my company and probably elsewhere. Product name must be unique.

dsever:
@valentijnscholten  can we somehow test it, or can you explain internals that are using product_name as unique? I wanted to test upgrade, but I'm missing dataset that can be in use.

devGregA:
I agree with @valentijnscholten. I believe it is much easier to accommodate naming products in a unique manner rather than introducing confusion on metrics and matching filters.

valentijnscholten:
All my Jenkins pipeline scripts are finding product if by using product name, same for downstream dashboards and other data aggregation scripts. Of course I can fix those, but I'm pretty sure there will be many more people.

dsever:
Thanks guys, for your view. Than I will close this PR, we will than probably use it for our internal use case only.

Hijerboa:
Perhaps this is something that could be added as an optional setting which is disabled by default?

StefanFl:
I agree with @valentijnscholten and @devGregA. Since @dsever accepted their opinion, I close this PR.

PR #4671, opened 2021-06-17
Helm upgrade, config tracking

dsever:
This helm improvement introduces helm feature to track configurations and secrets, to avoid manual intervention in case of changing configuration of the deployments. 

How to enable feature:

`trackConfig=enabled (default disabled)`

system will track, if there is any change in config/secret and if there is, it will respin pods (django, celery).

e.g. 

`helm upgrade defectdojo  --set trackConfig=enabled --set-string extraConfigs.DD_FEATURE_AUTHORIZATION_V2=true .`


valentijnscholten:
Sounds look a good idea

madchap (approved):
lgtm, thanks @dsever

alles-klar (approved):

PR #4680, opened 2021-06-18
Trustwave Fusion API Parser

SupaJuke:
In contrary to the existing Trustwave parser that works in the CSV file, this parser supports the import of the reports that are directly from the Trustwave Fusion API, which are in the JSON format.


dojo/tools/trustwave_fusion_api/parser.py
```diff
@@ -0,0 +1,100 @@
...
+    date_str = vuln['updatedOn']
+    date_str = date_str[:len(date_str) - 3] + date_str[-2:]
+    finding.date = datetime.strptime(date_str, "%Y-%m-%dT%H:%M:%S.%f%z")
+
+    if "CVE-NO-MATCH" not in vuln['kb']['cves']:
+        cves = ""
+        for cve in vuln['kb']['cves']:
+            cves += f"{cve}, "
+        cves = cves[:len(cves) - 2]  # removing the comma and the blank space
+        finding.cve = cves
```
damiencarol, line 98: this can't work. the attribute `cve` can hold only one CVE for now :/

dojo/tools/trustwave_fusion_api/parser.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+
+    # Endpoint
+    if vuln['location']['url']:
+        endpoint = Endpoint.from_uri(vuln['location']['url'])
+        finding.saved_endpoints = [endpoint]
+
+    finding.title = vuln['name']
+    finding.description = vuln['classification']
+    finding.severity = f"{vuln['severity'][0]}{vuln['severity'][1:].lower()}"
```
damiencarol, line 86: Sould we use the method `.title()` here?
Also for other parsers we created a specific method to convert it with unit tests.
Could you do it for this parser?

damiencarol (changes requested):

dojo/tools/trustwave_fusion_api/parser.py
SupaJuke: I think it might be better to include all the cves because the link to the cves will still work: it will link you to the first cve provided.

SupaJuke: This way if the uploader wants to see other cves, they can still do it manually, and the information would not be lost.

valentijnscholten: Yeah, but you can't because the field is an integer field. You could add the other cves to the description.

valentijnscholten: Can you also add some vulnerabilities/testcases that actually have cves to the logic van be tested/asserted?

SupaJuke: Sure, will do that.

valentijnscholten:
any reason why you closed this @SupaJuke , I think it was close to ready?

damiencarol:
@SupaJuke if you think your PR need more work, you can mark it as 'draft'

PR #4681, opened 2021-06-18
feat: AuditJS parser added

SoaAlex:
AuditJS Scan parser.

Working for AuditJS using SonaType OSSIndex database with JSON output format.

I'm a beginner so there might be many things wrong that would be happy to solve if you tell me.


SoaAlex:
My parser.py file got unstaged right before I  commit, rewritting it again and pushing as soon as I can...

SoaAlex:
Spent a few minutes rewritting it. Run test again and it seems fine. Hope I didn't miss anything

dojo/settings/settings.dist.py
```diff
@@ -789,6 +789,7 @@ def generate_url(scheme, double_slashes, user, password, host, port, path, param
     'Anchore Engine Scan': ['title', 'severity', 'component_name', 'component_version', 'file_path'],
     'Anchore Grype': ['title', 'severity', 'component_name', 'component_version'],
     'Aqua Scan': ['severity', 'cve', 'component_name', 'component_version'],
+    'AuditJS': ['unique_id_from_tool', 'title', 'description', 'component_version'],
```
valentijnscholten, line 792: Unique id from tool must be unique, otherwise it's probably better to use vuln_id_from_tool

valentijnscholten (changes requested):

dojo/settings/settings.dist.py
SoaAlex: The ID provided by AuditJS is issued from SonaType OSS Index, which has unique IDs for the vulnerabilities in their database (like 3a3bf289-21dc-4c84-a46e-39280f80bb01)
So they are unique right ?

valentijnscholten: If they wereunique, you wouldn't have to use true other girls in the dedupe Algo. In theory a project can have multiple instances of the same vulnerability, making the id not unique inside the product/application.

SoaAlex: Alright thanks, I see.
Will change to vuln_id_from_tool then.

SoaAlex: Done.
Also tested importing the same scan file and the algorithm was able to find duplicates.

dojo/unittests/tools/test_auditjs_parser.py
```diff
@@ -37,19 +37,3 @@ def test_auditjs_parser_with_many_vuln_has_many_findings(self):
...
-    #         parser = AuditJSParser()
-    #         findings = parser.get_findings(testfile, Test())
-    #         testfile.close()
-    #         for finding in findings:
-    #             for endpoint in finding.unsaved_endpoints:
-    #                 endpoint.clean()
-    #         self.assertTrue(
-    #             "AuditJS report contains errors:" in str(context.exception)
-    #         )
-    #         self.assertTrue("ECONNREFUSED" in str(context.exception))
```
damiencarol: For this one you could provide a file which is not JSON.

valentijnscholten (approved):

dojo/fixtures/test_type.json
```diff
@@ -726,5 +726,12 @@
         },
         "model": "dojo.test_type",
         "pk": 195
+    },
+    {
+        "fields": {
+            "name": "AuditJS Scan"
+        },
+        "model": "dojo.test_type",
+        "pk": 196
```
damiencarol, line 735: This part is not needed. We load parser dynamically now.

dojo/tools/auditjs/parser.py
```diff
@@ -0,0 +1,125 @@
...
+                    if 'reference' in vulnerability:
+                        references = vulnerability['reference']
+
+                    finding = Finding(
+                        title=title,
+                        test=test,
+                        cve=cve,
+                        cwe=cwe,
+                        cvssv3=cvssVector,
+                        cvssv3_score=cvss,
```
damiencarol, line 102:
```suggestion
```
If you set the CVSS vector, the upper layers will manage the score automatically.

damiencarol (changes requested):
You should investigate using the module `cvss`. You re-implemented the algo of score and severity in the parser but we rely on module `cvss`.

dojo/tools/auditjs/parser.py
SoaAlex: Yes, that's right, but some findings in AuditJS do not have a CVSS Vector specified, so I prefer passing it anyway as the CVSS score is always there.

dojo/unittests/tools/test_auditjs_parser.py
SoaAlex: OK, I'll do that

dojo/fixtures/test_type.json
SoaAlex: OK removed. It was automatically generated by CookieCutter

SoaAlex:
At least I served as a bas practice example for future contributors 👍 
OK I'll change the code to use module CVSS.

dojo/tools/auditjs/parser.py
SoaAlex: Does DefectDojo compute severity automatically if I only pass the CVSS score and not vector ?
Therefore, I could delete the get_severity function (which I already simplified).

dojo/unittests/tools/test_auditjs_parser.py
```diff
@@ -0,0 +1,50 @@
...
+        parser = AuditJSParser()
+        findings = parser.get_findings(testfile, Test())
+        testfile.close()
+        for finding in findings:
+            for endpoint in finding.unsaved_endpoints:
+                endpoint.clean()
+        self.assertEqual(1, len(findings))
+        self.assertEqual("mysql", findings[0].component_name)
+        self.assertEqual("2.0.0", findings[0].component_version)
+        self.assertEqual(9.5, findings[0].cvssv3_score)
```
damiencarol, line 26: It's better to test that the vector is ok. the score is managed by upper layers if the vector is good.

damiencarol (changes requested):
The attribute `id` in the reports seems to be our attribute `unique_id_from_tool` instead of `vuln_id_from_tool`. So the internal de-duplication should be different.

SoaAlex:
> The attribute `id` in the reports seems to be our attribute `unique_id_from_tool` instead of `vuln_id_from_tool`. So the internal de-duplication should be different.

I previously changed from unique_id to vuln_id. I'll revert then.

dojo/unittests/tools/test_auditjs_parser.py
SoaAlex: I'll do that

SoaAlex:
For example, in auditjs_many_vulns, there are 2 vulnerabilities with the same ID in the same report (`3a5ca7f8-cd91-4681-a8b5-285ef96915f2`, CWE-200).
The reason AuditJS find 2 of them is because there are 2 versions of the same component but with the same vulnerability.
With the unique_id_from_tool version, it is considered duplicate, whereas with the hash_code, it doesn't because I take into account the component version.

What is the best approach ? Consider duplicate a vulnerability with 1 finding regardless of the version or have 2 findings ?

dojo/tools/auditjs/parser.py
damiencarol: For now , no. So the current version of the code is good.

damiencarol:
The best approach here is "Consider duplicate a vulnerability regardless of the version".
I mean the report have the same vulnerability 2x. It's ok and it happen when a project have multiple versions of the same lib that is impacted by the same security issue. So to have the parser reflect that is perfectly normal.
So I think you can use directly `unique_id_from_tool` as the key for the internal de-duplication or let the parser generate 2 findings to keep the link with the component+version *but* you will need to make sure that the de-duplication configuration rely only on `unique_id_from_tool` to mark the second vulnerability as a duplicate.

Keep in mind that the internal de-duplication feature that parsers implement exists to limit the number of findings that are duplicates. It reduces number of records in DB and the needs to process a lot of data with report that have many findings.

SoaAlex:
Thank you for your answer :) 
It confirms what I was thinking and I agree with you, it's better to use only `unique_id_from_tool`.

The only downside as you said is that it only records one version of the component, but it's worth the lighter workload for the parser and the DB.

Gonna switch back to `DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL` in next commit, and use `unique_id_from_tool` as dupe_key for internal de-duplication.

damiencarol (approved):

damiencarol:
Are we good or you wanna continue to work on this one? For me it's good enough.

SoaAlex:
> @SoaAlex Are we good or you wanna continue to work on this one? For me it's good enough.

It's good for me too. I don't see any improvments I could make for now.

damiencarol:
@SoaAlex ok, let's merge. it will be part of `2.0.0` so we can improve later with some first feedback.

PR #4689, opened 2021-06-20
Authorization V2: Frontend for Groups

Hijerboa:
Adds frontend UI features for the groups features added by @StefanFl. 

Adds:

- Group listing page
image
- Group view/edit/delete pages
image
- Ability to add Users, Products, and Product Types to groups with relevant permissions:
image
- Group listings to user, product, and product type pages
image
- Listing of groups a user is in to their profile page, so they can access them and edit them if they have the proper permissions
image

To do:

- [x] Implement UI for users to manage groups that are not staff


StefanFl:
Great contribution, thanks a lot @Hijerboa. A lot of stuff, will take a bit of time to review.

dojo/group/queries.py
```diff
@@ -44,8 +45,26 @@ def get_authorized_group_users(permission):
...
         return Dojo_Group_User.objects.all()
 
     groups = get_authorized_groups(permission)
-    return Dojo_Group_User.objects.filter(dojo_group__in=groups)
+    return Dojo_Group_User.objects.filter(group__in=groups)
+
+
+def get_authorized_group_users_for_user(user):
+    groups = get_authorized_groups(Permissions.Group_View)
+    groups = Dojo_Group_User.objects.filter(user=user, group__in=groups)
```
StefanFl, line 57: This would be even better if we sort by group name.

```suggestion
    groups = Dojo_Group_User.objects.filter(user=user, group__in=groups).order_by('group__name')
```

dojo/user/queries.py
```diff
@@ -50,3 +50,7 @@ def get_authorized_users_for_product_and_product_type(users, product, permission
             Q(id__in=product.prod_type.authorized_users.all()) |
             Q(is_superuser=True) |
             Q(is_staff=True))
+
+
+def get_groups_for_user(user):
```
StefanFl, line 55: This method is not needed anymore, after the suggested change in `user/views.py`

dojo/user/views.py
```diff
@@ -197,6 +199,7 @@ def alertcount(request):
 def view_profile(request):
     user = get_object_or_404(Dojo_User, pk=request.user.id)
     form = DojoUserForm(instance=user)
+    groups = get_groups_for_user(user)
```
StefanFl, line 202: There is already another method you can use:
```suggestion
    groups = get_authorized_group_users_for_user(user)
```

StefanFl:
Sorry guys, but I have just seen something that needs to be fixed. The name of the model class `Dojo_Group_User` is not good and should be changed to `Dojo_Group_Member`. Than it is similar to `Product_Member` and `Product_Type_Member` which implement the same concept. My fault. 

@Hijerboa I prepare the changes and give it you as another PR for your branch.

PR #4690, opened 2021-06-21
Trustwave Fusion API Parser

SupaJuke:
A parser for vulnerability reports from Trustwave Fusion's API


kiblik:
https://github.com/SupaJuke/django-DefectDojo/pull/1

dojo/tools/trustwave_fusion_api/parser.py
```diff
@@ -0,0 +1,110 @@
...
+        if file is None:
+            return None
+
+        data = file.read()
+        tree = json.loads(data)
+
+        if tree:
+            return self.get_items(tree, test)
+
+    def get_items(self, tree, test):
```
damiencarol, line 31:
```suggestion
        tree = json.load(file)
```

dojo/tools/trustwave_fusion_api/parser.py
```diff
@@ -0,0 +1,101 @@
...
+        cves = ""
+        for cve in vuln["kb"]["cves"]:
+            cves += f"{cve}, "
+        cves = cves[: len(cves) - 2]  # removing the comma and the blank space
+
+    finding.description = description + "; CVEs: " + cves
+    finding.severity = vuln["severity"].title()
+
+    # Date
+    date_str = vuln["updatedOn"]
```
damiencarol, line 97:
```suggestion
    date_str = vuln["createdOn"]
```
The date attribute is the date of creation. I think it's better to use "createdOn" here

damiencarol (changes requested):

damiencarol (approved):

damiencarol:
@SupaJuke it seems that the reports have data for component, see the attribute `"applicationCpe":"None",
             "applicationName":"None",` of the report .
I think you could improve this parser to support attribute `component_name` and `component_version` with this data.

dojo/tools/trustwave_fusion_api/parser.py
```diff
@@ -0,0 +1,105 @@
...
+    finding.description = description + "; CVEs: " + cves
+    finding.severity = vuln["severity"].title()
+
+    # Date
+    date_str = vuln["createdOn"]
+    date_str = date_str[: len(date_str) - 3] + date_str[-2:]
+    finding.date = datetime.strptime(date_str, "%Y-%m-%dT%H:%M:%S.%f%z")
+
+    # Component name
+    if vuln["location"]["applicationName"] != "None":
```
valentijnscholten, line 102: I see a lot of references to fields like `vuln["xxx"]["yyy"]` which could fail if xxx is not there or yyy is not there. Are we sure these are always there? I'd prefer to make it foolproof by checking first.
The `component_name` field currently in Defect Dojo is not populated with CPE but with the name, i.e. in https://github.com/valentijnscholten/django-DefectDojo/blob/2cd0de4e95290517428e98a2aa93744c89bc846e/dojo/tools/dependency_check/parser.py#L92 we parse the CPE and extract the name. Maybe in the future we need to add a cpe field to Defect Dojo.

valentijnscholten (changes requested):
see comment

dojo/tools/trustwave_fusion_api/parser.py
SupaJuke: Ok, I could definitely add a checker to check if the field exists and not "None."

damiencarol: I agree, also if the data is a CPE, parsing it with `cpe` module is the good way to do it.

dojo/unittests/tools/test_trustwave_fusion_api_parser.py
```diff
@@ -62,6 +62,7 @@ def test_parse_file_with_multiple_vuln_has_multiple_findings(self):
         self.assertEqual("Info", finding.severity)
         self.assertIsNone(finding.cve)  # should be none since CVE is "CVE-NO-MATCH"
         self.assertIsNone(finding.component_name)
+        self.assertIsNone(finding.component_version)
```
damiencarol: do the assertEqual here.
`self.assertEqual("comp", finding.component_name)`

dojo/tools/trustwave_fusion_api/parser.py
```diff
@@ -0,0 +1,115 @@
...
+
+def get_item(vuln, test):
+    finding = Finding(
+        test=test,
+        unique_id_from_tool=vuln["id"],
+        nb_occurences=1,
+    )
+
+    # Endpoint
+    if vuln["location"]["url"] and vuln["location"]["url"] != "None":
```
damiencarol, line 61:
```suggestion
    if "url" in vuln["location"] and vuln["location"]["url"] != "None":
```

SupaJuke:
According to https://www.flake8rules.com/rules/W503.html, the practice I am doing (that is violating the flake8 formatting) will soon be the best practice. Do I keep this or do I change it to one that would be considered anti-pattern?

dojo/unittests/tools/test_trustwave_fusion_api_parser.py
SupaJuke: There is no comp name here though: the parser will detect the applicationName (in the location field of the JSON file) being "None," so it will pick nothing up. Thus, the field here should be None. Am I missing something here though?

damiencarol: Sorry maybe not the best finding.
I saw another finding in the report that have component data.

damiencarol: if you take a look at:
https://github.com/DefectDojo/django-DefectDojo/blob/77bd0431df417ad1ca57cbd8447cee841a01e0a4/dojo/unittests/scans/trustwave_fusion_api/test_cve.json#L17
I can see that you have component name+version as CPE string.
You can do something like this:
```
             if "applicationCpe" in vuln["location"]:
                    cpe = CPE(vuln["location"]["applicationCpe"])
                    component_name = cpe.get_vendor()[0] + ':' if len(cpe.get_vendor()) > 0 else ''
                    component_name += cpe.get_product()[0] if len(cpe.get_product()) > 0 else ''
                    finding.component_name = component_name if component_name else None
                    finding.component_version = cpe.get_version()[0] if len(cpe.get_version()) > 0 else None
```

damiencarol:
@SupaJuke stick to the current validation configuration, it will be hard to change it and modify the code in the whole application to comply with this change.
We can modify the configuration of flake8 in another dedicated PR.

dojo/unittests/tools/test_trustwave_fusion_api_parser.py
SupaJuke: I see it now, thanks!

SupaJuke: Done!

dojo/tools/trustwave_fusion_api/parser.py
SupaJuke: Done!

damiencarol (approved):

damiencarol:
@valentijnscholten the modifications you requested has been done, could you take another look?

valentijnscholten (approved):

Issue #4730, opened 2021-06-29
Hadolint findings should be marked as static findings


christophe226:
**Bug description**
Hadolint parser has no specific configuration regarding static/dynamic status of its findings. So it uses the default status, i.e. "dynamic", whereas Hadolint is a SAST, not a DAST. For consistency, Hadolint findings should be "static", not "dynamic". This wrong "dynamic" value has impacts on deduplication.

**Steps to reproduce**
1. Import Hadolint report of your choice into DefectDojo
2. Go to a newly created Hadolint finding: it is marked as "Dynamic"

**Expected behavior**
On step 2. I should see "Static" for every Hadolint finding.

**Deployment method** *(select with an `X`)*
- [X] Kubernetes
- [ ] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: Ubuntu 18.04

valentijnscholten:
next step is to make a PR here upstream? :-)

StefanFl:
PR has been merged

PR #4738, opened 2021-06-29
Mark Hadolint findings as static

christophe226:
Fix issue #4730.

Issue description: Hadolint parser has no specific configuration regarding static/dynamic status of its findings. So it uses the default status, i.e. "dynamic", whereas Hadolint is a SAST, not a DAST. For consistency, Hadolint findings should be "static", not "dynamic". This wrong "dynamic" value has impacts on deduplication.


damiencarol (changes requested):
@christophe226 could you rebase your PR on `dev`? it seems you have some modifications not linked to this PR in your branch

PR #4753, opened 2021-06-30
Enable to specify SAML2 mapping using env variables

dsever:
1. Problem:
I can't find the better way how to specify  SAML attrib mappins, in current format (for djangosaml2):

`SAML_ATTRIBUTE_MAPPING = {
    'uid': ('username', ),
    'mail': ('email', ),
    'cn': ('first_name', ),
    'sn': ('last_name', ),
}`

using env variables, specially this is  important in case of k8s deployment type.

2. Workaround:
Usage of old way of mapping, than in configuration to use additional function to rewrite it into the new format.

Old format: 
`
DD_SAML2_ATTRIBUTES_MAP=(dict, {
        # Change Email/UserName/FirstName/LastName to corresponding SAML2 userprofile attributes.
        # format: SAML attrib:django_user_model
        'Email': 'email',
        'UserName': 'username',
        'Firstname': 'first_name',
        'Lastname': 'last_name'
    }),
`
and linux env var representation:
`
DD_SAML2_ATTRIBUTES_MAP=  email=email,username=User.username,first_name=first_name,last_name=last_name
`
Additionally in this PR additional logger for SAML is added, and additional switch:

`DD_SAML2_ALLOW_UNKNOWN_ATTRIBUTE`, default False

That allows attributes that are not defined in *attribute_maps* to be parsed.


dojo/settings/settings.dist.py
```diff
@@ -119,11 +119,13 @@
     DD_SAML2_CREATE_USER=(bool, False),
     DD_SAML2_ATTRIBUTES_MAP=(dict, {
         # Change Email/UserName/FirstName/LastName to corresponding SAML2 userprofile attributes.
-        'Email': ('email', ),
-        'UserName': ('username', ),
-        'Firstname': ('first_name', ),
-        'Lastname': ('last_name', )
+        # format: SAML attrib:django_user_model
```
alles-klar, line 122: @Maffooch I think you had two variables on the right side, will this solution work for you too? Or do you use local_settings.py or another way?

dojo/settings/settings.dist.py
```diff
@@ -697,10 +710,10 @@ def generate_url(scheme, double_slashes, user, password, host, port, path, param
     SAML_LOGOUT_REQUEST_PREFERRED_BINDING = saml2.BINDING_HTTP_POST
     SAML_IGNORE_LOGOUT_ERRORS = True
     SAML_DJANGO_USER_MAIN_ATTRIBUTE = 'username'
-    # SAML_DJANGO_USER_MAIN_ATTRIBUTE_LOOKUP = '__iexact'
+    SAML_DJANGO_USER_MAIN_ATTRIBUTE_LOOKUP = '__iexact'
```
alles-klar, line 713: Would be great to have this configurable and disable it by default, otherwise it may cause a security problem.

alles-klar:
Good Idea @dsever!

Maybe we should also extend the documentation and add a part about the linux env var representation.

dojo/settings/settings.dist.py
dsever: Yes, sure my mistake.

dojo/settings/settings.dist.py
Maffooch: I use local settings, so this can easily be overwritten there. I wonder if the env var mapping can be done with something like this to be consistent with the current  tuple mapping model.
```
DD_SAML2_ATTRIBUTES_MAP=(dict, {
         'Email': ['email'],
         'UserName': ['username'],
         'Firstname': ['first_name'],
         'Lastname': ['last_name']
 }),
```

dsever: I assume it is the question of capabilities of https://django-environ.readthedocs.io/en/latest/, I was looking at the module, but not use is this the case

alles-klar:
@dsever is it still a draft?

dsever:
> @dsever is it still a draft?

If you agree with this approach, I can mark it as official PR.

alles-klar:
> > @dsever is it still a draft?
> 
> If you agree with this approach, I can mark it as official PR.

It is the best option we have at the moment, so yes :)

alles-klar:
@dsever Can you please update the documentation too?

dsever:
> @dsever Can you please update the documentation too?

Sure, will do it.

Maffooch (approved):

alles-klar (approved):

alles-klar:
Thanks @dsever!

PR #4756, opened 2021-06-30
Fix Github Issue Creation

Maffooch:
There is no `GitHub-body.tpl' file (and never was from what it looks like) so use the Jira one instead since it already does a good job at listing all pertinent info.


devGregA (approved):

mtesauro (approved):
Approved

PR #4758, opened 2021-07-01
Reopen JIRA ticket when an original finding is reopened

FallenAtticus:
This change will force the original finding, having a jira issue linked to it, to update the jira issue as well. Having it that way, the linked jira issue of the orignial finding will be re-opened as well.

### DIFF
--- a/dojo/utils.py
+++ b/dojo/utils.py
@@ -369,7 +369,7 @@ def set_duplicate_reopen(new_finding, existing_finding):
     existing_finding.verified = new_finding.verified
     existing_finding.notes.create(author=existing_finding.reporter,
                                     entry="This finding has been automatically re-openend as it was found in recent scans.")
-    existing_finding.save()
+    existing_finding.save(push_to_jira=True)
 
 
 def do_apply_rules(new_finding, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


Maffooch (changes requested):
I don't believe this function is JIRA specific. It should not assume that it is. There are some JIRA helper methods at `dojo/jira_link/helper.py` to identify if there is a valid JIRA connection associated with the finding.

Maffooch:
@FallenAtticus Any thoughts on my comment? Will close this in a week otherwise.

valentijnscholten:
I think if a finding gets reopened in JIRA, it makes sense to also reopen it in JIRA. Using save to jira is probably the generic approach to get the issue in jira updated from defect dojo.

Maffooch:
@valentijnscholten doesn't that function reopen duplicates? I don't think duplicates should be pushed to JIRA. I may be misinterpreting here though.

valentijnscholten:
This code is called when a new finding is imported that is a duplicate of an existing (original) finding, which is closed/mitigated. In this case the original gets reactivated. If that original has a jira issue linked, it makes sense to reopen that jira issue as well. But we can't just always say `push_to_jira=True` as it could create a new jira issue if there is none. Also this push to jira is not explicitly chosen by the user, so we may have to push only the status to JIRA and not overwrite the description etc by pushing from save()

PR #4759, opened 2021-07-01
Downgrade to Ubuntu 18.04

dsever:
Downgrade ubuntu to 18.04 in GH k8s actions

### DIFF
--- a/.github/workflows/k8s-testing.yml
+++ b/.github/workflows/k8s-testing.yml
@@ -100,7 +100,7 @@ jobs:
   setting_minikube_cluster:
     name: Kubernetes Deployment
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
 
     needs: build
 


StefanFl:
Hi @dsever, great to see the GH actions working again. Just out of interest: Why don't we go to Ubuntu 20.04, which is also an LTS version?

dsever:
> Hi @dsever, great to see the GH actions working again. Just out of interest: Why don't we go to Ubuntu 20.04, which is also an LTS version?

To be honest have no idea why 20.04 in combination with minikube is causing  this issues. So quick win to enable k8s deployment test is to downgrade ubuntu version for the runners

StefanFl:
Ok, sounds good to me

StefanFl (approved):
Ubuntu 18.04 will get updates for another 2 years. Enough time to refine the solution.

PR #4761, opened 2021-07-01
Fixed CWE integer parsing problem.

kiliczsh:
CWE value was preventing for successful import, fixed with removing several unnecessary code.


damiencarol:
@kiliczsh could you add more checks in the unit tests that check the date and CWE ?

damiencarol (approved):

valentijnscholten (approved):

PR #4766, opened 2021-07-01
UI: Make all select fields searchable

kiblik:


Maffooch:
This is really cool! As expected, most of the unit tests were likely not expecting anything like this, so fixing those will be a large task. Very well worth it though!

damiencarol (approved):

Maffooch (approved):

PR #4770, opened 2021-07-02
Dockle parser findings marked static and dynamic

StefanFl:
Findings from the Dockle parser are marked as static and dynamic findings, which causes problems with reimports. This PR marks the findings as static only.

Additionally I removed the prefix "Found " from the title, as it doesn't add any information but makes the titles more crowded and distracts from the useful information.


mtesauro (approved):
Approved

valentijnscholten (approved):

PR #4783, opened 2021-07-07
Added force password reset functionality

blakeaowens:
Adds functionality to force a given user to reset their password upon login


valentijnscholten:
I'm surprised this isn't built-in to Django or some existing library/app.

StefanFl:
@blakeaowens I like this functionality very much. There is one thing I can't find: How do you actually force a password for a user? I see no call of `enable_force_password_reset`.

blakeaowens:
> @blakeaowens I like this functionality very much. There is one thing I can't find: How do you actually force a password reset for a user? I see no call of `enable_force_password_reset`.

Hello Stefan, sorry for my delayed response. There is a boolean field within UserContactInfo for enabling a forced password reset. This can be enabled within the AddDojoUser/EditDojoUser forms.

The enable_force_password_reset helper was added for rate limiting, which is not covered here, but in the PR mentioned above by @Maffooch.

dojo/models.py
```diff
@@ -378,6 +392,7 @@ class UserContactInfo(models.Model):
     slack_username = models.CharField(blank=True, null=True, max_length=150, help_text="Email address associated with your slack account", verbose_name="Slack Email Address")
     slack_user_id = models.CharField(blank=True, null=True, max_length=25)
     block_execution = models.BooleanField(default=False, help_text="Instead of async deduping a finding the findings will be deduped synchronously and will 'block' the user until completion.")
+    force_password_reset = models.BooleanField(default=False, help_text='Forces this user to reset password upon first sign in.')
```
StefanFl, line 395: @blakeaowens Might be a bit picky, but shouldn't the help_text say "Forces this user to reset password at next sign in."? The flag can be set for existing users as well, not only for new users.

blakeaowens: Sure, I will make that change!

dojo/settings/settings.dist.py
```diff
@@ -463,6 +463,30 @@ def generate_url(scheme, double_slashes, user, password, host, port, path, param
 
 LEGACY_API_V1_ENABLE = env('DD_LEGACY_API_V1_ENABLE')
 
+AUTH_PASSWORD_VALIDATORS = [
```
StefanFl, line 466: @blakeaowens This is the first time DefectDojo enforces password rules, hooray. We should have some form of documentation about that so users are not confused, either in our GitHub pages docs or in the dialogues where users can set a password.

dojo/db_migrations/0115_usercontactinfo_force_password_reset.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.1.13 on 2021-07-07 04:25
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dojo', '0114_cyclonedx_vuln_uniqu'),
```
StefanFl, line 9: @blakeaowens We have merged 2 PRs with db migrations, the last one is `0116_test_type_active`. I am afraid you have to rename this file and update the dependency.

blakeaowens: Yes, I see that now. Will take care of it!

devGregA (approved):

mtesauro (approved):
Approved

devGregA:
@blakeaowens  good stuff,  and this works / I'm okay with merging. We may want to consider allowing users to customize their policies... but maybe we shouldn't give people the option to make things less secure. Just a thought.

PR #4784, opened 2021-07-07
Add Rate Limiting to Login Page

Maffooch:
Dependent on #4783

Add the ability to rate limit the login page. 

Uses https://github.com/jsocol/django-ratelimit

### DIFF
--- a/docs/content/en/integrations/rate_limiting.md
+++ b/docs/content/en/integrations/rate_limiting.md
@@ -46,7 +46,7 @@ In the event of a brute force attack, a users credentials could potentially be c
 
 In an attempt to circumvent that event, setting `DD_RATE_LIMITER_ACCOUNT_LOCKOUT` will force a user to reset their password upon the next attempted login. 
 
-#### Multi-Process Behvaior
+#### Multi-Process Behavior
 
 When using configurations with multiple uwsgi processes, the rate limiting package uses the default cache that is memory based and local to a process.
 
--- a/docs/content/en/integrations/rate_limiting.md
+++ b/docs/content/en/integrations/rate_limiting.md
@@ -46,6 +46,10 @@ In the event of a brute force attack, a users credentials could potentially be c
 
 In an attempt to circumvent that event, setting `DD_RATE_LIMITER_ACCOUNT_LOCKOUT` will force a user to reset their password upon the next attempted login. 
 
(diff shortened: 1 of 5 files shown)


valentijnscholten:
Did you look into if / how this would work if you have multiple django containers/pods or when using multiple uwsgi processes? I think we need to explain in the docs what works and doesn't work.

Maffooch:
Didn't know that was even possible. How can I go about setting up an environment to test that?

valentijnscholten:
The docker compose in release mode already has 2 processes: https://github.com/valentijnscholten/django-DefectDojo/blob/d30234cba2278b73848c8f1f85362a0adbd24c21/docker/entrypoint-uwsgi.sh#L36
I _think_ the default cache used by the ratelimiter is not cross-process, typically people use some external cache like memcached or redis for that.

Maffooch:
It appears to block requests on a per process basis (guessing based on the 500 followed by 2 200's)
image

Here is what the documentation has to say about caching
```
After activating django-ratelimit, you should ensure that your cache backend is setup to be both persistent and work across multiple deployment worker instances (for instance UWSGI workers). Read more in the Django docs on caching.
```

Does dojo use any sort of caching method out of the box?

valentijnscholten:
No, it only uses the `default` cache, which is in memory and local to a process/worker. That's also the reason why we can't easily add query caching for example or other forms of global caching. If we were to switch to `redis` as default celery broker, we could also use `redis` as a cross worker/process cache. I think for it's best to just make the limitation very clear in our docs.
Maybe something could be achieved by doing sticky sessions in `nginx`.

docs/content/en/integrations/rate_limiting.md
```diff
@@ -46,6 +46,10 @@ In the event of a brute force attack, a users credentials could potentially be c
 
 In an attempt to circumvent that event, setting `DD_RATE_LIMITER_ACCOUNT_LOCKOUT` will force a user to reset their password upon the next attempted login. 
 
+#### Multi-Process Behvaior
```
valentijnscholten, line 49:
```suggestion
#### Multi-Process Behavior
```

valentijnscholten (changes requested):

devGregA (approved):

mtesauro (approved):
Approved.

PR #4790, opened 2021-07-08
APIv2: Users: multiple usernames search

kiblik:


dojo/filters.py
```diff
@@ -1997,11 +1998,30 @@ class UserFilter(DojoFilter):
...
         model = Dojo_User
         fields = ['is_staff', 'is_superuser', 'is_active', 'first_name',
-                  'last_name', 'username']
+                  'last_name', 'username', 'email']
         exclude = ['password', 'last_login', 'groups', 'user_permissions',
                    'date_joined']
 
 
+class ApiUserFilter(DojoFilter):
+    username = ModelMultipleChoiceFilter(field_name='username', to_field_name='username',
```
valentijnscholten, line 2007: It would be better to have some free form search where a search can be done on any username or part thereof. With the ModelMultipleChoiceFilter you will get a 400 or 500 error if you put in a username that doesn't exist.

valentijnscholten (changes requested):

dojo/filters.py
kiblik: @valentijnscholten, I'm not sure what kind of search field you expect. How it could cover multiple usernames in one request.

Maffooch: I believe he is asking for just a text field rather than a dropdown to select the user. These filters work with http query so an invalid query would result in a 4/500.

Heres an example `username = CharFilter(field_name='username', lookup_expr='icontains')` so that if I query for a user 'kib', then your username will be returned. If there are multiple users with 'kib' in their username, then they will also appear in the filtered results.

valentijnscholten: I think you could achieve this by copying the CharFieldInFilter I implemented for multiple tag search:
```
    tags = CharFieldInFilter(field_name='tags__name', lookup_expr='in',
                             help_text='Comma seperated list of exact tags')
```

Maffooch:
Couple errors here
```
2021-08-28T18:53:06.0484262Z [36muwsgi_1         |[0m ======================================================================
2021-08-28T18:53:06.0485560Z [36muwsgi_1         |[0m FAIL: test_user_multiple_usernames (dojo.unittests.test_rest_framework.UsersTest)
2021-08-28T18:53:06.0486824Z [36muwsgi_1         |[0m ----------------------------------------------------------------------
2021-08-28T18:53:06.0487739Z [36muwsgi_1         |[0m Traceback (most recent call last):
2021-08-28T18:53:06.0488840Z [36muwsgi_1         |[0m   File "/app/dojo/unittests/test_rest_framework.py", line 1261, in test_user_multiple_usernames
2021-08-28T18:53:06.0490007Z [36muwsgi_1         |[0m     self.assertEqual(r.json()['count'], 2, r.content[:1000])
2021-08-28T18:53:06.0491108Z [36muwsgi_1         |[0m AssertionError: 0 != 2 : b'{"count":0,"next":null,"previous":null,"results":[]}'
2021-08-28T18:53:06.0491911Z [36muwsgi_1         |[0m 
2021-08-28T18:53:06.0492730Z [36muwsgi_1         |[0m ----------------------------------------------------------------------
2021-08-28T18:53:06.0493761Z [36muwsgi_1         |[0m Ran 1576 tests in 1142.082s
2021-08-28T18:53:06.0494368Z [36muwsgi_1         |[0m 
2021-08-28T18:53:06.0495109Z [36muwsgi_1         |[0m FAILED (failures=1, skipped=189)
```

PR #4791, opened 2021-07-08
Languages overhaul

StefanFl:
DefectDojo has a hidden gem, being able to show the languages used in a project including their lines of code. Although the model classes and the display on the production details page are already there, there were no means to fill it with data. This PR:

- Provides the standard API methods for model classes `Languages` and `Language Type`
- Provides another API method to be able to import languages for a project from a JSON file produced by https://github.com/AlDanial/cloc
- Updates the fixture for language types with all languages and their colors from https://github.com/github/linguist/blob/master/lib/linguist/languages.yml. All id's from the previous fixture are not changed.
- Implements a db migration script to load the fixture, so that existing installations will get all new language types
- Implements a management command to update language types if desired
- Implememts some slight enhancements for showing the languages
- Provides unit tests and documentation


mtesauro (approved):
I love this idea and the PR to match the idea.

You addressed a shortfall I noticed a while ago but never did anything about so THANKS.

Approved.

devGregA (approved):

damiencarol:
@StefanFl this is a good patch.
I was wondering why you *mocked* the add_language method for Checkmarx parser and I saw why.
I have an idea, maybe we can modify a little bit what is returned by the parsers to return stats about languages (like Checkmarx, Semgrep, ...) and use your modifications in the *importer* part. This could be a clean way to export the language data from parsers.

damiencarol (approved):

StefanFl:
@damiencarol I understand your idea, but I can't see enough information coming from the parsers. Checkmarx returns the language per finding, but you don't get the full picture which languages are used in the project including lines of code. Same for semgrep.

PR #4792, opened 2021-07-09
Skip false positives from detect-secrets report

pablosantiagolopez:
This Pull Request includes a fix in [detect-secrets](https://github.com/Yelp/detect-secrets) parser to skip the findings established as False Positive. For that, I check the value of the `is_secret` field:

- If the `is_secret` field is not present, the finding could be a real secret.
- If the `is_secret` field it's true, the finding it's a real secret.
- If the `is_secret` field it's false, the finding it's a false positive.

The reference to the `is_secret` field can be checked in this [link](https://github.com/Yelp/detect-secrets/blob/b206dc29f274839ebd3c522fed19448c9b8f3f70/detect_secrets/core/potential_secret.py#L40).

### DIFF
--- a/dojo/tools/detect_secrets/parser.py
+++ b/dojo/tools/detect_secrets/parser.py
@@ -25,8 +25,6 @@ class DetectSecretsParser(object):
             find_date = dateutil.parser.parse(data.get('generated_at'))
         for detect_file in data.get('results'):
             for item in data.get('results').get(detect_file):
-                if 'is_secret' in item and item['is_secret'] is False:
-                    continue
                 type = item.get('type')
                 file = item.get('filename')
                 hashed_secret = item.get('hashed_secret')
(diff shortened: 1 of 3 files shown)


valentijnscholten:
I wonder if these should be skipped, or imported as false positive?
Also, could you add/update the testcases to reflect the change?

damiencarol:
@valentijnscholten I think it's better to import them as false positives.
We can have this use case:
1. first import with finding: is_secret is not here because the system detected it
2. second report with same finding but `is_secret` is `true` because it was tagged as false positive in the tool

Could be better for the logic to compare with the same findings.

damiencarol (changes requested):
I think it's better to import them as false positives.
We can have this use case:

    first import with finding: is_secret is not here because the system detected it
    second report with same finding but is_secret is true because it was tagged as false positive in the tool

Could be better for the logic to compare with the same findings.

Could you modify your PR to use the attribute false_p instead of skipping the finding?

pablosantiagolopez:
Solved! Thank you for your feedback @valentijnscholten @damiencarol

damiencarol (approved):

valentijnscholten (approved):

PR #4793, opened 2021-07-09
Add flag to enable/disable a parser

damiencarol:
Add a flag `active` in `Test_Type` entity to disable or enable a parser. `True` by default.

I made the modification aligned with the current behavior, i.e. default to "all parser enabled", but it will be easy to block one parser by API or directly in DB for now.

### DIFF
--- a/dojo/db_migrations/0115_test_type_active.py
+++ b/dojo/db_migrations/0115_test_type_active.py
@@ -1,0 +1,17 @@
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dojo', '0114_cyclonedx_vuln_uniqu'),
+    ]
+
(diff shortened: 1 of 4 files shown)


damiencarol:
@devGregA it is the mandatory base to move forward with this feature like add UI and/or a way to do it by the settings, etc...

Maffooch (approved):

StefanFl:
@damiencarol I would have expected that an inactive parser won't appear in the list of choices. I can't see if this happens.

damiencarol:
@StefanFl this part is hard. Because to do that we need to load from DB as the structure to load it is used in serializers.
I'm working on a more global patch that simplify things. This PR is a thread-off to have the feature without waiting 6 months.

StefanFl (approved):

PR #4794, opened 2021-07-09
Enable import history by default

valentijnscholten:
Was implemented 3-4 months ago, got some bugfixes so should be good to enable by default now.


damiencarol (changes requested):
your base seems not up-to-date

valentijnscholten:
yeah I messed up, fixed now @damiencarol

Maffooch (approved):

damiencarol (approved):

PR #4804, opened 2021-07-10
Remove Finding Images/Migrate to Files

Maffooch:
Remove Finding Images. This was intended to happen before 2.0.0, but it's a bit late. 

Changes:
* Replace Finding Images with File Uploads
* Files that are images will appear on reports and in Jira tickets
* Refactor Product Type reports as it was messy and broke the format a bit
* Add volume to celery worker in compose to allow access to the media file (pushing images to jira)


devGregA (changes requested):
@Maffooch can we use a template tag / snippet for the repeated template code or no?

devGregA (approved):

mtesauro (approved):
Approved

PR #4823, opened 2021-07-15
Fix file upload on Kubernetes deployment

TheocharisPetros:
Currently on Kubernetes deployment (via the Official HELM chart) the file and image upload functionalities do not work.

Once you upload an image or a file, the document is stored on the /app/media path of the uwsgi server but it is not accessible by the Nginx component (serving the file via the /usr/share/nginx/html/media path) resulting in error 404 once you try to view the image.

The above issue happens because there is no shared volume between uwsgi container and the Nginx container, resulting to Nginx not been able to fetch and serve the files.

### DIFF
--- a/helm/defectdojo/templates/django-deployment.yaml
+++ b/helm/defectdojo/templates/django-deployment.yaml
@@ -46,6 +46,8 @@ spec:
       volumes:
       - name: run
         emptyDir: {}
+      - name: media
+        emptyDir: {}
       {{- if  .Values.django.uwsgi.certificates.enabled }}
       - name: cert-mount
         configMap:
@@ -90,6 +92,8 @@ spec:
         volumeMounts:
         - name: run
           mountPath: /run/defectdojo
+        - name: media
+          mountPath: /app/media/
         {{- if  .Values.django.uwsgi.certificates.enabled }}
         - name: cert-mount
           mountPath: {{ .Values.django.uwsgi.certificates.certMountPath }}
@@ -169,6 +173,8 @@ spec:
         volumeMounts:
         - name: run
           mountPath: /run/defectdojo
+        - name: media
+          mountPath: /usr/share/nginx/html/media
         ports:
         - name: http
           protocol: TCP


valentijnscholten:
@dsever @alles-klar @madchap

dsever:
This is true...but there is bigger problem, emptyDir volume type doesn't ensure persistence, so definitely data integrity is questionable.

> An emptyDir volume is first created when a Pod is assigned to a node, and exists as long as that Pod is running on that node.

So basically it has to be multi read/write volume type (like NFS) for multipod implementation or e.g. cinder type for single pod (than it kills cloud native principles), or at the other way  your application will not have unique view of the data, and you will have probable E500 because of broken paths. 

I would recommend to postpone this PR, and take some focus to solve multi read/write for the volumes (I started some activities but never finished :( )

TheocharisPetros:
> This is true...but there is bigger problem, emptyDir volume type doesn't ensure persistence, so definitely data integrity is questionable.
> 
> > An emptyDir volume is first created when a Pod is assigned to a node, and exists as long as that Pod is running on that node.
> 
> So basically it has to be multi read/write volume type (like NFS) for multipod implementation or e.g. cinder type for single pod (than it kills cloud native principles), or at the other way your application will not have unique view of the data, and you will have probable E500 because of broken paths.
> 
> I would recommend to postpone this PR, and take some focus to solve multi read/write for the volumes (I started some activities but never finished :( )

You are completely right, we have already identified the issue and I will update my pull request with the solution.

TheocharisPetros:
In order to provide persistence storage capabilities for uploaded files and images the HELM chart has been updated.

A PVC is created based on the values defined under “django.uwsgi.persistentVolume.”
•	Enabled
•	Size 
•	StorageClass

The relevant additions have been added into the Django.deployment.yaml . Please note that in case django.uwsgi.persistentVolume.enabled is set to “false” an empyDir is created instead (no persistence of data).
A storage class provider is used; thus the solution is compliant with Cloud Native principles since the defined storage class is handled by the relevant provisioner.

In order to retain the data in case of upgrade or uninstall, a HELM annotation has been hardcoded to the PVC ("helm.sh/resource-policy": keep). I reviewed the project’s HELM deployment documentation and found out that you were not able to use the relevant annotation on MySQL. On our case the annotation works properly (tested), I will try to review the issue for MySQL once I have some free time and provide a solution.

Please let me know if comments are required to the values.yamls and relevant deployment documentation.

@dsever please for your input!

helm/defectdojo/templates/pvc.yaml
```diff
@@ -0,0 +1,19 @@
...
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  annotations:
+    "helm.sh/resource-policy": keep
+  name: {{ include "defectdojo.fullname" . }}-media
+spec:
+  accessModes:
+  - ReadWriteOnce
+{{- with .Values.django.uwsgi.persistentVolume.storageClass }}
```
dsever, line 12: Why this when .Values.django.uwsgi.persistentVolume.storageClass is not list or array...

helm/defectdojo/templates/pvc.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.django.uwsgi.persistentVolume.enabled -}}
+{{- if not .Values.django.uwsgi.persistentVolume.existingClaim -}}
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  annotations:
+    "helm.sh/resource-policy": keep
+  name: {{ include "defectdojo.fullname" . }}-media
+spec:
+  accessModes:
```
dsever, line 10: Overhere, in case of multiple replicas of django pod, there will be an issue. By this policy you are not able to write read to the same volume from multiple pods.

TheocharisPetros: Updated accessMode to "ReadWriteMany".

helm/defectdojo/templates/pvc.yaml
TheocharisPetros: Changed to storageClassName: "{{ .Values.django.uwsgi.persistentVolume.storageClass }}

helm/defectdojo/templates/pvc.yaml
dsever: It is not so simple just to convert into RWM, because underlay needs to support it, that is the problem, that is the reason I mentioned NFS, there is no clear solution without StorageClass that supports RWM. I assume for cloud solution like Google or AWS it is already there but for local installation it is responsibility of k8s admin to come up with something.   

@madchap @alles-klar  what do you think?

TheocharisPetros: You are right but I believe the usage of storageClass with any kind of internal or external provisioner should be at least an option if not the "default solution".

alles-klar: We have to mention that in the docs and provide some documentation but then it should be ok.

alles-klar:
A complete different approach and a bit out of scope but if you don't know it: https://django-storages.readthedocs.io/en/latest/index.html would be also an option.

TheocharisPetros:
Hello,

A comment has been added above the persistent volume of uswgi with a quick description.

In addition, there is an environmental variable for the definition of MEDIA_ROOT used in order to define the path in which the uploaded_files will be stored.

Currently if the environmental variable DD_MEDIA_ROOT was defined into the extraConfigs it would not change the path which would default to /media/app.

"kindIs "invalid" function has been used in order to exclude possible nil issues.

@alles-klar , @dsever  please for your input/review .

TheocharisPetros:
> A complete different approach and a bit out of scope but if you don't know it: https://django-storages.readthedocs.io/en/latest/index.html would be also an option.

Thank you for sharing!

On a quick review I did it does not support NFS and based on the developer's feedback FTP does not work properly with large files.

Based on the above, I am not sure if it would be a practical solution for an on premise installation.

helm/defectdojo/values.yaml
```diff
@@ -145,6 +145,12 @@ django:
       periodSeconds: 10
       successThreshold: 1
       timeoutSeconds: 5
+    # Persistent Volume on UWSGI is used in order to store files uploaded to defectdojo
+    # In case the persistentVolume is not enabled an emptyDir will be created and uploaded files will be deleted in case the pod is removed   
+    persistentVolume:
```
alles-klar, line 150: Would be good to have .Values.django.uwsgi.persistentVolume.existingClaim here, at least commend out.

helm/defectdojo/templates/django-deployment.yaml
```diff
@@ -90,6 +97,14 @@ spec:
...
         - name: run
           mountPath: /run/defectdojo
+        - name: media
+        {{- if (kindIs "invalid" .Values.extraConfigs) }}
+          mountPath: /app/media
+        {{else if not (kindIs "invalid" .Values.extraConfigs.DD_MEDIA_ROOT)}}
+          mountPath: {{.Values.extraConfigs.DD_MEDIA_ROOT}}
+        {{else}}
+          mountPath: /app/media
+        {{end}}
```
alles-klar, line 107: looks a bit complex, the following should have the same meaning:
```
        - name: media
        {{if not (kindIs "invalid" .Values.extraConfigs.DD_MEDIA_ROOT)}}
          mountPath: {{.Values.extraConfigs.DD_MEDIA_ROOT}}
        {{else}}
          mountPath: /app/media
        {{end}}
```

TheocharisPetros: My initial idea was to validate the DD_MEDIA_ROOT presence as you exactly mentioned but a nil error would come up in case "extraConfigs" was commented out or missing in the values.yaml.

This is the reason I followed the committed  solution. 

Please for your input :).

alles-klar: got it :)

helm/defectdojo/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "2.1.0-dev"
 description: A Helm chart for Kubernetes to install DefectDojo
 name: defectdojo
-version: 1.6.10
+version: 1.6.11
```
madchap, line 5: This probably warrants a bump to 1.7.0 in my opinion.

madchap (changes requested):
This is a great add, and certainly a good iteration. May I ask that you write a bit of documentation in KUBERNETES.md, under `Kubernetes Production`? Feel free to write a bit around the caveats you've had, or even maybe the actual implementation that you tested with so that readers get a good feel for it :-) Thanks!

Maffooch:
@TheocharisPetros please see conflicts

PR #4837, opened 2021-07-19
Helm extra volumes

dsever:
This basically gives option to replace or add particular files into  containers, including celery. 

So consumer of the helm can create secret or configMap from local file or just mount hostPath (not recommended), as k8s resource, and reference it using extraVolumes. 

This is usecase when dd k8s operator wants to (for. e.g.):
- replace CSS (like branding)
- inject local_setting.py to tune DD when something is not exposed using env. var....

@alles-klar  @madchap 

What do  you think?


SoaAlex:
I was exactly looking for this feature. Really nice work thanks

helm/defectdojo/values.yaml
```diff
@@ -109,6 +109,39 @@ celery:
       # concurrency: 8
       # prefetch_multiplier: 128
 
+  # A list of extra volumes to mount. This
```
madchap, line 112: Would it make sense to copy this whole help into a separate file and reference here, especially since it is put in twice?

madchap (approved):
lgtm thanks @dsever

helm/defectdojo/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "2.1.0-dev"
 description: A Helm chart for Kubernetes to install DefectDojo
 name: defectdojo
-version: 1.6.10
+version: 1.6.11
```
madchap, line 5: Should be 1.6.12 now.

helm/defectdojo/values.yaml
dsever: It is not the same, because for celery there is single container, I wanted to follow practice to have explanation together with values.

alles-klar (approved):

alles-klar:
@madchap did you tested it on your installation? Asking because I am not sure if we want to merge it before the release on Tuesday?

valentijnscholten:
I think it's mainly extra stuff with little chance of breaking existing stuff, so could be good to merge.

helm/defectdojo/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "2.2.0-dev"
 description: A Helm chart for Kubernetes to install DefectDojo
 name: defectdojo
-version: 1.6.11-dev
+version: 1.6.12-dev
```
StefanFl, line 5: @madchap Why is the version changed to 1.6.12-dev? The current version in master is 1.6.10, in dev we changed to 1.6.11-dev, so I assumed the next master will have version 1.6.11.

valentijnscholten: Not sure if this was changed at some point, but I think the linter expects a version bump every time something changes in the chart.

valentijnscholten: I changed it into 1.6.11-dev, let's see what happens :-)

valentijnscholten: ✖︎ defectdojo => (version: "1.6.11-dev", path: "helm/defectdojo") > Chart version not ok. Needs a version bump!

valentijnscholten: going to set it to 1.6.12-dev and merge it for the release. someone who knows helm should figure out this linting stuff as it seems to restrive atm

Issue #4840, opened 2021-07-19
Pagination issue on paused engagements


christophe226:
**Bug description**
On web view about engagements, pagination does not work for Paused engagements. Clicking on "page 2" or "Next" has no effect. Modifying "Page Size" has no effect either. Whereas the same actions work perfectly for Active engagements.

**Steps to reproduce**
1. Work on a product with more than 10 paused engagements
2. On this product, click on 'Engagements / View engagements'
3. In "Paused Engagements" section, click on page index "2"
4. Verify that you stay with the same engagements as in Page 1, i.e. page index has not changed.
5. Again in "Paused Engagements" section, click on "Page Size" and choose "25" for instance.
6. Verify that you still have only 10 results.

**Expected behavior**
I expect Page Index and Page Size work as well as in "Active Engagements" section.
The problem comes probably from wrong parameter names in GET request: queued_engs_countpage=1&queued_engs_countpage_size=25 instead of queued_engspage=1&queued_engspage_size=25

**Deployment method** *(select with an `X`)*
- [ ] Docker
- [X] Kubernetes
- [ ] GoDojo
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: Ubuntu 18.04

StefanFl:
PR is merged

PR #4846, opened 2021-07-21
Fix Pagination issue on paused engagements #4840

christophe226:
Original issue #4840 description:
On web view about engagements, pagination does not work for Paused engagements. Clicking on "page 2" or "Next" has no effect. Modifying "Page Size" has no effect either. Whereas the same actions work perfectly for Active engagements.

Fix: in view_engagements.html, there was prefix="queued_engs_count" instead of prefix="queued_engs"


Maffooch:
Was the pipeline file intended to be changed here as well?

christophe226:
> Was the pipeline file intended to be changed here as well?

Actually, I changed pipeline.py in my branch, as I started from a branch with a already modified pipeline.py. My changes were just to come back on the same pipeline.py as in dev branch. I should have done a rebase, but I did not manage, so I did this manual operation, sorry

damiencarol:
@Maffooch seems clean now.

StefanFl (approved):

Maffooch (approved):

PR #4851, opened 2021-07-22
AWS Prowler Scan parser Upgrade and Enhancements

dkade:
AWS Prowler scanner enhancements

Detected and fixed issues:
- Inside the same scan findings from different regions were being marked as duplicates;
- Missing useful information in finding description and fields;
- dedup_key was not strong;
- Deduplication was not working, so same test scan would duplicate the number of findings.

All those have been fixed.

Using the same scan we can see that before it only found 18 issues when in reality there are 60, because the key to check duplicates was wrongly classifying by ignoring AWS Region. 

Before:
Before_Finding

After:
After_Finding
After_Finding_Dedup
Marked_as_duplicates


dkade:
wow just noticed that the CSV files in the unit tests are missing a bunch of fields from the prowler scan. They seem to have been made for an older prowler version. Updating unit testing for AWS Prowler Scan.

dkade:
Please not the code in master for this parser isn't working properly, Critical issues are being imported as High and a lot of information was missing when importing to Dojo. 

This PR fixes them.

damiencarol (approved):

damiencarol:
@dkade you made an excellent contribution. Yes some test data are fake so some unit tests where not perfect.

StefanFl (approved):

Issue #4521, opened 2021-05-17
Cargo audit


zakrush:
**Scanner Name**
Cargo audit. 
Hello, It helpfull importer of `cargo audit` for rust dependencies. 
https://crates.io/crates/cargo-audit

It can generate only `text` or `json` format.
`cargo audit --json >> ./cargo-audit.json`

**Sample File**
I attached json output into txt file
[cargo-audit.txt](https://github.com/DefectDojo/django-DefectDojo/files/6493736/cargo-audit.txt)

ansidorov:
I will make a parser for cargo audit

ansidorov:
@damiencarol I think we can close this issue.

damiencarol:
Fixed in `dev`/`2.0.0`

zakrush:
@damiencarol  @ansidorov  Hello. Now I testing this integration:
I get this error.
```
uwsgi_1         | [23/Jul/2021 09:47:44] ERROR [dojo.importers.utils:27] list index out of range
uwsgi_1         | Traceback (most recent call last):
uwsgi_1         |   File "/app/./dojo/importers/utils.py", line 16, in parse_findings
uwsgi_1         |     parsed_findings = parser.get_findings(scan, test)
uwsgi_1         |   File "/app/./dojo/tools/cargo_audit/parser.py", line 33, in get_findings
uwsgi_1         |     cve = advisory.get('aliases')[0]
uwsgi_1         | IndexError: list index out of range
uwsgi_1         | [23/Jul/2021 09:47:44] WARNING [dojo.importers.utils:28] Error in parser: list index out of range
uwsgi_1         | [23/Jul/2021 09:47:44] ERROR [dojo.engagement.views:616] list index out of range
uwsgi_1         | Traceback (most recent call last):
uwsgi_1         |   File "/app/./dojo/engagement/views.py", line 601, in import_scan_results
uwsgi_1         |     test, finding_count, closed_finding_count = importer.import_scan(scan, scan_type, engagement, user, environment, active=active, verified=verified, tags=tags,
uwsgi_1         |   File "/app/./dojo/importers/importer/importer.py", line 290, in import_scan
uwsgi_1         |     parsed_findings = importer_utils.parse_findings(scan, test, active, verified, scan_type)
uwsgi_1         |   File "/app/./dojo/importers/utils.py", line 16, in parse_findings
uwsgi_1         |     parsed_findings = parser.get_findings(scan, test)
[...]

PR #4858, opened 2021-07-23
cargo audit parser repaired

zakrush:
repaire cargo-audit parser fail.

Describe here https://github.com/DefectDojo/django-DefectDojo/issues/4521#issuecomment-885528078

PR #4859, opened 2021-07-23
fix aliases for cargo audit parser

ansidorov:
Resolve issue https://github.com/DefectDojo/django-DefectDojo/issues/4521#issuecomment-885528078


zakrush:
Hi/ My pull is more correctly. 
https://github.com/DefectDojo/django-DefectDojo/pull/4858

In this case you get error into hash summ.

Issue #4752, opened 2021-06-30
Incorrect False positives and Risk accepted metrics


zakrush:
**Slack us first!**
Hello. I write about problem here:
https://owasp.slack.com/archives/C2P5BA8MN/p1624892081234100

**Be informative**
As additional into slack I find the same behaviour with Risk Accepted findings. Into Metrics I see 0 **Risk Accepted** findings, but I have 1 Risk Accepted finding

**Bug description**
No error. Metrics into product, or metrics dushboard has incorrect info. 

**Steps to reproduce**
1. Create 2 two or more findings into product
2. SetUp **false Positive** for finding and **Risk Accepted** for other
1. Go to 'Metrics' of products see that here 0 'False Positives'
2. Go to Metrics - Simple Metrics
3. Click into _View Details_ of products
4. Scroll 
3. Scroll down to `Metric Counts` - `Accepted Findings` 
4. See here is 0

**Expected behavior**
I expected that I can see metrics with False Positives and RiskAccepted

**Deployment method** *(select with an `X`)*
- [X ] Docker
- [ ] Kubernetes
- [ ] GoDojo
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: [e.g. Ubuntu 18.04]
 - DefectDojo is 1.15.0 from master

**Sample scan files** (optional)
If applicable, add sample scan files to help reproduce your problem.

**Screenshots** (optional)
![DD_Metrics_Bug](https://user-images.githubusercontent.com/33844164/123970372-bd227680-d9c1-11eb-83be-541aab74f976.gif)

**Console logs** (optional)
If applicable, add console logs to help explain your problem.

**Additional context** (optional)

zakrush:
I also foud the same behavior with Closed Findings into Metrics
Here is ZERO when I have Closed Findings
![image](https://user-images.githubusercontent.com/33844164/124109092-7428fb80-da6f-11eb-85fd-46ac2c12cbcf.png)

devGregA:
@zakrush  can you please confirm that this persists after 10 minutes? Some metrics are cached but I don't know which off the top of my head.

zakrush:
@devGregA It's persist more than 10 minutes. I am observing this problem so far. 
False Positives result are incorrect into product.

XiChen-Tibco:
Inactive findings within the Product Metrics is incorrect as well.

StefanFl:
[...]

PR #4860, opened 2021-07-23
Fix for Issue#4752

XiChen-Tibco:
Solution for issue #4752 by commenting out unnecessary variable check (Inactive finding within product metric) and modifying the date range (0 accepted finding issue).

### DIFF
--- a/dojo/metrics/views.py
+++ b/dojo/metrics/views.py
@@ -187,18 +187,19 @@ def finding_querys(prod_type, request):
                                                  test__engagement__product__prod_type__in=prod_type).prefetch_related(
             'test__engagement__product')
         # capture the accepted findings in period
-        accepted_findings = Finding.objects.filter(risk_acceptance__created__date__range=[start_date, end_date],
+        # set the start date of the month manually for acceepted_findings and accepted_findings_counts
+        accepted_findings = Finding.objects.filter(date__range = [start_date.replace(day=1), end_date], risk_accepted = True,
                                                    test__engagement__product__prod_type__in=prod_type). \
             prefetch_related('test__engagement__product')
-        accepted_findings_counts = Finding.objects.filter(risk_acceptance__created__date__range=[start_date, end_date],
+        accepted_findings_counts = Finding.objects.filter(date__range = [start_date.replace(day=1), end_date], risk_accepted = True,
                                                           test__engagement__product__prod_type__in=prod_type). \
             prefetch_related('test__engagement__product')
     else:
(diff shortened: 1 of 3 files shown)


XiChen-Tibco (commented):
Update for flake8 issue

dojo/metrics/views.py
```diff
@@ -187,18 +187,19 @@ def finding_querys(prod_type, request):
                                                  test__engagement__product__prod_type__in=prod_type).prefetch_related(
             'test__engagement__product')
         # capture the accepted findings in period
-        accepted_findings = Finding.objects.filter(risk_acceptance__created__date__range=[start_date, end_date],
+        # set the start date of the month manually for acceepted_findings and accepted_findings_counts
+        accepted_findings = Finding.objects.filter(date__range=[start_date.replace(day=1), end_date], risk_accepted=True,
                                                    test__engagement__product__prod_type__in=prod_type). \
             prefetch_related('test__engagement__product')
-        accepted_findings_counts = Finding.objects.filter(risk_acceptance__created__date__range=[start_date, end_date],
+        accepted_findings_counts = Finding.objects.filter(date__range=[start_date.replace(day=1), end_date], risk_accepted=True,
```
Maffooch, line 194: Does hard coding the start date affect the filters?

dojo/product/views.py
```diff
@@ -339,7 +339,7 @@ def finding_querys(request, prod):
                                          active=True,
                                          is_mitigated=False)
     filters['inactive'] = findings_qs.filter(date__range=[start_date, end_date],
-                                             false_p=False,
+                                             # No need for verify if it is False positive, comment out
```
Maffooch, line 342: Remove the line rather than commenting it out

Maffooch (changes requested):

dojo/metrics/views.py
XiChen-Tibco: Yep, for some reason, the original "start_date" variable is not the beginning of the month, so I add .replace(day=1) to make sure it starts from the first day of that month.

Maffooch: So the following url will only find findings from 08/03/2021 and 08/04/2021 instead of 08/01/2021 and 08/04/2021?

/product/1/metrics?start_date=08/02/2021&end_date=08/03/2021

XiChen-Tibco: It should not affect the product page metrics(product/1/metrics), this part of code is involving the overall metrics display (/metrics/product/type/1).
But I noticed another bugs on that page with the given url you have, and I'm trying to fix that right now.

Maffooch: The same logic applies to both metric pages

PR #4863, opened 2021-07-26
OAuth2: Default group and setting for staff members

StefanFl:
This PR implements 2 enhancements for new users created with OAuth2 logins:

- You can define a default group in the system settings, new users will be member of that group with the defined role.
- You can define an email pattern in the system settings to decide if a new users will be staff.

Additionally there comes an exception handler for the API, so the data of the response will not be a HTML document anymore but an error message.

There seem to a lot of changes in `models.py`, which are not actually changes but I had to move some classes before the system settings.


Maffooch (approved):

mtesauro (approved):
Approved

nu-vi:
Does this work for SAML or OAuth only?

StefanFl:
@nu-vi To be honest I don't really know. I haven't worked with SAML and don't know the framework and can't test it. My gut feeling is the pipeline works only for OAuth.

PR #4878, opened 2021-07-27
Master into dev/2.1.0 2.2.0 dev

Maffooch:


helm/defectdojo/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "2.1.0-dev"
+appVersion: "2.2.0-dev"
 description: A Helm chart for Kubernetes to install DefectDojo
 name: defectdojo
 version: 1.6.10
```
StefanFl, line 5: I think we should merge that PR even though the version of the helm chart needs to be updated. We can create another PR with the updated chart version.

StefanFl (approved):

helm/defectdojo/Chart.yaml
Maffooch: I fear that would break the unit tests. Could be incorrect though. @valentijnscholten @madchap what are your thoughts?

valentijnscholten: I usually bump it, so 1.6.11

valentijnscholten (changes requested):

Issue #4861, opened 2021-07-23
don't work migrations after update to 2.0


zakrush:
**Slack us first!**
https://owasp.slack.com/archives/C2P5BA8MN/p1627056588112000
here is describe my problem ^

**Be informative**
I upgraded from 1.15 to 2.0.3
I have deduplication on endpoints, and when I import scan I add id of endpoint. After update when script is importing scan it created new endpoint instead needed. My deduplication is broken. 

Then I saw that I needed to apply migration. Results of migration here:

```
user@defect-dojo:/var/www/defectdojo$ docker exec b3847dcc9fee python manage.py migrate
[23/Jul/2021 18:47:33] INFO [dojo.models:3613] enabling audit logging
Operations to perform:
  Apply all migrations: admin, auditlog, auth, authtoken, contenttypes, django_celery_results, dojo, sessions, sites, social_django, tagging, watson
Running migrations:
  No migrations to apply.
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
user@defect-dojo:/var/www/defectdojo$ docker-compose exec uwsgi ./manage.py migrate
[23/Jul/2021 18:48:53] INFO [dojo.models:3613] enabling audit logging
Operations to perform:
  Apply all migrations: admin, auditlog, auth, authtoken, contenttypes, django_celery_results, dojo, sessions, sites, social_django, tagging, watson
Running migrations:
  No migrations to apply.
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
user@defect-dojo:/var/www/defectdojo$ docker-compose exec uwsgi ./manage.py makemigrations
[23/Jul/2021 18:49:32] INFO [dojo.models:3613] enabling audit logging
Migrations for 'dojo':
  dojo/db_migrations/0114_auto_20210723_1849.py
    - Alter field default_issue_type on jira_instance
user@defect-dojo:/var/www/defectdojo$ docker-compose exec uwsgi ./manage.py migrate
[23/Jul/2021 18:49:50] INFO [dojo.models:3613] enabling audit logging
Operations to perform:
  Apply all migrations: admin, auditlog, auth, authtoken, contenttypes, django_celery_results, dojo, sessions, sites, social_django, tagging, watson
Running migrations:
[...]

PR #4887, opened 2021-07-28
Fix Endpoint clean in migration

kiblik:
- fix bug #4861
- add test
- description into documentation


Maffooch:
I think a better solution would be to add a note in the upgrade docs to run the [endpoint migration script](https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/management/commands/endpoint_migration.py) rather then making a duplicate migration.

zakrush:
@Maffooch In my case endpoint migration didn't help me when I upgraded to 2.0.3.

kiblik:
@zakrush, @Maffooch was referring to migration which I wrote as part of this PR

@Maffooch, I removed migration and added description in to documentation

Maffooch (approved):
@zakrush correct. There was a bug in the endpoint migration fix. This PR corrects the bug.

valentijnscholten (approved):

PR #4891, opened 2021-07-29
Install chromium driver in Python

StefanFl:
Today the integration tests were not running anymore, because the latest chrome driver didn't support the current stable chromium browser, see https://github.com/DefectDojo/django-DefectDojo/runs/3192403317?check_suite_focus=true

According to https://chromedriver.chromium.org/downloads/version-selection the method how the chromium driver was selected is not reliable:

> In addition, the version of ChromeDriver for the current stable release of Chrome can be found at https://chromedriver.storage.googleapis.com/LATEST_RELEASE. However, using on this file may be less reliable than methods described above.

The Python library [chromedriver-autoinstaller](https://pypi.org/project/chromedriver-autoinstaller/) installs automatically the driver for the currently installed chromium browser and seems to be the better solution.

### DIFF
--- a/Dockerfile.integration-tests
+++ b/Dockerfile.integration-tests
@@ -28,14 +28,7 @@ RUN \
   rm -rf /var/lib/apt/lists && \
   true
 
-RUN pip install --no-cache-dir selenium requests
-
-# Installing Chromium Driver and Selenium for test automation
-RUN LATEST_VERSION=$(curl -s https://chromedriver.storage.googleapis.com/LATEST_RELEASE) && \
-    wget -O /tmp/chromedriver.zip https://chromedriver.storage.googleapis.com/$LATEST_VERSION/chromedriver_linux64.zip && \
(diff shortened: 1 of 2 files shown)


devGregA (approved):

Maffooch (approved):

alles-klar:
@StefanFl Do you have a good idea for people running the test in environments without internet connection? Also a proxy configuration isn't supported :(

StefanFl:
@alles-klar Regarding the proxy: Does https://github.com/yeongbin-jo/python-chromedriver-autoinstaller/issues/18#issuecomment-872327054 help? But no iidea how it would work without internet connection to be honest.

alles-klar:
I prefer moving the installation back to the build step. It is much more efficient because you don't have to download and install chromedriver in each test run. Testing to install chromedriver in the dockerbuild with the autoinstaller wasn't successful yet, but would be the best solution.

PR #4903, opened 2021-08-02
Restrict deletion of users when used in product related objects

StefanFl:
Deleting a user currently can cascade down to product level or objects related to products like engagements, tests or findings. This PR restricts the deletion of a user when they are used in products related relations like contacts or leads. 

Having done this it seems to be safe to allow to delete users via the API.


dsever:
@StefanFl can you maybe add on the top of this, option to reassign objects to some other user?

Maffooch:
I like this. @dsever has a great idea to make it even better, but I'm thinking it could have some unintended affects without a very granular approach. For example, imagine deleting a user that interacts with multiple products. Who could be assigned that does not allow the new assignee to see more than they should? Assigning objects one by one could be super exhaustive as well.

valentijnscholten:
A compromise for now could be to generate a insightful error message explaining exactly which objects are preventing the user from being deleted.

StefanFl:
@dsever's idea would be a PR on its own and need to be thought through, how to implement it. 

The database (at least MySQL in my tests) returns a helpful error message, that gets displayed in the UI and is returned by the API:
2021-08-11 08_38_17-All Users _ DefectDojo

PR #4910, opened 2021-08-04
Change Dependency Check parser to make "Location" field more informative

W0uldYk1ndlY:
By default field "Location" on Findings page contains only name of the dependency, and no information about its location. It makes dependency check difficult when trying to analyze big monolith applications with many files with vulnerable dependencies. 

With this addition filed "Location" now display path for file, where vulnerable dependencies has been found (example "/lib/cropper/package-lock.json?lodash").


httpnotonly:
+

whitespots:
Why was it like this?)
path=name

whitespots (approved):
Looks good

sarosbacz:
Faced the same problem, yes

httpnotonly (approved):
+

SecurityEdge2:
good feature

Maffooch (changes requested):
Please see failed unit tests

zakrush:
IMHO No good Idea. I have the same problem, but I resolve this by other methods. 

If you get nodejs project, did `npm run build` and then run Dependency Check, you give very long filepath, It's be not readble.

W0uldYk1ndlY:
> Please see failed unit tests

Hello!

PR fails test_parse_file_with_multiple_vulnerabilities_has_multiple_findings (dojo.unittests.tools.test_dependency_check_parser.TestDependencyCheckParser) test. 

I can rewrite it to match new data in location field. Tell me, please, does it acceptable?

W0uldYk1ndlY:
> IMHO No good Idea. I have the same problem, but I resolve this by other methods.
> 
> If you get nodejs project, did `npm run build` and then run Dependency Check, you give very long filepath, It's be not readble.

Hello!

We don't encountered with any problems with our projects. But if you managed this problem with another method, can you, please, tell me about it?

zakrush:
> We don't encountered with any problems with our projects. But if you managed this problem with another method, can you, please, tell me about it?

We are also having monorepo with many services into one repo. 
So. I generate different jobs for every service. Then I upload scan result with endpoints, where hostname = product, path = service. Into hash I added enpdoints for deduplications.

valentijnscholten:
the dependency path can be something wild like `<filePath>/var/lib/jenkins/workspace/vice-middleware-security_develop/offer-service-ear/target/client-offer-service-ear-1.0-SNAPSHOT-deployment-prod.zip/jboss/standalone/deployments/h2-console.war</filePath>` which is impossible to read or display in a useful way on screen.

a solution could be to include the full path in the description so at least it's available somewhere.

@zakrush I've seen other people using (abusing) the endpoint mechanism to add an extra layer of abstraction between services / containers / micro services. I wonder if we should add a proper layer of abstraction to facilitate this as endpoints can have some side effects maybe for example with dedupe.

W0uldYk1ndlY:
> the dependency path can be something wild like `<filePath>/var/lib/jenkins/workspace/vice-middleware-security_develop/offer-service-ear/target/client-offer-service-ear-1.0-SNAPSHOT-deployment-prod.zip/jboss/standalone/deployments/h2-console.war</filePath>` which is impossible to read or display in a useful way on screen.
> 
> a solution could be to include the full path in the description so at least it's available somewhere.
> 
> @zakrush I've seen other people using (abusing) the endpoint mechanism to add an extra layer of abstraction between services / containers / micro services. I wonder if we should add a proper layer of abstraction to facilitate this as endpoints can have some side effects maybe for example with dedupe.

Hello! 

Sorry for long answer. Please tell me, should i close this PR and open Feature request instead?

Maffooch (approved):

damiencarol (approved):

PR #4912, opened 2021-08-05
Fix heading for intsights

kiblik:


StefanFl (changes requested):
Thanks a lot for that @kiblik. The 2 following sections (Manual Import and Automated Import) need to be one level higher as well, as they belong to the Intsight parser and need 4 hashes. Could you please fix that as well?

StefanFl (approved):

valentijnscholten (approved):

PR #4915, opened 2021-08-06
AWS Prowler Scan Fix and Enhancement

dkade:
This is the same as [#4851](https://github.com/DefectDojo/django-DefectDojo/pull/4851) but now PR to `dev` and with a clean base.

@StefanFl I did this new branch as the old one was too messy because of wrong base since the beginning.

Detected and fixed issues:
- Inside the same scan findings from different regions were being marked as duplicates;
- Missing useful information in finding description and fields;
- dedup_key was not strong;
- Deduplication was not working, so same scan would duplicate the number of findings.
- Noticed the parser was using an older version of Prowler in the CSV export, some fields were removed and others were added;
- Fixed the unit tests. CSV files were missing many fields.

Working for PROWLER_VERSION=2.4.0-07042021

CSV fiels taken from CSV Export of PROWLER_VERSION=2.4.0-07042021

Using the same scan we can see that before it only found 18 issues when in reality there are 60, because the key to check duplicates was wrongly classifying by ignoring AWS Region. 

Before:
Before_Finding

After:
After_Finding
After_Finding_Dedup
Marked_as_duplicates


valentijnscholten:
Since this was approved and merged in the original PR against master, I am merging this into dev straightaway

Issue #4920, opened 2021-08-07
Issues with helm versioning at release time when merging back from master to dev


madchap:
Mods' ref: https://owasp.slack.com/archives/GQKK61742/p1627404283002600

PR #4923, opened 2021-08-08
[GHA - release] Bump chart version when merging master back to dev

madchap:
Submitted against `master` for immediate effect, addresses #4920.

An issue we've been having is at https://github.com/DefectDojo/django-DefectDojo/runs/3173934689?check_suite_focus=true

This PR will check the current chart version with 
```
CURRENT_CHART_VERSION=$(grep -oP 'version: (\K\S*)?' helm/defectdojo/Chart.yaml)
```

and then smartly increment it and also adding a `-dev` suffix (per @valentijnscholten's feedback)
```
sed -r "s/version: \S+/$(echo "version: $CURRENT_CHART_VERSION" | awk -F. -v OFS=. 'NF==1{print ++$NF}; NF>1{$NF=sprintf("%0*d", length($NF), ($NF+1)); print}')/" helm/defectdojo/Chart.yaml
```

(I did not come up with that last regex, some people are smarter than me)

Since we're now adding a `-dev` suffix to the chart's version, that suffix is removed automatically when we submit the first PR from `dev`->`master`.


valentijnscholten:
I don't understand the wild regex/sed/awks so probably just merge it so we can do some test runs?

madchap:
Sure, I am good in testing it for real. The stuff worked manually on my branch but you know…

PR #4926, opened 2021-08-09
Add support for CloudSQL Auth Proxy's IAM login and private IPs

dhozac:
These flags allow the use of databases that reside in private networks as well as using IAM-based authentication rather than username and password (see e.g. https://cloud.google.com/sql/docs/postgres/authentication for more details).


Maffooch:
@dsever @alles-klar @madchap
I know close to nothing about helm. Any thoughts?

alles-klar (approved):

valentijnscholten:
need to bump the chart version to make it pass our (strict) linting check

PR #4928, opened 2021-08-09
Remove nonexisting file reference in sample data

Maffooch:
There are two references to a file uploads on a finding that do not exist in the fixtures file. Removing them allows the fixtures to run again without failure


mtesauro (approved):
Thanks for the fix that was breaking loading data on the demo server.

Approved.

valentijnscholten (approved):

Issue #3121, opened 2020-10-28
There are more mandatory fields required than showed in Swagger


vira-vira:
**Attempted to**
Create a new issue in Defect Dojo through API v2 using the minimal parameters marked as mandatory in the documentation. We use a POST Request with the endpoint /findings/
Documentation:
 http://127.0.0.1:8080/api/v2/doc/
Screenshots of Swagger:

<img width="1266" alt="Screen Shot 2020-10-29 at 09 16 27" src="https://user-images.githubusercontent.com/32620981/97572902-bcb38500-19c7-11eb-9161-8788fdd8e9a2.png">
<img width="1166" alt="Screen Shot 2020-10-29 at 09 16 49" src="https://user-images.githubusercontent.com/32620981/97572910-c0470c00-19c7-11eb-86b7-0ac0151bd3bb.png">
<img width="1171" alt="Screen Shot 2020-10-29 at 09 17 08" src="https://user-images.githubusercontent.com/32620981/97572913-c1783900-19c7-11eb-8e54-73ca8b72e198.png">
<img width="1185" alt="Screen Shot 2020-10-29 at 09 17 25" src="https://user-images.githubusercontent.com/32620981/97572919-c210cf80-19c7-11eb-98f8-a4390a168d30.png">

In Swagger, it shows that the minimum mandatory fields required for a POST Request in /findings are:
- test
- found_by
- title
- severity
- description
- mitigation
- impact

**The expected behavior was**
A 201 Created with a JSON response as stated in the docs.
**We Got**
<h1> 500 Error </h1>.
In the server logs there is this traceback:
![image](https://user-images.githubusercontent.com/32620981/97574556-2b91dd80-19ca-11eb-9b0a-33c3c27bac49.png)

**Workaround**
We created POST request with the following parameters:

- test
- found_by
- title
- severity
- description
- mitigation
- impact
- numerical_severity
- duplicate
- verified
- false_p

valentijnscholten:
@vidhan13j07 could you check again in `dev`? We have fixed lots of schema inconsistencies and added an OpenAPI v3 schema as well.

devGregA:
@valentijnscholten  I can have this validated.

XiChen-Tibco:
@vira-vira Hi, could you give us more details about how to reproduce this error?

valentijnscholten:
merged into dev / 2.2.0

valentijnscholten:
[...]

PR #4931, opened 2021-08-10
API: list all mandatory fields for FindingCreateSerializer

XiChen-Tibco:
Fix the required parameter issue within swagger for issue #3121.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1156,6 +1156,8 @@ class ProductSerializer(TaggitSerializer, serializers.ModelSerializer):
     #  -> List[int] as return type doesn't seem enough for drf-yasg
     @swagger_serializer_method(serializer_or_field=serializers.ListField(child=serializers.IntegerField()))
     def get_findings_list(self, obj) -> List[int]:
+        print("HERE")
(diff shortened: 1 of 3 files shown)


dojo/api_v2/serializers.py
```diff
@@ -1058,8 +1058,13 @@ class FindingCreateSerializer(TaggitSerializer, serializers.ModelSerializer):
 
     class Meta:
         model = Finding
-        exclude = ['images']
+        fields = '__all__'
         extra_kwargs = {
+            'mitigation': {'required': True},
+            'impact': {'required': True},
```
Maffooch, line 1064: I don't believe these fields are required anywhere in the app anymore, but they were when the issue was created.

Maffooch (approved):

valentijnscholten (approved):

valentijnscholten:
Tbh I haven't checked, but thought the '__all__' thing did that. But maybe not since only images was excluded?

Maffooch:
No the current `fileds = __all__` and the old `exlcude = ['images']`are the exact same since images have been removed

valentijnscholten:
yeah that's what I meant

PR #4933, opened 2021-08-10
SonarQube: use severity from issue instead of rule

zeeshan811:
Changed the code to show updated severity of the vulnerability rather than the vulnerability severity of the rule for better behavior for sonarqube_api

### DIFF
--- a/dojo/tools/sonarqube_api/importer.py
+++ b/dojo/tools/sonarqube_api/importer.py
@@ -75,8 +75,6 @@ class SonarQubeApiImporter(object):
                 line = issue.get('line')
                 rule_id = issue['rule']
                 rule = client.get_rule(rule_id)
-                #severity = self.convert_sonar_severity(rule['severity'])
-                #Changed the code to show updated severity of the vulnerability rather than the vulnerability severity of the rule for better behavior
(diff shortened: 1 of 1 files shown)


dojo/tools/sonarqube_api/importer.py
```diff
@@ -75,7 +75,9 @@ def import_issues(self, test):
                 line = issue.get('line')
                 rule_id = issue['rule']
                 rule = client.get_rule(rule_id)
-                severity = self.convert_sonar_severity(rule['severity'])
+                #severity = self.convert_sonar_severity(rule['severity'])
```
StefanFl, line 78: Please remove both comments. If someone is interested in the history of the change they can use GitHub to find out what has changed and why.

StefanFl (changes requested):

dojo/tools/sonarqube_api/importer.py
zeeshan811: Removing the comments

zeeshan811 (commented):
Check.

StefanFl:
Succeeded by #4934

PR #4935, opened 2021-08-12
Fix for issue parsing empty elements within ZAP report (DefectDojo#4855)

iainrawson:


dojo/tools/zap/parser.py
```diff
@@ -155,16 +155,16 @@ def __init__(self, item_node):
         for instance in item_node.findall('instances/instance'):
             for node in instance.iter():
                 if node.tag == "uri":
-                    if node.text != "":
+                    if node.text != "" and node.text is not None:
```
Maffooch, line 158: Would look better if it were 
`if node.text and node.text != "":`

valentijnscholten:
Thanks, could you update one of the sample files used in the unit tests to cover this case for future regression protection?

iainrawson:
I've added a sample empty <param> element to one of the test files.  Is there any documentation on how to run the tests locally?

StefanFl (approved):

PR #4954, opened 2021-08-18
Fix newline characters in the summary (Jira issue)

Zilborg:
Added `.replace('\r', '').replace('\n', '')` to summary.

Alternative: using `autofix` in JIRA config. (see https://jira.readthedocs.io/en/master/_modules/jira/config.html?highlight=autofix)

This bug depends on Finding titles. In my case, the Semgrep import put newlines characters in title.

Log fragment:
```
celeryworker_1  | 2021-08-18T12:02:33.806344825Z 
celeryworker_1  | 2021-08-18T12:02:34.778273729Z [18/Aug/2021 12:02:34] ERROR [dojo.jira_link.helper:711] JiraError HTTP 400 url: https://grib-test.atlassian.net/rest/api/2/issue
celeryworker_1  | 2021-08-18T12:02:34.778312271Z 	text: The summary is invalid because it contains newline characters.
celeryworker_1  | 2021-08-18T12:02:34.778337818Z 	
celeryworker_1  | 2021-08-18T12:02:34.778347416Z 	response headers = {'Server': 'AtlassianProxy/1.19.3.1', 'cache-control': 'no-cache, no-store, no-transform', 'Content-Type': 'application/json;charset=UTF-8', 'Strict-Transport-Security': 'max-age=315360000; includeSubDomains; preload', 'Date': 'Wed, 18 Aug 2021 12:02:34 GMT', ....................many headers}
celeryworker_1  | 2021-08-18T12:02:34.778365961Z 	response text = {"errorMessages":[],"errors":{"summary":"The summary is invalid because it contains newline characters."}}
celeryworker_1  | 2021-08-18T12:02:34.778377652Z Traceback (most recent call last):
celeryworker_1  | 2021-08-18T12:02:34.778384836Z   File "/app/dojo/jira_link/helper.py", line 670, in add_jira_issue
celeryworker_1  | 2021-08-18T12:02:34.778391789Z     new_issue = jira.create_issue(fields)
```

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -617,7 +617,7 @@ def add_jira_issue(obj, *args, **kwargs):
                 'project': {
                     'key': jira_project.project_key
                 },
-                'summary': jira_summary(obj),
+                'summary': jira_summary(obj).replace('\r', '').replace('\n', ''),
                 'description': jira_description(obj),
                 'issuetype': {
                     'name': jira_instance.default_issue_type


valentijnscholten:
Interesting find. Any reason you didn't opt for the `autofix` option? My first thought is that would be 'safer' as it might fix other cases as well which we haven't thought about yet? On the other hand, the jira python library seems to be no longer maintained, so we may have to move to another one possibly.

damiencarol:
We should instead move the fix in the `jira_summary`:
https://github.com/DefectDojo/django-DefectDojo/blob/053b4a82fd9a9b67d9ebca8add25c78252195f44/dojo/jira_link/helper.py#L493-L500
Because it is used also here:
https://github.com/DefectDojo/django-DefectDojo/blob/053b4a82fd9a9b67d9ebca8add25c78252195f44/dojo/jira_link/helper.py#L787

damiencarol (changes requested):
See my comment

valentijnscholten:
> We should instead move the fix in the `jira_summary`:

Committed, let's await test suite.

valentijnscholten (approved):

damiencarol (approved):

Zilborg:
> Interesting find. Any reason you didn't opt for the `autofix` option? My first thought is that would be 'safer' as it might fix other cases as well which we haven't thought about yet? It does seem to fix some other corner cases as well around reporters/assignee being mandatory etc.

Sorry for late answer! And thank you for merged fix! 
Actually, the `autofix` is global variable and I haven't enough time for research how it works. It may affect other functions. The `replace` is easier and local solution for me.

I also added _cutter_ for long titles, that the tasks in the Jira looks better. 
And added `display_tag` for converting markdown from Defect dojo fields (**Impact**, **Mitigation**) to Jira syntax via this[ render file](https://github.com/miyuchina/mistletoe/blob/master/contrib/md2jira.py). Now it looks like at screenshot Jira View.

DefectDojo syntax
Screenshot 2021-09-16 at 10 18 17

Jira view
Screenshot 2021-09-16 at 10 16 19

P.S. It has bug with comment view block, which contains new lines. =)

If this modifications are interesting for production version, I will create correct pull requests)

PR #4676, opened 2021-06-17
SonarQube: support multiple SQ projects for one product

kiblik:
# Idea

Right now, DD supports to [store multiple SQ configurations (SQ Tool Config + Project Key)](https://github.com/DefectDojo/django-DefectDojo/blob/31801d9341255915b6023ea552e6e32e920b60fc/dojo/models.py#L1539-L1550), however:
- UI allows configuring only one SQ config
- SQ parser automatically uses the only first config - Check [this](https://github.com/DefectDojo/django-DefectDojo/blob/31801d9341255915b6023ea552e6e32e920b60fc/dojo/tools/sonarqube_api/importer.py#L48), [this](https://github.com/DefectDojo/django-DefectDojo/blob/31801d9341255915b6023ea552e6e32e920b60fc/dojo/tools/sonarqube_api/updater.py#L108) or [this](https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/tools/sonarqube_api/updater_from_source.py#L33)

Usecase (which should be solved by this PR): Allow to pull results from different SQ projects in one product - one SQ project = one engagement

# Parser 

Logic of parser:
  1. No SQ Tool Config - fail (from api_client)
  2. One SQ Tool Config, No Product SQ Config - use product name as key
  3. Multiple SQ Tool Configs, No Product SQ Config - fail (from api_client)
  4. One Product SQ Config, No Project key, No selected Product SQ config - use it and use product name as key
  5. One Product SQ Config, With Project key, No selected Product SQ config - use it
  6. Multiple Product SQ Configs, No selected Product SQ config - fail (from importer)
  7. Selected Product SQ config, No Project key -  use it and use product name as key
  8. Selected Product SQ config, With Project key - use it

SQ Tool Config = URL + API-key/user+pass
Product SQ Config = Project Key + SQ Tool Config

Logic of reimport/updater: 
If Product SQ Config was used, config id is saved in test -> it will be reused in case of reimport

# DB migration

- If sb hasn't set SonarQube config until now, the first one was used by default. So we are going to "hardcode" it because the empty config is not allowed anymore
- Test will store SQ config - it will be reused during re-upload (retest)

# GUI

| What/Where | Before | After |
| --- | --- | --- |
| All Products - Add+view SQ | image | image |
| Product Tab - Add+view SQ | image | image |
| View Product - add+view SQ | image | image |
| Add/Edit Product | image | image |
| Add SQ | x | image |
| View SQ | x | image |
| Edit SQ | x | image |
[...]

PR #4962, opened 2021-08-20
Add Cobalt.io API parser/importer

ericcornelissen:
This adds a (second) parser for Cobalt.io which can be used to import findings from [the Cobalt API](https://docs.cobalt.io). This is as opposed to the existing Cobalt.io parser (ref. #1215) which can be used to import findings from the CSV file that Cobalt.io users can get in the webapp. _Full disclosure, I'm a software engineer at Cobalt.io._

As was [suggested to me on Slack](https://owasp.slack.com/archives/C014H3ZV9U6/p1628070452009800?thread_ts=1627995536.004600&cid=C014H3ZV9U6), the implementation is based on that of the SonarQube API importer (ref. #4676). As such, this adds more than just a parser, namely:

- A Cobalt.io configuration at the product level, including menu items and webpages to manage this configuration.
- A field to the "Import Scan" form to select the Cobalt.io configuration (required if there's more than one for a product).

## Usage Details

Before the parser can be used one has to add some tokens for the Cobalt.io API as a _Tool Configuration_ (see first image below). In particular, this requires adding your Cobalt.io API token and what is called a Cobalt.io Org token. This org token is a token that determines from which organisation in Cobalt findings will be returned. The org token is inputted into the "Extras" field of the tool configuration.

Next, at the product level you have to configure the Cobalt.io assets from which to import findings (see third image below). Users can configure multiple assets for one product if they so choose, but will always import findings from only one asset at a time. The motivation to import findings from a specific asset rather than a specific pentest is this: the latter would require updating the configuration every time there's a new pentest in Cobalt.

Lastly, to import findings you go to the Import Scan form, select the parser being added here, select the asset from which to import findings, and hit import - no file required. (see fourth image below)

## Implementation Details

The bulk of the changes are Cobalt.io equivalents of the configuration objects and logic for SonarQube - if I did anything wrong here I'm happy to change it.

The database migration is just to accommodate storing the Cobalt.io configuration at the product level.

On top of that the parser itself for the Cobalt.io API is included. Some notable things about the parser that may require extra attention:

- The created date is set to the date the finding was created in Cobalt, not sure if that's the desired behaviour.
- Findings in the "new" and "triaging" state in Cobalt.io are not yet 100% verified by the pentesters so the `verified` flag is set to `false`. Alternatively these could be excluded entirely. 
- Findings marked as "duplicate" in Cobalt.io are imported with the `duplicate` flag set. These could also just be excluded entirely.
- Findings marked as "wont_fix" in Cobalt.io are imported with the `risk_accepted` flag set. These could also just be excluded entirely.
- Findings marked as "invalid" in Cobalt.io are imported with the `false_p` flag set. These could also just be excluded entirely.

#### Tool config form for Cobalt.io

tool config form

#### Product config for Cobalt.io

to add Cobalt io configuration

cobaltio config form

#### Import form with Cobalt.io

select cobalt asset on import


valentijnscholten:
Great PR, thanks for the elaborate description and screenshots. I think it was mentioned earlier that Defect Dojo should focus on it's core task, which is parsing of scan / pentest reports. As opposed to try and integrate with lots of tools. For me I think it could be very helpful if Defect Dojo could "easily" pull in findings from tools like SonarQube, Cobalt.io, etc. 
I do think if we continue on this path, we should make the implementation a little more generic so we don't have to add a new database column for each new tool for example.

ericcornelissen:
I completely agree that the approach I used here is not great in the long run. If there will be more "importers" like this and SonarQube's there will be a lot of near-duplication and indeed a lot of database columns - also, things like the dropdown in [the second preview](https://user-images.githubusercontent.com/3742559/130249869-2e450218-21c4-480e-8e9b-a1a5ebc5751c.png) will grow out of hand. If we can find a way to do it generically, I personally think it would make sense to support importers.

It's a small sample but the SonarQube parser and this Cobalt parser are very similar. The global tool configuration is already generic. The product-level configurations are identical in approach (one input + a dropdown for the tool configuration), as is their usage in the import form (just a dropdown to select the product-level configuration). If it is possible to make the product-level configuration form generic and change the import scan form fields based on the selected parser, then we should have a decent solution. What do you think?

In the meanwhile, as I expect such a change will take quite some time, would it make sense to add a "manual" parser for the Cobalt.io API (already present in 78d68b3c769974d00301193ffe189785b8a7d6be)? Would it be possible to change such a parser into something like what is currently being added here? Or perhaps that would become a third parser (such that there's one for CVS, one for the API manually, and one for the API automated)?

dojo/tools/cobalt_api/parser.py
```diff
@@ -0,0 +1,182 @@
+import json
+from datetime import datetime
+from dojo.models import Endpoint, Finding
+from dojo.tools.cobalt_api.importer import CobaltApiImporter
+
+__author__ = 'ericcornelissen'
```
damiencarol, line 6:
```suggestion
```
We are using VCS to track authorship. Meta attributes are not mandatory.

dojo/tools/cobalt_api/parser.py
```diff
@@ -0,0 +1,182 @@
...
+                verified=self.is_verified(cobalt_state),
+                false_p=self.is_false_p(cobalt_state),
+                duplicate=self.is_duplicate(cobalt_state),
+                out_of_scope=self.is_out_of_scope(cobalt_state),
+                risk_accepted=self.is_risk_accepted(cobalt_state),
+                is_mitigated=self.is_mitigated(cobalt_state),
+                last_status_update=last_status_update,
+                numerical_severity=self.convert_severity_numerical(cobalt_severity),
+                static_finding=False,
+                dynamic_finding=False,
```
damiencarol, line 78: How is it possible? Most of parsers are static or dynamic. Does Cobalt mix the both?

damiencarol: I was thinking that Cobalt.io is a "Pentest as a service" so dynamic?

damiencarol (changes requested):

damiencarol:
@ericcornelissen I think it's good to merge this one after some fixes but **after `2.2.0` release**.

For now we don't have to much parsers that "gather/collect" findings from external systems.
If we want to add more feature like that we will need a more generic way to configure them (like @valentijnscholten suggested) and keep the specific code for the "importer" part in the module `dojo/tools/xxx` 
 
The main challenge for now is that this kind of integration (like for SQ) bring to much specific code in upper layers (for ex: `dojo/api_v2/serializers.py`). Which is a bad thing if we do it too much.

dojo/tools/cobalt_api/parser.py
ericcornelissen: That makes sense to me, I updated it.

To clarify, I initially opted to set `dynamic_finding` to `False` as well since I wasn't sure a pentest counts as a "DAST **tool**".

https://github.com/DefectDojo/django-DefectDojo/blob/3f443735126eb987cb7d8e47fb54f427cd8652bd/dojo/models.py#L1821-L1823

damiencarol: I guess, maybe a third flag is missing here, like `pentest_finding`.

ericcornelissen: That sounds like a better solution, should I add that as part of this Pull Request?

damiencarol: no, it will be too big. let's use `dynamic_finding` and we will work on another flag another time.

damiencarol:
The migration part require modifications
```
2021-08-31T07:49:50.7412474Z [36minitializer_1        |[0m [31/Aug/2021 07:49:50] INFO [dojo.models:3634] enabling audit logging
2021-08-31T07:49:52.1902336Z [36minitializer_1        |[0m CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0121_cobaltio_product, 0121_user_restrict in dojo).
```

dojo/tools/cobalt_api/parser.py
damiencarol: Had some discussion with the team, use `dynamic_finding`.

dojo/tools/cobalt_api/parser.py
```diff
@@ -0,0 +1,180 @@
...
+        return cobalt_state == "wont_fix"
+
+    def is_verified(self, cobalt_state):
+        return cobalt_state != "new" and cobalt_state != "triaging"
+
+    def safe_title(self, title):
+        if len(title) > 255:
+            return title[0:252] + "..."
+        else:
+            return title
```
damiencarol, line 180: Use the function `shorten` of the package `textwrap` here ( https://docs.python.org/3/library/textwrap.html#textwrap.shorten ).

damiencarol (changes requested):

dojo/tools/cobalt_api/parser.py
ericcornelissen: Thanks for the suggestion, removed this helper and added `textwrap.shorten` where this helper used to be called.

dojo/tools/cobalt_api/parser.py
```diff
@@ -0,0 +1,175 @@
...
+                steps_to_reproduce=steps_to_reproduce,
+                active=self.is_active(cobalt_state),
+                verified=self.is_verified(cobalt_state),
+                false_p=self.is_false_p(cobalt_state),
+                duplicate=self.is_duplicate(cobalt_state),
+                out_of_scope=self.is_out_of_scope(cobalt_state),
+                risk_accepted=self.is_risk_accepted(cobalt_state),
+                is_mitigated=self.is_mitigated(cobalt_state),
+                last_status_update=last_status_update,
+                numerical_severity=self.convert_severity_numerical(cobalt_severity),
```
damiencarol, line 75:
```suggestion
```
`numerical_severity` is managed in the upper layers.

dojo/tools/cobalt_api/parser.py
```diff
@@ -0,0 +1,175 @@
...
+    def convert_severity_numerical(self, cobalt_severity):
+        """Convert severity value"""
+        if cobalt_severity == "low":
+            return "S3"
+        elif cobalt_severity == "medium":
+            return "S2"
+        elif cobalt_severity == "high":
+            return "S1"
+        else:
+            return "S4"
```
damiencarol, line 152:
```suggestion
```
See the above comment

damiencarol (changes requested):

damiencarol (approved):

damiencarol:
I think this PR is ok now. Let's merge it like this and work further on improving the importer/reimporter part to be more generic.

Maffooch (approved):
I agree this would ideally be more generic, but that would require a lot of overhaul for this, and the SonarQube integration. Merging now to prevent the PR from sitting forever.

Issue #3666, opened 2021-01-20
deduplication config not checked at startup


valentijnscholten:
when someone makes a mistake in the dedupe hashcode field configuration, this is silently ignored and will result in confusing deduplication problems for security analyst who maybe uploading reports but do not have access to settings.py config.

we should check the config at startup. the logic is already there in models.py, but only checked during dedupe:

```
                    deduplicationLogger.debug(
                        "compute_hash_code - configuration error: some elements of HASHCODE_FIELDS_PER_SCANNER are not in the allowed list HASHCODE_ALLOWED_FIELDS. "
                        "Using default fields")
```

PR #4963, opened 2021-08-20
Check deduplication config on startup

jhewi:
This script will solve issue #3666 by checking for deduplication configuration at startup.


valentijnscholten:
I think just logging something on INFO level won't get much attention. It should be WARNING or ERROR, but even better would be to let the startup fail I think?

Maffooch (approved):
Finally got this to pass tests.

damiencarol (approved):

PR #4965, opened 2021-08-21
Implementation for Engagement_Presets and Network_Locations API

jpbowie:
Added support to the v2 API for Engagement_Presets and Network_Locations

### DIFF
--- a/dojo/api_v2/permissions.py
+++ b/dojo/api_v2/permissions.py
@@ -1,7 +1,8 @@
 import re
 
 from rest_framework.exceptions import ParseError
-from dojo.models import Endpoint, Engagement, Finding, Product_Type, Product, Test, Dojo_Group
+from dojo.models import Endpoint, Engagement, Finding, Product_Type, Product, Test, Dojo_Group, Engagement_Presets, \
+    Network_Locations
 from django.shortcuts import get_object_or_404
 from rest_framework import permissions
 from dojo.authorization.authorization import user_has_permission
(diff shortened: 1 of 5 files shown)


StefanFl (changes requested):
Thanks for your submission @jpbowie. With some changes for the permissions it will be good for merging. 

Just out of interest, I never saw why the EngagementPresets are needed. What is your use case for them?

jpbowie:
I simplified the engagement presets permissions implementation and removed the network locations implementation in favor of the model permissions as requested.  I removed the  engagement presets validation as it was simply replacing a null scope with an empty space to prevent errors, the user can do that in the api.

I'm thinking about using the engagement presets api with some automation so when a new engagement is created based on a preset, specific tests within the preset can be performed automatically and the results can then be uploaded as part of the engagement.  However, for this to be possible, the tests that are part of the preset need to be available and I didn't find an existing api that would allow me to determine that.

StefanFl:
There is currently no functionality implemented concerning `Engagement_Presets`. You can save new ones and edit them, but that's it.

jpbowie:
There appears to be enough functionality to implement what I was thinking (except for the API support).  I have been able to Add/Edit/Delete an Engagement_Preset, assign an Engagement_Preset to an existing or new engagement (both in the UI and the engagement API), and an Engagement_Preset panel appears on the engagement screen with details when an Engagement_Preset has been assigned to the engagement.

StefanFl (approved):

PR #4977, opened 2021-08-24
Aws prowler fix

dkade:
There was an issue with severity, now it's working properly


StefanFl:
Hello @dkade, you need to have a look at the unit tests. They don't work with your change.

Another thought: I haven't used AWS Prowler before but it seems to be a very useful tool. But CSV is quite a poor file format, wouldn't it be better in the long run if DD would support the JSON output of AWS Prowler?

dkade:
I'm using the orignal CSV from Aaron that's why it is this way. It makes all the sense to switch to JSON but for that I need time and for now I need to have this working in production.

StefanFl:
I have a JSON example now and can implement the JSON import soon. Nevertheless please fix the unit tests so we can have both options.

dkade:
Yes just noticed I fixed the severity but not the unit tests, will do this now.

StefanFl (approved):

PR #4979, opened 2021-08-25
Initialization of test types 

StefanFl:
Test types are currently initialized by a fixture which is out of date. There are scanners in the list that don't exist and others are missing. An entry for a test type will be created if it doesn't exist by the scanner factory when using it the first time. But this is not enough for integration scenarios, where you want to create a test before an import. There you need an id for the test type that might not exist in the database yet.

This PR adds an initialization of the test types to the initializer container and can now remove a lot of entries in the fixture and an integration test case.


dsever:
@StefanFl 

This approach I proposed to @damiencarol, but there are few gaps in implementation:
1. Importers are not using the same mechanism (coming directly from factory) as creation if test types (coming from DB ), and it is important when test is created first and then reimport is in use.
2. There are few test types that are not part of factory (e.g pen test, API scan)

Descriptions should be imported into the table as well: 
https://github.com/dsever/django-DefectDojo/blob/remove_fixtures_tools_type/dojo/management/commands/dump_test_types.py

StefanFl:
@dsever 

1. I think this is addressed by this PR, now there will be a Test_Type for every parser
2. That hasn't changed with this PR. The additional Test_Types are still there and can be you used if a user wants to create a Test and Findings manually without a parser
3. Where would you use the descriptions in the user interface? Might be something for an additional PR.

dsever:
> @dsever
> 
> 1. I think this is addressed by this PR, now there will be a Test_Type for every parser

There could be problem with consistency of scanners in case of upgrade. 
See https://github.com/dsever/django-DefectDojo/blob/remove_fixtures_tools_type/dojo/management/commands/dump_test_types.py#L66

> 2. That hasn't changed with this PR. The additional Test_Types are still there and can be you used if a user wants to create a Test and Findings manually without a parser

Sorry I've missed to see that you haven't removed fixture completely  

> 3. Where would you use the descriptions in the user interface? Might be something for an additional PR.
This should be additional column in DB, and reflected to the interface and API.

PR #4994, opened 2021-08-28
make notifications async again

valentijnscholten:
When an import is done, or some bulk action is performed, Defect Dojo sometimes sends out notifications. Some of these can be quite slow, for example sending an e-mail via SMTP. With lots of users, this can slow down the UI quite a lot. Do an import, wait 30 seconds for the notifications to be sent.

In past, we made the notifications asynchronous for the above reason. However `django-tagulous` had a bug, or missing feature, where it didn't support the pickling used by our celery workers. So we had to make the notifications synchronous again in https://github.com/DefectDojo/django-DefectDojo/pull/3678.

In `django-tagulous` 1.2.0 pickling support was added, which means we can make the notifications asynchronous again in this PR.

We also remove some unused code from earlier attempts to work around the pickling problem.

### DIFF
--- a/dojo/decorators.py
+++ b/dojo/decorators.py
@@ -2,8 +2,6 @@ from functools import wraps
 from dojo.models import Finding, Dojo_User
 from django.db import models
 from django.conf import settings
-from django.forms.models import model_to_dict
-from django.db.models.query import QuerySet
 
 from ratelimit.exceptions import Ratelimited
 from ratelimit.core import is_ratelimited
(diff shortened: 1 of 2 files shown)


StefanFl (approved):
Release 2.2.0?

valentijnscholten:
I'm good with 2.3.0

damiencarol (approved):
To merge only in 2.3.0

Maffooch (approved):

PR #5025, opened 2021-09-01
Better support SARIF ruleId attribute

damiencarol:
According to specification, attribute `ruleId` could be missing from reports. (cf https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317643 )

> Not all existing analysis tools emit the equivalent of a ruleId in their output. 

This PR allow to import reports for tools that don't emit `ruleId` properly (like Gitleaks `7.5.0` which we use for a new unit tests) 

### Work done
- [x] better support attribute `ruleId` when tools doesn't set it cleanly (missing, reference but no data, etc...)
- [x] implement `mitigation` attribute

### DIFF
--- a/dojo/tools/sarif/parser.py
+++ b/dojo/tools/sarif/parser.py
@@ -150,11 +150,6 @@ def cve_try(val):
 
 
 def get_item(result, rules, artifacts, run_date):
-    mitigation = result.get('Remediation', {}).get(
-        'Recommendation', {}).get('Text', "")
-    references = result.get('Remediation', {}).get(
-        'Recommendation', {}).get('Url')
-
     # if there is a location get it
     file_path = None
     line = -1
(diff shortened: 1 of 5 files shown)


dojo/tools/sarif/parser.py
```diff
@@ -185,18 +196,19 @@ def get_item(result, rules, artifacts, run_date):
     # we add a special 'None' case if there is no CWE
     cwes = [0]
     if rule is not None:
```
StefanFl, line 198: You have 2 times `if rule is not None:`, here and in line 179. Can you put the code under one `if`?

dojo/tools/sarif/parser.py
```diff
@@ -185,18 +196,19 @@ def get_item(result, rules, artifacts, run_date):
...
-        cwes_extracted = get_rule_cwes(result)
+        cwes_extracted = get_rule_cwes(rule)
+        cwes_properties_extracted = get_result_cwes_properties(result)
```
StefanFl, line 200: `get_result_cwes_properties` is now only called when there are rules. Wouldn't it make sense to get CWEs from the results if there are no rules as well?

StefanFl (commented):
See my 2 questions.

dojo/tools/sarif/parser.py
damiencarol: modified.

dojo/tools/sarif/parser.py
damiencarol: Modified following your comment

damiencarol:
@StefanFl addressed your comments and also implemented `mitigation` attribute.

StefanFl (approved):

valentijnscholten (approved):

PR #5030, opened 2021-09-02
safety-parser: configurable offline mode

alles-klar:
When using the safety parser in environments without internet connection it take a while until the parser using the fallback offline mode. Now you can configure to use offline mode only.

On behalf of DB Systel GmbH


damiencarol:
@alles-klar I don't think we should continue to query the DB online. I made a PR to the tool upstream to have all the data in the report and to remove completely the external call.
The team that maintain Safety DB didn't merged or answered yet. So better to remove completely the call and just update the file when we release a version as Safety DB update this file only once a month.

damiencarol:
As discussed IRL with @alles-klar , we will remove this external call in this PR and wait PR https://github.com/pyupio/safety/pull/340 . After that I will update the parser to support the new data

valentijnscholten (approved):

PR #5035, opened 2021-09-03
Add a deduplication configuration for Aquasecurity's Cloudsploit

axelpavageau:


damiencarol (approved):

dojo/settings/settings.dist.py
```diff
@@ -978,6 +978,7 @@ def saml2_attrib_map_format(dict):
     'CargoAudit Scan': ['cve', 'severity', 'component_name', 'component_version', 'vuln_id_from_tool'],
     'Checkmarx Scan': ['cwe', 'severity', 'file_path'],
     'Checkmarx OSA': ['cve', 'component_name'],
+    'Cloudsploit Scan': ['title', 'severity', 'description'],
```
ptrovatelli, line 981: hello @axelpavageau. Is the title enough to identify a vulnerability or have you ever seen 2 vulns with the same title but a different description? i've checked the unit test sample but it is a little simple. If you have bigger samples can you check? 

If the title is enough (or if it's enough in 99% of cases), I'd advise not to use the description in the key, because in case of minor description change (which will happen when the tool version changes), that will make defectdojo fail to identify that it is actually still the same vulnerability. 

Also it's good practice to define the `HASHCODE_FIELDS_PER_SCANNER` fields in the same order as in HASHCODE_ALLOWED_FIELDS (see why in https://defectdojo.github.io/django-DefectDojo/usage/features/#hash_code-computation-configuration)  . so should be: 
`'Cloudsploit Scan': ['title', 'description', 'severity'],`

or

`'Cloudsploit Scan': ['title', 'severity'],`

Thanks!

ptrovatelli (changes requested):
see comment inline

dojo/settings/settings.dist.py
axelpavageau: hello @ptrovatelli  👋 
In cloudsploit's reports `title` contains the name of the test, so you can have multiple findings with the same title, which don't want to consider duplicates. 
```
[
  {
    "plugin": "acmValidation",
    "category": "ACM",
    "title": "ACM Certificate Validation",
    "description": "ACM certificates should be configured to use DNS validation.",
    "resource": "arn:aws:acm:us-east-1:012345678901:certificate/foo",
    "region": "us-east-1",
    "status": "OK",
    "message": "*.foo.net is using DNS validation."
  },
  {
    "plugin": "acmValidation",
    "category": "ACM",
    "title": "ACM Certificate Validation",
    "description": "ACM certificates should be configured to use DNS validation.",
    "resource": "arn:aws:acm:us-east-1:012345678901:certificate/bar",
    "region": "us-east-1",
    "status": "OK",
    "message": "foo.net is using DNS validation."
  }
]
```
This is why I think we need to keep the description (which contains the resource affected by the finding) as part of the deduplication fields.

I'll update the fields' order immediately

valentijnscholten: The title is not enough. In my opinion the `resource`  should be part of the hash_code. And severity should never be. Severity is only needed for some rare scanners that don't have enough data to otherwise make a good hash_code.
So I suggest:
```
'Cloudsploit Scan': ['title', 'resource'],
```
If `resource` is not globally unique (I see some short vpc ids), we should include the `region`.

axelpavageau: just to clear things up : do the fields configured in the de-duplication configuration refer the the attributes of the defect dojo finding, or to the attributes on the object in the source json ?
In the example source json above, `description` is identical between the 2 findings, so it's not a good candidate for deduplication. But in the parser, concatenate various fields to build the description https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/cloudsploit/parser.py#L31-L33, which I think makes it a good candidate for deduplication

valentijnscholten: ha, good point. My suggestion is not possible because hash_code fields are Defect Dojo fields and not parser fields :-)
A shortcoming of the hash_code mechanism. So the only choice is to use the description (or any other field we could use for that).

valentijnscholten (approved):

valentijnscholten (changes requested):
please remove severity

axelpavageau:
> please remove severity

done

valentijnscholten (approved):

ptrovatelli (approved):

PR #5036, opened 2021-09-03
Fix SonarQube migration error (#5028)

### LINKED ISSUE #5028: Migration from 2.1.0 to 2.2.0 fails during dojo.0120_sonarqube_test_and_clean step
**Bug description**
Migration command tool from 2.1.0 to 2.2.0 fails (python manage.py migrate) with this error: 
`Applying dojo.0120_sonarqube_test_and_clean...Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
psycopg2.errors.NotNullViolation: column "sonarqube_tool_config_id" contains null values`

It seems that the table dojo_sonarqube_product contains rows with "null" value of "sonarqube_tool_config_id".
Since we don't use Sonarqube , the deletion of all rows fix the problem.

**Steps to reproduce**
1. Migrate from 2.1.0 to 2.2.0
2. Run the command "python manage.py migrate"

**Expected behavior**
The migration process should proceed even with "null" sonarqube_tool_config_id

- [ ] Docker
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: EKS

kiblik:


valentijnscholten:
@kiblik thanks, could you base it against master? We already had 2 people reporting the issue, so it might be good to have a hotfix 2.2.1 release.

StefanFl (approved):

damiencarol (approved):

PR #5044, opened 2021-09-06
Feature/veracode pipeline import

jpbowie:
Added import for Veracode Pipeline JSON file output.
Found an example of the result file for testing here:
https://github.com/veracode/veracode-pipeline-scan-results-to-sarif/blob/master/test_resource/resultsToSarif.json


damiencarol (changes requested):
As Veracode pipeline support SARIF format conversion, it will be better to leverage the existing SARIF parser.

https://github.com/veracode/veracode-pipeline-scan-results-to-sarif#output-results-sarif

```
- name: Convert pipeline scan output to SARIF format
  id: convert
  uses: Veracode/veracode-pipeline-scan-results-to-sarif@v0.1.2
  with:
    pipeline-results-json: results.json
    output-results-sarif: veracode-results.sarif
    source-base-path-1: "^com/veracode:src/main/java/com/veracode"
    source-base-path-2: "^WEB-INF:src/main/webapp/WEB-INF"
    finding-rule-level: "3:1:0"
```

jpbowie:
The SARIF conversion support is only for a specific Github action provided by Veracode.  There isn't native SARIF support from the Pipeline Scanner.

damiencarol:
As explained in my previous comment andnaccording to the Veracode documentation, the best way to do scan is to use the Veracode tools and actions (see https://help.veracode.com/r/Github_Pipeline_Scan_Examples):

```YAML
name: pipeline-scan

on: push

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - name: build
        run: gradle clean build
  pipeline-scan:
    runs-on: ubuntu-latest
    steps:
      - name: Set up JDK 1.8
        uses: actions/setup-java@v1
        with:
          java-version: 1.8
      - name: Download the Pipeline Scanner
        uses: wei/curl@master
        with:
          args: -O https://downloads.veracode.com/securityscan/pipeline-scan-LATEST.zip
      - name: Unzip the Pipeline Scanner
        run: unzip pipeline-scan-LATEST.zip
      - name: Run Pipeline Scanner
        run: java -Dpipeline.debug=true -jar pipeline-scan.jar --veracode_api_id "${{secrets.VERACODE_API_ID}}" --veracode_api_key "${{secrets.VERACODE_API_KEY}}" --file "example.jar" --fail_on_severity="Very High, High"

```

This steps will produce a JSON file that you can convert in SARIF format with a specific Github action maintained by the Veracode team.

```YAML
...
- name: Convert pipeline scan output to SARIF format
  id: convert
  uses: Veracode/veracode-pipeline-scan-results-to-sarif@v0.1.2
  with:
    pipeline-results-json: results.json
    output-results-sarif: veracode-results.sarif
    source-base-path-1: "^com/veracode:src/main/java/com/veracode"
    source-base-path-2: "^WEB-INF:src/main/webapp/WEB-INF"
    finding-rule-level: "3:1:0"

- name: upload sarif file to repository
  uses: github/codeql-action/upload-sarif@v1
  with: # Path to SARIF file relative to the root of the repository
    sarif_file: veracode-results.sarif
```
Ref here : https://github.com/veracode/veracode-pipeline-scan-results-to-sarif#output-results-sarif

It's better to use the tools and actions created and maintained by Veracode team instead of creating new code maintained by us.

PS: You can also customize the name of the output file with this option:
```
  -jf JSON_OUTPUT_FILE, --json_output_file JSON_OUTPUT_FILE
                         Filename (in the current directory) to save results JSON. (default: results.json)
 ```

**Comment by Maffooch** [comment]
I agree with @damiencarol here. Maybe a good compromise could be to add a subsection to the [SARIF parser in the docs](https://defectdojo.github.io/django-DefectDojo/integrations/import/#sarif) outlining which tools (just Veracode pipeline for now) are compatible and can be used with dojo?

**Comment by valentijnscholten** [comment]
What if you run the Veracode pipeline elsewhere, not in GitHub. i.e. manually, from their saas product or in gitlab. I think the sarif converter is only available as a GitHub action is what @jpbowie states?

**Comment by jpbowie** [comment]
That was my understanding from the documentation, it is a GitHub Action.  If you are not using GitHub, you'll need to find another solution to convert the native pipeline scanner JSON output to SARIF.

File: dojo/tools/veracode_pipeline/parser.py [hunk 1]
```diff
@@ -0,0 +1,95 @@
...
+        return "Veracode Pipeline Scan Results"
+
+    def get_findings(self, filename, test):
+        if filename is None:
+            return
+
+        tree = json.load(filename)
+
+        if 'findings' not in tree:
+            return
```
damiencarol, line 36: Generate a `ValueError` and create a unit tests that check that

dojo/tools/veracode_pipeline/parser.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def get_label_for_scan_types(self, scan_type):
+        return "Veracode Pipeline Scan"
+
+    def get_description_for_scan_types(self, scan_type):
+        return "Veracode Pipeline Scan Results"
+
+    def get_findings(self, filename, test):
+        if filename is None:
+            return
```
damiencarol, line 31:
```suggestion
```
It will silently swallow an error (filename is never null for this parser)

dojo/tools/veracode_pipeline/parser.py
```diff
@@ -0,0 +1,95 @@
...
+    @classmethod
+    def __json_flaw_to_severity(cls, flaw):
+        return cls.vc_severity_mapping.get(int(flaw['severity']), 'Info')
+
+    @classmethod
+    def __json_flaw_to_finding(cls, scan_id, flaw, test):
+        finding = Finding()
+        finding.test = test
+        finding.mitigation = ''
+        finding.impact = ''
```
damiencarol, line 63:
```suggestion
```
Better no data instead of fake broken ones. These attributes are not mandatory.

dojo/tools/veracode_pipeline/parser.py
```diff
@@ -0,0 +1,95 @@
...
+        finding.static_finding = True
+        finding.dynamic_finding = False
+        finding.unique_id_from_tool = cls.__json_flaw_to_unique_id(scan_id, flaw)
+        finding.severity = cls.__json_flaw_to_severity(flaw)
+        finding.cwe = int(flaw['cwe_id'])
+        finding.title = flaw['issue_type']
+        finding.description = flaw['display_text']
+        finding.date = test.target_start
+        finding.is_mitigated = False
+        finding.mitigated = None
```
damiencarol, line 73:
```suggestion
```
Same comment

dojo/unittests/tools/test_veracode_pipeline_parser.py
```diff
@@ -0,0 +1,43 @@
...
+
+        finding = findings[1]
+        self.assertEqual('99f326aa-d648-4c11-ac0c-a583460c370d|2', finding.unique_id_from_tool)
+        self.assertEqual("Medium", finding.severity)
+        self.assertTrue(finding.static_finding)
+        self.assertFalse(finding.dynamic_finding)
+        self.assertEqual(456, finding.cwe)
+        self.assertEqual('Another Issue Type', finding.title)
+        self.assertEqual('another display text', finding.description)
+        self.assertEqual('different_filename', finding.sourcefile)
```
damiencarol, line 40: Most of the parsers use `file_path` and `sast_source_file_path`
`sourcefilepath` is deprecated.

dojo/tools/veracode_pipeline/parser.py
```diff
@@ -0,0 +1,95 @@
...
+        finding.sourcefile = _source_file_path
+        finding.sast_source_file_path = _source_file_path
+        finding.line_number = _source_line_number
+        finding.source_line = _source_line_number
+        finding.sast_source_line = _source_line_number
+        _sast_source_obj = _source_file_function
+        finding.sast_source_object = _sast_source_obj if _sast_source_obj else None
+
+        if 'flaw_match' in flaw and 'flaw_hash' in flaw['flaw_match']:
+            finding.hash_code = flaw['flaw_match']['flaw_hash']
```
damiencarol, line 93: Parser never set `hash_code`. it is an internal attribute that is managed entirely outside of the parser code.
You must not set it.
if your value `flaw['flaw_match']['flaw_hash']` is some kind of identifier, you should consider using `vuln_id_from_tool` and/or `unique_id_from_tool`

damiencarol (changes requested):

damiencarol:
> What if you run the Veracode pipeline elsewhere, not in GitHub. i.e. manually, from their saas product or in gitlab. I think the sarif converter is only available as a GitHub action is what @jpbowie states?

Ok. Yes you're right.

@jpbowie  Is it your use case? If so yes this parser will only be useful for you. Please take also into account that we already have another parser dedicated to the more common Veracode tool (not the pipeline)

Maffooch:
@jpbowie what are your thoughts on this?

damiencarol:
@jpbowie any progress on this one?

jpbowie:
Is there something specific I should be working on?

damiencarol:
@jpbowie don't you see my feedback on your PR? Do you want to close it?

PR #5049, opened 2021-09-07
Endpoint: Fix host validation

### LINKED ISSUE #5048: Endpoint validation does not match RFC URL standard
**Bug description**
The Host validation regex found [here](https://github.com/DefectDojo/django-DefectDojo/blob/00b425742b783ada0f432241c2812ac1257feb73/dojo/models.py#L1199) does not support all valid host names according to [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2). For example using the host  "_invalid._host.com" will not pass the regex even tho it should. See the Screenshots section to see why I think it should pass.

This leads to other problems like internal server errors when using the reimport api to import findings with these kinds of endpoints (starting with charakters other than [A-Za-z0-9]).

**Steps to reproduce**
1. Use the reimport API to import a findings with the endpoint "http://_invalid._host.com"

**Expected behavior**
The validation should not fail

**Deployment method**
Docker

**Screenshots** (optional)
image
image
image

**Console logs** (optional)
These logs were supplied by @EndPositive in Issue secureCodeBox/secureCodeBox#618
```
[30/Aug/2021 20:17:01] ERROR [django.request:224] Internal Server Error: /api/v2/reimport-scan/
  File "/app/./dojo/importers/reimporter/reimporter.py", line 162, in process_parsed_findings
    endpoint.clean()
  File "/app/./dojo/models.py", line 1218, in clean
    raise ValidationError(errors)
django.core.exceptions.ValidationError: ['Host "_invalid._host.com" has invalid format']
[...]
```

kiblik:
Fixes #5048

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1196,7 +1196,7 @@ class Endpoint(models.Model):
                 self.userinfo = None
 
         if self.host:
-            if not re.match(r'^[A-Za-z0-9][A-Za-z0-9_\.\-\+]+$', self.host):
+            if not re.match(r'^[A-Za-z0-9_\-\+][A-Za-z0-9_\.\-\+]+$', self.host):
                 try:
                     validate_ipv46_address(self.host)
                 except ValidationError:
--- a/dojo/unittests/test_endpoint_model.py
+++ b/dojo/unittests/test_endpoint_model.py
@@ -65,6 +65,8 @@ class TestEndpointModel(TestCase):
         endpoint.clean()
         endpoint = Endpoint(host='456_desktop')
         endpoint.clean()
+        endpoint = Endpoint(host='_invalid._host.com')
+        endpoint.clean()
 
     def test_invalid(self):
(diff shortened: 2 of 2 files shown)


damiencarol (approved):

StefanFl (approved):

dojo/models.py
```diff
@@ -1196,7 +1196,7 @@ def clean(self):
                 self.userinfo = None
 
         if self.host:
-            if not re.match(r'^[A-Za-z0-9][A-Za-z0-9_\.\-\+]+$', self.host):
+            if not re.match(r'^[A-Za-z0-9_\-\+][A-Za-z0-9_\.\-\+]+$', self.host):
```
damiencarol, line 1199: Should we remove the underscore? We are validating the host name here @kiblik

kiblik: @damiencarol, no. Underscore was added there to accept also hosts like `_invalid._host.com`

damiencarol: ok. so permissive way. :) I'm good with that

Issue #4586, opened 2021-06-01
Internal Server Error when merging Findings


37b:
https://github.com/DefectDojo/django-DefectDojo/blob/9e88e0480c9afd58b21354d7de38ea6f12fd0f2c/dojo/forms.py#L574

```
File "./dojo/forms.py", line 575, in __init__
--
self.fields.keyOrder = ['finding_to_merge_into', 'findings_to_merge', 'append_description', 'add_endpoints', 'append_reference']
AttributeError: 'dict' object has no attribute 'keyOrder'```

valentijnscholten:
Should be changed into `field_order` due to django 3:https://docs.djangoproject.com/en/3.0/ref/forms/api/#notes-on-field-ordering

valentijnscholten:
Are you able to do a PR and test it?

37b:
Yep I can do that

valentijnscholten:
fixed in dev / 2.3.0

PR #5064, opened 2021-09-08
fix error on merging findings due to django3 changes

37b:
fixes https://github.com/DefectDojo/django-DefectDojo/issues/4586


valentijnscholten:
@37b Thanks. We (not you) should have picked this up earlier. Can you look at removing that extra file that came along in the commit?

damiencarol (approved):

valentijnscholten (approved):

cw-acroteau:
Hi! I just manually applied this fix on my DefectDojo instance because the merging feature is critical for us. However, there is a bug in the fix. See PR #5144 for details.

PR #5068, opened 2021-09-09
Fix errors with Spotbugs 4.4.x

damiencarol:
Some data changed with version 4.4.0, the parser was only compatible with version 4.2.x

## Work done
- [x] improve robustness of the parser with missing attributes
- [x] add fix from #4292
- [x] fix type of CWE
- [x] add more unit tests

### DIFF
--- a/dojo/tools/spotbugs/parser.py
+++ b/dojo/tools/spotbugs/parser.py
@@ -1,8 +1,3 @@
-__author__ = 'bakalor'
-__maintainer__ = "Igor Bakalo"
-__email__ = "bigorigor.ua@gmail.com"
-__status__ = "Development"
-
 import re
 import html2text
 from defusedxml import ElementTree as ET
@@ -10,6 +5,7 @@ from dojo.models import Finding
 
 
 class SpotbugsParser(object):
+    """Parser for XML ouput file from Spotbugs (https://github.com/spotbugs/spotbugs)"""
 
     def get_scan_types(self):
         return ["SpotBugs Scan"]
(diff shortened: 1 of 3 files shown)


valentijnscholten (approved):

dojo/tools/spotbugs/parser.py
```diff
@@ -86,41 +82,52 @@ def get_findings(self, filename, test):
             for message in bug.itertext():
                 desc += message + '\n'
 
-            dupe_key = bug.get('instanceHash')
-
-            title = bug.find('ShortMessage').text
+            shortmessage_extract = bug.find('ShortMessage')
+            if shortmessage_extract:
+                title = shortmessage_extract.getText()
```
StefanFl, line 87: Something isn't working with the title in my test. Will investigate tomorrow, no time today anymore. For now just want to make sure the PR is not merged.

StefanFl (changes requested):

dojo/tools/spotbugs/parser.py
Maffooch: The title in my testing appears to be fine. The new scan (4.4.0) in this PR does not use `ShortMessage`, but the[ other sample scan ](https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/unittests/scans/spotbugs/many_findings.xml)does. Here's a screenshot of both versions titles.

version 4.4.0
image

Version 4.2.x
image

StefanFl: In the version 4.2 example you see the problem. The title of the first one should be **Potential Command Injection** which is the content of `ShortMessage`.  

It works for me when I change the detection of the title to:

```
shortmessage_extract = bug.find('ShortMessage')
if shortmessage_extract is not None:
    title = bug.find('ShortMessage').text
else:
    title = bug.get('type')
```

StefanFl (changes requested):

StefanFl:
I tested the SpotBugs scan on a project in my company. I get the additional information with SpotBugs 4.4 as well. The report doesn't look very much different from SpotBugs 4.2. Nevertheless it makes sense to make the SpotBugs import more robust.

dojo/tools/spotbugs/parser.py
StefanFl: The unit tests for the old version of the reports doesn't have a test for the title. With such a test you would have noticed the title has changed with this fix to a poorer version.

damiencarol:
@StefanFl how did you tested it? with the jar alone? I tested it with Maven and the dictionary were missing leading to the bugs reported.

StefanFl:
I use Maven as well together with the findsecbugs plugin.  This is my configuration:

```
  <plugin>
	<groupId>com.github.spotbugs</groupId>
	<artifactId>spotbugs-maven-plugin</artifactId>
	<version>${version.spotbugs.maven}</version>
			<configuration>
				<effort>Max</effort>
				<threshold>Low</threshold>
				<failOnError>true</failOnError>
				<includeFilterFile>spotbugs-include-filter.xml</includeFilterFile>
				<excludeFilterFile>spotbugs-exclude-filter.xml</excludeFilterFile>
				<plugins>
					<plugin>
						<groupId>com.h3xstream.findsecbugs</groupId>
						<artifactId>findsecbugs-plugin</artifactId>
						<version>1.10.1</version>
					</plugin>
				</plugins>
			</configuration>
	<dependencies>
	<!-- Overwrite dependency on SpotBugs if you want to specify the version of SpotBugs.
		SpotBugs itself is frequently several versions ahead of the spotbugs-maven-plugin -->
		<dependency>
			<groupId>com.github.spotbugs</groupId>
			<artifactId>spotbugs</artifactId>
			<version>${version.spotbugs}</version>
		</dependency>
	</dependencies>
</plugin>
```

where `version.spotbugs` is 4.4.0

damiencarol:
@StefanFl found the error with title and fixed it. Made more unit tests and fixes.

dojo/tools/spotbugs/parser.py
damiencarol: I added unit tests for title and saw the pb. You were right @StefanFl , I fixed my PR

StefanFl (approved):

PR #5071, opened 2021-09-09
SARIF parser - add more information to findings

StefanFl:
see title 😃

### DIFF
--- a/dojo/tools/sarif/parser.py
+++ b/dojo/tools/sarif/parser.py
@@ -139,19 +139,21 @@ def cve_try(val):
 
 
 def get_title(result, rule):
-    title = get_message_from_multiformatMessageString(result['message'], rule)
-    if rule is not None:
-        if title is None and 'shortDescription' in rule:
+    title = None
+    if 'message' in result:
+        title = get_message_from_multiformatMessageString(result['message'], rule)
+    if title is None and rule is not None:
+        if 'shortDescription' in rule:
             title = get_message_from_multiformatMessageString(rule['shortDescription'], rule)
-        elif title is None and 'fullDescription' in rule:
+        elif 'fullDescription' in rule:
             title = get_message_from_multiformatMessageString(rule['fullDescription'], rule)
-        elif title is None and 'name' in rule:
+        elif 'name' in rule:
             title = rule['name']
-        elif title is None:
+        elif 'id' in rule:
             title = rule['id']
 
     if title is None:
-        title = ''
+        raise ValueError('No information found to create a title')
 
     return textwrap.shorten(title, 150)
 
--- a/dojo/tools/sarif/parser.py
+++ b/dojo/tools/sarif/parser.py
@@ -149,10 +149,76 @@ def cve_try(val):
         return None
 
 
+def get_title(result, rule):
+    title = get_message_from_multiformatMessageString(result['message'], rule)
+    if rule is not None:
+        if title is None and 'shortDescription' in rule:
+            title = get_message_from_multiformatMessageString(rule['shortDescription'], rule)
+        elif 'fullDescription' in rule:
+            title = get_message_from_multiformatMessageString(rule['fullDescription'], rule)
+        elif title is None and 'name' in rule:
+            title = rule['name']
+        elif title is None:
+            title = rule['id']
+
+    if title is None:
+        title = ''
+
+    return textwrap.shorten(title, 150)
+
+
+def get_snippet(result):
+    snippet = None
+    if 'locations' in result:
+        location = result['locations'][0]
+        if 'physicalLocation' in location:
+            if 'region' in location['physicalLocation']:
+                if 'snippet' in location['physicalLocation']['region']:
+                    if 'text' in location['physicalLocation']['region']['snippet']:
+                        snippet = location['physicalLocation']['region']['snippet']['text']    
+    return snippet
+
+
+def get_description(result, rule):
+    description = ''
+    message = ''
+    if 'message' in result:
+        message = get_message_from_multiformatMessageString(result['message'], rule)
+        description += '**Result message:** {}\n'.format(message)
+    if get_snippet(result) is not None:
+        description += '**Snippet:** {}\n'.format(get_snippet(result))
+    if rule is not None:
+        if 'name' in rule:
+            description += '**Rule name:** {}\n'.format(rule.get('name'))
+        if 'shortDescription' in rule:
+            shortDescription = get_message_from_multiformatMessageString(rule['shortDescription'], rule)
+            if shortDescription != message:
+                description += '**Rule short description:** {}\n'.format(shortDescription)
+        if 'fullDescription' in rule:
+            fullDescription = get_message_from_multiformatMessageString(rule['fullDescription'], rule)
+            if fullDescription != message and fullDescription != shortDescription:
+                description += '**Rule full description:** {}\n'.format(fullDescription)
+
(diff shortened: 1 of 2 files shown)


StefanFl:
@damiencarol

dojo/unittests/tools/test_sarif_parser.py
```diff
@@ -297,25 +326,43 @@ def test_flawfinder(self):
...
+            description = '''**Result message:** buffer/sscanf:The scanf() family's %s operation, without a limit specification, permits buffer overflows (CWE-120, CWE-20).
+**Snippet:**       if (sscanf(argv[i], "%[^=]=%s", name, val) == 2) {
+**Rule name:** buffer/sscanf
+**Rule short description:** The scanf() family's %s operation, without a limit specification, permits buffer overflows (CWE-120, CWE-20).
+'''
+            self.assertEqual(description, finding.description)
             self.assertEqual("src/cli_main.cc", finding.file_path)
             self.assertEqual(482, finding.line)
             self.assertEqual("FF1021", finding.vuln_id_from_tool)
+            self.assertEqual('https://cwe.mitre.org/data/definitions/120.html', finding.references)
```
damiencarol, line 368: better to use `self.assertIn('http://...', finding.references)`

dojo/unittests/tools/test_sarif_parser.py
```diff
@@ -297,25 +326,43 @@ def test_flawfinder(self):
             finding = findings[0]
             self.assertEqual("random/setstate:This function is not sufficiently random for security-related functions such as key and nonce creation (CWE-327).", finding.title)
             self.assertEqual("Medium", finding.severity)
-            self.assertEqual("This function is not sufficiently random for security-related functions such as key and nonce creation (CWE-327).", finding.description)
+            description = '''**Result message:** random/setstate:This function is not sufficiently random for security-related functions such as key and nonce creation (CWE-327).
+**Snippet:**       is.setstate(std::ios::failbit);
```
damiencarol, line 330: Snippet can break the Markdown format.
We should use triple back-tick to escape snippet of code.

dojo/tools/sarif/parser.py
```diff
@@ -149,10 +138,92 @@ def cve_try(val):
...
+
+
+def get_description(result, rule):
+    description = ''
+    message = ''
+    if 'message' in result:
+        message = get_message_from_multiformatMessageString(result['message'], rule)
+        description += '**Result message:** {}\n'.format(message)
+    if get_snippet(result) is not None:
+        description += '**Snippet:** {}\n'.format(get_snippet(result))
```
damiencarol, line 178: We need to add ``` to escape the snippet code.

dojo/tools/sarif/parser.py
```diff
@@ -149,10 +138,92 @@ def cve_try(val):
...
+            title = get_message_from_multiformatMessageString(rule['shortDescription'], rule)
+        elif title is None and 'fullDescription' in rule:
+            title = get_message_from_multiformatMessageString(rule['fullDescription'], rule)
+        elif title is None and 'name' in rule:
+            title = rule['name']
+        elif title is None:
+            title = rule['id']
+
+    if title is None:
+        title = ''
```
damiencarol, line 154: Not a fan of producing fake blank data.

damiencarol (commented):
Many improvements. it's a good move forward.

dojo/unittests/tools/test_sarif_parser.py
StefanFl: Very good idea, done.

dojo/tools/sarif/parser.py
StefanFl: Very good idea, done.

dojo/tools/sarif/parser.py
StefanFl: Me neither, but title cannot be None. Shall we throw an error, when there is no title?

dojo/unittests/tools/test_sarif_parser.py
StefanFl: Why? This is exactly what I expect to be in the references.

damiencarol: most of the parsers use this syntax for references:
```Markdown
* [URL TITLE](https:///example.com/1)
* [URL2 TITLE](https:///example.com/2)
```
I want to keep them aligned.

dojo/tools/sarif/parser.py
damiencarol: I think we need some `\n` chars to make it works as expected.

dojo/tools/sarif/parser.py
damiencarol: I think maybe. what bother me is that the blank title could introduce some confusion for people (is it a bug (no title)?) and the code of DD (default de-duplication algorithm rely on the title I think)

dojo/unittests/tools/test_sarif_parser.py
StefanFl: I still don't understand to be honest. This is a test case, it just checks that it gets exactly the value it expects from the scan file.

dojo/tools/sarif/parser.py
StefanFl: Then I will change it to raise a `ValueError`. A lot of fields are tried to get a title. When all this fails, the report might be unusable anyway.

StefanFl: Done

damiencarol: awesome

dojo/unittests/tools/test_sarif_parser.py
damiencarol: Ok just saw that we take the reference from rule data and we can't have more than one URL or a text instead. So we are not in the same case as the other parsers.

dojo/tools/sarif/parser.py
StefanFl: Where do you need the `\n` characters? In one of my njsscan examples the snippet is

```
"snippet": {
    "text": "      await db.query(sql`\n        INSERT INTO users (id, email, first_name, last_name, password_hash)\n        VALUES (${join([\n          user.id,\n          user.email,\n          user.firstName,\n          user.lastName,\n          user.passwordHash,\n        ])})\n      `)"
    },
```

The description in the database is

```
**Result message:** Untrusted input concatinated with raw SQL query can result in SQL Injection.
**Snippet:** ```      await db.query(sql`
        INSERT INTO users (id, email, first_name, last_name, password_hash)
        VALUES (${join([
          user.id,
          user.email,
          user.firstName,
          user.lastName,
          user.passwordHash,
        ])})
      `)```
**Rule name:** NodeSqliInjection
```

In the UI it looks like this:
2021-09-12 16_58_39-View Finding _ DefectDojo

It looks fine to me.

damiencarol: Could you add some line feed to have the same look as for other parser (like the Dependency Check one ?)
image

StefanFl: I see, makes sense. With an additional space it looks like this now:
2021-09-12 20_35_57-View Finding _ DefectDojo

Maffooch (approved):
This is looking great!

dojo/tools/sarif/parser.py
damiencarol: awesome!

damiencarol (approved):

Issue #5070, opened 2021-09-09
Filter findings + sort by date result in "No findings found."


asbiin:
When you are on findings list (all findings, or a product's finding list), and you filter the list to, for instance filter only active findings, and apply filter, then if you sort the list by clicking the "Date" column, then the list disappear.

**Steps to reproduce**
1. Go to 'All findings'
2. Open the filters
3. Choose a filter, for instance "Active=yes", or "Verified=yes" (I did not tried all combinations)
4. Do not change the "Ordering"
5. Click on "Apply Filters"
6. Sort the result by clicking on the "Date" column header

-> the result is an empty list with `No findings found.` message

**Deployment method** *(select with an `X`)*
- [ ] Docker
- [x] Kubernetes
- [ ] GoDojo
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - DefectDojo version: 2.1.0

StefanFl:
Weird, but I can reproduce it. 

When I open the filters again, the field  **Tag name contains** is set with `¬_tags=¬_test__tags=¬_test__engagement__tags=¬_test__engagement__product__tags=¬_tag=`.

valentijnscholten:
Just saw the same, happens also when sorting by name.

valentijnscholten:
fixed in dev / 2.3.0

PR #5072, opened 2021-09-09
fix incorrect encoding of urls in finding list headings

valentijnscholten:
fixes https://github.com/DefectDojo/django-DefectDojo/issues/5070


StefanFl (approved):

valentijnscholten:
Not sure if this is the best way. I always learned that when building querystrings, every value should be url encoded seperately and only after that construct the querystring. But that would leave us still with "&not_...." causing problems. So escaping the whole things seems like the easy way out.

Maffooch (approved):

PR #4934, opened 2021-08-11
SonarQube: use severity from issue instead of rule

zeeshan811:
This will reflect updated finding severity when importing sonarqube findings using sonarqube api

PR #5073, opened 2021-09-09
fix custom severity for sonarqube

37b:
SonarQube allows admins to change the default severity for rules on a per-quality gate basis. The SonarQube API importer uses the default severity based on the data from the rule, rather than the (potentially) customized severity based on the issue.

This PR changes the severity derivation from rule to issue to allow custom severities to propagate into Defect Dojo.


valentijnscholten:
see https://github.com/DefectDojo/django-DefectDojo/pull/4934

Maffooch:
#4934 has been merged

PR #5074, opened 2021-09-10
Changed Dropdown Button Color to White

blakeaowens:
Old dropdown buttons were light blue, which blended too well with the header color. Changing them to white makes them more visible.

### DIFF
--- a/dojo/static/dojo/css/dojo.css
+++ b/dojo/static/dojo/css/dojo.css
@@ -52,14 +52,14 @@ a {
 .panel-default .btn-primary {
     background-color: transparent!important;
     border-style: none!important;
-    color: #546474!important;
-    border: 1px solid #546474!important;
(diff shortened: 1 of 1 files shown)


Maffooch (approved):
I think this looks more consistent than the old way. Good change!
Screen Shot 2021-09-09 at 9 24 16 PM

valentijnscholten:
Not sure yet if I like it. When you click on filtering, it gets a grey border which sticks around even after closing / unfocusing:
image

StefanFl:
We have the same effect already:

2021-09-10 16_32_10-Test Type List _ DefectDojo

Happens in Chrome but not in Firefox. It just more visible in the white version. Apart from that I personally like the white version. Looks much fresher.

StefanFl (approved):

PR #5076, opened 2021-09-10
Switch Finding.publish_date to Date type

damiencarol:
This break some UI things.
Also all the parsers have only date data and not date+time.


ptrovatelli:
hi @damiencarol ; thanks for mentioning me, I'm the one who created this field indeed. I used datetime because we had the exact time at the source. why not keep the exact time as we have it and truncate before sending to JIRA if JIRA doesn't support datetime?

ptrovatelli:
see https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/checkmarx_osa/parser.py#L57

damiencarol:
@ptrovatelli some UI things are broken. it seems that some checks are enable to parse/control date+time.
Could you help me move forward with this PR ?

valentijnscholten:
I don't think it has any relationship to JIRA is the publish date field is not propagated to JIRA. It's more the UI validation that prevents the user from saving the finding (and having it pushed to JIRA as part of that).

damiencarol:
I agree, but how do we control the "validation" part. I don't see anything in the view code to do that

Maffooch:
What if this could be circumvented by changing the date object to a date time object in the `save()` function for the finding? That would remove the inconsistency between parsers.

ptrovatelli:
hi guys; how about this? https://github.com/DefectDojo/django-DefectDojo/pull/5176

damiencarol:
Kinda lost here. I don't find where we can modify the validation of publish_date in the UI and modify it to validate a date instead.

valentijnscholten:
I vote for changing it into a DateField, easiest solution and publish "time" is irrelevant fin my opinion.

valentijnscholten:
@damiencarol there's a bunch of unrelated commits in the PR, can you take a look?

damiencarol:
@valentijnscholten done. rebased on last dev commit.

valentijnscholten:
Some migration issue it seems. Can you let us know what you tested this migration with? Ideally we have a test case for it, but maybe some manual testing is sufficient. We need to be sure the migration doesn't break if someone has timestamps in the dataset. I guess it will work as any existing value will have implicit 00:00:00.000 in there.

damiencarol:
@valentijnscholten let me check on my side. I will fix tests.

valentijnscholten:
Did you run tests on an instance with data populated in this field?

damiencarol:
@valentijnscholten Yes. But I think I mixed things with the rebase -_-

damiencarol:
@valentijnscholten doing more tests. But I had a database with findings with data.

valentijnscholten (approved):
Approved, but migration needs a number bump

Maffooch (approved):

PR #5082, opened 2021-09-11
Added Dependency Check Suppression parsing

emresaglam-dremio:
* Dependency Suppressions are now parsed.
* If the suppression has no notes, it'll tag it
  [missingdoc]
* If the suppression has notes, it'll populate
  the mitigation field with it.
* Also added some tests related to parsing
* Fixed Docker.md with correct examples

Overall I modified the dependency checker to parse suppression fields coming from the XML file. 

If it has notes about the suppression, it will populate the migrations with the note, it'll mark the finding Inactive, and Risk Accepted. 
If it doesn't have notes, it'll still mark it as Inactive, Risk Accepted but will also tag the finding as "missingdoc"

I also added some tests related. 

I don't do much PRs as I don't consider myself a good programmer. I hope I'm doing everything right :) Thanks!

### DIFF
--- a/dojo/tools/dependency_check/parser.py
+++ b/dojo/tools/dependency_check/parser.py
@@ -162,6 +162,10 @@ class DependencyCheckParser(object):
         dependency_filename, dependency_filepath = self.get_filename_and_path_from_dependency(dependency, related_dependency, namespace)
         # logger.debug('dependency_filename: %s', dependency_filename)
 
+        # In the begining there were no tags!
+        # Tags will be used to (well duh...) tag the missingdoc and related fields.
+        tags = []
+
         if dependency_filename is None:
             return None
 
@@ -172,6 +176,8 @@ class DependencyCheckParser(object):
         else:
             cwe_field = self.get_field_value(vulnerability, 'cwe', namespace)
         description = self.get_field_value(vulnerability, 'description', namespace)
+        # I need the notes field since this is how the suppression is documented.
+        notes = self.get_field_value(vulnerability, 'notes', namespace)
 
         cve = name[:28]
         if cve and not cve.startswith('CVE'):
@@ -196,6 +202,10 @@ class DependencyCheckParser(object):
         # startswith CWE-XXX
         stripped_name = re.sub(r'^CWE-\d+', '', stripped_name).strip()
 
+        if component_name is None:
+            logger.error("component_name was None!")
+            return None
+        
         title = '%s:%s | %s(in %s)' % (component_name.split(':')[-1], component_version,
             (stripped_name + ' ' if stripped_name else '') + (description if len(stripped_name) < 25 else ''),
             dependency_filename)
@@ -238,7 +248,33 @@ class DependencyCheckParser(object):
                                      'source: {1}\n' \
                                      'url: {2}\n\n'.format(name, source, url)
 
-        mitigation = 'Update ' + component_name + ':' + component_version + ' to at least the version recommended in the description'
+        if related_dependency is not None:
+            # If we have a related vulnerability. I want to tag it as related
+            tags.append("related")
+
+        # Check if this is a vulnerability or a suppressed vulnerability.
+        if vulnerability.tag == "{}suppressedVulnerability".format(namespace):
+            #logger.info("This is a suppressed vulnerability!")
+            # If there are no notes in the suppressed vulnerability, we need to know and correct it
+            # Tag the suppressed vulneability as missingdoc
+            if notes == "":
+                # We suppressed this vulnerability, but never documented it. Not cool!
+                # Tag it so that we will have a way to find and document it for the future. 
+                notes = "Document on why we are suppressing this vulnerability is missing!"
+                tags.append("missingdoc")
+            mitigation = '**This vulnerability is mitigated and/or suppressed:** {}'.format(notes)
+            # Since we are now confident that this is a suppressed vulnerability, mark it as mitigated
+            # risk accepted and inactive. Poor man's risk acceptance for now.
+            # TODO: Figure out a way to document the Risk acceptance field(s) à la "Add Risk Acceptance" in the UI.
+            is_mitigated = True
+            risk_accepted = True
+            active = False
+
+        else:
+            mitigation = 'Update {}:{} to at least the version recommended in the description'.format(component_name, component_version)
+            is_mitigated = False
+            risk_accepted = False
+            active = True
 
         return Finding(
             title=title,
@@ -249,6 +285,10 @@ class DependencyCheckParser(object):
             description=description,
             severity=severity,
             mitigation=mitigation,
+            tags=tags,
+            is_mitigated = is_mitigated,
+            risk_accepted = risk_accepted,
+            active = active,
             static_finding=True,
             references=reference_detail,
             component_name=component_name,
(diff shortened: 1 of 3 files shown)


valentijnscholten:
It's hard to decide whether they should be risk accepted or false positive if suppressed. Risk accepted is only allowed to be set accompanied by a Risk_Acceptance model instance. Or "simple risk acceptance" should be enabled on the engagement. Maybe it's better to just set the finding as active=False and that's it? So no false_p or mitigated or risk_accepted.

emresaglam-dremio:
> It's hard to decide whether they should be risk accepted or false positive if suppressed. 
A suppression can be because of three reasons. False Positive, Risk Acceptance and Mitigations. 

For example we might have a library that is flagged that we use only at build time. I accept the Risk with this one. 
Or we might have a vulnerable library that we never use the vulnerable part. It's mitigated. 
Or False Positive is well... False Positive.

Unless we have a tag for specific reason, we can never inherit this properly from the Dependency Scan. The work around for that is to have a standard per `<note>` tag's text. This is not flexible as my standard would be different than someone else's.

> Risk accepted is only allowed to be set accompanied by a Risk_Acceptance model instance. Or "simple risk acceptance" should be enabled on the engagement. Maybe it's better to just set the finding as active=False and that's it? So no false_p or mitigated or risk_accepted.

This wouldn't work for my model as it'll be double effort on how and why we are suppressing the vulnerabilities flagged in DD. My DD users are not only engineers ;) My goal is to be able to give access to the support for, "hey we scanned you here are the issues" types of questions.

valentijnscholten:
Why do you want to import them if they are suppressed anyway. Just to have an official record in Defect Dojo? I think it's fine to come up with some note form that instructs Defect Dojo to set the finding as false positive, mitigated or risk accepted. As I said risk accepted can only be set with a Risk_Acceptance model instance unless the engagement allows simple risk acceptance.

emresaglam-dremio:
> Why do you want to import them if they are suppressed anyway. Just to have an official record in Defect Dojo?

The reason is this ⬆️ since I want to consolidate all our vulnerabilities. And also what I wrote at the end of my previous comment. I want to empower our support to be able to use DD per component and figure out appropriate answers for customers.

Maffooch:
> Maybe it's better to just set the finding as active=False and that's it?
This is the most community friendly approach in my opinion. This coupled with a `suppressed` tag or something to that nature per finding would be ideal.

emresaglam-dremio:
I’ll remove the other flags and send the changes. I’ll keep them for my purposes. 

Any strong opinions on missingdoc tag? 

I might open a can of worms and implement per scanner optional flags :)

valentijnscholten:
Yeah using tags might be a good idea as well. A new status boolean would have a big impact. I still think it might be good for the future to switch to a resolutionbased model like JIRA has. It becomes easier then to have customized resolutions etc without affecting the database schema.

damiencarol:
Read the discussion with big interest. 

> Maybe it's better to just set the finding as active=False and that's it?

I completely agree with this comment. It seems here that to implement the good data in the DD model we should take into account a lot of things. It's because these three use cases `False Positive`, `Risk Acceptance` and `Mitigations` are not implemented in the same way.

So better to just set the `active` flag for now and maybe also set the `false_p` (false positive) when the finding is suppressed.

Even with this minimum modification we're improving a lot the Dependecy Check parser.

> A new status boolean would have a big impact. I still think it might be good for the future to switch to a resolutionbased model like JIRA has. It becomes easier then to have customized resolutions etc without affecting the database schema.

100% agree. And we need to design this part to be generic and easy to implement in parsers, like dict and managed in upper layers by the (re)importer.

@emresaglam-dremio could you update the PR this way (only manipulate `active=False` when suppressed and also setting `false_p` flag when it's true)?

emresaglam-dremio:
> @emresaglam-dremio could you update the PR this way (only manipulate active=False when suppressed and also setting false_p flag when it's true)?

I'll do that today. There is not a good way to implement a reliable false positive detection from the dependency check reports. And my code doesn't flag anything as false positive. I'll leave it that way.

Maffooch (approved):
This looks good to me now. Thanks!

dojo/tools/dependency_check/parser.py
```diff
@@ -196,6 +200,10 @@ def get_finding_from_vulnerability(self, dependency, related_dependency, vulnera
         # startswith CWE-XXX
         stripped_name = re.sub(r'^CWE-\d+', '', stripped_name).strip()
 
+        if component_name is None:
+            logger.error("component_name was None!")
+            return None
+
```
valentijnscholten, line 206: ERROR log level is reserved for serious flaws that need to be looked into by the sysadmin. so this should be warning at most. and if you want it to be more useful I would add some context info about the finding that is being parsed or something relevant.

dojo/tools/dependency_check/parser.py
```diff
@@ -238,7 +246,21 @@ def get_finding_from_vulnerability(self, dependency, related_dependency, vulnera
...
                                      'url: {2}\n\n'.format(name, source, url)
 
-        mitigation = 'Update ' + component_name + ':' + component_version + ' to at least the version recommended in the description'
+        if related_dependency is not None:
+            tags.append("related")
+
+        if vulnerability.tag == "{}suppressedVulnerability".format(namespace):
+            if notes == "":
+                notes = "Document on why we are suppressing this vulnerability is missing!"
+                tags.append("missingdoc")
```
valentijnscholten, line 255: not sure if we want to tag this also, but if we do, can we make it something more relevant, i.e. no_suppression_reason or something?

valentijnscholten: and can we add a finding to the unit test to cover this case?

valentijnscholten (changes requested):
see notes

valentijnscholten:
Can you update the description of this PR as well as the findings are no longer risk accepted when supressed?
```
If it has notes about the suppression, it will populate the migrations with the note, it'll mark the finding Inactive, and Risk Accepted.
If it doesn't have notes, it'll still mark it as Inactive, Risk Accepted but will also tag the finding as "missingdoc"
```

dojo/tools/dependency_check/parser.py
emresaglam-dremio: > not sure if we want to tag this also, but if we do, can we make it something more relevant, i.e. no_suppression_reason or something?

I think so too. `missingdoc` is too generic. I'll change it with `no_suppression_document`.

dojo/tools/dependency_check/parser.py
emresaglam-dremio: It's now logging as warning. I also changed the logic to use `dependency_filename` when `component_name` cannot be found. It works really good. It doesn't drop the finding. Thanks for flagging this.

dojo/tools/dependency_check/parser.py
emresaglam-dremio: > and can we add a finding to the unit test to cover this case?

There is a test for that. However it's not for the tag, it's for the Mitigation field. Well, the tags are... complicated :) I will have to spend some time for this.

dojo/tools/dependency_check/parser.py
```diff
@@ -284,61 +310,28 @@ def get_findings(self, filename, test):
...
-                        relatedDependencies = dependency.find(namespace + 'relatedDependencies')
-                        if relatedDependencies:
-                            for relatedDependency in relatedDependencies.findall(namespace + 'relatedDependency'):
-                                finding = self.get_finding_from_vulnerability(dependency, relatedDependency, vulnerability, test, namespace)
-                                self.add_finding(finding, dupes)
-
         return list(dupes.values())
 
-
+# From Emre - I left the original comments here below and moved them to the bottom.
```
damiencarol, line 334: Just remove them. It make no sense to have a report in the code as comment

dojo/tools/dependency_check/parser.py
```diff
@@ -358,3 +351,35 @@ def get_findings(self, filename, test):
 #     <software>cpe:2.3:a:oracle:rapid_planning:12.1:*:*:*:*:*:*:*</software>
 #     <software>cpe:2.3:a:netapp:snap_creator_framework:-:*:*:*:*:*:*:*</software>
 # </vulnerableSoftware>
+
+
+# TODO relateddependencies are ignored in this parser, but should be imported because you might miss vulnerable dependencies otherwise
```
damiencarol, line 356: Move this comment in the comment of the parser

damiencarol (changes requested):
There are few minor fixes to do

dojo/tools/dependency_check/parser.py
emresaglam-dremio: I feel like this TODO is done already since the parser is importing related dependencies and tagging them as related in DD. I'll add a test for the related tag. Are you OK for me to remove this TODO comment blob?

emresaglam-dremio: It looks like I already did add a test for the related tags. :) Line 734 in dojo/unittests/tools/test_dependency_check_parser.py

damiencarol (approved):

damiencarol:
@valentijnscholten up for review I think.

dojo/tools/dependency_check/parser.py
```diff
@@ -238,8 +246,22 @@ def get_finding_from_vulnerability(self, dependency, related_dependency, vulnera
...
+                notes = "Document on why we are suppressing this vulnerability is missing!"
+                tags.append("no_suppression_document")
+            mitigation = '**This vulnerability is mitigated and/or suppressed:** {}'.format(notes)
+
+            active = False
+            tags.append("suppressed")
+
+        else:
+            mitigation = 'Update {}:{} to at least the version recommended in the description'.format(component_name, component_version)
+            description += '\nFilepath: ' + str(dependency_filepath)
```
valentijnscholten, line 263: I would add these lines even when the vulnerability is suppressed (after the mitigation line set above)

valentijnscholten (changes requested):
see comment

valentijnscholten:
Can you add some short explanation to `docs/content/en/integrations/import.md` as well?

dojo/tools/dependency_check/parser.py
valentijnscholten: resolved.

valentijnscholten (approved):
just waiting for the tests to become green so we can merge it.

damiencarol:
@emresaglam-dremio one unit test is failling. could you fix it?

valentijnscholten:
> @emresaglam-dremio one unit test is failling. could you fix it?

that was my earlier attempt at fixing it with some copy-and-pasting via the GitHub web UI :-)

PR #5084, opened 2021-09-12
Make scan type list rely on Dynamic parser infra.

damiencarol:
Simplify code around scan type lists and parser loading/validating making some serializer/forms don't rely on DB but on code directly.


Maffooch:
Why is there such an urgency to remove the test_type table from the DB?

damiencarol:
@Maffooch Sorry the text was confusing. I will change the text of the PR. The goal is to have  Test_type less "fixed.
I mean, as Test_type could depends on new parsers loaded dynamically, it doesn't make sense to rely on the data in the DB for the UI.

dojo/tools/cobalt_api/parser.py
```diff
@@ -5,20 +5,26 @@
...
     """
     Import from Cobalt.io API /findings
     """
 
     def get_scan_types(self):
-        return ["Cobalt.io API Import"]
+        return [SCAN_COBALTIO_API]
 
     def get_label_for_scan_types(self, scan_type):
-        return "Cobalt.io API Import"
```
valentijnscholten, line 17: if you change the scan_type value you probably going to have to write a data migration to convert existing entries?

valentijnscholten: oh wait you're extracting string constants it seems nvm

dojo/management/commands/initialize_test_types.py
```diff
@@ -1,9 +1,14 @@
...
 class Command(BaseCommand):
     help = 'Initializes Test_Types'
 
     def handle(self, *args, **options):
-        initialize_test_types()
+        # called by the initializer to fill the table with test_types
+        for scan_type in PARSERS:
+            test_type, created = Test_Type.objects.get_or_create(name=scan_type)
+            if created:
+                test_type.save()
```
Maffooch, line 14: Saving is not necessary with `get_or_create`

dojo/product/views.py
```diff
@@ -1160,6 +1160,9 @@ def edit_meta_data(request, pid):
 @user_is_authorized(Product, Permissions.Finding_Add, 'pid', 'staff')
 def ad_hoc_finding(request, pid):
     prod = Product.objects.get(id=pid)
+    test_type, created = Test_Type.objects.get_or_create(name="Pen Test")
+    if created:
+        test_type.save()
```
Maffooch, line 1165: Saving is not necessary with `get_or_create`

Maffooch (commented):
Okay that sounds better. I was under the impression the goal was to remove the entire table from the database.

dojo/tools/factory.py
```diff
@@ -23,20 +19,12 @@ def register_parser(scan_type, parser):
...
-    if scan_type in PARSERS:
-        # create dynamicaly in DB
-        test_type, created = Test_Type.objects.get_or_create(name=scan_type)
-        if created:
-            test_type.save()
-        if not test_type.active:
-            raise ValueError(f"Parser {scan_type} is not active")
+def get_parser(scan_type):
+    """Return a parser by the scan type"""
+    if True:
```
StefanFl, line 24: Why do you use `if True:` here? The `else` path can never be reached.

damiencarol: My idea was to add a settings with a regex to enable/disable parser. I forgot it 😅

dojo/management/commands/initialize_test_types.py
damiencarol: I just moved the code but I can refactor it this way.

damiencarol: Fixed

dojo/product/views.py
damiencarol: I just moved the code but I can refactor it this way.

damiencarol: Fixed

dojo/tools/factory.py
damiencarol: Fixed with my last commits. I added a new setting to filter/allow parsers instead of a always `True` value like in the previous releases.

Maffooch (approved):
Things look good to me

dojo/tools/factory.py
StefanFl: Test_Types have a flag `active`, wouldn't it be better to use that instead of a new setting? Admins can set that easily via the UI instead of editing settings.

damiencarol: Made a modification to support the flag `active`. Also I created few unit tests to check that.
You're right, maybe it's better to support both.
Thanks for all your feedback, this make the PR better :)

StefanFl:
@damiencarol I will approve this merge request if you do me one favour 😃: The list of parsers at the import page is not sorted. Please use `get_choices_sorted()` for the scan types in `dojo/engagement/views.py/import_scan_results`.

damiencarol:
I tried but it's not that simple. To do that we need to modify `TestForm` to fake the link between `scan_type` and the `test_type` attribute of a `Test`.
I failed to found an elegant way to do it. Will try again today.

StefanFl:
Not sure we talk about the same thing. I meant using `get_choices_sorted()` in line https://github.com/DefectDojo/django-DefectDojo/blob/b44af80ddfb4535eba85b72e5ec707e96db61996/dojo/engagement/views.py#L664 

That does the trick.

damiencarol:
@StefanFl done.

StefanFl (approved):

PR #5099, opened 2021-09-14
ZAP parser maintenance

damiencarol:
# Work done
- [x] remove query and fragment in endpoint because with some configuration, the tool generate them on-the-go and it produces a lot of fake endpoints
- [x] don't parse URLs but rely on `Endpoint.from_uri()`
- [x] implement scanner confidence
- [x] escape HTML encoding in data with `html2text` (clean description, mitigation and references)
- [x] add many checks in unit tests
- [X] implement `vuln_id_from_tool`

This PR fixes issue #4697 

image

The upload/re-upload is better after this PR

image

References are better managed and HTML is well translated to Markdown

(new)
image


Maffooch (approved):
Looks good!

valentijnscholten (approved):

PR #5103, opened 2021-09-14
Fix ssl_labs and nsp parsers, correct occurences of `self.items`.

JOT85:
In e515c0e4030efec056e0ab5536ffe7b6fcf715db, the expected behaviour of the `get_findings` method changed. Instead of setting `self.items`, it is instead expected to return the items.

The `ssl_labs` and `nsp` parsers didn't seem to have been updated to this behaviour, causing errors when they are used. This PR fixes this behaviour.

In addition, `self.items` was sometimes set or mentioned in comments but not used, I've removed these.

Also, you'll notice the *FIXME* comment above the loop in the `ssl_labs` parser. The loop always returns at the end of the first iteration so this should be replaced with code to simply get and parse the first item if this behaviour is kept. However, I feel like it would make more sense to trigger an error if there are multiple reports, since some won't be handled?
Another alternative would be to extract only the entry with the correct host matching the endpoint of the import, however that seems like overkill for something that's not likely to happen!

### DIFF
--- a/dojo/tools/blackduck_component_risk/parser.py
+++ b/dojo/tools/blackduck_component_risk/parser.py
@@ -20,8 +20,7 @@ class BlackduckComponentRiskParser(object):
 
     def get_findings(self, filename, test):
         """
-        Function initializes the parser with a file and sets the
-        self.items (eventually).
+        Function initializes the parser with a file and returns the items.
         :param filename: Input in Defect Dojo
         :param test:
         """
--- a/dojo/tools/npm_audit/parser.py
+++ b/dojo/tools/npm_audit/parser.py
@@ -24,7 +24,6 @@ class NpmAuditParser(object):
 
     def parse_json(self, json_output):
         if json_output is None:
-            self.items = []
             return
         try:
             data = json_output.read()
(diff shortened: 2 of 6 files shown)


damiencarol (approved):

damiencarol:
Could be good also to have unit tests for these parsers, @JOT85 do you mind creating some?

JOT85:
I'll write some tests, the problem will be that I don't have example datasets for most import types. But I'll write a test which runs all the parsers it has an example for, then examples can be easily added.

JOT85:
Ignore me, there's already example files and tests!

damiencarol:
don't hesitate to create draft pull request also.

valentijnscholten:
Only ssl_parser and nsp need tests it seems :-)

damiencarol:
@valentijnscholten  I saw 2 other parsers without tests:
* "Dawnscanner"
* "Bundler audit"

damiencarol:
@JOT85 any news on this one?

JOT85:
@damiencarol I have started work on it, it should be done by the end of the week.

Maffooch:
Hi @JOT85  how are things coming on this?

JOT85:
@damiencarol thanks for writing the other two unit tests.

This took so long because I embarked on trying to complete the integration tests in [tests/Import_scanner_test.py](https://github.com/DefectDojo/django-DefectDojo/blob/master/tests/Import_scanner_test.py).

I will still complete these. If you'd like to merge this now, I could create another PR once the integration tests are completed?

PR #5111, opened 2021-09-16
Fix Bundler Audit for version 0.9.x.x and unittests maintenance

damiencarol:
# Work done
- [x] **Bundle Audit parser:** Fixes broken with version 0.9.x.x (format change attribute `CVE`)
- [x] **Bundle Audit parser:** implements `component_name+version` and `cve` 
- [x] remove few `print()` and `logger.warn()` that polluted logs. Switching some piece of code with normal `if/else` branch instead of exception catching
- [x] renaming every unit tests that didn't followed conventions


valentijnscholten (approved):

dojo/tools/burp/parser.py
```diff
@@ -234,7 +234,7 @@ def get_item(item_node, test):
     cwes = do_clean_cwe(item_node.findall('vulnerabilityClassifications'))
     if len(cwes) > 1:
         # FIXME support more than one CWE
-        logger.warning(f"more than one CWE for a finding {cwes}. NOT supported by parser API")
+        logger.debug(f"more than one CWE for a finding {cwes}. NOT supported by parser API")
```
StefanFl, line 237: Why do you suppress this message? Might be interesting for a production environment to see why there is no CWE in a finding.

dojo/tools/cyclonedx/parser.py
```diff
@@ -167,7 +167,7 @@ def manage_vulnerability(
         cwes = self.get_cwes(vulnerability, ns)
         if len(cwes) > 1:
             # FIXME support more than one CWE
-            LOGGER.warning(
+            LOGGER.debug(
```
StefanFl, line 170: Why do you suppress this message? Might be interesting for a production environment to see why there is no CWE in a finding. The same for the other parsers in this PR.

dojo/tools/burp/parser.py
damiencarol: No. It's not what you're thinking. I created this log to notify that the data in the report have more than one CWE but the current data structure only supports one.
But many many parsers have this design problem. It doesn't make sense to flood the logs with these messages because our implementation is not perfect.
Also, the parsers that generate these warnings use the first one when there is more than one CWE.

damiencarol: I really think that we should modify our implementation of Finding to support more than zero or only one CWE.
Maybe an array.

StefanFl (approved):

PR #5121, opened 2021-09-19
Custom Test_Type for parsers (modify "Found by" dynamicaly)

### LINKED ISSUE #3797: Display actual scanner name in 'Found By' column
When importing scan results, DefectDojo displays the importer name in the 'Found By' column. This is fine for specific importers, however for generic importers like SARIF or GitLab SAST it would be more useful to see the actual scanner name in the 'Found By' column (if available in the SARIF or GitLab report). 

Especially if you are using multiple tools that generate the same type of reports (i.e. multiple tools that generate a GitLab SAST report), just seeing `GitLab SAST Report` in the list of findings is not very useful. More useful would be to see the actual scanner names like `PMD`, `ESLint`, `Brakeman`, or `Fortify`.

image

damiencarol:
## Description

Add feature to customize `Test` from parser data.

This will allow some parsers to handle meta-format like SARIF.

Fixes #3797

## Implementation

1. Test: added a `scan_type` attribute to save the scan_type used during import as string
2. Importer/Reimporter: add a new method `get_tests()` for parsers that are able to build findings **and** tests
3. The importer now take into account the data of the parsers to create the Test.

The core of the modifications are in `dojo/importers/importer/importer.py`.
I implemented like this:
```
 *if the parser support creating Test* =>
   1. we parse the findings and tests
   2. we create the test with these data (with a customized type and name)
 * else *
   <old behavior>
```

Once this PR is merged I plan to add more attributes like description and more parsers that support this new interface like the generic parser.

## Screenshots
For ex:
image

The name of the Test can come from the data now (ex for SARIF test file):
image

### DIFF
--- a/dojo/importers/importer/importer.py
+++ b/dojo/importers/importer/importer.py
@@ -27,12 +27,12 @@ deduplicationLogger = logging.getLogger("dojo.specific-loggers.deduplication")
 
 class DojoDefaultImporter(object):
 
-    def create_test(self, scan_type, test_type_str, engagement, lead, environment, tags=None,
+    def create_test(self, scan_type, test_type_name, engagement, lead, environment, tags=None,
                     scan_date=None, version=None, branch_tag=None, build_id=None, commit_hash=None, now=timezone.now(),
                     sonarqube_config=None, cobaltio_config=None):
 
         test_type, created = Test_Type.objects.get_or_create(
-            name=test_type_str)
+            name=test_type_name)
 
         if created:
             logger.info('Created new Test_Type with name %s because a report is being imported', test_type.name)
--- a/dojo/importers/importer/importer.py
+++ b/dojo/importers/importer/importer.py
@@ -309,7 +309,7 @@ class DojoDefaultImporter(object):
             #
             # we also aggregate the label of the Test_type to show the user the original scan_type
             # only if they are different. This is to support meta format like SARIF
-            # so a report that have the label 'CodeScanner' will be changed to 'SARIF > CodeScanner'
+            # so a report that have the label 'CodeScanner' will be changed to 'CodeScanner Scan (SARIF)'
             test_type_name = scan_type
             if len(tests) > 0:
                 test_type_name = tests[0].type + " Scan"
@@ -318,12 +318,17 @@ class DojoDefaultImporter(object):
             test = self.create_test(scan_type, test_type_name, engagement, lead, environment, scan_date=scan_date, tags=tags,
                                 version=version, branch_tag=branch_tag, build_id=build_id, commit_hash=commit_hash, now=now,
                                 sonarqube_config=sonarqube_config, cobaltio_config=cobaltio_config)
+            # This part change the name of the Test
+            # we get it from the data of the parser
             test_raw = tests[0]
             if test_raw.name:
                 test.name = test_raw.name
                 test.save()
 
             logger.debug('IMPORT_SCAN parser v2: Parse findings (aggregate)')
+            # currently we only support import one Test
+            # so for parser that support multiple tests (like SARIF)
+            # we aggregate all the findings into one uniq test
             parsed_findings = []
             for test_raw in tests:
                 parsed_findings.extend(test_raw.findings)
--- a/dojo/importers/importer/importer.py
+++ b/dojo/importers/importer/importer.py
@@ -19,18 +19,20 @@ import dojo.jira_link.helper as jira_helper
 import base64
 import logging
 
+from dojo.tools.factory import get_parser
+
 logger = logging.getLogger(__name__)
 deduplicationLogger = logging.getLogger("dojo.specific-loggers.deduplication")
 
 
 class DojoDefaultImporter(object):
 
-    def create_test(self, scan_type, engagement, lead, environment, tags=None,
+    def create_test(self, scan_type, test_type_str, engagement, lead, environment, tags=None,
                     scan_date=None, version=None, branch_tag=None, build_id=None, commit_hash=None, now=timezone.now(),
                     sonarqube_config=None, cobaltio_config=None):
 
         test_type, created = Test_Type.objects.get_or_create(
-            name=scan_type)
+            name=test_type_str)
 
         if created:
             logger.info('Created new Test_Type with name %s because a report is being imported', test_type.name)
@@ -39,6 +41,7 @@ class DojoDefaultImporter(object):
             engagement=engagement,
             lead=lead,
             test_type=test_type,
+            scan_type=scan_type,
             target_start=scan_date if scan_date else now.date(),
             target_end=scan_date if scan_date else now.date(),
             environment=environment,
@@ -67,15 +70,15 @@ class DojoDefaultImporter(object):
         logger.debug('starting import of %i items.', len(items) if items else 0)
         i = 0
         for item in items:
-            sev = item.severity
-            if sev == 'Information' or sev == 'Informational':
-                sev = 'Info'
+            # FIXME hack to remove when all parsers have unit tests for this attribute
+            if item.severity.lower().startswith('info') and item.severity != 'Info':
+                item.severity = 'Info'
 
-            item.severity = sev
-            item.numerical_severity = Finding.get_numerical_severity(sev)
+            item.numerical_severity = Finding.get_numerical_severity(item.severity)
 
-            if minimum_severity and (Finding.SEVERITIES[sev] >
+            if minimum_severity and (Finding.SEVERITIES[item.severity] >
                     Finding.SEVERITIES[minimum_severity]):
+                # finding's severity is below the configured threshold : ignoring the finding
                 continue
 
             item.test = test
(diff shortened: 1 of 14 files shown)


valentijnscholten:
Can you explain a bit how you solved the problem? It's hard to digest from the GitHub diff with 11 files changed but only a few lines without context.

damiencarol:
Added more description. @valentijnscholten

damiencarol:
merged #5125

StefanFl:
Wanted to try it out before doing the actual code review, but got an exception while trying to import a SARIF report:

```
uwsgi_1         | [21/Sep/2021 06:24:50] ERROR [dojo.engagement.views:614] local variable 'test_raw' referenced before assignment
uwsgi_1         |   File "/app/./dojo/engagement/views.py", line 599, in import_scan_results
uwsgi_1         |     test, finding_count, closed_finding_count = importer.import_scan(scan, scan_type, engagement, user, environment, active=active, verified=verified, tags=tags,
uwsgi_1         |   File "/app/./dojo/importers/importer/importer.py", line 312, in import_scan
uwsgi_1         |     if test_raw.name:
uwsgi_1         | UnboundLocalError: local variable 'test_raw' referenced before assignment
```

damiencarol:
@StefanFl missed one line in the "merge" :sweat: (Fixed)
Will add unit test to check this code path.

valentijnscholten:
I think I understand the first two points, changed the wording slightly. What is meant by point 3? 
I'm not sure, but it looks like filtering and reports etc need some updates as I expect we want to user to be able to choose to filter by any test type (scan type) used by any import? i.e. if I import a SARIF report with tool name "Valentijns Mighty Security Scanner", the user should be able to filter / search for findings based on that tool name?

damiencarol:
> I think I understand the first two points, changed the wording slightly. What is meant by point 3?

Added details to 3. I meant to create Test. The importer customize the name and the Test_Type from the data.

> I'm not sure, but it looks like filtering and reports etc need some updates as I expect we want to user to be able to choose to filter by any test type (scan type) used by any import? i.e. if I import a SARIF report with tool name "Valentijns Mighty Security Scanner", the user should be able to filter / search for findings based on that tool name?

Yes. Everything works as before. Users will be able to filter by `Valentijns Mighty Security Scanner` if the data in SARIF report have this value.

> Or are you just creating new Test_Types for the tool name found in a SARIF report? Would that result in conflict between for example semgrep import and semgrep import via SARIF?

This is possible that the data in the SARIF report is exactly the same as the current `scan_type` of one parser. So currently it could be taken into account by the de-duplication algorithm. Maybe it's the behavior that we want. I plan to implement the same feature for the generic parsers (CSV and JSON). This way we can have custom de-duplication algorithm instead of having only one de-duplication algorithm for every SARIF report uploaded. Knowing that it will be different tools that produce SARIF and that the data of these tools are different we can have different de-dup algorithm even if it's from SARIF format.
But it's up to us. I mean we can decide that it's what we want.
The opposite (having every SARIF data using the same de-dup algorithm) will lead to different challenges.

StefanFl:
Both the current implementation and this PR don't work properly for re-imports and produce weird results. That's because it uses the default legacy deduplication algorithm. If I set the deduplication to `DEDUPE_ALGO_HASH_CODE` for SARIF the current implementation works well. But this approach doesn't work with this PR, because I would have to change the settings for each new SARIF tool.

damiencarol:
> Both the current implementation and this PR don't work properly for re-imports and produce weird results. That's because it uses the default legacy de-duplication algorithm. If I set the de-duplication to DEDUPE_ALGO_HASH_CODE for SARIF the current implementation works well. But this approach doesn't work with this PR, because I would have to change the settings for each new SARIF tool.

This is how the current implementation works. You need to create configuration for de-duplication to works, this PR doesn't change this fact. This approach will work with this PR. Because instead of having `SARIF` all the time, you will have a custom label coming from the report.
For ex with real data:
```bash
$ jq ".runs[]|.tool.driver.name" dojo/unittests/scans/sarif/* | sort | uniq -c
      1 "dependency-check"
      1 "Dockle"
      1 "Flawfinder"
      1 "Gitleaks"
      4 "LGTM.com"
      1 "mobsfscan"
      2 "nodejsscan"
      1 "OtherScanner"
      1 "Python Security Analysis"
      1 "semgrep"
      2 "Shell Script Analysis"
      1 "SpotBugs"
  ```

 So it will be easier for users to customize the de-duplication algorithm by using this label. I really think that users will not use too many tools that produce SARIF, so not much configuration here.

> Either we should make DEDUPE_ALGO_HASH_CODE the default de-duplication algorithm or we need a different approach for the SARIF parser, eg. like setting the name of the test instead of changing the test_type.

100% agree with you. We should work on having some kind of default algorithm to cover when the configuration is not here.
Something like this:
1. Check if we have a de-duplication algorithm for the `Test_Type`
2. If not, as it's a SARIF liked format, check if we have a de-duplication algorithm for the `Test_Type`
3. If not check if we have a default de-duplication algorithm

PS: also @valentijnscholten pointed something interesting. If the label in the report is the same as an existing scan_type used by a parser it will have collision. Maybe we could add a suffix like ` (SARIF)` to the Test_Type generated from SARIF.

**Comment by StefanFl** [comment]
I thought about the same as @valentijnscholten, the generated Test_Type should have the ` (SARIF)` suffix, to be transparent how this test was imported and to avoid collisions We should definitely do this.

If you implement your idea how to find the deduplication algorithm and set the deduplication for SARIF to `DEDUPLICATION_ALGORITHM_PER_PARSER` we don't have to change the global default. Then we can separate the discussion about the global default from this PR.

**Comment by damiencarol** [comment]
Ok. LGTM. Let's do that

**Comment by damiencarol** [comment]
Ok. LGTM. Let's do that.

**Comment by damiencarol** [comment]
> I thought about the same as @valentijnscholten, the generated Test_Type should have the (SARIF) suffix, to be transparent how this test was imported and to avoid collisions We should definitely do this.

@valentijnscholten just pointed the fact that we could have collisions. Adding a suffix was an easy proposal from me to that but I think it's a bad idea in the end. Because I want to use this feature for all parsers that need to customize the test not only SARIF. So in the end a lot of `Test_type` will have suffix. It could be confusing.

What I can do is customizing the de-duplication algorithm to take into account if the Test was generated with a generic format like SARIF. Like for ex searching in `DEDUPLICATION_ALGORITHM_PER_PARSER` with a 2-part key.
For ex, for `SARIF` and `Dockle` we could have :
```Python
DEDUPLICATION_ALGORITHM_PER_PARSER["SARIF|Dockle]" = DEDUPE_ALGO_HASH_CODE
```
this way we don't mix the findings. And if `SARIF|Dockle` doesn't exists, we use `SARIF`. 

@StefanFl what do you think?

StefanFl:
I would like to be able to see in the user interface, that the findings where imported via the SARIF format. What would be other use cases you see for that behaviour?

Maffooch:
I like this. Here's a list of where my opinions fit with others
- Naming
  -  Agree with @StefanFl that appending the format of the parser used is a good approach. This is crucial in discerning between formats that include lots of details vs those that do not (ex. Contrast has an XML format that is way better than the CSV format)
- Deduplication
  - I was under the impression there was already a standard dedupe. Having separate dedupe configs seems like a good move to eliminate any unanticipated mixtures.

damiencarol:
Fixed the label to include the original format. It should be better now and more clear.
I add a security that keep things simple if there are the same.

image

StefanFl:
> Adding a suffix was an easy proposal from me to that but I think it's a bad idea in the end. Because I want to use this feature for all parsers that need to customize the test not only SARIF. So in the end a lot of Test_type will have suffix. It could be confusing.

We will end up having additionnal Test_Types anyway, at least for SARIF. The existing Test_Types have the suffic ** Scan** most of the time. When I import Dockle or KICS scans via SARIF, I have these Test_Types:
- Dockle
- Dockle Scan
- KICS
- KICS Scan

If we would add the suffic ** (SARIF)** to the end of the new Test_Types we would know where the additional ones are coming from.

I still don't understand for which other scan types we would use this feature.

StefanFl:
Regarding deduplication: Can you add `'SARIF': DEDUPE_ALGO_HASH_CODE,` to `DEDUPLICATION_ALGORITHM_PER_PARSER` please? Otherwise re-imports for SARIF produce a mess and are pretty unusable.

damiencarol:
> Regarding deduplication: Can you add 'SARIF': DEDUPE_ALGO_HASH_CODE, to DEDUPLICATION_ALGORITHM_PER_PARSER please? Otherwise re-imports for SARIF produce a mess and are pretty unusable.

Done. I agree it's a good thing to have a default configuration.

damiencarol:
> If we would add the suffix ** (SARIF)** to the end of the new Test_Types we would know where the additional ones are coming from.

Change to `<data scan_type> Scan (<root scan_type>)`. This give us `KICS Scan (SARIF)`

> I still don't understand for which other scan types we would use this feature.

Currently (I don't have the whole list in mind):
 * Generic parser, SCB team want to use this one to customize the test type
 * AWSScout2Parser, customize description
 * MobSFParser, customize description
 * ScoutSuiteParser, customize description

dojo/importers/importer/importer.py
```diff
@@ -19,18 +19,20 @@
...
+from dojo.tools.factory import get_parser
+
 logger = logging.getLogger(__name__)
 deduplicationLogger = logging.getLogger("dojo.specific-loggers.deduplication")
 
 
 class DojoDefaultImporter(object):
 
-    def create_test(self, scan_type, engagement, lead, environment, tags=None,
+    def create_test(self, scan_type, test_type_str, engagement, lead, environment, tags=None,
```
StefanFl, line 30: `test_type_name` would be a better name of that parameter

dojo/importers/importer/importer.py
```diff
@@ -292,13 +295,48 @@ def import_scan(self, scan, scan_type, engagement, lead, environment, active, ve
...
+            logger.debug('IMPORT_SCAN parser v2: Create Test and parse findings')
+            tests = parser.get_tests(scan_type, scan)
+            # for now we only consider the first test in the list and artificially aggregate all findings of all tests
+            # this is the same as the old behavior as current import/reimporter implementation doesn't handle the case
+            # when there is more than 1 test
+            #
+            # we also aggregate the label of the Test_type to show the user the original scan_type
+            # only if they are different. This is to support meta format like SARIF
+            # so a report that have the label 'CodeScanner' will be changed to 'SARIF > CodeScanner'
+            test_type_str = scan_type
```
StefanFl, line 313: `test_type_name` would be more descriptive here as well.

dojo/importers/utils.py
```diff
@@ -1,30 +1,10 @@
...
 
 logger = logging.getLogger(__name__)
 
 
-def parse_findings(scan, test, active, verified, scan_type):
-    try:
-        parser = get_parser(scan_type)
-        parsed_findings = parser.get_findings(scan, test)
-        return parsed_findings
-    except SyntaxError as se:
```
StefanFl, line 14: This exception handling seems to be gone completely now, do we really not need it? Errors in parsers happen often, eg. when they change their format. Where will errors be logged now?

StefanFl (changes requested):

dojo/importers/importer/importer.py
damiencarol: Changed.

dojo/importers/importer/importer.py
damiencarol: Changed.

dojo/importers/utils.py
damiencarol: The exception is still correctly handled.
Ex:
image

The log are still there:
```
uwsgi_1         | [25/Sep/2021 16:33:51] DEBUG [dojo.importers.importer.importer:304] IMPORT_SCAN parser v2: Create Test and parse findings
uwsgi_1         | [25/Sep/2021 16:33:51] ERROR [dojo.engagement.views:614] name 'b' is not defined
uwsgi_1         |   File "/app/./dojo/engagement/views.py", line 599, in import_scan_results
uwsgi_1         |     test, finding_count, closed_finding_count = importer.import_scan(scan, scan_type, engagement, user, environment, active=active, verified=verified, tags=tags,
uwsgi_1         |   File "/app/./dojo/importers/importer/importer.py", line 305, in import_scan
uwsgi_1         |     tests = parser.get_tests(scan_type, scan)
uwsgi_1         |   File "/app/./dojo/tools/sarif/parser.py", line 41, in get_tests
uwsgi_1         |     a = b / 0
uwsgi_1         | NameError: name 'b' is not defined
uwsgi_1         | [25/Sep/2021 16:33:51] DEBUG [dojo.jira_link.helper:41] JIRA is disabled, not doing anything
uwsgi_1         | [pid: 149|app: 0|req: 17/21] 172.21.0.1 () {60 vars in 1378 bytes} [Sat Sep 25 16:33:50 2021] POST /product/10/import_scan_results => generated 71016 by
```
The previous version just logged and raised again the same exception
Not sure why it was required.

damiencarol: I'm testing the two version but honestly I prefer when we have less code.
The function without the try/catch was just 2 lines. Not very useful.

StefanFl: I checked, the views and the serializers are catching the exception and care about logging. So we don't need the code you deleted.

StefanFl (approved):

Maffooch (approved):

valentijnscholten (changes requested):
I haven't followed all discussions, but I think the description of the PR needs an update and we should add some info to the docs about how sarif scans (or generic etc) are handled. That will also help to approve it.

damiencarol:
@valentijnscholten done

StefanFl (approved):

valentijnscholten (approved):

Issue #5115, opened 2021-09-17
Outdated images in the DefectDojo Reports


chanduaki:
Found a outdated images in the defectdojo documents,it makes confusing to the people

**OLD Version Report**
![image](https://user-images.githubusercontent.com/54622987/133817834-352f57e8-8b3f-49e3-9138-723f5ffc4988.png)
**New Version Reporting Format**
![image](https://user-images.githubusercontent.com/54622987/133817883-af562104-7f53-45a6-8c01-02356eac80f8.png)

PR #5122, opened 2021-09-19
Documentation update for settings and reports

### LINKED ISSUE #5033: OAuth2/SAML2 Documentation tells you to hardcode credentials into settings.dist.py instead of recommending Environmental Variables
[OAuth2/SAML2 Documentation](https://defectdojo.github.io/django-DefectDojo/integrations/social-authentication/)

`dojo/settings/settings.dist.py` is where Environmental variables and their defaults are stored. If the environmental variables is not set in the runtime environment of the application, then the default is used.

**Problem**
Hard-coding secret values can lead to embedded secrets if they build a docker image with them. Recommending users to `edit the dojo/settings/settings.dist.py file and edit/replace the following information` is poor security practice and can lead to users inadvertently exposing their secrets to others if they upload their docker image to a non-private repository.

**Reproduction**
1. Setup OAuth2/SAML2 per the [documentation](https://defectdojo.github.io/django-DefectDojo/integrations/social-authentication/)
1. Docker compose / build
1. Docker push to docker.io
1. Secrets are now accessible by all on docker.io

**Fix**
These values should be set within the runtime environment and using environmental variables. Instead of recommending to edit the defaults to a hard-coded secret, it should recommend to use environmental variables.

[kubernetes](https://kubernetes.io/docs/concepts/configuration/secret/), [AWS](https://docs.aws.amazon.com/AmazonECS/latest/userguide/specifying-sensitive-data-secrets.html), [GCP](https://cloud.google.com/run/docs/configuring/secrets), [Azure](https://docs.microsoft.com/en-us/azure/container-instances/container-instances-environment-variables#secure-values), etc all allow you to inject secrets as environmental variables. This should be the default and recommend way of setting these secret values.

If just using docker, then recommend storing them in a safe place, like a secrets manager or password manager: 
Vault:
`docker run -e  DD_SOCIAL_AUTH_...="$(vault read ...)"`
LastPass:
`docker run -e  DD_SOCIAL_AUTH_...="$(lpass show ...)"`

StefanFl:
Fixes #5033 and #5115


docs/content/en/integrations/social-authentication.md
```diff
@@ -52,47 +52,45 @@ to be created. Closely follow the steps below to guarantee success.
     on the left side of the screen. Click **Create Credentials**, and
     choose **OAuth Client ID**:
 
-![image](../../images/google_1.png)
+    ![image](../../images/google_1.png<>)
```
damiencarol, line 55:
```suggestion
    ![image](../../images/google_1.png)
```
?

damiencarol (approved):

valentijnscholten (approved):

PR #5127, opened 2021-09-21
SQ: Fix broken migration 0120

kiblik:


Maffooch:
Django usually disallows editing past migrations due to the probable event of corrupting/deleting relevant data. Here are some other options you could take
- Create a new migration that address the issue with migration 0120
- Create a management script that could be run after installation to fix the db

kiblik:
@Maffooch, I do not understand why it didn't fail when we edited `0120` in version 2.2.1: https://github.com/DefectDojo/django-DefectDojo/pull/5036. Maybe because we didn't edit the content of `operations` but we changed the content of function `sq_clean`.

New migration will not fix the issue because users who are using version 2.2.0 (and do not use SonarQube) are already stacked with unfinished `0120`. I performed this test. Users who are migrating from the earlier version will not have this problem.

Management sounds usable but it will require also rerun migration (to finish 0120-0124).

Or we can edit `sq_clean` again. We will add there
```python
    try:
        migrations.RemoveField(
            model_name='test',
            name='sonarqube_config')        
    ...
```
There are 3 options:
- user finished migration before so there will be no effect for him
- user never tried this migration, action will fail but it is in `try`, so it will pass
- user tried this migration (he is running 2.2.0) and his migration is failing because `sonarqube_config` already exists. So removing this field (before its re-adding) will solve his problem

valentijnscholten:
You can modify existing migrations, Django won't complain. You just have to know what you are doing. In this case the 0120 migration is broken so we need to fix it if we can. Otherwise we need to provide SQL queries or a management command to repair those broken instances.
Did you test the option with `try:`, not sure if that will work?

kiblik:
> Did you test the option with `try:`, not sure if that will work?

I'm just trying to do it. I will inform about the results.

valentijnscholten:
Great find this low level schema editor, I think it could work.

One thing left is to add a new migration (0125 at the moment) to clear any empty row that might have been created for `Sonarqube_Product` table. I have lots of entries, all with NULL values for `sonarqube_project_key` and `sonarqube_tool_config`.

So something like
```
Sonarqube_Product.objects.filter(sonarqube_project_key__is_null=True, sonarqube_tool_config__is_null=True).delete()
```

kiblik:
> Great find this low level schema editor, I think it could work.
> 
Thank you.
Well, it could, but I will do not understand why it is still failing.

> One thing left is to add a new migration (0125 at the moment) to clear any empty row that might have been created for `Sonarqube_Product` table. I have lots of entries, all with NULL values for `sonarqube_project_key` and `sonarqube_tool_config`.
> 
> So something like
> 
> ```
> Sonarqube_Product.objects.filter(sonarqube_project_key__is_null=True, sonarqube_tool_config__is_null=True).delete()
> ```

Fact 1: Using empty `sonarqube_project_key` was and still is the valid scenario (check https://github.com/DefectDojo/django-DefectDojo/pull/4676, case no. 4 and no. 7). Implementation:
https://github.com/DefectDojo/django-DefectDojo/blob/57f64dba864b36508197ce7a6638260675be7fb7/dojo/tools/sonarqube_api/importer.py#L83-L86
So there is no reason to remove any `Sonarqube_Product` entry with `sonarqube_project_key__is_null=True`.

Fact 2:
- If `0120` passed on the instance successfully before, there can not be any `Sonarqube_Product` entry with `sonarqube_tool_config__is_null=True` because setting `null=False, blank=False` was there from the beginning of #4676 (compare [this](https://github.com/DefectDojo/django-DefectDojo/blob/57f64dba864b36508197ce7a6638260675be7fb7/dojo/db_migrations/0018_sonarqube_api_integration.py#L56) and [this](https://github.com/DefectDojo/django-DefectDojo/pull/4676/files#diff-87574859f016de36cc117fa758708d5a7b9dcbbda6c14a6ee7261aa736107edcR36))


- If `0120` didn't pass on the instance successfully before, the instance is stacked by waiting for the passable implementation of `0120`.

As I know, there are only 2 options to have some entries with `NULL` in  `sonarqube_tool_config` and passed `0120`:
- user used `migration` with `--fake`
- user changed code (for example by commenting out some parts)

So I do not see the reason for this kind of migration script.

However, if we want to be 100% sure, adding 
```python
Sonarqube_Product.objects.filter(sonarqube_tool_config__is_null=True).delete()
```
will not break anything. After the previous bug, all kind of "double check" is good.

valentijnscholten:
I thought it was working as the unit tests are green, but looks like the initializer fails or something around that.

valentijnscholten:
So we decided to go with a management command to help people fix their instance if they are stuck at migration `0120`. I think the number of users is small. The new `0125` migration is an additional cleanup needed for everyone just in case there is some code not handling `None` values very well for these fields.

Maffooch (approved):
Much better approach

kiblik:
Give me 1 more hour, I would like to perform one more test

kiblik:
I have some issue with the computer. I will run the test in morning

Maffooch:
FYI a new release is coming tomorrow between 12:00pm - 1:00pm central time.

kiblik:
It was good I performed one more test - I found bug which I fixed right now

Issue #4900, opened 2021-08-02
Add <component name> to the <column visibility> options


DeshengH:
Hi, We are using DefectDojo to address the vulnerabilities identified by a SCA (software composition analysis) scanner. However, since multiple vulnerabilities can be related to one component, we'd like to review the vulnerabilities on the component level.

We also tried to perform the review on the 'Components' section, but it seems to be complicated to deal with the components per project. 

Therefore, if the component name can be displayed within the finding list, we will have a clear view about which finding belongs to which component, like what we have on the original SCA scanner. 

Thanks

DeshengH:
FYI @damiencarol :)

valentijnscholten:
Did you consider using the grouping functionality that allows grouping of findings based on component_name? It's per test at the moment.
Regarding column visibility I think ideally we have a more dynamic/configurable way to display columns as there are already lots of columns.
What are you missing in the components view to help your usecase?

damiencarol:
@valentijnscholten I know thus use case. @DeshengH is a security researcher in my company. They use the finding view to export in csv. They want to have the component in the same report. Maybe it's not the best view to generate report.

StefanFl:
Putting more columns in the view doesn't seem to be the right solution. I would add an export button, which exports Excel, CSV and PDF (the current report implementation that's already there as a button). Will be finished at the weekend so we can have it in the next release.

damiencarol:
LGTM. Will share internally (my company will be the first user of these new columns)

StefanFl:
@damiencarol Does the new CSV export fix this issue?

PR #5148, opened 2021-09-24
Export findings to CSV and Excel

StefanFl:
The finding lists now provide an export functionality to download CSV and Excel files with the filtered findings:
2021-09-24 08_37_45-Findings _ DefectDojo

This is a solution for #4900

Sample reports:
- [findings.xlsx](https://github.com/DefectDojo/django-DefectDojo/files/7224349/findings.xlsx)
- [findings.csv](https://github.com/DefectDojo/django-DefectDojo/files/7224350/findings.csv)

This PR uses the implementation to select findings that was introduced with the quick reports. This implementation didn't respect the filters and reported too many findings. This has been fixed with one exception: Filtering for tags and finding_groups is not trivial and is not yet implemented.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -73,4 +73,3 @@ django-test-migrations==1.1.0
 djangosaml2==1.3.3
 drf-spectacular==0.19.0
 django-ratelimit==3.0.1
-XlsxWriter==3.0.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -73,3 +73,4 @@ django-test-migrations==1.1.0
 djangosaml2==1.3.3
 drf-spectacular==0.19.0
 django-ratelimit==3.0.1
+XlsxWriter==3.0.1
--- a/tests/base_test_class.py
+++ b/tests/base_test_class.py
@@ -79,6 +79,10 @@ class BaseTestCase(unittest.TestCase):
             desired = webdriver.DesiredCapabilities.CHROME
             desired['goog:loggingPrefs'] = {'browser': 'ALL'}
 
+            # set automatic downloads to test csv and excel export
+            prefs = {"download.default_directory": '/tmp'}
+            dd_driver_options.add_experimental_option("prefs", prefs)
+
             # change path of chromedriver according to which directory you have chromedriver.
             print('starting chromedriver with options: ', vars(dd_driver_options), desired)
             dd_driver = webdriver.Chrome('chromedriver', chrome_options=dd_driver_options, desired_capabilities=desired)
@@ -321,7 +325,7 @@ class BaseTestCase(unittest.TestCase):
 
             The addition of the trigger exception is due to the Report Builder tests.
             """
-            accepted_javascript_messages = r'(zoom\-in\.cur.*)404\ \(Not\ Found\)|Uncaught TypeError: Cannot read properties of null \(reading \'trigger\'\)'
+            accepted_javascript_messages = r'(zoom\-in\.cur.*)404\ \(Not\ Found\)|Uncaught TypeError: Cannot read properties of null \(reading \'trigger\'\)|Uncaught TypeError: Cannot read properties of null \(reading \'innerHTML\'\)'
 
             if (entry['level'] == 'SEVERE'):
                 # print(self.driver.current_url)  # TODO actually this seems to be the previous url
@@ -351,7 +355,7 @@ class BaseTestCase(unittest.TestCase):
         print('tearDownDriver: ', cls.__name__)
         global dd_driver
         if dd_driver:
-            if not dd_driver_options.experimental_options or not dd_driver_options.experimental_options['detach']:
+            if not dd_driver_options.experimental_options or not dd_driver_options.experimental_options.get('detach'):
                 print('closing browser')
                 dd_driver.quit()
 
--- a/tests/base_test_class.py
+++ b/tests/base_test_class.py
@@ -324,6 +324,7 @@ class BaseTestCase(unittest.TestCase):
             but http://localhost:8080/static/dojo/img/zoom-in.cur still produces a 404
 
             The addition of the trigger exception is due to the Report Builder tests.
+            The addition of the innerHTML exception is due to the test for quick reports in finding_test.py
             """
             accepted_javascript_messages = r'(zoom\-in\.cur.*)404\ \(Not\ Found\)|Uncaught TypeError: Cannot read properties of null \(reading \'trigger\'\)|Uncaught TypeError: Cannot read properties of null \(reading \'innerHTML\'\)'
 
--- a/tests/finding_test.py
+++ b/tests/finding_test.py
@@ -70,7 +70,7 @@ class FindingTest(BaseTestCase):
         driver.find_element_by_id("downloadMenu").click()
         driver.find_element_by_id("csv_export").click()
 
-        time.sleep(2)
+        time.sleep(4)
         self.assertTrue(Path("/tmp/findings.csv").is_file())
 
     def test_excel_export(self):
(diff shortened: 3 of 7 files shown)


requirements.txt
```diff
@@ -73,3 +73,4 @@ django-test-migrations==1.1.0
 djangosaml2==1.3.3
 drf-spectacular==0.19.0
 django-ratelimit==3.0.1
+XlsxWriter==3.0.1
```
damiencarol, line 76: We have already have a dependency on module `openpyxl`.

damiencarol (changes requested):

requirements.txt
StefanFl: Yes, we should have only one library. I will have a look which one to keep.

valentijnscholten:
Is there a reason for the custom url / query parsing (introduced earlier). I would say `django-filters` should do that for us?

valentijnscholten (changes requested):
Can you add some integration tests that trigger the exports?

StefanFl:
> Is there a reason for the custom url / query parsing (introduced earlier). I would say django-filters should do that for us?

I used it because it's there. I don't know how to achieve filtering this part of the query string with django-filter. Someone else would have to take over to change that, but maybe in a second step when this PR is merged.

devGregA:
Does the current CSV export not capture all the findings (bottom left of screenshot)?

StefanFl:
> Does the current CSV export not capture all the findings (bottom left of screenshot)?

These exports only export the findings that are currently on the screen with the columns in the list. If we wanted to have more attributes exported with that, we would have to put all the attributes in the HTML list. 

I don't know where these buttons come from.

devGregA:
Interesting, just played around and tested the old buttons. Stefan's version seems much better. Not asking Stefan too, but we should likely remove those in the future, as I'm not sure what the value even is of grabbing things from a single page.

requirements.txt
StefanFl: The Excel export is now using openpyxl and I removed XlsxWriter from the requirements.

StefanFl:
> Can you add some integration tests that trigger the exports?

Done

devGregA (approved):

valentijnscholten:
The pre-existing records come from the javascript datatables plugin. I guess someone added those thinking "something is better then nothing". It doesn't really hurt, but can be a bit confusing at first. The PDF button can be helpful if you have a small amount of findings and you want to do a quick report. It looks ugly though as it has no stylesheet.

valentijnscholten (dismissed):

StefanFl:
> Interesting, just played around and tested the old buttons. Stefan's version seems much better. Not asking Stefan too, but we should likely remove those in the future, as I'm not sure what the value even is of grabbing things from a single page.

This is actually really easy to do, I found the buttons now and removed them with my latest commit to this PR.

tests/finding_test.py
```diff
@@ -362,9 +394,13 @@ def add_finding_tests_to_suite(suite, jira=False, github=False, block_execution=
     suite.addTest(ProductTest('test_add_product_finding'))
     # TODO add some more findings with different statuses
     suite.addTest(FindingTest('test_list_findings_all'))
-    suite.addTest(FindingTest('test_list_findings_closed'))
-    suite.addTest(FindingTest('test_list_findings_accepted'))
-    suite.addTest(FindingTest('test_list_findings_open'))
+    # The next 3 tests don't work and need to be reworked
+    # suite.addTest(FindingTest('test_list_findings_closed'))
+    # suite.addTest(FindingTest('test_list_findings_accepted'))
+    # suite.addTest(FindingTest('test_list_findings_open'))
```
valentijnscholten, line 400: what's wrong with the tests, I can help maybe if needed to fix them.

tests/finding_test.py
```diff
@@ -51,6 +53,36 @@ def test_list_findings(self, suffix):
...
+
+    def test_excel_export(self):
+        driver = self.driver
+        driver.get(self.base_url + "finding")
+
+        driver.find_element_by_id("downloadMenu").click()
+        driver.find_element_by_id("excel_export").click()
+
+        time.sleep(2)
+        self.assertTrue(Path("/tmp/findings.xlsx").is_file())
```
valentijnscholten, line 84: I think this is a server side assert? And I guess you looked at doing a client side quick, but it's not possible or hard?

tests/base_test_class.py
```diff
@@ -321,7 +325,7 @@ def assertNoConsoleErrors(self):
 
             The addition of the trigger exception is due to the Report Builder tests.
             """
-            accepted_javascript_messages = r'(zoom\-in\.cur.*)404\ \(Not\ Found\)|Uncaught TypeError: Cannot read properties of null \(reading \'trigger\'\)'
+            accepted_javascript_messages = r'(zoom\-in\.cur.*)404\ \(Not\ Found\)|Uncaught TypeError: Cannot read properties of null \(reading \'trigger\'\)|Uncaught TypeError: Cannot read properties of null \(reading \'innerHTML\'\)'
```
valentijnscholten, line 328: should we add a comment to explain where this new "ignore exception" is coming from?

tests/finding_test.py
valentijnscholten: or maybe it is client side, just saw the prefs now above

valentijnscholten:
Great PR, but still some comments from me :-)

tests/finding_test.py
StefanFl: I didn't look what's wrong with them. They got never executed properly. Each of these tests calls `test_list_findings` with a parameter for the suffix of the URL like `finding/open` or `finding/closed`. But in `test_list_findings` this suffix is ignored and it executes the same test for all findings 4 times. Have a look at https://github.com/DefectDojo/django-DefectDojo/blob/master/tests/finding_test.py#L17-L32 and you will see.

It's completely unrelated with this PR, just something I noted when I saw it while writing more tests.

StefanFl: It should be fixed in a separate PR

valentijnscholten: ah oke, at first flance it looked like this PR broke. maybe put that comment in there and leave it for another PR indeed.

tests/finding_test.py
StefanFl: What do you mean with client or server side assert, is there a difference? The implementation comes from the Python `unittest` package and raises an exception when the expression is false:

```
    def assertTrue(self, expr, msg=None):
        """Check that the expression is true."""
        if not expr:
            msg = self._formatMessage(msg, "%s is not true" % safe_repr(expr))
            raise self.failureException(msg)
```

tests/base_test_class.py
StefanFl: Yes we should 😃, it's done with https://github.com/DefectDojo/django-DefectDojo/pull/5148/commits/01e854a9105e4d679911cc06aaa0cf2898108256

StefanFl:
Another idea, we could add a list of endpoints, separated by newlines for Excel and semicolon for CSV. Maybe not all of them but max 5 followed by `...` when there are more so this cell doesn't get too long. What do you think?

tests/finding_test.py
valentijnscholten: In some unit tests there is sometimes a mix of things being checked. In the api tests for example some asserts are on the data returned to the client (so "client-side") and some asserts are on server-side data like things in the database.
At first glance the `/tmp/findings.xlsx` assert looked like "server-side" to me. But I was wrong, so forget about it :-)

tests/finding_test.py
StefanFl: I wrote issue #5154 to address this.

damiencarol (approved):

StefanFl:
>>Is there a reason for the custom url / query parsing (introduced earlier). I would say django-filters should do that for us?

> I used it because it's there. I don't know how to achieve filtering this part of the query string with django-filter. Someone else would have to take over to change that, but maybe in a second step when this PR is merged.

see issue  #5155

PR #5156, opened 2021-09-25
Make integration test for exports more resilient

StefanFl:
The integration test for #5148 after it was merged into dev. Problem seems to be the waiting time to download the exported files, which was 4 seconds. This PR makes the test wait for a maximum of 10 seconds, but checks every second to not delay the following tests more than necessary.

### DIFF
--- a/tests/finding_test.py
+++ b/tests/finding_test.py
@@ -70,8 +70,13 @@ class FindingTest(BaseTestCase):
         driver.find_element_by_id("downloadMenu").click()
         driver.find_element_by_id("csv_export").click()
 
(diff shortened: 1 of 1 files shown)


valentijnscholten (approved):

damiencarol:
@StefanFl do we have other tests that do this kind of test? f yes, we should refactor it as a function and share it.

damiencarol (approved):

StefanFl:
> Do we have other tests that do this kind of test? f yes, we should refactor it as a function and share it.

Not that I would know

PR #5158, opened 2021-09-27
Fix broken documentation links in installation.md

savek-cc:
Fix broken documentation links

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -27,23 +27,22 @@ Steps to reproduce the behavior:
 A clear and concise description of what you expected to happen.
 
 **Deployment method** *(select with an `X`)*
-- [ ] Docker
+- [ ] Docker Compose
 - [ ] Kubernetes
 - [ ] GoDojo
-- [ ] setup.bash / legacy-setup.bash
 
 **Environment information**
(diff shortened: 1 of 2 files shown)


valentijnscholten (approved):

damiencarol (approved):

devGregA:
I think this is broken because dev has not made it into master. These links exist in dev.

devGregA:
Oh I see, but it breaks it in prod.... Not really sure what to do about that as it will create merge conflicts. I guess we could submit doc changes to dev and prod at the same time?

devGregA (changes requested):
looking into how best to handle due to the above.

devGregA:
I believe the solution is to switch the actions workflow to master instead of dev https://github.com/DefectDojo/django-DefectDojo/blob/master/.github/workflows/gh-pages.yml#L6. Thinking over best way to get fixed immediately.

valentijnscholten:
Earlier we decided to build the docs from dev so we can see the latest and spot any errors on github pages. So if we change these "broken" links to point to dev, we will be fine I think.

valentijnscholten (changes requested):
@savek-cc the docs are built from dev so links should best be to the dev branch. Can you update the PR?

savek-cc:
768px-Acoustic_modem

PR #5170, opened 2021-09-28
Use django-filter for quick reports and CSV and Excel reports

### LINKED ISSUE #5155: Investigate how django-filter can be used for quick reports and CSV/Excel exports

StefanFl:
fixes #5155

With this PR quick reports and the exports work with all filter setting.

This PR includes some refactoring to move the filtering of findings into its own method and removal of 2 superfluous filters.


Maffooch (approved):
This is very clean. Much better than using the datatables method of getting CSVs.

alles-klar (approved):

PR #5182, opened 2021-09-29
Yet another parser: Azure Security Center

StefanFl:
Microsoft's Azure Security Center gathers recommendations about the configuration of a subscription and lets user export its findings in CSV format.


dojo/settings/settings.dist.py
```diff
@@ -1024,7 +1024,8 @@ def saml2_attrib_map_format(dict):
     'Scout Suite Scan': ['title', 'severity', 'description'],
     'AWS Security Hub Scan': ['unique_id_from_tool'],
     'Meterian Scan': ['cwe', 'component_name', 'component_version', 'description', 'severity'],
-    'Github Vulnerability Scan': ['unique_id_from_tool']
+    'Github Vulnerability Scan': ['unique_id_from_tool'],
+    'Azure Security Center Scan': ['title', 'description', 'severity'],
```
valentijnscholten, line 1028: why is severity needed?

valentijnscholten (approved):

dojo/tools/azure_security_center/parser.py
```diff
@@ -0,0 +1,91 @@
...
+
+                findings.append(Finding(
+                    title=title,
+                    cwe=1032,  # Security Configuration Weaknesses
+                    test=test,
+                    description=description,
+                    severity=severity,
+                    references=azure_portal_recommendation_link,
+                    mitigation=remediation_steps,
+                    date=datetime.strptime(status_change_date[0:10], '%Y-%m-%d').date(),
```
damiencarol, line 84: The `date` is the first date saw by the tool. So I guess a better candidate for this one is `firstEvaluationDate` attribute of the report. It could lead to some bug with de-duplication but As we are configuring the de-duplication algo to use uniq id from tool is ok.

damiencarol (approved):

dojo/tools/azure_security_center/parser.py
StefanFl: `firstEvaluationDate` is the date when a check for a resource has run the first time. A lot of checks will have the healthy state, so it doesn't give information at which date the check failed for the first time. This is why I think the `status_change_date` might be better, as we import only unhealthy checks.

damiencarol:
@StefanFl global comment: you implemented the "Recommendations" report part. But most of the users will try the "Security Alert" first I think. (I've done that) It could be confusing. Any reason you implemented the recommendation format and not the other one?

StefanFl:
@damiencarol I have just started using the Security Center and saw only the recommendations. They come out of the box with the Security Center while for the alerts you seem to have the Defenders active. 

But now I see this new scanner is named wrongly, because it needs to have the Recommendations in the name. Just created some sample alerts and they have a different format.

PR #5187, opened 2021-09-29
Optimize view finding

StefanFl:
For some scans the view finding page can currently look like this:

2021-09-29 16_34_46-View Finding _ DefectDojo

This finding of Dockle for example has no file or component and is not connected to Jira or GitHub. The second table shows there is 1 occurence of this finding in the scan and the finding is not in a group. Both information are not really important. With this PR the second table is only shown when any of the fields has a content, including nb_occurences is greater than 1.

This makes the user more focused on the important information of the finding.

### DIFF
--- a/dojo/templates/dojo/view_finding.html
+++ b/dojo/templates/dojo/view_finding.html
@@ -451,7 +451,7 @@
         </div>
         {% endif %}
     {% endif %}
-    {% if finding.static_finding or finding.line > 0 or finding.has_jira_configured or finding.has_jira_issue or finding.github_issue or finding.github_conf_new %}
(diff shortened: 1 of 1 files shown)


damiencarol:
@StefanFl `nb_occurences` is not mandatory. Could this lead to a bug?

StefanFl:
> `nb_occurences` is not mandatory. Could this lead to a bug?

The template engine seems to be lenient here. I tested it, it does work with `null` values and doesn't even log a warning.

valentijnscholten (approved):
yeah, templates render fine even with printed variables being None.

Maffooch (approved):

PR #5194, opened 2021-09-30
SonarQube API Import: set dedup algo

kiblik:
DD was closing and creating the same findings because the default dedup algorithm doesn't check `file_path` and `line`


damiencarol:
@kiblik thanks for this PR, we are facing this issue at my day job.

damiencarol (approved):

kiblik:
I'm going to add also unittest for changing of `hash_code`

Btw, it is getting answer for [this question](https://github.com/DefectDojo/django-DefectDojo/blob/c548d60ccabdb9146c78490e05c11cb4205cd9df/dojo/unittests/test_deduplication_logic.py#L12)

kiblik:
However, I'm still thinking about the performance impact.

kiblik:
I'm going to split it because `hash_code`s will be more complex a questionable topic

StefanFl (approved):

StefanFl:
@damiencarol Is your approval still valid?

valentijnscholten:
recalculating hash_code will open pandora's box regarding deduplication, so we can't approve that currently.

kiblik:
When I saw the results of unittest I realised it will open Pandora's box. Because of this, I removed the `hash_code` recalculation and I keep here only setting of dedup algo for SonarQube.

damiencarol:
@kiblik  so could change the description also ?

damiencarol:
@valentijnscholten I'm merging this as @kiblik removed the hashcode part.

PR #4688, opened 2021-06-20
[Bandit parser] Fix internal de-duplication

damiencarol:
# Work done

Tested demo instance and loaded big Bandit reports (more than 3000 findings by report) and found few pbs.

## What was implemented

- [x] Fix internal de-duplication to highly reduce the number of findings (group by `vuln_id_from_tool`)
- [x] improve description when grouping finding
- [x] implement `references`
- [x] add more test data and unit tests (a big report)

bandit_dup

PR #5206, opened 2021-10-03
Bandit parser: add de-duplication algorithm

damiencarol:
# Work
- [x] global maintenance (add more test data from recent version 2021-10)
- [x] add de-duplication algorithm `DEDUPE_ALGO_HASH_CODE`
- [x] implement `references` attribute
- [x] supersede https://github.com/DefectDojo/django-DefectDojo/pull/4688

image


dojo/tools/bandit/parser.py
```diff
@@ -19,29 +17,31 @@ def get_description_for_scan_types(self, scan_type):
     def get_findings(self, filename, test):
         data = json.load(filename)
 
-        dupes = dict()
+        res = list()
```
StefanFl, line 20: Can we have a more descriptive variable name here, eg. `results`?

StefanFl (changes requested):

valentijnscholten (approved):
I'm good assuming you'll take Stefan's feedback into account :-)

dojo/tools/bandit/parser.py
damiencarol: Fixed.

StefanFl (approved):

PR #5207, opened 2021-10-04
Fix BulkEdit severity dropdown after #4766

Maffooch:
After #4766, all select fields were dynamically converted to select picker fields. It did not work so well for those that were contained within a drop down. 

Before:
https://user-images.githubusercontent.com/46459665/135797454-84f9fcdc-3046-4e63-8cc0-1b095b790490.mov

After:
https://user-images.githubusercontent.com/46459665/135798169-3f636d72-c21b-4f0a-9fa3-b91e4abf0dae.mov

### DIFF
--- a/dojo/templates/dojo/findings_list_snippet.html
+++ b/dojo/templates/dojo/findings_list_snippet.html
@@ -71,7 +71,7 @@
                     </button>
                     {% endif %}
                 </div>
-                <ul class="dropdown-menu" aria-labelledby="dropdownMenu1" id="bulk_edit">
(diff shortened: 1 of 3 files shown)


valentijnscholten:
Nice, just in time. Back then I "knew" there was something with the PR, but couldn't scientifically prove it :-)

valentijnscholten:
@Maffooch `view_test.thml` needs the same changes, that page still doesn't use the same codebase for listing findings.

valentijnscholten (commented):
.

valentijnscholten (changes requested):
.

valentijnscholten (approved):

damiencarol (approved):

PR #5213, opened 2021-10-04
[HELM] Mount media to permanent storage

dsever:
This PR is based on https://github.com/DefectDojo/django-DefectDojo/pull/4823 from @TheocharisPetros (kudos), and includes:

- mounting of media to nginx and uwsgi container
- kubernetes explanation
- default values

Does not include:
- volume creation from storage class, is uses already existing PVC, just to allow more freedom for different types of k8s clusters

Together with this, I will add example in Community Contribs, one example how minIO can be utilized for enable r/w many.

### DIFF
--- a/helm/defectdojo/values.yaml
+++ b/helm/defectdojo/values.yaml
@@ -234,7 +234,7 @@ django:
 
   extraVolumes: []
 
-  # This feature needs more preparation before can be enabled, please visit Kubernetes.md#Persistent volumes
+  # This feature needs more preparation before can be enabled, please visit KUBERNETES.md#media-persistent-volume
   mediaPersistentVolume:
     enabled: true
     name: media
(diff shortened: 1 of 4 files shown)


helm/defectdojo/values.yaml
```diff
@@ -233,6 +233,14 @@ django:
   # @type: array<map>
   extraVolumes: []
 
+  # This feature needs more preparation before can be enabled, please visit KUBERNETES.md#media-persistent-volume
+  mediaPersistentVolume:
+    enabled: true
+    name: media
+    type: emptyDir
+    # could be emptyDir (not for production) or pvc
```
alles-klar, line 241: if the comment is related to the line type: emptyDir, please place the comment above.

helm/defectdojo/values.yaml
```diff
@@ -233,6 +233,14 @@ django:
...
+    persistentVolumeClaim: media
+    # in case if pvc specified
```
alles-klar, line 243: same here

dsever:
Regarding the test, I tested it manually. Of I will have enough

> Thanks @dsever!
> 
> I think our integration tests doesn't cover it at the moment because we do not run them in the kubernetes tests. Did you tested it with the default values?
> 
I tested it manually,  if I will have enough time, I would  like to take a look at `helm test` and see are there missing parts.

PR #5221, opened 2021-10-05
APIv2: allow create/list/view/delete User Contact Info

kiblik:


StefanFl:
@kiblik Every endpoint shall have a test in https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/unittests/test_rest_framework.py. So we need one for the `UserContactInfo` as well.

PR #5224, opened 2021-10-05
APIv2: Allow to set the first password when the user is created through API

kiblik:
Allow to set the first password when the user is created through API


valentijnscholten (changes requested):
I think this will expose all users passwords via the API, can you make it a write_only field and add a testcase that verifies the password (or its hash) is not exposed?

kiblik:
> I think this will expose all users passwords via the API, can you make it a write_only field and add a testcase that verifies the password (or its hash) is not exposed?

it was "write_only" from the beginning so it works correctly. I'm happy to add more tests.

valentijnscholten:
> it was "write_only" from the beginning so it works correctly. I'm happy to add more tests.

Oh, missed that. The test I was looking for is to make sure the password is not returned when retrieving users, thanks for that.

valentijnscholten (approved):

StefanFl (approved):

PR #5227, opened 2021-10-06
Add support for context region snippets in SARIF parser

Kjeld-P:
Some tools that output SARIF files put the only available code snippet in the **contextRegion** propery instead of the region propery. An example is CodeQL with the `--sarif-add-file-contents` flag set.
This change will attempt to extract a snippet in the contextRegion propery if none was found in the region propery.

As per the [specification](http://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/sarif-v2.0-csprd01.html) section 3.21.5, a contextRegion will only exist if the region exists, hence the order of checking.

### DIFF
--- a/dojo/tools/sarif/parser.py
+++ b/dojo/tools/sarif/parser.py
@@ -185,6 +185,10 @@ def get_snippet(result):
                 if 'snippet' in location['physicalLocation']['region']:
                     if 'text' in location['physicalLocation']['region']['snippet']:
                         snippet = location['physicalLocation']['region']['snippet']['text']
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Thanks, can you add an example sarif report. Or add a contextRegion in an existing report?

valentijnscholten:
Thanks, can you update the unit test to make sure the codepath you added is tested?

valentijnscholten:
No worries, happy with your contribution! Can you review the small piece of doc I added?

PR #5234, opened 2021-10-06
Bandit parser: fix de-duplication settings

damiencarol:


valentijnscholten:
@damiencarol bandit is used in dedupe tests, so I guess those need to be updated

damiencarol:
@valentijnscholten done. Added in initializer, unit-tests and uwsgi.

PR #5241, opened 2021-10-07
User: allow superuser (no staff) to list and view users

kiblik:
Superuser (who is not the staff) was able to edit and delete users however he wasn't able to list and view them.

### DIFF
--- a/dojo/templates/base.html
+++ b/dojo/templates/base.html
@@ -325,7 +325,7 @@
                                 </ul>
                                 <!-- /.nav-second-level -->
                             </li>
-                            {% if request.user.is_staff %}
+                            {% if request.user.is_staff or request.user.is_superuser %}
                                 <li>
                                     <a href="{% url 'users' %}" aria-disabled="true" aria-expanded="false" aria-label="Users">
                                         <i class="fa fa-user fa-fw"></i>
--- a/dojo/templates/dojo/users.html
+++ b/dojo/templates/dojo/users.html
@@ -42,7 +42,7 @@
                     <table id="users"
                         class="tablesorter-bootstrap table table-condensed table-striped">
                         <tr>
-                            {% if request.user.is_staff %}
+                            {% if request.user.is_staff or request.user.is_superuser %}
                             <th></th>
                             {% endif %}
                             <th class="nowrap">{% dojo_sort request 'First Name' 'first_name' %}</th>
@@ -60,7 +60,7 @@
                         </tr>
                         {% for u in users %}
                             <tr>
-                                {% if request.user.is_staff %}
+                                {% if request.user.is_staff or request.user.is_superuser %}
                                 <td>
                                     <ul>
                                         <li class="dropdown" style="list-style:none;position:absolute">
--- a/dojo/user/views.py
+++ b/dojo/user/views.py
@@ -243,7 +243,7 @@ def change_password(request):
         'form': form})
 
 
-@user_passes_test(lambda u: u.is_staff)
+@user_passes_test(lambda u: u.is_staff or u.is_superuser)
 def user(request):
     users = Dojo_User.objects.all() \
         .select_related('usercontactinfo', 'global_role') \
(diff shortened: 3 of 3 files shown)


kiblik:
Btw, this can happen in multiple places in DD.

StefanFl:
@Maffooch It is slighty different, users who are not **Active** cannot login. That's the way to deactivate a users without deleting the user object.

I personally see **Staff** literally, people who are employees of the company and therefore have some more permissions than users who are not **Staff** (external users who have access to DD but are not employees). This reflects very good how DD works currently and is easy to explain. 

The new authorization is about permissions regarding products and product types. It is not designed to address the all other areas, eg. viewing/editing users or test types.

@kiblik I still don't understand your use case, why you need to have superusers that are not staff. DD works fine as long as superusers are staff as well. If we break that up, we should be consequent and check all occurrences of `user.is_staff` to see if we need to add an `or user.is_superuser`

StefanFl:
There is a new issue to make sure superusers are staff users as well: #5271

valentijnscholten:
This is from the offcial django docs:
image

So the staff flag controls whether a user can login to the admin site. It works like that in my instance and the demo instance:

image

If the user has no permissions, he/she can't do anything. But it does allow them login, which makes no sense. The "staff" thing has caused us nothing but confusion. And we're not the only ones if you google for "django staff admin superuser" permission for example: https://stackoverflow.com/questions/20469174/whats-the-difference-between-staff-admin-superuser-in-django

StefanFl:
I am with you @valentijnscholten, we should get rid of the staff users. The admin portal should not be relevant for any user. In our next meeting I want to discuss a different approach, where you can give users or groups the permissions to do selected administrative tasks. The good thing is, we can use the Django permission as the backend for it.

StefanFl:
Screenshot doesn't work for me as well

StefanFl:
@devGregA  To be honest, I don't get what your use case is, when users can see a page where they cannot do anything. But I think we need to understand the use cases, before we can decide how to design a new solutions. Thinking of roles like "security staff" help, because it makes it clearer what these people want to do with DefectDojo.

devGregA:
I will have to check if the pr made it in, but it was an env variable that blocked the actual URL. @Maffooch did that get submitted? @valentijnscholten  I'm confused on your statement. If a superuser is both staff and superuser they will have full access to django admin because they are superuser. Staff doesn't play a part if I understand you. You can go to django admin as staff, but have no permissions, and we don't use any of the functions in django admin (unless there is a PR I missed, which is very possible). We used to use it for the password reset form, which we've since implemented as part of the actual dojo UI.

valentijnscholten:
Looks like we crossposted. Although we spent limited amount of time to think about "security team" roles, I think the roles we have in place are a good start for those scenario's. For example the authorization migration script migrates `staff` users into being `global owner` which means they can manage all products and permissions, but not the configuration of Defect Dojo itself. Not sure about tools, credentials, etc. So maybe we need another permission or global role. But let's not call it staff as it confuses everyone as it's a Django internal permission.

devGregA:
@valentijnscholten Not all security people are developers. That is definitely the minority of our users, even though I would generalize and say they are more technically inclined. People get in to trouble with Dojo when we don't put in good guard rails, and choose to do any administration from django admin, which no one really should be for the intended use case of dojo.

Again, I'm not opposed to doing away with the staff status, using a different model, etc, but I do want to preserve the use case.

On intended usage, I'm not overly concerned, and I'm not opposed to your suggested solution, but specifically on intended usage, I believe our mistake was in lack of documentation. Django was created as content management system for news organizations that were 'perfectionists' and didn't want to use wordpress. The point I'm trying to get to is that specifically in regards to intended use of django, is that we are so far outside intended use, and have been since inception, that I don't think it makes sense to measure dojo against django. Has it caused confusion?  Absolutely, and we should address that. 

Regarding the django admin portal, a narrow view of staff was used because in django 1.0 managing permissions was a nightmare. So we only used the password reset form. I've provided screenshots, and tested that users with only 'staff' permission (not superuser) do not have access to django admin. They can login, like you mentioned, but they see nothing by default and have no access. Maybe you have non-standard usage around admin that gives staff extra permission? I do think we should block django admin as a default, because for the majority of users, it poses a risk in regards to bricking their instance. I understand that there maybe a usage that is beneficial, which is why we try so hard to not kill use cases, but I do think the default should be what favors the  majority of the community, which can be overwritten and still accessed by flipping the environment variable.

Issue #5236, opened 2021-10-07
Filters for Findings have no dropdowns


StefanFl:
**Bug description**
When you open the filters in any of the Finding lists, the dropdown fields (Date, Product, ...) have no dropdowns and cannot be selected.

Report by @37b on Slack: https://owasp.slack.com/archives/C2P5BA8MN/p1633566424351200

StefanFl:
might have been caused by #5170

PR #5245, opened 2021-10-07
add some basic dropdown tests

valentijnscholten:
to help prevent regressions like https://github.com/DefectDojo/django-DefectDojo/issues/5236 I added some basic tests to click on these dropdowns during the integration tests.
funny thing is that even selenium seems to think the dropdown is working normally, as it succesfully is able to select a visible "High" text, even it's not visible to the human eye :-)
So some funky stuff going on with these bootstrap-select and select2 dropdowns. Still I think good to have these extra clicks in the tests to at least make sure the elements are there and don't generate any javascript errors.


damiencarol:
@Maffooch I think yes. it's common when you work on integration tests, I'm doing the same ;) .

PR #5264, opened 2021-10-09
integration tests: run as matrix

valentijnscholten:
WIP
The integration tests take a long time to run i.e. ~26minutes.
All tests are setup as separate tests, so they can be run in parallel.
This PR tries out the matrix strategy of GitHub Actions to see if it works to have all tests run as separate jobs.

First tests show that the tests are now completed in ~13 minutes, which is a nice improvement.

It does mean it generate 21 GitHub Action jobs, each starting their own instance of Defect Dojo. Feels a little bit overkill :-)

Ideally we don't list all the test files separately in two places (or at all), but let's first see how this new approach turns out.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -140,7 +140,7 @@ jobs:
       - name: Integration tests
         run: docker-compose up --exit-code-from integration-tests integration-tests
         env:
-          DD_INTEGRATION_TEST_FILENAME: ${{ matrix.docker-image }}
+          DD_INTEGRATION_TEST_FILENAME: ${{ matrix.test-case }}
 
       - name: Logs
         if: always()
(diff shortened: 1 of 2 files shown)


valentijnscholten:
126? It's 21 I think. There is a limit of 60 parallel jobs as part of our "Teams" plan: https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

StefanFl:
Every integration test starts 6 containers (rabbitmq, mysql, nginx, uwsgi, celery_beat, celery_worker). So 21 integration tests times 6 containers makes 126 containers in total.

Apart from nginx each of the other 5 containers uses 100 - 300 MB memory while running the integration tests in my local environment. This is quite a strain for GitHub's resources.

damiencarol:
I think it's a great improvement.

devGregA:
I'm for it. We're not violating what GitHub allows, so if it makes things better I'm all for it.

PR #5265, opened 2021-10-10
Upgrade to Django 3.2

valentijnscholten:
Django 3.1 is already in extended support, getting only security fixes. Extended support will end December 1st. So we need to upgrade to Django 3.2, which is an LTS version. 

image

At first glance we shouldn't expect too many problems.


valentijnscholten:
At second glance Defect Dojo won't start due to a [yet to be fixed incompatiblity](https://github.com/radiac/django-tagulous/issues/144) in `django-tagulous` :-)

valentijnscholten:
Django 3.1 became End Of Life today. We are still waiting for a fix in `django-tagulous` to be released. There is a fix, but it's not released yet. This PR updates to Django 3.2 and installs `django-tagulous` from a specific git commit, with the fix. I need your opinion on if we want to merge this so we are on a supported Django version, or do we want to wait for `django-tagulous` to release their fix? The bug report has been open for almost 2 months now with no response from the maintainer, so could take some more weeks, taking it past our 2.5.0 release next week.

StefanFl (approved):
The requirement for a specific commit is ugly, but we should do it nevertheless to be able to move forward with Django.

Maffooch (approved):

mtesauro:
+1 on what Stefan said ^

Not the best option but better than not updating Django

PR #5268, opened 2021-10-12
ScoutSuite parser: refactor parser interface

damiencarol:
## Work done
- [x] enable the multi-test interface and customize Test name, description, type, etc... (it now more clear which kind of cloud was tested)
- [x] Deprecate the old parser for previous versions upstream (deactivate it by default in settings, will remove the parser in versions >= 3.x.x)
- [x] improve de-duplication algorithm to use `vuln_id_from_tool` attribute

For ex if ScoutSuite check AWS configuration:
image


damiencarol:
I need to improve more part like the de-duplication algorythm

PR #5275, opened 2021-10-13
Allow User Profiles to be read-only

dsever:
Admin can specify (DD_)USER_PROFILE_EDITABLE = False (default True)

This brings option, not to allow normal user to change their profile information.  
 
Some modification for dev entrypoint of uwsgi, because I'm using trick to observe dojo/setting/settings.py  and I'm touching it when reload is needed.


Maffooch (approved):
@damiencarol please see

valentijnscholten:
Considering the upcoming release and 3 approvals I am merging this one.

alles-klar:
@dsever Do you have an idea how the tests can work when the local file system from integration tests and uwsgi are not mounted? I think it is very bad to have the precondition that the filesystem from you host ist mounted, writable and shared between both containers.

Can we maybe move the tests with configuration changes to the unit tests?

dsever:
I don't see any other way, I did research, and found out that I can only use this approach to reconfigure application on the fly and test it.

This is part of integration tests so that's why I wanted to have it there. 

You have an issue when not using docker compose to do integration tests?

alles-klar:
I am using the integration tests to test my defectdojo instance with production configuration from "outside".

Can't we run these security tests in the unit tests only? We can check there if we have security problems. We don't test this settings also in the ui, but for me it is worth it, when the setup is less complex.

Lets look in the future: We are getting more and more tests, that test different configurations. Would it make sense so create a configuration file (template-env) and that using it also for the integration tests? The we could run two test in parallel, one with configuration A and one with configuration B? This does only make sense for features with two states.

dsever:
Makes sense to me, the problem I see here is actually all integration test are testing admin PoV, but not from user perspective with different configurations. But it is about moderators to decide.

valentijnscholten:
The `settings` file way of working is a big pain as the settings can't be changed at run-time. Now it's only 2 scenario's we want to test, but that could easily explode to n!  or 2^n or N^n scenario's. But if we switch to database based settings, users will lose the way to override settings via `local_settings.py` for example.

In this particular case we could run the test in its own testcase and launch the django containerwith the desired values for DD_USER_PROFILE_EDITABLE. But that can only work now that we have 1 GitHub Action job per testcase (which is currently blocking our GH actions when it is running as it start 20 or so jobs and we can only have 20 running in parallel).

PR #5283, opened 2021-10-15
test(integration): Install Chromedriver during dockerbuild

alles-klar:
Moves the installation of chromedriver into dockerbuild. This increases the speed of integration tests, because we do not have to download and install chromedriver on each run and can use the caching from docker. Furthermore it allows to run the test in environments without internet connection. 

Related to #4891

On behalf of DB Systel GmbH

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,3 @@
 .git
 .gitignore
 *.md
-.env*


StefanFl:
Great job, thank you @alles-klar

PR #5295, opened 2021-10-18
Moving CURL to build step

dsever:
It is not good practice to have `curl` utility in production.  This PR just moves fetching of the file in the build step.


valentijnscholten:
Well if we remove the comment, the two build containers don't have to be equal anymore ;-) As said on Slack, let's put this PR on hold for a bit.

Issue #5204, opened 2021-10-01
Add Forgot Password flow


valentijnscholten:
https://www.ordinarycoders.com/blog/article/django-password-reset

PR #5302, opened 2021-10-19
Add "Forgot password" functionality

kiblik:
Solution for https://github.com/DefectDojo/django-DefectDojo/issues/5204

I used original templates (original wording) but I modified style (CSS)

image

image

image

image

image

image

Or

image


Maffooch:
@kiblik I still need to test this out, but it looks great so far! Stylistic requests by Stefan and Valentijn are good and I request them as well

@StefanFl I use https://github.com/mailhog/MailHog for a quick docker mail server. Pretty easy to set up as well. I can assist if needed

Maffooch:
@kiblik we are planning to release tomorrow afternoon (12 CT) do you think this can be ready by then?

kiblik:
I really wanted, but I will not be able to finish it today.

kiblik:
Do we prefer to use different "notification" and "reset" email addresses? I'm thinking about renaming `mail_notifications_from` to something more general (like `mail_from`) and it can be used by both.

valentijnscholten:
Yeah, one from address should suffice

kiblik:
If an SMTP server is not available, the original implementation fails when the user click "Reset my password" (HTTP-500).
I can override this behaviour but what do we expect? What kind of error message?

kiblik:
New locations. Is it better?

image

kiblik:
> If an SMTP server is not available, the original implementation fails when the user click "Reset my password" (HTTP-500). I can override this behaviour but what do we expect? What kind of error message?

I wrote the error message to the log and returned standard page (We sent you email) even it wasn't send

valentijnscholten:
> New locations. Is it better?
> 
> image

Better. It's still not the best of login pages, but that's not really related to this PR. Thanks for improving it.

valentijnscholten:
> > If an SMTP server is not available, the original implementation fails when the user click "Reset my password" (HTTP-500). I can override this behaviour but what do we expect? What kind of error message?
> 
> I wrote the error message to the log and returned standard page (We sent you email) even it wasn't send

I think there should be some error displayed to the user. Maybe even a 500 error. I think we should have some 500 error page anyway stating the at people should contact their administrator. The admins have a way to disable the forgot password button, so they should make sure when enabling it, there is a email from configured. 

Other thought, just use some dummy email addres like `no-configured@defectdojo`?

Or add a django system on startup? Or just make the email address mandatory in system settings?

StefanFl:
@kiblik, looks great now, thanks for all your work here. Nevertheless one question: An email is only sent, when the user had a password before. When I add a new user without a password, the email for the password change is not sent. Is that intended respectively something we can influence?

kiblik:
> @kiblik, looks great now, thanks for all your work here. Nevertheless one question: An email is only sent, when the user had a password before. When I add a new user without a password, the email for the password change is not sent. Is that intended respectively something we can influence?

@StefanFl, to be honest, I haven't tested this use-case. But as I see, it is implemented as you mention - procedure check if the password was already set it up by `.has_usable_password()` check
https://github.com/django/django/blob/8a7ac78b706797a03d26b88eddb9d1067ed35b66/django/contrib/auth/forms.py#L277-L281
I can override this behaviour but it will automatically allow creating passwords and login directly (skipping IdP).

I can add EnvVar which will allow this behaviour but I prefer to keep it by default disabled.

StefanFl:
We shouldn't override this. Fine for me with this explanation.

StefanFl (approved):

PR #5307, opened 2021-10-20
main(tests): remove rabbitmq from unit-tests

alles-klar:
Using build in sqlite database for unit tests instead of a separate rabbitmq.


alles-klar:
@valentijnscholten and @damiencarol works now. What do you think of removing rabbitmq from the unit tests?

StefanFl (approved):

valentijnscholten:
I had some doubts, but at the same time I attempted to let GHA run the unit tests without starting rabbitmq in the past so that must mean I like the idea ;-)
We still have to look at running all the unit tests in the foreground to make sure all code is run/covered/tested.

valentijnscholten (approved):

PR #5336, opened 2021-10-25
Use endpoint status to determine vulnerable endpoints

StefanFl:
Until now the endpoint status hasn't been used to determine if an endpoint is vulnerable. With this PR an endpoint is marked as vulnerable, when it has active / verified findings and their endpoint status is not mitigated. Additionally this PR contains some tlc for the endpoint dialogues and a bugfix for editing metadata.

2021-10-25 19_59_52-All Endpoints _ DefectDojo

2021-10-25 19_58_43-Endpoint _ DefectDojo

2021-10-25 19_59_25-Host _ DefectDojo


StefanFl:
Nobody seems to like the locks, not even my wife 😄 The new version uses thumbs up and down:

2021-10-30 10_29_29-Host _ DefectDojo

@kiblik and @valentijnscholten What do you think?

valentijnscholten (approved):

valentijnscholten:
I agree with your wife.

alles-klar (approved):

PR #5357, opened 2021-10-29
Change Anchore Grype parser to allow matcher lists

gvwalker:
Anchore Grype JSON file layout changes throw an error when trying to import due to matchers now being an array instead of an object.

Modified the parser to support either a list or object.


valentijnscholten:
Turns out the unit tests are half useless currently as they are not checking for the results to be present. I have fixed that and added extra unit test. But cannot push to this PR for some reason. Will create a new one so it can be part of the release today.

valentijnscholten:
replaced by https://github.com/DefectDojo/django-DefectDojo/pull/5369

PR #5359, opened 2021-11-01
Remove safety parser and db

StefanFl:
At the request of pyup.io, we had to remove the parser for Safety scans.


Maffooch:
You beat me by 3 minutes 😂 I like yours better though. Will approve after tests pass.

Maffooch (approved):

PR #5361, opened 2021-11-01
order findings last 7 days by -date

valentijnscholten:
On the dashboard there's the number of findings created in the last 7 days. When clicking on the number the user is presented a list of findings ordered by `numerical severity, -date, title`.

I think it's make more sense to order just by `-date` in this case.

### DIFF
--- a/dojo/templates/dojo/dashboard.html
+++ b/dojo/templates/dojo/dashboard.html
@@ -46,7 +46,7 @@
                         </div>
                     </div>
                 </div>
-                <a href="{% url 'all_findings' %}?duplicate=2&date=2">
+                <a href="{% url 'all_findings' %}?duplicate=2&date=2%o=-date">
                     <div class="panel-footer">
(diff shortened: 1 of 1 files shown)


dojo/templates/dojo/dashboard.html
```diff
@@ -46,7 +46,7 @@
                         </div>
                     </div>
                 </div>
-                <a href="{% url 'all_findings' %}?duplicate=2&date=2">
+                <a href="{% url 'all_findings' %}?duplicate=2&date=2%o=-date">
```
StefanFl, line 49: Shouldn't it be `&o=-date`?

valentijnscholten: Yes looks like some encoding artifact while copy and pasting. Will fix

StefanFl (approved):

Maffooch (approved):

PR #5364, opened 2021-11-01
Fix problem with duplicate projects in GitLab pipeline

StefanFl:
see https://owasp.slack.com/archives/C2P5BA8MN/p1635770164087400

The GitLab pipeline to update product access fails if there is a product with the name of a GitLab project exists but is in the wrong product type.

I cannot test it so it needs a very thorough code review.


Maffooch (approved):
I think this will be okay

alles-klar (approved):

PR #5382, opened 2021-11-03
Auth V2 - Remove legacy authorization part 1: Remove legacy auth from templates

StefanFl:
Auth V2 - Remove legacy authorization part 1

This first of 4-5 PR's removes all decisions in the templates on whether to display UI elements based on legacy authorisation.


valentijnscholten (approved):

Maffooch (approved):
Thanks for making the time to do this!

PR #5383, opened 2021-11-03
Fix migration of API configurations

StefanFl:
plus an issue with authorization, which leads to 404 messages in some cases


valentijnscholten:
Why do we need raw sql here?

valentijnscholten:
because some fields are removed from `models.py` I assume

valentijnscholten:
Did you try the official way to access models, i.e. `Test= apps.get_model('dojo', 'Test')`? This should take into account the version of the Test model present at the point where 0131 is run. I think running raw sql may be a risk of its own that it breaks in some unforeseen scenario's where people are using obscure database like oracle or sqlite or whatever.

StefanFl:
That brings a different error message but still doesn't work. The fields are not there anymore in the model Test class, where would the information about the missing fields come from?

I can't see how to avoid the SQL statement. It is completely SQL standard, so I think it should be safe.

valentijnscholten:
Django builds a complete model/tree/history of all migrations. The field would come from the migration files 0001-0131.

StefanFl:
This is what I tried: 

```
    sq_products = Sonarqube_Product.objects.filter(sonarqube_tool_config__isnull=False)
    for sq_product in sq_products:
        api_scan_configuration = Product_API_Scan_Configuration()
        api_scan_configuration.product = sq_product.product
        api_scan_configuration.tool_configuration = sq_product.sonarqube_tool_config
        api_scan_configuration.service_key_1 = sq_product.sonarqube_project_key
        api_scan_configuration.save()

        Test = apps.get_model('dojo', 'Test')
        tests = Test.objects.filter(sonarqube_config=sq_product)
        for test in tests:
            test.api_scan_configuration = api_scan_configuration
            test.sonarqube_config = None
            test.save()

        sq_product.delete()
```

This raised: 

```
...
django-defectdojo_migration-initializer-1   |   File "/app/dojo/db_migrations/0131_migrate_sonarcube_cobalt.py", line 18, in migrate_sonarqube
django-defectdojo_migration-initializer-1   |     tests = Test.objects.filter(sonarqube_config=sq_product)
...
django-defectdojo_migration-initializer-1   | ValueError: Cannot query "MW SonarQube | mwplatobackend": Must be "Sonarqube_Product" instance.
```

Any ideas?

valentijnscholten:
You should use `get_model` for all Model classes in the migration

valentijnscholten:
Here's an example: https://github.com/valentijnscholten/django-DefectDojo/blob/495b5c810f9cfc748ca10e2b2699f7542d53b30f/dojo/db_migrations/0120_sonarqube_test_and_clean.py

StefanFl:
Don't know why, but this version works. Thanks @valentijnscholten

valentijnscholten (approved):

devGregA (approved):

Maffooch:
Changed the base from master to a dedicated hotfix branch to avoid releasing things from dev that were introduced since 2.4.0

PR #5387, opened 2021-11-04
Remove models for legacy api classes

StefanFl:
Now that the migration `0131_migrate_sonarcube_cobalt` has been changed and is not using imports for `Sonarqube_Product` and `Cobaltio_Product` anymore, these classes can be removed from the model.


StefanFl:
Unit tests reported a problem with the test data fixture. This was due to auth_permissions in the fixture. They point to content_types which are not initialized and there was no content_type 118 anymore because we have only 117 tables after the removal in this PR. Since we don't use the auth_permissions currently in DefectDojo, I deleted them in the test data fixture.

valentijnscholten (approved):

StefanFl:
> Would have been interesting to see if they could have been removed as part of 2.4.1 now that you fixed 0131 migration. I assume so.

For a split second I thought about it yesterday evening to put the removal in the 0132 migration. Would have worked, but I just wanted to get 2.4.1 done as quick as possible.

Maffooch (approved):

PR #5397, opened 2021-11-05
Integration test idempotency

CharlieSears:
## Problem(s)
Two tests were reliably failing when re-running integration tests on the same database following an unsuccessful previous run. This is mitigated by simply destroying and re-initalizing the database, but that didn't appear to be the intention in documentation or code.

1. tests.test_user.test_create_user was failing if the user already existed as the "A user with that username already exists." text was not within a 'success' type field.
2. tests.test_user.test_user_notifications_change was failing if the test had already run as the checkboxes were already selected, when clicked again they were un-selected and the test only validated they were 'checked'.
3. ~docker/entrypoint-integration-tests.sh attempts to run tests/user_standard_test.py which does not exist. This run also sets the value `Frue` in local_settings.py which is not a valid value causing parse errors.~

## Solution(s):
1. The `or` case on tests.test_user.test_create_user now correctly checks that the "A user with that username already exists." message exists within an element with the `help-block` class. A helper method was added for this similar to success/fail message checks.
2. The test now checks the current state and validates that the state has changed after save. This was based on the 'changed' name of the test. I'm open to setting a desired state before executing and ensuring we go from 0 to 1 for example instead of just seeing the bit flipped.
3. ~The test setup for the missing file was removed and the unused reload function was removed. I believe based on the lack of this test in GitHub actions for integration tests and the existence of read-only vs editable user tests in test.test_user this is not necessary. Please let me know if that is not the case and I can remove this edit for this PR.~

### DIFF
--- a/docker/entrypoint-integration-tests.sh
+++ b/docker/entrypoint-integration-tests.sh
@@ -39,6 +39,10 @@ function success() {
     echo "Success: $1 test passed\n"
 }
 
+function reload() {
+  touch /app/dojo/settings/settings.py
+}
+
 echo "IT FILENAME: $DD_INTEGRATION_TEST_FILENAME"
 if [[ ! -z "$DD_INTEGRATION_TEST_FILENAME" ]]; then
     test=$DD_INTEGRATION_TEST_FILENAME
@@ -210,6 +214,20 @@ else
     else
         fail $test
     fi
+
+    test="Read only user profile test"
+    echo "Preparing configuration: USER_PROFILE_EDITABLE=True"
+    echo "USER_PROFILE_EDITABLE=Frue" > /app/dojo/settings/local_settings.py
+    reload
+
+    echo "Running $test"
+    if python3 tests/user_standard_test.py ; then
+        success $test
+    else
+        fail $test
+    fi
+    echo "USER_PROFILE_EDITABLE=True" > /app/dojo/settings/local_settings.py
+    reload
 
 
 # The below tests are commented out because they are still an unstable work in progress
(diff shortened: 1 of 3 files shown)


docker/entrypoint-integration-tests.sh
```diff
@@ -215,20 +211,6 @@ else
...
-
-    echo "Running $test"
-    if python3 tests/user_standard_test.py ; then
-        success $test
-    else
-        fail $test
-    fi
-    echo "USER_PROFILE_EDITABLE=True" > /app/dojo/settings/local_settings.py
-    reload
-
```
valentijnscholten, line 231: these should not be removed, but the test should be added to the GHA

valentijnscholten (changes requested):
see comment

docker/entrypoint-integration-tests.sh
CharlieSears: I can add this back and add the file to the GitHub Action yaml, but the file does not exist so it will be unable to complete the action.

I opened this issue for that missing file: #5400

valentijnscholten: ok yeah those #5400 and #5401 must be resolved first, but better to remove the changes here so we can merge the rest of your PR

valentijnscholten (approved):

docker/entrypoint-integration-tests.sh
dsever: This tests are already in GHA, but I missed to execute local tests...
https://github.com/DefectDojo/django-DefectDojo/blob/master/.github/workflows/integration-tests.yml#L100
https://github.com/DefectDojo/django-DefectDojo/blob/master/tests/user_test.py#L182

CharlieSears:
Looking into that test failure.

valentijnscholten (approved):

StefanFl (approved):

PR #5410, opened 2021-11-08
Authz: Allow global maintainers/owner to add Product Types

valentijnscholten:
We use a Product Types each external client that we build software for. Ideally some non-staff users can add new Product Types, or in our case the CICD process should be able to add new product types. Currently only staff users can add new product types. But I don't want to use a staff user for my CICD deployments or assign staff permissions to architect or project managers. I think Global Owners and Global Maintainers can basically edit all the findings / engagements / products, so I suggest to also let them add new Product Types.

This PR also extracts some repeated code into a central method. The specialty about this `Permission.Product_Type_Add` is that there is no object id to check the permissions against. Luckily @StefanFl agreed the PR was a good idea, let's hope my implementation pleases the German department of the Defect Dojo team :-)

The PR also changes the wording of `Guest` users in the docs to `Regular` users which I think is a more accurate name for these types of users.

### DIFF
--- a/dojo/templatetags/authorization_tags.py
+++ b/dojo/templatetags/authorization_tags.py
@@ -1,7 +1,7 @@
 from django import template
 from crum import get_current_user
 from dojo.authorization.roles_permissions import Permissions
-from dojo.authorization.authorization import user_has_permission
+from dojo.authorization.authorization import user_has_global_permission, user_has_permission
 
 register = template.Library()
 
@@ -9,3 +9,8 @@ register = template.Library()
 @register.filter
 def has_object_permission(obj, permission):
     return user_has_permission(get_current_user(), obj, Permissions[permission])
+
+
+@register.filter
+def has_permission(permission):
+    return user_has_global_permission(get_current_user(), Permissions[permission])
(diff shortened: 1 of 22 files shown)


dojo/templatetags/authorization_tags.py
```diff
@@ -1,11 +1,16 @@
...
 register = template.Library()
 
 
 @register.filter
 def has_object_permission(obj, permission):
     return user_has_permission(get_current_user(), obj, Permissions[permission])
+
+
+@register.filter
+def has_permission(permission):
```
StefanFl, line 15:
```suggestion
def has_global_permission(permission):
```

Then it fits to the other checks invented with this PR.

StefanFl (changes requested):

dojo/templatetags/authorization_tags.py
valentijnscholten: Yeah maybe. I chose this one as I think the other one should be called has_object_permission. But I will change it and we'll see later if we want to do a big revamp or not.

StefanFl: With the authorization for configuration we will get a `has_configuration_permission` as well.Then we will have a nice triplet. I agree, the object based permission checks should have `object` as part of their names. I would like to do it after the legacy authorization has been removed.

StefanFl (approved):

Maffooch (approved):
Good cleanup

alles-klar (approved):

PR #5436, opened 2021-11-12
fixed nuclei templates parser

yelsanya:
Nuclei seems to change names of variables in json reports, fixed to match these names


yelsanya:
Sorry pushed wrong fix

PR #5438, opened 2021-11-12
fixed nuclei report parser

yelsanya:
Nuclei seems to change names of variables in json reports, fixed nuclei report parser to match these names

### DIFF
--- a/dojo/tools/nuclei/parser.py
+++ b/dojo/tools/nuclei/parser.py
@@ -24,12 +24,12 @@ class NucleiParser(object):
         else:
             dupes = {}
             for item in data:
-                template_id = item.get('templateID')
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Thanks. Can you add a unit tests with these new names/tags? And retain support for the old tags as well. We can't support all versions of all parsers, but in this case I think it's quite easy to retain backwards compatability.

yelsanya:
Sure, I will tag you when it will be done

Maffooch (changes requested):
please rebase you changes to the `dev` branch. 
It would also be best to make the parser backward compatible with the old format so that both can be used.

StefanFl:
superseded by #6103

PR #5441, opened 2021-11-12
Link Parent Jira issue to finding's Jira issue

prakashar11:
~~Requirement: Currently in DefectDojo, when Engagement to Epic mapping is enabled, the finding summary is considered as the Epic name and sent to Jira. However, usually development teams have created a separate Epic in their Jira board for tracking the AppSec findings and hence there is a need to configure Epic name associated with each Engagement. Also, even though Tracker field in the Engagement can be configured with Epic link, code in existing version of DefectDojo, doesn't use the Epic link configured in Tracker field to create finding in a given Epic.~~

**Requirement:**

In environments where Jira next gen is used for tracking issues, product development teams would typically create a parent issue (which can translates to an Epic in Jira next gen's functionality to manage roadmap), and link Jira issue created for the vulnerability assessment finding. To support this requirement, enable configuring parent Jira issue at engagement level and enable linking of the same to the Jira issue created for a findings in that engagement.

~~Proposed changes: Changes submitted through this PR enables the functionality to get Epic key & id from the Epic link and uses the same while pushing a finding to create Jira issue~~

**Proposed changes:**

1. Add `enable_parent_issue_linking` in the `Jira_project` model - changes in `dojo/models.py & dojo/forms.py`. 

Screen Shot 2022-01-25 at 2 22 18 PM

2. Add `jira_parent_issue` in `jira-epic-form` so that parent Jira issue can be configured per Engagement. Parsing logic in `dojo/templates/base.html` to make sure parent issue selected even if full Jira instance URL is included in the input

Screen Shot 2022-01-25 at 2 33 38 PM

3. Updates in dojo/jira_link/helper.py to associate parent Jira issue with a Jira issue

- Added new function `add_issue_to_parent` - populates `parent_dict` and updates Jira issue. Care taken to support updating a Jira issue with no parent issue and also to make sure `enable_engagement_epic_mapping` functionality is not impacted
- Modified `add_jira_issue` to update parent issue while creating a Jira issue. 
- Update `process_jira_epic_form` for handling parent issue display in Jira epic form

4. Bulk edit to support associating finding's Jira issue with parent Jira issue configured in each engagement. Changes related to the same are in - `dojo/finding/views.py`, `dojo/templates/dojo/findings_list_snippet.html` & `dojo/templates/dojo/view_test.html`

Screen Shot 2022-01-25 at 2 53 53 PM
Screen Shot 2022-01-25 at 3 19 31 PM

**Unit test cases to be developed:**

1. With `enable_parent_issue_linking` set to `True`, for a given finding, create Jira issue with & without parent Jira issue
2. With `enable_parent_issue_linking` set to `True`, bulk editing of finding (at product level), create & update Jira issue with & without parent Jira issue configured 
3. With `enable_parent_issue_linking` set to `True`, bulk editing of finding (at test cycle level), create & update Jira issue with & without parent Jira issue configured

~~Testing: Tested with Jira cloud~~


valentijnscholten:
Thanks for the idea, but the implementation doesn't align with how the JIRA integration works in Defect Dojo. The way this could/should work is that users should be able to link existing epics to an engagement. Or the engagement should have some kind of 'jira summary' field to use as a title for the epic. Or users just rename the epic in JIRA after it has been created.

prakashar11:
Thanks for the feedback; yes, the existing Tracker filed in Engagement model would have link to existing Epic. However, currently this Epic link configured is not considered while adding a Jira issue to Epic. Hence these proposed changes. Also, in an organization where there are multiple dev teams with their own Jira boards, renaming Epic in Jira creates friction between AppSec team and dev teams

Maffooch:
The title of the epic in jira is correlated to the name of the engagement in dojo. 

If I understand the need for this, you're trying to add a link to en existing epic so that epic mapping will not create a new epic, but attach itself to the existing epic?

If that is the goal, then adding this change in add/update jira issue is repetitive and honsetly not the correct place. Please add it here instead
https://github.com/DefectDojo/django-DefectDojo/blob/813689833f0d03a57a26f318492ea13f426a64b6/dojo/jira_link/helper.py#L1092

prakashar11:
> The title of the epic in jira is correlated to the name of the engagement in dojo.
> 
> If I understand the need for this, you're trying to add a link to en existing epic so that epic mapping will not create a new epic, but attach itself to the existing epic?
> 
> If that is the goal, then adding this change in add/update jira issue is repetitive and honsetly not the correct place. Please add it here instead
> 
> https://github.com/DefectDojo/django-DefectDojo/blob/813689833f0d03a57a26f318492ea13f426a64b6/dojo/jira_link/helper.py#L1092

Thank you; yes, the requirement is to use an existing Epic for an engagement so that when findings in that engagement are pushed to Jira, then Jira issues are created in existing Epic. Have updated the code to reflect the same

valentijnscholten:
If you want to link an existing epic, please implement something similar to what is implemented for linking existing tickets to findings. Just parsing an url is not a good solution. It could be from a different jira instance, and it will fail if someone runs Jira in a subfolder for example https://host/jira

prakrama-bb:
> If you want to link an existing epic, please implement something similar to what is implemented for linking existing tickets to findings. Just parsing an url is not a good solution. It could be from a different jira instance, and it will fail if someone runs Jira in a subfolder for example https://host/jira

Would getting Epic using the following be agreeable?
`epic = engagement.tracker.split(jira_project.jira_instance.url + '/browse/')[1]`

prakashar11:
> If you want to link an existing epic, please implement something similar to what is implemented for linking existing tickets to findings. Just parsing an url is not a good solution. It could be from a different jira instance, and it will fail if someone runs Jira in a subfolder for example https://host/jira

Updated logic to get epic as follows; hope it is acceptable
`epic = engagement.tracker.split(jira_project.jira_instance.url + '/browse/')[1]`

Maffooch:
This looks reasonable! A couple more things before approval:

- add this feature to the documentation here
  - https://github.com/DefectDojo/django-DefectDojo/blob/master/docs/content/en/integrations/jira.md
  - Add an example of what format is acceptable/will work
- Add a unit test for this
  - https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/unittests/test_jira_config_engagement_epic.py

valentijnscholten (changes requested):
I still don't like it. It feels dirty to just use some field and use it in the JIRA integration code. People might use the field for other trackers like github. I propose to add an 'existing epic' field to the JIRA form for an engagement. Similar to what can be done when editing an issue. There is also parsing logic that works with JIRA instances served under a subfolder as mentioned.

prakashar11:
> The field is probably from before we had proper JIRA integration. It might also conflict with engagements that already have an epic and then get this field set. People might use the field for other trackers like github. I propose to add an 'existing epic' field to the JIRA form for an engagement. Similar to what can be done when editing an issue. 

Sure, agree that a new field to the engagement Jira form to configure existing Epic, is a cleaner solution. However the help tip for the 'Tracker' field reads as  'Link to epic or ticket system with changes to version' and I couldn't find any code to use the value set in this field, hence earlier proposed reusing this tracker field. 

> There is also parsing logic that works with JIRA instances served under a subfolder as mentioned.

Also, please help me by pointing to the code that implement 'parsing logic that works with JIRA instances served under a subfolder'

valentijnscholten:
> > There is also parsing logic that works with JIRA instances served under a subfolder as mentioned.
> 
> Also, please help me by pointing to the code that implement 'parsing logic that works with JIRA instances served under a subfolder'

It's in javescript, but basically take the part after the latest slash, so `split('/')[-1]`

prakashar11:
> > > There is also parsing logic that works with JIRA instances served under a subfolder as mentioned.
> > 
> > 
> > Also, please help me by pointing to the code that implement 'parsing logic that works with JIRA instances served under a subfolder'
> 
> It's in javascript, but basically take the part after the latest slash, so `split('/')[-1]`: https://github.com/valentijnscholten/django-DefectDojo/blob/7e640cb6345059fdce621b2b99f21ba2ec155b7f/dojo/templates/base.html#L764

Updated code by introducing a new field in Jira epic form to configure parent issue. Used Jira next gen API to update parent issue. Please let me know whether approach taken is agreeable

valentijnscholten:
Not sure if the description still matches the implementation? The term "parent" is usually used for subtasks in JIRA that are created un a parent issue. Epics are something else and usually use the epic_link field.

prakashar11:
> Not sure if the description still matches the implementation? The term "parent" is usually used for subtasks in JIRA that are created un a parent issue. Epics are something else and usually use the epic_link field. Might be better to use the epic_link term to make it clear it's about epics? Also could you add testcases, that will also help us understand the intended/desired behaviour.

As I understand, Jira next gen uses 'parent' synonymously with 'epic' ([1][2]), to create hierarchy while using Jira roadmap feature. 

Sure, will add test cases and update PR

Ref:
[1] https://confluence.atlassian.com/jirasoftwarecloud/manage-epics-in-team-managed-projects-960142560.html
[2] https://community.atlassian.com/t5/Jira-Software-questions/Next-Gen-Story-and-Epic-Link/qaq-p/910068)

valentijnscholten:
DD currently doesn't support next gen projects, did you not experience other problems?

prakashar11:
> DD currently doesn't support next gen projects, did you not experience other problems?

Not experienced any issue so far with DD Jira next gen bidirectional integration except for this -> closing of issue in Jira doesn't update DD finding status (we are not using this capability as we rely on scanning tool to update DD finding status). If you have few test cases in mind for me to run in our DD Jira next gen setup, let me know & I can report back the test results.

valentijnscholten:
You need to make sure everything you add also works for JIRA Server and non-nextgen projects, as that is what we have been developing against unitl now. You need to add testcases for new functionality that you add, so adding issues to existing epics. The description still is out of sync with what you have implemented in this PR. So please update that to avoid confusion and reviewers having to spend extra time understanding what is going on in the PR.

devGregA:
I'm sorry for getting around to this so late, but I don't see a way forward on this PR. Closing. Please feel free to reach out to me to discuss.

PR #5444, opened 2021-11-13
Allow to force login form

kiblik:
When `SHOW_LOGIN_FORM` is `False`, `SOCIAL_LOGIN_AUTO_REDIRECT` is `True` and there is only one "Social" or "SAML", the login page is skipped and the request is automatically redirected to IdP. However, sometimes it is useful to have available the standard method as last resort (e.g. `admin` user need to log in because of a change of some settings or permissions).

This PR allows forcing login page with form when `<DD_HOST>/login?force_login_form` is accessed.


valentijnscholten (approved):

Maffooch (approved):
Good change!

PR #5450, opened 2021-11-15
main(helm): remove deprecated stable repo

alles-klar:
Removing the deprecated helm stable repo. Also updating outdated dependencies. Not sure why renovate doesn't update them...

On behalf of DB Systel GmbH


Maffooch:
@alles-klar you're the helm guy, so let us know when you're all done with this.
@dsever please review if you can

dsever:
> @alles-klar you're the helm guy, so let us know when you're all done with this. @dsever please review if you can

I assume this is not ready, yet

alles-klar:
True, I broke my local kubernetes installation, so I have to fix it first. I will notify you if it is ready.

valentijnscholten:
I see you sneaked in a k8s upgrade as well... :-)

alles-klar:
> I see you sneaked in a k8s upgrade as well... :-)

Yes, but tests for 1.16 failed. Not sure why, maybe the mysql helm chart isn't compatible. But 1.16 is EOL, so I think we don't have to support it anymore. Do you agree?

dsever:
> > I see you sneaked in a k8s upgrade as well... :-)
> 
> Yes, but tests for 1.16 failed. Not sure why, maybe the mysql helm chart isn't compatible. But 1.16 is EOL, so I think we don't have to support it anymore. Do you agree?

It's not that simple, talking out of my head at least 1.19 version needs to be tested, because it still uses beta ingress resource, while after 1.20 it is stable, and as well minimal tested version should be validated against gap and aws k8s managed services

alles-klar:
@dsever Can you take a look please? From my perspective it is ready. I prefer to make the update to helm 3 in another PR.

dsever:
I will check it today, helm 3 is already prepared in my repo.

.github/workflows/k8s-testing.yml
```diff
@@ -108,9 +108,22 @@ jobs:
 
     strategy:
       matrix:
-        databases: [pgsql, mysql]
-        brokers: [redis, rabbit]
-        k8s: ['v1.16.0','v1.20.0']
+        include:
```
dsever, line 111: Personally I'm fun of  deep testing approach, to test all combination.

readme-docs/KUBERNETES.md
```diff
@@ -52,7 +52,6 @@ helm repo update
```
dsever, line 52: We should  have in documentation what version are supported. All this changes needs to be documented during the release, because migration to bitnami for some users could be problematic.

.github/workflows/k8s-testing.yml
alles-klar: I don't think we really need it and with "Green IT" in mind it better to use less resources.

dsever: Ok, I can survive it :)

docs/content/en/getting_started/upgrading.md
```diff
@@ -84,7 +84,9 @@ notifications but also for sending the reset password emails. It is highly recom
 value if you are satisfied. If you installed DefectDojo earlier, you can expect `"from@example.com"` there. A fresh
 installation will use `"no-reply@example.com"`
 
+This release [updates](https://github.com/DefectDojo/django-DefectDojo/pull/5450) our helm dependencies. There is a breaking change if you are using the mysql database from the helm chart because we replaced the deprecated chart from the stable repo with a chart from bitnami. If you have persistance enabled, ensure to backup your data before upgrading. All data get lost when replacing the mysql chart during the upgrade. For data migration take a look at the mysql backup and restore process.
```
dsever, line 87: Have you tested upgrade? What will happen if you define existing Claim  `primary.persistence.existingClaim`?

alles-klar: Upgrade with default values works. I don't know what happens with an existing claim.

dsever: Sorry I was not testing it, but when you upgrade existing (generated from stable) mysql to one from bitnami, will data be preserved, or it will create completely new (empty) PVC? 

So my question is, in case if it will not use same PVC (new one will be created) and prior upgrade you specify exitingClaim name (one that is created from stable) what will happen?  Just wondering, in case it works it can be use for data migration, even I'm fully with you about the statement of backup/restore. 

I'm reading you statement that all data will be lost, if persistence is not enable data will be lost anyhow.

alles-klar: I don't know. I wrote that the data will be lost to warn people and to assume the worst case - maybe there is a way to prevent it. Anyway I don't think people are running a productive defectdojo with the mysql chart (it is deprecated for a while now!). Most will use an external database. And if there is someone, it will be easier to restore your data from the backup than testing such a migration scenario. But feel free to write a migration guide :)

readme-docs/KUBERNETES.md
alles-klar: I added a section, but I don't want to write the exact versions in the docs. It will outdated soon because someone will forget to update it.

docs/content/en/getting_started/upgrading.md
dsever: That is ok.

dsever (approved):
Now it looks ok

valentijnscholten (approved):
approving by proxy of @dsever

StefanFl (approved):

PR #5466, opened 2021-11-18
Add a tag for each vulnerability

shubhindia:
This adds a tag for each Trivy vulnerability which is present in the report json.
Which makes it easy to filter the vulnerabilities according to their `Type`
e.g. `python-pkg, node-pkg, debian`

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -6,10 +6,14 @@ categories:
   - title: '💣 Breaking changes'
     labels:
       - 'Breaking Changes'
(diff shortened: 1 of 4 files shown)


valentijnscholten:
Please base your PR against `dev`

shubhindia:
@valentijnscholten done

Maffooch:
@shubhindia you've got one extra file in your changes log

shubhindia:
Closing this as it is no longer a general use-case. We have made the changes to our specific use and this doesn't seem relevant enough for general use.

Issue #5356, opened 2021-10-29
Endpoint with protocol length > 20 characters causes error in report uploads (Tested with nessus)


cristiano-corrado:
When importing an endpoint which has **protocol** length > 20 characters the import fails

error in UI :

```
An exception error occurred during the report import:value too long for type character varying(20)
```

Causes of the problem is a restrictive **max_length=20** in the Endpoint model :

```
class Endpoint(models.Model):
    protocol = models.CharField(null=True, blank=True, max_length=20,
                                 help_text="The communication protocol/scheme such as 'http', 'ftp', 'dns', etc.")
```

Example string causing error:
as seen below the string is 21 chars
```
fw1-topology-download://xx-0-0-0-0.xx-xx-x.xxx.xxxx
```

By truncating the endpoint as the example below is possible to upload the Nessus report :

```
fw1://xx-0-0-0-0.xx-xx-x.xxx.xxxx
```

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [x]  Kubernetes
- [ ] GoDojo


**Logs** 

```
                                                                                                                                                                         │
│ uwsgi Traceback (most recent call last):                                                                                                                                                                                                                   │
│ uwsgi   File "/app/./dojo/engagement/views.py", line 602, in import_scan_results                                                                                                                                                                           │
│ uwsgi     test, finding_count, closed_finding_count = importer.import_scan(scan, scan_type, engagement, user, environment, active=active, verified=verified, tags=tags,                                                                                    │
│ uwsgi   File "/app/./dojo/importers/importer/importer.py", line 299, in import_scan                                                                                                                                                                        │
│ uwsgi     new_findings = self.process_parsed_findings(test, parsed_findings, scan_type, user, active,                                                                                                                                                      │
│ uwsgi   File "/app/./dojo/importers/importer/importer.py", line 125, in process_parsed_findings                                                                                                                                                            │
│ uwsgi     ep, created = endpoint_get_or_create(                                                                                                                                                                                                            │
│ uwsgi   File "/app/./dojo/endpoint/utils.py", line 79, in endpoint_get_or_create                                                                                                                                                                           │
│ uwsgi     return Endpoint.objects.get_or_create(**kwargs)                                                                                                                                                                                                  │
│ uwsgi   File "/usr/local/lib/python3.8/site-packages/django/db/models/manager.py", line 85, in manager_method                                                                                                                                              │
[...]

PR #5471, opened 2021-11-18
Nessus: store only standard protocol names

kiblik:
Fix https://github.com/DefectDojo/django-DefectDojo/issues/5356

Parser use protocol names from `svc_name` only if it is a well-known name of the schema.
In other cases, the protocol will be `tcp` or `udp`

Note: Change was prepared only for `nessus xml`. `nessus csv`, `nessus_was xml` and `nessus_was csv` weren't changed because the example of a report is missing.


Maffooch:
@kiblik please see https://github.com/DefectDojo/django-DefectDojo/tree/master/dojo/unittests/scans/nessus_was

kiblik:
> @kiblik please see https://github.com/DefectDojo/django-DefectDojo/tree/master/dojo/unittests/scans/nessus_was

@Maffooch, I know where `nessus_was xml` example reports are. I just wanted to be sure that it will be created on the real problem from the real report. Not just an edited version based on my thoughts.

Maffooch (approved):
Gotcha makes sense

valentijnscholten:
Can we try to get some example reports or do an educated guess? I think it can get quite confusing if there are too many differences between the parsing of different report formats. Looking at the code change I think it should be easy to port it to the other report formats?

kiblik:
If "educated guess" is not too risky, I have no problem doing it and preparing a parser for `nessus_was xml`.

And what about `nessus* csv`?

kiblik:
Result of "educated guess": Nessus WAS is for _Web Application Scanning_ so there will not be protocol other than "http" or "https"

kiblik:
Btw, is it not a better idea to merge nessus-parser and nessus_was-parser into one? I checked differences and it looks there are only `row.get(...)` or `if '...' in row:` so there is nothing to break

valentijnscholten:
> Btw, is it not a better idea to merge nessus-parser and nessus_was-parser into one? I checked differences and it looks there are only `row.get(...)` or `if '...' in row:` so there is nothing to break

Sounds like a good idea, although I don't know Nessus reports to well. Might be better to do in a separate PR anyway?

kiblik:
So can we merge this PR and consolidation of `nessus*` will be done separately?

valentijnscholten:
I guess. Can you create an issue that lists the things left to be changed/fixed?

kiblik:
Issue created, can we merge this?

valentijnscholten (approved):

valentijnscholten:
Done, thanks for you contributions @kiblik

PR #5476, opened 2021-11-20
Add support for pushing tags to jira

Maffooch:
I noticed today that tags within dojo do not get pushed into jira. This makes that happen.


valentijnscholten (approved):
Nice! Actually yesterday I noticed this too and put it on my todo list. Removing it again :-)

StefanFl (approved):
Not that I really understand it, but it seems to be well tested.

PR #5478, opened 2021-11-20
New report format for Trufflehog3, use hashcode for dedupe

### LINKED ISSUE #5422: Trufflehog3 import error
**Bug description**
Trying import a report from Trufflehog3:2.0.4, 3.0.1, 3.0.3 and see this 
`An exception error occurred during the report import:'reason'`

**Steps to reproduce**
1. Scan any project with trufflehog3 and get more that 10 findings
2. Import the report 
3. See error `An exception error occurred during the report import:'reason'`

**Expected behavior**
No error

- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Screenshots**
image

StefanFl:
fixes #5422

The reporting format of Trufflehog3 seems to have changed. With this PR the scanner supports both formats.


valentijnscholten (approved):
A pity these scanners don't includ a `version` or `format` property in their reports, but still a good PR.

valentijnscholten:
@StefanFl I think it also changes the hashcode algorithm and requires a hashcode recompute after upgrading?

StefanFl:
@valentijnscholten Not really, the hashcode algorithm hasn't been changed. The settings change is only about using the hashcode for deduplication. Findings imported with an older version of Trufflehog3 would still get the same hashcode as before.

Maffooch (approved):

Issue #5312, opened 2021-10-21
Reimporting scans on an empty test gives less results than full import


malexmave:
## Description
I have a semgrep results JSON with 9 findings. Depending on the way I import it into DefectDojo, not all findings are imported correctly.

## Steps to Reproduce
All instructions assume a fresh standard dockerized install without any changes to the settings. 

Method 1, which will give all findings:
1. Create a new product in DD
2. Select "Import Scan Results"
3. Fill out the form and upload the [semgrep results file](https://gist.github.com/malexmave/c1d9869abf069a1f0191fa3c6d0c2fa6)
4. Observe that all 9 findings are recognized

Method 2, which will give incomplete results:
1. Create a new product in DD
2. Create a new engagement for that product
3. Create a new test of type "semgrep" for that engagement
4. Select the test (it should be empty)
5. Select "Re-upload scan"
6. Upload the [results JSON](https://gist.github.com/malexmave/c1d9869abf069a1f0191fa3c6d0c2fa6)
7. Observe that only 5 findings are recognized

Note that the results JSON contains 9 findings based on 5 rules - it seems like some sort of deduplication logic takes into account the used rule, but not the place where the rule matched. Thus, matches of the same rule in different places in the code are not recognized as distinct. However, you will find that the results are not even marked as "duplicate findings" or anything like that, they are simply discarded.

## Expected behavior
I would expect both methods to give the same results.

## Deployment method
- [x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

## Environment information
 - Operating System: Mac OS 11.6
 - DefectDojo version (see footer) or commit message: 2.3.1
 - Standard settings (no changes to settings.dist.py or the system settings from the defaults)

## Additional context
This problem is not exclusive to the semgrep scan type. We found it because the secureCodeBox DefectDojo integration uses the second workflow: It creates product, engagement and test and then runs a re-import instead of doing a regular import, because this lets it set a description of the test (which is not possible otherwise, it seems). For secureCodeBox scan types that are not supported by DefectDojo, it uses the `generic` parser. When running a git-repo-scanner job and importing the results, we get the same result: The [findings file](https://gist.github.com/malexmave/a2f1b8d266e8ce6dadb7c31859be16d9) contain four repositories, which are correctly recognized on a full import (Method 1), but only one is recognized on the reimport (Method 2).

damiencarol:
[...]

PR #5483, opened 2021-11-22
Use hash_code for deduplication of detect-secrets scans

StefanFl:
Reimporting of detect-secrets scans swallowed some findings, due to using the legacy deduplication algorithm, see https://github.com/DefectDojo/django-DefectDojo/issues/5312#issuecomment-975171871.


Maffooch (approved):

valentijnscholten:
The list gets longer and longer, should we consider the legacy dedupe mostly broken for reimport since reimport was changed to use the dedupe / hash_code algorithm?

StefanFl:
I agree @valentijnscholten, reimport doesn't work properly for a lot of scanners when you don't set the deduplication algorithm explicitly. Hash_code deduplication should be a good default for static scanners, but I have no idea about dynamic scanners.

valentijnscholten (approved):

PR #5484, opened 2021-11-22
Checkmarx: parse and set false positive, active and verified fields correctly

### LINKED ISSUE #5480: Checkmarx: false positive imported as active/verified
**Bug description**
In checkmarx GUI, it's possible to mark findings as false positive. Since 2019 the `false_p` value is correctly valued from the report `FalsePositive` field for checkmarx. However if "active/verified" is set to "true" during the import (in the GUI or the API), then the false positives are imported as active/verified.

**Steps to reproduce**
1. Import single_finding_false_positive.xml attached (api or gui) with active=true, verified=true

**Expected behavior**
- The checkmarx parser should set active/verified accordingly to the state of the finding in checkmarx (see tests.xlsx attached, tab checkmarx_status)
- (won't be implemented right away) The active/verified values that are given during the import in the gui or the API should only be a default value in case no value is given by the parser. If the report specifically sets a value for a given finding, that value should be taken into account instead of the default value. 

- [x ] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
2.5.0-dev
[2021-11-17 22:33:25 +0100] a740bdb38: Bump markdown from 3.3.4 to 3.3.5 (#5460) [ (HEAD -> dev, origin/dev)]

**Sample scan files**
See tests.xlsx for tests done with different values of active/verified in the report and during the import:
- single_finding_false_positive.xml : one false positive finding (see tab checkmarx_false_p)
- single_finding.xml : one not false positive finding (see tab checkmarx_normal)
- 5_zap_sample_one_single_vuln.xml: a zap finding (see tab ZAP)
- checkmarx_states.xml : one finding in every existing state in checkmarx (see expected mapping in tab checkmarx_status)

[tests.zip](https://github.com/DefectDojo/django-DefectDojo/files/7579773/tests.zip)

**Screenshots**

Other tools that may suffer that kind of issues: those parsers set a value for active and/or verified. 
`The proposed behavior could improve the functioning of those tools tools` -> no, won't change anything global other than checkmarx parser
- whitehat_sentinel
- sonarqube_api
- mozilla_observatory
- kics
- intsights
- h1
- generic
- dependency_check
- crashtest_security
- cobalt_api
- checkmarx_osa
- aws_prowler
- awssecurityhub

ptrovatelli:
fixes #5480


ptrovatelli:
fixes https://github.com/DefectDojo/django-DefectDojo/issues/5480

dojo/unittests/dojo_test_case.py
```diff
@@ -441,6 +441,49 @@ def import_scan_with_params(self, filename, scan_type='ZAP Scan', engagement=1,
 
         return self.import_scan(payload, expected_http_status_code)
 
+    def import_checkmarx_scan_with_params(self, filename, scan_type='Checkmarx Scan', engagement=1, minimum_severity='Low', active=True, verified=True,
```
valentijnscholten, line 444: sorry for barking in on a draft PR, but I think you could just use the existing import_scan_with_params and provide the scan_type parameter?

ptrovatelli: you're right :)

valentijnscholten (approved):

valentijnscholten (approved):

valentijnscholten:
Every time I look at this code / logic, I get confused. When looking back https://github.com/DefectDojo/django-DefectDojo/pull/4014 it wasn't fully clear and although a unit test was added, it doesn't actually test this logic.

Should we switch to a more intuitive schema where the input parameters in API / UI have 3 values:
- TRUE: All findings marked as active, regardless of what the parser sets
- FALSE: All findings marked as inactive regardless of what the parser sets 
- NONE: Use value from parser, which defaults to active if the parser doesn't set anything

Maffooch:
I've understood the behavior to be if the active/verified flags were not set at import time, then the it would be up to the parser to figure it out. 
https://github.com/DefectDojo/django-DefectDojo/blob/1a1cf79d303046e6c21bbc2cb591058b1b875001/dojo/importers/importer/importer.py#L88-L92

valentijnscholten:
> hello @valentijnscholten . Yes it would be ideal to have 3 values but it will require a lot of changes . How about going through with this PR that will fix a blatant issue with checkmarx and then open an issue in order to improve things?

Yeah, that's fine. That was my idea as well and I approved the PR already.

ptrovatelli:
@valentijnscholten however it seems we are not in line with which is default and which is override :) 
you said: 
TRUE: All findings marked as active, regardless of what the parser sets
FALSE: All findings marked as inactive regardless of what the parser sets
NONE: Use value from parser, which defaults to active if the parser doesn't set anything

it means that the parser is the default value, but if something is given during the import other than none, the value given during the import is used. 

I think it should be the opposite because the value from the parser is finer : it can be different for each finding so it makes more sense to use this as the override that will tune the default value given by the import form. 

do we really have a use case to import all findings with inactive status whater the status from the parser? why would we import that kind of report?

Maffooch (approved):

PR #5487, opened 2021-11-23
Fixed multiple comma/space separated tags issue via API call

blakeaowens:
Previously, an API call with tags in the format of "test:tag1,test:tag2,test:tag3" would import as one tag of "test:tag1,test:tag2,test:tag3"

With this fix, tags of this format will now import as regular looking tags, i.e. "test:tag1", "test:tag2", "test:tag3"

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -27,6 +27,7 @@ from django.contrib.auth.password_validation import validate_password
 from django.utils import timezone
 import datetime
 import six
+import re
 from django.utils.translation import ugettext_lazy as _
 import json
 import dojo.jira_link.helper as jira_helper
@@ -104,21 +105,16 @@ class TagListSerializerField(serializers.ListField):
         if not isinstance(data, list):
             self.fail('not_a_list', input_type=type(data).__name__)
 
-        # data_safe = []
+        data_safe = []
         for s in data:
             if not isinstance(s, six.string_types):
                 self.fail('not_a_str')
 
             self.child.run_validation(s)
 
-            # if ' ' in s or ',' in s:
-            #     s = '"%s"' % s
+            data_safe += re.split('[, ]', s)
 
-            # data_safe.append(s)
-
-        # internal_value = ','.join(data_safe)
-
-        internal_value = tagulous.utils.render_tags(data)
+        internal_value = tagulous.utils.render_tags(data_safe)
 
         return internal_value
         # return data
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -27,7 +27,6 @@ from django.contrib.auth.password_validation import validate_password
 from django.utils import timezone
 import datetime
 import six
-import re
 from django.utils.translation import ugettext_lazy as _
 import json
 import dojo.jira_link.helper as jira_helper
@@ -112,7 +111,7 @@ class TagListSerializerField(serializers.ListField):
 
             self.child.run_validation(s)
 
-            data_safe += re.split('[, ]', s)
+            data_safe += tagulous.utils.parse_tags(s, max_count=0, space_delimiter=True)
 
         internal_value = tagulous.utils.render_tags(data_safe)
 
--- a/dojo/unittests/test_rest_framework.py
+++ b/dojo/unittests/test_rest_framework.py
@@ -353,7 +353,7 @@ class BaseClass():
                 tag_list = []
                 for tag_string in self.payload.get('tags', None):
                     tag_list += tagulous.utils.parse_tags(tag_string, max_count=0, space_delimiter=True)
-                
+
                 self.assertEqual(len(tag_list), len(response.data.get('tags', None)))
                 for tag in tag_list:
                     # logger.debug('looking for tag %s in tag list %s', tag, response.data['tags'])
--- a/dojo/unittests/test_rest_framework.py
+++ b/dojo/unittests/test_rest_framework.py
@@ -37,6 +37,7 @@ from drf_spectacular.settings import spectacular_settings
 import logging
 import pathlib
 import json
+import tagulous
 from dojo.authorization.roles_permissions import Permissions
 
 
@@ -302,7 +303,7 @@ class BaseClass():
             # print(open_api3_json_schema)
             # validator = ResponseValidator(spec)
 
-            check_for_tags = False
+            tag_list = []
             if hasattr(self.endpoint_model, 'tags') and self.payload and self.payload.get('tags', None):
                 # create a new instance first to make sure there's at least 1 instance with tags set by payload to trigger tag handling code
                 logger.debug('creating model with endpoints: %s', self.payload)
@@ -312,7 +313,8 @@ class BaseClass():
                 # print('response:', response.content[:1000])
                 check_for_id = response.data['id']
                 # print('id: ', check_for_id)
-                check_for_tags = self.payload.get('tags', None)
+                for tag_string in self.payload.get('tags', None):
+                    tag_list += tagulous.utils.parse_tags(tag_string, max_count=0, space_delimiter=True)
 
             response = self.client.get(self.url, format='json')
             # print('response')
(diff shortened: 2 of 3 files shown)


valentijnscholten:
To have multiple tags, you need to provide a list of tags.

valentijnscholten (changes requested):
i don't think we want to change this behaviour.

blakeaowens:
> i don't think we want to change this behaviour.

Both still work (i.e. a list of tags, a single entry with multiple comma separated tags, or a mixture of both). Examples of current working method vs. my changes below:

Here are screenshots from the demo site: (Can be seen [here](https://demo.defectdojo.org/engagement/22))
IMG_7491
IMG_4623

Here are screenshots from my local instance with expected behavior:
Screen Shot 2021-11-23 at 2 28 39 PM
Screen Shot 2021-11-23 at 2 28 55 PM

valentijnscholten:
Check out the failing unit tests to see the expected behavior.

Maffooch:
This looks like additional support for tags rather than a change in behavior. After looking at the unit tests, I think this change makes sense as a "auto splitting" feature. 

https://github.com/DefectDojo/django-DefectDojo/blob/813689833f0d03a57a26f318492ea13f426a64b6/dojo/unittests/test_rest_framework.py#L1076

I would expect this to be two separate tags

valentijnscholten:
I don't understand why we want this autosplitting behaviour. I implemented the `tagulous` library and tried to keep it clean and backwards compatible with what we had before and prefer to stick to a clean implementation.

This: 
`tags": ["mytag, yourtag"]` 
is an example of a single string in a list, which should be treated as a single string. It's even inside quotes, so I think it's bad to just deviate from all the standards and start splitting it. What if there is no space next to the comma? What if there are two comma's? What if someones edits an existing finding that has a tag in it with comma's? It would result in the tags getting splitted when doing a PUT or PATCH request. If you GET the data for an object with tags and the PUT/PATCH it back into defect dojo, nothing should change. I spent a lot of time on the tag implementation and wouldn't want anyquick and dirty things in there.

What usecase can you not serve with the current implemetnation in `dev`/ 2.4.1?

blakeaowens:
> I don't understand why we want this autosplitting behaviour. The idea has always been to support as many usacases as possible, and some people may want to use comma's in tags: https://www.google.com/search?q=use+commas+in+tags I implemented the `tagulous` library and tried to keep it clean and backwards compatible with what we had before and prefer to stick to a clean implementation.
> 
> This: `tags": ["mytag, yourtag"]` is an example of a single string in a list, which should be treated as a single string. It's even inside quotes, which are there explicitly to have it handled as a single tag. I think it's bad to just deviate from all the standards and start splitting it. What if there is no space next to the comma? What if there are two comma's? What if someones edits an existing finding that has a tag in it with comma's? It would result in the tags getting splitted when doing a PUT or PATCH request. If you GET the data for an object with tags and the PUT/PATCH it back into defect dojo, nothing should change. I spent a lot of time on the tag implementation and wouldn't want anyquick and dirty things in there. It would also create a difference between the UI and API behaviour.
> 
> What usecase can you not serve with the current implemetnation in `dev`/ 2.4.1?

I do not understand this design. "tags": ["mytag, yourtag"] would clearly indicate two tags to me. Likewise, "tags": ["ci/cd", "api"] would also clearly indicate two tags. If you notice in my implementation, the list of tags are tokenized by both spaces and/or commas.

I would also argue that a string containing commas could easily be interpreted as a list, for example:
"tags":  ["tag1,tag2,tag3"] versus "tags":  ["tag1", "tag2", "tag3"] should be interpreted the same way by the serializer.

This is not a change removing existing functionality, but a change bringing additional functionality.

I can also add some checks to touch those corner cases in which you mentioned that may not be covered already.

valentijnscholten:
> I do not understand this design. "tags": ["mytag, yourtag"] would clearly indicate two tags to me. Likewise, `"tags": ["ci/cd", "api"]` would also clearly indicate two tags. 

I disagree. `"tags": ["mytag, yourtag"]` is a list with one tag containing a comma.
 
> This is not a change removing existing functionality, but a change bringing additional functionality.

Wrong. It's removing the possibility to have a comma inside a tag. For example:

image

Would no longer be possible. Upon editing this object using the api, the tags would get split into 3 tags rendering them useless.
 
> I can also add some checks to touch those corner cases in which you mentioned that may not be covered already.
> 
> Edit: I just checked, and all of these produce the expected result of `"tags": ["tag1", "tag2"]` `"tags": ["tag1,,tag2"]` `"tags": ["tag1, tag2"]` `"tags": ["tag1 tag2"]`
It doesn't matter as we want to support comma's inside tags.

damiencarol:
@valentijnscholten is right. We can't have a broken implementation with some auto-magic splitting that will prevent us to use space or comma in tag values. I don't want this feature

damiencarol (changes requested):
not a good feature and break the use of tags with comma or space in it

blakeaowens:
> > I do not understand this design. "tags": ["mytag, yourtag"] would clearly indicate two tags to me. Likewise, `"tags": ["ci/cd", "api"]` would also clearly indicate two tags.
> 
> I disagree. `"tags": ["surname, firstname"]` is a list with one tag containing a comma.
> 
> > This is not a change removing existing functionality, but a change bringing additional functionality.
> 
> It's removing the possibility to have a comma inside a tag. For example:
> 
> image
> 
> Would no longer be possible. Upon editing this object using the api, the tags would get split into 3 tags rendering them useless.
> 
> Same for:
> 
> image
> 
> > I can also add some checks to touch those corner cases in which you mentioned that may not be covered already.
> > Edit: I just checked, and all of these produce the expected result of `"tags": ["tag1", "tag2"]` `"tags": ["tag1,,tag2"]` `"tags": ["tag1, tag2"]` `"tags": ["tag1 tag2"]`
> 
> It doesn't matter as we want to support comma's inside tags.
> 
> `tagulous` explicitly supports comma's and spaces inside tags. It has more features, for example [tag trees](https://radiac.net/projects/django-tagulous/documentation/models/tag_trees/), which use a forward slash .
> 
> I'd like to stay as close as possible to the `tagulous` upstream and intended functionality so we don't get any surprises/deviations or blocked upgrades.
> 
> If you want to split tags by comma or spaces, you can do so before inserting them into the `tags` list in the request. Is there any usecase you cannot serve with the current implementation in `dev` / 2.4.1?
> 
> If you want to discuss it more with other members of the team, please come to our Slack channel.

While I see the dilemma with this feature addition and your usage examples, they directly contradict the `tagulous` tag string parsing documentation, which can be found [here](https://radiac.net/projects/django-tagulous/documentation/parser/). The documentation clearly states _commas take priority_.

Also, tags cannot contain spaces through the UI, only through the API, which further supports some sort of change is necessary to keep consistency with the `tagulous` documentation and how tag strings are interpreted in DefectDojo as of now.

valentijnscholten:
The tagulous parser is used when parsing multiple tags in a single string. In the API we parse multiple tags from a list of strings which are interpreted as one tag per string. I don't know how to make it more clear as I now have explained it three times.

In the mean time you do not mention why you want this behaviour changed, even though I have asked now multiple times if there is a usecase that cannot be handled by the current implementation.

The UI supports comma's and spaces by using quotes as specified in the tagulous docs.

image

blakeaowens:
> The tagulous parser is used when parsing multiple tags in a single string. In the API we parse multiple tags from a list of strings which are interpreted as one tag per string. I don't know how to make it more clear as I now have explained it three times.
> 
> In the mean time you do not mention why you want this behaviour changed, even though I have asked now multiple times if there is a usecase that cannot be handled by the current implementation.
> 
> The UI supports comma's and spaces by using quotes as specified in the tagulous docs.
> 
> image

Alrighty.

I believe my most recent commit reflects all of our comments here. The Tagulous rules in relation to parsing tag strings are followed EXACTLY as listed in the Tagulous documentation.

Want commas or spaces in your tag name? Escape them in double quotations.
Want multiple tags in one tag string? Delimit them with either commas or spaces, however commas will take higher priority if both commas and spaces are present.

You have asked for the use case that our current implementation does not support, but I thought it was clear by the title of the PR. I cannot include multiple tags in one tag string from the command line. Please see the example in the initial PR.

It is not that I am AGAINST commas or spaces in tags, it is that our current implementation leaves out the FIRST parsing rule for tags from the documentation I previously linked, which is the entire reason I created this PR.

While I admit my first implementation did not portray much thought in regards to these special cases we discussed in this comment thread, I believe it does now. Likewise, the most recent commit DOES NOT REMOVE any functionality that already exists.

valentijnscholten:
With the latest changes the tags still get split up if you GET an object and then POST it back because the rendering of the tags is not changed in accordance with the parsing of tags.

Anyway, as said multiple times now we do NOT want to change the existing behaviour. The unit tests were made to capture existing behaviour and make sure the implementation is backwards compatible. At the moment we don't have any desire to change this. We may revisit this when creating Defetc Dojo 3.0 and/or APIv3.

PR #5490, opened 2021-11-23
Update Nessus WAS parser to catch the lack of a port in CSV Parser

Maffooch:


valentijnscholten:
The sslscan parser also sets ports based on protocol. Should this be something generic somewhere? Maybe even inside endpoint? I mean to avoid that at some point we have 10 places where https gets translated into port 443? Or to avoid that for some parsers the port is always set even for default values and for other parsers it doesn't get set? Or am I going to far now?

Maffooch:
Ideally, the [from_uri](https://github.com/DefectDojo/django-DefectDojo/blob/813689833f0d03a57a26f318492ea13f426a64b6/dojo/models.py#L1400) helper should be used that fills all fields of the Endpoint model. Assigning port based on protocol is used as a very last resort.

I guess it can be generalized

Maffooch:
So I checked this out and found that sslabs actually sets protocol based on the port. It was also the only instance of that I found.

https://github.com/DefectDojo/django-DefectDojo/blob/813689833f0d03a57a26f318492ea13f426a64b6/dojo/tools/sslscan/parser.py#L71

Setting port based on protocol was not found anywhere else besides in this PR. Not much generalizing to do now.

valentijnscholten:
Just curious, is there are reason why we want to set the default ports for nessus (and sslscan https), but not for other scanners?

Maffooch:
For Nessus, there are instances were a port is not recorded in the report. So instead of leaving it blank, we can set a default based on protocol. 

Likely the same thought on SSLabs. No protocol in report, so use other data to fill in the blanks. 

Why other parsers are not doing? Maybe their reports are better at including a port/protocol?

damiencarol:
I think `EndPoint.from_uri()` do the job. (fill protocol etc...)
https://github.com/DefectDojo/django-DefectDojo/blob/813689833f0d03a57a26f318492ea13f426a64b6/dojo/models.py#L1414-L1422
So better *not* have specific code in parsers.
Maybe it's worth to check if you should remove the specific code in `sslscan` parser

kiblik:
I suppose also showing endpoints (in UI) works well.
- If you do not set port = nothing special, it means you use default port
- If you set it to default = doesn't need to be printed, it is default
- If you set it as non-default = it is shown (because it is different)
```python
>>> e=Endpoint(protocol='https',host='foo.bar')
>>> str(e)
'https://foo.bar'
>>> e=Endpoint(protocol='https',host='foo.bar',port=443)
>>> str(e)
'https://foo.bar'
>>> e=Endpoint(protocol='https',host='foo.bar',port=8443)
>>> str(e)
'https://foo.bar:8443'
```
Also endpoints without ports and with default port are in DD equal (useful during deduplication)
```python
>>> bool(Endpoint(protocol='https',host='foo.bar') == Endpoint(protocol='https',host='foo.bar',port=443))
True
```
So adding a default port is not beneficial from my perspective.

valentijnscholten:
I was thinking about the usecase where someone might query the endpoints to find all endpoints with port 443. In that case it would only get the endpoints from 1 or 2 parsers and not the rest.

valentijnscholten:
Is there anything in the unit test that could/should be added/asserted to ensure the port is set as expected?

Maffooch:
I would think they would be covered by this test
https://github.com/DefectDojo/django-DefectDojo/blob/813689833f0d03a57a26f318492ea13f426a64b6/dojo/unittests/test_endpoint_model.py#L82

mtesauro (approved):
Approved.

devGregA (approved):

PR #5501, opened 2021-11-26
Set default group for all new users

StefanFl:
Currently you can specify a default group for users in the System Settings, but this will only be set for new users that are created via OAuth2. With this PR the functionality is moved from the OAuth2 pipeline to post save receiver and will be executed for all users. A db migrations is unfortunately needed because I had to change the help texts in the model.


valentijnscholten:
Great idea. Please send my thanks to the person who suggested this via a Slack DM ;-P

valentijnscholten (approved):

Maffooch (approved):

PR #5504, opened 2021-11-26
fix(helm): allow disabling initializer job

qlimenoque:
There is `initializer.run: true` in default **values.yaml**, but it isn't used anywhere.

I'd like to add the opportunity to disable the initializer job.
Also, it would be great to make it possible to add annotations to the initializer job.

UPD: annotations support for init job was added :)

### DIFF
--- a/helm/defectdojo/templates/initializer-job.yaml
+++ b/helm/defectdojo/templates/initializer-job.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.initializer.run  }}
 {{- $fullName := include "defectdojo.fullname" . -}}
 apiVersion: batch/v1
 kind: Job
@@ -86,3 +87,4 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
   backoffLimit: 1
+{{- end }}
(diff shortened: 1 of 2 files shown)


valentijnscholten:
Is there a reason you want to disable it? To save resources? It should run at least each upgrade, but in the future possibly there may be the assumption that it runs every time Defect Dojo is restarted.

qlimenoque:
Well, I would like to add annotation at least if this job is required for correct work. 

Could we add this feature, please? :)

qlimenoque:
It tries to connect to vault with default vault role because I have some vault secrets in configMap, but the process fails and I can’t overwrite the role name with annotations

valentijnscholten:
I'm not against it, just checking if there is any problem with the initializer that needs fixing.

qlimenoque:
Thank you! Initializer works fine. Is this PR good for you or should I remove condition for disabling initializer?

dsever:
You can leave as it is, I see use case to disable init job, when DB needs to be restored after DD deployment.

valentijnscholten (approved):

alles-klar (approved):

Issue #3958, opened 2021-02-28
Re-importing the same report leaves the duplicates in status mitigated


ptrovatelli:
**Bug description**
When we import a report which includes itself duplicates, when re-importing the same report, the duplicates get mitigated.
- first import: 
    - 1 active, verified
    - 1 inactive, duplicate
- after reimport: 
    - 1 active, verified
    - 1 inactive, mitigated, duplicate

It doesn't seem correct to me. I'd expect to have : 
 - 1 active, verified
 - 1 inactive, duplicate (same status as before the reimport as we have just re-imported the same report)

The problem is that when matching the new findings to the existing findings, we always match the new findings to the same single original finding (the one that's not duplicate). Consequently, the duplicates from the original report are flagged as mitigated because no new finding was matched against them. 

i think the issue is here (serializers.py and test/views.py) 
```
                if findings:
                    # existing finding found
                    finding = findings[0]
```
Intead of working on the first finding that matches, we should work on all of them (might need a bit of tuning in order not to re-save the same findings multiples times...)

Sample data: with checkmarx parser attached

**Steps to reproduce**
- Import the report once
- Re-import the same report
**Expected behavior**
The status after the re-import is the same as after the initial import

**Deployment method** *(select with an `X`)*
- [ ] Kubernetes
- [x ] Docker
- [ ] setup.bash / legacy-setup.bash

**Environment information**
 - DefectDojo Commit Message: 
```
$ git show -s --format="[%ci] %h: %s [%d]"
[2021-02-26 17:14:50 +0100] 4f789bc7: fix flake8, fix Q queries, add  unit tests [ (HEAD -> reimport-configurable-dedupe, myOrigin/reimport-configurable-dedupe)]
```

**Sample scan files** (optional)
See attached
[checkmarx_duplicate_in_same_report.zip](https://github.com/DefectDojo/django-DefectDojo/files/6057969/checkmarx_duplicate_in_same_report.zip)

**Screenshots** (optional)

**Console logs** (optional)

**Additional context** (optional)
I've found this while working on https://github.com/DefectDojo/django-DefectDojo/pull/3753 and the provided report will produce the issue after this is merged. The problem was present before PR 3753 but the test data provided might not replicate it.

devGregA:
It looks like this was fixed via https://github.com/DefectDojo/django-DefectDojo/pull/3753

devGregA:
[...]

PR #5506, opened 2021-11-26
Checkmarx parser aggregation and deduplication with query id

jcaillon:
Hello,

The goal of this PR is to fix the aggregation mechanism for the checkmarx parser.

## Details of the issue

At the moment, when we import a checkmarx report, the deduplication mechanism can find duplicates among the findings of this report. This is explained in this issue: #3958.
However it does not make much sense to find duplicates inside a single report because the "Checkmarx Scan" already implements a way to deduplicate similar findings by aggregating them into a single finding (each individual finding description gets merged into the aggregated finding description). This happens because we use different discriminants for the hash code of the dedup algo and disciminants for the findings aggregation algo. 

In this PR the aggregation key has been updated to match the dedup key so there is no possibility to find duplicate inside a single report (findings will be correctly aggregated instead of being created and marked as duplicates). Also added an extra information to the each checkmarx findings which is the query ID used in checkmarx (put it in vuln_id_from_tool field).

So this PR does not exactlty fix the issue #3958 but instead it makes it irrelevant because it can not happen anymore (we can't have duplicate inside a single report).

I have also slightly improved the format of the description for aggregated findings so the end user can better identify each original findings that were aggregated (c.f. screenshots below).

## Reproducing the original issue

I've done the tests with the .xml pushed in the PR.

Import the report:

image

2 findings are marked as duplicate because they have the same hashcode in the dedup algo (cwe/severity/file):

image

Now if we reimport the same report:

image

## Tests with the changes

Import the report:

image

Nothing is marked as duplicated. The aggregation algo is doing its work and aggegates the 2 findings that were previouslt marked as duplicated.

This is how it looks for the CVE id 346

image

For non aggregated results we have the standard description:

image

Now reimporting the report does not cause any issues (notice the reimport counter):

image

FYI this is the same report using the **Checkmarx Scan detail** scan type:

image

with an example of a finding details:

image

All unit tests are running fine

image


valentijnscholten:
Looks like a well thought PR! I don't know enough of Checkmarx to approve, so I'll leave it for the others to review.

Maffooch:
I love finding aggregation. Some people think it messes up the metrics though. Can you add a feature flag similar to the one used on the Qualys WAS parser? 

https://github.com/DefectDojo/django-DefectDojo/blob/1a1cf79d303046e6c21bbc2cb591058b1b875001/dojo/settings/settings.dist.py#L1295

jcaillon:
Actually in the case of defect dojo the aggregation enable/disable is already handled through 2 different scan type:

- Checkmarx Scan will do findings aggregation
- Checkmarx Detailed Scan will not aggregate anything

The changes done will not modify the behavior of the current aggregation because it was already using title as aggregation key which is equivalent to the query ID (but the latter will be more consistant overall).

Maffooch (approved):
Okay great looks good!

jcaillon:
@Maffooch after some discussions with @ptrovatelli I changed this PR a bit: the hashcode fields now remain unchanged (to avoid asking the users to rebuild hashcode for existing checkmarx tests) and instead the aggregation key has been changed to match the dedup key.
The result is the same, i.e. can't have duplicates in a single report. But now the issues that were wrongfully flagged as duplicated are now simply aggregated.
I have updated the PR description to make it more clear.

jcaillon:
PR rebased on dev

damiencarol (approved):
LGTM

Issue #5513, opened 2021-11-28
django-doctor findings


django-doctor:
`Product_Type` is referenced in the file, but not defined anywhere:

https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/db_migrations/0066_django_tagulous.py#L41

valentijnscholten:
Good find. This case is hard to test as I think it's impossible to create products now with `prod_type_id==0`.
Can you try again with the changes in https://github.com/DefectDojo/django-DefectDojo/pull/5514 ?

django-doctor:
good fix :)

FWIW I found this problem while testing new features we're writing at https://django.doctor/ (our test rig runs all our checkers against 666 codebases, your included).


Looking at https://django.doctor/DefectDojo/django-DefectDojo I see also see a few more issues (and some false positives which you can ignore). Example of something that probably wants attention is duplicate url name: https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/notifications/urls.py#L7

I'll leave that in your hands

valentijnscholten:
https://django.doctor/rsumner31/django-DefectDojo

django-doctor:
PRs solve the problem so closing this issue :)

FYI you can prevent similar issues by using Django Doctor on GitHub https://github.com/marketplace/django-doctor/

PR #5514, opened 2021-11-28
Fix missing import/model in 0066_django_tagulous.py

valentijnscholten:
`Product_Type` was not defined as raised in #5513


django-doctor:
:+1:

mtesauro (approved):
Approved

damiencarol (approved):

PR #5515, opened 2021-11-28
fix duplicate notification urls

valentijnscholten:
some non-unique names where used for urls. Only the first one of these 3 was used, so I renamed the bottom 2.

as raised in #5513


django-doctor:
LGTM!

mtesauro (approved):
Approved

damiencarol (approved):

PR #5518, opened 2021-11-29
Auth V2 - Remove legacy authorization part 5: Removal of authorized users

StefanFl:
This is the final PR for the removal of the legacy authorization, it removes the `authorized_users` from Products and Product Types.

Additionally it has a bugfix for a previous step, where I missed one occurence of a legacy staff permission.


docs/content/en/getting_started/upgrading.md
```diff
@@ -73,6 +58,8 @@ Replace the first step above with this one: `docker-compose build`
 
 ## Upgrading to DefectDojo Version 2.5.x.
 
+Legacy authorization has been completely removed with version 2.5.0
```
valentijnscholten, line 61: should we remind users about the migration that should have happened, but can also be run again using the management command?

valentijnscholten (approved):

docs/content/en/getting_started/upgrading.md
StefanFl: With this PR they won't have the chance anymore to use the management command to migrate the users. They would have to do it before they update to release 2.5.0. Is this what you want to say? Would definitely make sense.

valentijnscholten: Just had the feeling it was good to bring it up here, but didn't yet realize it would have to run before the upgrade. Should be mentioned as (possible) breaking change then I guess.

StefanFl: Have added a remark to the release notes.

Maffooch (approved):

PR #5527, opened 2021-11-29
move unittests outside dojo folder

valentijnscholten:
The unittests are inside the dojo folder, which I think is against recommendation to keep tests in a separate folder, outside of the app sourcecode.

Also, the current `django` docker image that we build is `939MB` large. `115MB` of this is occupied by `unittests`. So by moving the unittests outside the `dojo` folder we also save on image size (an build time and GitHub Actions build times).

Also added `--no-install-recommends` where it was not yet used for `apt-get install` (but this didn't work on all occassions).

Also extracted a `get_unit_tests_path` method so if we ever move the unittests again, we only have to update one place maximum.

I used `git mv` to move the unit tests which should help with any PRs affecting unit tests that are now moved by this PR.


mtesauro:
Love this idea.  Waiting for the tests to complete...

valentijnscholten:
@mtesauro it's green now

mtesauro (approved):
Phew, quite a few changes.  Nice job.

Approved.

StefanFl (approved):
Great job, thanks @valentijnscholten 

I approve even with conflicts in the PR. When you have resolved this conflicts, you want to be fast merging it, before the next changes come for unit tests.

valentijnscholten:
conflicts resolved, tests green.

PR #5533, opened 2021-12-01
add logic if external redis is used

sandroded:
hi guys!
I've discovered the issue, that secret name for message broker is hardcoded in the helm chart when redis is used. I've added the logic to fix it. 
I fixed the few typos in the values.yaml: missing port in the mysql.service and name for the django debug value, and name of the key in the rabbimq secret.


valentijnscholten:
Thanks. Can you base your PR against `dev`? There have been some chart changes recently there.

helm/defectdojo/values.yaml
```diff
@@ -266,6 +266,7 @@ mysql:
   secretKey: mysql-password
   mysqlDatabase: defectdojo
   service:
+    port: 3306
```
dsever, line 269: From next release, this port is not in use anymore. @alles-klar can give you more information

sandroded: Shall it be removed? Because it is referenced it all deploymets at the moment as .Values.mysql.service.port

alles-klar: after the rebase you will find the value (mysqlmaster.service.port), see https://github.com/DefectDojo/django-DefectDojo/pull/5450

alles-klar:
Why is it a problem to have the secret name hardcoded?

helm/defectdojo/values.yaml
valentijnscholten: if the rebase is messy you can try to create a new PR  from a new branch if that's easier.

sandroded: I'll close this one, because I created a new one: https://github.com/DefectDojo/django-DefectDojo/pull/5534

PR #5543, opened 2021-12-03
Move unit test leftovers

StefanFl:
Moving the unit tests was very close to two other PR's. Because of that we had some leftovers in `dojo/unit_tests`, that are moved now with this PR.


valentijnscholten:
Thanks, missed those when merging dev into the unit tests moving PR I think. I am going to merge them straightaway to reduce the chance of any conflicts with other PRs.

PR #5544, opened 2021-12-03
Add organizational blocks around metrics.js files

Maffooch:
Add some blocks around the metrics page to organize them and easily swap the file out for custom colors

Here are some docs on template inheritance: https://docs.djangoproject.com/en/3.2/ref/templates/language/#template-inheritance

### DIFF
--- a/dojo/templates/dojo/dashboard-metrics.html
+++ b/dojo/templates/dojo/dashboard-metrics.html
@@ -196,7 +196,9 @@
         <script src="{% static "JUMFlot/javascripts/jquery.flot.mouse.js" %}"></script>
         <script src="{% static "JUMFlot/javascripts/jquery.flot.bubbles.js" %}"></script>
     {% endif %}
-    <script src="{% static "dojo/js/metrics.js" %}"></script>
+    {% block metrics %}
+        <script src="{% static "dojo/js/metrics.js" %}"></script>
+    {% endblock metrics %}
     <script>
         $(function () {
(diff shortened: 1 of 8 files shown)


valentijnscholten:
Is there a rule / guideline on when a block needs ` {{block.super}}` and when it doesn't? A quick glance seems to show most blocks have it currently?

Maffooch:
if `{{ block.super }}` is present
    The child block  will **append** the parent block (same block name in the template that is being extended)  

if `{{ block.super }}` is **not** present
    The child block will **overwrite** the parent block

I did not add them to these new `metrics` blocks because they are the "parent" block in this case. The `{{ block.super }}` tag can be added here, but it would not make a difference since there is not another `metrics` block that is being extended.

StefanFl (approved):
I don't really understand the blocks and what this changes, but it seems not to do any harm 😃

Maffooch:
Template blocks are crucial for the success of plugins in dojo! They can get pretty complex sometimes, so I suggest reading the docs I linked above

valentijnscholten:
Blocks in general are good to have for people who want to extend our templates and append/overwrite stuff. I would expect the `block.super` to only be needed in subtemplates and not in our templates as we are not really appending or overwriting anything. On the other hand if someone decides to add something to the `content` block in `base.html` that needs to be present in all subtemplates, it will work now because of all those `block.super`-s.

valentijnscholten (approved):

PR #5547, opened 2021-12-04
Respect `scan_date` at import time for all findings imported

Maffooch:
I noticed that the `scan_date` field in the importer is not respected when setting the date of a finding.
This PR sets the date of all findings imported to the date specified at import time, but only if the user changes it from the default. Default behavior of setting the finding date to "today" is maintained.

For example, here are some scenarios of how this PR affects the date of a finding:

1. The parser does **not** set the date, and scan date is **not** specified at import
    Finding date will be the default value of scan date (time of import)
2. The parser  **sets**  the date, and the scan date is **not** specified at import
    Finding date will be whatever the parser sets
3. The parser does **not** set the date, and the scan date **is** specified at import
    Finding date will be whatever the user specified for scan date
4. The parser **sets** the date, and the scan date **is** specified at import
    Finding date will be whatever the user specified for scan date

The bottom line is whatever the user specifies for the scan date is respected for all findings imported. if the scan date is left at default, or not specified at all, the parser will need to set the date, or the date of the finding will be set to the default (import time)

image


dojo/importers/importer/importer.py
```diff
@@ -91,6 +91,10 @@ def process_parsed_findings(self, test, parsed_findings, scan_type, user, active
                 item.active = active
             if item.verified:
                 item.verified = verified
+            # Only set date value of the test if they were NOT set by the parser
+            # scan_date is set to both target_start and target_end in the creat_test function
+            if item.date:
```
StefanFl, line 96: Unfortunately this doesn't work:
- When the parser sets a `date` explicitly, `item.date` is true and it will be overwritten with `test.target_start`. This is not what we want.
- When the parser doesn't set a `date`, the model sets `current_date` as a default

So basically the `date` is set here, but you don't know why. I thought about a logic comparing the dates, but couldn't find one that works, as long as there is the default in the model.

StefanFl (changes requested):

dojo/importers/importer/importer.py
Maffooch: I did not think the date would actually be present until the finding was saved (and the default date would be applied)

Do you think it would be good to set to
```
if **not** item.date:
    item.date = test.target_date
```

StefanFl: I tried that, it doesn't work as well, because the model has set the date to current date as a default already. It would work when we remove the default, but then we might get other side effects.

Maffooch: I've got a quick idea that is a bit more fleshed out. Will commit it and ping you

StefanFl: I am afraid we need more test cases:
- Parser sets date explicitly
- Import sets date different from today

Maffooch:
@StefanFl unit tests will fail here (the one I made), but think the logic is sound now and ready for some more test coverage

dojo/importers/importer/importer.py
StefanFl: The first branch works in all but one cases: When the date for the import is set to something different than today but the parser has set the date explicitely to today. Then the parser date would be overriden. But I think this shouldn't happen in reality and we can ignore that.

I am honestly not sure about the else branch, do we want to override the date the parser has set, when the user set a date for the import? When I think about it while writing, I think it makes sense. Would like to hear the opinion of other maintainers.

Do we need to the same for re-imports?

Maffooch: The import `scan_date` should overwrite the date the parser recorded. It is the user taking it into their hands to change it, so the findings should record as such.

I am not sure about re-import as I feel it's primary use is for automated imports as well as keeping an up to date view of what's vulnerable.

The main motivation for this PR is because when importing older reports, the date of the test was often far in the future than the findings themselves. I do no think many would be re-importing older scan reports.

StefanFl: I am really open about it. @valentijnscholten and @damiencarol, you know import and reimport the best, what do you think?

valentijnscholten: I think the idea of this PR is good, but we need some test cases (haven't checked if there are any). We want to make sure it fully works as expected. Otherwise it gets confusing, for example what we saw with the `active/verifed` being overwritten or not by the parsers or importers. I think during reimport the date of the finding was already set during the initial import, so probably no need to add logic there.

I haven't checked the code, but could it work to remove the default from the default and only set a default in `save()` as safeguard? That way the `if not item.date` could work.

Maffooch: There is a very basic test (just showing that the date at import time does overwrite the default date), but the scan that is used in tests (zap) does not set the date on the report level,  so I would need to test with another scanner (I was using Acunetix from the unit tests when I wrote this)

Sounds like leaving re-import alone is the move though.

StefanFl (approved):

mtesauro (approved):
Approved

dojo/importers/importer/importer.py
```diff
@@ -96,6 +96,24 @@ def process_parsed_findings(self, test, parsed_findings, scan_type, user, active
...
+                if target_start_date != now_date:
+                    # Import scan_date was set, and the parser has not overwritten it
+                    logger.debug('Parser did not set date, import override it ' + str(target_start_date) + ' : ' + str(now_date))
+                    item.date = test.target_start
+            else:
+                # The parser has set the date already
+                if target_start_date != now_date:
+                    # The date set by import scan_date should overwrite scan report date
+                    logger.debug('Parser set date, import override it ' + str(target_start_date) + ' : ' + str(now_date))
+                    item.date = test.target_start
```
valentijnscholten, line 116: the if and else parts seem to do the identical thing, except logging a different message?

The PR description says:

_This PR sets the date of all findings imported to the date set at import time, but only if the parser fails to do so._

But the code seems to overwrite the date set by the parser in all cases? Looks like the unit tests check for that as well?

Not sure about how others use the scan_date, but I would expect that a date set by the parser will always be respected and not overwritten. So the scan_date is only used if the parser doesn't set a date, like the PR description mentions.

unittests/dojo_test_case.py
```diff
@@ -469,13 +469,17 @@ def import_scan_with_params(self, filename, scan_type='ZAP Scan', engagement=1,
         if test_title is not None:
             payload['test_title'] = test_title
 
+        if scan_date is not None:
+            payload['scan_date'] = scan_date
+        else:
+            payload['scan_date'] = str(datetime.date.today())
+
```
valentijnscholten, line 476: I think we should also test for the scenario where scan_date is None, what used to be the default in the test cases. Not sure how anybody else uses the scan_date, but I always leave it as None because it will get defaulted to today by the serializer. If we make the change proposed here, we have defaults set in two places, if one of them changes it might lead to confusing results or missed scenario's in the tests.

valentijnscholten:
_The import scan_date should overwrite the date the parser recorded. It is the user taking it into their hands to change it, so the findings should record as such._ 

I am not sure I agree. The field is mandatory in the UI and mandatory (defaulted to today) in API. So if it's always used to overwrite any date set by the parser, the user will have no way to actually use the date set by the parser. When importing a scan the user (or the cicd script) doesn't always know when the scan was performed, so it can't always set the scan_date correctly. So it would be better to have the parser set it. 

Either way, the PR description differs from the PR implementation at the moment.

valentijnscholten (changes requested):
Don't want nag, but I really think we should give this some second thought. For example reports from AWS or other cloud security scans usually contain findings with different "first found" dates. This PR now overwrites all of them with the date chosen by the user or the api script (or today if no date is provided in the API).
I don't think that is what we want, or am I misreading the code?

I think we have similar case to the active/verified handling of overwriting. The user has no choice between "overwrite always" or "use this as a default when the parser can't figure it out".

dojo/importers/importer/importer.py
Maffooch: I will change the PR description to be more descriptive, but no, the date is not overwritten in **all** cases. This is outlined in the unit tests created in this PR.

The motivation for this change is due to a scan being completed on given date (say 01/01/2005) and then imported to dojo at a later date. Regardless of setting the scan_date variable, the date of the findings will not change from those imported in the scan report. If the scan report does not specify the date, the date of the finding will always be the default date (the time of import) regardless of what value is specified in the scan_date. 

The default date given to a test (and all findings) is the date of import. This is true in a number of cases
 - the scan_date field is not supplied at all (API)
 - The scan_date is auto populated, and the user does not change it (UI)

The only way for a finding to have a date other than the date contained in the scan report (if the scanner even supplies it) or the default date (generated at import time)  is to manually set the date for every single finding imported.

unittests/dojo_test_case.py
Maffooch: I cannot think of a possible scenario when the scan date could be None. It will always have a value because there are defaults set. The default in the test cases was actually the date 2020-06-04. I can set it to actually be None (so it relies on the default) in the test case, but I think having a default in two places (not in general, but specifically one in code, and one in test) is a good thing. If someone tries to change the default in the code, the unit tests will fail. This a good thing as it will be easier to catch someone making big changes. 

Your use of the scan_date is the common use case. Leaving it as None (not specifying it at all) will **not** affect the date of the findings at all whether they were set by the parser, or set by the default value. This is reflected in the code as well.

Maffooch:
I will update the PR description to better describe the implementation

> So if it's always used to overwrite any date set by the parser, the user will have no way to actually use the date set by the parser.

This is not true. Please see the unit tests that outline all possible use cases of where a findings date can come from. 

> When importing a scan the user (or the cicd script) doesn't always know when the scan was performed, so it can't always set the scan_date correctly. So it would be better to have the parser set it.

Since the scan_date field is not mandatory when using the API, a CI/CD script can simply not set the scan_date (and it will default to the time of import), and the finding date will either come from the scan report, or the default value.

> For example reports from AWS or other cloud security scans usually contain findings with different "first found" dates. This PR now overwrites all of them with the date chosen by the user or the api script (or today if no date is provided in the API).

Half of this correct. The dates will be overwritten by the user if they specify the scan_date. Since it is not mandatory, users can not use it. The default generated date will not overwrite a date that was taken from the scan report (see the two unit tests on this)

> I don't think that is what we want, or am I misreading the code?

You are misreading the code. Please deploy this and test by hand. I found that the Acunetix scans used in the unit tests have a date from the report, and the ZAP scans used in the unit tests do not specify a date in the report.

unittests/dojo_test_case.py
valentijnscholten: Still I think it would be better to have the until test reflect what most users are doing, not sending a scan_date. IT won't change what happens today, the tests will still pass.

valentijnscholten:
Ok, I don't have the time to analyze all the possible code paths here so have to assume it does what it does. I am wondering though what happens when an import starts at 5minutes before midnight and runs 10 or 15 minutes? And what happens when a DD instance is not in UTC? Couple of years ago I had to change most instance to `.localtime()` and `.localdate()` to avoid "off by one" date errors.

dojo/importers/importer/importer.py
damiencarol: > For example, here are some scenarios of how this PR affects the date of a finding:
>
>    * The parser does not set the date, and scan date is not specified at import
>    Finding date will be the default value of scan date (time of import)
>    * The parser sets the date, and the scan date is not specified at import
>    Finding date will be whatever the parser sets
>    * The parser does not set the date, and the scan date is specified at import
>    Finding date will be whatever the user specified for scan date
>    * The parser sets the date, and the scan date is specified at import
>    Finding date will be whatever the user specified for scan date

I fully agree.

Specifically this comment:
> The import scan_date should overwrite the date the parser recorded. It is the user taking it into their hands to change it, so the findings should record as such.

It's why I think this PR is good. More control to the user and a way to override the data of the parser *OR* let the parser do the job and use the data as-is.

Also:
> I think the idea of this PR is good, but we need some test cases (haven't checked if there are any). We want to make sure it fully works as expected. Otherwise it gets confusing, for example what we saw with the active/verifed being overwritten or not by the parsers or importers. I think during reimport the date of the finding was already set during the initial import, so probably no need to add logic there.

I think @valentijnscholten is right. To make this PR works we need to be good on the unit tests part. We must have the 4 cases covered.

> I haven't checked the code, but could it work to remove the default from the field and only set a default in save() as safeguard? That way the if not item.date could work.

Same here, I haven't checked the code but I expect that the `save()` logic is changed to allow the logic described in the PR description.

Also, @Maffooch it could be good to make it clear in the documentation how we handle the date. Could you add a part in the documentation that clearly state the 1/2/3/4 cases? Thanks.

Maffooch:
Thanks for the comments @damiencarol @valentijnscholten 

I'll add the uses cases to documentation. 

The defaults are set in [forms.py](https://github.com/DefectDojo/django-DefectDojo/blob/1a1cf79d303046e6c21bbc2cb591058b1b875001/dojo/forms.py#L408-L410) and [API serializers](https://github.com/DefectDojo/django-DefectDojo/blob/1a1cf79d303046e6c21bbc2cb591058b1b875001/dojo/api_v2/serializers.py#L1176) and neither one accommodates timezones. It would be good to change those (and all instances of time related activities) to support timezones. But for now, I will change just these two.

I'm not sure how I would check for timezone accommodation in the unit tests though.

valentijnscholten:
Testing for timezones in unit tests would be a big job, as we're not doing that anywhere yet. I have an idea on how to implement the requirements without relying on comparisons to "today" in these places. But don't have the time to do it or show it, so I'll just leave it up to the rest.

Maffooch:
> Looking at the code if the import is started at 23:55 and runs for 10 or 15 minutes you will get some findings that have their date overwritten, and some keep their date.

Hmm that is a tough one. Struggling to think of way to overcome that. Wouldn't that happen without this code change as well though?

> Testing for timezones in unit tests would be a big job, as we're not doing that anywhere yet. I have an idea on how to implement the requirements without relying on comparisons to "today" in these places. But don't have the time to do it or show it, so I'll just leave it up to the rest to review/approve.

Agreed, the timezone test would be very time consuming. I think simulating various timezones via hardcode and checking count the count of findings at various dates and validating dates would do it. 

I have updated the default date in UI/API and in tests to accommodate timezones. Also added documentation

valentijnscholten:
> > Looking at the code if the import is started at 23:55 and runs for 10 or 15 minutes you will get some findings that have their date overwritten, and some keep their date.
> 
> Hmm that is a tough one. Struggling to think of way to overcome that. Wouldn't that happen without this code change as well though?

Before this PR I think the date field of findings is never overwritten by scan_date? You could fix it by getting `now()` in the importer and pass it on to process parsed findings so it uses the same "now" for all findings. Not sure if that makes the code better/nicer. Also not saying it should prevent merging as maybe most would consider it a corner case, but from a product owner point of view I personally would like to have that scenario working 100% correctly at some point :-)

Maffooch:
Oh that is actually what is already happening (see now=now)
https://github.com/DefectDojo/django-DefectDojo/blob/1a1cf79d303046e6c21bbc2cb591058b1b875001/dojo/importers/importer/importer.py#L352-L356

I did not notice this until now, but I think it solves that corner case

valentijnscholten:
> Oh that is actually what is already happening (see now=now)
> 
> https://github.com/DefectDojo/django-DefectDojo/blob/1a1cf79d303046e6c21bbc2cb591058b1b875001/dojo/importers/importer/importer.py#L352-L356
> 
> I did not notice this until now, but I think it solves that corner case

Ah crap, I didn't realize it was already there as it's not shown in the diff. But SOMEBODY (*me) already added it earlier it seems. The prophecy fulfilled or how do they say that in English.

devGregA:
For reference: https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/forms.py#L376

I think the UI should be consistent with the API. I agree with the change, but I think the field should not be required in the UI, and should not be pre-populated.

docs/content/en/integrations/importing.md
```diff
@@ -118,4 +118,23 @@ A classic way of reimporting a scan is by specifying the ID of the test instead:
...
+DefectDojo offers a plethora of supported scanner reports, but not all of them contain the 
+information most important to a user. The `scan_date` field is a flexible smart feature that
+allows users to set the completion date of the a given scan report, and have it propagate
+down to all the findings imported. This field is **not** mandatory, but the default value for 
+this field is the date of import (whenever the request is processed and a successful response is returned). 
+
+Here are the following use cases for using this field:
+
+1. The report **does not** set the date, and `scan_date` is **not** set at import
+    - Finding date will be the default value of `scan_date`
```
StefanFl, line 134: What is the "default value of scan_date" when it's not set? If it's the current date, then it should say "Finding date will be current date"

docs/content/en/integrations/importing.md
```diff
@@ -118,4 +118,23 @@ A classic way of reimporting a scan is by specifying the ID of the test instead:
     "scan_type": 'ZAP Scan',
     "test": 123,
 }
-```
\ No newline at end of file
+```
+
+## Using the `scan_date` field
```
StefanFl, line 123: Is this only about `scan_date` in the API or about **Scan Completion Date** in the UI as well?

dojo/api_v2/serializers.py
```diff
@@ -1202,7 +1202,7 @@ def get_findings_list(self, obj) -> List[int]:
 
 
 class ImportScanSerializer(serializers.Serializer):
-    scan_date = serializers.DateField(default=datetime.date.today)
+    scan_date = serializers.DateField()
```
StefanFl, line 1205: The `scan_date` is still a mandatory field. This doesn't fit to the documentation. If it should be optional, I think you need to set `required=False`, see https://www.django-rest-framework.org/api-guide/fields/#required

StefanFl (changes requested):
Plus see my comments about the documentation as well

docs/content/en/integrations/importing.md
Maffooch: This is in the lat sentence of the first paragraph

StefanFl (approved):

unittests/dojo_test_case.py
```diff
@@ -518,6 +521,11 @@ def reimport_scan_with_params(self, test_id, filename, scan_type='ZAP Scan', eng
         if test_title is not None:
             payload['test_title'] = test_title
 
+        if scan_date is not None:
+            payload['scan_date'] = scan_date
+        else:
+            payload['scan_date'] = str(timezone.localtime(timezone.now()).date())
```
valentijnscholten, line 527: I think this should now be None as there no default anymore

Maffooch: Yes good catch

Issue #5516, opened 2021-11-28
Authentication via OAuth2/SAML2


hkelley:
Okta setup notes that might improve the documentation:

- SOCIAL_AUTH_REDIRECT_IS_HTTPS = True was mandatory for me.  When I didn't use it,  a ":443" was appended to back of the parameters when returning the state code, and thus confusing DefectDojo.   I don't know if this was Okta or my nginx config, but simpler to use HTTPS and avoid any extra redirects.
- Sign-in redirect URI in Okta app setup should be the user's URL to the DefectDojo app,  not http://localhost:8000.   
- In some Okta tenants,  the trailing "/default" in `` DD_SOCIAL_AUTH_OKTA_OAUTH2_API_URL`` is invalid.   Try it without that if you get a 400 error about the endpoint/resource not being available on the server.   I think this a difference between "corporate" and "consumer" Okta tenants, but I forget the licensing specifics.

valentijnscholten:
Thanks, could you do a PR with proposed changes? (see `/docs` folder)

hkelley:
PR #5550

PR #5550, opened 2021-12-04
Okta doc edits

hkelley:
As discussed in #5516

### DIFF
--- a/docs/content/en/integrations/social-authentication.md
+++ b/docs/content/en/integrations/social-authentication.md
@@ -96,7 +96,7 @@ OKTA
 ----
 
 Okta can bet set up in a similar fashion to that of Google, using Okta as an OAuth2 provider. Upon login to 
-DefectDojo with an Okta identity token, a new DefectDojos user will be created if one does not already exist. 
+DefectDojo with an Okta identity token, a new DefectDojo user will be created if one does not already exist. 
 Okta OAuth2 identities are matched to DefectDojo users by email address, not by username. Follow along below.  
 These images are taken from an Okta Developer Edition tenant so options for Custom Authorization Servers are shown.  
 The corresponding steps for Okta Workforce tenants are mentioned in the text below and the overall procedure is the same.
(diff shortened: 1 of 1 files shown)


docs/content/en/integrations/social-authentication.md
```diff
@@ -140,12 +143,16 @@ carries the same attributes and a similar procedure. Follow along below.
...
+If you are using an Okta Workforce (not Developer) tenant then use only "/oauth2" for the API URL,  do not include the "default" 
+because your tenant has only one authorization server.
+
+For production deployments where you are using HTTPS to access DefectDojo, also add/set the following.  Note that this is not a "DD_*" 
+environment variable so it is further down in the the settings file.
+
+    {{< highlight python >}}
+    SOCIAL_AUTH_REDIRECT_IS_HTTPS = True
+    {{< /highlight >}}
+
```
valentijnscholten, line 155: I suggest to add this as a `DD_` variable and set it default to `False` as the defautl docker-compose deployment has no https. Is that something you could add to the PR?

valentijnscholten:
Thanks for the PR, made one little comment.

docs/content/en/integrations/social-authentication.md
hkelley: I'm happy to take a crack at that.  I searched through this repo and I only see `SOCIAL_AUTH_REDIRECT_IS_HTTPS` mentioned in this documentation.   Would the change you propose be to [godojo](https://github.com/DefectDojo/godojo) within env.go ?   I believe it is what generates `settings.dist.py`.

I see where the `DD_SOCIAL_AUTH_REDIRECT_IS_HTTPS`  would be added to the template but haven't yet found the  generation of the remainder of the file.

StefanFl (changes requested):
This branch shall be merged into `dev` and not `master`.

docs/content/en/integrations/social-authentication.md
Maffooch: Hi @hkelley 

`SOCIAL_AUTH_REDIRECT_IS_HTTPS` is from the [social-auth]( https://python-social-auth.readthedocs.io/en/latest/configuration/settings.html#processing-redirects-and-urlopen) package we use for SSO. 

The change would need to go into `settings.dist.py`.

PR #5551, opened 2021-12-05
Fix staff permission to add product types

StefanFl:
#5410 gave non-staff users the permission to add product types when they have a global role. But it removed the permission to add product types for staff users regardless if they have a global role, which was not intended.


valentijnscholten:
Maybe I thought `AUTHORIZATION_STAFF_OVERRIDE` should be `True` to allow this, can't remember.

valentijnscholten (approved):

Maffooch (approved):

PR #5553, opened 2021-12-06
Add asynchronous re/imports (disabled by default)

Maffooch:
I was trying to upload a Nessus scan with 15k lines and had a few findings that had over 5k endpoints attached. Took over 20 minutes to import. This was super annoying, so I made imports async.

What this does is batch imports and findings into celeryworkers based on a configurable bucket size (default is 100)
so that processing endpoints/statuses can continue to happen totally after all findings have been imported. 

Findings are imported in a semi-async fashion. The semi is because a successful import waits for all workers to finish before returning a successful response. This still reduces wait times significantly as finding "buckets" can be process simultaneously

I was not sure how to handle unit tests, but I have tested for race conditions, data loss, and celery worker overload extensively.

**This feature is disabled by default**


dsever:
@Maffooch https://owasp.slack.com/archives/C014H3ZV9U6/p1637871478056900

Maffooch:
@kiblik I apologize for not seeing the slack discussion sooner. Does this implementation fit the use case you were after?

mtesauro (approved):
Impressive work.  Thanks for making it able to be set in setting.py 

Approved.

dojo/importers/importer/importer.py
```diff
@@ -356,10 +309,34 @@ def import_scan(self, scan, scan_type, engagement, lead, environment, active, ve
...
+            # After all tasks have been started, time to pull the results
+            logger.info('IMPORT_SCAN: Collecting Findings')
+            for results in results_list:
+                serial_new_findings = results.get()
+                new_findings += [next(serializers.deserialize("json", finding)).object for finding in serial_new_findings]
+            logger.info('IMPORT_SCAN: All Findings Collected')
+            # Indicate that the test is not complete yet as endpoints will still be rolling in.
+            test.percent_complete = 50
+            test.save()
+            importer_utils.update_test_progess(testß)
```
kiblik, line 334: `ß` looks like a typo

kiblik (changes requested):

kiblik:
@Maffooch, I'm happy to see any improvement in speed so not be sorry :)
However, the answer is no. 

Your scenario: You are able to process the content of a file (by the parser) quickly and you have a problem with slow write speed into the database.
Our scenario: We are questing `SonarQube API Import`. We have a problem with the "long-run" of `parsed_findings = parser.get_findings(scan, test)` because is it performing so many API calls.

Note: I preferred to force `async` by adding API (or GUI) parameter. You decided to set it globally for the whole DD. Hard to say what is better. It can be different case-by-case.

dojo/importers/importer/importer.py
Maffooch: I'm not sure how to even type that key... Thanks for catching that! Fixed.

Maffooch:
Interesting. When I first ran into this issue with my massive scan file, I thought parsing was the bottleneck, but I eventually found that the actual file parsing only took 2-4 seconds consistently. 

Your use case would be specific to API importers. I would suggest making _just_ the API parsers run async rather than adding a flag for async to dojo's API/UI.

I found that making smaller processes async actually slowed them down, so making the the parsing element of all imports async (in all cases besides the API importers) would actually hurt performance across the board

dojo/importers/utils.py
```diff
@@ -82,3 +91,72 @@ def construct_imported_message(scan_type, finding_count=0, new_finding_count=0,
...
+            finding.endpoint_status.add(eps)
+            finding.endpoints.add(ep)
+    logger.debug('IMPORT_SCAN: ' + str(len(endpoints)) + ' imported')
+
+
+# This function is added to the async queue at the end of all finding import tasks
+# and after endpoint task, so this should only run after all the other ones are done
+@dojo_async_task
+@app.task()
+def update_test_progess(test):
```
dsever, line 160: Small typo

Maffooch: Where is the typo? Is it the same one @kiblik pointed out?

dsever: line 160 -> `update_test_progess`

Maffooch: Omg I could not see that. Thanks!

devGregA (approved):

dsever:
@Maffooch 

I did test with different number of celery-workers, i would say sweet spot is 3, but now parsing takes the most time inside of  import. 

```
Celery number, time in (ms):

==1 worker==
207280
229792
187019

==3 workers==

150036
169638
169024

==6 workers==

140174
135639
112309

==12==

155014
112248
131125
```

valentijnscholten:
This PR was merged before I had a chance to review it. I like that the imports are being made faster and that it's behind a feature flag. But there are some concerns, some of which prevented me from making the imports async myself.

**Bugs**

*finding groups*
When findings are grouped during import or reimported, they are pushed to JIRA is push_to_jira is True. This PR executed this code multiple times (for each chunk), so finding groups can be pushed to JIRA multiple times. Could be each time with different content, or overwriting eachothers content. Sometimes multiple JIRA issues will be created. Hard to see from the code what exactly will happen.
The root cause is that the whole process_parsed_findings is executed multiple times. Might be better to only chunk the for loop inside process_parsed_findings, which will also help with the next item.

*duplicate clusters*
In the past I have spent considerable amounts of time to make sure findings were imported in sequential order. This makes sure that the result of an import is always the same.
Any duplicates inside the same report are always imported in the same order, making sure the oldest/first finding is the original and that there are no "circular duplicates" where A->B->A.
These cirular duplicates can cause problem with deleting findings or the tests/engagements they are part of.
This async way of importing findings in parallel may violate this depending on the size of the report, performance, etc.
Maybe we should lower the requirements for ourselves and declare that "duplicates inside the same report" are not supported and can lead to unexpected results. This should be mentioned in the docs (or we should fix it).


**Concerns**

*test coverage*
Currently there are no tests running with DD_ASYNC_FINDING_IMPORT set to True. This means the chunking of findings and async handling of endpoint status is not tested for these codepaths.
We have lots of tests to check that imports work as expected, reimports work as expected, findings get created/closed/reactivated as expected, etc, etc.
I think the async codepaths should be covered as well. Doing things async can lead to unexpected results, race conditions, etc.
If we can't do this short term, we should manage expectations by marking the feature as "in development" or "preview" or "beta" in the docs.

*return type variation*
The process_parsed_findings method used to return a tuple with lists of findings. After this PR the return type depends on the value of DD_ASYNC_FINDING_IMPORT (or sync). If that is True, the return type changes into a tuple of serialized json objects.
This means every method calling process_parsed_findings has to be aware of this and possibly check the value of DD_ASYNC_FINDING_IMPORT or sync. I suggest to restore the old process_parsed_findings method and make a wrapper method process_parsed_findings_task that does all the things related to it being a celery task. But it might even be better to do the chunkinging inside process_parsed_findings as described in the previous bullet.

*not awaiting completion of endpoint status updates*
The endpoint status updates are performed asynchronously and the result is not awaited. The import can complete before all endpoint statuses are updated. Depending on the circumstances this could lead to unexpceted results, or at least to stale information in the UI. Not sure of the extend/severity of this. I had a plan for a PR to return statistics of each import in the response. If not all updates are completed before the import completes, we cannot return correct statistics.

*tagulous errors*
tagulous sometimes raises an exception when an model or queryset is passed as a parameter to a celery task. we see this sometimes with notifications, which uses a similar pattern to the async import (passing queryset of findings to celery). so there's a chance this problem occurs at some point. I haven't found the root cause, so not sure if it will, but I haven't been able to solve it for notifications.  The error is reported in https://github.com/radiac/django-tagulous/issues/142, but we have seen users on Slack having this and I have seen it myself from time to time. This could lead to half imported reports.


**Suggestions**

*bulk update*
For bulk updates to models we have the mass_model_updater method which is very efficient in updating multiple models in on go (or batches). It's probably more efficient on cpu/memorty/broker to use this, instead of lots of small queries.

*close old findings*
Close old findings also updated endpoints, so could benefit from the same improvements. Currently these are not chunked.


Considering these concerns, and current "lack" of test coverage, I suggest to mention the concerns in the docs and mark the async feature as "in preview" or "under development" or "experimental" or ...

PR #5559, opened 2021-12-06
add support for more than one hostname in defectdojo

nobletrout:
This is a simple hack to deal with a real annoyance I had with leveraging DD's API for submitting within k8s and having a different public/gated web frontend.

It won't help those who already use this helm deployment model. But it seems easy enough to adjust to it.


alles-klar (approved):

alles-klar:
@nobletrout great idea. Sorry for the late review.

dsever:
Guys, it has to be implemented for ingress too.

nobletrout:
> Guys, it has to be implemented for ingress too, or will not fly

Sorry, I wasn't using the default ingress so missed that. Happy to hear you guys find this useful.

Maffooch:
No clue on helm stuff. Please "at" me when you guys think it is ready

PR #5573, opened 2021-12-08
feat(helm-chart): celery worker app_settings

qlimenoque:
This PR allows setting custom settings for celery worker.

Values didn't work because there were hardcoded values in configMap template.


alles-klar:
Thanks for the PR. Can you add this values also to values.yaml please, so people see that they can change them?

qlimenoque:
They are in the **values.yaml** already in master branch :)
This PR makes possible to map that values to configMap

alles-klar (approved):

StefanFl (approved):

PR #5574, opened 2021-12-08
add scan_date fix also for reimport, fix validation

### LINKED ISSUE #5576: reimport with files is broken
No unit test coverage for reimport it seems.

image

### LINKED ISSUE #5584: api: future date accepted as scan_date
In the UI you cannot upload a scan report and select a `scan_date` in the future.
In the API this is possible. There is code in the API to check and forbid a future `scan_date` but it is not executed due to a typo (and no longer work with the current implementation of `scan_date` after the recent changes)

https://github.com/DefectDojo/django-DefectDojo/blob/ba1a415219ff20e8b4e909ef14f750de9b80297e/dojo/api_v2/serializers.py#L1484-L1488

valentijnscholten:
Following up on #5547 this PR adds the same fix for reimport (which can call import when auto creating objects):
- add `scan_date` logic to reimport that does an initial import
- add `scan_date` logic to reimport for new findings created during reimport
- add test cases for reimport
- make scan_date optional on reimport form
- fix #5584 

While working on it, I also realized some cleanups/simplifications:
- simplify `scan_date` handling, mainly to remove confusing comparisons with "today"
- remove `item.updated/created = now` which has no effect due to `auto_now`
- when `scan_date` is provided, do not add current time of today to avoid confusion
- update test case descriptions
- remove duplicate/unused update_timestamps method
- let api send `datetime` instead of `date` for `scan_date` to align with UI logic
- remove superfluous scan_date params from other (unit)tests.
- fix #5576


dojo/api_v2/serializers.py
```diff
@@ -1327,16 +1327,16 @@ def validate(self, data):
 
     def validate_scan_data(self, value):
         # scan_date is no longer deafulted to "today" at import time, so set it here if necessary
-        if not value.date:
-            return None
+        # if not value.date:
+        #     return None
```
damiencarol, line 1331: I'm not a big fan of commented code

valentijnscholten: Agreed, forgot to remove it

valentijnscholten: Turns out this code is never executed: #5584

dojo/api_v2/serializers.py
```diff
@@ -1280,13 +1279,15 @@ def save(self, push_to_jira=False):
         _, test_title, scan_type, engagement_id, engagement_name, product_name, product_type_name, auto_create_context = get_import_meta_data_from_dict(data)
         engagement = get_or_create_engagement(engagement_id, engagement_name, product_name, product_type_name, auto_create_context)
 
+        # have to make the scan_date_time timezon aware otherwise uploads via the API would fail (but unit tests for api upload would pass...)
```
StefanFl, line 1282: Typo in timezone (missing e)

dojo/api_v2/serializers.py
```diff
@@ -1416,14 +1414,16 @@ def save(self, push_to_jira=False):
         engagement = get_target_engagement_if_exists(None, engagement_name, product)
         test = get_target_test_if_exists(test_id, test_title, scan_type, engagement)
 
+        # have to make the scan_date_time timezon aware otherwise uploads via the API would fail (but unit tests for api upload would pass...)
```
StefanFl, line 1417: Typo in timezone (missing e)

StefanFl (changes requested):
I know, requesting changes about typos in comments is a bit picky 😃 Apart from that I like the PR.

dojo/api_v2/serializers.py
valentijnscholten: actually that word triggers the python runtime to go into overdrive mode

valentijnscholten: typo fixed.

dojo/api_v2/serializers.py
valentijnscholten: typo fixed.

StefanFl (approved):

devGregA (approved):

PR #5575, opened 2021-12-08
Fix typo

fabaff:
`migrate_sruveys` -> `migrate_surveys`


valentijnscholten:
thanks

valentijnscholten:
oh, merged into master. should have been dev, but not so bad for just a typo fix.

PR #5579, opened 2021-12-08
Fix: Solar Appscreener - add line property

zapililirad:
Parcer specified finding.sast_source_line property instead of finding.line - fixed.
Now specify both.


valentijnscholten:
Can you rebase on or merge latest dev? the unit tests files have been moved.

PR #5586, opened 2021-12-09
Add UI label to PR labeler and release drafter

kiblik:
It is good to know about changes in UI

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -39,6 +39,8 @@ categories:
     labels:
       - 'dependencies'
       - 'maintenance'
+  - title: '🖌 Updates in UI'
+    label: 'ui'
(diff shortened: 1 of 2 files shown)


.github/release-drafter.yml
```diff
@@ -39,6 +39,8 @@ categories:
     labels:
       - 'dependencies'
       - 'maintenance'
+  - title: '🖌 Updates in UI'
+    label: 'ui'
```
valentijnscholten, line 43: label or labels?

kiblik: I used the same syntax which is used for `documentation` or `apiv2`

StefanFl (approved):

devGregA (approved):

PR #5591, opened 2021-12-11
simplify and add comments for `auto_create_context`

valentijnscholten:
as requested by some I have simplified the logic a bit and added some comments to help understand the logic that checks and performs the `auto_create_context` actions.


dojo/api_v2/permissions.py
```diff
@@ -184,11 +184,12 @@ def has_permission(self, request, view):
             # engagement_id doesn't exist
             raise serializers.ValidationError("Engagement '%s' doesn''t exist" % engagement_id)
 
-        if check_import_engagement_permission(request.user, product, product_name, engagement_name, auto_create_context):
-            return True
-
-        if check_import_product_permission(request.user, product, product_name, product_type, product_type_name, auto_create_context):
-            return True
+        if not auto_create_context:
+            raise serializers.ValidationError("Engagement '%s' doesn't exist in Product %s" % (engagement_name, product_name))
```
StefanFl, line 188: This error message is misleading when the product doesn't exist, because it doesn't mention a missing product.

dojo/api_v2/permissions.py
```diff
@@ -278,30 +279,24 @@ def has_permission(self, request, view):
...
-            return True
+        if not auto_create_context:
+            # avoid NoneType errors on logging
+            if not engagement_name:
+                engagement_name = ''
 
-        if check_import_product_permission(request.user, product, product_name, product_type, product_type_name, auto_create_context):
-            return True
+            if test_title:
+                raise serializers.ValidationError("Test '%s' with scan_type '%s'  doesn't exist in Engagement %s" % (scan_type, test_title, engagement_name))
```
StefanFl, line 294: - `scan_type` and `test_title` are swapped in the error message
- There are 2 blanks before "doesn't"
- The error message is misleading when the product or product_type are not existing, because it doesn't mention it.

dojo/api_v2/permissions.py
```diff
@@ -278,30 +279,24 @@ def has_permission(self, request, view):
...
+            raise serializers.ValidationError("Test with scan_type '%s' doesn't exist in Engagement %s" % (scan_type, engagement_name))
+        else:
+            # the test doesn't exist, so we need to check if the user has requested and is allowed to use auto_create
+            if check_auto_create_permission(request.user, product, product_name, engagement, engagement_name, product_type, product_type_name):
+                return True
```
StefanFl, line 299: Not sure how to reach the error message in line 301, but it's not correct. You don't need the `test_title`. It can be empty when the `scan_type` is correct. _[Couldn't put a comment on line 301]_

dojo/importers/reimporter/utils.py
```diff
@@ -177,45 +177,49 @@ def get_target_test_if_exists(test_id=None, test_title=None, scan_type=None, eng
 
 
 def get_or_create_product(product_name=None, product_type_name=None, auto_create_context=None):
+    # try to find the procut (withing the provided product_type)
```
StefanFl, line 180: Typo in `product`

StefanFl (changes requested):
One more thing I couldn't comment in the changed code:

When Product and Engagement exist, Test has been manually created and should be found by `scan_type`: This doesn't work because the `scan_type in `manually created Tests is empty, only the test_type_id is set.

valentijnscholten:
>When Product and Engagement exist, Test has been manually created and should be found by scan_type: This doesn't work because the scan_type in manually created Tests is empty, only the test_type_id is set.

Sounds like a pre-existing bug? I thought the idea was that `scan_type` always contains the string representation of the Test_Type, with some suffix maybe for SARIF style tests?
I also notice you can't filter anywhere on `scan_type`, so you can only filter on all SARIF findings and not on SARIF Gitleaks for eaxmple? All this is a bit beyond this PR I think.

dojo/api_v2/permissions.py
```diff
@@ -373,11 +360,63 @@ def has_object_permission(self, request, view, obj):
...
+    if not engagement_name:
+        raise ValidationError("engagement_name parameter missing")
+
+    if engagement:
+        # existing engagement, nothing special to check
+        return user_has_permission(user, engagement, Permissions.Import_Scan_Result)
+
+    if product and product_name and engagement_name:
         if not user_has_permission(user, product, Permissions.Engagement_Add):
             raise PermissionDenied('No permission to create engagements in product %s', product_name)
```
StefanFl, line 421: The parameter needs to be in apostrophies. There are 3 more following.

StefanFl (commented):
- There is a bug in `dojo/importers/utils.py` line 121: `if product.product_type.name ...`: The Product_Type variable in Product is `prod_type`

StefanFl (approved):

devGregA (approved):

PR #5608, opened 2021-12-15
Remove Objects_Engagement and introduce flag to enable/disable Objects_Product

StefanFl:
Instances of `Objects_Engagements` could only be viewed but there was no way to add, edit or delete them. So the class and its functionality were removed.

`Objects_Product` don't seem to be used very often and can now be enabled/disabled with a flag in the System Settings to hide them in the user interface.


dojo/models.py
```diff
@@ -409,6 +409,11 @@ class System_Settings(models.Model):
         blank=False,
         verbose_name='Enable user profile for writing',
         help_text="When turned on users can edit their profiles")
+    enable_product_tracking_files = models.BooleanField(
+        default=True,
+        blank=False,
```
devGregA, line 414: I believe when you set blank=False for a boolean field it will not allow the checkbox to be left blank.

devGregA: I may have this confused with required looking at the file in it's totality, going to go ahead and approve.

devGregA (approved):

valentijnscholten (approved):

PR #5612, opened 2021-12-15
APIv2: Add preview of elements that will be affected by DELETE

kiblik:
`GET apiv2/{model}/{id}/delete_preview`

https://owasp.slack.com/archives/C014H3ZV9U6/p1639241564073500

## Example

```json
{
  "count": 123,
  "next": "http://DD/api/v2/users/1/delete_preview/?offset=2&limit=4",
  "previous": "http://DD/api/v2/users/1/delete_preview/?offset=10&limit=4",
  "results": [
      {
        "model": "User",
        "id": 1,
        "name": "admin"
      },
      {
        "model": "Product_Type_Member",
        "id": 1,
        "name": "Product_Type_Member object (1)"
      },
      {
        "model": "Alerts",
        "id": 4,
        "name": "Alerts object (4)"
      },
      {
        "model": "Alerts",
        "id": 2,
        "name": "Alerts object (2)"
      }
  ]
}
```

## TODO
- [x]  add pagination
- [x]  fix schema

### DIFF
--- a/dojo/api_v2/mixins.py
+++ b/dojo/api_v2/mixins.py
@@ -1,0 +1,35 @@
+from rest_framework.response import Response
+from django.db import DEFAULT_DB_ALIAS
+from django.contrib.admin.utils import NestedObjects
+from drf_spectacular.utils import extend_schema
+from drf_yasg.utils import swagger_auto_schema
+from rest_framework.decorators import action
+from rest_framework import status
+import itertools
+
+
+class DeletePreviewModelMixin:
+    @extend_schema(
+        methods=['GET'],
+    )
+    @swagger_auto_schema(
+        method='get',
+    )
+    @action(detail=True, methods=["get"])
+    def delete_preview(self, request, pk=None):
+        object = self.get_object()
+
+        collector = NestedObjects(using=DEFAULT_DB_ALIAS)
+        collector.collect([object])
+        rels = collector.nested()
+
+        def flatten(elem):
+            if isinstance(elem, list):
+                return itertools.chain.from_iterable(map(flatten, elem))
+            else:
+                return [elem]
+
+        rels = [{"model": type(x).__name__, "id": x.id, "name": str(x)} for x in flatten(rels)]
+
+        return Response(rels,
+                        status=status.HTTP_200_OK)
--- a/dojo/api_v2/mixins.py
+++ b/dojo/api_v2/mixins.py
@@ -29,7 +29,14 @@ class DeletePreviewModelMixin:
             else:
                 return [elem]
 
-        rels = [{"model": type(x).__name__, "id": x.id, "name": str(x)} for x in flatten(rels)]
+        rels = [
+            {
+                "model": type(x).__name__,
+                "id": x.id if hasattr(x, 'id') else -1,  # Because of "'Token' object has no attribute 'id'" and None is not Int
+                "name": str(x)
+            }
+            for x in flatten(rels)
+        ]
 
         return Response(rels,
                         status=status.HTTP_200_OK)
--- a/dojo/api_v2/mixins.py
+++ b/dojo/api_v2/mixins.py
@@ -37,9 +37,9 @@ class DeletePreviewModelMixin:
             {
                 "model": type(x).__name__,
                 "id": x.id if hasattr(x, 'id') else None,
-                "name": str(x)
+                "name": str(x) if not isinstance(x, Token) else "<APITokenIsHidden>"
             }
-            for x in flatten(rels) if not isinstance(x, Token)
+            for x in flatten(rels)
         ]
 
         # next part is inspired by original implementation of ListModelMixin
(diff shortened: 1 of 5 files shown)


damiencarol:
@kiblik I'm not a big fan of these previews. Sometimes the preview are really big and just crash (timeout).
We have this pb at my day job: some projects have so much data that it's impossible to use this "preview" screens.

kiblik:
> @kiblik I'm not a big fan of these previews. Sometimes the preview are really big and just crash (timeout). We have this pb at my day job: some projects have so much data that it's impossible to use this "preview" screens.

We can set it as an optional API call (by env variable). 
And you don't have to use it before you call `DELETE ...`. It is up to you.

Maffooch (approved):
I think this is a good idea. It only affects the API, and it is more like a dry run of a delete rather than just doing it and not understanding what is actually deleted.

Admittedly, the previews in the UI do often crash with just a small amount of data, but this PR would not make that issue better or worse.

Maffooch:
Oops I did not see that tests were not passing. Keeping the approval for now

kiblik:
@Maffooch, I used `WIP:` in the title because it is not ready
Some parts are not done, tests are failing, there is some new conflict, so please do not merge it yet.
But thanks for the support :)

damiencarol:
> @Maffooch, I used WIP: in the title because it is not ready
Some parts are not done, tests are failing, there is some new conflict, so please do not merge it yet.
But thanks for the support :)

@kiblik better to use the 'Draft' feature of GitHub => https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request

kiblik:
> > @Maffooch, I used WIP: in the title because it is not ready
> > Some parts are not done, tests are failing, there is some new conflict, so please do not merge it yet.
> > But thanks for the support :)
> 
> @kiblik better to use the 'Draft' feature of GitHub => https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request

Thank you. I knew drafts but the "Convert to draft" button was too small and I missed it .

kiblik:
The main part of implementation is done but small help is still needed. I was able to force swagger to add fields for pagination (by adding `...Serializer(many=True)`) but this step also added other GET parameters (parameters used in the original Model). For example `User`:
image

valentijnscholten:
Is this happening for both Open API 2 and 3 specs? Looks like maybe it gets confused by having both `many=True` and `detail=True` on the same method.

kiblik:
> Is this happening for both Open API 2 and 3 specs? Looks like maybe it gets confused by having both `many=True` and `detail=True` on the same method.

I haven't checked Open API 2 before. Now I see that paging fields are not there - Btw, does it need to be fixed? How long Open API 2 swagger will be there?

`detail=True` allows to use format `/api/v2/users/{id}/delete_preview/` (otherwise there would be `/api/v2/users/delete_preview/`). `many=True` is there because API call return list (one only one element) of affected elements and allows to use pagination.

If `detail` would be `False`, `self.get_object()` is not possible to use.

valentijnscholten:
@kiblik lots of changes to drf_spectacular in https://github.com/DefectDojo/django-DefectDojo/pull/6046, could you retry?

kiblik:
Still the same

kiblik:
After better reading of [FAQ](https://drf-spectacular.readthedocs.io/en/latest/faq.html#my-action-is-erroneously-paginated-or-has-filter-parameters-that-i-do-not-want), I found solution: Add `filter_backends=[]` to `@action`.
image


However, this is the solution only for APIv2 (`drf-spectacular`). I wasn't able to find a solution for APIv1 (`drf_yasg`). Should I spend more time digging deeper or is it an acceptable solution?
image

valentijnscholten:
> After better reading of [FAQ](https://drf-spectacular.readthedocs.io/en/latest/faq.html#my-action-is-erroneously-paginated-or-has-filter-parameters-that-i-do-not-want), I found solution: Add `filter_backends=[]` to `@action`.


Nice find!
 
> However, this is the solution only for OpenAPI3 (`drf-spectacular`). I wasn't able to find a solution for OpenAPI2 (`drf_yasg`). Should I spend more time digging deeper or is it an acceptable solution?

I think asking the `drf-yasg` team how this could be solved is a small task we could do? It would be really impractical if we have an incorrect API specs and would have to clear document that or add some description to the affected endpoints.

kiblik:
We have received the reaction https://github.com/axnsan12/drf-yasg/issues/787#issuecomment-1105787846
> ...
> - spectacular response suggests there is a `pagination_class` present on that view/action. The pagination logic gets activated by doing `Serializer(many=True)`.
> - we make no guarantees about correctness when using both `spectacular` and `yasg` decorators at the same time. your mileage may vary.

Maybe I will try to debug `yasg` how and why the "pagination engine" is not activated in our case even if we are using `Serializer(many=True)`

kiblik:
I finally found the solution.

More details: https://github.com/axnsan12/drf-yasg/issues/787#issuecomment-1116777245

StefanFl (approved):

Maffooch (approved):
Double approved! Good addition @kiblik

PR #5613, opened 2021-12-16
Updating engineer metrics and removing research metrics

StefanFl:
On the way to remove the significance of the `is_staff` flag, I had a look at metrics:

- Engineer metrics where restricted to staff users. This isn't necessary anymore because every user can import findings, if they have the permissions. Still only superusers can select other users to view, non-superusers can only see their metrics. While I was at it, I gave the user selection dialogue a revamp.
- Research metrics where very specific, only Findings with the test_type `Security Research` are shown here. This should be removed because it seems to be a specific use case for one installation. If it's still needed, I am happy to remove the removal from this PR.
- Found a little bug in the user filter, where authorized users where not removed yet.


dojo/templates/dojo/users.html
```diff
@@ -53,7 +53,7 @@ <h3 class="has-filters">
                             <th class="nowrap">Phone Number(s)</th>
                             <th class="nowrap, text-center">{% dojo_sort request 'Active' 'is_active' %}</th>
                             <th class="nowrap, text-center">{% dojo_sort request 'Staff' 'is_staff' %}</th>
```
damiencarol, line 55: should we remove this also? notify that it will be deprecated soon?

damiencarol (approved):

dojo/templates/dojo/users.html
StefanFl: That's for the next release, there are some steps to do until then.

damiencarol: ok

valentijnscholten (approved):
I like the cleanup. But for a future PR we could think about letting non-superusers view metrics for other users as well? For example owners and maintainers might want to see these?
It's probably not an wasy change to make as you need to decide which users are shown in the dropdowns. I think JIRA struggles with this and has a global "Browse User" permissions used on most places. But maybe we can think of some nice query based on permissions for the product/producttype.

valentijnscholten:
@devGregA @Maffooch Do you agree with the removal?

StefanFl:
> But for a future PR we could think about letting non-superusers view metrics for other users as well? For example owners and maintainers might want to see these?

Something we can think about in another release.

kiblik (changes requested):

devGregA:
@valentijnscholten will take a good look at this by tomorrow evening (US time).

Maffooch:
I am fine with this. All yours @devGregA

Maffooch (approved):

dojo/metrics/views.py
```diff
@@ -775,15 +767,14 @@ def engineer_metrics(request):
 
 
 # noinspection DjangoOrm
-@user_passes_test(lambda u: u.is_staff)
 @cache_page(60 * 5)  # cache for 5 minutes
 @vary_on_cookie
 def view_engineer(request, eid):
     user = get_object_or_404(Dojo_User, pk=eid)
     if not (request.user.is_superuser or
             request.user.username == 'root' or
```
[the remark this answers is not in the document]
StefanFl: Yes, this hard coded user name should be removed. Okay for you @devGregA?

devGregA:
looks good to me

devGregA (approved):

Issue #5607, opened 2021-12-14
Choctaw Hog Scanner replacement through Rusty Hog


manuel-sommer:
[Rusty Hog](https://github.com/newrelic/rusty-hog) is a secret scanner built in Rust for performance, and based on TruffleHog which is written in Python. Rusty Hog provides the following binaries:
- Ankamali Hog: Scans for secrets in a Google doc.
- Berkshire Hog: Scans for secrets in an S3 bucket.
- **Choctaw Hog:** Scans for secrets in a Git repository.
- Duroc Hog: Scans for secrets in a directory, file, and archive.
- Essex Hog: Scans for secrets in a Confluence wiki page.
- Gottingen Hog: Scans for secrets in a JIRA issue.
- Slack Hog: Scans for secrets in a Slack Channel.

We could rename Choctaw Hog to Rusty Hog and advance the parser from choctaw hog to rusty hog to also include the other secret scanners. Or do we rather want a separate parser for each Hog?

I have already tested an upload of a Gottingen Hog output file to Choctaw Hog and the parser processes it. Only [this method](https://github.com/DefectDojo/django-DefectDojo/blob/ba1a415219ff20e8b4e909ef14f750de9b80297e/dojo/tools/choctaw_hog/parser.py#L43) would have to be updated.

[gottingen.zip](https://github.com/DefectDojo/django-DefectDojo/files/7715262/gottingen.zip)

manuel-sommer:
I will close this issue as the PR got merged.

PR #5614, opened 2021-12-16
rustyhog replaces choctawhog and gottingenhog is added #5607

manuel-sommer:
This PR renames ChoctawHog to Rustyhog https://github.com/newrelic/rusty-hog as ChoctawHog is a part of RustyHog. Furthermore, GottingenHog is added to Rustyhog. Thus, Rustyhog now is able to process ChoctawHog and GottingenHog


manuel-sommer:
In case this gets merged, I will advance it with Essex Hog in another PR

dojo/tools/rusty_hog/parser.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # create the finding object
+    finding = Finding(
+        title=title,
+        test=test,
+        severity='High',
+        cwe=cwe,
+        description=description,
+        mitigation="Please ensure no secret material nor confidential information is kept in clear within git repositories.",
+        file_path=file_path,
```
StefanFl, line 82: Can you set `line` as well, e.g. if `new_line_num` is given in the report?

dojo/tools/rusty_hog/parser.py
```diff
@@ -0,0 +1,89 @@
...
+        static_finding=True,
+        dynamic_finding=False
+    )
```
StefanFl, line 85: It would be helpful to see in the list, which scanner has found the finding (ChoctawHog or GottingenHog). We have had a PR for the Trivy parser recently setting a tag with the kind of scanner. This could be done here as well.

unittests/tools/test_rusty_hog_parser.py
```diff
@@ -0,0 +1,41 @@
+from ..dojo_test_case import DojoTestCase
+from dojo.tools.rusty_hog.parser import RustyhogParser
+from dojo.models import Test
+
+
+class TestRustyhogParser(DojoTestCase):
```
StefanFl, line 6: The test cases shouldn't only check the number of findings but some or all of the expected contents as well.

dojo/tools/rusty_hog/parser.py
```diff
@@ -0,0 +1,89 @@
...
+        items = {}
+        for node in tree:
+            item = get_item(node, test)
+            unique_key = "Finding {}".format(tree.index(node))
+            items[unique_key] = item
+
+        return list(items.values())
+
+
+def get_item(vulnerability, test):
```
StefanFl, line 43: Can you find out if data is coming from a subparser that's not supported and throw an exception?

StefanFl (changes requested):
First of all, thank you for this PR.

This is a breaking change, it needs documentation in 2 files:

- In https://github.com/DefectDojo/django-DefectDojo/blob/master/docs/content/en/getting_started/upgrading.md please start a section **Upgrading to DefectDojo Version 2.6.x** describing the breaking change removing ChoctawHog by RustyHog.
- In https://github.com/DefectDojo/django-DefectDojo/blob/master/docs/content/en/integrations/parsers.md please remove the ChoctawHog and add the RustyHog parser, describing which subparsers are supported.

valentijnscholten:
Is it also an option to add a migration to rename the existing Test_Type so history is maintained and deduplication / reimport works?

damiencarol:
1) @manuel-sommer does the output format is consistent between all these scanners?
If yes to will make this format a meta-format and it maybe a better idea to manage it like SARIF format.
For SARIF parser we create new Test_type dynamically. This way it will be possible to keep `"Choctaw Hog Scan"` string and stay compatible with existing findings.
@valentijnscholten it will solve what your pointing at.
2) @manuel-sommer  We need to check also if the `settings.py` need modifications.

dojo/tools/rusty_hog/parser.py
```diff
@@ -0,0 +1,125 @@
...
+    def parse_json(self, json_output):
+        try:
+            data = json_output.read()
+            try:
+                tree = json.loads(str(data, 'utf-8'))
+            except:
+                tree = json.loads(data)
+        except:
+            raise Exception("Invalid format")
+        return tree
```
Maffooch, line 30: This does not appear to be used. Can you either use it in places where `json.load()` is used?

dojo/tools/rusty_hog/parser.py
```diff
@@ -0,0 +1,125 @@
...
+                if vulnerability.get('parent_commit_hash') is not None:
+                    description += "\n**Parent commit hash:** {}".format(vulnerability.get('parent_commit_hash'))
+                if vulnerability.get('old_file_id') is not None and vulnerability.get('new_file_id') is not None:
+                    description += "\n**Old and new file IDs:** {} - {}".format(
+                                    vulnerability.get('old_file_id'),
+                                    vulnerability.get('new_file_id'))
+                if vulnerability.get('old_line_num') is not None and vulnerability.get('new_line_num') is not None:
+                    description += "\n**Old and new line numbers:** {} - {}".format(
+                                    vulnerability.get('old_line_num'),
+                                    vulnerability.get('new_line_num'))
```
Maffooch, line 87: It would be great to use the line num info in the finding `line` field. Since this would only be applicable for Choctaw Hogs scans, it can be added to the finding after initialization by checking if `line` variable previously defined above is no longer equal to `''`

Maffooch (changes requested):

dojo/tools/rusty_hog/parser.py
```diff
@@ -0,0 +1,127 @@
...
+                        vulnerability.get('reason'),
+                        vulnerability.get('issue_id'),
+                        vulnerability.get('location'))
+            # create the finding object
+            finding = Finding(
+                title=title,
+                severity='High',
+                cwe=cwe,
+                description=description,
+                mitigation="Please ensure no secret material nor confidential information is kept in clear within git repositories.",
```
StefanFl, line 118: Just saw this while testing the PR. The mitigation text shall be different for each parser. For Gottingen Hog it shall reference to Jira and not git.

Apart from this the PR is fine for me.

StefanFl (approved):

manuel-sommer:
@Maffooch or @valentijnscholten could you review and merge this?

Maffooch (approved):

manuel-sommer:
@damiencarol I will make another MR to improve this

unittests/scans/rusty_hog/gottingenhog_one_vuln.json
```diff
@@ -0,0 +1,11 @@
+[
+  {
+    "stringsFound": [
+      "password: jeans"
+    ],
+    "issue_id": "TEST-123",
```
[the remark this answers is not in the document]
manuel-sommer: issue_id referse to the JIRA Ticket. However, a JIRA Ticket can have multiple findings. Thus, this is not a good candidate anymore as it will not be unique in these cases.

PR #5617, opened 2021-12-16
Reversed x-axis on Product's Detailed Metrics Page

blakeaowens:
Previously, some of the plots on the Detailed Metrics page displayed dates with the most recent date on the far left side of the x-axis.
Screen Shot 2021-12-16 at 1 41 02 PM

With this change, these plots will display the most recent dates on the far right side of the x-axis.
Screen Shot 2021-12-16 at 1 44 27 PM


Maffooch (approved):

mtesauro (approved):
Probably good that we're consistent.

Approved.

PR #5620, opened 2021-12-17
Solar change deduplication model

zapililirad:
Solar Appscreener parcer:
Correct settings to apply hashcode as deduplication algoritm
Add line field to finding in parcer

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -1040,6 +1040,7 @@ HASHCODE_FIELDS_PER_SCANNER = {
     'Meterian Scan': ['cwe', 'component_name', 'component_version', 'description', 'severity'],
     'Github Vulnerability Scan': ['unique_id_from_tool'],
     'Azure Security Center Recommendations Scan': ['unique_id_from_tool'],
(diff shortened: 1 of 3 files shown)


dojo/settings/settings.dist.py
```diff
@@ -1073,6 +1074,7 @@ def saml2_attrib_map_format(dict):
     'Hadolint Dockerfile check': True,
     'Semgrep JSON Report': True,
     'Generic Findings Import': True,
+    'Solar Appscreener Scan': True,
```
StefanFl, line 1077: This line is not necessary, because CWE is not in the hash_code computation

StefanFl (changes requested):
Thanks for this PR, one small change requested.

dojo/settings/settings.dist.py
damiencarol: I agree. please remove this change only and we can move forward.

damiencarol, line 1077:
```suggestion
```

zapililirad:
done

StefanFl (approved):

PR #5626, opened 2021-12-21
update tagulous to 1.3.1 instead of hotfix commit

valentijnscholten:
their bug is fixed in this release so we no longer need to install that specific hotfix commit


StefanFl (approved):

mtesauro (approved):
Approved

Issue #5618, opened 2021-12-17
`close_old_findings` only from findings with the same `service`


rucciva:
**Is your feature request related to a problem? Please describe**
I'm currently trying to import scans from a product which in reality consisted of multiple repository and multiple services. I modeled this product as a product in defect dojo, which will have many CI/CD engagements as many as CI/CI job that run triggered by changes in repositories or services of the product. I enabled deduplication on product level and in each scans upload enabled the `close_old_findings` flag.  I set the `service` attribute of scans upload to repositories url or service base url which the scans is targeted too. Deduplications across many engagements work great. The problem is in `close_old_findings`, which will close scans from older engagements which might not includes scan reports from the same repository or service.

**Describe the solution you'd like**
add filter by `service` in [here](https://github.com/DefectDojo/django-DefectDojo/blob/2796e2cbd532edc3c2d270fb215b422786d977d3/dojo/importers/importer/importer.py#L188-L200) so that product that has many repositories or services can still works with `close_old_findings`

**Describe alternatives you've considered**
the alternative would be to have one engagement only for each repository or service being scanned and enable engagement level deduplication. But that would seem weird since the engagement will never be closed.

another alternative is to have one product per repository or service. the downside will be losing real product level metrics in `product/{id}/metrics` page

**Additional context**
Add any other context, screenshots, sketch, code snippet, etc. about the feature request here.

StefanFl:
Thank you @rucciva for bringing this to our attention. I will have a look at it.

damiencarol:
@StefanFl @rucciva could it be better to and `service` attribute as a possible attribute for de-duplication?

rucciva:
> @StefanFl @rucciva could it be better to add `service` attribute as a possible attribute for de-duplication?

From what i read [here](https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/settings/settings.dist.py#L1082),  `service` has been used in de-duplication, but only if dedup is done based on `hash_code` in .

But i personally think that there should be one more explicit abstraction under _engagement_ that denotes the component being scanned (repository, service, or any other kind of component) where de-duplication and close_old_findings should be limited to. De-duplication and close_old_findings across scans only make sense when the things being scanned is the same thing, while product and engagement it self is too broad for that.

StefanFl:
Closing old findings have to acknowledge the `service` attribute like deduplication does already. This is why I think that's a bug instead of an enhancement.

[...]

PR #5631, opened 2021-12-22
Close old findings of same service only

StefanFl:
implements #5618

When a `service` is specified and **Close old findings** is set as well while importing a scan, only the old findings of the same service shall be closed.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1230,10 +1230,15 @@ class ImportScanSerializer(serializers.Serializer):
         default=None,
         queryset=User.objects.all())
     tags = TagListSerializerField(required=False)
-    close_old_findings = serializers.BooleanField(required=False, default=False)
+    close_old_findings = serializers.BooleanField(required=False, default=False,
+        help_text="Select if old findings no longer present in the report get closed as mitigated when importing. "
+                  "If service has been set, only the findings for this service will be closed.")
     push_to_jira = serializers.BooleanField(default=False)
     environment = serializers.CharField(required=False)
-    version = serializers.CharField(required=False)
+    version = serializers.CharField(required=False,
+        help_text="A service is a self-contained piece of functionality within a Product. "
+                  "This is an optional field which is used in deduplication and closing of old findings when set. "
+                  "This affects the whole engagement/product depending on your deduplication scope.")
     build_id = serializers.CharField(required=False)
     branch_tag = serializers.CharField(required=False)
     commit_hash = serializers.CharField(required=False)
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1235,16 +1235,16 @@ class ImportScanSerializer(serializers.Serializer):
                   "If service has been set, only the findings for this service will be closed.")
     push_to_jira = serializers.BooleanField(default=False)
     environment = serializers.CharField(required=False)
-    version = serializers.CharField(required=False,
-        help_text="A service is a self-contained piece of functionality within a Product. "
-                  "This is an optional field which is used in deduplication and closing of old findings when set. "
-                  "This affects the whole engagement/product depending on your deduplication scope.")
+    version = serializers.CharField(required=False)
     build_id = serializers.CharField(required=False)
     branch_tag = serializers.CharField(required=False)
     commit_hash = serializers.CharField(required=False)
     api_scan_configuration = serializers.PrimaryKeyRelatedField(allow_null=True, default=None,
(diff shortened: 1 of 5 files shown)


valentijnscholten:
This looks like maybe a good default, might be that people want to have some control over it in the future (per request).
Should the docs get a small update with this behaviour?

StefanFl:
> This looks like maybe a good default, might be that people want to have some control over it in the future (per request).

Not sure what users should have control over, to be honest. The `service` field has a defined purpose and that needs to be reflected in deduplication and closing old findings.

> Should the docs get a small update with this behaviour?

I am afraid this would need a larger update.  **Close old findings** is not described for imports currently, only reimports and API mentions it. There are a few other things, like a really old screenshot. I would like to update documentation for import and reimport in a separate PR.

valentijnscholten:
The help_text of the field(s) would be be easy to update, same for a help_text in the serializer field(s).

StefanFl:
> The help_text of the field(s) would be be easy to update, same for a help_text in the serializer field(s).

Makes sense, done

dojo/importers/importer/importer.py
```diff
@@ -179,18 +179,32 @@ def close_old_findings(self, test, scan_date_time, user, push_to_jira=None):
...
+                                                .exclude(hash_code__in=new_hash_codes) \
+                                                .filter(test__engagement__product=test.engagement.product,
+                                                    test__test_type=test.test_type,
+                                                    active=True, service=service)
+            else:
+                old_findings = Finding.objects.exclude(test=test) \
+                                                .exclude(hash_code__in=new_hash_codes) \
+                                                .filter(test__engagement__product=test.engagement.product,
+                                                    test__test_type=test.test_type,
+                                                    active=True)
```
valentijnscholten, line 207:
```suggestion
            old_findings = Finding.objects.exclude(test=test) \
                                                .exclude(hash_code__in=new_hash_codes) \
                                                .filter(test__engagement__product=test.engagement.product,
                                                    test__test_type=test.test_type,
                                                    active=True)

            if service:
                old_findings = old_findings.filter(service=service)
```

valentijnscholten (changes requested):
See my suggestion, it's cleaner I think. Same for the other location where this additional filtering takes place.

dojo/importers/importer/importer.py
StefanFl: Thank you @valentijnscholten, your suggestion is not only better but made me find a bug as well. The new version works with all combinations of having or not having a service when importing scans.

valentijnscholten: What's the 'service=True' for?

StefanFl: This was a mistake of course, should have filtered for `is_null` what it's doing now. That was the reason why the unit test for API's initially failed.

dojo/api_v2/serializers.py
```diff
@@ -1230,10 +1230,15 @@ class ImportScanSerializer(serializers.Serializer):
...
+    close_old_findings = serializers.BooleanField(required=False, default=False,
+        help_text="Select if old findings no longer present in the report get closed as mitigated when importing. "
+                  "If service has been set, only the findings for this service will be closed.")
     push_to_jira = serializers.BooleanField(default=False)
     environment = serializers.CharField(required=False)
-    version = serializers.CharField(required=False)
+    version = serializers.CharField(required=False,
+        help_text="A service is a self-contained piece of functionality within a Product. "
+                  "This is an optional field which is used in deduplication and closing of old findings when set. "
+                  "This affects the whole engagement/product depending on your deduplication scope.")
```
Maffooch, line 1241: Is this help text intended to go under the version field?

StefanFl: Good find @Maffooch, of course it should go under the `service` field. Done

dojo/importers/importer/importer.py
StefanFl: @valentijnscholten Added some test cases

valentijnscholten:
Thanks. In a perfect world there would also be findings in the unit test that are _not_ closed because they have a different value for `service`.

StefanFl:
> Thanks. In a perfect world there would also be findings in the unit test that are not closed because they have a different value for service.

The world is not perfect but in this case nearly :wink: The test cases `test_import_param_close_old_findings_with_and_without_service_1` and `test_import_param_close_old_findings_with_and_without_service_2` are doing exactly that.

valentijnscholten (approved):

valentijnscholten:
OK, misread the diff. Approving.

PR #5633, opened 2021-12-23
Update to Django 3.2.10

StefanFl:
dependabot missed release 3.2.10 of Django, maybe because we declined 4.0

This Django version fixes a vulnerability.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,7 @@ django-slack==5.17.6
 django-tagging==0.5.0
 django-watson==1.6.0
 django-prometheus==2.1.0
-Django==3.2.9
+Django==3.2.10
 djangorestframework==3.12.4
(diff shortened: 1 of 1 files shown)


damiencarol (approved):

valentijnscholten:
not the first time dependabot missed something. i raised 2 support cases already but they only keep limited logs so the answer is always "we don't have logs anymore for when this happened"

valentijnscholten (approved):

damiencarol:
> not the first time dependabot missed something. i raised 2 support cases already but they only keep limited logs so the answer is always "we don't have logs anymore for when this happened"

It could be a side effect. Maybe it's due to the fact that we have a lot of dependency in the same repository (Python app, docker files, Helm charts, JS components...)

Issue #5636, opened 2021-12-24
Import "Checkmarx Scan" failed when filename is empty


rucciva:
**Bug description**
I don't know if this is a bug in checkmarx or not but its report can contain node with empty filename. When uploaded to defect dojo, it will throw exception

**Steps to reproduce**
Steps to reproduce the behavior: Upload scans from checkmarx result than contains empty filename on some of the query

**Expected behavior**
No Error

**Deployment method** *(select with an `X`)*
- [x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Ubuntu 20.04
 - DefectDojo version (see footer) or commit message: 2.5.0

**Logs** 

```
uwsgi_1         | [24/Dec/2021 05:05:41] ERROR [dojo.engagement.views:625] expected str, bytes or os.PathLike object, not NoneType
uwsgi_1         | Traceback (most recent call last):
uwsgi_1         |   File "/app/./dojo/engagement/views.py", line 610, in import_scan_results
uwsgi_1         |     test, finding_count, closed_finding_count = importer.import_scan(scan, scan_type, engagement, user, environment, active=active, verified=verified, tags=tags,
uwsgi_1         |   File "/app/./dojo/importers/importer/importer.py", line 331, in import_scan
uwsgi_1         |     parsed_findings = parser.get_findings(scan, test)
uwsgi_1         |   File "/app/./dojo/tools/checkmarx/parser.py", line 97, in get_findings
uwsgi_1         |     self.process_result_file_name_aggregated(dupes, findingdetail, query, result, find_date)
uwsgi_1         |   File "/app/./dojo/tools/checkmarx/parser.py", line 113, in process_result_file_name_aggregated
uwsgi_1         |     title = "{} ({})".format(titleStart, ntpath.basename(sinkFilename))
uwsgi_1         |   File "/usr/local/lib/python3.8/ntpath.py", line 216, in basename
uwsgi_1         |     return split(p)[1]
uwsgi_1         |   File "/usr/local/lib/python3.8/ntpath.py", line 185, in split
uwsgi_1         |     p = os.fspath(p)
uwsgi_1         | TypeError: expected str, bytes or os.PathLike object, not NoneType
```

**Sample scan files (truncated)**
```
<?xml version="1.0" encoding="utf-8"?>
[...]

PR #5638, opened 2021-12-24
Checkmarx parser: fix empty filename

damiencarol:
- [x] fix a bug when file name is empty #5636
- [x] improve language stats data (number of files+fix stats in the UI)
- [x] remove some useless fake data (mitigation, impact and references blank)


dojo/tools/checkmarx/parser.py
```diff
@@ -45,21 +42,17 @@ def get_description_for_scan_types(self, scan_type):
...
 
-    # FIXME get rid of local variables
-    language_list = []
-    mitigation = 'N/A'
-    impact = 'N/A'
-    references = ''
-
     def get_findings(self, filename, test):
         cxscan = ElementTree.parse(filename)
+        # FIXME get rid of local variable
```
StefanFl, line 47: Will you fix this within this PR?

dojo/tools/checkmarx/parser.py
```diff
@@ -79,8 +72,10 @@ def get_findings(self, filename, test):
 
                 if language is not None:
                     findingdetail = "{}**Language:** {}\n".format(findingdetail, language)
-                    if language not in self.language_list:
-                        self.language_list.append(language)
+                    if language not in language_list:
+                        language_list[language] = 1
+                    else:
+                        language_list[language] = language_list[language] + 1
```
StefanFl, line 78: How much sense makes it to count the number of files for a language? You only get the number of files with vulnerabilities, right?

dojo/tools/checkmarx/parser.py
```diff
@@ -110,7 +105,10 @@ def process_result_file_name_aggregated(self, dupes, findingdetail, query, resul
         titleStart = query.get('name').replace('_', ' ')
         description, lastPathnode = self.get_description_file_name_aggregated(query, result)
         sinkFilename = lastPathnode.find('FileName').text
-        title = "{} ({})".format(titleStart, ntpath.basename(sinkFilename))
+        if sinkFilename:
+            title = "{} ({})".format(titleStart, sinkFilename.split("/")[-1])
```
StefanFl, line 109: What does happen, when a filename doesn't have a `/`? And does this work when the check is done on Windows (if this is possible)?

dojo/tools/checkmarx/parser.py
```diff
@@ -210,7 +203,7 @@ def process_result_detailed(self, dupes, findingdetail, query, result, find_date
             # pathId is the unique id from tool which means that there is basically no aggregation except real duplicates
             aggregateKeys = "{}{}{}{}{}".format(categories, cwe, name, sinkFilename, pathId)
             if title and sinkFilename:
-                title = "{} ({})".format(title, ntpath.basename(sinkFilename))
+                title = "{} ({})".format(title, sinkFilename.split("/")[-1])
```
StefanFl, line 206: see my question on line 109

StefanFl (commented):
There are some questions

dojo/tools/checkmarx/parser.py
damiencarol: Yes. but better than what we have for now: it was only the language so the graphs were bugged.
My opinion is we don't have enough data so we should completely get ride of this.

dojo/tools/checkmarx/parser.py
damiencarol: the format is not dependent of the OS. the data in the report are always like this: `WebGoat/webgoat-lessons/missing-function-ac/src/main/java/org/owasp/webgoat/plugin/Users.java`.
This version is better as it doesn't fail with empty strings.

dojo/tools/checkmarx/parser.py
damiencarol: I fixed 3 of 4. I can complete this.

damiencarol: pushed

dojo/tools/checkmarx/parser.py
StefanFl: I wouldn't miss it.

dojo/tools/checkmarx/parser.py
StefanFl: ok

devGregA (approved):

dojo/tools/checkmarx/parser.py
damiencarol: Let's do that in another PR. for now it keeps things as before, I will ask users and remove them if not required anymore

StefanFl (approved):

PR #5640, opened 2021-12-25
Revamp of the Anchore Grype parser

StefanFl:
This PR is basically a rewrite of the Anchore Grype parser:

- The reports contain a lot of information that was not stored in the findings so far. Now the findings are much richer than before.
- Internal deduplication was based on CVE only. When you have several components with the same CVE, they had been stored in one finding under one component name. Now the internal deduplication is done by the title, which contains CVE, component name and component version.

### DIFF
--- a/dojo/tools/anchore_grype/parser.py
+++ b/dojo/tools/anchore_grype/parser.py
@@ -39,6 +39,7 @@ class AnchoreGrypeParser(object):
             rel_datasource = None
             rel_urls = None
             rel_description = None
+            rel_cvss = None
             related_vulnerabilities = item.get('relatedVulnerabilities', None)
             if related_vulnerabilities:
                 related_vulnerability = related_vulnerabilities[0]
@@ -46,14 +47,9 @@ class AnchoreGrypeParser(object):
                 rel_datasource = related_vulnerability.get('dataSource', None)
                 rel_urls = related_vulnerability.get('urls', None)
                 rel_description = related_vulnerability.get('description', None)
-            rel_cvss = related_vulnerability.get('cvss', None)
+                rel_cvss = related_vulnerability.get('cvss', None)
 
-            match_matcher = None
             matches = item['matchDetails']
-            if matches:
-                match_matcher = matches[0]['matcher']
-                if match_matcher.endswith('-matcher'):
-                    match_matcher = match_matcher.replace('-matcher', '')
 
             artifact = item['artifact']
             artifact_name = artifact.get('name', None)
@@ -63,6 +59,7 @@ class AnchoreGrypeParser(object):
             cve = self.get_cve(vuln_id, rel_id)
             finding_title = f'{cve} in {artifact_name}:{artifact_version}'
 
+            finding_tags = None
             finding_description = f'**Vulnerability Id:** {vuln_id}'
             if vuln_namespace:
                 finding_description += f'\n**Vulnerability Namespace:** {vuln_namespace}'
@@ -73,12 +70,20 @@ class AnchoreGrypeParser(object):
             if rel_description and rel_description != vuln_description:
                 finding_description += f'\n**Related Vulnerability Description:** {rel_description}'
             if matches:
-                if len(matches) == 1:
+                if type(item["matchDetails"]) is dict:
+                    finding_description += f"\n**Matcher:** {matches['matcher']}"
+                    finding_tags = [matches['matcher'].replace('-matcher', '')]
+                elif len(matches) == 1:
                     finding_description += f"\n**Matcher:** {matches[0]['matcher']}"
+                    finding_tags = [matches[0]['matcher'].replace('-matcher', '')]
                 else:
                     finding_description += '\n**Matchers:**'
+                    finding_tags = []
                     for match in matches:
                         finding_description += f"\n- {match['matcher']}"
+                        tag = match['matcher'].replace('-matcher', '')
+                        if tag not in finding_tags:
+                            finding_tags.append(tag)
 
             if artifact_purl:
                 finding_description += f'\n**Package URL:** {artifact_purl}'
@@ -90,9 +95,9 @@ class AnchoreGrypeParser(object):
 
             finding_mitigation = None
             if vuln_fix_versions:
-                finding_mitigation = 'Upgrade to version: '
+                finding_mitigation = 'Upgrade to version:'
                 if len(vuln_fix_versions) == 1:
-                    finding_mitigation += vuln_fix_versions[0]
+                    finding_mitigation += f' {vuln_fix_versions[0]}'
                 else:
                     for fix_version in vuln_fix_versions:
                         finding_mitigation += f'\n- {fix_version}'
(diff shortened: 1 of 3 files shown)


dojo/tools/anchore_grype/parser.py
```diff
@@ -23,44 +22,143 @@ def get_findings(self, file, test):
...
-                    component_version=purl.version,
-                    vuln_id_from_tool=cve,
-                    nb_occurences=1,
-                )
+                            title=finding_title,
+                            description=finding_description,
+                            cve=finding_cve,
+                            cwe=1352,
+                            cvssv3=finding_cvss3,
+                            cvssv3_score=finding_cvss3_score,
```
damiencarol, line 149: If you provide the vector, the score is calculated automatically.
```suggestion
```
So better to provide the vector only.

dojo/tools/anchore_grype/parser.py
```diff
@@ -23,44 +22,143 @@ def get_findings(self, file, test):
...
+            vuln_id = vulnerability["id"]
+            vuln_namespace = vulnerability.get('namespace', None)
+            vuln_datasource = vulnerability.get('dataSource', None)
+            vuln_severity = self._convert_severity(vulnerability["severity"])
+            vuln_urls = vulnerability.get('urls', None)
+            vuln_description = vulnerability.get('description', None)
+            vuln_fix_versions = None
+            fix = vulnerability.get('fix', None)
+            if fix:
+                vuln_fix_versions = fix.get('versions', None)
```
damiencarol, line 35:
```suggestion
            if 'fix' in vulnerability:
                vuln_fix_versions = vulnerability.get('fix').get('versions')
```
Better, you can do it in one line:
```
vuln_fix_versions = vulnerability.get('fix', []).get('versions')
```
But the previous line is more readable.

dojo/tools/anchore_grype/parser.py
```diff
@@ -23,44 +22,143 @@ def get_findings(self, file, test):
...
-            severity = self._convert_severity(item["vulnerability"]["severity"])
-            purl = PackageURL.from_string(item["artifact"]["purl"])
-            description = ""
-            description += f"\n**CVE:** {cve}"
-            if type(item["matchDetails"]) is dict:
-                description += f'\n**Matcher:** {item["matchDetails"]["matcher"]}'
+            vulnerability = item['vulnerability']
+            vuln_id = vulnerability["id"]
+            vuln_namespace = vulnerability.get('namespace', None)
+            vuln_datasource = vulnerability.get('dataSource', None)
```
damiencarol, line 28: The method `get()` return `None` if the key is not in the dictionary.
So the correct form is:
```suggestion
            vuln_namespace = vulnerability.get('namespace')
            vuln_datasource = vulnerability.get('dataSource')
```

dojo/tools/anchore_grype/parser.py
```diff
@@ -23,44 +22,143 @@ def get_findings(self, file, test):
...
+                rel_urls = related_vulnerability.get('urls', None)
+                rel_description = related_vulnerability.get('description', None)
+                rel_cvss = related_vulnerability.get('cvss', None)
+
+            matches = item['matchDetails']
+
+            artifact = item['artifact']
+            artifact_name = artifact.get('name', None)
+            artifact_version = artifact.get('version', None)
+            artifact_purl = artifact.get('purl', None)
```
damiencarol, line 57:
```suggestion
```
If you want the parser to not use anymore PURL data, just remove it completely.

dojo/tools/anchore_grype/parser.py
```diff
@@ -70,3 +168,20 @@ def _convert_severity(self, val):
...
+            return vuln_id
+        elif rel_id and rel_id.startswith('CVE'):
+            return rel_id
+        else:
+            return vuln_id
+
+    def get_cvss(self, cvss):
+        if cvss:
+            for cvss_item in cvss:
+                if cvss_item['version'].startswith('3'):
```
damiencarol, line 183: Please consider using the package `cvss` to be aligned with other parsers.
https://defectdojo.github.io/django-DefectDojo/contributing/how-to-write-a-parser/#do-not-parse-cvss-by-hand-vector-score-or-severity

damiencarol (changes requested):

dojo/tools/anchore_grype/parser.py
StefanFl: The `purl` is still used as part of the description.

dojo/tools/anchore_grype/parser.py
StefanFl: The `cvss` item is a list of CVSS2 and CVSS3 entries. The parser needs to find the CVSS3 entry. I don't see how the cvss library helps to find this entry. We can test all cvss vectors with the cvss library, but just using the version number as provided in the json seems to be easier.

dojo/tools/anchore_grype/parser.py
StefanFl: done

dojo/tools/anchore_grype/parser.py
StefanFl: done

dojo/tools/anchore_grype/parser.py
StefanFl: done

dojo/tools/anchore_grype/parser.py
damiencarol: Using `cvss` module provide us a check of the vectors out-of-the-box which means that the parser could be able to see not-well formatted vectors when reading the report and not crash the upper layers (which check this data with a regex in Django model).

unittests/tools/test_anchore_grype_parser.py
```diff
@@ -86,3 +88,165 @@ def test_grype_parser_with_new_matcher_list(self):
...
+- http://www.trustix.org/errata/2004/0050
+- http://bugzilla.redhat.com/bugzilla/show_bug.cgi?id=136304
+- https://exchange.xforce.ibmcloud.com/vulnerabilities/17583
+- https://oval.cisecurity.org/repository/search/definition/oval%3Aorg.mitre.oval%3Adef%3A10497
+- https://lists.apache.org/thread.html/rc713534b10f9daeee2e0990239fa407e2118e4aa9e88a7041177497c@%3Cissues.guacamole.apache.org%3E'''
+        self.assertEqual(references, finding.references)
+        self.assertEqual('libgssapi-krb5-2', finding.component_name)
+        self.assertEqual('1.17-3+deb10u3', finding.component_version)
+        self.assertEqual('CVE-2004-0971', finding.vuln_id_from_tool)
+        self.assertEqual(['dpkg'], finding.tags)
```
damiencarol, line 128: better use:
```suggestion
        self.assertIn('dpkg', finding.tags)
```
More future proof if we add more tags

StefanFl: The idea is to check exactly what's in the test data. If someone changes the test data without the test case, I think it's ok when it fails.

dojo/tools/anchore_grype/parser.py
StefanFl: done

unittests/tools/test_anchore_grype_parser.py
damiencarol: ok

dojo/tools/anchore_grype/parser.py
```diff
@@ -70,3 +166,21 @@ def _convert_severity(self, val):
...
+        else:
+            return vuln_id
+
+    def get_cvss(self, cvss):
+        if cvss:
+            for cvss_item in cvss:
+                vector = cvss_item['vector']
+                cvss_objects = cvss_parser.parse_cvss_from_text(vector)
+                if len(cvss_objects) > 0 and type(cvss_objects[0]) == CVSS3:
+                    vector = cvss_item['vector']
```
damiencarol, line 184: Line 181 and 184 are the same.

StefanFl: Well spotted, I have removed the second line

damiencarol (approved):

devGregA (approved):

PR #5642, opened 2021-12-27
New parser for pip-audit

StefanFl:
To my knowledge DefectDojo currently doesn't support an open source dependency scanner for Python. pip-audit (https://github.com/trailofbits/pip-audit) seems to be a good candidate we should integrate.


dojo/tools/pip_audit/parser.py
```diff
@@ -0,0 +1,74 @@
...
+    def requires_file(self, scan_type):
+        return True
+
+    def get_findings(self, scan_file, test):
+
+        scan_data = scan_file.read()
+        try:
+            data = json.loads(str(scan_data, 'utf-8'))
+        except:
+            data = json.loads(scan_data)
```
damiencarol, line 26: Please use the simple form 
```suggestion
        data = json.loads(scan_file)
```

dojo/tools/pip_audit/parser.py
```diff
@@ -0,0 +1,74 @@
...
+                            title=title,
+                            cve=cve,
+                            cwe=1352,
+                            severity='Medium',
+                            description=description,
+                            mitigation=mitigation,
+                            component_name=component_name,
+                            component_version=component_version,
+                            static_finding=True,
+                            dynamic_finding=False,
```
damiencarol, line 70:
```suggestion
                            dynamic_finding=False,
                            vuln_id_from_tool=vuln_id,
```

dojo/tools/pip_audit/parser.py
```diff
@@ -0,0 +1,74 @@
...
+            data = json.loads(str(scan_data, 'utf-8'))
+        except:
+            data = json.loads(scan_data)
+
+        findings = list()
+        for item in data:
+            vulnerabilities = item.get('vulns', [])
+            if vulnerabilities:
+                component_name = item['name']
+                component_version = item.get('version', None)
```
damiencarol, line 33:
```suggestion
                component_version = item.get('version')
```
`None` is the default value.

dojo/tools/pip_audit/parser.py
```diff
@@ -0,0 +1,74 @@
...
+                        cve = None
+
+                    mitigation = None
+                    if vuln_fix_versions:
+                        mitigation = 'Upgrade to version:'
+                        if len(vuln_fix_versions) == 1:
+                            mitigation += f' {vuln_fix_versions[0]}'
+                        else:
+                            for fix_version in vuln_fix_versions:
+                                mitigation += f'\n- {fix_version}'
```
damiencarol, line 56: Why not using the form:
```
if 'fix_versions' in vulnerability:
```
It's easier to read and understand what the code do.

damiencarol: Also this code does not cover if the array is empty

dojo/tools/pip_audit/parser.py
```diff
@@ -0,0 +1,74 @@
...
+        scan_data = scan_file.read()
+        try:
+            data = json.loads(str(scan_data, 'utf-8'))
+        except:
+            data = json.loads(scan_data)
+
+        findings = list()
+        for item in data:
+            vulnerabilities = item.get('vulns', [])
+            if vulnerabilities:
```
damiencarol, line 31: We should generate an `Info` finding with only the data for the component.
It could be useful if there is no vulnerabilities to have at least the data for components.

unittests/tools/test_pip_audit_parser.py
```diff
@@ -0,0 +1,77 @@
...
+        self.assertEqual('PYSEC-2019-128 in twisted:18.9.0', finding.title)
+
+        finding = findings[4]
+        self.assertEqual('PYSEC-2020-260 in twisted:18.9.0', finding.title)
+
+        finding = findings[5]
+        self.assertEqual('PYSEC-2019-129 in twisted:18.9.0', finding.title)
+
+        finding = findings[6]
+        self.assertEqual('PYSEC-2020-259 in twisted:18.9.0', finding.title)
```
damiencarol, line 77: Please test the `vuln_id_from_tool`

damiencarol (changes requested):

damiencarol:
@StefanFl there is a unique Id in the reports, why not using this Id in the de-duplication algorithm?

dojo/tools/pip_audit/parser.py
StefanFl: I like to get the data for an item first and then process it. It covers the empty array, there is even one finding in the unit test cases for it (lxml).

dojo/tools/pip_audit/parser.py
StefanFl: I wouldn't like to use DefectDojo to store a BOM (bill of material), we should stick to manage vulnerabilities.

StefanFl:
> @StefanFl there is a unique Id in the reports, why not using this Id in the de-duplication algorithm?

I am not sure about the rules about how unique these PYSEC numbers are. A CVE can affect more than one component and/or more than one version of a component. Actually the `title` will be enough, because it contains id, component name and component title.

dojo/tools/pip_audit/parser.py
StefanFl: done

dojo/tools/pip_audit/parser.py
StefanFl: done

unittests/tools/test_pip_audit_parser.py
StefanFl: done

damiencarol:
@StefanFl I made some search about those Ids. I think we can use them as `unique_id_from_tool` and use them for de-duplication algorithm.
Those Ids are unique and they come from OSV database. So there are the same as our `unique_id_from_tool` attribute.

StefanFl:
According to https://ossf.github.io/osv-schema/ one id can have a list of affected packages. This is why I want to keep the `title` to calculate the hash_code, which has id, component name and component version.

damiencarol:
I think you're right. In the JSON schema the affected package could be different, so technically one vulnerability could affect few different package.

> This is why I want to keep the title to calculate the hash_code, which has id, component name and component version.

It's better to make it explicit in the de-duplication algorithm configuration. I mean relying on id+component_name+component_version instead of the relying on the title that is an aggregation and that could be changed in future release.

something like this:
```Python
'pip-audit Scan': ['vuln_id_from_tool', 'component_name', 'component_version'],
```

StefanFl:
Good idea, done

damiencarol (approved):

devGregA (approved):

PR #5645, opened 2021-12-27
Show descriptions for scan types on pages for import and re-import

StefanFl:
The pages to import and re-import show a list of available scan types. Currently it shows the name of the scan type 2 times next to each other:

2021-12-27 18_24_48-Import Scan Results _ DefectDojo

With this PR the list shows the name of the scan type with its description, which I think would have been the initial idea of that list:

2021-12-27 18_20_20-Import Scan Results _ DefectDojo

Additionally 3 descriptions have changed because they where wrong or outdated and a method has been removed that was not used anymore.


mtesauro (approved):
I am smacking myself that I didn't this of this earlier. Great idea. 

Approved.

damiencarol (approved):
Thanks @StefanFl . Yes, you're right, my idea to add a "description" in the parsers was to do exactly this kind of things.
Now this is a big list (more than 130 parsers...). It could be good to load it dynamically when we select the parser.

PR #5648, opened 2021-12-30
Display "Banner" on dashboard when skipping login form

CharlieSears:
### "Problem"
When the login screen is skipped for social login via `DD_SOCIAL_AUTH_SHOW_LOGIN_FORM` banners are never visible to users.

### Solution
Added a check if the login screen is displayed and banner is configured. When login screen has been skipped, and the banner has been configured, display the banner on the dashboard.

Added a new alert class `.alert-banner` to mirror the `.well` class used for login page banners.

### For Consideration
This implementation will _always_ display the banner anytime the user goes to the dashboard. This differs from the current once per form-based-login display of the banner. This implementation is based on the _assumption_ that when enabling a banner, one would want it to be displayed regularly to users until the banner is disabled.

### Images
Login Page
DefectDojoLoginBanner

Dashboard Addition
DefectDojoDashboardBanner

### DIFF
--- a/dojo/home/views.py
+++ b/dojo/home/views.py
@@ -4,6 +4,8 @@ from typing import Dict
 
 from dateutil.relativedelta import relativedelta
 
+from django.conf import settings
+from django.contrib import messages
 from django.urls import reverse
 from django.http import HttpResponseRedirect, HttpResponse, HttpRequest
 from django.shortcuts import render
@@ -11,7 +13,7 @@ from django.utils import timezone
 
 from django.db.models import Count, Q
 from dojo.utils import add_breadcrumb, get_punchcard_data
-from dojo.models import Answered_Survey
+from dojo.models import Answered_Survey, BannerConf
 from dojo.authorization.roles_permissions import Permissions
 from dojo.engagement.queries import get_authorized_engagements
 from dojo.finding.queries import get_authorized_findings
@@ -51,6 +53,16 @@ def dashboard(request: HttpRequest) -> HttpResponse:
             .filter(Q(engagement__isnull=True) | Q(engagement__in=engagements))
     else:
         unassigned_surveys = None
+
+    if not settings.SHOW_LOGIN_FORM:
+        banner_config = BannerConf.objects.get()
+        if banner_config and banner_config.banner_enable:
+            messages.add_message(
+                    request,
+                    messages.INFO,
+                    banner_config.banner_message,
+                    extra_tags="alert-banner",
+                )
 
     add_breadcrumb(request=request, clear=True)
     return render(request, 'dojo/dashboard.html', {
--- a/dojo/home/views.py
+++ b/dojo/home/views.py
@@ -17,6 +17,9 @@ from dojo.models import Answered_Survey, BannerConf
 from dojo.authorization.roles_permissions import Permissions
 from dojo.engagement.queries import get_authorized_engagements
 from dojo.finding.queries import get_authorized_findings
+import logging
+
+logger = logging.getLogger(__name__)
 
 
 def home(request: HttpRequest) -> HttpResponse:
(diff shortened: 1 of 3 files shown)


dojo/home/views.py
```diff
@@ -52,6 +54,16 @@ def dashboard(request: HttpRequest) -> HttpResponse:
     else:
         unassigned_surveys = None
 
+    if not settings.SHOW_LOGIN_FORM:
+        banner_config = BannerConf.objects.get()
```
StefanFl, line 58: `BannerConf.DoesNotExist` and `BannerConf.MultipleObjectsReturned` should be caught, in case the BannerConf object doesn't exist or there are more than one.

StefanFl (changes requested):
Thanks @CharlieSears for this PR. I have one small change request.

dojo/home/views.py
CharlieSears: Good call. I added some unit tests around the logic as well so it's not as fragile

StefanFl:
Great to see some unit tests. Best practice for unit tests is not to have any side effects. Using `override_settings` and `patch` would do this for your tests:

```python
    @override_settings(SHOW_LOGIN_FORM=True)
    @patch('dojo.models.BannerConf.objects.get')
    def test_banner_message_not_displayed_when_showing_login_form_even_if_banner_is_enabled(self, mock):
        mock.return_value = BannerConf(banner_enable=True, banner_message='This message should not be shown')

        response = self._request("user1")
        banner_messages = [m.message for m in get_messages(response.wsgi_request) if m.extra_tags == 'alert-banner']

        self.assertFalse(banner_messages)

    @override_settings(SHOW_LOGIN_FORM=False)
    @patch('dojo.models.BannerConf.objects.get')
    def test_banner_message_not_displayed_when_banner_is_disablled(self, mock):
        mock.return_value = BannerConf(banner_enable=False, banner_message='This message should not be shown')

        response = self._request("user1")
        banner_messages = [m.message for m in get_messages(response.wsgi_request) if m.extra_tags == 'alert-banner']

        self.assertFalse(banner_messages)

    @override_settings(SHOW_LOGIN_FORM=False)
    @patch('dojo.models.BannerConf.objects.get')
    def test_banner_message_not_displayed_when_banner_object_does_not_exist(self, mock):
        mock.side_effect = BannerConf.DoesNotExist()

        with self.assertLogs('dojo', DEBUG) as logManager:
            response = self._request("user1")
            banner_messages = [m.message for m in get_messages(response.wsgi_request) if m.extra_tags == 'alert-banner']

            self.assertFalse(banner_messages)
            logRecords = [record for record in logManager.records if record.levelno == DEBUG and
                record.message == 'Login form skipped and no banner is configured, nothing to display.']
            self.assertEqual(1, len(logRecords))

    @override_settings(SHOW_LOGIN_FORM=False)
    @patch('dojo.models.BannerConf.objects.get')
    def test_banner_message_not_displayed_when_multiple_banner_objects_exist(self, mock):
        mock.side_effect = BannerConf.MultipleObjectsReturned()

        with self.assertLogs('dojo', DEBUG) as logManager:
            response = self._request("user1")
            banner_messages = [m.message for m in get_messages(response.wsgi_request) if m.extra_tags == 'alert-banner']

            self.assertFalse(banner_messages)
            logRecords = [record for record in logManager.records if record.levelno == WARN and
                record.message == 'Login form skipped but multiple banner objects exist, which is unexpected.']
            self.assertEqual(1, len(logRecords))

    @override_settings(SHOW_LOGIN_FORM=False)
    @patch('dojo.models.BannerConf.objects.get')
    def test_banner_message_displayed_when_banner_is_enabled_and_login_form_skipped(self, mock):
        expectedMessage = 'This message should be displayed on the banner!'
        mock.return_value = BannerConf(banner_enable=True, banner_message=expectedMessage)

        response = self._request("user1")
        banner_messages = [m.message for m in get_messages(response.wsgi_request) if m.extra_tags == 'alert-banner']

        self.assertEqual(1, len(banner_messages))
        self.assertEqual(expectedMessage, banner_messages[0])
```

valentijnscholten:
It would be good to have multiple places dealing with getting the banner config from the database. Can we reuse the existing template tag? -> https://github.com/valentijnscholten/django-DefectDojo/blob/e1f89693ca4c59706bfec6634c1dbe6c6ac9a443/dojo/templatetags/get_banner.py#L9-L10

But I don't think it should be a banner that is shown permanently on the dashboard (as long as it is configured) If we want that, we might be better of creating a dedicated system wide banner as JIRA has.

Maffooch:
This is a good fix, but I agree with Valentijn. 

The login banner is intended to only be used/seen on the login page. The banner could definitely be missed due to a session existing before the login banner is created/updated.

I like the idea of a system wide banner, but it would need to be a separate from the login banner, and also in a separate PR.

CharlieSears:
Thank you all for not only the great feedback on getting unit tests in a better place, but the functionality.

If you'd all agree I'd be happy to approach this as a new setting for a system wide banner similar to Jira and closing this PR. That would allow more flexibility in notifications as well.

I started this path after the desire to present our users with some notifications of known delays or new changes.

Maffooch:
Hi @CharlieSears are you planning the systemwide banner changes in this PR, or a different one? Anything we can do to help guide you?

CharlieSears:
@Maffooch , I am planning those changes in a different PR. I'll close this one with this comment.

As far as direction, location I was thinking above the top nav elements entirely as opposed to within the content container.

Additionally, any thoughts on making the banner dismissible? I was on the fence, but now leaning towards allowing the banner to be dismissed to clean up the UI, I'm uncertain if that's the functionality JIRA has today or not.

PR #5652, opened 2021-12-30
Fix Report Nuclei file

pabloalexsandroalmeida83:
I created this fix, because Nuclei changed the report fields, besides including some fields to populate the "Finding" screen


damiencarol:
@pabloalexsandroalmeida83 we will need more checks in the unit tests

damiencarol (changes requested):
Need new checks in the unit tests

PR #5664, opened 2022-01-03
Fix Nuclei report and redirect import scan

pabloalexsandroalmeida83:
**Update on import of a Nuclei report.
Correct redirection after a report import**

Captura de tela de 2022-01-03 12-24-34


dojo/tools/nuclei/parser.py
```diff
@@ -24,12 +24,12 @@ def get_findings(self, filename, test):
         else:
             dupes = {}
             for item in data:
-                template_id = item.get('templateID')
+                template_id = item.get('template-id')
                 info = item.get('info')
```
Maffooch, line 28: This looks like a format change. Can you make both formats acceptable to maintain backward compatible? It would be best to keep the previous test scan as well as adding the new one to ensure that both formats continue to be supported.

dojo/engagement/views.py
```diff
@@ -641,7 +641,7 @@ def import_scan_results(request, eid=None, pid=None):
 
             if not error:
                 return HttpResponseRedirect(
-                    reverse('view_test', args=(test.id, )))
+                    reverse('product_open_findings', args=(pid, )))
```
Maffooch, line 644: I do not believe this is where imports should return to. In the case of a product having existing findings before an import, the user would have no idea which findings were imported in the most recent import than those that already existed.

Maffooch (changes requested):

dojo/tools/nuclei/parser.py
pabloalexsandroalmeida83: Yes, I can.

dojo/engagement/views.py
pabloalexsandroalmeida83: The problem that this page he redirects is unfinished, causing always error, what idea would you have for that? I could help with this task.

Maffooch: Can you elaborate on what unfinished means here? It has been set to redirect to the view test page for years without issue.

valentijnscholten: @pabloalexsandroalmeida83 what error are you seeing? The view_test page is like 100 years old, so should be working :-) Maybe fix the error in the parser and then see if the view_test page works.

valentijnscholten:
Please base your PR against the `dev` branch.

damiencarol (changes requested):
Please rebase on `dev` branch

StefanFl:
superseded by #6103

PR #5284, opened 2021-10-15
Fix: filesystem not to be writable by the defectdojo user

dsever:
- set directories to 755 but files to 644

Previously: 
```
drwxrwx--- 1 defectdojo root    4096 Oct  8 02:41 .
drwxr-xr-x 1 root       root    4096 Oct  8 11:04 ..
drwxrwx--- 1 defectdojo root    4096 Oct  8 11:04 dojo
-rwxrwx--- 1 defectdojo root     257 Oct  6 20:34 manage.py
drwxr-xr-x 3 defectdojo nogroup 4096 Oct  8 02:41 media
-rwxrwx--- 1 defectdojo root    1644 Oct  6 20:34 requirements.txt
drwxrwx--- 1 defectdojo root    4096 Oct  8 02:41 tests
-rwxrwx--- 1 defectdojo root     178 Oct  6 20:34 unit-tests.sh
-rwxrwx--- 1 defectdojo root    1423 Oct  6 20:34 wsgi.py
```
So user defectdojo can write to any of this file/folders, this is not good practice (related to immutability of docker containers, but explanation can be found here https://sysdig.com/blog/dockerfile-best-practices/#1-3 )

So with this PR user defectdojo (1001) can only read/execute files

```
drwxr-xr-x 1 root       root    4096 Oct 15 10:25 .
drwxr-xr-x 1 root       root    4096 Oct 15 10:25 ..
drwxr-xr-x 3 root       root    4096 Oct 15 10:25 docker
drwxr-xr-x 1 root       root    4096 Oct 15 10:21 dojo
-rwxr-xr-x 1 root       root     257 Jun 23  2020 manage.py
drwxr-xr-x 3 defectdojo nogroup 4096 Oct 15 10:28 media
-rw-r--r-- 1 root       root    1644 Oct 13 20:19 requirements.txt
drwxr-xr-x 1 root       root    4096 Oct 15 10:21 tests
-rwxr-xr-x 1 root       root     178 Aug  2 11:22 unit-tests.sh
-rw-r--r-- 1 root       root    1423 Jun 23  2020 wsgi.py
```

PR #5681, opened 2022-01-05
Updated django dockerfile to work with bind mounts for local_settings.py

mtesauro:
Recent change (Pull Request #5284) changed /app to be owned by root.  This broke the ability for docker-compose installs to bind mount a local_settings.py into the container.

To address this, I've added a defectdojo user and group with a UID and GID of 1337.  This allows /apps/dojo/settings to be writable by the defectdojo group so local_settings.py can be copied in by the entrypoint script.

I also moved the UID from 1001 to 1337 so that it has a much reduced chance of colliding with a OS user if you want to chown the host files that are bind mounted as a host OS user with a UID that matches the UID in the container.  1337 would require 337 local users vs 1 local user for 1001.

### DIFF
--- a/Dockerfile.django
+++ b/Dockerfile.django
@@ -144,4 +144,5 @@ ENV \
 ENTRYPOINT ["/entrypoint-uwsgi.sh"]
 
 FROM django as django-unittests
+USER root
 COPY unittests/ ./unittests/
--- a/Dockerfile.django
+++ b/Dockerfile.django
@@ -28,7 +28,8 @@ RUN pip3 wheel --wheel-dir=/tmp/wheels -r ./requirements.txt
 
 FROM base as django
 WORKDIR /app
-ARG uid=1001
+ARG uid=1337
+ARG gid=1337
 ARG appuser=defectdojo
 ENV appuser ${appuser}
 RUN \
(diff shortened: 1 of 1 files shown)


dsever:
@mtesauro I'm not sure this the the right way, to let entrypoint to change anything inside of container. From my point of view would be safer to bind directly into **settings** to preserve immutable nature, but I don't see nice way to do it using docker-compose (like conditional mount) or to use kind of template generator or ti use hack like this https://stackoverflow.com/questions/61952572/docker-compose-bind-volume-only-if-exists.

In k8s we are able to mount/overwrite single file, probably it is possible for docker-compose as well. 

From security perspective, if someone will break into app, can always dump env vars to gain access to sensitive part of appl. But what can do potentially include some code/module that will cause XSS, but need somehow force to reload app.

But to summarize, currently I don't see better way than you proposed.

Dockerfile.django
```diff
@@ -88,14 +89,21 @@ RUN \
...
+    addgroup --gid ${gid} ${appuser} && \
+    adduser --no-create-home --disabled-password \
+            --gecos 'DefectDojo' --uid ${uid} --gid ${gid} ${appuser} && \
     chown -R root:root /app && \
     chmod -R u+rwX,go+rX,go-w /app && \
+    # Allow for bind mounting local_settings.py and other setting overrides
+    chown root:${appuser} /app/dojo/settings && \
+    chmod 775 /app/dojo/settings && \
+    # Allow ${appuser} to run entrypoint scripts
+    chown root:defectdojo /entrypoint-* && \
```
dsever, line 102: Why here not root:root and to go with 755?

Dockerfile.django
```diff
@@ -88,14 +89,21 @@ RUN \
...
+    chmod g+x /entrypoint-* && \
     mkdir /var/run/${appuser} && \
-    chown ${appuser} /var/run/${appuser} && \
-	  chmod g=u /var/run/${appuser} && \
-    mkdir -p media/threat && chown -R ${uid} media
+    chown ${appuser}:${appuser} /var/run/${appuser} && \
+    mkdir -p media/threat && chown -R ${uid}:${gid} media
```
dsever, line 106: Can be here `chown -R ${appuser}:${appuser} media` to be consistent

Dockerfile.django
mtesauro: Since we have a defectdojo group (technically ${appuser} group), I thought it was more explicit to list the group as defectdojo with write permissions.  An `ls -lah /app/dojo/settings` will let you know that defectdojo is expected to write in this directory.  It's also the only non-root group owned directory so it's very explicit.

Dockerfile.django
mtesauro: That makes sense to me.  I considered it but was trying to keep the scope limited to just the settings/bind mount problem.  I'll go ahead and do a commit to add that in.  100% agree on your thinking.

PR #5682, opened 2022-01-05
Removal of `is_staff` for various functionalities

StefanFl:
This PR removes some checks of the `is_staff` flag, where the functionality can either be used by all users or gets restricted to superusers:

- Staff users where allowed to delete any notes with the API. This has been restricted to superusers.
- Bulk edits and deletes for Endpoints, Tests and Findings where restricted to staff users. Since the bulk operations respect the permissions of the user for each object, it is save to make it available to all users.
- There was an unnecessary check that only staff users can import Findings without providing an Engagement or Product. This would not work anyway, so an exception is raised in this case.
- Only staff users could be selected as the lead for an Engagement. This has been changed to all users, but only active users can be selected.
- System notifications where send to all superusers and all staff users. They now will be sent to superusers only.
- The request to generate a report without giving a valid object to specify the content of the report was permitted for staff users, but generated an exception later in the code. Now the exception is raised explicitly.
- `is_staff` was used to check if a user is logged in to answer a survey. This has been changed to check `is_authenticated`.

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -722,7 +722,7 @@ class FindingViewSet(prefetch.PrefetchListMixin,
         else:
             return Response({"error": "('note_id') parameter missing"},
                 status=status.HTTP_400_BAD_REQUEST)
-        if note.author.username == request.user.username or request.user.is_staff:
+        if note.author.username == request.user.username or request.user.is_superuser:
             finding.notes.remove(note)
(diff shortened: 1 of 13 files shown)


damiencarol (approved):

Maffooch (approved):
Not sure on what order you want stuff merged, so leaving it up to you :)

PR #5685, opened 2022-01-05
Edgescan integration

ShayVD:
# Context

This adds an Edgescan Parser and API Importer in a similar fashion to Sonarqube and Cobalt.io.

# Changes

Added Edgescan Parser, API, and Importer.
Added tests and test data.
Added documentation.

# Edit
Updated the Tool_Type initializer to be able to handle a `None` value from the Parser.


dojo/settings/settings.dist.py
```diff
@@ -1040,6 +1040,8 @@ def saml2_attrib_map_format(dict):
     'Azure Security Center Recommendations Scan': ['unique_id_from_tool'],
     'Solar Appscreener Scan': ['title', 'file_path', 'line', 'severity'],
     'pip-audit Scan': ['vuln_id_from_tool', 'component_name', 'component_version'],
+    'Edgescan Scan': ['title', 'severity', 'cve', 'cwe'],
+    'Edgescan API Scan': ['title', 'severity', 'cve', 'cwe'],
```
valentijnscholten, line 1044: Does severity need to be in here? Usually there will be no 2 findings with the same title, same cve, same cwe but different severity?

ShayVD: Good point. It has been reduced to just 'title' and 'severity'.

valentijnscholten: Why does severity need to be in there?

ShayVD: It is not always certain that our Findings will have a cve and cwe so I went with severity instead. Do you recommend removing severity altogether?

damiencarol:
@ShayVD it seems that the format is the same between the 2 parsers. It's possible to combine and have the same parser use the same code.

dojo/settings/settings.dist.py
ShayVD: This has been changed to just use `unique_id_from_tool`

ShayVD:
> @ShayVD
> 
> 1. it seems that the format is the same between the 2 parsers, is it correct? If yes, it's possible to combine both and have the same parser that use the same code.
> 2. are the Id in the file unique by project ? or is it by instance of Edgescan? if yes for one of the 2, it should be put in `unique_id_from_tool` instead of `vuln_id_from_tool`

@damiencarol 
You are correct about the parsers so I have merged them into one.
`unique_id_from_tool` is also now used as it is unique by project.

dojo/tools/edgescan/parser.py
```diff
@@ -0,0 +1,79 @@
...
+        finding.url = vulnerability["location"]
+        finding.severity = ES_SEVERITIES[vulnerability["severity"]]
+        finding.description = vulnerability["description"]
+        finding.mitigation = vulnerability["remediation"]
+        finding.active = True
+        finding.verified = True
+        finding.false_p = False
+        finding.duplicate = False
+        finding.out_of_scope = False
+        finding.numerical_severity = Finding.get_numerical_severity(ES_SEVERITIES[vulnerability["severity"]])
```
damiencarol, line 73:
```suggestion
```
This should not be implemented in the parser layer.

dojo/tools/edgescan/parser.py
```diff
@@ -0,0 +1,79 @@
...
+        finding.tags = vulnerability["asset_tags"]
```
damiencarol, line 74: There is no data in the unit tests covering tags. The data is always null. We will need a report with some tags and unit tests to cover this data.

unittests/tools/test_edgescan_parser.py
```diff
@@ -0,0 +1,53 @@
...
+            self.assertEqual(0, len(findings))
+
+    def test_parse_file_with_one_vuln_has_one_findings(self):
+        with open("unittests/scans/edgescan/one_vuln.json") as testfile:
+            parser = EdgescanParser()
+            findings = parser.get_findings(testfile, Test())
+            self.assertEqual(1, len(findings))
+            finding = findings[0]
+            testfile.seek(0)
+            payload = json.loads(testfile.read())[0]
```
damiencarol, line 24: This is not a good way to test. Please use fixed values.

unittests/tools/test_edgescan_parser.py
```diff
@@ -0,0 +1,53 @@
...
+            self.assertEqual(finding.title, payload["name"])
+            self.assertEqual(finding.date, payload["date_opened"][:10])
+            self.assertEqual(finding.cwe, int(payload["cwes"][0][4:]))
+            self.assertEqual(finding.cve, str(payload["cves"]))
```
damiencarol, line 28: I took a look at the data of the unit tests and the CVE are weird.
Like this:
image
CVE are more like `CVE-2021-44228` (see https://www.cvedetails.com/cve/CVE-2021-44228/)

dojo/tools/edgescan/parser.py
```diff
@@ -0,0 +1,79 @@
...
+        finding.unique_id_from_tool = vulnerability["id"]
+
+        finding.unsaved_endpoints = [Endpoint.from_uri(vulnerability["location"])]
+
+        return finding
```
damiencarol, line 79: It seems that there are CVSS vectors in the reports, you need to implement `cvssv3 ` with the vector
image

damiencarol (changes requested):

dojo/tools/edgescan/parser.py
ShayVD: Fixed in 387fc77. If the cvss version is v3 then it will extract the vector and score otherwise it is left blank.

ShayVD: If the cvss version is v3 then the vector and score are added otherwise, it is left blank.

unittests/tools/test_edgescan_parser.py
ShayVD: Fixed in 4977b14.

unittests/tools/test_edgescan_parser.py
ShayVD: Fixed in 4977b14.

dojo/tools/edgescan/parser.py
ShayVD: Fixed in 4977b14.

dojo/tools/edgescan/parser.py
ShayVD: Fixed in 387fc77.

dojo/tools/edgescan/parser.py
```diff
@@ -0,0 +1,79 @@
...
+            finding.cve = str(vulnerability["cves"])
+        finding.url = vulnerability["location"]
+        finding.severity = ES_SEVERITIES[vulnerability["severity"]]
+        finding.description = vulnerability["description"]
+        finding.mitigation = vulnerability["remediation"]
+        finding.active = True
+        finding.verified = True
+        finding.false_p = False
+        finding.duplicate = False
+        finding.out_of_scope = False
```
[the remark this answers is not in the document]
ShayVD: Fixed in 387fc77.

dojo/tools/edgescan/parser.py
```diff
@@ -0,0 +1,79 @@
...
+        finding.date = vulnerability["date_opened"][:10]
+        if vulnerability["cwes"]:
+            finding.cwe = int(vulnerability["cwes"][0][4:])
+        if vulnerability["cves"]:
+            finding.cve = str(vulnerability["cves"])
+        finding.url = vulnerability["location"]
+        finding.severity = ES_SEVERITIES[vulnerability["severity"]]
+        finding.description = vulnerability["description"]
+        finding.mitigation = vulnerability["remediation"]
+        finding.active = True
```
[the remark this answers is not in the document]
ShayVD: Fixed in 387fc77.

dojo/tools/edgescan/api_client.py
```diff
@@ -0,0 +1,78 @@
...
+            proxies=self.get_proxies(),
+        )
+
+        if response.ok:
+            return response.json()
+        raise Exception(
+            "Unable to get asset findings due to {} - {}".format(
+                response.status_code, response.content.decode("utf-8")
+            )
+        )
```
damiencarol, line 63: You have a better schema with `requests` module.
If you really want to throw an error if the API return a break, use `raise_for_status(...)` method  (https://fr.python-requests.org/en/latest/api.html?highlight=raise#requests.Response.raise_for_status)

dojo/tools/edgescan/parser.py
```diff
@@ -0,0 +1,80 @@
+import json
+
+from dojo.models import Endpoint, Finding
+from dojo.tools.edgescan.importer import EdgescanImporter
+
+ES_SEVERITIES = {1: "Info", 2: "Low", 3: "Medium", 4: "High", 5: "Critical"}
+SCAN_EDGESCAN_API = 'Edgescan API Scan'
+SCAN_EDGESCAN = 'Edgescan Scan'
```
damiencarol, line 8: Call them `SCANTYPE`

dojo/tools/edgescan/parser.py
```diff
@@ -0,0 +1,80 @@
...
+        finding.date = vulnerability["date_opened"][:10]
+        if vulnerability["cwes"]:
+            finding.cwe = int(vulnerability["cwes"][0][4:])
+        if vulnerability["cves"]:
+            finding.cve = vulnerability["cves"][0]
+        if vulnerability["cvss_version"] == 3:
+            if vulnerability["cvss_vector"]:
+                finding.cvssv3 = vulnerability["cvss_vector"][9:]
+            if vulnerability["cvss_score"]:
+                finding.cvssv3_score = vulnerability["cvss_score"]
```
damiencarol, line 68:
```suggestion
```
The score is automatically calculated in upper layers, just let the system calculate it.

damiencarol (changes requested):
Made some comments, it's a good quality PR.
Last request: don't use Exception but `ValueError`

dojo/tools/edgescan/api_client.py
ShayVD: Fixed in 5b04908.

dojo/tools/edgescan/parser.py
ShayVD: Fixed in 5b04908.

dojo/tools/edgescan/parser.py
ShayVD: Fixed in 5b04908.

ShayVD:
> Made some comments, it's a good quality PR. Last request: don't use Exception but `ValueError`

@damiencarol Thanks I appreciate the feedback!

dojo/tools/edgescan/api_client.py
```diff
@@ -0,0 +1,72 @@
...
+            self.options = self.get_extra_options(tool_config)
+        else:
+            raise Exception('Edgescan Authentication type {} not supported'.format(tool_config.authentication_type))
+
+    @staticmethod
+    def get_extra_options(tool_config):
+        try:
+            return json.loads(tool_config.extras)
+        except (JSONDecodeError, TypeError):
+            return {}
```
damiencarol, line 44: We are silently swallowing the errors which is not good

dojo/tools/edgescan/parser.py
```diff
@@ -0,0 +1,78 @@
...
+    def get_label_for_scan_types(self, scan_type):
+        return scan_type
+
+    def get_description_for_scan_types(self, scan_type):
+        if scan_type == SCANTYPE_EDGESCAN_API:
+            return "Edgescan findings can be imported by API."
+        return "Edgescan findings can be imported in JSON format."
+
+    def requires_file(self, scan_type):
+        return False if scan_type == SCANTYPE_EDGESCAN_API else True
```
damiencarol, line 28:
```suggestion
        return scan_type == SCANTYPE_EDGESCAN
```

dojo/tools/edgescan/parser.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def requires_tool_type(self, scan_type):
+        return 'Edgescan' if scan_type == SCANTYPE_EDGESCAN_API else None
```
damiencarol, line 31:
```suggestion
        return scan_type == SCANTYPE_EDGESCAN_API
```

unittests/tools/test_edgescan_parser.py
```diff
@@ -0,0 +1,64 @@
...
+        with open("unittests/scans/edgescan/one_vuln.json") as testfile:
+            parser = EdgescanParser()
+            findings = parser.get_findings(testfile, Test())
+            self.assertEqual(1, len(findings))
+            finding = findings[0]
+            self.assertEqual(finding.title, "Cross-site scripting (reflected)")
+            self.assertEqual(finding.date, "2014-12-05")
+            self.assertEqual(finding.cwe, 75)
+            self.assertEqual(finding.cve, "CVE-2021-5300")
+            self.assertEqual(finding.cvssv3, "AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:L/A:N")
```
damiencarol, line 25: The CVSS vector format is not good. The field need the complete form with the prefix (like `CVSS:3.0/AV:A/AC:H/PR:L/UI:R/S:C/C:N/I:L/A:N` **not** `AV:A/AC:H/PR:L/UI:R/S:C/C:N/I:L/A:N`)

damiencarol (changes requested):

Maffooch:
Tests should pass once you correct the unit tests you wrote per Damiens feedback.
Here is how to run and test them locally https://github.com/DefectDojo/django-DefectDojo/blob/master/readme-docs/DOCKER.md#run-the-tests-with-docker

dojo/tools/edgescan/parser.py
ShayVD: Fixed in d9d8bf7.

unittests/tools/test_edgescan_parser.py
ShayVD: Fixed in d377a80.

ShayVD:
Hey @damiencarol @valentijnscholten @Maffooch , sorry I've been busy for a couple of weeks. I've made the requested changes!

dojo/tools/edgescan/parser.py
```diff
@@ -0,0 +1,78 @@
...
+            if file:
+                serialized_data = file.read()
+                try:
+                    data = json.loads(str(serialized_data, "utf-8"))
+                except:
+                    data = json.loads(serialized_data)
+            else:
+                data = EdgescanImporter().get_findings(test)
+        except:
+            raise Exception("Invalid details provided")
```
damiencarol, line 44: Please don't try to manage by hand the encoding.

Just use `json.load(file)`
```suggestion
        if file:
            data = json.load(file)
        else:
            data = EdgescanImporter().get_findings(test)
```

dojo/settings/settings.dist.py
```diff
@@ -1091,6 +1093,8 @@ def saml2_attrib_map_format(dict):
     'Hadolint Dockerfile check': True,
     'Semgrep JSON Report': True,
     'Generic Findings Import': True,
+    'Edgescan Scan': True,
+    'Edgescan API Scan': True,
```
damiencarol, line 1097: It seems that loading from the API and getting data from a file load the same format/data.
Should we consider having only one scan_type?

damiencarol (changes requested):

dojo/tools/edgescan/parser.py
ShayVD: Fixed in 424bfaf.

dojo/settings/settings.dist.py
ShayVD: Fixed in 1bbd69a.

damiencarol (approved):

StefanFl (approved):

PR #5699, opened 2022-01-06
Removal of `AUTHORIZATION_STAFF_OVERRIDE`

StefanFl:
The functionality using the flag `AUTHORIZATION_STAFF_OVERRIDE` has been removed. The same result can be achieved with giving the staff users a global Owner role. To make that easier administrators can run a migration script with `./manage.py migrate staff_users`. This script creates a group for all staff users and sets the global Owner role, if `AUTHORIZATION_STAFF_OVERRIDE` is set to True.

The migration script already prepares the configuration permissions for staff users as well.


Maffooch (approved):

mtesauro (approved):
Approved

PR #5700, opened 2022-01-06
Revert "Updated all dockerfiles to match the django dockerfile so UID/GIDs are consistent across dockerfiles"

Maffooch:
Reverts DefectDojo/django-DefectDojo#5691

### DIFF
--- a/Dockerfile.django
+++ b/Dockerfile.django
@@ -27,15 +27,11 @@ COPY requirements.txt ./
 RUN pip3 wheel --wheel-dir=/tmp/wheels -r ./requirements.txt
 
 FROM base as django
-
 WORKDIR /app
-
-# Setup defectdojo user and group with 1337 UID/GID
 ARG uid=1337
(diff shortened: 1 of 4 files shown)


mtesauro (approved):
Approved

Maffooch:
Waiting until tests pass.

Maffooch:
K8's work now, but integration tests do not. They should work after reverting #5681 

Merging this now so the other can be reverted

dsever:
> K8's work now, but integration tests do not. They should work after reverting #5681
> 
> Merging this now so the other can be reverted

What is the problem with k8s?

Maffooch:
After #5691 was merged, K8 test did not pass any longer on the dev branch. Reverting #5691 made sure the unit tests for K8 deploys works again.

dsever:
Probably hadcoded djangosecuritycontext defined in helm values.

PR #5711, opened 2022-01-09
Rubocop: add parser

damiencarol:
SAST tool for Ruby

image

```
gem install rubocop
rubocop -f json > report.json
```

See:
 * https://rubocop.org/
 * https://docs.rubocop.org/


StefanFl (commented):
Hi @damiencarol, I have 2 comments:

- Do you known the meaning of the attribute `corrected`? Is it possible that this attribute is `True` and there should be no Finding?
- The parser should have an entry under `DEDUPLICATION_ALGORITHM_PER_PARSER` with `DEDUPE_ALGO_HASH_CODE` in `settings.dist.py`. Otherwise reimports might not work properly.

damiencarol:
@StefanFl I will add the entry in the settings for de-duplication.

=> For the `corrected` flag, I made some tests and I'm unable to produce them. When I fix a finding, they just disappear from the report.

StefanFl (approved):

Maffooch (approved):

PR #5713, opened 2022-01-09
Configuration authorization 3: Finish making more configuration permissions editable

StefanFl:
This is a rather large PR, sorry about that. But it is the last part to make configuration permissions editable. 

2022-01-09 17_46_40-View User _ DefectDojo

There are 3 areas left which can only edited by superusers: System Settings, Notifications and the Configuration Permissions.

One notable point: Credentials had a functionality about Selenium, which was partially removed. Now it's completely removed.


dojo/forms.py
```diff
@@ -3177,19 +3217,28 @@ def __init__(self, *args, **kwargs):
...
+            Permission_Helper(name='tool configuration', app='dojo', view=True, add=True, change=True, delete=True),
             Permission_Helper(name='tool type', app='dojo', view=True, add=True, change=True, delete=True),
             Permission_Helper(name='user', app='auth', view=True, add=True, change=True, delete=True),
         ]
 
-        self.permission_fields = permission_fields_1 + questionnaire_permissions + permission_fields_2
+        self.permission_fields = permission_fields_1 + github_permission + google_sheet_permission + \
+                                 permission_fields_2 + jira_permissions + permission_fields_3 + \
+                                 questionnaire_permissions + permission_fields_4 + rules_permissions + \
+                                 permission_fields_5
```
Maffooch, line 3241: Can we append to a single large list of `Permissions_Helper` objects rather than creating many small lists and then adding them all together at the bottom? Would look much cleaner

Maffooch:
This is very large, but it does look solid

dojo/forms.py
StefanFl: @Maffooch Does this look better now?

valentijnscholten (approved):

Maffooch (approved):

PR #5737, opened 2022-01-13
Bugfix SLA days remaining

psbelin:
The SLA days remaining are currently calculated using the Finding's date, even if the field `sla_start_date` is set. It's curious, because the message that appears when you move the mouse over the SLA number is well calculated using the `sla_start_date`. But the number of the SLA days was being calculated differently. It seems like a bug that would be fixed with this PR.
With this change, the SLA will be calculated using the date from sla_start_date when it's available.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -2216,7 +2216,7 @@ class Finding(models.Model):
         from dojo.utils import get_system_setting
         sla_age = get_system_setting('sla_' + self.severity.lower())
         if sla_age:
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

StefanFl (approved):
Makes sense to me. Anyone else wants to have a look?

damiencarol:
@StefanFl I'm not sure of the impacts it could have. Somebody that know better the metrics part should take a look at this.

PR #5743, opened 2022-01-13
added some logic changes, attempts to speed dashboard loading a little

nobletrout:
made some logic changes and reordering of some code
some reformatting


dojo/product/views.py
```diff
@@ -1052,7 +1052,11 @@ def edit_technology(request, tid):
                                  'Technology changed successfully.',
                                  extra_tags='alert-success')
             return HttpResponseRedirect(reverse('view_product', args=(technology.product.id,)))
+        else:
+            logger.debug('delete_product: POST INVALID FORM')
+            logger.error(form.errors)
```
valentijnscholten, line 1057: just having an error on a form does not warrant the logging at the highest severity level. should be debug.

dojo/product/views.py
```diff
@@ -1379,6 +1382,11 @@ def delete_engagement_presets(request, pid, eid):
                                      'Engagement presets and engagement relationships removed.',
                                      extra_tags='alert-success')
                 return HttpResponseRedirect(reverse('engagement_presets', args=(pid,)))
+            else:
+                logger.debug('delete_product: POST INVALID FORM')
+                logger.error(form.errors)
```
valentijnscholten, line 1387: same

dojo/product/views.py
```diff
@@ -1667,7 +1674,11 @@ def edit_product_group(request, groupid):
                     return HttpResponseRedirect(reverse('view_group', args=(group.group.id, )))
                 else:
                     return HttpResponseRedirect(reverse('view_product', args=(group.product.id, )))
+        else:
+            logger.debug('delete_product: POST INVALID FORM')
+            logger.error(form.errors)
```
valentijnscholten, line 1679: same

dojo/settings/settings.dist.py
```diff
@@ -301,6 +301,8 @@ def generate_url(scheme, double_slashes, user, password, host, port, path, param
         }
     }
 
+# https://docs.djangoproject.com/en/4.0/ref/settings/#std:setting-CONN_MAX_AGE
+DATABASES['default']['CONN_MAX_AGE'] = 60
```
valentijnscholten, line 305: introducing database pooling should probably be in a PR of itself as it's a major thing. at 60s it probably won't do much except for high traffic sites.

dojo/settings/template-local_settings
```diff
@@ -58,3 +58,10 @@ DEBUG_TOOLBAR_PANELS = [
 
 import debug_toolbar
 EXTRA_URL_PATTERNS = [url(r"^__debug__/", include(debug_toolbar.urls))]
+
+# debug toolbar doesn't work without this enabled for local dev
+if DEBUG:
+    import os  # only if you haven't already imported this
+    import socket  # only if you haven't already imported this
+    hostname, _, ips = socket.gethostbyname_ex(socket.gethostname())
+    INTERNAL_IPS = [ip[:-1] + '1' for ip in ips] + ['127.0.0.1', '10.0.2.2']
```
valentijnscholten, line 67: works fine for me without this. where does the 10.0.2.2 come from? I think the internal ips is meant on how to identify the clients that are allowed to see the toolbar. so not sure this is related to the hostname or ip of defect dojo itself.
it might have the adverse effect that it's not shown for people using 192.168.x.x or similar.

dojo/test/queries.py
```diff
@@ -5,21 +5,25 @@
 from dojo.authorization.authorization import get_roles_for_permission, user_has_global_permission
 
 
-def get_authorized_tests(permission, product=None):
+def get_authorized_tests(permission, product=None, testID=None):
```
valentijnscholten, line 8: what is this param for? please rename it to a pythonic way, i.e. test_id

dojo/test/views.py
```diff
@@ -70,7 +70,8 @@ def view_test(request, tid):
         available_note_types = find_available_notetypes(notes)
     files = test.files.all()
     person = request.user.username
-    findings = Finding.objects.filter(test=test).order_by('numerical_severity')
+    findings = Finding.objects.filter(test_id=tid).order_by('numerical_severity')\
+        .defer('url', 'description', 'impact', 'steps_to_reproduce', 'severity_justification', 'references', 'payload', 'line', 'file_path', 'files')
```
valentijnscholten, line 74: I don't think this really helps when showing a couple of findings. If in the future someone wants to show these fields it will result in extra queries.

dojo/utils.py
```diff
@@ -663,8 +663,12 @@ def get_punchcard_data(objs, start_date, weeks, view='Finding'):
                                         .values('date') \
                                         .annotate(count=Count('id')) \
                                         .order_by('date')
+        else:
+            raise ValueError("view parameter only supports values 'Finding' or 'Endpoint'")
+
         # return empty stuff if no findings to be statted
-        if severities_by_day.count() <= 0:
+        # don't make an additional sql query, use len()
+        if len(severities_by_day) <= 0:
```
valentijnscholten, line 671: severities_by_day isn't queried to count() is faster. len will result in loading all the findings into memory, creation of Finding objects etc which is slower.

valentijnscholten (changes requested):
thanks for the PR, but left some remarks. I'm not a fan of slamming different stuff into a big PR, could you look at seperating out the database pooling into a PR and the debug toolbar stuff into a PR (not sure we need/want that as mentioned).

dojo/test/queries.py
StefanFl: I don't like that parameter, the queries shall be as universal as possible. And I can't see in this PR where it shall be used.

dojo/utils.py
```diff
@@ -1505,20 +1509,16 @@ def __init__(self, product_id, title=None, tab=None):
         self.product = Product.objects.get(id=product_id)
         self.title = title
         self.tab = tab
-        self.engagement_count = Engagement.objects.filter(
-            product=self.product, active=True).count()
-        self.open_findings_count = Finding.objects.filter(test__engagement__product=self.product,
+        self.engagement_count = Engagement.objects.filter(product=self.product, active=True).count()
+        self.open_findings_count = Finding.objects.filter(test__engagement__product=product_id,
```
StefanFl, line 1513: Why did you switch to `product_id`, when other selects still use `self.product`?

dojo/home/views.py
```diff
@@ -23,28 +23,26 @@ def home(request: HttpRequest) -> HttpResponse:
...
+        .only('severity', 'id', 'created', 'mitigated').prefetch_related('risk_acceptance_set')
 
     engagement_count = engagements.filter(active=True).count()
 
     today = timezone.now().date()
+    beginning_date = today - timedelta(days=6)
+
+    finding_count = findings.filter(created__gte=beginning_date, created__lte=today).order_by().count()
+
+    mitigated_count = findings.filter(mitigated__gte=beginning_date, mitigated__lte=today).only('id').order_by().count()
```
StefanFl, line 38: Does `.only('id').order_by()` make sense when we only count the mitigated findings?

dojo/home/views.py
```diff
@@ -23,28 +23,26 @@ def home(request: HttpRequest) -> HttpResponse:
...
-    findings = findings.filter(duplicate=False)
+    findings = get_authorized_findings(Permissions.Finding_View).filter(duplicate=False).distinct()\
+        .only('severity', 'id', 'created', 'mitigated').prefetch_related('risk_acceptance_set')
 
     engagement_count = engagements.filter(active=True).count()
 
     today = timezone.now().date()
+    beginning_date = today - timedelta(days=6)
+
+    finding_count = findings.filter(created__gte=beginning_date, created__lte=today).order_by().count()
```
StefanFl, line 36: Does `.order_by()` make sense when we only count the findings?

dojo/home/views.py
```diff
@@ -23,28 +23,26 @@ def home(request: HttpRequest) -> HttpResponse:
...
-    finding_count = findings\
-        .filter(created__date__range=date_range)\
-        .count()
-    mitigated_count = findings\
-        .filter(mitigated__date__range=date_range)\
-        .count()
-    accepted_count = findings\
-        .filter(risk_acceptance__created__date__range=date_range)\
-        .count()
+    accepted_count = findings.filter(risk_acceptance__created__gte=beginning_date, risk_acceptance__created__lte=today).only('id').order_by().count()
```
StefanFl, line 40: Does `.only('id').order_by()` make sense when we only count the accepted findings?

dojo/home/views.py
StefanFl, line 36: Is `__gte`  and `__lte` faster than `range`?

StefanFl (changes requested):

Maffooch:
Great feedback from Val and Stefan. Nothing more to add from my side

PR #5744, opened 2022-01-14
added asynchronous deletion for engagements, tests, and products.

nobletrout:
This can take a long time to process depending on database size and
scoping and this makes it much less painless with the tradeoff that you
might see data disappear over time. Ideally a "pending deletion" flag
would be set in the top level engagement, product, or test..


dojo/engagement/views.py
```diff
@@ -289,13 +290,16 @@ def delete_engagement(request, eid):
...
-                engagement.delete()
-                messages.add_message(
-                    request,
-                    messages.SUCCESS,
-                    'Engagement and relationships removed.',
-                    extra_tags='alert-success')
-                create_notification(event='other',
+                if form.cleaned_data['asynchronous_delete']:
+                    dojo.tasks.asynch_delete.delay(Engagement, engagement.id)
+                else:
```
valentijnscholten, line 295: Will this work if the checkbox is not ticked? I remember unticked checkboxes not being sent?
Probably a message should still be added saying the deletion was succesfully queued?
Please update the unit tests to reflect these changes.

valentijnscholten (changes requested):
see comments

Maffooch:
I love this idea, but I wonder if making this a System Setting would be better. My biggest gripe with the delete page is how long the GET request takes to render with larger datasets. Waiting to click the async button until after the page is full rendered does not save the user much time.

valentijnscholten:
And the api would need to support async deletes too.

nobletrout:
I can work on that. The get taking forever is also my biggest gripe but I
was going to see what this reception was like first.

How about two system settings:
1. No dangerzone
2. Asynchronous deletes default ?

I’d like to have a “I know what I am doing here and I don’t want you tell
me about all the findings related to things just delete the stuff please
thank you”

nobletrout:
making some notes:

It appears I'm not the only person and this is not the only project that suffers from this issue:

https://code.djangoproject.com/ticket/10919

Looks like Django is now supporting something called ASGI or planning it? I'm not a django expert
https://docs.djangoproject.com/en/4.0/howto/deployment/asgi/

Maffooch:
Im nervous about the no danger zone setting. A simple misclick could totally ruin somebodies day. Another approach that would could incorporate the async delete would be to display the danger zone, but do not query the DB for the objects that are about to be deleted. That would allow for a fast danger zone, and scheduled delete that could take a while.

I have no idea if that is possible, but could be worth exploring

valentijnscholten:
> Looks like Django is now supporting something called ASGI or planning it? I'm not a django expert https://docs.djangoproject.com/en/4.0/howto/deployment/asgi/

That is about non-blocking IO which is not the same as doing something async.

> Another approach that would could incorporate the async delete would be to display the danger zone, but do not query the DB for the objects that are about to be deleted. That would allow for a fast danger zone, and scheduled delete that could take a while.

The danger zone could be 2 steps. 

1) Display the danger zone page with only text. Two buttons: a) display related items to be deleted b) yolo, just delete all the things
2) optional page which displays the related items similar to what we have currently -> confirm -> delete all the things

mtesauro:
I think @valentijnscholten 's 2 step approach is the best I've seen so far on this thread.  Allows for confirmation and YOLO if you want or details if that's more your thing.

damiencarol:
I agree. Sometime you really want to delete a product and everything behind (when you test, etc...) and sometime you really want to see what ill be deleted.
Maybe we could just give hint like the number of objects that the confirmation will delete.

Like `"You will destroy 56798 object(s), do you confirm this action?"` is not the same as `"You will destroy 2 object(s), do you confirm this action?"` :D

nobletrout:
Hey guys, I tried to get the integration/unit tests working on this PR. But I'm getting unrelated failures (first something about a JIRA check failing and now test_login). Can someone please help me figure out a fast track to running unit tests?
Thanks

Maffooch:
Here are some docs on running the unit tests https://github.com/DefectDojo/django-DefectDojo/blob/master/readme-docs/DOCKER.md#run-the-tests-with-docker

nobletrout:
It appears that #6183 duplicates this effort and has been merged in. woohoo!

PR #5745, opened 2022-01-14
Kubernetes hostname changes

nobletrout:
fixed my fork, made git close pr #5559 sorry.


alles-klar:
@dsever I suggest we continue with this PR because it fixes the problem from @nobletrout and doesn't change to much. I agree with you that there is room for improvement, but maybe we continue with it in later PR, when someone need the feature.

alles-klar (approved):

dsever:
@alles-klar, problem I see here is that PR breaks upgrades. Previously there was def as: `key:value` and now is `key: list`.  

At least it has to be documented.

alles-klar:
@nobletrout I added an upgrade note, but I think we must also update the [kubernetes instructions](https://github.com/DefectDojo/django-DefectDojo/blob/master/readme-docs/KUBERNETES.md). Can you please take a look and test/update it?

dsever:
Guys finally I had time to test it in minikube, sorry about that (my bad). So few comments from my side, it works as expected, that is good. 

But I'm thinking about something else, this change introduces many changes but actually it changes only  `DD_ALLOWED_HOSTS:` it is not touching items like liveness probes, and it break changes (if someone has configured CICD way of working). Addition to that `hosts` is list, and for liveness probes and TLS certs it is taking first element:

`{{- $hostName := index .Values.hosts 0}}` 

So if order is not respected ti will break the deployment. 

Maybe would be better to introduce new value like e.g  `alternative_hosts` that will be for DD_ALLOWED_HOSTS translated to:
`host` + `alternative_hosts` (list) 

So user will not need to break/adapt current installation, not pay attention of the order, but required functionality will be still there.

dsever (commented):
Please take a look at my last comment.

nobletrout:
Makes sense I was just trying to avoid adding an additional value.

helm/defectdojo/templates/configmap.yaml
```diff
@@ -1,4 +1,8 @@
 {{- $fullName := include "defectdojo.fullname" . -}}
+{{- $hosts := .Values.host -}}
```
dsever, line 2: Can you move it to tpl?

dsever:
Now it looks good to me, I don't think  integration tests that fails has anything to do with your changes.

dsever:
@nobletrout can you try to rebase it.

dsever:
Am I right, you did rebase against master instead of dev?

nobletrout:
ugh, looks like i did. the github UI isn't very clear about how that sort of stuff works.

nobletrout:
git you make everything very hard.

dsever (approved):
It looks ok now

helm/defectdojo/values.yaml
```diff
@@ -20,7 +20,14 @@ createPostgresqlSecret: false
 # Option to use "postgresql" or "mysql" database type, by default "mysql" is chosen
 # Set the "enable" field to true of the database type you select (if you want to use internal database) and false of the one you don't select
 database: postgresql
+# List of allowed hosts, first entry is used for http header
```
alles-klar, line 23: This comment is not correct anymore. It isn't a list. Can you please update it?

nobletrout: done

PR #5756, opened 2022-01-15
Hide "Staff" flags in UI and API plus documentation of configuration permissions

StefanFl:
This PR does 3 things:

- Hiding the **Staff** flag when viewing or editing of users, if `FEATURE_CONFIGURATION_AUTHORIZATION` is set to `True`.
- Documentation of configuration permissions
- Removing a few `is_staff` flags that have been overseen before (in `dojo/templates/base.html` and ` dojo/templates/dojo/rules.html`)

This is the final PR for configuration permissions and `is_staff` removal for the release 2.7.0. It is feature complete, but I maybe its better to leave the default of `FEATURE_CONFIGURATION_AUTHORIZATION` set to `False` for now. Then nothing changes for the users for now, unless they change the flag it. Or shall we get live with it right away?


valentijnscholten (changes requested):
Might be best to leave it off by defaul tin 2.7.0 to allow ourselves to test it in real life? Also, should we add some docs about `staff users` and `superusers` and what they mean and don't mean in the new setup?

StefanFl:
> Might be best to leave it off by defaul tin 2.7.0 to allow ourselves to test it in real life? Also, should we add some docs about staff users and superusers and what they mean and don't mean in the new setup?

It is off by default at the moment exactly for that reason 😃 I will turn it on in my productive installation to see how it works. 

I would like to change the documentation anyway, when the configurable permissions will be the default. Then the staff flag won't have no meaning anymore and won't be visible in the UI or the API. At that point I will add some integration tests as well.

valentijnscholten:
I know not everyone agrees, but I think we should allow people to access the `/admin` interface, which needs the staff flag?

StefanFl:
> I know not everyone agrees, but I think we should allow people to access the /admin interface, which needs the staff flag?

I agree with you for superusers. For them it might make sense to access the /admin interface. Not so much for non-superusers, for them the admin interface is empty anyway.

I would change this PR to make sure a superuser is always staff as well. Would that work for you @valentijnscholten?

valentijnscholten:
yeah, was thinking about that as well.

StefanFl:
With the latest commit superusers will always be staff as well.

valentijnscholten:
Should it be added to `initialize_permissions` or some migration that does a one time updates of superusers? Will it also work if someone creates a superuser via `manage.py`?

StefanFl:
>Should it be added to initialize_permissions or some migration that does a one time updates of superusers? 

That would only be needed if an instance already has superusers which are not staff. They would have a lot problems in the current version because of all the `is_staff` flags. With the configuration permissions this problems will be gone anyway. If there is a case, you can do a save of that superusers and he will be made staff by the signal. The only difference will be if a superuser can access the admin interface. I don't see a migration necessary for this case.

>Will it also work if someone creates a superuser via manage.py?

Yes it does, I have tested it. The signals work there as well.

valentijnscholten (approved):

Maffooch (approved):

PR #5757, opened 2022-01-15
DependencyCheck maintenance

damiencarol:
## Work done

- [x] Fixes 2 bugs/useless code:

1. Call to method DependencyCheckParser.get_field_value with too few arguments; should be no fewer than 3.
https://github.com/DefectDojo/django-DefectDojo/blob/76175aca446e077884bdb5e1d8e2a671a0840775/dojo/tools/dependency_check/parser.py#L92

`get_field_value` needs 3 arguments.

2. The logic in severity part is odd

https://github.com/DefectDojo/django-DefectDojo/blob/76175aca446e077884bdb5e1d8e2a671a0840775/dojo/tools/dependency_check/parser.py#L227-L233

Line 228

- [x] add real report from a recent version (6.5.3)
- [x] reverted back title management


valentijnscholten:
The changes to the `title` field I implemented a while ago are not so good in hindsight. Too long. Do you have an idea on how to improve them?

damiencarol:
> The changes to the title field I implemented a while ago are not so good in hindsight. Too long. Do you have an idea on how to improve them?

I will take a look. I generated more recent report, maybe new data will help this part.

> Also I don't think we should remove the example report sections (unless they are from very old versions). I've found them very helpful while implementing the parser. Maybe they should be strores somewhere else, but without good examples it's hard to maintain the parser.

It's exactly what I shared with you in PM. I'm pushing to have more **real** file reports. IT will provide both (help to implement new features in the parser and good examples)

StefanFl:
> The changes to the title field I implemented a while ago are not so good in hindsight. Too long. Do you have an idea on how to improve them?

I like a simple "[cve] in [component_name]:[component:version]"

damiencarol:
@StefanFl @valentijnscholten the date have a simple `name` attribute. We should rely on this only.
like this:
image
We already have the component+version in another attribute

damiencarol:
It gives something like that
image

StefanFl:
I wouldn't need the brackets and not the NPM bit as well. In other parsers we started using the tag for the source, very nice in filters.

valentijnscholten:
I think that is what it used to be before. But only having the CVE, CWE or npm advisory id in the title is not very descriptive. I tried to make it more descriptive, but I think it became very ugly because of that. So definitely needs some changes. Problem seems to be that the CVEs don't have a title field we can use. But maybe the npm advisories do have it somewhere as they seem to have a title field: https://github.com/advisories?query=type%3Areviewed+ecosystem%3Anpm (github powers npm advisories since a while)

damiencarol:
I took a look at NPM data in the reports and it's a mess.
The title field is in the references...
```
<references>
  <reference>
    <source>Advisory 1006884: Inefficient Regular Expression Complexity in marked</source>
    <name>- https://github.com/markedjs/marked/security/advisories/GHSA-rrrm-qjm4-v8hf
- https://nvd.nist.gov/vuln/detail/CVE-2022-21680
- https://github.com/markedjs/marked/commit/c4a3ccd344b6929afa8a1d50ac54a721e57012c0
- https://github.com/markedjs/marked/releases/tag/v4.0.10
- https://github.com/advisories/GHSA-rrrm-qjm4-v8hf</name>
  </reference>
</references>
```

damiencarol:
@StefanFl @valentijnscholten data seems inconsistent between sources (NPM/Java/Python). So any logic we add will break the title part.
I really think we should just take the `name` attribute of the vulnerability as-is.
If you're ok with that I can simplify the logic behind the title to just keep the `name` and put the source as a tag.

valentijnscholten:
That's fine. Still think that looking for the npm advisory title in the references and trying to include it would be informative.

StefanFl:
`name` would just be `CVE-2020-9488` (from the sample scan), correct? Other dependency parser have the component name and version in it, I personally like that.

valentijnscholten:
if there is no title, it makes perfect sense to include the component name and maybe version. otherwise you just have CVE which means you have nothing. I suggest we revert back to what it was before my PR which changed the title.

StefanFl:
I agree with @valentijnscholten

damiencarol:
ok, will make some test and try your suggestion.
we will see. But I really think the best move is to stop creating fake data (like aggregating few attribute to make the UI feels better) because every time a tool modify the format of the report it breaks the parsers.

valentijnscholten:
I don't understand why you think this is fake data. The CVE is also a field that is strictly speaking not the title. And already shown in another column. What do you suggest, leave the title empty? I think we're wasting our time and should just revert to the old title before my before which made everyone happy.

StefanFl:
Hi @damiencarol, do you want to make a proposal for the title in the code and we will go from there?

StefanFl:
@damiencarol Title is good for me now. It's always a compromise, this is a good one.

While you're at it, what do you think about 2 cosmetic changes? I like it when keywords in description or references are bold, that makes it easier to distinguish them from the actual content. This would be:

- Line 211: `description += '\n**Filepath:** ' + str(dependency_filepath)`
- Lines 204 - 206: 
```
                reference_detail += '**Name:** {0}\n' \
                                    '**Source:** {1}\n' \
                                    '**URL:** {2}\n\n'.format(name, source, url)
```

StefanFl:
References are weirdly formatted, but I am sure that's not the fault of the parser. I will investigate that.

damiencarol:
Foe references I can also take a look. Maybe make a raw block

StefanFl:
Found the bug why the references formatting is not working. Definitely not related to this parser, could happen with every other parser as well, dependent on the data. I will fix it.

damiencarol:
Ok. good to know. cosmetic changes broke the unit tests, I need to update them

StefanFl:
Can you make name, source and url start with capitals?

damiencarol:
@StefanFl done. And I added more checks around severity values like `Critical`

dojo/tools/dependency_check/parser.py
```diff
@@ -261,11 +209,11 @@ def get_finding_from_vulnerability(self, dependency, related_dependency, vulnera
 
         else:
             mitigation = 'Update {}:{} to at least the version recommended in the description'.format(component_name, component_version)
-            description += '\nFilepath: ' + str(dependency_filepath)
+            description += '\n**Filepath**: ' + str(dependency_filepath)
```
StefanFl, line 212:
```suggestion
            description += '\n**Filepath:** ' + str(dependency_filepath)
```

I know I am picky, just looks a bit nicer.

dojo/tools/dependency_check/parser.py
```diff
@@ -239,18 +187,18 @@ def get_finding_from_vulnerability(self, dependency, related_dependency, vulnera
             reference_detail = ''
             for reference_node in references_node.findall(namespace +
                                                           'reference'):
-                name = self.get_field_value(reference_node, 'name', namespace)
-                source = self.get_field_value(reference_node, 'source', namespace)
-                url = self.get_field_value(reference_node, 'url', namespace)
-                reference_detail += 'name: {0}\n' \
-                                     'source: {1}\n' \
-                                     'url: {2}\n\n'.format(name, source, url)
+                name = reference_node.findtext(f"{namespace}name")
```
StefanFl, line 190: I have a suggestion for the references: 

```python
            for reference_node in references_node.findall(namespace + 'reference'):
                ref_source = reference_node.findtext(f"{namespace}source")
                ref_url = reference_node.findtext(f"{namespace}url")
                ref_name = reference_node.findtext(f"{namespace}name")
                if ref_url == ref_name:
                    reference_detail += f'**Source:** {ref_source}\n' \
                                        f'**URL:** {ref_url}\n\n'
                else:
                    reference_detail += f'**Source:** {ref_source}\n' \
                                        f'**URL:** {ref_url}\n' \
                                        f'**Name:** {ref_name}\n\n'
```

* There was a bug, because the variable `name` is reused here and overwrites the variable carrying the name of the vulnerability, which is later used for the title.
* Preserves the order of the 3 attributes as they are in the XML report.
* `URL` is an abbreviation and is written in capitals in English.
* In a real life example I used to test it, the `URL` and the `name` where the same a few times, so the references don't repeat the same data here anymore. Not completely sure if it may confuse people when the `name` is missing sometimes.

dojo/tools/dependency_check/parser.py
```diff
@@ -10,172 +11,127 @@
...
         if dependency_filename is None:
             return None
 
-        name = self.get_field_value(vulnerability, 'name', namespace)
-        cwes_node = vulnerability.find(namespace + 'cwes')
-        if cwes_node is not None:
-            cwe_field = self.get_field_value(cwes_node, 'cwe', namespace)
+        tags = []
+        name = vulnerability.findtext(f'{namespace}name')
+        cwes_node = vulnerability.find(f'{namespace}cwes')
```
StefanFl, line 126: The variable `cwes_node` is unused, so this line can be deleted.

StefanFl (changes requested):
Thank you @damiencarol, this PR is a huge improvement of the parser, including the `title`.

There is one more thing. The report has the source of the vulnerability in `<vulnerability source="NVD">` It would be great to save this information in the description, like:

> 
> Included in Log4j 1.2 is a SocketServer class that is vulnerable to deserialization of untrusted data which can be exploited to remotely execute arbitrary code when combined with a deserialization gadget when listening to untrusted network traffic for log data. This affects Log4j versions up to 1.2 up to 1.2.17.
> 
> **Source:** NVD
>
> **Filepath:** .m2/repository/log4j/log4j/1.2.7/log4j-1.2.7.jar

dojo/tools/dependency_check/parser.py
damiencarol: no you're right

dojo/tools/dependency_check/parser.py
damiencarol: we should use it to set the CWE to the first value.

StefanFl:
@damiencarol As discussed I made a few changes, please have a look at them to confirm these are ok for you.

damiencarol:
I'm good with them so far.

StefanFl (approved):

damiencarol:
@valentijnscholten we made the change to the tile, could you review again?

valentijnscholten (approved):

PR #5760, opened 2022-01-16
First step to internationalization 

shipko:
Hi guys!

I want to add localization in project. 
Why it need?
This will make it possible to translate the interface on other language. If company needs interface on other language - it can generate locale file. 
On first step I decided prepare base.html file. If my pull request will be interesting for you, I continue work


damiencarol:
@shipko  Don't you need a localization file to match the strings to localize?

devGregA:
Hi @shipko nice to meet you! I've always wanted to pursue multilingual support. Wouldn't it be better to start with the models rather than the templates?

shipko:
@damiencarol on first step it's not require. Django will use words wrapped in translatization function. 

If somebody wants to use other language can execute django-admin makemessages -l <your locale>, replace LANGUAGE_CODE in settings.dist.py and translate language file

shipko:
@devGregA of course. This not last commit. After that I will start edit models.py

shipko:
@devGregA wrapped models.py to translatization form.

damiencarol:
wow. this will be a big DB migration. (every model that have some description)

shipko:
That's all @devGregA @damiencarol 
I edited base.html and models.py
For the beginning I think that completely done first step for i18n

Also i notice that no required migration, because we changed 'verbose_name' and 'help_text' parameters.

Please approve my PR or write that I need to fix. Because I don't want to resolve  merge conflicts in soon

damiencarol:
@shipko could you add some languages files? I want to test this PR

shipko:
@damiencarol of course!
This is [test branch](https://github.com/shipko/django-DefectDojo/tree/translate-for-damien) where I added some required files for testing.
This files changed only some elements in left bar menu. Another elements should  be translated in /dojo/locale/ru/LC_MESSAGES/django.po file.

I want to notice, that this files from test branch will not be merged in dev branch.

shipko:
@damiencarol you have news?

damiencarol:
I'm still testing. I need to catch-up with the team as it will impact the whole UI.

valentijnscholten:
I would expect that we would define text keys where we use texts, i.e. `menu.engagements.all` and then we would have language files that contain the actual texts for these keys. I think if we do just `{% trans "Engagements by Product" %}` it becomes error prone when somebody changes texts in templates, but not in the language files. So would expect something like {% trans "menu.engagements.all" %}

damiencarol:
@valentijnscholten I think it's good to have English as key. switching to "menu.engagement.all" will block us to use string interpolation in some string. I think it's a good first step.
@shipko @valentijnscholten I'm not a fan of renaming the `gettext()` function to `_`. If you're ok, I prefer to keep `gettext()`

damiencarol:
@shipko please add a first PO file to show contributors how it could be done to add a language.

shipko:
@damiencarol 
If I correct understand, you want to replace `_()` to `gettext()` equivalent. 
I used `_()` function to simplify construction. Because this function is will be widely use. When we need pluralization - we will be use `ngettext`.

I understand your logic - it more consistently, but we have much more code in you variant. Aren't we will lost the code clarity?

valentijnscholten:
Using gettext seems very verbose, Django itself suggests _() in the docs

damiencarol:
@shipko ok.

shipko:
@damiencarol  @devGregA  @valentijnscholten please leave feedback about this pr. I constantly resolve merge conflicts. It's boring

valentijnscholten:
I have no experience with i18n, especially not in Django. I guess if this PR follows the [docs](https://docs.djangoproject.com/en/4.0/topics/i18n/translation/) it should be OK.
One thing I am wondering is how it would work if a string has variables/parameters/interpolation.
Something like the example in the docs:
image
But strangely enough I can't find how that would actually work in the `.po` file.

devGregA:
@shipko  looking at this now

devGregA:
@Maffooch would you also please review?

damiencarol:
It's good enough for me as a first step. I made a couple of tests by adding a new language. I will accept it before the end of the week if everything is ok.

Maffooch (approved):
I've done a quick performance test with this to be sure making translations via the templates would not be too impacting. Django's template loader can be super slow with certain operations. 

I found that there really is no discernible difference, and that this is the standard way of doing django internalization. Approving for now

shipko:
@Maffooch thanks!

@damiencarol @valentijnscholten @devGregA what about you?

valentijnscholten (approved):

PR #5772, opened 2022-01-18
feat(helm): add ingressClassName value to ingress resource

KarstenSiemer:
Hello,

this PR will introduce the ingressClassName to the ingress template.
Docs on this: https://kubernetes.io/docs/concepts/services-networking/ingress/

I added a semverCompare since it is available beginning k8s 1.18 and only on v1 of ingress resource and we also make this check for deciding what apiGroup to use.

if `.Values.django.ingress.ingressClassName` is either non-existent or empty, the key will not appear in ingress resource

### DIFF
--- a/helm/defectdojo/templates/django-ingress.yaml
+++ b/helm/defectdojo/templates/django-ingress.yaml
@@ -28,6 +28,11 @@ metadata:
   {{- end }}
 {{- end }}
 spec:
+{{- if semverCompare ">=1.19.0" .Capabilities.KubeVersion.GitVersion }}
+   {{- if .Values.django.ingress.ingressClassName }}
+   ingressClassName: {{ .Values.django.ingress.ingressClassName }}
+   {{- end }}
+{{- end }}
 {{- if .Values.django.ingress.activateTLS }}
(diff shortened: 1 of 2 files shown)


KarstenSiemer:
`Lint Helm chart` fails because of some unrelated error

dsever:
@valentijnscholten ?

valentijnscholten:
Please revise against dev branch

KarstenSiemer:
@valentijnscholten Hi, did the changes. Thanks :-)

valentijnscholten:
thanks, no need to bump the chart version in dev. better not to actually.

KarstenSiemer:
@valentijnscholten okay, reverted the change and cleaned up commit history

dsever:
@KarstenSiemer  This looks nice, I wanted to push the same, to avoid post patching (as we do now).

valentijnscholten (approved):
approving on behalf of @dsever

alles-klar (approved):

alles-klar:
Thanks @KarstenSiemer

PR #5789, opened 2022-01-20
Bugfix: For Snyk parser, preserve file paths with @ in package name 

SafeEval:
The Snyk parser uses the `file_path` field to show dependency paths. Many package names include ampersands in them, which are not being preserved due to truncation when trimming out version numbers. This PR preserves the packages names in the `file_path` field.

### Before
snyk-parser-before

### After
snyk-parser-after


valentijnscholten (approved):
lgtm, thanks!

damiencarol (approved):

PR #5797, opened 2022-01-21
Add support for including multiple setting files via extra_settings

Maffooch:
I have found that only a single settings file is acceptable for copying a settings file from the host machine in `docker/extra_settings` to the container.

This PR allows for multiple files to be copied over at once via bound mounts

### DIFF
--- a/docker/entrypoint-celery-beat.sh
+++ b/docker/entrypoint-celery-beat.sh
@@ -11,32 +11,6 @@ do
 done
 echo
 
-# Allow for bind-mount setting.py overrides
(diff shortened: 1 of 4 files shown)


valentijnscholten:
Should the pre existing pieces that copy only a single file be removed? They will also be copied by the new code I think?

damiencarol (approved):

valentijnscholten (approved):

mtesauro (approved):
Approved

mtesauro:
Cool idea - I like it :+1:

PR #5799, opened 2022-01-21
Facelift: borders for panels with lists

StefanFl:
In some views there are separate panels, where some of them have borders and others, with lists and paginations, don't have borders or partially have borders. One example is the list of Findings in the Test views:

2022-01-21 17_22_24-Test _ DefectDojo

With this PR the user interface looks more consistent, having borders around every block, when there are multiple blocks on one page. This is the list of Findings in a Test with the PR:

2022-01-21 17_20_01-Test _ DefectDojo

These dialogues have been touched (all in the Product view):
- Engagement list (lists of active / paused / closed Engagements)
- Engagement details (list of Tests)
- Test details (list of Findings)
- Endpoint / Host details (list of Findings)

### DIFF
--- a/dojo/static/dojo/css/dojo.css
+++ b/dojo/static/dojo/css/dojo.css
@@ -593,6 +593,11 @@ form ul#id_accepted_findings input {
   border-color: #546474;
 }
 
+.pagination-in-panel {
+    margin-left: 10px!important;
+    margin-right: 10px!important;
+}
+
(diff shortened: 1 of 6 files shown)


mtesauro (approved):
So glad you have UI skills - I've spent too much time in vi and the terminal (plus, I am just not good at it)

Approved.

valentijnscholten (approved):

PR #5809, opened 2022-01-22
Rendering of references is corrupted when the same URL occurs more than one time

StefanFl:
When there is the same URL more than one time in the references of a finding, the rendering of the references produced weird results. That's becauce the first URL in the references got replaced multiple times. With this PR only one occurrence of a URL gets rendered with a proper HTML link, all others stay untouched.

Someone could imagine a more sophisticated solution, but it's a rare case and there is at least one link for each URL to click on.


valentijnscholten (approved):

mtesauro (approved):
This approach makes sense to me.

Approved

PR #5815, opened 2022-01-24
Modified file (jira_link/helper.py) to add a label to the Jira indica…

techylinux:
Tried to modified the below python script (jira_link/helper.py) to add a label to the Jira indicating the source branch. And yeah we build the new change and any Jira ticket created gets labeled respective to their branch name (the report which is being imported from source).Tested locally and it is indeed working find and label is getting added automatically which sort of eliminates extra manual work (bulk edit).It's painful to modify this piece of code each time we upgrade to new release. So, why not to implement this directly in DD?

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -476,14 +476,25 @@ def log_jira_message(text, finding):
 def get_labels(obj):
     # Update Label with system setttings label
     labels = []
+ 
+    # Put a branch tag in as a label.
+    if type(obj) == Finding and obj.test.branch_tag:
+        labels.append(obj.test.branch_tag.replace(" ", "_"))
+ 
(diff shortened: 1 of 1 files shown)


techylinux:
HI Team my bad noticed a strange behavior with this change only reports which are imported from JFrog the corresponding Jira tickets are getting created while for Sonar-qube and ZAP scan, Jira is not getting created and getting the following error
Screenshot 2022-01-25 124723
:

Closing this PR need to investigate more on this.

damiencarol:
it seems that if the tags are `None` it break the helper. could be easy to fix.

techylinux:
> it seems that if the tags are `None` it break the helper. could be easy to fix.

techylinux:
I am reopening instead of creating again new one..I will investigate on this until this can be kept on hold

PR #5816, opened 2022-01-24
Add PostgreSQL as an option for Docker Compose

StefanFl:
This PR introduces the options to use PostgreSQL as a database and Redis as a celery broker. The profiles are completely flexible, 4 profiles are available:

- mysql-rabbitmq
- mysql-redis
- postgres-rabbitmq
- postgres-redis

Downside of it all: You cannot start docker compose anymore so quickly but you have to set parameters for profiles and environment variables. To cover this, there are 4 shell scripts:

- `dc-build.sh`: Builds the docker images.
- `dc-up.sh`: Starts the containers in the foreground, you can supply 1 or 2 profiles
- `dc-up-d.sh`: Starts the containers in the background, you can supply 1 or 2 profiles
- `dc-down.sh`: Stops the containers when started in the background, you have to supply the first profile

Users have to pass 1 or 2 parameters to the `up` scripts, but the first profile can be set by the environment variable `DD_PROFILE` as well.

As @mtesauro pointed out, you need a quite recent version of docker. Service profiles have been introduced with Docker Compose version 1.28.0 on 2021-01-20.

Thanks to @mtesauro for his work on profiles, which made this PR a lot better during development.


Maffooch:
Wow this looks like it turned out to be way more work than expected. 

Here is some comments/answers to your questions:
- Does not seem like there is a better option besides have comments everywhere. @mtesauro what are your thoughts?
- The `.env` file was used a while ago for setup.bash (the original way of installing dojo before docker)
- I am not sure on this one, but I would assume it would be a no. I was unable to find any evidence that it is supported.

StefanFl:
The comments are not a great solution. They are still there for `.env` and `docker-compose.yml`. For the override files I have chosen now a different solution. PGAdmin can be started with a profile now and so the override files need not to be changed anymore when using PostgreSQL.

mtesauro:
I've looked at this PR about 4 times over the last couple of days and keep thinking, there's a better way.  I generally like what's here.  It does make the compose.yml files a bit more cluttered/complex but not too much. After re-looking at the changed files, I had a bit of an idea:

@StefanFl @Maffooch What if we use profiles + environmental variables to all us to support all 4 supported composes in a single docker-compose?

The 4 supported composes are:
1. MySQL + RabbitMQ
2. Postgres + RabbitMQ
3. MySQL + Redis
4. Postgres + Redis

So, if we do profiles, we can have a single compose start with one of those 4 options by either adding an `--profile mysql` argument to `docker-compose up -d` or use `COMPOSE_PROFILES` env var to select a profile.  I figured out and tested 2 different ways to achieve that.

**Pros**
- Single compose handles all the supported ways to run DD
- Easy to switch to a different configuration with ENV variables
- Add a conditional to the [script](https://github.com/DefectDojo/django-DefectDojo/blob/master/docker/setEnv.sh) to setup compose so it takes a profile as an command argument and sets a default if none is provided

**Cons**
- We'd have to bump the file version of docker-compose from 3.7 to 3.8
- Requires a rather recent version of docker-compose to support 3.8

I have a working version of what I'm taking about in [this Gist](https://gist.github.com/mtesauro/f10b1252e0a7958c6e465dcfb1f98d43) - the composes in the Gist were simplified to the most essential bits.  I have a full working version but the ENV vars, build, volumes just clutter things up so were removed.

If we collectively like the idea, I'm very willing to flesh it out fully including the additions to setEnv.sh  I can even include a version with pgadmin container like is in Stephan's PR if that's useful.

Thoughts?

Maffooch:
I had never seen those profiles before! Very cool. 

I think it could work for this case here

mtesauro:
@StefanFl The depends_on cheats were my first idea - and yeah, it's pretty ugly but it works ;-)

I've attached a full example - build stanzas are still removed so up'ing is quick. Containers fixed at v 2.6.2 - I also created empty directories and touched some files to avoid errors for bind mounts (see tree command below). Well, looks like I can't attach a yaml directly so I just tarballed the entire directory: [profile-example.tar.gz](https://github.com/DefectDojo/django-DefectDojo/files/7980854/profile-example.tar.gz)
 
Notes:
- There's potential work in combining the volumes for RabbitMQ and Redis into a single volume (since both won't exist at the same time).  I've not tested that yet.  
- Without setting a profile, you get no running containers, just a couple of volumes created. 
- I didn't worry about setting ENV vars correctly when testing as I figured if the containers are launched, we've proved the profiles work. Any required messing with ENV vars can come later.

One last thing, I created two throwaway scripts to try out the various profiles - we'd need to do something 'more formal' to setEnv.sh if we go this route.  They are copy/pasted below for your convenience.

"up" script
```
#/bin/bash

# Placeholder to avoid errors - likely won't work
export DD_DATABASE_URL="mysql://defectdojo:defectdojo@database:3306/defectdojo"

echo "Can also run 'docker-compose --profile $1 up -d'"
echo ""
echo "Starting with profile [$1]..."

COMPOSE_PROFILES="$1" docker-compose up -d
```

"down" script
```
#!/bin/bash

docker-compose down
```
Example run with 'up' script: (container names simplified)
```
$ ./up postgres
Can always set a DD_PG_URL and DD_MYSQL_URL for testing/automation

Can also run 'docker-compose --profile postgres up -d'

Starting with profile [postgres]...

Creating network "profile-example_default" with the default driver
Creating celeryworker ... done
Creating rabbitmq     ... done
Creating uwsgi        ... done
Creating postgres     ... done
Creating init         ... done
Creating nginx        ... done
Creating celerybeat   ... done
$
```

Directory structure:
```
$ tree ./
./
├── db
│   └── docker-entrypoint-initdb.d
├── docker
│   └── extra_settings
├── docker-compose.yml
├── down
└── up
```

StefanFl:
I continued on @mtesauro's approach and now the profiles are completely flexible and there is no need anymore to comment/uncomment anything. 4 profiles are available:

- mysql-rabbitmq
- mysql-redis
- postgres-rabbitmq
- postgres-redis

Downside of it all: You cannot start docker compose anymore so quickly but you have to set parameters for profiles and environment variables. To cover this there are 4 shell scripts:

- `dc-build.sh`: Builds the docker images.
- `dc-up.sh`: Starts the containers in the foreground, you can supply 1 or 2 profiles
- `dc-up-d.sh`: Starts the containers in the background, you can supply 1 or 2 profiles
- `dc-down.sh`: Stops the containers when started in the background, you have to supply the first profile 

And as @mtesauro pointed out before, you need a quite recent version of docker. Service profiles have been introduced with Docker Compose version 1.28.0 on 2021-01-20.

To be done: The `docker/setEnv.sh` script doesn't stops the containers properly at the moment.

.gitignore
```diff
@@ -81,7 +81,7 @@ dojo/uploads/threat/*
 *.sqlite
 *.db
 celerybeat.pid
-*.env*
+./*.env*
```
valentijnscholten, line 84: I use this `dojo/settings/.env.prod` in my local environment, which now becomes unignored. Not sure if this is bad or can be avoided. I think we should try to avoid accidentally commiting "prod credentials" into git.

StefanFl: I changed it back to how it was and added an exception rule for the docker environment files

valentijnscholten:
Now that we have this, we could also run the integration tests against redis/postgres. Do we want that?

StefanFl:
Not sure what to do about tests to be honest. Testing with 2 different configurations would be great, but the checks take a long time already and I am afraid we will see more unintended fails in the integration tests.

valentijnscholten:
It's quite easy to run 2 instances of the integration tests, or extend the matrix. We can try? I think in the past we have seen something that didn't work on postgres, not sure what it was. Possibly a (data) migration of some sorts.

StefanFl:
A lot of times one integration test fails because of timing issues. That started when integration tests started running in parallel. We already fire up 172 docker containers in parallel. I am afraid starting 344 containers in parallel will make things worse.

valentijnscholten:
The issues with the integration tests have been presented for at least 2 years and we should fix those to be honest.

StefanFl:
Another thing: I found out that you can install pgAdmin easily locally. Therefore I would take it out of the docker-compose files because they are complex enough. Would anyone mind if I remove it?

valentijnscholten:
> https://github.com/DefectDojo/django-DefectDojo/blob/9be7b81b39ea72c533655d52141a748f4cdfaedc/tests/dedupe_test.py#L95
> 
> is one candidate that seems to fail more often than others. @valentijnscholten can you have a look at it?
> But there where problems with 502 errors while pulling images and k8s deployments as well.

That wait was the solution to a long lasting timing issue in this spot about 2 years or so back. But that was more a client side timing issue as the datatable is added later via javascript. Seems now that dedupe by celery is just not completed in time, but haven't looked at it. Not sure if testing deduplication makes sense via the UI tests.

The 502 and other errors are just some GHA problems I think we're seeing lately.
The k8s problem is also a long lasting problem that sometimes the health check doesn't work, or defect dojo doesn't start correctly or there is some (virtual) networking issue preventing the health check / up check to reach defect dojo in the cluster.

mtesauro:
> Another thing: I found out that you can install pgAdmin easily locally. Therefore I would take it out of the docker-compose files because they are complex enough. Would anyone mind if I remove it?

Nope.  I think a pgAdmin container is a convenience we can leave out.  If people want/need it, I agree that it's easy enough to run it locally.

StefanFl:
This time the 5th run of the integration tests was successful. Failures of the first 4 runs:

Run 1: 
- test_test.py / test_add_and_delete_stub_finding / mysql-rabbitmq
- dedupe_test.py / test_delete_findings / postgres-redis

Run 2:
- test_test.py / test_add_and_delete_stub_finding / postgres-redis
- dedupe_test.py / test_delete_findings / postgres-redis

Run 3:
- finding_test.py / test_delete_finding / postgres-redis

Run 4:
- group_test.py / setUpClass / postgres-redis (The process started from chrome location /usr/bin/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)

Very unstable, 9 out of 10 runs where failing with different reasons 😞

StefanFl:
I checked some of the fails of integration tests after merges, it is normally one of these 3:
- test_test.py / test_add_and_delete_stub_finding
- dedupe_test.py / test_delete_findings
- finding_test.py / test_delete_finding

On the other hand, 2 out 2 runs of sequential integration tests did run properly now. Maybe just a coincidence, maybe not.

Maybe it would make sense to think about something in between, where 4 or 6 sets of tests run in parallel.

mtesauro:
So the non-testing side of this looks good to me.

@StefanFl Has your thinking changed on the tests from your last comment ^?

valentijnscholten:
Looking at the logs it seems that every test is running independently on (its own?) runner:
`Job is about to start running on the hosted runner: GitHub Actions 15 (hosted)`
But still it seems that if we spin up more tests, the chances of failures increase.

And some of the failures I haven't seen before like the `test_test.py / test_add_and_delete_stub_finding` and `dedupe_test.py / test_delete_findings`. 

If we only run 4-6 tests in parallel, it might actually be slower sometimes compared to sequential. Every "batch" has to build the containers and start them before running the tests. This can take a couple minutes if there are cache misses during building. Also the initializer takes some time. So might be easier/faster to just run them sequentially. 

What about running them both sequentially AND in parallel to make full comparison? Probably will lead to more errors maybe? Or maybe the sequential ones still work fine?

valentijnscholten:
Looking at the raw logs I'm seeing some suspicious stuff. Possibly unrelated, still would be good to have not so many errors in the logs:

`2022-02-10T11:19:59.3155995Z [35;1muwsgi_1              |[0m /entrypoint-uwsgi-dev.sh: 10: [: True: unexpected operator`

and

```
2022-02-10T11:19:59.3166601Z [35;1muwsgi_1              |[0m Can't listen for client connections: [Errno 98] Address already in use
2022-02-10T11:19:59.3166959Z [35;1muwsgi_1              |[0m Traceback (most recent call last):
2022-02-10T11:19:59.3167314Z [35;1muwsgi_1              |[0m   File "/app/./dojo/wsgi.py", line 49, in <module>
2022-02-10T11:19:59.3167659Z [35;1muwsgi_1              |[0m     debugpy.listen(("0.0.0.0", debugpy_port))
2022-02-10T11:19:59.3168083Z [35;1muwsgi_1              |[0m   File "/usr/local/lib/python3.8/site-packages/debugpy/__init__.py", line 113, in listen
2022-02-10T11:19:59.3168511Z [35;1muwsgi_1              |[0m     return api.listen(address)
2022-02-10T11:19:59.3168932Z [35;1muwsgi_1              |[0m   File "/usr/local/lib/python3.8/site-packages/debugpy/server/api.py", line 143, in debug
2022-02-10T11:19:59.3169488Z [35;1muwsgi_1              |[0m     log.reraise_exception("{0}() failed:", func.__name__, level="info")
2022-02-10T11:19:59.3169926Z [35;1muwsgi_1              |[0m   File "/usr/local/lib/python3.8/site-packages/debugpy/server/api.py", line 141, in debug
2022-02-10T11:19:59.3170329Z [35;1muwsgi_1              |[0m     return func(address, settrace_kwargs, **kwargs)
2022-02-10T11:19:59.3170771Z [35;1muwsgi_1              |[0m   File "/usr/local/lib/python3.8/site-packages/debugpy/server/api.py", line 243, in listen
2022-02-10T11:19:59.3171160Z [35;1muwsgi_1              |[0m     raise RuntimeError(str(endpoints["error"]))
2022-02-10T11:19:59.3171590Z [35;1muwsgi_1              |[0m RuntimeError: Can't listen for client connections: [Errno 98] Address already in use
2022-02-10T11:19:59.3171955Z [35;1muwsgi_1              |[0m subprocess 47 exited with code 1
```

```
2022-02-10T11:18:19.4282314Z [36minitializer_1        |[0m ls: cannot access '/app/docker/extra_settings/*.py': No such file or directory
2022-02-10T11:18:19.4309132Z [36minitializer_1        |[0m ============================================================
2022-02-10T11:18:19.4309622Z [36minitializer_1        |[0m      Overriding DefectDojo's local_settings.py with multiple
2022-02-10T11:18:19.4309989Z [36minitializer_1        |[0m      Files: 
2022-02-10T11:18:19.4310355Z [36minitializer_1        |[0m ============================================================
2022-02-10T11:18:19.4327951Z [36minitializer_1        |[0m cp: cannot stat '/app/docker/extra_settings/*.py': No such file or directory
```

StefanFl:
Current test setup is running all MySQL tests in parallel and after that all PostgreSQL tests in parallal. Should at least not be worse than before concerning failures. But takes longer of course.

mtesauro:
@StefanFl Looks like the testing issues are sorted out.  Would you agree?

valentijnscholten:
We can experiment with the testing suites in another PR if needed. But running the two scenarios sequentially completely destroys all the timing benefits we achieved earlier by parallelizing the tests. I'd rather run the two big scenario's in parallel and all the tests inside them sequentially. Or for now skip the postgresql instance of the suite.

StefanFl:
@mtesauro, @valentijnscholten I have implemented another version now: The profile gets selected randomly. Now we don't have more tests than before but we still make sure integration tests run for MySQL and PostgreSQL. This should be a good compromise.

valentijnscholten (changes requested):
Sorry, but I don't like it. Tests should be as predictable/reproducible as possible.

StefanFl:
Then I would suggest we use the fully parallelized integration tests. With the changes of @valentijnscholten they seem to be reliable enough. If more problems occur, we can change it in a different PR.

valentijnscholten (approved):

mtesauro (approved):
Approved

PR #5820, opened 2022-01-25
Modified the helper.py code 

techylinux:
This is with reference to previous PR where after doing some more testing found that JIRA ticket was not getting created in case of JFROG and ZAP scan. Now, the things are working fine.

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -484,10 +484,7 @@ def get_labels(obj):
     system_settings = System_Settings.objects.get()
     system_labels = system_settings.jira_labels
     if system_labels is None:
-        if len(labels) == 0:
-            return
-        else:
-            return labels
+        return labels
     else:
         system_labels = system_labels.split()
     if len(system_labels) > 0:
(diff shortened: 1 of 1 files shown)


dojo/jira_link/helper.py
```diff
@@ -476,14 +476,22 @@ def log_jira_message(text, finding):
...
+        return labels
+    else:
         system_labels = system_labels.split()
+    if len(system_labels) > 0:
         for system_label in system_labels:
             labels.append(system_label)
-        # Update the label with the product name (underscore)
-        labels.append(prod_name(obj).replace(" ", "_"))
+    # Update the label with the product name (underscore)
+    labels.append(prod_name(obj).replace(" ", "_"))
```
valentijnscholten, line 494: this is skipped if there are no system labels, might be better to always add it

techylinux: Sure Valentijn. But what I found that with this code change either the labelling is done automatically if the report contains the branch information or else while importing new report if we mention branch tag, the labels are added for all those findings in Jira which was not possible with the default helper.py code. With the default code only in description (field containing branch ) tagging was happening.

image

valentijnscholten: I understand, but my comment is about the product name not being added to the labels if there are no system labels. I think product name should always be added as label, as that was the intended behaviour I think. Currently this also doesn't happen, but if we are working on this code we might as well fix that.

techylinux: Sure will work with one of my colleague to fix this.

PR #5823, opened 2022-01-25
Make unit test for rest framework more stable

StefanFl:
The unit test for the rest framework fails in some constellations with PostgreSQL. I am not completely sure why, but the adaptions in this PR make the assertions more straightforward and the test doesn't fail anymore.


damiencarol (approved):

mtesauro (approved):
Approved

PR #5830, opened 2022-01-27
Async Import: Fix group_by flag

Maffooch:
Add `user` kwarg to finding group assignment helper method to all for the `group_by` flag to work during async imports. The requirement for a kwarg is because the `get_current_user()` crum function does not return anything when the context is async

### DIFF
--- a/dojo/decorators.py
+++ b/dojo/decorators.py
@@ -22,7 +22,7 @@ def we_want_async(*args, func=None, **kwargs):
         logger.debug('dojo_async_task %s: running task in the foreground as sync=True has been found as kwarg', func)
         return False
 
-    user = get_current_user()
+    user = kwargs.get('async_user', get_current_user())
     logger.debug('user: %s', user)
 
     if Dojo_User.wants_block_execution(user):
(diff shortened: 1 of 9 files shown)


damiencarol (approved):

mtesauro (approved):
Approved

valentijnscholten:
Not sure if this will scale as every piece of code doing something with the current user is affected. Might be for now it's only this case. Is the audit log affected? Ideally the current user is passed on to a celery task when it's created. This was something I looked at in the past, something like this: https://stackoverflow.com/questions/46792488/passing-user-object-in-celery-using-django

Maffooch:
Yeah that could be a good idea. Could be added in as kwarg. Could be added here https://github.com/DefectDojo/django-DefectDojo/blob/f9b802abb659a14b12ad83f28c5b68dead267dc4/dojo/decorators.py#L25

Maffooch:
@valentijnscholten I added the changes to the decorator, but it required a lot more changes, I think it should be more scalable from here though as we can rely on arbitrary kwargs

valentijnscholten:
This was not exactly what I meant, but it also works :-) Might be good enough for now.
If we wanted to make it transparent / compatible with any code running outside celery tasks, we could let tasks inherit from a base task which puts the user value into the place where `get_current_user` expects it: https://stackoverflow.com/questions/32141103/middleware-for-celery-tasks

Issue #5826, opened 2022-01-26
Deduplication algorithm does not work for SARIF and Rusty Hog


manuel-sommer:
Hi all,

I uploaded multiple RustyHog files (e.g. Essex Hog and Gottingen Hog), but the DefectDojo deduplication algorithm is wrong as no finding are deduplicated even though I uploaded one file multiple times. 

Tested on branch dev: v. 2.7.0-dev ( release mode )

damiencarol:
@manuel-sommer we need to check the configuration and if the parser provide the data.

manuel-sommer:
@damiencarol, that's exactly what I have already tried out. Didn't work out well yet. Still investigating.

manuel-sommer:
It looks like this issue also applies for the SARIF scanner. 
Import this file multiple times:
https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/scans/sarif/dockle_0_3_15.sarif 
![SARIF Deduplication](https://user-images.githubusercontent.com/47991713/153305341-146884ea-c2d7-4d68-a49c-8135cc30f0a8.png)

manuel-sommer:
It looks like the same issue applies for some SARIF scans. 
Use the following scanfile to reproduce the issue: 
https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/scans/sarif/dockle_0_3_15.sarif
![SARIF](https://user-images.githubusercontent.com/47991713/153359710-54e251fa-c6fa-4f72-9a0b-44bf34584686.png)

PR #5832, opened 2022-01-27
Deduplication Algorithm for RustyHog 

manuel-sommer:
#5826

Screenshot from 2022-01-27 09-51-16

This only deduplicates Choctaw Hog. I have tried various settings, but Gottingen Hog and Essex Hog were never deduplicated. 
@damiencarol, could you give me input here?


damiencarol:
It could be a bug in the de-duplication algorithm selection. I need to pull your PR to test it.

valentijnscholten:
What is in the debug logs? None of these scanners seem to have dedupe settings, which I think means the legacy algo is used which won't dedupe if there is no file_path or no line number for the finding.

manuel-sommer:
Hi @valentijnscholten 
All these three scanners are subscanners of the Rusty Hog scanner. 
https://github.com/newrelic/rusty-hog
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/rusty_hog/parser.py 
That's why I have set deduplication for Rusty Hog.

valentijnscholten:
ok. This thing read like a bug report instead of a PR, that's what got me confused.

manuel-sommer:
Here the bug report: https://github.com/DefectDojo/django-DefectDojo/issues/5826 

Indeed this is a PR, but I need input on how to continue here. Also, I have already asked in the slack channel.

damiencarol:
@manuel-sommer this scanner use the new interface for parsers. if you can check which dedup algo is used

manuel-sommer:
This PR is closed because of PR #5903

PR #5833, opened 2022-01-27
Wrong var in include_executive_summary block for product_type

Safren-tutu:
After enabling `"include_executive_summary": true` in Api_V2 (`product_types` -> `/product_types/{id}/generate_report/`) we started getting 500 errors by api requests. 
In logs:
```
uwsgi_1 | File "./dojo/api_v2/views.py", line 1618, in report_generate
uwsgi_1 | if test.environment:
uwsgi_1 | AttributeError: 'NoneType' object has no attribute 'environment'
```
Rename test var to t solved the problem.


StefanFl (approved):
Good find, thank you @Safren-tutu

mtesauro (approved):
Approved

PR #5839, opened 2022-01-28
copy python files from extra_settings

tiagoposse:
The new approach to copy files from `/app/docker/extra_settings` will fail in **kubernetes** when overwriting the folder through a volume mount. README.md will no longer be there and as such the number of files will be 1, instead of 2 as expected by the if conditions. I propose this be moved to enumerate and copy all *.py files found in the folder.

Additionally, this PR also deals with celery beat and worker, for which the local_settings.py file is being copied only after executing `python3 manage.py dbshell`, causing late-defined variables to not be counted on. In my use case this causes the DB password to not be passed

### DIFF
--- a/docker/entrypoint-celery-beat.sh
+++ b/docker/entrypoint-celery-beat.sh
@@ -12,9 +12,9 @@ done
 echo
 
 # Allow for bind-mount multiple settings.py overrides
-FILES=$(ls -I README.md /app/docker/extra_settings)
+FILES=$(ls /app/docker/extra_settings/*.py)
 NUM_FILES=$(echo "$FILES" | wc -l)
-if [ "$NUM_FILES" -gt "1" ]; then
+if [ "$NUM_FILES" -gt "0" ]; then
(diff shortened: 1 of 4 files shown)


Maffooch:
@tiagoposse I like the idea of just copying to python files over instead of the entire directory

I am not crazy about putting DB creds in a local settings file though. We generally recommend using environment variables to set credentials

tiagoposse:
@Maffooch I'd love if you could expand on your opinion. env vars vs config files to store sensitive data is a very hot topic. I myself have been leaning to an approach using injection into config files (hashicorp Vault approach), but I'm always happy to have my mind enlightened

valentijnscholten (approved):

valentijnscholten:
i've approved as regardless of storage of db credentials it's good to first try to copy over any extra settings before doing any work that might use them. not sure if there's anything k8s to think about @alles-klar as that was mentioned as where it's not working at the moment with 2.6.0

PR #5842, opened 2022-01-28
Option not to create user automatically using SOCIAL_AUTH

dsever:
Adding (DD)_SOCIAL_AUTH_CREATE_USER option to:
True, user will be created automatically
False, user will not be created automatically 

Please test it, I have test it on my local instance, but double check would be useful.

### DIFF
--- a/dojo/pipeline.py
+++ b/dojo/pipeline.py
@@ -8,6 +8,7 @@ from social_core.backends.google import GoogleOAuth2
 from dojo.authorization.roles_permissions import Permissions, Roles
 from dojo.product.queries import get_authorized_products
 
+USER_FIELDS = ['username', 'email']
 
 def social_uid(backend, details, response, *args, **kwargs):
     if settings.AZUREAD_TENANT_OAUTH2_ENABLED and isinstance(backend, AzureADTenantOAuth2):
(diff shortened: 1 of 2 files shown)


StefanFl:
Unfortunately I have no OAuth server set up to be able to test it. And to be honest, I have no idea why it works, because I don't know about the social auth framework. I can't even see how users have been created before that PR.

valentijnscholten:
Looks like it has some code copied from the create_user function built into django social auth? Would be nice if that could be avoided by using `super` or delegating to the built in code, but not sure if that's possible.

dsever:
That's right, that is quick solution, I can wrap it, to be more clear? 
But currently I'm more interested about adapting soc. Auth, because this is the first time I'm adapting auth pipelines

dsever:
@valentijnscholten  Can you check it now?

dojo/pipeline.py
```diff
@@ -110,3 +112,10 @@ def update_product_access(backend, uid, user=None, social=None, *args, **kwargs)
             if product_name not in project_names:
                 product = Product.objects.get(name=product_name)
                 Product_Member.objects.filter(product=product, user=user).delete()
+
+
+def create_user(strategy, details, backend, user=None, *args, **kwargs):
+    if not settings.SOCIAL_AUTH_CREATE_USER:
+        return
```
valentijnscholten, line 119: I assume you tested this, but should it officially be `return None` or `return {}` instead of returning "nothing"?

dsever: @valentijnscholten  I was asking my self  the same, but looks like it is not expecting anything to be returned back even in original code...

https://github.com/python-social-auth/social-core/blob/master/social_core/pipeline/user.py#L73
https://github.com/python-social-auth/social-core/blob/master/social_core/pipeline/user.py#L85

valentijnscholten (approved):

StefanFl (approved):

PR #5846, opened 2022-01-29
Allow empty report files (in API v2 and UI)

damiencarol:
Some security tools generate reports as empty file.

For example, Nuclei report format is JSON lines. So an empty report is a possible.

This modification let's the API v2 accept empty files.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1288,7 +1288,7 @@ class ImportScanSerializer(serializers.Serializer):
     endpoint_to_add = serializers.PrimaryKeyRelatedField(queryset=Endpoint.objects.all(),
                                                          required=False,
                                                          default=None)
-    file = serializers.FileField(required=False)
+    file = serializers.FileField(allow_empty_file=True, required=False)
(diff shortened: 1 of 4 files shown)


alles-klar:
If we allow empty files, we can handle it central or we need to ensure that every parser can handle an empty report -> we need a new unit test for each parser.

damiencarol:
@alles-klar having unit tests for each parser is already what we have today, I spent a lot of time (with Val and Stefan now) to have more unit tests based on real report data.. We also have high level unit tests that are able to do the "upload scan" feature from UI and API v2. (you can see I used this part to had a unit test that load an empty report for Nuclei tool)

StefanFl:
@alles-klar The parsers who don't accept an empty file throw an error message like when they get a file in a wrong format. For me that's completely acceptable.

StefanFl (approved):

valentijnscholten:
Not sure if we should let the parser decide if an empty report is the best solution, but don't have a strong opinion.

valentijnscholten (approved):

PR #5849, opened 2022-01-31
Product type metrics update

dsever:
Minor change to align across the DD, previously was Engineer with this commit is Reporter. 

image


StefanFl (approved):

Maffooch (approved):

mtesauro:
Wow, that "engineer" is left over from the original code from Rackspace.

dsever:
So I'm killing the history?

mtesauro:
LOL.  Yeah, and it's a good thing.  :-)

I forgot that "Engineer" was in that page and I think "Reporter" makes way more sense generally.

PR #5861, opened 2022-02-01
Customize documentation URL

kiblik:
If somebody is using his own documentation how to use DefectDojo in his own company.

Just set `DD_DOCUMENTATION_URL` as an environmental variable.


Maffooch (approved):
Interesting change. Is our documentation really not that good?? 😉

kiblik:
> Interesting change. Is our documentation really not that good?? 😉

I suppose it is good (I'm also trying to participate in it when I'm doing changes). But we have documentation which is more focused on our users (no information about installation, deep info about handling each kind of finding following our internal security processes).

dojo/context_processors.py
```diff
@@ -16,7 +16,8 @@ def globalize_oauth_vars(request):
             'SOCIAL_AUTH_KEYCLOAK_LOGIN_BUTTON_TEXT': settings.SOCIAL_AUTH_KEYCLOAK_LOGIN_BUTTON_TEXT,
             'SAML2_ENABLED': settings.SAML2_ENABLED,
             'SAML2_LOGIN_BUTTON_TEXT': settings.SAML2_LOGIN_BUTTON_TEXT,
-            'SAML2_LOGOUT_URL': settings.SAML2_LOGOUT_URL}
+            'SAML2_LOGOUT_URL': settings.SAML2_LOGOUT_URL,
+            'DOCUMENTATION_URL': settings.DOCUMENTATION_URL}
```
StefanFl, line 20: I am bit surprised to see the URL here. What is the reason for that?

dojo/templates/base.html
```diff
@@ -165,7 +165,7 @@
                             <li>
                                 <a href="{% url 'swagger-ui_oa3' %}?docExpansion=none" target="_blank"><i class="fa fa-book fa-fw" title="OpenAPI v3 Docs"></i> API v2 OpenAPI3 Docs </a>
                             </li>
-                            <li><a href="https://defectdojo.github.io/django-DefectDojo" target="_blank">
+                            <li><a href="{{ DOCUMENTATION_URL }}" target="_blank">
```
StefanFl, line 168: There is another link to the documentation at the bottom of the page. Did you leave it unchanged on purpose?

kiblik: No. Thanks. I'm going to change it as well.

dojo/context_processors.py
kiblik: If was the easiest way how to propagate this setting to all pages. What do you recommend?

kiblik: Or I can put it in to `system_settings`. Maybe it is the best idea.

dojo/templates/base.html
StefanFl: Might make sense, only wanted to be sure it hasn't been forgotten 😃

dojo/context_processors.py
StefanFl: To me `settings.dist.py` is the right place for a configuration like this. Just finding out how to use it in a template.

kiblik: I already changed it to `system_settings`. I prefer settings that can be changed on running instances.

kiblik:
Ok, so rollback and use better name

StefanFl:
That would be the more appropriate solution, thank you.

dojo/context_processors.py
```diff
@@ -2,7 +2,7 @@
 from django.conf import settings
 
 
-def globalize_oauth_vars(request):
+def globalize_vars(request):
```
kiblik, line 5: @StefanFl, is this better? 😄

StefanFl: 👍

StefanFl (approved):

PR #5873, opened 2022-02-02
nessus: make parser more robust on missing fields

whitespots:
Ensure the parser doesn't fail when any of the following fields are missing:
- Synopsis
- Solution
- Description
- See Also

### DIFF
--- a/dojo/tools/nessus_was/parser.py
+++ b/dojo/tools/nessus_was/parser.py
@@ -42,10 +42,10 @@ class NessusWASCSVParser(object):
             if not title:
                 continue
             # get description, mitigation, impact etc. from respective columns
-            description = row.get('Synopsis')
-            mitigation = str(row.get('Solution'))
-            impact = row.get('Description')
-            references = row.get('See Also')
+            description = row.get('Synopsis', '')
+            mitigation = str(row.get('Solution', ''))
+            impact = row.get('Description', '')
+            references = row.get('See Also', '')
             cvssv3 = row.get('CVSSv3', None)
             protocol = row.get('Protocol').lower() if 'Protocol' in row else None
             port = row.get('Port', None)


valentijnscholten:
I've updated the PR title + description to make it more informative.

Maffooch (approved):

dojo/tools/nessus_was/parser.py
```diff
@@ -42,10 +42,10 @@ def get_findings(self, filename, test: Test):
             if not title:
                 continue
             # get description, mitigation, impact etc. from respective columns
-            description = row.get('Synopsis')
-            mitigation = str(row.get('Solution'))
-            impact = row.get('Description')
-            references = row.get('See Also')
+            description = row.get('Synopsis', '')
+            mitigation = str(row.get('Solution', ''))
+            impact = row.get('Description', '')
```
damiencarol, line 47: these 2 attributes are not mandatory in the model. Are you sure you prefer blank string instead of just letting them as `NULL` is the DB?
I think it could be better to change the code as `row.get('Description')` which produce `None` as data

damiencarol (changes requested):

valentijnscholten:
@whitespots whitespots edited 5 days ago
> It's impossible now to parse the usual nessus report. Also it does not show any valuable error. So, anyway, this change would not be bad

The above was your original description. Can you explain what problem you are trying to solve? Is it that 'str' will fall on a None value? Or something else?

whitespots:
It had some problems with the rules engine. I just tried to make some "refactoring" :)
It is impossible to import the nessus report with any rule enabled

kiblik:
I suppose if code was failing on `None`s, it is better to hand how they are used.

So,  on line 70 replace
```python
find.description += row.get('Plugin Output')
```
by
```python
find.description = (find.description or '') + row['Plugin Output']
```

and on file 73 replace
```python
description = description + str(row.get('Plugin Output'))
```
by
```python
description = (description or '') + str(row['Plugin Output'])
```

PR #5882, opened 2022-02-04
Make alert the default notification in API

dsever:
This small Enhancement is switching from `slack` to `alert` as default values.

Why: slack is not a convenient option, because is not there by default and to be more consistent with the GUI option. 

Previously:
image

 After:
image


Maffooch (approved):

dojo/api_v2/serializers.py
```diff
@@ -1788,6 +1788,9 @@ class FindingNoteSerializer(serializers.Serializer):
 
 
 class NotificationsSerializer(serializers.ModelSerializer):
+
+    DEFAULT_NOTIFICATION = NOTIFICATION_CHOICES[-1]
```
StefanFl, line 1792: This is not necessarily stable, if another target for notifications would be introduced. Can `alert` be addressed explicitly?

StefanFl (changes requested):

dojo/api_v2/serializers.py
dsever: This is not stable in any case, because if you look at the HTML template, order is hardcoded :(

StefanFl: Initializing the default with

```suggestion
    DEFAULT_NOTIFICATION = ("alert", "alert")
```

would make the intention clear and the code easier to read. And it's more consistent with the model, where it is like `product_type_added = MultiSelectField(choices=NOTIFICATION_CHOICES, default='alert', blank=True)`

dsever: Makes sense, but then it should be in Notifications class, so default values are in the single place

StefanFl (approved):

StefanFl:
I think that's a great solution now.

PR #5883, opened 2022-02-05
Ignore warning from django-auditlog

StefanFl:
**django-auditlog** started to produce the warning `auditlog.LogEntry.additional_data: (django_jsonfield_backport.W001) You are using Django 3.1 or newer, which already has a built-in JSONField.` with its release 1.0.0. That happens because they are importing **django-jsonfield-backport** now. According to https://github.com/laymonage/django-jsonfield-backport/blob/master/README.rst:

> This package is fully compatible with the JSONField from Django 3.1. That means you just need to change your imports and edit your migrations when you finally upgrade to Django 3.1. If you leave them as they are, this package will use the built-in JSONField and system warnings will be raised.

That is why we can ignore the warning.


valentijnscholten (approved):

mtesauro (approved):
Approved

PR #5884, opened 2022-02-05
Copy only *.py files from extra_settings

StefanFl:
This is a follow-up to #5797 and #5839. Files will be copied from `docker/extra_settings` to `dojo\settings` if there a `*.py` files in the source directory. In that case currently all files are copied, including a `README.md`. This copied readme file appears then in the list of changed files to commit to git, which is unnecessary.

With this PR only `*.py` files are copied from `docker/extra_settings` to `dojo\settings`.


mtesauro (approved):
Approved

manuel-sommer (commented):
looks good

damiencarol (approved):

PR #5885, opened 2022-02-07
main(docker): sync nginx and django docker baselayer

alles-klar:
Sync the base layers for caching.

---

On behalf of DB Systel GmbH

### DIFF
--- a/Dockerfile.nginx
+++ b/Dockerfile.nginx
@@ -1,15 +1,18 @@
+
 # code: language=Dockerfile
 
 # The code for the build image should be identical with the code in
 # Dockerfile.django to use the caching mechanism of Docker.
 
-FROM python:3.8.12-slim-bullseye@sha256:3d3edc52cfae3ed6fb8303559f10184f962a8069194b2dee93baaac66ebedeb5 as build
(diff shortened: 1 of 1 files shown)


valentijnscholten:
I think at some point we decided to let go of this requirement and removed related comments. It can't hurt to sync it, but sometimes people forget or you need to install tools you don't need such as `gcc` in this case.

alles-klar:
I don't remember this decision and the comments are still there. For me this caching is important to make build faster. Installing packages you don't need is not really an argument if you use caching and you have to install it for the other image anyway. 

If we decide to let go this requirement, we should do it clean and remove/merge the hole base/build layer.

valentijnscholten:
I am not saying we must not make the layers equal, was just explaining there seemed to be a majority not caring enough about it to enforce it. But yeah the comments are still there, so we must enforce it ;-)

Even if the base for nginx differs from django, it will still be cached. Only the first build without cache is slower.
The base layer needs to stay there because only the wheels are needed for the final image, the rest is discarded to save space.

valentijnscholten (approved):

damiencarol (approved):

PR #5886, opened 2022-02-07
fix(helm): update redis keys

alles-klar:
In on of the last updates, bitnami has change the name of some keys. We need to update them too.

[Bitnami Redis Chart](https://github.com/bitnami/charts/tree/master/bitnami/redis)

---

On behalf of DB Systel GmbH

### DIFF
--- a/helm/defectdojo/values.yaml
+++ b/helm/defectdojo/values.yaml
@@ -383,6 +383,7 @@ rabbitmq:
     fsGroup: 1001
     runAsUser: 1001
 
+# For more advance options check the bitnami chart documentation: https://github.com/bitnami/charts/tree/master/bitnami/redis
 redis:
   enabled: false
   transportEncryption:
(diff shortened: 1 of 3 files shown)


helm/defectdojo/values.yaml
```diff
@@ -381,11 +381,11 @@ redis:
...
-  existingSecret: defectdojo-redis-specific
-  secretKey: redis-password
-  password: ""
-  cluster:
-    slaveCount: 1
+  auth:
+    existingSecret: defectdojo-redis-specific
+    existingSecretPasswordKey: redis-password
+    password: ""
+  architecture: standalone
```
dsever, line 388: While in past there was cluster way of working by default just add comment with the reference to bitnami manual. e.g. for more advance options check link.

And add it in doc (upgrade notes), because this are breaking changes during the upgrade. (but it looks like no-one is using redis, so this bug wasn't  detected earlier)

dsever (approved):
Approved

mtesauro (approved):
Approved

PR #5890, opened 2022-02-08
make JIRA authentication docs more explicit

valentijnscholten:
Seems that some people were not aware that username is needed when authenticating with API Token


mtesauro (approved):
I don't see how your doc change caused failures in the integration tests so I'm approving :-)

Approved

PR #5894, opened 2022-02-09
Format of reviewer form 

dsever:
When usernames are anonymized  it is very hard to find the right user for peer review. So it is adapted to use `generate_full_name()`

Before:
image

After:
Screenshot from 2022-02-09 15-29-38

### DIFF
--- a/dojo/forms.py
+++ b/dojo/forms.py
@@ -1555,8 +1555,8 @@ class ClearFindingReviewForm(forms.ModelForm):
 
 
 class ReviewFindingForm(forms.Form):
-    reviewers = forms.ModelMultipleChoiceField(queryset=Dojo_User.objects.filter(is_active=True),
-                                               help_text="Select all users who can review Finding.")
+
+    reviewers = forms.MultipleChoiceField(help_text="Select all users who can review Finding.")
     entry = forms.CharField(
         required=True, max_length=2400,
         help_text='Please provide a message for reviewers.',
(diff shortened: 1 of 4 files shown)


mtesauro:
Love this change. The username vs full name has caused me issues in the past.  Great find & update.

mtesauro (approved):
Approved

dojo/forms.py
```diff
@@ -1571,9 +1571,18 @@ def __init__(self, *args, **kwargs):
...
         if finding is not None:
-            self.fields['reviewers'].queryset = get_authorized_users_for_product_and_product_type(None, finding.test.engagement.product, Permissions.Finding_Edit)
+            queryset = get_authorized_users_for_product_and_product_type(None, finding.test.engagement.product, Permissions.Finding_Edit)
+            self.fields['reviewers'].choices = self._get_choices(queryset)
+
+    @staticmethod
+    def _get_choices(queryset):
+        l_choices = []
+        for item in queryset:
+            l_choices.append((item.pk, Dojo_User.generate_full_name(item)))
```
valentijnscholten, line 1584: should be `item.get_full_name()`

valentijnscholten (changes requested):
The screenshots are pixelated for me. The only thing I can think of is that we need to harmonize this everywhere. We should use the same solution / method / format everywhere.

dojo/forms.py
dsever: But `item` is type  `'django.contrib.auth.models.User'` not  proxy class `Dojo_User`  and actually it is calling original method that is formed as `First_Name  Last_Name` (missing username).

So this will not work out of box (harmonized across other forms).

I assume then `get_authorized_users_for_product_and_product_type` should be reworked first https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/user/queries.py#L25

dsever: I adapted `get_authorized_users_for_product_and_product_type` now it should be ok, I hope it will not break anything, but not results (format) looks fine.

dsever: @StefanFl can you please take a look because it  touches `get_authorized_users_for_product_and_product_type`

valentijnscholten: Good point @dsever . I think in general it's best if we return the `Dojo_User`  from our functions as it will have provide the extra abilities we sometimes want/need.

Maffooch:
I can make out the images and love the change! Agreed that it should be propagated everywhere in the app.

dsever:
@valentijnscholten  @StefanFl  can we proceed with this?

valentijnscholten:
Do you mean approving/merging this PR, or changing other places in the UI do behave the same? Ideally we should update as much places as possible to align everything.

dsever:
> Do you mean approving/merging this PR, or changing other places in the UI do behave the same? Ideally we should update as much places as possible to align everything.

To  merge this and then proceed with the othesr, to do smaller changes as possible.

valentijnscholten (approved):

valentijnscholten:
stefan looks in agreement to, merging.

PR #5898, opened 2022-02-10
Don't show 'all alerts' and 'clear all alerts' when no notification found

shipko:
It's strange that in the absence of alerts, we suggest looking at all the alerts or clearing them. 

I suggest showing a big message instead

Before:
Снимок экрана 2022-02-12 в 12 31 57

After:
Снимок экрана 2022-02-12 в 12 31 57

### DIFF
--- a/dojo/templates/base.html
+++ b/dojo/templates/base.html
@@ -764,10 +764,16 @@
                                                htmlEscape(elem.fields.source) + '</span></div></li><li class="divider"></li>');
                   $("#" + dropdown_alert).text(display_title);
                 });
-                $('.dropdown-alerts').append('<li><a class="text-center" href="{% url "alerts" %}"><strong>See All Alerts </strong>' +
-                                             '<i class="fa fa-angle-right"></i></a></li>');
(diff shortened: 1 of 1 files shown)


valentijnscholten (changes requested):
see above

shipko:
No problem :)

StefanFl:
I like the idea, thank you @shipko. One thing: Wouldn't it look better, if you make the font for **No alerts found** the same as for the other message, so `<strong>` instead of `<h3>`?

Maffooch:
Good suggestion from Stefan ^

shipko:
This is a two different variant: 
With using <strong>
image

With using <h3>
image

@StefanFl @Maffooch I think it would be better using <h3>. It's just nicely

StefanFl:
I still like the first one with `<strong>` more. Then all font sizes are the same on the start page.

StefanFl (approved):

mtesauro (approved):
Approved

mtesauro:
@valentijnscholten We good with your requested change?

valentijnscholten (approved):

PR #5899, opened 2022-02-10
Add Postgres NULL char validation/removal in Endpoint cleaning

Maffooch:
When first importing a scan file that contained a NULL char to my sandbox env (running MySQL) I did not experience any issues. After attempting to import to production (running Postgres) I was greeted with the following 

Screen Shot 2022-02-10 at 5 29 11 PM

Screen Shot 2022-02-10 at 5 29 35 PM

This error is totally dependent on the DB connected to dojo, so a universal fix is sorta out the window. Ideally, removing information from an endpoint is not acceptable, but I figured there was no way to import a scan with a NULL char to a Postgres backend anyways, so might as well remove the offending bits of the URL. 

Screen Shot 2022-02-10 at 5 30 56 PM
 
 I added some validation errors to make the log reader aware of what is going on

`uwsgi_1         | [10/Feb/2022 23:29:53] WARNING [dojo.importers.utils:133] DefectDojo is storing broken endpoint because cleaning wasn't successful: ['Query "count=50&query=http://test.com/n?\x00.php&" has invalid format - It contains the NULL character. The following action was taken: Postgres does not accept NULL character. Attempting to remove...']`

For those that run MySQL, they will see a similar error message like the one below and the endpoint (and NULL chars) will be preserved 

`uwsgi_1         | [10/Feb/2022 23:29:53] WARNING [dojo.importers.utils:133] DefectDojo is storing broken endpoint because cleaning wasn't successful: ['Query "count=50&query=http://test.com/n?\x00.php&" has invalid format - It contains the NULL character. The following action was taken: No action taken as MySQL allows for storing the NULL character']`

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1281,7 +1281,7 @@ class Endpoint(models.Model):
 
     def clean(self):
         errors = []
-        null_char_list = ["0x00", "%00", "\x00"]
+        null_char_list = ["0x00", "\x00"]
         db_type = connection.vendor
         if self.protocol or self.protocol == '':
             if not re.match(r'^[A-Za-z][A-Za-z0-9\.\-\+]+$', self.protocol):  # https://tools.ietf.org/html/rfc3986#section-3.1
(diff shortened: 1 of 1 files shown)


devGregA (approved):

Maffooch:
It looks like unit tests do not like input validation errors because the importer does not manage the exceptions

valentijnscholten:
Did you consider urlencoding, or:

image

The reason I am asking is that the null character might be part of the exploit payload. If it is removed, it might confuse the analyst who might think there is no vulnerability?

Maffooch:
I like the idea of replacing the null chars with "%00"
That is actually the way the tool I am using reports the endpoint with, but it gets converted to "/xx" at some point

mtesauro (approved):
Approved.

What do you think @valentijnscholten ?

valentijnscholten (approved):

PR #5689, opened 2022-01-05
feat: Rework sslyze parser for version 5

kibernautas:
The sslyze parser changed schema when it updated to v5, so I have rewritten the parser partially in order to parse the JSON file. 

- [x] rework parser for v5
- [x] create new unit test cases for v5
- [x] create fallback scan option for v3

PR #5900, opened 2022-02-11
fix sslyze test type name in sample data

valentijnscholten:
Fix sslyze test type name in sample data (changed in #5689)


mtesauro (approved):
Approved

Maffooch (approved):

PR #5903, opened 2022-02-12
Use scan_type to determine hash_code and deduplication algorithms

### LINKED ISSUE #5826: Deduplication algorithm does not work for SARIF and Rusty Hog

StefanFl:
Fixes #5826

Deduplication for SARIF and Rusty Hug doesn't work reliably. At the moment, the `test_type` is used to determine the algorithms for hash_code calculation and deduplication. But for SARIF and Rusty Hog a `test_type` is generated during import like `Dockle (SARIF)´. There will be no settings for this `test_type`.

There is also a `scan_type` in the test, which will be set to **SARIF** or **Rusty Hog Scan**. The deduplication works reliably using the `scan_type` to determine the algorithms for hash_code calculation and deduplication. 

PR #5832 can be closed, the changes are part of this PR as well.


StefanFl:
@manuel-sommer Please have a look if the settings for hash code calculation are ok for you.

damiencarol:
This is not how the were intended to work. Using the `test_type` is intentional. My goal was to have this algo to choose de-duplication algorithm:
1. check if we have a configuration for `test_type`
2. if not, check if we have a configuration for the `scan_type`
3. if not,fallback to the default algorithm
The goal is to be able to customize for each `test_type` or `scan_type` because `test_type` are generated dynamically.

StefanFl:
@damiencarol I couldn't find how step 2 was implemented. At least for imports always `test_type.name` was used. Please check how it is implemented now in `models.py`. This should do what you have intended.

manuel-sommer (commented):
I retested #5826 with this PR. It looks good for both Rusty Hog and SARIF:
Screenshot from 2022-02-14 10-32-04

damiencarol:
> @damiencarol I couldn't find how step 2 was implemented. At least for imports always test_type.name was used. Please check how it is implemented now in models.py. This should do what you have intended.

Great!

> What about the hash_code fields? They where always searched by test_type.name before and now have been switched to scan_type.

Same thing for hash_code fields. I think we should use test_type.name first and fallback to the scan_type

damiencarol (approved):

StefanFl:
Ok @damiencarol, I will change the hash_code calculation to the fallback as well.

StefanFl:
Now we have the same fallback mechanism for `HASHCODE_FIELDS_PER_SCANNER` and `HASHCODE_ALLOWS_NULL_CWE`. While doing this I refactored the calculation of hash_codes to make it better understandable.

damiencarol (approved):
Good work

valentijnscholten (approved):

PR #5904, opened 2022-02-12
Replace CVE with Vulnerability references in Findings

StefanFl:
This PR does several things:
- New models for `Vulnerability_Reference` (for Findings) and `Vulnerability_Reference_Template` (for Finding_Templates) have been introduced. They hold references to other repositories with vulnerability advices like NVD (CVE), GitHub Advisories (GHSA) or Sonatype OSS index (OSV). Links to repositories can be configured in `settings.dist.py`.
- The attribut `cve` in the Finding or Finding Template is synchronized with the first Vulnerability Reference of the list, to keep backward compatibility. This is done when adding or editing Findings and for imports/reimports
- Finding lists and Finding view show the first Vulnerability Reference where there has been the CVE before. This will still make sense in the future, because a lot of parsers only generate one Vulnerability Reference. Additional Vulnerability References (items 2..n of the list of Vulnerability References) are shown in the Finding view in a separate row.
- Finding filters in API and UI filter for Vulnerability References
- 3 parsers (CycloneDX, Grype and pip-audit) have been adapted to use the new feature.

**Example**

* CycloneDX delivers in this example a vulnerability with the id `SNYK-JAVA-COMFASTERXMLJACKSONCORE-32111` and within this vulnerability an additional reference `CVE-2018-7489`.
* The finding has a list of Vulnerability References, containing 2 items: `['SNYK-JAVA-COMFASTERXMLJACKSONCORE-32111', 'CVE-2018-7489']`
* The attribut `cve` of the finding contains the first item of the list for backwards compatibility: `SNYK-JAVA-COMFASTERXMLJACKSONCORE-32111`. This is still used for deduplication, search and other things that have not yet been changed.
* The first item of the list of Vulnerability References is shown in the list of Findings:
2022-03-15 08_16_10-Findings _ DefectDojo
* The first item of the list of Vulnerability References is shown in the top row of the Finding details view, the second item of the list (could be more if the parser would provide more) is shown in a separate row as Additional Vulnerability References:
2022-02-12 15_58_11-View Finding _ DefectDojo

**What has not been done for now:**

The `cve` attribute in Findings and Finding_Templates shall be removed in the future. To be able to that we will need to some more things in future PR's:
- Change a lot of parsers who write the `cve` attribute.
- Change the calculation of `hash_codes`
- Change the search of Findings with the search engine
- Some other things we will find on our way


StefanFl:
> Is the idea the the first reference is stored in the finding model and any subsequent/extra references are stored in the new table?

Exactly, lots of parser will provide only one value anyway, others like Grype or CycloneDX have a main reference and additional ones.

valentijnscholten:
Would it be easier to just have a plain 1-N relationship (so not storing the "first" reference in the finding model itself). That would make some cases easier to implement I think:

- What if someone removes the "first" reference? 
- What if someone wants to make one of the extra references the "first" reference (swap)
- How to perform searches via the API without having to specify multiple search clauses to both filter on the "first" reference AND (or?) the extra references?

Thinking out loud, would it make sense to use a field of type `tags`, which brings some benefits such as search, serialization, ...
Probably not, but just putting it out there...

StefanFl:
There will be not too many cases where we will have additional references. The CVE is used by a lot of scanners to calculate their hash_codes and is used in the search function. I am afraid this will get much more complicated if we don't have the first reference anymore.

I see your points, let me think about it and look what solutions there are. This is still a draft to get your feedback and gather ideas.

damiencarol:
> Would it be easier to just have a plain 1-N relationship (so not storing the "first" reference in the finding model itself). That would make some cases easier to implement I think:
> 
>     * What if someone removes the "first" reference?
> 
>     * What if someone wants to make one of the extra references the "first" reference (swap)
> 
>     * How to perform searches via the API without having to specify multiple search clauses to both filter on the "first" reference AND (or?) the extra references?

Fully agree with this parT. It's time to make this part (references) simple with a 1-N relation, so no more cve attribute.

> Thinking out loud, would it make sense to use a field of type `tags`, which brings some benefits such as search, serialization, ... Probably not, but just putting it out there...

I'm not so sure about this part. Libraries to support tags were changing a lot. I prefer a simple Django model entity.

damiencarol:
> There will be not too many cases where we will have additional references. The CVE is used by a lot of scanners to calculate their hash_codes and is used in the search function. I am afraid this will get much more complicated if we don't have the first reference anymore.
> 
> I see your points, let me think about it and look what solutions there are. This is still a draft to get your feedback and gather ideas.

As said during one of our contributor meetings, let make this first step simple. We have many parsers that only manage one CVE by finding. we can keep this behaviors keep the hash_code algorithm as before.

dojo/models.py
```diff
@@ -2545,6 +2543,11 @@ class FindingAdmin(admin.ModelAdmin):
     x: "Endpoint: " + str(x.endpoint)
 
 
+class Additional_Vulnerability_Reference(models.Model):
+    finding = models.ForeignKey(Finding, editable=False, on_delete=models.CASCADE)
+    vulnerability_reference = models.TextField(max_length=50, blank=False, null=False)
```
damiencarol, line 2548: We need a type somewhere.

StefanFl: Where would the type come from and what would we need it for?

StefanFl:
@valentijnscholten @damiencarol I will go for a hybrid approach. A parser can either set the CVE field in the Finding or the list of Vulnerability References. This would be a first step that makes us more flexible. That will be this PR. After that I can change existing parsers step by step to use the list of Vulnerability References until the CVE field is not written any more and then we can get rid of that field. Does this make sense?

valentijnscholten:
I think if we adjust the importer / reimporter to take the cve field set by the parser and store it in the new table, we could remove the field.
But it would leave all templates to be adjusted, which could be done by making `cve` a property of the `Finding` model.
But that leaves some places where queries are built with cve as filter, those should be changed. But this might be just a couple of fixes in `filters.py` possibly.
But then we have the API, people filtering on `cve` would expect it to keep working without having to adjust their queries.

So might be substantial amount of working to finish the job completely.

If we go with your proposal, we may want to always copy the `cve` to the new table as well?

Also I think the new table needs a `type` field, so we can filter easily on `cve`, `rhsa`, github identifier etc. Without having to do partial string matching / regex matching.

StefanFl:
> I think if we adjust the importer / reimporter to take the cve field set by the parser and store it in the new table, we could remove the field.

I think this wouldn't work because the parsers set the CVE field in the Finding model. So we can remove it only when no parser sets this anymore. 

> But it would leave all templates to be adjusted, which could be done by making cve a property of the Finding model.
> But that leaves some places where queries are built with cve as filter, those should be changed. But this might be just a couple of fixes in filters.py possibly.
> But then we have the API, people filtering on cve would expect it to keep working without having to adjust their queries.

> So might be substantial amount of working to finish the job completely.

Exactly, a lot of work and there are lots of unit tests involved as well that need to be changed.

> If we go with your proposal, we may want to always copy the cve to the new table as well?

I didn't really think about that, to be honest. But it seems to be a good idea to synchronize the CVE field with the list. I will see how to deal with it while implementing things.

> Also I think the new table needs a type field, so we can filter easily on cve, rhsa, github identifier etc. Without having to do partial string matching / regex matching.

I still don't see the need for the type. Where exactly do we do this filtering? And in case we would do filtering somewhere, an `icontains` would do the job quite well.

valentijnscholten:
The finding only gets saved by the importer, so we can handle cave varies there before saving to the database.

We are not filtering anywhere because we only have cve currently 🙂. But we do need to think indeed if we need/want to filter on type. If so, icontains could get very slow, but startswith might be ok if properly indexed.

damiencarol:
I agree with @valentijnscholten . We need a type field with wlaue like "cve", "ghsa", etc. For filtering but also to validate if we want to enforce some validations.
Regarding the modifications from cve to Vulnerability References, I prefer to do the reverse. 
1) we modify each parsers to switch from cve to Vulnerability References one by one (I can help)
2) in the importer we can populate the cve field with Vulnerability References to stay compatible with existing page/template
3) once no page/template use the old deprecated cve field we can remove it.

StefanFl:
> I agree with @valentijnscholten . We need a type field with wlaue like "cve", "ghsa", etc. For filtering but also to validate if we want to enforce some validations.

I am still not sold on the idea of the type. Wouldn't it in the end be very similar like the start of the reference? At least for the ones I know so far. We should build it, when we need it. So far there is no filtering and no checks. 

> Regarding the modifications from cve to Vulnerability References, I prefer to do the reverse.
>
>   1. we modify each parsers to switch from cve to Vulnerability References one by one (I can help)
>    2. in the importer we can populate the cve field with Vulnerability References to stay compatible with existing page/template
>    3. once no page/template use the old deprecated cve field we can remove it.

That's basically what I started today 😃

dojo/models.py
damiencarol: filtering and some deduplication logic will be more simple if we have the type of reference.
For ex: "cve" or "ghsa"

StefanFl: The type would typically be the prefix of the vulnerability reference itself, at least for all sources I know about. I still think it adds more complexity and redundant data with little benefit.

- **Filtering:** We have 117 `icontains` in `filters.py`, wouldn't this work for the vulnerability references as well? Otherwise we get one more field on some already overcrowded filter forms.
- **Deduplication:** To calculate the `hash_code` we will most likely just concetanate the vulnerability references. How would the type come into play for that?

valentijnscholten:
I think it's great we get support for multiple identifiers and recognize it's been a lot of work. Still to me it _feels_ not 100% complete currently. Would it be doable to remove the `cve` field and adjust the UI and API? I think the parsers can be modified later.

In the current state it might confuse people. Suppose we have finding with:

`` 
cve: SNYK-12345
vulnerability references: SNYK-12345, CVE-2022-7890
`` 

If uses searches for `cve=CVE-2022-7890` via the API, the finding won't be returned? It would already if the `ApiFilter` gets extended with search for vulnerability references.

StefanFl:
> Would it be doable to remove the cve field and adjust the UI and API? I think the parsers can be modified later.

I don't see how this should work in one go. There are so many areas where the `cve` field is used. What happens if a parser assigns a value to this field? Wouldn't it would lead to an exception? Plus changing the hash_code calculation, search engine and some more places I don't know yet. 

My idea is to do this step by step and not working on one PR for a very long time and making it far too large.

> If a user searches for cve=CVE-2022-7890 via the API, the finding won't be returned? It would already help if the ApiFilter gets extended with search for vulnerability references.

This is a very good idea, I will work on it within this PR.

valentijnscholten:
> > Would it be doable to remove the cve field and adjust the UI and API? I think the parsers can be modified later.
> 
> I don't see how this should work in one go. There are so many areas where the `cve` field is used. What happens if a parser assigns a value to this field? Wouldn't it would lead to an exception? 

Not if the importer/reimport code move the cve into the vulnerability references and not store the cve field.

> Plus changing the hash_code calculation, search engine and some more places I don't know yet.
Good point, could maybe be as simple as adding a readonly `cve` property, not sure.

StefanFl:
>>    What happens if a parser assigns a value to this field? Wouldn't it would lead to an exception?

> Not if the importer/reimport code move the cve into the vulnerability references and not store the cve field.

I still don't understand how that should work. Parsers generate Findings with code like

```python
items.append(
    Finding(
        test=test,
        title=title,
        cve=vuln_id,
        ...
    )
)
```

When the `cve` attribute doesn't exist anymore, I get a `TypeError: Finding() got an unexpected keyword argument 'cve'`.

valentijnscholten:
I think there's a way to just remove the `cve` field before saving to the database in the importers.

But as discussed on chat also search is affected. And I think the cve field maybe has to remain for a bit in the API as other systems might be using it to push findings to DD. So maybe we need to go with the PR as is.

StefanFl:
I am already working on filtering on vulnerability references and to remove the `cve` attribute from the API. Then the functionality for users will basically be in the final state and future changes are predominantly internal. Currently updating unit and integration tests.

It is a changing break for the API, but it will have to come anyway and then it is done.

StefanFl:
@valentijnscholten and @damiencarol I think that's a good package now. Ready for review.

dojo/finding/views.py
```diff
@@ -1133,6 +1145,11 @@ def apply_template_to_finding(request, fid, tid):
             finding.last_reviewed = timezone.now()
             finding.last_reviewed_by = request.user
             finding.tags = form.cleaned_data['tags']
+
+            finding.cve = None
+            Vulnerability_Reference.objects.filter(finding=finding).delete()
+            finding_helper.save_vulnerability_references(finding, form.cleaned_data['vulnerability_references'].split())
```
damiencarol, line 1151: I prefer to have the line 1150 in the the `save_vulnerability_references()` method.
you choose to create a method to manage the vulnerability_references attribute.

dojo/importers/utils.py
```diff
@@ -168,3 +168,29 @@ def add_endpoints_to_unsaved_finding(finding, test, endpoints, **kwargs):
...
+def handle_vulnerability_references(finding):
+    # Synchronize the cve field with the unsaved_vulnerability_references
+    # We do this to be as flexible as possible to handle the fields until
+    # the cve field is not needed anymore and can be removed.
+    if finding.unsaved_vulnerability_references and finding.cve:
+        # Make sure the first entry of the list is the value of the cve field
+        finding.unsaved_vulnerability_references.insert(0, finding.cve)
+    elif finding.unsaved_vulnerability_references and not finding.cve:
+        # If the cve field is not set, use the first entry of the list to set it
+        finding.cve = finding.unsaved_vulnerability_references[0]
```
damiencarol, line 182: you clearly fixed that the CVE should be always the first unsaved_vulnerability_references ?
if yes, what do we do if we don't have a CVE but another reference?

dojo/tools/anchore_grype/parser.py
```diff
@@ -36,19 +36,19 @@ def get_findings(self, file, test):
...
+            vulnerability_references = None
             related_vulnerabilities = item.get('relatedVulnerabilities')
             if related_vulnerabilities:
                 related_vulnerability = related_vulnerabilities[0]
-                rel_id = related_vulnerability.get('id')
                 rel_datasource = related_vulnerability.get('dataSource')
                 rel_urls = related_vulnerability.get('urls')
                 rel_description = related_vulnerability.get('description')
                 rel_cvss = related_vulnerability.get('cvss')
+            vulnerability_references = self.get_vulnerability_references(vuln_id, related_vulnerabilities)
```
damiencarol, line 51: Attribute `relatedVulnerabilities` is an array. it could be a good opportunity to add every items.

unittests/tools/test_pip_audit_parser.py
```diff
@@ -29,10 +29,12 @@ def test_parser_many_vulns(self):
...
+        description = 'aiohttp is an asynchronous HTTP client/server framework for asyncio and Python. In aiohttp before version 3.7.4 there is an open redirect vulnerability. A maliciously crafted link to an aiohttp-based web-server could redirect the browser to a different website. It is caused by a bug in the `aiohttp.web_middlewares.normalize_path_middleware` middleware. This security problem has been fixed in 3.7.4. Upgrade your dependency using pip as follows "pip install aiohttp >= 3.7.4". If upgrading is not an option for you, a workaround can be to avoid using `aiohttp.web_middlewares.normalize_path_middleware` in your applications.'
```
damiencarol, line 32: This is good.

dojo/importers/utils.py
StefanFl: If there is another reference (like GHSA or SNYK), it will be currently in the `cve` field, as well as being the first element of the vulnerability references, if the parser puts it there. There are parser who already did it before.

In the UI and the API this `cve` field is not used anymore, they are using the `Vulnerability_References`. It will be removed, after all parsers, the search and the hash_code algorithm are changed in subsequent PR's.

dojo/models.py
StefanFl: Currently nothing has changed for deduplication, because it still uses the `cve` field. In the next PR I will work on the `hash_code` calculation, we should discuss these changes then.

dojo/tools/anchore_grype/parser.py
StefanFl: It could be done, but I wouldn't do it in this PR.

dojo/finding/views.py
StefanFl: This `delete` is not necessary here. The `save_vulnerability_references()` method handles the old Vulnerability References anyway. I will remove this line.

components/package.json
```diff
@@ -32,7 +32,7 @@
     "jquery.cookie": "1.4.1",
     "jquery.flot.tooltip": "^0.9.0",
     "jquery.hotkeys": "jeresig/jquery.hotkeys#master",
-    "jszip": "^3.9.0",
+    "jszip": "^3.9.1",
```
StefanFl, line 35: I have no idea why this is marked as a change. `jszip` is at version 3.9.1 in `dev` as well

components/yarn.lock
```diff
@@ -715,10 +715,10 @@ jquery.hotkeys@jeresig/jquery.hotkeys#master:
   resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.6.0.tgz#c72a09f15c1bdce142f49dbf1170bdf8adac2470"
   integrity sha512-JVzAR/AjBvVt2BmYhxRCSYysDsPcssdmTFnzyLEts9qNwmjmu4JTAMYubEfwVOSwpQ1I1sKKFcxhZCI2buerfw==
 
-jszip@^3.9.0:
-  version "3.9.0"
-  resolved "https://registry.yarnpkg.com/jszip/-/jszip-3.9.0.tgz#511b26718aca5f3a553d1f8e072dfd0efdc74481"
-  integrity sha512-Vb3SMfASUN1EKrFzv5A5+lTaZnzLzT5E6A9zyT7WFqMSfhT2Z7iS5FgSOjx2Olm3MDj8OqKj6GHyP2kMt1Ir6w==
+jszip@^3.9.1:
```
StefanFl, line 718: I have no idea why this is marked as a change. `jszip` is at version 3.9.1 in `dev` as well

helm/defectdojo/templates/django-ingress.yaml
```diff
@@ -29,9 +29,9 @@ metadata:
 {{- end }}
 spec:
 {{- if semverCompare ">=1.19.0" .Capabilities.KubeVersion.GitVersion }}
-   {{- if .Values.django.ingress.ingressClassName }}
-   ingressClassName: {{ .Values.django.ingress.ingressClassName }}
-   {{- end }}
+  {{- if .Values.django.ingress.ingressClassName }}
```
StefanFl, line 32: I have no idea why this is marked as a change. It is exactly what was in the merged PR https://github.com/DefectDojo/django-DefectDojo/pull/6140

requirements.txt
```diff
@@ -37,7 +37,7 @@ python-dateutil==2.8.2
 pytz==2022.1
 redis==4.2.2
 requests==2.27.1
-sqlalchemy==1.4.34  # Required by Celery broker transport
+sqlalchemy==1.4.35  # Required by Celery broker transport
```
StefanFl, line 40: I have no idea why this is marked as a change, it is the same version as in `dev`.

requirements.txt
```diff
@@ -50,8 +50,8 @@ Python-jose==3.3.0
 gitpython==3.1.27
 debugpy==1.6.0
 python-gitlab==3.3.0
-google-api-python-client==2.42.0
-google-auth==2.6.2
+google-api-python-client==2.43.0
```
StefanFl, line 53: I have no idea why this is marked as a change, it is the same version as in `dev`.

requirements.txt
```diff
@@ -50,8 +50,8 @@ Python-jose==3.3.0
...
+google-auth==2.6.3
```
StefanFl, line 54: I have no idea why this is marked as a change, it is the same version as in `dev`.

damiencarol:
@StefanFl I think the PR is good like this. Any idea why GitHub sees so many diffs?
Could you try to rebase on the last dev branch to force it to re-check?

StefanFl:
I had to squash all commits into one. Otherwise a rebase with dev wasn't really possible because of the long history and large amount of commits in this PR. Now it all seems to be fine. Only the flake8 check isn't working, it seems to have a problem with the squashing the commits. I checked, the PR is compliant with flake8.

StefanFl:
UPDATE: The flake8 problem is not related to this PR, it happens on other PR's as well and seems to be caused by https://github.com/actions/checkout/issues/760

Maffooch (approved):

Maffooch:
Leaving the last review for @damiencarol

Issue #5926, opened 2022-02-17
the "Harbor Vulnerability Scan" not in duplicate delete


TriompheL:
**Bug description**
I found the " Harbor Vulnerability Scan" no in  https://github.com/DefectDojo/django-DefectDojo/blob/9be7b81b39ea72c533655d52141a748f4cdfaedc/dojo/settings/settings.dist.py#L1016
So the harbor scan I imported cannot be duplicate to delete.
![image](https://user-images.githubusercontent.com/27794837/154396429-23193dee-1ff9-4554-88c6-15ac51a5c088.png)

**Steps to reproduce**

**Expected behavior**

**Deployment method** *(select with an `X`)*
- [ X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: centos 7
 - DefectDojo version (see footer) or commit message: [use `git show -s --format="[%ci] %h: %s [%d]"`]

**Logs** 
Use `docker-compose logs` (or similar, depending on your deployment method) to get the logs and add the relevant sections here showing the error occurring (if applicable).

**Sample scan files**
If applicable, add sample scan files to help reproduce your problem.

**Screenshots**

**Additional context** (optional)

manuel-sommer:
https://github.com/DefectDojo/django-DefectDojo/pull/5931

PR #5931, opened 2022-02-17
harbor vulnscan deduplication algorithm #5926

manuel-sommer:
Screenshot from 2022-02-17 11-45-15


valentijnscholten:
I think `severity` should be removed and possiblt also `description`. The `title` seems to be unique? Either way the `description` won't make it "more unique".

StefanFl:
I agree with @valentijnscholten, the `title` should be good enough. Let's keep it simple.

valentijnscholten (approved):

PR #5932, opened 2022-02-17
main(docker): remove duplicated default envs from dockerfile

alles-klar:
Reduce redundancy through removing environ variables that have default values in settings.dist.py. 

---

On behalf of DB Systel GmbH


Dockerfile.django
```diff
@@ -106,19 +106,12 @@ RUN \
...
 USER ${uid}
 ENV \
+  # Only variables that are not defined in settings.dist.py
   DD_ADMIN_USER=admin \
   DD_ADMIN_MAIL=admin@defectdojo.local \
   DD_ADMIN_PASSWORD='' \
-  DD_ADMIN_FIRST_NAME=Administrator \
+  DD_ADMIN_FIRST_NAME=Admin \
   DD_ADMIN_LAST_NAME=User \
-  DD_ALLOWED_HOSTS="*" \
```
StefanFl, line 114: `DD_ALLOWED_HOSTS` is used in `settings.dist.py` but there is no default value for it in there.

alles-klar: There is, but not in the standard list.... 
ALLOWED_HOSTS = tuple(env.list('DD_ALLOWED_HOSTS', default=['localhost', '127.0.0.1']))

StefanFl: Ah, now I understand that line. But don't we change it now from `*` to `['localhost', '127.0.0.1']`?

alles-klar: For users that use the docker container without our docker-compose or helm chart yes, but in that case * is a very bad default. For docker-compose and helm these values are overwritten with default values from there.

valentijnscholten (approved):

mtesauro (approved):
Approved

PR #5934, opened 2022-02-17
dc-up.sh: use correct message

kiblik:


StefanFl (approved):
Classical copy-and-paste error, thanks @kiblik for spotting this

PR #5942, opened 2022-02-19
Fix indentation in docker-compose.override.debug.yml

kiblik:
For some reason, the structure in `docker-compose.override.debug.yml` was indented more then in other `docker-compose.override.*.yml` files


mtesauro (approved):
Approved

valentijnscholten (approved):

Issue #5943, opened 2022-02-19
Reimport break endpoint_status (problem of late deduplication of endpoint_statuses)


kiblik:
**Bug description**

After reimporting one of the reports which contain endpoints, I received HTTP-500. After some time (when deduplication in celery was done), everything was all right. 

**Steps to reproduce**
1. create Product
2. create Eng
3. import Test - [Gitlab DAST](https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/scans/gitlab_dast/gitlab_dast_one_vul.json)
4. reimport the same test
5. check the list of endpoints (pop-up bubble when you have the cursor over icon next to finding) - you can see the same endpoint twice
6. open finding - HTTP 500 - "MultipleObjectsReturned at /finding/1 get() returned more than one Endpoint_Status -- it returned 2!"
7. wait a minute (for deduplication in celery)
8. everything is fine

**Expected behavior**

No HTTP 500

**Deployment method** *(select with an `X`)*
- [x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Commit message: `[2022-02-18 09:08:26 +0100] c9c5ae2cb: dc-up.sh: use correct message (#5934) [ (HEAD, upstream/dev)]`

**Logs** 
```python
django-defectdojo-uwsgi-1         | ERROR:django.request:Internal Server Error: /finding/1
django-defectdojo-uwsgi-1         | Traceback (most recent call last):
django-defectdojo-uwsgi-1         |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
django-defectdojo-uwsgi-1         |     response = get_response(request)
django-defectdojo-uwsgi-1         |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
django-defectdojo-uwsgi-1         |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
django-defectdojo-uwsgi-1         |   File "/app/./dojo/authorization/authorization_decorators.py", line 30, in _wrapped
django-defectdojo-uwsgi-1         |     return func(request, *args, **kwargs)
django-defectdojo-uwsgi-1         |   File "/app/./dojo/finding/views.py", line 372, in view_finding
django-defectdojo-uwsgi-1         |     return render(
[...]

PR #5944, opened 2022-02-19
Endpoint_status: fix reimport

kiblik:
Solution for https://github.com/DefectDojo/django-DefectDojo/issues/5943

Core of fix was small change of
```python
eps, created = Endpoint_Status.objects.get_or_create(
    finding=finding,
    endpoint=ep,
    date=finding.date)
```
to
```python
eps, created = Endpoint_Status.objects.get_or_create(
    finding=finding,
    endpoint=ep)
if created:
    eps.date = finding.date
```

I added also migration, the test of migration and I disabled the old migration tests. Any future changes of code should not affect these tests and they are quite slow. Disabling should not bring any problem and it should speed up unittesting.

TODO:
- [x] unittest


kiblik:
This of code 
```python
        ep = None
        try:
            ep, created = endpoint_get_or_create(...)
        except (MultipleObjectsReturned):
            pass

        eps = None
        try:
            eps, created = Endpoint_Status.objects.get_or_create(...)
        except (MultipleObjectsReturned):
            pass
```
has been in DD for a really long time. It doesn't follow best practices.

I suppose we did a lot of improvements (migrations and processing of endpoints) which should avoid situations when this can ever happen. So I'm going to finally remove it.

Please report it as a [bug](https://github.com/DefectDojo/django-DefectDojo/issues/new?assignees=&labels=bug&template=bug_report.md&title=MultipleObjectsReturned%20of%20Endpoint%20or%20EndpointStatus) and feel free to assign it directly to me if you will ever receive `MultipleObjectsReturned` in this part of the code.

valentijnscholten:
> ```python
>  ep = None
>         try:
>             ep, created = endpoint_get_or_create(...)
>         except (MultipleObjectsReturned):
>             pass
> ```

Instead of pass, we could maybe add some "self healing" code that merges all the endpoints / statuses into one?

For the rest of the PR:
- Is there a test case for reimport that covers the fixed behaviour? i.e. endpoint_statuses with different dates (no longer) being created?
- Could you look at skipping the migration test usings `@skip` or similar instead of commenting it out?

kiblik:
> > ```python
> >  ep = None
> >         try:
> >             ep, created = endpoint_get_or_create(...)
> >         except (MultipleObjectsReturned):
> >             pass
> > ```
> 
> Instead of pass, we could maybe add some "self healing" code that merges all the endpoints / statuses into one?
> 
> For the rest of the PR:
> 
>     * Is there a test case for reimport that covers the fixed behaviour? i.e. endpoint_statuses with different dates (no longer) being created?
> 
>     * Could you look at skipping the migration test usings `@skip` or similar instead of commenting it out?

@valentijnscholten, good idea, thanks. 
I will implement a "self-healing" code.
Tests are in todo list.

kiblik:
@valentijnscholten, do I need to implement migration if I will implement self-healing? There will be the same piece of code.

kiblik:
@valentijnscholten, I added 
```python
        constraints = [
            models.UniqueConstraint(fields=['finding', 'endpoint'], name='endpoint-finding relation')
        ]
```
I suppose, there will be no change for `MultipleObjectsReturned`

damiencarol (approved):

valentijnscholten:
Thanks @kiblik for all the work. I think it looks OK, but there is of course some risk that the migration will fail to add the unique constraint. And we might run into exceptions where duplicate endpoints or endpoint statuses are created. So I think we would need someone to test it on real life data soon if we want to include it in the release next week.

kiblik:
> Thanks @kiblik for all the work. I think it looks OK, but there is of course some risk that the migration will fail to add the unique constraint. And we might run into exceptions where duplicate endpoints or endpoint statuses are created. So I think we would need someone to test it on real life data soon if we want to include it in the release next week. @Maffooch are you able to test this branch maybe on a copy of some real life data? I myself am not using endpoints (yet).

No problem, I will wait for the test.

dojo/api_v2/serializers.py
```diff
@@ -718,10 +719,16 @@ class Meta:
     def create(self, validated_data):
         endpoint = validated_data['endpoint']
         finding = validated_data['finding']
-        status = Endpoint_Status.objects.create(
-            finding=finding,
-            endpoint=endpoint
-        )
+        try:
+            status = Endpoint_Status.objects.create(
```
valentijnscholten, line 723: would `get_or_create` work here to make it simpler?

valentijnscholten:
> @valentijnscholten, so can I remove this?
> 
> ```python
> try:
>     ...
> except (MultipleObjectsReturned):
>     ...
> ```

I think so as the new unique constraint should  assure there is at most 1 matching row.

dojo/api_v2/serializers.py
kiblik: Well, It depends on our definition/understanding of the `POST` method. If we will say that
- `POST` request creates a new element if it doesn't exist and updates if exists, we can use `get_or_create` and I can remove `try ... except IntegrityError ...` part.
- `POST` request always creates a new element, we can not use `get_or_create`. Or we can use it but if `create == False` we should raise `ValidationError`.

Maffooch:
I would recommend testing with the sample data dataset https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/fixtures/defect_dojo_sample_data.json and using the management command to create the statuses (assuming they are not already there) https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/management/commands/create_endpoint_status.py

dojo/api_v2/serializers.py
valentijnscholten: Ah good point, I think we should indeed stay true to the restful meaning of POST

valentijnscholten:
I think the real hair would be in testing with real life data as that might contain the duplicate rows that are too be fixed by this PR. The sample data will already be correct by itself so won't trigger the important code paths.

kiblik:
Ouu, `PATCH` and `PUT` need to be fixed and tested as well

valentijnscholten (approved):
Looks like nobody can/will test it, so maybe we should let is "bake" in `dev` then :-)

PR #5949, opened 2022-02-22
Harmonize user format

dsever:
This PR changes:
-  from User model to Dojo_User
- notes to show `name last (username)`

The idea is to harmonize user format across the app.


valentijnscholten (approved):

mtesauro (approved):
Approved

Issue #4198, opened 2021-04-02
Right evaluation of endpoint status


kiblik:
Endpoint evaluates status (if it is vulnerable and active and/or mitigated) based on [Finding fields](https://github.com/DefectDojo/django-DefectDojo/blob/b3393ecf09562da93b987e1eb598e823ff25e92a/dojo/models.py#L1163-L1199). However the real status depends (also) on the content of `Endpoint_Status`.

I suppose this should be redesigned when https://github.com/DefectDojo/django-DefectDojo/pull/4188 / https://github.com/DefectDojo/django-DefectDojo/pull/4473 will be finished. 

Another reason: https://owasp.slack.com/archives/C2P5BA8MN/p1621324919005200

valentijnscholten:
Yeah, the endpoint_status thing wasn't 100% completed I think. There are also some performance issues around it which we need to tackle.

kiblik:
Before implementation check https://owasp.slack.com/archives/C2P5BA8MN/p1645522693583909

kiblik:
Fixed in https://github.com/DefectDojo/django-DefectDojo/pull/6193

PR #5953, opened 2022-02-23
Endpoint: Remove endpoint.mitigated field

kiblik:
Remove leftover `endpoint.mitigated` field. Any endpoint that is mitigated will get their `endpoint_statuses` mitigated as part of this PR via a migration.

Part of the solution of https://github.com/DefectDojo/django-DefectDojo/issues/4198


damiencarol (approved):

Maffooch (approved):
This looks good to me, but I am nervous to merge the day before a release. @valentijnscholten leaving this to you if think it's okay to merge

dojo/endpoint/views.py
```diff
@@ -397,14 +397,21 @@ def endpoint_bulk_update_all(request, pid=None):
                 if skipped_endpoint_count > 0:
                     add_error_message_to_response('Skipped mitigation of {} endpoints because you are not authorized.'.format(skipped_endpoint_count))
 
+                eps_count = 0
+
                 for endpoint in endpoints:
-                    endpoint.mitigated = not endpoint.mitigated
-                    endpoint.save()
+                    for eps in endpoint.endpoint_status:
```
valentijnscholten, line 403: might be good to prefetch the endpont_status relation to make it performant on large number of endpoints?

valentijnscholten:
If it's obsolete, why does it need a migration? Could it lead to surprises with endpoint_statuses being closed?

kiblik:
> If it's obsolete, why does it need a migration? Could it lead to surprises with endpoint_statuses being closed?

If somebody marked endpoint as mitigated (by GUI or API), I suppose his action/decision should be migrated.
On the other hand, mitigation should be changed by (re)importer, not by a user (if he/she is not doing manual pentest) and he/she should use flags "false positive", "out of scope" and "risk accepted"

dojo/endpoint/views.py
kiblik: Hmmm... and maybe use of [`.update()`](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#django.db.models.query.QuerySet.update) would be the best idea

valentijnscholten:
Looking at the PR it seems the endpoint.mitigated field is no longer shown in the UI? This migration would then use the value of this "hidden/no longer visible" field to start mitigating endpoint_statuses. This might be confusing for some users? Not sure, just thinking out loud. In theory the affected endpoint statuses should be small.

valentijnscholten:
I don't know the history of changes around this field to make a good judgement. I think impact will be small so might as well go for it.

dojo/endpoint/views.py
valentijnscholten: @kiblik are you planning on changing this so we can merge it?

kiblik: I will prepare it today/tomorrow.

kiblik:
The last requested change was implemented. Feel free to merge it.

PR #5955, opened 2022-02-24
Add CVSS3 to nessus XML parser

bjhijmans:
Currently only the nessus CSV parser parses cvss3 vectors, this adds the same functionality to the XML parser


kiblik:
Do you mind, you can add also the unittest for it?

bjhijmans:
> Do you mind, you can add also the unittest for it?

I'm not sure how to do that, I don't have access to a report I can share and none of the current test files have that field set. I guess I can add a cvss3 field to an existing finding in one of the test files, but that doesn't seem very neat.

Maffooch:
@bjhijmans please base your PR against the dev branch.

damiencarol (changes requested):
need data for unit tests

bjhijmans:
I've added a unit test for the parser.

dojo/tools/nessus/parser.py
```diff
@@ -197,6 +197,10 @@ def get_findings(self, file, test):
                     cwe = None
                     if item.findtext("cwe"):
                         cwe = item.find("cwe").text
+                    cvssv3 = None
+                    if item.findtext("cvss3_vector"):
+                        cvssv3 = CVSS3(str(item.find("cvss3_vector").text)).clean_vector(output_prefix=False)
```
damiencarol, line 202: the attribute `cvssv3` need the prefix.
Also please use the form `item.findtext("cvss3_vector")` directly

damiencarol (changes requested):

dojo/tools/nessus/parser.py
bjhijmans: > Also please use the form item.findtext("cvss3_vector") directly
What do you mean by that?

damiencarol: Something like this:
```Python
                    cvssv3 = None
                    if item.findtext("cvss3_vector"):
                        cvssv3 = CVSS3(item.findtext("cvss3_vector")).clean_vector()
```

bjhijmans: @damiencarol Thank you. I made those changes. 

However that does mean that the csv parser in the same file is wrong because it doesn't return the prefix, which is why I was confused when you mentioned it should.

bjhijmans:
I fixed the issues, but I noticed that pipelines seem to be failing.

Also #6172 seems to be trying to do the same thing.

damiencarol (approved):

StefanFl (approved):

unittests/tools/test_nessus_parser.py
```diff
@@ -152,3 +152,20 @@ def test_parse_some_findings_samples(self):
...
+            for endpoint in finding.unsaved_endpoints:
+                endpoint.clean()
+        self.assertEqual(48, len(findings))
+        finding = findings[3]
+        self.assertEqual("Medium", finding.severity)
+        self.assertIsNone(finding.cwe)
+        print(finding.unsaved_endpoints)
+        endpoint = finding.unsaved_endpoints[0]
+        self.assertEqual("http", endpoint.protocol)
+        self.assertEqual("CVSS3.0:AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:N/A:N", finding.cvssv3)
```
StefanFl, line 171: This line lets the unit test fail.

StefanFl (changes requested):

PR #5962, opened 2022-02-25
update result template in nuclei parser

fadhilthomas:
These updates related to breaking changes on nuclei
https://github.com/projectdiscovery/nuclei/releases/tag/v2.5.3

### DIFF
--- a/dojo/tools/nuclei/parser.py
+++ b/dojo/tools/nuclei/parser.py
@@ -29,9 +29,9 @@ class NucleiParser(object):
                 name = info.get('name')
                 severity = info.get('severity').title()
                 type = item.get('type')
-                matched = item.get('matched')
+                matched = item.get('matched-at')
                 if '://' in matched:
-                    endpoint = Endpoint.from_uri(matched-at)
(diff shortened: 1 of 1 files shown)


valentijnscholten:
There's already a PR in progress: https://github.com/DefectDojo/django-DefectDojo/pull/5664, but that seems to have stalled.
Are you able to update your PR to:
- support both old and new version of nuclei
- have unit tests for both versions

tequreq:
yes please upgrade parser because parser is not working.  @valentijnscholten

damiencarol:
@fadhilthomas could you add data for the new version and add unit tests?

fadhilthomas:
@damiencarol sure, will add the unit test.

damiencarol:
Last time was #5689 but it could be something as simple as checking first `template-id` and fall back to `templateID` if the first one is not here. (same logic for "matched at" change)
Also some parser check the version in the meta data of the report. There is no automatic way to do it. Just keep it simple.

StefanFl:
superseded by #6103

PR #5984, opened 2022-03-02
add support_request template

valentijnscholten:
As discussed on slack most GitHub issues created as bugs are really support requests. 
This PR allows creation of support requests, which get labeled as such so we don't have to keep relabeling bug reports to support requests :-)

Against `master` for immediate effect.


mtesauro (approved):
Approved

StefanFl (approved):

PR #5990, opened 2022-03-03
Fix missing DD_CELERY_BROKER_PATH variable

dsever:
This is fixing missing DD_CELERY_BROKER_PATH variable.


alles-klar:
Not sure if "//" is the best default but as is worked until now, let's leave it untouched.

valentijnscholten (approved):

PR #5997, opened 2022-03-04
base.html cleanup and formatting

Maffooch:
I was looking at base.html and noticed I was having a hard time reading it, so I cleaned it up.

The diff looks real large mainly because of change in indention and expanding ugly one liners


mtesauro (approved):
Approved

valentijnscholten (approved):

Issue #5999, opened 2022-03-07
npm/yarn audit parser fails on CWE field


valentijnscholten:
I don't know if this is something new, but `yarn` imports are failing when multiple CWE values are references for an advisory:
```
          "cwe":"[\"CWE-918\",\"CWE-1333\"]",
```

dvdgsng:
Experiencing the same with:

```
docker run -it node:16.13.0 yarn --version
1.22.15

```

grohmio:
Same problem with single CWE values, e.g.

`"cwe":"[\"CWE-79\"]"`

valentijnscholten:
this PR fixes that as well, covered by a unit test. I think yarn gets their advisories from upstream, which is where something might have changed. it's a weird format anyway.

valentijnscholten:
Fixed in 'dev' / 2.9.0

dvdgsng:
Thanks for fixing this so quickly, looking forward to v2.9.0 <3

valentijnscholten:
Reopened as of today it seems `npm audit` reports are now suffering from the same issue: 
`"cwe": "[\"CWE-22\",\"CWE-59\"]",`

valentijnscholten:
finally fixed also yarn.... now they have seemed to change the format again....

```
  File "./dojo/tools/utils.py", line 10, in get_npm_cwe
    if cwe_node.startswith('CWE-'):
AttributeError: 'list' object has no attribute 'startswith'
```

```
        "cwe": [
          "CWE-346",
          "CWE-453"
        ],
```

let's wait if this stabilizes.

mtesauro:
Yeah, what a mess.

valentijnscholten:
merged into `dev` for `2.9.0`

PR #6000, opened 2022-03-07
yarn: fix CWE parsing

valentijnscholten:
fix yarn parser to handle wildly formatted list of CWEs: #5999

        # Somehow multiple CWEs end up in the json report like this
        # [\"CWE-173\",\"CWE-200\",\"CWE-601\"]
        # which becomes after json load:
        # ["CWE-173","CWE-200","CWE-601"]
        # we parse this and take the first CWE

EDIT: Seems to happen also for single CWE values

        # [\"CWE-173\"]


dvdgsng:
Any idea what DefectDojo is doing when running an import that could cause such an issue out of the blue? We are running v2.7.1 and haven't touched the deployment for 25 days. So we would expect it to be rather stable.

valentijnscholten:
yarn has changed their output format or logic, or gets data from another source formatted in a different way.

dvdgsng:
AFAICT we haven't changed the node version either :/ need to verify though

valentijnscholten:
It might be a bug / side effect in yarn that they may fix at some point.

dojo/tools/yarn_audit/parser.py
```diff
@@ -54,6 +55,25 @@ def get_item(item_node, test):
...
+        # we parse this and take the first CWE
+
+        cwe = ast.literal_eval(cwe)[0][4:]
+        # https://docs.python.org/3/library/ast.html#ast.literal_eval
+        # Safely evaluate an expression node or a string containing a Python literal or container display.
+        # The string or node provided may only consist of the following Python literal structures:
+        # strings, bytes, numbers, tuples, lists, dicts, sets, booleans, None and Ellipsis.
+        # This can be used for safely evaluating strings containing Python values from untrusted sources
+        #  without the need to parse the values oneself. It is not capable of evaluating arbitrarily
+        #  complex expressions, for example involving operators or indexing.
```
damiencarol, line 75: I think it's better to use `json.loads()`. The data in the report is not Python code.
It's JSON object in a JSON string. so better to load it this way (with the JSON module)

damiencarol (changes requested):

dojo/tools/yarn_audit/parser.py
```diff
@@ -54,6 +54,18 @@ def get_item(item_node, test):
...
+    if cwe.startswith('CWE-'):
+        cwe = int(cwe[4:])
+    elif cwe.startswith('['):
+        # Somehow multiple CWEs end up in the json report like this
+        # [\"CWE-173\",\"CWE-200\",\"CWE-601\"]
+        # which becomes after json load:
+        # ["CWE-173","CWE-200","CWE-601"]
+        # we parse this and take the first CWE
+
+        cwe = json.loads(cwe)[0][4:]
```
damiencarol, line 67:
```suggestion
        cwe = json.loads(cwe)[0]
```
Should it be like this?

valentijnscholten: No, cwe is an integer field.

damiencarol (approved):

devGregA (approved):

Issue #4285, opened 2021-04-15
Support for npm audit v7+


valentijnscholten:
**Scanner Name**
npm7 audit is not supported because it is missing lots of data in the reports: https://github.com/npm/npm-audit-report/issues/45

valentijnscholten:
Progress has been made, waiting for https://github.com/npm/metavuln-calculator/pull/34 to be merged.

jake-cryptic:
Hi, any update on this? The PR was merged and a new release made: https://github.com/npm/metavuln-calculator/releases/tag/v3.1.0

valentijnscholten:
They still don't include `cve`, but it seems it is hard for them to get that data and include it in the report. We could do a first implementation without `cve`.
They also don't include any other `id` we could easily use, except that we could parse the `url` field and hope for the best.

valentijnscholten:
Some more info here: https://github.com/jeemok/better-npm-audit#npm-version-6-and-7-and-8

manuel-sommer:
NPM Audit is already covered. I guess, this can be closed @mtesauro

valentijnscholten:
I don't think npm 7 or higher is supported by the parser.

manuel-sommer:
Do you have a sample file, then I will fix the parser ?

manuel-sommer:
I guess this can be closed @mtesauro

PR #6016, opened 2022-03-11
npm/yarn audit: fix CWE parsing

valentijnscholten:
`npm audit` reports are suffering from the same thing as `yarn audit` reports in #5999

CWE values are references for an advisory:
```
          "cwe":"[\"CWE-918\",\"CWE-1333\"]",
```
After fixing that, the format changed without warning into:
```
        "cwe": [
          "CWE-346",
          "CWE-453"
        ],
```

So now this PR supports all 3 possible formats for both `yarn audit` and `npm audit`.
New example files are real life examples.


valentijnscholten:
hmmm after fixing this and making the unit tests pass, it seems the output has changed again. now it is a proper list:

```
        "cwe": [
          "CWE-346",
          "CWE-453"
        ],
```

let's wait a couple of days to see what else they have up their sleeves :-)

damiencarol:
> let's wait a couple of days to see what else they have up their sleeves :-)

XD

dojo/tools/utils.py
```diff
@@ -0,0 +1,30 @@
...
+            cwe_string = cwe_node[4:]
+            if cwe_string:
+                return int(cwe_string)
+        elif cwe_node.startswith('['):
+            cwe = json.loads(cwe_node)
+            if cwe:
+                return int(cwe[0][4:])
+
+    # Use CWE-1035 as fallback (vulnerable third party component)
+    return 1035
```
damiencarol, line 30: This choice is not always true. If we want to use this function for all parsers, sometimes it's not component-oriented. could we just return an empty array?

valentijnscholten: this function is for npm/yarn only

StefanFl:
@valentijnscholten Which version of `npm` do you use? With the most recent version 8.5.4 I get a different looking JSON file without the advisories.

valentijnscholten:
Npm 6. Anything above that is not supported, https://github.com/DefectDojo/django-DefectDojo/issues/4285

StefanFl (approved):

damiencarol (approved):

PR #6022, opened 2022-03-14
[Helm] Fix celery deployments

### LINKED ISSUE #6021: [Helm] Celery beat & Celery worker deployment password is broken
**Bug description**

In the `values.yml` file we can customize the name of the existing secret for rabbitmq or redis:

- rabbitmq: https://github.com/DefectDojo/django-DefectDojo/blob/caf5c91b3f8870d5f466dfaaf5a3a096f8812ad9/helm/defectdojo/values.yaml#L359-L367
- redis: https://github.com/DefectDojo/django-DefectDojo/blob/caf5c91b3f8870d5f466dfaaf5a3a096f8812ad9/helm/defectdojo/values.yaml#L387-L394

In the snippets above, we can see that they are all by default `defectodjo-{{ service }}-specific`. If we decide to change those values, then the deployment will fail because of the celery deployments.  
In the following snippets, we can see the problematic lines:

- celery-beat: https://github.com/DefectDojo/django-DefectDojo/blob/caf5c91b3f8870d5f466dfaaf5a3a096f8812ad9/helm/defectdojo/templates/celery-beat-deployment.yaml#L102-L121

- celery-worker: https://github.com/DefectDojo/django-DefectDojo/blob/caf5c91b3f8870d5f466dfaaf5a3a096f8812ad9/helm/defectdojo/templates/celery-worker-deployment.yaml#L97-L116

Now, I have a few issues with this:

1. What is the point of naming them `defectdojo-{{ .Values.celery.broker }}-specific` if just above we are doing a `{{- if eq .Values.celery.broker "rabbitmq" }}`? Why don't we name them `defectdojo-rabbitmq-specific` directly. (applicable to the redis service as well. rabbitmq was just used as an example)
2. If we decide to use rabbitmq and change the `existingPasswordSecret` for some reason, then the deployment would fail since `{{ .Values.rabbitmq.enabled }}` will evaluate to true. As such, the `secretKeyRef` would use the "hardcoded" `defectdojo-rabbitmq-specific` value instead of the one we customized.
  I don't really see the use of that `if` statement. If we enable and use the service, then use the hardcoded value. If we do not, then use the value we defined?? I am going to guess it assumes that if we have `enabled: False` then we don't specify any further config for the service, which is valid, i guess, but what if we are just testing something and setting enabled to `false`? Then the deployment would use the custom secret we have configured, which may not exist, if we have the service `enabled: False`.
  Worst case scenario, I think it should always be set to `{{ .Values.rabbitmq.existingPasswordSecret }}` so that we can actually customize the secret name.

Interestingly enough, this is NOT the case for postgres or mysql. There, the correct values are used. 

**Steps to reproduce**
1. Define a custom `existingSecret` for any of "utility" services: redis or rabbitmq
2. Deploy the chart

**Expected behavior**

The celery pods should use the custom secrets set for the services, not hardcoded ones.

- [ ] Docker Compose
- [x] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version (see footer) or commit message: 2.4.1 (but the same configuration is present in the latest release)

mirceanton:
Fixes #6021


dsever:
@mirceanton  is this PR ready?

mirceanton:
@dsever i noticed i have a failing test and i didn't have time to look into it yet, but the main idea is simple. The if statement itself should be re-written to match the other services.

I will try to take another look at it today or tomorrow hopefully. Maybe someone else can throw a quick look at it too?

mirceanton:
Accidentally closed this PR by force pushing to undo some changes, hehe 😅 

Opened PR #6054 to fix the issue.

PR #6029, opened 2022-03-16
APIv2: remove configuration_url from ToolConfigurationSerializer

kiblik:
remove obsolete field

https://owasp.slack.com/archives/C2P5BA8MN/p1647430817715979

### DIFF
--- a/docs/content/en/getting_started/upgrading.md
+++ b/docs/content/en/getting_started/upgrading.md
@@ -61,6 +61,10 @@ godojo installations
 
 If you have installed DefectDojo on "iron" and wish to upgrade the installation, please see the [instructions in the repo](https://github.com/DefectDojo/godojo/blob/master/docs-and-scripts/upgrading.md).
 
+## Upgrading to DefectDojo Version 2.9.x.
+
+**Breaking change for APIv2:** From API endpoint `/api/v2/tool_configurations/` was removed `configuration_url` because of redundancy.
+
 ## Upgrading to DefectDojo Version 2.8.x.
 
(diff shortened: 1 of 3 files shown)


valentijnscholten:
Isn't this field used to set the `url` field?

kiblik:
> Isn't this field used to set the `url` field?

No, `url` is still [there](https://github.com/DefectDojo/django-DefectDojo/blob/caf5c91b3f8870d5f466dfaaf5a3a096f8812ad9/dojo/models.py#L972). 
`configuration_url` is not used in the rest of the code. Only in this one place (+ unittest).

Example from swagger (without the change in this PR).
```json
{
  "configuration_url": "string",
  "name": "string",
  "description": "string",
  "url": "string",
  "authentication_type": "API",
  "extras": "string",
  "username": "string",
  "password": "string",
  "auth_title": "string",
  "ssh": "string",
  "api_key": "string",
  "tool_type": 0
}
```

valentijnscholten:
That's what I mean. People can use it to set the url field. Like an alias.

StefanFl:
I agree with @valentijnscholten, it might not be a big deal but it's still a breaking change. We should have a short note in the upgrade documentation about it.

docs/content/en/getting_started/upgrading.md
```diff
@@ -61,6 +61,10 @@ godojo installations
 
 If you have installed DefectDojo on "iron" and wish to upgrade the installation, please see the [instructions in the repo](https://github.com/DefectDojo/godojo/blob/master/docs-and-scripts/upgrading.md).
 
+## Upgrading to DefectDojo Version 2.9.x.
+
+**Breaking change for APIv2:** From API endpoint `/api/v2/tool_configurations/` was removed `configuration_url` because of redundancy.
```
StefanFl, line 66: This would sound a bit better:

**Breaking change for APIv2:** `configuration_url` was removed from API endpoint `/api/v2/tool_configurations/` due to redundancy.

StefanFl (changes requested):

StefanFl (approved):

valentijnscholten (approved):

PR #6041, opened 2022-03-18
Logging enhancement 

dsever:
It adds options to define customized way of uwsgi logging, that can be changed without need to rebuild the images. 

For instance user can specify Apache format by defining (in compose or helm value):

`DD_UWSGI_LOGFORMAT: '"%(addr) - %(user) [%(ltime)] "%(method) %(uri) %(proto)" %(status) %(size) "%(referer)" "%(uagent)"'`

To produce:
`"172.19.0.1 - - [18/Mar/2022:12:52:21 +0000] "GET /alerts/count HTTP/1.1" 200 183 "http://localhost:8080/system_settings" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.109 Safari/537.36 OPR/84.0.4316.31"`

The problem I still see that default is not populating (user) variable, I tried to do it manually in middleware with `set_logvar` but is not working as well (still empty). But it works fine if I define alternative logvar in the code like user1.
like 

``DD_UWSGI_LOGFORMAT: '"%(addr) - %(user1) [%(ltime)] "%(method) %(uri) %(proto)" %(status) %(size) "%(referer)" "%(uagent)"'``

### DIFF
--- a/docker/entrypoint-uwsgi-dev.sh
+++ b/docker/entrypoint-uwsgi-dev.sh
@@ -14,8 +14,6 @@ if [ ${DD_DEBUG} == "True" ]; then
   DD_UWSGI_NUM_OF_THREADS=1
 fi
 
-echo $DD_UWSGI_LOGFORMAT
-
 exec uwsgi \
   "--${DD_UWSGI_MODE}" "${DD_UWSGI_ENDPOINT}" \
   --protocol uwsgi \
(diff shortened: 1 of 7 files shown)


StefanFl:
`%(user)` shall be taken from REMOTE_USER of the `wsgi_request` structure, according to https://uwsgi-docs.readthedocs.io/en/latest/LogFormat.html#offsetof. Maybe you can't override that with `set_logvar`. I couldn't find any information if Django can set this user.

Maybe it's best to go with a variable like `dd_user`, if you can't find any other solution.

dsever:
Am I happy with this? No. So I'm considering this as minor improvement of the security logging.

damiencarol (approved):

PR #6053, opened 2022-03-22
Finding group enhancements

dsever:
- This PR shows group real names names instead of Y and N.
- removes 'Experimental notification'
- and moves configuration (enable/disable) from config file to DB
image
image
image


dojo/utils.py
```diff
@@ -2124,3 +2124,8 @@ def get_enabled_notifications_list():
         if get_system_setting('enable_{}_notifications'.format(choice[0])):
             enabled.append(choice[0])
     return enabled
+
+
+def is_finding_feature_group_enabled():
```
StefanFl, line 2129:
```suggestion
def is_finding_groups_enabled():
```
The `feature_` part is not necessary here (would be in the wrong place anyway). And `groups` with an `s` in the end lets it fit better to the model

StefanFl (changes requested):

dojo/utils.py
dsever: Done

dojo/models.py
```diff
@@ -455,6 +455,11 @@ class System_Settings(models.Model):
         blank=False,
         verbose_name=_('Enable Product Tracking Files'),
         help_text=_("With this setting turned off, the product tracking files will be disabled in the user interface."))
+    enable_feature_findings_groups = models.BooleanField(
```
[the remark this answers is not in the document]
dsever: Done

valentijnscholten (changes requested):
I think this PR should either migrate the existing setting into the DB for existing installs and/or add something to the upgrade notes. Ideally if users have disabled finding groups, it should remain disabled after upgrading to the version that includes this PR.

dsever:
> I think this PR should either migrate the existing setting into the DB for existing installs and/or add something to the upgrade notes. Ideally if users have disabled finding groups, it should remain disabled after upgrading to the version that includes this PR.

This is true, will try to add it

dsever:
@StefanFl  @valentijnscholten  Changes have been implemented

dsever:
> Thanks. Experience has shown it's better to separate schema migrations and data migrations. Data migrations sometimes fail and having them in a separate migration allows for easy retry without having to comment out any schema parts that were successful. Could you split them up?

@valentijnscholten, Done

StefanFl (approved):

valentijnscholten (approved):

PR #6054, opened 2022-03-22
Fix flawed logic in celery deployment env vars

mirceanton:


mirceanton:
@dsever
I accidentally closed #6022 when force pushing to undo some changes 😅 , so I opened this new one.

To my (rather limited) testing, this seems fine and it **should** work.

Basically, the logic that was implemented before was something along the lines of: 

https://github.com/DefectDojo/django-DefectDojo/blob/caf5c91b3f8870d5f466dfaaf5a3a096f8812ad9/helm/defectdojo/templates/celery-beat-deployment.yaml#L102-L121


**L106**: Why would rabbitmq be disabled if the broker is set to "rabbitmq"? wouldn't that if statement always evaluate to `true` in a correct deployment?
**L107**: Why use `defectdojo-{{ .Values.celery.broker }}-specific` when we already know that `{{ .Values.celery.broker }}` is "rabbitmq" due to the `if` condition?
**L108**: Same as **L107**, shouldn't this be `rabbitmq-password` directly? what's the point of templating when we already know the value?

**L105->L109**: If rabbitmq is enabled, then ignore the custom values the user has set for secrets and just use the default ones??
**L109->L111**: If rabbitmq is NOT enabled, then use the custom values the user has set for the secrets??

This doesn't make sense to me.  
We use default values if the service is enabled, so the user is likely to have set a custom secret, so the deployment will not work. On the other hand, if the service is not enabled, so likely the user has not configured it then try to use the `existingSecret`?

Am I getting this wrong?


What I tried to implement now is to simply always try to use the custom `existingSecret` values and if they are not configured then default to the `defectdojo-<SERVICE>-specific` secret name. This makes more sense in my head. Opinions?

mirceanton:
@dsever on my local branch, all tests passed: https://github.com/mirceanton/django-DefectDojo/actions/runs/2025191532

Here, it seems one failed for some reason. Ideas?

helm/defectdojo/templates/celery-beat-deployment.yaml
```diff
@@ -102,32 +102,24 @@ spec:
...
-            {{- else }}
-              name: {{ .Values.redis.existingSecret }}
-              key: {{ .Values.redis.secretKey }}
-            {{- end }}  
-            {{- end }}
+              {{- if eq .Values.celery.broker "rabbitmq" }}
+              name: {{ .Values.rabbitmq.existingPasswordSecret | default "defectdojo-rabbitmq-specific" }}
+              key: {{ .Values.rabbitmq.secretPasswordKey | default "rabbitmq-password" }}  
+              {{- end -}}
+              {{- if eq .Values.celery.broker "redis" }}
```
dsever, line 109: Any reason not to use else, broker can be redis or rabbit. nothing else

helm/defectdojo/templates/celery-beat-deployment.yaml
```diff
@@ -102,32 +102,24 @@ spec:
...
-            {{- if .Values.redis.enabled }}
-              name: defectdojo-{{ .Values.celery.broker }}-specific
-              key: {{ .Values.celery.broker }}-password
-            {{- else }}
-              name: {{ .Values.redis.existingSecret }}
-              key: {{ .Values.redis.secretKey }}
-            {{- end }}  
-            {{- end }}
+              {{- if eq .Values.celery.broker "rabbitmq" }}
+              name: {{ .Values.rabbitmq.existingPasswordSecret | default "defectdojo-rabbitmq-specific" }}
```
dsever, line 106: I would say path from values is wrong should be `.Values.rabbitmq.auth.existingPasswordSecret`

helm/defectdojo/templates/celery-beat-deployment.yaml
```diff
@@ -102,32 +102,24 @@ spec:
...
+              key: {{ .Values.rabbitmq.secretPasswordKey | default "rabbitmq-password" }}  
+              {{- end -}}
+              {{- if eq .Values.celery.broker "redis" }}
+              name: {{ .Values.redis.existingSecret | default "defectdojo-redis-specific" }}
```
dsever, line 110: The same is for redis

helm/defectdojo/templates/celery-beat-deployment.yaml
mirceanton: I tried to follow the same pattern I saw above, in the `cloudsql-proxy` container `args` section, but yeah, that was a mistake on my part they're different. unsure what I was thinking 😆

helm/defectdojo/templates/celery-beat-deployment.yaml
mirceanton: [Fixed](https://github.com/DefectDojo/django-DefectDojo/pull/6054/commits/11a4447cf457d9c791ee2251119a6a862f14c9fe).

helm/defectdojo/templates/celery-beat-deployment.yaml
mirceanton: [Fixed](https://github.com/DefectDojo/django-DefectDojo/pull/6054/commits/11a4447cf457d9c791ee2251119a6a862f14c9fe).

mirceanton:
Hmm, upon further testing, it appears that the nginx container inside the defectdojo pod fails the healthcheck. I am unsure why. Currently troubleshooting.

It seems to be failing with codes `500` or `502`.

alles-klar:
> @dsever I accidentally closed #6022 when force pushing to undo some changes 😅 , so I opened this new one.
> 
> To my (rather limited) testing, this seems fine and it **should** work.
> 
> Basically, the logic that was implemented before was something along the lines of:
> 
> https://github.com/DefectDojo/django-DefectDojo/blob/caf5c91b3f8870d5f466dfaaf5a3a096f8812ad9/helm/defectdojo/templates/celery-beat-deployment.yaml#L102-L121
> 
> **L106**: Why would rabbitmq be disabled if the broker is set to "rabbitmq"? wouldn't that if statement always evaluate to `true` in a correct deployment? **L107**: Why use `defectdojo-{{ .Values.celery.broker }}-specific` when we already know that `{{ .Values.celery.broker }}` is "rabbitmq" due to the `if` condition? **L108**: Same as **L107**, shouldn't this be `rabbitmq-password` directly? what's the point of templating when we already know the value?
> 
> **L105->L109**: If rabbitmq is enabled, then ignore the custom values the user has set for secrets and just use the default ones?? **L109->L111**: If rabbitmq is NOT enabled, then use the custom values the user has set for the secrets??
> 
> This doesn't make sense to me. We use default values if the service is enabled, so the user is likely to have set a custom secret, so the deployment will not work. On the other hand, if the service is not enabled, so likely the user has not configured it then try to use the `existingSecret`?
> 
> Am I getting this wrong?
> 
> What I tried to implement now is to simply always try to use the custom `existingSecret` values and if they are not configured then default to the `defectdojo-<SERVICE>-specific` secret name. This makes more sense in my head. Opinions?

@mirceanton I think this logic comes from the usecase of an external redis/rabbitmq (outside of the helm chart). Does it make sense to you?

mirceanton:
@alles-klar It makes more sense, yes, but I still don't think it is correct. In this case, it still doesn't allow me to use a custom secret name for an "internal" rabbitmq/redis.

alles-klar:
Sounds valid.

alles-klar (approved):

italvi:
@dsever @alles-klar why was this PR closed without being implemented? This is a change we would require as we want to have two different systems using two different secrets but with the current implementation it is not possible to use a custom secret

dsever:
@italvi see https://github.com/DefectDojo/django-DefectDojo/pull/6149

italvi:
@dsever but it only closed it for redis and not rabbitmq. There it is still wrong. And it does not make sense that I have to disable rabbitmq, s.t. it uses own variables and not -specific. I will open a PR for it then?

PR #6059, opened 2022-03-23
Add Hydra Parser

fhoeborn-cb:
Adds a Defect Dojo parser capable of importing the result content of the [THC Hydra Scanner](https://github.com/vanhauser-thc/thc-hydra) to DefectDojo.

Side Note: Hydra doesn't supply a severity (as its purpose is to find weak logins), so we assume a static severity of "High"


manuel-sommer (changes requested):
Hi, thank you for contributing. 

If you want to add a parser, you can stick to this document:
https://github.com/DefectDojo/django-DefectDojo/blob/master/docs/content/en/contributing/how-to-write-a-parser.md

Could you also add documentation with your PR:
https://github.com/DefectDojo/django-DefectDojo/blob/master/docs/content/en/integrations/parsers.md

fhoeborn-cb:
Thanks for the hint. I missed that, I added documentation

docker-compose.yml
```diff
@@ -13,6 +13,7 @@ services:
       context: ./
       dockerfile: Dockerfile.nginx
     image: "defectdojo/defectdojo-nginx:${NGINX_VERSION:-latest}"
+    platform: linux/x86_64
```
StefanFl, line 16: This line shouldn't be here, please remove it.

docker-compose.yml
```diff
@@ -131,6 +132,7 @@ services:
           target: /app/docker/extra_settings
   mysql:
     image: mysql:5.7.37@sha256:66d52e6baa8093820c09fec56992a5ee734f17e9fad8ef5ffc31597b231bd048
+    platform: linux/x86_64
```
StefanFl, line 135: This line shouldn't be here, please remove it.

dojo/tools/hydra/parser.py
```diff
@@ -0,0 +1,91 @@
...
+        username = raw_finding['login']
+        password = raw_finding['password']
+
+        finding = Finding(
+            test=test,
+            title="Weak username / password combination found for " + host,
+            date=parse_datetime(metadata.date),
+            severity="High",
+            description=host + " on port " + str(port) + " is allowing logins with easy to guess username " + username + " and password " + password,
+            static_finding=metadata.static_finding,
```
StefanFl, line 62: It would be easier to read if you would set `static_finding` here directly, it is a constant anyway.

dojo/tools/hydra/parser.py
```diff
@@ -0,0 +1,91 @@
...
+            dynamic_finding=metadata.dynamic_finding,
```
StefanFl, line 63: It would be easier to read if you would set `dynamic_finding` here directly, it is a constant anyway.

dojo/tools/hydra/parser.py
```diff
@@ -0,0 +1,91 @@
...
+            service=metadata.service_type,
+        )
+
```
StefanFl, line 66: Since you have `host` and `port`, wouldn't it make sense to build an Endpoint with the Finding?

dojo/tools/hydra/parser.py
```diff
@@ -0,0 +1,91 @@
...
+    def __parse_json(json_output):
+        report = json.load(json_output)
+
+        if 'generator' not in report or 'results' not in report:
+            raise ValueError(" Unexpected JSON format provided. That doesn't look like a Hydra scan!")
+
+        return report
+
+    @staticmethod
+    def __hyperlink(link: str) -> str:
```
StefanFl, line 79: I can't see where this method is called.

dojo/tools/hydra/parser.py
```diff
@@ -0,0 +1,91 @@
...
+        return '[' + link + '](' + link + ')'
+
+    @staticmethod
+    def __endpoint_status(status: str) -> str:
```
StefanFl, line 83: I can't see where this method is called.

dojo/settings/settings.dist.py
```diff
@@ -1070,6 +1070,7 @@ def saml2_attrib_map_format(dict):
     'Harbor Vulnerability Scan': ['title'],
     'Rusty Hog Scan': ['title', 'description'],
     'StackHawk HawkScan': ['vuln_id_from_tool', 'component_name', 'component_version'],
+    'Hydra Scan': ['title', 'description'],
```
StefanFl, line 1073: Not at this line but in the same file: Deduplication works better, if you set `DEDUPLICATION_ALGORITHM_PER_PARSER` to `DEDUPE_ALGO_HASH_CODE` for this parser.

StefanFl (changes requested):
Hi @fhoeborn-cb, thank you for this new parser. I have some remarks, but they should be easy to implement.

dojo/tools/hydra/parser.py
fhoeborn-cb: You got me, forgot to remove that :-)

dojo/tools/hydra/parser.py
fhoeborn-cb: Endpoint is now added as well

dojo/tools/hydra/parser.py
fhoeborn-cb: Adjusted

dojo/tools/hydra/parser.py
fhoeborn-cb: Adjusted

fhoeborn-cb:
Thanks for the review!
Adjusted the found issues

StefanFl (approved):

fhoeborn-cb:
@manuel-sommer Did I miss some requested change from your review?

dojo/tools/hydra/parser.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class HydraScanMetadata:
+    def __init__(self, generator):
+        self.date = generator['built']
+        self.command = generator['commandline']
+        self.schema_version = generator['jsonoutputversion']
+        self.service_type = generator['service']
+        self.tool_version = generator['version']
+        self.server = generator['server']
```
valentijnscholten, line 14: are these fields always present? a lot of parser fail at some point due to missing fields.

valentijnscholten (changes requested):
a

dojo/tools/hydra/parser.py
```diff
@@ -0,0 +1,75 @@
...
+            finding = self.__extract_finding(raw_finding, metadata, test)
+            findings.append(finding)
+
+        return findings
+
+    def __extract_finding(self, raw_finding, metadata: HydraScanMetadata, test) -> Finding:
+        host = raw_finding['host']
+        port = raw_finding['port']
+        username = raw_finding['login']
+        password = raw_finding['password']
```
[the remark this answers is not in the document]
fhoeborn-cb: Yes, according to tool documentation those should always be there.

dojo/tools/hydra/parser.py
fhoeborn-cb: Yes, according to tool documentation those should always be there.

valentijnscholten: I think we usually instruct authors to use something like `self.date = generator.get('built')` to reduce the chances of future exceptions on missing/renamed fields.

fhoeborn-cb: Adjusted it accordingly - if some vital information is missing for findings (even though that should not happen) a warning is logged and that finding skipped

Maffooch (approved):

PR #6064, opened 2022-03-23
[ADD] Allow HTML img tag resizing

X0x1RG9f:
By default, it is possible to add some HTML tag. However, width, height and style attribute are not allowed.
This commits add possibility to add these attributes. Style can only be used in addition with background-color tag (in case of white PNG).

### DIFF
--- a/dojo/templatetags/display_tags.py
+++ b/dojo/templatetags/display_tags.py
@@ -36,16 +36,21 @@ markdown_tags = [
     "img",
     "a",
     "sub", "sup",
+    "center",
 ]
 
 markdown_attrs = {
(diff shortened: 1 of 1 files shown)


StefanFl:
Hi @X0x1RG9f, please post a snippet how to use the tags, to make it easier for me to test it.

X0x1RG9f:
Hi @StefanFl,

Code can be used as follow in the descriptions (Product Types, Products, Findings, etc.):
- `GitHub`
- `<center>GitHub</center>`
- `GitHub`

valentijnscholten (approved):

PR #6066, opened 2022-03-23
[FIX] Issues on disconnection and connection using Social Auth

X0x1RG9f:
Fix two issues that appear when SOCIAL_AUTH and "SHOW_LOGIN_FORM = False" are combined (no /login form):
- When disconnecting, as return URL is not /login and does not contains next parameter, an error 500 is triggered.
- When connecting, message "You have logged out" is displayed after logging in.


dsever:
@X0x1RG9f  how to test it, to trigger error, can be it replicated using SAML auth?

X0x1RG9f:
> @X0x1RG9f how to test it, to trigger error, can be it replicated using SAML auth?

I don't know if it can be replicated with SAML. Test is easy, activating the two parameters in my first post. SOCIAL AUTH (I used Gsuite) and "SHOW_LOGIN_FORM = False". Then, when you are logged in and try to logout, you will trigger a 500 error.

dsever:
> > @X0x1RG9f how to test it, to trigger error, can be it replicated using SAML auth?
> 
> @dsever I don't know if it can be replicated with SAML. Test is easy, activating the two parameters in my first post. SOCIAL AUTH (I used Gsuite) and "SHOW_LOGIN_FORM = False". Then, when you are logged in and try to logout, you will trigger a 500 error.

Ok, I will give a try.

dojo/finding/views.py
```diff
@@ -687,7 +687,7 @@ def copy_finding(request, fid):
                                 title='Copying of %s' % finding.title,
                                 description='The finding "%s" was copied by %s to %s' % (finding.title, request.user, test.title),
                                 product=product,
-                                url=request.build_absolute_uri(reverse('copy_finding', args=(finding_copy.id, ))),
+                                url=request.build_absolute_uri(reverse('finding_copy', args=(finding_copy.unsaved_vulnerability_ids, ))),
```
damiencarol: `unsaved_vulnerability_ids` is an array that don't have finding id. I don't think this will work

damiencarol (changes requested):

damiencarol (approved):

devGregA:
@X0x1RG9f  thank you for the PR. Looks good. Will merge after the release for further testing.

PR #6074, opened 2022-03-23
[FIX] Typo fix on migrate_staff_users

TheocharisPetros:
Currently there is a typo in the documentation while the migrate_staff_users is the script that needs to be executed, the underscore was missing in the current documentation.


mtesauro (approved):
Approved

StefanFl (approved):

PR #6077, opened 2022-03-24
Fix entrypoint when running in Kubernetes

MXfive:
Adds a check to see if `/app/docker/extra_settings` exists before trying to enumerate settings files from it.


valentijnscholten:
please rebase against `dev` and check if it's not already fixed there

MXfive:
Aha yes it is, thanks, is that branch part of a published Docker build pipeline somewhere that I can pull from?

PR #6079, opened 2022-03-24
[FIX] Wrong Endpoint 'Date Discovered'

X0x1RG9f:
When creating a new Finding in the past for an Endpoint (new or already existing one), the 'Date Discovered' of this Endpoint (for the corresponding Finding) is always set to the today's date as defaulted in models.py files.
This date should be fixed to the same value than the Finding's date, because the 'Date Discovered' is linked between the Endpoint and the Finding


valentijnscholten (approved):

mtesauro (approved):
Approved

Issue #6092, opened 2022-03-29
SAML2 force_authn environment variable


italvi:
**Is your feature request related to a problem? Please describe**
I want to make SSO with SAML2 possible.

**Describe the solution you'd like**
As an admin, I want to set an evironment variable to force_authn in SAML2, so that an user do not have to provide his credentials always.

**Additional context**
Provide a new Environment Variable, e.g., DD_SAML2_FORCE_AUTH and set force_authn to env('DD_SAML2_FORCE_AUTH') in the SAML_CONFIG dictionary

italvi:
@dsever Done, but unfortunately the checker says it does not work. Seems like I did something wrong?

StefanFl:
merged for release 2.9.0

PR #6093, opened 2022-03-29
Update settings.dist.py to force_auth for SAML2

italvi:
Make SSO over SAML2 possible (#6092)

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -129,7 +129,7 @@ env = environ.Env(
     DD_SOCIAL_AUTH_GITHUB_ENTERPRISE_SECRET=(str, ''),
     DD_SAML2_ENABLED=(bool, False),
     # Force Authentication to make SSO possible with SAML2
-    DD_SAML2_FORCE_AUTH =(bool, True),
+    DD_SAML2_FORCE_AUTH=(bool, True),
     DD_SAML2_LOGIN_BUTTON_TEXT=(str, 'Login with SAML'),
     # Optional: display the idp SAML Logout URL in DefectDojo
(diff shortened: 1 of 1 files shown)


dojo/settings/settings.dist.py
```diff
@@ -848,7 +850,7 @@ def saml2_attrib_map_format(dict):
                 'name_id_format': saml2.saml.NAMEID_FORMAT_TRANSIENT,
                 'want_response_signed': False,
                 'want_assertions_signed': True,
-                'force_authn': True,
+                'force_authn': env('DD_SAML2_FORCE_AUTH'),
```
dsever, line 853: Do not use env variables directly, config should support config py files as well 

Please take a look:

https://github.com/italvi/django-DefectDojo/blob/patch-1/dojo/settings/settings.dist.py#L826

dsever (changes requested):
Please take a look at my comment

dojo/settings/settings.dist.py
dsever: Actually this is my mistake :/ can you fix this as well?

StefanFl (approved):
Hi @italvi, thank you for your contribution, lgtm

dsever (approved):
Looks ok

nobletrout:
Is this a breaking change for existing SAML logins?

dsever:
@nobletrout  do you have some issues after the upgrade?

nobletrout:
No I haven't upgraded yet was reading release notes and raising a concern

italvi:
The default value is still True, therefore it should not have any impact on existing SAML Logins.

PR #6094, opened 2022-03-29
Bulk notes for the findings 

dsever:
This PR brings finding bulk update for notes, the use case is simple. When analyst walks though findings and wants to make multiple findings as FP, probably would be clever to leave some comment why he did it. 

Please  test it.

- [x] Add new feature
- [ ] Add tests


StefanFl:
Hi @dsever, is this PR good for review or do you want to add tests before we review?

dsever:
> Hi @dsever, is this PR good for review or do you want to add tests before we review?

Give me day or 2 to add 1-2 test.

dojo/finding/views.py
```diff
@@ -1908,11 +1907,24 @@ def finding_bulk_update_all(request, pid=None):
                             else:
                                 add_external_issue(finding, 'github')
 
+                if form.cleaned_data['notes']:
+                    logger.info('Setting bulk notes')
```
StefanFl, line 1911: I think this should be a `debug` log.

dojo/finding/views.py
```diff
@@ -1908,11 +1907,24 @@ def finding_bulk_update_all(request, pid=None):
...
+                    for finding in finds:
+                        finding.notes.add(note)
+                        finding.save()
+
                 if form.cleaned_data['tags']:
                     for finding in finds:
                         # tags = tagulous.utils.render_tags(form.cleaned_data['tags'])
                         tags = form.cleaned_data['tags']
-                        logger.debug('bulk_edit: setting tags for: %i %s %s', finding.id, finding, tags)
+                        logger.info('bulk_edit: setting tags for: %i %s %s', finding.id, finding, tags)
```
StefanFl, line 1927: I think this should be a `debug` log.

dojo/templates/dojo/view_test.html
```diff
@@ -604,12 +604,15 @@ <h4 class="has-filters"> Findings ({{findings.total_count}}) <small>{{ test.id|g
                             <b>Push to JIRA</b> <input id="id_push_tojira" name="push_to_jira" type="checkbox" alt="Select to push to JIRA" title="Push_To_Jira"/><br/>
 
                         {% endif %}
+                        <label style="display: block">Notes</label>
+                        {{ bulk_edit_form.notes }}
                         <label style="display: block">Tags</label>
                         {% comment %}
                             Quick hack to make bulk edit work without refactoring the whole bulk edit form
+                            {{ bulk_edit_form.media.css }}
```
StefanFl, line 612: Why did you move these 2 lines?

StefanFl (changes requested):

dojo/finding/views.py
dsever: Fully agree, I wanted to do the same but few lines above I've seen `info` so I have followed the pattern

dojo/templates/dojo/view_test.html
dsever: To be honest, I have no idea what those two lines are doing, in `finding_list_snippet.html` this lines are under the comment. So I simply tested with and without this two lines, I can't determinate the difference, so I decide to turn it off but preserve.

PR #5509, opened 2021-11-26
Fix files API

StefanFl:
fixes #5502

The API to upload files to engagements, tests and findings is actually working, but the implementation made it hard to figure out how and generated a not very useful swagger description. With the new implementation the swagger description is consistent with other file uploads. There are 2 changes to the API:

- The return code for a successful ``post`` has been changed to 201 (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201)
- The ``patch`` method didn't work at all and has been deleted.

PR #6098, opened 2022-03-30
Show file selection dialogue in OpenAPI3 Swagger

### LINKED ISSUE #5891: openapi3 spec doesn't provide file upload for import/reimport
https://demo.defectdojo.org/api/v2/oa3/swagger-ui/?docExpansion=none#/import-scan/import_scan_create

This used to be a "Browse" or similar button to select a file to upload.

image

StefanFl:
fixes #5891 

Swagger for OpenAPI3 now looks like this:

2022-03-30 06_07_27-Defect Dojo API v2


valentijnscholten:
hmm I thought we already had set this option, but looks like it never made it to the PR back then: https://github.com/DefectDojo/django-DefectDojo/pull/5509#issuecomment-981133818

mtesauro (approved):
Approved

PR #6100, opened 2022-03-30
Fix Acunetix360 parser error

TheocharisPetros:
Currently Acunetix360 report may include/reference multiple CWEs for a single finding.

Based on the above, while importing such kind of report it would fail since only a single CWE can be referenced on DefectDojo.

The parser has been updated in order to split the CWE list and set the CWE reference the first of the list.


StefanFl:
Hi @TheocharisPetros , thank you for this PR. Please provide a unit test for the list of CWE's as well, so that we have tests for both behaviours.

components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "defectdojo",
-  "version": "2.9.0-dev",
+  "version": "2.9.0",
```
damiencarol, line 3: This should not be in the PR

damiencarol (changes requested):

components/package.json
TheocharisPetros: I will close this PR and create a new one, somehow VS Code rebased the branch with accept both.

damiencarol:
@TheocharisPetros I updated the PR and fixed the conflicts but I see now that we don't have unit tests to test the fix, could you add some?

PR #6101, opened 2022-03-30
feat(core): dedupe algorithm considers endpoints

alles-klar:
Adding a new option to consider endpoints during deduplication. At the moment the feature is disabled by default (empty list of fields) because we will have to fix some of our tests first. I suggest to fix the test and activate the feature in a later PR, so people can start testing the feature. A good default would be DEDUPE_ALGO_ENDPOINT_FIELDS=['host', 'path'].

---
On behalf of DB Systel GmbH


devGregA:
This makes sense to me. I approve. @valentijnscholten  would you please share your thoughts?

devGregA (approved):

valentijnscholten:
I like the idea, but:
- If we don't fix / add tests now, we probably never will. Tests are also a great way to document what the feature does;
- Wondering if the performance is OK in the current state. I think you need to prefetch the endpoints with the existing findings to avoid an extra query for each finding and/or each endpoint;
- Optional? If you want to make it really fast, which we might want because we already get complaints about imports being slow, we should hash the endpoints as well to you can filter on it in the db query that gets existing findings.

alles-klar:
Thanks for the feedback. 

> I like the idea, but:
> 
> * If we don't fix / add tests now, we probably never will. Tests are also a great way to document what the feature does;

Problem is that I don't want to change tests and add new functionality in on PR because then you can't be sure that nothing is broken. But I can create another PR that fixes the current tests. There are some bugs related to endpoints...

> * Wondering if the performance is OK in the current state. I think you need to prefetch the endpoints with the existing findings to avoid an extra query for each finding and/or each endpoint;

Prefetching sound as a good idea.

> * Optional? If you want to make it really fast, which we might want because we already get complaints about imports being slow, we should hash the endpoints as well to you can filter on it in the db query that gets existing findings.

My thoughts were that we are currently not sure how we want to use endpoints/services etc. in the future. Using hashing makes it faster but also very hard to change things because we have to update the hashes every time. I suggest we start with this flexible version and maybe switch to hashes if we are sure we found the best version. If someone wants to have it fast, he can still include the endpoints in the finding hash...

Maffooch (commented):
This is a great first step. Prefetching is a must for sure. I agree with holding out on hashing for the time being.

damiencarol:
100% agree with @valentijnscholten comment on unit tests/integration tests.
This will be important/mandatory for this one.

alles-klar:
> This is a great first step. Prefetching is a must for sure. I agree with holding out on hashing for the time being.

@Maffooch and @valentijnscholten I thought about about prefetching but didn't found a point where it makes sense. Each field is only used once or copied. Can you take a look if I miss something?

alles-klar:
Tests are fixed but the setup is quite complex...

I also decided it makes sense to enable the feature by default because it is the behavior most users would expect. But I am interested in you opinion.

damiencarol:
> Tests are fixed but the setup is quite complex...
> 
> I also decided it makes sense to enable the feature by default because it is the behavior most users would expect. But I am interested in you opinion.

I agree

Maffooch (approved):
This looks great @alles-klar

kiblik:
I really like this idea however I have some questions:
- Can we use `protocol` instead of `scheme`? Just to follow the naming convention in model `Endpoint`. To be honest, I would prefer to change it in the model but it could have a huge impact.
- Is it really necessary to convert Endpoints to string, then decompose by `hyperlink.parse(...)` and then compare? Is not comparing Endpoint fields enough? It looks like 2 useless steps for me which can decrease performance.

devGregA:
@kiblik we spoke and agree with your feedback. Happy to adjust, but just for the sake of time and improvement merging now. Please feel free to submit a PR updating if you'd like.

PR #6102, opened 2022-03-30
HELM: Fix probes when SOCIAL_LOGIN_AUTO_REDIRECT is True

kiblik:
If `SOCIAL_LOGIN_AUTO_REDIRECT=True`, `/login?next=/` is redirected to `SOCIAL_LOGIN` handler. If it fails (for example because IdP is down) also `SOCIAL_LOGIN` mechanism fails and the container is marked as not ready or not live. This needs to be avoided because the health of DD is independent of the health of IdPs. `force_login_form` will forbid redirection.


dsever (approved):

valentijnscholten (approved):
good find

PR #6104, opened 2022-03-31
[FIX] Tags were not listed in the "All Products" page

X0x1RG9f:
Tags are not listed because too much "L"!


valentijnscholten:
looks like minimist update somehow ended up in this PR after rebasing (or not rebasing).
but it's actually a good thing as #6085 was merged into master accidentally.

StefanFl (approved):

Issue #6111, opened 2022-04-01
'view_tool_product' not found when I added a tool or clicked on view tools on product dashboard settings.


pmagnemi20:
### Discussed in https://github.com/DefectDojo/django-DefectDojo/discussions/6026

<div type='discussions-op-text'>

<sup>Originally posted by **worapojc** March 16, 2022</sup>
Hi all,

I've configured **SonarQube** in **Tools Configuration**.

```
API connection successful with message: You have access to 8 projects.
Tool Configuration successfully updated.
```
I tried to add SonarQube integration for the product via _Settings -> Add Tools_ to import from SonarQube project.
However, it returns HTTP 500. Please see the stacktrace that I copied from uWSGI service.

How do I solve this issue?

> django.urls.exceptions.NoReverseMatch: Reverse for 'view_tool_product' not found. 'view_tool_product' is not a valid view function or pattern name.
> ERROR:django.request:Internal Server Error: /product/1/tool_product/all
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
>     response = get_response(request)
>   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
>     response = wrapped_callback(request, *callback_args, **callback_kwargs)
>   File "/app/./dojo/authorization/authorization_decorators.py", line 30, in _wrapped
>     return func(request, *args, **kwargs)
>   File "/app/./dojo/tool_product/views.py", line 50, in all_tool_product
>     return render(request, 'dojo/view_tool_product_all.html', {
>   File "/usr/local/lib/python3.8/site-packages/django/shortcuts.py", line 19, in render
>     content = loader.render_to_string(template_name, context, request, using=using)
>   File "/usr/local/lib/python3.8/site-packages/django/template/loader.py", line 62, in render_to_string
>     return template.render(context, request)
>   File "/usr/local/lib/python3.8/site-packages/django/template/backends/django.py", line 61, in render
>     return self.template.render(context)
>   File "/usr/local/lib/python3.8/site-packages/django/template/base.py", line 170, in render
>     return self._render(context)
>   File "/usr/local/lib/python3.8/site-packages/django/template/base.py", line 162, in _render
>     return self.nodelist.render(context)
>   File "/usr/local/lib/python3.8/site-packages/django/template/base.py", line 938, in render
>     bit = node.render_annotated(context)
[...]

PR #6113, opened 2022-04-03
Fix view tool product error

valentijnscholten:
fixes https://github.com/DefectDojo/django-DefectDojo/issues/6111

`'view_tool_product' not found when I added a tool or clicked on view tools on product dashboard settings.`

### DIFF
--- a/dojo/templates/dojo/view_tool_product_all.html
+++ b/dojo/templates/dojo/view_tool_product_all.html
@@ -43,9 +43,7 @@
                                   </td>
                                   <td>{{ tool.tool_configuration.tool_type.name }}</td>
                                   <td>
-                                      <div class="btn-group"><!--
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Not sure what happened with this view, and if there is a testcase missing, but removing this commented out code resolves the error from https://github.com/DefectDojo/django-DefectDojo/issues/6111 @StefanFl @kiblik

StefanFl (approved):
That is really weird. I removed the `view_tool_product` a while ago, because there was no url attached to it. But why can a commented piece of code make a problem. Anyway, great you found and fixed it.

valentijnscholten:
It's commented in html, the Django template still is trying to render it.

mtesauro (approved):
Approved

Issue #5988, opened 2022-03-03
Automatic assignment of users to corresponding Azure AD groups


stellastarssda:
**Is your feature request related to a problem? Please describe**

When using a corporate Azure Active Directory to authenticate users to Defect Dojo, they are not automatically assigned to the groups that exist in Azure AD.

**Describe the solution you'd like**

As an admin, I want users to be assigned to the groups they belong to. If the group does not yet exist in DefectDojo, it should be generated automatically. 

As a user, after signing in with Azure AD, I want to see all products and product types relevant for the groups I belong to.

**Additional context**

Preliminary work to this topic has been described in https://github.com/DefectDojo/django-DefectDojo/issues/2919
Concept of groups:  https://github.com/DefectDojo/django-DefectDojo/pull/4503

The creation of groups and the assignment of users to groups will be added by myself as a function to https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/pipeline.py#L71 which will be called in https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/settings/settings.dist.py#L437.

kiblik:
This issue can be closed. It was implemented in #6128

PR #6128, opened 2022-04-05
Import azuread groups

wurstbrot:
Add import azureAD Groups as described in https://github.com/DefectDojo/django-DefectDojo/issues/5988.

image

_is_azure_ is not showing up in the OpenAPI description, ideas?

Do I have to add migrations by myself (like in the PR)?

Most code and diagram created with Stella Stars.


wurstbrot:
The error in the unit test is:
`uwsgi_1         | 	text: Login failed
uwsgi_1         | 	</li></ul></li><li>password<ul class="errorlist"><li>Unable to authenticate to JIRA. Please check the URL, username, password, captcha challenge, Network connection. Details in alert on top right. JiraError HTTP 401
uwsgi_1         | 	text: Login failed
uwsgi_1         | 	</li></ul></li></ul>`
Which I assume is not comming due to my changes.

valentijnscholten:
Great addition. I recently used AzureAD groups in Dependency Track via OIDC and I didn't need to query the graph API to get the group names. I could set the `groups` claim to contain the `sAMAccountName` of the groups which is basically the human readable name. Would that work here maybe?

wurstbrot:
> > The error in the unit test is:
> 
> The actual error is lower in the logs: https://github.com/DefectDojo/django-DefectDojo/runs/5835554090?check_suite_focus=true#step:8:2493
> 
> ```
> uwsgi_1         | ======================================================================
> uwsgi_1         | FAIL: test_create (unittests.test_rest_framework.DojoGroupsTest)
> uwsgi_1         | ----------------------------------------------------------------------
> uwsgi_1         | Traceback (most recent call last):
> uwsgi_1         |   File "/app/unittests/test_rest_framework.py", line 83, in wrapper
> uwsgi_1         |     f(self, *args, **kwargs)
> uwsgi_1         |   File "/app/unittests/test_rest_framework.py", line 357, in test_create
> uwsgi_1         |     self.assertEqual(201, response.status_code, response.content[:1000])
> uwsgi_1         | AssertionError: 201 != 500 : b'{"message":"Internal server error, check logs for details"}'
> ```

Thx, fixed

wurstbrot:
Regarding the request from @StefanFl and @valentijnscholten:
I detect groupIDs based on the pattern now and only fetch more information in case it is a groupID. Does this suites your needs?

>To solve the second issue I would prefer a mapping of the AD groups to DD groups. Then users still have control over amount and naming of DD groups. Lots of other tools do it this way.
Dependency Track for example allows such a mapping in the UI. I do aggree that this is a good addition, which should be taggled in an extra issue (due to my limited python knowledge I am not able to do that).

Regarding limitation of group imports: The provided groups in the _access_token_ should be handled by the IDP/AzureAD from my point of view (see for example https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-add-app-roles-in-azure-ad-apps#assign-users-and-groups-to-roles).

wurstbrot:
Hi @StefanFl ,
one of my clients has a naming convention of "team-<teamname> for azureAD groups. A group filter via regex could be applied here to limit the amount of groups in case the IDP is not able to. 
An other client doesn't have such a convention, here the filter would need to be a list expressed as regex.
Would such an ENV-var (e.g. `DD_SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_GROUPS_FILTER`) be suffient for you?
A downside would be the need to restart defectdojo in order to apply the new groups filter via ENV-var.

StefanFl:
I am not completely sure how much a group filter would help me. Nevertheless it would be great if you would implement it. Then I would like to see how this works in a productive environment (by having it in our next release) and then see if we need an additional mapping mechanism.

wurstbrot:
@StefanFl , I added the filter.

I assume the Flake8 error, is like you commented in other PRs, not related to this PR:
```
+ git log -1
current commit
fatal: unsafe repository ('/github/workspace' is owned by someone else)
[...]
````

**Comment by StefanFl** [comment]
Was quite busy last week but will find time to review the PR at the weekend.

File: dojo/pipeline.py [hunk 1]
```diff
@@ -69,6 +75,63 @@ def modify_permissions(backend, uid, user=None, social=None, *args, **kwargs):
...
+                if settings.AZUREAD_TENANT_OAUTH2_GROUPS_FILTER == "" or re.search(settings.AZUREAD_TENANT_OAUTH2_GROUPS_FILTER, group_name):
+                    group_names.append(group_name)
+                else:
+                    logger.debug("Skipping group " + group_name + " due to AZUREAD_TENANT_OAUTH2_GROUPS_FILTER " + settings.AZUREAD_TENANT_OAUTH2_GROUPS_FILTER)
+                    continue
+
+                group_member, is_member_created = Dojo_Group_Member.objects.get_or_create(group=group, user=user, defaults={
+                    'role': Role.objects.get(id=Roles.Writer)})
+
+                if settings.AZUREAD_TENANT_OAUTH2_CLEANUP_GROUPS:
```
StefanFl, line 110: Do we really need this parameter? What sense would it make to leave a user in a AD group in DefectDojo when they are not in this group in Azure anymore?

dojo/pipeline.py
```diff
@@ -69,6 +75,63 @@ def modify_permissions(backend, uid, user=None, social=None, *args, **kwargs):
...
+def is_group_id(group):
+    if re.search(r'^[a-zA-Z0-9]{8,}-[a-zA-Z0-9]{4,}-[a-zA-Z0-9]{4,}-[a-zA-Z0-9]{4,}-[a-zA-Z0-9]{12,}$', group):
+        return True
+    else:
+        return False
+
+
+def cleanup_old_azureAD_groups_for_user(user, group_names):
+    all_groups = Dojo_Group.objects.all()
+    for group in all_groups:
```
StefanFl, line 126: Why do you iterate over all groups? Wouldn't it make more sense just to iterate over the group memberships of that user?

StefanFl (changes requested):

dojo/pipeline.py
```diff
@@ -69,6 +75,63 @@ def modify_permissions(backend, uid, user=None, social=None, *args, **kwargs):
...
+        member = Dojo_Group_Member.objects.filter(group=group, user=user)  # A user shouldn't have multiple roles (see code above) and therefore only one member will be returned
+        logger.debug("Group to delete: " + str(group))
+        if member:
+            if str(group) not in group_names and group.is_azure:
```
valentijnscholten, line 130: slightly better would be to include `group__is_azure==True` to the `Dojo_Group_Member` filter query.

docs/content/en/integrations/social-authentication.md
```diff
@@ -191,6 +191,26 @@ user, such as 'staff' or 'superuser'
 5.  Restart your Dojo, and you should now see a **Login with Azure AD**
     button on the login page which should *magically* work
 
+### Automatic Import of User-Groups
+To import groups from AzureAD users, the following env. var needs to be set:  
+
+    {{< highlight python >}}
+    DD_SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_GET_GROUPS=True
+    {{< /highlight >}}
+This allows to automatically limit the products a user can interact with.
```
valentijnscholten, line 200: I don't think this setting does anything to limit the products a user can interact with. It just adds users to the groups copied from Azure AD.

docs/content/en/integrations/social-authentication.md
```diff
@@ -191,6 +191,26 @@ user, such as 'staff' or 'superuser'
 5.  Restart your Dojo, and you should now see a **Login with Azure AD**
     button on the login page which should *magically* work
 
+### Automatic Import of User-Groups
+To import groups from AzureAD users, the following env. var needs to be set:  
+
+    {{< highlight python >}}
+    DD_SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_GET_GROUPS=True
```
valentijnscholten, line 198: the GET_GROUPS as variable name might be confusing. Does it get the groups from the Azure AD Graph API and store them in DD? Does it get the groups from the token an update the membership for the user in DD? Or Both?
In the code I can see it does both. Might be better to have 2 flags as some people actually might not want to call the graph api because the tokens involved don't allow querying the graph api for groups.

dojo/pipeline.py
```diff
@@ -69,6 +75,63 @@ def modify_permissions(backend, uid, user=None, social=None, *args, **kwargs):
...
+                    continue
+
+                group_member, is_member_created = Dojo_Group_Member.objects.get_or_create(group=group, user=user, defaults={
+                    'role': Role.objects.get(id=Roles.Writer)})
+
+                if settings.AZUREAD_TENANT_OAUTH2_CLEANUP_GROUPS:
+                    cleanup_old_azureAD_groups_for_user(user, group_names)
+            except:
+                logger.error("Could not call microsoft graph API or save groups to member")
+                traceback.print_exc()
```
valentijnscholten, line 114: If there is no permission to query the graph API, this will lead to lots and lots of exceptions/errors in the logs?

docs/content/en/integrations/social-authentication.md
valentijnscholten: Another scenario that I see people using as that they want users to be added to any groups specified in the AAD token, but they do not want new groups to be automatically created. So instead of regex filter, they don't want any new groups to be created. I know this can be achieved by using the regex filter to contain all the groups you have in DD, but that would mean the regex needs to be updated everytime a group is added/removed.

dojo/pipeline.py
valentijnscholten: what if there is a group "SecurityAnalysts" that was created manually in DD, so is_azure is false. Then a AAD users signs in which has that role in its token. The user is added to the group "SecurityAnalysts". Then the user is removed from that group in AAD. Then on the next login the user will not be removed from the group in DD because it is not an azure group. Not sure if this the expected behaviour.

dojo/pipeline.py
```diff
@@ -69,6 +75,63 @@ def modify_permissions(backend, uid, user=None, social=None, *args, **kwargs):
...
+
+                group, created_group = Dojo_Group.objects.get_or_create(name=group_name, is_azure=True)
+                if settings.AZUREAD_TENANT_OAUTH2_GROUPS_FILTER == "" or re.search(settings.AZUREAD_TENANT_OAUTH2_GROUPS_FILTER, group_name):
+                    group_names.append(group_name)
+                else:
+                    logger.debug("Skipping group " + group_name + " due to AZUREAD_TENANT_OAUTH2_GROUPS_FILTER " + settings.AZUREAD_TENANT_OAUTH2_GROUPS_FILTER)
+                    continue
+
+                group_member, is_member_created = Dojo_Group_Member.objects.get_or_create(group=group, user=user, defaults={
+                    'role': Role.objects.get(id=Roles.Writer)})
```
valentijnscholten, line 108: Not sure, but I don't think Writer is a valid role for group membership of users.

dojo/pipeline.py
wurstbrot: Correct. In case there is no permission, you woudn't activate `AZUREAD_TENANT_OAUTH2_ENABLED`.

docs/content/en/integrations/social-authentication.md
wurstbrot: From my point of view, it shouldn't be the target to have groupIDs in DefectDojo. Therefore, Azure AD should provide valid groups (which is not possible in all cases) or the graph should be able to be called. In case group IDs are returned and the graph api can not be called, what would be the value of the IDs in DefectDojo? Therefore, I do not recommend such a filter.

The second senario might be a case for an other issue. The best case would be UI driven with a database like in Dependency Track. I recommend to make one step at a time.

docs/content/en/integrations/social-authentication.md
wurstbrot: Is it better to understand now?

dojo/pipeline.py
valentijnscholten: Is it  a valid usecase to have
a) azure ad oauth enabled
b) return group ids in token
c) do not give permission to graph api to retrieve groups
I think so, but some of the things I mentioned in comments yesterday could be future PRs, same here.

docs/content/en/integrations/social-authentication.md
valentijnscholten: yes, some things could be a future PR. Maybe we should suggest in the docs to use the application roles feature of Azure AD, I believe they result in human readable names in the tokens.

docs/content/en/integrations/social-authentication.md
valentijnscholten: I suggest:
"This will ensure the user is added to all the groups found in the Azure AD Token. Any missing groups will be created in Defect Dojo (unless filtered). This group synchronization allows for product access via groups to limit the products a user can interact with".

valentijnscholten:
Great PR. Left some comments for consideration, some of them are just thinking out loud and could be for a future PR.

dojo/pipeline.py
wurstbrot: I am not sure how to make it better. Shall we document that case under the cleanup flag?

dojo/pipeline.py
```diff
@@ -69,6 +75,63 @@ def modify_permissions(backend, uid, user=None, social=None, *args, **kwargs):
...
+        return True
+    else:
+        return False
+
+
+def cleanup_old_azureAD_groups_for_user(user, group_names):
+    all_groups = Dojo_Group.objects.all()
+    for group in all_groups:
+        member = Dojo_Group_Member.objects.filter(group=group, user=user)  # A user shouldn't have multiple roles (see code above) and therefore only one member will be returned
+        logger.debug("Group to delete: " + str(group))
```
valentijnscholten, line 128:
```suggestion
        logger.debug("Group membership to delete: " + str(group))
```

wurstbrot: That line is not there anymore because I am iterating over the membership groups directly.

dojo/pipeline.py
wurstbrot: When I login with with the used DefectDojo version "v. 2.9.0-dev ( 84afb379 )", I get the "Group role" "Writer" assgined for the groups of the user.
From my point of view, it works.

dojo/pipeline.py
```diff
@@ -69,6 +75,63 @@ def modify_permissions(backend, uid, user=None, social=None, *args, **kwargs):
...
+        group_names = []
+        if 'groups' not in kwargs['response'] or kwargs['response']['groups'] == "":
+            logger.warn("No groups in response. Stopping to update groups of user based on azureAD")
+            return
+        group_IDs = kwargs['response']['groups']
+        for group_from_response in group_IDs:
+            logger.debug("Analysing Group_ID " + group_from_response)
+            request_headers = {'Authorization': 'Bearer ' + token}
+            try:
+                if is_group_id(group_from_response):
```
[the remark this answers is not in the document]
wurstbrot: yes

dojo/pipeline.py
valentijnscholten: According to the docs there's no writer role. Also can't find it in the demo instance. Or am I missing something @StefanFl

StefanFl: Well spotted @valentijnscholten, there is only `Owner`, `Maintainer` or `Reader`.

wurstbrot: I changed it to `Maintainer`.

StefanFl:
@wurstbrot You have to rename the db migration and change the reference in there, because there's a new db migration on the `dev` branch.

wurstbrot:
> @wurstbrot You have to rename the db migration and change the reference in there, because there's a new db migration on the `dev` branch.

Done

StefanFl (approved):
I think the PR is good to go in the next release. Looking forward to see how it will work in my productive environment.

dojo/group/utils.py
```diff
@@ -31,7 +32,7 @@ def group_post_save_handler(sender, **kwargs):
         group.auth_group = auth_group
         group.save()
         user = get_current_user()
-        if user:
+        if user and not(settings.AZUREAD_TENANT_OAUTH2_GET_GROUPS):
             # Add the current user as the owner of the group
```
valentijnscholten, line 36: Is there a reason to make the user owner? Is the idea to make sure there is at least 1 owner? Making a user owner allows them to add other members to the group. Although I am not sure what permissions are needed for a user to be able to access the UI elements for this.

docs/content/en/integrations/social-authentication.md
```diff
@@ -191,6 +191,28 @@ user, such as 'staff' or 'superuser'
...
+To import groups from Azure AD users, the following environment variable needs to be set:  
+
+    {{< highlight python >}}
+    DD_SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_GET_GROUPS=True
+    {{< /highlight >}}
+
+This will ensure the user is added to all the groups found in the Azure AD Token. Any missing groups will be created in DefectDojo (unless filtered). This group synchronization allows for product access via groups to limit the products a user can interact with.
+Do not activate `Emit groups as role claims` within the Azure AD "Token configuration".
+
+To prevent authorization creep, old Azure AD groups a user is not having anymore can be deleted with the following environment parameter:
```
valentijnscholten, line 204: add some comment about that the user will remain in local groups.

dojo/pipeline.py
valentijnscholten: Maybe. Currently there is no way for users/admins to see which groups are "local groups" and which groups are "azure groups". Safest would be to remove the user from all types of groups to avoid permission creep?

valentijnscholten:
Some loose ends, added some comments. Could you also resolve any discussions that have been resolved/completed so it is clear when the PR is "done"?

dojo/pipeline.py
wurstbrot: @valentijnscholten Do you recommend to keep the azuread flag in case we remove the user membership to the group in any case or should we remove it?

wurstbrot: @valentijnscholten ?

valentijnscholten: it might be useful to know where the group came from, but then it should be visible somewhere probably

wurstbrot: I have removed the check for azuread but I keep azuread for future options. Is it ok now?

wurstbrot: ping @valentijnscholten

wurstbrot: @Maffooch could you perform a review?

kiblik:
I had a similar idea with SAML and OpenID in my mind for a longer time.

Do you mind, is it possible to make it a bit more universal - to be able to adopt a similar approach for other SOCIALs?
What do I mean:
change
```python
     is_azure = models.BooleanField(default=False, null=False, help_text="Group imported from Azure AD.", verbose_name="Is Azure AD Group")
```
to
```python
    AZURE = 'azure'
    SOCIAL_CHOICES = (
        (AZURE, _('Azure')),
    )
...
    is_social = models.CharField(max_length=10, choices=SOCIAL_CHOICES, blank=True, null=True)
```
Plus adopt this change in the rest of the code.

kiblik:
Maybe it is a good idea to add the new field also here: https://github.com/DefectDojo/django-DefectDojo/blob/d55fb295e3eae8167af320c545445b2098422737/dojo/api_v2/views.py#L102

wurstbrot:
> > both recommendations are added and tested.
> 
> @wurstbrot I can see no added commit.

sorry, pushed now

Maffooch (approved):
This looks great @wurstbrot

PR #6151, opened 2022-04-12
Endpoint Meta Importer - Import UTF-8-BOM files

Maffooch:
I learned that windows attached something the byte order mark onto files that are generated on windows. The endpoint meta importer did not accept that file encoding.


mtesauro (approved):
Approved

PR #6160, opened 2022-04-14
Improve notifications on Exceptions 

dsever:
In current implementation in case of Exceptions, e.g SonarQube availability, all users will receive information about it, even those who are not entitled for particular product. 

This PR just detects if product is not defined, and then sends notifications only to admins. 
And enhances sonarqubeAPI parser to distribute product as part of Exception (obj).

### DIFF
--- a/dojo/notifications/helper.py
+++ b/dojo/notifications/helper.py
@@ -95,7 +95,6 @@ def create_notification(event=None, **kwargs):
                 logger.debug('Product is not specified, making it silent')
                 users = users.filter(is_superuser=True)
 
-
             for user in users:
                 logger.debug("Authorized user for the product %s", user)
                 # send notifications to user after merging possible multiple notifications records (i.e. personal global + personal product)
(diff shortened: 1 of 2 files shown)


damiencarol (approved):

StefanFl (approved):

dojo/notifications/helper.py
```diff
@@ -73,16 +83,26 @@ def create_notification(event=None, **kwargs):
...
             if product:
                 users = get_authorized_users_for_product_and_product_type(users, product, Permissions.Product_View)
+
             elif product_type:
                 users = get_authorized_users_for_product_type(users, product_type, Permissions.Product_Type_View)
+            else:
+                # nor product_type nor product defined, we should not make noise and send only notifications to admins
+                logger.debug('Product is not specified, making it silent')
+                users = users.filter(is_superuser=True)
+
```
StefanFl, line 97: Here is an actual flake8 error, hidden so far by the git problem

StefanFl (changes requested):

dojo/notifications/helper.py
damiencarol: fixed

PR #6172, opened 2022-04-21
Nessus: Push CVSSv3 into Impact field if it present.

satandyh:
Nothing more just push CVSSv3 into Impact field if it present of course.
Used only for xml parser version


StefanFl:
@satandyh There is an attribute in the findings called `cvssv3`. If this is used, the score will be calculated when saving the finding and stored in the attribute `cvssv3_score`. We should populate this field instead of writing the information in the `impact`.

damiencarol:
> @satandyh There is an attribute in the findings called `cvssv3`. If this is used, the score will be calculated when saving the finding and stored in the attribute `cvssv3_score`. We should populate this field instead of writing the information in the `impact`.

I agree, @satandyh  why using another attribute?

damiencarol (changes requested):
Use the dedicated `cvssv3` attribute

Maffooch:
Closing as #5955 adds the CVSS vector to the dedicated CVSS score field

PR #6194, opened 2022-04-27
Enlarge RAM Request

wurstbrot:
Hi,

on all my defectdojo instances, the uwsgi DefectDojo container operates with ~250MB. Therefore, I recommend to set the request higher to not get into resource conflicts. 

Kind regards
Timo

### DIFF
--- a/helm/defectdojo/values.yaml
+++ b/helm/defectdojo/values.yaml
@@ -189,7 +189,7 @@ django:
     resources:
       requests:
         cpu: 100m
-        memory: 128Mi
+        memory: 256Mi
       limits:
         cpu: 2000m
         memory: 512Mi


wurstbrot:
sorry, I found instances with only 130MB usage, now.

wurstbrot:
> sorry, I found instances with only 130MB usage, now.

... 130MB is for the celery instance ...

damiencarol:
@wurstbrot so you're PR is still needed?

wurstbrot:
From my point of view yes

dsever:
This information is for the scheduler, but in general it is not the problem to increase it. It is ok to me.

dsever (approved):
Approved

damiencarol (approved):

PR #6197, opened 2022-04-27
[HELM] Network policies

dsever:
This brings light network policies option, that blocks k8s defaults to have open communication between all pods within the cluster. When the option is enabled, communication is allowed only within the DD application components within the same namespace, and from an ingressController toward Django pods from the outside of a namespace. 

Default: disabled


Maffooch:
Seems like a good addition for good reasoning. I need to get more familiar with helm so I can actually give a review. For now, I must defer to @alles-klar

alles-klar:
Sorry for the delay, will try to test it next week.

alles-klar (approved):

StefanFl (approved):

PR #6207, opened 2022-04-28
Fix yaml indentation for ingressClassName

achupryn:


dsever:
It has already been fixed in DEV repo

PR #6211, opened 2022-04-29
Read only authorized users in filters and calendar

StefanFl:
Until now some filters and the calendar view listed all DefectDojo users. This might not be desired, e.g. if an instance has external users that should not be visible for everyone.

With this PR a user can only see the users that are authorized for the same products and product types as themselve, including users with global roles and administrators.

### DIFF
--- a/dojo/views.py
+++ b/dojo/views.py
@@ -85,8 +85,8 @@ def action_history(request, cid, oid):
 
     history = LogEntry.objects.filter(content_type=ct,
                                       object_pk=obj.id).order_by('-timestamp')
-    history = LogEntryFilter(request.GET, queryset=history)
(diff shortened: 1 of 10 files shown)


dojo/views.py
```diff
@@ -99,6 +99,7 @@ def action_history(request, cid, oid):
                   {"history": paged_history,
                    'product_tab': product_tab,
                    "filtered": history,
+                   "log_entry_filter": log_entry_filter,
```
valentijnscholten, line 102: is this used somewhere?

StefanFl: The `log_entry_filter` was unused, but I activated it in the history view in `action_history.html`.

devGregA (approved):

valentijnscholten (approved):

PR #6220, opened 2022-04-30
Hash_code calculation with Vulnerability Ids

StefanFl:
With this PR the calculation of hash_codes gets changed from `cve` to `vulnerability_ids`.

### DIFF
--- a/dojo/tools/cargo_audit/parser.py
+++ b/dojo/tools/cargo_audit/parser.py
@@ -39,9 +39,9 @@ class CargoAuditParser(object):
                 date = advisory.get('date')
 
                 if len(advisory.get('aliases')) != 0:
-                    cve = advisory.get('aliases')[0]
+                    vulnerability_reference = advisory.get('aliases')[0]
                 else:
-                    cve = None
+                    vulnerability_reference = None
 
                 package_name = item.get('package').get('name')
                 package_version = item.get('package').get('version')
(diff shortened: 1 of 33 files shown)


dojo/tools/cargo_audit/parser.py
```diff
@@ -39,9 +39,9 @@ def get_findings(self, filename, test):
                 date = advisory.get('date')
 
                 if len(advisory.get('aliases')) != 0:
-                    cve = advisory.get('aliases')[0]
+                    vulnerability_id = advisory.get('aliases')[0]
```
damiencarol, line 42: `aliases` attribute seems to be an array, could we use it as-is?

StefanFl: What do you thing of a larger change: The `id` of the finding in the file is something like `RUSTSEC-2020-0091` which can point to this URL: https://rustsec.org/advisories/RUSTSEC-2020-0091, see https://github.com/DefectDojo/django-DefectDojo/blob/fef0d194ff985358be8e7ebbd4c13a02b9c5a539/unittests/scans/cargo_audit/many_findings.json#L26. 

I think we should use the id and the list of aliases as vulnerability ids.

damiencarol: I agree. now we need to make it real :D

StefanFl: Done

damiencarol (approved):

Maffooch (approved):

PR #6222, opened 2022-05-01
Removes Sponsors Section

devGregA:
After OWASP removed dedicated project funds, we stopped accepting sponsors, but kept the section to make sure no one was shorted on their contributions. In January we met all requirements for display and it now makes sense to remove to avoid any confusion.


Maffooch (approved):

mtesauro (approved):
Approved.

PR #6165, opened 2022-04-18
Enhanced documentation for Docker scripts

StefanFl:
After some users had problems using the scripts for Docker, this PR enhances the documentation with examples.

PR #6227, opened 2022-05-02
dc command improvements

dsever:
In this PR:
- check for docker-compose version
- if not specified use default profile mysql-rabbitmq

### DIFF
--- a/dc-build.sh
+++ b/dc-build.sh
@@ -1,21 +1,7 @@
 #/bin/bash
 
-main=`docker-compose  version  --short | cut -d '.' -f 1`
-minor=`docker-compose  version  --short | cut -d '.' -f 2`
-current=`docker-compose  version  --short`
-
-echo 'Checking docker-compose version'
-if [[ $main -lt 1 ]]; then
-  echo "$current is not supported docker-compose version, please upgrade to minimal supported version:1.28"
-  exit 1
-elif [[ $main -eq 1 ]]; then
-  if [[ $minor -lt 28 ]]; then
-    echo "$current is not supported docker-compose version, please upgrade to minimal supported version:1.28"
-    exit 1
-  fi
-fi
-
-echo 'Supported docker-compose version'
+bash ./docker/docker-compose-check.sh
+if [[ $? -eq 1 ]]; then exit 1; fi
 
 if [ $# -eq 0 ]
 then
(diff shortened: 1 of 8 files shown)


dc-build.sh
```diff
@@ -1,5 +1,22 @@
...
+  exit 1
+elif [[ $main -eq 1 ]]; then
+  if [[ $minor -lt 28 ]]; then
+    echo "$current is not supported docker-compose version, please upgrade to minimal supported version:1.28"
+    exit 1
+  fi
+fi
+
+echo 'Supported docker-compose version'
+
```
valentijnscholten, line 19: Could this some kind of function/script that is included to avoid duplication?

dc-up-d.sh
```diff
@@ -2,12 +2,38 @@
...
+        echo "No profile supplied, running default: mysql-rabbitmq"
+        PROFILE="mysql-rabbitmq"
+        echo "Other supported profiles:
+          mysql-rabbitmq*
+          mysql-redis
+          postgres-rabbitmq
+          postgres-redis
+
+        Usage example: ./dc-up-d.sh mysql-redis
+        "
```
valentijnscholten, line 36: I suggest to also update the docs to have an example like `./dc-up-d.sh mysql-redis` (instead of assuming / relying on DD_PROFILE to be set).

valentijnscholten (changes requested):
nice one, provided some comments.

dc-build.sh
dsever: Yes, I was thinking the same. I will do it

dc-up-d.sh
StefanFl: https://github.com/DefectDojo/django-DefectDojo/pull/6165/files introduced some examples in the docs.

valentijnscholten: ok, didn't notice that. still my suggestion would to make the first command in each line contain the profile name as well and mention the DD_PROFILE thing only after the examples

mtesauro:
+1 on the comments from @valentijnscholten about deduping via a function/script

Otherwise looks good pending resolution of the comments.

dc-build.sh
dsever: Done

dc-up-d.sh
dsever: Added into doc, and did small changes in order.

mtesauro (approved):
Approved.

Nice additions/tweaks - thanks!

Maffooch (approved):

PR #6231, opened 2022-05-02
Performance optimization: Don't read all endpoints or findings in filters

StefanFl:
Some filters read all endpoints or findings. In installations with a lot of endpoints and findings, this can lead to performance problems. This filter attributes have been changed from lists to CharFilters with `icontains` lookups.


devGregA (approved):

mtesauro (approved):
Approved

PR #6233, opened 2022-05-03
Fixed INFO rating for neutral/positive Mozilla Observatory findings

0x4d4e:
Mozilla Observatory findings/tests can have a numeric rating of 0 (neutral finding) or even positive (defense in depth measures). However, the current implementation of the parser rates everything >= -10 as "Low". Neutral or positive Findings should be rated with "Info" instead.


Maffooch (approved):

mtesauro (approved):
Approved

PR #6240, opened 2022-05-03
[FIX] Issue 6230 Fix: Snyk parser cannot handle Nonetype CVSSv3 / cvssScore

### LINKED ISSUE #6230: Snyk Scan parse: cvssScore null value causes exception for operator
**Description** 

cvssScore values of null cause an exception in the Snyk Scan parser, which permits only integer or float. 

**Expected behavior**
The vulnerability should be parsed without its cvssScore component, with field populated appropriately.

**Environment information**
 - Operating System: [e.g. Ubuntu 18.04
 - DefectDojo version [2.9.1](https://hub.docker.com/layers/defectdojo-django/defectdojo/defectdojo-django/2.9.1/images/sha256-7f3d39336319ceb16fa8a7d740104822088aaa9f0d7f52cd0e38a868a159f95b?context=explore)
 - 
**Logs** 

**An exception error occurred during the report import:'<=' not supported between instances of 'NoneType' and 'float' **

**Sample scan files**

P440Boyd:
I'm unable to determine the best alternative (None or empty string) for the fields, but Severity can be pulled from the tool's assessment should a Nonetype be present. 

CVSSv3 vector is a little more complex.

Fixes #6230


dojo/tools/snyk/parser.py
```diff
@@ -118,7 +120,7 @@ def get_item(self, vulnerability, test):
 
         # CVSSv3 vector
         if 'CVSSv3' in vulnerability:
```
damiencarol, line 122: Better to have this version:
```Python
if vulnerability.get("CVSSv3"):
```

damiencarol (changes requested):

damiencarol (commented):
[@P](https://github.com/P440Boyd) there is files that should not be in the PR :/

P440Boyd:
> [@P](https://github.com/P440Boyd) there is files that should not be in the PR :/

Bah, which files? I amended parser.py, test_snyk...py and added some testdata JSON. Which file is added that shouldn't be there? First time here so may be a mistake.

damiencarol:
these modifications shouldn't be here 
image

P440Boyd:
> these modifications shouldn't be here image

Bah - Absolutely agreed. I didn't change them manually, any idea why they'd be there from GitHub? I must have done something not quite right in the fork process.

damiencarol:
come from the merge. just revert this modifications

damiencarol (approved):

P440Boyd:
Thanks for fixing the modifications. Wasn't around yesterday. Noticed unittest was seeking 3 findings after modification, but the test scan I pushed only contains 1. That should resolve unittests.

damiencarol:
Happy to help, it should be fine now.

Maffooch (approved):

PR #6193, opened 2022-04-26
Endpoint_Status: Optimize DB (remove redundancy)

kiblik:
|              |                                    | F1(226) | F2(227)            | F3(228)                                  | F4(229)                                | F5(230)                                | F6 (231) mitigated  |
|--------------|------------------------------------|----|---------------|-------------------------------------|-----------------------------------|------------------------------------|---------------|
| Host foo.bar | Endpoint(4) ftp://foo.bar          | x  | x             | x                                   | x                                 | x                                  | x             |
|              | Endpoint(5) http://foo.bar         | x  | mitigated     | not mitigated<br>but false positive | not mitigated<br>but out of scope | not mitigated<br>but risk accepted | x             |
| Host bar.foo | Endpoint(6) http://bar.foo         | x  | mitigated     | x                                   | x                                 | x                                  | x             |
|              | Endpoint(7) https://bar.foo        | x  | not mitigated | x                                   | x                                 | x                                  | x             |
|              | Endpoint(8) <br>https://bar.foo/f6 | x  | x             | x                                   | x                                 | x                                  | not mitigated |

PR #6242, opened 2022-05-03
Speed-up unittests: skip irrelevant migration tests

kiblik:


Maffooch (approved):

StefanFl:
@kiblik Why wouldn't we delete these tests? Is there any chance they will become relevant again?

kiblik:
@StefanFl, originally I wanted to comment-out these lists. Like here:
https://github.com/DefectDojo/django-DefectDojo/pull/5944/commits/893d75d103f8912d7b03fcd1ac233227913c7493#diff-3e6c30d576b482a6b176c7fa3b7874560e27b470f768c49ab20ec4ad709003b2L93
But @valentijnscholten recommended using `@skip` https://github.com/DefectDojo/django-DefectDojo/pull/5944#issuecomment-1046189313

Yes, delete is also the option.
I do not think they will be used again (because they are testing specific migrations) but sometimes it is good to keep them as inspiration for the creation of new tests. I know, they will be still available in git history but is it easier to be inspired and copy-paste+modify-as-you-need from existing code.

damiencarol:
@kiblik please remove if it's not useful anymore. Not a big fan of dead code

StefanFl:
I agree with @damiencarol, it is better to remove unused code. Sometimes you stumble over these things while refactoring and then it causes unnecessary work.

kiblik:
Ok, in that case, I will totally remove the file because there is no other active scan.
However, it will be recreated soon because I'm using this file in [other PR](https://github.com/DefectDojo/django-DefectDojo/pull/6193/files#diff-3e6c30d576b482a6b176c7fa3b7874560e27b470f768c49ab20ec4ad709003b2)

damiencarol (approved):

PR #6288, opened 2022-05-11
add extraEnv to helm chart

jeffgran:
This adds an value to the helm chart's `values.yaml` file which can be used to inject extra environment variables into the containers. 

It's different from the existing `extraConfigs` value because this is injected into the container definition, not into the configmap. That matters because you can't dynamically pull values from one configmap into another configmap - so if your parameters are stored elsewhere we need this mechanism to add these values to the containers.

An example (very similar to what I'm currently using with this changeset) is in the docs in the values file (below). We have the built-in postgres container disabled but have another template creating our postgres cluster, which then stores its host name in a configmap, and this allows us to inject that host into the containers.

```
# To add (or override) extra variables which need to be pulled from another configMap, you can
# use extraEnv. For example:
# extraEnv:
# - name: DD_DATABASE_HOST
#   valueFrom:
#     configMapKeyRef:
#       name: my-other-postgres-configmap
#       key: cluster_endpoint
```

### DIFF
--- a/helm/defectdojo/templates/celery-beat-deployment.yaml
+++ b/helm/defectdojo/templates/celery-beat-deployment.yaml
@@ -134,6 +134,9 @@ spec:
             secretKeyRef:
               name: {{ $fullName }}
               key: DD_SECRET_KEY
+        {{- if .Values.extraEnv }}
+        {{- toYaml .Values.extraEnv | nindent 8 }}
+        {{- end }}
         resources:
           {{- toYaml .Values.celery.beat.resources | nindent 10 }}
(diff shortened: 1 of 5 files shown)


dsever:
Please do changes against dev, and fix linting problems

jeffgran:
@dsever thanks for taking a look! Sorry, I didn't notice the `dev` branch instruction in the contributing guidelines before. I rebased this branch against dev, and it looks like there are not linting errors now.

dsever (approved):
Looks good, approved

mtesauro (approved):
Approved

PR #6298, opened 2022-05-12
Added a reminder for users about Authentication Configuration

murilo-melo-caramelo:
Added a rebuild reminder for saving the changes in defect dojo authentication, some users may not think about it when realizing changes and it can prevent them from opening future issues about this problem.

### DIFF
--- a/docs/content/en/integrations/social-authentication.md
+++ b/docs/content/en/integrations/social-authentication.md
@@ -5,6 +5,8 @@ draft: false
 weight: 3
 ---
 
+> For every case below, you need to rebuild your DefectDojo to apply any changes in authentication configuration.
+
 ## Auth0
 
 In the same way as with other identity providers, it's now possible to


dsever:
Should not be definitly rebuil, it should be restart. And keep in mind different installation types have different process. 

You should create or against dev

murilo-melo-caramelo:
Tried 4 different methods, only worked after rebuild, did some restarts on my instance and the config hasn't updated

dsever:
Can you first describe your environment? Could be that you are using wrong approach of changing the configuration.

Maffooch:
Agreed with @dsever I am not sure how it works with helm, but with docker-compose, there is an external settings file for the sole purpose of updating configs without needing to rebuild

https://github.com/DefectDojo/django-DefectDojo/tree/master/docker/extra_settings

dsever:
> Agreed with @dsever I am not sure how it works with helm, 

With helm even simpler https://github.com/DefectDojo/django-DefectDojo/blob/master/helm/defectdojo/values.yaml#L13

StefanFl:
It definitely works without rebuilding the docker files. I have set the configuration for Azure AD while deploying the Helm chart and it works as expected.

devGregA:
@murilo-melo-caramelo thank you for the PR, but I don't see this one getting merged.

PR #6301, opened 2022-05-13
Fix get slack id at 1st notification and disclaimer in slack notifications

Antoningggg:
- Fix non working get_slack_user_id function
- Switch on system_settings.disclaimer length as the var cannot be null


Maffooch (approved):

mtesauro (approved):
Approved

PR #6305, opened 2022-05-16
Vulnerability Id: Another batch of parsers

StefanFl:
In the series of PRs to move from CVE to Vulnerability Ids, this PR changes some parsers which were currently writing CVE data.


mtesauro (approved):
Approved

damiencarol (approved):

PR #6307, opened 2022-05-16
Support postgresql ha

### LINKED ISSUE #6096: Support HA PostgreSQL
The DefectDojo Helm chart includes a dependency on the [postgresql](https://github.com/bitnami/charts/tree/master/bitnami/postgresql) Helm chart from Bitnami. This Helm chart does not support high availability (HA), it only supports master/slave replication.

But DefectDojo does not support configuring database writes to the PostgreSQL master and database reads to the PostgreSQL slaves, so the master/slave replication does not increase the global availability of the DefectDojo product.

In case the PostgreSQL master pod fails (e.g. because of a maintenance operation on the underlying node), the database is completely unavailable and DefectDojo is unusable (even with multiple PostgreSQL slaves).

A solution would be to use the [postgresql-ha](https://github.com/bitnami/charts/tree/master/bitnami/postgresql-ha) Helm chart from Bitnami. This Helm chart supports high availability (HA) with database node failover.

bgoareguer:
Fix #6096


Maffooch (commented):
This looks okay to be me, but I am still not very familiar with Kubernetes

dsever:
this looks as good improvement, but still I need to test it

alles-klar:
Hi @bgoareguer,

thanks for your work!

Before taking a deeper look at the content, I want to discuss some general thoughts:
Each database makes the helm chart more complex and harder to maintain. All setups, I know so far, are using an external database from their cloud provider in production (ha) setups. Are you using the postgresql-ha chart in production? Do you maybe know more production installations running the database in kubernetes?

If ha is required, does it make sense to remove the current postgresql chat? Maybe it is better to accept the ha overhead and reduce complexity?

dsever:
> Hi @bgoareguer,
> 
> thanks for your work!
> 
> Before taking a deeper look at the content, I want to discuss some general thoughts: Each database makes the helm chart more complex and harder to maintain. All setups, I know so far, are using an external database from their cloud provider in production (ha) setups. Are you using the postgresql-ha chart in production? Do you maybe know more production installations running the database in kubernetes?
> 
> If ha is required, does it make sense to remove the current postgresql chat? Maybe it is better to accept the ha overhead and reduce complexity?

@alles-klar  we are running everything in k8s, in prod including pgsql. 

I have some mix feeling about it, is in k8s env HA mandatory at all:
1. K8s provides HA on pod level,  (PV should be on some shared storage)
2. Does DD really needs to have that level of availability

But pgpool is nice  to have, I see this as improvement for installations where localstorage is in use.

bgoareguer:
> Before taking a deeper look at the content, I want to discuss some general thoughts: Each database makes the helm chart more complex and harder to maintain. All setups, I know so far, are using an external database from their cloud provider in production (ha) setups. Are you using the postgresql-ha chart in production? Do you maybe know more production installations running the database in kubernetes?
> 
No I am not using postgresql-ha in production yet, I was waiting for it to be supported by the DefectDojo chart.

bgoareguer:
>     1. K8s provides HA on pod level,  (PV should be on some shared storage)
I agree that, if using shared storage, K8S will reschedule the pod by itself but the database will be unavailable until the new pod is ready. With the postgresql-ha chart, there should be no service interruption at all.

dsever:
> > ```
> > 1. K8s provides HA on pod level,  (PV should be on some shared storage)
> > ```
> 
> I agree that, if using shared storage, K8S will reschedule the pod by itself but the database will be unavailable until the new pod is ready. With the postgresql-ha chart, there should be no service interruption at all.

It  is not really the true, it take some tome to convergent from slave to master. But agree it will be quicker than to restart pod because  of DB consistency check.

alles-klar:
If you guy think users need ha, I am fine with it. But adding one testrun in the github actions would be great. I don't think we need to test each combination but one test should include the ha configuration.

bgoareguer:
> If you guy think users need ha, I am fine with it. But adding one testrun in the github actions would be great. I don't think we need to test each combination but one test should include the ha configuration.

I added a testrun with pgsqlha + rabbitmq

alles-klar (approved):

dsever (approved):
Approved

PR #6321, opened 2022-05-20
Add fixes to bugcrowd parser and endpoint uri util

Maffooch:
I found that the bug crowd parser had a couple pitfalls where information that was expected to be in the scan report was not actually there. Additionally, I found a strange endpoint deciphering bug where if given a uri like `subdomain.domain.com:1234` the `Endpoint.from_uri()` function will attempt to create an endpoint with the protocol being `subdomain.domain.com` and the path would be `1234`. Im nervous to attempt to modify the path because it is sort of the catch all for uri attributes. Hopefully what I have implemented will suffice for now.

### DIFF
--- a/dojo/tools/bugcrowd/parser.py
+++ b/dojo/tools/bugcrowd/parser.py
@@ -163,8 +163,8 @@ class BugCrowdParser(object):
 
     def get_endpoint(self, url):
         stripped_url = url.strip()
-        if '://' in stripped_url:  # is the host full uri? 
-            endpoint = Endpoint.from_uri(stripped_url) 
-        else: 
-            endpoint = Endpoint.from_uri('//' + stripped_url) 
+        if '://' in stripped_url:  # is the host full uri?
+            endpoint = Endpoint.from_uri(stripped_url)
+        else:
+            endpoint = Endpoint.from_uri('//' + stripped_url)
         return endpoint
--- a/dojo/tools/bugcrowd/parser.py
+++ b/dojo/tools/bugcrowd/parser.py
@@ -57,9 +57,7 @@ class BugCrowdParser(object):
             finding.impact = pre_description.get('impact', '') + '\n' + row.get('vrt_lineage', '')
             finding.steps_to_reproduce = pre_description.get('steps_to_reproduce', None)
             finding.references = References
-            priority = row.get('priority', 0)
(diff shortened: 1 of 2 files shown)


dojo/tools/bugcrowd/parser.py
```diff
@@ -57,7 +57,9 @@ def get_findings(self, filename, test):
             finding.impact = pre_description.get('impact', '') + '\n' + row.get('vrt_lineage', '')
             finding.steps_to_reproduce = pre_description.get('steps_to_reproduce', None)
             finding.references = References
-            finding.severity = self.convert_severity(int(row.get('priority', 0)))
+            priority = row.get('priority', 0)
+            priority = priority if priority != '' else 0
+            finding.severity = self.convert_severity(int(priority))
```
damiencarol, line 62: Better to have this logic moved in the `convert_severity` method

damiencarol (changes requested):

kiblik:
It is not a bug. It is a future. ;)

The original implementation was trying to follow the standard of URI. Please check https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax
It is the reason, why the function is called `from_uri`.

If your parser returns endpoints like `foo.bar:123`, please add `//` in front, as it is used in:
https://github.com/DefectDojo/django-DefectDojo/blob/d55fb295e3eae8167af320c545445b2098422737/unittests/test_endpoint_model.py#L46
https://github.com/DefectDojo/django-DefectDojo/blob/d55fb295e3eae8167af320c545445b2098422737/dojo/tools/generic/parser.py#L56-L61
https://github.com/DefectDojo/django-DefectDojo/blob/d55fb295e3eae8167af320c545445b2098422737/dojo/tools/nuclei/parser.py#L37-L40
https://github.com/DefectDojo/django-DefectDojo/blob/d55fb295e3eae8167af320c545445b2098422737/dojo/tools/openscap/parser.py#L96-L104
https://github.com/DefectDojo/django-DefectDojo/blob/d55fb295e3eae8167af320c545445b2098422737/dojo/tools/sslscan/parser.py#L68-L74

Maffooch:
Ooh learn something every day! Thanks @kiblik :)

damiencarol (approved):

devGregA (approved):

PR #6324, opened 2022-05-21
Add possibility for arbitrary files to get copied to extra settings

Maffooch:
On one of my dev servers, I am unable to use remote SAML metadata files and instead use XML files. This was restricted to just python files with #5884 as the motive was to remove the possibility of the `README.md` file. I think this approach gives the best of both worlds

### DIFF
--- a/docker/entrypoint-celery-beat.sh
+++ b/docker/entrypoint-celery-beat.sh
@@ -4,7 +4,7 @@ umask 0002
 id
 
 # Allow for bind-mount multiple settings.py overrides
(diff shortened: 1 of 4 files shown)


StefanFl:
@Maffooch I don't understand your use case. Why are the XML files needed within DefectDojo? Are the part of the configuration of your instance?

Maffooch:
The use case is for local SAML metadata files that need to be in the uwsgi container so they can be opened and read by the SAML integration. I could of course just copy the file in by hand, but that does not persist through a restart.

mtesauro (approved):
Approved

StefanFl (approved):

PR #6326, opened 2022-05-21
Fix CSV and Excel export for findings

StefanFl:
The Excel export for findings wasn't working anymore after the introduction of the vulnerability ids. With this PR the Excel export works again and both CSV and Excel export contain the vulnerability ids now.

Additionally the integration tests are waiting 5 seconds now for the generation of the file.


mtesauro (approved):
Approved

damiencarol (approved):

PR #6330, opened 2022-05-24
Always populate unique_id_from_tool

madeoninfo:
Created a unique_id_from_tool that is always populated and identifies a finding with a specific component, even if there is no CVE or XRAY report

### DIFF
--- a/dojo/tools/jfrog_xray_api_summary_artifact/parser.py
+++ b/dojo/tools/jfrog_xray_api_summary_artifact/parser.py
@@ -1,3 +1,4 @@
+import hashlib
 import json
 import re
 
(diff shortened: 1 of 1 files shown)


Maffooch:
@madeoninfo I like this idea as it would ultimately help with duplication for the Jfrog parser.  I am somewhat nervous of overriding something that should be coming from the tool and nowhere else. 

@StefanFl @damiencarol what are your thoughts?

StefanFl:
I don't see why the `unique_id_from_tool` is set in this parser. As @Maffooch said, this field should be set with a unique id that comes from the tool and you can find the finding in the tool using this id. But the parser does set this field already in many cases and this PR just adds one case. 

Currently there are no settings for deduplication for this parser, so the default settings are used, which don't work very reliable. It would be good to think about a setting how to calculate the `hash_code` and to use the `hash_code` for deduplication.

damiencarol:
I agree, the parser should not try to put fake id in `unique_id_from_tool ` field.
@StefanFl the deduplication conf is not set? Maybe that's the reason for this PR, @madeoninfo  ?

Maffooch:
Closing as this is not an optimal solution

PR #6337, opened 2022-05-26
Replace user id with full_name

dsever:
Replaces user id with a full user name inside of the notification  message.

Previously 
```
User name surname (user) has requested that users 43 review the
```
Now
```
User name surname (user) has requested that user(s) user surname (user) review the...
```


StefanFl (approved):

mtesauro (approved):
Approved

PR #6338, opened 2022-05-28
Product based SLA

37b:
This PR includes functionality to set the SLA per Product (if wanted). I am looking for others to test this out and provide feedback.

During migration the current SLA settings are ~~copied~~ moved into a new Default SLA Configuration that is used for all Products that otherwise do not specify a custom SLA.

Additional SLA Configurations can be created and applied to Products individually.

Currently there is __no delete function__ as there should always be at least one config - is there a Django way of preventing deletion of the first record in the table? This is the one that is created during migration using the existing values and is used to set SLA config for all products that do not use a different config.

Will add REST API support before PR is completed.


37b:
@Maffooch I have another Django learner question. This latest PR is failing because the sites are not migrated with the initializer. If I run `python manage.py migrate sites` manually it seems to clear up the 500 error

Note: This is after the app is up and the initializer has run all the migrations

```defectdojo@c197baefe5a2:/app$ python manage.py migrate sites
[30/May/2022 17:38:54] INFO [dojo.models:3864] enabling audit logging
Operations to perform:
  Apply all migrations: sites
Running migrations:
  Applying sites.0001_initial... OK
  Applying sites.0002_alter_domain_unique... OK
```

Maffooch:
@37b I'm going to try rebasing to the dev branch so the commit is a little cleaner. It does not always work so well though. I will change it back if it messy

Maffooch:
This is a pretty neat idea! I'll take a closer look at it when conflicts are fixedx

37b:
Closing while I refactor

PR #6360, opened 2022-06-02
Remove CWE from Dependency Track Hash Code Field

TheocharisPetros:
Dependency Track upgraded the export of results format in order to follow the relevant updated FPF format, so now multiple CWEs are included into to the vulnerabilities export. For information visit their release log https://docs.dependencytrack.org/changelog/ version 4.5.0 . 

In order to not change their API and relevant integrations (such as DefectDojo) they created a work around. So, in case there are multiple CWEs in a single finding they would always use the first one in the list (under the same JSON objects "cweId"and "cweName"), while the total count of CWEs and relevant information will be available in a new array. For more information https://github.com/DependencyTrack/dependency-track/issues/1467#issuecomment-1128161152. 

Close old findings and deduplication features breaks since findings with multiple CWEs may include a different CWE than the one used in order to calculate the original hash (due to the new CWE holding the first place in the array).

Based on the above I would propose to change the "HASHCODE_FIELDS_PER_SCANNER" for "Dependency Track Finding Packaging Format (FPF) Export" in order to not include CWE by default.

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -1077,7 +1077,7 @@ HASHCODE_FIELDS_PER_SCANNER = {
     'SonarQube API Import': ['title', 'file_path', 'line'],
     'Dependency Check Scan': ['vulnerability_ids', 'cwe', 'file_path'],
     'Dockle Scan': ['title', 'description', 'vuln_id_from_tool'],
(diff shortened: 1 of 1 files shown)


StefanFl:
@TheocharisPetros Please rebase your branch against dev, we have fixed a problem with the helm charts

Maffooch (approved):
Seems logical to me. Please see Stefans comment ^ to clear unit tests

damiencarol (approved):

TheocharisPetros:
@StefanFl  done

PR #6362, opened 2022-06-03
Fix for Kubernetes tests

StefanFl:
Bitnami has changed the index.yaml for pre-2022 charts, see https://github.com/bitnami/charts/issues/10539. The Kubernetes tests were not running anymore because of this.

This PR uses the pre-2022 index.yaml and fixes this problem. Nevertheless it shows DefectDojo's Kubernetes installation comes with old charts, that we should update from time to time.


dsever (approved):
Cool

alles-klar (approved):

PR #6394, opened 2022-06-10
Set connection charset

olafz:
In the `docker-compose` file, the MySQL character set is explicitly [set to `utf8mb4`](https://github.com/DefectDojo/django-DefectDojo/blob/master/docker-compose.yml#L142). However, the applications connection string does not include the character set. This PR fixes this.

When importing scan results containing UTF-8 characters, the import used to fail when trying to insert certain characters. After the change in this PR was applied, the problem is gone.

### DIFF
--- a/docker/environments/mysql-rabbitmq.env
+++ b/docker/environments/mysql-rabbitmq.env
@@ -1,4 +1,4 @@
-DD_DATABASE_URL=mysql://defectdojo:defectdojo@mysql:3306/defectdojo
+DD_DATABASE_URL=mysql://defectdojo:defectdojo@mysql:3306/defectdojo?charset=utf8mb4
 DD_DATABASE_ENGINE=django.db.backends.mysql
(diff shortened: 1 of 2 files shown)


olafz:
There is an error in the Helm chart linting, but I'm not sure how (if?) this is related to this PR:

```
Error: Error linting charts: Error processing charts
New chart version: 1.6.33
------------------------------------------------------------------------------------------------------------------------
 ✖︎ defectdojo => (version: "1.6.33", path: "helm/defectdojo") > Chart version not ok. Needs a version bump!
------------------------------------------------------------------------------------------------------------------------
Error linting charts: Error processing charts
```

Maffooch:
@olafz helm linting is failing here because the PR is based against master. Please rebase it to the dev branch

Maffooch (approved):
Good change as we will need to move to MySQL 8 at some point as utf8mb4 is the suggested default

PR #6399, opened 2022-06-13
main(helm): update dependencies and helm

alles-klar:
Update helm dependencies to fix issues with retention policy: https://github.com/bitnami/charts/issues/10539

Update helm to V3.

---
On behalf of DB Systel GmbH


damiencarol (approved):

alles-klar:
ToDos: 
- Testing upgrade process, maybe writing docs for it?
- Remove Helm 2 instructions

StefanFl:
@alles-klar Do we deploy new versions of the tools with this change and if yes, which versions get deployed? I am in particular worried about MySQL, we cannot accidentally upgrade to MySQL 8, which would require a manual database migration, as far as I know.

Maffooch:
@alles-klar I am wondering if this should be pushed to a hot fix branch so we can publish a helm chart under 2.11.1

alles-klar:
> @alles-klar Do we deploy new versions of the tools with this change and if yes, which versions get deployed? I am in particular worried about MySQL, we cannot accidentally upgrade to MySQL 8, which would require a manual database migration, as far as I know.

  - name: mysql
    version: 8.8.12 -> 9.1.7
    appversion: 8.0.27 -> 8.0.29
  - name: postgresql
    version: 10.13.4 -> 11.6.5
    appversion: 11.13.0 -> 14.3.0
  - name: rabbitmq
    version: 8.24.2 -> 10.1.5
    appversion: 3.9.8 -> 3.10.5
  - name: redis
    version: 15.5.4 -> 16.12.0
    appversion: 6.2.6 -> 6.2.7

Mysql was already on version 8, but postgesql is may need a migration?

alles-klar:
> @alles-klar I am wondering if this should be pushed to a hot fix branch so we can publish a helm chart under 2.11.1

We can, but we should do some testing first. There were a lot of changes. I was not able to test yet, if every migration from the old chart to the current version works smooth.

StefanFl:
So MySQL seems to be fine, when we are already on a version 8. I am not sure about PostgreSQL. It seems you need to run `pg_upgrade` (https://www.postgresql.org/docs/current/pgupgrade.html) when you upgrade to another major release.

dsever (changes requested):
Can you take a look at my comments

helm/defectdojo/Chart.yaml
```diff
@@ -8,3 +8,25 @@ maintainers:
   - name: madchap
     email: defectdojo-project@owasp.org
     url: https://github.com/DefectDojo/django-DefectDojo
+dependencies:
```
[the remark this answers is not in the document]
alles-klar: Great idea 👍

alles-klar:
I added some upgrade instructions and tested the upgrade, works in my setup.

I downgraded rabbitmq, so we don't have a major upgrade. Same for postgres but only for the container -> We stay on postgres 11 but use the latest helm chart.

Redis doesn't work at all, but this is another story. It is also disabled in our github workflow for a while:
```
HELM_REDIS_BROKER_SETTINGS: ” \
    --set redis.enabled=false \
```

docs/content/en/getting_started/upgrading.md
```diff
@@ -83,6 +83,24 @@ This requires a one-time rebuild of the Django-Watson search index. Execute the
 
 `./manage.py buildwatson`
 
+**Upgrade instructions for helm chart with postgres enabled**:
```
StefanFl, line 86: Please write, when this has to be executed. After I have done my normal update routine or does this replace my normal update routine. Mine for example is in a GitLab CI pipeline, that runs after it gets a changed helm chart.

StefanFl (changes requested):

docs/content/en/getting_started/upgrading.md
alles-klar: I don't know what your update routine does because the migrations should run during the helm update process but if you have one, that should run after the update. The instructions must be executed before the helm upgrade, otherwise you get an error message caused by the statefullset.

Anyway, I added some more lines. Is it more clear now?

StefanFl: So I do two helm upgrades, first the one you described here and then the helm upgrade I do normally? Sorry, but I am not really familiar with helm and some other users might not be as well?

StefanFl (approved):

Issue #6329, opened 2022-05-23
CSRF verification failed. Request aborted.


isrc-team:
We have installed DefectDojo with the Docker option in Debian 11.

Reverse proxy has been configured to protect the machine with a public ip. 

After accessing with the proxy ip and login we get the following error:

![image](https://user-images.githubusercontent.com/106073386/169848082-83f740e0-f0b7-4b37-8b5e-24009381406c.png)

After reviewing previous problems in this platform, we have realized that we have to add the **CSRF_TRUSTED_ORIGINS** variable, but we do not know in which file to do it, since the installation has been done on top of the docker.

This is the structure of the DefectDojo folder: 
![image](https://user-images.githubusercontent.com/106073386/169855951-bf0c38af-3818-42c7-96aa-1fd253b933cb.png)

Thank you,

dsever:
You can overwrite default settings using localsettings please follow https://defectdojo.github.io/django-DefectDojo/getting_started/configuration/#local_settingspy-not-with-kubernetes

isrc-team:
Thank you!
We have been able to solve the problem, with the documentation provided.

dsever:
Closing

surajanuja:
> Thank you! We have been able to solve the problem, with the documentation provided.

@isrc-team Can you share how did you resolve this issue? Did you made any changes in settings.dist.py or any other settings file you have created and what is the settings that you have added?

mtesauro:
As the linked PR above shows, you need CSRF_TRUSTED_ORIGINS

You can set that in local_settings.py or use an environmental variable like:
`DD_CSRF_TRUSTED_ORIGINS="https://my.domain.com"`

More details at https://docs.djangoproject.com/en/4.0/ref/settings/#csrf-trusted-origins

PR #6403, opened 2022-06-13
Add variable DD_CSRF_TRUSTED_ORIGINS

kiblik:
More people (e.g. https://github.com/DefectDojo/django-DefectDojo/issues/6329) started to need the ability to define `CSRF_TRUSTED_ORIGINS`. It is good to have it as EnvVar.


damiencarol:
@kiblik shouldn't we have a default value?

dsever:
@damiencarol, default values are not required here. But still I have some doubts about the formatting.

dsever:
@kiblik can you inside of settings add link to https://docs.djangoproject.com/en/4.0/ref/settings/#std-setting-CSRF_TRUSTED_ORIGINS because format is different for django 3.2 and 4. Beside that looks good, I tested it manually for K8S

kiblik:
@damiencarol, there is a default value: empty list. It is the same as the original default value for `CSRF_TRUSTED_ORIGINS`:
```python

damiencarol (approved):

PR #6409, opened 2022-06-15
Updated Mozilla Observatory parser to parse findings without 'name' element

0x4d4e:
Older versions of the Mozilla Observatory result file had the element 'name' within each finding. In the current version (seems to have been this way for some time) the 'name' element is no longer included. The updated parser simply uses the 'key' of the finding, which is/was identical to the 'name' used previously by the parser.

Thus, this version is able to parse Mozilla Observatory result files with and without name, by using the finding's key.

I also updated two of the unittest input scanfiles with freshly created scans and adapted the test cases/asserts accordingly.


dojo/tools/mozilla_observatory/parser.py
```diff
@@ -48,7 +48,7 @@ def get_findings(self, file, test):
                 severity=self.get_severity(int(node['score_modifier'])),
                 static_finding=False,
                 dynamic_finding=True,
-                vuln_id_from_tool=node['name']
```
Maffooch, line 51: Was the name attribute removed? Might be better to attempt to use the name attribute if available, and then use the key value as a backup. Increased support there.
`node.get('name', key)` should work here

damiencarol (changes requested):
please create new unit tests instead of replacing the existing ones

dojo/tools/mozilla_observatory/parser.py
0x4d4e: Yes, the name is no longer included in the output of the local scan. https://github.com/mozilla/http-observatory/blob/0b7928ee9d09a3f5dffdb81f27d3e5f80ef2024c/httpobs/scanner/local.py#L76

My observation is based on the output of running Mozilla Observatory locally (using the included script). I'm not certain if other options behave the same. However, in all other examples of files that I have seen (also from the unit tests), the name attribute was always identical to the key.

Would you prefer using the key as 'primary source' in all cases or your option as only using it as fallback?

0x4d4e:
@damiencarol I have added a new scan output file and unittest in addition to the existing tests. Waiting for @Maffooch 's feedback above, otherwise I don't expect any further changes.

dojo/tools/mozilla_observatory/parser.py
Maffooch: Got it. If the name is no longer present, then searching for key as the 'primary source' and then name as the fallback would be fine.

0x4d4e: I changed it to your initial suggestion `node.get('name', key)` - it should be the most fail safe solution with both formats as it does not change the behavior for files with the name attribute. Although key should (!) be identical to the name attribute also in the older format version and thus safe to use, I cannot be certain (since the only 'old' version files I currently have are the unit test files). A quick test with observatory-cli (https://github.com/mozilla/observatory-cli) showed that the json output here still _does_ contain the name attribute, but also identical to the key.

Maffooch (changes requested):
@0x4d4e please remove the docker changes. Should be good to go after that!

0x4d4e:
Argl, I managed to avoid adding those file for so long just to slip up on the last commit :D Should be removed now.

Maffooch (approved):

PR #6413, opened 2022-06-15
Per-Project SLA Config

37b:
This PR includes functionality to set the SLA per Product (if wanted). I am looking for others to test this out and provide feedback.

During migration the current SLA settings are copied moved into a new Default SLA Configuration that is used for all Products that otherwise do not specify a custom SLA.

Additional SLA Configurations can be created and applied to Products individually.


37b:
> @37b The integration tests are failing for adding a product. I imagine selenium is barfing on the extra element because it likely uses xpath to identify elements.

Saw that. Looking at a fix when I get back from holiday.

devGregA:
@37b targeting next release for getting this merged. Thank you for your PR!

37b:
@Maffooch Can I get your feedback on this? I think I covered all the bases, but would like some scrutiny. Thanks.

docker-compose.override.debug.yml
```diff
@@ -33,6 +33,7 @@ services:
     volumes:
       - './dojo/static/dojo:/usr/share/nginx/html/static/dojo'
   mysql:
+    platform: linux/amd64
```
Maffooch, line 36: 😉

dojo/sla_config/views.py
```diff
@@ -0,0 +1,99 @@
...
+from dojo.authorization.authorization import user_has_configuration_permission_or_403
+from dojo.authorization.authorization_decorators import user_is_configuration_authorized
+from dojo.forms import SLAConfigForm
+from dojo.models import SLA_Configuration, System_Settings
+from dojo.utils import add_breadcrumb
+
+logger = logging.getLogger(__name__)
+
+
+@user_is_configuration_authorized('dojo.add_sla_configuration', 'superuser')
```
Maffooch, line 19: This seems a little too restrictive imo. I think allowing owner roles to touch these settings would be a bit better

dojo/forms.py
```diff
@@ -247,6 +247,11 @@ class ProductForm(forms.ModelForm):
                                        queryset=Product_Type.objects.none(),
                                        required=True)
 
+    sla_configuration = forms.ModelChoiceField(label='SLA Configuration',
+                                        queryset=SLA_Configuration.objects.all(),
+                                        required=True,
+                                        initial='Default')
+
```
Maffooch, line 254: This a good place for this, but I fear product maintainers could swap an SLA conf to something a bit more forgiving (thinking the worst of people here)

Possible solution would be to not display this field unless user has owner role (similar to how jira options are toggled based on enablement)

Maffooch (commented):
Apologies for taking so long on this @37b 

Overall, this is terrific! I left a few nitpicky things

dojo/forms.py
37b: How would the value be set during creation? I'm still learning Django's idioms...

docker-compose.override.debug.yml
37b: 😖

dojo/sla_config/views.py
StefanFl: Owner roles exist only for authorization of objects like Products or Findings. This is the correct way for authorization of a configuration. The `superuser` bit is only legacy and will have to be removed after #6446 is merged.

StefanFl (approved):

StefanFl (changes requested):
The configuration authorization cannot be managed for users or groups yet. The fields need to be added in https://github.com/DefectDojo/django-DefectDojo/blob/fbd6cc78e6880ae7312193a31c6a4a0f1eb62420/dojo/user/utils.py#L64.

dojo/db_migrations/0164_custom_sla.py
```diff
@@ -0,0 +1,69 @@
...
+from django.db import migrations, models
+
+logger = logging.getLogger(__name__)
+
+
+def save_existing_sla(apps, schema_editor):
+    SLA_Configuration = apps.get_model('dojo', 'SLA_Configuration')
+    SLA_Configuration.objects.create(name='Default',
+                                     description='The Default SLA Configuration. Products not using an explicit SLA Configuration will use this one.',
+                                     critical=7,
```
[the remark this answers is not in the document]
37b: @StefanFl  This was my initial approach but I couldn't get the data copied in the migration script so I fell back to this "dumb" implementation. Do you have some guidance on how to accomplish this?

dsever:
For me it has more sense to assign, adapted SLA per product type than single product.

37b:
@dsever I thought about that at first but it's less flexible. We have different SLAs for different risk profiles which don't closely align with product type (in our org). 

With this approach the SLA could be named to match the product type so at least to casual users they are aligned for your use case while under the hood they are separate.

JoshBrodieTM:
> For me it has more sense to assign, adapted SLA per product type than single product.

I'm highly anticipating SLA per product. We've taken the suggestion in documentation of having Product Types be Business Units. In each business unit, any given product will have a criticality rating. I'm looking to iterate over our products and set the SLA per product based on the criticality rating. Changing the rating per Product Type would require regrouping products into Product Types by criticality. Definitely agree with the above that SLA per Product allows the same SLA to be set across a Product Type without preventing this use case.

dsever:
> > For me it has more sense to assign, adapted SLA per product type than single product.
> 
> I'm highly anticipating SLA per product. We've taken the suggestion in documentation of having Product Types be Business Units. In each business unit, any given product will have a criticality rating. I'm looking to iterate over our products and set the SLA per product based on the criticality rating. Changing the rating per Product Type would require regrouping products into Product Types by criticality. Definitely agree with the above that SLA per Product allows the same SLA to be set across a Product Type without preventing this use case.

From my point of view non of this is adaptable enough  to solve SLA problem in general, I was thinking something like this https://owasp.slack.com/archives/C014H3ZV9U6/p1655809862156139 we have already created backlog for this in the company and would like to have it as PoC, to have kind of advance SLA logic.

37b:
> > > For me it has more sense to assign, adapted SLA per product type than single product.
> > 
> > 
> > I'm highly anticipating SLA per product. We've taken the suggestion in documentation of having Product Types be Business Units. In each business unit, any given product will have a criticality rating. I'm looking to iterate over our products and set the SLA per product based on the criticality rating. Changing the rating per Product Type would require regrouping products into Product Types by criticality. Definitely agree with the above that SLA per Product allows the same SLA to be set across a Product Type without preventing this use case.
> 
> From my point of view non of this is adaptable enough to solve SLA problem in general, I was thinking something like this https://owasp.slack.com/archives/C014H3ZV9U6/p1655809862156139 we have already created backlog for this in the company and would like to have it as PoC, to have kind of advance SLA logic.

@dsever Are you saying you want the SLA to change on the fly based on other criteria set on the product? That kind of approach can be done via scripts/API using the basic per-product SLA implementation. It could possibly also be added as a feature flag for those who want more complex logic built into the tool. The idea was to keep it simple and see what works/doesn't work so it can be iterated upon.

dsever:
> > > > For me it has more sense to assign, adapted SLA per product type than single product.
> > > 
> > > 
> > > I'm highly anticipating SLA per product. We've taken the suggestion in documentation of having Product Types be Business Units. In each business unit, any given product will have a criticality rating. I'm looking to iterate over our products and set the SLA per product based on the criticality rating. Changing the rating per Product Type would require regrouping products into Product Types by criticality. Definitely agree with the above that SLA per Product allows the same SLA to be set across a Product Type without preventing this use case.
> > 
> > 
> > From my point of view non of this is adaptable enough to solve SLA problem in general, I was thinking something like this https://owasp.slack.com/archives/C014H3ZV9U6/p1655809862156139 we have already created backlog for this in the company and would like to have it as PoC, to have kind of advance SLA logic.
> 
> @dsever Are you saying you want the SLA to change on the fly based on other criteria set on the product? That kind of approach can be done via scripts/API using the basic per-product SLA implementation.

Then brain\logic is outside of the platform, script needs to pick product/product type metadata and make an decision what to apply, this is not so convenient. I'm just describing our use case, and why we want to enable more programmable advance SLA definitions.

dojo/db_migrations/0164_custom_sla.py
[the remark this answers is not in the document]
37b: @StefanFl I was able to get it done using two migrations - I don't know why it did not work with just one.

Maffooch (approved):
I think this is good from a first iteration pov. The logic of changing SLA conf based on criticality and other vars within the product will likely mean something slightly different to everybody. Maybe it would be good to keep it a manual change for now, and then based on community use and standards, we could set some sane defaults.

dojo/sla_config/views.py
```diff
@@ -0,0 +1,97 @@
...
+        request=request)
+
+    return render(request,
+                  'dojo/edit_sla_config.html',
+                  {
+                      'form': form,
+                  })
+
+
+@user_is_configuration_authorized('dojo.add_sla_configuration')
```
StefanFl, line 83: Here we have the wrong permission.

```suggestion
@user_is_configuration_authorized('dojo.view_sla_configuration')
```

StefanFl (changes requested):

StefanFl:
IMO this PR makes DefectDojo better. I would like to see the SLA configuration per Product Type, but this should be easy to add in another PR. Having it even more configurable like @dsever proposed would be the next step.

StefanFl (approved):

StefanFl:
@Maffooch @dsever Any objections about merging this PR?

Maffooch:
No objections from me! Will merge this after release.

PR #6415, opened 2022-06-16
Update importing.md

derweiser:
On the Integrations pages, both Importing and Supported Reports has the same subtitle "DefectDojo has the ability to import scan reports from a large number of security tools."
I've made a suggestion to change this to make the Importing link more specific to the contents.


mtesauro (approved):
Approved

Maffooch (approved):

Issue #6422, opened 2022-06-17
DD_SITE_URL not set by helm chart


Gby56:
**Be informative**
There was an issue when I tried to push issues to Jira, the link stayed localhost, the helm chart was broken.

**Bug description**
Simply missing an env setting

**Steps to reproduce**

**Expected behavior**

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
DDojo version 2.10.0

**Logs** 
Use `docker-compose logs` (or similar, depending on your deployment method) to get the logs and add the relevant sections here showing the error occurring (if applicable).

**Sample scan files**
If applicable, add sample scan files to help reproduce your problem.

**Screenshots**

**Additional context** (optional)

dsever:
Merged into dev

PR #6423, opened 2022-06-17
Fix DD_SITE_URL in helm chart #6422

Gby56:
https://github.com/DefectDojo/django-DefectDojo/issues/6422

This is simply letting the user set a DD_SITE_URL in the uwsgi of django, and that will get copied over to the celery worker, I tested this with my currently deployed helm chart and it worked, Jiras finally show the link
I have set the default value explicitly, so it's visible directly and not in a yaml comment

### DIFF
--- a/helm/defectdojo/values.yaml
+++ b/helm/defectdojo/values.yaml
@@ -197,8 +197,6 @@ django:
   replicas: 1
   tolerations: []
   uwsgi:
-    env:
-      DD_SITE_URL: 'http://localhost:8080'
     livenessProbe:
       # Enable liveness checks on uwsgi container. Those values are use on nginx readiness checks as well.
       enabled: true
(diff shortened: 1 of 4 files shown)


dsever:
I will take a look, but first of all create PR against of DEV not MASTER branch.
I can agree that DD_SITE_URL could be one of the main configuration parameters, but before it will be merged into dev->master you can reference the same using: extraConfig helm section.

helm/defectdojo/values.yaml
```diff
@@ -197,6 +197,8 @@ django:
   replicas: 1
   tolerations: []
   uwsgi:
+    env:
```
dsever, line 200: As env is in use for other components, not only for  the  django uwsgi itself, this is not natural position for given config

Gby56: Done ! it seems odd to have it all the way down in the values.yaml but I understand yes, I added a comment on the default value to explain why you likely would change the value

dsever:
So, just to explain, extraConfig is designed in the way, everything that is defined there will be created as env variable in each django based container. So if you define something like:

`extraConfig:
   DD_SOME_VAR: something
`
helm char will automatically create given env variable with assigned value. 

So it hasn't be in helm explicitly defined, you don't need to to touch helm tempates.

But if you want to specify DD_SITE_URL in extraConfig, just do it, and you don't need to change any template.

https://github.com/DefectDojo/django-DefectDojo/blob/master/helm/defectdojo/templates/configmap.yaml#L46

Gby56:
Oh sorry right ! that should have been obvious, I forgot how extraConfigs worked, I'll remove the templates changes then, and just add an explicit DD_SITE_URL in the extraConfigs I guess ?

Gby56:
I have added a small piece in the kubernetes readme doc then

damiencarol (approved):

Maffooch (approved):

PR #6426, opened 2022-06-18
Enable/disable calendar

StefanFl:
A user wanted an option to disable the calendar due to privacy concerns. Because I think not a lot of users will need the Calendar, I have added this option to the system settings.


damiencarol (approved):

damiencarol:
LGTM, @StefanFl could you fix the lint error in the values file?

Maffooch (approved):

PR #6431, opened 2022-06-20
Display tags in several lists again

StefanFl:
A PR end of last year introduced a typo that let tags not to be displayed anymore in several list.


Maffooch (approved):

Maffooch:
I had a feeling this was something only I could accomplish 😭 #5491

mtesauro (approved):
Approved.

@Maffooch Turn down the repeat rate on your keyboard ;-)

PR #6432, opened 2022-06-20
Fix part 2: Add DD_SITE_URL to Helm chart 

Gby56:
This new PR is to move the option out of extraConfigs, so that it is not polluted.

### DIFF
--- a/helm/defectdojo/values.yaml
+++ b/helm/defectdojo/values.yaml
@@ -47,7 +47,7 @@ database: postgresql
 host: defectdojo.default.minikube.local
 
 # The full URL to your defectdojo instance, depends on the domain where DD is deployed, it also affects links in Jira
-site_url: 'http://localhost:8080'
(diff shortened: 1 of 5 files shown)


helm/defectdojo/values.yaml
```diff
@@ -46,6 +46,9 @@ database: postgresql
 # Primary hostname of instance
 host: defectdojo.default.minikube.local
 
+# The full URL to your defectdojo instance, depends on the domain where DD is deployed, it also affects links in Jira
+site_url: 'http://localhost:8080'
```
dsever, line 50: Would be good to comment out this. You already have defaults in the configMaps.

dsever (changes requested):
comment out site_url inside of values.

helm/defectdojo/values.yaml
Gby56: Done !

PR #6446, opened 2022-06-22
Remove legacy authorization for changing configuration

StefanFl:
The era of the staff user is finally over. Legacy authorization for configuration changes and settings to activate it have been removed.


Maffooch (approved):
Wow this is one big boy

StefanFl:
Yes, unfortunately it's a bit lengthy. But all these things were interconnected and at least it's not complicated.

mtesauro (approved):
Appoved

PR #6448, opened 2022-06-22
Create a notification if the template does not exist

coheigea:
If the template does not exist in the configured template directory then no alert is raised when you push to JIRA.

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -5,6 +5,7 @@ import io
 import json
 import requests
 from django.conf import settings
+from django.template import TemplateDoesNotExist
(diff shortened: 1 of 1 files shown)


Maffooch:
Are these the templates you are referring to?

https://github.com/DefectDojo/django-DefectDojo/tree/master/dojo/templates/issue-trackers

coheigea:
Yes @Maffooch  - the error came about because we created a folder but didn't realise the name of the file is hard-coded in DefectDojo. We saw an error in the logs but no notification...

StefanFl (approved):

Maffooch (approved):

Issue #6434, opened 2022-06-20
Notes are not added to new JIRAs


coheigea:
When pushing a new finding to JIRA, where the product has "Push notes" selected for the JIRA configuration, existing notes will not be pushed to the JIRA. Only notes that are created after the JIRA is created are pushed to JIRA.

Steps to reproduce:
 1. Enable "Push notes" in the JIRA configuration for a product.
 2. For a given finding add a note and then push to JIRA. Observe that no comment exists in JIRA for the note.
 3. Add a note to the finding and refresh the JIRA, and the second note is added as a comment, but the first note is missing.

It looks like for add_jira_issue in https://github.com/DefectDojo/django-DefectDojo/blob/c101e47b294863877cd68a82d0cc60f8017b45b1/dojo/jira_link/helper.py#L599 it never checks for existing notes to push as comments, like in the add_comment function. 

If you agree this makes sense to add I'll look at a PR for it.

PR #6449, opened 2022-06-22
Add all existing finding notes to the JIRA when created

coheigea:
Tested locally correctly.

Fixes: https://github.com/DefectDojo/django-DefectDojo/issues/6434

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -719,6 +719,13 @@ def add_jira_issue(obj, *args, **kwargs):
         issue = jira.issue(new_issue.id)
 
         logger.info('Created the following jira issue for %d:%s', obj.id, to_str_typed(obj))
+
+        # Add any notes that already exist in the finding to the JIRA
+        for find in findings:
+            if find.notes.all():
+                for note in find.notes.all().reverse():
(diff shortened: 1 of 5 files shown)


Maffooch (changes requested):
Great change! We need some unit tests here though. 

This would be a good place to add them. Be sure to read the block comment up top for use of  the VCR recording (mimics the response from the jira server)

https://github.com/DefectDojo/django-DefectDojo/blob/c101e47b294863877cd68a82d0cc60f8017b45b1/unittests/test_jira_import_and_pushing_api.py#L14-L32

coheigea:
@Maffooch It looks like modifying test_import_with_push_to_jira_add_comment in that file to set push_to_jira=false on the initial import, then add the comment and push to JIRA, should work. However, I don't think I can add a test without:

# if you need some credentials for the Defect Dojo JIRA Cloud instance, contact one of the moderators

coheigea:
@Maffooch @damiencarol How can we move forward with getting this merged?

Maffooch:
@coheigea This totally slipped through the cracks. Sorry about that! I will add a test for this today.

damiencarol (approved):

PR #6452, opened 2022-06-22
Fix: in reimporter, do not reactivate mitigated findings if they have the same date

Gby56:
This is clearly a bug, reimporting the same report over and over would reactivate the previously closed findings. (notice that the code before my change reactivates findings by default, even if the reimport adds a mitigated finding, it opens the existing finding and makes it active for no reason)
This can be improved by doing other date comparisons, so that triage done directly in the scanners can be reflected during reimport.

### DIFF
--- a/dojo/importers/reimporter/reimporter.py
+++ b/dojo/importers/reimporter/reimporter.py
@@ -140,7 +140,7 @@ class DojoDefaultReImporter(object):
                             logger.debug('%i: closing: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
                             finding.mitigated = item.mitigated
                             finding.is_mitigated = True
-                            finding.mitigated_by = None
+                            finding.mitigated_by = item.mitigated_by
                             finding.active = False
                             finding.verified = verified
                     if not finding.component_name or not finding.component_version:
--- a/dojo/importers/reimporter/reimporter.py
+++ b/dojo/importers/reimporter/reimporter.py
@@ -88,10 +88,12 @@ class DojoDefaultReImporter(object):
                         logger.debug("finding mitigated time: " + str(finding.mitigated.timestamp()))
                         if item.mitigated.timestamp() == finding.mitigated.timestamp():
                             logger.debug("New imported finding and already existing finding have the same mitigation date, will skip as they are the same.")
+                            unchanged_items.append(finding)
+                            unchanged_count += 1
                             continue
                         if item.mitigated.timestamp() != finding.mitigated.timestamp():
-                            logger.debug("New imported finding and already existing finding have different dates")
-
+                            logger.debug("New imported finding and already existing finding are both mitigated but have different dates")
+                            # TODO: implement proper date-aware reimporting mechanism
                     if not item.mitigated:
                         logger.debug('%i: reactivating: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
                         finding.mitigated = None
--- a/dojo/importers/reimporter/reimporter.py
+++ b/dojo/importers/reimporter/reimporter.py
@@ -82,7 +82,7 @@ class DojoDefaultReImporter(object):
                 finding = findings[0]
                 if finding.false_p or finding.out_of_scope or finding.risk_accepted:
                     logger.debug('%i: skipping existing finding (it is marked as false positive:%s and/or out of scope:%s or is a risk accepted:%s): %i:%s:%s:%s', i, finding.false_p, finding.out_of_scope, finding.risk_accepted, finding.id, finding, finding.component_name, finding.component_version)
-                elif finding.mitigated and finding.is_mitigated:
+                elif finding.is_mitigated:
                     if item.mitigated:
                         logger.debug("item mitigated time: " + str(item.mitigated.timestamp()))
                         logger.debug("finding mitigated time: " + str(finding.mitigated.timestamp()))
--- a/dojo/importers/reimporter/reimporter.py
+++ b/dojo/importers/reimporter/reimporter.py
@@ -40,7 +40,7 @@ class DojoDefaultReImporter(object):
         unchanged_count = 0
         unchanged_items = []
 
-        logger.debug('starting reimport of %i items.', len(items) if items else 0)
+        deduplicationLogger.debug('starting reimport of %i items.', len(items) if items else 0)
         from dojo.importers.reimporter.utils import (
             match_new_finding_to_existing_finding,
             update_endpoint_status,
@@ -83,7 +83,13 @@ class DojoDefaultReImporter(object):
                 if finding.false_p or finding.out_of_scope or finding.risk_accepted:
                     logger.debug('%i: skipping existing finding (it is marked as false positive:%s and/or out of scope:%s or is a risk accepted:%s): %i:%s:%s:%s', i, finding.false_p, finding.out_of_scope, finding.risk_accepted, finding.id, finding, finding.component_name, finding.component_version)
                 elif finding.mitigated or finding.is_mitigated:
-                    logger.debug('%i: reactivating: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
+                    deduplicationLogger.debug("item mitigated time: "+str(item.mitigated.timestamp()))
+                    deduplicationLogger.debug("finding mitigated time: "+str(finding.mitigated.timestamp()))
+                    if item.mitigated:
+                        if item.mitigated.timestamp() == finding.mitigated.timestamp():
+                            deduplicationLogger.debug("New imported finding and already existing finding have the same mitigation date, will skip as they are the same.")
+                            continue
+                    deduplicationLogger.debug('%i: reactivating: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
                     finding.mitigated = None
                     finding.is_mitigated = False
                     finding.mitigated_by = None
@@ -111,7 +117,7 @@ class DojoDefaultReImporter(object):
                         # If there is only one chunk, then do not bother with async
                         if len(chunk_list) < 2:
                             reactivate_endpoint_status(endpoint_statuses, sync=True)
-                        logger.debug('IMPORT_SCAN: Split endpoints into ' + str(len(chunk_list)) + ' chunks of ' + str(chunk_list[0]))
+                        deduplicationLogger.debug('IMPORT_SCAN: Split endpoints into ' + str(len(chunk_list)) + ' chunks of ' + str(chunk_list[0]))
                         # First kick off all the workers
                         for endpoint_status_list in chunk_list:
                             reactivate_endpoint_status(endpoint_status_list, sync=False)
@@ -123,7 +129,7 @@ class DojoDefaultReImporter(object):
                     reactivated_count += 1
                 else:
                     # existing findings may be from before we had component_name/version fields
-                    logger.debug('%i: updating existing finding: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
+                    deduplicationLogger.debug('%i: updating existing finding: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
                     if not finding.component_name or not finding.component_version:
                         finding.component_name = finding.component_name if finding.component_name else component_name
                         finding.component_version = finding.component_version if finding.component_version else component_version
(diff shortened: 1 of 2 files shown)


valentijnscholten (changes requested):
Deduplicationlogger is for deduplication, not for reimport which sole purpose is almost to avoid duplicates.

Gby56:
Ah sorry, the normal logger.debug didn't show up as I was testing things

StefanFl:
Hi @Gby56, I don't understand which problem do you want to solve with this PR. Please elaborate a bit more when this problem happens and what the outcome is.

Gby56:
Sure !

I'm using Veracode XML reports, the strategy we use is to have 1 engagement and 1 test and keep reimporting every day, to prevent creating duplicates.

I noticed simply that importing twice the exact same report would yield a completely different status, it opened ALL the issues on the second upload, which makes zero sense to me, as the first upload would properly create inactive issues that are marked mitigated in the XML report.

This is because the skipping of issues is only done if they are false positive or out of scope or risk accepted. But the way Veracode does it, and the parser, is not using these 3 statuses for a normal "not a concern" triage.
"Mitigated by design" in Veracode does not truly map to a risk acceptance or a false positive. It is an actual finding but something else mitigated that, false positive would truly be a faulty scanner imo.

That's why my fix is simply looking at "mitigated" but not false positive or other statuses, and comparing the mitigated dates, to see if maybe the mitigation applied by triaging in the scanner is newer or just different than the DefectDojo triage. I don't see a lot of date comparison in DefectDojo and I think it should be an appropriate strategy, if the scanner triaged something newer than DefectDojo, let it take over, it's the source of truth after all.

StefanFl:
Thanks. I will have a look at it but not sure it will be today.

StefanFl (dismissed):
The reimport logic is mind-boggling but I think this change make sense. @damiencarol please have look too.

Gby56:
Yes I think the tests might need readjusting, also I thought about improvements for the logic I added:
For now it's only leaving findings intact on exact same mitigated date, but maybe the findings should get updated if the mitigated date is newer than the one stored in defect dojo too ?

StefanFl:
> Yes I think the tests might need readjusting

Definitely, the PR cannot be merged with failing tests.

> also I thought about improvements for the logic I added: For now it's only leaving findings intact on exact same mitigated date, but maybe the findings should get updated if the mitigated date is newer than the one stored in defect dojo too ?

That makes sense to me

Gby56:
Alright, tests failing are currently:

```
FAIL: test_import_0_reimport_1_active_verified_reimport_0_active_verified (unittests.test_import_reimport.ImportReimportTestUI)
```

```
FAIL: test_import_0_reimport_1_active_verified_reimport_0_active_verified_statistics_no_history (unittests.test_import_reimport.ImportReimportTestAPI)
```

```
FAIL: test_import_0_reimport_1_active_verified_reimport_0_active_verified_statistics (unittests.test_import_reimport.ImportReimportTestAPI)
```

```
FAIL: test_import_0_reimport_1_active_verified_reimport_0_active_verified (unittests.test_import_reimport.ImportReimportTestAPI)
```

Gby56:
Note: trying to run a single unit test via docker-compose is not easy to find, I'm struggling with running all the tests every time and it's super long, so I'm trying to narrow it down. So far I think it's always the `assertTestImportModelsCreated` that is wrong, because less issues are created or closed or touched, I have to re-evaluate and guess the right numbers...

Gby56:
The issue was just in the logging statement, the object was None if before the if block :)
Should be all good @damiencarol @StefanFl @valentijnscholten , unless you'd like for me to try and implement the date detection that goes something like:
- get finding mitigated date in new report
- get finding mitigated date in defectdojo
- if new report newer than defectdojo, take new information and update the existing
  - potentially the new mitigation has updated comments/info ?
- if older, do not take action, `continue`

damiencarol:
reviewing it...

dojo/importers/reimporter/reimporter.py
```diff
@@ -83,9 +83,9 @@ def process_parsed_findings(self, test, parsed_findings, scan_type, user, active
                 if finding.false_p or finding.out_of_scope or finding.risk_accepted:
                     logger.debug('%i: skipping existing finding (it is marked as false positive:%s and/or out of scope:%s or is a risk accepted:%s): %i:%s:%s:%s', i, finding.false_p, finding.out_of_scope, finding.risk_accepted, finding.id, finding, finding.component_name, finding.component_version)
                 elif finding.mitigated or finding.is_mitigated:
-                    logger.debug("item mitigated time: " + str(item.mitigated.timestamp()))
-                    logger.debug("finding mitigated time: " + str(finding.mitigated.timestamp()))
                     if item.mitigated:
+                        logger.debug("item mitigated time: " + str(item.mitigated.timestamp()))
+                        logger.debug("finding mitigated time: " + str(finding.mitigated.timestamp()))
```
StefanFl, line 88: Wouldn't we have the same problem here, if `item` is mitigated but `finding` isn't? Maybe line 85 needs an `and` instead of an `or`.

StefanFl (changes requested):

dojo/importers/reimporter/reimporter.py
Gby56: I agree, the original code was with the `or` condition but I didn't really understand why, honestly the whole block that reactivates the finding should be gone, and have much more conditions imo
```python
                    logger.debug('%i: reactivating: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
                    finding.mitigated = None
                    finding.is_mitigated = False
                    finding.mitigated_by = None
                    finding.active = True
                    finding.verified = verified
```

Gby56:
I started testing around with the Veracode parser, it's more complex because it depends on each parser's logic... veracode saves the whole "history" of mitigations in the XML, and the parser only picks the last one in the list (which is the first one, so the oldest date...) this meant that the logic was flawed, it would not close a finding that is opened for some reason.

Gby56:
@damiencarol I'd really like to get that fix in the next release btw, does defectdojo have a schedule for releases ?

dojo/importers/reimporter/reimporter.py
```diff
@@ -124,6 +134,15 @@ def process_parsed_findings(self, test, parsed_findings, scan_type, user, active
                 else:
                     # existing findings may be from before we had component_name/version fields
                     logger.debug('%i: updating existing finding: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
+                    if not (finding.mitigated and finding.is_mitigated):
+                        logger.debug('Item matches a finding that is currently open.')
+                        if item.mitigated:
+                            logger.debug('%i: closing: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
+                            finding.mitigated = item.mitigated
+                            finding.is_mitigated = True
+                            finding.mitigated_by = None
```
StefanFl, line 143: Why do you set `finding.mitigated_by` to `None`? Maybe the attribute is set in `item`.

StefanFl (changes requested):

StefanFl (approved):

damiencarol:
> @damiencarol I'd really like to get that fix in the next release btw, does defectdojo have a schedule for releases ?

We have a monthly release. It could be too late for this one.
I reviewed this PR and I don't think it helps. Not want to be harsh but I don't think checking the date of the mitigation will fix completely the pb.
Something is missing. If we do re-upload multiple time of the same report we should not see any changes. even if the dates of the mitigation are differents

Gby56:
> > @damiencarol I'd really like to get that fix in the next release btw, does defectdojo have a schedule for releases ?
> 
> We have a monthly release. It could be too late for this one. I reviewed this PR and I don't think it helps. Not want to be harsh but I don't think checking the date of the mitigation will fix completely the pb. Something is missing. If we do re-upload multiple time of the same report we should not see any changes. even if the dates of the mitigation are differents

If you try with a Veracode report or other parsed formats, you'll see that reimporting the same exact report twice will FLIP the status of issues, just because they are not false positive, out of scope, or risk accepted, they are "just" mitigated.

This was half-fixed in https://github.com/DefectDojo/django-DefectDojo/pull/3858 where only these few issue statuses were ignored and left "as-is", but still, if a reimported issue is "just mitigated" in the uploaded report, and the current one in defectdojo is mitigated too, it will reopen the defectdojo issue. This makes zero sense to me 😓

Gby56:
On first import:
first_import

After reimporting the SAME exact file:
after_reimport_same_scan

Maffooch:
This seems okay to me. Agreed with Damien that it would be a good idea to wait until after the release to merge this.

coheigea:
LGTM can we get it merged please? It solves two problems:

- Re-importing a mitigated finding will re-open the old mitigated finding, because it doesn't take into account that the re-imported finding is mitigated.
- If we mitigate a finding in DefectDojo, then this is ignored when re-importing the finding.

Maffooch (approved):
@damiencarol @StefanFl I am good with this one to be merged, but will leave to you for final action

Gby56:
We're trying to figure out how to make a test to import the same report twice with @coheigea, since that should always give 0 changes.
The definition of "same report" really means zero changes to mitigations, mitigation dates in the report of course. 

I'd also like to make an overall test where:

1. You import a scan report with open and mitigated findings
2. You mitigate a finding in DD
3. You reimport the same scan, nothing changes
4. You close a finding in the report that is still open in DD, and check that it has been closed after reimporting
5. You close a finding in the report that was already closed by DD, effectively you have a newer mitigation date from the report and maybe you add a comment from the report (if triage in scanners allows for comments)

This should cover very common workflows where you reimport over and over the scans, they might get triaged only in DD or in the scanner or both

Gby56:
@damiencarol added a unit test too, which failed in https://github.com/DefectDojo/django-DefectDojo/pull/6566 

I haven't really checked if "untouched" had to be explicitly set in the reimporter, when the same report is reimported, it won't report that all the findings have been left untouched, it's blank on the screenshot, is that an issue ?

I played around with the veracode `mitigated_finding.xml` test file, I switched it to "proposed" mitigation on the last import
image

Gby56:
After trying to address the issue with the "untouched" count I gave up, I think it breaks some counting logic and the tests would start failing for no reason. The original implementation just ignored and skipped false pos/risk accepted/out of scope without counting anyways

StefanFl:
I am a bit lost now what the outcome of the discussion is. Do we think this PR makes DefectDojo better and we should merge it or are there still issues that need to be addressed?

Gby56:
I have implemented tests and it fixes a very basic bug, try reimporting the same report with mitigated issues for yourself, they will reopen if they are not false positive, out of scope or risk accepted. So it's missing a huge piece of the implementation imo

dojo/importers/reimporter/reimporter.py
```diff
@@ -82,13 +82,24 @@ def process_parsed_findings(self, test, parsed_findings, scan_type, user, active
...
                 if finding.false_p or finding.out_of_scope or finding.risk_accepted:
                     logger.debug('%i: skipping existing finding (it is marked as false positive:%s and/or out of scope:%s or is a risk accepted:%s): %i:%s:%s:%s', i, finding.false_p, finding.out_of_scope, finding.risk_accepted, finding.id, finding, finding.component_name, finding.component_version)
-                elif finding.mitigated or finding.is_mitigated:
-                    logger.debug('%i: reactivating: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
-                    finding.mitigated = None
-                    finding.is_mitigated = False
-                    finding.mitigated_by = None
-                    finding.active = True
-                    finding.verified = verified
+                elif finding.mitigated and finding.is_mitigated:
```
damiencarol, line 85:
```suggestion
                elif finding.is_mitigated:
```
we can simplify this.

damiencarol (changes requested):

damiencarol (approved):

damiencarol:
> I am a bit lost now what the outcome of the discussion is. Do we think this PR makes DefectDojo better and we should merge it or are there still issues that need to be addressed?

@StefanFl we made a review live with @Gby56 . everything is good. we also saw that this part need more work. I can discuss that with you later but the PR as is is good and move in the right direction. @Gby56 also shared that he will push more PR.

I guess more parser use the mitigation data the more we will need to add features on this part.

italvi:
@Gby56 this only solves a reimport, right? When using the import-scan API adding a scan twice, my already mitigated findings are getting reopened because "This finding has been automatically re-opened as it was found in recent scans".

Gby56:
Hi @italvi ! Yes only reimport, the logic for comparing existing findings to new ones was reopening things even if they were marked closed, your finding had to be false positive or out of scope to stay closed, which was a bit wild.
If there is a flaw in the import API I'd be surprised though, I don't recall such an issue, I tried uploading reports daily and likely with the same findings and nothing was reopened...

italvi:
Hi @Gby56, maybe I am doing it wrong? I am just closing them by clicking on "close finding", it seems more reasonable as I have to write there a rational for closing the finding. By clicking on "Edit" and then use the checkbox for "false positive" there won't be any comment, why it is an false positive.

PR #6455, opened 2022-06-23
Setting the date of Bugcrowd findings to the 'submitted_at' value

coheigea:
When you import findings from Bugcrowd it just sets the current date. Instead it should get the date from submitted_at in the imported CSV.

### DIFF
--- a/dojo/tools/bugcrowd/parser.py
+++ b/dojo/tools/bugcrowd/parser.py
@@ -2,6 +2,7 @@ import csv
 import hashlib
 import io
 
+from dateutil import parser
 from dojo.models import Endpoint, Finding
 
 
@@ -58,6 +59,9 @@ class BugCrowdParser(object):
             finding.steps_to_reproduce = pre_description.get('steps_to_reproduce', None)
             finding.references = References
             finding.severity = self.convert_severity(row.get('priority', 0))
+
+            if row.get('submitted_at'):
+                finding.date = parser.parse(row.get('submitted_at'))
 
             if url:
                 finding.unsaved_endpoints = list()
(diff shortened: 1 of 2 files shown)


StefanFl (approved):

37b:
@coheigea does the `submitted_at` date take into account the time that is spent in triage by the BugCrowd team?

coheigea:
@37b I will check with Bugcrowd. I suspect it does not take the triaging into account. IMO a good enhancement would be to add a configuration switch per-product, which would allow you to choose whether to set the SLAs from when an issue was first submitted to DefectDojo, versus the date it was found in the tool. WDYT? Then the change in this PR would be configurable.

37b:
> @37b I will check with Bugcrowd. I suspect it does not take the triaging into account. IMO a good enhancement would be to add a configuration switch per-product, which would allow you to choose whether to set the SLAs from when an issue was first submitted to DefectDojo, versus the date it was found in the tool. WDYT? Then the change in this PR would be configurable.

Yeah something like that makes sense. I know in the past we've had bugs sitting in "Blocked" while there's lots of back and forth before it actually turns to an accepted Finding. That time shouldn't count against the developers' SLA.

I think we use the Generic Importer for BugCrowd right now to get around some of those quirks.

dojo/tools/bugcrowd/parser.py
```diff
@@ -59,6 +60,9 @@ def get_findings(self, filename, test):
             finding.references = References
             finding.severity = self.convert_severity(row.get('priority', 0))
 
+            if row.get('submitted_at'):
+                finding.date = datetime.strptime(row.get('submitted_at'), '%Y-%m-%d %H:%M:%S %Z')
+
```
damiencarol, line 65: Could you switch to `dateutil` module instead?

Maffooch (approved):

dojo/tools/bugcrowd/parser.py
coheigea: Done @damiencarol

damiencarol (approved):

PR #6470, opened 2022-06-27
Populate CWE finding field from CycloneDX JSON file

coheigea:


dojo/tools/cyclonedx/parser.py
```diff
@@ -349,6 +349,14 @@ def _get_findings_json(self, file, test):
                 if vulnerability_ids:
                     finding.unsaved_vulnerability_ids = vulnerability_ids
 
+                # if there is some CWE
+                cwes = vulnerability.get('cwes')
+                if cwes and len(cwes) > 1:
+                    # FIXME support more than one CWE
+                    LOGGER.debug(f"more than one CWE for a finding {cwes}. NOT supported by parser API")
```
StefanFl, line 356: Why don't we take just the first CWE? Better than none.

StefanFl (approved):

StefanFl (changes requested):
see comment

dojo/tools/cyclonedx/parser.py
coheigea: @StefanFl It does actually take the first CWE in all cases. This if statement just logs a warning about there being more than one CWE, the next if statement takes the first CWE.

coheigea:
@StefanFl Please review again

StefanFl (approved):

damiencarol (approved):

Issue #6468, opened 2022-06-27
API Authentication fails after restart/redeployment


0x4d4e:
**Bug description**

We are running DefectDojo on ECS Fargate using a template that is based on the included docker-compose file. After resuming the service after it was paused due to inactivity on our test environments or after redeploying the service (e.g. due to a new version) accessing DefectDojo via the API/access key fails with the exception included below. 

The exception message points at an issue with the request caching?

API access works again after a user authenticates via the web interface.

**Steps to reproduce**

1. Resume/Redeploy the service (i.e. all included containers)
2. Attempt to access the DefectDojo API via the previously generated (and working) API key
3. *Error*

**Expected behavior**
Accessing the API via the API key should immediately work after resuming/redeploying the containers.

**Deployment method** *(select with an `X`)*
- [x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo
- [X] ECS Fargete (based on Docker-Compose template)

**Environment information**
 - ECS Fargate with official defectdojo Docker images
 - DefectDojo version 2.11.0 (noticed this from 2.8.x onwards, could have also existed earlier)

**Logs** 
The following logs are output by the uwsgi container when attempting to access the instance via the API after a restart/redeployment.

````
[27/Jun/2022 11:42:09] ERROR [django.request:224] Internal Server Error: /api/v2/product_types/
Traceback (most recent call last):
File "/app/./dojo/request_cache/__init__.py", line 53, in wrapper
result = getattr(cache, key)
AttributeError: 'RequestCache' object has no attribute '('get_authorized_users', <Permissions.Finding_View: 1402>, <object object at 0x7f7f5b4115b0>)'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
File "/app/./dojo/request_cache/__init__.py", line 53, in wrapper
result = getattr(cache, key)
AttributeError: 'RequestCache' object has no attribute '('get_groups', <SimpleLazyObject: <django.contrib.auth.models.AnonymousUser object at 0x7f7f5a02f160>>, <object object at 0x7f7f5b4115b0>)'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
response = get_response(request)
File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 167, in _get_response
callback, callback_args, callback_kwargs = self.resolve_request(request)
[...]

PR #6478, opened 2022-06-28
Added middleware to set request.user from AuthToken

0x4d4e:
See https://github.com/DefectDojo/django-DefectDojo/issues/6468

PoC to set request.user in a custom middleware when TokenAuthentication is used for API access.

TODO: 
* better alternatives?
* middleware ordering?
* exception handling
* tests!


Maffooch:
I have run into this issue as well, but I fear the added middleware approach is a bit too hackish for me. I'll do some thinking on this.

@StefanFl I did not fully understand this PR until looking closely at the associated issue. This issue has occured in my prod env when the VM would restart do to unattended upgrades. On reboot, our pipeline would fuss and fail for a few requests, and then resolve itself. My suspicion is that that the [request_cache](https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/request_cache/middleware.py) is being dumped on restart. What are your thoughts?

0x4d4e:
I also think that caching somehow fixes this issue (at least for the same user - not sure since we are currently only using the admin to access the API). However, the self-healing in repeated requests does not work reliably in all instances.

Furthermore, all API requests using Token authentication are missing the user attribute. This also affects the audit logging, which shows no user in the log output for API requests.

We are currently using the hack on top of the latest release to fix the restart issue and so far it works reliably in multiple environments.

StefanFl:
I agree this issue should be addressed and I agree with both of you that this approach might not be the best one. Unfortunately I don't have a better idea at the moment :-(

devGregA:
All, I would suggest we disable the cache. I am against taking user input in this fashion which could be malicious or have other security implications. This issue can be fixed by disabling the cache.

Maffooch:
@0x4d4e I may have fixed this in  a PR recently. Are you able to reproduce this issue when running 2.14.3?

0x4d4e:
@Maffooch Seems to be fixed - updating the version on a stack without my patch worked without the previously required manual login. Thanks!

Out of interest, did removing the cache fix the issue or were some other changes required as well? I tried removing (some of) the caching annotations previously but saw no difference in the behavior.

PR #6492, opened 2022-07-01
Fix parser horusec 

srburton:
**I'm doing this update because when trying to import a report from Horusec you get the following error:**

WhatsApp Image 2022-07-01 at 14 06 30

**I noticed that the reports of the new version of Horusec the "line" comes empty where it ends up breaking the parser of the integer:**

WhatsApp Image 2022-07-01 at 14 09 19

Screenshot_2

### DIFF
--- a/dojo/tools/horusec/parser.py
+++ b/dojo/tools/horusec/parser.py
@@ -46,7 +46,9 @@ class HorusecParser(object):
         test.findings = [self._get_finding(node, report_date) for node in data.get("analysisVulnerabilities")]
         return [test]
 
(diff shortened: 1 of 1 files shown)


renatosazup (approved):

Maffooch (approved):

Maffooch (changes requested):
Just noticed the tests were failing. Please fix those

srburton:
I opened a new PR: https://github.com/DefectDojo/django-DefectDojo/pull/6563

Issue #6289, opened 2022-05-12
Wrong severity mapping for native semgrep JSON parser


Gby56:
https://github.com/DefectDojo/django-DefectDojo/blob/612a0f682ae0a4e007196de4f438782464b3ae3d/dojo/tools/semgrep/parser.py#L67

I think this is really wrong, WARNING level should be the same as the SARIF level https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127839
And this should be Medium in my opinion... or High, and ERROR should be Critical, this is all very weird to me
@damiencarol let me know what you think ? I haven't tried to ingest SARIF yet but I think it would reflect issues better

coheigea:
IMO we should make the following changes:

- Change SARIF "error" mapping to "High" from "Critical"
- Change Semgrep "WARNING" mapping to "Medium" from "Low".

That would ensure both Semgrep/Sarif map High/Medium to the same severity values in DefectDojo

coheigea:
PR submitted: https://github.com/DefectDojo/django-DefectDojo/pull/6508

coheigea:
@Gby56 The issue can be closed

PR #6508, opened 2022-07-04
Align severity mapping between SARIF/semgrep

coheigea:
Align severity mapping between SARIF and semgrep as proposed here:

https://github.com/DefectDojo/django-DefectDojo/issues/6289

- Change SARIF "error" to "High" from "Critical"
- Change Semgrep "warning" to "Medium" from "Low"

With this change both SARIF and Semgrep are consistent with how they map warning and errors.

### DIFF
--- a/dojo/tools/semgrep/parser.py
+++ b/dojo/tools/semgrep/parser.py
@@ -67,7 +67,7 @@ class SemgrepParser(object):
 
     def convert_severity(self, val):
         if "WARNING" == val.upper():
-            return "Low"
+            return "Medium"
         elif "ERROR" == val.upper():
             return "High"
         elif "INFO" == val.upper():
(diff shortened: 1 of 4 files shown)


damiencarol (approved):

damiencarol:
@coheigea please fix the unit tests

dojo/tools/semgrep/parser.py
```diff
@@ -67,7 +67,7 @@ def get_findings(self, filename, test):
 
     def convert_severity(self, val):
         if "WARNING" == val.upper():
-            return "Low"
+            return "Medium"
```
StefanFl, line 70: It looks weird to me, to map a Warning to Medium instead of Low.

coheigea: @StefanFl It's what sarif does. Ultimately we have the problem of mapping external severity levels that only cover 3 possibilities into the 5 in DefectDojo. The goal of the PR is to at least maintain consistency between the mapping levels, because for example we can ingest semgrep using both the semgrep and sarif parsers and end up with different severity levels...

Maffooch (approved):

Issue #6248, opened 2022-05-04
Discrepancy in Severity Mappings for Veracode SCA scans


cah-thomas-howard:
Veracode SCA scans given the following severities based upon CVSS scores:

![Screen Shot 2022-05-04 at 1 53 55 PM](https://user-images.githubusercontent.com/38532219/166802766-00a583c9-7cbb-439c-8ce9-18c0e3e73d75.png)

However, DefectDojo tags the severity using this criteria:

![Screen Shot 2022-05-04 at 1 58 10 PM](https://user-images.githubusercontent.com/38532219/166802847-a58c5bdb-ef57-4aec-ba46-cbc1dec4b6e2.png)

Given this, a CVSS score of 4.0 will put the severity at 'Medium' in DD but the description assigned to the finding will show CVS Score 4.0 (Low) since the severity value coming from Veracode is used within the description:

![Screen Shot 2022-05-04 at 2 20 26 PM](https://user-images.githubusercontent.com/38532219/166803508-1c8075e1-b73f-44cf-bbcf-7b555a8011d3.png)

Personally, I would think the CVSS breakdown within Veracode would be used since it's well defined, with Very High mapping to Critical and Low/Very Low mapping to Low.

Veracode SCA Severity mapping
https://docs.veracode.com/r/review_severity_exploitability

DefectDOJO Veracode parser SCA Severity Mapping
https://github.com/DefectDojo/django-DefectDojo/blob/612a0f682ae0a4e007196de4f438782464b3ae3d/dojo/tools/veracode/parser.py#L192

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

 - DefectDojo version 2.7.1

damiencarol:
> Personally, I would think the CVSS breakdown within Veracode would be used since it's well defined, with Very High mapping to Critical and Low/Very Low mapping to Low.

I agree. I think we can improve that. Will try to create a patch for that.

coheigea:
@damiencarol I think the problem in the Veracode parser is that it's manually mapping CVSS scores to DefectDojo severity. However, the Veracode XML actually contains the severity score. So we should set the DefectDojo severity according to:

- 5 -> Critical
- 4 -> High
- 3 -> Medium
- 2 -> Low
- 1 and 0 -> Info

WDYT? I can prepare a patch if you agree.

coheigea:
This issue can be closed

PR #6513, opened 2022-07-04
Take severity from Veracode rather than attempt to map the CVSS score…

coheigea:
… to a severity

See: https://github.com/DefectDojo/django-DefectDojo/issues/6248

### DIFF
--- a/dojo/tools/veracode/parser.py
+++ b/dojo/tools/veracode/parser.py
@@ -192,19 +192,6 @@ class VeracodeParser(object):
 
         return finding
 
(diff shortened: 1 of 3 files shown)


StefanFl (commented):
Doesn't that change affects any unit test?

coheigea:
@StefanFl I added a unit test that would have failed under the old code

StefanFl (approved):

damiencarol (approved):
this is a very good patch. we should not try to re-interpret the data of the tool

PR #6515, opened 2022-07-05
Update parser of DependencyTrack

italvi:
As cve was replaced by vuln_id the filtering for cves from NVD is unnecessary

### DIFF
--- a/unittests/tools/test_dependency_track_parser.py
+++ b/unittests/tools/test_dependency_track_parser.py
@@ -46,7 +46,7 @@ class TestDependencyTrackParser(DojoTestCase):
         self.assertEqual(1, len(findings[0].unsaved_vulnerability_ids))
         self.assertEqual('533', findings[0].unsaved_vulnerability_ids[0]))
         self.assertEqual(1, len(findings[1].unsaved_vulnerability_ids))
-        self.assertEqual('48', findings[1].unsaved_vulnerability_ids[0]))
+        self.assertEqual('48', findings[1].unsaved_vulnerability_ids[0])
         self.assertEqual(1, len(findings[2].unsaved_vulnerability_ids))
         self.assertEqual('CVE-2016-2097', findings[2].unsaved_vulnerability_ids[0])
         self.assertEqual(1, len(findings[3].unsaved_vulnerability_ids))
--- a/unittests/tools/test_dependency_track_parser.py
+++ b/unittests/tools/test_dependency_track_parser.py
@@ -43,8 +43,10 @@ class TestDependencyTrackParser(DojoTestCase):
         testfile.close()
         self.assertEqual(4, len(findings))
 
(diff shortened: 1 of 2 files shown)


damiencarol (approved):

unittests/tools/test_dependency_track_parser.py
```diff
@@ -43,8 +43,10 @@ def test_dependency_track_parser_has_many_findings(self):
         testfile.close()
         self.assertEqual(4, len(findings))
 
-        self.assertIsNone(findings[0].unsaved_vulnerability_ids)
-        self.assertIsNone(findings[1].unsaved_vulnerability_ids)
+        self.assertEqual(1, len(findings[0].unsaved_vulnerability_ids))
+        self.assertEqual('533', findings[0].unsaved_vulnerability_ids[0])
```
StefanFl, line 47: The vulnerability id is defined as "An id of a vulnerability in a security advisory associated with this finding. Can be a Common Vulnerabilities and Exposures (CVE) or from other sources." These examples are just numbers that don't fit to this definition.

StefanFl (changes requested):

unittests/tools/test_dependency_track_parser.py
damiencarol: you mean you prefer to build "CWE-{x}" string?

StefanFl: The vulnerability ids don't store CWEs but something like GHSA-57j2-w4cx-62h2 or CVE-2020-36518. They will contain the link to the source in the UI or could be googled.

damiencarol: so we completly remove CWE from this field?

Maffooch:
@italvi apologies for the lull on this one. I agree with Stefan's point that putting in just the number does not actually do much in this case. The vulnerability ID field requires some sort of identifier to build a hyperlink with (for example CVE or GHSA)

In my test instance I manually set the vuln id to `NPM-533` and it displays the value in plain text because a hyperlink could not be created. I tried to find what this link would look like, but could not figure out were `NPM-533` is supposed to be pointing. 

The only way we could go forward with this PR is to make the dep tracker parser take the `NPM` and `533` parts into the vulnerability ID field. 

image

italvi:
@Maffooch as every NPM Advisory is stored in Github Advisories and therefore we also find a reference in the vulnerability to the GHSA, maybe it would be useful to check in the References (see in the image below) for such a link and then take the ID from there? Would I need just the last part, e.g. GHSA-qh73-qc3p-rjv2?
image

Maffooch:
@italvi Sorry for the delayed response. I was on vacation for a bit. 

Yes I like your idea of searching the references for potential vulnerability IDs. Would be nice to capture the NVD and GHSA through regex or somethingg

PR #6523, opened 2022-07-07
Fix 500 error on adding finding to group name

coheigea:
When you try to add a finding to a group by group name, you get a 500 error. This is because the form uses a boolean instead of a charfield. The form name is a bit confusing, so I renamed it to make it clearer that it's an id.


Maffooch (approved):

mtesauro (approved):
Approved

PR #6524, opened 2022-07-07
Fix bug with creating two finding group JIRAs on group creation

coheigea:
When you create a finding group and select "Push to JIRA" it creates two JIRAs for the finding group.


mtesauro (approved):
Approved

Issue #6137, opened 2022-04-07
(UI) Merge Findings does not work if "Finding to Merge Into" is changed


aklyukin:
**Bug description**
It is not possible to merge finds if you change the finding in the "Finding to Merge Into" field. Because it is not possible to mark a finding(s) in the "Findings to Merge" field. I assume that the problem is only in the Web UI

**Steps to reproduce**
1. Select two findings for merge and click the "merge" button
2. Chose "Finding Action" - "Inactive"
3. Change "Finding to Merge Into" to another finding
4. Try to mark findning(s) in "Findings to Merge" drop down menu
5. Marked finds do not appear as added. Unable to press merge findings button - popup appears "please select an item in the list"

**Expected behavior**
Issues:
1. When I mark the Findings in the "Findings to Merge" drop-down list, these Findings should appear in the field
2. The Finding that is selected in the ""Finding to Merge Into"" field should not appear in the "Findings to Merge" list

**Deployment method** *(select with an `X`)*
- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - https://demo.defectdojo.org/
 - Operating System: CentOS Linux release 7.6
 - DefectDojo version: 2.7.1 - 2.9.0

**Screenshots**
<img width="878" alt="Screenshot 2022-04-07 at 04 47 42" src="https://user-images.githubusercontent.com/9791671/162104282-aa1977ab-f715-4571-b1cb-114bd21b89c0.png">

PR #6529, opened 2022-07-08
Fix UI bug with merging fingings

coheigea:
A fix for https://github.com/DefectDojo/django-DefectDojo/issues/6137

I simplified the code so that on first load it selects all findings to merge that aren't selected by default as the finding to merge into. Then when finding to merge in to is changed, it unselects it from the findings to merge.


Maffooch (approved):

mtesauro (approved):
Approved

PR #6532, opened 2022-07-08
Don't wrap lines when parsing Acunetix

coheigea:
When parsing Acunetix 360 findings, it wraps the description and other fields. This should be disabled as it looks odd. The burp parser does something similar.


StefanFl (approved):
Yes, looks much better :-)

damiencarol (approved):

Issue #6252, opened 2022-05-05
Veracode SCA Vulns with Same CVE Not Importing


cah-thomas-howard:
When parsing through SCA components, the 'dup_key' is set to the CVE ID. As a result, when two or more components have the same CVE, only the first one is imported.

Code in question:
![Screen Shot 2022-05-05 at 8 49 10 AM](https://user-images.githubusercontent.com/38532219/166932499-653acf6d-fc5a-4e1e-97fd-4c073f027144.png)

Veracode report showing two components with the same CVE:
![Screen Shot 2022-05-05 at 9 10 13 AM](https://user-images.githubusercontent.com/38532219/166932646-97d5e561-cbb7-4cf0-b80b-a720797c3255.png)

Only one component imported:
![Screen Shot 2022-05-05 at 9 07 11 AM](https://user-images.githubusercontent.com/38532219/166932966-f4645a9e-52b8-4e7a-a93e-894740a6f9e1.png)

Perhaps the dup_key could be a combination of the component_id and cve_id? That way each component will be imported but any dupe CVEs within the component vulns will be removed.

Code in question:
https://github.com/DefectDojo/django-DefectDojo/blob/612a0f682ae0a4e007196de4f438782464b3ae3d/dojo/tools/veracode/parser.py#L70

coheigea:
@StefanFl It can be closed now I think

damiencarol:
@coheigea I reviewed your PR and it's good as it fixes the bug but I really think that we should get ride of the internal aggregation and simplify the parser. we should not do the de-duplication in the parser.
It's ok for now but we can simplify code in later PRs

PR #6537, opened 2022-07-11
Fix de-duplication bug with Veracode SCA

coheigea:
A proposed fix for https://github.com/DefectDojo/django-DefectDojo/issues/6252

### DIFF
--- a/dojo/tools/veracode/parser.py
+++ b/dojo/tools/veracode/parser.py
@@ -1,4 +1,5 @@
 import re
+import uuid
 from datetime import datetime
 
 from defusedxml import ElementTree
@@ -75,10 +76,8 @@ class VeracodeParser(object):
             _version = component.attrib['version']
 
             for vulnerability in component.findall('x:vulnerabilities/x:vulnerability', namespaces=XML_NAMESPACE):
-                dupe_key = vulnerability.attrib['cve_id']
-                # Only process if we didn't do that before.
-                if dupe_key not in dupes:
-                    dupes[dupe_key] = self.__xml_sca_flaw_to_finding(test, report_date, _vendor, _library, _version, vulnerability)
+                # We don't have a Id for SCA findings so just generate a random one
+                dupes[str(uuid.uuid4())] = self.__xml_sca_flaw_to_finding(test, report_date, _vendor, _library, _version, vulnerability)
 
         return list(dupes.values())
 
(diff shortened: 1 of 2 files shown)


dojo/tools/veracode/parser.py
```diff
@@ -202,13 +202,13 @@ def _get_cwe(val):
...
     @classmethod
-    def __xml_sca_flaw_to_finding(cls, test, report_date, vendor, library, version, xml_node):
+    def __xml_sca_flaw_to_finding(cls, test, report_date, vendor, library, version, xml_node, dupe_key):
         # Defaults
         finding = Finding()
         finding.test = test
         finding.static_finding = True
         finding.dynamic_finding = False
-        finding.unique_id_from_tool = xml_node.attrib['cve_id']
+        finding.unique_id_from_tool = dupe_key
```
StefanFl, line 211: This is not a good way to use this field. The id here should come directly from the tool and shouldn't be generated. To fix deduplication we should calculate a meaningful hash_code with and use that for deduplication in `settings.dist.py`.

unittests/tools/test_veracode_parser_dedup.py
```diff
@@ -0,0 +1,46 @@
...
+        self.assertEqual("Medium", finding.severity)
+        self.assertEqual(1, len(finding.unsaved_vulnerability_ids))
+        self.assertEqual("CVE-2012-6153", finding.unsaved_vulnerability_ids[0])
+        self.assertEqual(20, finding.cwe)
+        self.assertEqual("commons-httpcore", finding.component_name)
+        self.assertEqual("3.1", finding.component_version)
+        self.assertEqual("CVE-2012-6153_commons-httpcore_3.1", finding.unique_id_from_tool)
+        self.assertEqual(4.3, finding.cvssv3_score)
+
+    def enable_dedupe(self, enable=True):
```
StefanFl, line 43: You shouldn't change the settings in a unit test, it might have effects on other test cases. There must be better ways to achieve this, maybe with a @patch or a mock.

StefanFl (changes requested):

Maffooch (dismissed):

coheigea:
@StefanFl @Maffooch Thanks for the review - I updated the PR to just generate a random key for the dupes map in the Veracode parser, and removed the unique_id_from_tool to be the CVE id. This means the de-duplication logic falls back to compute_hash_code_legacy for SCA, which solves the problem.

Maffooch (approved):

mtesauro (approved):
Approved

PR #6551, opened 2022-07-12
Only group by if the name/component is not none

coheigea:
With a tool like Veracode which contains both SAST and SCA findings, if you group on import by component name/version then it will group all SAST findings under one entry with None:None. Similarly, if you group by filepath, it groups all SCA findings under a single entry with FilePath: None.

I think it only makes sense to group if the thing you are grouping by is not None.


Maffooch (approved):
Makes sense to me.  Good change.

Issue #6558, opened 2022-07-13
Auto_create_context in reimport scan changes the engagement deduplication


Gby56:
When trying to create engagements dynamically with different deduplication modes, I realized that the setting would be switched back to "disabled" for the deduplicate only in engagement, which is not ideal

**Expected behavior**
Either let the user input that param in the reimport-scan API or don't change it

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version 2.12.0

**Logs** 
Use `docker-compose logs` (or similar, depending on your deployment method) to get the logs and add the relevant sections here showing the error occurring (if applicable).

**Sample scan files**
If applicable, add sample scan files to help reproduce your problem.

**Screenshots**

**Additional context** (optional)

manuel-sommer:
This can be closed @mtesauro

PR #6562, opened 2022-07-14
Deduplication for Engagement only when auto_create_context = True

37b:
#6558 

Added `deduplication_on_engagment` parameter to `import_scan` and `reimport_scan` APIs to enable `deduplication_on_engagement` on newly created Engagements

### DIFF
--- a/docker-compose.override.dev.yml
+++ b/docker-compose.override.dev.yml
@@ -27,6 +27,7 @@ services:
     volumes:
       - './dojo/static/dojo:/usr/share/nginx/html/static/dojo'
   mysql:
+    platform: linux/amd64
     ports:
       - target: ${DD_DATABASE_PORT}
         published: ${DD_DATABASE_PORT}
--- a/docker-compose.override.dev.yml
+++ b/docker-compose.override.dev.yml
@@ -27,7 +27,6 @@ services:
     volumes:
       - './dojo/static/dojo:/usr/share/nginx/html/static/dojo'
   mysql:
-    platform: linux/amd64
     ports:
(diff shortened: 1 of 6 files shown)


37b:
Will resubmit when tests are passing.

docker-compose.override.dev.yml
```diff
@@ -27,6 +27,7 @@ services:
     volumes:
       - './dojo/static/dojo:/usr/share/nginx/html/static/dojo'
   mysql:
+    platform: linux/amd64
```
Maffooch, line 30: Why is this added?

requirements.txt
```diff
@@ -18,6 +18,7 @@ django-slack==5.17.8
 django-tagging==0.5.0
 django-watson==1.6.2
 django-prometheus==2.2.0
+django-rest-multiple-models==2.1.3
```
Maffooch, line 21: Where is this used in this PR?

Maffooch (commented):
Good PR. Couple things for clarification

docker-compose.override.dev.yml
37b: I keep adding that to get around local dev for arm/64 (Mac) ...I'll remove it.

requirements.txt
37b: It wasn't!...but I kept getting errors while running the app until I added that. Looks like a rebase may have fixed it. I'll remove it and repush.

Maffooch (approved):

StefanFl (approved):
Good enhancement

PR #6563, opened 2022-07-14
Fix parser horusec

srburton:
**I'm doing this update because when trying to import a report from Horusec you get the following error:**

WhatsApp Image 2022-07-01 at 14 06 30

**I noticed that the reports of the new version of Horusec the "line" comes empty where it ends up breaking the parser of the integer:**

WhatsApp Image 2022-07-01 at 14 09 19

Screenshot_2


coheigea (changes requested):
Can you add a unit test to unittests/tools/test_horusec_parser.py for it?

renatosazup (approved):

Maffooch (changes requested):
Please fix Flake8 and add unit tests

renatosazup:
Ok @Maffooch

mtesauro (approved):
Approved

renatosazup (approved):

renatosazup (approved):

damiencarol:
@srburton @renatosazup : the comment of @Maffooch is still there, we need a unit test with a report that don't have the line (empty value) to show that the fix works.

srburton:
@damiencarol added unit test!

Maffooch:
@srburton Thanks for the new scan report!

Please test against it in the horusec unit test file [here](https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/tools/test_horusec_parser.py)

srburton:
@Maffooch check my fix, please?

renatosazup (approved):

Maffooch (approved):

Maffooch:
@damiencarol leaving merge to you if it looks good

PR #6564, opened 2022-07-14
Add file upload extension allow list, Force authorization to download file

Maffooch:
In order to make the management of arbitrary files more secure,  I have added the following behaviors:
- Users must be able to edit the object a file belongs to download the file
- A whitelist of supported file types is configurable by a setting/env_var
  - `['.txt', '.pdf', '.json', '.xml', '.yml', '.png', '.jpeg']`

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -232,7 +232,8 @@ env = environ.Env(
     # Feature toggle for new authorization for configurations
     DD_FEATURE_CONFIGURATION_AUTHORIZATION=(bool, True),
     # List of acceptable file types that can be uploaded to a given object via arbitrary file upload
-    DD_FILE_UPLOAD_TYPES=(list, ['.txt', '.pdf', '.json', '.xml', '.yml', '.png', '.jpeg']),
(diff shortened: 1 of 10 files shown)


dojo/settings/settings.dist.py
```diff
@@ -231,6 +231,8 @@
     DD_DELETE_PREVIEW=(bool, True),
     # Feature toggle for new authorization for configurations
     DD_FEATURE_CONFIGURATION_AUTHORIZATION=(bool, True),
+    # List of acceptable file types that can be uploaded to a given object via arbitrary file upload
+    DD_FILE_UPLOAD_TYPES=(list, ['.txt', '.pdf', '.json', '.xml', '.yml', '.png', '.jpeg']),
```
coheigea, line 235: There are quite a few other file extensions in the unittests folder, for example:

.csv, .html, .sarif, .xslx, .html, .js, .nessus, .zip

Maffooch: Good suggestion

mtesauro (approved):
Approved.

dzmitry-savitski:
Is there a way to get file content via API after this change? It seems like authorization is hardcoded now, and there's no API endpoint to get file contents.

Maffooch:
Hi @dzmitry-savitski 
This is a good point. I forgot that files could be retrieved by the API. Will work on a solution

PR #6566, opened 2022-07-15
Testing: Import and reimport Veracode test

Gby56:
Related to https://github.com/DefectDojo/django-DefectDojo/pull/6452
This test is to show that current reimporter fails to import the same veracode report twice without changes. It should fail for now, and work correctly once #6452 is merged

### DIFF
--- a/unittests/test_import_reimport.py
+++ b/unittests/test_import_reimport.py
@@ -384,7 +384,7 @@ class ImportReimportMixin(object):
         notes_count_before = self.db_notes_count()
 
         # # reimport exact same report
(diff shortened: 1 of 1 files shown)


Gby56:
Weirdly, the new test `test_import_veracode_reimport_veracode_active_verified` passed without any issues, where it should have failed...

Gby56:
Found out why: the many_findings in veracode tests doesn't have a mitigated static flaw, the only one there is "proposed" 🤦

Gby56:
Test failing as expected

Issue #6575, opened 2022-07-18
"engagement_added" notification is sent twice


bgoareguer:
**Bug description**
After activating email notifications for the "Engagement Added" event, I receive 2 different emails.
- the Subjects are not the same
- the link to the engagement in the bodies point to the correct engagement but are not displayed identically

**Steps to reproduce**
1. Go to a specific product
2. Activate the email notification for the "Engagement Added" event
3. Create a new engagement
4. You will receive 2 emails

**Expected behavior**
A single email should be sent.

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [x] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Ubuntu 20.04
 - DefectDojo version: 2.12.0

**Additional context** (optional)
It seems that `create_notification` with event `engagement_added` is called twice:
- [dojo/product/views.py:993](https://github.com/DefectDojo/django-DefectDojo/blob/2.12.0/dojo/product/views.py#L993)
- [dojo/utils.py:1747](https://github.com/DefectDojo/django-DefectDojo/blob/2.12.0/dojo/utils.py#L1747)

The email sent from [dojo/utils.py:1747](https://github.com/DefectDojo/django-DefectDojo/blob/2.12.0/dojo/utils.py#L1747) seems better than the one sent from [dojo/product/views.py:993](https://github.com/DefectDojo/django-DefectDojo/blob/2.12.0/dojo/product/views.py#L993).

StefanFl:
merged for release 2.13.0

PR #6579, opened 2022-07-19
Don't create two notifications when an engagement is added

coheigea:
Fixes https://github.com/DefectDojo/django-DefectDojo/issues/6575


Maffooch:
Where is the other notification being added?

coheigea:
@Maffooch Here: https://github.com/DefectDojo/django-DefectDojo/blob/2.12.0/dojo/utils.py#L1747

Maffooch (approved):
Those signals can be real sneaky sometimes

mtesauro (approved):
Approved

PR #6581, opened 2022-07-19
Parse Acunetix360 State field for risk acceptance and false positive …

coheigea:
…findings

Right now it just marks all findings as active, even if they are already triaged by Acunetix360.


Maffooch (approved):

mtesauro (approved):
Approved

PR #6590, opened 2022-07-21
Expose prod_type on ApiFindingFilter

### LINKED ISSUE #6589: Specific filter not exposed on ApiFindingFilter but on FindingFilter (test__engagement__product__prod_type)
The field `test__engagement__product__prod_type` is exposed as a filter on the `FindingFilter` class:
https://github.com/DefectDojo/django-DefectDojo/blob/fbd6cc78e6880ae7312193a31c6a4a0f1eb62420/dojo/filters.py#L1165-L1167

That's great, because it allows people to filter findings by a product type in the UI, via `/finding/open?test__engagement__product__prod_type=<id>`.

Unfortunately, this field is not exposed as a filter on `ApiFindingFilter` though.
Only `test__engagement__product` is exposed, see:
https://github.com/DefectDojo/django-DefectDojo/blob/fbd6cc78e6880ae7312193a31c6a4a0f1eb62420/dojo/filters.py#L1072

**Expected behaviour**
I would expect the behaviour between the UI filters and API filters to be similar, in the sense that the `test__engagement__product__prod_type` field is also exposed on the API, and not just in the UI.

**Deployment method**
- [x] Docker Compose
- [x] Kubernetes
- [x] GoDojo

**Environment information**
 - Operating System: MacOS 12.04
 - DefectDojo version: 2.12.0

fw-work:
This PR exposes the `test__engagement__product__prod_type ` on the `ApiFindingFilter`.
Previously it was already exposed on the `FindingFilter` and is used in the UI, e.g. here: `/finding/open?test__engagement__product__prod_type=<id>`.

Fixes #6589.

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -1070,6 +1070,7 @@ class ApiFindingFilter(DojoFilter):
     test__test_type = NumberInFilter(field_name='test__test_type', lookup_expr='in', label='Test Type')
     test__engagement = NumberInFilter(field_name='test__engagement', lookup_expr='in')
     test__engagement__product = NumberInFilter(field_name='test__engagement__product', lookup_expr='in')
+    test__engagement__product__prod_type = NumberInFilter(field_name='test__engagement__product__prod_type', lookup_expr='in')
     finding_group = NumberInFilter(field_name='finding_group', lookup_expr='in')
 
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

PR #6614, opened 2022-07-27
Correct wording on copy text, add more places to copy

Maffooch:
- Correct the wording on the copy operation on engageements
- Add to the number of places that an object can be copied from

### DIFF
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -368,7 +368,7 @@ def copy_engagement(request, eid):
     return render(request, 'dojo/copy_object.html', {
         'source': engagement,
         'source_label': 'Engagement',
-        'destination_label': 'Engagement',
+        'destination_label': 'Product',
         'product_tab': product_tab,
         'form': form,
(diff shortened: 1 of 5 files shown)


StefanFl:
@Maffooch: Why is this checkbox `Done` needed?

StefanFl:
@damiencarol see https://owasp.slack.com/archives/GQKK61742/p1658870922995239

Maffooch:
I needed some kind of form to prevent a misclick of the copy button from actually copying anything without a second confirmation. Instead of making a new form for this specific page, I used the `DoneForm` that only has a checkbox that has a required field

StefanFl:
> I needed some kind of form to prevent a misclick of the copy button from actually copying anything without a second confirmation. Instead of making a new form for this specific page, I used the DoneForm that only has a checkbox that has a required field

Ok

StefanFl (approved):

damiencarol (approved):

Issue #6588, opened 2022-07-21
Reassign an engagement to another product


gzot:
Hello,

Using the web interface, I can edit an engagement (https://demo.defectdojo.org/engagement/[ID]/edit).
Last option is to change the associated product but when I submit the form ("Done" button), the new product is not saved.

Can we move an engagement to another product using the web interface ?

Note : It is possible to do the same thing using the PATCH verb on the following API endpoint : https://demo.defectdojo.org/api/v2/engagements/[ID]/

X0x1RG9f:
@gzot, this should be fixing the issue: https://github.com/DefectDojo/django-DefectDojo/pull/6617

manuel-sommer:
PR was merged. can we close this @mtesauro ?

PR #6617, opened 2022-07-28
[FIX] Reassign an engagement to another product

X0x1RG9f:
On the Engagement Edit page, it is possible to change the product. However, this has no effect on real engagement's product assignment. Adding the line to fix the issue.

https://github.com/DefectDojo/django-DefectDojo/issues/6588

### DIFF
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -221,6 +221,7 @@ def edit_engagement(request, eid):
         if form.is_valid():
             # first save engagement details
             new_status = form.cleaned_data.get('status')
+            engagement.product = form.cleaned_data.get('product')
             engagement = form.save(commit=False)
             if (new_status == "Cancelled" or new_status == "Completed"):
                 engagement.active = False


Maffooch (approved):

StefanFl:
I am not sure if this is a good idea. Deduplication works on a product level per default. So the old product might have findings that were marked as duplicates which are not duplicates anymore after moving the engagement and for the the new product deduplication would need to be run.

@X0x1RG9f Is there a strong use case to move an engagement to another product? If yes, the deduplication issue needs to be solved; if no we should remove the `product` field from editing engagements.

Maffooch:
This is a fix for #1512 that stopped working at some point

StefanFl:
@Maffooch Well, yes, but aren't the problems with the deduplication relevant?

Maffooch:
Most likely yes. I believe we could manually kick the dedupe action after saving the engagement, but I am not sure how effective it would be without modification.

devGregA:
I see Stefan's point, but I would assume most users would expect the copy to not modify data, rather than us trying to modify it smartly.

devGregA (approved):

kiblik:
I know, this is already merged but I also suppose there should not be able easily to change the product for engagement.

There is high chance, that `Engagement` contains `Test`. This `Test` may contain `Finding` and `Finding` may contain `Endpoint`. But `Endpoint` are directly connected to `Product`, not only to `Finding`. Change of `Product` for `Engagement` can create a lot of problems if nothing else is changed. Same for `Product_API_Scan_Configuration` and `DojoMeta`.

Issue #6620, opened 2022-07-28
Create an API import for Bugcrowd


Gby56:
**Scanner Name**
Bugcrowd, for bug bounty and vulnerability disclosure programs.
The current import is for a CSV that can only be downloaded in the UI, not really automatable.

**Sample File**
See the api docs https://docs.bugcrowd.com/api/2021-10-28/#tag/submission_resource/operation/listSubmissions

Gby56:
I'm working on this

manuel-sommer:
The PR was merged, could you close this @mtesauro ?

Issue #6580, opened 2022-07-19
Bugcrowd parser breaks on submission titles that contain special characters


Gby56:
https://github.com/DefectDojo/django-DefectDojo/blob/fbd6cc78e6880ae7312193a31c6a4a0f1eb62420/dojo/tools/bugcrowd/parser.py#L55

I had an occurrence today when trying to import a CSV that had a title containing : `xxxxxx http://xxx.domain.example.com`

The defectdojo I get is:
`An exception error occurred during the report import:invalid scheme: 'xxxxxx http'. Only alphanumeric, "+", "-", and "." allowed. Did you meant to call URL.from_text()?`

I think `:` is not allowed and breaks a lot of things

manuel-sommer:
The PR was merged, could you close this @mtesauro ?

PR #6621, opened 2022-07-28
#6620 Create API importer for Bugcrowd

Gby56:
Implements https://github.com/DefectDojo/django-DefectDojo/issues/6620

Needs testing too


dojo/settings/settings.dist.py
```diff
@@ -1257,6 +1259,7 @@ def saml2_attrib_map_format(dict):
     'Gitleaks Scan': DEDUPE_ALGO_HASH_CODE,
     'pip-audit Scan': DEDUPE_ALGO_HASH_CODE,
     'Edgescan Scan': DEDUPE_ALGO_HASH_CODE,
+    'Bugcrowd API': DEDUPE_ALGO_HASH_CODE,
```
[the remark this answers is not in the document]
Gby56: because I blindly copy-pasted for now :P But yeah it makes sense, there is an id from the tool !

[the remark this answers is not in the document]
Gby56: Done !

Gby56:
I have added a regex parsing for endpoints because I kept getting errors I have mentioned in https://github.com/DefectDojo/django-DefectDojo/issues/6580 ! Submissions can have random stuff in their bug_url, not just urls which is annoying...

Maffooch:
@Gby56 This is a great start here! There a couple things left to do before we can approve
 - Fix flake 8 errors
 - Add some unit tests. This would be [good starting place](https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/tools/test_sonarqube_api_parser.py)

dojo/tools/bugcrowd_api/api_client.py
```diff
@@ -0,0 +1,104 @@
...
+
+class BugcrowdAPI:
+    """
+    A simple client for the bugcrowd.io API
+    """
+
+    bugcrowd_api_url = "https://api.bugcrowd.com"
+
+    def __init__(self, tool_config):
+        tool_type, _ = Tool_Type.objects.get_or_create(name='Bugcrowd API')
```
damiencarol, line 15: Shouldn't we use `get` and generate a notification if there is no configuration? Better than creating an empty one that will fail a few lines below

dojo/tools/bugcrowd_api/api_client.py
```diff
@@ -0,0 +1,104 @@
...
+        :param program:
+        :param target:
+        :return:
+        """
+        output = {'data':[]}
+
+        if target:
+            next = '{}/submissions?filter%5Bprogram%5D={}&filter%5Btarget%5D={}&page%5Blimit%5D=100&page%5Boffset%5D=0&include=monetary_rewards,target&filter%5Bduplicate%5D=false&sort=submitted-desc'.format(self.bugcrowd_api_url, program, quote(target))
+        else:
+            next = '{}/submissions?filter%5Bprogram%5D={}&page%5Blimit%5D=100&page%5Boffset%5D=0&include=monetary_rewards,target&filter%5Bduplicate%5D=false&sort=submitted-desc'.format(self.bugcrowd_api_url, program)
```
damiencarol, line 35: please use [urllib.parse.urlencode](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode)

dojo/tools/bugcrowd_api/api_client.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def get_headers(self):
+        headers = {
+            'Accept': 'application/vnd.bugcrowd+json',
+            'Authorization': 'Token {}'.format(self.api_token),
+            'User-Agent': 'DefectDojo',
+            'Bugcrowd-Version': '2021-10-28'
+        }
+
+        return headers
```
damiencarol, line 104: Make it a constant. better

dojo/tools/bugcrowd_api/importer.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class BugcrowdApiImporter(object):
+    """
+    Import from Bugcrowd API
+    """
+
+    def get_findings(self, test):
+        client, config = self.prepare_client(test)
+        logger.info("Fetching submissions program " + str(config.service_key_1) + " and target " + str(config.service_key_2))
```
damiencarol, line 16:
```suggestion
        logger.debug("Fetching submissions program " + str(config.service_key_1) + " and target " + str(config.service_key_2))
```

dojo/tools/bugcrowd_api/parser.py
```diff
@@ -0,0 +1,193 @@
...
+            bugcrowd_duplicate = entry["attributes"]["duplicate"]
+            bugcrowd_severity = entry["attributes"]["severity"]
+
+            title = entry["attributes"]["title"]
+            if not pattern_title_authorized.match(title):
+                char_to_replace = {':': ' ','"': ' ','@': 'at'}
+                for key, value in char_to_replace.items():
+                    title = title.replace(key, value)
+
+            date = self.get_created_date(entry["attributes"]["submitted_at"])
```
damiencarol, line 64: Better to use `dateutil.parse` module

dojo/tools/bugcrowd_api/api_client.py
Gby56: Ah ! I copied over from the edgescan importer I think, so I didn't really know these other functions

damiencarol (changes requested):
There is a few changes needed, also unit tests are missing.

dojo/tools/bugcrowd_api/api_client.py
damiencarol: forget my comment , this part is good.

dojo/tools/bugcrowd_api/api_client.py
Gby56: If I need the self.api_token I'm not sure how I can make it a constant of the class ?

damiencarol: at least make it a constant of the module.
And in the code, just add 'Auth' header.
Something is not good here. I think you should switch to use sessions and set the auth in the session.

Gby56: What difference does it make between using session vs just a constant header ? Is there caching in the client or something ?

damiencarol: yes. and the code is clear and structured. auth will be set only one time when building the session.
I can show you. it will also mean fewer lines, fewer bugs, less maintenance... etc

Gby56: Done ! I definitely saw that a lot of clients don't use session yet lol

Gby56:
@Maffooch all good :) I copied tests from Edgescan

damiencarol:
@Gby56 

1. when I said use urlencode method instead, I was refering to something like this:

```Python
import urllib.request
import urllib.parse
params = urllib.parse.urlencode({'spam': 1, 'eggs': 2, 'bacon': 0})
url = "http://www.musi-cal.com/cgi-bin/query?%s" % params
with urllib.request.urlopen(url) as f:
    print(f.read().decode('utf-8'))
```
Which is easier to read and less bug prone (instead of pushing the string encoded directly in the code).

2. We should convert this method `get_findings(self, program, target)` to be a generator function.
If I understand well the data returned by GitLab API, the data are paginated. 
Which means that the function could return multiple time data. Perfect use case for a generator function.

Gby56:
> @Gby56
> 
>     1. when I said use urlencode method instead, I was refering to something like this:
> 
> 
> ```python
> import urllib.request
> import urllib.parse
> params = urllib.parse.urlencode({'spam': 1, 'eggs': 2, 'bacon': 0})
> url = "http://www.musi-cal.com/cgi-bin/query?%s" % params
> with urllib.request.urlopen(url) as f:
>     print(f.read().decode('utf-8'))
> ```
> 
> Which is easier to read and less bug prone (instead of pushing the string encoded directly in the code).
> 
>     2. We should convert this method `get_findings(self, program, target)` to be a generator function.
>        If I understand well the data returned by GitLab API, the data are paginated.
>        Which means that the function could return multiple time data. Perfect use case for a generator function.

Done ! The bugcrowd api uses an annoying query format for JSONAPI, with brackets everywhere, this definitely looks better now

Gby56:
Implemented a generator pattern for the api client function, no gain in performance since it stays sequential, removes the "accumulator" logic from the function that's all
A better pattern would be to build out the work pool initially, looking at a given total number of elements if it's returned by the API, and then split the work across workers.
Otherwise, when the total number of pages is unknown, it would be possible to split between 2 workers (or more) by just taking odd/even page numbers for example ?

Gby56:
Added multithreading worker pool to fetch all the pages, I fetch a first page to get the total, then pre-compute the urls, then dispatch
```
Execution time: 17.77815866470337 seconds multi threaded
Execution time: 19.320730924606323 seconds generator sequentially
```

dojo/tools/bugcrowd_api/parser.py
```diff
@@ -118,27 +129,31 @@ def convert_endpoint(self, url):
         if result:
             return Endpoint.from_uri(result.group(0))
         else:
-            return Endpoint.from_uri('')
+            return Endpoint.from_uri("")
```
damiencarol, line 130: Not sure this is ok.
@kiblik any feedback?

dojo/tools/bugcrowd_api/parser.py
```diff
@@ -118,27 +129,31 @@ def convert_endpoint(self, url):
...
-            "informational",      # The finding is not yet verified by the pentest team
+            "informational",  # The finding is not yet verified by the pentest team
         ]
 
         if entry["attributes"]["state"] in allowed_states:
             return True
         else:
-            print(entry["attributes"]["state"] + 'NOT IN ALLOWED STATES !')
+            print(
+                entry["attributes"]["state"]
```
damiencarol: We need to remove every `print()`

damiencarol (changes requested):

dojo/tools/bugcrowd_api/parser.py
Gby56: Not my prints 🥲 but seriously though, how do parsers report errors to users ? I wasn't sure how I can raise an exception that bubbles back to the UI

damiencarol: there is a need to add more data back to the importer a feature is missing IMO. for now just raise a `ValueError` or use `LOGGER.debug()`.
Print is a no-go

dojo/tools/bugcrowd_api/parser.py
kiblik: Hi @Gby56,
unfortunately, I have to agree with @damiencarol.

I checked the implementation of the parser but I couldn't find why your lines need to use regex for searching/validating URL and with using of `Endpoint.from_uri(...)`. Plus I'm also not sure what will even `Endpoint.from_uri("")` return (with high probability `ValidationError`).

In my opinion,
```python
            if bug_url != "" and bug_url is not None:
                endpoint = self.convert_endpoint(bug_url)
                finding.unsaved_endpoints = [endpoint]
```
could be replaced by
```python
            if bug_url:
                endpoint = Endpoint.from_uri(bug_url)
                finding.unsaved_endpoints = [endpoint]
```
and `convert_endpoint(...)` could be removed.
 
In your JSONs, I wasn't able to see any `bug_url` which needs this kind of "special care" :)

By the way, as I remember correctly, the URL should not be mentioned in the description since the `Endpoint` and `Endpoint_Status` work quite well.

Gby56: Hello ! Yes the regex seems a bit odd but I kept getting a bug on my own private submissions, which I can't share of course, but let me try to find an example and maybe the fix could be implemented in the endpoint from_uri method instead, it was a weird one

Gby56: And I agree I'll remove the empty endpoint for sure, that was creating a useless endpoint I think

kiblik: If URL is not valid, `Endpoint.from_uri(...)` will raise `ValidationError` so 
```python
            if bug_url:
                try:
                    endpoint = Endpoint.from_uri(bug_url)
                    finding.unsaved_endpoints = [endpoint]
```
should be good.

damiencarol: 100% agree. good feedback @kiblik

Gby56: The error I keep getting with the `Endpoint.from_uri` is 
```An exception error occurred during the report import:invalid scheme: ' http'. Only alphanumeric, "+", "-", and "." allowed. Did you meant to call URL.from_text()?```

Comment (Gby56) [reply dojo/tools/bugcrowd_api/parser.py:130]: I'll try to ignore the exception but how can I return a message to the user in the UI, to tell them that a URL was not parsed, but not fail the whole parser ?

Comment (kiblik) [reply dojo/tools/bugcrowd_api/parser.py:130]: It looks like there problem with space at the beginning of the URL. I see you used `strip()` in your implementation. Can you try to call `Endpoint.from_uri(bug_url.strip())`. It may fix the whole issue.

Comment (Gby56) [reply dojo/tools/bugcrowd_api/parser.py:141]: I'll leave the logger because the parser is really going to get tested with a lot of weird inputs from submissions, so I think it's better not to fully break the import for 1 single submission

Comment (kiblik) [reply dojo/tools/bugcrowd_api/parser.py:141]: Even if you add `strip()`, feel free to use `try ... except ...`. It is a good approach.

But setting `finding.unsaved_endpoints = []` is not necessary. It is `[]` by default.

Comment (Gby56) [reply dojo/tools/bugcrowd_api/parser.py:141]: I keep seeing a weird error now though, Mysql doesn't seem too happy about a protocol ?
```
uwsgi_1         |   File "/app/./dojo/engagement/views.py", line 626, in import_scan_results
uwsgi_1         |     test, finding_count, closed_finding_count, _ = importer.import_scan(scan, scan_type, engagement, user, environment, active=active, verified=verified, tags=tags,
uwsgi_1         |   File "/app/./dojo/importers/importer/importer.py", line 324, in import_scan
uwsgi_1         |     new_findings = self.process_parsed_findings(test, parsed_findings, scan_type, user, active,
uwsgi_1         |   File "/app/./dojo/decorators.py", line 47, in __wrapper__
uwsgi_1         |     return func(*args, **kwargs)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/celery/local.py", line 188, in __call__
uwsgi_1         |     return self._get_current_object()(*a, **kw)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/celery/app/task.py", line 392, in __call__
uwsgi_1         |     return self.run(*args, **kwargs)
uwsgi_1         |   File "/app/./dojo/importers/importer/importer.py", line 136, in process_parsed_findings
uwsgi_1         |     importer_utils.add_endpoints_to_unsaved_finding(item, test, item.unsaved_endpoints, sync=True)
uwsgi_1         |   File "/app/./dojo/decorators.py", line 47, in __wrapper__
uwsgi_1         |     return func(*args, **kwargs)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/celery/local.py", line 188, in __call__
uwsgi_1         |     return self._get_current_object()(*a, **kw)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/celery/app/task.py", line 392, in __call__
uwsgi_1         |     return self.run(*args, **kwargs)
uwsgi_1         |   File "/app/./dojo/importers/utils.py", line 135, in add_endpoints_to_unsaved_finding
uwsgi_1         |     ep, created = endpoint_get_or_create(
uwsgi_1         |   File "/app/./dojo/endpoint/utils.py", line 81, in endpoint_get_or_create
uwsgi_1         |     return Endpoint.objects.get_or_create(**kwargs)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/models/manager.py", line 85, in manager_method
uwsgi_1         |     return getattr(self.get_queryset(), name)(*args, **kwargs)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/tagulous/models/tagged.py", line 242, in get_or_create
uwsgi_1         |     return super(TaggedQuerySet, self).get_or_create(defaults, **safe_fields)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/models/query.py", line 588, in get_or_create
uwsgi_1         |     return self.create(**params), True
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/tagulous/models/tagged.py", line 223, in create
uwsgi_1         |     obj = super(TaggedQuerySet, self).create(**safe_fields)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/models/query.py", line 453, in create
uwsgi_1         |     obj.save(force_insert=True, using=self.db)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/models/base.py", line 739, in save
uwsgi_1         |     self.save_base(using=using, force_insert=force_insert,
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/models/base.py", line 776, in save_base
uwsgi_1         |     updated = self._save_table(
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/models/base.py", line 881, in _save_table
uwsgi_1         |     results = self._do_insert(cls._base_manager, using, fields, returning_fields, raw)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/models/base.py", line 919, in _do_insert
uwsgi_1         |     return manager._insert(
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/models/manager.py", line 85, in manager_method
uwsgi_1         |     return getattr(self.get_queryset(), name)(*args, **kwargs)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/models/query.py", line 1270, in _insert
uwsgi_1         |     return query.get_compiler(using=using).execute_sql(returning_fields)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/models/sql/compiler.py", line 1416, in execute_sql
uwsgi_1         |     cursor.execute(sql, params)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 66, in execute
uwsgi_1         |     return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 75, in _execute_with_wrappers
uwsgi_1         |     return executor(sql, params, many, context)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
uwsgi_1         |     return self.cursor.execute(sql, params)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/utils.py", line 90, in __exit__
uwsgi_1         |     raise dj_exc_value.with_traceback(traceback) from exc_value
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
uwsgi_1         |     return self.cursor.execute(sql, params)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/db/backends/mysql/base.py", line 73, in execute
uwsgi_1         |     return self.cursor.execute(query, args)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/MySQLdb/cursors.py", line 206, in execute
uwsgi_1         |     res = self._query(query)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/MySQLdb/cursors.py", line 319, in _query
uwsgi_1         |     db.query(q)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/MySQLdb/connections.py", line 254, in query
uwsgi_1         |     _mysql.connection.query(self, query)
uwsgi_1         | django.db.utils.DataError: (1406, "Data too long for column 'protocol' at row 1")
```

Comment (Gby56) [reply dojo/tools/bugcrowd_api/parser.py:140]: I think the endpoint model in defectdojo is too restrictive

Comment (kiblik) [reply dojo/tools/bugcrowd_api/parser.py:140]: - `Endpoint` doesn't accept only well-known protocols. It accepts also unknown protocols.
- `Endpoint` uses well-known protocols but only in `__str__` function - it can decide to hide the port (e.g. if the protocol is HTTP and the port is 80 or if the protocol is HTTPS and the port is 443)
- I fully support extending the length of the protocol in reasonable cases (there are 20 characters right now). Can tell us what is your protocol?
- We tried to avoid so obscure protocol names - please check https://github.com/DefectDojo/django-DefectDojo/pull/4696#discussion_r655532107
- `Endpoint` tried to reflect RFC definition for URI/URL. 
- We already implemented less strict validation even if it doesn't follow RFC. E.g. https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/test_endpoint_model.py#L68-L74. More info here https://github.com/DefectDojo/django-DefectDojo/pull/4604. But we would like to avoid very unusual formats only because some other developer in some other project doesn't follow RFCs.

Comment (Gby56) [reply dojo/tools/bugcrowd_api/parser.py:140]: Gotcha ! That's good to know then, I guess it's mostly just the input that I send into from_uri that is creating issues, I need to isolate the URI/URL first and then try to create an Endpoint from that
I have tried to use that regex for validation of a URI no matter the protocol etc... but it's a lot of trouble for no reason, just to isolate a URI in a random string https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java

Comment (kiblik) [reply dojo/tools/bugcrowd_api/parser.py:140]: I understand it can be hard. The question is: _Do we need to solve this issue in DD?_
If some software/API is providing a field with the name _bug_url_, I suppose it should be a valid URL every time.
Is it possible to open an issue/question in _Bugcrowd_: _Why are they sending non-URL information in the field called `bug_url`?_

Comment (Gby56) [reply dojo/tools/bugcrowd_api/parser.py:140]: I agree, it's just that I want to make a powerful extractor for this integration, because of the human-nature of it. It won't produce a good amount of data otherwise and I'd think it's a shame :/ I'm even considering to try and extract URLs from the submission title, because I see in our internal findings that most researchers don't bother submitting properly...

Comment (Gby56) [reply dojo/tools/bugcrowd_api/parser.py]: You're right, I give up :P I think some submissions have disgusting inputs... I won't parse the description to find URLs either, but I'm still getting errors from the db column... `django.db.utils.DataError: (1406, "Data too long for column 'protocol' at row 1")`

Comment (Gby56) [reply dojo/tools/bugcrowd_api/parser.py]: I just saw, we have people submitting bug_urls like a csv too, like `url1, url2, url3`

Comment (kiblik) [reply dojo/tools/bugcrowd_api/parser.py]: As I remember correctly, the length of the field is not checked in `Endpoint.from_uri(...)` but in `Endpoint.save()`. So you can try to call:
```python
endpoint = Endpoint.from_uri(bug_url)
endpoint.protocol = endpoint.protocol[:20]
```

kiblik: But keep in mind, it will cut everything over 20 so 
```
abcdeabcdeabcdeabcdex://host/
```
and 
```
abcdeabcdeabcdeabcdey://host/
```
will be merged to `abcdeabcdeabcdeabcde://host/`. Not so sure it is the best solution. I prefer to solve it in a different way.

Gby56: Yeah, I dumped the protocols I find when parsing all the submissions we have.... and here's a clear bug :P 
The quick fix by cutting to 20 chars works, but looks at this:
Someone submitted `aaa-bbbbbb.cccccc.dddddd.eeeeee.com:443` as a bug_url
The protocol is weirdly parsed, and basically, the path becomes 443, and the protocol becomes `aaaa-bbbbb.ccccc`

Gby56: that's a bug on https://pypi.org/project/hyperlink/ https://github.com/python-hyper/hyperlink/issues/170

kiblik: Oooo, I see. I highly recommend to use this workaround https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/generic/parser.py#L61-L66

kiblik: `aa.bb.cc:443` is not a valid URI/URL. But this ^ workaround works well.

Gby56: Alright I did that, it kinda works, the hosts are still broken because some just have "API Call" in the hostname, but it's fine I guess for now...

Gby56:
Code is functional, could be improved a bit with URI parsing and handling more use cases (bug urls in Bugcrowd are not always URL/URIs, sometimes just hosts, sometimes just paths without the host which is almost unusable)

Gby56:
@damiencarol @Maffooch @kiblik if I can get a last review, it should be all good 🤔

dojo/tools/bugcrowd_api/api_client.py
```diff
@@ -0,0 +1,140 @@
...
+                )
+                return f'With {total_subs} submissions, you have access to the "{ program_names }" programs, \
+                    you can use these as Service key 1 for filtering submissions \
+                        You also have targets "{ target_names }" that can be used in Service key 2'
+        else:
+            raise Exception(
+                "Connection failed (error: {} - {})".format(
+                    response_subs.status_code, response_subs.content.decode("utf-8")
+                )
+            )
```
damiencarol, line 122: use `response.raise_for_status()` at line 87 and 92 instead.

dojo/tools/bugcrowd_api/api_client.py
```diff
@@ -0,0 +1,140 @@
...
+
+                # Otherwise, keep updating next link
+                next = "{}{}".format(self.bugcrowd_api_url, data["links"]["next"])
+            else:
+                next = "over"
+                raise Exception(
+                    "Unable to get bugcrowd submissions due to {} - {}".format(
+                        response.status_code, response.content.decode("utf-8")
+                    )
+                )
```
damiencarol, line 80: please, as shared in IRL review, use `response.raise_for_status()` instead of shadowing the root exception.

dojo/tools/bugcrowd_api/api_client.py
```diff
@@ -0,0 +1,140 @@
...
+        submission_number = len(submissions)
+        return f'You have access to "{submission_number}" submissions (no duplicates)\
+            in Bugcrowd in the Program code "{api_scan_configuration.service_key_1}" \
+            and Target "{api_scan_configuration.service_key_2}" (leave service key 2 empty to get all submissions in program)'
+
+    def get_headers(self):
+
+        self.default_headers["Authorization"] = "Token {}".format(self.api_token)
+
+        return self.default_headers
```
damiencarol, line 140: This should be in the session object.

damiencarol (changes requested):

dojo/tools/bugcrowd_api/api_client.py
Gby56: ah yes, it's not in use so I'm deleting that

dojo/tools/bugcrowd_api/api_client.py
Gby56: Done

dojo/tools/bugcrowd_api/api_client.py
Gby56: Done

damiencarol (approved):

Gby56:
Not sure why one test failed, is that normal ?

kiblik:
@Gby56, sorry for the delay. I wrote my comments.
I'm a little bit confused that some of them haven't been identified by Flake8

coheigea:
@kiblik We'd like to see this one merged for the next release if possible - are there outstanding issues?

kiblik (changes requested):

kiblik:
Ouuuu, sorry @coheigea & @Gby56. 6 days ago, I wrote my comments but I forgot to click "Submit Review" and they haven't been visible to everybody.

dojo/tools/bugcrowd_api/parser.py
```diff
@@ -0,0 +1,275 @@
...
+                duplicate=bugcrowd_duplicate,
+                out_of_scope=self.is_out_of_scope(bugcrowd_state),
+                risk_accepted=self.is_risk_accepted(bugcrowd_state),
+                is_mitigated=self.is_mitigated(bugcrowd_state),
+                static_finding=False,
+                dynamic_finding=True,
+                unique_id_from_tool=unique_id_from_tool,
+            )
+
+            if bug_endpoint:
```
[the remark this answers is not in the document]
coheigea: Good point - I did find this on error on testing: uwsgi_1         | UnboundLocalError: local variable 'bug_endpoint' referenced before assignment

dojo/tools/bugcrowd_api/parser.py
```diff
@@ -0,0 +1,275 @@
...
+        """Get the date of when a finding was created"""
+        return self.convert_log_timestamp(date)
+
+    def get_latest_update_date(self, log):
+        """Get the date of the last time a finding was updated"""
+        last_index = len(log) - 1
+        entry = log[last_index]
+        return self.convert_log_timestamp(entry["timestamp"])
+
+    def convert_endpoint(self, url):
```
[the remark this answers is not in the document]
Gby56: Done !

dojo/tools/bugcrowd_api/api_client.py
```diff
@@ -0,0 +1,125 @@
...
+
+    bugcrowd_api_url = "https://api.bugcrowd.com"
+    default_headers = {
+        "Accept": "application/vnd.bugcrowd+json",
+        "User-Agent": "DefectDojo",
+        "Bugcrowd-Version": "2021-10-28",
+    }
+
+    def __init__(self, tool_config):
+        tool_type, _ = Tool_Type.objects.get_or_create(name="Bugcrowd API")
```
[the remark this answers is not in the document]
Gby56: removed, I'll just call the function

dojo/tools/bugcrowd_api/api_client.py
```diff
@@ -0,0 +1,125 @@
...
+            total_subs = str(data_subs["total_hits"])
+
+            progs = list(filter(lambda prog: prog["type"] == "program", data))
+            program_names = ", ".join(
+                list(map(lambda p: p["attributes"]["code"], progs))
+            )
+            # Request targets to validate the org token
+            response_targets = self.session.get(
+                url="{}/targets".format(self.bugcrowd_api_url)
+            )
```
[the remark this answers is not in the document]
Gby56: added

Gby56:
@kiblik Hello ! I have implemented all the suggestions, and improved the endpoint issue. I check before saving them that they are not broken, and completely skip if they are. The reason behind that is because of Bugcrowd's API/web ui, lots of  researchers submit weird stuff in that field and I'd prefer to have an actual clean endpoint, than 20 broken ones.

dojo/tools/bugcrowd_api/api_client.py
```diff
@@ -0,0 +1,132 @@
...
+                    filter(lambda prog: prog["type"] == "target", data_targets)
+                )
+                target_names = ", ".join(
+                    list(map(lambda p: p["attributes"]["name"], targets))
+                )
+                return f'With {total_subs} submissions, you have access to the "{ program_names }" programs, \
+                    you can use these as Service key 1 for filtering submissions \
+                        You also have targets "{ target_names }" that can be used in Service key 2'
+            else:
+                raise Exception("Bugcrowd did not return a valid targets response")
```
kiblik, line 116: It is better to give some useful information to the user about what is wrong (error code/message). 
Like here https://github.com/DefectDojo/django-DefectDojo/blob/dc20ee1bf9f99953b50c610bd9840053e8fd12f1/dojo/tools/sonarqube_api/api_client.py#L69-L71

dojo/tools/bugcrowd_api/api_client.py
```diff
@@ -0,0 +1,132 @@
...
+        else:
+            raise Exception(
```
kiblik, line 118: More useful message here as well

dojo/tools/bugcrowd_api/parser.py
```diff
@@ -0,0 +1,275 @@
...
+                risk_accepted=self.is_risk_accepted(bugcrowd_state),
+                is_mitigated=self.is_mitigated(bugcrowd_state),
+                static_finding=False,
+                dynamic_finding=True,
+                unique_id_from_tool=unique_id_from_tool,
+            )
+
+            if bug_endpoint:
+                try:
+                    finding.unsaved_endpoints = [bug_endpoint]
```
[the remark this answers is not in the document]
kiblik: This also has a little bit of misleading error message. Because if there will be an `Exception` it will be because of creating list and/or setting variable. There will be no problem with _the parsing_ of an endpoint.

dojo/tools/bugcrowd_api/parser.py
```diff
@@ -0,0 +1,227 @@
...
+                bug_endpoint.clean()
+                try:
+                    finding.unsaved_endpoints = [bug_endpoint]
+                except Exception as e:
+                    logger.error(
+                        "{} bug url from bugcrowd failed to parse to endpoint, error= {}".format(
+                            str(bug_endpoint), e
+                        )
+                    )
+            except:
```
kiblik, line 138: If there is any issue in format, `Endpoint.clean()` it should raise only `ValidationError`.
https://github.com/DefectDojo/django-DefectDojo/blob/dc20ee1bf9f99953b50c610bd9840053e8fd12f1/dojo/models.py#L1426-L1500
It is not best practice to use a bare `except:`
> When catching exceptions, mention specific exceptions whenever possible instead of using a bare `except:` clause

Source: https://peps.python.org/pep-0008/#programming-recommendations

https://youtu.be/hgI0p1zf31k?t=111

dojo/tools/bugcrowd_api/parser.py
```diff
@@ -0,0 +1,227 @@
...
+
+            if not pattern_title_authorized.match(title):
+                char_to_replace = {":": " ", '"': " ", "@": "at"}
+                for key, value in char_to_replace.items():
+                    title = title.replace(key, value)
+
+            date = dateutil.parser.parse(entry["attributes"]["submitted_at"])
+
+            bug_url = ""
+            bug_endpoint = Endpoint()
```
kiblik, line 72: Empty `Endpoint` like `Endpoint()` will raise an exception, when you call `.clean()`
https://github.com/DefectDojo/django-DefectDojo/blob/dc20ee1bf9f99953b50c610bd9840053e8fd12f1/dojo/models.py#L1442-L1449
I suppose, you should set `bug_endpoint`  to `None`

And this part (later in the code):
```python
    try:
        bug_endpoint.clean()
        finding.unsaved_endpoints = [bug_endpoint]
```
should be called after
```python
    if bug_endpoint:
```

kiblik (changes requested):
Hi @Gby56,

I'm not sure that merging `master` to this branch was the best step. I suppose you wanted to merge `dev`.

I'm adding also a couple of comments to your changes.

Gby56:
@kiblik yeah I did the upstream fetch on my fork and it pulled the master, but now that I merged dev into my branch it's not working either, anyways I copied over the exact files from dev and there are no conflicts here, it's weird how the diff is showing

components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "defectdojo",
-  "version": "2.14.0-dev",
+  "version": "2.15.0-dev",
```
Gby56, line 3: why is it screaming about that in the diff, but the actual dev branch in defectdojo is on 2.15.0-dev ?!?!

kiblik: Unfortunately, I'm not able to help you here. I have no idea.
But I would try `git rebase dev`

Gby56:
I think it’s fine though ? It’s the same as the dev branch anyways

dojo/tools/bugcrowd_api/api_client.py
```diff
@@ -0,0 +1,134 @@
...
+                    filter(lambda prog: prog["type"] == "target", data_targets)
+                )
+                target_names = ", ".join(
+                    list(map(lambda p: p["attributes"]["name"], targets))
+                )
+                return f'With {total_subs} submissions, you have access to the "{ program_names }" programs, \
+                    you can use these as Service key 1 for filtering submissions \
+                        You also have targets "{ target_names }" that can be used in Service key 2'
+            else:
+                raise Exception(
```
kiblik, line 116: When[ I was talking ](https://github.com/DefectDojo/django-DefectDojo/pull/6621#discussion_r969682196)about

> It is better to give some useful information to the user about what is wrong (error code/message). 
> Like here https://github.com/DefectDojo/django-DefectDojo/blob/dc20ee1bf9f99953b50c610bd9840053e8fd12f1/dojo/tools/sonarqube_api/api_client.py#L69-L71

I meant adding information from the HTTP response. The user does not need to know only what does happen (_... test not successful ..._) but also why it does happen (_unable to do DNS resolution_ (my DNS service is down?), _unable to open TCP connection_ (I need to add firewall rule?), _HTTP 401_ (my API key expired?), _HTTP 403/404_ (mistake in URL?), ...).

dojo/tools/bugcrowd_api/api_client.py
kiblik: Same here

kiblik (changes requested):

dojo/tools/bugcrowd_api/api_client.py
Gby56: Done ! I have added the response.text to the exception message, even though I'm not able to trigger that bug to validate things, but it should provide more than enough info (also the raise for status will give some info earlier anyways, 401 will trigger that)

Gby56:
the init file probably didn't get linted with black, the string delimiters were just different

Gby56:
@kiblik tell me if everything is ok for you, I have added the feedback from the http responses in case of a failure as requested

kiblik:
Yes @Gby56, it looks good to me. Thank you for the changes and your effort.
Unfortunately, I do not have "approval" or "merge" rights here, so you will need to wait or ask somebody else.

Maffooch (approved):

PR #6650, opened 2022-08-03
Fix a bug that caused re-imports to drop service name of findings

SStorm:
If findings had a service name set, the re-importer would override it with an empty one, unless it is specified. This behaviour seems incorrect. 

This matches behaviour of the importer.


damiencarol (approved):
good catch!

Maffooch (approved):

PR #6660, opened 2022-08-08
Fix shebang on shell scripts

gnustavo:
The first line of many shell scripts are missing the "!" character,
which make them not to be taken as proper shebangs.


dsever:
Good catch!

dsever (approved):
Approved

damiencarol (approved):

PR #6675, opened 2022-08-09
Parse the report timestamp for CycloneDX json

coheigea:
For CycloneDX XML, it sets the finding date as the date of the report timestamp, however it doesn't do this for JSON. This PR is to make it consistent. In addition the PR fixes two issues:

1. The "jake.json" testcase for CycloneDX is not actually run, due to a previous change as to how vulnerability IDs are parsed.
2. The dc-unittest.sh fails on my machine (mac) due to a recent shebang change

### DIFF
--- a/dc-unittest.sh
+++ b/dc-unittest.sh
@@ -1,4 +1,4 @@
-#!/bin/env bash
+#!/usr/bin/env bash
 
 unset PROFILE
 unset TEST_CASE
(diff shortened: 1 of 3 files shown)


damiencarol (approved):

damiencarol:
@coheigea awesome patch as always.

Maffooch (approved):

PR #6679, opened 2022-08-09
feat: add parser and importer for BlackDuck API

damiencarol:
# Work done

- [x] add importer to load data directly from API (config, doc etc...)
- [x] add parser
- [x] add real obfuscated data and unit tests

This parser is different from other BlackDuck parsers. It uses directly the API and rely on Python package https://pypi.org/project/blackduck/ maintained by the editor of BlackDuck (Synopsys)


mtesauro (approved):
Approved

Maffooch (approved):

PR #6687, opened 2022-08-10
feat: add SARIF basic de-duplication data

### LINKED ISSUE #6598: Deduplication for SARIF import
DefectDojo's SARIF parser does not take into account values provided in the [partialFingerprints property](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317655) for deduplication. It would be nice if it did. The legacy deduplication algorithm used by the SARIF parser includes the line number of the finding, making it sensitive to unrelated code changes.

As a first cut, it would be helpful if the SARIF parser hashed all of the `partialFingerprints` into the finding's `finding.unique_id_from_tool` value and switch to using the `DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE` algorithm (falling back on the legacy fields if no fingerprints are provided).

In a more sophisticated implementation, it would be helpful to be able to have the caller of the import API to specify which fingerprints in the SARIF file they would like to include in the hash, as we have seen that the quality of fingerprints varies within a single tool. For example, one tool provides two fingerprint for every result, one of which is sensitive to line numbers.

Some tools, contrary to the SARIF specification, put their fingerprinting information in the [fingerprints property](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317654) instead of the `partialFingerprints` one. It would be helpful if the SARIF parser could handle this case, too, but it isn't clear what the general solution for this might be.

The SARIF spec gives some guidelines for how to make use of fingerprinting information generated by tools in [Appendix B. (Normative) Use of fingerprints by result management systems](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317932).

damiencarol:
Add data in `unique_id_from_tool` based on [fingerprints](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#AppendixFingerprints) in the SARIF reports.
Also this PR enable de-duplication based on this data or fallback to the old hash method.

I also added unit tests for previous reports.

Fix #6598 
Reference: https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#AppendixFingerprints

### DIFF
--- a/dojo/tools/sarif/parser.py
+++ b/dojo/tools/sarif/parser.py
@@ -332,4 +332,45 @@ def get_item(result, rules, artifacts, run_date):
     if run_date:
         finding.date = run_date
 
+    # manage fingerprints
+    # fingerprinting in SARIF is more complete than in current implementation
+    # SARIF standard make it possible to have multiple version in the same report
+    # for now we just take the first one and keep the format to be able to compare it
+    if result.get("fingerprints"):
+        hashes = get_fingerprints_hashes(result["fingerprints"])
+        first_item = next(iter(hashes.items()))
+        finding.unique_id_from_tool = first_item[1]['value']
+    elif result.get("partialFingerprints"):
+        # for this one we keep an order to have id that could be compared
+        hashes = get_fingerprints_hashes(result["partialFingerprints"])
+        sorted_hashes = sorted(hashes.keys())
+        finding.unique_id_from_tool = "|".join([f'{key}:{hashes[key]["value"]}' for key in sorted_hashes])
     return finding
+
+
+def get_fingerprints_hashes(values):
+    """
+    Method that generate a `unique_id_from_tool` data from the `fingerprints` attribute.
+     - for now, we take the value of the last version of the first hash method.
+    """
+    fingerprints = dict()
+    for key in values:
+        if "/" in key:
+            key_method = key.split("/")[-2]
+            key_method_version = int(key.split("/")[-1].replace("v", ""))
+        else:
+            key_method = key
+            key_method_version = 0
+        value = values[key]
+        if fingerprints.get(key_method):
+            if fingerprints[key_method]["version"] < key_method_version:
+                fingerprints[key_method] = {
+                    "version": key_method_version,
+                    "value": value,
+                }
+        else:
+            fingerprints[key_method] = {
+                "version": key_method_version,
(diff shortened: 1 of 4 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

dojo/tools/sarif/parser.py
```diff
@@ -332,4 +332,45 @@ def get_item(result, rules, artifacts, run_date):
...
+def get_fingerprints_hashes(values):
+    """
+    Method that generate a `unique_id_from_tool` data from the `fingerprints` attribute.
+     - for now, we take the value of the last version of the first hash method.
+    """
+    fingerprints = dict()
+    for key in values:
+        if "/" in key:
+            key_method = key.split("/")[-2]
+            key_method_version = int(key.split("/")[-1].replace("v", ""))
```
kdyck-cb, line 360: The SARIF spec states that the `fingerprints` and `partialFingerprints` property keys are to be [versioned hierarchical strings](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317441). 

The spec is a little unclear on how many levels are allowed in a hiearchical string. The formal syntax shows only a single slash character is allowed:

```
hierarchical string = component, { "/", component };
```

But the text description seems to suggest that many levels are allowed:

> The first component represents the top level of the hierarchy, the second component represents the second level, and so on.

The spec also states that the version component is optional:

> NOTE 1: A versioned hierarchical string does not need to include a version component. The syntax permits but does not require it.

This code appears to be assuming that:
1. The last component is a version
2. There is only a single level in the hierarchy, and choosing only the final component.

damiencarol: Thank you for you comments, very interesting.
1. the spec states that ` are said to be “versioned.” This means that if the last component of the string is of the form: version component = "v", non negative integer;  ` so yes, the last component should be the version. I can add more tests to check that
2. for now the current code is compatible with all the reports we have in unit tests, I will continue to check with new version of the tools. it's possible that some tools didn't implemented well and will change a little bit the data. We will see

PR #6692, opened 2022-08-11
Add parser and importer for Vulners.com reports API 

vankyver:
Adding vulners.com reports parser.


unittests/tools/test_vulners_parser.py
```diff
@@ -0,0 +1,12 @@
...
+from dojo.models import Test
+
+
+class TestVulnersParser(DojoTestCase):
+
+    def test_parse_many_findings(self):
+        testfile = open("unittests/scans/vulners/vulns_list.json")
+        parser = VulnersParser()
+        findings = parser.get_findings(testfile, Test())
+        self.assertEqual(75, len(findings))
```
damiencarol, line 12: Please add more checks on findings

dojo/tools/vulners/parser.py
```diff
@@ -0,0 +1,99 @@
...
+                severity=severity,
+                impact=severity,
+                description=vuln.get("description", title),
+                mitigation=component.get("cumulativeFix"),
+                static_finding=False,  # by definition
+                dynamic_finding=True,  # by definition
+                # false_p=False,
+                # duplicate=False,
+                out_of_scope=False,
+                vuln_id_from_tool=id,
```
damiencarol, line 74: it seems that it should be better to use `unique_id_from_tool` for this data.

dojo/tools/vulners/api_client.py
```diff
@@ -0,0 +1,78 @@
+import requests
```
damiencarol, line 1: It seems that you implemented from scratch the client, any reason to not use the Pypi official package https://pypi.org/project/vulners/#description ?

damiencarol (changes requested):

dojo/tools/vulners/api_client.py
vankyver: Just to do not include one more library to project

damiencarol: it's better to use this package honestly. somebody will need to maintain this part. it's better to let vulners do it itself (this package is promoted through their doc), also the package is small (in kb)

vankyver: Right now vulners SDK does not support using of another API address instead of https://vulners.com
Few days ago was release of https://github.com/vulnersCom/vulners-proxy, so I can tell DD to use proxy instead of main API
As soon as SDK support proxy, I'll change raw requests to vulners SDK

dojo/tools/vulners/parser.py
vankyver: I kept **vulnerability_id_from_tool** and added vulners.com to **VULNERABILITY_URLS**, because I believe it's exactly public vulnerability ID like [RHSA-2018:2285](https://access.redhat.com/errata/RHSA-2018:2285) or [CESA-2021:0348](https://lists.centos.org/pipermail/centos-announce/2021-February/060805.html) and not just internal identifier

vankyver:
@damiencarol Looks like all required fixes done. Can we finalise review?

dojo/tools/vulners/parser.py
damiencarol: ok. thanks

dojo/tools/vulners/api_client.py
damiencarol: Made a review of the official SDK and it support proxies.

https://github.com/vulnersCom/api/blob/f56dab4ffcdb400da58b680d7feb89cae3550984/vulners/api.py#L97-L105

image

damiencarol:
@vankyver :

Made a review of the official SDK and it support proxies.

https://github.com/vulnersCom/api/blob/f56dab4ffcdb400da58b680d7feb89cae3550984/vulners/api.py#L97-L105

image

vankyver:
> @vankyver :
> 
> Made a review of the official SDK and it support proxies.
> 
> https://github.com/vulnersCom/api/blob/f56dab4ffcdb400da58b680d7feb89cae3550984/vulners/api.py#L97-L105
> 
> image

@damiencarol I've added fixes with including Vulners SDK

dojo/tools/vulners/api_client.py
vankyver: @damiencarol, issues fixed. Will you have a chance to check it out?

Maffooch (approved):

devGregA (approved):

mtesauro:
@damiencarol Wanted to give you a chance to look at this before it gets merged since you had comments above.

devGregA:
@vankyver will merge after the release.

damiencarol:
@vankyver will review this WE. saw a couple of things to slightly modify

dojo/tools/vulners/api_client.py
```diff
@@ -0,0 +1,33 @@
...
+            self.api_key = tool_config.api_key
+            self.vulners_api_url = tool_config.url or self.DEFAULT_URL
+        else:
+            raise Exception('Vulners.com Authentication type {} not supported'.format(tool_config.authentication_type))
+
+    def get_client(self):
+        return vulners.VulnersApi(
+            api_key=self.api_key,
+            server_url=self.vulners_api_url,
+            persistent=False
```
damiencarol, line 24: why this? forcing to false make creating a connection for each call to the API

damiencarol: also `server_url` already have a default value in the lib, why forcing it?

(from the 2.0.4)
```Python
def __init__(
        self, api_key, proxies=None, persistent=True, server_url="https://vulners.com"
    ):
        """
        Create VScanner API object.

        :param api_key:
            string with Vulners API key. You can obtain one from the https://vulners.com
        :param proxies:
            dict for proxy supporting. Example: {"https": "myproxy.com:3128"}
        """
```

Also, regarding our discussion, I was expecting thta you needed the support for proxies.
Something like:
```
    def get_client(self):
        return vulners.VulnersApi(
            api_key=self.api_key,
            proxies=self.proxies,  # ex: {"https": "myproxy.com:3128"}
```

dojo/settings/settings.dist.py
```diff
@@ -1568,6 +1571,7 @@ def saml2_attrib_map_format(dict):
     'PYSEC': 'https://osv.dev/vulnerability/',
     'SNYK': 'https://snyk.io/vuln/',
     'RUSTSEC': 'https://rustsec.org/advisories/',
+    'VNS': 'https://vulners.com/',
```
damiencarol, line 1574:
```suggestion
```
Not needed, see my other comments

dojo/tools/vulners/api_client.py
```diff
@@ -0,0 +1,33 @@
+import vulners
+
+
+class VulnersAPI:
+    """
+    A simple client for the Vulners API
+    """
+
+    DEFAULT_URL = "https://vulners.com"
```
damiencarol, line 9: Not needed, see my other comments.

Maffooch:
@vankyver please see feedback from Damien

manuel-sommer (changes requested):
- Please add also the vulners.com parser to the docs: https://github.com/DefectDojo/django-DefectDojo/blob/master/docs/content/en/integrations/parsers.md

dojo/tools/vulners/api_client.py
damiencarol: taking a look.

manuel-sommer:
As @vankyver did not reply over a month, would it be an option to make another PR including these changes plus the requested changes from @damiencarol and close this PR?

Maffooch:
@manuel-sommer that is cool with me!

damiencarol:
@manuel-sommer , @Maffooch : if @vankyver is ok, I can "finish" the PR.

Maffooch:
@damiencarol either way works! This PR has two approvals so you could change and merge at your timeline

vankyver:
Hi, sorry, hard month, going to come back with changes asap.

dojo/tools/vulners/api_client.py
vankyver: > This is better but the current PR doesn't support proxies. is it still one of your goal?

Proxy in this PR is not exactly a default http proxy, it's name of vulners sub-project https://github.com/vulnersCom/vulners-proxy
so server_url can be the url of vulners-proxy

But proxy in ```vulners.VulnersApi(
            api_key=self.api_key,
            proxies=self.proxies,``` is a normal http-proxy

dojo/settings/settings.dist.py
vankyver: Sorry, didn't find other comments regarding this feature. I used this to highlight links that follows to CVE and other advisories located at vulners.com.

dojo/tools/vulners/api_client.py
vankyver: removed

manuel-sommer (approved):
thank you for adding the parser to the documentation

mtesauro:
@damiencarol Was looking at this and was about to merge since it has 2 approvals but couldn't tell from the thread if there was work you still wanted to do.

So, I'll wait to give you a chance to chime it or address the changes you requested. :+1:

vankyver:
@damiencarol From my point of view, everything is ready, so I hope that now we can complete the pull request.

damiencarol (approved):

damiencarol:
Great work @vankyver

Issue #6569, opened 2022-07-15
Add native AWS SQS support to Django Docker image


jclRatepay:
**Is your feature request related to a problem? Please describe**
In my company we are trying to run as much stuff on AWS and try to make use of as much AWS managed services as possible. So we also tried to bring DefectDojo to run as a AWS Fargate service with the additional requirement to use an existing AWS SQS queue as the queue for the celery worker. During the setup and after a lot of googling I was able to find out how to configure the celery-worker and celery-beat container instances so they will connect with my already existing AWS SQS queue. However, I also encountered the problem, that for being able to connect to my queue some additional Python modules are required. After finding a way how to install those via the container's entrypoint, it turned out that one of the modules - pycurl to be exactly - needs some further packages ("libcurl-devel" and "openssl-devel") installed via apt package manager, which is not possible to achieve as none root user via commands given to the container's entrypoint - which means that the official docker images for DefectDojo currently provided via e.g. the Docker Hub Repository can not be used with AWS SQS.

**Describe the solution you'd like**
I would suggest to simply extend the Django Docker File to install the Python modules required for invoking the AWS API natively and (at least) also install the former mentioned "libcurl-devel" and "openssl-devel" apt packages when building the Docker image.

**Describe alternatives you've considered**
While as mentioned before there are ways to manually post-install the Python modules on container startup, there is no alternative for making the apt packages available. The only alternative would then be to checkout the repo, extend and build the image including the required packages and modules by our own - which of cause would make maintenance and updating the application much more complicated.

**Additional context**
A nice overview of apt-packages and modules required could be found at https://www.yippeecode.com/topics/aws-sqs-message-broker-for-python-celery/

jclRatepay:
The requirement has been implemented via PR #6696

PR #6696, opened 2022-08-12
Add native AWS SQS support to Django Docker image

jclRatepay:
Extended Dockerfile.django and requirements.txt files to additionally install apt packages and python libraries required to support AWS SQS queue connections for Celery as described in Issue https://github.com/DefectDojo/django-DefectDojo/issues/6569. The extended setup has successfully been tested in the scenario described in the former mentioned issue.


damiencarol (approved):

mtesauro (approved):
Approved

PR #6698, opened 2022-08-12
Adding parser for Veracode SCA (SourceClear) JSON/CSV files

coheigea:
This is a new parser for Veracode SCA (SourceClear). It supports both CSV files that can be exported from the UI, as well as the JSON exports available via the API.


unittests/tools/test_veracode_sca_parser.py
```diff
@@ -0,0 +1,72 @@
...
+        testfile = open("unittests/scans/veracode_sca/veracode_sca.json")
+        parser = VeracodeScaParser()
+        findings = parser.get_findings(testfile, Test())
+        self.assertEqual(1, len(findings))
+
+        finding = findings[0]
+        self.assertEqual("High", finding.severity)
+        self.assertTrue(finding.active)
+        self.assertFalse(finding.is_mitigated)
+        self.assertEqual("Apache Calcite Avatica", finding.component_name)
```
damiencarol, line 64:
```suggestion
        self.assertEqual("avatica-core", finding.component_name)
```
in the data it seems we have the data like this `maven:org.apache.calcite.avatica:avatica-core:1.11.0:` , in dojo we have only component name and version not the group. so it's better to set this data that will be aligned with others parsers.

damiencarol (changes requested):
It's a very interesting PR.
But few things bother me:
1. it seems the CSV and JSON have the same data but not in the same format. but the parsers set values differently. for ex in JSON we set unique_id but in CSV we set vuln_id.
2. it seems that the report have Issue and Vuln data but the parser filter to only have Issue when the type is Vuln. so should we just consider data of vuln? this way de-duplication will be easier. (we can use the id of the vulnerability directly for de-duplication)

coheigea:
@damiencarol Thanks for the feedback, I've made all the changes suggested, apart from your second point. The vulnerability data alone isn't enough to parse, for example the library information isn't in the vulnerability section.

damiencarol:
> @damiencarol Thanks for the feedback, I've made all the changes suggested, apart from your second point. The vulnerability data alone isn't enough to parse, for example the library information isn't in the vulnerability section.

My comment was more to align data on both parser (for ex use the vuln id in both parser for unique_id_from_tools attribute) you can still use data from the issue for things like components. But the PR is good enough IMO. let release this one. more PR will come if modifications are needed

damiencarol (approved):

mtesauro (approved):
Approved

PR #6699, opened 2022-08-12
Questionnaire cleanup efforts

Maffooch:
I was looking through the questionnaires and noticed a good amount of ugly. Here is what I've fixed
- Various clean ups in code formatting in the views
- Changed some redirects to make sense (ex. redirect to question list after adding a new question)
- The active bar would display outside of the div on most screen sizes
  - Before: 
  - image
  - After: 
  - image
- Question selection form 
  - Every questions selected was being displayed in the dropdown box and would make the page a ridiculous size
  - Before:
  - image
  - After:
  - image

### DIFF
--- a/dojo/forms.py
+++ b/dojo/forms.py
@@ -3055,6 +3055,19 @@ class AddGeneralQuestionnaireForm(forms.ModelForm):
         model = General_Survey
         exclude = ('num_responses', 'generated')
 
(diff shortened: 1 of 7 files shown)


mtesauro (approved):
Approved

damiencarol (approved):

PR #6701, opened 2022-08-12
Add BleachCSS for Login Banner

Maffooch:
On the release of Bleach 5.0.0, the CSS module was separated into a different package. When upgrading to the Bleach 5.0 in dojo, I had overlooked that the login banner sanitizes CSS and not just HTML. This is the fix


mtesauro (approved):
Approved

damiencarol (approved):

PR #6712, opened 2022-08-15
Update API endpoint for file upload

Maffooch:
Per the discussion around #6564, I had forgotten to update the URL returned by the API to actually download the file from the server. Previously, the response returned the URL directly that is not permission controlled. As of #6564, the ability to download files from the direct location is totally blocked, so downloading the file is not actually possible.

In this PR, the access controlled URL is returned instead

Before:
image

After:
image

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -762,6 +762,21 @@ class EngagementToFilesSerializer(serializers.Serializer):
     engagement_id = serializers.PrimaryKeyRelatedField(queryset=Engagement.objects.all(), many=False, allow_null=True)
     files = FileSerializer(many=True)
 
+    def to_representation(self, data):
+        engagement = data.get('engagement_id')
+        files = data.get('files')
+        new_files = []
+        for file in files:
+            new_files.append({
+                'id': file.id,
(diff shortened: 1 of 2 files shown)


dzmitry-savitski:
I believe this PR will just fix the URL. But it's still not possible to retrieve file content using API. There's no such API endpoint. And the "access_file/***" endpoint won't accept API token.

damiencarol (approved):

mtesauro (approved):
Approved

mtesauro:
@Maffooch I'm good with merging this (as is Damien) - wanted to give you a chance to see the comment above.

If you're good with it, feel free to merge it since it has the necessary 2 approvals.

Maffooch:
Maybe a new PR would be better. The only think left to do would be to create an API endpoint to download a file object directly. It would likely involve 2 API calls to download a file (1 to get the file id from a given object, and the second tom download the file) so the solution is not graceful.

Overall better then having all files totally unrestricted though

dzmitry-savitski:
The only thing I wanted to bring up is the current situation. With this security update, there's no way to download files programmatically anymore. I have to run my scripts in manual mode (passing the session cookie) until the new API endpoint is created.

PR #6713, opened 2022-08-16
[FIX] Tag bugs when adding finding from template

X0x1RG9f:
Tags not taken into account when adding a finding from template


Maffooch:
please add some sort of unit tests. This would be a good spot -> https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/test_apply_finding_template.py

devGregA:
@X0x1RG9f happy to merge with unit tests.

Maffooch (approved):

PR #6714, opened 2022-08-16
Use the CVSS parser to parse the CVSS vector

coheigea:
In a few of the parsers we just set the CVSS vector directly rather than parsing it.

### DIFF
--- a/dojo/tools/acunetix/parser.py
+++ b/dojo/tools/acunetix/parser.py
@@ -4,6 +4,7 @@ import logging
 import dateutil
 import html2text
 import hyperlink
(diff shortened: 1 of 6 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

damiencarol:
@coheigea weird that no unit tests fail. I would expect to to see some change in unit tests.

coheigea:
@damiencarol Why? It should only fail if the CVSS vector being parsed is invalid, right? I guess I could have added some tests to catch this.

PR #6725, opened 2022-08-16
changed type to regular django type to also pick up creation events f…

### LINKED ISSUE #6626: Default group configuration not working with Oauth2.0
**Bug description**
Currently using DefectDojo version 2.12.0 and  trying to add new user  signed up using google Oauth2.0 into a default group with some default permission (automatically without any intervention from our side) which is completely possible as per the docs.
But somehow this is not working
But when I am creating a new user manually  then  default group configuration is working perfectly fine.
Am I missing certain configurations for oauth2.0 users or Default group configuration do not work for user signed up using oauth2.0 ? 

**Steps to reproduce**
1. Create a group and assign some permissions to that group .
2. Go to system settings and configure default group settings
3. After saving those settings try to signin using oauth2.0
4. You will find out that default group is not assigned to the new user.

**Expected behavior**
A default group should have been assigned to the new user
- [`X` ] Kubernetes
Deployment method is kubernetes

**Environment information**
 - DefectDojo version :2.12.0

lme-nca:
changed the type of user_post_save to generic Django user so all users created (ldap, oauth, ui) will trigger this function. This should fix #6626


cody-m-tibco:
@lme-nca please see the failing tests. You may have to base your change from the dev branch to make the helm lint test pass

lme-nca:
Hi @cody-m-tibco I changed the base to dev and fixed a unused import. Let me know if there is anything else i'm supposed to do.

Maffooch (approved):
I think this will be okay based on how the User model is wrapped by the Dojo_User model
@StefanFl what do you think?

StefanFl:
Hard to tell without testing it. The failing unit tests look like it might not work for the normal creation of users anymore. @lme-nca  Have you tested it with creating users via OAuth2 and creating users with the user interface?

lme-nca:
@StefanFl Hi Stefan, you were right it was not being triggered for Dojo users (created through the UI) now. I expected this to work because of inheritance, however I am not that familiar with how Python handles this. It seems the easiest solution is just adding both types as separate annotations. I tested this and it now works for both Dojo as well as Django Users. 

I have not tested it with Oauth, as I don't have a test setup for this, however as this also creates a Django user type i'm confident this would fix the issue.

Maffooch:
I am nervous about double dipping here. For example, if you edit a user in the UI (should be a Dojo_User) would the post save be triggered once for the Dojo_User, and then a second time for the Django user?

lme-nca:
@Maffooch well clearly it wasn't being triggered for the Dojo_User, because the user was not added in the group and also the notification templates were not created, this triggered the test to fail.

however alternatively we could remove the sender restriction all together and do something like this: https://stackoverflow.com/questions/7792287/how-to-use-django-model-inheritance-with-signals i.e. 
`if isinstance(user, Dojo_User) or isinstance(user, User)` 

Personally I don't like this solution.

lme-nca:
@Maffooch @StefanFl The test are passing now, ready for review ?

Maffooch (approved):

lme-nca:
@Maffooch could you add a second reviewer for approval ?

lme-nca:
created a new pull request from the DEV branch: https://github.com/DefectDojo/django-DefectDojo/pull/6918

PR #6740, opened 2022-08-19
Add support for specifying custom JIRA fields

coheigea:
Several of our JIRA projects have required fields, which means that it's not possible to push DefectDojo issues automatically to JIRA. This adds the ability to set a custom JIRA field per product/engagement. The value is a JSON dictionary, which can map field names to values. 

For example, some JIRA projects we use have a "Required in Release" JIRA field that takes an array, and the corresponding value is  { "customfield_12345": [{"name":"7.1.1"}] }


Maffooch:
@coheigea have you seen the JSONfield in Django 3.2 yet? It works and behaves just like a python dict

coheigea:
Thanks @Maffooch , that's a much better idea, PR updated - please review.

coheigea:
@Maffooch Any updates please?

damiencarol (approved):
LGTM

devGregA (approved):

PR #6768, opened 2022-08-30
Add the ability to specify the Epic name and priority in JIRA

coheigea:
When creating a JIRA Epic, it uses the engagement name as the epic name (and JIRA summary), as well as the default priority specified in JIRA for epics. 

This task is to add the ability to configure the Epic Name and Priority in DefectDojo. It will fall back to the existing defaults if not specified.


Maffooch:
Does this have the possibility of creating duplicate epics in jira? For example, If a custom epic name of `epic_name_1` was pushed initially, and then a user edits the epic, and overwrites the the epic name with `epic_name_2`, what would happen?

coheigea:
@Maffooch No it doesn't create duplicate epics, it just updates the name of the existing epic.

devGregA:
@coheigea planning to merge after 2.15 release on tuesday.

Maffooch (approved):
Cannot get that last test to even run anymore...

mtesauro (approved):
Approved

PR #6770, opened 2022-08-30
Corrected deduplication settings for 'Anchore Enterprise Policy Check'

fhoeborn-cb:
There was no deduplication setting explicitly defined for 'Anchore Enterprise Policy Check' which lead to deduplication not working correctly for such reports.


Maffooch (approved):

mtesauro (approved):
Approved

Issue #6761, opened 2022-08-26
Links (Ctrl + K) do not work in Products, Engagement and Findings


manuel-sommer:
**Bug description**
You can't click on inserted links within Products, Engagements and Findings. You are not forwarded to the inserted URL. 

**Steps to reproduce**
1. Create a Product, Engagement or Finding
2. Create a Link (Ctrl + K)
3. Try to click on the Link to get redirected --> You don't get redirected

**Expected behavior**
You should be forwarded to the link

**Deployment method** *(select with an `X`)*
- [X] Docker Compose

**Environment information**
 - DefectDojo version: v. 2.14.0-dev ( 70b8ece9 )  and also in release 2.13.1

**Browsers**
- Firefox Version 104.0
- Google Chrome Version 104.0.5112.101

**Screenshots**
Engagement (you can't click onto the link, thus it is marked orange):
![Screenshot from 2022-08-26 08-46-58](https://user-images.githubusercontent.com/47991713/186839944-04212577-112f-43f9-8ef3-3a8377f2e826.png)
![Screenshot from 2022-08-26 08-45-34](https://user-images.githubusercontent.com/47991713/186839985-6d8a4a70-665e-470c-993d-335112c6a022.png)

Product:(you can't click onto the link, thus it is marked orange)
![Screenshot from 2022-08-26 08-47-58](https://user-images.githubusercontent.com/47991713/186840156-c3283e7e-2343-418b-bafa-cf755dc0e729.png)
![Screenshot from 2022-08-26 08-48-14](https://user-images.githubusercontent.com/47991713/186840313-4e07489c-cc25-4cb8-9250-be8a6136aacd.png)

manuel-sommer:
Also links provided from scanners are not clickable anymore.

damiencarol:
can confirm, it's a bug. don't know what's going on.

manuel-sommer:
It looks like this problem was introduced here: https://github.com/DefectDojo/django-DefectDojo/commit/eab684362098113555378fc6100e1b27cbaa11dd

manuel-sommer:
@damiencarol , the problem is https://github.com/DefectDojo/django-DefectDojo/commit/e1ef1348cea1e0e5fa4f96b6728eb1c3f1c9507d 

If you use DefectDojo 2.13.1 with bleach 4.1.0, the links work fine.

damiencarol:
I think we will revert it for this release

PR #6771, opened 2022-08-30
Revert of (#6475)

damiencarol:


Maffooch:
Why revert bleach? I think the correct reversion is #6167

manuel-sommer:
Hi @damiencarol , do you want to you want to close #6761 with this PR?
Please could you also revert #6701 with this PR?

Maffooch:
Maybe a better solution would be to determine why bleach is sanatizing the markdown fields rather than reverting to older versions

Maffooch:
Fixed by #6848

PR #6775, opened 2022-08-31
Fix bug that caused scan imports to fail

ShayVD:
# Context

If no options were provided to the API scan configuration for edgescan then the scan import would fail when searching for a key in the options object, since no object was available. 

# Changes

Check if options object is available first before checking for keys.


Maffooch (approved):

mtesauro (approved):
Approved.

mtesauro:
Going to go ahead and merge this - not sure why Helm link it failing given this doesn't touch the Helm stuff.

PR #6777, opened 2022-08-31
Corrected deduplication settings for 'Twistlock Image Scan'

fhoeborn-cb:
There was no deduplication setting explicitly defined for 'Twistlock Image Scan' which lead to deduplication not working correctly for such reports.


damiencarol (approved):

mtesauro (approved):
Approved

Issue #6779, opened 2022-08-31
API Risk Acceptance doesn't De-Activate finding


kareem-DA:
**Bug description**
Calling the Risk Acceptance API point doesn't set the finding to inactivate.  

**Steps to reproduce**
1.Call API endpoint "/api/v2/findings/accept_risks/" with the data 
```
[
  {
    "vulnerability_id": "CVE-2022-32792",
    "justification": "test api request",
    "accepted_by": "admin"
  }
]
````
2. Finding risk accepted by this will be marked as risk accepted, but will still be active

**Expected behavior**
Risk Accepted findings should be set as inactive

**Deployment method** *(select with an `X`)*
- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Ubuntu 20.04
 - DefectDojo version2.13.1

**Logs** 
Use `docker-compose logs` (or similar, depending on your deployment method) to get the logs and add the relevant sections here showing the error occurring (if applicable).

**Sample scan files**
If applicable, add sample scan files to help reproduce your problem.

**Screenshots**

**Additional context** (optional)
From the source code: Line 110 from dojo/risk_acceptance/api.py 
```
findings.update(risk_accepted=True)
```

The findings are risk accepted but active is not set to false

tonnyhideyori:
same as this one https://github.com/DefectDojo/django-DefectDojo/issues/6776

coheigea:
This can be closed

PR #6780, opened 2022-08-31
Set finding to inactivate after creation of risk acceptance

kareem-DA:
Finding's don't get de-activated when the risk is accepted from the API.  See bug #6779


Maffooch (approved):

mtesauro (approved):
Approved

PR #6781, opened 2022-08-31
Doc: Update local_settings.py description

kiblik:
To avoid situations like https://owasp.slack.com/archives/C2P5BA8MN/p1661972243567139?thread_ts=1661970538.645869&cid=C2P5BA8MN
> In the documentation it says it's in the dojo/settings directory but it doesn't have any local_settings.py file.


damiencarol (approved):

dsever (approved):
Approved

PR #6786, opened 2022-09-01
fix npm audit JFrog Artifactory import bug

TheRealArlie:
We are using JFrog Artifactroy as NPM repository. Here we encountered problems while parsing the npm audit results in Defect Dojo. We could nail it down to the problem that Artifactroy seems to write `"cwe": null` in some advisories. We could fix the problem by adapting the code as seen in the PR. Please feel free to comment. We would love to see the fix being merged into the DoJo code.


Maffooch (changes requested):
Please add a new unit test that can test a redacted report with cwe=null so that this bug can be tested for in the future

TheRealArlie:
Sure, sry the project is so complex and huge. Could you please guide me to the file where I can add the test?

TheRealArlie:
Never mind, I guess I found the right place. Please have a look at them and decide if this fits your needs.

Maffooch (approved):

mtesauro (approved):
Approved

PR #6794, opened 2022-09-05
helm: defectdojo-deployment: support defaultMode parameter for django.extraVolumes

pna-nca:
.Values.django.extraVolumes is useful to pass aritrary files from K8s to
DefectDojo containers. However, it does not have customizable
"defaultMode" parameter. This makes file permissions having default
value which in some cases (i.e. private key) can be considered as
insecure.

This commit adds .Values.django.extraVolumes.defaultMode parameter with
default mode set to 0644 (this should be suitable for most needs).


dsever:
PR needs to go against  dev not master

PR #6796, opened 2022-09-05
Translate metrics

shipko:
In this PR was added localization for metrics pages.


shipko:
@damiencarol I think we have some problem with execution Integration tests / test (tests/finding_test.py, postgres-redis) (pull_request)

On my laptop this test is passed

PR #6806, opened 2022-09-07
Adding Closed Notes to Jira

Roooodie:
current condition : All notes that we sent through the finding detail are sent to jira. But, the closed notes we sent when we close the finding is not sent to jira even if that notes are saved to the finding notes in the finding detail page.
proposed changes : sent the close finding notes to jira also


Maffooch (approved):

mtesauro (approved):
Approved

PR #6807, opened 2022-09-07
fix: using Capabilities.APIVersions.Has for new ingress api

ChristianSpiess:
The helm `semverCompare` dosn't always works as expected. The `Capabilities.APIVersions.Has` function is predestined for the apiVersion switch. It is backward compatible to older cluster and older versions of helm.


dsever:
This is duplicate to https://github.com/DefectDojo/django-DefectDojo/pull/6759

ChristianSpiess:
Sure - thx.
I'll close this as this is a dup #6759

PR #6809, opened 2022-09-07
tools: NeuVector: introducing NeuVector (REST) scan type

pna-nca:
This commit adds the support of NeuVector
(https://github.com/neuvector/neuvector) tool for importing scan results. Scan results can be exported via REST API in JSON format (that is why the tool is named 'NeuVector (REST)'). There is no GUI for that at the moment.

Scan results are just a list of issues found in packages installed in a container or an image. Very similar to Twistlock.

NeuVector also provides compliance scan results. This is not supported by the introduced tool.

### DIFF
--- a/dojo/tools/neuvector/parser.py
+++ b/dojo/tools/neuvector/parser.py
@@ -1,0 +1,117 @@
+import json
+import logging
+
+from dojo.models import Finding
+
+logger = logging.getLogger(__name__)
+
+NEUVECTOR_SCAN_NAME = 'NeuVector (REST)'
+NEUVECTOR_IMAGE_SCAN_ENGAGEMENT_NAME = 'NV image scan'
+NEUVECTOR_CONTAINER_SCAN_ENGAGEMENT_NAME = 'NV container scan'
+
+
+class NeuVectorJsonParser(object):
+    def parse(self, json_output, test):
+        tree = self.parse_json(json_output)
+        items = []
+        if tree:
+            items = [data for data in self.get_items(tree, test)]
+        return items
+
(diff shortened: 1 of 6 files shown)


kareem-DA:
I could use this.  Thanks for adding

dojo/tools/neuvector/parser.py
```diff
@@ -0,0 +1,117 @@
...
+                unique_key = node['name'] + str(package_name + str(
+                    node['package_version']) + str(node['severity']))
+                items[unique_key] = item
+        return list(items.values())
+
+
+def get_item(vulnerability, test):
+    severity = convert_severity(vulnerability['severity']) if 'severity' in vulnerability else "Info"
+    vector = vulnerability['vectors_v3'] if 'vectors_v3' in vulnerability else "CVSSv3 vector not provided. "
+    fixed_version = vulnerability['fixed_version'] if 'fixed_version' in vulnerability else "There seems to be no fix yet. Please check description field."
```
Maffooch, line 51:
```suggestion
    fixed_version = vulnerability.get('fixed_version', 'There seems to be no fix yet. Please check description field.')
```
Something like this across the file would be much cleaner

Maffooch (changes requested):
Overall, this is a great first pass

dojo/tools/neuvector/parser.py
pna-nca: Thank you, I've pushed the updated version.

Maffooch (approved):

mtesauro (approved):
Approved

PR #6811, opened 2022-09-07
Add support for connecting to Postgres DBMS over mutual TLS

pna-nca:


dsever:
I'm just wondering why not service mesh then?

pna-nca:
> I'm just wondering why not service mesh then?

I followed the straightforward approach and stumbled across various issues and came up with this solution as a result. I would vote myself for more clear implementation.

dsever:
> I followed the straightforward approach and stumbled across various issues and came up with this solution as a result. I would vote myself for more clear implementation.

I consider istio the more clean solution, have you tried to do the same using istio?

pna-nca:
> > I followed the straightforward approach and stumbled across various issues and came up with this solution as a result. I would vote myself for more clear implementation.
> 
> I consider istio the more clean solution, have you tried to do the same using istio?

Will have a look into it, thanks!

pna-nca:
> > I followed the straightforward approach and stumbled across various issues and came up with this solution as a result. I would vote myself for more clear implementation.
> 
> I consider istio the more clean solution, have you tried to do the same using istio?

Two points here:
- DefectDojo lacks support of specific authentication method to access DBMS. This commit fixes that. Using some third-party solution as a workaround can be okay until this is not implemented on the side of DefectDojo. Even if workaround is more clear, DefectDojo should not abandon such support.
- Istio does not support the discussed connection type, so it does not work even as a workaround. See https://github.com/istio/istio/issues/29761

devGregA:
@pna-nca I believe are helm people are away right now :( I apologize this can't be reviewed sooner.

dsever:
@pna-nca  I will try to review it in the next few days.

alles-klar:
Thanks for your work @pna-nca and sorry for the long delay.

I think the current approach mixes tls (for encryption) and mtls (for authentication). Of course tls is required for mtls, but in a lot of situations tls is sufficient. So we should be able to configure these two things separate. Can you take a look and ensure that we can activate tls without mtls and tls with mtls?

A point where I am not so happy with, is the modification of the entrypoint. It makes it harder to maintain and test, but I don't have a better solution for it at the moment.

dsever:
I'm still having mix feelings about that, it is not the way how things should be done in k8s, it will not scale. This is the reason why cloud native is recommending to use service mesh.

Maffooch:
Closing due to inactivity

PR #6816, opened 2022-09-07
fix(helm): fix postgresql svc name

ryok-0319:
## What

Fix PostgreSQL service name in the Helm chart

## Background

When `postgresql.architecture` is `replication`, 2 services with suffixes are created
- `${release-name}-postgresql-primary`
- `${release-name}-postgresql-read`

*Those suffixes are also customizable via `primary.name` and `readReplicas.name`, but the default values are `-primay` and `-read`.

However, the current chart always tries to use `${release-name}-postgresql` regardless of `postgresql.architecture`


dsever:
@ryok-0319 can you please take a look at the tests, some of them failed

mtesauro:
GHA has been flaky - I re-ran them and they passed...

dsever (approved):
Approved

PR #6854, opened 2022-09-16
Fix log_user_login_failed if username is missing

kiblik:
`log_user_login_failed()` was failing if `username` as missing in `credentials`


Maffooch (approved):

mtesauro (approved):
Approved

PR #6855, opened 2022-09-16
Parse mitigation status for Veracode SCA findings

coheigea:
The Veracode SCA mitigation part is not parsed, meaning that findings aren't being marked as mitigated in DD.


Maffooch (approved):

mtesauro (approved):
Approved

PR #6858, opened 2022-09-18
Use psycopg2 instead of psycopg2-binary.

Ayrx:
This PR changes DefectDojo to use the `psycopg2` package instead of `psycopg2-binary`.

This fixes a [problem](https://github.com/psycopg/psycopg2/issues/1360) with running DefectDojo on `aarch64` machines, with the main motivation here being able to do dev work on an Apple Silicon Macbook. 

The `psycopg2` package also [recommends](https://pypi.org/project/psycopg2-binary/) build from source instead of relying on the binary package.

> The binary package is a practical choice for development and testing but in production it is advised to use the package built from sources.


Maffooch:
There was some talk about moving to psycopg3 recently. That may be the better approach here

mtesauro:
I agree with @Maffooch - if we're going to change how we handle that dependency, might as well upgrade to psycopg3

Ayrx:
I'm not familiar enough with the DefectDojo codebase to make that change, since I believe the APIs aren't compatible, so someone else will have to take that on. :)

devGregA:
@Ayrx thank you for your PR, we're going to work internally on going to psycops3, so I think adding the binary is moving in the wrong direction.

Ayrx:
It occured to me that you will have the same issue with psyocpg3. The recommended installation method for production sites is the local installation method which still requires `libpq-dev`.

https://www.psycopg.org/psycopg3/docs/basic/install.html#local-installation

damiencarol:
@Ayrx so? are modifying your PR to switch to `psyocpg3` or we abandon this PR?

Maffooch (approved):
psycopg3 is not ready for Django yet, so this is a good step in the right direction :)

mtesauro:
Thanks @Maffooch for doing the research on this :+1:

mtesauro (approved):
Approved

PR #6867, opened 2022-09-20
Fix accepted by name

JoshBrodieTM:
Removed get_full_name from accepted_by in two templates, as accepted_by doesn't get set to a user, it's free text entry. Prior to this fix, "Accepted By" value was blank on these two pages.


Maffooch (approved):

mtesauro (approved):
Approved

mtesauro:
Merging - this PR has nothing to do with Helm

PR #6875, opened 2022-09-21
added Null value check in vectorString of Github Scan

L3m0nb4tt3ry:
Github Scan Parser fails if the vectorString is null. Github API might return null value for vectorString if it's not available. Added a check to only parse if vectorString is not `None`


Maffooch:
@L3m0nb4tt3ry I was hoping to push this directly into a hot fix, but there is a lot of extra baggage on the commit logs after rebasing. Will redo this PR based against another branch and git credit to you in the title

PR #6885, opened 2022-09-22
Fix improper reactivation in reimporter, using is_mitigated

Gby56:
Due to incomplete fix in https://github.com/DefectDojo/django-DefectDojo/pull/6452
Explainer:
The initial fix was too narrow in its decision/conditions, only looking at items/findings that had a `mitigated` timestamp.
Instead, `is_mitigated` should be used, to simply check if the incoming item is already mitigated or not, and if the existing finding is mitigated or not. I have left the timestamp comparison IF it exists on the item.

### DIFF
--- a/dojo/importers/reimporter/reimporter.py
+++ b/dojo/importers/reimporter/reimporter.py
@@ -99,7 +99,7 @@ class DojoDefaultReImporter(object):
                         else:
                             # even if there is no mitigation time, skip it, because both the current finding and the reimported finding are is_mitigated
                             continue
-                    if not item.is_mitigated:
+                    else:
                         logger.debug('%i: reactivating: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
                         finding.mitigated = None
                         finding.is_mitigated = False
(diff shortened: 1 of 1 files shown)


dojo/importers/reimporter/reimporter.py
```diff
@@ -84,17 +84,22 @@ def process_parsed_findings(self, test, parsed_findings, scan_type, user, active
...
+                                continue
+                            if item.mitigated.timestamp() != finding.mitigated.timestamp():
+                                logger.debug("New imported finding and already existing finding are both mitigated but have different dates, not taking action")
+                                # TODO: implement proper date-aware reimporting mechanism, if an imported finding is closed more recently than the defectdojo finding, then there might be details in the scanner that should be added
+                                continue
+                        else:
+                            # even if there is no mitigation time, skip it, because both the current finding and the reimported finding are is_mitigated
                             continue
-                    if not item.mitigated:
+                    if not item.is_mitigated:
```
coheigea, line 102: Couldn't we just have an else: here instead?

coheigea (approved):

Maffooch (approved):

mtesauro (approved):
Approved

PR #6888, opened 2022-09-22
Fix url creation for github for file location

Demaz93:
The URL inside the file location (in Finding view) generated from Repo URL (in Engagement configuration) is only based on `source_code_management_uri` but scanning results come front different commit hash or branches that are provided by CI/CD platform. I added (only for GitHub) a way to generate the right URL for GitHub repositories gathering hash commit or branch from engagement config.


Maffooch (approved):

mtesauro (approved):
Approved

mtesauro:
Merging this as it has nothing to do with the helm charts

Issue #6897, opened 2022-09-25
Promote "Potential finding" to real Finding error when Jira configured


drJabber:
**Bug description**
I think there is a bug in JIRAFindingForm.init code in line
https://github.com/DefectDojo/django-DefectDojo/blob/d8f759d43f13af1523a1b12e5080ede455d4ca8a/dojo/forms.py#L2688
```python

if self.instance and self.instance.has_jira_group_issue:
            self.fields['push_to_jira'].widget.attrs['checked'] = 'checked'
            self.fields['jira_issue'].help_text = 'Changing the linked JIRA issue for finding groups is not (yet) supported.'
            self.initial['jira_issue'] = self.instance.finding_group.jira_issue.jira_key
            self.fields['jira_issue'].disabled = True
```
Because when Jira configured for product and I tried to promote "Potential Finding" (Stub_Finding) to real finding - there is an exception like "no has_jira_group_issue attribute" in stub_finding
I think there should be code in condition like a couple of lines before
```python
if self.instance:
            if hasattr(self.instance, 'has_jira_group_issue') and self.instance.has_jira_group_issue:
```

**Steps to reproduce**
1. Configure Jira Instance in DD
2. Configure project to push all issues to Jira
3. Create engagament
4. Create Test in engagement
5. Add "Potential finding" in test (below findings list)
6. Press "promote to finding"
7. Fill all necessary fields in finding form
8. Press "promote finding" button
9. Got an error 

**Expected behavior**
Potential finding should be converted to Finding and new finding data should be pushed to Jira to create new Jira issue 

**Deployment method** *(select with an `X`)*
- [ x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Ubuntu 20.04
 - DefectDojo version (see footer) or commit message: 2.14.3

**Logs** 
something like "no has_jira_group_issue attribute"

coheigea:
This can be closed, as a fix is merged.

PR #6898, opened 2022-09-26
Fix error when promoting stub finding to JIRA

coheigea:
A fix for https://github.com/DefectDojo/django-DefectDojo/issues/6897

Tested that it gets pushed to JIRA successfully with the change.


Maffooch (approved):

mtesauro (approved):
Approved

Issue #6170, opened 2022-04-19
API - Tool Product Settings


guidorapisarda:
**Description**
POSTing to the API endpoint "tool_product_settings" with "product" parameter other than 1 (for example, using "product":2 in the body of the request) causes Dojo to ignore it and instead creating on product 1.
Also, it's required a parameter that is not listed.
![image](https://user-images.githubusercontent.com/12972491/164108760-03c4c3ec-1f04-4070-98d8-0d79a2039054.png)
![image](https://user-images.githubusercontent.com/12972491/164108551-925c2965-ade6-487b-a37a-cdf2cb7fefee.png)

**Bug description**
POSTing to the API endpoint "tool_product_settings" with "product" parameter other than 1 (for example, using "product":2 in the body of the request) causes Dojo to ignore it and instead creates it on product 1.

**Steps to reproduce**
Just using curl to create the POST request and sending it triggers this bug.

**Expected behavior**
Create the tool setting in the correct product.

**Deployment method** *(select with an `X`)*
- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Amazon Linux 2
 - DefectDojo version (see footer) or commit message: v2.9.0

**Logs** 
N/A

**Sample scan files**
N/A

**Screenshots**
Request:
![image](https://user-images.githubusercontent.com/12972491/164104734-3e52886f-b88e-4829-affc-9f7591df6673.png)

Response:
![image](https://user-images.githubusercontent.com/12972491/164110931-687275aa-7f3c-41bd-bb92-babf35890b27.png)

**Additional context** (optional)

guidorapisarda:
The issue is now fixed, thanks!

PR #6904, opened 2022-09-27
[Fix] API - Tool Product Settings (#6170)

jpbowie:
Corrected issue #6170 by making the product editable in the Tool_Product_Settings model.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -811,6 +811,7 @@ class ToolConfigurationSerializer(serializers.ModelSerializer):
 
 class ToolProductSettingsSerializer(serializers.ModelSerializer):
     setting_url = serializers.CharField(source='url')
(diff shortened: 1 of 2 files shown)


Maffooch:
@jpbowie I don't believe this is the appropriate fix for this. The issue is stating that the product ID is appearing to be ignored in the POST request of the API. Could this be an issue with the API endpoint rather than the model itself?

Also please base your changes against the `dev` branch

jpbowie:
I forgot to change the request to the dev branch, sorry.  I think the change makes the Tool_Product_Settings model consistent with other models that have a product foreign key relationship.  The uneditable product with a default value of 1 as assigned in the model also explains why no matter what product value the user provided, it was always assigned to product 1.

jpbowie:
I think I found a way to allow the product to be specified in the API and saved by the serializer without the changes to the model.  Will this work?

Maffooch (approved):
Yup this is much better! Thanks :)

damiencarol (approved):

Issue #6866, opened 2022-09-20
Broken report generator


jakealbasan:
The report builder doesn't work properly when multiple severities are selected for filtration. 

During a report generation, if multiple severity selected, the report builder only generates a report for single severity items. For example, if critical and high vulnerability findings are filtered, the report builder generates the report with only "High" vulnerability findings.

However, If no severity is selected at all, the generated report works as expected and includes all findings. 

**Steps to reproduce**
1. Click on reports
2. Enter a name for report options and leave the rest with defaults
3. Drag the findings widget to the center. Under filters, select critical and high severity
4. Select a product to limit the finding counts
5. Apply filters
6. Click "RUN"
7. Findings report has only high severity items

**Expected behavior**
The generated report should list both critical and high severity findings

**Deployment method** *(select with an `X`)*
- [x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: [CentOS 7.9]
 - DefectDojo version:  [2.14.1]

**Logs** 
There are no error logs during this process. However, the below logs are generated during the filtering and RUN button clicked. Logs are from uwsgi container.

"[pid: 1|app: -|req: -/-] 10.1.x.102 (admin) {58 vars in 1081 bytes} [Tue Sep 20 14:20:41 2022] GET /alerts/count => generated 14 bytes in 11 msecs (HTTP/1.1 200) 6 headers in 171 bytes (1 switches on core 3)
[pid: 1|app: -|req: -/-] 10.1.x.102 (admin) {68 vars in 1434 bytes} [Tue Sep 20 14:20:41 2022] POST /reports/custom => generated 46635 bytes in 557 msecs (HTTP/1.1 200) 6 headers in 182 bytes (1 switches on core 1)
[pid: 36|app: -|req: -/-] 10.1.x.102 (admin) {58 vars in 1081 bytes} [Tue Sep 20 14:22:53 2022] GET /alerts/count => generated 14 bytes in 13 msecs (HTTP/1.1 200) 6 headers in 171 bytes (1 switches on core 3)
[...]

PR #6906, opened 2022-09-27
Fixing broken report generation when more than one item is selected

coheigea:
A fix for https://github.com/DefectDojo/django-DefectDojo/issues/6866

When multiple severities are selected, it only picks the first one. This is because it's appending to a list which doesn't modify the internal list.

Tested that the new report contains both high and critical severities when they are selected.


Maffooch (approved):

mtesauro (approved):
Approved

PR #6916, opened 2022-09-29
Add the ability to only create finding groups when you have more than…

coheigea:
… one finding in the group

It only makes sense to me on importing a scan to create finding groups, if you have more than one finding in the group. Otherwise, you can end up with as many finding groups as there are findings. This PR adds a new checkbox to control whether to always create finding groups for every finding, or whether to do it only when you have more than one finding in the group.

Tested:
 * When importing multiple findings, it only creates finding groups where the finding group contains > 1 entry. Tested it only creates one JIRA per finding group, and one JIRA for findings that aren't in a finding group.
 * Tested that when the checkbox is ticked, the old behaviour is still valid - to create a finding group for all findings.
 * Tested the behaviour without grouping works as expected.

### DIFF
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -627,6 +627,7 @@ def import_scan_results(request, eid=None, pid=None):
             environment = Development_Environment.objects.get(id=environment_id)
 
             group_by = form.cleaned_data.get('group_by', None)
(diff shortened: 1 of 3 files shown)


Maffooch (approved):

blakeaowens:
Tested this and everything works as expected. Thanks @coheigea!

blakeaowens (approved):

PR #6924, opened 2022-09-30
Fix defect dojo

manikandan-rajappan:


Maffooch:
What is the purpose of this?

PR #6926, opened 2022-09-30
Fixed Tag serialization to support multiple tags correctly

fhoeborn-cb:
Fixed a bug, that a list of tags was not supported correctly in the API (e.g. tags="tag1, tag2" would result in one tag "tag1, tag2" instead in two tags "tag1" and "tag2")


fhoeborn-cb:
Closed as there is still something wrong with that fix - gonna reopen / create new PR when all that is resolved

PR #6928, opened 2022-10-02
SonarQube API Importer: Add sonarcloud global org id, code clean up

Maffooch:
When SonarCloud integration was initially implemented, each API scan config needed to have an org ID in the service key2 slot. In most cases, this is the same org ID. This PR makes the ORG ID set at the Tool Configuration level with the option to override the global value at the API Scan Config level

Also did some code cleanup and docs


blakeaowens (approved):
Looks good to me.

PR #6930, opened 2022-10-02
Add validation error when users attempt to send email without valid SMTP server

Maffooch:
When sending a 'forgot password' button, I was greeted with a 500 error. I was used to having mailhog there on dev servers..

### DIFF
--- a/dojo/user/views.py
+++ b/dojo/user/views.py
@@ -590,7 +590,7 @@ class DojoPasswordResetForm(PasswordResetForm):
                 connection.open()
                 connection.close()
         except Exception:
-            raise ValidationError("SMTP server is not conifugred correctly...")
+            raise ValidationError("SMTP server is not configured correctly...")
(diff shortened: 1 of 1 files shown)


dojo/user/views.py
```diff
@@ -577,10 +579,18 @@ def send_mail(self, subject_template_name, email_template_name,
...
+        super().send_mail(subject_template_name, email_template_name, context, from_email, to_email, html_email_template_name)
+
+    def clean(self):
+        try:
+            connection = get_connection()
+            if isinstance(connection, EmailBackend):
+                connection.open()
+                connection.close()
+        except Exception:
+            raise ValidationError("SMTP server is not conifugred correctly...")
```
blakeaowens, line 593: Configured spelled incorrectly.

blakeaowens (approved):
LGTM

PR #6931, opened 2022-10-02
Rev actions to latest ubuntu version per deprecation

Maffooch:
https://github.com/actions/runner-images/issues/6002

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -11,7 +11,7 @@ on:
 # Both builds have to be one worflow as otherwise one publish will overwrite the other
 jobs:
   deploy:
-    runs-on: ubuntu-18.04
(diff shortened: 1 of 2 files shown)


blakeaowens (approved):
Looks good to me.

damiencarol:
@Maffooch do we want that? it will force us to do maintenance on the doc part.

Maffooch:
Unfortunately do not have much of a choice tbh :/ the deprecation is happening either way

The gh-pages action will be phased out eventually after a few more versions as the the docs are being built in another repo that deploys too https://documentation.defectdojo.com. the gh-pages action is keeping https://defectdojo.github.io/django-DefectDojo/ alive for folks on older releases.

damiencarol:
ok, so moving forward

damiencarol (approved):

Issue #6935, opened 2022-10-03
Add "podLabels" option to helm chart


enidevops:
**Is your feature request related to a problem? Please describe**
As part of our asset management we want to have specific tags for our applications (in this case for the pods) like application name or organizational unit.

**Describe the solution you'd like**
Would it be possible to add something like Sonarqube does with their helm chart:
https://github.com/SonarSource/helm-chart-sonarqube/blob/master/charts/sonarqube/values.yaml (line 460)

```
# Additional labels to add to the pods:
# podLabels:
#   key: value
podLabels: {}
```

With this option it would be pretty easy to add additional tags to the pods.

dsever:
@enidevops be free to come up with a PR

dsever:
Resolved in https://github.com/DefectDojo/django-DefectDojo/pull/6941

PR #6941, opened 2022-10-04
added option to add labels using podLabels: {} #6935

enidevops:
As proposed in #6935, I have added the option to add custom pod labels.


alles-klar (approved):

dsever (approved):
Approved

PR #6945, opened 2022-10-05
Parse state, cwe and createdAt in the GitHub parser

coheigea:
Added support in the GitHub parser to:

1. Parse createdAt so that the findings have the correct creation date
2. Parse cwes to set the cwe for the finding
3. Parse state so that fixed issues don't show as active in DefectDojo


damiencarol (approved):

mtesauro (approved):
Approved

PR #6947, opened 2022-10-05
dojo: tool: introduce NeuVector compliance scans import support

pna-nca:
This commit makes DefectDojo to support compliance scans performed by NeuVector. Such scan results can only be fetched via REST API using endpoints like /v1/scan/workload/{id} and /v1/host/{id}/compliance. The latter one returns the results in a slightly different format. Both of them are supported.


damiencarol:
Please check you Flake8 issues

pna-nca:
> Please check you Flake8 issues

Thank you, it should be fixed now.

damiencarol (approved):

mtesauro (approved):
Approved

PR #6950, opened 2022-10-06
Make version optional for cyclonedx

coheigea:
version is optional for a CycloneDX components, but DefectDojo throws an error if version isn't present.
https://cyclonedx.org/docs/1.4/json/#components_items_version


damiencarol:
good catch.

damiencarol (approved):

mtesauro (approved):
Approved

PR #6968, opened 2022-10-11
Use DEDUPE_ALGO_HASH_CODE for Acunetix360

coheigea:
Use DEDUPE_ALGO_HASH_CODE with title,description, same as for the Acunetix scanner.
Tested dedup works OK for the same report on different engagements.


damiencarol (approved):

mtesauro (approved):
Approved

PR #6970, opened 2022-10-11
Enable filtering Findings on steps_to_reproduce

CharlieSears:
# Issue
It is not possible to filter `Findings` on `steps_to_reproduce`.

There appears to be a typo in the filter expression in filters.py at [`step_to_reproduce = CharFilter(lookup_expr='icontains')`](https://github.com/DefectDojo/django-DefectDojo/blob/e6a88af6109c8d9fdeee9dddc87a319d74292e88/dojo/filters.py#L1108)

# Solution
1. Fix typo replacing `step_to_reproduce` with `steps_to_reproduce`.
2. Add tests

I could take or leave the tests, there are no other examples of api filters being tested, but it didn't feel right to make a change without a test.

No other occurrences of `step_to_reproduce` exist in the repository.

# Examples
### Original
both `step_to_reproduce` and `steps_to_reproduce`
image

### With Fix
image

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -1105,7 +1105,7 @@ class ApiFindingFilter(DojoFilter):
     references = CharFilter(lookup_expr='icontains')
     severity = CharFilter(method=custom_filter, field_name='severity')
     severity_justification = CharFilter(lookup_expr='icontains')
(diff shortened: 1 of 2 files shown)


damiencarol (approved):

mtesauro (approved):
Approved

PR #6973, opened 2022-10-12
Add a link back to the vulnerability in the references for Acunetix360

coheigea:
It's useful in DefectDojo to be able to click on a link in the report to go back to the dashboard in Acunetix for the finding.

### DIFF
--- a/dojo/tools/acunetix360/parser.py
+++ b/dojo/tools/acunetix360/parser.py
@@ -38,8 +38,10 @@ class Acunetix360Parser(object):
             if sev not in ['Info', 'Low', 'Medium', 'High', 'Critical']:
                 sev = 'Info'
             mitigation = text_maker.handle(item.get("RemedialProcedure", ""))
-            references = "https://online.acunetix360.com/issues/detail/" + item["LookupId"] + "\n"
-            references += text_maker.handle(item.get("RemedyReferences", ""))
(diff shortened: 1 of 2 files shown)


dojo/tools/acunetix360/parser.py
```diff
@@ -38,7 +38,8 @@ def get_findings(self, filename, test):
             if sev not in ['Info', 'Low', 'Medium', 'High', 'Critical']:
                 sev = 'Info'
             mitigation = text_maker.handle(item.get("RemedialProcedure", ""))
-            references = text_maker.handle(item.get("RemedyReferences", ""))
+            references = "https://online.acunetix360.com/issues/detail/" + item["LookupId"] + "\n"
+            references += text_maker.handle(item.get("RemedyReferences", ""))
```
Maffooch, line 42: I am not super familiar with acunetix, but is it possible that the lookupID is not present in the report? Maybe something like this would be a tad safer
```suggestion
            references = text_maker.handle(item.get("RemedyReferences", "")) 
            if "LookupId" in item:
                references = f"https://online.acunetix360.com/issues/detail/{item["LookupId"]}\n" + references
```

coheigea:
Thanks @Maffooch , I've applied the change.

Maffooch (approved):
Thanks @coheigea :)

mtesauro (approved):
Approved

PR #6974, opened 2022-10-12
Fix issues with Bugcrowd API parser not marking not_applicable/not_re…

coheigea:
…producible as inactive

There are some bugs with the Bugcrowd API parser:

- It doesn't mark not_applicable (risk accepted) findings as inactive.
- It doesn't mark not_reproducible findings as inactive


mtesauro (approved):
Approved

PR #7005, opened 2022-10-18
Update the Cobalt API parser's API client

ericcornelissen:
Update the Cobalt.io API parser's `CobaltAPI` client to fetch the maximum allowed number of findings and assets.

This fixes a bug in the parser due to changes to the Cobalt.io API that resulted in users not being able to
1. import findings from all of their assets, and
2. import all findings from any given asset.


mtesauro (approved):
Approved

PR #7012, opened 2022-10-21
Don't save vulnerability Ids on a re-import if they're already define…

coheigea:
…d for the finding

I noticed on calling the API that it was returning multiple duplicate vulnerability Ids for a Finding, after calling re-import multiple times. This is because on each re-import it's saving a new VulnerabilityId for the same finding + CVE.

### DIFF
--- a/dojo/importers/reimporter/reimporter.py
+++ b/dojo/importers/reimporter/reimporter.py
@@ -233,7 +233,8 @@ class DojoDefaultReImporter(object):
                         file_upload.save()
                         finding.files.add(file_upload)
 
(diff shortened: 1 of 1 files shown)


damiencarol:
@coheigea the logic is good but it will produce 1+N requests :/ , I'm ok to merge this one but it can trigger some performance problems and will require more work.

coheigea:
@damiencarol I've re-worked it so that the re-importer only calls this utils method if the finding has an unsaved vulnerability Id.

Maffooch (approved):

mtesauro (approved):
Approved

PR #7013, opened 2022-10-21
Add support for ZAP "XML with requests and responses" format

malexmave:
I have [added a new XML report format to ZAP](https://github.com/zaproxy/zap-extensions/pull/3983) that includes full request and response headers and bodies. This makes ZAP findings a lot more useful, as they now show the entire req/response pair, making it easier to understand what exactly ZAP did and why it believes the response to be a problem. This PR adds support for the new format to DefectDojo.

Here's an example of what the result looks like in DefectDojo, with my patch applied:
Results in DefectDojo

As you can see, there is a small inconsistency in how the data is displayed: The response does not have an empty line between HTTP headers and body. This is due to the fact that [the model removes blank lines](https://github.com/malexmave/django-DefectDojo/blob/feature/zap-request-response/dojo/models.py#L2777-L2778). As changing this would potentially break things in other places, I chose not to change this behavior as part of this pull request, even though it means that there are some cosmetic problems with the output - but I'd be happy to make that small change if you believe that it should be made, of course.

As this is my first PR to this project, I'm not sure if anything else is required - let me know if I forgot anything :).

### DIFF
--- a/docs/content/en/integrations/parsers.md
+++ b/docs/content/en/integrations/parsers.md
@@ -1283,4 +1283,4 @@ Import Yarn Audit scan report in JSON format. Use something like `yarn audit --j
 
 ### Zed Attack Proxy
 
-ZAP XML report format.
+ZAP XML report format (with or without requests and responses).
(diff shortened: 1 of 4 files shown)


damiencarol (approved):
Your patch is good.

I agree, this lines doesn't make sense:
image

Curious to know why we have them.

Maffooch (approved):

PR #7027, opened 2022-10-26
Make requests of a peer review available to those with read access

Maffooch:
When users have read only access, It makes sense for them to be able to request another user to review a finding


damiencarol (approved):

mtesauro (approved):
Approved

Issue #7030, opened 2022-10-26
500 error when creating a Pentest finding with 'push to jira option' enabled through webui


devsecopsale:
**Bug description**
When a finding is created and 'push to jira' is enabled, I get a  500 error.
The only way to create the jira ticket is by creating the finding first and enable 'push to jira' flag later.

**Steps to reproduce**
1. create a finding and enable 'push to jira' option
2.  see the error

**Expected behavior**
Ticket and finding should be created. None of that happens.

**Deployment method** *(select with an `X`)*
- [ x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version 2.15.1

** Logs **
ker_1  | [26/Oct/2022 11:53:08] INFO [celery.worker.strategy:161] Task dojo.tasks.async_dupe_delete[cfd6b9df-588b-4cd0-8b75-ef1f53074ca5] received
celeryworker_1  | [26/Oct/2022 11:53:08] INFO [celery.app.trace:131] Task dojo.tasks.async_dupe_delete[cfd6b9df-588b-4cd0-8b75-ef1f53074ca5] succeeded in 0.012051589787006378s: None
uwsgi_1         | [pid: 1|app: -|req: -/-] X.X.X.X (admin) {56 vars in 1048 bytes} [Wed Oct 26 11:53:20 2022] GET /alerts/count => generated 13 bytes in 24 msecs (HTTP/1.1 200) 6 headers in 171 bytes (1 switches on core 1)
uwsgi_1         | [pid: 31|app: -|req: -/-] X.X.X.X (admin) {56 vars in 1048 bytes} [Wed Oct 26 11:53:28 2022] GET /alerts/count => generated 13 bytes in 26 msecs (HTTP/1.1 200) 6 headers in 171 bytes (1 switches on core 0)
uwsgi_1         | [26/Oct/2022 11:53:37] ERROR [django.request:224] Internal Server Error: /test/2/add_findings
uwsgi_1         | Traceback (most recent call last):
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
uwsgi_1         |     response = get_response(request)
uwsgi_1         |   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
uwsgi_1         |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
[...]

PR #7035, opened 2022-10-27
Fix for error on pushing an adhoc finding to JIRA

coheigea:
A fix for https://github.com/DefectDojo/django-DefectDojo/issues/7030


mtesauro (approved):
Approved

PR #7067, opened 2022-11-02
Fixed system notifications

ksotik:
Before this fix django tried to get one row from database for system notifications settings, but there are two rows when user_id=None - system and template notifications settings. As a result there was used Notifications() instance with default settings `(alert, alert)` and system notifications settings was ignored.


dsever:
Please do against the dev repo.

Issue #7059, opened 2022-11-01
Add support for Edit Finding Status when Close Finding


italvi:
**Is your feature request related to a problem? Please describe**
When editing a finding I can choose the status of a finding, e.g. false positive but I dont have to provide a reason for the finding being a false positive, which make it easy for a developer to suppress findings.

**Describe the solution you'd like**
I would like to use the "close finding" Features, which forces to write a note. NEW: When closing a finding I would like to choose, additionally to providing a reason, an option why the finding is being closed, e.g. false positive. I attached an edited image how I would expect it.
![Example](https://user-images.githubusercontent.com/58322186/199355199-a3fb1ba8-01ad-4da8-b87d-e4e447e510c7.jpg)

I already tried to implement it, but according to my understanding you are using the Model = Notes in the forms.py when closing a finding and therefore this would result in a bigger change, which I am not capable of implement as I am lacking knowledge of the code here.

italvi:
i tried to fix this issue in #7068

italvi:
was fixed by #7068

PR #1539, opened 2019-09-04
Add note-types to organize finding notes

piyarathnalakmali:
When submitting a pull request, please make sure you have completed the following checklist:

- [x] Your code is flake8 compliant 
- [x] Your code is python 3.5 compliant
- [x] If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR. ([Feature Documentation](https://github.com/DefectDojo/django-DefectDojo/files/3574357/Feature.Documentation.-.DefectDojo.Note.Type.Modification.pdf))
- [x] Model changes must include the necessary migrations in the dojo/dd_migrations folder.
- [x] Add applicable tests to the unit tests.

**The note-type modification has no effect on new or existing installations, or the existing functionalities of the DefectDojo application. All the changes and modifications done by this improvement will be taken into account only if the system admin configures any note-types.**

In the current DefectDojo application there is a functionality for the users to add notes to findings. In this modification a new field “note type” is introduced. For each note added to a finding there should be an associated note-type.

This improvement is useful when it is required to store notes in an organized manner and to restrict closing of findings unless the finding has fulfilled some requirements of notes. 

- As an example, for each finding there is a requirement to know what are the use cases affected by the particular vulnerability. In such case, the admin can add a new note-type as "Use case", thereby allowing the users to add a "Use case" type note to describe the affected use cases.

As part of this feature, a system admin can define note-types according to their requirements. Admin has all the privileges to add, edit, disable and enable note-types as required.

In the previous DefectDojo application, to close a finding user has to add a closing note explaining the reason. With the current improvement, in order to close a finding notes representing all mandatory note-types should also be available. This will ensure that every closed finding has all the details about that finding and its impact on the product. Also any other user can easily verify a closed finding since all notes are properly organized.  

[Feature Documentation - DefectDojo Note Type Modification  .pdf](https://github.com/DefectDojo/django-DefectDojo/files/3574357/Feature.Documentation.-.DefectDojo.Note.Type.Modification.pdf)

**WSO2 uses DefectDojo as the centralized Vulnerability Management System. This modification is a contribution done by WSO2.** WSO2 will use this feature to mandate note types such as: 
- Use case
- Resolution details
- Impact

PR #7068, opened 2022-11-02
Add Status of Findings when Close Finding

italvi:
This is a try to resolve the issue discussed in #7059 

Result:
grafik

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -559,7 +559,7 @@ class FindingViewSet(prefetch.PrefetchListMixin,
                     finding.mitigated = timezone.now()
                 finding.mitigated_by = request.user
                 finding.active = False
-                finding.false_p = finding_close.validated_data.get('false_p, False)
+                finding.false_p = finding_close.validated_data.get('false_p', False)
                 finding.duplicate = finding_close.validated_data.get('duplicate', False)
                 finding.out_of_scope = finding_close.validated_data.get('out_of_scope', False)
 
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -559,6 +559,9 @@ class FindingViewSet(prefetch.PrefetchListMixin,
                     finding.mitigated = timezone.now()
                 finding.mitigated_by = request.user
                 finding.active = False
+                finding.false_p = finding_close.validated_data['false_positive']
+                finding.duplicate = finding_close.validated_data['duplicate']
+                finding.out_of_scope = finding_close.validated_data['out_of_scope']
 
                 endpoints_status = finding.endpoint_status.all()
                 for e_status in endpoints_status:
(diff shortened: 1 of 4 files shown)


italvi:
@alles-klar as far as I know you are an expert to the k8 deployment: Do you know why it fails?

mtesauro:
I re-ran the failed ones and it seems to be working now.

dojo/api_v2/views.py
```diff
@@ -559,6 +559,9 @@ def close(self, request, pk=None):
                     finding.mitigated = timezone.now()
                 finding.mitigated_by = request.user
                 finding.active = False
+                finding.false_p = finding_close.validated_data['false_p']
+                finding.duplicate = finding_close.validated_data['duplicate']
+                finding.out_of_scope = finding_close.validated_data['out_of_scope']
```
Maffooch, line 564:
```suggestion
                finding.false_p = finding_close.validated_data.get('false_p, False)
                finding.duplicate = finding_close.validated_data.get('duplicate', False)
                finding.out_of_scope = finding_close.validated_data.get('out_of_scope', False)
```

italvi:
@Maffooch you are right with the .get. Was a bit lazy from my side to do it in one file but not another. I implemented your suggestion and added a missing apostrophe.

italvi:
I have 2 improvements that I would like to add in the future:

1. For the duplicates field there should be the possiblity to choose from findings which is the "original"
2. Remove the possibility to edit the status in the EditForm, s.t. nobody can just change the status without adding a note.

What do you think? Would this be also good PR?

Maffooch (approved):

Maffooch:
> For the duplicates field there should be the possiblity to choose from findings which is the "original"

This seems sensical to me when manually setting the flag.

> Remove the possibility to edit the status in the EditForm, s.t. nobody can just change the status without adding a note.

I think this would be a little annoying tbh. Especially for smaller teams or folks just getting to know dojo. I think it would be best to leave that to the users

italvi:
> > Remove the possibility to edit the status in the EditForm, s.t. nobody can just change the status without adding a note.
> 
> I think this would be a little annoying tbh. Especially for smaller teams or folks just getting to know dojo. I think it would be best to leave that to the users

@Maffooch could you explain me https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/finding/views.py#L744 then? It seems to me that somebody had the inital thought of checking whether the finding includes any notes? If so, then the contributor mad a mistake with `closing_disabled!=0`, which should be `closing_disabled==0 ` as notes will be 0 when there are no notes and therefore closing_disabled will remain 0

Maffooch:
Looks like something to do with note types from this #1539 
From just a brief skim of the PR description, it looks like if you have a specific note type configured, it requires that a note of that type be present in the finding  before it is actually closed. I like this approach

italvi:
@mtesauro could you please review the PR too or is something missing?

mtesauro:
@italvi LGTM - approving and merging

mtesauro (approved):
Approve

PR #7079, opened 2022-11-05
Fix flaw in logic in edit_finding

italvi:
If I understand it correctly the closing should be disabled if a special note is missing. So if note_type exists the closing should be NOT disabled. Therefore it makes more sense to have closing_disable = 1 and set it to 0.
Alternative: It has to be `if closing_disabled == 0` because it was initialed with it.


italvi:
seems like this was a misunderstanding because of the variable name closing_disabled

PR #7083, opened 2022-11-08
Feature/add tool ggshield Parser

iamnihal:
This PR will add ggshield (https://github.com/GitGuardian/ggshield) secret scanner JSON Report as a new scanner/importer.

### DIFF
--- a/docs/content/en/integrations/parsers.md
+++ b/docs/content/en/integrations/parsers.md
@@ -533,6 +533,10 @@ Example:
 }
 ```
 
+### Ggshield
+
+Import [Ggshield](https://github.com/GitGuardian/ggshield) findings in JSON format.
+
 ### Gosec Scanner
 
 Import Gosec Scanner findings in JSON format.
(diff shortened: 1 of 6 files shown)


Maffooch (approved):
This looks good to me
@damiencarol what do you think?

Gby56:
@iamnihal Hello ! Cool idea to add ggshield, is this intended for use with the pre-commit hook ? Do you think we could add an example in the docs, how to reconfigure the pre-commit to upload the results to a Defectdojo instance ?

I think ggshield isn't reporting the issues found via the gitguardian API, so we don't really get reporting through it, defectdojo could be a temporary solution. https://github.com/GitGuardian/ggshield/issues/421 I have reported the problem to GGuardian

damiencarol (approved):

iamnihal:
Hello @Gby56!!
This PR is not intended to be used with the pre-commit hook as ggshield currently doesn't provide the scan's output in pre-commit scan mode and hence in this case it's not possible to integrate ggshield with Defectdojo. However, pre-push hook scan mode supports JSON output.

Yes, ggshield is not reporting when used via Gitguadian API. Thank you @Gby56, for raising the issue in ggshield. If they add a feature to provide the scan's output in the pre-commit scan mode, then we can think of integrating Defectdojo with ggshield (in the pre-commit hook).

PR #7095, opened 2022-11-08
Adjust possibility of out_of_scope when edit finding

italvi:
When editing a finding currently only it is checked whether a finding is set to false positive when it is being closed but it could also be closed because it is out of scope. This is added in this PR


Maffooch (approved):
@italvi good one

PR #7102, opened 2022-11-09
Make SLA calculation of a Finding Group use the most severe finding

Maffooch:
When using finding groups within a test, the SLA calculation for groups passes the entire group to the SLA calculation for a single finding. There is an expected mismatch there. This PR takes the most severe finding I the group and passes that finding the SLA calculation function so that the group has the fastest approaching SLA.

### DIFF
--- a/dojo/templatetags/display_tags.py
+++ b/dojo/templatetags/display_tags.py
@@ -243,7 +243,9 @@ def group_sla(group):
         return ""
 
     # if there is at least 1 finding, there will be date, severity etc to calculate sla
-    return finding_sla(group)
+    # Get the first finding with the highests severity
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

mtesauro:
@Maffooch Looks like there's something deprecated in our Helm Lint GH Action setup :-(

Maffooch:
Yes, Helm chart unit testing has bee pretty broken for a few months now :/ I change it slightly each release in hopes of finding the right fix, but no dice.

This error happens because the target branch is not `dev`, and the problematic helm test runs and (expectedly) fails :/

Issue #7098, opened 2022-11-09
Incorrect logic with close_old_findings in the importer


coheigea:
When close_old_findings=true, in the importer it has this logic:
```
        # TODO I don't think these criteria are 100% correct, why are findings with the same hash_code excluded?
        # Would it make more sense to exclude duplicates? But the deduplication process can be unfinished because it's
        # run in a celery async task...
        if test.engagement.deduplication_on_engagement:
            old_findings = Finding.objects.exclude(test=test) \
                                            .exclude(hash_code__in=new_hash_codes) \
                                            .filter(test__engagement=test.engagement,
                                                test__test_type=test.test_type,
                                                active=True)
        else:
            # TODO BUG? this will violate the deduplication_on_engagement setting for other engagements
            old_findings = Finding.objects.exclude(test=test) \
                                            .exclude(hash_code__in=new_hash_codes) \
                                            .filter(test__engagement__product=test.engagement.product,
                                                test__test_type=test.test_type,
                                                active=True)
```
What I noticed is that when I have an existing engagement in a product, and import a second engagement with the same test type, it will incorrectly mitigate all the issues in the first engagement! The logic above in the else statement gets all findings that aren't in the test being uploaded and whose hash-codes don't match in the same product, and deactivates them. 

The deduplication switch doesn't make sense here, deduplication should be left for the async task. Instead it should only close all old findings in the same engagement.

coheigea:
This can be closed as the PR is applied

PR #7108, opened 2022-11-10
Only close old findings in the importer in the same engagement.

coheigea:
A fix for https://github.com/DefectDojo/django-DefectDojo/issues/7098

Tested:
- Importing a new test doesn't close old findings of the same type in a different engagement in the same product.
- Importing a new test does close old findings of the same type in the same engagement.


Maffooch (approved):
This change makes sense to me. Good one

mtesauro (approved):
Approved

Issue #6963, opened 2022-10-10
Jira 9.0+ is not supported


pna-nca:
**Bug description**
An error occurs when trying to integrate DefectDojo with Jira 9.0+, on altering product settings.

**Steps to reproduce**
0. Be sure that target Jira instance has version 9.0+.
1. Configure Jira instance in DefectDojo settings.
2. Go to the existing product and select "Settings --> Edit"
3. In Jira integration settings provide the valid settings. Press "Submit".
4. See non-descriptive error that Jira settings were not saved.

**Expected behavior**
A clear and concise error message why the error occurred.
Working without errors with Jira instances 9.0+. :)

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version: 2.15.0

**Logs** 
No logs without DEBUG mode.
In DEBUG mode the error is "invalid JIRA Project Config, can't retrieve metadata". See also below.

**Additional context** (optional)
Jira's createmeta endpoint is deprecated since 2020, see https://confluence.atlassian.com/jiracore/createmeta-rest-endpoint-to-be-removed-975040986.html
However, Jira python module still uses it, see https://github.com/pycontribs/jira/issues/1462
DefectDojo uses jira.createmeta() call in get_jira_meta() function in jira_link/helper.py file.
Moreover, DefectDojo does not handle exception properly which may be triggered by jira.createmeta().
See pull request #6962.

fderoubaix:
Hello,

I'm currently experiencing the same problem (DefectDojo v2.11 / Jira server v9.2 and Kubernetes deployment).

When I try to edit an existing product (id 42) in DefectDojo and set Jira integration, i've got this message : "JIRA Project config not stored due to errors."

In the logs of the uwsgi container there is this line that appears:
`- WARNING [django.request:224] Not Found: /jira/webhook/XXX`

I checked the webhook id on both sides and everything matches

In the logs of the nginx container there are a lot of connection tests with random users of my Jira instance (I don't know why the requests go through with these random users instead of the account registered in the configuration in Jira instance settings) : 

```
- "GET /alerts/count HTTP/1.1" 200 13 "https://defectdojo.xxx.com/product/42/edit" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:105.0) Gecko/20100101 Firefox/105.0" "xx.xx.xx.xx"

[...]

PR #7110, opened 2022-11-10
Support jira 9

pmilosev:
Fixes: https://github.com/DefectDojo/django-DefectDojo/issues/6963
Depends on: https://github.com/pycontribs/jira/pull/1527

If you don't have the update in the mentioned dependency than behaves as it used to - works until Jira 9 and breaks afterwards.
If you update the dependency than it will continue working after Jira 9.
With the updated dependency it will throw deprecation warnings on Jira versions [8.4 - 9).

Tested manually against Jira  7.13.8, 8.22.6 and 9.3.1.


pmilosev:
I see there are some code formating rules that I need to correct - will reopen another pul request

PR #7116, opened 2022-11-10
Fix helm chart using external databases

johnvandeweghe:
The helm chart does not currently allow external databases without modification of the ```values.yaml``` file by hand. This makes deploying the helm chart from the repository impossible if you want to use an external database.

Helm does not allow setting the values if they are not defaulted in values.yaml. Having the values defaulted will not affect the chart because they are each in templated ```if``` blocks.

It's possible this failure was introduced with Helm 3 specifically not allowing the use of ```.Values``` that aren't in the ```values.yaml```.

This is something I noticed while setting it up, so I made a PR. I didn't think it should be against dev, because it would be nice to get this in without waiting for an application release. If you need be to rebase or make a fresh branch I can.

### DIFF
--- a/helm/defectdojo/Chart.yaml
+++ b/helm/defectdojo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "2.16.1"
 description: A Helm chart for Kubernetes to install DefectDojo
 name: defectdojo
-version: 1.6.45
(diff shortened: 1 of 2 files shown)


johnvandeweghe:
I was wrong and misinterpreting an error, closing this.

PR #7117, opened 2022-11-11
Fix deprecations of GHA (set-output and node12)

Maffooch:
- GitHub has begun adding warnings to actions that are using the `set-output` directive in favor of a different format
  - https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ 
- GitHub is also deprecating the use of Node12 in favor of Node16, so some actions needed to be updated or replaced 
  - https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -64,7 +64,7 @@ jobs:
           
       # export docker images to be used in next jobs below
       - name: Upload image ${{ matrix.docker-image }} as artifact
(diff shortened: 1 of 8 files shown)


mtesauro (approved):
Approved

mtesauro (approved):
Approved (after updates)

damiencarol (approved):

PR #7135, opened 2022-11-14
Fix verified/active logic in the re-importer

coheigea:
There is a bug in the re-importer if a new finding is added. It will always set active/verified according to the parameters in the re-importer call and ignore what is set in the original tool. This change aligns the re-importer logic with what the importer does.


Gby56 (approved):
Great !

Maffooch:
Hi @coheigea 

I have been thinking about this PR for a few days now and am very conflicted. The intention of the active/verified flags on the import/re-import features is to allow a manual override of whatever the tool provides. The reason for this is because DefectDojo is supposed to be the central source of truth for vulns across all tools. 

With this PR, there is no way to override those statuses if the tool being used happens to provide those statuses. For example, a user could import a scan report and want all findings to come in as active and verified, and then be surprised when they are not all active/verified. I see this problem being very prevalent for users that are new to DefectDojo that have been tracking remediation progress in other tools until trying DefectDojo.

This is further complicated by the nature of the boolean either being True/False so that removes the possibility of allowing for a "Give me the default or whatever comes from the tool" situation. Maybe an additional flag would be a good idea to identify that you are trying to override the statuses coming from the tool? What are your thoughts on that?

mtesauro:
@coheigea FWIW, I had similar thoughts to what @Maffooch said in the previous comment. The additional flag idea he suggests makes sense to me.

damiencarol:
> For example, a user could import a scan report and want all findings to come in as active and verified, and then be surprised when they are not all active/verified. I see this problem being very prevalent for users that are new to DefectDojo that have been tracking remediation progress in other tools until trying DefectDojo.

This comment is important. I think the intended behavior was to allow to "force" to "verified"/"Active" in the UI or the API.
So the boolean in the UI  is maybe not the best UI component.

Maybe we should rework the UI part and switch to dropdown like this:

image

Gby56:
Yep a dropdown would be more explicit perhaps, but then... I thought about the implication of making such a big API change, it could potentially break a lot of user's workflows... I don't think the reimporter is as used as the importer but still, if we make the change on both it will be quite breaking

damiencarol:
The change will impact only the UI. For the API part and the rest of the code, I think there is the way to check if the attribute is defined.

Gby56:
Right, what I meant was on the block of code that forces the values, it can't force to active if the item is not active, and it can't force to verified if the item is not verified ?
The comment in the code seems to say "else final status is active" but I don't see how that's possible.
If we correct that, it will impact the internal logic a lot.

```
            if item.active:
                item.active = active
            if item.verified:
                item.verified = verified
```

Maffooch:
I think the drop down is the best approach. I don't think the change would be all that impacting either:

UI:
```
Force to True -> active/verified = True
Force to False -> active/verified = False
Not Specified -> active/verified = None (this would be the default as it would allow users to understand the relationship between sec tool and dojo better to decided if they want override or not)
```

API:
```
Active/Verified supplied -> active/verified = value supplied
Active/Verified *not* supplied -> active/verified = None (this should be the default behavior, aka passing None to the importer/re-importer)
```

In the import/re-import logic:
```
if active is not None:
    # indicates an overide. Otherwise, do not change the value of item.active
    item.active = active

if verified is not None:
    # indicates an overide. Otherwise, do not change the value of verified
    item. verified = verified
```

dojo/api_v2/serializers.py
```diff
@@ -1542,8 +1542,8 @@ class ImportScanSerializer(serializers.Serializer):
     def save(self, push_to_jira=False):
         data = self.validated_data
         close_old_findings = data.get('close_old_findings')
-        active = data.get('active')
-        verified = data.get('verified')
+        active = data.get('active', None)
+        verified = data.get('verified', None)
```
damiencarol, line 1546: This is not needed, by default `.get()` returns None if not here

coheigea:
@Maffooch Please review again, I've implemented your suggestions. I think the test failures are not related to my PR, as they are happening for other PRs as well.

tests/dedupe_test.py
```diff
@@ -515,8 +515,8 @@ def test_import_service(self):
         driver.find_element(By.PARTIAL_LINK_TEXT, "Path Test 1").click()
         driver.find_element(By.ID, "dropdownMenu1").click()
         driver.find_element(By.LINK_TEXT, "Re-Upload Scan").click()
-        driver.find_element(By.XPATH, '//*[@id="base-content"]/form/div[3]/div/div').click()
-        driver.find_element(By.XPATH, '//*[@id="base-content"]/form/div[4]/div/div').click()
+        # driver.find_element(By.XPATH, '//*[@id="base-content"]/form/div[3]/div/div').click()
+        # driver.find_element(By.XPATH, '//*[@id="base-content"]/form/div[4]/div/div').click()
```
damiencarol, line 519: please remove them or add a comment that explicitly say that we don't want to force active/verified

damiencarol (approved):
See my minor comment but the PR is good. It improve and clarify a lot the active/verified management

Maffooch (approved):
Great work here @coheigea 

Will merge after the release tomorrow

PR #7153, opened 2022-11-16
Fix helm unit test duplicate 'key' value in secretKeyRef

nnamzlas:
This PR removes the duplicate `key` value in the `secretKeyRef` of the `DD_DATABASE_PASSWORD` environment variable. The change should keep the same behaviour as previously when it comes to default values but ensures that only a single `key` value exists.

### DIFF
--- a/helm/defectdojo/templates/tests/unit-tests.yaml
+++ b/helm/defectdojo/templates/tests/unit-tests.yaml
@@ -50,9 +50,12 @@ spec:
               key: postgresql-postgres-password
               {{- else if eq .Values.database "mysql" }}
               name: {{ .Values.mysql.auth.existingSecret }}
+              {{- if .Values.mysql.auth.secretKey }}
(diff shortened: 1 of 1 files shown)


alles-klar (approved):

mtesauro:
@alles-klar Looking at the changes, I can't see why the unit-tests would fail - any idea why those and the Helm linting are failing.

I went ahead and re-ran both of those checks but the failures continued.

alles-klar:
Maybe because the target branch was first the master branch?
@nnamzlas The PR should be based on the dev branch. Which branch did you use?

dsever:
> Maybe because the target branch was first the master branch? @nnamzlas The PR should be based on the dev branch. Which branch did you use?


I think you are on the right track, it should be dev->dev

nnamzlas:
Ohh, yeah, I did base this on the master branch. Created #7203 to use the correct source branch. I'll close this one and we can continue there :)

PR #7161, opened 2022-11-18
Remove enforcement of unnecessary notes

### LINKED ISSUE #7160: Note_Type enforces unnecessary notes
**Bug description**
If closing a note and having note_type activated the current check enforces to include one more note than necessary

**Steps to reproduce**
1. Activate Note_Type and make it mandatory
2. Go to a finding and close it, as you can see in the image you can now only add a note but not close it: 
image
3. Add the note, now the finding can be closed but an additional note is required for that: 
image

**Expected behavior**
When all note_types are already in the finding the finding should be closed

italvi:
When note_types activated it should be checked if up to one note_type is missing. When including this note no note_types would be missing and therefore all mandatory notes are included and an additional note is not necessary.
resolve #7160

### DIFF
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -430,7 +430,7 @@ def close_finding(request, fid):
                 'Note Saved.',
                 extra_tags='alert-success')
 
-            if len(missing_note_types) == 0:
+            if len(missing_note_types) <= 1:
                 finding.active = False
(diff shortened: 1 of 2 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

PR #7163, opened 2022-11-18
Bugcrowd API parser fixes

Gby56:
I noticed that my implementation did a duplicate detection in the wrong way, since Bugcrowd itself marks submissions as duplicates, but they mean something else: another researcher made a submission about the same thing.
Since my API client is not returning the duplicate submissions, but only the "first come, first served" submission, there is no reason to use the duplicate field, and it was throwing a lot of issues.
So I ripped out the line `duplicate=bugcrowd_duplicate,`

I'm also fixing the links in the defectdojo findings in markdown, I had to propagate the api scan configuration object back to the parser so it can get the service key 1, which is the program name required in the url link.

I also have removed the "Informational" state from the active issues. It is considered to be a closed subject when Informational is used, so a lot of issues were shown active for no reason.

Lastly, I have added an inclusion of fields in the API call, to try to get the external Jira issues linked (but it's not implemented by the API yet apparently) and also getting the program, but it's not giving the name of it yet.

### DIFF
--- a/dojo/tools/bugcrowd_api/api_client.py
+++ b/dojo/tools/bugcrowd_api/api_client.py
@@ -44,7 +44,7 @@ class BugcrowdAPI:
             "filter[program]": program,
             "page[limit]": 100,
             "page[offset]": 0,
-            "include": "monetary_rewards,target",
+            "include": "monetary_rewards,target,program,external_issues",
             "sort": "submitted-desc",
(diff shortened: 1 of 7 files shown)


coheigea (approved):
LGTM

mtesauro (approved):
Approved

PR #7165, opened 2022-11-18
Enable notify for sla in system settings

italvi:
I included the settings with an or and made it per default false, s.t. we do not make a breaking change for already existing configs. For new persons to DD it will not be necessary to set ENV Variable for the notification anymore and the current user still have the value of their ENV Variable and can decide for themselves whether they want to use the system setting and remove the ENV variable


italvi:
the db migration was generated by the manage.py makemigrations command and I just added the file here. Why does the uni test fail?

dsever:
Would be great to add a comment to the settings file, like moved to system_setting and will be dedicated from settings file

Issue #5907, opened 2022-02-14
trivy defectdojo image scan


saldam72:
Hello team,
I have scanned DefectDojo 2.6.2 official image on Docker Hub Repository with Trivy 0.23 and I have found some vulnerabilities of CRTICAL (14) and HIGH (53) severity.

In case of need, I'm available to send you reports.

P.S. I have scanned also 2.7.1 and I have found again some vulnerabilities of CRTICAL (11) and HIGH (50) severity.

Regards,
Sal

valentijnscholten:
Did you triage the issues? These containers scans usually generate lots of noise due to incorrect version detection. We're running the latest version of alpine and all python packages, so not much more we can do. There may be some vulnerabilities in some python packages, but these have been triaged by use and are updated in `dev` already and part of the next release.
If you feel there are some real vulnerabilities still in dev, or critical/highs in master, that need attention, please let us know via the security disclosure process mentioned in our bug report template.

saldam72:
Hi Valentijn,
Trivy is the open source version of Aqua Security and is a scanner for known vulnerabilities in container images.

We are also running a static and dynamic application security testing and if we find any vulnerabilities we will let you know via the security disclosure process.

regards,
Sal

manuel-sommer:
I guess this can be closed @mtesauro .

PR #7166, opened 2022-11-19
Alpine based django docker image

gietschess:
**Description**

* Dockerfile for an alpine based django docker image
* Helm-Chart flag to choose between the default debian and alpine image

This should also satisfy the issue #5907

I've not worked with github workflows and docker-compose so far so i bet there's a way to combine those files for both distros but that's something where i could use some help.

**Test results**

After changing one line in the integration test tests/finding_test.py they ran pretty fine.
That should review someone who might can say why there's no alert but the text message shown on the page.

**Documentation**

Not really documentation for that change but fixed two links in the readme-docs/CONTRIBUTING.md

### DIFF
--- a/.github/workflows/integration-tests-django-alpine.yml
+++ b/.github/workflows/integration-tests-django-alpine.yml
@@ -1,0 +1,150 @@
+name: Integration tests
+# pull requests:
+# push:
+#      run on every push, which is when something gets merged also
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - master
+      - dev
+      - release/**
+      - hotfix/**
+
+env:
+  DD_DOCKER_REPO: defectdojo
+
+jobs:
+  build:
+    # build with docker so we can use layer caching
+    name: Build Image
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        docker-image: [django-alpine, nginx, integration-tests]
+    steps:
+      # - name: Login to DockerHub
+      #   uses: docker/login-action@v1
+      #   with:
+      #     username: ${{ secrets.DOCKERHUB_USERNAME }}
+      #     password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+
+      - name: Read Docker Image Identifiers
+        id: read-docker-image-identifiers
+        run: echo "IMAGE_REPOSITORY=$(echo ${{ github.repository }} | tr '[:upper:]' '[:lower:]')" >> $GITHUB_ENV
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          buildkitd-flags: --debug
+          driver-opts: image=moby/buildkit:master # needed to get the fix for https://github.com/moby/buildkit/issues/2426
+
+      - name: Build
+        id: docker_build
+        uses: docker/build-push-action@v3
+        env:
+          docker-image: ${{ matrix.docker-image }}
+        with:
+          context: .
+          push: false
+          tags: |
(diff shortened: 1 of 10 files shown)


dsever:
Like you read my mind, something I wanted to work on :)

Maffooch:
I am not super knowledgeable about OS related stuff. My knee-jerk reaction would be that it would likely not be wise to have Kubernetes and docker compose running different OS in the container. Either both should stay on Debian, or both move to alpine. 

@mtesauro what are your thoughts?

damiencarol:
@dsever @Maffooch my 2 cents: if it's possible to switch to Alpine, we should do it. lighter image and more up-to-date (security angle) images.

mtesauro:
While I'm not convinced that musl libc in Alpine is better than GNU libc (libc6) in Debian, it's definitely true that the images are smaller with Alpine:

```
REPO    |  TAG                   |  IMAGE ID      |  SIZE    | LOCATION      |
------------------------------------------------------------------------------
python  |  3.8.15-alpine3.16     |  c001bad9ce01  |  46.8MB  |  PR #7166     |
python  |  3.8.15-slim-bullseye  |  94fb3cc064e8  |  124MB   |  Main branch  |
python  |  3.11.0-slim-bullseye  |  626d06aea490  |  128MB   |  Dev branch   |
```
From [Docker Hub's entry for Python](https://hub.docker.com/_/python):

> This variant is useful when final image size being as small as possible is your primary concern. The main caveat to note is that it does use [musl libc](https://musl.libc.org/) instead of [glibc and friends](https://www.etalabs.net/compare_libcs.html), so software will often run into issues depending on the depth of their libc requirements/assumptions. See [this Hacker News comment thread](https://news.ycombinator.com/item?id=10782897) for more discussion of the issues that might arise and some pro/con comparisons of using Alpine-based images.

So there's a chance that we may have issues switching to musl libc - it's unknown as we've historically used Debian (libc) Python images to build and run our containers.

However, since the dev branch is targeting [Python 3.11.0](https://github.com/DefectDojo/django-DefectDojo/blob/dev/Dockerfile.django#L8), I'd suggest we should at least update this PR to use [3.11.0-alpine](https://hub.docker.com/layers/library/python/3.11.0-alpine/images/sha256-a71e8ff8638a437e4cd8ad80a609235af54bc04d75e1aa2832146595e31d83bc?context=explore) if the consensus is that we want to trade a smaller image for potential incompatibilities with musl libc.

I don't particularly find it problematic that we have 2 different Linuxes as base images (Alpine vs Debian) even if we end up with a container that is 77.2 MB bigger but tried and tested for years.  As a practical bonus, I find Debian an easier environment to debug issues in the Defectdojo/Django container over Alpine but that's likely mostly personal preference.

As far as having newer/more secure images with Alpine vs Debian - we have Dependabot and Renovate in place so that shouldn't be much of an issue overall as those tools make sure we're always running the latest version of the image.

FWIW, running `docker scan` against the base images in the table above results were "you are currently using the most secure version of the selected base image" - however, the Debian variants did have some older and vulnerable libraries (requiring local/terminal access).  So, on the security front, points go to Alpine.

So, I'd probably leave things as they are but won't fight a migration to Alpine if that's what we collectively decide.

Maffooch:
@devGregA @alles-klar what are your thoughts?

alles-klar:
I had the same idea 1 or 2 years ago. Long story short: if nothing fundamental has changed it is a fail at all. 
It ends in a much longer build time, a larger image and less performance.  
Let me check my notes for the exact reasons.

alles-klar:
Here is a blog article which explains the problems: https://pythonspeed.com/articles/alpine-docker-python/

mtesauro:
So, it seems the consensus is to close this and keep things as is 

@Maffooch @alles-klar @damiencarol  Now's the last chance to say you want this otherwise I'll close this in the next day or two.

alles-klar:
I talked with @gietschess and he told me, he didn't had the problems I had two years ago. So I would like to give it a second chance. Give me some time to check it out.

PR #7169, opened 2022-11-21
Fixing link generation code

coheigea:
I noticed for SAST, it doesn't generate a correct link back to GitHub when the repo is configured in the engagement, because the if statement for branch/commit hash etc only checks if the value is not None. Instead it should check for not empty.


mtesauro (approved):
Approved

PR #7171, opened 2022-11-21
Add a parameter to the importer/reimporter to set source_code_managem…

coheigea:
…ent_uri

Right now we can't set the github URL correctly using the importer/reimporter.


mtesauro (approved):
Approved

PR #7178, opened 2022-11-23
Add the ability to set the default assignee for a JIRA issue

coheigea:
It would be very useful to be able to set the default JIRA assignee for a product.

Tested:

- If the new default assignee field is blank, the JIRA is created with no assignee in JIRA.
- If the new default assignee field is specified, the JIRA is automatically assigned to this JIRA user on creation.


Maffooch (approved):
Great change

mtesauro (approved):
Approved

PR #7172, opened 2022-11-22
Implement create_finding_groups_for_all_findings in the Reimporter

coheigea:
A follow-on from https://github.com/DefectDojo/django-DefectDojo/pull/7150.

PR #7181, opened 2022-11-24
Add engagement end-date to API

### LINKED ISSUE #7156: Set engagement end date when auto_create_context with import/reimport API
Hi, 

I'm facing the need to customize the end date of an engagement created with `auto_create_context=true` when using the `import/reimport` API endpoints. 

It would be very convenient to be able to pass an additional `engagement_end_date` parameter for this. 

```
"curl --location --request POST <...>/api/v2/reimport-scan
            ...
            --form 'engagement_name=xxx' 
            --form 'engagement_end_date=2022-12-31'
            --form 'auto_create_context=true' 
            ...
```

In the codebase. the created engagement's end date is mandatory and a hardcoded value of 365 (one year) is used... 
So it should make sense to have a parameter for this one ?

https://github.com/DefectDojo/django-DefectDojo/blob/1bb23ddc071bfa947c9e9e840ea428e6cbfb2984/dojo/importers/reimporter/utils.py#L215

It might be a hassle to do date validation (eg. not before start date), so passing the number of days in a `engagement_period_days` parameter is fine too.

italvi:
Added the possibility to add an end_date for engagements, s.t. not a year is fix. If no end_date is provided defaults to old value of delta 365. Also added check whether start_date > engagement_date.
Closes #7156

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1498,6 +1498,7 @@ class ImportScanSerializer(serializers.Serializer):
     product_type_name = serializers.CharField(required=False)
     product_name = serializers.CharField(required=False)
     engagement_name = serializers.CharField(required=False)
+    engagement_end_date = serializers.DateField(required=False)
     source_code_management_uri = serializers.URLField(max_length=600, required=False, help_text="Resource link to source code")
     engagement = serializers.PrimaryKeyRelatedField(
         queryset=Engagement.objects.all(), required=False)
@@ -1566,9 +1567,10 @@ class ImportScanSerializer(serializers.Serializer):
         group_by = data.get('group_by', None)
         create_finding_groups_for_all_findings = data.get('create_finding_groups_for_all_findings', True)
 
+        engagement_end_date = data.get('engagement_end_date', None)
         _, test_title, scan_type, engagement_id, engagement_name, product_name, product_type_name, auto_create_context, deduplication_on_engagement, do_not_reactivate = get_import_meta_data_from_dict(data)
         engagement = get_or_create_engagement(engagement_id, engagement_name, product_name, product_type_name, auto_create_context,
-                                              deduplication_on_engagement, source_code_management_uri=source_code_management_uri)
+                                              deduplication_on_engagement, source_code_management_uri=source_code_management_uri, target_end=engagement_end_date)
 
         # have to make the scan_date_time timezone aware otherwise uploads via the API would fail (but unit tests for api upload would pass...)
         scan_date_time = timezone.make_aware(datetime.combine(scan_date, datetime.min.time())) if scan_date else None
@@ -1644,6 +1646,7 @@ class ReImportScanSerializer(TaggitSerializer, serializers.Serializer):
     product_type_name = serializers.CharField(required=False)
     product_name = serializers.CharField(required=False)
     engagement_name = serializers.CharField(required=False)
+    engagement_end_date = serializers.DateField(required=False)
     source_code_management_uri = serializers.URLField(max_length=600, required=False, help_text="Resource link to source code")
     test = serializers.PrimaryKeyRelatedField(required=False,
         queryset=Test.objects.all())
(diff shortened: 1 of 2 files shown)


dojo/api_v2/serializers.py
```diff
@@ -1735,16 +1739,15 @@ def save(self, push_to_jira=False):
                                                 commit_hash=commit_hash, push_to_jira=push_to_jira,
                                                 close_old_findings=close_old_findings,
                                                 group_by=group_by, api_scan_configuration=api_scan_configuration,
-                                                service=service, do_not_reactivate=do_not_reactivate,
-                                                create_finding_groups_for_all_findings=create_finding_groups_for_all_findings)
```
Maffooch, line 1739: Was this intentional?

Maffooch (changes requested):

dojo/importers/reimporter/utils.py
```diff
@@ -197,7 +197,7 @@ def get_or_create_product(product_name=None, product_type_name=None, auto_create
 
 
 def get_or_create_engagement(engagement_id=None, engagement_name=None, product_name=None, product_type_name=None, auto_create_context=None,
-                             deduplication_on_engagement=False, source_code_management_uri=None):
```
damiencarol, line 200: For retro compatibility add the default value here

dojo/api_v2/serializers.py
```diff
@@ -1498,6 +1498,7 @@ class ImportScanSerializer(serializers.Serializer):
     product_type_name = serializers.CharField(required=False)
     product_name = serializers.CharField(required=False)
     engagement_name = serializers.CharField(required=False)
+    engagement_end_date = serializers.DateField(required=False)
```
coheigea, line 1501: Can you add a help_text here to explain what the field does? Same for the Reimporter

coheigea (changes requested):

dojo/importers/reimporter/utils.py
italvi: @damiencarol  could you please explain your comment further? I already had target_end=None, it seems like git shows the diff strange, because I changed the line, so it looks like I removed it. This is the reason why I have the check for instance being probably none. Would it be better to have the default value of +365days in the definition?

dojo/api_v2/serializers.py
italvi: @Maffooch no, thanks for pointing out. Something went wrong when sync with the PR #7172

dojo/api_v2/serializers.py
italvi: @coheigea are the help text enough now or did I understood the intention of them wrong?

dojo/importers/reimporter/utils.py
```diff
@@ -213,8 +213,14 @@ def get_or_create_engagement(engagement_id=None, engagement_name=None, product_n
         if not product:
             raise ValueError('no product, unable to create engagement')
 
+        target_start = timezone.now().date()
+        if isinstance(target_end, type(None)):
+            target_end = (timezone.now() + timedelta(days=365)).date()
+        if target_start > target_end:
+            target_end = (timezone.now() + timedelta(days=365)).date()
```
coheigea, line 220: Couldn't these be combined in the same if statement?

coheigea (changes requested):

dojo/importers/reimporter/utils.py
italvi: you are right. This happened because I introduced the second if afterwards and changed the target_end value.

coheigea (approved):

Maffooch (approved):

mtesauro:
@damiencarol Was reviewing this one and about to approve but wanted to give you a chance to chime in on this PR.

dojo/importers/reimporter/utils.py
damiencarol: As you are adding a new parameter I was expecting that you put a default value. But it's ok

damiencarol (approved):

PR #7188, opened 2022-11-27
Added Custom Password Requirements to System Settings

blakeaowens:
Introduced functionality to allow admin users to set custom password requirements from system settings. All previously stated password requirements still exist as the default.

Within system settings, a new field has been added to set minimum password length, as well as fields to dictate whether numbers, uppercase/lowercase letters, and special characters are required within user passwords.
Screenshot 2022-11-26 at 7 49 14 PM

This functionality has been added to both changing user passwords, as well as setting the password when adding a new user.
Screenshot 2022-11-26 at 7 52 27 PM
Screenshot 2022-11-26 at 7 53 41 PM

When these settings are changed, all help text popups and form validation errors reflect the requirements set within system settings.
Screenshot 2022-11-26 at 8 03 19 PM
Screenshot 2022-11-26 at 8 03 33 PM
Screenshot 2022-11-26 at 8 03 44 PM


Maffooch:
One additional thing that would be nice to add would be a max password length. Some vendor risk questions ask about max length because of the "risk" of DoS attacks.

I think a default of 24 characters would be good default.

mtesauro:
FWIW, I'd make the max 48 - my password manager is set to 25 - if nothing else, I found out how many sites have a max less than that so we can do better ;-)

mtesauro (approved):
Approved

Maffooch (approved):

PR #7190, opened 2022-11-28
JFrog xray api corrections

madeoninfo:
- Removed populating the to-be-deprecated cve and replaced it with populating unsaved_vulnerability_ids
- Added population of the vuln_id_from_tool  (XRAY-xxxxxx)
- Improved the unique_id_from_tool to make sure that all cases are covered (With unique id, only cve, and no ids provided by Xray)
- Title changed to the summary of the Xray finding
- Minor code refactoring changes

image


madeoninfo:
image

dojo/tools/jfrog_xray_api_summary_artifact/parser.py
```diff
@@ -102,6 +95,15 @@ def get_item(vulnerability, service, test):
         unique_id_from_tool=unique_id_from_tool
     )
 
+    # Add vulnerability ids
```
madeoninfo, line 98: Included both cve and xray vulnerability ids in the finding.

dojo/tools/jfrog_xray_api_summary_artifact/parser.py
```diff
@@ -74,22 +69,20 @@ def get_item(vulnerability, service, test):
...
-        title = impact_path.name + ":" + impact_path.version
-        unique_id_from_tool = None
+        unique_id_from_tool = str(hash(impact_path.name + impact_path.version + vulnerability['summary']))
+        vuln_id_from_tool = ""
 
     finding = Finding(
+        vuln_id_from_tool=vuln_id_from_tool,
         service=service,
-        title=title,
+        title=vulnerability['summary'],
```
madeoninfo, line 84: Now that the XRAY and CVEs are persisted in the right locations the titles were changed to be easier to read

dojo/tools/jfrog_xray_api_summary_artifact/parser.py
```diff
@@ -74,22 +69,20 @@ def get_item(vulnerability, service, test):
     if len(impact_paths) > 0:
         impact_path = decode_impact_path(impact_paths[0])
 
-    # The unique_id_from_tool is set only when a given component (SHA) has a specific unique Finding (XRAY or CVE)
     if 'issue_id' in vulnerability:
-        title = vulnerability['issue_id'] + " - " + impact_path.name + ":" + impact_path.version
-        unique_id_from_tool = vulnerability['issue_id'] + " " + impact_path.sha
+        unique_id_from_tool = str(hash(impact_path.name + impact_path.version + vulnerability['issue_id']))
```
madeoninfo, line 73: The unique id from the tool has been corrected to depict a unique hash that is derived from the affected component at a  given version and the XRAY id, CVE, or description (if Xray does not provide an id). Hash code might also have a performance improvement impact when discovering duplicates and decupling.

damiencarol: shouldn't the unique_id_from_tool  be `vulnerability['issue_id']` ?

madeoninfo: I totally agree and it is the first priority, unfortunately, not all results from JFrog Xray have an XRAY issue_id. There are 3 cases that need to be considered in the respective order of preference :

1. JSON from XRay includes an XRAY-xxxxxx id (issue_id)

image

2. No XRAY-xxxxxx id exists in the results -> use the CVE if it exists

image

3. Neither the issue_id exists nor a CVE -> use the finding description
 
image

madeoninfo: For the last case the unique id from tool is a hash that combines the component name, version and the title.

image

One question that I need to ask you @damiencarol is in regard to the definition of the unique_id_from_tool. For **artifacts** (i.e. microservices) that have findings in dependant **components** is it:

_( NOTE : Available ID = XRAY, CVE or title of finding 

or reference let's consider that we have the following cases, there the only thing that differs is the artifact:

Finding 1 : **Microservice A**, Available ID A, Component A, Version A
Finding 2 : **Microservice B**, Available ID A, Component A, Version A

)_

a) a unique finding that refers to a specific **component** at a specific **version**. In this case, they will have the same unique_id_from_tool 

b) a unique finding that refers to a specific **component** at a specific **version** AND in relation to a specific **artifact**. In this case, they will have a different unique_id_from_tool

Maffooch: I would think option B would be the preferred choice. It will no doubt allow for more findings to be created (less duplicates identified) but I would much prefer that to fixing a vuln in Microservice A, and then forgetting about Microservice B since that was identified as a duplicate.

Maffooch:
@madeoninfo this looks good to me so far. Left one comment above.

dojo/tools/jfrog_xray_api_summary_artifact/parser.py
madeoninfo: I agree with you @Maffooch and will alter the creation of the unique id to include the **Artifact**. 

Regarding the resolution of **Findings**. If one has the same **Finding** across multiple **Artifacts** (Microservices) it would make sense to store this information somewhere for the developers to have an easier way to group them and address them by bumping up the **Component** dependencies or working on a specific **Artifact** at the time.

For now, the only way I see is to use the **Location**, unless there is a way that I am not aware of. Maybe the inclusion of an **Artifact** entity can be considered?

madeoninfo: In fact, I will opt to use the sha_256 key of the artifact to make sure that the exact binaries are identical for a given **Artifact** and not only the name and version.

image

madeoninfo:
After adding the MD5 hash id I am having some strange issues where the duplicates are not identified, even though the unique ids are the same. It might be my setup so please don't merge yet! Will have a look over the weekend

madeoninfo:
Same issue with sha256 hash. I double-checked the database and with 2 identical unique ids, the duplicate is not detected. It is only noted as a similar finding. Any idea why this is the case?

image

image

image

madeoninfo:
One thing that I noticed is that all necessary services are up but Celery is reported in the settings that it is not running

image

image

madeoninfo:
It was the celery worker on overdrive. It managed to catch up after a few hours and the duplicates have been discovered! 

image

Good to go as is ;)

damiencarol:
@madeoninfo awesome work. I'm reviewing your PR.
One advice: always check the celery status or your tests will be irrelevant.

madeoninfo:
Hi @damiencarol, I just realised something. The top level Artifact (microservice) is also a component. It would make more sense to have this mentioned as a component in DefectDojo instead of the second level component. Reporting would be more intuitive for developers and they could still see the 2nd level dependent component in the path. Is there a way to pass the 2 options as a parameter or do I need to create a separate parser?

madeoninfo:
image

Sorry for changing my mind about the parser level of component tracking and grouping. It makes much more sense to have the 1st level component (Artifact/Microservice) the one that is tracked in DefectDojo. When looking into the findings it is much more intuitive for the developers to focus on one Artifact at a time to resolve the 3rd party, 2nd level, component findings that are affecting the 1st level component. Filtering the findings is easier and the reporting is more intuitive and easy to understand by the developers and the client.

madeoninfo (commented):
All comments resolved and ready for your approval

Maffooch (approved):

damiencarol:
@madeoninfo this is good work. Sadly I didn't time to provide more feedback. Some comments you've made are right.
Feel free to push more PR to improve the component part.

PR #7198, opened 2022-11-29
Add middleware to alert users of missing trailing slash

Maffooch:
As the title says, the API does not respond very well to POST requests that do not have a trailing slash. Simply redirecting whatever data in the payload would not be a good idea either, so I took the "update the response" in transit route


alles-klar (approved):

eatasaur:
I think this change might have caused issues on my side. Now on the uwsgi container I seem to be getting issues:
```python
AttributeError: 'WSGIRequest' object has no attribute 'user'
```

PR #7203, opened 2022-11-29
Fix helm unit test duplicate 'key' value in secretKeyRef

nnamzlas:
This PR removes the duplicate key value in the secretKeyRef of the DD_DATABASE_PASSWORD environment variable. The change should keep the same behaviour as previously when it comes to default values but ensures that only a single key value exists.
Replaces #7153


dsever:
@nnamzlas  can you please rebase this one

nnamzlas:
Done @dsever

alles-klar (approved):

Maffooch (approved):

PR #7208, opened 2022-11-30
Fixed bug with the horusec parser's date parser.

bjhijmans:
`2022-02-02` is 10 characters long, but it only took the first 9.

### DIFF
--- a/dojo/tools/horusec/parser.py
+++ b/dojo/tools/horusec/parser.py
@@ -27,7 +27,7 @@ class HorusecParser(object):
 
     def get_findings(self, filename, test):
         data = json.load(filename)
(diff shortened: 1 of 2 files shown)


Maffooch:
Hi @bjhijmans please add a unit test that checks for the length of the date

bjhijmans:
I added a test. This one failed because the date was read as `2022-05-0` which couldn't be parsed, The first test file has a date of `2021-10-19` which was truncated to `2021-10-1` which was interpreted as `2021-10-01` . That 's why it didn't fail.

Maffooch (approved):

damiencarol (approved):

PR #7155, opened 2022-11-17
Dark Mode

salvianf:
This PR add button in navbar that can change to dark mode using inverted filter CSS  
resolve #6124

PR #7212, opened 2022-12-01
Dev salvian dark mode

### LINKED ISSUE #6124: Dark Mode Support
Although a cosmetic feature, implementing a dark mode would be beneficial for the eyes of the poor souls spending many hours on DefectDojo.

salvianf:
This PR add button in navbar that can change to dark mode using inverted filter CSS
resolve #6124. This PR is a continuation of #7155.


salvianf:
Hi @Gby56 @Maffooch @mtesauro, this is my new PR. Please review it. Thanks.

damiencarol (approved):

devGregA (changes requested):
Sorry, will weigh in this shortly.

dojo/static/dojo/css/dojo.css
```diff
@@ -1031,7 +1031,7 @@ div.custom-search-form {
 
 @media (min-width: 320px) {
     #page-wrapper {
-        margin: 75px 0 0 0;
+        margin: 0 0 0 0;
```
Gby56, line 1034: Why the change in margins ? I didn't run your code locally so I can't really make sense of that, but I thought it was only a css filter so maybe that's not required ?

Gby56: Update: I realized, it might be to apply the filter properly everywhere ? otherwise you had a white line ?

Gby56 (approved):

devGregA:
Hi everyone, sorry for the delay. This PR is hard to accept, because we offer this feature in the [commercial version](https://documentation.defectdojo.com/proprietary_plugins/) of DefectDojo. In our contributing guidelines, we do [recommend reaching out for new features](https://github.com/DefectDojo/django-DefectDojo/blob/master/readme-docs/CONTRIBUTING.md#submisison-pre-approval) so that people don’t waste effort. I am very open to your feedback on how we can handle this better as I truly don’t want to discourage any member of our community, but the commercial version is what makes the OS version possible.

Please let me know your thoughts and feedback.

Gby56:
Hello,

I've been thinking about that feature for a while and couldn't explain to myself how it was not already in the community DefectDojo... now I know haha

I really think these kinds of features are nowadays a bare minimum for accessibility and quality of life improvements, sorry if I'm being a bit blunt... I don't think this should be a commercial-grade feature tbh.

To be fair, the PR is only setting the mode client side and not saving that in user preferences so it's a tiny bit less "fancy", but not by much.

mtesauro:
FWIW, I don' t really think this is PR-worthy, especially when the UI could use a general overhaul and I'd prefer to spend time on that over adding dark mode to the existing UI.

<matt's 2 cents>
I use this extension to 'dark-ify' many sites that don't include a dark mode and it has generally worked great for me.  TBH, that's a primary reason why I'd not work on a dark mode for DefectDojo's current UI.
https://chrome.google.com/webstore/detail/dark-mode/dmghijelimhndkbmpgbldicpogfkceaj/related

Here's what the demo site looks like:
image

If that one doesn't work well for you, there's loads of dark mode plugins to try.
<\matt's 2 cents>

devGregA:
I don't see it as an accessibility feature, but I do see it as quality of life feature. Both when we open sourced Dojo and now, our goal is for Dojo to be a force for good to make security easier and accessible to all. The app, the algorithms, and even the integrations (SSO, JIRA, etc.) all open sourced. We appreciate all contributions and the community that's developed around Dojo. The tool wouldn't be where it is today without the Dojo Community. @Gby56, you have submitted some phenomenal PRs and I truly appreciate your involvement. We've tried to be very careful to make sure there aren't any barriers to a person or organization achieving their goals with the open source version. Where we think it makes sense to draw the line is quality-of-life features and out-of-the-box automation because these features aren't essential for achieving one's goals with Dojo, but provide a means to sustain and improve the platform for both open source and commercial users.

Gby56:
I understand your point of view much better, I had the impression that the paid DefectDojo's version was mainly attractive for customers because of the support and the deployment / maintenance taken care of, and also supporting the community, which is also an attractive choice.

But also, now that I think about it, this is a tool that handles sensitive data, and therefore, it will be deployed on-prem most of the time.

This makes it difficult to have a sustainable commercial version, and the licensed on-prem version is the best way to satisfy customer requirements, if they can't be in someone else's cloud.

So, the fine line between a community and a commercial on-prem version is super hard to navigate, because on-prem is the most used deployment model, both having the burden of running and maintenance of the system.

Therefore you really need to at least convince with features, or a cheap licensing that would mostly say "to support the project".

PR #7217, opened 2022-12-01
Add additional error catching for severity in Checkov Parser

Maffooch:


dojo/tools/checkov/parser.py
```diff
@@ -77,7 +77,7 @@ def get_item(vuln, test, check_type):
         resource = vuln['resource']
 
     severity = 'Medium'
-    if 'severity' in vuln:
+    if 'severity' in vuln and vuln['severity'] is not None:
         severity = vuln['severity'].capitalize()
```
damiencarol, line 81: One thing that could had been done is:

```Python
if vuln.get('severity'):
```

This way you cover every cases. Even even the attribute is here but have empty string.

PR #7222, opened 2022-12-02
Helm changes for using the alpine-based django docker image

gietschess:
**Description**

After review with @alles-klar i've split the pr #7166 .

* Changes in the helm files to make the alpine django image usable by feature flag.

**Test results**

See #7166

### DIFF
--- a/helm/defectdojo/templates/_helpers.tpl
+++ b/helm/defectdojo/templates/_helpers.tpl
@@ -106,22 +106,18 @@ Create chart name and version as used by the chart label.
   Builds the repository names for use with local or private registries
 */}}
 {{- define "celery.repository" -}}
-{{- printf "%s" .Values.repositoryPrefix -}}/defectdojo-django
 {{- printf "%s/defectdojo-django%s" .Values.repositoryPrefix (include "defectdojo.django.distro" .) -}}
 {{- end -}}
 
 {{- define "django.nginx.repository" -}}
-{{- printf "%s" .Values.repositoryPrefix -}}/defectdojo-nginx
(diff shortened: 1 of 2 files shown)


dsever:
@alles-klar  @gietschess  why we don't just use standard way, keep distro as part of tags like in 
https://hub.docker.com/_/python

So for DD could be: 

2.16.2--slim-bullseye (tagged as latest for now)
or 
2.16.2-slim-alpine
 
Anyway anyhow, we would have pinned versions in the future, I hope this weekend I'll finish it.

gietschess:
hi @dsever , we've also discussed that.
I've changed the image name instead of the tag in order to keep the tags of defectdojo-django and -nginx the same.

If you prefer changing the tags just let me know and i'll provide the helm changes.

Maffooch:
I cannot get the unit tests to pass on this one. It is failing in the same place consistently

gietschess:
hi @Maffooch ,
my fault, there was an error building the image names.

Maffooch (approved):

dsever:
Hi, 

Why this one is closed?

gietschess:
ah damn, sorry, forgot that this one was still open and deleted the branch. i've recreated it

dsever:
> ah damn, sorry, forgot that this one was still open and deleted the branch. i've recreated it

Thanks, this PR is valuable, but still is on my todo list

alles-klar:
I created a PR to push the alpine images: https://github.com/DefectDojo/django-DefectDojo/pull/7496. In the current implementation we must only update the tag to use the alpine image. But let's wait and discuss it there, how we name the images.

PR #7236, opened 2022-12-05
Improved the design of the custom html report findings

madeoninfo:
- Made the finding main header smaller to include the counter and the DefectDojo ID number
- Included the score in the severity
- Added the CVE if it exists with the relevant web link
- Location: Added the Service and fixed the issue with the file path width

image

I hope you like the changes, it produces smaller in size reports and includes additional and important information related to a finding.

### DIFF
--- a/dojo/templates/dojo/custom_html_report_finding_list.html
+++ b/dojo/templates/dojo/custom_html_report_finding_list.html
@@ -30,10 +30,6 @@
                     </h5>
                 </div>
             </div>
-            {% if finding.title %}
-                <h6>Title</h6>
-                <pre>{{ finding.title|markdown_render }}</pre>
-            {% endif %}
             <div class="table-responsive">
                 <table class="table-striped table table-condensed table-hover centered">
                     <tr><th>Title</th></tr>
--- a/dojo/templates/dojo/custom_html_report_finding_list.html
+++ b/dojo/templates/dojo/custom_html_report_finding_list.html
@@ -19,7 +19,7 @@
             <div class="panel-heading finding-title">
                 <div class="clearfix">
                     <h5>
-                        Finding {{ forloop.counter }} - DefectDojo ID {{ finding.id }}
+                        Finding {{ forloop.counter }} - {{ finding.title }}
                         {% if finding.tags %}
                             <sup>
                                 {% for tag in finding.tags.all %}
@@ -31,11 +31,7 @@
                 </div>
             </div>
             <div class="table-responsive">
-                <table class="table-striped table table-condensed table-hover centered">
-                    <tr><th>Title</th></tr>
-                    <tr><td>{{ finding.title }}</td></tr>
-                </table>
-                <table id="heading" class="table-striped table table-condensed table-hover centered">
+                 <table id="heading" class="table-striped table table-condensed table-hover centered">
                     <tr>
                         <th>Severity</th>
                         <th>Status</th>
@@ -55,6 +51,7 @@
                         {% if finding.vulnerability_ids %}
                             <th>Vulnerability IDs</th>
                         {% endif %}
+                        <th>Dojo ID</th>
                     </tr>
                     <tr>
                         <td>
@@ -97,6 +94,7 @@
                         {% if finding.vulnerability_ids %}
                             <td>{{ finding.vulnerability_ids }}</td>
                         {% endif %}
+                        <td>{{ finding.id }}</td>
                     </tr>
                 </table>
             </div>
--- a/dojo/templates/dojo/custom_html_report_finding_list.html
+++ b/dojo/templates/dojo/custom_html_report_finding_list.html
@@ -137,7 +137,6 @@
         {% endif %}
 
         {% if finding.steps_to_reproduce %}
-            <pre>{{ finding.impact|markdown_render }}</pre>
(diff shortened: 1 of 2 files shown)


Maffooch (dismissed):

dojo/templates/dojo/custom_html_report_finding_list.html
```diff
@@ -19,7 +19,7 @@ <h4>{{ finding.severity|capfirst }}</h4>
             <div class="panel-heading finding-title">
                 <div class="clearfix">
                     <h5>
-                        Finding {{ forloop.counter }}: {{ finding.title }}
+                        Finding {{ forloop.counter }} - DefectDojo ID {{ finding.id }}
```
Maffooch, line 22: Just realized the title of the finding is removed here, and not displayed any where else in the report. Was this intended?

dojo/templates/dojo/custom_html_report_finding_list.html
```diff
@@ -96,14 +108,16 @@ <h5>
...
-        <pre>{{ finding.description|markdown_render }}</pre>
+        {% if finding.description %}
+            <h6>Description</h6>
+            <pre>{{ finding.description|markdown_render }}</pre>
+        {% endif %}
 
-        <h6>Mitigation</h6>
-        <pre>{{ finding.mitigation|markdown_render }}</pre>
         {% if finding.get_report_requests %}
+            <h6>Mitigation</h6>
```
Maffooch, line 119: Putting the mitigation in this if block will only allow the mitigation section to appear in the report if the finding has request/response pairs. That means it will only be available for dynamic tools.

dojo/templates/dojo/custom_html_report_finding_list.html
```diff
@@ -113,22 +127,25 @@ <h6>Sample Request(s): Displaying {{finding.get_report_requests.count}} of {{fin
...
             {% endfor %}
-            {% endif %}
+        {% endif %}
 
-        <h6>Impact</h6>
-        <pre>{{ finding.impact|markdown_render }}</pre>
         {% if finding.steps_to_reproduce %}
-        <h6>Steps to Reproduce</h6>
-        <pre>{{ finding.steps_to_reproduce|markdown_render }}</pre>
+            <h6>Impact</h6>
```
Maffooch, line 133: Impact should not be dependent on steps to reproduce being present.

dojo/templates/dojo/custom_html_report_finding_list.html
```diff
@@ -77,15 +82,22 @@ <h5>
...
                         {% endif %}
-                        <td>{{ finding.id }}</td>
+                        {% if finding.cve %}
+                            <td>
+                                <a target="_blank"
+                                    href="https://nvd.nist.gov/vuln/detail/{{ finding.cve }}">
+                                    <i class="fa fa-external-link"></i> {{ finding.cve }}
+                                </a>
+                            </td>
+                        {% endif %}
```
Maffooch, line 100: I don't believe the CVE field is really used anymore. Please use the vulnerability ID instead.

Maffooch (changes requested):

dojo/templates/dojo/custom_html_report_finding_list.html
madeoninfo: It was, due to the description covering the finding, but on second thoughts it might be important. I'll add it, but not in the main heading as it can be really bulky

madeoninfo:
Doing a few more tests @Maffooch locally. Will let you know if all is ok

dojo/templates/dojo/custom_html_report_finding_list.html
```diff
@@ -77,15 +90,17 @@ <h5>
...
+                        {% if finding.cwe %}
                             <td>
                                 <a target="_blank"
                                     href="https://cwe.mitre.org/data/definitions/{{ finding.cwe }}.html">
                                     <i class="fa fa-external-link"></i> {{ finding.cwe }}
                                 </a>
                             </td>
                         {% endif %}
-                        <td>{{ finding.id }}</td>
+                        {% if finding.vulnerability_ids %}
```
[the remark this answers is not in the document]
madeoninfo: Managed to test it and it is the right field 

image

madeoninfo (commented):
Gone through all of your comments @Maffooch and resolved them

dojo/templates/dojo/custom_html_report_finding_list.html
madeoninfo: Vulnerability ids were used instead

image

Maffooch:
Most revisions look good! I am still weary about the title though. Do you have a screenshot?

madeoninfo:
> Most revisions look good! I am still weary about the title though. Do you have a screenshot?

image

Maffooch:
Hmm.. This could just be personal preference, but I think it is much more intuitive to have the title as the header,  and then the dojo ID as a table entry (similar to the way it was before)

@mtesauro  @devGregA thoughts?

madeoninfo:
Hi @Maffooch, for this PR the importance is on having the necessary data on the report. I have reverted the design of the title and the id to the old style.

Maffooch (approved):
Will make sure it gets into this release

PR #7237, opened 2022-12-06
Add govulncheck parser and unittests

ECOMMPAY:
Govulncheck reports known vulnerabilities that affect Go code. It uses static analysis of source code or a binary's symbol table to narrow down reports to only those that could affect the application. 
https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck
Command to scan the project in the current directory and generate a report for the parser: 
govulncheck -json . >> report.json


ECOMMPAY:
HI @Maffooch 
Thanks for your review!
1. We will add parser information very soon
2. The "affected" field is required because OSV (open source vunlerability) with CVE must have a vulnerable package anyway. Thus, when generating a report, this field will always exist, as well as the fields 'url' (link to information about CVE from the vulnerability database where it was found), 'title', 'cve', 'references' and other. Based on this, we do not introduce additional checks on dict index.
I could be wrong so please correct me if you know more.

https://ossf.github.io/osv-schema/

ECOMMPAY:
HI @Maffooch
I understood the importance of safely accessing dictionary elements.
I will definitely add this to the govulncheck parser
But I would like to hear your opinion on one issue.
I suggest get dictionary elements with [ ] in try/except blocks. If some field is not in the report, then the exception will not lead to the termination of the parser. This approach will provide higher readability of the code. The example you posted above is awesome, but in my implementation, using the get() method on a multidimensional dictionary will degrade the quality of the code
What do you think?

Maffooch:
I don't have a strong opinion one way or other tbh. Maybe the try/catch blocks would allow for more logging possibilities to make the issue more clear rather than silently ignoring and moving on

ECOMMPAY:
I updated the parser code, added try/except where the dictionary is accessed and in case of an exception in the logs it will be displayed

Maffooch (approved):

dojo/tools/govulncheck/parser.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def get_findings(self, scan_file, test):
+        findings = []
+        scan_data = scan_file.read()
+        # remove intro from developer
+        scan_data = scan_data[scan_data.find(b'{'):]
+        try:
+            data = json.loads(scan_data)
+        except Exception:
+            return findings
```
damiencarol, line 37: This is not good, you're doing exception swallowing.

dojo/tools/govulncheck/parser.py
```diff
@@ -0,0 +1,63 @@
...
+        return "Import Govulncheck Scanner findings in JSON format."
+
+    @staticmethod
+    def get_location(data, node):
+        return [f"{x['Pos']['Filename']}:{x['Pos']['Line']}:{x['Pos']['Column']}" for x in
+                data['Calls']['Functions'][str(node)]['CallSites'] if x['Parent'] == 1]
+
+    def get_findings(self, scan_file, test):
+        findings = []
+        scan_data = scan_file.read()
```
damiencarol, line 31: Please use the direct form:
```Python
json.load(scan_file)`
```

dojo/tools/govulncheck/parser.py
```diff
@@ -0,0 +1,63 @@
...
+                    vuln_methods = set(
+                        first_elem[0]['OSV']['affected'][0]['ecosystem_specific']['imports'][0]['symbols'])
+                    impact = set(self.get_location(data, first_elem[0]['CallSink']))
+                    for elem in elems:
+                        impact.update(self.get_location(data, elem['CallSink']))
+                        vuln_methods.update(elem['OSV']['affected'][0]['ecosystem_specific']['imports'][0]['symbols'])
+                except Exception as e:
+                    logger.warning('Skip vulnerability due %r', e)
+                    continue
+                d['impact'] = '; '.join(impact) if impact else NO_IMPACT
```
damiencarol, line 60: `impact` attribute is optional. So **it's better to not provide the data or set it to none**

dojo/tools/govulncheck/__init__.py
```diff
@@ -0,0 +1 @@
+# Denis Timoshenko
```
damiencarol, line 1:
```suggestion
```
Please let this file empty. We track authorship with Git

unittests/tools/test_govulncheck_parser.py
```diff
@@ -0,0 +1,48 @@
...
+        testfile.close()
+
+        self.assertEqual(2, len(findings))
+
+        with self.subTest(i=0):
+            finding = findings[0]
+            self.assertEqual("GO-2022-1039", finding.title)
+            self.assertEqual("Info", finding.severity)
+            self.assertEqual("CVE-2022-41715", finding.cve)
+            self.assertEqual("stdlib", finding.component_name)
```
damiencarol, line 33: Test also the component version. most screens use both atributes together.

damiencarol (changes requested):

dojo/tools/govulncheck/parser.py
ECOMMPAY: HI! @damiencarol
Thanks for your review!
I need to use read() to further remove extra text in the report before the json data. After that, the cleaned data is fed to the loads() input
Direct load() from a file object gives an error due to the fact that the report stores not only data in json format, but also text at the beginning
Is it possible to do without direct load() ?

ECOMMPAY:
HI @damiencarol 
Thanks again for your review
I fixed all your comments

ECOMMPAY:
Hi @damiencarol
Any updates to the parser review?

dojo/tools/govulncheck/parser.py
```diff
@@ -0,0 +1,68 @@
...
+    def get_version(data, node):
+        return data['Requires']['Modules'][str(node)]['Version']
+
+    def get_findings(self, scan_file, test):
+        findings = []
+        try:
+            data = json.load(scan_file)
+        except Exception as e:
+            logger.warning("Invalid JSON format. %r", e)
+            return findings
```
damiencarol, line 40: you're still doing exception swallowing

damiencarol (changes requested):

dojo/tools/govulncheck/parser.py
ECOMMPAY: Maybe use something like: raise ValueError("Invalid JSON format") ?

ECOMMPAY:
Hi @damiencarol
I updated parsing of empty files and tried to avoid swallowing exceptions.
What do you think of my decision?

dojo/tools/govulncheck/parser.py
damiencarol: or simply not catching the exception.

kiblik:
Hi. We are ordering the list of parsers alphabetically in the documentation. Can you move it a little bit higher, please?

docs/content/en/integrations/parsers.md
```diff
@@ -1292,3 +1292,7 @@ Import Yarn Audit scan report in JSON format. Use something like `yarn audit --j
 ### Zed Attack Proxy
 
 ZAP XML report format (with or without requests and responses).
+
+### Govulncheck
```
kiblik, line 1296: Hi. We are ordering the list of parsers alphabetically in the documentation. Can you move it a little bit higher, please?

kiblik (changes requested):

docs/content/en/integrations/parsers.md
ECOMMPAY: Hi!
No problem, I will do it

ECOMMPAY:
Hi @damiencarol 
Are there any updates or comments?
The problem with exception swallowing was solved by raising ValueError
Also I fixed parser.md

Maffooch:
@damiencarol are you good with this PR?

damiencarol (changes requested):
Two part are missing:
* Setting the unique_id_from_tool in Finding (in the report it's the id attribute)
* setting of the de-duplication algorithm.

jaskaransinghdr6j:
Hey guys, any updates on this?

Maffooch:
parser.md was deleted as part of #7418 .. Not sure why it showing as being deleted again in the diff. Not sure if it really matters

mtesauro (approved):
Approved

unittests/tools/test_govulncheck_parser.py
```diff
@@ -0,0 +1,65 @@
...
+        testfile = open("unittests/scans/govulncheck/many_vulns.json")
+        parser = GovulncheckParser()
+        findings = parser.get_findings(testfile, Test())
+        testfile.close()
+
+        self.assertEqual(3, len(findings))
+
+        with self.subTest(i=0):
+            finding = findings[0]
+            self.assertEqual("GO-2022-1144", finding.title)
```
damiencarol, line 33: We also need to test the attribute `unique_id_from_tool`

dojo/tools/govulncheck/parser.py
```diff
@@ -0,0 +1,69 @@
...
+                        vuln_methods = set(
+                            first_elem[0]['OSV']['affected'][0]['ecosystem_specific']['imports'][0]['symbols'])
+                        impact = set(self.get_location(data, first_elem[0]['CallSink']))
+                        for elem in elems:
+                            impact.update(self.get_location(data, elem['CallSink']))
+                            vuln_methods.update(
+                                elem['OSV']['affected'][0]['ecosystem_specific']['imports'][0]['symbols'])
+                    except Exception as e:
+                        logger.warning('Skip vulnerability due %r', e)
+                        continue
```
damiencarol, line 65: This will lead to silently removing Findings, if users starts to use re-import it will break the logic.
Why don't you just remove the try/except. It will make code easier to read.
Let upper layers manage exceptions. Also it will be hard to understand why a report is not fully imported and why some finding are missing.

damiencarol (changes requested):

damiencarol (approved):

blakeaowens (approved):

PR #7266, opened 2022-12-12
Enable remote RabbitMQ

cybercapsicum:
This PR adds the following features:
- A new value `celery.brokerHost` for specifying an external broker
  - Becomes the new default for also specifying an external Redis server
  - An example use case is users that may prefer [Amazon MQ](https://aws.amazon.com/amazon-mq/) or similar services
    - This service is untested, but it advertises RabbitMQ compatibility ([link](https://docs.aws.amazon.com/amazon-mq/latest/developer-guide/getting-started-rabbitmq.html)).
- A slight modification to the main configmap to reference the external hostname in the same way as Redis


dsever:
Nice, I will test it it in day/2

alles-klar (approved):

dsever (approved):
Approved

alfianabdi:
Hi, how to configure DD to use RabbitMQ with TLS ?

PR #7269, opened 2022-12-13
Importer + Re-Importer are setting active/verified to false by default

coheigea:
The recent fix I did to the importer + reimporter mean that the active/verified booleans are actually set to false by default, instead of None.


Maffooch (approved):

mtesauro (approved):
Approved

PR #7273, opened 2022-12-14
Change Endpoint.host_mitigated_endpoints not to include active Findings

coheigea:
I noticed in the Endpoints tab for a product that an Endpoint will be shown as mitigated, but an endpoint associated with a Host will not be shown as mitigated, when the Finding is inactive (but not mitigated). This change excludes inactive findings from Endpoint.host_mitigated_endpoints to make it consistent.


Maffooch (approved):

mtesauro (approved):
Approved

PR #7289, opened 2022-12-16
Add endpoints to Qualys dedupe settings

Maffooch:
issues spanning  multiple endpoints were being swallowed by reimport


mtesauro (approved):
Approved.

damiencarol (approved):

alles-klar:
@Maffooch Why do you added  the endpoints hashcode default config? We have a separate logic for endpoint deduplication: DEDUPE_ALGO_ENDPOINT_FIELDS. We may need to fix the bug here and not falling back to the "old" logic and adding the endpoint to the hashcode calculation.

Maffooch:
@alles-klar I do not see how to enable the matching of the endpoints using `DEDUPE_ALGO_ENDPOINT_FIELDS `. Is it enabled by default?

alles-klar:
Yes, but default takes only host and path for duplication.

damiencarol:
@Maffooch do you mind working on another PR that move forward this path?

PR #7291, opened 2022-12-18
Fix for overscroll issue

blakeaowens:
The extra padding added when expanding the sidebar caused the new footer to expand outside of the main page wrapper. This should (hopefully) fix the issue without impacting anything else.

Example of the issue:
overscroll-issue

[sc-177]


mtesauro (approved):
Approved

damiencarol (approved):

PR #7255, opened 2022-12-09
Translate product type

shipko:
Yet another part page translation. Now - product type pages

PR #7301, opened 2022-12-19
Fix regression in blocktranslate

coheigea:
The following commit introduced some regressions:

commit 7942d70f23c231ebaf29fa0bed82e03cdae4362d
Author: Dmitry <mukovkin@yandex.ru>
Date:   Sat Dec 10 10:04:02 2022 +0700

    Translate product type (#7255)

You can't reference a property in blocktranslate, so there was a bunch of empty text in the UI in various places.

### DIFF
--- a/dojo/templates/dojo/dashboard-metrics.html
+++ b/dojo/templates/dojo/dashboard-metrics.html
@@ -27,7 +27,7 @@
     {{ block.super }}
     <div class="row navbar-fixed-top">
         <div class="col-md-12">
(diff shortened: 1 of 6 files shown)


Maffooch:
@shipko please take a look

shipko:
Yes, @coheigea is absolutely right. 
I noticed this problem earlier, but not thinked that reason is my MR. 
I will check my previous commits and fix another places

coheigea:
@shipko I already checked all the places where blocktrans is used and fixed them in my PR, maybe some were there for a while

Maffooch (approved):

shipko:
@coheigea thanks you so much

PR #7307, opened 2022-12-20
Set Veracode "Fixed" findings as mitigated

coheigea:
We are seeing some recent failures with Veracode, where findings with "Fixed" status but no mitigation information are not being marked as mitigated.


Maffooch (approved):

mtesauro (approved):
Approved

PR #7308, opened 2022-12-20
Prepare for bugfix branch and release cadence

Maffooch:
As discussed in the moderators meeting, we would like to provide weekly bug fixes to users to prevent them having to wait a month for fixes. Please see the new branching model diagram in the diff

[sc-129]


Maffooch:
@damiencarol @alles-klar @dsever 

Does everything look clear here? Am I missing anything?

devGregA (approved):

mtesauro:
FWIW, I like this idea but wanted to give others a chance to chime in so held back on an actual 'approval'

coheigea (approved):
It will be a good change for the project, to avoid having to wait a month to pick up a fix.

damiencarol:
Seems all good.

PR #7316, opened 2022-12-23
Survey Bugs: Submit multichoice question, 500 when editing name

Maffooch:
Two issues:
- Updating the name of a survey question resulted in a 500 due to an incorrect url reverse
- Remove some javascript that hid the title of the question for multichoice questions (led to the submit button not even working)

[sc-264] [sc-122]


mtesauro (approved):
Approved

devGregA (approved):

PR #7318, opened 2022-12-23
Add last login field to users page

Maffooch:
Taking advantage of the last login field that is already stored by Django user. Also added a banner at login to specify when the last login was

image

image

image

[sc-282]


devGregA (approved):

mtesauro (approved):
Approved

PR #7327, opened 2022-12-27
Add the ability to add extra fixtures

cybercapsicum:
This Pull Request adds additional configuration options to the DefectDojo Helm chart by mounting extra fixtures as a ConfigMap volume to the initializer job. The `extra_*.json` files from this volume are copied to the `/app/dojo/fixtures/` directory to be run by the existing [`entrypoint-initializer.sh`](https://github.com/DefectDojo/django-DefectDojo/blob/33d387e7682a410b336f528078a0ade38c42c00e/docker/entrypoint-initializer.sh#L100).


cybercapsicum:
The current script will only import fixtures if the admin user doesn't exist ([ref](https://github.com/DefectDojo/django-DefectDojo/blob/33d387e7682a410b336f528078a0ade38c42c00e/docker/entrypoint-initializer.sh#L60)), which is only the case on a fresh installation.

cybercapsicum:
In my test environment, the extra fixtures were correctly imported.

helm/defectdojo/templates/extra-fixtures.yaml
```diff
@@ -0,0 +1,28 @@
...
+    {{- $models := dict }}
+    {{- $output := list }}
+    {{- range $j, $data := $fixture.data }}
+    {{- if empty $data.model }}
+    {{- printf "'model' key missing from extraFixture %d, data item %d" $i $j | fail }}
+    {{- end }}
+    {{- if empty $data.fields }}
+    {{- printf "'fields' key missing from extraFixture %d, data item %d" $i $j | fail }}
+    {{- end }}
+    {{- $_ := set $models $data.model (get $models $data.model | add1) }}
```
cybercapsicum, line 23: I would especially appreciate feedback on this line, which increments the `pk` value for every additional item in the fixture with that same `model`. Is generating the `pk` this way at all problematic? I am not very familiar with the project, so I would appreciate some help here.

helm/defectdojo/templates/extra-fixtures.yaml
```diff
@@ -0,0 +1,28 @@
...
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    helm.sh/chart: {{ include "defectdojo.chart" . }}
+data:
+  {{- range $i, $fixture := .Values.extraFixtures }}
+  {{- printf "extra_%03d.json" $i | nindent 2 }}: |
+    {{- $models := dict }}
+    {{- $output := list }}
+    {{- range $j, $data := $fixture.data }}
+    {{- if empty $data.model }}
+    {{- printf "'model' key missing from extraFixture %d, data item %d" $i $j | fail }}
```
cybercapsicum, line 18: I decided it was important to fail helm installations if extra fixtures are missing the required fields. Is this important?

cybercapsicum (commented):
Some lines I would appreciate feedback on

helm/defectdojo/values.yaml
```diff
@@ -495,6 +495,16 @@ redis:
...
 
+# Extra fixtures can be mounted to the initializer job by adding an item to the extraFixtures list:
+# NOTE  Each fixture must have a 'data' section, which must include a 'model' string and a 'fields' map
+# NOTE  These fixtures are only applied to an uninitialized database
+# extraFixtures:
+# - data:
+#   - model: dojo.system_settings
+#     fields:
+#       enable_deduplication: true
+extraFixtures: []
```
dsever, line 506: Can you provide example how extraFixuters should look like?

dsever: Is this the usecase? If it is, you should use api to bootstrap your app.

cybercapsicum: It's the simplest usecase I could think of, but users could use this to pre-populate their databases with anything from additional test types to custom surveys

cybercapsicum: The API would be better for most post-install configurations like this. I may go that direction for what I initially made this for, but this feature is still potentially useful. Whether it's useful enough to maintain is up to the you all

cybercapsicum:
Just updated to create a `Secret` rather than a `ConfigMap`, so users aren't limited by the sensitivity of data. I also added the ability to set the `pk` so users have more control. If the `pk` isn't set, it will increment like normal.

dsever:
I would like to close this PR without merging it, to keep the helm simple as possible. @alles-klar @cybercapsicum do you agree?

alles-klar:
I agree. This feature will be hard to maintain and we have the API for most use cases. 

Anyway, thanks for your work @cybercapsicum!

dsever:
Thanks @cybercapsicum for PR, Closing PR for now

PR #7329, opened 2022-12-28
Added exact product name filter

blakeaowens:
There is currently no way to filter or perform product get via the API on an exact product name. This change introduces a new field to the ProductFilter and ApiProductFilter to filter/get a product by exact name.

I do not see a point in adding this to any other objects, as you can have duplicate engagements, findings, etc. with identical names.

[sc-230]


mtesauro (approved):
Approved

devGregA (approved):

PR #7338, opened 2022-12-30
Optimize call db query for product page

shipko:
In this PR I optimized working with db. 
For /product/<id> page I decreased count from 43 to 31 queries (~28%). 

You could see screenshots below as proof
Before
Before

After
After


shipko:
Please recheck PR

Maffooch (approved):
Nice one!

mtesauro (approved):
Approved

PR #7346, opened 2023-01-02
Add gzip for nginx

shipko:
In this PR I added gzip compression by default for nginx (also TLS). This decrease packet size for download by client and increase speed load


mtesauro (approved):
Approved

Maffooch (approved):

PR #7370, opened 2023-01-05
Fixing the sla_deadline function to work with the current date and no…

coheigea:
…t the date the finding was created

I noticed that JIRAs were created with the incorrect due date - which were the date the finding was created + the number of SLA days left at the time of JIRA creation.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -2655,7 +2655,7 @@ class Finding(models.Model):
     def sla_deadline(self):
         days_remaining = self.sla_days_remaining()
         if days_remaining:
-            return self.date + relativedelta(days=days_remaining)
+            return get_current_date() + relativedelta(days=days_remaining)
         return None
 
     def github(self):


Maffooch:
I am unsure about this change. Imagine a scenario of manually adding finding from a manual pen test that spans many weeks, but were all entered into DefectDojo at the same time. This would give vulns a longer SLA than it should have.

coheigea:
@Maffooch The scenario I ran into was for an existing finding created last month, which shows the correct number of days left in the SLA as against the current date in the UI, but when I create a JIRA it sets the due date as due in a few weeks as it's calculating current SLA remaining days against the date the issue was originally found on.

Maffooch (commented):
I did not realize this function is used [solely for calculating the due date for jira](https://github.com/DefectDojo/django-DefectDojo/blob/4173bb38345dfed796303be697a04c4ec88f96e1/dojo/jira_link/helper.py#L712). I imagined it would be used in other places as well! This change makes sense to me now.

I think this would be a good bug fix branch thing. Change the head now

coheigea:
Rebased off bugfix.

PR #7373, opened 2023-01-06
In the re-importer, if the new finding is mitigated then make sure th…

coheigea:
…e old finding's endpoints are all mitigated

If you re-import a finding that was mitigated upstream, then the endpoints will stay active on the original finding. If de-duplication is enabled you see a single mitigated finding with an active endpoint.


Maffooch:
@coheigea I think this would be a good candidate for the `bugfix` branch to be released this upcoming Monday. Would you mind rebasing? If not, it can go in `dev` instead

coheigea:
@Maffooch Thanks, I've rebased now against bugfix

coheigea:
@Maffooch I rebased so that only my commit was in the PR against bugfix, but the lint helm chart failed. Did I do the right thing?

Maffooch:
Hmm. I need to add an exclusion to the helm chart test so that it does not run against this branch (similar to the dev branch)

You're the first one on this branch, so I'll need to work all these bugs out. Will start on this in a few hours

PR #7399, opened 2023-01-11
Avoid unnecessary save for dynamic findings on reimport

coheigea:
On re-import, if an open duplicate finding is found it will save the finding if the finding has no component name/version, which will be the case for dynamic findings. It should only save if the new finding has a defined component name/version.


Maffooch (approved):

mtesauro (approved):
Approved

PR #7410, opened 2023-01-12
Set CSRF_TRUSTED_ORIGINS when it is not the default value

Maffooch:
Basing this issues against dev since this upgrade is not on the bug fix branch yet

Ran into this today, but somehow did not when testing Django 4.x
```
SystemCheckError: System check identified some issues:

ERRORS:
?: (4_0.E001) As of Django 4.0, the values in the CSRF_TRUSTED_ORIGINS setting must start with a scheme (usually http:// or https://) but found []. See the release notes for details.
```

I found that not setting the setting at all resolved the issue. For some reason, setting the value to the default (an empty list) is explicitly saying "so not trust any origins" but setting it all means the opposite.


mtesauro (approved):
Approved.

devGregA (approved):

PR #7411, opened 2023-01-13
Create env var DD_JIRA_DEFAULT_ISSUE_TYPE

Gby56:
**Description**
I'd like to be able to configure Jira through the helm chart, because otherwise, the deployment requires a lot of manual set-up after deployment, it's not fully configured. This PR might need to be extended to configure much more than just the default ticket type, but maybe the whole Jira configuration should be (and with support for multiple Jira instances too perhaps?)

**Test results**
No tests yet, I hope this won't be needed for simple changes, and the helm chart would mostly just inject things in extraConfigs: {}
Ideally you extend the test suite in `tests/` and `dojo/unittests` to cover the changed in this PR.
Alternatively, describe what you have and haven't tested.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -3361,10 +3361,13 @@ class JIRA_Instance(models.Model):
                                         ('Bug', 'Bug'),
                                         ('Security', 'Security')
                                     )
-    default_issue_type = models.CharField(max_length=15,
-                                          choices=default_issue_type_choices,
-                                          default='Bug',
-                                          help_text=_('You can define extra issue types in settings.py'))
+    if hasattr(settings, 'JIRA_DEFAULT_ISSUE_TYPE'):
+        default_issue_type = settings.JIRA_DEFAULT_ISSUE_TYPE
+    else:
+        default_issue_type = models.CharField(max_length=15,
+                                            choices=default_issue_type_choices,
+                                            default='Bug',
+                                            help_text=_('You can define extra issue types in settings.py'))
     issue_template_dir = models.CharField(max_length=255,
                                       null=True,
                                       blank=True,
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -220,7 +220,7 @@ env = environ.Env(
     # DD_FEATURE_FINDING_GROUPS feature is moved to system_settings, will be removed from settings file
     DD_FEATURE_FINDING_GROUPS=(bool, True),
     DD_JIRA_TEMPLATE_ROOT=(str, 'dojo/templates/issue-trackers'),
-    DD_JIRA_DEFAULT_ISSUE_TYPE=(str,'Bug'),
+    DD_JIRA_DEFAULT_ISSUE_TYPE=(str, 'Bug'),
     DD_TEMPLATE_DIR_PREFIX=(str, 'dojo/templates/'),
 
     # Initial behaviour in Defect Dojo was to delete all duplicates when an original was deleted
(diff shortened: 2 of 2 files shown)


Maffooch:
I agree with this change. In my experience, changing the list of jira issue types leads to migration errors on every upgrade because Django thinks the database needs to change. Very annoying.

Gby56:
I think for the helm deployment model, we would have to create new values in the configmap for example, and then, pass them to the initializer job, which already is used to apply fixtures as mentioned by @valentijnscholten 
We have to modify the `entrypoint-initializer.sh` script to support  some input, because the fixtures are currently built directly inside the container image via a COPY, which won't be possible in Helm.

Gby56:
@Maffooch `because Django thinks the database needs to change` what do you mean by that ? Would an initializer job fix the issue ?... I'm thinking about using the command `python3 manage.py loaddata <fixture.json>`, but I'm not sure that would fix your issue, just making sure.
I guess Django "thinks the database needs to change" is after the initializer job runs, you start the django pod and there it crashes ? I have never seen such a bug. Could you clarify, so I make sure my idea would prevent the bug ?

Maffooch:
I think the issue lies with the default choices being a column in the database. It makes modification of the setting very brittle. Take these lines for example:

https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/models.py#L3353-#L3367

If the field were just a regular char field with no choices, there is an unlimited amount of customization there. The only way to gatekeep that would be in the form in the UI/API to only list and accept fields defined in the system settings var `JIRA_ISSUE_TYPE_CHOICES_CONFIG`

For more context,  this is the issue I was talking about:

I add a new type of jira issue:
```
JIRA_ISSUE_TYPE_CHOICES_CONFIG += (
    ('New Type', 'New Type'),
)
```

I restart the server and get this error:
```
Migrations for 'dojo':
init  |   dojo/db_migrations/0178_alter_jira_instance_default_issue_type.py
init  |     - Alter field default_issue_type on jira_instance
init  | Traceback (most recent call last):
init  |   File "/app/manage.py", line 11, in <module>
init  |     execute_from_command_line(sys.argv)
init  |   File "/usr/local/lib/python3.11/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
init  |     utility.execute()
init  |   File "/usr/local/lib/python3.11/site-packages/django/core/management/__init__.py", line 413, in execute
init  |     self.fetch_command(subcommand).run_from_argv(self.argv)
init  |   File "/usr/local/lib/python3.11/site-packages/django/core/management/base.py", line 354, in run_from_argv
init  |     self.execute(*args, **cmd_options)
init  |   File "/usr/local/lib/python3.11/site-packages/django/core/management/base.py", line 398, in execute
init  |     output = self.handle(*args, **options)
init  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
init  |   File "/usr/local/lib/python3.11/site-packages/django/core/management/base.py", line 89, in wrapped
init  |     res = handle_func(*args, **kwargs)
init  |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
init  |   File "/usr/local/lib/python3.11/site-packages/django/core/management/commands/makemigrations.py", line 190, in handle
init  |     self.write_migration_files(changes)
init  |   File "/usr/local/lib/python3.11/site-packages/django/core/management/commands/makemigrations.py", line 228, in write_migration_files
init  |     with open(writer.path, "w", encoding='utf-8') as fh:
init  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
init  | PermissionError: [Errno 13] Permission denied: '/app/dojo/db_migrations/0178_alter_jira_instance_default_issue_type.py'
```

The issue is not necessarily that the migration could not be created, but if it were created, the next time I were to upgrade, I would have two 0178 migrations and then sorting that out would be a nightmare.

Issue #7403, opened 2023-01-11
Nessus reimport mitigating endpoint that still vunerable


aflima:
Nessus reimport mitigating endpoint that still vunerable
When a scan made via nessus is reimported, defectdojo is mitigating an endpoint that is not effectively mitigated and make finding open with all endpoints wrongly mitigated.

**Steps to reproduce**
1. Create an engagment and import .nessus scan result in some test;
2. After import, reupload scan results
[Machine.zip](https://github.com/DefectDojo/django-DefectDojo/files/10392960/Machine.zip)
 the same .nessus file;
3. You will see that for some reason Dojo mitigate some endpoints even using the same file.

**Expected behavior**
Dojo needs to correctly validate which endpoints were mitigated at reimport time.

**Deployment method** *(select with an `X`)*
- [x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: 
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.1 LTS
Release:        22.04
Codename:       jammy
 - DefectDojo version v. 2.18.1 ( release mode )

**Sample scan files**
[Machine.zip](https://github.com/DefectDojo/django-DefectDojo/files/10392970/Machine.zip)

**Screenshots**
![image](https://user-images.githubusercontent.com/2230195/211824133-3748000d-5d3e-4f11-9cd7-7eaee00bf5c5.png)
![image](https://user-images.githubusercontent.com/2230195/211824347-070b75d5-28e7-4f95-9b52-ae1117604bf4.png)
![image](https://user-images.githubusercontent.com/2230195/211824465-5d0a8ecb-c8f6-4c24-82bd-e317d014e9d0.png)
![image](https://user-images.githubusercontent.com/2230195/211824569-d1b364f8-ac39-4b2f-8f27-f746449d17e4.png)
![image](https://user-images.githubusercontent.com/2230195/211824684-dcc091f1-cbb5-4a36-bdfa-a8fabb2db196.png)
![image](https://user-images.githubusercontent.com/2230195/211824801-1ce82698-cef1-4985-8276-ff0f56056277.png)

coheigea:
I found the bug here: https://github.com/DefectDojo/django-DefectDojo/pull/7416

coheigea:
This can be closed as it's fixed in 2.18.2.

sixpacksecurity:
[...]

PR #7416, opened 2023-01-13
Clean unsaved endpoints before comparing them to existing endpoints

coheigea:
A fix for https://github.com/DefectDojo/django-DefectDojo/issues/7403

The problem here is that the port is a String in the Endpoint, something that is cleaned up by clean(). Comparison fails compared to the existing endpoint as a result, and the endpoint gets incorrectly mitigated on reimport.

### DIFF
--- a/dojo/importers/reimporter/reimporter.py
+++ b/dojo/importers/reimporter/reimporter.py
@@ -73,6 +73,10 @@ class DojoDefaultReImporter(object):
             if service:
                 item.service = service
 
+            if item.dynamic_finding:
+                for e in item.unsaved_endpoints:
+                    e.clean()
+
             item.hash_code = item.compute_hash_code()
             deduplicationLogger.debug("item's hash_code: %s", item.hash_code)
 


kiblik:
And I suppose we should clean them also [in importer](https://github.com/DefectDojo/django-DefectDojo/blob/ffdc12563be93fc27b02644be47de6bee31a85bf/dojo/importers/importer/importer.py#L73-L104) (not only in reimporter)

coheigea:
@kiblik Done, thanks. The importer is already cleaned here - https://github.com/DefectDojo/django-DefectDojo/blob/1bb23ddc071bfa947c9e9e840ea428e6cbfb2984/dojo/importers/utils.py#L129

kiblik (approved):

Maffooch (approved):

coheigea:
@Maffooch I don't see this change in the 2.18.2 release notes? https://github.com/DefectDojo/django-DefectDojo/releases/tag/2.18.2

Maffooch:
Corrected

Issue #7429, opened 2023-01-16
In API findings/{id}/close on swagger and dexternal api return status 500 'Finding' object has no attribute 'endpoint_status'


poteddy:
v. 2.16.2 ( release mode )

When using api to close a finding. The default swagger throws a 500 error
sent using api swagger
{
  "is_mitigated": true,
  "mitigated": "2023-01-16T15:39:09.809929Z"
}

[16/Jan/2023 16:30:21] ERROR [dojo.api_v2.exception_handler:32] 'Finding' object has no attribute 'endpoint_status'
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/app/./dojo/api_v2/views.py", line 563, in close
    endpoints_status = finding.endpoint_status.all()
AttributeError: 'Finding' object has no attribute 'endpoint_status'
[16/Jan/2023 16:30:21] ERROR [django.request:224] Internal Server Error: /api/v2/findings/17080/close/
[pid: 1|app: -|req: -/-] 10.8.159.213 (admin) {64 vars in 1359 bytes} [Mon Jan 16 16:30:21 2023] POST /api/v2/findings/17080/close/ => generated 59 bytes in 61 msecs (HTTP/1.0 500) 7 headers in 212 bytes (1 switches on core 0)

poteddy:
Ok I maybe incorrect but I believe merged pr #6193 affects this.

coheigea:
PR - https://github.com/DefectDojo/django-DefectDojo/pull/7433

coheigea:
Pr merged

coheigea:
This can be closed

mtesauro:
Thanks @coheigea :+1:

PR #7433, opened 2023-01-17
Use Finding.status_finding instead of endpoint_status

coheigea:
A fix for https://github.com/DefectDojo/django-DefectDojo/issues/7429

Looks like some instances were missed as part of https://github.com/DefectDojo/django-DefectDojo/pull/6193/files


Maffooch (approved):
Thanks for taking this on @coheigea

devGregA (approved):

PR #7444, opened 2023-01-18
Improve transaction behaviour on imports

### LINKED ISSUE #6217: Duplicates deletion error
**Bug description**
I found the issue with duplicates deletion. When I enabled this feature I started to face such errors sometimes during report uploading:
`{"message":"Internal server error, check logs for details"}`
In django logs I found the error:
```
ERROR [dojo.api_v2.exception_handler:32] insert or update on table "dojo_test_import_finding_action" violates foreign key constraint "dojo_test_import_fin_finding_id_28fe8e2d_fk_dojo_find"
DETAIL: Key (finding_id)=(742929) is not present in table "dojo_finding".
django.db.utils.IntegrityError: insert or update on table "dojo_test_import_finding_action" violates foreign key constraint "dojo_test_import_fin_finding_id_28fe8e2d_fk_dojo_find"
```
As I understand DD tried to find a finding that was deleted as a duplicate. 

@valentijnscholten commented on Slack:
"The job to delete duplicates runs every minute, so if you have an import that runs longer than that, there may be a race condition that newly imported findings which are duplicates of existing findings are deleted before the import is complete. I can't remember exactly what the transactional model is in django, but probably  not everything runs in a transaction so duplicates become visible to the job that deletes duplicates before the import is completed."

**Steps to reproduce**
1. Create a project.
2. Enable **Deduplicate findings** and **Delete duplicates** features in settings.
3. Import a report with a lot of findings (more than 100).
4. Import a few reports with duplicates to the project.
5. After some imports you will face an error in logs: 
`{"message":"Internal server error, check logs for details"}`

**Expected behavior**
The report should be imported successfully.

- [ ] Docker Compose
- [`X`] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version 2.9.1

**Logs** 
```
ERROR [dojo.api_v2.exception_handler:32] insert or update on table "dojo_test_import_finding_action" violates foreign key constraint "dojo_test_import_fin_finding_id_28fe8e2d_fk_dojo_find"
DETAIL: Key (finding_id)=(742929) is not present in table "dojo_finding".
django.db.utils.IntegrityError: insert or update on table "dojo_test_import_finding_action" violates foreign key constraint "dojo_test_import_fin_finding_id_28fe8e2d_fk_dojo_find"
```

fhoeborn:
This changes the transaction behavior for imports / reimports of test results to fix #6217.
A test import or reimport now runs in one transaction instead of committing on every .save() (e.g. finding creation).


Maffooch:
HI @fhoeborn this is a good change. Don't worry about the integration tests failing until you move this PR out of draft. The tests are a finicky and the failures are not due to your change

fhoeborn:
Hi @Maffooch, thanks for the warm words :-) to be honest, I think in this case the integration tests catched something - for some reason this change seems to break the deduplication function. I'll take another look into that

maxi-bee:
@fhoeborn you should sign your commits shame on you

fhoeborn:
> @fhoeborn you should sign your commits shame on you

I'll do with some self punishment ;-) Fixed that now.

dojo/importers/reimporter/reimporter.py
```diff
@@ -300,15 +302,16 @@ def process_parsed_findings(self, test, parsed_findings, scan_type, user, active
...
                     jira_helper.push_to_jira(findings[0].finding_group)
                 else:
                     jira_helper.push_to_jira(findings[0])
 
-        if is_finding_groups_enabled() and push_to_jira:
-            for finding_group in set([finding.finding_group for finding in reactivated_items + unchanged_items if finding.finding_group is not None]):
-                jira_helper.push_to_jira(finding_group)
+            if is_finding_groups_enabled() and push_to_jira:
+                for finding_group in set([finding.finding_group for finding in reactivated_items + unchanged_items if finding.finding_group is not None]):
+                    jira_helper.push_to_jira(finding_group)
```
damiencarol, line 314: I'm not sure you should modify this. If really sending data to JIRA break because some data are not persistented, this part should be extracted and done after `post_process_finding_save` method. but It could not be the case.

sameeraksc:
we also have the same issue. any plan to fix it soon :-(

fhoeborn:
Maybe an update from me:
I changed the transaction behaviour to do the import / reimport in one transaction (instead of commiting directly after every .save() operation) - this also means, that some actions should happen after the (whole) transaction is commited (e.g. https://github.com/fhoeborn/django-DefectDojo/blob/fix-race-condition/dojo/models.py#L2794)
The Unittests are using TestCase as base class instead of TransactionTestCase for performance reasons - this also means that the transaction.on_commit hooks are not executed by default. To do so and not sacrifice performance of the whole unittest suite, I used the testing tools of django to execute those callbacks (https://docs.djangoproject.com/en/4.1/topics/testing/tools/#django.test.TestCase.captureOnCommitCallbacks).
This worked perfectly well for all dedupe tests - but for the JIRA unittests it seems from my tests, that the callbacks get scheduled, but never executed. The main difference seems to be, that they inherit from APITestCase from the rest_framework (and VCRTestCase from vcr_unittest) - though I see no direct reason, why this should break it.

I am currently stuck with those failing Unittests and a bit out of ideas, what might be the reason for the failure. If anyone has a good idea, what might be the reason - I am happy to hear about those.

dojo/importers/reimporter/reimporter.py
fhoeborn: Yeah, that change was bogus - a leftover from some desperate testing

Maffooch:
Closing due to stale

valentijnscholten:
It's hard to oversee the overall implications of this change. All postprocessing of findings, pushing to jira, deduplication is al delayed until all findings have been created. There might be implications on other import running at the same time waiting to locks, or deduplications from other imports seeing or not seeing current data. In other products I sometimes see that imports are done inside transactions, but with a certain batch size to prevent the transaction running for hours for example.

Issue #7367, opened 2023-01-05
Test tags are not updated when using reimport_scan


gietschess:
**Bug description**
When using the reimport_scan api call and updating a test tag value for a existing test its value will not be updated.

**Steps to reproduce**
1. Use reimport_scan to add a test containing any test tag
2. Use reimport_scan again with the same scan and change the test tag value

**Expected behavior**
Test tag value is updated to the new value.

**Deployment method**
- [ ] Docker Compose
- [x] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: [Amazon Linux 2, k8s 1.21]
 - DefectDojo version 2.18.0

**Sample api calls**
```bash
curl -X 'POST' \
  'https://defectdojo.local/api/v2/reimport-scan/' \
  -H 'accept: application/json' \
  -H 'Content-Type: multipart/form-data' \
  -F 'minimum_severity=Info' \
  -F 'active=true' \
  -F 'verified=true' \
  -F 'do_not_reactivate=false' \
  -F 'scan_type=Trivy Scan' \
  -F 'file=@alpine312.txt;type=application/json' \
  -F 'product_type_name=container' \
  -F 'product_name=os' \
  -F 'engagement_name=alpine' \
  -F 'test_title=alpine' \
  -F 'auto_create_context=true' \
  -F 'deduplication_on_engagement=true' \
  -F 'push_to_jira=false' \
  -F 'close_old_findings=true' \
  -F 'close_old_findings_product_scope=false' \
  -F 'version=3.12' \
  -F 'service=alpine' \
  -F 'tags=image_tag=3.12' \
  -F 'create_finding_groups_for_all_findings=true'
```
```bash
curl -X 'POST' \
  'https://defectdojo.local/api/v2/reimport-scan/' \
  -H 'accept: application/json' \
  -H 'Content-Type: multipart/form-data' \
  -F 'minimum_severity=Info' \
  -F 'active=true' \
  -F 'verified=true' \
  -F 'do_not_reactivate=false' \
  -F 'scan_type=Trivy Scan' \
  -F 'file=@alpine317.txt;type=application/json' \
  -F 'product_type_name=container' \
  -F 'product_name=os' \
  -F 'engagement_name=alpine' \
  -F 'test_title=alpine' \
  -F 'auto_create_context=true' \
  -F 'deduplication_on_engagement=true' \
  -F 'push_to_jira=false' \
  -F 'close_old_findings=true' \
[...]

PR #7446, opened 2023-01-18
Handle test tags when reimporting with reimport_scan api

gietschess:
**Description**

see issue #7367

tags are created while using the reimport_api and creating a new, not already existing test.
when the test exists the newly set tags are ignored.

with this pr tags will be overwritten when tags are set while reimporting a existing test object.
if no tags are set the existing tags will remain.

**Test results**

the integration_tests locally ran without any errors.

**Documentation**

i've specified the behaviour in the api help text.


Maffooch (approved):
Huh I guess fourth time is a charm 😂

mtesauro (approved):
Approved

PR #7447, opened 2023-01-18
Add the old finding in the reimporter to the list of unchanged findin…

coheigea:
…gs if risk acceptance matches

If you have a risk accepted finding, on a re-import it will get additionally marked as mitigated, which isn't necessary. Instead if risk acceptance or false positive or out of scope matches with the new finding, we should add it to the list of unchanged findings.


Maffooch (approved):

mtesauro (approved):
Approved

Issue #7354, opened 2023-01-03
Adding Betterscan.io to "Scan type" on "import scan results" screen


marcinguy:
**Is your feature request related to a problem? Please describe**
Is it possible to add Betterscan.io to "Scan type" on "import scan results" screen. Betterscan produces SARIF, so it can be mapped to it directly I hope. I think users can also find it useful to see Betterscan.io scan type vs generic SARIF type. Can improve usability. Was asked by users about it at Betterscan project, hence write here.

**Describe the solution you'd like**
As a user I would like to see "Scan type" being also Betterscan.io on "import scan results" screen

I can also make a PR for this (if I figure it out)

**Describe alternatives you've considered**
Not having Betterscan.io name there. But not everyone will know that SARIF will also work for Betterscan.

**Additional context**

I mean "Scan type" field on this screen

![image](https://user-images.githubusercontent.com/20355405/210394447-9aa0af1d-86d8-4caf-9b4a-b361eab49e18.png)

With guidance I can open PR for this.

Thanks,

marcinguy:
Made a PR for it. Hope it is fine. Thanks,

mtesauro:
#7398  for reference

manuel-sommer:
@mtesauro I guess we can close this issue.

PR #7452, opened 2023-01-18
Adding Betterscan.io scan type to SARIF parser

marcinguy:
**Description**

Adding Betterscan.io as a new "Scan Type"

https://github.com/DefectDojo/django-DefectDojo/issues/7354

Used current SARIF implementation

**Test results**

Use SARIF parsing, so no need for test

**Documentation**

Not needed, since explained in the UI

### DIFF
--- a/docs/content/en/integrations/parsers.md
+++ b/docs/content/en/integrations/parsers.md
@@ -85,6 +85,9 @@ Azure Security Center recommendations can be exported from the user interface in
 
 JSON report format
 
+### Betterscan
+
+Betterscan findings can be exported in SARIF from Betterscan and imported via SARIF format.
 
 ### Blackduck API
 
--- a/dojo/tools/sarif/parser.py
+++ b/dojo/tools/sarif/parser.py
@@ -18,12 +18,14 @@ class SarifParser(object):
     """
 
     def get_scan_types(self):
-        return ["SARIF"]
+        return ["SARIF", "Betterscan.io"]
 
     def get_label_for_scan_types(self, scan_type):
         return scan_type  # no custom label for now
 
     def get_description_for_scan_types(self, scan_type):
+        if scan_type == "Betterscan.io":
+            return "Import Betterscan.io report. Currently only SARIF Format is supported."
         return "SARIF report file can be imported in SARIF format."
 
     def get_findings(self, filehandle, test):


Maffooch (dismissed):

damiencarol:
1) @marcinguy could you add a SARIF report generated by Betterscan.io ?
2) I still think this PR is not a good way to proceed. Our SARIF Parser take into account the TestType if it's define in the report.
This give this label for ex for a report generated by LGTM.com:
image
You should work on your side to generate report data that clearly show that Betterscan.io generated the report.

Something like this:
```JSON
{
   "$schema":"https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
   "version":"2.1.0",
   "runs":[
      {
         "tool":{
            "driver":{
               "name":"Betterscan.io",
               "organization":"Betterscan",
               "version":"1.2.9",
 <...>
```

It's better to improve your soft and let DefectDojo do the job instead of relying on a specific code in DefectDojo to support better Betterscan.io reports.

The maximum we can do is specify in the documentation (only!) tha we support Batterscan.io through the SARIF importer.
 
I'm rejecting this PR if this PR change the code instead of only improving documentation.

damiencarol (changes requested):
I'm rejecting this PR because this PR change the code instead of only improving documentation.

marcinguy:
Hi @damiencarol 

The concept was to show "Betterscan.io" as an option in import scan type (Adding Betterscan.io to "Scan type" on "import scan results" - https://github.com/DefectDojo/django-DefectDojo/issues/7354)

So it is here on the list/dropbox/combobox:

https://user-images.githubusercontent.com/20355405/210394447-9aa0af1d-86d8-4caf-9b4a-b361eab49e18.png

1. Sample SARIF (renamed to sample_sarif.txt, since could not upload *.sarif)

[sample_sarif.txt](https://github.com/DefectDojo/django-DefectDojo/files/10528040/sample_sarif.txt)

2.  Yes I generate a valid SARIF. It has those values. So it should show properly.

```
{
  "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
  "version": "2.1.0",
  "runs": [
    {
      "tool": {
        "driver": {
          "name": "Betterscan",
          "informationUri": "https://betterscan.io",
          "semanticVersion": "0.9.9",
          "rules": [
            {
              "id": "0",
              "name": "BetterscanRule",
              "shortDescription": {
                "text": "vmdetect"
```

I hope this provides answers to your questions.

To repeat: This PR is about just this combo box (Scan type)

marcinguy:
Well, do as you wish.

It shows after import:

image

**BUT** not in the scan type before the import:

image

It would have to be selected as SARIF type.

The whole concept was to have it there and be listed in here:

image


If you will not do it, feel free to close.

Maffooch:
This PR may be served in finding a different way to pull the newly created serif Test Types into that drop down list after they are created by the serif parser

damiencarol:
As discussed, better to close it for now.

Issue #7437, opened 2023-01-17
AWS Security Hub Scan - Bug on reupload


gnxbr:
**Bug description**
When I reupload a scan of type **AWS Security Hub Scan**, even with the same findings of the previous version, I get the message:
"_No findings were added/updated/closed/reactivated as the findings in Defect Dojo are identical to those in the uploaded report._", but the status of the findings are changed to "Inactive, Mitigated"

**Steps to reproduce**
1. Get the sample [here](https://raw.githubusercontent.com/DefectDojo/sample-scan-files/master/awssecurityhub/example.json)
2. Create a test of type "AWS Security Hub Scan", inside some product engagement
3. Upload the sample
4. Check the status of the findings: "Active"
5. Reupload the sample. You will get the message: "_No findings were added/updated/closed/reactivated as the findings in Defect Dojo are identical to those in the uploaded report._"
7. Check the status: "Inactive, Mitigated"

**Expected behavior**
When the same findings are reuploaded, the status of the findings should not be changed

**Deployment method**
- [x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Amazon Linux 2 (Linux 5.10.149-133.644.amzn2.x86_64)
 - DefectDojo version (see footer) or commit message: [2022-12-06 12:08:52 -0600] 4173bb383: Merge pull request #7244 from DefectDojo/release/2.17.0 [ (HEAD -> master, tag: 2.17.0, origin/master, origin/HEAD)]

coheigea:
Can you try with the latest 2.18.2 release? There were a few bugs fixed since 2.17.0 in the reimporter.

gnxbr:
I found the bug, it is [in this line](https://github.com/DefectDojo/django-DefectDojo/blob/release/2.18.2/dojo/tools/awssecurityhub/parser.py#L48). It was not fixed yet.

coheigea:
Actually I think the bug is that in the first import if is_Mitigated is set to true, then active should be set to False. As "Compliance" has Passed according to the imported json, then the issues should be marked as inactive as well as mitigated.

If you agree I will submit a PR.

gnxbr:
Locally, I solved the problem doing it:

`if finding.get('Compliance', {}).get('Status', 'PASSED') == 'PASSED':`

In this way, the finding will be Mitigated whenever:
* Finding status changed to "PASSED"
* Finding not present in the reupload

[...]

PR #7454, opened 2023-01-19
If findings are mitigated for AWS Security Hub then set them to inactive

coheigea:
A fix for https://github.com/DefectDojo/django-DefectDojo/issues/7437

It fixes two bugs in the AWS Security Hub parser:

1. When a finding is mitigated, set it to inactive.
2. The code was setting all issues to mitigated!


Maffooch (approved):

mtesauro (approved):
Approved

PR #7457, opened 2023-01-19
Update localization in all PRs

kiblik:
This PR force developer to run `./manage.py makemessages` on all PRs (as we run flake8 or unit tests).

If any change can affect localization but the developer does not commit also update of `django.po`, test fails.

Commit with missing update of `django.po` like [this](https://github.com/DefectDojo/django-DefectDojo/pull/7457/commits/bd80b35cf23d9147c68f09768d7e3a5ddbdeb1bc) will fail like [this](https://github.com/DefectDojo/django-DefectDojo/actions/runs/3960062363/jobs/6783689824#step:7:205).

### DIFF
--- a/Dockerfile.django
+++ b/Dockerfile.django
@@ -58,6 +58,7 @@ RUN \
     postgresql-client \
     # libcurl4-openssl-dev is required for installing pycurl python package
     libcurl4-openssl-dev \
+    gettext \
     && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists && \
--- a/docker/entrypoint-unit-tests-devDocker.sh
+++ b/docker/entrypoint-unit-tests-devDocker.sh
@@ -13,7 +13,7 @@ unset DD_DATABASE_URL
 # Unset the celery broker URL so that we can force the other DD_CELERY_BROKER settings
 unset DD_CELERY_BROKER_URL
 
-python3 manage.py makemessages -l en --no-obsolete --extension html,txt,py,tpl --add-location file
+python3 manage.py makemessages -l en --no-obsolete --extension html,txt,py,tpl --add-location file --ignore=docs/*
 python3 manage.py makemigrations dojo
 python3 manage.py migrate
 
--- a/docker/entrypoint-unit-tests-devDocker.sh
+++ b/docker/entrypoint-unit-tests-devDocker.sh
@@ -13,6 +13,7 @@ unset DD_DATABASE_URL
 # Unset the celery broker URL so that we can force the other DD_CELERY_BROKER settings
 unset DD_CELERY_BROKER_URL
 
+python3 manage.py makemessages -l en --no-obsolete --extension html,txt,py,tpl
 python3 manage.py makemigrations dojo
 python3 manage.py migrate
 
--- a/docker/entrypoint-unit-tests-devDocker.sh
+++ b/docker/entrypoint-unit-tests-devDocker.sh
@@ -13,7 +13,7 @@ unset DD_DATABASE_URL
 # Unset the celery broker URL so that we can force the other DD_CELERY_BROKER settings
 unset DD_CELERY_BROKER_URL
 
-python3 manage.py makemessages -l en --no-obsolete --extension html,txt,py,tpl
+python3 manage.py makemessages -l en --no-obsolete --extension html,txt,py,tpl --add-location file
 python3 manage.py makemigrations dojo
 python3 manage.py migrate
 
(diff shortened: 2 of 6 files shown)


kiblik:
@shipko, how do you see this?

kiblik:
We should add `--add-location file`. It will remove line numbers - because it could require too many changes in `django.po` (almost in every PR).

CharlieSears:
I dig it! I'm a big fan of adding `--add-location file` as well. Though it feels like hijacking the unit tests for source control. If someone is just running unit tests, won't they get a failure for localization when that may be coming after they've solved their test issue?

Would it be possible to add an action to detect a lack of localization updates? Like [test-helm-chart](https://github.com/DefectDojo/django-DefectDojo/blob/master/.github/workflows/test-helm-chart.yml)?

kiblik:
> I dig it! I'm a big fan of adding `--add-location file` as well. Though it feels like hijacking the unit tests for source control. If someone is just running unit tests, won't they get a failure for localization when that may be coming after they've solved their test issue?

Unittests designed to run locally (`docker/entrypoint-unit-tests-devDocker.sh` and `docker/unit-tests.sh`) will only update the content of `django.po`. So it will not fail (you can run unit test without updating `django.po`), it will prepare changes in `django.po` and then you can commit prepared changes.
> 
> Would it be possible to add an action to detect a lack of localization updates? Like [test-helm-chart](https://github.com/DefectDojo/django-DefectDojo/blob/master/.github/workflows/test-helm-chart.yml)?

Unittest which runs as GitHub action on PRs (`docker/entrypoint-unit-tests.sh`) will fail if there is missing update of `django.po`.

This behaviour copies behaviour with `manage.py makemigrations`. It will generate migration and not fail if you run it locally however it fails if it was committed and pushed with `django.po` update.

We are not able to identify if the developer forgot to add `_(...)` or `{% trans "..." %}` because it is hard to determine. If he/she/they used `_(...)` or `{% trans "..." %}`, we are able to detect if an update of `django.po` wasn't performed.

Does it answer questions? Is it acting as you asked or did I miss something?

shipko:
@kiblik a great idea!

I absolutely agree with add `--add-location file` - it optimize our process. 

But I think this idea that we should require to update locale file on every changes `_(...)` or `{% trans "..." %}` is not so good. Because:
1. We raise the entry threshold for providing PR, requiring additional manipulations
2. We add more commit history that raise size of git repo
3. We raise a time to market metric for some PR

My suggestion is to make a new repo `defectDojo-translation` that will run git workflow on every PR `django-DefectDojo`. 
Advantages of this approach:
1. Contributors will not notice any changes in the PR process
2. We can support many another languages without effect (many-many commits) on main repo
3. By default en locale is available in code. If any members will require another locale it can download from `defectDojo-translation` repo

For examples: for my russian locale on every release update I just will copy ru folder and run project. It simple and fast.

devGregA:
I very much like @shipko's idea and approach and can have someone from the core team help facilitate any new repos @Maffooch

kiblik:
I can only agree with these arguments. However, so far I have no knowledge of how to create cross-repository GitHub actions. Somebody with knowledge of designing these think would be helpful here.

Maffooch:
Here is an example of cross repo GHA calls
https://github.com/DefectDojo/django-DefectDojo/blob/master/.github/workflows/submodule-update.yml

kiblik:
@shipko, do I understand correctly that you would like to remove `dojo/locale/en/LC_MESSAGES/django.po` from this repo?

Btw, what about adding `https://github.com/DefectDojo/defectDojo-translation` as submodule "mounted" to `dojo/locale`? Just not sure, when it should be cloned/updated. During building image? It can be too soon because somebody may add new phrases after the official release.

shipko:
@kiblik yes, I suggest to remove dojo/locale/*/*/django.po files from this repo.

Updating the base image when updating the localization file is our ideal pipeline. But is really need us? 

We can update submodule for every release (maybe also bugfix release).
If somebody will want to update locale file in that moment he will can update it manually.

 When container is starting we need to execute `python3 manage.py compilemessages` in entrypoint-initializer.sh

peeqonet:
Any update?
I'm ok to do it manually and i can change the source on my own environment but none of the existing solutions work.

Maffooch:
Closing due to stale

kiblik:
This issue became stale because we have been waiting for the reaction of core developers about the next steps [here](https://github.com/DefectDojo/django-DefectDojo/pull/7457#issuecomment-1398816861) and in [Slack](https://owasp.slack.com/archives/C014H3ZV9U6/p1694030448468349?thread_ts=1694016810.899579&cid=C014H3ZV9U6) as well.

PR #7470, opened 2023-01-20
Make Finding verified field default to false

Maffooch:
When using a tool that does not supply any status on active/verified, I noticed that all findings being imported were defaulting to verified. This is a poor default states as verified is intended to mean that the finding has been verified by a human as legit.

This PR changes the default value of verified from true to false


coheigea:
@Maffooch  "This PR changes the default value of verified from false to true" - From true to false you mean?

Maffooch:
Yup definitely meant from true to false 😅

devGregA (approved):

PR #7472, opened 2023-01-23
A fix to update tags if specified in the re-importer

coheigea:
A fix for https://github.com/DefectDojo/django-DefectDojo/issues/7367


gietschess:
hi @coheigea ,
i did something similar in pr #7446 which was merged to dev instead of bugfix.

coheigea:
OK @gietschess , will close then.

PR #7479, opened 2023-01-23
Disable GHA unit tests on push to branch, limit only to PR

Maffooch:
I have been noticing that unit tests have been running twice on during the merge process:
1. In the PR itself on each commit
2. After merging from contributor branch X to dev/bugfix/master

This is redundant and clogs the actions queue. 

I suspect this may also fix the helm lint tests on master since the primary reason for failing is that the test is running post merge to master (checking master helm version for an increased helm version after it has already been bumped)


damiencarol:
Good catch.

PR #7482, opened 2023-01-23
Generic: Handle incorrect json

kiblik:
Refuse to process incorrect generic JSON - if some keys are missing or if it is not allowed to modify them


dojo/tools/generic/parser.py
```diff
@@ -52,14 +53,30 @@ def _get_test_json(self, data):
...
             # check for required keys
             required = ['title', 'date', 'severity', 'description']
             missing = []
             for field in required:
-                if field not in data:
+                if field not in item.keys():
                     missing.append(field)
             if missing:
-                raise Exception(f"Required fields are missing: {missing}")
+                raise ValidationError(f"Required fields are missing: {missing}")
```
damiencarol, line 79: Please use `ValueError` instead. Using a Django error could be misleading.
```suggestion
                raise ValueError(f"Required fields are missing: {missing}")
```

damiencarol (changes requested):

dojo/tools/generic/parser.py
kiblik: I see your point. But if we will use only `ValueError`, the user will receive `HTTP-500` with 
```json
{
  "message": "Internal server error, check logs for details"
}
```
and he/she/they will have to contact the admin for logs to identify small mistakes.
In the case of `ValidationError`, the user will be able to see the error message in the response and identify the problem and solve it.

What about using `ValueError`s in parsers and "rewrap" them to `ValidationError` in (re)importer?

Maffooch (approved):

devGregA (approved):

blakeaowens (approved):
LGTM

mtesauro (approved):
Approved

mtesauro:
@damiencarol It looks like you're the last person with proposed changes to this.

If it's good by you, feel free to merge.

damiencarol (approved):

damiencarol:
@kiblik the final version is good!

PR #7487, opened 2023-01-24
Deduplication hashcode fields configuration UI

Gby56:
**Description**

This is a first try to implement an actual UI/API to configure deduplication hash code generation.
It could enable people to play around with their duplicate matching and maybe improve things a bit more.
Sometimes it's obvious that a finding is a duplicate of another scanner, and you can tell that it's just 1 field that messes up the hash matching.

**Improvements**
- Maybe improve the first-loading mechanism, it's a bit of a dirty trick, on the first time you load the page of dedupe configs, I load the settings.dist.py and see if they are missing, and create one for each that has been configured. It won't create one for a scanner that isn't configured specifically to use x or y fields. (but should give the user the whole list of scanners, if they'd like to set one up that was completely "default")


devGregA:
@Gby56  I have some bad news https://documentation.defectdojo.com/proprietary_plugins/tuner/#deduplication-settings, but I may have a solution. Would you please email me? greg@defectdojo.com

PR #7491, opened 2023-01-25
New env var DD_HASHCODE_FIELDS_PER_SCANNER

Gby56:
**Description**

Simply a new env var that enables users to hardcode their hashcode fields another way, which will be possible for Kubernetes deployments too.
It copies from `DD_CELERY_BROKER_TRANSPORT_OPTIONS`, because it's loading a JSON String from the env, into a dict, and replaces the settings.dist.py values with the ones in the dict

**Test results**

Not much testing to add I think, it works well, no exception handling needed because we don't want the startup to work if the value isn't valid.

**Documentation**

A lot of env vars are kinda undocumented, I think I'll push another commit to mention it.

### DIFF
--- a/docs/content/en/usage/features.md
+++ b/docs/content/en/usage/features.md
@@ -152,7 +152,10 @@ DEDUPE_ALGO_LEGACY
 
 The hash_code computation can be configured for each parser using the
 parameter `HASHCODE_FIELDS_PER_SCANNER` in
-`settings.dist.py`.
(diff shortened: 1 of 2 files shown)


Gby56:
Might be a bug... the variable `HASHCODE_FIELDS_PER_SCANNER` doesn't seem like it's modified

Maffooch (approved):

mtesauro (approved):
Approved

PR #7498, opened 2023-01-26
For findings that are out of SLA display them as negative values so t…

coheigea:
…hat we can sort findings according to SLA

I'm not sure @Maffooch if fixes at the end of the month should go against dev or bugfix?

The problem here is that if you want to sort findings by SLA in the dashboard, the out of SLA findings may appear buried in the finding list as they are not negative values. Instead it should display the out of SLA issues first, followed by the issues still in SLA sorted by the days remaining.

### DIFF
--- a/dojo/templatetags/display_tags.py
+++ b/dojo/templatetags/display_tags.py
@@ -270,7 +270,6 @@ def finding_sla(finding):
         status_text = 'Remediation for ' + severity.lower() + ' findings in ' + str(sla_age) + ' days or less since ' + finding.get_sla_start_date().strftime("%b %d, %Y") + ')'
         if find_sla and find_sla < 0:
             status = "red"
-            find_sla = abs(find_sla)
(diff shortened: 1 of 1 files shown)


Gby56 (approved):

Maffooch (approved):
Good change! I totally spaced the procedure of feature and bug release falling on the same day. I believe thee release process can be tweaked to merge bugfix and dev into master when doing a feature release from dev. 

There is a bit of time to figure that out though since feature release is in another week due to the way the dates fell

coheigea:
OK thanks. Do we have a check by the way to rebase bugfix from dev after each push to master?

Maffooch:
Sorta. I have been following this model: https://documentation.defectdojo.com/contributing/branching-model/#release-and-hotfix-model

bug fix branch is merged to master and then master is merged back into bug fix and then into dev

mtesauro (approved):
Approved

PR #7507, opened 2023-01-27
Add screenshots of what slack notifications look like

Maffooch:
I was curious what the slack notifications looked like, so I went and found out. Adding the screenshots here to save others some time 

[sc-487]


mtesauro (approved):
Approved

devGregA (approved):

PR #7522, opened 2023-01-31
Refresh lock file on helm changes

dsever:
**Description**

When a contributor pushes any changes to the helm, it will detect and automatically refresh chart.lock file.

### DIFF
--- a/.github/workflows/refresh_helm_lock_file.yaml
+++ b/.github/workflows/refresh_helm_lock_file.yaml
@@ -24,7 +24,7 @@ jobs:
         with:
           version: v3.4.0
 
-      - name: Execute generate new Chart.lock gile
+      - name: Execute generate new Chart.lock file
         run: |
           helm repo add bitnami https://charts.bitnami.com/bitnami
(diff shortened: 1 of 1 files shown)


Maffooch (approved):
@alles-klar final approval up to you

alles-klar:
For the history: We have three options:

1. check if the chart has been changed and update on PR to the dev lock file (have no idea how to do it)
2. As part of the test detect that the chart file has been changed but lock file hasn't
3. Completely drop lock file, so if a consumer wants to use helm from repo code directly needs to do helm dep update manually. 

I would prefer option 1 and integrate it in the current release workflow: https://github.com/DefectDojo/django-DefectDojo/blob/dev/.github/workflows/release-2-tag-docker-push.yml

The current implementation will run on every helm related pull request, which will result in many merge conflicts.

dsever:
The idea is to have fresh lock file not in a release, in dev as well, in case we will track only charm.yaml then amount of conflicts will not be so significant.

alles-klar:
Well, let's try.

alles-klar (approved):

PR #7530, opened 2023-02-01
Add suppression evaluation in SARIF importer

TheRealArlie:
**Description**

The SARIF parser currently ignores the `suppressions` flag in SARIF. We had a project scanning code using semgrep and expoting the results in SARF. They were wondering why they see a finding in dojo that should be suppressed.

The SARIF parser now evaluates a non empty `suppressions` entry in a result as false positive as specified here:
https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127852

**Test results**

I have adapted the unittest.

**Documentation**

No further documentation needed. 

**Checklist**

- [X] Make sure to rebase your PR against the very latest `dev`.


Maffooch:
This seems good to me. @damiencarol what do you think?

damiencarol:
@Maffooch taking a look

damiencarol:
@Maffooch @TheRealArlie we should also make the flag `active=True`

TheRealArlie:
Hi @damiencarol, good point. I added it to the commit.

Maffooch (approved):

dojo/tools/sarif/parser.py
```diff
@@ -302,6 +302,12 @@ def get_item(result, rules, artifacts, run_date):
     if kind != 'fail':
         return None
 
+    # if finding is suppressed, mark it as False Positive
+    # Note: see https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127852
+    suppressed = False
+    if "suppressions" in result and result.get("suppressions") is not None:
```
damiencarol, line 308:
```suggestion
    if result.get("suppressions"):
```
This format is better, it will cover the case where the attribute is here but empty.

>  if the array is empty, a consumer SHALL treat the result as not suppressed.

So if the array is empty a simple if manage it better.

damiencarol (changes requested):

dojo/tools/sarif/parser.py
TheRealArlie: Good point. Thank you for being so attentive. I adapted It and outsourced the test case for suppressions. Hope the test are successful.

damiencarol (approved):

damiencarol:
I like what you've done with unit tests.

PR #7534, opened 2023-02-02
Use the product name as the description in the importer

coheigea:
**Description**

When using the importer, you can create a product with no description. Then if you try to edit the product in the UI it errors if you don't add a description. I think description should be optional in the UI as well.

### DIFF
--- a/dojo/importers/reimporter/utils.py
+++ b/dojo/importers/reimporter/utils.py
@@ -189,7 +189,7 @@ def get_or_create_product(product_name=None, product_type_name=None, auto_create
             member.role = Role.objects.get(is_owner=True)
             member.save()
 
-        product = Product.objects.create(name=product_name, prod_type=product_type)
+        product = Product.objects.create(name=product_name, prod_type=product_type, description=product_name)
         member = Product_Member()
         member.user = get_current_user()
         member.product = product


Maffooch:
I think it would be a better idea to make the product have a description via the importer rather than making it non required in the other places.

coheigea:
@Maffooch Can do, but this will break every API client who is not setting the description

Maffooch:
I don't think it would break anything really. This is only applicable for the auto create context on the re/import endpoints. Just set the description to the same thing as the title.

damiencarol:
@Maffooch I think the patch to make it optional was good. Mandatory description is weird IMO.

coheigea:
I've already change the PR to just re-use the description in the importer - let me know if I should revert :-)

Maffooch:
The PR as it stands currently is good for a bugfix in the upcoming release, but setting the field to not required should go in dev under the next release

coheigea:
Yep sounds good @Maffooch

Maffooch (approved):

devGregA (approved):

Maffooch:
@coheigea I brought this up with Greg to think about changing the description to optional

coheigea:
@Maffooch OK let me know, and I'll submit a new PR

PR #7535, opened 2023-02-02
Add a link to the source code in the JIRA description

coheigea:
**Description**

When pushing a semgrep finding to JIRA, it would be nice to add a link to the source code, link can be seen in DefectDojo UI,


mtesauro (approved):
Approved

PR #7540, opened 2023-02-03
Add the ability to group by the finding title

coheigea:
**Description**

I think it would be useful to be able to group by Finding Title. For example, you might want to push to a single JIRA all semgrep findings of a particular type, and not care about the filepath.


damiencarol (approved):

mtesauro (approved):
Approved

PR #7558, opened 2023-02-06
Add default Dockerfile operating system

CharlieSears:
**Description**

Building locally with `docker-compose --profile mysql-rabbitmq --env-file ./docker/environments/mysql-rabbitmq.env up --build` would fail with `failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/.../Dockerfile.nginx: no such file or directory`

**Documentation**

Updated references to Dockerfile.django and Dockerfile.nginx

Not sure of the best location to document the `DEFECT_DOJO_OS` environment variable but leaving it undocumented doesn't seem ideal. Open to advice if nothing comes to mind before this has eyes on it.

### DIFF
--- a/Dockerfile.nginx-alpine
+++ b/Dockerfile.nginx-alpine
@@ -2,7 +2,7 @@
 # code: language=Dockerfile
 
 # The code for the build image should be identical with the code in
-# Dockerfile.django to use the caching mechanism of Docker.
+# Dockerfile.django-alpine to use the caching mechanism of Docker.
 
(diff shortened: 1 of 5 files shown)


Maffooch:
I would say somewhere in here
https://github.com/DefectDojo/django-DefectDojo/blob/dev/readme-docs/DOCKER.md#building-images

Maffooch:
Going to run with this for now and add docs later

PR #7567, opened 2023-02-07
Only mitigate endpoints that are not already mitigated

coheigea:
**Description**

This is a performance improvement in the re-importer to:

1. Only mitigate endpoint statuses if it's not already mitigated
2. Only reactivate endpoint statuses if it's already mitigated


coheigea:
@Maffooch Somehow, only 6 checks ran on this PR?

Maffooch:
Hmm I guess the bugfix actions got overwrote by the ones on the dev? Not sure what happened on that one. I corrected this issue on master hoping they would trigger here, but I guess not

damiencarol:
@coheigea trying to add a comment to try to trigger the whole thing.

Maffooch (approved):
This just saves a write per endpoint status. If unit tests break then I will fix

mtesauro (approved):
Approved

PR #7315, opened 2022-12-22
Notifications: Split templates per type

kiblik:
This change should give more freedom in the customisation of notification engines (and implementation of a new one). Some notifications could be more customised only for some specific engine (in other engines, default template (other.tpl) will be used).

Note:
Command for the update of `django.po`: `./manage.py makemessages -l en --no-obsolete --extension html,txt,py,tpl`

PR #7593, opened 2023-02-10
Fix missing information in notifications

coheigea:
There is a regression in 2.19.0 caused by https://github.com/DefectDojo/django-DefectDojo/pull/7315. Alerts are missing key information, i.e. on slack:
```
DefectDojo
APP  [3:01 PM]
The engagement "" has been created in the product "". It can be viewed here:
```
It's the same fix as for https://github.com/DefectDojo/django-DefectDojo/pull/7301
I tested most of the changes for mails + alerts.


damiencarol (approved):

mtesauro (approved):
Approved.

Thanks for the i18n/translation additions too!

PR #7603, opened 2023-02-10
Add parser for Codechecker

drJabber:
**Codechecker**

Codechecker SAST analyzer
image

Import Codechecker static analyzer report in JSON format: https://codechecker.readthedocs.io/en/latest/
Report format described here: https://codechecker.readthedocs.io/en/latest/analyzer/user_guide/#parse

One could make Codechecker JSON report using command like this: 
```shell
CodeChecker parse /path/to/codechecker/analyzer/output/directory -e json -o /path/to/output/file.json
```

Before this step you should build your project with Codechecker build process interception, 
```shell
odeChecker log -b "make -j8" -o ./my.project.codechecker.log
```

then analyze it
```shell
CodeChecker analyze ./codechecker.log -o /path/to/codechecker/analyzer/output/directory

### CODE REVIEW

**Comment by kiblik** [comment]
Out of curiosity. If only `HASHCODE_FIELD` is `unique_id_from_tool`, isn't it better to set `DEDUPLICATION_ALGORITHM` to `DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL`?

**Comment by drJabber** [comment]
> Out of curiosity. If only `HASHCODE_FIELD` is `unique_id_from_tool`, isn't it better to set `DEDUPLICATION_ALGORITHM` to `DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL`?

May be you are right, 
but could you explain the difference in deduplication?  may be it will be faster? I don't understand completely...

**Comment by kiblik** [comment]
> May be you are right, but could you explain the difference in deduplication? may be it will be faster? I don't understand completely...

The calculation of hashcode is performed doesn't mether if you define fields
https://github.com/DefectDojo/django-DefectDojo/blob/4dfb3a43651846696823e2d3d7bd5ec485b54276/dojo/importers/reimporter/reimporter.py#L81
https://github.com/DefectDojo/django-DefectDojo/blob/4dfb3a43651846696823e2d3d7bd5ec485b54276/dojo/models.py#L2394-L2442

Question is if there is significant differents between `compute_hash_code_legacy` and concat of fields + `hash_fields`

Using `DEDUPE_ALGO_HASH_CODE` vs. `DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL` looks might similar (is there some time difference in SQL SELECT for `CharField(null=True,max_length=64)` and `CharField(null=True,max_length=500)`?)
https://github.com/DefectDojo/django-DefectDojo/blob/4dfb3a43651846696823e2d3d7bd5ec485b54276/dojo/importers/reimporter/utils.py#L25-L34
https://github.com/DefectDojo/django-DefectDojo/blob/4dfb3a43651846696823e2d3d7bd5ec485b54276/dojo/models.py#L2183-L2188
https://github.com/DefectDojo/django-DefectDojo/blob/4dfb3a43651846696823e2d3d7bd5ec485b54276/dojo/models.py#L2234-L2238

**Comment by drJabber** [comment]
thank you @kiblik :)
So, `DEDUPE_ALGO_HASH_CODE` computes hash from field set and stores in hash_code[64] field, 
`DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL` takes precomputed unique ID (an my case - also hash), stored in unique_id_from_tool[500] field
i.e. there is not difference in algorithm, just in hash compute mode

File: dojo/tools/codechecker/parser.py [hunk 1]
```diff
@@ -0,0 +1,134 @@
...
+            return self.get_items(tree)
+
+    def parse_json(self, json_output):
+        data = json_output.read()
+        try:
+            tree = json.loads(str(data, 'utf-8'))
+        except:
+            tree = json.loads(data)
+
+        return tree
```
damiencarol, line 36:
```suggestion
        return json.loads(data)
```

dojo/tools/codechecker/parser.py
```diff
@@ -0,0 +1,134 @@
...
+        title = vuln['checker_name']
+    elif 'message' in vuln:
+        title = vuln['message']
+    else:
+        title = unique_id_from_tool
+
+    finding = Finding(title=title,
+                      description=description,
+                      severity=severity,
+                      scanner_confidence=None,
```
damiencarol, line 106: this attribute is not mandatory
```suggestion
```

unittests/tools/test_codechecker_parser.py
```diff
@@ -0,0 +1,80 @@
...
+        testfile = open(
+            get_unit_tests_path() + "/scans/codechecker/cc-report-1-vuln.json"
+        )
+        parser = CodeCheckerParser()
+        findings = parser.get_findings(testfile, Test())
+        self.assertEqual(1, len(findings))
+        finding = findings[0]
+        self.assertEqual("clang-diagnostic-sign-compare", finding.title)
+        self.assertEqual("Medium", finding.severity)
+
```
damiencarol, line 26: Some checks on file_path and line attributes are missing.

damiencarol (changes requested):

unittests/tools/test_codechecker_parser.py
drJabber: thank you, Damien, 
added some extra checks

dojo/tools/codechecker/parser.py
```diff
@@ -0,0 +1,128 @@
...
+    if line is not None and column is not None:
+        description += 'Location in file: line {}, column {}\n'.format(line, column)
+
+    sast_source_line = line
+
+    sast_object = None
+
+    severity = get_mapped_severity(vuln.get('severity', 'UNSPECIFIED'))
+
+    mitigation = ''
```
damiencarol, line 76: `mitigation` attribute is optional. So better to not set it.
It seems that there is no mitigation data in the report, so better remove this part.
```suggestion
```

dojo/tools/codechecker/parser.py
```diff
@@ -0,0 +1,128 @@
...
+        title = vuln['checker_name']
+    elif 'message' in vuln:
+        title = vuln['message']
+    else:
+        title = unique_id_from_tool
+
+    finding = Finding(title=title,
+                      description=description,
+                      severity=severity,
+                      mitigation=mitigation,
```
damiencarol, line 101:
```suggestion
```
cf: previous comment

unittests/scans/codechecker/cc-report-1-vuln.json
```diff
@@ -0,0 +1,45 @@
...
+                "original_path": "/opt/_ga/openvpn/src/openvpn/push.c"
+            },
+            "line": 576,
+            "column": 20,
+            "message": "comparison of integers of different signs: 'int' and 'unsigned long'",
+            "checker_name": "clang-diagnostic-sign-compare",
+            "severity": "MEDIUM",
+            "report_hash": "d5e8aec375bc1d89630e5e848b442109",
+            "analyzer_name": "clang-tidy",
+            "category": "clang",
```
damiencarol, line 18: These attributes seems good candidates for tags. Please push one of them is the tags.

dojo/tools/codechecker/parser.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def get_mapped_severity(severity):
+    switcher = {
+        'CRITICAL': 'Critical',
+        'HIGH': 'High',
+        'MEDIUM': 'Medium',
+        'LOW': 'Low',
+        'STYLE': 'Informational',
+        'UNSPECIFIED': 'Informational',
```
damiencarol, line 126:
```suggestion
        'STYLE': 'Info',
        'UNSPECIFIED': 'Info',
```
According to:
https://github.com/DefectDojo/django-DefectDojo/blob/4dfb3a43651846696823e2d3d7bd5ec485b54276/dojo/models.py#L46-L47

This is a NO-GO

dojo/tools/codechecker/parser.py
```diff
@@ -0,0 +1,128 @@
...
+    severity = get_mapped_severity(vuln.get('severity', 'UNSPECIFIED'))
+
+    mitigation = ''
+
+    references = ''
+
+    review_status = vuln.get('review_status', 'unreviewed')
+    verified = review_status == 'confirmed'  # bug confirmed by reviewer
+    risk_accepted = review_status == 'intentional'  # not confirmed, not a bug, there are some reasons to make this code in this manner
+    false_positive = review_status in ['false_positive', 'suppressed']  # this finding is false positive
```
damiencarol, line 83: One big important attribute is missing here => `active`
More important than `verified` or others.

dojo/tools/codechecker/parser.py
```diff
@@ -0,0 +1,128 @@
...
+
+    hash = hashlib.sha256()
+    unique_id = vuln['report_hash'] + '.' + vuln['analyzer_result_file_path'] + description
+    hash.update(unique_id.encode())
+    unique_id_from_tool = hash.hexdigest()
```
damiencarol, line 88: This is interesting.
It seems the report data doesn't have an unique id.
So we you try to create one artificially.
I see another way to do it that could be cleaner but it will require some work outside of the parser infra. So maybe in another PR.
This part is good for now.

dojo/tools/codechecker/parser.py
```diff
@@ -0,0 +1,128 @@
...
+
+    sast_source_line = line
+
+    sast_object = None
+
+    severity = get_mapped_severity(vuln.get('severity', 'UNSPECIFIED'))
+
+    mitigation = ''
+
+    references = ''
```
damiencarol, line 78: Same comment as for `mitigation`. Make it simple. If we don't have references, just remove this part.

unittests/tools/test_codechecker_parser.py
```diff
@@ -0,0 +1,95 @@
...
+    def test_parse_file_with_one_vuln_has_one_finding(self):
+        testfile = open(
+            get_unit_tests_path() + "/scans/codechecker/cc-report-1-vuln.json"
+        )
+        parser = CodeCheckerParser()
+        findings = parser.get_findings(testfile, Test())
+        self.assertEqual(1, len(findings))
+        finding = findings[0]
+        self.assertEqual("clang-diagnostic-sign-compare", finding.title)
```
damiencarol, line 24: Some checks on `active` attribute are missing.

damiencarol (changes requested):

unittests/tools/test_codechecker_parser.py
drJabber: Thank you, Damien, 
fixed code to follow your suggestions

damiencarol (approved):

Maffooch (approved):

PR #7639, opened 2023-02-16
Fix missing GitLab scanners in settings file

jahrome:
This prevents the use of environment variables to set the deduplication algorithm for those scanners.


Maffooch (approved):

mtesauro (approved):
Approved

PR #7648, opened 2023-02-20
Add languages option

Arukara:
add languages


Maffooch:
@shipko please take a look

PR #7659, opened 2023-02-20
Correct "if" checks in docker release process

Maffooch:
There were some extra if checks that may have contributed to alpine and debian images being built across all steps of the docker release action


Maffooch:
Turns out the alpine image overwrote the debian image. Ended up pushing straight to master to correct the mixed up tags. Closing this as it is redundant

PR #7665, opened 2023-02-21
BlackduckAPI: remove test_product_connection

kiblik:
This fix should avoid error: `'BlackduckAPI' object has no attribute 'get_asset'`

More details (and story behind): https://owasp.slack.com/archives/C2P5BA8MN/p1676985422740079?thread_ts=1675768685.949799&cid=C2P5BA8MN

### DIFF
--- a/dojo/tools/api_blackduck/api_client.py
+++ b/dojo/tools/api_blackduck/api_client.py
@@ -30,7 +30,7 @@ class BlackduckAPI:
     #     if not response.ok:
     #         raise Exception(f'Unable to connect and search in BlackDuck due to ...')
     #     return f'You have access to project "..."'
- 
+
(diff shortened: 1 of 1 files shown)


Maffooch:
Thanks for attaching the slack link. The extra context was helpful. I think commenting out that function and adding a #TODO comment would be better for the sake of this proposed solution sometime in the future:

> It would be nice if somebody who actively use BlackDuck will be able to correctly implement functions test_connection and test_product_connection

Otherwise it will likely be forgotten about

Maffooch (approved):

devGregA (approved):

damiencarol:
> 

I can test it on my side. We have a real instance of BlackDuck

PR #7668, opened 2023-02-21
Add default announcement banner

Maffooch:
Adds a default announcement banner to be created by the initializer


alles-klar:
@Maffooch This on causes a bug. If the init container starts more than once, multiple announcements are created, which results in errors, for example when you edit a user.

Is suggest to revert this MR.

PR #7680, opened 2023-02-23
Added amd64 build config for Apple Silicon

corrupt:
**Description**

Added a matrix dimension for the buildx platform parameter to build both amd64 and arm64 images.

**Test results**

Without Docker Hub secrets, I cannot test. But the workflow change is really straight-forward.

**Documentation**
As far as I can see there is no architecture mentioned in the docs. If anything, this will make the tutorial in `Docker.md` work on Apple Silicon.

### DIFF
--- a/.github/workflows/release-x-manual-docker-containers.yml
+++ b/.github/workflows/release-x-manual-docker-containers.yml
@@ -29,6 +29,7 @@ jobs:
       matrix:
           docker-image: [django, nginx]
           os: [alpine, debian]
+          platform: [amd64, arm64]
     steps:
       - name: Login to DockerHub
         uses: docker/login-action@v2
@@ -89,6 +90,7 @@ jobs:
           context: .
           cache-from: type=local,src=/tmp/.buildx-cache-${{ env.docker-image }}
           cache-to: type=local,dest=/tmp/.buildx-cache-${{ env.docker-image }}
+          platforms: ${{ matrix.platform }}
 
       - name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}


Maffooch:
Since there is no change to the image name, how will users know which platform they would be getting from docker hub? Or is docker smart enough to just figure it out? I am unsure how this works with docker

corrupt:
> Since there is no change to the image name, how will users know which platform they would be getting from docker hub? Or is docker smart enough to just figure it out? I am unsure how this works with docker

That's the whole point. Docker will always pull the image that is compatible to your architecture unless you force it to do otherwise. Buildx will create a single manifest containing all target architectures automatically. The result will look something like this on Docker Hub.

image

Maffooch:
Dang that is cool! Will give time for @mtesauro to look at

Maffooch (approved):
Did some research and this should be fine

devGregA (approved):

damiencarol:
@devGregA @Maffooch I see this one has been merged but no new images in DockerHub. Did we missed something?

corrupt:
> @devGregA @Maffooch I see this one has been merged but no new images in DockerHub. Did we missed something?

It's been merged into `dev`. I assume we'll have for it to merge into `master`.

Maffooch:
Yes, the images are pushed to docker hub from master

Unfortunately, building on arm64 a total failure :/ I ended up reverting so the release could be completed. Details on the runs below:
https://github.com/DefectDojo/django-DefectDojo/actions/runs/4346888907

corrupt:
That seems to be an oversight on my part. I forgot to add multi-arch builds to the debian-based images but did add them for the alpine-based ones which fail now. Does it make sense to create a new PR? I cannot test the workflow before merge. Debian images should work, alpine I'll have to investigate.

ManAnRuck:
@corrupt did you have any updates or some issues. I'm just asking to maybe support if possible

PR #7685, opened 2023-02-23
Eliminate possibility of adding single finding to multiple risk exceptions

Maffooch:
There was a possibility to add the same finding to multiple risk exceptions. This was discouraged in the UI, but it was possible to do so by unaccepting, creating a new risk acceptance, and then adding the finding back to the original risk acceptance.


devGregA (approved):

blakeaowens (approved):
LGTM

PR #7700, opened 2023-02-27
Update JIRA when a finding is edited

coheigea:
**Description**

When you edit a finding, the JIRA is not updated. 

**Test results**

Tested that the description + title of a JIRA are updated when the finding is edited.


Maffooch:
@coheigea this could be very alarming to users not expecting this new behavior. Could you add a checkbox to the Jira config option toggling this behavior? Same thing for #7707

coheigea:
@Maffooch Done, please review again. I'll update the other PR to use this config when this PR is applied.

Maffooch (approved):
Thanks for the fast response! Will move to get it merged as soon as I can to make #7707 easier

mtesauro (approved):
Approved

PR #7701, opened 2023-02-27
Remove Github OAuth dependencies (Credit to @hackerhumble)

Maffooch:
Description
Support for Github integration is removed in release https://github.com/DefectDojo/django-DefectDojo/pull/7691 with commit https://github.com/DefectDojo/django-DefectDojo/commit/659b3e9d590f8465f14d13824332cd79eb3466a5.

But there are still references of 'GITHUB_OAUTH2_ENABLED' which is breaking the login page. Removed the references in views.py as they are no longer required.


Maffooch:
Seems like it is breaking if the basic auth login form is disabled. Looks like if basic auth is disabled, the page will be broken

PR #7706, opened 2023-02-28
Fixing extraVolumeMounts for deployments

preetsindhal:
**Description**

Helm Chart doesn't support extraVolumeMounts for deployments


preetsindhal:
this can be closed, i was able to use the `extraVolumes: [] ` format provided in values file.
thanks.

PR #7718, opened 2023-02-28
Make sure that we search for finding groups for "group by" in the sam…

coheigea:
…e test

**Description**

I noticed that when grouping findings in one product, they appeared in a finding group in a different product/test which shared the same name.

**Test results**

Tested that finding groups are jailed to the test associated with the finding.


Maffooch (approved):

mtesauro (approved):
Approved

damiencarol:
@coheigea good catch.

PR #7719, opened 2023-02-28
Add prefetching to technologies endpoint

Maffooch:
Reduce the number of calls in technologies endpoint from 3 to 1 (technologies, product ID query, user ID query
[sc-613]


mtesauro (approved):
Approved

devGregA (approved):

PR #7721, opened 2023-02-28
Add prefetching to engagement presets endpoint

Maffooch:
Reduce the number of calls in engagement presets endpoint from 4 to 1 (engagement presets, product type ID query, test type ID queries, network locations ID queries) [sc-614]


mtesauro (approved):
Approved

devGregA (approved):

PR #7722, opened 2023-02-28
Add prefetching to product API scan config presets endpoint

Maffooch:
Reduce the number of calls in product API scan config endpoint from 3 to 1 (product API scan config, product ID query, tool configuration ID query)  [sc-615]


mtesauro (approved):
Approved

devGregA (approved):

Issue #5973, opened 2022-03-01
docker-compose up fails with execution permission denied


r0bag:
**Be informative**
After following the docker installation instructions from github I am able to succesfully run the "docker-compose build" command, however the "docke-compose up" fails.

**Bug description**
I am receiving a "standard_init_linux.go:228: exec user process caused: permission denied" error message while running the "docke-compose up" command. 

**Expected behavior**
No execution permission errors.

**Deployment method** *(select with an `X`)*
- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Linux Centos8 4.18.0-348.7.1.el8_5.x86_64 #1 SMP Wed Dec 22 13:25:12 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
 - DefectDojo version (see footer) or commit message:  #5869 from DefectDojo/release/2.7.1 [ (HEAD -> master, tag: 2.7.1, origin/master, origin/HEAD)]

**Logs** 
--==LOGS==--
mysql_1         | 2022-03-01T08:10:16.743633Z 0 [Note] Event Scheduler: Loaded 0 events
mysql_1         | 2022-03-01T08:10:16.743893Z 0 [Note] mysqld: ready for connections.
mysql_1         | Version: '5.7.37'  socket: '/var/run/mysqld/mysqld.sock'  port: 3306  MySQL Community Server (GPL)
celerybeat_1    | standard_init_linux.go:228: exec user process caused: permission denied
initializer_1   | standard_init_linux.go:228: exec user process caused: permission denied
celeryworker_1  | standard_init_linux.go:228: exec user process caused: permission denied
uwsgi_1         | standard_init_linux.go:228: exec user process caused: permission denied
django-defectdojo_uwsgi_1 exited with code 1
django-defectdojo_celerybeat_1 exited with code 1
django-defectdojo_initializer_1 exited with code 1
django-defectdojo_celeryworker_1 exited with code 1
nginx_1         | Basic auth is off (HTTP_AUTH_PASSWORD not provided)
nginx_1         | 2022/03/01 08:10:23 [emerg] 1#1: host not found in upstream "uwsgi:3031" in /run/defectdojo/uwsgi_server:1
nginx_1         | nginx: [emerg] host not found in upstream "uwsgi:3031" in /run/defectdojo/uwsgi_server:1
django-defectdojo_nginx_1 exited with code 1
[...]

PR #7734, opened 2023-03-01
Update permissions for entrypoints

WojTecH94:
**Description**

It is a fix for my problems with "permission denied" error running Defect Dojo on Oracle Linux 8 (RHEL) in rootless mode.
More details: https://github.com/DefectDojo/django-DefectDojo/issues/5973

**Test results**

DD works as expected, not a big of a change.


dsever:
This should not be the case, not good practice, it breaks immutability

dsever:
@WojTecH94  can you try with this one, it is just PoC?

https://github.com/dsever/django-DefectDojo/tree/docker-compose-improvements

Only docker-compose.yaml I was playing with.

Issue #7756, opened 2023-03-03
Support longer "issue type" names for JIRA connector


nv-pipo:
**Is your feature request related to a problem? Please describe**
When trying to configure JIRA connector for DD, the "issue type" is restricted to 15 characters
![image](https://user-images.githubusercontent.com/7769945/222759463-960eb24c-bb12-44f2-989c-c54a9f12265e.png)

**Describe the solution you'd like**
Please allow for longer "issue types", 256 characters? for us 20 should be enough.

**Describe alternatives you've considered**
Use our own inhouse developed Jira connector.

coheigea:
PR submitted: https://github.com/DefectDojo/django-DefectDojo/pull/7761

coheigea:
Fix is released, the issue can be closed

PR #7761, opened 2023-03-06
Allowing a max length of 255 for default JIRA issue types

coheigea:
**Description**

A fix for https://github.com/DefectDojo/django-DefectDojo/issues/7756
Max length of 255 comes from https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issue-type-schemes/#api-rest-api-2-issuetypescheme-post


Maffooch:
@coheigea I think the following entry in the migration is forcing another migration to be created since the model definition would not match the database schema:
`('Spike123456789abcderfg', 'Spike123456789abcderfg'),`

coheigea:
thanks @Maffooch , indeed that was just a test value I used.

PR #7770, opened 2023-03-06
Improve Risk Acceptance api endpoint

Maffooch:
The risk acceptance API endpoint showed the full media URL path. It has been replaced with the access controlled download link. Also added an api path to download the risk acceptance proof directly


mtesauro (approved):
Approval

blakeaowens (approved):
LGTM

PR #7771, opened 2023-03-07
Add extra path in Engagement endpoint for checklists

Maffooch:
Add extra path in Engagement endpoint for checklists because there was not a way to get or create those checklists from the api


mtesauro (approved):
Approved

blakeaowens (approved):
LGTM

kiblik:
@Maffooch, I'm just working on #9751 and I found out that this PR is using the non-existing function `get_authorized_engagement_checklists`. Can you check it, please?

Maffooch:
@kiblik Thanks for the heads up. I found this the other day as well. I believe at the time of implementation, I attempted to add a full endpoint with model viewsets for Engagement Checklists, but then backed up to only add a few new paths to the Engagement viewset. Since the viewset was not actually called anywhere, there was never an error thrown for trying to use the non existent `get_authorized_engagement_checklists` method.

In a later PR (#10503) I removed the `EngagementCheckListViewset ` viewset that calls the `get_authorized_engagement_checklists ` method. You could do that same in your PR, or just pull in the latest changes from the release when it drops later today

PR #7787, opened 2023-03-10
drf-spectacular : Load resources from static path

kiblik:
Add an option to load `drf-spectacular` resources from the static path.

Useful if
- can not access external resources because of internal corporate policy
- need to load everything locally (e.g. if you are developing DD in the aeroplane)


Maffooch (approved):

mtesauro (approved):
Approved

PR #7797, opened 2023-03-13
Add link to finding group from inside finding

coheigea:
**Description**

It would be useful to link directly inside a finding to the finding group.

**Test results**

Tested this works OK


Maffooch (approved):

mtesauro (approved):
Approved

PR #7804, opened 2023-03-13
feat: New page to display all API scan configurations

Gby56:
**Description**

Currently, for API Importers, there is no easy way to show how a Tool configuration is used across all products in the instance.
This is primordial to make sense of the overall Vuln management system, to provide an overview of the "mapping"

Further work should also implement a "pre-ingestion" view to show all the discovered scans from all configured API importers, and let the user simply "link" these scans to products, to automate the ingestion. (That's also a work item I think, there is no automated cron ingestion)

**Test results**

image

### DIFF
--- a/dojo/api_scan_configurations/urls.py
+++ b/dojo/api_scan_configurations/urls.py
@@ -1,0 +1,6 @@
+from django.urls import re_path
+from . import views
+
+urlpatterns = [
+    re_path(r'^api_scan_configurations$', views.api_scan_configurations, name='api_scan_configurations'),
+]
--- a/dojo/api_scan_configurations/views.py
+++ b/dojo/api_scan_configurations/views.py
@@ -1,15 +1,9 @@
 # #  product
 import logging
 
-from django.contrib import messages
-from django.urls import reverse
-from django.http import HttpResponseRedirect
 from django.shortcuts import render
 from dojo.models import Product_API_Scan_Configuration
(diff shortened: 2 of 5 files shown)


shipko:
Hi @Gby56!
I see that you don't using django localization feature. It's a bad practice. Please wrap your text in {% trans "" %} tag. 

More information you may see https://docs.djangoproject.com/en/dev/topics/i18n/translation/

Gby56:
@shipko done ! thanks for reminding me, I completely forgot about it, turns out there are no translations yet anyways but still 😅

damiencarol (approved):

devGregA:
Hi @Gby56! I’m sure you are excited to see me on another PR.

We moved to consider DefectDojo feature complete almost a year ago (although I’ll be very honest that I’m very bad with approximating time so +- 6 months). I’ve gone back through our documentation, contributing guidelines, and PR templates, to be more clear, as it wasn’t as clear as it should be, and ultimately that is on me.

Accepting the few API parsers DefectDojo has currently was likely a mistake and has created some maintainability pain. We're not going to remove the existing API connectors to make sure we do not impact current functionality, but accepting new API parsers / connectors / API parser enhancements is problematic (but open to revisiting in the future). To maintain a high quality API connector, it is necessary to have a license to the tool. To get that license requires partnership with the author or vendor (working on this).

PR #7812, opened 2023-03-14
Remove numpy

shipko:
**Description**
I noticed that DD used numpy dependency. After some research I found that this library is using only in one place for one function. Using this huge library for one method - is bad idea. And I decided to replace numpy method and remove this library from project.


Maffooch (approved):

mtesauro (approved):
Approved

damiencarol:
@mtesauro this is exactly the change §I was talking about. I'm glad a contributor was able to push it.

PR #7820, opened 2023-03-15
Allow pushing to JIRA and specifying the JIRA URL from the finding gr…

coheigea:
…oup edit page

**Description**

There is no way to specify a JIRA for a finding group, unlike for findings. This improvement adds the ability to both push to JIRA in the finding group edit/view page, as well as to specify a JIRA URL to link instead.

**Test results**

Tested all scenarios I could think of with JIRA.

Screenshot 2023-03-15 at 11 08 53


Maffooch (approved):

mtesauro (approved):
Approved

PR #7822, opened 2023-03-15
Fix code warnings about logger.warn

coheigea:
**Description**

Just a cleanup to fix some deprecation errors about logger.warn.


Maffooch:
Possibly. Will take a look

Maffooch:
I think it should be okay now if you pull latest changes

mtesauro (approved):
Approved

PR #7830, opened 2023-03-16
Entrypoints permissions fix

WojTecH94:
Description

It is a fix for problems with "permission denied" error running Defect Dojo on Oracle Linux 8 (RHEL) in rootless mode. Permissions are preserved from the host system during copy so we have to make sure that the files will have execution permission in container.
More details: https://github.com/DefectDojo/django-DefectDojo/issues/5973

Test results

DD works as expected, not a big of a change.


Maffooch:
successor to #7734

mtesauro (approved):
Approved

PR #7836, opened 2023-03-17
Adding/Simplify Assign Issue for Cloud/Premises

Roooodie:
**Description**

before: update function send "name" key field for assigning issue. but it will have a problem if users use Jira cloud where the "name" key is deprecated (https://developer.atlassian.com/cloud/jira/platform/deprecation-notice-user-privacy-api-migration-guide/)

proposed: using the Jira library function assign_issue, it will use user search in the jira for users that have a similar name then take and use the "name" (for premise) or "accountId" (for cloud) then send it using the correct key.


Maffooch (approved):

mtesauro (approved):
Approved

PR #7844, opened 2023-03-20
Coheigea/sla backoff

coheigea:
**Description**

Add exponential backoff strategy for SLA notifications. By default the old strategy of alerting every day is used. It can get annoying though to have a JIRA clogged up with repeated SLA notifications.

**Test results**

I tested that the expression works OK locally and the UI loads, but otherwise didn't test it, as it's not something easily testable without being deployed.


Maffooch:
Annoyance can be quite effective for motivating progress 😉 

I am unsure whether so many db changes are good for a bugfix release though. This seems more like a dev branch thing

coheigea:
Good point @Maffooch , I rebased to dev instead

Maffooch (approved):
Pretty genius way to check If a number is a power of two! I didn't get it at first until I tested 😂

mtesauro (approved):
Approved

PR #7895, opened 2023-03-28
feat: fix jira errors handling

kokhanevych-macpaw:
**Description**

I found undefined variables for two functions. It produces an error in a case when **except JIRAError** block is triggered.


Maffooch (approved):

mtesauro (approved):
Approved

PR #7915, opened 2023-03-30
Update detect_secrets parser and include FP findings in importer

adrianasantex:
**Description**

I have developed the detect-secrets parser and the global importer to mitigate findings that are set on the report as false positive.


adrianasantex:
Can you help us @coheigea ??

adrianasantex:
Now it works :) @Maffooch

damiencarol (approved):
Thanks for improving this parser. `active` attribute is an important one.

Maffooch (approved):

blakeaowens (approved):

PR #7916, opened 2023-03-30
Don't create a notification alert when a finding is a duplicate

coheigea:
**Description**

We noticed a problem when de-duplication is enabled and when multiple engagements report the same CVE in a product with push_to_jira enabled. The first Finding creates the JIRA, but then for all subsequent Findings we get notification alerts that they can't push to JIRA, as the deduplication logic sets the Findings to inactive before push_to_jira is called.

The fix is just to check when push_to_jira fails if the finding is already marked as duplicate, and just log this rather than create an alert.

**Test results**

Test it works locally.


Gby56 (approved):

Maffooch (approved):

mtesauro (approved):
Approved

PR #7934, opened 2023-04-04
arm64v8 support with Docker Compose

cneill:
**Description**

* Adding arm64v8 support for Apple M1/M2 laptops by adding the `docker-compose.override.arm64.yml` file and updating the `./dc-up.sh` and `dc-build.sh` scripts to detect your system architecture automatically
* Making some other minor changes to the `./dc-*` scripts
    * Removing the `./dc-start-d.sh` script (redundant since you can just pass `-d` to the `./dc-start.sh` script)
    * Allowing multiple CLI args to be passed to all of them
* Changing the default guidance to using the `postgres-redis` profile rather than `mysql-rabbitmq`
* This necessarily bumps MySQL to version `8.0.32` since [that's all that's currently offered for arm64v8](https://hub.docker.com/r/arm64v8/mysql)

I debated swapping `docker-compose` out for `docker compose` since V1 will be [EoL'd in June](https://docs.docker.com/compose/compose-v2/), but this is already a large enough PR.

**Note that because `arm64v8` images aren't currently being shipped to Docker Hub, the images will always be built locally.**

**Test results**

No (core) code changes.

**Documentation**

Updated `README.md` and `readme-docs/DOCKER.md`.


devGregA (approved):

mtesauro (approved):
Approved

Maffooch:
Closing for now due to stale

PR #7954, opened 2023-04-11
Finding hot edit

Sh1nZ0u:
Description

Add hot actions on finding_view page - some buttons for status change. All buttons update status and page using ajax.
Add localization to finding_view.html

Before
before

After
after

ajax_update

### DIFF
--- a/dojo/finding/urls.py
+++ b/dojo/finding/urls.py
@@ -53,6 +53,8 @@ urlpatterns = [
         name='view_finding'),
     re_path(r'^finding/(?P<fid>\d+)/edit$',
         views.edit_finding, name='edit_finding'),
+    re_path(r'^finding/(?P<fid>\d+)/edit_view',
+        views.finding_hot_edit, name='finding_hot_edit'),
     re_path(r'^finding/(?P<fid>\d+)/touch',
         views.touch_finding, name='touch_finding'),
     re_path(r'^finding/(?P<fid>\d+)/simple_risk_accept',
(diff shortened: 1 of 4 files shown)


Maffooch:
hi @Sh1nZ0u 
Thank you for the contribution! I am a little weary of this feature due to the ease of accidental status changes. A secondary confirmation message is not too much different than the actions already found in the drop down or editing the finding directly though. I am not sure of the best approach here

shipko:
@Maffooch @Sh1nZ0u hi!
Thanks for this PR. 
Sorry @Maffooch but I don't agree with you. The current UX of handling found bugs is quite complex and requires a lot of overhead. In my experience, my colleagues have complained about just that. I see that this functionality speeds up this rather routine activity. Of course, I am not the maintainer of this project, but it seems to me that such functionality is necessary here

damiencarol:
> hi @Sh1nZ0u Thank you for the contribution! I am a little weary of this feature due to the ease of accidental status changes. A secondary confirmation message is not too much different than the actions already found in the drop down or editing the finding directly though. I am not sure of the best approach here

I backup this comment, people will click too fast on these buttons.

Maffooch:
Maybe putting those hot actions in this drop down menu would accommodate both concerns?
image

Maffooch:
Thanks again for your contribution! Unfortunately, we are closing as this is not in line with the style and flow of other operations in DefectDojo

PR #7955, opened 2023-04-11
Fix processing of test/eng fields during re-import

kiblik:


Maffooch:
@kiblik I am not sure about this change. I have seen some use cases where an engagement is treated as a development branch, and each test corresponded to a feature branch. This is especially true in the cases of larger repos where product <-> repo relationships are not possible

kiblik:
> @kiblik I am not sure about this change. I have seen some use cases where an engagement is treated as a development branch, and each test corresponded to a feature branch. This is especially true in the cases of larger repos where product <-> repo relationships are not possible

No problem. I understand your argument and it makes sense. If somebody would like to do this kind of update, he still can make another API call for an update of engagement fields.
I accidentally bumped into this implementation part when I was debugging something else.

But in that case, at least
- `test.engagement.version = version` should be removed - it doesn't make any sense under `if branch_tag:`
- `if branch_tag:` should be changed to `if commit_hash:`

Maffooch:
Agreed on the last bit. Something screwy happened there at some point 🥴

kiblik (commented):
Dump editing eng

Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

cneill (approved):

PR #7958, opened 2023-04-11
Update DRF Filtersets fields to lists rather than tuples

Maffooch:
I saw in the DRF docs that `filterset _fields` were all lists. The ones in dojo are all tuples. I could not find when that change occurred, but we should match whatever is in the docs

https://www.django-rest-framework.org/api-guide/filtering/#djangofilterbackend


blakeaowens (approved):

mtesauro (approved):
Approved

cneill (approved):

PR #7960, opened 2023-04-12
Fix wide grants for gitlab auth

shipko:
**Description**
I noticed that DD requires some wide grants than it was needed. This PR is solving this problem.

I remove grants email, profile - this providing grant "openid". Grant "api" is replace to read_repository


Maffooch:
Good idea. Adding to the docs would be a good idea too so people are not quite as surprised if their SSO no longer works. In the upgrade notes would prolly be best

shipko:
@Maffooch agree with you. If developer enable gitlab_auto_import_project setting this functional will not working. I will write it in documentation.

shipko:
@Maffooch  done!

Maffooch (approved):

mtesauro (approved):
Approved

devGregA (approved):

cneill (approved):

PR #7964, opened 2023-04-12
bugfix - Add support for current close old findings behavior to UI

sarahgibs:
**Description**

A couple months ago import_scan changed its default behavior for closing findings. The `close_old_findings` flag now only covers the same engagement. If you want it to cover all findings in the same product you have to send a second flag along with the first, `close_old_findings_product_scope`. The Import Scan page does not reflect this change, and currently only supports closing findings within the same engagement. 

This PR adds a check box for 'close_old_findings_product_scope' to the Import Scan page. Instead of requiring both check boxes to be selected, which could be confusing, the checkboxes are presented as two separate choices. One to close old findings at the engagement scope, and one for closing old findings at the product scope. The backend code ensures both flags are set to true when 'close_old_findings_product_scope' is selected.

This issue isn't operating system dependent, but I replicated it using docker compose on my Mac OS 13.3.1.

**Test results**

Tested closing old findings at both the engagement and product level using the UI.
Existing unit tests passed.

### DIFF
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -622,6 +622,11 @@ def import_scan_results(request, eid=None, pid=None):
             api_scan_configuration = form.cleaned_data.get('api_scan_configuration', None)
             service = form.cleaned_data.get('service', None)
             close_old_findings = form.cleaned_data.get('close_old_findings', None)
+            # close_old_findings_prodct_scope is a modifier of close_old_findings.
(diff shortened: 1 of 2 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #7996, opened 2023-04-20
Strip spaces from tags before pushing to jira

Maffooch:
Jira does not allow spaces in their tags. Remove the spaces in tags coming from dojo before they are pushed out to jira


mtesauro (approved):
Approved

blakeaowens (approved):

cneill (approved):

PR #7998, opened 2023-04-21
Metrics multi type and etc

Sh1nZ0u:
**Description**

A little changes for improvement of metrics work:
- remove excess tables and agregate them to total findings count table
metrics_after_2
- add active/closed graph by week
metrics_after_1
- and change dashboard graph to weekly too
graph_after
- you can choose several product types at one metric
metrics_choice_after
- and now you can get statistic by one or several products
- actually set current month as default
- rewrite metric count functions and delete a lot of excess sql queries
**before**
sql_before
**after**
sql_after


mtesauro:
These are solid ideas however we're looking at completely overhaul/replace the library used for all the metrics in 3.0. 

If you're not aware about the restrictions that are in place for DefectDojo v3, I wrote up a discussion item at https://github.com/DefectDojo/django-DefectDojo/discussions/8002 that spells out what we're thinking.

We put the 'consider-for-3.0' tag on this so your good ideas won't be lost while we plan for DefectDojo v3.

PR #8011, opened 2023-04-25
Don't send the response headers to slack in a JIRA alert

coheigea:
**Description**

On a JIRA transition error, DefectDojo is sending the full response headers in the slack alert. I think it's sufficient if the response headers just get logged locally in DefectDojo, and just send the text in the alert. Otherwise on a few errors it pollutes the slack channel with noise.

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -413,9 +413,14 @@ def jira_transition(jira, issue, transition_id):
             jira.transition_issue(issue, transition_id)
             return True
     except JIRAError as jira_error:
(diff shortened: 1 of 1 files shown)


coheigea:
Test is failing with "open nginx/nginx-alpine_img: no such file or directory"

Maffooch:
I have reran these tests 7 times... rebuilding the containers will likely do it

Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #8013, opened 2023-04-25
Feature/pep8 refactoring code

renejal:


Maffooch:
@renejal please base your PR against dev per the contributing guidehttps://github.com/DefectDojo/django-DefectDojo/blob/master/readme-docs/CONTRIBUTING.md

Maffooch:
Dropping this comment here for the sake visibility:

>@renejal I think we are on the same page now. I think the course of action should be as follows
> - Close https://github.com/DefectDojo/django-DefectDojo/pull/8013
> - Update the code base one by one to be in compliance with Black, Pep8, and Flake8
> - Once the code base is in compliance, add the pre commit hooks and update the unit tests to ensure the code base stays in compliance.

PR #8025, opened 2023-04-28
Optimized calling popover plugin

shipko:
**Description**
I noticed that when we have over 250 products page /products is loaded to slow in Chromium based browser. After doing research on this topic, I fixed it in this MR


Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

devGregA (approved):

PR #8050, opened 2023-05-03
[Github parser] Parse dependabot alert number and add link in description

Gby56:
It's currently quite unusable to view dependabot alerts because they don't give a direct link into the Github platform, to maybe triage things there, get more details... view the PR attached etc...

Funny thing is, Github doesn't return a direct link to the alert, so you must either return the repo URL, or at least the full name with the org/owner, and build from there.

Another thing that I can't fix unfortunately: directly linking to the file in Github is hard, because yet again, Github's API isn't telling you the branch where the issue was detected. Dependabot opens PRs properly though, so maybe you could introspect the PR it has opened and find out the target branch... if someone reads this and wants to bother


Maffooch (approved):

mtesauro (approved):
Approved

devGregA (approved):

blakeaowens (approved):

PR #8083, opened 2023-05-10
Only alert on a JIRA error where the JIRA is active

coheigea:
**Description**

If a team resolve a JIRA upstream of DefectDojo, and if the JIRA workflow is such that a JIRA can't be edited when in a resolved/rejected state, then DefectDojo alerts when trying to resolve the Finding if push_to_jira is enabled. Instead we should just log the error, if the upstream JIRA is already resolved/rejected then an alert is not required.

**Test results**

Tested an alert does not happen any more in this scenario.


mtesauro (approved):
Approved

cneill (approved):

blakeaowens (approved):

PR #8091, opened 2023-05-11
Fix bug with mitigated findings

coheigea:
**Description**

When parsing a mitigated finding in the re-importer that is already mitigated, it doesn't add it to the list of unchanged items, so it always tries to close the old finding (which causes a JIRA ping if JIRA is enabled for the finding). There is a second bug fixed where newly mitigated findings aren't added to the mitigated list. 

**Test results**

Tested it manually. I fixed the test logic which was incorrect IMO, reimporting a newly mitigated finding should show an issue being closed and not remaining "untouched".


mtesauro (approved):
Approved

devGregA (approved):

cneill (approved):

PR #8094, opened 2023-05-11
chore: allow specifying revision history limit for deployments

celonis-eg:

**Description**

Allows setting the number of revisions of a replicaset maintained by a deployment in Kubernetes. Sets a default value of retained revisions for each deployment (worker, beat, and django).

**Test results**

n/a as this only touches the helm templates.

**Documentation**

The values file is updated with sane defaults.

**Checklist**


helm/defectdojo/values.yaml
```diff
@@ -124,6 +124,7 @@ celery:
       limits:
         cpu: 2000m
         memory: 256Mi
+    revisionHistoryLimit: 1
```
dsever, line 127: Maybe we should keep default k8s setting as is 10, and let user to specify less

dsever (changes requested):
Please take a look

helm/defectdojo/values.yaml
celonis-eg: Done. I put it in values as opposed to using default in the template or excluding it entirely from the template. I thought it'd be more appropriate to be specific about the value as opposed to letting the API load said default or using the [default directive](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/) from helm. But if there is a strong opinion one way or another, I can change it!

dsever: The thing is if some of the users are relaying on history feature (10) , it will break their understanding of their current setup

helm/defectdojo/values.yaml
```diff
@@ -124,6 +124,7 @@ celery:
       limits:
         cpu: 2000m
         memory: 256Mi
+    revisionHistoryLimit: 10
```
dsever, line 127: Can we simplify helm to use same settings for the all deployments, like in root level but applicable for all deployments.

dsever (commented):
Please take a look at my additional comments

helm/defectdojo/values.yaml
celonis-eg: Sure, why not?

celonis-eg: I updated the chart to support both.

helm/defectdojo/templates/django-deployment.yaml
```diff
@@ -1,3 +1,4 @@
+{{- $revisionHistoryLimit := or .Values.django.revisionHistoryLimit .Values.revisionHistoryLimit }}
```
dsever, line 1: Hi, so maybe I wasn't clear in my proposal about the simplification. I was thinking just to have one definition like `.Values.revisionHistoryLimit` and ommit under specific deployments in values.  Like is in the e.g . https://github.com/bitnami/charts/blob/main/bitnami/apache/values.yaml .

dsever (commented):
I'm trying to understand what would be the usecase of keeping different number of rev. for different types of deployments?

helm/defectdojo/templates/django-deployment.yaml
nnamzlas: I think this is actually an even better solution than just exposing a single value. If you want to set a standard value for all deployments it works exactly as you would want, while also providing more flexibility to users that want to specify different limits for different parts of the overall deployments 🤔

alles-klar: It is more flexible but also more complex. If there is a use case for it, the flexible one is ok. Personally can't imagine a relevant use case for this. I mean it is just the revision history and not nothing imported to customize per deployment.

celonis-eg: fwiw i left the conditional rendering (but not the conditional assignment) in so that users who did not specify this value  in any other customizations. This will keep any templates from being unexpectedly modified by an opinionated statement of the default history limit by these templates and defer to Kubernetes's defaults. Hope that is okay.

helm/defectdojo/templates/celery-beat-deployment.yaml
```diff
@@ -11,6 +11,9 @@ metadata:
     helm.sh/chart: {{ include "defectdojo.chart" . }}
 spec:
   replicas: {{ .Values.celery.beat.replicas }}
+  {{- if .Values.revisionHistoryLimit }}
```
dsever, line 14: Just final polishing, so  values by default parameters is disabled so then value is 10.  What is the reason with this approach you are not using `{{ .Values.revisionHistoryLimit  | default 10 }}`  to save 6 LoC  of flow control. It is not a big deal but code would be more readable

dsever (approved):
Approved

mtesauro (approved):
Approved.

grendel513 (approved):

blakeaowens (approved):

PR #8119, opened 2023-05-16
Make PostgreSQL database the default for containerized deploys

Maffooch:
MySQL supports case insensitivity out of the box which makes searches and exact match operations more challenging/ineffective. We have been talking about not supporting MySQL in v3 already as well. This is a good first step to test the waters
[sc-963]

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,8 @@ git clone https://github.com/DefectDojo/django-DefectDojo
 cd django-DefectDojo
 # building
 ./dc-build.sh
-# running (for other profiles besides mysql-rabbitmq look at https://github.com/DefectDojo/django-DefectDojo/blob/dev/readme-docs/DOCKER.md)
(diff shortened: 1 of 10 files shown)


testaccount90009:
Out of curiosity - for those that are on a docker-compose default mysql deployed version, is there any proposed plan for migrating the mysql data into a postgresql database?

Maffooch:
@testaccount90009 I am speculating here as dojo v3 is still in the design phase, but the migration will more than likely be a generalized process to convert dojo v2 objects into dojo v3 objects. It would probably not matter what database was being used as dojo would be normalizing the data above the db level

grendel513 (approved):
Looks good

mtesauro (approved):
Approved

blakeaowens (approved):

cneill (approved):

PR #8125, opened 2023-05-18
Revamp of the false positive history feature

adiffpirate:
> Reopening #6023

The current fp history implementation is kinda old and clunky, that's why I feel most people doesn't use it. Hopefully this PR will change that.
Personally, I use it for the following workflow:

  - Not use dedup;
  - One engagement per branch;
  - Reimport everything (to avoid a bunch of duplicates);
  - Mark the same finding as false positive in the entire product regardless of which branch/engagement has found it (to avoid marking the same thing as fp over and over again).

But it should be generic enough to fit other use cases.

Anyway, this PR makes the following changes:

  - [x] FP History now takes the deduplication algorithm into account in order to determine if a finding is false positive or not;
  - [x] FP History now only works when deduplication is **disabled** (it doesn't make sense to use both);
  - [x] The FP History method (`do_false_positive_history` at `dojo/utils.py`) retroactively mark all findings in the product as false positive if the finding passed as argument is (or is being marked as) a false positive;
  - [x] New method at `dojo/utils.py` called `match_finding_to_existing_findings` that takes one finding as argument and returns a list of existing findings based on its deduplication algorithm and a custom filter (this method can return matched findings in three diferents scopes: product wide, engagement wide, test wide). I feel like this is a useful method to have as it can be used for multiple purposes (including replace the similar, but less versatile, `match_new_finding_to_existing_finding` method at `dojo/importers/reimporter/utils.py`).
  - [x] Retroactively (un)mark all findings in the product as false positive when editing finding.
  - [x] Retroactively (un)mark all findings in the product as false positive when bulk editing findings.
  - [x] Unit tests
  - [x] Integration tests

Edit¹: The retroactive behaviour is now opt-in, toggleable via system settings.


mtesauro:
@adiffpirate I tried re-running the failed test and they remain red so there might be some more work to do here.

adiffpirate:
Hi @mtesauro. Appreciate the retry.

Probably something went wrong with the migrations, I'll take a deeper look and let you know 😉

opus-luiz-monteiro:
@mtesauro All fixed :)

mtesauro:
@adiffpirate Thanks, I'll take a look shortly.  Thanks for the fast turnaround.

dojo/finding/helper.py
```diff
@@ -351,12 +351,15 @@ def post_process_finding_save(finding, dedupe_option=True, false_history=False,
...
+        # Only perform false positive history if deduplication is disabled
+        if system_settings.enable_deduplication:
+            deduplicationLogger.warning("skipping false positive history because deduplication is also enabled")
+        else:
             from dojo.utils import do_false_positive_history
             do_false_positive_history(finding, *args, **kwargs)
-        else:
-            deduplicationLogger.debug("skipping false positive history because it's disabled in system settings")
+    else:
+        deduplicationLogger.debug("skipping false positive history because it's disabled in system settings")
```
Maffooch, line 362: This seems sorta excessive. Please remove
```suggestion
```

dojo/finding/views.py
```diff
@@ -818,6 +822,30 @@ def edit_finding(request, fid):
...
+
+                        for fp in existing_fp_findings:
+                            logger.debug('FALSE_POSITIVE_HISTORY: Reactivating false positive %i: %s', fp.id, fp)
+                            fp.active = new_finding.active
+                            fp.verified = new_finding.verified
+                            fp.false_p = False
+                            fp.out_of_scope = new_finding.out_of_scope
+                            fp.is_mitigated = new_finding.is_mitigated
+                            fp.last_reviewed = timezone.now()
+                            fp.last_reviewed_by = request.user
```
Maffooch, line 846: This might get a little confusing as it would create instances of findings being "reviewed" by someone who may not have any knowledge of the findings existence

dojo/finding/views.py
```diff
@@ -1936,6 +1968,31 @@ def finding_bulk_update_all(request, pid=None):
...
+
+                                    for fp in existing_fp_findings:
+                                        logger.debug('FALSE_POSITIVE_HISTORY: Reactivating false positive %i: %s', fp.id, fp)
+                                        fp.active = find.active
+                                        fp.verified = find.verified
+                                        fp.false_p = False
+                                        fp.out_of_scope = find.out_of_scope
+                                        fp.is_mitigated = find.is_mitigated
+                                        fp.last_reviewed = timezone.now()
+                                        fp.last_reviewed_by = request.user
```
Maffooch, line 1993: Same comment as ^

> This might get a little confusing as it would create instances of findings being "reviewed" by someone who may not have any knowledge of the findings existence

dojo/models.py
```diff
@@ -357,7 +357,25 @@ class System_Settings(models.Model):
     enable_mail_notifications = models.BooleanField(default=False, blank=False)
     mail_notifications_to = models.CharField(max_length=200, default='',
                                              blank=True)
-    false_positive_history = models.BooleanField(default=False, help_text=_("DefectDojo will automatically mark the finding as a false positive if the finding has been previously marked as a false positive. Not needed when using deduplication, advised to not combine these two."))
+
+    false_positive_history = models.BooleanField(
+        default=False, help_text=_(
+            "(EXPERIMENTAL) DefectDojo will automatically mark the finding as a "
+            "false positive if an equal finding (according to its dedup algorithm) "
```
Maffooch, line 364:
```suggestion
            "false positive if an equal finding (according to its dedupe algorithm) "
```

dojo/utils.py
```diff
@@ -50,55 +50,151 @@
...
+        ).exclude(id=finding.id).order_by('id')
+        deduplicationLogger.debug(query.query)
+        return query
+
+    elif deduplication_algorithm == 'legacy':
+        # This is the legacy reimport behavior. Although it's pretty flawed and
+        # doesn't match the legacy algorithm for deduplication, this is left as is for simplicity.
+        # Re-writing the legacy deduplication here would be complicated and counter-productive.
+        # If you have use cases going through this section, you're advised to create a deduplication configuration for your parser
+        logger.debug("Legacy dedup. In case of issue, you're advised to create a deduplication configuration in order not to go through this section")
```
Maffooch, line 185:
```suggestion
        logger.debug("Legacy dedupe. In case of issue, you're advised to create a deduplication configuration in order not to go through this section")
```

dojo/utils.py
```diff
@@ -50,55 +50,151 @@
...
+        return (
+            Finding.objects.filter(
+                **custom_filter,
+                title=finding.title,
+                severity=finding.severity,
+                numerical_severity=Finding.get_numerical_severity(finding.severity)
```
Maffooch, line 191: Default fields are specified here:
https://github.com/DefectDojo/django-DefectDojo/blob/386d2e69a4ada567e6b849e48b2978e22dcd19d1/dojo/models.py#L2442-L2445

Maffooch (changes requested):
A few nitpicky things, but overall a great improvement!

dojo/utils.py
adiffpirate: This one I didn't get.
You think we should change the filter by adding `cwe` + `line` + `file_path` + `description` and removing `severity` + `numerical_severity`?
The code is currently like this because it was based on the legacy reimport:
https://github.com/DefectDojo/django-DefectDojo/blob/dev/dojo/importers/reimporter/utils.py#L42C1-L52

Maffooch: Hmm that is a good point. My thought was that if FP history was based on dedupe algorithms then the default  algorithm should based on the default dedupe algorithm.  

I suppose since true dedupe cannot be enabled with FP history, the chances of a "default algorithm mismatch" are pretty slim. I think you made the right call keeping the the same algorithm as the reimoprter

@mtesauro what are your thoughts?

mtesauro: That is a very interesting point.

Given that you can have FP on and dedup off, relying on the legacy method seems to be the safest bet given how the code works today.

I agree with both of you (@Maffooch & @adiffpirate ) - lets go with the legacy method. :+1:

Maffooch (approved):
Will hold on merging this until after the minor release this upcoming Monday

mtesauro (approved):
Approved

grendel513 (approved):

mtesauro (changes requested):
Purposefully blocking this getting merged too soon. I'll remove this once the next release is out.

blakeaowens (approved):

PR #8156, opened 2023-05-24
Update rate limiting on the login page to only block on unsafe requests

Maffooch:
I noticed that users are getting 403 errors on the login page if any users are hitting it at the same time via VPN


cneill (approved):

mtesauro (approved):
Approved

devGregA (approved):

blakeaowens (approved):

PR #8158, opened 2023-05-25
Make requests of a peer review available to those with read access

Maffooch:
A continuation of #7027 
[sc-1028]

### DIFF
--- a/dojo/templatetags/authorization_tags.py
+++ b/dojo/templatetags/authorization_tags.py
@@ -52,3 +52,11 @@ def group_has_configuration_permission(group, codename):
         if permission.codename == codename:
             return True
     return False
+
+
+@register.simple_tag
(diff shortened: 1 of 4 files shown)


dojo/templatetags/authorization_tags.py
```diff
@@ -52,3 +52,11 @@ def group_has_configuration_permission(group, codename):
         if permission.codename == codename:
             return True
     return False
+
+
+@register.simple_tag
+def user_can_clear_peer_review(finding, user):
+    finding_under_review = finding.under_review
+    user_that_request_review = user == finding.review_requested_by
```
cneill, line 60: How about `user_requesting_review` to avoid the "that"?

blakeaowens (approved):

PR #8162, opened 2023-05-25
Unify push to jira logic for finding/groups  review/open/close

Maffooch:
Sometimes when working with the jira integration, there are instances of a finding being pushed to jira when modifying the status of a finding via the edit page, but not when using the open/close/review buttons. This is confusing and creates disparity.

Here is what this PR does
 - [x] Update single/grouped findings in Jira when a single finding is closed/open
- [x] Update single/grouped findings in Jira when a single finding has a requested review, cleared review, or actually reviewed
- [x] Correct statuses of findings under review to be Active instead of Inactive. This makes more sense as the decision for whether the finding is valid/mitigated is still up for debate and should not be marked as inactive until confirmed via review
- [x] Correct statuses of findings that have been reviewed when "not fixed" was selected. Previously, the finding would not be updated and would stay in an inactive state. This is crazy. The behavior now is the finding will be active and verified since it is confirmed the finding was not actually fixed and should this fact has been verified.
- [x] Correct statuses of findings that have been reviewed when "Closed" was selected. Previously, the finding would not be marked as verified, but that is the whole point of the review. So now the finding will be marked as inactive, mitigated, and verified.
- [x] Express Jira instance creation would not save the "automatically sync findings" flag. It does now
[sc-1040]

### DIFF
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -454,7 +454,7 @@ def close_finding(request, fid):
                 # Determine if the finding is in a group. if so, not push to jira
                 finding_in_group = finding.has_finding_group
                 # Check if there is a jira issue that needs to be updated
-                jira_issue_exists = finding.has_jira_issue or finding.finding_group.has_jira_issue
+                jira_issue_exists = finding.has_jira_issue or (finding.finding_group and finding.finding_group.has_jira_issue)
                 # Only push if the finding is not in a group
                 if jira_issue_exists:
                     # Determine if any automatic sync should occur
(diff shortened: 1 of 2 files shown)


mtesauro (approved):

devGregA (approved):

blakeaowens (approved):

blakeaowens (approved):

Maffooch (commented):
Do not merge until @37b has a chance to review

37b (commented):
Makes sense to me.

grendel513 (approved):

Issue #8165, opened 2023-05-26
Alert message missing important information to solve error


BoBeR182:
**Bug description**
Alert message not clear to resolve which project is missing Sonarqube API

**Steps to reproduce**
1. Go to 'Alerts'
2. Read the error.
3. <html><body>

  | Other | SonarQube API import issue | There are no API Scan Configurations for this Product.Please add at least one API Scan Configuration for SonarQube to this Product. | May 26, 2023, 1:41 p.m.
-- | -- | -- | -- | --

</body>
</html>Other 	[SonarQube API import issue](https://defectdojo.example.com:8443/alerts) 	

There are no API Scan Configurations for this Product.
Please add at least one API Scan Configuration for SonarQube to this Product.
	May 26, 2023, 1:41 p.m.

**Expected behavior**
Please provide a link or name for the project that is missing the Sonarqube API scan config

**Deployment method** *(select with an `X`)*
- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Debian
 - DefectDojo version :  v. 2.21.1 ( release mode ) 

**Screenshots**
![image](https://github.com/DefectDojo/django-DefectDojo/assets/525433/14f1c515-59de-4627-ab88-5760af8d53ae)

**Additional context** (optional)
We have over 2500 projects. Which one is misconfigured, since theses are all pipeline scans its hard to find it.

kiblik:
Fixed in [2.23.0](https://github.com/DefectDojo/django-DefectDojo/releases/tag/2.23.0).
Can be closed

PR #8166, opened 2023-05-26
API parsers: Make errors in importers more verbose (add product name and id)

kiblik:
**Description**

Solution for https://github.com/DefectDojo/django-DefectDojo/issues/8165

**Test results**

- [x] Updated 

**Documentation**

Not necessary


damiencarol:
@kiblik this is a lot of copy/pasted code. I think it could be better to create a typed exception and launch it with the good attribute and in the end catch this exception in upper layers to provide unified way to handle it.

damiencarol (approved):

Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #8181, opened 2023-05-30
Fix Detect Secrets parser when .baseline is not audited

adrianasantex:
**Description**

There was an error when importing not audited reports on Detect Secrets that marked as inactive secrets which haven't "is_secret" field.


Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

cneill (approved):

PR #8189, opened 2023-05-31
Accept hash_code field in the Generic parser

p-l-:
This is useful when the tool knows exactly how to compute the `hash_code` field and does it by itself.

**Checklist**

### DIFF
--- a/dojo/tools/generic/parser.py
+++ b/dojo/tools/generic/parser.py
@@ -84,7 +84,7 @@ class GenericParser(object):
                 'scanner_confidence', 'unique_id_from_tool', 'vuln_id_from_tool', 'sast_source_object',
                 'sast_sink_object', 'sast_source_line', 'sast_source_file_path', 'nb_occurences',
                 'publish_date', 'service', 'planned_remediation_date', 'planned_remediation_version',
-                'effort_for_fixing', 'tags'
+                'effort_for_fixing', 'tags', 'hash_code',
             }.union(required)
             not_allowed = sorted(set(item).difference(allowed))
             if not_allowed:


kiblik:
Are you aware that Dojo might recalculate this value in the future? It might be better to set "hashcode fields" for your Test Type in Dojo.

p-l-:
Wouldn't it be better to accept that projects compute their own values? And to only compute the field on DefectDojo if it was not provided?

kiblik:
I can talk only based on my experience and some observations with deduplication. If we would like to know more, we need to hear words from somebody who designed this part and implemented it.

How I see it: `hash_code` is the core parameter used in DefectDojo to be able to perform proper deduplication. Deduplication is run by DefectDojo itself so it makes sense that the calculation of `hash_code` is controlled by DefectDojo (without any outside interference). If there would be any change in the deduplication algorithm (mainly because of some possible bugs - you never now), DefectDojo should have full capability to recalculate `hash_code`s if it needs it. It is the reason why there is also a command [`docker-compose exec uwsgi ./manage.py dedupe --hash_code_only`](https://documentation.defectdojo.com/usage/features/#hashcode-generation--regeneration).

If would are sure that would know the proper way, how to identify finding uniquely, please use `unique_id_from_tool` or `vuln_id_from_tool` for it and set `DEDUPLICATION_ALGORITHM` or `HASHCODE_FIELDS` for your `Test_Type`. But keep managing `hash_code` only by DefectDojo.

Just some additional facts. Right now, Dojo is recalculating `hash_code`:
- If it does not exist (in `.save(...)`): https://github.com/DefectDojo/django-DefectDojo/blob/9e7ecf0869a83d4895b4f48adad9847d321d56f4/dojo/models.py#L2750-L2754
- During reimport: https://github.com/DefectDojo/django-DefectDojo/blob/9e7ecf0869a83d4895b4f48adad9847d321d56f4/dojo/importers/reimporter/reimporter.py#L108

Maffooch:
@kiblik is spot on here. You are looking for the `unique_id_from_tool` field and it serves the purpose you are looking for. A deduplication algorithm would need to be set on your end to use that field for it to be used with dedupe. The reason we do not have a default dedupe algorithm on the generic parser is because of the very wide spectrum of use cases.

p-l-:
Thanks a lot @kiblik (& @Maffooch for the confirmation).

PR #8198, opened 2023-06-02
bugfix - Close Old Findings will close all findings even when all findings are present in current import

sarahgibs:
This was reported to slack as a bug.

It looks like this [PR](https://github.com/DefectDojo/django-DefectDojo/pull/7915) introduced a bug with closing old findings. When you import the same report twice, it reports opening and closing all findings from the report. In the end usually a small number of the findings from the report end up active, while most of the should-be-active findings are closed.

To reproduce:
- Turn on `Deduplicate findings` in system settings
- In a new product (also works in an existing one, but easier to see in a new one), import a scan. (I used generic findings type).
- In either the same or a new engagement, import the same scan again, making sure the appropriate "close old findings" setting is selected.
- Notice that the banner reads "Generic Findings Import processed a total of x findings and closed x findings."
- If it was working correctly 0 findings would have been closed.
- I have tested locally with and without the added code in importer.py, and the behavior changes from broken to working when it is removed.

Screenshot 2023-05-30 at 4 17 33 PM

This bug fix has been tested locally to verify that the import functionality does not close the findings that should remain open. This has been tested with close findings within engagement and close findings within product scope. 
Screenshot 2023-06-02 at 4 52 54 PM

 

The change was to limit the number of calls to `test.finding_set.values()`, while keeping the functionality that was added in 7915 intact. It's not clear why that would have caused an issue, but calling it more than once in that function would lead to the findings being mitigated.


kiblik:
Is it possible to write a test for it, please?

sarahgibs:
> Is it possible to write a test for it, please?

I'll see if I can do that today or tomorrow.

Maffooch:
+1 on the unit test. It likely would have prevented this issue. I am unsure how this code really changes anything, but if it works, it works

sarahgibs:
Question on tests, should close old findings only work when deduplication is turned on? Because right now that's how it functions, but I'm not sure if that's intentional.

Maffooch:
deduplication and close old findings have always had weird side effects when enabled together. They basically do the same thing. Close old findings work at import time, and deduplication works post import. They tend to stomp on each other in some cases.

I was not aware of the requirement of Close Old Findings only be functional with dedupe. Is that enforced anywhere?

sarahgibs:
It is not enforced in the UI. 
When going through the Import Scan Results workflow, if you choose "close old findings" when deduplication is not on it adds the findings to the list if they are the same as existing findings. In that case no findings are closed and you end up with an extra copy of the open findings. If the list you upload contains completely different findings, it does close the findings that are not present in the new import. 
The problem we encounter with weekly scans being imported is that there are a lot of repeat findings from week to week, and we do not need 52 copies of the same finding. We have deduplication and close old findings turned on, and it has worked as expected for the last year that we have been using it this way (only one copy of the open findings stay open, and they are closed when they are no longer reported). 
Last week I noticed that findings that should still be open were being closed by results import. We tried to roll back our instance, but changes to the database schema mean that other functionality now doesn't work on the older version.

sarahgibs:
Added integration tests for closing old findings. Covers both within same engagement and same product. There are two sets, one with dedupe on and one with dedupe off. They both pass with the small change this PR introduced, and fail on the current bug fix branch.

kiblik:
When I wrote the test, I meant unittest, no integration test.
But performing an integration test is a good idea as well.

sarahgibs:
Do you have a guide to writing your unit tests? 
I would like to comply, as DefectDojo continues to be broken for my team.

kiblik:
@sarahgibs, there is no exact answer for this question but these might help:
- They are stored here https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/
- How you can perform them https://github.com/DefectDojo/django-DefectDojo/blob/master/readme-docs/DOCKER.md#run-the-tests-with-docker-compose
- How to write them for parsers (not exactly your case): https://documentation.defectdojo.com/contributing/how-to-write-a-parser/#unit-tests
- And 2 the closest to your case:
  - Deduplication logic: https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/test_deduplication_logic.py
  - Reopen during deduplication: https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/test_utils_deduplication_reopen.py

sarahgibs:
Added unit tests, they do not use deduplication.

Maffooch (approved):

devGregA (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #8207, opened 2023-06-05
Azure Group Mapping: Make group mapping less atomic

Maffooch:
When playing around with the azure group mapping, I found that there was an object ID in the groups response that did not match any resource I had in tenant. I tried deleting the application and all groups, and still the same object ID was blocking  my groups from being created in dojo. 

This change updates the exception handling to raise on each error, and then continue attempting the operation for the rest of the groups in the list. This way, if one apple gets in the bunch, the feature is still useful

Also had a very hard time figuring out what needed to be set on the Azure side to make this process work. Added some docs to clear this up for future folks
[sc-1078]


grendel513 (approved):

mtesauro (approved):
Approved

PR #8220, opened 2023-06-08
Update banner checks

Maffooch:
Behavior on when banners were created is inconsistent

### DIFF
--- a/docker/entrypoint-initializer.sh
+++ b/docker/entrypoint-initializer.sh
@@ -16,7 +16,7 @@ initialize_data()
 create_announcement_banner() 
 {
 # Load the announcement banner
-if [ -z "$CREATE_INITIAL_BANNER" ]; then
+if [ -z "$DD_CREATE_INITIAL_BANNER" ]; then
 echo "Creating Announcement Banner"
 cat <<EOD | python3 manage.py shell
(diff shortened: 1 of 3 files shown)


damiencarol (approved):

kiblik:
Not sure about the "Signals" part. Because this will avoid any Announcement (not only "Cloud and On-Premise Subscriptions Now Available! Click here for more details").
We are using Announcements for informing developers when we are planning to do upgrades. New users would miss this information.

kiblik:
Plus 
```shell
if [ -z "$DD_CREATE_INITIAL_BANNER" ]; then
```
might have the opposite behaviour to
```python
    if settings.CREATE_INITIAL_BANNER:
```

Maffooch:
> Not sure about the "Signals" part. Because this will avoid any Announcement (not only "Cloud and On-Premise Subscriptions Now Available! Click here for more details").
We are using Announcements for informing developers when we are planning to do upgrades. New users would miss this information.

Good point. Will think of alternatives

About your second comment, I am unsure what could be done about that. The same argument could be true for all env vars that are imported as django settings

devGregA (approved):

mtesauro (approved):

blakeaowens (approved):

cneill (approved):

PR #8221, opened 2023-06-08
Checkmarx: Add safeguards for "null" values in result types

Maffooch:
Checkmarx results may contain references to other result types even if they are not present. For example:
```
 "scanResults": {
  "sast": {
      ...
  },
  "sca": null,
  "kics": null
 }
```
Causes an error because the null value was not checked to be valid before it was attempted to be read


mtesauro (approved):
Approved

grendel513 (approved):

blakeaowens (approved):

devGregA (approved):

PR #8233, opened 2023-06-09
Tenable parser cleanups and improvements

Maffooch:
Went through the old Nessus/NessusWas parsers to look for any deviations between the original parsers and the consolidated parser. Here is what I found:
- CSV
  - Endpoint related fields may have been getting set to ""(empty string) in cases of rows being empty
  - Removed the duplicated code of adding plugin output to description
  - Added default "Info" severity in case one is not found in the report 
  - Reverted the dupe key calculation to what was originally in place
- XML
  - Returned the check for `NessusClientData_v2` data (though I believe it may be very outdated now)
  - Added some light formatting to the description output
[sc-1089]

### DIFF
--- a/dojo/tools/tenable/csv_format.py
+++ b/dojo/tools/tenable/csv_format.py
@@ -50,32 +50,21 @@ class TenableCSVParser(object):
             # skip entries with empty titles
             if not title:
                 continue
-
             # Severity: Could come from "Severity" or "Risk"
             if "Severity" in row:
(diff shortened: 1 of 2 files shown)


devGregA (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

cneill (approved):

PR #8249, opened 2023-06-14
Fixing type-error in Finding._age

coheigea:
**Description**

There is a bug in the Finding age calculation due to the fact that at least some parsers save a date and not a datetime object:

```
  File "/app/dojo/models.py", line 2695, in _age
    diff = get_current_date() - start_date
           ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
TypeError: unsupported operand type(s) for -: 'datetime.date' and 'datetime.datetime'
```

**Test results**

Added tests which verify the fix, as they fail before the fix is applied

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -2689,6 +2689,10 @@ class Finding(models.Model):
             else:
                 days = get_work_days(self.date, get_current_date())
         else:
+            from datetime import datetime
+            if isinstance(start_date, datetime):
+                start_date = start_date.date()
+
             if self.mitigated:
                 diff = self.mitigated.date() - start_date
             else:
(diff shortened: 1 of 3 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

Maffooch:
@coheigea I pulled these tests out as they are now failing when running a few days later. The SLA age is dynamic and did work with the age supplied when the test was ran, but now the finding is older, and the test is no longer working

coheigea:
@Maffooch Yes I spotted that. The test in the Veracode parser is correct as it works off the mitigated date, but the one in the Veracode SCA parser fails. I'm just about to submit another PR for a related issue so I'll include it in that.

PR #8253, opened 2023-06-14
Fix invalid OpenAPI schema + Add integration test

### LINKED ISSUE #7760: Empty is not a valid openapi 3.0 schema enum value for date-time and date fields
**Bug description**
OpenApi specification from (https://demo.defectdojo.org/api/v2/oa3/schema/?format=yaml) contains the following schema definition:  
[...]
/api/v2/findings/:
    get:
      operationId: findings_list
      parameters:
[...]
      - in: query
        name: created
        schema:
          type: string
          format: date-time
          nullable: true
          enum:
          - ''
          - 1
          - 2
          - 3
          - 4
          - 5
          - 6
          - 7
Here the empty enum value is not allowed...
Or at least openapi-generator-cli validate -i api.yaml  complains that 
[...]
	- attribute paths.'/api/v2/findings/'(get).parameters.[created].schemas.enum=`` is not of
	  type `date-time`
After removing the empty enum value (- '') the validation succeeded and I was able to generate a client.

**Expected behavior**
I expect without further manual modification of the download yaml openapi schema to generate an api client, 
Note: swagger 2.0 api endpoints works as expected, only oa3 has this bug.

**Environment information**
 - Tested with defectdojo demo site (api version: 2.9.2)

**Some hints**
- The problematic part is with DateRangeFilter in Filters.py:364. As '' valus refers to any date it can be the default value for date/date-time filters. However, for MetricsDateRangeFilter the '' value refers to 'Past 30 days'.
- A quick solution would be to remove format specifier from date/date-time fields with empty enum value.
- Or just make such fields are all nullable (that is already true for date-time but not for date fields) and remove empty enum anyway. In this case the missing field can be rendered as given with empty enum.

kiblik:
This PR  fixes #7760 and will identify issues similar to it.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -34,6 +34,7 @@ jobs:
           "tests/check_various_pages.py",
           "tests/notifications_test.py",
           "tests/tool_config.py",
+          "openapi-validatator",
         ]
         profile: ["mysql-rabbitmq", "postgres-redis"]
         os: [alpine, debian]
--- a/Dockerfile.integration-tests-debian
+++ b/Dockerfile.integration-tests-debian
@@ -11,7 +11,8 @@ RUN \
     curl \
     unzip \
     gpg \
-    jq \ 
+    maven \
+    jq \
     && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists && \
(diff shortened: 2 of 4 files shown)


kiblik:
Question: This fixed one filter. Are we going to adopt the same approach (`''` -> `None`) also for other filters?

Maffooch:
> Question: This fixed one filter. Are we going to adopt the same approach ('' -> None) also for other filters?

I think that sounds reasonable. @cneill @mtesauro thoughts?

mtesauro:
That makes sense to me. 

Let's give @cneill a bit of time to chime since it's a US holiday this week.

kiblik:
> 

@cneill, can we know your opinion?

kiblik:
@jozseftiborcz, I prepared the implementation based on your recommendation in #7760. Can you check it, please?

cneill (approved):
LGTM!

devGregA (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #8261, opened 2023-06-15
Request Review improvements

Maffooch:
When requesting reviews on findings, it is possible to select users with read permission that will not have access to actually provide the review. This fix limits the users that can provide reviews to those with at least write access to a given finding. 

In a system with many users, it is pretty annoying to have to select all of them for each review, so I added a checkbox to add all eligible reviews.


cneill (approved):

mtesauro (approved):
Approved

grendel513 (approved):

devGregA (approved):

PR #8263, opened 2023-06-16
SARIF: Add some extra logic around codeFlows

Maffooch:


devGregA (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

damiencarol (approved):

dojo/tools/sarif/parser.py
```diff
@@ -235,7 +236,7 @@ def get_description(result, rule):
             if fullDescription != message and fullDescription != shortDescription:
                 description += '**Rule full description:** {}\n'.format(fullDescription)
 
-    if 'codeFlows' in result:
+    if len(result.get('codeFlows', [])) > 0:
```
damiencarol, line 239: you know that this form is better, right?

```Python
if result.get('codeFlows'):
```

PR #8279, opened 2023-06-19
sla_deadline doesn't work with mitigated findings

coheigea:
**Description**

The sla_deadline function doesn't work with mitigated findings, as it just applies sla days remaining to the current date, instead of the mitigated date.

Also I added a mitigated date for the Veracode SCA parser when findings are mitigated, as the date isn't contained in the JSON.

**Test results**
 
New tests added in PR.


Maffooch (approved):
All green :)

grendel513 (approved):

PR #8282, opened 2023-06-20
Update files with PEP8 standards in folder dojo/tools #001

ajtortolero:
Update files with PEP8 standars in folders dojo/tools/

Folder affected
dojo/tools/dawnscanner
dojo/tools/dependency_check
dojo/tools/dependency_track
dojo/tools/detect_secrets
dojo/tools/dockerbench

### DIFF
--- a/dojo/tools/dependency_track/parser.py
+++ b/dojo/tools/dependency_track/parser.py
@@ -353,7 +353,7 @@ class DependencyTrackParser(object):
         data = file.read()
         try:
             findings_export_dict = json.loads(str(data, "utf-8"))
-        except BaseException:
+        except Exception:
             findings_export_dict = json.loads(data)
 
         # Exit if file is an empty JSON dictionary
--- a/dojo/tools/dependency_track/parser.py
+++ b/dojo/tools/dependency_track/parser.py
@@ -83,9 +83,7 @@ class DependencyTrackParser(object):
     }
     """
 
-    def _convert_dependency_track_severity_to_dojo_severity(
-        self, dependency_track_severity
-    ):
+    def _convert_dependency_track_severity_to_dojo_severity(self, dependency_track_severity):
         """
         Converts a Dependency Track severity to a DefectDojo severity.
         :param dependency_track_severity: The severity from Dependency Track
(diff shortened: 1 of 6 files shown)


dojo/tools/dependency_track/parser.py
```diff
@@ -242,16 +345,15 @@ def get_description_for_scan_types(self, scan_type):
...
         if file is None:
             return list()
 
         # Load the contents of the JSON file into a dictionary
         data = file.read()
         try:
-            findings_export_dict = json.loads(str(data, 'utf-8'))
-        except:
+            findings_export_dict = json.loads(str(data, "utf-8"))
+        except BaseException:
```
Maffooch, line 356: Is this an accurate exception?

ajtortolero: Hi @Maffooch, flake8 applies this change to comply with this [rule](https://www.flake8rules.com/rules/E722.html) and according to the [BaseException](https://docs.python .org/ 3/library/exceptions.html#BaseException) documentation  is a base class for any exception.

Maffooch: Please change the use of `BaseException` to the standard catch all `Exception`

Reason for this is that the `BaseException` class will also catch system exits and keyboard interrupts that may be not be desired/expected. The python docs you linked also mention that the `BaseException` class should not be directly extended (and therefore not directly used either) but that the `Exception` be used instead.

> The base class for all built-in exceptions. It is not meant to be directly inherited by user-defined classes (for that, use [Exception](https://docs.python.org/3/library/exceptions.html#Exception)).

Maffooch, line 356:
```suggestion
        except Exception:
```

Maffooch (changes requested):

dojo/tools/dependency_track/parser.py
ajtortolero: Changes apply in commit https://github.com/DefectDojo/django-DefectDojo/pull/8282/commits/2de95f283b3df8d72cc1577b35f2e202386dd84c

mtesauro (approved):
Approved

blakeaowens (approved):

alles-klar (approved):

Issue #8281, opened 2023-06-20
DrHeader Parser: Validation Error if uploading a scan result with 0 findings


manuel-sommer:
**Bug description**
KeyError in case of importing a non active test result

**Steps to reproduce**
Scan a target with DrHeader which is not active and upload the result to DefectDojo using DrHeader parser.

**Expected behavior**
DefectDojo parses the testresult without running into a ValidationError

**Deployment method:** Docker Compose
**Environment information:**  Ubuntu 20.04
**DefectDojo version (see footer) or commit message:**
[2023-06-16 09:58:53 -0500] https://github.com/DefectDojo/django-DefectDojo/commit/8ee0a7da7cd0e46f228c3e2c8a6c115ca469550b: Bump boto3 from 1.26.153 to 1.26.154 (https://github.com/DefectDojo/django-DefectDojo/pull/8266) [ (HEAD -> fix_testssl, origin/dev, dev)]

**Logs**
[20/Jun/2023 03:49:33] ERROR [dojo.engagement.views:703] ['Expecting value: line 1 column 1 (char 0)']
Traceback (most recent call last):
  File "/app/dojo/importers/importer/importer.py", line 309, in import_scan
    parsed_findings = parser.get_findings(scan, test)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/tools/drheader/parser.py", line 18, in get_findings
    data = json.load(filename)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/__init__.py", line 293, in load
    return loads(fp.read(),
           ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #8283, opened 2023-06-20
fix drheader parser #8281

manuel-sommer:
fix drheader parser #8281

### DIFF
--- a/dojo/tools/drheader/parser.py
+++ b/dojo/tools/drheader/parser.py
@@ -19,12 +19,6 @@ class DrHeaderParser(object):
         try:
             data = json.load(filename)
         except ValueError as err:
-            find = Finding(title="The uploaded file was empty.",
-                           test=test,
-                           description="The uploaded file was empty.",
-                           severity="Information",
-                           static_finding=False)
-            items.append(find)
             data = {}
         for item in data:
             findingdetail = ''
@@ -36,6 +30,5 @@ class DrHeaderParser(object):
                            description=message,
                            severity=severity,
                            static_finding=False)
-
             items.append(find)
         return items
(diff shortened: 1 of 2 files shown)


manuel-sommer:
@madchap could you retrigger the pipeline? 
The PR is ready to review.

kiblik:
I'm not sure, this is way, how "empty reports" should be processed.
I suppose, it should be handled this way
```python
        try:
            data = json.load(filename)
        except ValueError as err:
            data = {}
```

manuel-sommer:
Your opinions @mtesauro or @Maffooch ?

damiencarol:
@manuel-sommer : Not sure the pb is from dojo side. DrHeader should not produce empty files. Empty files are not valid JSON files.

manuel-sommer:
@damiencarol: I use drheader the following:
`drheader scan single targetasdf.de > output.json 
`
If the target is not available at the time of scanning, output.json is empty. I will look at the recommended way to use jq.

Maffooch:
Creating a finding in the case of an empty file is not good here. Adding the try catch in case of an empty file to report 0 findings for the sake of not throwing 500s in a pipeline is a good compromise as @kiblik suggested

manuel-sommer:
@Maffooch could you review?

Maffooch (approved):

manuel-sommer:
@grendel513 could you review this PR?

manuel-sommer:
@mtesauro could you review this PR?

mtesauro (approved):
Approved

cneill (approved):

damiencarol:
I maintain that we should push upstream, the current version is not good:
```
try:
            data = json.load(filename)
        except ValueError as err:
            data = {}
```
If the command line generate an error or anything, the report will be invalid JSON and dojo will close all findings in the scenario of re-import (because 0 findings as output).

PR #8287, opened 2023-06-20
fix: dont create sast object when nothing given

gotbadger:
**Description**

When importing a gitlab sast report you end up with something like.
Screenshot 2023-06-20 at 15 20 19

Gitlab sast data model for location looks like this:

```
"location": {
            "type": "object",
            "description": "Identifies the vulnerability's location.",
            "properties": {
              "file": {
                "type": "string",
                "description": "Path to the file where the vulnerability is located."
              },
              "start_line": {
                "type": "number",
                "description": "The first line of the code affected by the vulnerability."
              },
              "end_line": {
                "type": "number",
                "description": "The last line of the code affected by the vulnerability."
              },
              "class": {
                "type": "string",
                "description": "Provides the name of the class where the vulnerability is located."
              },
              "method": {
                "type": "string",
                "description": "Provides the name of the method where the vulnerability is located."
              }
            }
          },
```

Its debatable if we should even use sink or source objects in the finding since all we have is class/method but this PR takes a more limited aproach and just uses these when we have class and method information. This results in a more clean output on the finding page with just the location instead of the same filename/line information being duplicated 3 times.

**Checklist**

### DIFF
--- a/dojo/tools/gitlab_sast/parser.py
+++ b/dojo/tools/gitlab_sast/parser.py
@@ -94,18 +94,23 @@ class GitlabSastParser(object):
         file_path = location['file'] if 'file' in location else None
 
         line = location['start_line'] if 'start_line' in location else None
-        if 'end_line' in location:
-            line = location['end_line']
-
-        sast_source_line = location['start_line'] if 'start_line' in location else None
 
         sast_object = None
+        sast_source_file_path = None
+        sast_source_line = None
         if 'class' in location and 'method' in location:
             sast_object = f"{location['class']}#{location['method']}"
         elif 'class' in location:
             sast_object = location['class']
         elif 'method' in location:
             sast_object = location['method']
+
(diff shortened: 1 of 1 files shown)


Maffooch:
Hi @gotbadger 

Is this ready for review? Everything looks good to me!

gotbadger:
@Maffooch any thoughts on creating source/sink stuff vs not?

Maffooch:
Your solution is fine

I can see some value in recording that the source and sink are the same line (likely why gitlab does not report both source and sink when they align) but it would likely be intuitive already and be redundant. However, if the source and sink are in different places, then both should be recorded.

Maffooch (approved):

mtesauro (approved):
Approved

damiencarol (approved):
This is a good patch.

grendel513 (approved):

PR #8306, opened 2023-06-26
Docs: Fix link to key management

krizon:
**Description**

Replace `/api/v2/key/` link with `api/key-v2` as the original link links to 404 and the latter is the correct one.


Maffooch (approved):
Good find :)

PR #8314, opened 2023-06-26
Password validators: Fix validators, add tests

kiblik:


blakeaowens:
Would you mind also testing the password reset functionality from the login page? I did this without setting up SMTP by printing the context here https://github.com/DefectDojo/django-DefectDojo/blob/f1fefe4b81e3de78c50385a827809cce9e2fe219/dojo/user/views.py#L637 and manually forming the reset link. That was actually the main breaking point in the previous implementation that needed to be fixed.

Is there any possibility you have this completed before Friday afternoon?

kiblik:
> Would you mind also testing the password reset functionality from the login page? I did this without setting up SMTP by printing the context here
> 
> https://github.com/DefectDojo/django-DefectDojo/blob/f1fefe4b81e3de78c50385a827809cce9e2fe219/dojo/user/views.py#L637
> 
> and manually forming the reset link. That was actually the main breaking point in the previous implementation that needed to be fixed.
> Is there any possibility you have this completed before Friday afternoon?

Not sure, I understand you. How password validators are connected to the password reset link?

blakeaowens:
The reset password form rendered from the password reset link shares same validators. That is actually where the log snippet I sent you in my PR generating the 500 error came from.

kiblik:
image
Yes, validators work also in reset form.

It would be good to prepare some integration test for it but I do not have enough capacity for it right now.

blakeaowens (approved):
Hello, @kiblik sorry for the delay! I have tested this myself and everything looks good. Thank you for the PR -- approved!

grendel513 (approved):

Maffooch (approved):

devGregA (approved):

PR #8318, opened 2023-06-27
Fix occurrence where product exists, but cannot be found

Maffooch:
For some reason, I have seen cases where the product exists, but is not fetched by the `get_target_product_if_exists` function, but it does exist. 

Updating the `create` function to `get_or_create` makes this process safer across the board


mtesauro (approved):
Approved

grendel513 (approved):

blakeaowens (approved):

PR #8325, opened 2023-06-29
Disable NGINX leaking its version if TLS is not terminated on NGINX

fhoeborn:
**Description**

This disables NGINX leaking its version via the Server header if TLS is not terminated in the NGINX (e.g. in a Load Balancer in front of NGINX) 

**Test results**

Manually verified that Server header doesn't leak version


mtesauro (approved):
Approved

devGregA (approved):

Maffooch (approved):

cneill (approved):

PR #8346, opened 2023-07-04
Semgrep parser unique id from tool

kiberdruzhinnik:
**Description**

Semgrep parser has fingerprint field to track unique vulnerabilities. Current implementation does not handle this properly.

**Test results**

-

**Documentation**

-

**Checklist**

**Extra information**

### DIFF
--- a/dojo/tools/semgrep/parser.py
+++ b/dojo/tools/semgrep/parser.py
@@ -20,6 +20,10 @@ class SemgrepParser(object):
         dupes = dict()
 
         for item in data["results"]:
+            unique_id_from_tool = ""
+            if "extra" in item and "fingerprint" in item["extra"]:
+                unique_id_from_tool = item["extra"]["fingerprint"]
+            
             finding = Finding(
                 test=test,
                 title=item["check_id"],
@@ -31,7 +35,7 @@ class SemgrepParser(object):
                 dynamic_finding=False,
                 vuln_id_from_tool=item["check_id"],
                 nb_occurences=1,
-                unique_id_from_tool=item["extra"]["fingerprint"],
+                unique_id_from_tool=unique_id_from_tool,
             )
 
             # manage CWE
--- a/dojo/tools/semgrep/parser.py
+++ b/dojo/tools/semgrep/parser.py
@@ -31,6 +31,7 @@ class SemgrepParser(object):
                 dynamic_finding=False,
                 vuln_id_from_tool=item["check_id"],
                 nb_occurences=1,
+                unique_id_from_tool=item["extra"]["fingerprint"],
             )
 
             # manage CWE
--- a/dojo/tools/semgrep/parser.py
+++ b/dojo/tools/semgrep/parser.py
@@ -20,10 +20,6 @@ class SemgrepParser(object):
         dupes = dict()
 
         for item in data["results"]:
-            unique_id_from_tool = ""
-            if "extra" in item and "fingerprint" in item["extra"]:
-                unique_id_from_tool = item["extra"]["fingerprint"]
-            
             finding = Finding(
                 test=test,
                 title=item["check_id"],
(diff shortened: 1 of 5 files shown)


mtesauro:
@kotasher Thanks for the PR.

What version of Semgrep are you using? command-line/cli, one of the commercial variants, ...

Also, software release/version number of it if you have that handy would be great so we know what prompted this change.

kotasher:
@mtesauro it is the latest one (1.30.0, docker/cli, free version). Actually it has fingerprint field since [0.86.0](https://github.com/returntocorp/semgrep/blob/develop/CHANGELOG.md#0860---2022-03-24).

dojo/tools/semgrep/parser.py
```diff
@@ -31,6 +31,7 @@ def get_findings(self, filename, test):
                 dynamic_finding=False,
                 vuln_id_from_tool=item["check_id"],
                 nb_occurences=1,
+                unique_id_from_tool=item["extra"]["fingerprint"],
```
Maffooch, line 34: Might be a better idea to pull this out of the finding instantiation to check if the value is present/valid first. Assigning a potentially empty unique ID could cause all sorts of issues

Maffooch (changes requested):

dojo/tools/semgrep/parser.py
kotasher: should it be assigned "" or None if no fingerprint actually parsed?

Maffooch: Not assigning at all would be ideal. Something like this:

```
             finding = Finding(
                 test=test,
                 title=item["check_id"],
                 severity=self.convert_severity(item["extra"]["severity"]),
                 description=self.get_description(item),
                 file_path=item['path'],
                 line=item["start"]["line"],
                 static_finding=True,
                 dynamic_finding=False,
                 vuln_id_from_tool=item["check_id"],
                 nb_occurences=1,
             )

             unique_id_from_tool = item.get("extra", {}).get("fingerprint")
             if unique_id_from_tool:
                 finding.unique_id_from_tool = unique_id_from_tool
```
This would ensure that the finding would only know about about a unique id if the value is neither a None or an empty string

kotasher: @Maffooch check again, please.

Maffooch (approved):

damiencarol (changes requested):
could add unit test that check the new data

kotasher:
@damiencarol I'm not very familiar with unit testing but I attached some semgrep reports from version 1.30.0 containing current implementation of vulnerability fingerprinting to avoid close/open issues when matching code is just moved around or reindented.

damiencarol:
> @damiencarol I'm not very familiar with unit testing but I attached some semgrep reports from version 1.30.0 containing current implementation of vulnerability fingerprinting to avoid close/open issues when matching code is just moved around or reindented.

This is good but... only half of the job :D 

Take a look here to see an example:

https://github.com/DefectDojo/django-DefectDojo/blob/6ff5931f08f022c4e569537fad022558f7317915/unittests/tools/test_semgrep_parser.py#L97-L110

Just add a new test that use your file

kotasher:
> > @damiencarol I'm not very familiar with unit testing but I attached some semgrep reports from version 1.30.0 containing current implementation of vulnerability fingerprinting to avoid close/open issues when matching code is just moved around or reindented.
> 
> This is good but... only half of the job :D
> 
> Take a look here to see an example:
> 
> https://github.com/DefectDojo/django-DefectDojo/blob/6ff5931f08f022c4e569537fad022558f7317915/unittests/tools/test_semgrep_parser.py#L97-L110
> 
> Just add a new test that use your file

done

damiencarol (approved):

dojo/tools/semgrep/parser.py
damiencarol: all good, approved it.

PR #8351, opened 2023-07-05
Add signal to update Finding `found_by` column

Maffooch:
When updating the test type of a test, all of the findings under that test will now have an outdated `found_by` value. This column is shown in all of the finding lists, so it is important that it is accurate.

### DIFF
--- a/dojo/test/signals.py
+++ b/dojo/test/signals.py
@@ -1,0 +1,24 @@
+import contextlib
+from django.db.models import signals
+from django.dispatch import receiver
+import logging
+from dojo.models import Test, Finding
+
+logger = logging.getLogger(__name__)
+
+
+@receiver(signals.pre_save, sender=Test)
+def update_found_byt_for_findings(sender, instance, **kwargs):
+    with contextlib.suppress(sender.DoesNotExist):
+        obj = sender.objects.get(pk=instance.pk)
+        # Check if the test type has changed
+        if obj.test_type != instance.test_type:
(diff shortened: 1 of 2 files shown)


Maffooch:
Found By is really nice for duplicates across tools. I believe that was the main motivation for using a list for this field? That was before my time, so taking a guess

image

mtesauro:
@Maffooch Your guess is correct - that was the motivation for including that.

It's also just generally useful since you have a feel for how much you trust what various tools report.

damiencarol:
Yes in theory because I think most of the users have 2 lines if they import reports from 2 different tools.

grendel513 (approved):

mtesauro (approved):
Approved

devGregA (approved):

dojo/test/signals.py
```diff
@@ -0,0 +1,24 @@
...
+from django.db.models import signals
+from django.dispatch import receiver
+import logging
+from dojo.models import Test, Finding
+
+logger = logging.getLogger(__name__)
+
+
+@receiver(signals.pre_save, sender=Test)
+def update_found_byt_for_findings(sender, instance, **kwargs):
```
cneill, line 11: Should be `update_found_by_for_findings` (extra "t" in there)

cneill (changes requested):
Minor typo, otherwise looks good

blakeaowens (approved):

cneill (approved):

Issue #5535, opened 2021-12-02
passwords from secret management tool when installing via helm


tiagoposse:
**Is your feature request related to a problem? Please describe**
security around kubernetes secrets is subject to debate, but a part of the community prefers to use external secret management tools (like hashicorp vault) to store secrets.
The current helm chart supports passwords:
- defined directly in the values file (ends up being committed to git repo - insecure)
- load from kubernetes secrets (more secure, but still not as much as with external store)

**Describe the solution you'd like**
Configuration of passwords should be possible via extra file.
Something like:
```
postgresql:
  loadFromFile: false | true

redis:
  loadFromFile: false | true

extraConfigs:
  DD_EXTRA_ENV: <path to env vars with passwords>
```

**Additional context**
I've attempted to do this and was able to disable the requirement of secrets by the chart, but I hit a bump when the initializer did not read vars from DD_EXTRA_ENV and so db password was wrong. 
I've tried with `DD_EXTRA_ENV=extra_settings/.env.prod` (and others). I can probably get around this by providing a custom settings.py, but this seems unreasonable for such a little issue.
I've then changed line 235 of dojo/settings/settings.dist.py from:
if os.path.isfile(root('dojo/settings/.env.prod')) or 'DD_ENV_PATH' in os.environ:
`    env.read_env(root('dojo/settings/' + env.str('DD_ENV_PATH', '.env.prod')))`
to
`    env.read_env(root('dojo/settings/' + env.str('DD_ENV_PATH', '.env.prod')), overwrite=True)`

which caused the password to be correctly loaded, but initializer was still triggering a wrong password.

I would not mind to open an MR for this if you're at all interested, but I might need a bit of help debugging.

damiencarol:
@tiagoposse did try with the `local.settings.py` ? It's a better solution when you want to override settings with a file.

tiagoposse:
@damiencarol I was able to get to this finally. I was able to inject the secrets into `/app/docker/extra_settings/local_settings.py`. Thanks for the tip! I'm preparing a PR to the chart.

dsever:
But you have already option to inject extra secrets. That will end inside of secret resourse

tiagoposse:
@dsever correct if I'm wrong, but through `extraSecrets`, the data will end up in a kubernetes secret, which I'm trying to avoid. Or is the idea to pass something like:
```
extraSecrets:
[...]

PR #8385, opened 2023-07-15
support csi volumes in helm chart

tiagoposse:
This PR:

Adds support for csi volumes, extending currently supported types
Add .Values.externalDatabasePassword and .Values.externalBrokerPassword to remove mandatory usage of secrets from the chart
By adding the two values, it becomes possible to pass DD_DATABASE_PASSWORD and DD_CELERY_BROKER_PASSWORD via csi volumes and external secret systems (e.g. vault).

Closes https://github.com/DefectDojo/django-DefectDojo/issues/5535

### DIFF
--- a/helm/defectdojo/templates/celery-beat-deployment.yaml
+++ b/helm/defectdojo/templates/celery-beat-deployment.yaml
@@ -61,6 +61,11 @@ spec:
           {{- else if (eq .type "hostPath") }}
           type: {{ .pathType | default "Directory" }}
           path: {{ .hostPath }}
+          {{- else if (eq .type "csi") }}
+          driver: secrets-store.csi.k8s.io
+          readOnly: true
+          volumeAttributes:
+            secretProviderClass: {{ .secretProviderClass }}
           {{- end }}
       {{- end }}
       containers:
@@ -114,6 +119,7 @@ spec:
             name: {{ $fullName }}-extrasecrets
             optional: true
         env:
+        {{- if not .Values.externalBrokerPassword }}
         - name: DD_CELERY_BROKER_PASSWORD
           valueFrom:
             secretKeyRef:
@@ -124,6 +130,8 @@ spec:
               name: {{ .Values.redis.auth.existingSecret | default "defectdojo-redis-specific" }}
               key: {{ .Values.redis.auth.existingSecretPasswordKey | default "redis-password" }}
             {{- end }}
+        {{- end }}
+        {{- if not .Values.externalDatabasePassword }}
         - name: DD_DATABASE_PASSWORD
           valueFrom:
             secretKeyRef:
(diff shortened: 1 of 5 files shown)


alles-klar (approved):
I assume you tested it. The code looks good to me if you fix the linting issues.

dsever:
I have mix-feelings about it, helm needs to be generic and i this case it is addressing CSI as secret-csi, when csi is general term in k8s and there are plenty of CSI drivers (https://kubernetes-csi.github.io/docs/drivers.html). 

Can we still preserve same functionalities to enable user to consume CSI driver but in generic way, like specify driver as well?

cneill (commented):
There are 2 main problems I have here:

* This solves a narrow use case (only supports the "Secrets Store" CSI driver) and adds complexity to `extraVolumes` without sufficient documentation to explain the use case it solves for
* I don't believe the change to `externalDatabasePassword` / `externalBrokerPassword` will work as intended. Unless there is additional configuration in the Secrets Store driver that allows for automatic injection of environment variables based on the volume mount, it seems we would still need an `env`/`envFrom` ([ref](https://secrets-store-csi-driver.sigs.k8s.io/topics/set-as-env-var.html)). If a specific driver configuration is required, that should be documented. If I've misunderstood how the driver works, please let me know.

Zooming out a bit: the more options we support with the `extraVolumes` mechanism, the more I want to do away with it entirely and just let the user specify the YAML directly in the `volumes` and `volumeMounts` blocks. Right now it solves for a small subset of total functionality but saves the user from having to address both `volumes` and `volumeMounts` on their own. If we start supporting these as one-offs, we might ultimately have to add support for all the rest of the [29 VolumeSource options](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#volume-v1-core) and/or the [121 production CSI driver options](https://kubernetes-csi.github.io/docs/drivers.html).

I'm open to ideas on how to make `extraVolumes` more flexible, but copying and pasting more options into every `Deployment`/etc doesn't seem like the right way to go.

Maffooch:
Closing due to stale. If this is an error, please reopen and see provided feedback

petrzjunior:
@Maffooch I would like to see this reopened/reworked/finished.

We too use Secret Manager CSI driver to mount secrets but this volume type is not allowed. I totally agree with @cneill

> the more I want to do away with it entirely and just let the user specify the YAML directly in the volumes and volumeMounts blocks

This is how Helm charts usually handle this. For example Bitnami let's user specify a free-form:
- `extraVolumes`: https://github.com/bitnami/charts/blob/main/bitnami/redis/templates/master/application.yaml#L261-L263
- `extraVolumeMounts`: https://github.com/bitnami/charts/blob/main/bitnami/redis/templates/master/application.yaml#L484-L486

It is a good practice for Helm charts to allow adding basically all keys in all Kubernetes manifests in case someone needs to add something. Honestly, the user usually known best how what the actual YAML looks like. From my experience, I usually need to tune quite a lot of knobs in the Helm charts I use due to the constrained (security) environment I work in.

PR #8431, opened 2023-07-26
OpenAPI validator: Fetch binary from the official docker image

kiblik:
Original implementation downloaded script responsible for compilation of `openapi-generator-cli` (during first run).

New implementation copy precompiled binary from the official Docker images.

Benefits:
- Renovate is able to detect the newly released `openapi-generator`. The previous version of pinning is not supported by Renovate.
  - If the new version of `openapi-generator` breaks integration tests, it happens only in Renovate pull request.
- Shorter build time (`openapi-generator-cli` does not need to be compiled)
- Smaller image (compiler and useless artefact are not included)


cneill:
This seems like it could still result in breaking changes when they bump the image since there's an implicit `:latest` here, and this relies on them not changing the location of the jar within the Docker image. It also means we won't get a log message when there's a new version we _might_ want to bump to. I lean towards leaving things as they are now with the explicit version pin, but if others feel strongly about it we can change it.

Dockerfile.integration-tests-debian
```diff
@@ -1,6 +1,7 @@
 
 # code: language=Dockerfile
 
+FROM openapitools/openapi-generator-cli:v6.6.0@sha256:54381220aecf2e77bb4b6694c4e1a03e733b49453292cd1af6f48b510f1f008a as openapitools
```
kiblik, line 4: > This seems like it could still result in breaking changes when they bump the image since there's an implicit :latest here, and this relies on them not changing the location of the jar within the Docker image. It also means we won't get a log message when there's a new version we might want to bump to. I lean towards leaving things as they are now with the explicit version pin, but if others feel strongly about it we can change it.

@cneill, but the version is explicitly pinned by this line. In this case `6.6.0`

cneill: Oops! Sorry, I somehow missed this line on my first read of the changes. Works for me.

cneill (approved):

mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

PR #8545, opened 2023-08-22
Endpoint model/view corrections

Maffooch:
This PR fixes a handful of issues observed with the endpoint model and it's views

## Endpoint Equality across products

Deleting a product type with two products that have endpoints with matching contents (protocol + host + port + path + query) result in a mixup of which endpoint belongs where:

Screenshot 2023-08-21 at 11 20 06 AM

This is the error produced:
```
update or delete on table "dojo_product" violates foreign key constraint "dojo_endpoint_product_id_b86dc143_fk_dojo_product_id" on table "dojo_endpoint"
DETAIL:  Key (id)=(10) is still referenced from table "dojo_endpoint".
```

This issue originates from the endpoint equality method. An endpoint was deemed equal to another if the string representation was the same, but failed to take into account that the endpoints could be from different products. The solution there is to incorporate product matching if they are specified. Instances of unsaved endpoints still use the equality method when determining if a given finding already has existing endpoints (at this time, one finding has as a product, and one does not)

## Endpoint filtering on the Findings page

At some point, this filter was removed for the sake of performance (to prevent the page loading all endpoint objectsts in the system on each request) but was actually necessary when filtering findings by endpoint. The correct this, I added a hidden option for "endpoints" on the FindingFilter. This change allows clicking the number of active/verified findings in the screenshot below to actually work:

Screenshot 2023-08-22 at 11 28 15 AM

## Endpoint Views query "Active Verified" findings

Fix multiple places in the endpoint UI that reference "Active Verified" findings, but the query to fetch those findings does not filter for verified findings at all. I added matching helping methods for 
- `active_findings` to also include `active_verified_findings`
- `active_findings_count` to also include `active_verified_findings_count`
- `host_active_findings` to also include `host_active_verified_findings`
- `host_active_findings_count` to also include `host_active_verified_findings_count`

### DIFF
--- a/dojo/endpoint/views.py
+++ b/dojo/endpoint/views.py
@@ -124,12 +124,12 @@ def process_endpoint_view(request, eid, host_view=False):
         endpoints = endpoint.host_endpoints()
         endpoint_metadata = None
         all_findings = endpoint.host_findings()
-        active_findings = endpoint.host_active_findings()
+        active_verified_findings = endpoint.host_active_verified_findings()
     else:
         endpoints = None
         endpoint_metadata = dict(endpoint.endpoint_meta.values_list('name', 'value'))
         all_findings = endpoint.findings.all()
-        active_findings = endpoint.active_findings()
+        active_verified_findings = endpoint.active_verified_findings()
 
     if all_findings:
         start_date = timezone.make_aware(datetime.combine(all_findings.last().date, datetime.min.time()))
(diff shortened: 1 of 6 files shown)


cneill (approved):

blakeaowens (approved):

mtesauro (approved):
Approved

kiblik:
Can I ask to write some tests for it [here](https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/test_endpoint_model.py), please?

Maffooch:
@kiblik of course you can ask! You could also request changes on the PR as well. We value your opinion :)

kiblik:
> @kiblik of course you can ask! You could also request changes on the PR as well. We value your opinion :)

I meant exactly this set of tests https://github.com/DefectDojo/django-DefectDojo/pull/8545/commits/3cc1bbaf9bf2cf070619a0d90019509444ce9155 :)

For now, the is nothing else on my mind. Thank you.

mtesauro:
@kiblik Are you good to approve this one now?

kiblik:
> @kiblik Are you good to approve this one now?

I do not have permission to approve PRs but I'm good for me.

grendel513 (approved):

PR #8547, opened 2023-08-23
updated helm chart to allow setting a different port for external redis

Crayeth:
**Description**

I've updated the values.yaml file and the configmap template in the helm chart to be able to set a different port for external redis. It was hardcoded to the default 6379.

**Test results**

tested it with empty values.yaml (no external redis), values.yaml with an external redis without specifying port (defaulted to 6379) and with port 6380 (azure cache for redis ssl port).

**Documentation**

documentation added inline to values.yaml

**Checklist**


cneill (approved):
Thanks for the update!

grendel513 (approved):

mtesauro (approved):
Approved

Maffooch (approved):

PR #8574, opened 2023-08-29
fix 500 error when attempting to order tags

FelixHernandez:
## Fix 500 error when attempting to order tags

**Description**

This bugfix validates the endpoint string value for the request in the server and only applies if a correct value exists to make the filter by endpoint

**Test results**

The test now avoids the 500 error that was showing the filter before with the same request.

### DIFF
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -269,7 +269,7 @@ def findings(
     endpoint = None
     if "endpoints" in request.GET:
         endpoints = request.GET.getlist("endpoints", [])
-        if len(endpoints) == 1:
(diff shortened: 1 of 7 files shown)


cneill (approved):

Maffooch (approved):

grendel513 (approved):

devGregA (approved):

coheigea:
@Maffooch FYI this issue is not just for ordering tags, it's when you search for anything in the findings page (for example if you select a particular test type). Could we get a bugfix out for it sooner rather than later, as finding search is completely broken?

PR #8594, opened 2023-09-01
Bugfix Bulk edit drop-down

FelixHernandez:
## Bugfix Bulk edit drop-down

**Description**

This bugfix in the severity dropdown remains the parent opened to select more options from the same form. Before the fix, the main dropdown closes when the user selected an option from a nested dropdown.


mtesauro (approved):
Approved

blakeaowens (approved):
Tested and works fine for me. Thanks!

cneill (approved):

grendel513 (approved):

PR #8599, opened 2023-09-04
Update celery to 5.3.4

kiblik:


mtesauro:
@kiblik Thanks for this - I went ahead and merged the one from Dependabot as I saw that one first. 

I'm going to close this since it's sorted out.

PR #8622, opened 2023-09-06
Feature/add django csp

renejal:


kiblik:
I would like to double-check https://github.com/DefectDojo/django-DefectDojo/blob/master/readme-docs/CONTRIBUTING.md#submitting-pull-requests

PR #8662, opened 2023-09-13
Fix Product Search to Filter When You Search "Product"

FelixHernandez:
## Fix Product Search to Filter When You Search "Product"
[sc-1447]
**Description**

This bugfix allows the user easily fiind the titled "Product" products. 
**Test results**
When the user types "Product" or with following spaces (case insensitive) the search is in the product column.
image

### DIFF
--- a/dojo/templates/dojo/product.html
+++ b/dojo/templates/dojo/product.html
@@ -314,7 +314,6 @@
     <script>
         // DataTables setup
         $(document).ready(function() {
-            var productFilterCalled = false;
             date =  new Date().toISOString().slice(0, 10);
             var fileDated = 'Product_List_' + date;
             var buttonCommon = {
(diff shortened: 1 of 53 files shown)


dojo/templates/dojo/product.html
```diff
@@ -314,6 +314,7 @@ <h3 class="has-filters">
     <script>
         // DataTables setup
         $(document).ready(function() {
+            var productFilterCalled = false;
```
cneill: Looks like this line can be removed now
```suggestion

```

cneill (approved):

grendel513 (approved):

Maffooch (approved):

mtesauro (approved):
Approved

PR #8694, opened 2023-09-20
Allow POST for risk_acceptance + unittest

kiblik:


kiblik:
Implemented in https://github.com/DefectDojo/django-DefectDojo/pull/9415

PR #8707, opened 2023-09-21
Add a "tag findings" parameter to the import-scan and reimport-scan API endpoints

FelixHernandez:
## Add a "tag findings" parameter to the import-scan and reimport-scan API endpoints
[sc-2329]
**Description**
This feature adds a new parameter to import-scan and reimport-scan API endpoints the new parameter is `apply_tags_to_findings` False as default.


grendel513 (approved):

dojo/api_v2/serializers.py
```diff
@@ -2130,6 +2130,11 @@ class ImportScanSerializer(serializers.Serializer):
     product_type_id = serializers.IntegerField(read_only=True)
 
     statistics = ImportStatisticsSerializer(read_only=True, required=False)
+    apply_tags_to_findings = serializers.BooleanField(
+        help_text="If set to True, the tags will be applied to the findings",
+        required=False,
+        default=True,
+    )
```
Maffooch, line 2137: How is this new variable being used in the `import_scan` and `reimport_scan` functions?

dojo/importers/importer/importer.py
```diff
@@ -363,6 +363,9 @@ def import_scan(self, scan, scan_type, engagement, lead, environment, active=Non
             test_import = importer_utils.update_import_history(Test_Import.IMPORT_TYPE, active, verified, tags, minimum_severity,
                                                                 endpoints_to_add, version, branch_tag, build_id, commit_hash,
                                                                 push_to_jira, close_old_findings, test, new_findings, closed_findings)
+            if apply_tags_to_findings:
+                for finding in test_import.findings_affected:
+                    finding.tags.set(tags)
```
Maffooch, line 368: Using `.set` here could override tags that may have been set at the parser level. It would be better to iterate over the tags list and add them the finding that way to ensure this operation is strictly additive

dojo/importers/reimporter/reimporter.py
```diff
@@ -69,6 +70,8 @@ def process_parsed_findings(
             deduplication_algorithm,
         )
         for item in items:
+            if apply_tags_to_findings:
+                item.tags.set(tags)
```
Maffooch, line 74: Is this needed here as well as the `reimport_scan` function? I think it would be best to remove this instance of adding the tags the findings

Maffooch (changes requested):

dojo/importers/importer/importer.py
FelixHernandez: I updated the code to add the tag list instead set

dojo/importers/reimporter/reimporter.py
grendel513: I do think it is a good feature to have on re-imports.  It provides users the opportunity to apply tags that either were not available or they missed during the initial import.

mtesauro (approved):
Approved

Maffooch (changes requested):
Requesting changes as they were not for any UI work. @FelixHernandez please see the unresolved comment on my previous review. Pasting it here as well

Maffooch (approved):

grendel513 (approved):

cneill (approved):

PR #8728, opened 2023-09-25
Sanitization + Modernization: Findings, Groups, Notifications

Maffooch:
In an effort to have a better understanding of what behaviors are currently supported, and in what capacity, this PR converts some of the larger function based views into class based views. This cleanup/modernization work is done to help prepare the code for larger v3 changes in the future.


dojo/group/urls.py
```diff
@@ -3,11 +3,11 @@
...
-    re_path(r'^group$', views.group, name='groups'),
-    re_path(r'^group/(?P<gid>\d+)$', views.view_group, name='view_group'),
-    re_path(r'^group/(?P<gid>\d+)/edit$', views.edit_group, name='edit_group'),
-    re_path(r'^group/(?P<gid>\d+)/delete', views.delete_group, name='delete_group'),
-    re_path(r'^group/add$', views.add_group, name='add_group'),
+    re_path(r'^group$', views.ListGroups.as_view(), name='groups'),
+    re_path(r'^group/add$', views.AddGroup.as_view(), name='add_group'),
+    re_path(r'^group/(?P<group_id>\d+)$', views.ViewGroup.as_view(), name='view_group'),
+    re_path(r'^group/(?P<group_id>\d+)/edit$', views.EditGroup.as_view(), name='edit_group'),
+    re_path(r'^group/(?P<group_id>\d+)/delete', views.DeleteGroup.as_view(), name='delete_group'),
```
cneill, line 10: I'm pretty indifferent on the `$` termination on these patterns, but we should probably be consistent across these?

dojo/finding/urls.py
```diff
@@ -3,56 +3,125 @@
...
+    ),
+    re_path(
+        r'^product/(?P<product_id>\d+)/finding/open$',
+        views.ListOpenFindings.as_view(),
+        name='product_open_findings'
+    ),
+    re_path(
+        r'^product/(?P<product_id>\d+)/findings$',
+        views.ListOpenFindings.as_view(),
+        name='view_product_findings_old'
```
cneill, line 56: Should this be `views.ListFindings` to match `/finding/`? Also, what makes this one `_old`?

dojo/group/urls.py
Maffooch: Consistency would be good. The inconsistency was there from before (the diff is a liar 🤔) but I can update the URLs to all match

dojo/finding/views.py
```diff
@@ -435,207 +260,940 @@ def prefetch_for_similar_findings(findings):
...
+
+    def get_template(self):
+        return "dojo/findings_list.html"
+
+    def add_breadcrumbs(self, request: HttpRequest, context: dict):
+        # show custom breadcrumb if user has filtered by exactly 1 endpoint
+        if "endpoints" in request.GET:
+            endpoints = request.GET.getlist("endpoints", [])
+            if len(endpoints) == 1 and endpoints[0] != '':
+                endpoint = endpoints[0]
```
cneill, line 407: To make it clearer that we're dealing with an ID here rather than an endpoint object
```suggestion
                endpoint_id = endpoints[0]
```

cneill (commented):
A few relatively minor comments here..

dojo/notifications/helper.py
```diff
@@ -264,29 +264,32 @@ def send_msteams_notification(event, user=None, *args, **kwargs):
         pass
 
 
-@dojo_async_task
-@app.task
+# @dojo_async_task
+# @app.task
```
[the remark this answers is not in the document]
Maffooch: Good catch. Should not be removed. Needs to be uncommented

dojo/finding/views.py
Maffooch: Good call. Making more changes around this to further clarify

mtesauro (approved):
Approved.

Note: AccessLint seems stuck - it's dropped a comment that it's OK with this PR so moving forward with things.

grendel513 (approved):

cneill (approved):

blakeaowens (approved):

PR #8735, opened 2023-09-26
Excel Export throws 500 error code

FelixHernandez:
## Excel Export throws 500 error code
[sc-2338]

**Description**
This fix allows the user export to excel and csv the findings properly. The error is now wrapped in a `try...catch` statement and the exception is being logged for visibility.

### DIFF
--- a/dojo/reports/views.py
+++ b/dojo/reports/views.py
@@ -871,19 +871,23 @@ def csv_export(request):
         if not first_row:
             fields = []
             for key in dir(finding):
-                if key not in excludes_list and (not callable(getattr(finding, key)) or key in allowed_attributes) and not key.startswith('_'):
-                    if not callable(getattr(finding, key)):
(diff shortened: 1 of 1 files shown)


grendel513 (approved):

Maffooch (changes requested):
The same fix will need to be applied when inserting values into each row:

CSV: https://github.com/DefectDojo/django-DefectDojo/blob/f3e590453bfc5dc719b5f2f1d82196dfdb7beab4/dojo/reports/views.py#L870-L882
Excel: https://github.com/DefectDojo/django-DefectDojo/blob/f3e590453bfc5dc719b5f2f1d82196dfdb7beab4/dojo/reports/views.py#L968-L981

Maffooch (approved):

cneill (approved):

mtesauro (approved):
Approved

PR #8764, opened 2023-09-29
add parser documentation for burp scan & burp-enterprise

paulOsinski:
Adding Parser documentation for Burp Scan & Burp Enterprise.  I'm not sure how to make a good example of acceptable HTML for the Burp-Enterprise Parser, and I don't know how valuable that would be for debugging anyway.  XML format is much easier to understand and correct errors with.


Maffooch (approved):

cneill (approved):

mtesauro (approved):
Approved

PR #8765, opened 2023-09-29
Validate undefined and null value in criticality column table products

FelixHernandez:
## Validate undefined and null value in criticality column table products
[sc-2297]

**Description**
Now the code validates that `data`  within render function, in criticality column, should have a value. According to datatables documentation:  https://datatables.net/manual/tech-notes/4#Resolution

### DIFF
--- a/dojo/templates/dojo/product.html
+++ b/dojo/templates/dojo/product.html
@@ -352,11 +352,9 @@
                       };
                       if (type === 'sort') {
                         return criticals[getDojoExportValueFromTag(data, 'i', 'data-content')]
-                      }
-                      else if (type === 'export') {
+                      } else if (type === 'export') {
                           return getDojoExportValueFromTag(data, 'i', 'data-content')
-                      }
-                      else if (data != undefined||data != null) {
+                      } else if (data != undefined||data != null) {
                           return data
                       }
                       return ""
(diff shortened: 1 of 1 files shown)


dojo/templates/dojo/product.html
```diff
@@ -350,8 +350,16 @@ <h3 class="has-filters">
...
+                      if (type === 'sort') {
+                        return criticals[getDojoExportValueFromTag(data, 'i', 'data-content')]
+                      }
+                      else if (type === 'export') {
+                          return getDojoExportValueFromTag(data, 'i', 'data-content')
+                      }
+                      else if (data != undefined||data != null) {
+                          return data
+                      }
+                      return ""
```
cneill, line 362: Just a minor styling suggestion:

```suggestion
                      if (type === 'sort') {
                        return criticals[getDojoExportValueFromTag(data, 'i', 'data-content')]
                      } else if (type === 'export') {
                          return getDojoExportValueFromTag(data, 'i', 'data-content')
                      } else if (data != undefined||data != null) {
                          return data
                      }
                      return ""
```

Maffooch: Agreed with this request

FelixHernandez: Changes made in this [commit](https://github.com/DefectDojo/django-DefectDojo/pull/8765/commits/cc45195efc05563e006525dc5cdfe537ecf5f7dc)

cneill (approved):

mtesauro (approved):
Approved

grendel513 (approved):

devGregA (approved):

PR #8783, opened 2023-10-03
Browser tests should use the latest Chrome stable release

dwalleck:
**Description**

This PR is a follow-up to #8755. There is no guarantee that the version of Chrome installed via apt has a valid chromedriver version. apt also does not keep previous versions of Chrome available for installation. To ensure that we always get a valid pairing of Chrome and chromedriver, this change instead targets the latest stable version of Chrome and chromedriver.

**Test results**

TBD

### DIFF
--- a/Dockerfile.integration-tests-debian
+++ b/Dockerfile.integration-tests-debian
@@ -66,8 +66,6 @@ RUN \
 
 RUN pip install --no-cache-dir selenium==4.9.0 requests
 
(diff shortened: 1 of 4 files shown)


mtesauro (approved):
Approved

grendel513 (approved):

cneill (approved):

blakeaowens (approved):

PR #8824, opened 2023-10-13
Consolidation of notification creation

kiblik:
Purpose of this PR: If some action triggers the creation of a notification during clicking in UI. The same notification should be triggered also if the same happens via API. Until now many actions have been avoiding these notifications. Implementation is done by Django Signals.

Some notes:
- General rule: if the object is removed in cascade (e.g. removal product triggers removal of all engagements, tests, ...) notification is created only for the highest removed element (to avoid spamming).
- The only exception to this ^ general rule is the removal of the Product. If Product Type is removed, notification is triggered for all Products as well
- PR hasn't announced any new notifications. All of them have been created based on existing implementation.
- There is a set of notifications which hasn't been migrated and they are performed only for UI actions because they would generate useless noise. E.g. "creation of findings" vs "creation of findings during scan imports". But "TODO" comments have been added.
- If AuditLog is enabled, the message might contain also the actor's name. If the user disabled Auditlog, the user will be notified without knowing who did it. This is a small step back compared to the previous implementation.
- Most of the actions have been with the label `event="other"`. Now notification engine accepts any event name as meaningful and for Alerts event is translated to source name. If event type does not exist as the field in Notification, the engine works with the event as with `other`

### DIFF
--- a/dojo/endpoint/signals.py
+++ b/dojo/endpoint/signals.py
@@ -15,9 +15,9 @@ def endpoint_post_delete(sender, instance, using, origin, **kwargs):
     if instance == origin:
         if settings.ENABLE_AUDITLOG:
             le = LogEntry.objects.get(
-                    action=LogEntry.Action.DELETE,
-                    content_type=ContentType.objects.get(app_label='dojo', model='endpoint'),
-                    object_id=instance.id
+                action=LogEntry.Action.DELETE,
+                content_type=ContentType.objects.get(app_label='dojo', model='endpoint'),
+                object_id=instance.id
             )
             description = _('The endpoint "%(name)s" was deleted by %(user)s') % {
                                 'name': str(instance), 'user': le.actor}
--- a/dojo/endpoint/signals.py
+++ b/dojo/endpoint/signals.py
@@ -1,4 +1,5 @@
 from auditlog.models import LogEntry
+from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.db.models.signals import post_delete
 from django.dispatch import receiver
@@ -7,13 +8,12 @@ from django.utils.translation import gettext as _
 
 from dojo.models import Endpoint
 from dojo.notifications.helper import create_notification
-from dojo.utils import get_system_setting
 
 
 @receiver(post_delete, sender=Endpoint)
 def endpoint_post_delete(sender, instance, using, origin, **kwargs):
     if instance == origin:
-        if get_system_setting('enable_auditlog'):
+        if settings.ENABLE_AUDITLOG:
             le = LogEntry.objects.get(
                     action=LogEntry.Action.DELETE,
                     content_type=ContentType.objects.get(app_label='dojo', model='endpoint'),
--- a/dojo/endpoint/signals.py
+++ b/dojo/endpoint/signals.py
@@ -1,0 +1,30 @@
+from auditlog.models import LogEntry
+from django.contrib.contenttypes.models import ContentType
+from django.db.models.signals import post_delete
+from django.dispatch import receiver
+from django.urls import reverse
+from django.utils.translation import gettext as _
+
+from dojo.models import Endpoint
+from dojo.notifications.helper import create_notification
+from dojo.utils import get_system_setting
+
+
+@receiver(post_delete, sender=Endpoint)
+def endpoint_post_delete(sender, instance, using, origin, **kwargs):
+    if instance == origin:
+        if get_system_setting('enable_auditlog'):
+            le = LogEntry.objects.get(
+                    action=LogEntry.Action.DELETE,
+                    content_type=ContentType.objects.get(app_label='dojo', model='endpoint'),
+                    object_id=instance.id
+            )
+            description = _('The endpoint "%(name)s" was deleted by %(user)s') % {
+                                'name': str(instance), 'user': le.actor}
+        else:
+            description = _('The endpoint "%(name)s" was deleted') % {'name': str(instance)}
+        create_notification(event='endpoint_deleted',  # template does not exists, it will default to "other" but this event name needs to stay because of unit testing
+                            title=_('Deletion of %(name)s') % {'name': str(instance)},
+                            description=description,
+                            url=reverse('endpoint'),
+                            icon="exclamation-triangle")
--- a/dojo/engagement/signals.py
+++ b/dojo/engagement/signals.py
@@ -40,9 +40,9 @@ def engagement_post_delete(sender, instance, using, origin, **kwargs):
     if instance == origin:
         if settings.ENABLE_AUDITLOG:
             le = LogEntry.objects.get(
-                    action=LogEntry.Action.DELETE,
-                    content_type=ContentType.objects.get(app_label='dojo', model='engagement'),
-                    object_id=instance.id
+                action=LogEntry.Action.DELETE,
+                content_type=ContentType.objects.get(app_label='dojo', model='engagement'),
+                object_id=instance.id
             )
             description = _('The engagement "%(name)s" was deleted by %(user)s') % {
                                 'name': instance.name, 'user': le.actor}
--- a/dojo/engagement/signals.py
+++ b/dojo/engagement/signals.py
@@ -1,0 +1,57 @@
+from auditlog.models import LogEntry
+from django.contrib.contenttypes.models import ContentType
+from django.db.models.signals import post_delete, post_save, pre_save
+from django.dispatch import receiver
+from django.urls import reverse
+from django.utils.translation import gettext as _
+
+from dojo.models import Engagement
+from dojo.notifications.helper import create_notification
+from dojo.utils import get_system_setting
+
+
+@receiver(post_save, sender=Engagement)
+def engagement_post_save(sender, instance, created, **kwargs):
+    if created:
+        title = 'Engagement created for ' + str(instance.product) + ': ' + str(instance.name)
+        create_notification(event='engagement_added', title=title, engagement=instance, product=instance.product,
+                            url=reverse('view_engagement', args=(instance.id,)))
+
+
+@receiver(pre_save, sender=Engagement)
+def engagement_pre_save(sender, instance, **kwargs):
+    old = sender.objects.filter(pk=instance.pk).first()
+    if old and instance.status != old.status:
+        if instance.status in ["Cancelled", "Completed"]:
+            create_notification(event='close_engagement',
+                                title=_('Closure of %s') % instance.name,
+                                description=_('The engagement "%s" was closed') % (instance.name),
+                                engagement=instance, url=reverse('engagement_all_findings', args=(instance.id, )))
+        elif instance.status in ["In Progress"]:
+            create_notification(event='reopen_engagement',
+                                title=_('Reopening of %s') % instance.name,
+                                engagement=instance,
+                                description=_('The engagement "%s" was reopened') % (instance.name),
+                                url=reverse('view_engagement', args=(instance.id, )))
+
+
+@receiver(post_delete, sender=Engagement)
+def engagement_post_delete(sender, instance, using, origin, **kwargs):
+    if instance == origin:
+        if get_system_setting('enable_auditlog'):
+            le = LogEntry.objects.get(
+                    action=LogEntry.Action.DELETE,
+                    content_type=ContentType.objects.get(app_label='dojo', model='engagement'),
+                    object_id=instance.id
+            )
+            description = _('The engagement "%(name)s" was deleted by %(user)s') % {
+                                'name': instance.name, 'user': le.actor}
+        else:
+            description = _('The engagement "%(name)s" was deleted') % {'name': instance.name}
+        create_notification(event='engagement_deleted',  # template does not exists, it will default to "other" but this event name needs to stay because of unit testing
+                            title=_('Deletion of %(name)s') % {'name': instance.name},
+                            description=description,
+                            product=instance.product,
+                            url=reverse('view_product', args=(instance.product.id, )),
+                            recipients=[instance.lead],
+                            icon="exclamation-triangle")
--- a/dojo/engagement/signals.py
+++ b/dojo/engagement/signals.py
@@ -27,7 +27,7 @@ def engagement_pre_save(sender, instance, **kwargs):
                                 title=_('Closure of %s') % instance.name,
                                 description=_('The engagement "%s" was closed') % (instance.name),
                                 engagement=instance, url=reverse('engagement_all_findings', args=(instance.id, )))
-        elif instance.status in ["In Progress"]:
+        elif instance.status in ["In Progress"] and old.status not in ["Not Started"]:
             create_notification(event='engagement_reopened',
                                 title=_('Reopening of %s') % instance.name,
                                 engagement=instance,
--- a/dojo/engagement/signals.py
+++ b/dojo/engagement/signals.py
@@ -23,12 +23,12 @@ def engagement_pre_save(sender, instance, **kwargs):
     old = sender.objects.filter(pk=instance.pk).first()
     if old and instance.status != old.status:
         if instance.status in ["Cancelled", "Completed"]:
-            create_notification(event='close_engagement',
+            create_notification(event='engagement_closed',
                                 title=_('Closure of %s') % instance.name,
                                 description=_('The engagement "%s" was closed') % (instance.name),
                                 engagement=instance, url=reverse('engagement_all_findings', args=(instance.id, )))
         elif instance.status in ["In Progress"]:
-            create_notification(event='reopen_engagement',
+            create_notification(event='engagement_reopened',
                                 title=_('Reopening of %s') % instance.name,
                                 engagement=instance,
                                 description=_('The engagement "%s" was reopened') % (instance.name),
--- a/dojo/engagement/signals.py
+++ b/dojo/engagement/signals.py
@@ -1,4 +1,5 @@
 from auditlog.models import LogEntry
+from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.db.models.signals import post_delete, post_save, pre_save
 from django.dispatch import receiver
@@ -7,7 +8,6 @@ from django.utils.translation import gettext as _
 
 from dojo.models import Engagement
 from dojo.notifications.helper import create_notification
-from dojo.utils import get_system_setting
 
 
 @receiver(post_save, sender=Engagement)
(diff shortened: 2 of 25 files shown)


kiblik:
Hi @grendel513 / @mtesauro / @devGregA / @Maffooch / @cneill / @blakeaowens. I would like to ask for a review.

dojo/finding/views.py
Maffooch, line 1: Leaving the finding notifications in the view is a good choice. Otherwise, most reimports would flood a users notifications with closes/reopens/creations

With that said, I am sorta against moving the notification operation to the signal. I think revising the TODO comments to explain why it should not be moved to the signals would be preferable to the "this might be moved to signals" comment. Someone else in the future may act on that todo

dojo/jira_link/views.py
Maffooch, line 1: Good improvement migrating from the blanket "other" event. However, now it is just a more specific blanket "jira" event. Please add a bit more detail about each operation. Here are some suggestions:
- jira_comment
- jira_config_added
- jira_config_edited
- jira_config_deleted

unittests/test_rest_framework.py
```diff
@@ -1586,7 +1586,7 @@ def __init__(self, *args, **kwargs):
         }
         self.update_fields = {"first_name": "test changed", "configuration_permissions": [219, 220]}
         self.test_type = TestType.CONFIGURATION_PERMISSIONS
-        self.deleted_objects = 19
+        self.deleted_objects = 25
```
Maffooch, line 1589: Is this increase due to more Alert objects existing in the db due to better notification coverage?

Maffooch (commented):
This is really solid work! I left a few nitpicky things during an initial code review before pulling it down and giving it a proper test. There is one thing I have not figured out though. How is this working:

>General rule: if the object is removed in cascade (e.g. removal product triggers removal of all engagements, tests, ...) notification is created only for the highest removed element (to avoid spamming). The only exception to this ^ general rule is the removal of the Product. If Product Type is removed, notification is triggered for all Products as well

I am guessing that when a child object is deleted via cascade from a parent object, the signal is not being sent from the child model, but rather the parent model. In an example of a product being deleted, an engagement deletion signal would not be sent because the engagement model is not sending signal, the product is. Is that correct?

If so, I did not find the accommodations of all product deletion notifications being dispatched when a product type is deleted

Maffooch:
I have tested locally, and everything appears to be working well :)

Once the tests are passing (it is just the new test you have added that is failing) and the other nitpicky things are added, I think this one is good to go!

kiblik:
> This is really solid work! I left a few nitpicky things during an initial code review before pulling it down and giving it a proper test. There is one thing I have not figured out though. How is this working:
> 
> > General rule: if the object is removed in cascade (e.g. removal product triggers removal of all engagements, tests, ...) notification is created only for the highest removed element (to avoid spamming). The only exception to this ^ general rule is the removal of the Product. If Product Type is removed, notification is triggered for all Products as well
> 
> 
> I am guessing that when a child object is deleted via cascade from a parent object, the signal is not being sent from the child model, but rather the parent model. In an example of a product being deleted, an engagement deletion signal would not be sent because the engagement model is not sending signal, the product is. Is that correct?
> 
> If so, I did not find the accommodations of all product deletion notifications being dispatched when a product type is deleted

@Maffooch, yes, you are right. That is correct. The difference is in `if instance == origin:`. In all models (except `Prod` and `Prod_Type`), notification is created only if the triggering model is the same as the model that will be removed. Mentioned condition is missing in `Prod` and `Prod_Type`. So notification will be created, does not matter, who triggered the deletion.

dojo/finding/views.py
kiblik: Is this better? https://github.com/DefectDojo/django-DefectDojo/pull/8824/commits/284c85d0232d5d74667e842ff3f75fa54165ba30

dojo/jira_link/views.py
kiblik: Sounds good https://github.com/DefectDojo/django-DefectDojo/pull/8824/commits/63b824145a2e0d6091810eda94ddfa06ad2ca793

unittests/test_rest_framework.py
kiblik: Yes, exactly.

dojo/finding/views.py
Maffooch: Yep that works!

Maffooch:
> @Maffooch, yes, you are right. That is correct. The difference is in if instance == origin:. In all models (except Prod and Prod_Type), notification is created only if the triggering model is the same as the model that will be removed. Mentioned condition is missing in Prod and Prod_Type. So notification will be created, does not matter, who triggered the deletion.

Aha that would explain why I was not finding it lol thanks for the clarification! I see what you're talking about now :)

Maffooch:
@kiblik I will approve after unit tests are passing

kiblik:
@Maffooch (and others :))
I finally had enough time to review, why this PR is failing.

The unittest which I added is not going through.

https://github.com/DefectDojo/django-DefectDojo/blob/4fad5a2589fc7cfe081ab822e5a959c700db4fe2/unittests/test_notifications.py#L67-L71

I found out it is because of the change announced in #9227.
>- Overrides user notification settings for request reviews with system level notification settings
>    - Before this change, if the system level notifications for finding review requests were enabled, but a given user did not have the notifications enabled on their personal scope, then that user would not receive a notification. This means that the system level notifications were being ignored for this action

I'm not sure (correct me if I'm wrong) but I suppose implementation in #9227 used a misinterpretation of "system notification". From what I know, "system notification" is not a set of rules which overrides the user's setting but it is a separate category of notification settings. It is connected to the whole system.
Some proofs:
- https://github.com/DefectDojo/django-DefectDojo/blob/82181253cd303efd1642994327e1d862c6b247ef/dojo/notifications/helper.py#L72-L78
- https://github.com/DefectDojo/django-DefectDojo/blob/82181253cd303efd1642994327e1d862c6b247ef/dojo/templates/dojo/notifications.html#L49-L52
- https://documentation.defectdojo.com/integrations/notifications/#microsoft-teams
  > Microsoft Teams does not provide an easy way to send messages to a personal channel. Therefore, DefectDojo can only send system scope notifications to Microsoft Teams.

In my opinion, if the user knowingly turns off some kind of notification, he/she/they should not receive it.

If we would like to be sure that all users receive notifications by default, set "notification templates" to them.
Yes, this does apply to users who already exist. In that case:
1. Update setting to all of them (e.g. via script and API)
2. Implement a more advanced editor for notification settings
3. Introduce a new type of notification settings (we already have "template" or "system"). We can call it "overrides". But please do not override the user's settings with system settings. Btw, right now, "system notifications" are performing only positive overrides - you can force users to receive all notifications from some category; you are not able to force them to not receive some category of notification. 

I'm going to propose PR which reverts the mentioned part for now.

Maffooch:
The "system level trump" really only applies to a two larger actions:
- Requests for review
- @ mention on a note [code](https://github.com/DefectDojo/django-DefectDojo/blob/82181253cd303efd1642994327e1d862c6b247ef/dojo/utils.py#L1375C1-L1382C36)
And then some niche actions that I did not realize were apart of this change:
- Engagement coming up or going stale [code](https://github.com/DefectDojo/django-DefectDojo/blob/82181253cd303efd1642994327e1d862c6b247ef/dojo/tasks.py#L32C1-L49C1)
- Engagement lead is notified when engagement is deleted. This is different from any engagements being deleted [code](https://github.com/DefectDojo/django-DefectDojo/blob/82181253cd303efd1642994327e1d862c6b247ef/dojo/engagement/views.py#L310C1-L316C65)
  - This applied to tests as well [code](https://github.com/DefectDojo/django-DefectDojo/blob/82181253cd303efd1642994327e1d862c6b247ef/dojo/test/views.py#L301C1-L307C65)

 My reasoning for the override here is that some folks could elect to not receive a notification for being mentioned in a comment or when they are requested for a review. That behavior lead to users never being notified of time sensitive actions. Those two actions should not be toggle able at any level as the default behavior should be notify folks if they are explicitly mentioned/requested.

mtesauro (approved):
Approved

mtesauro:
@kiblik #9699 has been merged,  Happy to merge this one once it's tests are green & has the approvals

Maffooch (approved):

grendel513 (approved):

blakeaowens (approved):

dojo/engagement/signals.py
```diff
@@ -0,0 +1,57 @@
...
+@receiver(pre_save, sender=Engagement)
+def engagement_pre_save(sender, instance, **kwargs):
+    old = sender.objects.filter(pk=instance.pk).first()
+    if old and instance.status != old.status:
+        if instance.status in ["Cancelled", "Completed"]:
+            create_notification(event='engagement_closed',
+                                title=_('Closure of %s') % instance.name,
+                                description=_('The engagement "%s" was closed') % (instance.name),
+                                engagement=instance, url=reverse('engagement_all_findings', args=(instance.id, )))
+        elif instance.status in ["In Progress"]:
```
cneill, line 30: If an Engagement is "Not Started" and I add a Test, this results in a notification that the Engagement was "reopened". I guess the check here should discard these notifications if `old`'s status was "Not Started"

No Tests under the `asdfadsf` Engagement, status of "Not Started":
Screenshot 2024-04-09 at 19 56 38

Test added, notification that `asdfadsf` was "reopened":
Screenshot 2024-04-09 at 19 57 05

View of Engagement after Test was added:
Screenshot 2024-04-09 at 19 57 12

dojo/endpoint/signals.py
```diff
@@ -0,0 +1,30 @@
...
+def endpoint_post_delete(sender, instance, using, origin, **kwargs):
+    if instance == origin:
+        if settings.ENABLE_AUDITLOG:
+            le = LogEntry.objects.get(
+                    action=LogEntry.Action.DELETE,
+                    content_type=ContentType.objects.get(app_label='dojo', model='endpoint'),
+                    object_id=instance.id
+            )
+            description = _('The endpoint "%(name)s" was deleted by %(user)s') % {
+                                'name': str(instance), 'user': le.actor}
```
cneill, line 23: I'm not sure how `le.actor` is populated here, but it says the endpoint was deleted by "None" whether I delete it via the API or the UI when acting as the `admin` user. I created a new user `test` and tried it again. This time I still get `None` when deleting via the API, but no notification at all when deleting from the UI. I'm not sure what's happening there...

Screenshot 2024-04-09 at 20 21 40

dojo/finding_group/signals.py
```diff
@@ -0,0 +1,31 @@
...
+def finding_group_post_delete(sender, instance, using, origin, **kwargs):
+    if instance == origin:
+        if settings.ENABLE_AUDITLOG:
+            le = LogEntry.objects.get(
+                    action=LogEntry.Action.DELETE,
+                    content_type=ContentType.objects.get(app_label='dojo', model='finding_group'),
+                    object_id=instance.id
+            )
+            description = _('The finding group "%(name)s" was deleted by %(user)s') % {
+                                'name': instance.name, 'user': le.actor}
```
cneill, line 23: In this case, `le.actor` is properly populated with the username when a Finding Group is deleted in the UI.

Screenshot 2024-04-09 at 20 27 29

dojo/product/signals.py
```diff
@@ -1,77 +1,38 @@
...
+                            url=reverse('view_product', args=(instance.id,)))
+
+
+@receiver(post_delete, sender=Product)
+def product_post_delete(sender, instance, **kwargs):
+    if settings.ENABLE_AUDITLOG:
+        le = LogEntry.objects.get(
+                action=LogEntry.Action.DELETE,
+                content_type=ContentType.objects.get(app_label='dojo', model='product'),
+                object_id=instance.id
```
cneill, line 28: Username is properly populated here when deleting via the UI (`TestProduct`), but not the API (`test1`)...

Screenshot 2024-04-09 at 20 31 54

dojo/product_type/signals.py
```diff
@@ -0,0 +1,39 @@
...
+def product_type_post_save(sender, instance, created, **kwargs):
+    if created:
+        create_notification(event='product_type_added',
+                            title=instance.name,
+                            product_type=instance,
+                            url=reverse('view_product_type', args=(instance.id,)))
+
+
+@receiver(post_delete, sender=Product_Type)
+def product_type_post_delete(sender, instance, **kwargs):
```
cneill, line 23: No notification is generated if a Product Type is deleted via either UI or API. You can see here that I've created the same `test1` Product Type multiple times without any notifications for deletions

Screenshot 2024-04-09 at 20 37 30

dojo/test/signals.py
```diff
@@ -1,13 +1,39 @@
...
+from django.utils.translation import gettext as _
+from django.urls import reverse
+
 from dojo.models import Test, Finding
+from dojo.notifications.helper import create_notification
+
 
-logger = logging.getLogger(__name__)
+@receiver(post_delete, sender=Test)
+def test_post_delete(sender, instance, using, origin, **kwargs):
```
cneill, line 15: No notifications are generated when deleting a Test by either UI or API (apologies for the confusing names used here - Product name=`test1`, Engagement name=`test`, Test name=`test1` 😅 )

Screenshot 2024-04-09 at 20 45 00

cneill (commented):
I found a number of unexpected results when testing this locally, primarily in the form of 1) missing notifications that I would have expected to see, and 2) notifications missing the actor in some cases.

kiblik:
> I found a number of unexpected results when testing this locally, primarily in the form of 1) missing notifications that I would have expected to see, and 2) notifications missing the actor in some cases.

Thank you for this feedback. I will check it and try to cover it better in unittests

dojo/engagement/signals.py
kiblik: Make sense.
Implemented and covered in unittests as well: https://github.com/DefectDojo/django-DefectDojo/pull/8824/commits/575bb4e99d5714f759c02cbe6d9918388e312b7f

kiblik:
> I found a number of unexpected results when testing this locally, primarily in the form of 1) missing notifications that I would have expected to see, and 2) notifications missing the actor in some cases.

I adjusted tests for better checking of username https://github.com/DefectDojo/django-DefectDojo/pull/8824/commits/184a691cb4bbc8bb08aedb75426fdb016bd86346 and tests pass.
I think that it might be some deeper issue connected to "multithreading", "multiprocessing" or "multiple pods".
I thought that signals are processed with the exact same thread but I might be wrong.
A deeper analysis is needed here.

cneill:
> > I found a number of unexpected results when testing this locally, primarily in the form of 1) missing notifications that I would have expected to see, and 2) notifications missing the actor in some cases.
> 
> I adjusted tests for better checking of username [184a691](https://github.com/DefectDojo/django-DefectDojo/commit/184a691cb4bbc8bb08aedb75426fdb016bd86346) and tests pass. I think that it might be some deeper issue connected to "multithreading", "multiprocessing" or "multiple pods". I thought that signals are processed with the exact same thread but I might be wrong. A deeper analysis is needed here.

Thanks for looking into this. I couldn't tell why I was seeing such different results during my testing for what should be similar behaviors, so further investigation seems worthwhile here.

kiblik:
> > > I found a number of unexpected results when testing this locally, primarily in the form of 1) missing notifications that I would have expected to see, and 2) notifications missing the actor in some cases.
> > 
> > 
> > I adjusted tests for better checking of username [184a691](https://github.com/DefectDojo/django-DefectDojo/commit/184a691cb4bbc8bb08aedb75426fdb016bd86346) and tests pass. I think that it might be some deeper issue connected to "multithreading", "multiprocessing" or "multiple pods". I thought that signals are processed with the exact same thread but I might be wrong. A deeper analysis is needed here.
> 
> Thanks for looking into this. I couldn't tell why I was seeing such different results during my testing for what should be similar behaviors, so further investigation seems worthwhile here.

Hi @cneill,
I was making different changes but I wasn't able to replicate reported issues from the UI.
However, I was able to replicate the issue from the API. There was a problem with the method of how AuthN with Token is performed. I found the open issue in https://github.com/jazzband/django-auditlog/issues/115#issuecomment-1539262735 with an offered solution. I implemented it and it worked. I wrote also unittest for this use case. https://github.com/DefectDojo/django-DefectDojo/pull/8824/commits/b8e1be569e9ef119eb8666d236967857c61c1453
From the way, how it is written it might solve also some other cases (not only access via API with Token).
I would like to ask you to test it now.
If it passes, I hope, this PR is ready for merge.

cneill (commented):
Most of the issues I found last time have been fixed! Thanks for digging into this. I think I found the source of the issue with endpoint deletions not triggering alerts: bulk delete (see video). I'm not sure whether that deserves its own PR or if it would be in scope for this one, I'll leave that up to you.

kiblik:
> Most of the issues I found last time have been fixed! Thanks for digging into this. I think I found the source of the issue with endpoint deletions not triggering alerts: bulk delete (see video). I'm not sure whether that deserves its own PR or if it would be in scope for this one, I'll leave that up to you.

Thank you for your feedback and for testing again.
- **Bulk removal of endpoint**: It hasn't been implemented before this PR. Only single removal was implemented. So we would not be speaking only about fixing a bug but implementing the new feature. I would prefer to implement it in the new PR because smaller changes are easier to review and this PR is focused only on the migration of the notification processor.
  - If you are interested in the reason why it is not working now: `post_delete` signals are sending `origin` to inform you who triggered deletion. In case of removal of one endpoint (`ep = Endpoint.objects.get(id=pk).delete()`) `origin` is the endpoint itself. In case of bulk removal (`eps = Endpoint.objects.filter(id__in=[pk1, pk2, pk3]).delete()`), `origin` is `QuerySet`. This will need some special handling. By the way, `Findings` support bulk removal so they will need this handling as well.
- **Removal of test:** This has been implemented before however I found a bug in the implementation. So this kind of notification was broken before this PR.
  - `dojo/test/views.py:delete_test` sends in `create_notification` list of `recipients` as users (as Django models) https://github.com/DefectDojo/django-DefectDojo/blob/2234848f02f243b5250894fccb6fc05ccaa1bc0d/dojo/test/views.py#L342 but `dojo/notifications/helper.py:create_notification` expects usernames (as strings) https://github.com/DefectDojo/django-DefectDojo/blob/2234848f02f243b5250894fccb6fc05ccaa1bc0d/dojo/notifications/helper.py#L33 So same as Endpoints, I would prefer fix this bug in separated PR. Sidenote: I noticed the same bug in some other places as well. I will fix them there as well. Interesting that nobody complained :).

If everybody agrees, I hope this PR is ready for merge.

cneill (approved):
Thanks for the thorough explanation! I agree that it would be better to address the work required for those additional fixes in one or more follow-on PRs :+1:

PR #8839, opened 2023-10-17
Fix findings and endpoints URLs in notifications

C0wnuts:
**Description**

I've discovered that for notifications of findings creation, findings reopening and endpoint deletion, the url indicated contains 2 times the first part of the URL. 
Example of malformed URL : `https://defectdojo.mydomain.frhttps://defectdojo.mydomain.fr/finding/99`

The problem is apparently linked to the use of the "request.build_absolute_uri" function.

I went to investigate how the URL contained in the notifications that don't have this bug was constructed and modified the ones that do in the same way.

**Test results**

I turned on Slack notifications before change to make sure of the problem.

Made changes and replayed notifications containing a buggy URL. Problem fixed.

### DIFF
--- a/dojo/endpoint/views.py
+++ b/dojo/endpoint/views.py
@@ -228,7 +228,7 @@ def delete_endpoint(request, eid):
                                     title='Deletion of %s' % endpoint,
                                     product=product,
                                     description='The endpoint "%s" was deleted by %s' % (endpoint, request.user),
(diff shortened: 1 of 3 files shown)


blakeaowens (approved):

mtesauro (approved):
Approved

grendel513 (approved):

PR #8844, opened 2023-10-17
Update Qualys Parser to aggregate endpoints rather than only store 1 endpoint per Finding

FelixHernandez:
## Update Qualys Parser to aggregate endpoints rather than only store 1 endpoint per Finding
[sc-2801]
**Description**

This bugfix now aggregates the endpoints in a list and validate the duplicated findings


dwalleck (approved):
Tested. Looks good to me!

mtesauro (approved):
Approved

cneill (approved):

PR #8852, opened 2023-10-18
add new parser xray

renejal:
From our organization we want to contribute with a new xray parser for https://jfrog.com/help/r/jfrog-cli/on-demand-binary-scan

image


kiblik:
I'm happy to see this parser. We bumped into the same missing piece so we wanted to propose our version of parser. But your is already covering more use cases.

I will add some of our reports + unit tests.

But, I'm not 100% sure about the parser's name. I fully understand that it is coming from the official name of the method, how to obtain the mentioned report. Just not sure.

renejal:
> I'm happy to see this parser. We bumped into the same missing piece so we wanted to propose our version of parser. But your is already covering more use cases.
> 
> I will add some of our reports + unit tests.
> 
> But, I'm not 100% sure about the parser's name. I fully understand that it is coming from the official name of the method, how to obtain the mentioned report. Just not sure.

.dryrunsecurity.yaml
mtesauro, line 1: This file isn't part of the parser so please remove from this PR

docs/content/en/getting_started/upgrading.md
mtesauro, line 1: Same for this file. 

Once these extra files are removed, I'm ready to approve this PR.

mtesauro (commented):
Clean up 2 extra files in this PR and it will be good to go

PR #8864, opened 2023-10-19
Adjust how error handling is done in Excel and CSV downloads

FelixHernandez:
## Adjust how error handling is done in Excel and CSV downloads
[sc-2848]

**Description**

This bugfix keeps the information regardless the column raises an error or not in the creation of the excel and csv files to export.
The `logger.debug` was changed by `logger.error`


dwalleck (approved):

dwalleck:
I don't know testable these snippets are, but since there are changes in behavior, should tests to be amended to reflect the changes?

mtesauro:
@dwalleck Before this fix, if the catch of the try/catch was triggered, nothing was added for that cell of data so it would produce a CSV/Excel where every column after that was shifted on to the left starting at the cell with the error.

This fix puts something in that cell when there's issues with the data to:
* Keep the columns lining up correctly
* Drop a messing in that problem column letting the user know that something it wrong with the source data.

This is kinda an edge case where something like a strange unicode character makes it into the DB that breaks the export to CSV/Excel.

So, this is more make the export perform correctly when errors occur vs a change in behavior.

mtesauro:
Tested locally and worked.

mtesauro (approved):
Approved

cneill (approved):

grendel513 (approved):

PR #8879, opened 2023-10-23
Update deprecated escape sequence

FelixHernandez:
## Update deprecated escape sequence
[sc-2671]

**Description**

This change updates the code, removing the deprecated escape sequence before this produces a bug.
Also the filters now only send the parameters requested as the following image shows.

**Test results**
image


dojo/templates/dojo/filter_snippet.html
```diff
@@ -54,3 +54,32 @@
     </form>
 
 </div>
+<script>
```
cneill, line 57: This snippet appears to be unrelated to the escape sequence fixes?

dojo/user/validators.py
```diff
@@ -75,7 +75,7 @@ def get_help_text(self):
 
 class SymbolValidator(object):
     def validate(self, password, user=None):
-        contains_special_character = re.findall('[()[\]{}|\\`~!@#$%^&*_\-+=;:\'\",<>./?]', password)  # noqa W605
+        contains_special_character = re.findall(r'[(){}\[\]|~!@#$%^&*_\-+=;:\'",<>\./?]', password)
```
cneill, line 78: Looks like the backtick (``` ` ```) is missing here

cneill (changes requested):

dojo/user/validators.py
FelixHernandez: I added the backtick in the [commit](https://github.com/DefectDojo/django-DefectDojo/pull/8879/commits/1e51a6351795768c064e63a76d213517375b983f)

mtesauro (approved):
Approved

cneill (approved):

grendel513 (approved):

Maffooch (approved):

PR #8927, opened 2023-11-02
Add GitHub workflow to protect certain files from modifications

dwalleck:

**Description**

As part of preparation for v3, there is a need for code stability in certain areas. This PR includes a script and workflow that will report back to the PR if files marked as protected. For now, modifications to protected files will comment only and not fail the PR. At some point in the near future, it will switch to being a gating check.


mtesauro (approved):
Approved

mtesauro:
We're going to have to follow this was a much smaller initial list of files to protect - that's a rather broad list.

dwalleck:
> We're going to have to follow this was a much smaller initial list of files to protect - that's a rather broad list.

I agree. Unfortunately, this was the ask. We still have a little time to talk about paring down the list. I would rather start with too little than start with too much. On the flip side, this job won't be blocking commits, so I can see why some would want to move ahead. I'm very open to suggestions.

grendel513 (approved):

Maffooch (approved):

PR #8929, opened 2023-11-02
changed function based views to class based views

FelixHernandez:
## changed function based views to class based views
[sc-2909]
Changed function views Import and reimport to class based views


mtesauro (approved):
Approved

grendel513 (approved):

Maffooch (approved):
Can follow up with there other instances of import/reimport in future PRs

cneill (approved):

PR #8961, opened 2023-11-08
Reporting: Exclude empty fields from report + correct host for HTML reports

Maffooch:
- Removed some fields in HTML reports that does not have any contents to avoid extraneous sections for a given finding
- Removed the FQDN from HTML reports as it often exposed proxy ports and prevented images from being rendered in the report


grendel513 (approved):

mtesauro (approved):
Approved

cneill (approved):

blakeaowens (approved):

Issue #8966, opened 2023-11-09
Add field "endpointid" to Endpoint 


manuel-sommer:
At the moment, parser have the following fields available as the endpoint: 
https://github.com/DefectDojo/django-DefectDojo/blob/2a909ecc0f7f3d10fd0f7681ad880def4c775331/dojo/models.py#L1508C5-L1508C5

However, there are parsers which could also make use of a unique_id endpoint as e.g. the IP changes. This is for example occurs whenever Wazuh agent is rolled out on a client and the client IP changes each day. But, the Wazuh agent itself is always the same. The same issue occurs in the following discussion: https://github.com/DefectDojo/django-DefectDojo/pull/8914#discussion_r1384714114

It would be nice if we could add another field to Endpoints to support also unique agent IDs (endpoint IDs).

PR #8967, opened 2023-11-09
add endpointid_field to endpoint #8966

manuel-sommer:
The description of this PR is mentioned in this issue #8966


kiblik:
I fully understand the reasons for this change and I'm also not satisfied with the `Endpoint` implementation now (even though I helped with the rewrite of this implementation).
However, I'm not 100% sure if this is the way, how it should evolve.

In many DASTs (like ZAP), many `Endpoint`s are created. Because each URL (with different query parameters) is a separate `Endpoint`. Yes, UI supports grouping them by `host` (`IP`), so you are able to see findings per each host, not only per `Endpoint`.
But there are real-world cases (like MS Defender or Nexpose) where you are able to say that one asset has multiple IPs and they are changing over time.

Adding one parameter to `Endpoint` will maybe satisfy needs of one parser. But it will not solve the root cause. It needs deeper discussion and more brainstorming about redesign. Maybe v3 is coming with some changes in this?

mtesauro:
@kiblik That's definitely something we're looking at changing as part of v3 (aka it's a known issue) - one of the things we're looking at that, since products will have arbitrary children is making endpoints just a child of the product and having them be much more broadly defined for cases like you mention. The same thing happens with Qualys if you have the Qualys agent installed on systems - it provides a unique UUID for that system separate from it's IP or hostname.

PR #8971, opened 2023-11-10
[ENHANCEMENT] Add Finding Violates SLA filtering to Report Builder

tomaszn:
This merge request introduces a new feature to the Report Builder.

It enables users to filter findings based on their SLA violation status.

The main use case is to generate reports covering only SLA breaches.

sla

**Test results**

Use cases tested manually:

- no value chosen: displays all findings
- "Any" chosen: displays all findings and the field is reset
- "False" and "True": display only a proper subset of findings

**Documentation**

Update not needed for this minor improvement.


mtesauro (approved):
Approved

mtesauro:
@tomaszn Though you failed the Dryrun Security "Sensitive Files Check", your changes are small enough that they won't impact the work we're doing concurrently.

Maffooch (approved):

cneill (approved):

PR #9002, opened 2023-11-15
Fix incorrect implementation of auditlog managment

kiblik:
During troubleshooting of unittests in #8824, I found out that settings `auditlog` via `System_Settings` is not behaving correctly. 
It is not possible to enable/disable auditlog dynamically in one of the running containers because it will have only a "local" effect.

Enabling/disabling have to be set before start (e.g. by environmental variable).


dojo/views.py
```diff
@@ -13,9 +13,10 @@
 from dojo.models import Engagement, Test, Finding, Endpoint, Product, FileUpload
 from dojo.filters import LogEntryFilter
 from dojo.forms import ManageFileFormSet
-from dojo.utils import get_page_items, Product_Tab, get_system_setting
+from dojo.utils import get_page_items, Product_Tab
 from dojo.authorization.authorization import user_has_permission, user_has_permission_or_403, user_has_configuration_permission_or_403
 from dojo.authorization.roles_permissions import Permissions
+from dojo.settings.settings import ENABLE_AUDITLOG
```
Maffooch, line 19: It would be better to use the registered version in `django.conf.settings` rather than importing from the file directly

docs/content/en/getting_started/upgrading.md
```diff
@@ -72,6 +72,12 @@ godojo installations
 
 If you have installed DefectDojo on "iron" and wish to upgrade the installation, please see the [instructions in the repo](https://github.com/DefectDojo/godojo/blob/master/docs-and-scripts/upgrading.md).
 
+## Upgrading to DefectDojo Version 2.29.x.
+
+**Breaking Change**
+
+Parameter `enable_auditlog` is not possible to set through System settings anymore. If you set this parameter or you need to set it to `False` (to disable audit logging), set environmental variable `DD_ENABLE_AUDITLOG` to `False`.
```
Maffooch, line 79: Another env var should be added to the `docker-compose.yml` file in all the containers ran by the django image. This should do the trick
```
DD_ENABLE_AUDITLOG: ${DD_ENABLE_AUDITLOG:-False}
```

dojo/models.py
```diff
@@ -37,7 +37,7 @@
 from django.db.models import JSONField
 import hyperlink
 from cvss import CVSS3
-from dojo.settings.settings import SLA_BUSINESS_DAYS
+from dojo.settings.settings import SLA_BUSINESS_DAYS, ENABLE_AUDITLOG
```
Maffooch, line 40: I know this was here before you added to it, but I must ask anyways
>It would be better to use the registered version in django.conf.settings rather than importing from the file directly

docs/content/en/getting_started/upgrading.md
kiblik: @Maffooch, is this text which I should add to the mentioned file?

dojo/views.py
kiblik: Done

dojo/models.py
kiblik: Done

docs/content/en/getting_started/upgrading.md
Maffooch: Correct. Somewhere in the `environment` blocks for the `uwsgi, celerybeat, celeryworker, and init` containers
https://github.com/DefectDojo/django-DefectDojo/blob/19c4e746a320c2f4ed6b724257cdb4f2a59ff7bd/docker-compose.yml#L50

kiblik: Done

Maffooch (approved):

mtesauro (approved):
Approved

devGregA (approved):

blakeaowens (approved):

PR #9003, opened 2023-11-15
Authenticated? Go home from login page

kiblik:
If a user is already logged in but because of an unknown reason ends up on the login page.
The login page for logged-in users is useless.

Possible reasons:
- Not the best bookmark
- Cached autocompletes


kiblik:
Ok, I suppose, this might be nice change but so many integration tests should be fixed. Maybe bit later.

Maffooch:
Good change! I am surprised by the number of failing tests...

Maffooch:

PR #9008, opened 2023-11-16
fix button color in risk acceptance bug

FelixHernandez:
## fix button color in risk acceptance bug
[sc-3141]

**Description**

This bugfix shows the Add selected Findings button in a visible color  in /engagement/id/risk_acceptance/id/edit url

**Test results**

now it shows like the following.
image

### DIFF
--- a/dojo/static/dojo/css/dojo.css
+++ b/dojo/static/dojo/css/dojo.css
@@ -52,14 +52,12 @@ a {
 }
 
 .panel-default .btn-primary {
-    background-color: transparent!important;
     border-style: none!important;
     color: #fff!important;
     border: 1px solid #fff!important;
 }
 
 .panel-default .btn-primary:hover {
-    background-color: transparent!important;
     color: #fff!important;
     border: 1px solid #fff!important;
 }
--- a/dojo/static/dojo/css/dojo.css
+++ b/dojo/static/dojo/css/dojo.css
@@ -51,13 +51,20 @@ a {
   color: #546474;
 }
 
+.btn-in-footer{
+    background-color: #546474!important;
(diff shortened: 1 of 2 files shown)


grendel513 (approved):

blakeaowens (changes requested):
Removing the `background-color: transparent!important` attribute from these classes will affect lots of other areas in DefectDojo.

It may be best to just modify the template to look something like this:
```
            <form method="post" class="form-horizontal">{% csrf_token %}
                <div class="show_add_findings_form">
                    {% include "dojo/form_fields.html" with form=add_findings_form %}
                </div>
                <div class="panel-body">
                    {% include "dojo/paging_snippet.html" with page=add_findings prefix="apage" %}
                </div>
                <div class="form-group">
                    <div class="col-sm-offset-2 col-sm-10">
                        <input class="btn" name="add_findings" type="submit" value="Add Selected Findings"/>
                    </div>
                </div>
            </form>
```

Which would produce a result like this (similar to the _Add Note_ button):
Screenshot 2023-11-16 at 9 43 17 PM

cneill (approved):

grendel513 (approved):

dojo/static/dojo/css/dojo.css
```diff
@@ -51,6 +51,11 @@ a {
   color: #546474;
 }
 
+.btn-in-footer{
```
blakeaowens, line 54: Since this is no longer used, please remove, then this looks good!

blakeaowens (approved):

mtesauro (approved):
Approved

Issue #7867, opened 2023-03-23
Deduplication not working


domcar:
**Problem description**
I added some ENV VARs to the dojo deployment to make use of deduplications, unfortunately is not appearing to be working

**Steps to reproduce**
I added the following env vars via chart:
```yaml
extraConfigs:
  DD_DEDUPLICATION_ALGORITHM_PER_PARSER: |
    {
    "gitleaks":"DEDUPE_ALGO_HASH_CODE",
    "trivy-image":"DEDUPE_ALGO_HASH_CODE",
    "kube-hunter":"DEDUPE_ALGO_HASH_CODE",
    "kubeaudit":"DEDUPE_ALGO_HASH_CODE",
    "semgrep":"DEDUPE_ALGO_HASH_CODE",
    "zap-api-scan":"DEDUPE_ALGO_HASH_CODE"
    }
  DD_HASHCODE_FIELDS_PER_SCANNER: |
    {
    "gitleaks": ["title", "severity"],
    "trivy-image": ["title", "severity"],
    "kube-hunter": ["title", "severity"],
    "kubeaudit": ["title", "severity"],
    "semgrep": ["title", "severity"],
    "zap-api-scan": ["title", "severity"]
    }
```
But still also the Endpoint is being used for duplicates.

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [X ] Kubernetes
- [ ] GoDojo

alexandrelimait:
did you fix it somehow? I have the same issue

domcar:
Nope, still nothing

Gby56:
Hey ! I'm pretty familiar with these features, I wrote the one for the fields. For the algorithm, definitely you're not using the right value, that's the variable name, you should use `unique_id_from_tool` or `legacy` or `hash_code` or `unique_id_from_tool_or_hash_code`

alexeykond:
Hi. Did anyone manage to override hashcode fields for gitleaks? 
HASHCODE_FIELDS_PER_SCANNER map doesn't contain 'Gitleaks Scan' item for some reasons and DD_HASHCODE_FIELDS_PER_SCANNER env variable cannot override the map.
settings.dist.py:
`if len(env('DD_HASHCODE_FIELDS_PER_SCANNER')) > 0:
    env_hashcode_fields_per_scanner = json.loads(env('DD_HASHCODE_FIELDS_PER_SCANNER'))
[...]

PR #9038, opened 2023-11-22
Improve DD_DEDUPLICATION_ALGORITHM_PER_PARSER DD_HASHCODE_FIELDS_PER_SCANNER

Gby56:
Addressing https://github.com/DefectDojo/django-DefectDojo/issues/7867

This is improving the customization via env variables for hashcode field configurations, and deduplication algorithms, to allow for unknown scanners (not hardcoded in `settings.dist.py`)


kiblik:
@Gby56, thanks for this PR.
If you are changing this, do you mind replacing `print` with `logger.info`?
It will require to add also this to the top of file:
```python
import logging
```

Gby56:
Good idea ! Yes I thought about it before, but couldn't see it used at all in the file so I went straight for print haha

kiblik:
> Good idea ! Yes I thought about it before, but couldn't see it used at all in the file so I went straight for print haha

It is used in the most cases in the rest of the project. I have no idea why it wasn't here.
Thank you.

cneill (approved):

Maffooch (approved):

mtesauro (approved):
Approved

grendel513 (approved):

PR #9046, opened 2023-11-23
Improve Slack integration to allow notifying private, public, and group channels  (fixes #8569)

### LINKED ISSUE #8569: Support any kind of Slack channel for sending user notifications
This is a question regarding the implementation of Slack notifications, and possible a bug or a feature request.

So, Slack API accepts both IDs and names:
https://api.slack.com/methods/chat.postMessage#channels

Is it possible that in this case we may no longer need `slack_username` and `slack_user_id` columns in the database, and having just `slack_channel` would be enough?

I'm successfully using a patch that simply takes whatever Slack channel the user entered in their profile. No lookups or conditions are needed anymore [here](https://github.com/DefectDojo/django-DefectDojo/blob/967bf1b5018a1ad824e4e18371027b3e1fc3a7a3/dojo/notifications/helper.py#L220).

It works with their name, and private / public / group channels (having names starting with P / C / G, respectively).

tomaszn:
This change enables users to set private, public, or group channels as the target for Slack notifications.

It achieves this by removing the `slack_user_id` intermediary field, which was filled with a response after querying the Slack API with the user's email, and which limited the notification targets to users' private Slack channels. Before removal, the value is copied into `slack_username` field, overwriting the email address, which is no longer needed (and no longer works).

Detailed explanations are in #8569.

**Test results**

* Unit tests pass.
* SLA breach notifications are delivered successfully when `slack_username` is set to a channel names, with and without a "#" prefix.
* Also success when a channel ID is used.

**Documentation**

No changes in the documentation are needed.
However, I noticed that the Slack administration page looks now differently, so I updated the relevant screen shot in the "Notifications" section.

**Checklist**

- [x] Unit tests pass.


dojo/db_migrations/0194_remove_usercontactinfo_slack_user_id.py
```diff
@@ -0,0 +1,16 @@
+# Generated by Django 4.1.11 on 2023-11-23 09:57
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
manuel-sommer, line 6: maybe you could merge the two db migrations 0193 and 0194 together?

tomaszn: Thanks. These migrations are in relevant separate commits.
Do you want me to merge all commits in this MR into one?

manuel-sommer: Yeah, I guess it makes sense to merge the two db migrations.

devGregA:
Hi @tomaszn, we reviewed this PR but weren't able to replicate the problems mentioned. Could you please provide additional details on the issues you experienced?

tomaszn:
@devGregA, sure. The simplest use case is setting up notifications for `#general` Slack channel. It is currently not possible, as far as I know.

dojo/db_migrations/0193_alter_usercontactinfo_slack_username.py
```diff
@@ -0,0 +1,41 @@
...
+                verbose_name="Slack channel",
+            ),
+        ),
+        migrations.RunPython(code=replace_slack_email_with_channel_id,
+                             reverse_code=unreplace_slack_email_with_channel_id),
+        migrations.RemoveField(
+            model_name="usercontactinfo",
+            name="slack_user_id",
+        ),
+    ]
```
kiblik, line 41: What about
```
 migrations.RemoveField(
            model_name="usercontactinfo",
            name="slack_username",
        ),
 migrations.RenameField(
            model_name="usercontactinfo",
            old_name="slack_user_id",
            new_name="slack_username",
        ),
```
Looks easier compare to `RunPython`
https://docs.djangoproject.com/en/5.0/ref/migration-operations/#renamefield

tomaszn: Indeed, thanks. And then `AlterField` needs to be moved to the end, to set up `verbose_name` etc.

Moreover, I should set `max_length=81` ([`#` plus 80 characters of the channel name](https://api.slack.com/methods/conversations.rename#naming)).

paulOsinski:
Hi @tomaszn , I tested creating and linking the Slack integration and I was able to post two different kinds of notifications:

- personal notifications to an individual user's Slackbot (which is what the Slack Email Address field is used for in the DefectDojo UI.  Not sure what this maps to on the backend).
- system-wide notifications to a shared channel

I was able to define the Slack Channel for posts on the Slack-app side, when I created and installed it to my workplace.

I updated the documentation with what I found: https://github.com/DefectDojo/django-DefectDojo/pull/9420/commits/750e548e5e3f6393e755e3ae8bc3cd56ece9f972

----

Currently, Dojo does not support any customized or filtered notifications, or posts to multiple slack channels.  Is that what you're intending to do by changing the notification targets from Slack Usernames to Slack Channels?

tomaszn:
Hello @paulOsinski, thank you for your input here, and for the update to the documentation.

> Currently, Dojo does not support any customized or filtered notifications, or posts to multiple slack channels. Is that what you're intending to do by changing the notification targets from Slack Usernames to Slack Channels?

Correct. And by creating "virtual" users and assigning them Slack channels I can customize notifications. Several channels can get notifications about events in independently configured sets of products.

> If you wanted to add that feature, I think it would be more appropriate to add a Slack Channel mapping field on a Product, rather than on each user's profile which is where the Slack email address currently exists.

In my case, I need several Slack channels to be informed about various events in the products. So a simple 1:1 mapping would not be enough.

I acknowledge that the current solution is not perfect and serves more as a workaround. A fully-featured notification manager capable of handling mappings and preferences is something we should aim for in the future.

Please bear in mind that this changeset was prepared in a feature freeze situation. The goal had to be obtained with a minimal changes not to disrupt work of others :)

mtesauro:
Close as stale

PR #9047, opened 2023-11-23
Trufflehog deduplication fix

Demaz93:
**Description**

With the old configuration findings will be aggregated instead of split one finding for each report entry.
With the legacy hash algorithm every time you reimport the report, each finding was closed and reopened each time creating a lot of comments.


dojo/tools/trufflehog/parser.py
```diff
@@ -166,7 +168,7 @@ def get_findings_v3(self, data, test):
                     severity = "Medium"
 
             dupe_key = hashlib.md5(
```
manuel-sommer, line 170: Maybe it is better to use a DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL with unique_id_from_tool here?

grendel513 (approved):

blakeaowens (approved):

Maffooch (approved):

mtesauro (approved):
Approved

Issue #8797, opened 2023-10-05
Active, accepted findings getting mitigated when reimporting


gietschess:
**Be informative**
This issue relates to mr #7447.
Risk accepted Findings are set to mitigated even if they still exists in the imported scan.

**Bug description**
We're mostly working with trivy findings in defectdojo.
When we're accepting a finding with the simple risk acceptance function the finding gets the status 'Inactive', 'Risk Accepted'.
After the next reimport it also gets the status 'Mitigated', even if the active finding still exists in the report.

**Steps to reproduce**
1. Import findings of type 'Trivy Scan' via the reimport api
2. Go to http://localhost:8080/test/some-test-id, click on the menu points in front of one finding and hit 'Accept Risk' ('Enable simple risk acceptance' has to be enabled in the products settings to do this)
4. Reimport the same findings of type 'Trivy Scan' again via the reimport api
5. See that the previously accepted Finding got the additional status 'Mitigated'

**Expected behavior**
The finding should be skipped by the import (which it correctly does).
The finding should get **not** the status 'Mitigated' later in the reimport process when its still active.

In the Trivy Scan parser there are no informations for false positive, out of scope or risk accepted so these fields are set to False for each finding by default. By that the check mentioned in the [merge request](https://github.com/DefectDojo/django-DefectDojo/pull/7447) will never be True and the finding won't be added to the list of unchanged findings.
Without that the finding will be added to the to_mitigate list later.
See:
[dojo/importers/reimporter/reimporter.py#L431-L433](https://github.com/DefectDojo/django-DefectDojo/blob/bef8f1483ca8be21d658073dcfb4070ed2f173a0/dojo/importers/reimporter/reimporter.py#L431-L433)
[dojo/importers/reimporter/reimporter.py#L507](https://github.com/DefectDojo/django-DefectDojo/blob/bef8f1483ca8be21d658073dcfb4070ed2f173a0/dojo/importers/reimporter/reimporter.py#L507)

**Deployment method**
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version: 2.27.0
 - Trivy version: 0.45.1

**Logs** 
[uwsgi.log](https://github.com/DefectDojo/django-DefectDojo/files/12819937/uwsgi.log)
```
[...]

PR #6893, opened 2022-09-23
Add do_not_reactivate option to persist triages from triage-less scanners

Gby56:
Explained in https://github.com/DefectDojo/django-DefectDojo/issues/6892

PR #9050, opened 2023-11-24
fixes reimporter to ensure that risk accepted findings do not get mitigated

lme-nca:
This tries to improve on the (incomplete) fix introduced here: https://github.com/DefectDojo/django-DefectDojo/pull/7447 
Also see the discussion on Slack: https://owasp.slack.com/archives/C014H3ZV9U6/p1700819414409409 
and related issue https://github.com/DefectDojo/django-DefectDojo/issues/8797 

**Previous behaviour:**

IF existing findings (false positive, out of scope or risk accepted):
	- IGNORE ALL OTHER values of findings from scanner (even if its mitigated) so mitigated findings from scanner that are not mitigated in Defectdojo (but do have one of the above flags would never have been closed)
	
- IF existingFinding.false_p == new_finding.false_p and existingFinding.out_of_scope == new_finding.out_of_scope
                            and existingFinding.risk_accepted == new_finding.risk_accepted)
       - Then: `unchanged_items.append(finding)`
       -  comes from(https://github.com/DefectDojo/django-DefectDojo/pull/7447)

- ELSE Mitigate the finding **incorrectly**  <--- THE BUG described in this ISSUE:
 -- caused by the fact the finding is **not** added to "unchanged_items.append(finding)"
 -- Therefore the following code adds it to the mitigation list
`        to_mitigate = (
            set(original_items) - set(reactivated_items) - set(unchanged_items)
        )`

elIF existing finding IS mitigated
   ..... // not changed

else (existing finding is not mitigated)
	- IF new_finding: false positive, out of scope or risk accepted
		-- Close existing findings and take over values of scanner

**New behaviour**

 First IF condition is completely removed, no special treatment needed for this case --> if existing findings (false positive, out of scope or risk accepted): 

IF existing finding IS mitigated
   ..... // not changed

else (existing finding is not mitigated)
 - simplified logic (removed unnecessary If statement)
	--  `if not (finding.mitigated and finding.is_mitigated):` this always resolves to TRUE in this else branch!!!
        -- this changed the indenting here
 
 - IF existingFinding.false_p == new_finding.false_p and existingFinding.out_of_scope == new_finding.out_of_scope
                            and existingFinding.risk_accepted == new_finding.risk_accepted)
       - Then dont touch the existing finding, this keeps  (https://github.com/DefectDojo/django-DefectDojo/pull/7447) alive, however question is if we want this like that ?
 - IF new_finding: false positive, out of scope or risk accepted
		-- Close existing findings and take over values of scanner 
 - ELSE
 	Do not touch the finding 

Overall this is some of the most complicated and unreadable code I have seen so-far in Defectdojo. I would gladly refactor this complete method ( it is way to long and the variable names are terrible), however I dont know if this would fit with the whole Defectdojo 3.0 work ? @mtesauro @Maffooch Let me know if you are open for this and I would gladly work on making this code more digestible. 

@coheigea @Gby56 you also both worked on this code, I am open to any suggesting or improvements ?


Gby56:
I'm reading through the code changes, so far it makes sense to me, it removes a lot of unnecessary conditions I think

It made me think about the decision matrix overall, given the existing finding and the new uploaded item, there is probably a table looking like that that we should define no ?
image

lme-nca:
So one of the failing tests "test_import_reimport_keep_false_positive_and_out_of_scope" is interesting. Here it becomes obvious there was some kind of "special" treatment going on for **existing_findings** for which
`finding.false_p or finding.out_of_scope or finding.risk_accepted`

This kind of worked in the cases of `finding.false_p or finding.out_of_scope` as these status also set "is_mitigated" to true, therefore they never ran into the problem described in the issue https://github.com/DefectDojo/django-DefectDojo/issues/8797 
Risk_accepted keeps "is_mitigated" = False And therefore got set to is_mitigated=True incorrectly. (again this test did not find it because it manually sets is_mitigated=True with risk acceptance)

However, I always assumed it worked like this:

if "do_not_reactivate" == TRUE --> Then Defectdojo is source of truth and Scanner cannot reopen findings. (regardless of false positive, out of scope, risk acceptance)
if "do_not_reactivate" == TRUE --> Then Scanner is source of truth and Scanner cannot reopen findings. (regardless of false positive, out of scope, risk acceptance)

"do_not_reactivate" was introduced here: https://github.com/DefectDojo/django-DefectDojo/pull/6893 but as I read was mostly intended to work-around "triage-less scanners"

@Gby56 I agree that if we could define a complete decision table that this would make it easier.

lme-nca:
The problem here is 
`                    # If the scanner says the findings is either (risk-accepted, false positive or out of scope)
                    # we take over these values and close the finding
                    elif item.risk_accepted or item.false_p or item.out_of_scope:
                        logger.debug('Reimported mitigated item matches a finding that is currently open, closing.')
                        logger.debug('%i: closing: %i:%s:%s:%s', i, finding.id, finding, finding.component_name, finding.component_version)
                        finding.risk_accepted = item.risk_accepted
                        finding.false_p = item.false_p
                        finding.out_of_scope = item.out_of_scope
                        finding.active = False
                        if verified is not None:
                            finding.verified = verified`

Currently the default behaviour is to just "mitigate" a finding if the scanner sets risk acceptance (or false_p or out of scope). It even sets risk_accepted regardless if "simple risk acceptance" is active it seems. And it does not set mitigated=true (even though Defectdojo always does this in case of " false_p or out of scope"

Gby56:
I think a decision tree might be better, I realized that a matrix might be a bit overwhelming to navigate it

lme-nca:
dont have a nice way to create this here in this text, so just a screenshot. This is roughly how the logic works not in: https://github.com/DefectDojo/django-DefectDojo/pull/9050/commits/d3d918b7375c43322d2971e36ff6944b5c8d1d35 
image

coheigea (approved):
It looks OK to me overall, the logic looks a lot cleaner, before the class was a mess. I don't have the scope to test it right now, but once the patch is applied and released I can test it more thorougjly

lme-nca:
Reminder to: @Gby56 @mtesauro @blakeaowens  this pull request is still here, we have been running it on our fork for the last 3 weeks and it seems stable.

dojo/importers/reimporter/reimporter.py
```diff
@@ -105,258 +114,244 @@ def process_parsed_findings(
...
-                        and finding.out_of_scope == item.out_of_scope
-                        and finding.risk_accepted == item.risk_accepted
-                    ):
-                        unchanged_items.append(finding)
-                        unchanged_count += 1
-                        continue
-                elif finding.is_mitigated:
+            if existing_findings:
+                # existing findings found
+                existing_finding = existing_findings[0]
```
Gby56, line 130: Just wondering if we should log something here for a warning... or if a loop is needed to process all the matched existing findings instead of just the first one ?

Gby56 (approved):
Overall this looks good to me, renaming "items" also helped a whole lot for readability! Thank you for taking the time and testing it out

dojo/importers/reimporter/reimporter.py
lme-nca: Yes I was wondering something similar... For now we could write a debug/log statement. The purpose of this pull request was not to significantly change the behaviour of the reimporter (which this might do). Although I agree that probably this case should be handled somehow, I think this should be handled in a seperate ticket/issue.

lme-nca:
Just giving this another bump....  @mtesauro @blakeaowens @Maffooch

lme-nca:
Hey everyone, this is still here.... @mtesauro @blakeaowens @Maffooch Willing to fix the conflict if there is still interest to merge this ?

mtesauro:
@lme-nca 
There's some work happening around this part of DefectDojo e.g. #10011 which may address this issue. That work is of a broader scope than this PR. Let that one get merged and let's see where things are at after.

gietschess:
> @lme-nca There's some work happening around this part of DefectDojo e.g. #10011 which may address this issue. That work is of a broader scope than this PR. Let that one get merged and let's see where things are at after.

hey there,
i've tested the behaviour of issue #8797 again with defectdojo version 2.34.2 and risk accepted findings are still mitigated after reimporting them even if they're still active in the uploaded trivy json file

lme-nca:
@mtesauro @Gby56 @coheigea @Maffooch : I have rebased this change on the latest state and introduced a similar fix. If an existing finding is risk accepted then we avoid it being mitigated and closed if its still in an existing scan.

A quick review would be much appreciated as currently our fork is stuck (As we have operated this original change there for the last 6months) and rebasing is not possible until we have the fix also in upstream.

mtesauro:
@lme-nca  Sorry that this one has gotten lost in the shuffle for so long.

Would you mind rebasing this so we can review and approve it?

lme-nca:
> @lme-nca Sorry that this one has gotten lost in the shuffle for so long.
> 
> Would you mind rebasing this so we can review and approve it?

Hi @mtesauro I rebased it. Another pull request of mine that has been stuck for a long time is: https://github.com/DefectDojo/django-DefectDojo/pull/9787 would appreciate a nudge there as well ?

Maffooch (approved):

cneill (approved):

blakeaowens (approved):

Issue #8450, opened 2023-08-01
view Vulnerable Hosts / view Vulnerable Endpoints empty despite existing vulnerable endpoints and hosts


furrnace:
**Bug description**

view Vulnerable Hosts / view Vulnerable Endpoints shows an empty list although view Hosts and view Endpoints do have hosts/endpoints in the list that are marked as vulnerable

**Steps to reproduce**
1. Go to view Endpoints, browse to see if there are vulnerable Endpoints (analoguously for Hosts/vulnerable hosts)
2. Go to view Vulnerable Endpoints, see that the list is empty (analoguously for Vulnerable Hosts)

**Expected behavior**
Vulnerable Endpoints should contain the endpoints marked as vulnerable in view Endpoints
Vulnerable Hosts should contain the hosts marked as vulnerable in view Hosts

**Deployment method** *(select with an `X`)*
- [ X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Debian 11
 - DefectDojo version (see footer) or commit message: 2.24.3

WojTecH94:
Similar situation in version v. 2.28.1
My suspection is that endpoints are cleared from vulnerabilities once mitigated, but then when the same vulnerability is reported again endpoint status is not updated and it stays mitigated.

manuel-sommer:
Can we close this issue @mtesauro ?

mtesauro:
Yes,

PR #9077, opened 2023-11-29
Fix for vulns not included in host/endpoint views after reopening

WojTecH94:
Fix for vulnerabilities not included in host/endpoint views after reopening. More info: https://github.com/DefectDojo/django-DefectDojo/issues/8450

After commenting the lines on test environment, vulnerabilities that were reopened are properly shown on endpoint and host view. I was trying to find out how "status_finding__mitigated" filter works but couldn't find it. In my opinion it is not necessary to check it when "mitigated__isnull" filter is also verified.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1716,7 +1716,6 @@ class Endpoint(models.Model):
             mitigated__isnull=True,
             false_p=False,
             duplicate=False,
-            status_finding__mitigated=False,
             status_finding__false_positive=False,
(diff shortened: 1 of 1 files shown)


Maffooch (approved):
Makes sense because `mitigated` is DateTime field on the finding model

mtesauro (approved):
Approved

devGregA (approved):

blakeaowens (approved):

PR #9079, opened 2023-11-29
[ENHANCEMENT] Allow filtering Products and Engagements on unset properties

tomaszn:
This merge request introduces an "Unknown" option to the existing filters in Product and Engagement lists for:
* Product lifecycle,
* Business Criticality,
* Platform, 
* Origin.

This is how it looks like:

Screenshot_2023-11-29_18-16-16

It follows the scheme already used in other filters, like the boolean ones:

Screenshot_2023-11-29_17-48-40

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -725,7 +725,7 @@ class EngagementDirectFilter(DojoFilter):
         queryset=Product_Type.objects.none(),
         label="Product Type")
     test__engagement__product__lifecycle = MultipleChoiceFilter(
-        choices=Product.LIFECYCLE_CHOICES, label='Product lifecycle')
+        choices=Product.LIFECYCLE_CHOICES, label='Product lifecycle', null_label='Unknown')
     status = MultipleChoiceFilter(choices=ENGAGEMENT_STATUS_CHOICES,
                                               label="Status")
 
@@ -791,7 +791,7 @@ class EngagementFilter(DojoFilter):
         queryset=Product_Type.objects.none(),
         label="Product Type")
     engagement__product__lifecycle = MultipleChoiceFilter(
-        choices=Product.LIFECYCLE_CHOICES, label='Product lifecycle')
+        choices=Product.LIFECYCLE_CHOICES, label='Product lifecycle', null_label='Unknown')
     engagement__status = MultipleChoiceFilter(choices=ENGAGEMENT_STATUS_CHOICES,
                                               label="Status")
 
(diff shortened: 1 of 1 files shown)


mtesauro:
@tomaszn  I was a bit confused by this one initially.

After poking around an instance of DefectDojo, I now get what you're doing with this PR but I wonder if "Unknown" is the best label/value to use in the dropdown.

When I first read it I thought by "Unknown" you meant "I don't care what business critically is" e.g I don't know what critically it is which would basically be any business criticality.

I'm thinking that "Not set" or "Unset" would be better than "Unknown"

tomaszn:
@mtesauro, thanks. I agree that this is confusing. On the other hand, "Not set" and similar can suggest that the filter is not set.

My better ideas for this value include "Unspecified", "Undeclared", "Empty", "(null)".

cneill:
I agree that "Unknown" is confusing, but as @tomaszn points out, boolean filters like the ones for "Active" or "Verified" already display this today for empty filter values. As far as I can tell, that label appears to come from the [django-filter `BooleanWidget`](https://github.com/carltongibson/django-filter/blob/00e88652d9308c4c6c4e707c3e0d735dd6843dca/django_filters/widgets.py#L162).

I'd prefer something like "Any" to "Unknown", and indeed we use that in some places already.

Screenshot 2023-12-05 at 14 18 17

tomaszn:
@cneill, Criticality field allows picking several values, which in case of picking all of them already means "Any".
What is impossible, is to pick objects with the "null" value (not set).

Booleans are an interesting case, as there two widgets used inconsistently, and both allow choosing only one value. None of them supports all the cases: <filter disabled>, Yes, No, Yes or No, <not set>.

image

cneill:
> @cneill, Criticality field allows picking several values, which in case of picking all of them already means "Any". What is impossible, is to pick objects with the "null" value (not set).
> 
> Booleans are an interesting case, as there two widgets used inconsistently, and both allow choosing only one value. None of them supports all the cases: /filter disabled/, Yes, No, Yes or No, /not set/.

Ah apologies, I misunderstood. I like "Empty" best among your suggested replacements.

tomaszn:
changed to Empty, thanks!

cneill (approved):

mtesauro (approved):
Approved

grendel513 (approved):

blakeaowens (approved):

PR #9118, opened 2023-12-05
API: replace local swagger (drf-yasg -> sidecar)

kiblik:
The first step to removal of drf-yasg

If somebody uses `DD_DEFAULT_SWAGGER_UI=False` (a.k.a. do not load swagger from CDN but from static resources), this will not point the user to drf-yasg statics but to drf_spectacular_sidecar.

Tested locally.

Inspired by https://drf-spectacular.readthedocs.io/en/latest/readme.html#self-contained-ui-installation

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -791,8 +791,8 @@ SPECTACULAR_SETTINGS = {
 }
 
 if not env('DD_DEFAULT_SWAGGER_UI'):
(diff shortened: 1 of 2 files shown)


cneill (approved):

mtesauro (approved):
Approved

grendel513 (approved):

Maffooch (approved):

PR #9119, opened 2023-12-06
Shellcheck workflow

cneill:
**Description**

This PR adds a GitHub Actions workflow to run [shellcheck](https://github.com/koalaman/shellcheck) on shell scripts modified as part of a PR.

[sc-2794]

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -33,7 +33,6 @@ jobs:
         shell: bash
         run: |
           set -o pipefail
-          set -x
 
           # Make sure we already put the proper shellcheck binary in place
           if [ ! -f "./shellcheck" ]; then
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -3,8 +3,9 @@ name: Shellcheck
 on:
   pull_request:
 env:
-  SHELLCHECK_VERSION: 'v0.9.0'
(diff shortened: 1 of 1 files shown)


.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,125 @@
+---
+name: Shellcheck
+on:
+  pull_request:
+env:
+  SHELLCHECK_REPO: 'koalaman/shellcheck'
+  SHELLCHECK_VERSION: 'v0.9.0'
+  SHELLCHECK_SHA: '038fd81de6b7e20cc651571362683853670cdc71'
```
kiblik, line 8: Are Renovate or Dependabot able to monitor new releases for this kind of definition?

mtesauro (approved):
Approved.

.github/workflows/shellcheck.yml
cneill: I suspect Renovate / Dependabot won't catch this, but thankfully shellcheck only seems to [release about once a year](https://github.com/koalaman/shellcheck/releases), and I don't think we'd miss out on much if we were one release behind.

kiblik:
Hi guys, I'm happy that this linter was merged quickly but existing issues have not been fixed.
It detects previous issues (and fails tests) in existing PRs which didn't even touch problematic lines.
E.g. https://github.com/DefectDojo/django-DefectDojo/pull/9002/files.

Issue #9192, opened 2023-12-19
Manage notifications for other users


tomaszn:
**Is your feature request related to a problem? Please describe**

In my setup, each project has a dedicated user that subscribes to Slack notifications for their team's projects. Thanks to #9046 they can subscribe to "normal", group Slack channels.

When users switch teams or leave, or new projects are created, this needs to be adjusted. I would like administrators to maintain and supervise all the notifications, and not bother the regular users.

**Describe the solution you'd like**
As an administrator, I want to enable Slack channel notifications for different projects, targeting different Slack channels, so the teams are notified about SLA breaches.

It would be good enough to employ Django Admin functionality to create, modify, and delete the notifications.

**Describe alternatives you've considered**
Currently, I am connecting to the database and modifying the records.

kiblik:
Hi @tomaszn,

using of API endpoints `GET /api/v2/notifications/?product=XX&user=YY` and `PATCH /api/v2/notifications/{id}/` is not enough for you?

tomaszn:
@kiblik Thanks, that it is a perfect option for automating the notification setup. 

However, for manual changes, it is even less convenient than editing the settings visually in a GUI database editor.

PR #9193, opened 2023-12-19
Django Admin pages for editing Notifications

tomaszn:
**Description**

This customizes the built-in (and enabled by default) Django Admin interface, and adds a visual way to add, remove, and modify notifications in DefectDojo.

This is a workaround for #9192.

This is how the notification list will look like:

Screenshot_2023-12-19_13-15-12

This is how the a single notification editor will look like:

Screenshot_2023-12-19_13-21-51

**Test results**

Tested manually.

**Documentation**

Django Admin functionality is not described in the documentation.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -3837,6 +3837,18 @@ class Notifications(models.Model):
                 result.risk_acceptance_expiration = merge_sets_safe(result.risk_acceptance_expiration, notifications.risk_acceptance_expiration)
 
         return result
+    
+    def __str__(self):
+        return f"Notifications about {self.product or 'all projects'} for {self.user}"
+
+
+class NotificationsAdmin(admin.ModelAdmin):
+    list_filter = ('user', 'product')
+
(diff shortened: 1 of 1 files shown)


dojo/models.py
```diff
@@ -3838,6 +3838,18 @@ def merge_notifications_list(cls, notifications_list):
 
         return result
 
+    def __str__(self):
+        return f"Notifications about {self.product or 'all projects'} for {self.user}"
```
kiblik, line 3842: `self.user` might be `None` (for system [non-user] notifications)

kiblik (approved):

dojo/models.py
tomaszn: Fixed by using "system notifications" phrase instead of the user string representation:

image

cneill (approved):

mtesauro (approved):
Approved

grendel513 (approved):

Maffooch (approved):

PR #9273, opened 2024-01-04
:sparkles: Add option for additional labels on initializer

### LINKED ISSUE #9267: Add option to set custom Labels to Initializer Job in Helm Chart

When injecting istio sidecars to the django-DefectDojo helm deployment it is currently not possible to skip injection for the initializer job, which as a result will make the initializer job unable to finish, as the istio-sidecar will not exit when the job does and therefore block any startup from the defectdojo services, as the initializer job is not finished.

In order to circumvent this in other Helm Deployments the sidecar injection can specifically be disabled on specific deployments, pods, etc. through the same label or annotation with a value of "false". This is not possible in the current Helm Chart of django-DefectDojo.

Add an option to set additional labels [here](https://github.com/DefectDojo/django-DefectDojo/blob/master/helm/defectdojo/templates/initializer-job.yaml#L7) similar to how it is handled for annotations [here](https://github.com/DefectDojo/django-DefectDojo/blob/master/helm/defectdojo/templates/initializer-job.yaml#L13-L16).

```yaml
  labels:
    defectdojo.org/component: initializer
    {{- with .Values.initializer.jobLabels }}
    {{- toYaml . | nindent 4 }}
    {{- end }}
```

In order to have a more universal approach the same could also be done for the deployments and not just for the initializer job.

There is multiple approaches to inject istio sidecars into the deployments: 

- Through a label on the Namespace (this will inject istio sidecars onto all Pods in the namespace and result in the above mentioned issue)
- Through annotations on Deployment/Job/Cronjob templates (this is deprecated by istio and will not work in the future)
- Through labels on Deployment/Job/Cronjob templates (this is not possible in the current Helm Chart of django-Defectdojo)

Annotations are currently the only option that will work with the current Helm Chart provided for django-Defectdojo, but as annotations are deprecated I would not consider this a viable option, as it is not future-proof. 

**Additional context**
[Istio Sidecar Injection](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/)
[Istio Resource Annotations with Deprecation Warning](https://istio.io/latest/docs/reference/config/annotations/)
[Istio Resource Labels](https://istio.io/latest/docs/reference/config/labels/)

Reading [this comment](https://github.com/DefectDojo/django-DefectDojo/issues/7235#issuecomment-1721531198) I am aware that changes to the Helm Chart rely on the community and I would be happy to draft a PR for this if it gets accepted.

prempador:
Closes #9267

**Description**

This PR adds the option to add additional labels to the initializer job. This can be done in the values file like:

```yaml
initializer:
    labels: {
        key: value
    }
```

Contrary to what I mentioned in the Issue, it is already possible to add labels to the Deployment Pods via `podLabels`, see [here](https://github.com/DefectDojo/django-DefectDojo/blob/master/helm/defectdojo/values.yaml#L65-L68). Which only made it necessary to add the option to the Job aswell. 

I saw 3 options to make this possible: 

1. Add `labels` directly to initializer, same as annotations and therefore scope the additional labels only to the initializer job (which is as of writing this the only job).
2. Add `jobLabels` as a global value, same as `podLabels`. This would be more future-proof in case of addition of more jobs. But to me this option would be a bit confusing in the end, as it is not really "jobLabels" and rather "jobPodLabels", as `spec.template.metadata.labels` sets the labels that apply to pods managed by the job (or deployment). But this would be more of a wording discussion.
3. Add `jobLabels` next to `jobAnnotations`, same as [here](https://github.com/DefectDojo/django-DefectDojo/blob/master/helm/defectdojo/values.yaml#L303-L305). But this would as option 2 be misleading in my opinion, as `jobAnnotations` are adding annotations to the job metadata and not to `spec.template.metadata.labels`.

In the initial form of this PR I chose number 1, but I am happy to adjust my code to option number 2 if there is strong opinions towards it.

### DIFF
--- a/helm/defectdojo/templates/initializer-job.yaml
+++ b/helm/defectdojo/templates/initializer-job.yaml
@@ -22,6 +22,9 @@ spec:
         defectdojo.org/component: initializer
         app.kubernetes.io/name: {{ include "defectdojo.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
+        {{- if .Values.initializer.labels }}
+        {{- toYaml .Values.initializer.labels | nindent 8 }}
+        {{- end }}
       annotations:
       {{- with .Values.initializer.annotations }}
         {{- toYaml . | nindent 8 }}
--- a/helm/defectdojo/values.yaml
+++ b/helm/defectdojo/values.yaml
@@ -304,6 +304,7 @@ initializer:
     helm.sh/hook: "post-install,post-upgrade"
   }
   annotations: {}
+  labels: {}
   keepSeconds: 60
   affinity: {}
   nodeSelector: {}


cneill (approved):

mtesauro (approved):
Approved

Maffooch (approved):

dsever (approved):

PR #9286, opened 2024-01-05
dependabot and renovate PRs against bugfix to release dependencies faster

manuel-sommer:
Faster release of dependency updates


mtesauro:
For passive library/component updates, the dev branch is the best choice since those updates soak a month to let any subtle bugs be discovered.

If there's a security critical issue for either of those tools, we can always cherrypick that PR into the bugfix branch.

PR #9287, opened 2024-01-05
Update finding templates to use test_type rather then found_by

FelixHernandez:
## Update finding templates to use test_type rather then found_by
[sc-2859]
This feature allows the user see the test_type information when the found_by is not available.

In the Findings list:
image

In the Finding view:
image

### DIFF
--- a/dojo/templates/dojo/findings_list_snippet.html
+++ b/dojo/templates/dojo/findings_list_snippet.html
@@ -616,10 +616,17 @@
                                             {% endif %}
                                         </td>
                                         <td>
-                                            <a target="#"
-                                            data-toggle="tooltip"
-                                            data-placement="bottom"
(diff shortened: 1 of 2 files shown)


mtesauro (approved):
Approved

Maffooch (approved):

kiblik:
I suppose, the content of `<th>` should be adjusted somehow. Otherwise, the information is misleading.

blakeaowens (approved):

grendel513 (approved):

Maffooch:
@kiblik the `found_by` field on the findings is a list of test types. Sometimes, this list is totally empty, so falling back on the test type of the test is necessary. I think going back and forth between test type and found by could cause some confusion

PR #9300, opened 2024-01-08
UserContactInfo admin improvements

tomaszn:
User's contact details can change, and the administrator should be able to edit them.

Currently, picking the right object to edit is difficult:

image

This PR configures Django Admin to display a table with the values:

image


tomaszn:
@cneill Please review if possible

mtesauro:
We had a conversation about the Django Admin in the last moderators meeting and since it doesn't enforce constraints for data changes, using it is a great way to corrupt your data.  This is why we don't enable it in our deploys.

We don't want to take any more PR against the Django admin and give the community the impression we are actively supporting that use case.

devGregA:
+1 What Matt says, but thank you for the PR. User data can be edited via the "User" management tabs.

Issue #9308, opened 2024-01-10
Kiuwan CSV Import silently drops some findings


flmarkus:
**Bug description**
When importing a kiuwan CSV report file. Some findings are dropped silently.

The parser in https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/kiuwan/parser.py does not contain the file column in the description. Therefore findings that have the same description and line content have the same hash even if they are found in different files. This will skip the last found file.

**Steps to reproduce**
1. Create a new project
2. Create a new engagement
3. Import the testfile `kiuwan-testscenario-missing-imports.csv` which containes two findings
4. Only the first finding is imported. The second one is dropped silently

**Expected behavior**
Both findings from the test file are imported and displayed.

**Deployment method**
- [ ] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo
- [X] DefectDojo Saas (Pro Account)

**Environment information**
 - DefectDojo Version 2.30.1 - Pro

**Sample scan files**
[kiuwan-testscenario-missing-imports.csv](https://github.com/DefectDojo/django-DefectDojo/files/13886789/kiuwan-testscenario-missing-imports.csv)
[kiuwan-testscenario-missing-imports-working.csv](https://github.com/DefectDojo/django-DefectDojo/files/13886791/kiuwan-testscenario-missing-imports-working.csv)

The second file contains the same import line from the first file to proof that the line itself does not contain any issues. It can imported successfully if the first line does not exist.

manuel-sommer:
This can be closed @mtesauro

flmarkus:
Thanks for the fast response @manuel-sommer.

Unfortunately this does not solve our issue.
I tested you changes from the lasted dev branch.

In the attached test file I manually edited the CVE number to be able to better distinguish the results in the UI.
In the real export from kiuwan the CVE is the same in both lines.

This file better refelcts this:
[kiuwan-testscenario-missing-imports.csv](https://github.com/DefectDojo/django-DefectDojo/files/13903019/kiuwan-testscenario-missing-imports.csv)

The only difference is the file name. I created a pull request highlighting this change.

flmarkus:
@mtesauro Is it possible to open this issue again? Or should I file a new one?

mtesauro:
@flmarkus Yes. Re-opened

manuel-sommer:
This can be closed again @mtesauro 
https://github.com/DefectDojo/django-DefectDojo/pull/9316

[...]

Issue #9307, opened 2024-01-10
Kiuwan CSV Import mixes Software characteristic and Vulnerability type columns


flmarkus:
**Bug description**
There are two types of CSV files that can be exported from kiuwan: Defects and Vulnerabilities.
Vulnerablilities contain two columns: `Software characteristic` and `Vulnerability type`.
Defects only contain `Software characteristic`.
Both type of exports can be imported without errors, but the `Software characteristic` column is displayed as `Vulnerability type`. This is missleading. And the `Vulnerability type` for vulnerabilities is missing.

**Steps to reproduce**
1. Create an project
1. Create an engagement 
1. Upload defects and vulnerabilities CSV Export from Kiuwan
1. See the missleading description content of the imported Findings

**Expected behavior**
The `Software characteristic` column should be contained as `Software characteristic` in the description.
The `Vulnerability type` should be added to the description as well. But only if the column exists which is the case for vulnerability exports.

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo
- [X] DefectDojo Saas (Pro Account)

**Environment information**
 - DefectDojo Version 2.30.1 - Pro

**Sample scan files**
[kiuwan-defects-report.csv](https://github.com/DefectDojo/django-DefectDojo/files/13886510/kiuwan-defects-report.csv)
[kiuwan-vulnerabilities-report.csv](https://github.com/DefectDojo/django-DefectDojo/files/13886511/kiuwan-vulnerabilities-report.csv)

manuel-sommer:
This can be closed @mtesauro

flmarkus:
Thanks for the fast reponse @manuel-sommer!

Unfortunately this change does not solve our issue and (I'm sorry to say this) makes it even worse.
I tested the changes from the dev branch.

Importing the defects file from the PR will fail with this error:
![image](https://github.com/DefectDojo/django-DefectDojo/assets/82434148/a486e104-1ac2-49cf-9eda-1d4b91fb277f)

The reason is, that defects do not have the vulnerability type column.
Please do not ship this version. It will make us unable to continue using DefectDojo for kiuwan defects.

flmarkus:
@mtesauro Is it possible to open this issue again? Or should I file a new one?

manuel-sommer:
Ah, sorry, I overlooked "_But only if the column exists which is the case for vulnerability exports._"

mtesauro:
[...]

PR #9316, opened 2024-01-11
Improve kiuwan import parser

flmarkus:
**Description**

Updated the kiuwan parser to fix some imports. 

1. Added the file to the description to have a unique hash for findings that are the same except the file name. This solves issue #9308
2. Added a check if the vulnerability column is present in the import. This solves issue #9307

**Test results**

After making the changes in the parser the import ran successfully for the attached import file from the issues.
Tested with larger CSV exports from kiuwan production instances as well.


cneill (approved):

mtesauro (approved):
Approved

Maffooch (approved):

grendel513 (approved):

Issue #9340, opened 2024-01-16
Notifications editor in Django Admin: can't create new product notifications


tomaszn:
#9193 introduced a notifications editor for admins, and it works great for editing existing notifications.

However, this still doesn't allow creating new per-product notifications for other users. Fields `user` and `product` are not present in the new/edit admin form.

It is caused by [marking these model fields as `editable=False`](https://github.com/DefectDojo/django-DefectDojo/blob/f68f6144013be528e7605970bdc5318c311b1ee9/dojo/models.py#L3782C9-L3782C9). I think, the reason was to hide them from [non-product-specific notification settings form](https://github.com/DefectDojo/django-DefectDojo/blob/f68f6144013be528e7605970bdc5318c311b1ee9/dojo/forms.py#L2578).

This can be fixed by making the fields editable, and adding them to excluded in `NotificationsForm`. Optionally, they could be hidden or made read only for editing existing notifications, so they appear only when a new one is being created.

There are workarounds for this issue, namely using `./manage.py shell` to create a notification object manually, or editing the database contents. Then the details can be adjusted in Django Admin.

PR #9341, opened 2024-01-16
Django Admin: fix creating new notifications

tomaszn:
This is a fix for #9340. It enables User and Product fields in `Add notifications` editor in Django Admin interface, so new notifications can be set up for any Product and User combination.

**Test results**

Tested manually. Screenshot:

sc


devGregA:
Hi @tomaszn, is it possible to incorporate these changes into notfications rather than admin? We're trying to keep people away from django admin for data integrity purposes.

tomaszn:
@devGregA, I fully understand your concern about data integrity. On the other hand, the alternative for now is editing the database records.

I can imagine such a rework of the View User page. Maybe two new sections, one for user's general notification settings, and another for per-product notifications. Plus a nice wizard that sets up notifications for all products that match some criteria like tags or Product Type. This could be also accessible for regular users to set up and review their notifications.

But that would be a bigger feature. As the code freeze is in progress, what do you think about tackling it after DefectDojo 3.0 is released, and proceeding with this for now?

PR #9275, opened 2024-01-04
🐛 Nikto: Support v2.5.0 format

manuel-sommer:
see issue #9274

Issue #9250, opened 2023-12-31
Scanner filter mechanism in import or reimport.


manuel-sommer:
Some scanners ship a lot of findings, especially scanners which detect vulnerabilities as an agent on enrolled machines. (e.g. Wazuh, MS Defender or Checkov over a shared infrastructure). 
These findings are then uploaded to one destination inside DefectDojo. It would be nice if the upload (import or reimport) would have an option to filter out results (e.g. Team ABC). Then, the result could be uploaded to different Engagements with different filters and the access could be managed on team level. 
This would bring the benefit that these findings are not visible to a huge communitty and are also easier manageable. 

A second scenario is that e.g. Harbor detects also findings which can't be remediated yet as there is no fix present. This could also be adjusted with a filter. Some teams would like to see these results to have a total overview about the security of their application, but some teams would not like to get these results as they only want to focus on issues they can remediate and see them as false positives.

manuel-sommer:
https://github.com/DefectDojo/django-DefectDojo/pull/9351#issuecomment-1974256271

manuel-sommer:
Hi @mtesauro,

as multiple stakeholders (see linked examples) rely on this feature, I would be happy if you could bring this again into consideration between the maintainers if it is possible to approve this feature before 3.X? If this is possible, I would go for a PR to help here. I would like it to improve multiple parsers to finetune the behavior based on different usecases.

manuel-sommer:
@mtesauro, Just a friendly reminder if I can start implementing this feature?

manuel-sommer:
another use case: https://github.com/DefectDojo/django-DefectDojo/issues/12079 
May I submit a PR here @mtesauro ?

Maffooch:
Hi @manuel-sommer I was initially intrigued by your idea, and imagined a bunch of neat filtering rules that could be leveraged by a communication channel with the parsers. That's when it hit me - this is an open door that we cannot walk back through. I believe this will lead to many functionalities for different parsers over time that the project will have to own and maintain forever onwards.

I have learned a lot from DefectDojo over the years, and one of the most valuable lessons is that additional customization often comes as the expense of ease of maintenance and longevity. This lesson has been so impactful for both @mtesauro and I in the open source space that we have unofficially picked up a mantra: "Saying No is just for today, but a Yes is forever" 

For that reason, we must say no to this idea today

PR #9351, opened 2024-01-17
✨ advance import-scan api to have an additional field to finetune vulnerability parsers

manuel-sommer:
see issue #9250

This additional field could help to customize / finetune some scanners directly when parsing the report.

Use cases:

- see my last comment in: https://github.com/DefectDojo/django-DefectDojo/pull/9275
- Harbor has a field which states if a finding has a fix available yet or not. This can be turned on or off within the parser: https://documentation.defectdojo.com/integrations/parsers/file/harbor_vulnerability/
- MSDefender findings could be filtered based on the additional field.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -2135,6 +2135,10 @@ class ImportScanSerializer(serializers.Serializer):
         help_text="If set to True, the tags will be applied to the findings",
         required=False,
     )
+    parser_custom_setting = serializers.CharField(
+        help_text="You can specify custom parser settings, but please take a look at the docs",
+        required=False,
+    )
 
     def save(self, push_to_jira=False):
         data = self.validated_data
@@ -2157,6 +2161,7 @@ class ImportScanSerializer(serializers.Serializer):
         source_code_management_uri = data.get(
             "source_code_management_uri", None
         )
+        parser_custom_setting = data.get("parser_custom_setting", None)
 
         if "active" in self.initial_data:
             active = data.get("active")
@@ -2247,6 +2252,7 @@ class ImportScanSerializer(serializers.Serializer):
                 title=test_title,
                 create_finding_groups_for_all_findings=create_finding_groups_for_all_findings,
                 apply_tags_to_findings=apply_tags_to_findings,
+                parser_custom_setting=parser_custom_setting,
             )
 
             if test:
@@ -2419,6 +2425,10 @@ class ReImportScanSerializer(TaggitSerializer, serializers.Serializer):
         help_text="If set to True, the tags will be applied to the findings",
         required=False
     )
+    parser_custom_setting = serializers.CharField(
+        help_text="You can specify custom parser settings, but please take a look at the docs",
+        required=False,
+    )
 
     def save(self, push_to_jira=False):
         logger.debug("push_to_jira: %s", push_to_jira)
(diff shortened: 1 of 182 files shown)


manuel-sommer:
@Maffooch and @kiblik could you take a look here? I don't know how to proceed. Still getting "Bad Request" when doing an import and obviously the parameter is not passed through to the actual scanner parsers.

kiblik:
> @Maffooch and @kiblik could you take a look here? I don't know how to proceed and need some guidance. Still getting "Bad Request" when doing an import and obviously the parameter is not passed through to the actual scanner parsers.

What does your request (and response) look like? Do you have also logs from Django?

manuel-sommer:
Lol, my import curl was malformed 😆

manuel-sommer:
Could you take a look please @mtesauro ? 
Would be nice if we could merge this also fast due to the number of files changed. Then, I don't have merge conflicts too often.

manuel-sommer:
I would make followup PRs for the other use cases.

manuel-sommer:
@mtesauro could you take a look at this PR? I would love if this could be a way to finetune specific parsers.
We could extend the unittests in a way that this feature stays limited for parsers and also unittests that the fine tune logic gets documented within the parser markdown files if this is used.

manuel-sommer:
As soon as I get a clear opinion if this has the chance to get merged, I will also resolve the conflicts. What is your opinion @Maffooch ?

mtesauro:
> @mtesauro could you take a look at this PR? I would love if this could be a way to finetune specific parsers. We could extend the unittests in a way that this feature stays limited for parsers and also unittests that the fine tune logic gets documented within the parser markdown files if this is used.

@manuel-sommer From the consensus I've gathered about this PR, the resolution is that:
1. It's generally seen as a good concept
2. It's a bit big for right now considering there's 160+ parsers that might need to take this into account.
3. We like this for a 3.0 feature - I've labelled the PR accordingly

manuel-sommer:
> > @mtesauro could you take a look at this PR? I would love if this could be a way to finetune specific parsers. We could extend the unittests in a way that this feature stays limited for parsers and also unittests that the fine tune logic gets documented within the parser markdown files if this is used.
> 
> @manuel-sommer From the consensus I've gathered about this PR, the resolution is that:
> 
>     1. It's generally seen as a good concept
> 
>     2. It's a bit big for right now considering there's 160+ parsers that might need to take this into account.
> 
>     3. We like this for a 3.0 feature - I've labelled the PR accordingly

@mtesauro: Do you have a rough estimated release date for 3.0? I would like to have this feature available as soon as possible to further be able to advance various parsers. If 3.0 will take a while, can I make this PR ready to merge it earlier?

manuel-sommer:
Hi @mtesauro,
may I ask for the reason why you closed this PR? I would love such a feature to be able to finetune some parsers

mtesauro:
@manuel-sommer 
> Hi @mtesauro, may I ask for the reason why you closed this PR? I would love such a feature to be able to finetune some parsers

Sure.

I was reviewing PRs in prep for the March minor release and realized this one wasn't going to be in 2.x so I closed it. It's labeled "Consider for 3.0" which is used for PRs that we're not working on for 2.x but want to consider for 3.x so there' no real reason to keep this open till we're ready to start on 3.x.

About 3.x timelines, I have no real idea. Every time I think I'm confident in a timeline we find something else that needs to be updated / adjusted / etc - latest is the upgrade of Django to the 4.2.10 which alters the timeline. So, it's beginning to feel like I just keep giving what I think is the timeline when I'm typing a reply only to have that change a short time after I write down the timeline somewhere.

It's probably safer to use the "Debian" wording for it's releases which is "when its ready". That's about the only accurate thing I can say right now today about the move to 3.x.

HTH

manuel-sommer:
Hi @mtesauro,

may I throw this feature again into consideration earlier than 3.x.x. I guess multiple users would be happy if we could finetune scanners with custom values based on the scanner through api settings. I have already collected multiple use cases. Also, I could prepare a documentation page in advance and collaborate with you to get this on the road (maybe through multiple pull requests to split the work). But it makes only sense to work on it if the maintainers have time to also review and merge it.

PR #9322, opened 2024-01-11
:tada: merge OpenVAS XML and CSV parsers

manuel-sommer:
one parser for two file formats

PR #9355, opened 2024-01-17
:sparkles: merge clair and clair klar together

manuel-sommer:
merge the two parsers clair and clair klar


mtesauro (approved):
Approved

grendel513 (approved):

blakeaowens (approved):

Maffooch (changes requested):
When changing the name of a test type, it is important to do the following as well:
  - remove the Clair Klar test type object - this is to remove the possibility that a user attempts to import a test to Clair Klar and get a "parser not found" error
  - for any existing Clair Klar finding, the `found_by` column needs to be updated to Clair - this field is used to display all the test types that this finding belongs (multiple in cases of detected duplicates between tools)
  - [Here is a migration that can be used as a reference](https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/db_migrations/0187_nessus_to_tenable.py)

Maffooch:
Also would be good to add a breaking change section to the upgrade notes for folks that could be using Clair Klar in an API based import. [Here is an example from then Nessus -> Tenable conversion was made](https://documentation.defectdojo.com/getting_started/upgrading/2.23/)

mtesauro:
Very good point about the breaking change @Maffooch

manuel-sommer:
@Maffooch, could you review it very closely? I don't have much experience with db migrations and don't want to messup the db. I also included database migrations for OpenVAS: https://github.com/DefectDojo/django-DefectDojo/pull/9322

manuel-sommer:
Could you please merge this PR before https://github.com/DefectDojo/django-DefectDojo/pull/9375 because of the db migrations.

manuel-sommer:
@Maffooch do you have time to take a look at this?

Maffooch (approved):
Great work on the migration @manuel-sommer! I apologize for not catching the need for the openvas migration on your previous PR

PR #9360, opened 2024-01-18
Endpoint Status: Accommodate Risk Exceptions

Maffooch:
When setting or removing a risk exception on a given finding, the endpoint statuses were not updated. With this PR, endpoint statuses on a finding that is being risk accepted will have a status of mitigated + risk accepted. When the risk exception expires or is removed, the endpoint statuses for a given finding will revert to active to match the finding

[sc-3960]


grendel513 (approved):

mtesauro (approved):
Approved

cneill (approved):

devGregA (approved):

PR #9361, opened 2024-01-18
Add "First Seen" Parser Flag

Maffooch:
Adds a "First Seen" parser flag to select parsers for users wanting the finding date to be reflected as the first time a vulnerability's detected rather then the last time

[sc-3965]


mtesauro (approved):
Approved

grendel513 (approved):

blakeaowens (approved):

dojo/tools/veracode_sca/parser.py
```diff
@@ -51,7 +52,15 @@ def _get_findings_json(self, file, test):
             if issue.get("issue_type") != "vulnerability":
                 continue
 
-            date = parser.parse(issue.get("created_date"))
+            # Get the date based on the first_seen setting
+            try:
+                if settings.USE_FIRST_SEEN:
+                    date = parser.parse(issue.get("created_date"))
+                else:
+                    date = parser.parse(issue.get("created_date"))
```
kiblik, line 60: These looks same for me

dojo/tools/veracode_sca/parser.py
```diff
@@ -166,9 +175,19 @@ def get_findings_csv(self, file, test):
...
-            )
+            # Get the date based on the first_seen setting
+            try:
+                if settings.USE_FIRST_SEEN:
+                    date = datetime.strptime(
+                        row.get("Issue opened: Scan date"), "%d %b %Y %H:%M%p %Z"
+                    )
+                else:
+                    date = datetime.strptime(
+                        row.get("Issue opened: Scan date"), "%d %b %Y %H:%M%p %Z"
```
kiblik, line 186: These as well

dojo/tools/veracode_sca/parser.py
Maffooch: They are the same, but this time it is intentional. In the unit test reports for this parser, there is not any available data in the report to satisfy the notion of a "last seen" date. Before this change, the `created_date` field was used, so it seemed appropriate to maintain that same behavior for both cases.

I am open to any suggestions that could make more sense though!

dojo/tools/veracode_sca/parser.py
Maffooch: They are the same, but this time it is intentional. In the unit test reports for this parser, there is not any available data in the report to satisfy the notion of a "last seen" date. Before this change, the `Issue opened: Scan date` field was used, so it seemed appropriate to maintain that same behavior for both cases.

I am open to any suggestions that could make more sense though!

cneill (approved):

kiblik:
One more parser ;) https://github.com/Maffooch/django-DefectDojo/pull/2

dojo/tools/veracode_sca/parser.py
kiblik: In that case, is it necessary to replace the original implementation if there is no change in behavior?

dojo/tools/veracode_sca/parser.py
kiblik: same as https://github.com/DefectDojo/django-DefectDojo/pull/9361/files#r1459718238

Maffooch:
Thanks @kiblik

PR #9117, opened 2023-12-05
:bug: fix dependencytrack deduplication

manuel-sommer:
Take a look at the following issue: https://github.com/DependencyTrack/dependency-track/issues/1915

Issue #8778, opened 2023-10-03
Dependency-Track Hash Code computation: Add severity in the hash code calculation of the finding


kmouzoul23:
**Problem description**
At the moment severity is not part of the Unique Hash code for findings found via Dependency Track. This configuration leads to misleading information within Defect Dojo when findings CVSS score changes.
DefectDojo/django-DefectDojo/tree/master/dojo/settings)
/settings.dist.py: 
'Dependency Track Finding Packaging Format (FPF) Export': ['component_name', 'component_version', 'vulnerability_ids'] 

**Steps to reproduce the behavior**:
- Upload a Dependency-Track Report
- Change the severity of a finding included in the report
- Reupload this Dependency-Track file to the same engagement
- View the result (finding with changed severity will be marked as Inactive, Duplicate)

**Expected behavior**
The duplicate finding (with changed severity) should be considered the active one in Defect Dojo

**Deployment method**
- [ ]  Docker Compose
- [x]  Kubernetes
- [ ]  GoDojo

quirinziessler:
@kmouzoul23 The hash code is without the severity on purpose. Otherwise it would cause a lot of false truth in DD. If you want to have this feature I would recommend you to open up a issue to the DependencyTrack project or even better create a PR for this there. Resturcturing the sync to use patch methods for existing findings will probably be the most easiest & effective way.

manuel-sommer:
https://github.com/DefectDojo/django-DefectDojo/pull/9117

I guess this can be closed @mtesauro

kepten:
I agree with @quirinziessler that this should have been addressed in DT and not in DD. CVE severities are sometimes unassigned at first or get recalculated afterwards but they should still be considered the same finding in DD.
I'm still investigating but this change seems to result in existing DT findings with old hash_code to be closed and then re-created with the new hash_code which totally messed up our SLA tracking.
@manuel-sommer  I think https://github.com/DefectDojo/django-DefectDojo/pull/9117 should be reverted because I'd imagine this will cause unnecessary trouble for lots of users once they happen to upgrade to DD v2.30.x.

manuel-sommer:
@kepten and @quirinziessler  I had a quick look at the code.
Do findings with unassigned severity get a critical severity? Maybe we should change this behavior in a way that they will become an info finding. Then, this makes more sense. Maybe I had it wrong in my mind. 
[...]

Issue #6774, opened 2022-08-31
Upload / Reimport of reports should consider severity


manuel-sommer:
**Is your feature request related to a problem? Please describe**
Some tools update the severity of findings. Here some examples:

- DependencyTrack tracks CVEs. Due to false positive reportings, CVEs are closed as rejected e.g. https://nvd.nist.gov/vuln/detail/CVE-2021-20095. Then, DependencyTrack updates the severity of these vulnerabilities to "unassigned". This is basically a "Informational" finding in DefectDojo. However, the new severity is not updated when DependencyTrack syncs the finding to DefectDojo and the finding remains open.
- DrHeader supports the way to set custom severities. If this changes, the severity is not updated in DefctDojo https://github.com/Santandersecurityresearch/DrHeader/pull/170 

**Describe the solution you'd like**
The severity is updated for all findings when they are uploaded (upload or reimport). This includes an SLA adjustment.

manuel-sommer:
https://github.com/DefectDojo/django-DefectDojo/issues/7728

valentijnscholten:
@manuel-sommer Do you have a proposal on how this should work? There are at least 20 parsers that have the `severity` field in their hash_code configuration. I think this is not needed anyway, but to make reimport update the severity field the `seveirty` field has to be remove from any hash_code configuration/calculation. It must be removed from allowed fields for hash_code configuration.

manuel-sommer:
What we could do is: update the hash_code config for a parser and recalculate the hash_code for this parser in a db_migration. As a preanalysis, we could evaluate which parsers could affect this feature as they have too less available fields to be unique and need the severity field.
We could also do an if-statement in the reimport: if field severity not available in settings, then update in a reimport.

manuel-sommer:
@valentijnscholten, shall I make a PR for an import if field severity is not available in settings? or what is your opinion?

devaroop:
I am having the same issue with AWS Security Hub and AWS Inspector2 Scan types where severity is not being updated on reimport of json reports.

Please help with a fix

PR #9370, opened 2024-01-19
Dependencytrack default severity

manuel-sommer:
Due to the deduplication adaption in this PR https://github.com/DefectDojo/django-DefectDojo/pull/9117 the default severity shouldn't be "Critical" as per discussion in https://github.com/DefectDojo/django-DefectDojo/issues/8778


Maffooch (approved):

dojo/tools/dependency_track/parser.py
```diff
@@ -201,8 +201,8 @@ def _convert_dependency_track_finding_to_dojo_finding(self, dependency_track_fin
         dependency_track_severity = dependency_track_finding['vulnerability']['severity']
         vulnerability_severity = self._convert_dependency_track_severity_to_dojo_severity(dependency_track_severity)
         if vulnerability_severity is None:
-            logger.warning("Detected severity of %s that could not be mapped for %s. Defaulting to Critical!", dependency_track_severity, title)
-            vulnerability_severity = "Critical"
+            logger.warning("Detected severity of %s that could not be mapped for %s. Defaulting to Informational!", dependency_track_severity, title)
```
manuel-sommer, line 204: @Maffooch and @mtesauro:

The problem we face with DependencyTrack findings is that single findings are not patched if the severity changes (e.g. if the severity is not yet classified).

Thus, we can deal with this in two ways:

 - The default severity for findings is Critical (which till now is "Critical, see this PR) Then, severities are not deduplicated and in general Teams will have a higher load of critical findings and they have to deal with them. This puts more pressure on the teams then necessary. Also, unclassified findings often don't have a patch available yet. On the other hand, switching severities are not deleted. Thus, it is easier to keep track on vulnerabilities.

- The default severity for findings is Informational and we activate the deduplication in settings.dist.py. Then, these findings are closed automatically, but have the real severity once classified right. In this case, findings are closed and a new finding is generated as soon as the severity changes. Keeping track on vulnerabilities is harder.

Both ways are not the 100% right way as the capabilities are still limited due to the missing patch functionality of finding's severities from DependencyTrack to DefectDojo.
Maybe as an interim solution, we could implement both ways and choose the right one with https://github.com/DefectDojo/django-DefectDojo/pull/9351. We could implement a small logic in DependencyTrack to switch between the two options via parser_custom_setting.

What are your opinions and did I forget something?

Maffooch: This PR and #9371 make sense to go together. This is a tricky problem to solve that is not just limited to dependency track.

manuel-sommer: But now, if me merge both PRs, we have another problem: Info findings will stay with severity "info" eventhough the severity cfhanges. So, info findings would have to be analyzed.

kepten: I agree with @manuel-sommer that merging the two would "hide" real CVEs (most ppl/companies ignore info severity findings) if they were created and synced to DD before their real severity was assigned.

Let me collect my thoughts on this issue and then brainstorm on solution ideas:
- CVE severities can change and this is a totally valid case (e.g. withdrawn or created with unassigned severity initially then set to the real one, etc) which both DT and DD should be prepared to handle. AFAIK DT updates the severities upon each sync with CVE DBs, so they are up-to-date in the import data, so it is DD which doesn't handle it properly yet.
- AFAIK `hash_code` calculation is used to find duplicates. I think duplicate issues are the ones which are about the _same vuln_ but might be coming from different sources (e.g https://nvd.nist.gov/vuln/detail/CVE-2023-52079 and https://github.com/kriszyp/msgpackr/security/advisories/GHSA-7hpj-7hhx-2fgx are the same vuln) and therefore potentially can have different metadata (like severity, references, etc). Therefore I think including severity into the hash_code is a bad idea since it'd break deduplication (e.g. the above two CVEs would have ended up as two different findings in DD when the severity was already updated in GHSA to medium but still unassigned in CVE). Therefore I think https://github.com/DefectDojo/django-DefectDojo/pull/9371 should be merged asap.
- Regarding this PR: I agree with it, I think unassigned CVEs should fall back to Info severity rather than Critical. However, unless severity patching is implemented in DD, this change would "hide" those findings which is I think worse than the current Critical one, so I'd only merge this, once there is another PR which solves severity patching
- And now we arrive to the root cause which needs a solution: how to handle severity (and potentially other metadata "patching" in DD). This is not a trivial problem indeed since duplicated findings can have different metadata as described above. AFAIK DD stores all duplicated findings as separate records in the DB and then a scheduled task might remove duplicates according to the setting [here](https://github.com/DefectDojo/django-DefectDojo/pull/9371). This means that if two findings which are duplicates have different metadata in the import, those metadata would be stored separately in their respective DD record, so nothing would be overridden. This should mean that simply overriding the metadata of existing records at import time would be a trivial solution and should work. However, that would cause issues if such metadata gets updated/patched which is used for hash calculation (e.g `vulnerability_ids` in the case of Dependency Track imports) because then the hash_code would need to be updated as well which can result in new findings to become duplicates of others and existing ones not to be duplicates anymore. This would result in an inconsistent state which could be resolved if after any hash_code updates, some tasks would update all the findings in the DB which either have the old or the new hash_code values and "fix" the duplicate connections (I think `duplicate_finding_id` field). This is my best idea so far but I might be missing something or misunderstood the internals of DD so feel free to correct me if I'm wrong or you have better alternative.

manuel-sommer: Linking this issue also: https://github.com/DefectDojo/django-DefectDojo/issues/6774

mtesauro (approved):
Approved

blakeaowens (approved):

manuel-sommer:
one more to go, then this is also ready to be merged.

grendel513 (approved):

PR #9371, opened 2024-01-19
Revert adding severity to Dependency Track hash_code calculation

kepten:
**Description**

This reverts https://github.com/DefectDojo/django-DefectDojo/pull/9117 as per the discussion in https://github.com/DefectDojo/django-DefectDojo/issues/8778#issuecomment-1900406114


dojo/settings/settings.dist.py
```diff
@@ -1207,7 +1207,7 @@ def saml2_attrib_map_format(dict):
     'SonarQube API Import': ['title', 'file_path', 'line'],
     'Dependency Check Scan': ['title', 'cwe', 'file_path'],
     'Dockle Scan': ['title', 'description', 'vuln_id_from_tool'],
-    'Dependency Track Finding Packaging Format (FPF) Export': ['component_name', 'component_version', 'vulnerability_ids', 'severity'],
+    'Dependency Track Finding Packaging Format (FPF) Export': ['component_name', 'component_version', 'vulnerability_ids'],
```
manuel-sommer, line 1210: @Maffooch and @mtesauro: 

The problem we face with DependencyTrack findings is that single findings are not patched if the severity changes (e.g. if the severity is not yet classified). 

Thus, we can deal with this in two ways:
- The default severity for findings is Critical (which till now is "Critical, see PR https://github.com/DefectDojo/django-DefectDojo/pull/9370) Then, severities are not deduplicated and in general Teams will have a higher load of critical findings and they have to deal with them. This puts more pressure on the teams then necessary. Also, unclassified findings often don't have a patch available yet. On the other hand, switching severities are not deleted. Thus, it is easier to keep track on vulnerabilities.
- The default severity for findings is Informational and we activate the deduplication in settings.dist.py. Then, these findings are closed automatically, but have the real severity once classified right. In this case, findings are closed and a new finding is generated as soon as the severity changes. Keeping track on vulnerabilities is harder. 

Both ways are not the 100% right way as the capabilities are still limited due to the missing patch functionality of finding's severities from DependencyTrack to DefectDojo. 
Maybe as an interim solution, we could implement both ways and choose the right one with https://github.com/DefectDojo/django-DefectDojo/pull/9351. We could implement a small logic in DependencyTrack to switch between the two options via parser_custom_setting.

What are your opinions and did I forget something?

manuel-sommer (commented):
@Maffooch and @mtesauro:

The problem we face with DependencyTrack findings is that single findings are not patched if the severity changes (e.g. if the severity is not yet classified).

Thus, we can deal with this in two ways:

-  The default severity for findings is Critical (which till now is "Critical, see PR 
    https://github.com/DefectDojo/django-DefectDojo/pull/9370) Then, severities are not deduplicated and in general Teams will have a higher load of critical findings and they have to deal with them. This puts more pressure on the teams then necessary. Also, unclassified findings often don't have a patch available yet. On the other hand, switching severities are not deleted. Thus, it is easier to keep track on vulnerabilities.

- The default severity for findings is Informational and we activate the deduplication in settings.dist.py. Then, these findings are closed automatically, but have the real severity once classified right. In this case, findings are closed and a new finding is generated as soon as the severity changes. Keeping track on vulnerabilities is harder.

Both ways are not the 100% right way as the capabilities are still limited due to the missing patch functionality of finding's severities from DependencyTrack to DefectDojo.
Maybe as an interim solution, we could implement both ways and choose the right one with https://github.com/DefectDojo/django-DefectDojo/pull/9351. We could implement a small logic in DependencyTrack to switch between the two options via parser_custom_setting.

What are your opinions and did I forget something?

mtesauro (approved):
Approved

grendel513 (approved):

blakeaowens (approved):

Issue #7468, opened 2023-01-20
Sysdig Scan Report Import Parser


ronkara:
**Is your feature request related to a problem? Please describe**
Hello,
I would like to know how I can go about requesting a feature update to this project that would create an importation parser that can read Sysdig container scanner report CSV files. The Sysdig report contains the following attributes and they are not tailorable or modifiable within the Sysdig application (you get what you see).

VulnerabilityID,Severity,Packagename,ImageID,Imagename,Imagetag,Vulnerabilitytype,CVSSv2vector,CVSSv2basescore,CVSSv3vector,CVSSv3basescore,Vulnlink,Fixversion,Vulnexception,Packageversion,Imageadded,Pod,Namespace,ContainerName,ContainerID,ClusterName,Deployment,Hostname

I was capable of modifying my report file to match the Generic Findings attribute fields as displayed below and though this succeeded in importing it is a cludgy solution at best and requires manually modifying the file each and every time or developing some external manual process to convert the data.

Date,Title,CweId,Url,Severity,Description,Mitigation,Impact,References,Active,Verified,CVE

**Describe the solution you'd like**
As an ISSO, I want to be able to utilize DefectDojo to import all three scan source types (Tenable and Burp are ok) into a single consolidate view so that reporting may be done to FedRAMP PMO and Federal agency security personnel on a monthly basis as required.

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions or features you have considered.
The only alternative I have at this time is to take the aforementioned scanning attributes and manually modify them into the Generic Findings parser importation format which defeats the required automated nature of creating our monthly POAM (Plan of Action and Milestone) reports (different formats for FedRAMP and Federal agencies).

**Additional context**
Add any other context, screenshots, sketch, code snippet, etc. about the feature request here.

Sysdig report attributes (spaces have been removed)
"VulnerabilityID,Severity,Packagename,ImageID,Imagename,Imagetag,Vulnerabilitytype,CVSSv2vector,CVSSv2basescore,CVSSv3vector,CVSSv3basescore,Vulnlink,Fixversion,Vulnexception,Packageversion,Imageadded,Pod,Namespace,ContainerName,ContainerID,ClusterName,Deployment,Hostname"

ronkara:
That appears to be the case but curious if the other one is being worked on and if there is any time line for making it available.  Also curious if my report in its details gives a better definition of the issue and the output format fields.

manuel-sommer:
[...]

Issue #6939, opened 2022-10-04
Add a Sysdig importer


e-minguez:
**Scanner Name**

Sysdig vulnerability reports

https://docs.sysdig.com/en/docs/sysdig-secure/vulnerabilities/reporting/

Sysdig Secure can schedule asynchronous reports about detected runtime vulnerabilities along with package and image data.

**Sample File**

https://github.com/e-minguez/sysdig-to-defectdojo/blob/main/examples/20220926-090023-2Bw41vBq8qfnCThYiYDhr6EoMKQ-2FIbycpRUupe7syyasV3bBSV0kV.json

** Notes **

I tried the other way around, converting the Sysdig report to a generic one https://github.com/e-minguez/sysdig-to-defectdojo but I guess it should be better to do it the opposite way.

coheigea:
Are you planning to contribute a pull request for this feature @e-minguez ?

damiencarol:
I took a look at the report format and it's configurable, it will be hard as most of the data could be optional.

manuel-sommer:
Can you close this @mtesauro ?

miles3719:
Since this shows up in google when you search for Sysdig and Defect Dojo, just pointing anyone who gets here, a parser was created

https://github.com/DefectDojo/django-DefectDojo/pull/8868

and by the time you read this, is in the main branch already

e-minguez:
cc @manuelbcd

PR #9377, opened 2024-01-21
:tada: advance sysdig to also parse json

manuel-sommer:
#6939 
#7468


mtesauro (approved):
Approved

dojo/tools/sysdig_reports/parser.py
```diff
@@ -11,48 +12,124 @@ class SysdigReportsParser(object):
...
             return ()
 
+    def parse_json(self, data, test):
+        try:
+            vulnerability = data["data"]  # json output of https://pypi.org/project/harborapi/
+        except (KeyError):
+            pass
+        # Early exit if empty
+        if 'data' not in locals() or vulnerability is None:
+            return list()
```
cneill, line 46: How about something like this simplified approach?

```suggestion
        vulnerability = data.get("data")
        if not vulnerability:
            return list()
```

cneill (approved):

grendel513 (approved):

PR #9385, opened 2024-01-22
Add support for DD_APPEND_SLASH

kiblik:
Avoid situations like https://owasp.slack.com/archives/C2P5BA8MN/p1705702146187379

Better context description: https://learndjango.com/tutorials/trailing-url-slashes-django

Documentation: https://docs.djangoproject.com/en/4.2/ref/settings/#append-slash


dojo/settings/settings.dist.py
```diff
@@ -40,6 +40,7 @@
     DD_SECURE_CONTENT_TYPE_NOSNIFF=(bool, True),
     DD_CSRF_COOKIE_SAMESITE=(str, 'Lax'),
     DD_SESSION_COOKIE_SAMESITE=(str, 'Lax'),
+    DD_APPEND_SLASH=(bool, False),
```
Maffooch, line 43: We should default to True here to match what it has always been for folks as well as the default set by Django

Maffooch (changes requested):

dojo/settings/settings.dist.py
kiblik, line 43:
```suggestion
    DD_APPEND_SLASH=(bool, True),
```

Maffooch (approved):

mtesauro (approved):
Approved

kiblik:
Hi @grendel513 / @devGregA / @cneill / @blakeaowens. I would like to ask for a review.

grendel513 (approved):

blakeaowens (approved):

PR #9403, opened 2024-01-24
Add CVSS vector string to JIRA description

Maffooch:
When pushing a finding to JIRA, the CVSS score is already being passed, but the vector string is equally important


cneill (approved):

mtesauro (approved):
Approved

PR #9441, opened 2024-01-30
Add API tests/lints/validators (like oasdiff)

kiblik:


Maffooch:
We have been working around this topic internally, and have been inspired by you! We are about to update the documentation, and will be producing that information in those docs. To make that easier, we would like to roll each piece into one workflow from the ground up

PR #9442, opened 2024-01-30
API: Fix typed-enum from spectral/limeflight

kiblik:
During add #9441 I bumped into the error which needed to be fixed.

OAS defines for some fields option to choose (ChoiceFilter). Parameters are defined in OAS as nullable strings but options are integers.

This will transfer definitions to
```yaml
  - '1'
  - '2'
  - ...
```
and still keep compatibility with existing systems.


Maffooch:

PR #9462, opened 2024-02-02
Parse GitHub vulnerability version

coheigea:
**Description**

The affected version from GitHub is not imported by the parser. I think this change is better than the one proposed in https://github.com/DefectDojo/django-DefectDojo/pull/9457 which just adds the range of impacted versions, and not the range of the vulnerable version detected.

**Test results**

Test added.

### DIFF
--- a/dojo/tools/github_vulnerability/parser.py
+++ b/dojo/tools/github_vulnerability/parser.py
@@ -66,6 +66,9 @@ class GithubVulnerabilityParser(object):
             if "vulnerableManifestPath" in alert:
                 finding.file_path = alert["vulnerableManifestPath"]
 
+            if "vulnerableRequirements" in alert and alert["vulnerableRequirements"].startswith("= "):
+                finding.component_version = alert["vulnerableRequirements"][2:]
+
             if "createdAt" in alert:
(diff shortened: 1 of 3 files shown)


Demaz93:
Theoretically, you are right, `vulnerableRequirements` fits better, but in case you have multiple occurrences in the same file you probably lose something to fix because you are looking for the exact provided version for example:
Vulnerability related to library XYZ with vulnerable version < 1.2.3
in my code, I use version 0.8.2 and version 1.1.1 (because related to some dependencies of different libraries)

In this case, GitHub provides only one vulnerability and inside Defectdojo I'll have one finding with only 1.1.1 or 0.8.2 vulnerable version, which is false because I have two occurrences and I have to fix both.
Instead in the other PR, you have range so you have to check all the occurrences for the same library.

What do you think @coheigea about this edge-case scenario?

coheigea:
@Maffooch please review

mtesauro (approved):
Approved

grendel513 (approved):

blakeaowens (approved):

Maffooch (approved):

PR #9473, opened 2024-02-05
finding sla expiration date field (part one)

blakeaowens:
- Addition of new `sla_expiration_date` field to the Finding model
 - Addition of model method to update the `sla_expiration_date` field, `set_sla_expiration_date()`
 - Call to the `set_sla_expiration_date()` method made on each Finding `.save()`
 - Updated `sla_days_remaining()` method to perform calculation based on the `sla_expiration_date` Finding model field
 - Updated `sla_deadline()` method to return the new `sla_expiration_date` Finding model field
 - Signal to asynchronously update all Findings under a Product when a given Product's SLA Configuration is changed from one SLA Configuration to another (i.e. Product A's SLA Configuration changes from Default to some other SLA Configuration)
 - Signal to asynchronously update all Findings under an SLA Configuration when an SLA Configuration is updated (i.e. Critical days changed from 7 to 5)
 - A bugfix for a 500 error occurring when deleting an SLA Configuration that is used by one or more products
 - Unit tests to test calculating a Finding's SLA expiration date, as well as SLA days remaining when: a new Finding is created and saved, a Finding is updated (i.e. severity), a Product's SLA configuration is changed, or an SLA configuration is saved

This PR is the first part of introducing the `sla_expiration_date` field. A future PR will introduce a migration to run the calculations for all Findings. We decided to split this up to give more warning about a potentially long database migration in that future PR.


mtesauro (approved):
Approved

devGregA (approved):

dojo/db_migrations/0200_finding_sla_expiration_date_product_async_updating_and_more.py
```diff
@@ -0,0 +1,31 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='finding',
+            name='sla_expiration_date',
+            field=models.DateField(blank=True, help_text="(readonly)The date SLA expires for this finding. Empty by default, causing a fallback to 'date'.", null=True, verbose_name='SLA Expiration Date'),
+        ),
+        migrations.AddField(
+            model_name='product',
+            name='async_updating',
+            field=models.BooleanField(default=False, help_text='Findings under this SLA configuration are asynchronously being updated'),
```
cneill, line 24: Should this be "Product or SLA configuration"? Seems like this is doing double duty for 1) SLA config itself being modified or 2) different SLA config being chosen for this Product?

```suggestion
            field=models.BooleanField(default=False, help_text='Findings under this Product or SLA configuration are asynchronously being updated'),
```

dojo/models.py
```diff
@@ -998,6 +1038,37 @@ class Product(models.Model):
         blank=False,
         verbose_name=_("Disable SLA breach notifications"),
         help_text=_("Disable SLA breach notifications if configured in the global settings"))
+    async_updating = models.BooleanField(default=False,
+                                            help_text=_('Findings under this SLA configuration are asynchronously being updated'))
```
cneill, line 1042: Should this be "Product or SLA configuration"? Seems like this is doing double duty for 1) SLA config itself being modified or 2) different SLA config being chosen for this Product?

```suggestion
                                            help_text=_('Findings under this Product or SLA configuration are asynchronously being updated'))
```

dojo/models.py
```diff
@@ -998,6 +1038,37 @@ class Product(models.Model):
...
+                self.sla_configuration = initial_sla_config
+
+        super(Product, self).save(*args, **kwargs)
+
+        # if the initial sla config exists and async finding update is not running
+        if initial_sla_config is not None and not self.async_updating:
+            # get the new sla config from the saved product
+            new_sla_config = getattr(self, 'sla_configuration', None)
+            # if the sla config has changed, update finding sla expiration dates within this product
+            if new_sla_config and (initial_sla_config != new_sla_config):
```
cneill, line 1060: Won't this be true in both the case when the same SLA config has been modified and the case where a new SLA config has been selected? Seems like the SLA config should be marked `async_updating` only in the first case, not the second?

cneill (commented):
A few questions/comments

dojo/models.py
blakeaowens: We have to block both directions. For example, if `SLA_Configuration A` is used by `Product A`, but a new `SLA_Configuration B` is applied to `Product A`, we need to lock `SLA_Configuration B` from being changed (while the asynchronous update occurs) _and_ `Product A` from changing to a different SLA configuration (also while the asynchronous update occurs).

blakeaowens: This is also true for the other way around. If `Product A` is using `SLA_Configuration A`, and someone changes `SLA_Configuration A`'s SLA days, we need to block all products utilizing `SLA_Configuration A` from changing to a different SLA configuration while that async task runs. The reason I blocked this is we have no way to cancel a celery task once it has started (I could not find a way, at least).

cneill (approved):

PR #9490, opened 2024-02-06
Be strict about Warnings during testing

kiblik:
This PR sets more strict rules for test environments to be motivated to fix issues as soon we know about them.
It also adds some exceptions (`filterwarnings`) that need to be solved - they should be solved one by one to avoid large PR.

- `RemovedInDjango50Warning` - Fixed in https://github.com/DefectDojo/django-DefectDojo/pull/9491
- `invalid escape sequence.*`
- `'cgi' is deprecated and slated for removal in Python 3.13` - Fixed in https://github.com/DefectDojo/django-DefectDojo/pull/9108
- `DateTimeField .+ received a naive datetime .+ while time zone support is active.`
- `unclosed file .+`

For integration tests:
- `executable_path has been deprecated, please pass in a Service object`
- `use options instead of chrome_options`
- `desired_capabilities has been deprecated, please pass in a Service object`
- `It is deprecated to return a value that is not None from a test case`

### DIFF
--- a/docker/entrypoint-integration-tests.sh
+++ b/docker/entrypoint-integration-tests.sh
@@ -29,6 +29,9 @@ export CHROMEDRIVER
 CHROME_PATH=/opt/chrome/chrome
 export CHROME_PATH
 
+# We are strict about Warnings during testing
+export PYTHONWARNINGS=error
+
 # Run available unittests with a simple setup
 # All available Integrationtest Scripts are activated below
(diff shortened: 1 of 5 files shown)


mtesauro (approved):
Approved

grendel513 (approved):

Maffooch (approved):

blakeaowens (approved):

PR #9491, opened 2024-02-06
Remove DD_USE_L10N

kiblik:
The Setting `USE_L10N` is overridden later in the `settings.dist.py`. It is not necessary to keep it as EnvVar because there is no real effect.


Maffooch (approved):
Makes sens. The defaults is True in later versions in django, and then deprecated in 5.x as it will always be enabled

https://docs.djangoproject.com/en/4.2/ref/settings/#use-l10n

cneill (approved):

mtesauro (approved):
Approved

grendel513 (approved):

PR #9493, opened 2024-02-06
Upgrade Django to 4.2.13

kiblik:
Upgrade to Django 5.x https://github.com/DefectDojo/django-DefectDojo/pull/9258 might be quite complated but we need to go to at least 4.2.13 because 4.1.x is not supported anymore (check https://www.djangoproject.com/download/)

Needs to be checked/considered:
- [ ] https://docs.djangoproject.com/en/4.2/howto/upgrade-version/#deployment
- [ ] https://docs.djangoproject.com/en/5.0/releases/4.2/
       - Might be handy: _[CharField.max_length](https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.CharField.max_length) is no longer required to be set on PostgreSQL, which supports unlimited VARCHAR columns._
       - We might need to change some lines for `makemigrations`: _The [makemigrations --check](https://docs.djangoproject.com/en/5.0/ref/django-admin/#cmdoption-makemigrations-check) option no longer creates missing migration files._

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -36,7 +36,7 @@ jobs:
           "tests/tool_config.py",
           "openapi-validatator",
         ]
-        profile: ["mysql-rabbitmq", "postgres-redis"]
+        profile: ["postgres-rabbitmq", "postgres-redis"]
         os: [alpine, debian]
       fail-fast: false
 
@@ -58,10 +58,10 @@ jobs:
       - name: Set integration-test mode
         run: ln -s docker-compose.override.integration_tests.yml docker-compose.override.yml
 
-      # phased startup with MySQL and RabbitMQ so we can use the exit code from integrationtest container
-      - name: Start Dojo MySQL + RabbitMQ
-        if: matrix.profile == 'mysql-rabbitmq'
-        run: docker compose --profile ${{ matrix.profile }} --env-file ./docker/environments/${{ matrix.profile }}.env up --no-deps -d mysql nginx celerybeat celeryworker mailhog uwsgi rabbitmq
+      # phased startup with PostgreSQL and RabbitMQ so we can use the exit code from integrationtest container
+      - name: Start Dojo PostgreSQL + RabbitMQ
+        if: matrix.profile == 'postgres-rabbitmq'
+        run: docker compose --profile ${{ matrix.profile }} --env-file ./docker/environments/${{ matrix.profile }}.env up --no-deps -d postgres nginx celerybeat celeryworker mailhog uwsgi rabbitmq
         env:
           DJANGO_VERSION: ${{ matrix.os }}
           NGINX_VERSION: ${{ matrix.os }}
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,7 +24,7 @@ django-slack==5.19.0
 git+https://github.com/DefectDojo/django-tagging@develop#egg=django-tagging
 django-watson==1.6.3
 django-prometheus==2.3.1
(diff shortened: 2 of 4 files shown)


Maffooch:
I think the best approach here is to remove tests for MySQL by removing the following [entry from the integration tests](https://github.com/DefectDojo/django-DefectDojo/blob/738dca4534382049ca7587a60fb6b33715213943/.github/workflows/integration-tests.yml#L39C19-L39C35) and then [replace all the mysql references with Postgres](https://github.com/DefectDojo/django-DefectDojo/blob/738dca4534382049ca7587a60fb6b33715213943/.github/workflows/rest-framework-tests.yml#L35C1-L51C104)

kiblik:
@Maffooch, noticed that you implemented https://github.com/DefectDojo/django-DefectDojo/pull/5899.
Now `errors.append(ValidationError('Query "{}" has invalid format - It contains the NULL character. The following action was taken: {}'.format(old_value, action_string)))` breaks test.
If correction (`query.replace(remove_str, '%00')`) was performed, is it still necessary to raise `ValidationError`?
I'm considering change it to `logger.error`.

kiblik:
Help needed

I'm facing unusual behavior. Based on unittests, in `test_rest_framework.py` I adjusted results for `deleted_objects`. For `Language_Type` from `2` to `1` and for `TestType` from `18` to `5`.
At first, this kind of adjustment should not be done (if there is no change in the data of the algorithm). But I wanted to see the results.

Now, the complicated part:
- Why it happen?
- Why unittests for `Alpine` pass but for `Debian` fail (with a request to roll back from `5` to `18` for `TestsTest` and not mentioning `Language_Type`)

Maffooch:
@kiblik this is a very strange error, and I do not understand why it is happening. I am hoping to get some time in the near future to pull this PR down and play around with it

kiblik:
@Maffooch, it looks like the mentioned issue is not connected to the upgrade to Django 4.2 but to migration tests from MySQL to Postgres
This fails in the same way https://github.com/DefectDojo/django-DefectDojo/pull/9885

Maffooch (approved):

kiblik:
I'm happy that we are ready, just be careful with timing:
> **Dropped support for MySQL 5.7**
> Upstream support for MySQL 5.7 ends in October 2023. Django 4.2 supports MySQL 8 and higher.
> Source: https://docs.djangoproject.com/en/5.0/releases/4.2/#dropped-support-for-mysql-5-7

Docker-compose.yml uses `mysql:5.7.44` right now.

Maffooch:
@kiblik I am glad you mentioned this in this PR :) this is a good opportunity to link to the discussion where this change is announced by @mtesauro 

https://github.com/DefectDojo/django-DefectDojo/discussions/9690

cneill:
Since our version of MySQL isn't supported by the version of Django we're moving to and it's our intention to deprecate MySQL with the next release, should we go ahead and remove it entirely as part of this PR?

mtesauro:
> Since our version of MySQL isn't supported by the version of Django we're moving to and it's our intention to deprecate MySQL with the next release, should we go ahead and remove it entirely as part of this PR?

I'd prefer to keep this PR only about the Django update and have a separate PR to remove MySQL from the various places.
This PR is complicated enough without adding to it's scope.

FWIW, I'll be working on a PR to remove MySQL & RabbitMQ from the compose files over the weekend.

cneill (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #9494, opened 2024-02-06
finding sla expiration date field (part two)

blakeaowens:


Maffooch (approved):

mtesauro (approved):
Approved

grendel513 (approved):

dojo/db_migrations/0201_populate_finding_sla_expiration_date.py
```diff
@@ -0,0 +1,129 @@
...
+from dojo.utils import get_work_days
+
+logger = logging.getLogger(__name__)
+
+
+def calculate_sla_expiration_dates(apps, schema_editor):
+    System_Settings = apps.get_model('dojo', 'System_Settings')
+
+    ss, _ = System_Settings.objects.get_or_create()
+    if ss.enable_finding_sla:
```
cneill, line 17: To reduce nesting here, you could simplify to this:

```suggestion
    if not ss.enable_finding_sla:
        return
```

dojo/db_migrations/0201_populate_finding_sla_expiration_date.py
```diff
@@ -0,0 +1,129 @@
...
+        Finding.objects.bulk_update(batch, ['sla_expiration_date'])
+        batch = []
+        logger.info('%s out of %s findings processed...', i, total_count)
+
+
+def reset_sla_expiration_dates(apps, schema_editor):
+    System_Settings = apps.get_model('dojo', 'System_Settings')
+
+    ss, _ = System_Settings.objects.get_or_create()
+    if ss.enable_finding_sla:
```
cneill, line 87: Same as above, this can be a no-op if not enabled to reduce nesting:
```suggestion
    if not ss.enable_finding_sla:
        return
```

dojo/models.py
```diff
@@ -3294,8 +3294,7 @@ def inherit_tags(self, potentially_existing_tags):
 
     @property
     def violates_sla(self):
-        days_remaining = self.sla_days_remaining()
-        return days_remaining < 0 if days_remaining else False
+        return (self.sla_expiration_date and self.sla_expiration_date > timezone.now())
```
cneill, line 3297: Looks like this should be reversed:
```suggestion
        return (self.sla_expiration_date and self.sla_expiration_date < timezone.now())
```

cneill (changes requested):
Just a couple notes

cneill (approved):

PR #9498, opened 2024-02-07
Remove filterwarnings for "unclosed file"

kiblik:
It looks like this was just some side-effect during other tests. It can be directly dropped.


mtesauro (approved):
Approved

Maffooch (approved):

grendel513 (approved):

Issue #9506, opened 2024-02-08
Deploying Defectdojo on AWS causes unhealthy ALB


lme-nca:
**Bug description**
When deploying on AWS the ALB healthcheck fails because it sets the "HTTP_HOST" header to the PODS IP address. The "HTTP_HOST" value of the ALB healthcheck can not be changed, this results in unhealthy loadbalancer status and the following error in the DJANGO logs:

`jango.core.exceptions.DisallowedHost: Invalid HTTP_HOST header: 'XX.XXX.XXX:8443'. You may need to add '10.134.200.129' to ALLOWED_HOSTS.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 165, in get_exception_response
    response = callback(request, exception=exception)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/views.py", line 29, in custom_bad_request_view
    return render(request, "400.html", {})
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/shortcuts.py", line 24, in render
    content = loader.render_to_string(template_name, context, request, using=using)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/template/loader.py", line 62, in render_to_string
    return template.render(context, request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/template/backends/django.py", line 61, in render
    return self.template.render(context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/template/base.py", line 173, in render
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 56, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/utils/deprecation.py", line 135, in __call__
    with context.bind_template(self):
  File "/usr/local/lib/python3.11/contextlib.py", line 137, in __enter__
[...]

PR #9508, opened 2024-02-08
added healthcheck middleware for ALBs

lme-nca:

**Description**

This fixes the following bug: https://github.com/DefectDojo/django-DefectDojo/issues/9506 

**Test results**

We are running this in our AWS deployment of AWS, if fixes the ALB unhealthy status and stops many ERROR log statements

**Documentation**

added commented out Annotation in HELM chart to indicate that his is available for AWS deployments

**Checklist**

**Extra information**

- enhancement
- performance
- feature
- bugfix
- maintenance (a.k.a chores)
- dependencies

# Contributors: Git Tips
## Rebase on dev branch
If the dev branch has changed since you started working on it, please rebase your work after the current dev.

On your working branch `mybranch`:
```
git rebase dev mybranch
```
In case of conflict:
```
 git mergetool
 git rebase --continue
 ```

When everything's fine on your local branch, force push to your `myOrigin` remote:
```
git push myOrigin --force-with-lease
```

To cancel everything:
```
git rebase --abort
```

## Squashing commits
```
git rebase -i origin/dev
```
- Replace `pick` by `fixup` on the commits you want squashed out
- Replace `pick` by `reword` on the first commit if you want to change the commit message
- Save the file and quit your editor

Force push to your `myOrigin` remote:
```
git push myOrigin --force-with-lease
```

### DIFF
--- a/dojo/middleware.py
+++ b/dojo/middleware.py
@@ -6,6 +6,7 @@ import logging
 from threading import local
 from django.db import models
 from django.urls import reverse
+from django.http import HttpResponse
 
 
 logger = logging.getLogger(__name__)
@@ -164,3 +165,21 @@ class AdditionalHeaderMiddleware:
     def __call__(self, request):
         request.META.update(settings.ADDITIONAL_HEADERS)
         return self.get_response(request)
+
+
+class HealthCheckMiddleware:
+    """
+    Middleware that will allow for a healthcheck to return UP without the caller being in the
+    DJANGO ALLOWED_HOSTS list. Needed for AWS ALB healthchecks and improves general k8 healthchecks
+    """
+
+    def __init__(self, get_response):
+
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if request.META['PATH_INFO'] == '/health':
+            return HttpResponse('UP!')
+        else:
+            response = self.get_response(request)
+            return response
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -860,6 +860,7 @@ INSTALLED_APPS = (
 # MIDDLEWARE
 # ------------------------------------------------------------------------------
 DJANGO_MIDDLEWARE_CLASSES = [
+    'dojo.middleware.HealthCheckMiddleware',
     'django.middleware.common.CommonMiddleware',
     'dojo.middleware.APITrailingSlashMiddleware',
     'dojo.middleware.DojoSytemSettingsMiddleware',
--- a/helm/defectdojo/values.yaml
+++ b/helm/defectdojo/values.yaml
@@ -201,6 +201,8 @@ django:
       # Depending on the size and complexity of your scans, you might want to increase the default ingress timeouts if you see repeated 504 Gateway Timeouts
       # nginx.ingress.kubernetes.io/proxy-read-timeout: "1800"
       # nginx.ingress.kubernetes.io/proxy-send-timeout: "1800"
+      # specific for AWS deployments Defectdojo has the /health endpoint for ALB healthchecks
+      # alb.ingress.kubernetes.io/healthcheck-path: /health
   nginx:
     tls:
       enabled: false
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -72,7 +72,13 @@ http {
       include /etc/nginx/wsgi_params;
       access_log off;
     }
-
+    # Used by AWS ALB health checks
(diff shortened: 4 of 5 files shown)


Maffooch:
Is there another place where this health check can live? Having it in the middleware is an extra check on every single request. I understand the reasoning is to eliminate modifying the allowed hosts lists

lme-nca:
> Is there another place where this health check can live? Having it in the middleware is an extra check on every single request. I understand the reasoning is to eliminate modifying the allowed hosts lists

I dont think so, part of the problem is that the "ALLOWED_HOST" list check is triggered in the middleware on every check, specifically by this call: https://github.com/django/django/blob/9c5e382b981608a26f2c55f1259d9e823fee5f15/django/middleware/common.py#L48 in the "common" middleware. 

Also see the original error I put in the Issue: 
`File "/usr/local/lib/python3.11/site-packages/django/utils/deprecation.py", line 135, in call
with context.bind_template(self):
File "/usr/local/lib/python3.11/contextlib.py", line 137, in enter
return next(self.gen)
^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/django/template/context.py", line 254, in bind_template
response = self.process_request(request)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
**File "/usr/local/lib/python3.11/site-packages/django/middleware/common.py", line 48, in process_request
updates.update(processor(self.request))
host = request.get_host()**
^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/django/http/request.py", line 152, in get_host
^^^^^^^^^^^^^^^^^^^^^^^
File "/app/dojo/context_processors.py", line 38, in bind_alert_count
if request.user.is_authenticated:
^^^^^^^^^^^^
raise DisallowedHost(msg)``

So @Maffooch unless you know another way to intercept this request before it reaches the Middleware, I think it kind of has to be fixed here.

mtesauro:
@lme-nca Really appreciate the PR but I agree with Cody here about the use of middleware.

I also have concerns about this being a vendor specific change that negatively impacts any one who doesn't happen to use AWS by having a potentially unnecessary middleware check on every request.

We've been working to ensure that all features of DD can be tested by our GH Actions and having vendor specific changes means to truly test that, we need an AWS account, launch infra, etc. This is why we stopped taking new API integrations - they can't be tested and we want the DefectDojo that the community uses to be as rock solid as possible.

With that said, we're not going to accept this PR for the reasons outlined above.

lme-nca:
Hi @mtesauro, @Maffooch, I cant say im not a little disappointed by that decision, please take into consideration the following points:

 I don't think this is a "vendor" specific change.... This change merely introduces a "HEALTH" endpoint that does not get blocked by the "ALLOWED_HOST" list. Currently most liveness checks in the Kuberenets deployment use the login url and synthetically set the correct host header (see https://github.com/DefectDojo/django-DefectDojo/blob/738dca4534382049ca7587a60fb6b33715213943/helm/defectdojo/templates/django-deployment.yaml#L215 ). This is even though these healthchecks don't call these endpoints using this hostname at all, but rather the internal ip. Imho they should/could also be using this health endpoint. I would therefore propose to use this /health endpoint for all the liveness checks, that way this becomes a generic change to introduce a /health endpoint that is not bothered by "ALLOWED_HOST" checks...

I also wonder about the "performance" impact of this change, Middleware already runs on every request so i dont really understand how 1 IF statement could have a big impact on performance in this regard.

As for testing this can easily be tested outside of AWS, just call the health endpoint without the correct host set and see that you receive a valid response...

Lastly, I admit this isn't the prettiest change, but even Django themselves fixed this problem by introducing Middleware, see: https://github.com/mozmeao/infra/issues/735 and https://pypi.org/project/django-allow-cidr/ . The only difference is that they whitelist a complete range for all endpoints........ So if you reject my solution I would like to ask you to come up with an alternative that can solve this problem ? I would then happily try and implement this.

HackityAnn:
Have to say that I agree with @lme-nca , his proposed solution would be the Django way for this (and the way that I will hack this in on our own DD instance). Rock solid also means being able to handle health checks in a good way, middleware gives us that solution.

kiblik:
Is it possible to reconsider this PR? To be honest, I suppose the only "vendor-specific" part in this PR is the title and comments. Publication of a super simple health check does not expose any sensitive information and this endpoint might be valuable also for other environments that can benefit from it (not only ALB).

PR #9513, opened 2024-02-09
Jira Webhook: Catch comments from other issue updates

Maffooch:
In a situation where a screen requiring a comment during an issue transition, and comment will will be sent under the `issue_updated` event. Some times an additional event for `comment_added` would occur, but this was not a universal case. To achieve this adopted behavior, the following actions were taken:
- Move the original `comment_added` operations to a function
- Add an additional query for identical notes created in the last 30 seconds to remove the possibility of duplicated notes in the event both webhook events are captured
- Updated to safer dict access methods to reduce chances of exceptions

[sc-4202]

### DIFF
--- a/dojo/jira_link/views.py
+++ b/dojo/jira_link/views.py
@@ -106,10 +106,12 @@ def webhook(request, secret=None):
                     for finding in findings:
                         jira_helper.process_resolution_from_jira(finding, resolution_id, resolution_name, assignee_name, jira_now, jissue)
                 # Check for any comment that could have come along with the resolution
-                check_for_and_create_comment(parsed)
+                if (error_response := check_for_and_create_comment(parsed)) is not None:
+                    return error_response
 
             if parsed.get('webhookEvent') == 'comment_created':
-                check_for_and_create_comment(parsed)
+                if (error_response := check_for_and_create_comment(parsed)) is not None:
+                    return error_response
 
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

dojo/jira_link/views.py
```diff
@@ -218,6 +135,109 @@ def webhook(request, secret=None):
...
+                "displayName":"Defect Dojo",
+                "active":true,
+                "timeZone":"Europe/Amsterdam"
+            },
+            "created":"2021-01-10T10:12:47.824+0100",
+            "updated":"2021-01-10T10:12:47.824+0100"
+        }
+        }
+    """
+    if (comment := parsed_json.get("comment", None)) is not None:
```
cneill, line 181: Flipping this logic would let you eliminate a level of nesting throughout this block

```suggestion
    comment = parsed_json.get("comment", None)
    if comment is None:
        return
    ....
```

cneill (commented):
Just a minor note, otherwise looks good

dojo/jira_link/views.py
Maffooch: Rip I didn't think about the rest of the function needed to be unindented now

cneill (approved):

devGregA (approved):

grendel513 (approved):

PR #9517, opened 2024-02-09
Modifying Bugcrowd API Parser to align to vendor documentation on wha…

grendel513:
Modifying Bugcrowd API Parser to align to vendor documentation on what the not_applicable state means.  It is now active == False and severity == 'Info'. 

[sc-4217]

PR #9519, opened 2024-02-09
improved Sonatype parser

reichertan:
**Description**

The Sonatype parser has been improved. More attributes are passed. If a component previously had several vulnerabilities, only the first vulnerability was created in DefectDojo and a reference was made to the rest. Now all vulnerabilities reported by Sonatype are also transferred to DefectDojo as vulnerabilities.


dojo/tools/sonatype/parser.py
```diff
@@ -1,154 +1,95 @@
...
+        vulnerability_id = security_issue.get("reference")
+        finding.unsaved_vulnerability_ids = [vulnerability_id]
+
+    if status == "Open":
+        pass
+    elif status == "Acknowledged":
+        finding.verified = True
+    elif status == "Not Applicable":
+        finding.verified = True
+        finding.out_of_scope = True
```
cneill, line 80: I'm not too familiar with Sonatype, but couldn't this also arguably be a `false_p` based on how the user manages things in Sonatype? We're starting to lean away from automatically marking Findings `out_of_scope` / `false_p` / `risk_accepted` at the parser level if it's ambiguous, since it can vary based on user assumptions. For example, [with a similar case in the BugCrowd parser](https://github.com/DefectDojo/django-DefectDojo/pull/9517/files#diff-78adb39511a2d994c7c200bb18f4e89fd0ca0769f94e41410edbcfe156803e30R148), we opted to mark those findings as inactive/"Info" severity.

reichertan: Yes you are right. I was unsure about that too. This is more our business logic and would probably be better off in our own code than in the official DefectDojo repo. I remove this block and the associated test.

cneill (approved):

mtesauro (approved):
Approved

grendel513 (approved):

Maffooch (approved):

Issue #7966, opened 2023-04-13
KICS partial import


Acsigen:
**Bug description**
When importing a KICS json report, not all vulnerabilities are displayed in DefectDojo when compared to html version of KICS.

**Steps to reproduce**
1. Perform KICS scan with output as JSON and HTML
2. Import the report to a DefectDojo project
3. Compare the number of findings between DefectDojo and KICS HTML

**Expected behavior**
The number of vulnerabilities should be the same

**Deployment method** *(select with an `X`)*
- [x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Ubuntu 22.04
 - DefectDojo version: 2.21.0

**Screenshots**
KICS HTML report:
![image](https://user-images.githubusercontent.com/13287878/231697013-da624512-fab8-4f26-9281-4e1d87701db7.png)

DefectDojo Vulnerabilities count after importing the KICS report:
![image](https://user-images.githubusercontent.com/13287878/231697274-06a94e94-921c-42ca-a2c2-3606a4fa3046.png)

**Workaround**:
Set the output format of KICS to Gitlab SAST and then import it in DefectDojo as Gitlab SAST.

security101:
Yes, I have the same observation. 22 distinct findings within the json file but only 7 imported / presented running DD v2.24.1 (docker-compose)

manuel-sommer:
Hey @security101 and @Acsigen ,

could you please provide a sample report?

Acsigen:
> Hey @security101 and @Acsigen ,
> 
> could you please provide a sample report?

Sadly I don't have a report now. I will upload one as soon as I have a new project comming my way.

manuel-sommer:
Is this the right link @Acsigen ?
https://github.com/Checkmarx/kics

manuel-sommer:
@quirinziessler maybe kics is interesting for you?

manuel-sommer:
could you close this PR please @mtesauro

PR #9542, opened 2024-02-14
:bug: fix kics, #7966

manuel-sommer:
See issue #7966 

--> I noticed that a couple of findings which are almost the same are deduplicated because of dupe_key being to unspecific

compare: https://github.com/DefectDojo/django-DefectDojo/blob/d698a7a1ff2a914754a541140b236dd3092d7e8e/unittests/scans/kics/many_findings.json#L303 with: https://github.com/DefectDojo/django-DefectDojo/blob/d698a7a1ff2a914754a541140b236dd3092d7e8e/unittests/scans/kics/many_findings.json#L314

Now, the number of findings is also correct:
https://github.com/DefectDojo/django-DefectDojo/blob/d698a7a1ff2a914754a541140b236dd3092d7e8e/unittests/scans/kics/many_findings.json#L380

### DIFF
--- a/dojo/tools/kics/parser.py
+++ b/dojo/tools/kics/parser.py
@@ -62,6 +62,7 @@ class KICSParser(object):
                         + category
                         + issue_type
                         + file_name
+                        + expected_value
                         + str(line_number)
                     ).encode("utf-8")
                 ).hexdigest()


dojo/tools/kics/parser.py
```diff
@@ -62,6 +62,7 @@ def get_findings(self, filename, test):
                         + category
                         + issue_type
                         + file_name
+                        + expected_value
```
quirinziessler, line 65: the "similarity_id" seems to be unique and may suit better than the expected value, which is not.

I would suggest to add `similarity_id = item.get("similarity_id")` below the "acutal_value" and replace `expected_value` by `similarity_id` in the dupe_key

quirinziessler (changes requested):

dojo/tools/kics/parser.py
manuel-sommer: I tested similarity_id and it will revert this fix again as in the above mentioned case the similarity_id is the same

grafik

manuel-sommer: @quirinziessler do you have a unittestfile to help in this case? 
You can upload it here: https://github.com/DefectDojo/django-DefectDojo/issues/7966

quirinziessler: @manuel-sommer sorry for this, probably it was only unique for my output file. I noticed one finding being surpressed by your idea but it seems it is just a issue with the deduplication. Please keep the expected_value and adjust the hash_code in settings.dist.py. This together should solve the whole issue.
I would recommend to filter by "file_path, line, severity, description, title".

mtesauro (approved):
Approved

grendel513 (approved):

grendel513 (approved):

cneill (approved):

Maffooch (approved):

PR #9544, opened 2024-02-14
Fix engagement/view if import fail

kiblik:
If `import_scan` fails in the following block, `test` is not set. The user should be redirected to where he/she/they came from.

https://github.com/DefectDojo/django-DefectDojo/blob/d698a7a1ff2a914754a541140b236dd3092d7e8e/dojo/engagement/views.py#L732-L752

I suppose the user should be redirected to the mentioned place with the prefilled form (to avoid the necessity to refill it again). I need to know how to do it here.

### DIFF
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -768,7 +768,7 @@ class ImportScanResultsView(View):
                 return HttpResponseRedirect(
                     reverse('view_test', args=(test.id, )))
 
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

cneill (approved):

grendel513 (approved):

PR #9558, opened 2024-02-15
Deduplication: Do not reopen original finding

Maffooch:
When using deduplication, I do not expect old findings to be reopened. This can create some very interesting SLA violations. For example, if a Finding is closed one year ago, but is rediscovered today as a regression, then the timeline of the original finding, and each duplicate in the chain does not make sense. It makes more sense to create a new finding, and start another duplicate chain

[sc-4263]


mtesauro (approved):
Approved

cneill (approved):

devGregA (approved):

PR #9560, opened 2024-02-15
Correct Endpoint "Hosts" views when the host field is `None`

Maffooch:
When visiting the "All Hosts" view, I was encountered with a 500 error page for the following exception:
```
  File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 56, in inner
               ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/sentry_sdk/integrations/django/views.py", line 84, in sentry_wrapped_callback
    return callback(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/endpoint/views.py", line 109, in all_endpoint_hosts
    return process_endpoints_view(request, host_view=True, vulnerable=False)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/endpoint/views.py", line 53, in process_endpoints_view
    ids = get_endpoint_ids(EndpointFilter(request.GET, queryset=endpoints, user=request.user).qs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/endpoint/views.py", line 95, in get_endpoint_ids
    key = e.host + '-' + str(e.product.id)
          ~~~~~~~^~~~~
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```
I am still not sure how there is an endpoint without a host as there are multiple places where the `clean` function on an endpoint is ran to ensure there is a host. I learned that [on model saves, the `clean` function is actually not executed](https://docs.djangoproject.com/en/5.0/ref/models/instances/#django.db.models.Model.clean). To ensure the database does not get into an unrecoverable state again, the `clean` function is now explicitly called when saving an endpoint. 

[sc-4260]

### DIFF
--- a/dojo/endpoint/views.py
+++ b/dojo/endpoint/views.py
@@ -92,7 +92,7 @@ def get_endpoint_ids(endpoints):
     hosts = []
     ids = []
     for e in endpoints:
-        key = e.host + '-' + str(e.product.id)
+        key = f"{e.host}-{e.product.id}"
         if key in hosts:
             continue
         else:
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1601,6 +1601,13 @@ class Endpoint(models.Model):
             models.Index(fields=['product']),
         ]
 
+    def save(self, *args, **kwargs):
+        # clean() is not called when an object is saved, so we should do it here to prevent
+        # the endpoint from getting into an inrecoverable state
+        # https://docs.djangoproject.com/en/5.0/ref/models/instances/#django.db.models.Model.clean
+        self.clean()
+        super(Endpoint, self).save(*args, **kwargs)
+
     def clean(self):
         errors = []
         null_char_list = ["0x00", "\x00"]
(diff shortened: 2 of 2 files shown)


Maffooch:
After working through some unit test failures, it dawned on me why it is possible to save endpoints even after they fail validation. The reason is that it is better to save a broken endpoint that can be corrected vs throwing the entire thing out. This is indicated by a little red flag on a broken endpoint:
image
The host in this example is totally busted, but as a user, I would rather know that it is busted and be able to be correct it rather than never seeing it in dojo to begin with

cneill (approved):

grendel513 (approved):

Issue #9562, opened 2024-02-16
uWSGI OOMKilled on Kubernetes


hoeg:
**Bug description**
Deploying Defect Dojo to a Kubernetes cluster causes the uWSGI container to consume a lot of memory resulting in the node killing the pod. This is due to the unbound number of file descriptors on the node. See https://github.com/unbit/uwsgi/issues/2299 for a description of the issue with uWSGI.

**Steps to reproduce**
- Deploy helm chart to a kubernetes cluster with nodes running Flatcar Container Linux by Kinvolk 3602.2.1 (Oklo)
- watch the pod get deployed and after <15 sec killed by the node due to OOM.

**Expected behavior**
Expected the pod to start up and not get OOMKilled by the node.

I locally build my own container adding the `--max-fd` argument to `docker/entrypoint-uwsgi.sh` and used that image in the my cluster, this resolved the issue.

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
 - Kubernetes nodes running: 
 ```
Kernel Version:             5.15.136-flatcar
OS Image:                   Flatcar Container Linux by Kinvolk 3602.2.1 (Oklo)
Operating System:           linux
Architecture:               amd64
Container Runtime Version:  containerd://1.6.21
Kubelet Version:            v1.28.3
Kube-Proxy Version:         v1.28.3
```
 - DefectDojo version: 2.30.4

**Logs** 
Logs from the defectdojo-django pod
```
$ k logs defect-dojo-defectdojo-django
Defaulted container "uwsgi" out of: uwsgi, nginx
[13/Feb/2024 08:50:57] INFO [dojo.models:4295] enabling audit logging
/usr/local/lib/python3.11/site-packages/coreapi/codecs/download.py:5: DeprecationWarning: 'cgi' is deprecated and slated for removal in Python 3.13
  import cgi
System check identified no issues (0 silenced).
*** Starting uWSGI 2.0.23 (64bit) on [Tue Feb 13 08:50:58 2024] ***
compiled with version: 10.2.1 20210110 on 29 January 2024 15:50:06
os: Linux-5.15.136-flatcar #1 SMP Mon Oct 23 16:44:45 -00 2023
nodename: defect-dojo-defectdojo-django
machine: x86_64
clock source: unix
detected number of CPU cores: 4
current working directory: /app
detected binary path: /usr/local/bin/uwsgi
[...]

PR #9564, opened 2024-02-16
Add --max-fd argument to uwsgi to stop it from getting OOMKilled in Kubernetes

hoeg:
**Description**

This PR fixes the issue described in issue https://github.com/DefectDojo/django-DefectDojo/issues/9562 regarding uWSGI that under some circumstances will take up an unnecessary amount of resources on a kubernetes node leading to the pod getting OOMKilled.

We introduce the possibility to set the `--max-fd` argument when starting up uWSGI to mitigate this issue. 

**Test results**

I have tested the fix on a kubernetes cluster where it prevented the pod from getting OOMKilled. For more information see https://github.com/DefectDojo/django-DefectDojo/issues/9562.

**Documentation**

It is not clear to me where the documentation should be updated.

### DIFF
--- a/docker/entrypoint-uwsgi.sh
+++ b/docker/entrypoint-uwsgi.sh
@@ -32,5 +32,5 @@ exec uwsgi \
   --buffer-size="${DD_UWSGI_BUFFER_SIZE:-8192}" \
   --http 0.0.0.0:8081 --http-to "${DD_UWSGI_ENDPOINT}" \
   --logformat "${DD_UWSGI_LOGFORMAT:-$DD_UWSGI_LOGFORMAT_DEFAULT}" \
-  --max-fd "${DD_UWSGI_MAX_FD:-1048576}
+  --max-fd "${DD_UWSGI_MAX_FD:-1048576}"
   # HTTP endpoint is enabled for Kubernetes liveness checks. It should not be exposed as a service.
--- a/docker/entrypoint-uwsgi.sh
+++ b/docker/entrypoint-uwsgi.sh
@@ -22,15 +22,31 @@ python3 manage.py check
 
 DD_UWSGI_LOGFORMAT_DEFAULT='[pid: %(pid)|app: -|req: -/-] %(addr) (%(dd_user)) {%(vars) vars in %(pktsize) bytes} [%(ctime)] %(method) %(uri) => generated %(rsize) bytes in %(msecs) msecs (%(proto) %(status)) %(headers) headers in %(hsize) bytes (%(switches) switches on core %(core))'
 
-exec uwsgi \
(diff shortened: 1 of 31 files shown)


dsever:
Please create PR against dev  or bugfix

mtesauro (approved):
Approved

helm/defectdojo/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "2.32.0-dev"
 description: A Helm chart for Kubernetes to install DefectDojo
 name: defectdojo
-version: 1.6.110-dev
+version: 1.6.111
```
Maffooch, line 5: Please do not update this manually. Release automation will cover this action

docker/entrypoint-uwsgi.sh
```diff
@@ -31,5 +31,6 @@ exec uwsgi \
   --wsgi dojo.wsgi:application \
   --buffer-size="${DD_UWSGI_BUFFER_SIZE:-8192}" \
   --http 0.0.0.0:8081 --http-to "${DD_UWSGI_ENDPOINT}" \
-  --logformat "${DD_UWSGI_LOGFORMAT:-$DD_UWSGI_LOGFORMAT_DEFAULT}"
+  --logformat "${DD_UWSGI_LOGFORMAT:-$DD_UWSGI_LOGFORMAT_DEFAULT}" \
+  --max-fd "${DD_UWSGI_MAX_FD:-1048576}"
```
Maffooch, line 35: Is this the default for uswgi? I could not find it online

Maffooch (changes requested):

docker/entrypoint-uwsgi.sh
hoeg: It is what the docker installation defaulted to. Would you rather that we omit the argument instead of forcing a default?

hoeg: I will remove the argument when the `DD_UWSGI_MAX_FD` is not set.

Maffooch:
@hoeg thanks for updating your commits. It think this will work! There is not a some extra changes unrelated to your file descriptor changes that should not be here. Once those are removed, I think this will be good to go :)

mtesauro:
Closing this PR as the parent issue was addressed by changing the k8s config rather then a code change on the DD side.

tmablunar:
I am trying to pick this is this issue up from my colleague @hoeg.

The described solution in k8s does not resolve the issue for us. We do not have any resource limits and requesting 4096Mi of memory. The pod is still OOMKilled. Using the suggested solution here however solves the problem.

PR #9568, opened 2024-02-16
Deprecate merge_sets_safe in dojo/utils.py

manuel-sommer:
This PR uses python functionality which was added in newer python versions (3.6+) and thus deprecates method merge_sets_safe in utils.py. Furthermore, itertools is not needed anymore in utils.py. Thus, we don't need to import itertools.

### DIFF
--- a/unittests/test_notifications.py
+++ b/unittests/test_notifications.py
@@ -54,3 +54,4 @@ class TestNotifications(DojoTestCase):
         self.assertEqual('mail' in merged_notifications.other, True)
         self.assertEqual('slack' in merged_notifications.other, True)  # default alert from global
         self.assertEqual(len(merged_notifications.other), 3)
+        self.assertEqual(merged_notifications.other, {'alert', 'mail', 'slack'})
--- a/unittests/test_notifications.py
+++ b/unittests/test_notifications.py
@@ -53,4 +53,4 @@ class TestNotifications(DojoTestCase):
         self.assertEqual('alert' in merged_notifications.other, True)
         self.assertEqual('mail' in merged_notifications.other, True)
         self.assertEqual('slack' in merged_notifications.other, True)  # default alert from global
-        self.assertEqual(len(merged_notifications.other), 4)
+        self.assertEqual(len(merged_notifications.other), 3)
--- a/unittests/test_notifications.py
+++ b/unittests/test_notifications.py
@@ -53,4 +53,4 @@ class TestNotifications(DojoTestCase):
         self.assertEqual('alert' in merged_notifications.other, True)
         self.assertEqual('mail' in merged_notifications.other, True)
         self.assertEqual('slack' in merged_notifications.other, True)  # default alert from global
-        self.assertEqual(len(merged_notifications.other), 3)
+        self.assertEqual(len(merged_notifications.other), 4)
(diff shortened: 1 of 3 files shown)


manuel-sommer:
I have no clue why this unittest fails.

manuel-sommer:
Why is Kubernetes Deployment failing?

mtesauro (approved):
Approved

unittests/test_notifications.py
```diff
@@ -53,4 +53,4 @@ def test_merge_notifications_list(self):
         self.assertEqual('alert' in merged_notifications.other, True)
         self.assertEqual('mail' in merged_notifications.other, True)
         self.assertEqual('slack' in merged_notifications.other, True)  # default alert from global
-        self.assertEqual(len(merged_notifications.other), 3)
+        self.assertEqual(len(merged_notifications.other), 4)
```
Maffooch, line 56: I would think functionality would not change if doing a simple todo. Why does this value increase?

Maffooch (changes requested):

unittests/test_notifications.py
manuel-sommer: This is also what I wrote in the comment above: _I have no clue why this unittest fails._

manuel-sommer: merged_notifications.other is now ['slack', 'mail', 'alert', 'alert']
before that it was: ['slack', 'mail', 'alert']

manuel-sommer: It is because of "default alert of global", see line 55 @Maffooch

manuel-sommer:
Before the change: merged_notifications.other = {'alert', 'mail', 'slack'}
After the change: merged_notifications.other = ['slack', 'mail', 'alert', 'alert']

kiblik:
> Before the change: merged_notifications.other = {'alert', 'mail', 'slack'}
> After the change: merged_notifications.other = ['slack', 'mail', 'alert', 'alert']
> I don't know exactly why alert is now duplicated.

In the first case, it is a set (no duplicates possible). In the second case, it is a list (duplicates allowed, order is important).
`[*..., *...]` concatnate 2 lists.

I suppose, we should use `{*..., *...}` in stread of `[*..., *...]`

```python
>>> a={'a','b'}
>>> b={'b','c'}
>>> c={*a, *b}
>>> c
{'a', 'b', 'c'}
>>> d=[*a, *b]
>>> d
['a', 'b', 'b', 'c']
```

manuel-sommer:
@Maffooch, through the comment of @kiblik your requested change is resolved. 
Thank you @kiblik for the hint.

mtesauro:
@manuel-sommer Got some feedback on this from the community/contributors:

Can you:
1. Add a description to the PR - why/what/etc
2. comment in the code to state what it is doing and why

Thanks!

manuel-sommer:
@mtesauro, I added a description to the PR.

Furthermore, I have no clue why unittests now fails, it is not PR related, but looks like a false positive.

manuel-sommer:
another closure to retrigger the pipeline, maybe this helps.

manuel-sommer:
Reopen to retrigger pipeline

manuel-sommer:
@Maffooch could you please review this PR?

Maffooch (approved):

cneill (approved):

grendel513 (approved):

PR #9570, opened 2024-02-16
Bugfix: checkmarx parser - datetime is no longer put into the Finding.date field

reichertan:
**Description**
We could no longer import Checkmarx reports because the finding date field had a datetime value. This fix resolves the issue.

checkmarx_import_error

### DIFF
--- a/unittests/tools/test_checkmarx_parser.py
+++ b/unittests/tools/test_checkmarx_parser.py
@@ -7,23 +7,21 @@ from dojo.tools.checkmarx.parser import CheckmarxParser
 from ..dojo_test_case import DojoTestCase, get_unit_tests_path
 
 
-def init(report_filename):
-    my_file_handle = open(report_filename)
-    product = Product()
-    engagement = Engagement()
-    test = Test()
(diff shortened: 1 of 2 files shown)


unittests/tools/test_checkmarx_parser.py
```diff
@@ -7,56 +7,58 @@
...
+    test = Test()
+    engagement.product = product
+    test.engagement = engagement
+    return my_file_handle, product, engagement, test
+
+
+def teardown(my_file_handle):
+    my_file_handle.close()
+
+
```
Maffooch, line 23: What is the benefit of moving these outside of the class?

reichertan: My IDE suggested converting the methods to functions and I accepted it without thinking twice. I have now revoked the change.

Maffooch (approved):

mtesauro (approved):
Approved

cneill (approved):

grendel513 (approved):

PR #9571, opened 2024-02-16
Jira: Append labels and respect priority on update

Maffooch:
A couple fields are overwritten by DefectDojo when findings are pushed to an existing jira ticket. This can be destructive for developers in the following ways:
- Priority: This field often reflects the timeline a particular issue may be fixed. Developers may have more specific context for why a vulnerability may not be as severe as initially thought.
- Labels: Labels could be used to sort issues in a given queue to determine who works on a given ticket. When a finding is pushed to jira again after creation, these new labels should not be overwritten

These fields should be respected to avoid stomping on any changes/process set by developers

[sc-4264]

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -863,9 +863,10 @@ def update_jira_issue(obj, *args, **kwargs):
             summary=jira_summary(obj),
             description=jira_description(obj),
             component_name=jira_project.component if not issue.fields.components else None,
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

cneill (approved):

grendel513 (approved):

devGregA (approved):

Issue #9584, opened 2024-02-20
fortify fpr instead of xml


schdief:
**Scanner Name**
currently defect dojo expects a xml report for fortify "Fortify Scan - Import Findings from XML file format", the default report format is the fpr file, this should be used instead

**Sample File**
fpr, had to zip it to upload it
[fortify_sast_front_end_20240220_629_9_1361_6018341.zip](https://github.com/DefectDojo/django-DefectDojo/files/14341092/fortify_sast_front_end_20240220_629_9_1361_6018341.zip)

manuel-sommer:
Hi @schdief, I guess fpr format is the wrong format to be able to parse within DefectDojo as it looks to me that fpr file formats can't be parsed in a decent way. If I open your pfr file in vscode, I can't view the content.

What is your opinion @mtesauro ? Can we close this?

schdief:
@manuel-sommer the fpr is a zip itself, you can just unzip it and then parse it

<img width="229" alt="image" src="https://github.com/DefectDojo/django-DefectDojo/assets/10979159/aba6e4c1-8c90-4e2a-92e8-90e5e1f8329e">

we currently parse it ourselves like this with kotlin:

```
 fun parseFortifyArchive(inputStream: InputStream): Pair<FortifyTestResult, List<String>> {
            var report: FortifyTestResult? = null
            val suppressions: MutableList<String> = mutableListOf()
            ZipInputStream(inputStream).use { zipInputStream ->
                generateSequence { zipInputStream.nextEntry }
                    .filter { entry -> entry.name.matches(Regex(".*filters/.*txt|audit.fvdl")) }
                    .mapNotNull { entry ->
                        when {
                            entry.name.endsWith("fvdl") -> parseFortifyXML(
                                zipInputStream.bufferedReader().readText()
                            )

                            entry.name.endsWith(".txt") -> zipInputStream.bufferedReader().readText()
                                .split("\n")
                                .filter { (!it.startsWith("#")) && it.isNotEmpty() }

                            else -> null
                        }
                    }.forEach {
                        @Suppress("UNCHECKED_CAST") //Only ever adds strings
                        when (it) {
                            is FortifyTestResult -> report = it
                            is List<*> -> suppressions.addAll(it as List<String>)
                        }
                    }
            }
            report?.let {
                return it to suppressions.toList()
            }
            throw NoSuchElementException("Couldn't parse audit.fvdl.")
        }
```

[...]

PR #9590, opened 2024-02-20
:sparkles: Fortify: Support .fpr format

manuel-sommer:
see issue #9584


mtesauro (approved):
Approved

docs/content/en/integrations/parsers/file/fortify.md
```diff
@@ -2,7 +2,8 @@
 title: "Fortify"
 toc_hide: true
 ---
-Import Findings from XML file format.
+You can either import the findings in .xml or in .fdr file format. </br>
```
schdief, line 5:
```suggestion
You can either import the findings in .xml or in .fpr file format. </br>
```

docs/content/en/integrations/parsers/file/fortify.md
```diff
@@ -2,7 +2,8 @@
...
+If you import a .fdr file, the parser will look for the file 'audit.fvdl' and analyze it. An extracted example can be found [here](https://github.com/DefectDojo/django-DefectDojo/tree/master/unittests/scans/fortify/audit.fvdl).
```
schdief, line 6:
```suggestion
If you import a .fpr file, the parser will look for the file 'audit.fvdl' and analyze it. An extracted example can be found [here](https://github.com/DefectDojo/django-DefectDojo/tree/master/unittests/scans/fortify/audit.fvdl).
```

dojo/tools/fortify/parser.py
```diff
@@ -18,6 +18,12 @@ def get_description_for_scan_types(self, scan_type):
         return "Import Findings from XML file format."
```
schdief, line 18: shouldn't this be adapted as well?

dojo/tools/fortify/parser.py
```diff
@@ -114,6 +120,51 @@ def get_findings(self, filename, test):
...
+                    Confidence = vuln.find(f"{namespace}InstanceInfo").find(f"{namespace}Confidence").text
+                    description = Type + "\n"
+                    description += "**ClassID:** " + ClassID + "\n"
+                    description += "**Kingdom:** " + Kingdom + "\n"
+                    description += "**AnalyzerName:** " + AnalyzerName + "\n"
+                    description += "**DefaultSeverity:** " + DefaultSeverity + "\n"
+                    description += "**InstanceID:** " + InstanceID + "\n"
+                    description += "**InstanceSeverity:** " + InstanceSeverity + "\n"
+                    description += "**Confidence:** " + Confidence + "\n"
+                    items.append(
```
schdief, line 156: I think we should also import:
- SourceLocation path="public/category.html" as "File path" in GUI
- line="222" as "Line number" in GUI

dojo/tools/fortify/parser.py
```diff
@@ -114,6 +120,51 @@ def get_findings(self, filename, test):
...
+                        Finding(
+                            title=Type + " " + ClassID,
+                            severity="High",
+                            static_finding=True,
+                            test=test,
+                            description=description,
+                            unique_id_from_tool=ClassID,
+                        )
+                    )
+        return items
```
schdief, line 166: can we also populate the test entry with data from the report like:
- Version>front_end_20240220_629_9_1361_6018341</Version>
- CreatedTS date="2024-02-20" time="05:34:11"/>
- end = created + ScanTime value="215"/>
- description based on:

```
  <NumberFiles>58</NumberFiles>
  <LOC type="Line Count">2942</LOC>
  <SourceBasePath>/var/lib/jenkins/workspace/SDM_SockShop_front-end_master/src</SourceBasePath>
  <SourceFiles> as list of files
```

unittests/tools/test_fortify_parser.py
```diff
@@ -75,3 +74,14 @@ def test_fortify_issue6082(self):
...
+    def test_fortify_many_fdr_findings(self):
+        testfile = open("unittests/scans/fortify/many_findings.fpr")
+        parser = FortifyParser()
+        findings = parser.get_findings(testfile, Test())
+        self.assertEqual(61, len(findings))
+        with self.subTest(i=0):
+            finding = findings[0]
+            self.assertEqual("Cross-Site Request Forgery 114E5A67-3446-4DD5-B578-D0E6FDBB304E", finding.title)
+            self.assertEqual("High", finding.severity)
+            self.assertEqual('114E5A67-3446-4DD5-B578-D0E6FDBB304E', finding.unique_id_from_tool)
```
schdief, line 87: this is actually the class id and not the finding id

schdief (changes requested):
great work, happy to see such a fast reaction time, added some comments for improvement

unittests/tools/test_fortify_parser.py
manuel-sommer: This check is right I would say.

dojo/tools/fortify/parser.py
manuel-sommer: I don't know if we can populate this info to the test itself. Maybe I don't understand exactly what you are talking about @schdief .

manuel-sommer:
Please review again @schdief

dojo/tools/fortify/parser.py
schdief: when you import a new scan result this is related to a test in defect dojo, currently you only write data for the findings

https://URL/finding/39

but it belongs to a 

https://URL/test/7

and this test has metadata itself

image

here we can add more data based on the fpr like version, start and end and within the description of the test also the number of files, loc and the list of sourcefiles from which sourcebaspath

schdief (approved):
lgtm, thanks again
as written, maybe we can enrich the test metadata too, to give more context about the scan, which is especially imprtant if there are no findings at all, to verify that the scan itself was valid (not just empty files scanned etc.)

what is the release cadence for defectdojo? when can we use this after merging?

dojo/tools/fortify/parser.py
manuel-sommer: Yes, please look at the parser, I can't affect the test, only the findings. The test is affected by your API calls.

schdief: I can see that the current implementation cannot update the test, but the GUI uses this form
https://URL/engagement/3/import_scan_results

to import scan results, e. g. of fortify, and at the same time creates a test container for it, maybe this should be another ticket/pr, but if the info doesn't get parsed, it won't be able to be stored later on

manuel-sommer:
> lgtm, thanks again as written, maybe we can enrich the test metadata too, to give more context about the scan, which is especially imprtant if there are no findings at all, to verify that the scan itself was valid (not just empty files scanned etc.)
> 
> what is the release cadence for defectdojo? when can we use this after merging?

https://documentation.defectdojo.com/contributing/branching-model/

devGregA (approved):

Maffooch (approved):

PR #9593, opened 2024-02-20
Extend npm_audit.md with unsupported v7 notice

paulOsinski:
As NPM_Audit files are not supported past v6, adding this note to documentation.  Currently NPM_Audit scans past v6 omit many relevant data fields from the JSON export, which creates 'stub Findings' which need to be updated through a different process.

Staging a docs update until we decide what to do: whether to update the NPM-audit parser to create those stub Findings or to leave things the way they are.

### DIFF
--- a/docs/content/en/integrations/parsers/file/npm_audit.md
+++ b/docs/content/en/integrations/parsers/file/npm_audit.md
@@ -2,8 +2,29 @@
 title: "NPM Audit"
 toc_hide: true
 ---
+
+**Note: This parser only supports import from NPM Audit v6 or older.**
+
 Node Package Manager (NPM) Audit plugin output file can be imported in
 JSON format. Only imports the \'advisories\' subtree.
 
+### File Types
+This parser expects a JSON file.  Can only import NPM Audit files from NPM Audit v6 or older due to missing relevant fields, including:
+
+- Finding created / updated dates
+- Relevant CVE number
+- Finding overview (description Field)
+- Recommendation
+- Issue reference
+- CWE
+- Exploitability
+
+See NPM's issue on GitHub for more information.  https://github.com/npm/npm-audit-report/issues/45
+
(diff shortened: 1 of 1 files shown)


grendel513 (approved):

mtesauro (approved):
Approved

docs/content/en/integrations/parsers/file/npm_audit.md
```diff
@@ -2,8 +2,29 @@
 title: "NPM Audit"
 toc_hide: true
 ---
+
+**Note: This parser only supports import from NPM Audit v6 or older.**
```
manuel-sommer, line 6: https://github.com/DefectDojo/django-DefectDojo/issues/4285

manuel-sommer: Could you rather resolve this issue?

manuel-sommer (changes requested):

docs/content/en/integrations/parsers/file/npm_audit.md
manuel-sommer: In case you have a v7 file, you can attach it to the linked issue and then I can help you advancing this parser.

mtesauro: @manuel-sommer We've got a follow-on PR which will add a parser for v7 - should drop in the next day or two.

The data provided by npm_audit changed changed pretty significantly in v7 (to less useful data) so we're going to keep the parsers separate to:
* allow people to choose which version they want to run
* understand why there's a significant difference in output. If these were combined, you could have one parser producing 2 very different looking findings if results from v7 and earlier were uploaded
* a separate parser avoids a migration (a bonus given the above reasons)

Thanks for asking though :+1:

Maffooch (approved):

cneill (approved):

PR #9626, opened 2024-02-25
Revert "Update Helm release postgresql-ha from 9.4.11 to v13 (helm/defectdojo/Chart.yaml)"

dsever:
Reverts DefectDojo/django-DefectDojo#9553

This should be reverted, The next move would be to configure renovate not to go with the big bang in case of the helm, but rather to work on minor and patch versions.


Maffooch (approved):

Maffooch:
@dsever I tried to base this against bugfix for so that this fix could be in all the versions, but the diff was not clean. I made this PR instead #9630

PR #9645, opened 2024-02-28
DRF Spectacular: Enable sidecar by default

### LINKED ISSUE #9629: OpenAPI 3 docs not showing example values
When looking at the Swagger (OpenAPI 3) docs, I noticed that for a bunch of api calls the example and also acceptable values are missing and just "string" is shown.
The issue is showing up on v2.28.0 and on the latest dev version. Versions between may also be affected.

Below some examples:
image
image
image
image

Maffooch:
Most of the endpoints in the drf-spectacular docs do not have any meaningful examples or response bodies present. This makes the API much more difficult to use. fixes #9629 

[sc-4511]

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -206,7 +206,7 @@ env = environ.FileAwareEnv(
     # if you want to keep logging to the console but in json format, change this here to 'json_console'
     DD_LOGGING_HANDLER=(str, 'console'),
     # If true, drf-spectacular will load CSS & JS from default CDN, otherwise from static resources
-    DD_DEFAULT_SWAGGER_UI=(bool, True),
+    DD_DEFAULT_SWAGGER_UI=(bool, False),
     DD_ALERT_REFRESH=(bool, True),
     DD_DISABLE_ALERT_COUNTER=(bool, False),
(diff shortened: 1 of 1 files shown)


cneill (approved):

mtesauro:
@Maffooch Based on the [comment](https://github.com/DefectDojo/django-DefectDojo/issues/9629#issuecomment-1969967206) in #9629 is setting `DD_DEFAULT_SWAGGER_UI` to false the solution we're going with or should this be updated?

Maffooch:
@mtesauro there has been some developments in this area. The swagger UI project has implemented a fix, and is targeting a release on Monday, so this issue should resolve itself upstream without any action from us. 

However, I think it may be wise to start pinning the swagger UI version via the sidecar app. They release monthly, so we could have a higher chance of catching things like this before we do our releases.

Here is a link to a similar response in the issue: https://github.com/DefectDojo/django-DefectDojo/issues/9629#issuecomment-1973485452

mtesauro:
@Maffooch Yeah, I agree that pinning makes sense to allow us to handle situations like this more gracefully.

grendel513 (approved):

devGregA (approved):

PR #9649, opened 2024-02-29
Bundler Audit Parser - Support for GHSA-Only Findings

rh0dy:
**Description**

Results from a `bundler audit` scan can include findings that contain a GHSA ID but not a CVE ID (see [here](https://github.com/rubysec/bundler-audit/commit/2eb783bcab18bd5732c6a6792b9880758f007872)). I think these types of findings occur for a few reasons, i.e. the GHSA hasn't been assigned a CVE ID yet or that it never will be assigned a CVE ID. 

Currently, the `bundler audit` parser doesn't support these types of findings, which feels like we may be missing out on importing valid vulnerabilities into DefectDojo.

This PR is to support parsing of `bundler audit` findings that may only contain a GHSA but not a CVE ID:

Pasted Graphic 9<br />

**Test results**

```
./dc-unittest.sh --profile postgres-redis --test-case unittests.tools.test_bundler_audit_parser.TestBundlerAuditParser

...

test_get_findings (unittests.tools.test_bundler_audit_parser.TestBundlerAuditParser.test_get_findings) ... ok
test_get_findings_version9 (unittests.tools.test_bundler_audit_parser.TestBundlerAuditParser.test_get_findings_version9) ... ok

Ran 2 tests in 0.009s

OK
Pre
```

**Documentation**

N/A.


mtesauro (approved):
Approved

dojo/tools/bundler_audit/parser.py
```diff
@@ -28,17 +28,19 @@ def get_findings(self, filename, test):
...
                     gem_name = field.replace("Name: ", "")
                 elif field.startswith("Version"):
                     gem_version = field.replace("Version: ", "")
                 elif field.startswith("Advisory"):
-                    advisory_cve = field.replace("Advisory: ", "")
+                    advisory_id = field.replace("Advisory: ", "")
                 elif field.startswith("CVE"):
-                    advisory_cve = field.replace("CVE: ", "")
+                    advisory_id = field.replace("CVE: ", "")
+                elif advisory_id is None and field.startswith("GHSA"):
```
manuel-sommer, line 42: Hi @rh0dy:
I guess `advisory_id is None` and then replacing it will result in:
`AttributeError: 'NoneType' object has no attribute 'replace'`
Did you test this?
Furthermore, I don't recommend to replace `("GHSA: ", "")` as this is relevant information.

dojo/tools/bundler_audit/parser.py
```diff
@@ -91,8 +93,8 @@ def get_findings(self, filename, test):
                     component_name=gem_name,
                     component_version=gem_version,
```
manuel-sommer, line 94: could you please use advisory_id to fill this into Finding.cve? 
https://github.com/DefectDojo/django-DefectDojo/blob/c5d896b2a606985f41669388962e47409f4866ba/dojo/models.py#L2206
I guess Finding.cve also can handle GHSA as far as I know.

unittests/tools/test_bundler_audit_parser.py
```diff
@@ -56,3 +56,11 @@ def test_get_findings_version9(self):
...
                 self.assertEqual("sprockets", finding.component_name)
                 self.assertEqual("2.2.3", finding.component_version)
+            with self.subTest(i=3):
+                finding = findings[3]
+                self.assertEqual("Gem nokogiri: Improper Handling of Unexpected Data Type in Nokogiri [GHSA-xc9x-jj77-9p9j]", finding.title)
+                self.assertEqual("Medium", finding.severity)
+                self.assertEqual(1, len(finding.unsaved_vulnerability_ids))
+                self.assertEqual("GHSA-xc9x-jj77-9p9j", finding.unsaved_vulnerability_ids[0])
+                self.assertEqual("nokogiri", finding.component_name)
+                self.assertEqual("1.15.2", finding.component_version)
```
manuel-sommer, line 66: Could you also please add then another unittest for "finding.cve"?

unittests/scans/bundler_audit/version_9.0.json
```diff
@@ -1 +1 @@
...
...release should either upgrade or use one of the work arounds immediately.\n\nWorkaround:\nIn Rails applications, work around this issue, set `config.assets.compile = false` and\n`config.public_file_server.enabled = true` in an initializer and precompile the assets.\n\n This work around will not be possible in all hosting environments and upgrading is advised.\n","cvss_v2":null,"cvss_v3":null,"cve":"2018-3760","osvdb":null,"ghsa":"pr3h-jjhj-573x","unaffected_versions":[],"patched_versions":[">= 2.12.5, < 3.0.0",">= 3.7.2, < 4.0.0",">= 4.0.0.beta8"],"criticality":null}},{"type":"unpatched_gem","gem":{"name":"nokogiri","version":"1.15.2"},"advisory":{"path":"/home/damien/.local/share/ruby-advisory-db/gems/nokogiri/GHSA-xc9x-jj77-9p9j.yml","id":"GHSA-xc9x-jj77-9p9j","url":"https://github.com/sparklemotion/nokogiri/security/advisories/GHSA-xc9x-jj77-9p9j","title":"Improper Handling of Unexpected Data Type in Nokogiri","date":"2024-02-04","description":"Nokogiri updates packaged libxml2 to v2.12.5 to resolve CVE-2024-25062.\n","cvss_v2":null,"cvss_v3":null,"cve":null,"osvdb":null,"ghsa":"GHSA-xc9x-jj77-9p9j","unaffected_versions":[],"patched_versions":[">= 1.16.2"],"criticality":null}}]}
```
manuel-sommer, line 1: @rh0dy could you please also refactor the json file?
There are e.g. vscode plugins to refactor the json file to make it more human readable.

manuel-sommer (changes requested):

dojo/tools/bundler_audit/parser.py
rh0dy: Hey @manuel-sommer, thanks for the review! I might be missing something here, can you check my understanding as I'm not sure how `AttributeError` would get raised & it shouldn't be removing relevant information from `GHSA`:
- I don't think it performs a `.replace` on `advisory_id`, instead it assigns a string value to `advisory_id` 
- If the finding doesn't contain a `CVE` it will use the `GHSA`, but only if `CVE` isn't set (so that CVE is prioritised)
- The `.replace` is for truncating the beginning part of the line so it's left with only the ID part, e.g. `GHSA: GHSA-j6w9-fv6q-3q52` becomes `GHSA-j6w9-fv6q-3q52`

unittests/tools/test_bundler_audit_parser.py
rh0dy: https://github.com/DefectDojo/django-DefectDojo/pull/9649/commits/626defd50904b720429a8dd37f0b7891242ee014

dojo/tools/bundler_audit/parser.py
rh0dy: https://github.com/DefectDojo/django-DefectDojo/commit/626defd50904b720429a8dd37f0b7891242ee014

dojo/tools/bundler_audit/parser.py
manuel-sommer: aaah, sorry, I guess my review here is wrong. you can mark it as "resolved".

manuel-sommer (approved):

dojo/tools/bundler_audit/parser.py
rh0dy: Will do! No worries & thank you for the fast review 🎉

cneill (approved):

Maffooch (approved):

devGregA (approved):

PR #9651, opened 2024-02-29
FindingGroupSerializer: not break schemas when JIRAIssue not available

### LINKED ISSUE #9643: JiraIssue and FindingGroup  in models should be decoupled
**Be informative**
I use OpenAPI spec to generate me a nice client without too much effort but the generated models (or the spec in Django) might be wrong

**Bug description**
I use https://openapi-generator.tech/ to generate me a Client  + Models from Swgger Json "/api/v2/oa3/schema/?format=json"
So far all good.

How ever when I try to list my Tests with
`TestsApi(client).tests_list(**kwargs)`

I got the following issue:
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for FindingGroup
jira_issue
  Input should be a valid dictionary or instance of JIRAIssue [type=model_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.6/v/model_type
```

This is because the generated models create something like

```
class FindingGroup(BaseModel):
    """
    FindingGroup
    """ # noqa: E501
    id: StrictInt
    name: Annotated[str, Field(strict=True, max_length=255)]
    test: StrictInt
    jira_issue: JIRAIssue
    __properties: ClassVar[List[str]] = ["id", "name", "test", "jira_issue"]
```

meaning that EVERY finding group MUST have a valid JiraIssue which is not the intended model.
Maybe there is something wrong in the Relationship between JiraIssue and FindingGroup

Workarround (at date)

- Edit the Pydantic model for this

```
class FindingGroup(BaseModel):
    """
    FindingGroup
    """ # noqa: E501
    id: StrictInt
    name: Annotated[str, Field(strict=True, max_length=255)]
    test: StrictInt
    jira_issue: Optional[JIRAIssue]
    __properties: ClassVar[List[str]] = ["id", "name", "test", "jira_issue"]
```

**Steps to reproduce**
1. Retreive shcema.json file from http://defectdojo.localhost/api/v2/oa3/schema/?format=json
2. Run in your favorise docker engine
   a. `podman run --rm -v ${PWD}:/local openapitools/openapi-generator-cli generate  -i /local/schema.json -g python -o /local/out/python`
3. Create some test with finding groups enabled
4. Run Following snippet (should be update to match package name)

```
from defectdojo_ci.openapi import ApiClient, Configuration
from defectdojo_ci.openapi.api import TestsApi

apiToken="xxxxx"
config = Configuration(host=defectdojo.localhost, api_key={"tokenAuth":apiToken}, api_key_prefix={"tokenAuth":"Token"})
client = ApiClient(configuration=config)

print(TestsApi(client ).tests_list(**kwargs).results)
```

**Expected behavior**
Print result

- [ ] Docker Compose
- [x] Kubernetes
- [ ] GoDojo

**Environment information**
 - OpenApi Schema from image defectdojo/defectdojo-django:2.31.4
 - 
**Logs** 
No servers logs because it's an issue in the schema models

kiblik:
Fix #9643

There is still an open question: Is this the only place with this kind of bug?
Check analysis in https://github.com/DefectDojo/django-DefectDojo/issues/9643#issuecomment-1971291731

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1394,7 +1394,7 @@ class DevelopmentEnvironmentSerializer(serializers.ModelSerializer):
 
 
 class FindingGroupSerializer(serializers.ModelSerializer):
-    jira_issue = JIRAIssueSerializer(read_only=True)
+    jira_issue = JIRAIssueSerializer(read_only=True, allow_null=True)
 
     class Meta:
         model = Finding_Group


mtesauro (approved):
Approved

cneill (approved):

blakeaowens (approved):

PR #9654, opened 2024-02-29
Metrics: Status unification

Maffooch:
There was some disparity regarding the flags that consist of a given status. A prime example is product type metrics only measuring verified findings while most of defectdojo does not. Unification of these queries to those in `dojo.finding.helper` not only makes status of a given finding more consistent from page to page, but it also will make any future changes to a status a much smaller lift, and reduce the chances of creating bugs

[sc-4459]


grendel513 (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

devGregA (approved):

PR #9503, opened 2024-02-07
Move PYTHONWARNINGS to EnvVar

kiblik:
- Move PYTHONWARNINGS to EnvVar
- Be strict in debug as well
- Do not fail but be more verbose in dev

PR #9659, opened 2024-03-01
Force to use DjangoDivFormRenderer

kiblik:
This setting is here to override default renderer of forms (use div-based, instred of table-based).
It has effect only on templates that use "{{ form }}" in the body. Only "Delete forms" now.
The setting is here to avoid RemovedInDjango50Warning.

More info:
- https://docs.djangoproject.com/en/4.1/ref/forms/renderers/#django.forms.renderers.DjangoTemplates
- https://docs.djangoproject.com/en/5.0/ref/forms/renderers/#django.forms.renderers.DjangoTemplates

Notes:
 - Remove this setting in Django 5.0 because DjangoDivFormRenderer will become deprecated and the same class will be used by default DjangoTemplates.
 - Discovered during debugging #9503

Btw, there is no visual difference.

Before:
image

After:
image


Maffooch (approved):

cneill (approved):

mtesauro (approved):
Approved

devGregA (approved):

PR #9661, opened 2024-03-01
StringAgg: use default Value('')

kiblik:
To avoid `RemovedInDjango50Warning - In Django 5.0, StringAgg() will return None instead of an empty string if there are no rows. Pass default=None to opt into the new behavior and silence this warning or default=Value('') to keep the previous behavior.`

Note: Discovered in #9503


cneill (approved):

mtesauro (approved):
Approved

devGregA (approved):

PR #9679, opened 2024-03-05
Func views to class based views in finding exports

FelixHernandez:
## Func views to class based views in finding exports
[sc-2860]

**Description**

This changes converted the function views (HTML, CSV and Excel) exports to classes
Also the CSV and Excel export now contains tags in a column.


grendel513 (approved):

mtesauro (approved):
Approved

Maffooch (approved):

cneill (approved):

PR #9691, opened 2024-03-06
Jira: Improve alerting on a per step basis

Maffooch:
Most of the push/update to jira logic is  surrounded in a single try/catch, so if any step of the process goes wrong, the entire process fails. This is not ideal in some cases. The case that caught my attention the most is when the default assignee on a product no longer exists in Jira, so when a new issue is created, the link between dojo and jira is never established because of the failed assignee step. This results in a new ticket in jira that is never linked back to defectdojo, and ultimately led to duplicate jira tickets.

[sc-4460]

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Any
 from dojo.utils import add_error_message_to_response, get_system_setting, to_str_typed
 import os
 import io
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

cneill (approved):
Good stuff! :+1:

grendel513 (approved):

PR #9692, opened 2024-03-06
Npm audit v7+ Parser

grendel513:
Adding parser for new format of npm audit json as of version 7.  Implementing as a separate parser to avoid collision, aid in deduplication and maintainability.


manuel-sommer:
https://github.com/DefectDojo/django-DefectDojo/issues/4285

mtesauro (approved):
Approved

dojo/tools/npm_audit_7_plus/parser.py
```diff
@@ -0,0 +1,237 @@
...
+        component_version=component_version,
+        false_p=False,
+        duplicate=False,
+        out_of_scope=False,
+        mitigated=None,
+        impact="No impact provided",
+        static_finding=static_finding,
+        dynamic_finding=False,
+        unique_id_from_tool=unique_id_from_tool,
+        vuln_id_from_tool=unique_id_from_tool
```
cneill, line 169: I think we usually only populate one of these with the same identifier. I tried to dig into what `source` actually means to `npm audit`, and as far as I can tell it is a stable identifier for either a CVE or GHSA, not an individual *instance* of an issue detected in a specific codebase. So I think it better [matches the intent of `vuln_id_from_tool`](https://github.com/DefectDojo/django-DefectDojo/blob/82181253cd303efd1642994327e1d862c6b247ef/dojo/models.py#L2444-L2453)

```suggestion
        vuln_id_from_tool=unique_id_from_tool
```

dojo/tools/npm_audit_7_plus/parser.py
```diff
@@ -0,0 +1,237 @@
...
+        # this is a top level key (a vulnerability)
+        title = item_node["name"]
+        cwe = "CWE-1035"  # default
+        component_name = title
+        component_version = "n/a"
+
+    elif item_node["via"] and isinstance(item_node["via"][0], dict):
+        title = item_node["via"][0]["title"]
+        component_name = item_node["nodes"][0]
+        component_version = item_node["range"]
```
cneill, line 122: I think we typically use the actual version of the dependency detected by the tool here, rather than a range of versions affected by the vulnerability. @Maffooch am I correct about this?

Unfortunately, it doesn't appear that `npm audit` actually provides this information...

dojo/tools/npm_audit_7_plus/parser.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def get_vuln_description(item_node, tree):
+    """Make output pretty of details."""
+    effects_handled = []
+    description = ""
+
+    description = (description + item_node["name"] + " " +
+                   item_node["range"] + "\n")
+    description = description + "Severity: " + item_node["severity"] + "\n"
```
cneill, line 187: Just a minor nit, but why not use `+=` here and in the cases below?

```suggestion
    description += item_node["name"] + " " + item_node["range"] + "\n"
    description += "Severity: " + item_node["severity"] + "\n"
```

dojo/tools/npm_audit_7_plus/parser.py
```diff
@@ -0,0 +1,237 @@
...
+
+    for via in item_node["via"]:
+        if isinstance(via, str):
+            description = (description +
+                           "Depends on vulnearble versions of " + via + "\n")
```
cneill, line 192:
```suggestion
                           "Depends on vulnerable versions of " + via + "\n")
```

dojo/tools/npm_audit_7_plus/parser.py
```diff
@@ -0,0 +1,237 @@
...
+    for effect in item_node["effects"]:
+        # look up info in the main tree
+        description = (description + "  " + tree[effect]["name"] + " " +
+                       tree[effect]["range"] + "\n")
+        effects_handled.append(tree[effect]["name"])
+        for ev in tree[effect]["via"]:
+            if isinstance(ev, dict):
+                if tree[effect]["name"] != ev["name"]:
+                    description = (description +
+                                   "  Depends on vulnearble versions of " +
```
cneill, line 219:
```suggestion
                                   "  Depends on vulnerable versions of " +
```

dojo/tools/npm_audit_7_plus/parser.py
```diff
@@ -0,0 +1,237 @@
...
+                                   ev["name"] + "\n")
+            else:
+                if tree[effect]["name"] != ev:
+                    description = (description +
+                                   "  Depends on vulnearble versions of " +
```
cneill, line 224:
```suggestion
                                   "  Depends on vulnerable versions of " +
```

dojo/tools/npm_audit_7_plus/parser.py
```diff
@@ -0,0 +1,237 @@
...
+        severity = "Critical"
+    else:
+        severity = "Info"
+
+    if item_node["via"] and isinstance(item_node["via"][0], str):
+        # this is a top level key (a vulnerability)
+        title = item_node["name"]
+        cwe = "CWE-1035"  # default
+        component_name = title
+        component_version = "n/a"
```
cneill, line 117: No need to specify `n/a` here ([ref](https://documentation.defectdojo.com/contributing/how-to-write-a-parser/#not-all-attributes-are-mandatory))

dojo/settings/settings.dist.py
```diff
@@ -1278,6 +1278,7 @@ def saml2_attrib_map_format(dict):
     'Tenable Scan': True,
     'Nexpose Scan': True,
     'NPM Audit Scan': True,
+    "NPM Audit v7+ Scan": True,
```
cneill, line 1281:
```suggestion
    'NPM Audit v7+ Scan': True,
```

cneill (changes requested):
Noticed a couple typos, some stylistic suggestions, and a few cases where I think this populates fields in a non-standard way.

dojo/tools/npm_audit_7_plus/parser.py
grendel513: the tool does not provide the specific version, just the range - i thought it would be best to capture it in the field rather than just the description (where it also is)

dojo/tools/npm_audit_7_plus/parser.py
grendel513: Thank you, addressed in this commit https://github.com/DefectDojo/django-DefectDojo/pull/9692/commits/31746fd05a45cc587dd51941ed680f5dff34a736

dojo/tools/npm_audit_7_plus/parser.py
grendel513: true, true: https://github.com/DefectDojo/django-DefectDojo/pull/9692/commits/31746fd05a45cc587dd51941ed680f5dff34a736

dojo/tools/npm_audit_7_plus/parser.py
Maffooch: A single version is the intended use case. Because that is not available, at least something should go in the component version field. The tricky thing I can foresee happening is if two vulnerabilities affect different version ranges (example: version 1.1-1.5 =/= 1.1-1.4) of the same component, it will be impossible to aggregate findings under a single component + version combo.

grendel513: Thanks for taking a look @Maffooch  - should I remove the assignment then? Seems like the prudent thing to do given it is t a specific version.

Maffooch: Sure, let's remove it

grendel513: Sounds good. new commit: https://github.com/DefectDojo/django-DefectDojo/pull/9692/commits/7eb7edb0a74c014499d66cbca6beead87eb2b703

@cneill @Maffooch all feedback addressed.  Thank you.

dojo/settings/settings.dist.py
Maffooch, line 1: Please add a deduplication algorithm here

dojo/tools/npm_audit_7_plus/parser.py
```diff
@@ -0,0 +1,230 @@
...
+        references=", ".join(references),
+        component_name=component_name,
+        false_p=False,
+        duplicate=False,
+        out_of_scope=False,
+        mitigated=None,
+        impact="No impact provided",
+        static_finding=static_finding,
+        dynamic_finding=False,
+        vuln_id_from_tool=unique_id_from_tool
```
Maffooch, line 165:
```suggestion
        vuln_id_from_tool=unique_id_from_tool,
        unique_id_from_tool=unique_id_from_tool,
```

Maffooch (approved):

Maffooch (changes requested):
Clicked the wrong button above 🤡

dojo/tools/npm_audit_7_plus/parser.py
```diff
@@ -0,0 +1,226 @@
...
+        component_name=component_name,
+        false_p=False,
+        duplicate=False,
+        out_of_scope=False,
+        mitigated=None,
+        impact="No impact provided",
+        static_finding=static_finding,
+        dynamic_finding=False,
+        vuln_id_from_tool=unique_id_from_tool,
+        unique_id_from_tool=unique_id_from_tool,
```
cneill, line 163: As [mentioned in my comment here](https://github.com/DefectDojo/django-DefectDojo/pull/9692#discussion_r1515386367), I think this should only populate `vuln_id_from_tool`. Am I misinterpreting the meaning of these 2 fields @Maffooch ? Maybe the variable should be called `vuln_id_from_tool` instead of `unique_id_from_tool` if my interpretation is correct

```suggestion
        vuln_id_from_tool=unique_id_from_tool,
```

dojo/settings/settings.dist.py
grendel513: Thank you, added in commit https://github.com/DefectDojo/django-DefectDojo/pull/9692/commits/8574b35643e39e9967406b0aab3c7940967c42fd

Maffooch (approved):

cneill (approved):
Thanks for making all those changes!

qsoul (approved):

blakeaowens (approved):

PR #9699, opened 2024-03-08
Add NOTIFICATIONS_SYSTEM_LEVEL_TRUMP

kiblik:
Revert a small part from #9227.

Full explanation in https://github.com/DefectDojo/django-DefectDojo/pull/8824#issuecomment-1986098912

### DIFF
--- a/docs/content/en/integrations/notifications.md
+++ b/docs/content/en/integrations/notifications.md
@@ -120,7 +120,7 @@ To activate notifications to Microsoft Teams, you have to:
 
 System notification settings (scope: system) describe the sending of notifications to superadmins. User notification settings (scope: personal) describe sending notifications to the specific user.
 
-However, there is a specific use-case when the user decides to disable notifications (to decrease noise) but the system setting is used to override this behavior. These overrides apply only to `user_mentioned` and `review_requested`.
+However, there is a specific use-case when the user decides to disable notifications (to decrease noise) but the system setting is used to override this behavior. These overrides apply only to `user_mentioned` and `review_requested` by default.
 
 The scope of this setting is customizable (see environmental variable `DD_NOTIFICATIONS_SYSTEM_LEVEL_TRUMP`).
 
--- a/docs/content/en/integrations/notifications.md
+++ b/docs/content/en/integrations/notifications.md
@@ -115,3 +115,13 @@ To activate notifications to Microsoft Teams, you have to:
 - Configure an Incoming Webhook in a Teams channel and copy the URL of the webhook to the clipboard
 - Activate `Enable Microsoft Teams notifications` in the System Settings
 - Paste the URL of the Incoming Webhook into the field `Msteams url`
+
+## Specific overrides
+
+System notification settings (scope: system) describe the sending of notifications to superadmins. User notification settings (scope: personal) describe sending notifications to the specific user.
+
+However, there is a specific use-case when the user decides to disable notifications (to decrease noise) but the system setting is used to override this behavior. These overrides apply only to `user_mentioned` and `review_requested`.
+
+The scope of this setting is customizable (see environmental variable `DD_NOTIFICATIONS_SYSTEM_LEVEL_TRUMP`).
+
+For more information about this behavior see the [related pull request #9699](https://github.com/DefectDojo/django-DefectDojo/pull/9699/)
--- a/unittests/test_notifications.py
+++ b/unittests/test_notifications.py
@@ -1,5 +1,8 @@
 from .dojo_test_case import DojoTestCase
 from dojo.models import Product, User, Notifications
+from dojo.models import DEFAULT_NOTIFICATION
+from dojo.notifications.helper import create_notification, send_alert_notification
+from unittest.mock import patch
 
 
 class TestNotifications(DojoTestCase):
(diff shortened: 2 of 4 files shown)


mtesauro (approved):
Approved

grendel513 (approved):

Maffooch (changes requested):
Dropped a comment in the other PR since that is where the bulk of the context around this PR is: https://github.com/DefectDojo/django-DefectDojo/pull/8824#issuecomment-1988747569

Maybe we could add an additional check for the `event` type to determine if the system level notifications should trump the user level. Something like this (I could get suggestions to work here):
```
if event in ["user_mentioned", "review_requested"]:
    # merge the system level notifications with the personal level
    # this allows for system to trump the personal
    merged_notifications = Notifications.merge_notifications_list([system_notifications, recipient_notifications])
    merged_notifications.user = recipient_notifications.user
    logger.debug('Sent notification to %s', merged_notifications.user)
    process_notifications(event, merged_notifications, **kwargs)
else:
    # Do not trump user preferences and send notifications as usual
    logger.debug('Sent notification to %s', recipient_notifications.user)
    process_notifications(event, recipient_notifications, **kwargs)
```

kiblik:
> Dropped a comment in the other PR since that is where the bulk of the context around this PR is: [#8824 (comment)](https://github.com/DefectDojo/django-DefectDojo/pull/8824#issuecomment-1988747569)
> 
> Maybe we could add an additional check for the `event` type to determine if the system level notifications should trump the user level. Something like this (I could get suggestions to work here):
> 
> ```
> if event in ["user_mentioned", "review_requested"]:
>     # merge the system level notifications with the personal level
>     # this allows for system to trump the personal
>     merged_notifications = Notifications.merge_notifications_list([system_notifications, recipient_notifications])
>     merged_notifications.user = recipient_notifications.user
>     logger.debug('Sent notification to %s', merged_notifications.user)
>     process_notifications(event, merged_notifications, **kwargs)
> else:
>     # Do not trump user preferences and send notifications as usual
>     logger.debug('Sent notification to %s', recipient_notifications.user)
>     process_notifications(event, recipient_notifications, **kwargs)
> ```

Thank you @Maffooch for your reaction.
Purely technically, yes, this sounds like an alternative that would work.
However,
- This creates an anti-pattern in the systematic handling of notification
  - I suppose it puts evaluation (of how and if a notification should be processed) logic in the place where it should not be.
- It would create undo documented behavior solving one (sorry, two) specific use-case(s).

grendel513:
I believe @Maffooch thought process and approach seems the best here

mtesauro:
@grendel513 Thanks for chiming in - I've been meaning to comment on this for far too long.

Basically, I agree with the idea of checking the event from @Maffooch - I'm thinking that if I were a user who opted out of a notification, I'd be irritated if I got one anyway - excluding mentions or specific requests for me.

Matt's 2 cents

kiblik:
Ok, I will change the implementation to what @Maffooch offered.
But I will add an option to customize a list of events via EnvVars (loaded in Settings) with default `["user_mentioned", "review_requested"]` - to allow admins to easily change this behavior.

kiblik:
Hi @mtesauro / @devGregA / @Maffooch / @cneill / @blakeaowens. I would like to ask for a review.
It is an enabler for #8824 and it would be nice to have them both in the following release.

Maffooch (approved):

mtesauro:
@kiblik 
> Ok, I will change the implementation to what @Maffooch offered. But I will add an option to customize a list of events via EnvVars (loaded in Settings) with default `["user_mentioned", "review_requested"]` - to allow admins to easily change this behavior.

Fair enough - that's a great middle ground. Thanks for that addition :+1:

mtesauro (approved):
Approved

unittests/test_notifications.py
```diff
@@ -55,3 +58,46 @@ def test_merge_notifications_list(self):
...
+        notif_system, _ = Notifications.objects.get_or_create(user=None, template=False)
+
+        last_count = 0
+        with self.subTest('user off, system off'):
+            notif_user.user_mentioned = ()  # no alert
+            notif_user.save()
+            notif_system.user_mentioned = ()  # no alert
+            notif_system.save()
+            create_notification(event="user_mentioned", title="user_mentioned", recipients=['admin'])
+            self.assertEqual(mock.call_count, last_count + 0)
```
cneill, line 74: The `+ 0` doesn't seem necessary here?

```suggestion
            self.assertEqual(mock.call_count, last_count)
```

docs/content/en/integrations/notifications.md
```diff
@@ -115,3 +115,13 @@ To activate notifications to Microsoft Teams, you have to:
 - Configure an Incoming Webhook in a Teams channel and copy the URL of the webhook to the clipboard
 - Activate `Enable Microsoft Teams notifications` in the System Settings
 - Paste the URL of the Incoming Webhook into the field `Msteams url`
+
+## Specific overrides
+
+System notification settings (scope: system) describe the sending of notifications to superadmins. User notification settings (scope: personal) describe sending notifications to the specific user.
+
+However, there is a specific use-case when the user decides to disable notifications (to decrease noise) but the system setting is used to override this behavior. These overrides apply only to `user_mentioned` and `review_requested`.
```
cneill, line 123:
```suggestion
However, there is a specific use-case when the user decides to disable notifications (to decrease noise) but the system setting is used to override this behavior. These overrides apply only to `user_mentioned` and `review_requested` by default.
```

cneill (changes requested):
Just a few small things, otherwise looks good

unittests/test_notifications.py
kiblik: From the mathematical perspective, there is no difference. It just increases readability in the context of this unit test  - all other tests use `+ number` notation.
But if it is the only issue I can remove it.

cneill (approved):

kiblik:
@mtesauro, can I ask for a merge please?

mtesauro:
@kiblik Sure thing - merging now

Issue #9700, opened 2024-03-08
Flot-axis still in Yarn.lock file under Components folder


CarpFlop:
Flot-axis was removed from everywhere except yarn.lock.

It's a malicious package with a Synk rating of 9.8/10 Criticality.
Please hotfix for part of latest release that was meant to address this issue!

Thank you guys for your hard work

CarpFlop:
https://github.com/DefectDojo/django-DefectDojo/issues/9187 Potentially false positive:

But since all references and usage of flot-axislabels was removed in latest update, potentially still worth cleaning up lingering reference in lockfile

mtesauro:
@CarpFlop  Agree that this appears to be lingering cruft vs a real issue as this library was removed in  #9540 

I do agree we should remove that reference if for no other reason to remove confusion.

manuel-sommer:
This can be close @mtesauro ;-) was already released.

PR #9540, opened 2024-02-13
Remove flot-axis JS library

FelixHernandez:
## Remove `flot-axis` JS library 
[sc-2602]

**Description**
Since this library will no longer be maintained, it has been removed to make the app be secure and without vulnerabilities.

PR #9707, opened 2024-03-10
remove flot-axis, issue 9700

manuel-sommer:
see issue #9700


mtesauro:
@manuel-sommer Can you rebase this against bugfix so this cruft can be removed sooner rather then later?  Library use was removed in #9540 so this can only cause confusion.

manuel-sommer:
I messed up something in the git magic. Here is a followup PR @mtesauro https://github.com/DefectDojo/django-DefectDojo/pull/9709

Issue #8120, opened 2023-05-17
Logs are written to stderr


joeppeeters:
Most logs (all?) seem to be send to `/dev/stderr`. 

After deploying DefectDojo on our Nomad infra it appeared quite noisy on our error monitoring. After investigation it seems to send a lot of (info level?) logs to the error device.

**Steps to reproduce**
Steps to reproduce the behaviour based on a clean checkout of this repo running on OSX:

1. Start dojo by: `./dc-up.sh`
2. Tail logs, filter stdout only: `docker logs django-defectdojo-uwsgi-1 -f 2>/dev/null`
3. Notice logs are empty
4. Tail logs, filter stderr only: `docker logs django-defectdojo-uwsgi-1 -f 1>/dev/null`
5. Notice logs come in

**Expected behavior**
Logs are send to the correct device

**Deployment method** *(select with an `X`)*
- [x] Docker Compose (presumably not relevant)
- [x] Nomad 
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: replay scenario on OSX
 - DefectDojo version: `v2.22.2`, `master`

**Observations**

affected containers:
 - `django-defectdojo-uwsgi-1`
 - `django-defectdojo-celerybeat-1`
 - `django-defectdojo-celeryworker-1`

expected behaviour on containers:
 - `django-defectdojo-nginx-1`
 - `django-defectdojo-mysql-1`
 - `django-defectdojo-rabbitmq-1`

which points to the Django/Python config?

**Other sources**
 - Thread on [Slack](https://owasp.slack.com/archives/C2P5BA8MN/p1684251739075289)
 - Similar symptoms described on [Stackoverflow](https://stackoverflow.com/questions/44977066/logging-stdout-with-python3-and-uwsgi)

dschniepp:
@joeppeeters how did you solve the issue? @dsever are you still on the topic?

anthonwellsjo:
Hello 👋🏻 

We've deployed on Kubernetes using [the documentation available](https://github.com/DefectDojo/django-DefectDojo/blob/dev/helm/defectdojo/README.md). We're observing the same issues with uwsgi and celery containers writing info logs to stderr. This generates a lot of noise for us:

<img width="1741" height="232" alt="Image" src="https://github.com/user-attachments/assets/56f8aec2-fa1f-4fa4-b679-b4b87b0156a4" />

[...]

PR #9708, opened 2024-03-10
Log to stdout

kiblik:
Fix for #8120

### DIFF
--- a/docker/entrypoint-celery-beat.sh
+++ b/docker/entrypoint-celery-beat.sh
@@ -27,5 +27,6 @@ python3 manage.py check
 
 exec celery --app=dojo \
     beat \
+  --logfile=stdout \
   --pidfile=/var/run/defectdojo/celery-beat.pid \
   --schedule=/var/run/defectdojo/celerybeat-schedule
--- a/docker/entrypoint-celery-worker.sh
+++ b/docker/entrypoint-celery-worker.sh
@@ -34,6 +34,7 @@ python3 manage.py check
 
 exec celery --app=dojo \
     worker \
+  --logfile=stdout \
   --loglevel="${DD_CELERY_LOG_LEVEL}" \
   --pool="${DD_CELERY_WORKER_POOL_TYPE}" \
   --concurrency="${DD_CELERY_WORKER_CONCURRENCY:-1}" \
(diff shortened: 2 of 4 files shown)


kiblik:
At the time, when I opened this PR, I was doing multiple tests with `log-master`, `logto`, `logto2`, ... to be able to achieve that Errors will be logged to `stderr` and everything else to `stdout`.
I checked many online resources (like https://github.com/unbit/uwsgi/issues/1601) but had no success.

I hoped I would have time in the future to come back with a fresh mind and test some new approaches but I had never time.

If I'm not wrong, `log-master = True` (a.k.a. `--log-master`) will push all logs only to`stdout`. If it is a suitable solution, I might test it one more time (to be sure about my previous sentence) and I can offer it as solution (change draft PR to open PR).

dschniepp:
I quickly [asked](https://github.com/DefectDojo/django-DefectDojo/issues/8120#issuecomment-2310425165) in the issue #8120 if still somebody else works on it, and prefers a different solution. From my point that sounds definitely like a valid approach and according to the discussion, currently the only one, which is easily implementable. The only question is, do we add just another flag or do we switch to `.ini` file like mozilla does https://github.com/mozilla/remote-settings/commit/a9f4da48944b8b3d8f59a235aa4cd0c1f6931387.

kiblik:
So after a small test: Yes, `log-master` redirects all application logs to `stdout` (instead of `stderr`). In my opinion, it might not be best but better than all logs in `stderr`.
However, `log-master` does not work for `celery(worker|beat)`. I tried `--logfile /dev/stdout` but without any luck.
Unfortunately, I do not have the capacity to investigate alternatives. But I will be happy if somebody is willing to help here.

barucijah:
@kiblik I have tried with --logfile=stdout and logs became info logs as they should be. 
Would you like to test it on your side?

kiblik:
Not sure if I'm doing sth wrong or if I'm testing with wrong methods but when I apply this: https://github.com/DefectDojo/django-DefectDojo/pull/9708/commits/0fc5d0815896453724da73c62efaa9249d9d4b24
I still see logs like 
```
[12/Sep/2024 16:42:19] INFO [celery.beat:632] beat: Starting...
```
in stderr, not in stdout.

barucijah:
@kiblik I tested my custom image on the instance deployed in Kubernetes. I tried to check the files, but I could not open them. However, I see in my Monitoring tool that logs are now shown as INFO, and before they were shown as ERRORS.
Here is an example of my change for the beat:
`exec celery --app=dojo \
    beat \
  --logfile=stdout \
  --pidfile=/var/run/defectdojo/celery-beat.pid \
  --schedule=/var/run/defectdojo/celerybeat-schedule`

Maffooch:

Issue #9712, opened 2024-03-11
Can't import Nexpose Scans without "Scan Completion Date"


StephanPillhofer:
**Bug description**
When you try to import a Nexpose report and don't specify a "Scan Completion Date", the import will fail. Pretty sure this is due to changes made in https://github.com/DefectDojo/django-DefectDojo/pull/9473. Doesn't happen with Burp imports for example.  Disabling Finding SLA's and trying the import again causes another import error.

**Steps to reproduce**
1. Create a new engagement
2. Upload test
3. Select Nexpose Scan
4. Don't specify a "Scan Completion Date"
5. Failes import

**Expected behavior**
Successful import

**Deployment method** *(select with an `X`)*
- [x ] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
Ubuntu 18.04
DefectDojo version v. 2.31.5

**Logs** 
```
[11/Mar/2024 12:58:49] ERROR [dojo.engagement.views:750] unsupported operand type(s) for -: 'datetime.date' and 'NoneType'

Traceback (most recent call last):

  File "/app/dojo/engagement/views.py", line 734, in post

    test, finding_count, closed_finding_count, _ = importer.import_scan(scan, scan_type, engagement, user, environment, active=active, verified=verified, tags=tags,

                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/app/dojo/importers/importer/importer.py", line 344, in import_scan

    new_findings = self.process_parsed_findings(test, parsed_findings, scan_type, user, active=active,

                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/app/dojo/decorators.py", line 48, in __wrapper__

    return func(*args, **kwargs)

           ^^^^^^^^^^^^^^^^^^^^^

  File "/usr/local/lib/python3.11/site-packages/celery/local.py", line 182, in __call__

    return self._get_current_object()(*a, **kw)

           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[...]

PR #9724, opened 2024-03-12
:bug: date is None when importing and reimporting, issue #9712

manuel-sommer:
see issue #9712

### DIFF
--- a/dojo/importers/importer/importer.py
+++ b/dojo/importers/importer/importer.py
@@ -38,8 +38,6 @@ class DojoDefaultImporter(object):
 
         if scan_date and not scan_date.tzinfo:
             scan_date = timezone.make_aware(scan_date)
-        elif scan_date is None:
-            scan_date = timezone.make_aware(datetime.today())
 
         if now and not now.tzinfo:
             now = timezone.make_aware(now)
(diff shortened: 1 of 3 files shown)


manuel-sommer:
@Maffooch, it looks like this same problem applies also to other scanners. (E.g. look at Qualys Scan and import https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/scans/qualys/monthly_pci_issue6932.csv)

manuel-sommer:
Basically: models.DateField may not be None

manuel-sommer:
@Maffooch, if there is not really a way to catch this fundamental problem in a more elegant way, I can also fix other parsers.

Maffooch:
@manuel-sommer maybe this issue could be fixed in the importer rather than in every single parser? I would think it would be a simple if statement to check if the finding date field is none. If so, insert today's date

manuel-sommer:
> @manuel-sommer maybe this issue could be fixed in the importer rather than in every single parser? I would think it would be a simple if statement to check if the finding date field is none. If so, insert today's date

Ah, I tried it differently but not yet that way. Gotta take a look to find the right place. 👀

manuel-sommer:
Could you take a look @Maffooch ?

Maffooch:
This looks fine generally. Tests are failing as it looks like there are instances of overwriting a date from a finding. I think you need to have the condition "if there is not a scan scan date supplied at import time, and not on supplied from the report at the parser level, then supply todays date at the importer level"

manuel-sommer:
@Maffooch, I spent some time and looked into it, but did not figure out exactly on how to fix it. I get your point, but don't have enough time to debug it. Feel free to close this PR and fix it in a new PR.

PR #9729, opened 2024-03-12
Ruff Linter: Resolve deprecation notice

Maffooch:
There are some warnings on the ruff linter action related to the deprecated default `check` action. The action must be explicitly set now

[sc-4736]

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -33,4 +33,4 @@ jobs:
         run: pip install -r requirements-lint.txt
 
       - name: Run Ruff Linter
-        run: ruff .
+        run: ruff check .


Maffooch:
Couple things to note to avoid confusion as I got tripped up by this during testing:
- Deprecation warning is still present in in the `Ruff Linter / ruff-linting (pull_request_target)` action since that action is run from the branch this PR is being merged into (in this case it is `bugfix`)
- The `Ruff Linter / ruff-linting (push)` action is correct now as that action is running from my `Maffooch-patch-6` branch

The warning will be resolved in all branches on the next release as the workflow file will be updated on all three branches at that time

cneill (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #9758, opened 2024-03-16
Ruff: add PERF

kiblik:
Add support for https://docs.astral.sh/ruff/rules/#perflint-perf in Ruff


Maffooch:

PR #9760, opened 2024-03-16
Flake8: Remove useless ignores

kiblik:
Remove `F405` and `F812`. Violations that are not in the code anymore.

`exclude` section can be removed because `dojo/unittests` and `['.git', '__pycache__']` is default for this parameter.

`max-complexity` can be removed because ignore for `C901` (Function is too complex) is already in place.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -12,10 +12,6 @@ ignore =
     C901
     # Bare except
     E722
-    # May be undefined
-    F405
-    # list comprehension redefines
-    F812
     E126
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

kiblik:
Hi @grendel513 / @devGregA / @blakeaowens. I would like to ask for a review.

grendel513 (approved):

PR #9766, opened 2024-03-17
[common] Postgresql and rabbit upgrade

dsever:
don't merge for now.

The problem with linting in the case of rabbit is related to different versions of helm dependency `common`  for chart rabbit and PostgreSQL, so basically always was in use old version coming from Postgres

I would propose using fresh helm definitions for Postgres and rabbit but the upgrade needs to be tested.
Another proposal is to instruct renovate just to propose patch upgrade, and not to break upgrade cadence anymore.


dsever:
After https://github.com/DefectDojo/django-DefectDojo/pull/9684 will be in dev I can continue with the analysis

mtesauro:
@dsever #9684 merged :smiley:

dsever:
Tests are green, now upgrade should be tested. (should be a part of regular test at some time )

mtesauro:
@dsever Checking back on this.

We've deprecated MySQL and RabbitMQ as of 2.36.0 and removed them from tests in 2.37.0. Since those are done, I wasn't sure if you wanted to revisit this or not.

mtesauro:
Closing for now since a bunch of helm changes have landed since this PR was made. Feel free to open a new one when you're ready.

PR #9772, opened 2024-03-18
Endpoints: Redefine vulnerable definition

Maffooch:
Redefine the vulnerability status of an endpoint to consider endpoint statuses rather than findings. An endpoint can only be considered mitigated when all endpoint statuses are marked as mitigated. This aligns with the behavior exhibited by the bulk mitigate feature on the view endpoint page

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1769,7 +1769,13 @@ class Endpoint(models.Model):
 
     @property
     def vulnerable(self):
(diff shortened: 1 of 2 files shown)


grendel513 (approved):

blakeaowens (approved):

mtesauro (approved):
Approved

PR #9788, opened 2024-03-19
API responses more detailed (optional)

FelixHernandez:
## API responses more detailed (optional in settings)
[sc-1235]
**Description**

The user now can turn on/off the setting called API expose error details in the system settings screen, When turned on, the API will expose error details in the response. When turned off, the response will be only `{"message":"Internal server error, check logs for details"}`

**Test results**
this is the new configuration.
Screenshot 2024-03-19 at 11 54 05 a m


Maffooch (changes requested):
Please revert the changes that remove logging. It is very important, and should not be removed

grendel513 (approved):

kiblik:
In my opinion, users should not have access to details of errors. I fully understand that this setting is opt-in but I suppose it is still avoiding best practice.
Only the admin should have (based on logs) access to details.

grendel513:
> In my opinion, users should not have access to details of errors. I fully understand that this setting is opt-in but I suppose it is still avoiding best practice. Only the admin should have (based on logs) access to details.

hi @kiblik - thanks for your comment.  The message does not include the full stack trace, just the message of the error encountered. Thus there is little risk to information disclosure or providing too much information on an error.  Additionally since it is optional and controlled by a setting, the individuals managing the installation can make the decision themselves, thus acknowledging and accepting the risk.

mtesauro (approved):
Approved

cneill (approved):

Maffooch (approved):

PR #9791, opened 2024-03-20
add cve values to openvas csv parser

manuel-sommer:
See https://github.com/DefectDojo/django-DefectDojo/discussions/9780

### DIFF
--- a/dojo/tools/openvas/csv_parser.py
+++ b/dojo/tools/openvas/csv_parser.py
@@ -1,6 +1,7 @@
 import csv
 import hashlib
 import io
+import re
 from dateutil.parser import parse
 from dojo.models import Finding, Endpoint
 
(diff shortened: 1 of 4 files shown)


mtesauro (approved):
Approved

Maffooch (changes requested):
The finding `cve` field will be deprecated in the near future. Please use the `vulnerability_id` field instead. It allows for having multiple CVEs on a single finding in the case of `CveColumnMappingStrategy`

dojo/tools/openvas/csv_parser.py
```diff
@@ -71,6 +72,31 @@ def map_column_value(self, finding, column_value):
...
 
+class CveColumnMappingStrategy(ColumnMappingStrategy):
+    def __init__(self):
+        self.mapped_column = "cves"
+        super(CveColumnMappingStrategy, self).__init__()
+
+    def map_column_value(self, finding, column_value):
+        if "," in column_value:
+            finding.description += "\n**All CVEs:** " + str(column_value)
+            finding.unsaved_vulnerability_ids.append(column_value.split(",")[0])
```
Maffooch, line 83: If there are multiple CVEs in this list, we should probably save more than just the first one

unittests/tools/test_openvas_parser.py
```diff
@@ -47,6 +47,40 @@ def test_openvas_csv_many_vuln(self):
...
+            parser = OpenVASParser()
+            findings = parser.get_findings(f, test)
+            for finding in findings:
+                for endpoint in finding.unsaved_endpoints:
+                    endpoint.clean()
+            self.assertEqual(13, len(findings))
+            finding = findings[2]
+            self.assertEqual("Apache HTTP Server Detection Consolidation", finding.title)
+            self.assertEqual("Info", finding.severity)
+            self.assertEqual(finding.unsaved_vulnerability_ids[0], "")
```
Maffooch, line 83: This seems like unexpected behavior. Why would we want an empty string to be here?

Maffooch (changes requested):

Maffooch (approved):

blakeaowens (approved):

cneill (approved):

PR #9796, opened 2024-03-20
Fix items not found in scout suite findings

jbschooley:
**Description**

Some Scout Suite findings do not include `items`. If this is the case, Defect Dojo will fail when importing a scan.

**Test results**

I imported a scan where 2 of the findings (`vpc-subnet-with-allow-all-egress-acls` and `vpc-subnet-with-allow-all-ingress-acls` were missing `items`. The scan imported successfully.

Tested using Docker on Windows 11 and Amazon ECS.


mtesauro (approved):
Approved

Maffooch (approved):

cneill (approved):

PR #9804, opened 2024-03-21
Tenable Parser: Support the new "workbench" format

FelixHernandez:
## Tenable Parser: Support the new "workbench" format
[sc-1532]

**Description**

Now the Tenable parser, is supporting the new workbench format.

- The CSV is deliniated with semicolons rather than commas
- Every single column name is different
- columns have “dot notation” to create the notion of full objects within the CSV rather than the fields of a single object. Example being “asset.network.name”

**Test results**
unittests were added and ran Ok.


grendel513 (approved):

mtesauro (approved):
Approved

Maffooch (approved):

cneill (approved):

PR #9805, opened 2024-03-21
Fix finding filter in API `not_test__tags'

FelixHernandez:
## Fix finding filter in API `not_test__tags'
[sc-4929]

The Finding filters `not_test__tags` and `test_tags` was returning the same dataset when used on a GET /finding/ request.

**Test results**

The `exclude='True'` was missing in the `not_test__tags` field in API Finding filter. The exclude parameter was added.
Now the filter works properly.


mtesauro (approved):
Approved

Maffooch (approved):

grendel513 (approved):

cneill (approved):

PR #9810, opened 2024-03-22
remove xlrd

manuel-sommer:
It looks like xlrd is not really used.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -35,7 +35,6 @@ lxml==5.1.0
 Markdown==3.5.2
 mysqlclient==2.1.1
 openpyxl==3.1.2
-xlrd==1.2.0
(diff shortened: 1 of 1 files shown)


kiblik:
I believe some tools can detect these obsolete packages. It might be a good idea to add it as one of GitHub checks

manuel-sommer:
This is also what I thought, but did not dive deeper regarding such tools and GH checks

mtesauro:
I can only find this in requirements.txt and unit test tools output files so that's likely the reason we removed it's use.

mtesauro (approved):
Approved

Maffooch (approved):

cneill (approved):

grendel513 (approved):

PR #9819, opened 2024-03-25
Make tags optional for import and reimport API endpoints

FelixHernandez:
## Make tags optional for import and reimport API endpoints
[sc-4953]

**Description**

The `tags` field in Import and reimport Api endpoints are optional. The `TagListSerializerField` now validates if the given value is empty, if it is, and the field has the `allow_empty` option as True, then the validation returns an empty list, making the field as valid.

**Test results**
The tests were made manually


Maffooch (approved):

cneill (approved):

blakeaowens (approved):

mtesauro (approved):
Approved

PR #9830, opened 2024-03-26
API Import/Reimport: Convert string tags to lists

Maffooch:
When supplying tags to the import and reimport endpoints through the swagger page, the resulting curl command on the tags form data looks like this:
 ```
...
  -F 'tags=one,two,three'
...
```
When the serializer processes this request, a single string with the value `one,two,three` will be applied to any objects requested to be tagged. This is misleading and incorrect. Instead, the string should be split with commas as the delimiters.

[sc-4993]

### DIFF
--- a/unittests/test_tags.py
+++ b/unittests/test_tags.py
@@ -161,7 +161,6 @@ class TagTests(DojoAPITestCase):
         response = self.get_finding_tags_api(finding_id)
 
         self.assertEqual(2, len(response.get('tags')))
-        # print("response['tags']:" + str(response['tags']))
         self.assertIn('one', str(response['tags']))
         self.assertIn('two', str(response['tags']))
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: unittests/test_tags.py [hunk 1]
```diff
@@ -160,13 +160,7 @@ def test_finding_create_tags_with_commas(self):
...
 
-        # the old django-tagging library was splitting this tag into 2 tags
-        # with djangotagulous the tag does no longer get split up and we cannot modify tagulous
-        # to keep doing the old behaviour. so this is a small incompatibility, but only for
-        # tags with commas, so should be minor trouble
-        #
-        # self.assertEqual(2, len(response.get('tags')))
-        self.assertEqual(1, len(response.get('tags')))
+        self.assertEqual(2, len(response.get('tags')))
         # print("response['tags']:" + str(response['tags']))
```
mtesauro, line 164: We should drop this commented out print line

Maffooch: Sure will do

blakeaowens (approved):

cneill (approved):

mtesauro (approved):
Approved

PR #9831, opened 2024-03-27
Remove 'version' from docker-compose

kiblik:
Solve Warning
```bash
WARN[0000] .../django-DefectDojo/docker-compose.yml: `version` is obsolete 
WARN[0000] .../django-DefectDojo/docker-compose.override.yml: `version` is obsolete
```

More info:
> The top-level `version` property is defined by the Compose Specification for backward compatibility. It is only informative.
> 
> Compose doesn't use `version` to select an exact schema to validate the Compose file, but prefers the most recent schema when it's implemented.

From https://github.com/compose-spec/compose-spec/blob/master/04-version-and-name.md

### DIFF
--- a/docker-compose.override.debug.yml
+++ b/docker-compose.override.debug.yml
@@ -1,5 +1,4 @@
 ---
-version: '3.8'
 services:
   uwsgi:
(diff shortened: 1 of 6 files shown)


Maffooch (approved):

cneill (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #9744, opened 2024-03-15
correcting cvss scoring behavior when updating/adding findings via ui and api [sc-4849] 

grendel513:
CVSSv3 Vector takes precedence, followed by CVSSv3 Score, and finally Severity in that order

PR #9858, opened 2024-04-01
CVSS Override: Revert #9744

Maffooch:
CVSS overrides are creating issues for reimport due to the calculation of hash codes. This functionality needs to be reverted for the time being

[sc-5098]


grendel513 (approved):

mtesauro (approved):
Approved

cneill (approved):

PR #9873, opened 2024-04-03
:bug: jake json output in cyclonedx not parsed

manuel-sommer:
Exisiting jake.json in cyclonedx is not considered in the parser.


mtesauro:
@manuel-sommer  We should be checking that CycloneDX files from tool vendors are actually valid per the spec before adjusting the CycloneDX parser. I just had a conversation with Steve Springett (person who created CycloneDX) about vendor tool support and his any my experience agree that vendors say "we output ClycloneDX" but the really output "CycloneDX-ish" files.

If I look at the the JSON unit test file for Jake at https://cyclonedx.github.io/cyclonedx-web-tool/validate, I get:
image

If it's not valid CycloneDX, it's probably better to create a Jake specific parser rather than make our CycloneDX parser parse invalid CycloneDX.

That validator above was the first online one I found - it may not be the best but I still think the CycloneDX parser should probably only parse valid versions and have vendor-specific parsers when the tool provides invalid formats.

I'll ask around and see if there's a better CycloneDX validator than that one I found online.

grendel513:
> @manuel-sommer We should be checking that CycloneDX files from tool vendors are actually valid per the spec before adjusting the CycloneDX parser. I just had a conversation with Steve Springett (person who created CycloneDX) about vendor tool support and his any my experience agree that vendors say "we output ClycloneDX" but the really output "CycloneDX-ish" files.
> 
> If I look at the the JSON unit test file for Jake at https://cyclonedx.github.io/cyclonedx-web-tool/validate, I get: image
> 
> If it's not valid CycloneDX, it's probably better to create a Jake specific parser rather than make our CycloneDX parser parse invalid CycloneDX.
> 
> That validator above was the first online one I found - it may not be the best but I still think the CycloneDX parser should probably only parse valid versions and have vendor-specific parsers when the tool provides invalid formats.
> 
> I'll ask around and see if there's a better CycloneDX validator than that one I found online.

@manuel-sommer i suggest a specific parser for Jake be implemented instead.

mtesauro:
OK. I got curious and ran this official CycloneDX validator from https://github.com/CycloneDX/cyclonedx-cli against the unit test files at `django-DefectDojo/unittests/scans/cyclonedx`

```
$ for i in `ls *.json`
> do
> echo "Checking $i"
> ../../../../cyclonedx validate --input-file "$i"
> done
Checking cyclonedx_cwe.json
Unable to validate against any JSON schemas.
BOM is not valid.
Checking issue_8022.json
BOM validated successfully.
Checking issue_9277.json
BOM validated successfully.
Checking jake2.json
BOM validated successfully.
Checking jake.json
Unable to validate against any JSON schemas.
BOM is not valid.
Checking log4j.json
BOM validated successfully.
Checking nested-component-log4j.json
BOM validated successfully.
Checking valid-vulnerability-1.4.json
BOM validated successfully.
```

So the net result of that was these are valid:

* issue_8022.json
* issue_9277.json
* jake2.json
* log4j.json
* nested-component-log4j.json
* valid-vulnerability-1.4.json

And these are invalid:

* cyclonedx_cwe.json
* jake.json

I also repeated for XML with these results (all XML unit test files are valid)

* cyclonedx_bom.xml
* dd_1_15_0.xml
* grype_dd_1_14_1.xml
* grype.xml
* jake.xml
* log4j.xml
* retirejs.latest.xml
* spec1_lowfirst.xml
* spec1.xml
* valid-vulnerability-1.4.xml

mtesauro:
@manuel-sommer So I love that you were fixing an issue with Jake's CycloneDX parser (Thanks for that :pray: ) but it looks like the right path forward is to:

(A) Look at jake.json and figure out how we got an invalid CycloneDX file as a unit-test file. I'd not spend a bunch of time on that but whatever is discovered either:
1. Remove jake.json and update related unit tests to use a valid file
2.  Run [Jake](https://github.com/sonatype-nexus-community/jake) against something to generate a valid jake.json file and update the unit test to use this newly created json output. 

2 is the best since we're sure we're getting recent & valid output from Jake. I have no clue how old/state jake.json is but that's one possibility.

(B) Remove the invalid json files and update the unit tests to use valid files instead.

Ideally, you could use the file generated in 2 above to replace the invalid ones as well. I've not had a chance to look at what unit tests use those invalid files but they'll need to be updated once the invalid JSON files are removed so they keep passing when getting called.

manuel-sommer:
I removed jake.json. I don't think I will make a PR to support jake.json as there also exists a jake2.json file which is valid. Maybe, jake.json was an outdated version of jake.

Regarding cyclonedx_cwe.json I was not able to detect on why the schema is invalid. Furthermore, it would be an option to include cyclonedx-cli to validate cyclonedx directly on the input, but it has no python package. I found https://pypi.org/project/cyclonedx-editor-validator/ but that validator is inconsistent and does not look like being stable.

mtesauro:
@manuel-sommer 
> I removed jake.json. I don't think I will make a PR to support jake.json as there also exists a jake2.json file which is valid. Maybe, jake.json was an outdated version of jake.
> 
> Regarding cyclonedx_cwe.json I was not able to detect on why the schema is invalid. Furthermore, it would be an option to include cyclonedx-cli to validate cyclonedx directly on the input, but it has no python package. I found https://pypi.org/project/cyclonedx-editor-validator/ but that validator is inconsistent and does not look like being stable.

Understandable about not making a separate Jake parser based on that invalid jake.json file.

I also didn't find a good Python equivalent to cyclonedx-cli in Python. Something we can do in the future.

So this PR now just removes that invalid jake.json file :+1:

mtesauro:
@manuel-sommer Before we approve & merge this PR, would you mind removing the cyclonedx_cwe.json file and it's matching unit test so all the invalid CycloneDX files are removed in this PR?

See https://github.com/search?q=repo%3ADefectDojo%2Fdjango-DefectDojo%20cyclonedx_cwe.json&type=code

manuel-sommer:
Done @mtesauro

mtesauro (approved):
Approved

grendel513 (approved):

blakeaowens (approved):

PR #9874, opened 2024-04-03
:hammer: restructure json in scout suite unittests

manuel-sommer:


cneill:
While you're tidying up these files, could you also rename them to `.json` instead of `.js`?

manuel-sommer:
> While you're tidying up these files, could you also rename them to `.json` instead of `.js`?

The problem is then, that it looks like the content itself is invalid due to the first line `scoutsuite_results =` as it doesn't follow the json standard. Shall I do it nevertheless?

cneill:
> > While you're tidying up these files, could you also rename them to `.json` instead of `.js`?
> 
> The problem is then, that it looks like the content itself is invalid due to the first line `scoutsuite_results =` as it doesn't follow the json standard. Shall I do it nevertheless?

Apologies, I missed that bit in the big diff and didn't realize the parser was referencing a JS file produced as part of an HTML report.

cneill (approved):

mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

PR #9882, opened 2024-04-04
PreDjango 4.2 fixes

kiblik:
Set of changes that will be needed during the upgrade to Django 4.2 and it is safe to perform them in 4.1.

To decrease the migration scope of #9493


mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

PR #9885, opened 2024-04-05
Use Postgres in rest-framework-tests

kiblik:


Maffooch:
I finally got a chance to look into this for a few today and am seeing some weird things. The reasoning for unexpected failures is the order In which tests are running. Generally speaking, one test should not have any impact on another, but that does not appear to be the case here.

When I isolate the `TestTests` class, and ran it about 10 times, there were no failures observed. However, when running in the presence of other classes, failures were very sporadic. It almost seems as if one class of tests is deleting elements at a similar time that the `TestTests` class is looking for a preview of what is to be deleted. This is all speculation, so I cannot really say for sure if this the issue or not. @kiblik what do you think about all of this?

Maffooch (approved):
@kiblik I think this one can go out as it is a blocker for the Django 4.2 upgrade. If tests become unstable from this, we can asses then


Do you agree?

grendel513 (approved):

mtesauro (approved):
Approved

kiblik:
Sorry @Maffooch, I missed your comment.

It looks like tests are not written "independent enough". See #10335. This fact was probably hidden and it became visible only because so different behavior of databases (still fully do not understand why).
It is probably reason why `--parallel` does not work https://github.com/DefectDojo/django-DefectDojo/blob/67a75710d6991a3ee3d4ed8dca7d0ef5c9cef46a/docker/entrypoint-unit-tests.sh#L74-L75

I agree to merge this PR. It is not the best to agree when we do not fully understand what is happening.

However, I suppose we should start also process of dipper audit of tests.

kiblik:
It looks like the last run of tests in this PR was a coincidence.
New and rebased PRs are failing right now (with the same error): https://github.com/DefectDojo/django-DefectDojo/actions/runs/9412741301/job/25928361674?pr=10086

Maffooch:
Yeah I think you are right... I am spending some time looking into these tests to see what we can do

PR #9889, opened 2024-04-08
Ruff: add and fix C4

kiblik:
Fix https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -2949,7 +2949,7 @@ def report_generate(request, obj, options):
         )
         report_name = "Engagement Report: " + str(engagement)
 
-        ids = set(finding.id for finding in findings.qs)
+        ids = set(finding.id for finding in findings.qs)  # noqa: C401
         ids = get_endpoint_ids(
             Endpoint.objects.filter(product=engagement.product).distinct()
         )
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -2692,7 +2692,7 @@ def finding_bulk_update_all(request, pid=None):
         finding_to_update = request.POST.getlist("finding_to_update")
         finds = Finding.objects.filter(id__in=finding_to_update).order_by("id")
         total_find_count = finds.count()
-        prods = set([find.test.engagement.product for find in finds])
+        prods = set(find.test.engagement.product for find in finds)  # noqa: C401
         if request.POST.get("delete_bulk_findings"):
             if form.is_valid() and finding_to_update:
                 if pid is not None:
@@ -2996,9 +2996,9 @@ def finding_bulk_update_all(request, pid=None):
 
                 error_counts = defaultdict(lambda: 0)
                 success_count = 0
-                finding_groups = set(
-                    [find.finding_group for find in finds if find.has_finding_group]
-                )
+                finding_groups = {
+                    find.finding_group for find in finds if find.has_finding_group
+                }  # noqa: C403
                 logger.debug("finding_groups: %s", finding_groups)
                 groups_pushed_to_jira = False
                 for group in finding_groups:
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -2996,9 +2996,9 @@ def finding_bulk_update_all(request, pid=None):
 
                 error_counts = defaultdict(lambda: 0)
                 success_count = 0
-                finding_groups = {
+                finding_groups = set(  # noqa: C401
                     find.finding_group for find in finds if find.has_finding_group
-                }  # noqa: C403
+                )
                 logger.debug("finding_groups: %s", finding_groups)
                 groups_pushed_to_jira = False
                 for group in finding_groups:
--- a/dojo/importers/importer/importer.py
+++ b/dojo/importers/importer/importer.py
@@ -257,7 +257,7 @@ class DojoDefaultImporter:
                 old_finding.save(dedupe_option=False, push_to_jira=push_to_jira)
 
         if is_finding_groups_enabled() and push_to_jira:
-            for finding_group in set([finding.finding_group for finding in old_findings if finding.finding_group is not None]):
+            for finding_group in set(finding.finding_group for finding in old_findings if finding.finding_group is not None):  # noqa: C401
                 jira_helper.push_to_jira(finding_group)
 
         return old_findings
--- a/dojo/importers/reimporter/reimporter.py
+++ b/dojo/importers/reimporter/reimporter.py
@@ -451,12 +451,10 @@ class DojoDefaultReImporter:
                     jira_helper.push_to_jira(findings[0])
 
         if is_finding_groups_enabled() and push_to_jira:
-            for finding_group in set(
-                [
-                    finding.finding_group
-                    for finding in reactivated_items + unchanged_items
-                    if finding.finding_group is not None and not finding.is_mitigated
-                ]
+            for finding_group in set(  # noqa: C401
+                finding.finding_group
+                for finding in reactivated_items + unchanged_items
+                if finding.finding_group is not None and not finding.is_mitigated
             ):
                 jira_helper.push_to_jira(finding_group)
 
@@ -540,12 +538,10 @@ class DojoDefaultReImporter:
                 mitigated_findings.append(finding)
 
         if is_finding_groups_enabled() and push_to_jira:
-            for finding_group in set(
-                [
-                    finding.finding_group
-                    for finding in to_mitigate
-                    if finding.finding_group is not None
-                ]
+            for finding_group in set(  # noqa: C401
+                finding.finding_group
+                for finding in to_mitigate
+                if finding.finding_group is not None
             ):
                 jira_helper.push_to_jira(finding_group)
 
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1656,7 +1656,7 @@ class Endpoint(models.Model):
         if self.path or self.path == '':
             while len(self.path) > 0 and self.path[0] == "/":  # Endpoint store "root-less" path
                 self.path = self.path[1:]
-            if any([null_char in self.path for null_char in null_char_list]):
+            if any(null_char in self.path for null_char in null_char_list):
                 old_value = self.path
                 if 'postgres' in db_type:
                     action_string = 'Postgres does not accept NULL character. Attempting to replace with %00...'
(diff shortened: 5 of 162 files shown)


Maffooch (approved):

dojo/api_v2/views.py
```diff
@@ -2945,7 +2945,7 @@ def report_generate(request, obj, options):
         )
         report_name = "Engagement Report: " + str(engagement)
 
-        ids = set(finding.id for finding in findings.qs)
+        ids = {finding.id for finding in findings.qs}
```
cneill, line 2948: Switching to this form means you have to look for a `:` in the comprehension to determine whether it's producing a `set` or a `dict`. Speaking as someone who no longer writes Python every day, I don't think this rule (`C401`) actually increases code clarity and I feel like we should disable it.

dojo/finding/views.py
```diff
@@ -2675,7 +2675,7 @@ def finding_bulk_update_all(request, pid=None):
         finding_to_update = request.POST.getlist("finding_to_update")
         finds = Finding.objects.filter(id__in=finding_to_update).order_by("id")
         total_find_count = finds.count()
-        prods = set([find.test.engagement.product for find in finds])
+        prods = {find.test.engagement.product for find in finds}
```
cneill, line 2678: Similar to my comment in `dojo/api_v2/views.py`, I think that this form is less clear and that we can do without the associated rule (`C403`).

dojo/finding/views.py
```diff
@@ -2979,9 +2979,9 @@ def finding_bulk_update_all(request, pid=None):
 
                 error_counts = defaultdict(lambda: 0)
                 success_count = 0
-                finding_groups = set(
-                    [find.finding_group for find in finds if find.has_finding_group]
-                )
+                finding_groups = {
+                    find.finding_group for find in finds if find.has_finding_group
+                }
```
cneill, line 2984: Same comment as above

dojo/models.py
```diff
@@ -2684,24 +2684,16 @@ def get_vulnerability_ids(self):
...
         else:
             vulnerability_ids = Vulnerability_Id.objects.filter(finding=self)
             deduplicationLogger.debug("get_vulnerability_ids after the finding was saved. Vulnerability references count: " + str(vulnerability_ids.count()))
             # convert list of vulnerability_ids to the list of their canonical representation
-            vulnerability_id_str_list = list(
-                map(
-                    lambda vulnerability_id: str(vulnerability_id),
-                    vulnerability_ids.all()
-                ))
+            vulnerability_id_str_list = [str(vulnerability_id) for vulnerability_id in vulnerability_ids.all()]
```
cneill, line 2696: This is a good example of *increasing* code clarity, which I strongly support 😄

dojo/product/views.py
```diff
@@ -574,7 +574,7 @@ def view_product_metrics(request, pid):
 
     product_tab = Product_Tab(prod, title=_("Product"), tab="metrics")
 
-    open_objs_by_age = {x: len([_ for _ in filters.get('open') if _.age == x]) for x in set([_.age for _ in filters.get('open')])}
+    open_objs_by_age = {x: len([_ for _ in filters.get('open') if _.age == x]) for x in {_.age for _ in filters.get('open')}}
```
cneill, line 577: This double comprehension is an example of when removing `set()` just makes the code more difficult to read, IMO.

dojo/importers/importer/importer.py
```diff
@@ -257,7 +257,7 @@ def close_old_findings(self, test, scan_date_time, user, push_to_jira=None, serv
                 old_finding.save(dedupe_option=False, push_to_jira=push_to_jira)
 
         if is_finding_groups_enabled() and push_to_jira:
-            for finding_group in set([finding.finding_group for finding in old_findings if finding.finding_group is not None]):
+            for finding_group in {finding.finding_group for finding in old_findings if finding.finding_group is not None}:
```
cneill, line 260: Same  comment as above

dojo/importers/reimporter/reimporter.py
```diff
@@ -451,13 +451,11 @@ def process_parsed_findings(
                     jira_helper.push_to_jira(findings[0])
 
         if is_finding_groups_enabled() and push_to_jira:
-            for finding_group in set(
-                [
+            for finding_group in {
```
cneill, line 454: Same comment as above

dojo/importers/reimporter/reimporter.py
```diff
@@ -540,13 +538,11 @@ def close_old_findings(
                 mitigated_findings.append(finding)
 
         if is_finding_groups_enabled() and push_to_jira:
-            for finding_group in set(
-                [
+            for finding_group in {
```
cneill, line 541: Same comment as above

dojo/reports/views.py
```diff
@@ -435,7 +435,7 @@ def generate_report(request, obj, host_view=False):
         report_title = "Product Report"
         findings = ReportFindingFilter(request.GET, product=product, queryset=prefetch_related_findings_for_report(Finding.objects.filter(
             test__engagement__product=product)))
-        ids = set(finding.id for finding in findings.qs)
+        ids = {finding.id for finding in findings.qs}
```
cneill, line 438: Same comment as above

dojo/reports/views.py
```diff
@@ -466,7 +466,7 @@ def generate_report(request, obj, host_view=False):
         template = 'dojo/engagement_pdf_report.html'
         report_title = "Engagement Report"
 
-        ids = set(finding.id for finding in findings.qs)
+        ids = {finding.id for finding in findings.qs}
```
cneill, line 469: Same comment as above

dojo/utils.py
```diff
@@ -1357,7 +1357,7 @@ def reopen_external_issue(find, note, external_issue_provider, **kwargs):
 def process_notifications(request, note, parent_url, parent_title):
     regex = re.compile(r'(?:\A|\s)@(\w+)\b')
 
-    usernames_to_check = set([un.lower() for un in regex.findall(note.entry)])
+    usernames_to_check = {un.lower() for un in regex.findall(note.entry)}
```
cneill, line 1360: Same comment as above

unittests/test_rest_framework.py
```diff
@@ -1130,7 +1130,7 @@ def test_duplicate(self):
         assert result.status_code == status.HTTP_200_OK, "Could not check duplicate status"
         result_json = result.json()
         # Should return all duplicates for id=3
-        assert set(x["id"] for x in result_json) == {2, 4, 5, 6}
+        assert {x["id"] for x in result_json} == {2, 4, 5, 6}
```
cneill, line 1133: Same comment as above

cneill (changes requested):
I'm fine with most of these changes (particularly the removal of unnecessary `lambda`s). It's debatable whether `[]` and `{}` are clearer than `list()` and `dict()`, but the *consistency* of using the former is probably worth it. However, I would revert the ones that change `set([])` and `set( ... generator ... )`s into `{}` form and disable those rules ([`C401`](https://docs.astral.sh/ruff/rules/unnecessary-generator-set/) and [`C403`](https://docs.astral.sh/ruff/rules/unnecessary-list-comprehension-set/)). I could go either way on [`C405`](https://docs.astral.sh/ruff/rules/unnecessary-literal-set/) - I don't think it's strictly *necessary*, but the enforced consistency is probably worthwhile for set literals.

Open to feedback from others on this question, but linter rules are inherently opinionated and these felt wrong to me.

dojo/api_v2/views.py
mtesauro: I agree with cneil on this one - seems to make code harder to understand at the expense of a couple of characters. That feels like a bad trade-off to me.

dojo/finding/views.py
mtesauro: :+1:

dojo/product/views.py
mtesauro: Agreed from my perspective

dojo/reports/views.py
mtesauro: :+1:

dojo/importers/reimporter/reimporter.py
mtesauro: :+1:

dojo/reports/views.py
mtesauro: :+1:

dojo/utils.py
mtesauro: :+1:

unittests/test_rest_framework.py
mtesauro: :+1:

kiblik:
> I'm fine with most of these changes (particularly the removal of unnecessary `lambda`s). It's debatable whether `[]` and `{}` are clearer than `list()` and `dict()`, but the _consistency_ of using the former is probably worth it. However, I would revert the ones that change `set([])` and `set( ... generator ... )`s into `{}` form and disable those rules ([`C401`](https://docs.astral.sh/ruff/rules/unnecessary-generator-set/) and [`C403`](https://docs.astral.sh/ruff/rules/unnecessary-list-comprehension-set/)). I could go either way on [`C405`](https://docs.astral.sh/ruff/rules/unnecessary-literal-set/) - I don't think it's strictly _necessary_, but the enforced consistency is probably worthwhile for set literals.
> 
> Open to feedback from others on this question, but linter rules are inherently opinionated and these felt wrong to me.

I had no opinion about it. But thank you for the feedback. Now I see your point and I agree.
I added `noqa`s to related lines to accept the used syntax.

dojo/finding/views.py
```diff
@@ -2979,9 +2979,9 @@ def finding_bulk_update_all(request, pid=None):
 
                 error_counts = defaultdict(lambda: 0)
                 success_count = 0
-                finding_groups = set(
-                    [find.finding_group for find in finds if find.has_finding_group]
-                )
+                finding_groups = {
+                    find.finding_group for find in finds if find.has_finding_group
+                }  # noqa: C403
```
cneill, line 2984: I think I may have missed this one in my initial comment yesterday

```suggestion
                finding_groups = set(
                    find.finding_group for find in finds if find.has_finding_group
                )  # noqa: C403
```

cneill (changes requested):
Thanks for revisiting these! 🙏  One minor tweak and this one should be ready to go

dojo/finding/views.py
kiblik: I do not know how but you are right. I missed this one. Thank you.

cneill (approved):

mtesauro (approved):
Approved

hblankenship (approved):

PR #9903, opened 2024-04-09
Ruff: Fix UP (after couple of merges)

kiblik:
After combination of set of MRs, new parts need to be fixed.

PR #9913, opened 2024-04-10
fix engagements filters in 'engagements by product view'

davidhernandeze:
sc-[5299]

## Description
Engagements filters inside the "Engagements By Product" view are not working properly. Searching by the exact name in the "Engagement name" filter is returning incorrect results.

## The Problem
Building relationship filters as we currently do in `EngagementFilter` generates a query that gets _filtered products with all the related engagements, where at least one engagement satisfies the filter criteria_.  The `engagements_all` view iterates over the engagements of each product to render the table rows. As the result, we ended up showing all engagements of the products that satisfied the filter criteria. For example, if there are two engagements that satisfied the filter criteria from two different products, the rendered table will show all engagements that belong to those two products.

## Solution
Using an additional filter when prefetching the engament_set will filter the engagements at the query level, so only _filtered engagements_ related to _filtered products_ will be obtained.

``` python
engagement_query = Engagement.objects.annotate(test_count=Count('test__id'))
filter_qs = products_with_engagements.prefetch_related(
   Prefetch('engagement_set', queryset=ProductEngagementsFilter(request.GET, engagement_query).qs)
)
```

Also, the paginator object had to be modified, to represent the correct amount of rows that are currently shown in the page.

``` python
prods.paginator.count = sum(len(prod.engagement_set.all()) for prod in prods)
```


davidhernandeze:
sc-[5299]

mtesauro:
@davidhernandeze You don't have to fix the Ruff linter errors, those were fixed in #9903 and outside the changes you've made here.

mtesauro (approved):
Approved

Maffooch (approved):

davidhernandeze:
I'm sorry. I made a mistake using my `dev` branch, which was causing me problems opening more pull requests using my fork. I will duplicate this PR with the correct branching strategy.

PR #9929, opened 2024-04-15
Remove pyproject.toml, add ruff.toml

cneill:
**Description**

Since we don't currently use `pyproject.toml` for its intended purpose of describing the entire DefectDojo application and its dependencies, but only to configure the `ruff` linter, this PR switches to using [the `ruff.toml` file](https://docs.astral.sh/ruff/configuration/) to configure `ruff`. This will prevent confusion with tools that might try and fail to parse our current `pyproject.toml` file.

**Test results**

This appears to work as expected when running `ruff` 0.3.7 locally - no linting issues reported.

[sc-5411]


mtesauro (approved):
Approved

Maffooch (approved):

grendel513 (approved):

blakeaowens (approved):

Issue #9939, opened 2024-04-17
A string literal cannot contain NUL (0x00) characters


quyen66:
While import Horusec scan result I got some error `A string literal cannot contain NUL (0x00) characters.`, Maybe it's about the value in json report file.

**Deployment method** *(select with an `X`)*
- [x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Ubuntu 22.04
 - DefectDojo version v. 2.30.2 ( release mode ) 

**Logs** 
Logs in uwsgi show

> [pid: 35|app: -|req: -/-] 172.16.251.9 (admin_soc) {52 vars in 856 bytes} [Wed Apr 17 02:08:24 2024] GET /alerts/count => generated 12 bytes in 37 msecs (HTTP/1.1 200) 7 headers in 212 bytes (1 switches on core 3)
[pid: 31|app: -|req: -/-] 172.16.251.9 (admin_soc) {52 vars in 856 bytes} [Wed Apr 17 02:08:35 2024] GET /alerts/count => generated 12 bytes in 38 msecs (HTTP/1.1 200) 7 headers in 212 bytes (1 switches on core 0)
[17/Apr/2024 02:08:40] ERROR [dojo.engagement.views:751] A string literal cannot contain NUL (0x00) characters.
Traceback (most recent call last):
  File "/app/dojo/engagement/views.py", line 735, in post
    test, finding_count, closed_finding_count, _ = importer.import_scan(scan, scan_type, engagement, user, environment, active=active, verified=verified, tags=tags,
                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/importers/importer/importer.py", line 345, in import_scan
    new_findings = self.process_parsed_findings(test, parsed_findings, scan_type, user, active=active,
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/decorators.py", line 48, in __wrapper__
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #9941, opened 2024-04-17
fix horusec null characters issue, #9939

manuel-sommer:
see issue #9939

### DIFF
--- a/dojo/tools/horusec/parser.py
+++ b/dojo/tools/horusec/parser.py
@@ -62,7 +62,7 @@ class HorusecParser(object):
                 data["vulnerabilities"]["details"].split("\n")[-1],
                 "**Code:**",
                 f"```{data['vulnerabilities']['language']}",
-                data["vulnerabilities"]["code"].replace("```", "``````"),
+                data["vulnerabilities"]["code"].replace("```", "``````").replace("\x00", ""),
                 "```",
             ]
         )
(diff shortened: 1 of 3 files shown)


Maffooch (approved):

grendel513 (approved):

blakeaowens (approved):

dojo/tools/horusec/parser.py
```diff
@@ -62,7 +62,7 @@ def _get_finding(self, data, date):
                 data["vulnerabilities"]["details"].split("\n")[-1],
                 "**Code:**",
                 f"```{data['vulnerabilities']['language']}",
-                data["vulnerabilities"]["code"].replace("```", "``````"),
+                data["vulnerabilities"]["code"].replace("```", "``````").replace("\x00", ""),
```
cneill, line 65: Here's what this Finding looks like today if you just remove these null characters:

Screenshot 2024-04-17 at 12 57 17

A couple thoughts:
* Ideally, we shouldn't throw away data reported by the tool like this - nulls might be meaningful sometimes (though they're not in this case as this seems like an obvious false positive, and they wouldn't be displayed anyway)
* Displaying random binary characters isn't the ideal behavior here, and this probably isn't the only tool where binary data could appear like this

I'm okay with approving this one for now since it fixes an open issue, but it seems like there might be a better approach to handling binary data like this more generally, such as using the `\uXXXX` formatting like the JSON example. Unfortunately, as this case demonstrates, it's not always obvious that a given field in a tool's report will *sometimes* contain binary data.

cneill (approved):

dojo/tools/horusec/parser.py
manuel-sommer: Good hint, thank you @cneill

PR #9961, opened 2024-04-18
RemoteUser: Hide from Swagger

kiblik:
`RemoteUser` is usually used behind AuthN proxy and users should not know about this mechanism from Swagger because it is not usable by users.
It should be hidden by default.


mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

cneill (approved):

Issue #9942, opened 2024-04-17
A string literal cannot contain NUL (0x00) characters - Anchore Grype


corvusmod:
**Bug description**
While import Anchore Grype scan result I got some error 
A string literal cannot contain NUL (0x00) characters., Maybe it's about the value in json report file.

Similar to: https://github.com/DefectDojo/django-DefectDojo/issues/9939

**Steps to reproduce**
Imported an error with dd-importer

**Expected behavior**
No error importing Anchore Grype reports

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version 2.27.4

**Logs** 

`
[pid: 1|app: -|req: -/-] 10.20.21.225 (-) {50 vars in 835 bytes} [Wed Apr 17 12:35:14 2024] GET //api/v2/tests/?title=Anchore+Grype&engagement=2209 => generated 607 bytes in 39 msecs (HTTP/1.1 200) 8 headers in 246 bytes (1 switches on core 0)
[17/Apr/2024 12:35:14] ERROR [dojo.api_v2.exception_handler:36] A string literal cannot contain NUL (0x00) characters.
Traceback (most recent call last):
  File "/app/dojo/api_v2/serializers.py", line 2502, in save
    ) = reimporter.reimport_scan(
        ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/importers/reimporter/reimporter.py", line 682, in reimport_scan
    ) = self.process_parsed_findings(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/decorators.py", line 48, in __wrapper__
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/celery/local.py", line 182, in __call__
    return self._get_current_object()(*a, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/celery/app/task.py", line 411, in __call__
    return self.run(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/importers/reimporter/reimporter.py", line 332, in process_parsed_findings
    item.save(dedupe_option=False)
  File "/app/dojo/models.py", line 2915, in save
[...]

PR #9962, opened 2024-04-18
fix anchore_grype null characters issue, #9942

manuel-sommer:
see #9942


manuel-sommer:
I tried to replace it with more meaningful, but was not successful. I am open to any hints.

Maffooch (approved):

cneill (approved):
This is an odd one... no idea why they're sticking nulls in the `version` field, but just discarding them seems like the right approach in this case.

mtesauro (approved):
Approved

blakeaowens (approved):

PR #9966, opened 2024-04-19
Helm postgresql: Upgrade image

kiblik:
Pin newer image until there is no psql migration procedure prepared.


mtesauro (approved):
Approved

PR #9967, opened 2024-04-19
Make the number of request/response pairs returned by the API configurable

hblankenship:
**Description**
[sc-5238]
Implemented a new variable, DD_MAX_REQRESP_FROM_API, to allow a limit to the number of request/response pairs that are returned via the API. The default value is -1, meaning no limit to maintain current functionality.

**Test results**

Tested with multiple values for MAX_REQRESP_FROM_API: 5, 10, 18, 100, -1, -100
See included unittest for more
All tests pass

**Documentation**

Updated the DOCKER.md file to include updated instructions for running single or more tests in the container


Maffooch (approved):

cneill (changes requested):
Just a minor comment to make the environment variable's purpose a little clearer. I'm also not sure it makes sense to have a `0` value return 0 results here - I doubt anyone would want to force returning 0 req/resp pairs globally, but I'll leave that up to you to decide.

mtesauro (approved):
Approved

blakeaowens (approved):

cneill (approved):

Issue #9911, opened 2024-04-10
Yarn Audit v2 and later version


egorbeliy:
**Scanner Name**
Yarn audit v2+
yarn npm audit --json -R -A
https://yarnpkg.com/cli/npm/audit

kiblik:
Can you provide an example report file, please?

egorbeliy:
Sample of logs:

{"value":"@babel/plugin-proposal-class-properties","children":{"ID":"@babel/plugin-proposal-class-properties (deprecation)","Issue":"This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-class-properties instead.","Severity":"moderate","Vulnerable Versions":"7.18.6","Tree Versions":["7.18.6"],"Dependents":["jscodeshift@virtual:95986a29f66ea5c154da709639e46d9f25ab769cdc542a6076b371e193f79e407bbee37bc9d3845bfa503bd700408966c85b5a74356facf4da9a113fd4ce89d9#npm:0.14.0"]}}
{"value":"@babel/plugin-proposal-nullish-coalescing-operator","children":{"ID":"@babel/plugin-proposal-nullish-coalescing-operator (deprecation)","Issue":"This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-nullish-coalescing-operator instead.","Severity":"moderate","Vulnerable Versions":"7.18.6","Tree Versions":["7.18.6"],"Dependents":["jscodeshift@virtual:95986a29f66ea5c154da709639e46d9f25ab769cdc542a6076b371e193f79e407bbee37bc9d3845bfa503bd700408966c85b5a74356facf4da9a113fd4ce89d9#npm:0.14.0"]}}
{"value":"@babel/plugin-proposal-optional-chaining","children":{"ID":"@babel/plugin-proposal-optional-chaining (deprecation)","Issue":"This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-optional-chaining instead.","Severity":"moderate","Vulnerable Versions":"7.21.0","Tree Versions":["7.21.0"],"Dependents":["jscodeshift@virtual:95986a29f66ea5c154da709639e46d9f25ab769cdc542a6076b371e193f79e407bbee37bc9d3845bfa503bd700408966c85b5a74356facf4da9a113fd4ce89d9#npm:0.14.0"]}}

[...]

PR #9985, opened 2024-04-22
:sparkles: implement yarn2 parser

manuel-sommer:
see #9911


mtesauro (approved):
Approved

dojo/tools/yarn_audit/parser.py
```diff
@@ -44,6 +48,38 @@ def get_items_yarn(self, tree, test):
                 )
         return list(items.values())
 
+    def get_items_yarn2(self, tree, test):
+        items = []
+        for element in tree:
+            value = "**value:** " + str(element.get("value"))
```
cneill, line 54: I don't think we should include Markdown formatting in `component_name` like this. Additionally, this will set the component name to "None" instead of leaving it empty if there's no `value`. I'd check if `value` is non-empty, and if so, include the `str()` value as-is.

dojo/tools/yarn_audit/parser.py
```diff
@@ -44,6 +48,38 @@ def get_items_yarn(self, tree, test):
...
+            child_dependents = ', '.join(set(child.get("Dependents")))
+            description += childissue + "\n"
+            description += "**Vulnerable Versions:** " + child_vuln_version + "\n"
+            description += "**Dependents:** " + child_dependents + "\n"
+            dojo_finding = Finding(
+                title=str(childid),
+                test=test,
+                severity=self.severitytranslator(severity=childseverity),
+                description=description,
+                component_name=str(value),
```
cneill, line 71: This `str()` is unnecessary

```suggestion
                component_name=value,
```

cneill (changes requested):
Just a couple comments around `component_name` here. It would also be nice to include the `URL` key in `references` if it's available.

cneill (approved):

egorbeliy (approved):

kostiq (approved):

egorbeliy:
anyone can check it to be approved?

Maffooch (approved):

blakeaowens (approved):

PR #10028, opened 2024-04-25
Better separation of vulnerabilieties in tenable scan 

WojTecH94:
No longer aggregate different endpoints into single vulnerability because it complicates finding management and reporting. More info:
https://github.com/DefectDojo/django-DefectDojo/discussions/9669

(2nd PR approach because it appears that changes in xml_parser cause too many unittests failures and it have to be done separately)

### DIFF
--- a/dojo/tools/tenable/csv_format.py
+++ b/dojo/tools/tenable/csv_format.py
@@ -104,12 +104,25 @@ class TenableCSVParser(object):
             mitigation = str(row.get("Solution", row.get("definition.solution", "N/A")))
             impact = row.get("Description", row.get("definition.description", "N/A"))
             references = row.get("See Also", row.get("definition.see_also", "N/A"))
+            # Endpoint related fields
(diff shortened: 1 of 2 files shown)


lme-nca:
Please dont merge this! imho the way the parser works now is wanted behaviour and we rely on it, and as already mentioned you could use the https://github.com/DefectDojo/django-DefectDojo/discussions/9669#discussioncomment-8669293 view for this use case.

WojTecH94:
Here is my point of view https://github.com/DefectDojo/django-DefectDojo/discussions/9669#discussioncomment-9238369

mtesauro (changes requested):
Keeping this from being merged while discussion is ongoing

manuel-sommer (changes requested):
Hi @WojTecH94,
could you take a look at this issue and fix it on the fly if the issue is a true positive reported bug? 
https://github.com/DefectDojo/django-DefectDojo/issues/10051

mtesauro:
Closed as stale

PR #10038, opened 2024-04-26
Filtering Performance: Add opt-in setting for converting to string ba…

Maffooch:
Today, the finding filters will query all products type, products, engagements, and all tags for all items that can be tagged. When loading the all finding pages, this will end up querying a large portion of the database.

This PR adds an opt in setting in the system settings to convert these filter options to string fields in  order to avoid making these queries when they are not specifically used 

[sc-5650]


blakeaowens (approved):

mtesauro (approved):
Approved

Issue #8127, opened 2023-05-19
Filters using numbers as strings


ibanezgomez:
Dear DefectDojo Team,

I hope this message finds you well. I wanted to bring to your attention a bug that I encountered while using the DefectDojo open source program. This bug pertains to the incorrect sorting of vulnerabilities listed under the product filters.

Bug Description:
When sorting the vulnerabilities in ascending or descending order, the program treats the number of vulnerabilities as text instead of an integer. As a result, the sorting functionality does not display the correct order of vulnerabilities.

Steps to Reproduce:

Log in to the DefectDojo program.
Go to the "Product" section.
Apply sorting in ascending or descending order based on the number of vulnerabilities.
Observe that the sorting does not display the correct order.
Expected Behavior:
The program should interpret the number of vulnerabilities as an integer and display the correct sorting order when applying ascending or descending sorting in the product filters.

Actual Behavior:
The program treats the number of vulnerabilities as text, leading to incorrect sorting of the vulnerabilities when applying ascending or descending sorting in the product filters.

I attach you a example:

<img width="1089" alt="image" src="https://github.com/DefectDojo/django-DefectDojo/assets/8628648/c494be09-2e73-4a80-9445-2e5a32ca2861">

Additional Information:

I have encountered this issue consistently on the latest version of DefectDojo.
This bug affects the usability of the program, as it hampers efficient vulnerability management.
Please let me know if you require any further information or if there is anything else I can do to assist in resolving this bug. DefectDojo is a valuable tool, and I appreciate the effort put into maintaining and improving it.

Thank you for your attention to this matter. I look forward to hearing from you and seeing this bug resolved in a future update.

Best regards,Samuel

andremralves:
Hey, I would like to take this issue. Do you have any other examples where you found an incorrect sorting? I'm willing to fix them as well.

manuel-sommer:
https://github.com/DefectDojo/django-DefectDojo/issues/6629

Issue #6629, opened 2022-07-29
'Active (Verified)' column sorts lexicographically instead of numerically in Engagements view


kdyck-cb:
### Description
When sorting by the *Active (Verified)* column in the various engagements lists, the items are sorted lexicographically (a.k.a alphabetically) instead of numerically. 

Here's an example taken from a 'Closed Engagements' list on the v 2.12.0 public demo environment
![image](https://user-images.githubusercontent.com/92729718/181757873-63f1d8b6-2400-495c-994f-9eb720850369.png)
[source](https://demo.defectdojo.org/product/1/engagements)

### Steps to reproduce

1. Go to the `/product/<product-id>/engagements` page.
2. Click on the *Active (Verified)* column on one of the engagement lists.
4. See that the items are sorted lexicographically instead of numerically.

### Expected behavior
Since the values in the *Active (Verified)* column are numeric, they should be sorted numerically, not lexicographically.

### Environment information

Observed in v 2.10.0 and v 2.12.0.

PR #10045, opened 2024-04-26
Fix numerical sorting in the old UI for Active/Verified findings

davidhernandeze:
[sc-5627]

As #6629 describes, there are views where sorting is made "lexicographically." The main problem I found is that sorting is performed by the front end (the page doesn't reload on sorting). Tables that don't perform sorting on the backend, like the ones displayed at `/product/{pid}/engagements,` will need another approach.

For the `All products` view, using a filter for findings_count on the backend allows us to sort numerically and sort over the complete dataset instead of being limited by the rows shown on the current page.

By implementing this change, we directly address the issue outlined in #8127, providing a more efficient sorting and filtering mechanism for the displayed data.

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -1182,6 +1182,7 @@ class ProductFilter(DojoFilter):
             ('origin', 'origin'),
             ('external_audience', 'external_audience'),
             ('internet_accessible', 'internet_accessible'),
+            ('findings_count', 'findings_count')
         ),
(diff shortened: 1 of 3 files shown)


Maffooch (approved):

cneill (approved):

blakeaowens (approved):

mtesauro (approved):
Approved

PR #10046, opened 2024-04-26
Fix engagements filters in 'engagements by product view' 

davidhernandeze:
It's the same as #9913 but with the correct branching.

sc-[5299]

## Description
The engagement filters inside the "Engagements By Product" view are not working properly. Searching by the exact name in the "Engagement name" filter returns incorrect results.

## The Problem
Building relationship filters as we currently do in `EngagementFilter` generates a query that gets _filtered products with all the related engagements, where at least one engagement satisfies the filter criteria_.  The `engagements_all` view iterates over the engagements of each product to render the table rows. As a result, we showed all product engagements that satisfied the filter criteria. For example, if two engagements fulfill the filter criteria from two different products, the rendered table will show all engagements that belong to those two products.

## Solution
When prefetching the engament_set, an additional filter will filter the engagements at the query level, so only _filtered engagements_ related to _filtered products_ will be obtained.

``` python
engagement_query = Engagement.objects.annotate(test_count=Count('test__id'))
filter_qs = products_with_engagements.prefetch_related(
   Prefetch('engagement_set', queryset=ProductEngagementsFilter(request.GET, engagement_query).qs)
)
```

Also, the paginator object had to be modified to represent the correct number of rows currently shown on the page.

``` python
prods.paginator.count = sum(len(prod.engagement_set.all()) for prod in prods)
```


Maffooch (approved):

davidhernandeze:
[sc-5299]

mtesauro (approved):
Approved

grendel513 (approved):

Maffooch (changes requested):
Please add support for filters without object lookups as well. Here is an example of doing so https://github.com/DefectDojo/django-DefectDojo/blob/f66e6dbb07bcddfb607ab81e76cca073080a6c2b/dojo/filters.py#L1065C1-L1093C75

hblankenship (approved):

Maffooch (approved):

PR #10048, opened 2024-04-27
Product Metrics: Performance Enhancements

blakeaowens:
[sc-5692]

This PR improves performance for Product metrics when there is a large number of Findings under a given DefectDojo Product. The following changes were made to improve performance:

### 1. Limiting Queries & Flattening of QuerySet(s)
Previously, the Product Metrics view relied on calls to the `finding_querys` utility. All of these queries have been instantiated at the beginning of the view as flattened lists, containing only the Finding field values required for metrics calculations. These original queries are then used throughout the Product metrics view (rather than relying on the `finding_querys` utility to fetch Findings over and over).

### 2. Removal of `in` Operator
The `in` operator in Python is time complexity `O(n)` on average (and called 3 times within the for-loop across all Findings). To eliminate this, three dictionaries in the format of `{ finding_id: True }` were constructed to eliminate the need to call Python's `in` operator. Instead, `finding in open_findings` can be replaced with `if open_findings.get(finding_id, None)`, which drastically improves performance.

### 3. Rework of `open_objs_by_age` Construction
The `open_objs_by_age` dataset, used to display Finding counts by age, was previously time complexity  `O(n^3)`. This performance bottleneck has been eliminated entirely by moving the construction of `open_objs_by_age` into an existing `O(n)` for-loop.

### 4. Removal of `|length` from Template
Previously, entire query sets were passed to the `product_metrics.html` template, then used to display total Finding counts in the headers/footers of plots using the `|length` display tag. These counts are now calculated in the view by taking the `len()` of the newly flattened lists mentioned in point `#1`.

Other minor organizational changes and cleanup made in addition to the 4 points above.


mtesauro (approved):
Approved

grendel513 (approved):

Maffooch (approved):

Maffooch (changes requested):
Holding until based on bugfix

blakeaowens:
Could not successfully rebase onto `bugfix` without introducing lots of unnecessary changes. New PR here: #10059

blakeaowens:
Could not successfully rebase onto `bugfix` without introducing lots of unnecessary changes. New PR here: #10059

PR #10066, opened 2024-04-30
Logic to add the user as owner for the group he creates

missy-tester:
**Description**

As per DefectDojo's documentation "If users create a new Group, they will be given the Owner role of the new Group by default. https://support.defectdojo.com/en/articles/8758189-about-permissions-roles#h_cd25e87a4e" This was not reflecting in DefectDojo's code. When a user created a new group, he was not added in the group at all.

Adding logic to add the user as owner whenever he creates the group.


mtesauro:
@missy-tester I'm don't believe this PR is needed. Using the public demo at https://demo.defectdojo.org (login details in the repo's README) I'm able to create a User Group and, in creating that group had "Group Owner" permissions as the documentation says. I doubled checked with the author of the page you linked above.

image


You effectively have to have superuser permissions to have full control of a groups membership and configuration but this is by design. We'll look at trying to clear up the documentation around these requirements but I don't see a need for the change in this PR.

I'm going to close this but feel free to provide follow-on comments if you feel I've misunderstood your goal with this PR and re-open if needed.

missy-tester:
This functionality works fine when you are an Admin/superuser. But when a normal user is give permission on group via "Configuration permissions" as shown below, the user gets permission to view and add groups but when he adds a group, he is not added to the same group. Due to this he is not able to view the group he created, resulting in a permission error. Ideally, when the user creates the group, he should be added as the owner of the group.
image
image
image

paulOsinski:
@missy-tester I tried replicating your issue on the demo server, but I wasn't able to get the 403 error.

1. Created an Example-User account with no permissions
2. While logged in as Admin, I gave View, Add Groups permissions to that Example-User 
3. Logged in as the Example-User
4. Example-User created a new Group
5. Example-User was redirected to the Group view automatically, and they could see that Example-User is the owner of the Group

PR #10071, opened 2024-04-30
remove non-working links from social-authentication.md

paulOsinski:
Using certain markdown links prevents the docs from building correctly locally, as 'https://the_hostname_you_have_gitlab_deployed:your_gitlab_port/profile/applications' etc are not real URLs. 

Although this problem might be due to my own Hugo settings/version, I think it would be less problematic to remove the non-existent URL from the docs altogether, so that readers understand that this is an example url and not intended to be a functioning link.


mtesauro (approved):
Approved

cneill (approved):

PR #10082, opened 2024-05-02
:bug: Fix RedHatSatellite components

manuel-sommer:
This Bugfix puts the packages into the description field and removes it from the components field within the RedHatSatellite parser. The reason is that RedHatSatellite sometimes delivers a huge package list which exceeds the 500 characters limit of the components field. We would have to make a database migration which is right now limited to 500 character to at least 10000 characters. That is why I decided to move the content to the description field.


cneill (approved):

Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #10085, opened 2024-05-02
Ruff: add and fix FBT001 & FBT003

kiblik:
Implement fixes for 2 out of 3 FBT rules

https://docs.astral.sh/ruff/rules/#flake8-boolean-trap-fbt


unittests/test_apiv2_limit_reqresp.py
```diff
@@ -16,7 +16,7 @@ def setUp(self: object):
         self.client = APIClient()
         self.client.credentials(HTTP_AUTHORIZATION="Token " + token.key)
 
-    def assertReqrespValue(self: object, value: int, expect_notequal: bool = False) -> None:
+    def assertReqrespValue(self: object, value: int, *, expect_notequal: bool = False) -> None:
```
Maffooch, line 19: I learned something on this one! Thanks :)

For others that have not seen this before: https://realpython.com/python-asterisk-and-slash-special-parameters/

dojo/management/commands/push_to_jira_update.py
```diff
@@ -27,5 +27,5 @@ def handle(self, *args, **options):
 
         for finding in findings:
             logger.info("Checking issue:" + str(finding.id))
-            jira_helper.update_jira_issue(finding, True)
+            jira_helper.update_jira_issue(finding, *[True])
```
Maffooch, line 30: I am a little confused by this one actually. What is it doing?

Maffooch (changes requested):

dojo/management/commands/push_to_jira_update.py
kiblik: During the calling of a function, you might add `*` in front of `list` (or `tuple` or `string` or maybe also something else). It will expand the variable.
```python
>>> print('x')
x
>>> print('x', 'y')
x y
>>> print(['x', 'y'])
['x', 'y']
>>> print(*['x', 'y'])
x y
>>> z = ['x', 'y']
>>> print(*z)
x y
```
It is handy if you have stored some values stored in a list and you need to pass them as separate values.

Btw, similar works for dict (but `**`):
```python
>>> def f(a, b):
...     print(f'a: {a}, b: {b}')
... 
>>> c = {'b':'1', 'a':'2'}
>>> f(**c)
a: 2, b: 1
```

In this case, it is a little dirty hack. `update_jira_issue` is defined as `update_jira_issue(obj, *args, **kwargs)` so until now it would receive `args = [True]`. I used this format to mute ruff and keep the same functionality.

Maffooch: Aha that is quite dirty! That makes sense though. I think an exclusion comment would be better suited though so. The point of ruff is to make things less dirty 😄

kiblik: Btw, I checked also `update_jira_issue` and to be honest, I'm not able to find, how `args` (so also mentioned `True`) is used. If somebody else will check it and confirm it, I would go with removal of this `True` which might be easier solution.
I just never touched Jira implementation so I wanted to need backward compatibility.

Maffooch: Looks unused to me. I think the removal of true works. It also is relatively low risk since this code in particular is not access at run time. It is ran by hand

kiblik: I just removed it.

Maffooch (approved):

mtesauro (approved):
Approved

cneill (approved):

unittests/test_apiv2_limit_reqresp.py
hblankenship: Just learned that myself. It's awful. But...yay, Python.

hblankenship (approved):

mtesauro:
@kiblik DOH! This one too. :frowning_face:

PR #10091, opened 2024-05-02
Ruff: add and fix T20

kiblik:
all `print`s replaced with `logger.info` or `logger.debug`

https://docs.astral.sh/ruff/rules/#flake8-print-t20

### DIFF
--- a/tests/Import_scanner_test.py
+++ b/tests/Import_scanner_test.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import re
 import shutil
@@ -9,6 +10,8 @@ from base_test_class import BaseTestCase
 from product_test import ProductTest
 from selenium.webdriver.common.by import By
 from selenium.webdriver.support.ui import Select
+
+logger = logging.getLogger(__name__)
 
 
 class ScannerTest(BaseTestCase):
@@ -42,13 +45,13 @@ class ScannerTest(BaseTestCase):
                 missing_tests += [test]
 
         if len(missing_tests) > 0:
-            print("The following scanners are missing test cases or incorrectly named")
-            print("Names must match those listed in /dojo/tools")
-            print("Test cases can be added/modified here:")
-            print("https://github.com/DefectDojo/sample-scan-files\n")
+            logger.info("The following scanners are missing test cases or incorrectly named")
+            logger.info("Names must match those listed in /dojo/tools")
+            logger.info("Test cases can be added/modified here:")
+            logger.info("https://github.com/DefectDojo/sample-scan-files\n")
             for test in missing_tests:
-                print(test)
-            print()
+                logger.info(test)
+            logger.info()
         assert len(missing_tests) == 0
 
     def test_check_for_doc(self):
@@ -70,13 +73,13 @@ class ScannerTest(BaseTestCase):
                     missing_docs += [tool]
 
         if len(missing_docs) > 0:
-            print("The following scanners are missing documentation")
-            print("Names must match those listed in /dojo/tools")
-            print("Documentation can be added here:")
-            print("https://github.com/DefectDojo/django-DefectDojo/tree/dev/docs\n")
+            logger.info("The following scanners are missing documentation")
+            logger.info("Names must match those listed in /dojo/tools")
+            logger.info("Documentation can be added here:")
+            logger.info("https://github.com/DefectDojo/django-DefectDojo/tree/dev/docs\n")
             for tool in missing_docs:
-                print(tool)
-            print()
+                logger.info(tool)
+            logger.info()
         assert len(missing_docs) == 0
 
     def test_check_for_forms(self):
@@ -108,13 +111,13 @@ class ScannerTest(BaseTestCase):
                     missing_forms += [tool]
 
         if len(missing_forms) > 0:
-            print("The following scanners are missing forms")
-            print("Names must match those listed in /dojo/tools")
-            print("forms can be added here:")
-            print("https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/forms.py\n")
+            logger.info("The following scanners are missing forms")
+            logger.info("Names must match those listed in /dojo/tools")
+            logger.info("forms can be added here:")
+            logger.info("https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/forms.py\n")
             for tool in missing_forms:
-                print(tool)
-            print()
+                logger.info(tool)
+            logger.info()
         assert len(missing_forms) == 0
 
     @unittest.skip("Deprecated since Dynamic Parser infrastructure")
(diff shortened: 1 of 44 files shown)


Maffooch (changes requested):
This one will need a little more TLC. Loggers cannot take more than one argument like print can:
```python
>>> import logging
>>> logger = logging.getLogger(__name__)
>>> logger.error("one", "two", "three")
--- Logging error ---
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/logging/__init__.py", line 1160, in emit
    msg = self.format(record)
          ^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/logging/__init__.py", line 999, in format
    return fmt.format(record)
           ^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/logging/__init__.py", line 703, in format
    record.message = record.getMessage()
                     ^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/logging/__init__.py", line 392, in getMessage
    msg = msg % self.args
          ~~~~^~~~~~~~~~~
TypeError: not all arguments converted during string formatting
Call stack:
  File "<stdin>", line 1, in <module>
Message: 'one'
Arguments: ('two', 'three')
```

kiblik:
> This one will need a little more TLC. Loggers cannot take more than one argument like print can:


Thanks @Maffooch for this. I can see it only in
- `dojo/decorators.py`
- `dojo/utils.py`
- `tests/base_test_class.py`
- `tests/notes_test.py`

I just fixed them. Have you found them somewhere else as well?

Maffooch:
There are a few that are commented out that would break if someone were to uncomment them. An example being `dojo/api_v2/prefetch/schema.py`

Thought it may be more appropriate to just remove. What do you think?

kiblik:
> There are a few that are commented out that would break if someone were to uncomment them. An example being `dojo/api_v2/prefetch/schema.py`
> 
> Thought it may be more appropriate to just remove. What do you think?

Agree. I will remove them

unittests/tools/test_checkmarx_one_parser.py
```diff
@@ -125,7 +128,7 @@ def test_sca_finding(finding):
             findings = parser.get_findings(testfile, Test())
             self.assertEqual(146, len(findings))
             for index in range(len(findings)):
-                print(index, findings[index])
+                logger.debug(index, findings[index])
```
Maffooch, line 131: Found one more 😉 
```suggestion
                logger.debug(f"{index} {findings[index]}")
```

Maffooch (changes requested):

Maffooch (approved):

unittests/tools/test_checkmarx_one_parser.py
cneill: Looks like this one might have been reverted in a rebase?

tests/zap.py
```diff
@@ -103,11 +106,11 @@ class Main:
...
         summary.add_row(["Medium", medium])
         summary.add_row(["High", high])
-        print(summary)
+        logger.info(summary)
 
         for url in sort_by_url:
-            print()
-            print(url)
+            logger.info()
+            logger.info(url)
```
cneill, line 113: I don't think we can use `logger.info` this way since it requires a positional argument. Perhaps something like this:

```suggestion
            logger.info("\n" + url)
```

tests/Import_scanner_test.py
```diff
@@ -42,13 +45,13 @@ def test_check_test_file(self):
...
-            print('https://github.com/DefectDojo/sample-scan-files\n')
+            logger.info('The following scanners are missing test cases or incorrectly named')
+            logger.info('Names must match those listed in /dojo/tools')
+            logger.info('Test cases can be added/modified here:')
+            logger.info('https://github.com/DefectDojo/sample-scan-files\n')
             for test in missing_tests:
-                print(test)
-            print()
+                logger.info(test)
+            logger.info()
```
cneill, line 54: We can't use `logger.info()` this way and I'm not sure if logging a simple "\n" is necessary here?
```suggestion

```

tests/Import_scanner_test.py
```diff
@@ -70,13 +73,13 @@ def test_check_for_doc(self):
...
-            print('https://github.com/DefectDojo/django-DefectDojo/tree/dev/docs\n')
+            logger.info('The following scanners are missing documentation')
+            logger.info('Names must match those listed in /dojo/tools')
+            logger.info('Documentation can be added here:')
+            logger.info('https://github.com/DefectDojo/django-DefectDojo/tree/dev/docs\n')
             for tool in missing_docs:
-                print(tool)
-            print()
+                logger.info(tool)
+            logger.info()
```
cneill, line 82: Same comment as above
```suggestion

```

tests/Import_scanner_test.py
```diff
@@ -108,13 +111,13 @@ def test_check_for_forms(self):
...
-            print('https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/forms.py\n')
+            logger.info('The following scanners are missing forms')
+            logger.info('Names must match those listed in /dojo/tools')
+            logger.info('forms can be added here:')
+            logger.info('https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/forms.py\n')
             for tool in missing_forms:
-                print(tool)
-            print()
+                logger.info(tool)
+            logger.info()
```
cneill, line 120:
```suggestion

```

tests/Import_scanner_test.py
```diff
@@ -149,13 +152,13 @@ def test_check_for_options(self):
...
-            print('https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/templates/dojo/import_scan_results.html\n')
+            logger.info('The following scanners are missing templates')
+            logger.info('Names must match those listed in /dojo/tools')
+            logger.info('templates can be added here:')
+            logger.info('https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/templates/dojo/import_scan_results.html\n')
             for tool in missing_templates:
-                print(tool)
-            print()
+                logger.info(tool)
+            logger.info()
```
cneill, line 161:
```suggestion

```

tests/Import_scanner_test.py
```diff
@@ -235,13 +238,13 @@ def test_engagement_import_scan_result(self):
...
-            print('https://github.com/DefectDojo/sample-scan-files\n')
+            logger.info('The following scan imports produced errors')
+            logger.info('Names of tests must match those listed in /dojo/tools')
+            logger.info('Tests can be added/modified here:')
+            logger.info('https://github.com/DefectDojo/sample-scan-files\n')
             for test in failed_tests:
-                print(test)
-            print()
+                logger.info(test)
+            logger.info()
```
cneill, line 247:
```suggestion

```

cneill (changes requested):
A few minor comments

blakeaowens (approved):

unittests/tools/test_checkmarx_one_parser.py
kiblik: @cneill, you are probably right. Thanks also for other suggestions.

cneill (approved):
Looks good!

mtesauro (approved):
Approved

PR #10111, opened 2024-05-03
Ruff: add and fix RET

kiblik:
New Ruff rule: RET

https://docs.astral.sh/ruff/rules/#flake8-return-ret

### DIFF
--- a/dojo/search/views.py
+++ b/dojo/search/views.py
@@ -401,6 +401,7 @@ def simple_search(request):
     operators: {'tags': ['anchore'], 'vulnerability_id': ['CVE-2020-1234']}
     keywords:  ['jquery']
     """
+    return None
 
 
 # it's not google grade parsing, but let's do some basic stuff right
@@ -448,8 +449,7 @@ def vulnerability_id_fix(keyword):
 
     if vulnerability_ids:
         return " ".join(vulnerability_ids)
-    else:
-        return keyword
+    return keyword
 
 
 def apply_tag_filters(qs, operators, skip_relations=False):
--- a/dojo/search/views.py
+++ b/dojo/search/views.py
@@ -31,6 +31,45 @@ max_results = settings.SEARCH_MAX_RESULTS
 
 
 def simple_search(request):
+
+    """
+    query:     some keywords
+    operators: {}
+    keywords:  ['some', 'keywords']
+
+    query:     some key-word
+    operators: {}
+    keywords:  ['some', 'key-word']
+
+    query:     keyword with "space inside"
+    operators: {}
+    keywords:  ['keyword', 'with', 'space inside']
+
+    query:     tag:anchore word tags:php
+    operators: {'tag': ['anchore'], 'tags': ['php']}
+    keywords:  ['word']
+
+    query:     tags:php,magento
+    operators: {'tags': ['php,magento']}
+    keywords:  []
+
+    query:     tags:php tags:magento
+    operators: {'tags': ['php', 'magento']}
+    keywords:  []
+
+    query:     tags:"php, magento"
+    operators: {'tags': ['php, magento']}
+    keywords:  []
+
+    query:     tags:anchorse some "space inside"
+    operators: {'tags': ['anchorse']}
+    keywords:  ['some', 'space inside']
+
+    query:     tags:anchore vulnerability_id:CVE-2020-1234 jquery
+    operators: {'tags': ['anchore'], 'vulnerability_id': ['CVE-2020-1234']}
+    keywords:  ['jquery']
+    """
+
     tests = None
     findings = None
     finding_templates = None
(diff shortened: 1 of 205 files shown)


mtesauro:
@kiblik Ironically, this is failing a Ruff check :smiley: 

Maybe a difference between current ruff config and when this was created :shrug:

mtesauro:
@kiblik FYI: Going to try to knock this one out next week after the release on Monday so we avoid future merge conflicts.

Thanks for keeping it alive and healthy for as long as you have :rocket:

dojo/search/views.py
```diff
@@ -401,6 +401,7 @@ def simple_search(request):
     operators: {'tags': ['anchore'], 'vulnerability_id': ['CVE-2020-1234']}
     keywords:  ['jquery']
     """
+    return None
```
cneill, line 404:
```suggestion

```
This only seems to be present because the docstring comment occurs after the `return response` line above - it's never actually reachable. Maybe just move the comment to the top where it would be more easily seen?

dojo/system_settings/views.py
```diff
@@ -116,7 +116,7 @@ def get_celery_status(
             context["celery_msg"] = "Celery needs to have the setting CELERY_RESULT_BACKEND = 'db+sqlite:///dojo.celeryresults.sqlite' set in settings.py."
             context["celery_status"] = "Unknown"
 
-        return None
+        return
```
cneill, line 119: And another one going from `return None` -> `return`...

dojo/tools/cyclonedx/json_parser.py
```diff
@@ -144,4 +144,4 @@ def _flatten_components(self, components, flatted_components):
             # tools don't provide it
             if "bom-ref" in component:
                 flatted_components[component["bom-ref"]] = component
-        return None
+        return
```
cneill, line 147: Another `return None` -> `return`...

dojo/user/validators.py
```diff
@@ -13,8 +13,7 @@ def validate(self, password, user=None):
             raise ValidationError(
                 self.get_help_text(),
                 code="password_too_short")
-        else:
-            return None
+        return
```
cneill, line 16: `return None` -> `return`...

dojo/user/validators.py
```diff
@@ -91,5 +85,4 @@ class DojoCommonPasswordValidator(CommonPasswordValidator):
     def validate(self, password, user=None):
         if get_system_setting("non_common_password_required"):
             return super().validate(password, user)
-        else:
-            return None
+        return None
```
cneill, line 88: For some reason this one's left alone (stays `return None`)?

cneill (commented):
Several comments on this one, but they're all basically the same comment (besides the one in `dojo/search/views.py`) - there seems to be inconsistency in using `return` vs. `return None` in different files. I didn't comment on all of them, but they should be relatively easy to find. I don't have a strong opinion on which one we choose, but we should stick to one way of doing things if we're going to introduce explicit returns everywhere.

I am _very_ supportive of early returning, so I'm happy to enable this rule. However, while reviewing some of these files, I can't help but note that we have some [extremely long and complicated functions](https://github.com/DefectDojo/django-DefectDojo/blob/ea7c0f557814046ceacf727c45885f23d6d3b205/dojo/search/views.py#L33-L404) that are difficult to read even with explicit `if`/`else` conditions breaking up the flow of statements and helping to identify indentation. With early returns, these become even harder to parse IMO. I also tend to prefer bailing early on error to avoid super long `if` blocks, like [this one](https://github.com/DefectDojo/django-DefectDojo/blob/ea7c0f557814046ceacf727c45885f23d6d3b205/dojo/search/views.py#L61), but those cases are going to be hard for `ruff` to detect and reverse automatically.

Based on the above, I'd love it if we could:
* Enable a cyclomatic complexity rule like [`C901`](https://docs.astral.sh/ruff/rules/complex-structure/), even if we have to set the `lint.mccabe.max-complexity` threshold pretty high initially. Starting at `50` would trigger 3 failures, `25` would trigger 26.
* Enable a max function length rule like [`PLR0915`](https://docs.astral.sh/ruff/rules/too-many-statements/), even if we have to set the `lint.pylint.max-statements`threshold pretty high initially. Starting at `100` would trigger 26 failures (basically the same ones as `C901`), `50` would trigger 105.

These rules seem like a high-leverage way to increase code readability. I realize these can't be auto-fixed with `ruff` though, so I certainly understand if you don't want  to take on such a large task 😅  No need to address in this PR, just a "wish list" of linter rules for future PRs.

mtesauro:
@kiblik Starting to look good. Thanks for those other 2 PRs related to this one.

Once the Ruff stuff is fixed, I'm ready to approve and I suspect the other reviewers as well :+1:

dojo/search/views.py
kiblik: @cneill, like this https://github.com/DefectDojo/django-DefectDojo/pull/10111/commits/54942b38bc659615b760ae8d8462b7c8aeed643a

mtesauro (approved):
Approved

cneill (approved):

mtesauro:
Closing to reopen - GHA are being weird today.

Maffooch (approved):

grendel513 (approved):

PR #10112, opened 2024-05-03
Ruff: add and fix some SIM rules

kiblik:
There are many SIM rules, it was easier to exclude some specific

https://docs.astral.sh/ruff/rules/#flake8-simplify-sim


ruff.toml
Maffooch, line 1: I think we should enable SIM115 and SIM117. What do you think?

ruff.toml
```diff
@@ -55,7 +56,7 @@ select = [
    "NPY",
    "AIR",
 ]
-ignore = ["E501", "E722", "F821"]
+ignore = ["E501", "E722", "F821", "SIM102", "SIM113", "SIM115", "SIM116", "SIM117"]
```
Maffooch, line 59: It might also help to set them like this to avoid conflicts in subsequent PRs,
```suggestion
ignore = [
    "E501",
    "E722",
    "F821",
    "SIM102",
    "SIM113",
    "SIM116",
]
```

ruff.toml
kiblik: Fully agree. I would like to apply as many best practices as possible.
However, in the "first wave", I covered only cases which I have been able to solve with auto-fix. SIM115 and SIM117 need manual resolution.
So I would prefer to merge this PR without SIM115 and 117 (keep them ignored) and they might be resolved in the future in separated PR (which will be prepared by hand - or some smart AI ;) )

Maffooch: Ahh I see that makes a lot of sense! I apologize for "sneaking" those rules out on my code suggestion below this 😅

mtesauro:
@kiblik FYI: We're just experimenting with SonarCloud so don't feel obligated to make that GH Action happy.

mtesauro:
@kiblik We finally got #10111 over the line and we're ready to focus on this one if you have the cycles to fix the merge conflicts.

Thanks again for doing these and being so patient as they get reviewed. Big changes in a very positive direction for the project :rocket:

PR #10178, opened 2024-05-10
:bug: fix sonarqube missing CWE for deduplication

quirinziessler:
During the import of Sonarqube findings I got aware of an issue related with the deduplication on the parser. For some kind of findings a cwe is not provided but is mandatory for the deduplication algorithm by settings.dist.py. The deduplication then falls back to the legacy dedupe hash model and the import takes longer than usual.
```
defectdojo-uwsgi-1  | [10/May/2024 07:38:28] WARNING [dojo.specific-loggers.deduplication:2746] Cannot compute hash_code based on configured fields because cwe is 0 for finding of title 'java:S3415_AYK2F0bHUOabgueBfBvD' found in file 'None'. Fallback to legacy mode for this finding.
```
I removed the cwe from the dedupe hash code and added other suitable values by extending the parser.

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -1186,7 +1186,7 @@ HASHCODE_FIELDS_PER_SCANNER = {
     'Checkmarx Scan': ['cwe', 'severity', 'file_path'],
     'Checkmarx OSA': ['vulnerability_ids', 'component_name'],
     'Cloudsploit Scan': ['title', 'description'],
-    'SonarQube Scan': ['cwe', 'severity', 'file_path'],
+    'SonarQube Scan': ['severity', 'file_path', 'vuln_id_from_tool'],
     'SonarQube API Import': ['title', 'file_path', 'line'],
     'Sonatype Application Scan': ['title', 'cwe', 'file_path', 'component_name', 'component_version', 'vulnerability_ids'],
     'Dependency Check Scan': ['title', 'cwe', 'file_path'],
--- a/dojo/tools/sonarqube/sonarqube_restapi_json.py
+++ b/dojo/tools/sonarqube/sonarqube_restapi_json.py
@@ -19,6 +19,7 @@ class SonarQubeRESTAPIJSON:
                     status = issue.get("status")
                     message = issue.get("message")
                     tags = issue.get("tags")
+                    tags.append("bug")
                     type = issue.get("type")
                     scope = issue.get("scope")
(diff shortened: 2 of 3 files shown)


Maffooch:
@quirinziessler how much longer is import taking when using a different a hash code computation? I believe there is one more conditional and a log statement added to the hashcode computation. Sonarqube is a pretty popular tool, and this update would impact a lot of folks when their hash codes are now using different values, and there is no longer a match with existing sonarqube findings.

I would recommend updating the config in your instance to allow the hash code calculation for sonarqube to accept null CWEs:
https://github.com/DefectDojo/django-DefectDojo/blob/2c7b5066736bfe13ed42227f6832ac540e188499/dojo/settings/settings.dist.py#L1275C1-L1289C29

quirinziessler:
@Maffooch I am aware of the widespread use of SQ and this is also why I opened this PR against bugfix. During testing I saw my import time dropping about 5 seconds until all processes finished for aprox 50 findings. This may sound like a small impact but after successful testing with around 200 SQ projects I can say this change should definitely part of the next release, so everyone can use DD out of the box and to keep the required adaptations as low as possible.

Furthermore the SQ ID should be preferred instead of the CWE for deduplication in my eyes. A file can have multiple findings with the same CWE in multiple lines, e.g. "missing input validation". By using the SQ ID those findings will be reflected in DefectDojo and not marked as duplicates like it is currently. Unfortunately the finding line is not consequently provided within the output of SQ so this is the only possible way to catch this behavior.

quirinziessler:
@Maffooch any updates here?

Maffooch:
Apologies for missing this. If the hashcode is the be changed here, I think it should align with how the [API based SonarQube parser](https://github.com/DefectDojo/django-DefectDojo/tree/master/dojo/tools/api_sonarqube) is handling deduplication. It uses `unique_tool_by_id` based on the rule key. 

A migration will also need to be written to update all of the hash codes for sonarqube findings. Something like this would need to go against the `dev` branch, as the migration could potentially take long time for users with hundreds of projects

quirinziessler:
All good @Maffooch. According to the dedupe of SQ and your statement this meets my changes. I can also prepare the same PR against the latest dev version. However I currently fail on how to write a db_migration for the hash codes as the models seem not to be accessible at migration time. Can you help me on this and maybe provide an example?

mtesauro:
Closing as stale - assuming they made the changes they needed in local_settings.py.

PR #10183, opened 2024-05-10
Object File Uploads: Add validations and download functionality

Maffooch:
When using the API to upload files, it was noticed that `FILE_UPLOAD_TYPES` settings was not being enforced. This PR adds enforcement of this setting. Additionally, these files should be downloaded to disk rather than being rendered by the browser

[sc-5953]

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import os
 import re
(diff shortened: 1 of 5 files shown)


grendel513 (approved):

blakeaowens (approved):

mtesauro (approved):
Approved

PR #10204, opened 2024-05-14
Improving date of discovery filter

hblankenship:
[sc-5885]

Fixed missing filter on Discovery Date. Now able to specify a range with from and to. 

**Test results**

Tested locally and includes a fix for when labels are 'missing' in the report builder html.

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -1539,6 +1539,7 @@ class PercentageRangeFilter(RangeFilter):
 
 class FindingFilterHelper(FilterSet):
     title = CharFilter(lookup_expr="icontains")
+    date = DateFromToRangeFilter(field_name='date', label="Date Discovered")
     on = DateFilter(field_name="date", lookup_expr="exact", label="On")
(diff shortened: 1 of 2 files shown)


dojo/filters.py
```diff
@@ -1538,7 +1539,6 @@ def filter(self, qs, value):
 
 class FindingFilterHelper(FilterSet):
     title = CharFilter(lookup_expr="icontains")
-    date = DateRangeFilter()
```
Maffooch, line 1541: This deletion looks like it is removing the `date` filter for all finding filters besides the report filters. Is this intended?

Maffooch (changes requested):

dojo/filters.py
hblankenship: Added this as DateRangeFromToFilter after discussion.

Maffooch (approved):

cneill (approved):

grendel513 (approved):

Issue #10198, opened 2024-05-13
Mend Scan Findings API response vulnerability_ids[] list contains +1 repeated vulnerability_id entries for each Finding during reimport-scan


testaccount90009:
![vuln_ids_repeat_2](https://github.com/DefectDojo/django-DefectDojo/assets/122134756/bfe29ce3-0eb6-4490-a8df-293cd166d9f2)
![vuln_ids_repeat](https://github.com/DefectDojo/django-DefectDojo/assets/122134756/01fddcd7-e73c-44ff-a230-01b18b007c2c)
**I've raised an issue in slack here - https://owasp.slack.com/archives/C2P5BA8MN/p1715638859962089 **

**Bug description**
**Steps to reproduce**

1. Make an API call to the Findings for a Mend Scan test type
2. Observe the Findings API response containing the mentioned issue of 6x vulnerability_id repeat in vulnerability_ids list

**Expected behavior**
Expected behavior is this would not have a 6x repeated entry for each CVE in the list of vuln_ids and instead have the vulnerability_id = CVE or unique_id_from_tool or vuln_id_from_tool, but not a repeat of a single CVE 6x as the vulnerability_ids list of the Findings. (See Screenshots)

**Deployment method** *(select with an `X`)*
- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: [e.g. Ubuntu 18.04]
 - DefectDojo version (see footer) or commit message: [use `git show -s --format="[%ci] %h: %s [%d]"`]

**Logs** 
Use `docker-compose logs` (or similar, depending on your deployment method) to get the logs and add the relevant sections here showing the error occurring (if applicable).

**Can see this error in the response when making API call to /findings/ in 2.34.1**

**Sample scan files**
If applicable, add sample scan files to help reproduce your problem.

**Screenshots**

**Additional context** (optional)

testaccount90009:
In Slack it was confirmed by another person (in direct messages) that they are seeing this duplicative behavior against reimports where the vulnerability_ids are being processed/retrieved.

testaccount90009:
Our automation runs the reimport-scan each night, so I tested again today and saw that instead of 6x, it's now 7x.  If tomorrow it is 8, then the below must be true...

I think there is a problem with the way the vulnerability_id at import gets appended to a vulnerability_ids list, instead of overwriting that dictionary/list.

[...]

PR #10214, opened 2024-05-15
fix for multi vuln ids when reimporting scans

hblankenship:
**Description**
[sc-5982]
Resolves https://github.com/DefectDojo/django-DefectDojo/issues/10198

To re-iterate, when reimporting scans into DefectDojo, vulnerability_ids were getting duplicated with each reimport. This fix implements the same method used by the apiv2 finding helper in the baseimporter.

**Test results**

Tested mutiple scan types to ensure vulnerability ids were still handled correctly and that duplicates were removed on reimport. This will not clear duplicate vulnerability ids until a reimport is done after this fix.

### DIFF
--- a/dojo/finding/helper.py
+++ b/dojo/finding/helper.py
@@ -650,7 +650,8 @@ def save_vulnerability_ids(finding, vulnerability_ids):
     Vulnerability_Id.objects.filter(finding=finding).delete()
 
     # Save new vulnerability ids
-    Vulnerability_Id.objects.bulk_create([Vulnerability_Id(finding=finding, vulnerability_id=vulnerability_id) for vulnerability_id in finding.unsaved_vulnerability_ids])
+    if finding.unsaved_vulnerability_ids:
+        Vulnerability_Id.objects.bulk_create([Vulnerability_Id(finding=finding, vulnerability_id=vulnerability_id) for vulnerability_id in finding.unsaved_vulnerability_ids])
 
     # Set CVE
     if vulnerability_ids:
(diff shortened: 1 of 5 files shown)


kiblik:
I'm not sure that this is the best idea. If I may suggest, I would:
- Add db_mig which would drop `(finding,vuln_id)` duplicities (for inspiration: https://gist.github.com/victorono/cd9d14b013487b8b22975512225c5b4c)
- Add db constraint to avoid this situation in the future: `class Meta: constraints = [models.UniqueConstraint(fields=['finding', 'vulnerability_id'])]`
- Do not remove all existing entries related to finding but
  - Remove only unnecessary `Vulnerability_Id.objects.filter(finding=finding).exclude(vulnerability_id__in=finding.unsaved_vulnerability_ids).delete()`
  - Call DB only for the creation of new entries, outside of the for loop (to save db calls) `Vulnerability_Id.objects. bulk_create([Vulnerability_Id(finding=finding, vulnerability_id=vulnerability_id) for vulnerability_id in finding.unsaved_vulnerability_ids])`

kiblik:
If there is an issue with duplicates during processing, I suppose `set()` might help here.
The best would be to define `unsaved_vulnerability_ids` as the `set`. But for a quick fix, at least using `bulk_create(set(...))` would remove duplicates.

Maffooch:
This is not feedback on the approach taken here, but this function is used in other places to set the vulnerability IDs correctly. It should be used directly instead of copy/pasting code from it
https://github.com/DefectDojo/django-DefectDojo/blob/2234848f02f243b5250894fccb6fc05ccaa1bc0d/dojo/finding/helper.py#L645C5-L645C27

hblankenship:
dropping this PR for bugfix PR

PR #10224, opened 2024-05-17
fix(filters): prevent crashes when `start_date == date.min`

p-l-:
**Description**

Some reports may include bogus values for `start_date`, like the `date.min` value. In that situation, DefectDojo would crash; the bug is not really in DefectDojo, but a bit of defensive programming won't hurt, and we want to avoid a crash even with invalid data.

When that happen, we'll use `date.min` as `start_date` rather than `date.min - timedelta(days=1)` which cannot exist in Python.

**Test results**

We do have an (internal, home made) scanner that (used to) produce such invalid result, and this patch fixes the issue.

**Documentation**

N/A

**Checklist**

- [x] Make sure to rebase your PR against the very latest ~~`dev`~~ `bugfix`.

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -2,7 +2,7 @@ import collections
 import decimal
 import logging
 import warnings
-from datetime import datetime, timedelta
+from datetime import date, datetime, timedelta
 
 import pytz
 import six
@@ -184,7 +184,10 @@ class FindingStatusFilter(ChoiceFilter):
             start_date = local_tz.localize(datetime.combine(
                 earliest_finding.date, datetime.min.time())
             )
-            self.start_date = _truncate(start_date - timedelta(days=1))
+            try:
+                self.start_date = _truncate(start_date - timedelta(days=1))
+            except OverflowError:  # this will fail when start_date is date.min
+                self.start_date = date.min
             self.end_date = _truncate(now() + timedelta(days=1))
         try:
             value = int(value)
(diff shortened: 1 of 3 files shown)


p-l-:
I cannot add the `bugfix` label to my PR, but that's what it is!

kiblik:
Can you also write a unittest for it? If your report contains sensitive info, feel free to anonymize it.

p-l-:
@kiblik sure thing. Could you guide me a bit on how to do that (or point to a documentation)? Where should I store the scan result to test?

kiblik:
> @kiblik sure thing. Could you guide me a bit on how to do that (or point to a documentation)? Where should I store the scan result to test?

https://documentation.defectdojo.com/contributing/how-to-write-a-parser/

p-l-:
OK I did read that but how can I be sure that the code I'm fixing is already tested somehow? Do I need to write a specific test for that? In the meantime I have added a test sample file anyway.

The provided test file would currently create this 500 error when navigating to e.g. `/finding/open`:
image

Maffooch:
@p-l- your unit test could first create a finding with the "bad date" and then invoke the filter class that you are making the change to in this PR in the same way that the `/finding/open` view does. This would ensure that the exception is handled if the test passes

p-l-:
@Maffooch Thanks. It seems that the results are not sent to the database, so my test won't work. Do you have any idea what the easier way to test it could be?

Maffooch:
Could put your test here: https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/test_finding_model.py

mtesauro:
Closing as stale

p-l-:
Yep sorry @mtesauro @Maffooch I did not find the time to create a working test :-/

mtesauro:
@p-l- 
> Yep sorry @mtesauro @Maffooch I did not find the time to create a working test :-/

No need to apologize - I appreciate the work you did to and we can always pick this back up and write a test. Life happens to everyone - feel free to do a PR anytime in the future. :+1:

PR #10226, opened 2024-05-20
Test_Import: Correct exception for multiple findings in the same object

Maffooch:
When a vulnerability is present in  a report multiple times, but with different status, there is an opportunity for the following exception:
```
duplicate key value violates unique constraint "dojo_test_import_finding_test_import_id_finding_i_e6f366e5_uniq"
DETAIL:  Key (test_import_id, finding_id)=(4108, 948) already exists.
```


mtesauro (approved):
Approved

grendel513 (approved):

cneill (approved):

devGregA (approved):

PR #10212, opened 2024-05-15
feat(settings): Add integrity checker

kiblik:
In many cases, users are changing `dojo/settings/settings.dist.py` file (they expect it is the way how the application should be configured).

This PR adds a checker that does not allow to start app if the change is detected. The error message points the user to documentation.

It is allowed only in debug mode when changes are expected.
Unittest which helps you in debug mode was added as well.

PR #10241, opened 2024-05-21
fix(integrity check): update hash + handle collectstatic

kiblik:
This PR fix: 
- "Conflict" (not updated hash in PR developed in the same time) with #8824
- Issue with running `collectstatic` (during any build): identified in https://github.com/DefectDojo/django-DefectDojo/pull/10212#issuecomment-2122192782


manuel-sommer (commented):
@kiblik : I don't have enough insights into the release process, but maybe it makes sense to add automatic recalculation for the release process.
Let's assume hash changes in both bugfix and dev branches. Then, the new release should have new hash codes automatically recalculated and commited for all branches. Is this automatically possible somehow?

Maffooch:
> I agree with @manuel-sommer that would be the ideal approach imo. Changes could be made

This should probably wait until after this PR is merged to get the `dev` branch corrected sooner

Maffooch (approved):

cneill (approved):
I'm a bit confused about why we have to create an exception for `collectstatic` since I can't see any reason the hash of `settings.dist.py` file would differ from our expected value in that step, but in the interest of getting this fixed quickly we can leave it for now.

quirinziessler:
> I'm a bit confused about why we have to create an exception for `collectstatic` since I can't see any reason the hash of `settings.dist.py` file would differ from our expected value in that step, but in the interest of getting this fixed quickly we can leave it for now.

@cneill please see my comment on #10212 - the integrity checker wrongly stopped the build process when running ./dc-build.sh when running in debug mode. Thanks @kiblik for looking into this so quickly.

mtesauro (approved):
Approved

cneill:
> > I'm a bit confused about why we have to create an exception for `collectstatic` since I can't see any reason the hash of `settings.dist.py` file would differ from our expected value in that step, but in the interest of getting this fixed quickly we can leave it for now.
> 
> @cneill please see my comment on #10212 - the integrity checker wrongly stopped the build process when running ./dc-build.sh when running in debug mode. Thanks @kiblik for looking into this so quickly.

Right, I can see that it has been [failing at that step](https://github.com/DefectDojo/django-DefectDojo/actions/runs/9178161879/job/25237407576?pr=10244) in our unit tests runs, but taking this branch and removing the `collectstatic` condition from the `settings.py` check still doesn't produce issues for me. Changing the profile/etc has no effect. I can only get it to fail when I modify the checksum to be intentionally incorrect.

quirinziessler:
> > > I'm a bit confused about why we have to create an exception for `collectstatic` since I can't see any reason the hash of `settings.dist.py` file would differ from our expected value in that step, but in the interest of getting this fixed quickly we can leave it for now.
> > 
> > 
> > @cneill please see my comment on #10212 - the integrity checker wrongly stopped the build process when running ./dc-build.sh when running in debug mode. Thanks @kiblik for looking into this so quickly.
> 
> Right, I can see that it has been [failing at that step](https://github.com/DefectDojo/django-DefectDojo/actions/runs/9178161879/job/25237407576?pr=10244) in our unit tests runs, but taking this branch and removing the `collectstatic` condition from the `settings.py` check still doesn't produce issues for me. Changing the profile/etc has no effect. I can only get it to fail when I modify the checksum to be intentionally incorrect.

@cneill it's all about when developing and testing things. When preparing a fix for an Issue e.g. for dedupe, this behavior hinders contributor to correctly test their stuff. Skipping the file is the best way to provide a smooth access for contributor to change Hash fields as well as other stuff there and test it correctly. Please see this as a Hotfix, until the hash code is created during release. It is really annoying right now.

hblankenship (approved):

PR #10253, opened 2024-05-22
feat(email notif): Scan_added - put findings to `<details>`

kiblik:
This PR makes the "scan added" email notification easier to read.
If there is a long list of findings, it might be harder to read it. From now on, notifications are "collapsed" into detail section and only the number of findings is visible. Users can click on the name of the category and see the full list.

Before:
Screenshot 2024-05-22 at 16 57 50

After:
Screenshot 2024-05-22 at 17 02 56

After, when you open 2 groups:
Screenshot 2024-05-22 at 17 03 15

No hard magic 🪄 , only simple HTML tags.


Maffooch (approved):

mtesauro (approved):
Approved

grendel513 (approved):

cneill (approved):

PR #10254, opened 2024-05-22
Importers: migrate to options class

Maffooch:
Moves the parsing of kwargs to a central `ImporterOptions` class where validation can occur and defaults be set


Maffooch:
This is more of a restructure/reorg rather and a rewrite. The functionalities are otherwise the same. I can take a look to see if that issue linked has a quick solution

mtesauro (approved):
Approved

grendel513 (approved):

dojo/importers/base_importer.py
```diff
@@ -237,58 +192,62 @@ def parse_dynamic_test_type_findings_from_tests(
...
+        self,
+        parsed_findings: List[Finding],
+        **kwargs: dict,
+    ) -> List[Finding]:
+        """
+        Processes findings in chunks within N number of processes. The
+        ASYNC_FINDING_IMPORT_CHUNK_SIZE setting will determine how many
+        findings will be processed in a given worker/process/thread
+        """
+        return self.process_findings(parsed_findings, sync=True, **kwargs)
```
cneill, line 246: Should this be `sync=False`?

```suggestion
        return self.process_findings(parsed_findings, sync=False, **kwargs)
```

cneill (commented):
Just a couple questions/comments

cneill (approved):

blakeaowens (approved):

Issue #10277, opened 2024-05-28
End of life Alpine Linux 3.16


sknabe-DBS:
According to the official website https://alpinelinux.org/releases/ alpine 3.16 has reached it's end of life on  2024-05-23.
Loos like 3.16 is still used in 2.34.4.

To receive security fixes etc. the alpine containers referenced in the Dockerfiles should upgraded to a newer version, i. e. 3.20.

Best regards,
Stephan Knabe

kiblik:
@mtesauro, this issue can be closed. Alpine has been updated to 3.20

PR #10280, opened 2024-05-28
fix(docker): Bump versions (python 3.11, alpine 3.20)

kiblik:
Looks like Renovate and/or Dependabot are not able to identify higher versions
- If Debian names are used
- If 2 versions are mentioned in the tag

This should be addressed somehow, for now, let's fix it by hand.

Fix for #10277

### DIFF
--- a/Dockerfile.django-alpine
+++ b/Dockerfile.django-alpine
@@ -5,7 +5,7 @@
 # Dockerfile.nginx to use the caching mechanism of Docker.
 
 # Ref: https://devguide.python.org/#branchstatus
-FROM python:3.11.3-alpine3.16@sha256:9efc6e155f287eb424ede74aeff198be75ae04504b1e42e87ec9f221e7410f2d as base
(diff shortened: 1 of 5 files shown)


kiblik:
> @kiblik I agree about bumping these to a later Alpine version but would like to stick with Python 3.11 since that's the tested version of Python.
> 
> I'd prefer to handle these one at a time - first do the Alpine part, then consider bumping the Python version instead of doing both in one PR.
> 
> Looking at the available Python 3.11 containers, this one seems to fit the bill: https://hub.docker.com/layers/library/python/3.11.9-alpine/images/sha256-df44c0c0761ddbd6388f4549cab42d24d64d257c2a960ad5b276bb7dab9639c7?context=explore
> 
> And uses Alpine 20 so we'd be good to 2026. Once this PR is merged we can look at rev'ing the Python version separately.
> 
> I wanted to comment to let others chime in before requesting a change to python:3.11.9-alpine

No problem @mtesauro. I updated all Pythons to 3.11.9 and all alpines to 3.20. PR is ready.

mtesauro (approved):
Approved

grendel513 (approved):

Maffooch (approved):

PR #10286, opened 2024-05-28
fix(docker): Bump versions (python 3.11, debian bookworm)

kiblik:
Same as #10280 but for Debian

### DIFF
--- a/Dockerfile.django-debian
+++ b/Dockerfile.django-debian
@@ -5,7 +5,7 @@
 # Dockerfile.nginx to use the caching mechanism of Docker.
 
 # Ref: https://devguide.python.org/#branchstatus
-FROM python:3.11.9-slim-bullseye@sha256:03b84f686938d82255f6e784e8c9222c1fad1ba4a9e609d79efed56da69b025b as base
+FROM python:3.11.9-slim-bookworm@sha256:8c1036ec919826052306dfb5286e4753ffd9d5f6c24fbc352a5399c3b405b57e as base
 FROM base as build
 WORKDIR /app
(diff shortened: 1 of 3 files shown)


mtesauro (approved):
Approved

grendel513 (approved):

cneill (approved):

Maffooch (dismissed):

kiblik:
> @kiblik I didn't notice the change to bookworm until conflicts, but isn't bullseye the more stable/tested of bookworm? I would rather we stick with bullseye unless there was something we really needed in bookworm that is not in bullseye. I think https://github.com/DefectDojo/django-DefectDojo/pull/10280 covers the version upgrade right?

@Maffooch 
> [Debian 12 (bookworm)](https://www.debian.org/releases/bookworm/) — current stable release
> [Debian 11 (bullseye)](https://www.debian.org/releases/bullseye/) — current oldstable release
> Source: https://www.debian.org/releases/

> Debian 12.5 was released on [February 10th, 2024](https://www.debian.org/News/2024/20240210). Debian 12.0 was initially released on June 10th, 2023.
> Source: https://www.debian.org/releases/bookworm/

> Debian 11 has been superseded by [Debian 12 (bookworm)](https://www.debian.org/releases/bookworm/).
> Source: https://www.debian.org/releases/bullseye/

#10280 upgraded all pythons to latest `3.11` (which is `3.11.9` right now) and alpines to `3.20`.
In my opinion, if we upgraded alpine to the latest we might debian as well.

Maffooch (approved):

PR #10293, opened 2024-05-29
Calm the deduplication of MsDefender

manuel-sommer:
I removed some information in the description. This information is already described within the endpoints to calm the deduplication of MsDefender ['title', 'description']. The problem is that the description changes as soon as a client gets a new IP Address.

### DIFF
--- a/dojo/tools/ms_defender/parser.py
+++ b/dojo/tools/ms_defender/parser.py
@@ -94,13 +94,10 @@ class MSDefenderParser:
         description += "productVendor: " + str(vulnerability['productVendor']) + "\n"
         description += "productVersion: " + str(vulnerability['productVersion']) + "\n"
         description += "machine Info: id: " + str(machine['id']) + "\n"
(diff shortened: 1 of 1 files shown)


manuel-sommer:
Retriggering the unittests:

mtesauro (approved):
Approved

grendel513:
@manuel-sommer - removing this from the description may mean the information is lost.  For example, if the endpoint is deleted (either on purpose or by mistake) the information will no longer be known - additionally, other parsers act the same way, providing as much information as possible in the description.  Just want to make sure this was considered.

manuel-sommer:
The machineid is still present within the finding. This information does not change and stays persistent eventhough the IP addresses will change. I know that also other parsers act like this, but in this case it just causes a lot of flickering.

grendel513 (approved):

cneill (approved):

Maffooch (approved):

PR #10301, opened 2024-05-29
Rebased PR for Vulnerability Ids

hblankenship:
[sc-5982]
Resolves https://github.com/DefectDojo/django-DefectDojo/issues/10198

To re-iterate, when reimporting scans into DefectDojo, vulnerability_ids were getting duplicated with each reimport. This fix implements the same method used by the apiv2 finding helper in the baseimporter.

Test results

Tested mutiple scan types to ensure vulnerability ids were still handled correctly and that duplicates were removed on reimport. This will not clear duplicate vulnerability ids until a reimport is done after this fix.


dojo/importers/base_importer.py
```diff
@@ -38,7 +38,7 @@
     Test_Import_Finding_Action,
     Test_Type,
     Tool_Configuration,
-    Vulnerability_Id,
+    Vulnerability_Id,  # noqa: F401
```
kiblik, line 41:
```suggestion
```
I do not understand usage of `noqa` here. If `Vulnerability_Id` is not used in the rest of the file, import might be simply removed.

hblankenship: It is used in a bunch of decorators in tests like in TestImporterUtils and the linter doesn't understand the decorator:

@patch('dojo.importers.base_importer.Vulnerability_Id', autospec=True)
    def test_handle_no_vulnerability_ids_references_and_cve(self, mock):
    
And I readily admit that I am unsure why or even how these tests are supposed to work....

grendel513 (approved):

mtesauro (approved):
Approved

cneill (approved):

Maffooch (approved):

PR #10333, opened 2024-06-04
feat(docker): Use Python 3.12 in docker images

kiblik:
This PR follows in #10286 and #10280

Whenever DD is considered as ready for Python3.12, this might be merged.

Issues to solve:
- [ ] `DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html`
        https://github.com/DefectDojo/django-DefectDojo/actions/runs/9372204332/job/25803362734?pr=10333


kiblik:
We can not keep this PR open because tests are stacked in an endless loop.

PR #10334, opened 2024-06-04
fix(polymorphic): Install package from git

kiblik:
We would like to use polymorphic with python 3.12 or django 4, we need to install this package directly from git.

Context:
- https://github.com/jazzband/django-polymorphic?tab=readme-ov-file#installation
- https://github.com/jazzband/django-polymorphic/issues/598
- https://github.com/jazzband/django-polymorphic/issues/599 (blocker for #10333)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ django-environ==0.11.2
 django-filter==24.3
 django-imagekit==5.0.0
 django-multiselectfield==0.1.13
-django-polymorphic==3.1.0
+git+https://github.com/jazzband/django-polymorphic.git@v4.0.0a#egg=django-polymorphic
 django-crispy-forms==2.3
 django_extensions==3.2.3
 django-slack==5.19.0


Maffooch:
I did some light reading on this one. The JazzBand account is in the process of taking over the polymorphic repo, and is still waiting on pypi access to transferred. I would imagine a formal release through pypi will be coming in the next couple weeks.

I suspect there will be a handful of pypi packages we use that will block the python3.12 upgrade, so rushing on this may end up creating more churn for use in the long run. What do you think @kiblik @mtesauro ?

kiblik:
> I did some light reading on this one. The JazzBand account is in the process of taking over the polymorphic repo, and is still waiting on pypi access to transferred. I would imagine a formal release through pypi will be coming in the next couple weeks.
> 
> I suspect there will be a handful of pypi packages we use that will block the python3.12 upgrade, so rushing on this may end up creating more churn for use in the long run. What do you think @kiblik @mtesauro ?

The sad part is that the ticket for the takeover has been open for a couple of weeks without any progress. https://github.com/pypi/support/issues/4164 It is hard to predict, how long we will need to wait.

Plus it looks like the installation process is broken in this moment: https://github.com/DefectDojo/django-DefectDojo/actions/runs/10405827519/job/28817490858?pr=10334#step:5:1135

Plus, `polymorphic` is not the only blocker for upgrading to Python3.12. Upgrade of `pygithub` is needed as well: https://github.com/DefectDojo/django-DefectDojo/pull/10473/files#diff-767da9c92328dd092f33c455f007b9c3a936854ccab95f8130cf821d18af82d0R1749
And it is blocked by missing tests: https://github.com/DefectDojo/django-DefectDojo/pull/9948#issuecomment-2285152659

My conclusion: There are multiple blockers. So missing an upgrade of this one would not solve the whole situation.

Maffooch:
We are closing for now, but are tracking the progress of polymorphic so that we can upgrade to python 3.12 at a later date

PR #10335, opened 2024-06-04
Shuffle tests

kiblik:
Test robustness of unittests. They should pass in any order.

### DIFF
--- a/docker/entrypoint-unit-tests-devDocker.sh
+++ b/docker/entrypoint-unit-tests-devDocker.sh
@@ -53,7 +53,7 @@ EOF
 
 echo "Unit Tests"
 echo "------------------------------------------------------------"
-python3 manage.py test unittests -v 3 --keepdb --no-input
+python3 manage.py test unittests -v 3 --keepdb --no-input --shuffle
 
 # you can select a single file to "test" unit tests
 # python3 manage.py test unittests.tools.test_npm_audit_scan_parser.TestNpmAuditParser --keepdb -v 3
--- a/docker/entrypoint-unit-tests.sh
+++ b/docker/entrypoint-unit-tests.sh
@@ -79,4 +79,4 @@ python3 manage.py migrate
 
 echo "Unit Tests"
 echo "------------------------------------------------------------"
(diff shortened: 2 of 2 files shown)


Maffooch:
I ran into some pickling errors with parallel turned out. I am not really sure what we would need to do to get that working. If shuffle works here, I think that would still make for a pretty big win 😀

kiblik:
> I ran into some pickling errors with parallel turned out. I am not really sure what we would need to do to get that working. If shuffle works here, I think that would still make for a pretty big win 😀

Yes, I was pleasantly surprised by it as well.
I do not see some easy fix for `--parallel` so I will keep only `--shuffle` here and I will try to fix `--parallel` in separate PR.

mtesauro (approved):
Approved

cneill (approved):

Maffooch (approved):

grendel513 (approved):

Maffooch:
@kiblik I think tests were passing here by coincidence like the last time we tried something like this.. Very frustrating... We're going to revert for now, and then fix more widely in the future. Once those fixes are applied, we can enable shuffle again

I found something out about this one
> I ran into some pickling errors with parallel turned out. I am not really sure what we would need to do to get that working.

The error raises from an exception/stacktrace during the test that is not able to be pickled. I think exceptions are intended to be handled, and then the test will fail via the assertions.

kiblik:
> @kiblik I think tests were passing here by coincidence like the last time we tried something like this.. Very frustrating... We're going to revert for now, and then fix more widely in the future. Once those fixes are applied, we can enable shuffle again
> 
> I found something out about this one
> 
> > I ran into some pickling errors with parallel turned out. I am not really sure what we would need to do to get that working.
> 
> The error raises from an exception/stacktrace during the test that is not able to be pickled. I think exceptions are intended to be handled, and then the test will fail via the assertions.

I suppose the problem with pickling is mainly in `--parallel` (#10408). But I agree that `--shuffle` is not behaving "well". I planned to offer to exclude `unittests.test_rest_framework` and run it separately without `--shuffle`. But full revert is probably better idea for now.

PR #9922, opened 2024-04-14
refactor generic parser

manuel-sommer:

PR #10338, opened 2024-06-05
add prowler v4 parser

kagahd:
add prowler v4 parser


docs/content/en/integrations/parsers/file/aws_prowler_v4.md
```diff
@@ -0,0 +1,102 @@
+---
+title: "AWS Prowler V3"
```
manuel-sommer, line 2: V4

manuel-sommer (changes requested):
small typo

kagahd:
It seems that only 1 unit test in `alpine` and also in `debian` failed:
`FAIL: test_delete_preview (unittests.test_rest_framework.TestsTest.test_delete_preview)`
However, I don't know how my code contribution could have any effect to make this unit test fail.
Any hints on this?

mtesauro:
@kagahd We've had issues with those tests failing somewhat randomly. Give is a day or so to look into what's up with those tests before spending more time on this PR to try to make them pass. 

Thanks for the contribution and your patience as we sort out those misbehaving tests.

kagahd:
It seems it's not a good idea to have **multiple** parsers for the **same** scanner even though the same scanner may have different report formats (e.g. `csv`, `json-asff`, `json-ocsf`, etc.) wich would require different parsers. 
It seems that DefectDojo is **not** designed to allow multiple parsers, called "tools", for the same scanner because deduplication and closing of findings that are missing in a report are bound to the "tool", hence the parser.

For example, if you import a prowler v3 report and then import a semantically identical prowler v4 report, DefectDojo would not detect any duplicates because two different parsers, "tools", were used to import both reports.
The same applies for closing findings that are missing in a report: if you import a prowler v3 report and then import an empty prowler v4 report, DefectDojo would not close any of the prowler v3 findings.

So I think the only viable solution to this problem is to implement **only one parser**, one "tool", for one scanner. That means that we would have only one parser, one "tool", for Prowler which must be agnostic of the different report formats that Prowler may produce or may have produced in former Prowler versions.

kagahd:
Dear @mtesauro, @Maffooch, @manuel-sommer & all the DefectDojo team,

I have a question regarding the implementation for only one Prowler parser for several report formats.
If there is only **one** Prowler scanner for **all** report formats, how do you define the different `DEDUPLICATION_ALGORITHM_PER_PARSER`? For the Prowler parsers v3 and v4 it is `DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL` which results in `unique_id_from_tool` but for the Prowler parser versions before that, called `AWS Prowler Scan` in DefectDojo, it is `DEDUPE_ALGO_HASH_CODE` which results in `hash_code`.

How about renaming the parser `AWS Prowler Scan` to something like `AWS Prowler legacy`? The current Prowler v3 and v4 parser could be merged into only one parser,  able to parse different formats of reports, i.e. `json` and `json-ocsf`, as both use the same `DEDUPLICATION_ALGORITHM_PER_PARSER`. We could or should call the parser for example `Prowler v3 plus`, without the prefix `AWS` because Prowler is able to create more than just AWS reports. The suffix `v3 plus` indicates that it can be used from Prowler version 3.

What do you think of the idea or what would you suggest as the best solution?

manuel-sommer:
Hi @kagahd,
you have multiple options:

- Take a look at RustyHog https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/rusty_hog/parser.py --> But here the problem is that you have only one deduplication parser available. Here the right parser is automatically detected. Try it in the GUI, it is pretty nice. 
- Take a look at Checkmarx Scan https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/checkmarx/parser.py --> Here you have one parser with 2 deduplications available.

Thus, maybe you can:
- Merge Prowler, Prowler3 and Prowler4 together to "Prowler" and use the functionalities of both RustyHog and Checkmarx Scan to resolve all problems. 
- For this, you would have to separate the different prowler versions into multiple files (classes), like I did in multiple PRs, e.g. https://github.com/DefectDojo/django-DefectDojo/pull/9922/files (I separated here the json and csv part).

kagahd:
I close this PR in favor of PR #10372

Issue #10343, opened 2024-06-05
 An exception error occurred during the report import ,during the sonar cloud integration with the defect dojo


Nsai1997:
Hi team, when click on import vulnerabilities while i am importing the vulnerabilites from the sonarqube to defect dojo I am getting the below error.  An exception error occurred during the report import: api_scan_configuration of type <class 'dojo.models.Product_API_Scan_Configuration'> is not one of [<class 'dojo.models.Tool_Configuration'>] Here are all the options submitted: None

Even I checked the API key in the tool configuration,Project key while importing the findings all are correct

Nsai1997:
Hi team,
Even I tried with the latest release 2.3.5 but facing same error.
![Uploading IMG_0920.png…](

Can you please help me in guide me the steps to resolve this issue

37b:
We're having the same issue with on-prem SonarQube. We're stuck until this is resolved.

Nsai1997:
![IMG_0920](https://github.com/DefectDojo/django-DefectDojo/assets/46392529/26200c39-2d3c-4e81-a16c-78e6ec614d4e)

mtesauro:
Please try [2.35.1](https://github.com/DefectDojo/django-DefectDojo/releases/tag/2.35.1) release that happened today. A fix (#10345) was in that release that made an attempt at fixing it though the fix wasn't able to be tested.

Since it appears none of the core contributors have access to an on-prem or cloud version of SonarQube, we're kinda stuck without more details to help figure out what's going on if 2.35.1 doesn't fix the issue.

Nsai1997:
Hi team,did we implement the fix for the above issue in the 2.35.1.As we need to follow a large process to setup so want to confirm before going ahead?

mtesauro:
The PR I mentioned (#10345 ) was in 2.35.1 so yes a fix for the issue was added to that version.

As I said above:
**made an attempt at fixing it though the fix wasn't able to be tested**
because
**it appears none of the core contributors have access to an on-prem or cloud version of SonarQube**

so we _believe_ the fix is valid but don't have a way to confirm that. This is one the unfortunate aspects of API integrations with DefectDojo - they are difficult to test and confirm they work as expected unless someone happens to have that API running in a place where it can be tested.

Unfortunately, we cannot provide any more assurance than the info above.

Nsai1997:
Hi Team,

After reviewing the new release, I've identified a couple of issues that need attention:

[...]

PR #10345, opened 2024-06-05
fix(imp-options): Wrong type-checker in validate_api_scan_configuration

kiblik:
Fix for
- #10343
- https://owasp.slack.com/archives/C2P5BA8MN/p1717610907551889
- https://owasp.slack.com/archives/C2P5BA8MN/p1717611262044439

### DIFF
--- a/dojo/importers/options.py
+++ b/dojo/importers/options.py
@@ -15,9 +15,9 @@ from dojo.models import (
     Endpoint,
     Engagement,
     Finding,
+    Product_API_Scan_Configuration,
     Test,
     Test_Import,
-    Tool_Configuration,
 )
 from dojo.utils import get_current_user, is_finding_groups_enabled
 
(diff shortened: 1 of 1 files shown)


Maffooch (approved):
Thank you for taking care of this one

manuel-sommer:
https://github.com/DefectDojo/django-DefectDojo/issues/10343

kiblik:
2 notes:
- I wrote a fix only based on the error message. I do not have access to SQ right now so I'm not able to perform an end-to-end test of this fix.
- I suppose `dojo/importers/options.py` might need some tests

Maffooch:
> I wrote a fix only based on the error message. I do not have access to SQ right now so I'm not able to perform an end-to-end test of this fix.

This is my situation as well, and the main reason this bug was introduced

> I suppose dojo/importers/options.py might need some tests

Most cases should be covered by the existing test suite as these options are invoked each time the importer class is defined. The issue is that we do not have a reliable way to test API based parsers that would use the `api_scan_configuration` field. I would say that this is an isolated incident

cneill (approved):

mtesauro (approved):
Approved

grendel513 (approved):

Issue #9241, opened 2023-12-29
Make Titlecase configurable


DarkR0ast:
**Problem**
DefectDojo currently always applys titlecasing to save the title for findings. This is very inconvinient when working with different languages which do not work with English titlecasing. Therefore the current implementation creates spelling errors which later need to be corrected manually when, for example, creating a report.

**Solution**
Include a configuration option using an env variable to enable/disable titlecasing.

manuel-sommer:
I looked into it. This is not very straight forward as this is how models.CharField of Django works by default. 
What is your opinion @Maffooch ?

StephanPillhofer:
@manuel-sommer I couldn't find any reference regarding that titlecase is applied by default. Can you maybe post a link?

Also, if it really is the default behaviour, why is the tranformation currently done again within the source code of DefectDojo? 

Thank you!

manuel-sommer:
https://github.com/DefectDojo/django-DefectDojo/blob/738dca4534382049ca7587a60fb6b33715213943/dojo/models.py#L2183

manuel-sommer:
Ah, my bad, sorry for the wrong info, I deleted the last comment (missinformation)

kiblik:
I would maybe propose removal (keep only truncate to 511) and we will see how the community will react.

manuel-sommer:
Will that change also affect deduplication of findings?

kiblik:
> Will that change also affect deduplication of findings?

Ufff, it will. And quite a lot :-/

kiblik:
Maybe making it option is not that bad idea

manuel-sommer:
Btw: I uploaded cyclonedx files and these were not saved as titlecase. Take a look here:
https://github.com/DefectDojo/django-DefectDojo/issues/9474

StephanPillhofer:
That's strange. Based on the code I woud assume titlecasing is always applied when saving a finding.

StephanPillhofer:
Any update on that matter?

manuel-sommer:
Where exactly do you edit the source code to fix the problem? [Here? ](https://github.com/DefectDojo/django-DefectDojo/issues/9241#issuecomment-1933799461)

StephanPillhofer:
yes, we just comment this line

StephanPillhofer:
@manuel-sommer 

[...]

PR #10357, opened 2024-06-07
Make titlecasing for finding title configurable

StephanPillhofer:
New feature to optionally disable titlecasing for the title of findings.

This implements/fixes https://github.com/DefectDojo/django-DefectDojo/issues/9241


StephanPillhofer:
recalculated settings hash to fix failing unit tests

StephanPillhofer:
I do understand your fears of wrong hashcodes.

However enabling this configuration option would allow DefectDojo to be used with so many other languages other than English out of the box. Titlecasing only works in English and causes errors in most other languages.

Do you maybe have an idea how to make the config more "hidden" or harder to accidentally change? I can also include documentation about the implications of changing this config (in code and in the official docs)

In the end it would be sad to limit DefectDojo to English findings only, especially if this missing config is the only thing holding it back. (for German e.g.)

Thank you for your thoughts!

Maffooch:
> Another idea could be to always create the titlecase version of the title of the finding right before it get's hashed, no matter the current setting. Therfore, even if the setting is changed just to play with it, the resulting hash will still be the same.

Just to be sure I understand this correctly, are you saying something like this?
```
orignial_title = self.title
 # Title Casing
from titlecase import titlecase
self.title = titlecase(self.title[:511])
...
>> Do the hash code stuff
...
if not settings.FINDING_TITLECASING_ENABLED:
    self.title = original_title
```

StephanPillhofer:
I had a different solution in mind when writing my comment but I now realize that that wouldn't have worked.

So yes, something like you wrote could solve the hashing issue:

```
...

orignial_title = self.title
 # Title Casing
from titlecase import titlecase
self.title = titlecase(self.title[:511])

...

if dedupe_option:
    if (self.hash_code is not None):
        deduplicationLogger.debug("Hash_code already computed for finding")
    else:        
        self.hash_code = self.compute_hash_code()
        deduplicationLogger.debug(
            "Hash_code computed for finding: %s", self.hash_code)
        if not settings.FINDING_TITLECASING_ENABLED:
            self.title = original_title

...
```

mtesauro:
Closed as stale

StephanPillhofer:
Since this PR has been closed, can I assume that such a feature will not be approved in the future? Otherwise I could update this PR with the changes discussed above.

mtesauro:
@StephanPillhofer The PR was closed as stale as part of a general review of PRs which hadn't seen activity in a while.

Such a feature could be considered in the future. The key issue is that such a change would need to happen in the 'view' layer of DefectDojo - in the Django templates so that the title change is only done for UI operations and the title stored in the DB is left unchanged. If the PR changes how titles are stored in the DB, there's some seriously negative side effects for existing DefectDojo instances.

Due to the fact that title is frequently used in the deduplication hash codes, making a change to how those are stored is a problem for existing calculated dedup hashes for anyone already using DefectDojo.  For people with large numbers of findings, recalculating hashes can represent a significant time/resource issue. The largest one I know of with public numbers is: 
image
(full deck [here](https://www.slideshare.net/slideshow/intro-to-defectdojo-at-owasp-switzerland/246800332))

I understand that in your case, you're current running with the titlecase code commented out so a change wouldn't impact your instance but most users won't be in that situation. We could wrap those lines in an if but there's an on going support issue for a new `DD_something` setting for a small portion of the community with may be impacted by this. 

So, we're open to a PR which addresses the issue your having but it would have to consider the impact on the other members of the community currently using DefectDojo - especially because its very possible to make a change which appears to not have any side effects but actually breaks deduplication for users silently or at least with no easy to notice warning.

As DefectDojo has gained popularity over the years, we've had to be more careful about changes which impact those currently using the applications.

Hope that helps explain our thinking around this.

PR #10361, opened 2024-06-07
fix(Risk_Acceptance): Remove redundancy in strings of Treatments

kiblik:
Make strings translatable and remove redundancy

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -3614,21 +3614,21 @@ class Risk_Acceptance(models.Model):
     TREATMENT_FIX = 'F'
     TREATMENT_TRANSFER = 'T'
 
-    TREATMENT_CHOICES = [
-        (TREATMENT_ACCEPT, 'Accept (The risk is acknowledged, yet remains)'),
-        (TREATMENT_AVOID, 'Avoid (Do not engage with whatever creates the risk)'),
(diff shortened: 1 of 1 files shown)


Maffooch:
Have you check if the API needs to be updated? I cannot tell without doing some testing

https://github.com/DefectDojo/django-DefectDojo/blob/501d172aee0c165c8ed8efe1c9fe496a006f437b/dojo/api_v2/serializers.py#L1479C1-L1485C72

Maffooch (approved):

mtesauro (approved):
Approved

kiblik:
> Have you check if the API needs to be updated? I cannot tell without doing some testing
> 
> https://github.com/DefectDojo/django-DefectDojo/blob/501d172aee0c165c8ed8efe1c9fe496a006f437b/dojo/api_v2/serializers.py#L1479C1-L1485C72

Yes, it works.

kiblik:
Btw, Django 5.0 supports more [flexible form (not just list of tuples but using of a dict)](https://docs.djangoproject.com/en/5.0/releases/5.0/#more-options-for-declaring-field-choices).
After [upgrade to 5.0](https://github.com/DefectDojo/django-DefectDojo/pull/10409) I would rewrite all models to dicts like here.

grendel513 (approved):

PR #10374, opened 2024-06-10
Entrypoint Scripts: Add container level breakouts to prevent doom loopings

Maffooch:
In cases where the `settings.dist.py` hash is incorrect in incoming PRs, containers will boot loop during unit tests until they timeout. By default this timeout is 360 minutes, or 6 hours. This prevents units tests from actually running for all other PRs

This PR adds a breakout condition after 10 seconds of not being able to query the database through Django's `db_shell` to allow unit tests to exit quicker. In addition, I also added some timeouts for GHA action jobs just in case we run into a similar situation down the road

[sc-6360]

### DIFF
--- a/.github/workflows/build-docker-images-for-testing.yml
+++ b/.github/workflows/build-docker-images-for-testing.yml
@@ -36,6 +36,7 @@ jobs:
       - name: Build
         id: docker_build
         uses: docker/build-push-action@v5
+        timeout-minutes: 10
         with:
           context: .
           push: false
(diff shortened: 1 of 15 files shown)


cneill (approved):

blakeaowens (approved):

mtesauro (approved):
Approved

mtesauro:
Thanks for putting some great guardrails on our GH Actions :rocket:

grendel513 (approved):

PR #10384, opened 2024-06-11
Allow setting --max-fd argument to uwsgi to stop it from getting OOMKilled in Kubernetes

tmablunar:
**Description**

Reopening https://github.com/DefectDojo/django-DefectDojo/pull/9564

This PR fixes the issue described in issue https://github.com/DefectDojo/django-DefectDojo/issues/9562 regarding uWSGI that under some circumstances will take up an unnecessary amount of resources on a kubernetes node leading to the pod getting OOMKilled.

We introduce the possibility to set the --max-fd argument when starting up uWSGI to mitigate this issue.

**Test results**

I have tested the fix on a kubernetes cluster where it prevented the pod from getting OOMKilled. For more information see https://github.com/DefectDojo/django-DefectDojo/issues/9562.

**Documentation**

It is not clear to me where the documentation should be updated.

### DIFF
--- a/helm/defectdojo/templates/configmap.yaml
+++ b/helm/defectdojo/templates/configmap.yaml
@@ -38,6 +38,7 @@ data:
   DD_UWSGI_PASS: unix:///run/defectdojo/uwsgi.sock
   DD_UWSGI_NUM_OF_PROCESSES: '{{ .Values.django.uwsgi.app_settings.processes | default 2 }}'
   DD_UWSGI_NUM_OF_THREADS: '{{ .Values.django.uwsgi.app_settings.threads | default 2 }}'
+  DD_UWSGI_MAX_FD: '{{ .Values.django.uwsgi.app_settings.max_fd | default 1048576 }}'
   DD_DJANGO_METRICS_ENABLED: '{{ .Values.monitoring.enabled }}'
   NGINX_METRICS_ENABLED: '{{ .Values.monitoring.enabled }}'
   METRICS_HTTP_AUTH_USER: {{ .Values.monitoring.user | default "monitoring" }}
--- a/helm/defectdojo/templates/configmap.yaml
+++ b/helm/defectdojo/templates/configmap.yaml
@@ -38,7 +38,7 @@ data:
   DD_UWSGI_PASS: unix:///run/defectdojo/uwsgi.sock
   DD_UWSGI_NUM_OF_PROCESSES: '{{ .Values.django.uwsgi.app_settings.processes | default 2 }}'
   DD_UWSGI_NUM_OF_THREADS: '{{ .Values.django.uwsgi.app_settings.threads | default 2 }}'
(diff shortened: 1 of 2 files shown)


tmablunar:
I can see that the tests are failing due to changes being made to files that that I (or @hoeg) is not allowed to alter. How should I handle this?

tmablunar:
> It's OK to have a failed test only for this one GH Action: image
> 
> Basically, we use that to alert reviewers that one of the files that 'important' to DefectDojo has been modified by someone who isn't a reviewer/core contributor. Those are files that have historically had bugs in them, require migrations or other things the reviewers need to keep in mind. It's sort of a 'heads up' to reviewers. Thanks for asking 👍
> 
> However, the k8s tests do need to pass - I'm going to re-kick them now but a trick you can use in the future is to close and them re-open the PR which will re-run the tests since you likely don't have permission to re-kick tests.

It does not seem like the k8s tests has been run even after closing and reopening the PR. At least the status is not shown. Could you re-kick them again or advise how to get them run?

helm/defectdojo/templates/configmap.yaml
```diff
@@ -38,6 +38,7 @@ data:
   DD_UWSGI_PASS: unix:///run/defectdojo/uwsgi.sock
   DD_UWSGI_NUM_OF_PROCESSES: '{{ .Values.django.uwsgi.app_settings.processes | default 2 }}'
   DD_UWSGI_NUM_OF_THREADS: '{{ .Values.django.uwsgi.app_settings.threads | default 2 }}'
+  DD_UWSGI_MAX_FD: '{{ .Values.django.uwsgi.app_settings.max_fd }}'
```
cneill, line 41: Could you add this new value to the `values.yaml` file so people know it's an available option? I'm not sure what the right default value would be, so if you want you can just leave it commented out to start with.

cneill (changes requested):

helm/defectdojo/templates/configmap.yaml
tmablunar: This is fixed now. Have requested a new review

helm/defectdojo/values.yaml
```diff
@@ -234,6 +234,7 @@ django:
     app_settings:
       processes: 2
       threads: 2
+      # max-fd: 102400 # Uncomment to set the maximum number of file descriptors. If not set will be detected by uwsgi
```
cneill, line 237:
```suggestion
      # max_fd: 102400 # Uncomment to set the maximum number of file descriptors. If not set will be detected by uwsgi
```

cneill (changes requested):
This should be good to go after a quick typo fix

helm/defectdojo/values.yaml
hoeg: Thank you, I added your suggestion 👍

Maffooch (approved):

cneill (approved):

grendel513 (approved):

mtesauro (approved):
Approved

PR #10386, opened 2024-06-11
Finding hash/dedupe changes

dogboat:
**Description**

This patch updates methods related to Finding hash computation and deduplication to support custom functionality. It refactors the setting of hash codes on a Finding into its own method. It adds a new method that, given a setting name, attempts to use the value of that setting to load a function and return it. It uses that method to attempt to load custom functionality in place of that new Finding set-hash method and the Finding Helper dedupe methods, falling back to existing functionality if no custom settings exist.

**Test results**

Findings still save as expected; if custom functionality is specified, it's called instead.


mtesauro (approved):
Approved

blakeaowens (approved):

Maffooch (approved):

cneill (approved):

PR #10389, opened 2024-06-12
:bug: fix trivy operator deduplication setting

manuel-sommer:
Provide the mitigation information as soon as it becomes available. Then, the finding can be fixed.


Maffooch:
@manuel-sommer I am curious about this one.. Do you have an example for why the mitigation needs to be added to deduplication settings? This also makes me wonder if maybe there is some info in the mitigation that would be better served in the description or something

manuel-sommer:
Hi @Maffooch. So the point is that we either need the description or the mitigation field to be used in the deduplication. The reason is the following scenario.
1. A CVE gets detected from Trivy Operator. But a fix is not yet existing. Thus, a developer does not have the information to fix the vulnerability:
grafik
2. A fix is now present which is recognized by the trivy operator. Then, this information should be parsed from DefectDojo and become available through deduplication as there is no mechanism right now to update an existing finding:
grafik

Maffooch:
Ahh I see.. This does make more sense now. I think the description would be more appropriate than the mitigation

manuel-sommer:
Done @Maffooch. Could we merge this fast please? It would be awesome if we could get it into the next release.

Maffooch (approved):

hblankenship (approved):

PR #10393, opened 2024-06-12
Jira Finding Group Templates: Correct object links

Maffooch:
When findings groups are pushed to Jira, the URLs to represent the product, engagement, test, and even findings, are not functioning correctly. The reason for this is because the variable used to define the URL for each of the objects above is not actually in scope:
```
{% url 'view_finding_group' finding_group.id as finding_group_url %}
{% url 'view_product' finding.test.engagement.product.id as product_url %}
{% url 'view_engagement' finding.test.engagement.id as engagement_url %}
{% url 'view_test' finding.test.id as test_url %}
```
`finding_group` is present, and that links works, but `finding` is not in scope, and therefore does not produce the correct URLs

[sc-6434]


mtesauro (approved):
Approved

cneill (approved):

blakeaowens (approved):

PR #10402, opened 2024-06-14
Fix 10401 enhanced naming of discovered findings filter and add same for mitigated

### LINKED ISSUE #10401: Duplicates in Finding filters
The findings filter has duplicates in regards of filtering for the creation of findings.
image

1. there is a filter to define a timeframe within "Date discovered" 
image
2. there are three filter options to define "before", "on" and "after"
image

Is there any purpose behind this duplicates? And if so, why is my second point simply described with "on" and not a bit detailed like "discovered on"?

I looked at PRs in the past and think this came with #9642. Can you maybe bring a bit light into this @FelixHernandez?

quirinziessler:
fixes #10401 and adds same functionality for mitigated.
image


quirinziessler:
If there is some comment on #10402 I would also style the discovered filter the same way as I did with mitigated. This would mean 2 filters less on the already imo overfilled filters menu.

quirinziessler:
closing because of wrong rebasing

PR #10407, opened 2024-06-14
fix(loc): Do not use _(...) in combination with format

kiblik:
For the correct handling of translations, in the strings that use string interpolation (placing variables to translated text), preparation steps need to be performed in the following order:
1. translate string with placeholders
2. replace placeholders with variables

If steps are executed in opposite order, it is not possible to perform translation.

Context: https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#internationalization-in-python-code

### DIFF
--- a/dojo/notifications/helper.py
+++ b/dojo/notifications/helper.py
@@ -127,11 +127,11 @@ def create_notification(event=None, **kwargs):
 def create_description(event, *args, **kwargs):
     if "description" not in kwargs.keys():
         if event == 'product_added':
-            kwargs["description"] = _('Product {title} has been created successfully.'.format(title=kwargs['title']))
+            kwargs["description"] = _('Product %s has been created successfully.') % kwargs['title']
         elif event == 'product_type_added':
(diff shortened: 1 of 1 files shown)


Maffooch:
@kiblik why do we not want to use _(...) in combination with format? Could you please add the reasoning to the PR description?

kiblik:
> @kiblik why do we not want to use _(...) in combination with format? Could you please add the reasoning to the PR description?

@Maffooch, no problem. I updated the description. I hope it is understandable and more valuable.

mtesauro (approved):
Approved

Maffooch (approved):

cneill (approved):
From the link above:
> Since string extraction is done by the xgettext command, only syntaxes supported by gettext are supported by Django. **In particular, Python [f-strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) are not yet supported** by xgettext, and JavaScript template strings need gettext 0.21+.

PR #10410, opened 2024-06-15
feat(unittests): Avoid run 'dojo_async_task' in async mode (during te…

kiblik:
if a function is running as part of unittest, it should not be performed in async mode:
- to be able to test functions decorated with `dojo_async_task`
- to test that the task will not fail in the background in celery

`dedupe_option=False` is needed in `unittests/test_utils_deduplication_reopen.py` to avoid triggering dedup at that moment.

### DIFF
--- a/dojo/decorators.py
+++ b/dojo/decorators.py
@@ -1,4 +1,5 @@
 import logging
+import sys
 from functools import wraps
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


Maffooch:
I think for cases where we want async tasks to run in the foreground, we do something like this https://github.com/DefectDojo/django-DefectDojo/blob/0bc597eafff463eb07cab39bb5b31043e917a72a/unittests/test_deduplication_logic.py#L131C1-L140C32

kiblik:
> I think for cases where we want async tasks to run in the foreground, we do something like this https://github.com/DefectDojo/django-DefectDojo/blob/0bc597eafff463eb07cab39bb5b31043e917a72a/unittests/test_deduplication_logic.py#L131C1-L140C32

Thanks @Maffooch. This works for me.

PR #10409, opened 2024-06-15
feat(django): Upgrade to 5.0

kiblik:
We released 4.2 so we can start migration to 5.0

Needs to be considered: https://docs.djangoproject.com/en/5.0/releases/5.0 

Needs to be fixed (to be able to upgrade)
- [ ] #10420

Post upgrade improvements (new useful features):
- [ ] choices: From list of tuples to dict https://docs.djangoproject.com/en/5.0/releases/5.0/#more-options-for-declaring-field-choices

PR #10420, opened 2024-06-17
fix(multiselectfield): Use original repo

kiblik:
https://github.com/goinnn/django-multiselectfield looks like alive again (https://github.com/goinnn/django-multiselectfield/issues/141) with more fixes (including Django 5 support - needed for #10409) than https://github.com/DefectDojo/django-multiselectfield.

`django-multiselectfield` in version `0.1.13` was released and it is pinned in `requirements.txt` now

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,9 +11,7 @@ django-dbbackup==4.1.0
 django-environ==0.11.2
 django-filter==24.2
 django-imagekit==5.0.0
-# This library is very outdated, but is a pillar of DefectDojo
-# django-multiselectfield==0.1.12
-git+https://github.com/DefectDojo/django-multiselectfield@master#egg=django-multiselectfield
+django-multiselectfield==0.1.13
 django-polymorphic==3.1.0
 django-crispy-forms==2.2
 django_extensions==3.2.3


Maffooch:
If we move away from the fork, I would prefer we wait for a pinned release. It looks like it should be happening any day now!
https://github.com/goinnn/django-multiselectfield/commit/d78d65219ec600a4078ad2b7d69e7af28c2681a3

kiblik:
> If we move away from the fork, I would prefer we wait for a pinned release. It looks like it should be happening any day now! [goinnn/django-multiselectfield@d78d652](https://github.com/goinnn/django-multiselectfield/commit/d78d65219ec600a4078ad2b7d69e7af28c2681a3)

Sorry for not updating the description. I already used the pinned version.

Maffooch (approved):
I should have looked at the diff again before commenting that 😅

kiblik:
There is missing git tag and release but it was already published to pip server: https://pypi.org/project/django-multiselectfield/#history

cneill (approved):

blakeaowens (approved):

hblankenship (approved):

PR #10426, opened 2024-06-18
Finding Reports: Support string based filtering

Maffooch:
This PR extends the string based filtering to the finding report filter. Currently, all of the following objects the user has access to are fetched on each page load:
- Product types
- Products
- Endpoints
- Engagements
- Tests
- Test Types
- Users,
- Tags (for each of the models)

On instances with large amounts of data, rendering this filter is as follows for my server 40k findings
- Object based filtering: ~15 seconds
- String based filtering: ~1.5 seconds

[sc-6622]

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -3012,8 +3012,8 @@ class ReportFindingFilterWithoutObjectLookups(ReportFindingFilterHelper, Finding
     review_requested_by = CharFilter(
         field_name="review_requested_by__username",
         lookup_expr="iexact",
-        label="Review Request By Username",
-        help_text="Search for Review Request By names that are an exact match")
+        label="Review Requested By Username",
+        help_text="Search for Review Requested By names that are an exact match")
     review_requested_by_contains = CharFilter(
         field_name="review_requested_by__username",
         lookup_expr="icontains",
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -3118,7 +3118,7 @@ class ReportFindingFilterWithoutObjectLookups(ReportFindingFilterHelper, Finding
             self.delete_tags_from_form(product_refs)
             self.delete_tags_from_form(engagement_refs)
             self.delete_tags_from_form(test_refs)
-        if self.engagement:
+        elif self.engagement:
             self.delete_tags_from_form(product_type_refs)
             self.delete_tags_from_form(product_refs)
             self.delete_tags_from_form(engagement_refs)
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -595,6 +595,10 @@ class FindingTagStringFilter(FilterSet):
         help_text="Search for tags on a Product that are an exact match, and exclude them",
         exclude=True)
 
+    def delete_tags_from_form(self, tag_list: list):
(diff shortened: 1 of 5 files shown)


mtesauro (approved):
Approved

dojo/filters.py
```diff
@@ -2961,10 +2972,161 @@ def __init__(self, *args, **kwargs):
...
+    last_reviewed_by_contains = CharFilter(
+        field_name="last_reviewed_by__username",
+        lookup_expr="icontains",
+        label="Last Reviewed By Username Contains",
+        help_text="Search for Last Reviewed By usernames that contain a given pattern")
+    review_requested_by = CharFilter(
+        field_name="review_requested_by__username",
+        lookup_expr="iexact",
+        label="Review Request By Username",
+        help_text="Search for Review Request By names that are an exact match")
```
cneill, line 3016:
```suggestion
        label="Review Requested By Username",
        help_text="Search for Review Requested By names that are an exact match")
```

dojo/filters.py
```diff
@@ -2961,10 +2972,161 @@ def __init__(self, *args, **kwargs):
...
+    review_requested_by_contains = CharFilter(
+        field_name="review_requested_by__username",
+        lookup_expr="icontains",
+        label="Review Request By Username Contains",
+        help_text="Search for Review Request By usernames that contain a given pattern")
```
cneill, line 3021:
```suggestion
        label="Review Requested By Username Contains",
        help_text="Search for Review Requested By usernames that contain a given pattern")
```

dojo/filters.py
```diff
@@ -2961,10 +2972,161 @@ def __init__(self, *args, **kwargs):
...
+    test__engagement__product__name = CharFilter(
+        field_name="test__engagement__product__name",
+        lookup_expr="iexact",
+        label="Product Name",
+        help_text="Search for Product names that are an exact match")
+    test__engagement__product__name_contains = CharFilter(
+        field_name="test__engagement__product__name",
+        lookup_expr="icontains",
+        label="Product name Contains",
+        help_text="Search for Product Typ names that contain a given pattern")
```
cneill, line 3061:
```suggestion
        help_text="Search for Product names that contain a given pattern")
```

dojo/filters.py
```diff
@@ -2961,10 +2972,161 @@ def __init__(self, *args, **kwargs):
...
+            "not_test__tags",
+            "not_test__tags_contains",
+        ]
+
+        if self.test:
+            self.delete_tags_from_form(product_type_refs)
+            self.delete_tags_from_form(product_refs)
+            self.delete_tags_from_form(engagement_refs)
+            self.delete_tags_from_form(test_refs)
+        if self.engagement:
```
cneill, line 3121: Seems like this should be an `elif` since it won't have any effect if the condition above was true.

```suggestion
        elif self.engagement:
```

grendel513 (approved):

cneill (changes requested):
Just a couple small things

blakeaowens (approved):

cneill (approved):

PR #10473, opened 2024-06-28
feat(docker): Use Python 3.12 in docker images

kiblik:
Next try for #10333


Maffooch:
Hey @kiblik I'm closing this one out for now until are ready to make this move

valentijnscholten:
reopening this as the final hurdle should be resolved by pygithub was updated.

kiblik:
Looks like we are finally ready 🎉

valentijnscholten (changes requested):
can you update https://github.com/DefectDojo/django-DefectDojo/blob/bugfix/readme-docs/CONTRIBUTING.md#python3-version as well?

I am also wondering if we should add another line somewhere to explicitly indicate Defect Dojo requires Python 3.12+?

kiblik:
> can you update https://github.com/DefectDojo/django-DefectDojo/blob/bugfix/readme-docs/CONTRIBUTING.md#python3-version as well?
> 

done (btw, I found one more)

> I am also wondering if we should add another line somewhere to explicitly indicate Defect Dojo requires Python 3.12+?

Not sure if it is explicitly necessary. As soon as you start collaborating, you will notice which `ruff` rules will fail and what is not working. And it is quite easy to notice the version in Docker images. How often is DefectDojo running outside of Docker?
I'm not against this idea. I just do not see it as necessary. But if you have some explicit proposal, I'm happy to add it.

valentijnscholten (approved):

kiblik:
@mtesauro, @Maffooch 🙏

kiblik:
@valentijnscholten, can we target this PR to `2.50.0` as well?

mtesauro:
We have a bunch of changes landing in 2.50.0 so I wonder if it makes sense to hold off till 2.51.0 for this change.

kiblik:
I believe, this might be reviewed and merged.

PR #10348, opened 2024-06-06
Feat(psql): Use psycopg3

kiblik:
Django 4.2 introduced support for Psycopg 3

> Psycopg 3 support[¶](https://docs.djangoproject.com/en/5.0/releases/4.2/#psycopg-3-support)
> Django now supports [psycopg](https://www.psycopg.org/psycopg3/) version 3.1.8 or higher. To update your code, install the [psycopg library](https://pypi.org/project/psycopg/), you don’t need to change the [ENGINE](https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-DATABASE-ENGINE) as django.db.backends.postgresql supports both libraries.
> 
> Support for psycopg2 is likely to be deprecated and removed at some point in the future.
> Source: https://docs.djangoproject.com/en/5.0/releases/4.2/#psycopg-3-support

This PR is blocked by #9493

PR #10491, opened 2024-07-02
fix(deps): build psycopg3 instead of using pre-build binary

gietschess:
**Description**

Since the update of the base python docker image from alpine:3.16 to 3.20 we've had segmentation faults while connecting to aws rds postgres databases. With the debug environment variable from the psycopg2 dependency we found connection issues while connecting to our database (see logs below).

Building the psycopg3 postgresql adapter instead of using the pre-build dependency as it's described in the [psycopg3 installation documentation](https://www.psycopg.org/psycopg3/docs/basic/install.html#local-installation) the connection could be established again as expected.

Also see the [slack thread](https://owasp.slack.com/messages/C2P5BA8MN/p1717509947946239) on this issue.

**Test results**

I've locally build and tested both the alpine and debian django +nginx dockerimages and tested them against the bitnami and aws rds postgres database.

**Extra information**

Additional logs:

Debug logs with the alpine image containing the psycopg2-binary 2.9.9 dependency:
```
[19] psyco_connect: dsn = 'dbname=dd client_encoding=UTF8 user=aws_admin password=bla host=defectdojo-rds-dev port=5432', async = 0
[19] connection_setup: init connection object at 0x7f18ab9af240, async 0, refcnt = [1] connection_setup: init connection object at 0x7f18ab9af240, async 0, refcnt = 11
[19] con_connect: connecting in SYNC mode
[19] conn_connect: new PG connection at 0x7f18aaf8d4b0
[19] conn_connect: PQconnectdb(dbname=dd client_encoding=UTF8 user=aws_admin password=bla host=defectdojo-rds-dev port=5432) returned BAD
[19] connection_init: FAILED
!!! uWSGI process 19 got Segmentation Fault !!!
```

Debug logs with the alpine image containing the self-build psycopg2 2.9.9 dependency:
```
[19] psyco_connect: dsn = 'dbname=dd client_encoding=UTF8 user=aws_admin password=bla host=defectdojo-rds-dev port=5432', async = 0
[19] connection_setup: init connection object at 0x7f87c53e3240, async 0, refcnt = 1
[19] con_connect: connecting in SYNC mode
[19] conn_connect: new PG connection at 0x7f87c49248a0
[19] conn_connect: server standard_conforming_strings parameter: on
[19] conn_connect: server requires E'' quotes: NO
[19] conn_connect: using protocol 3
[19] conn_connect: client encoding: UTF8
[19] clear_encoding_name: UTF8 -> UTF8
[19] conn_set_fast_codec: encoding=UTF8
[19] conn_set_fast_codec: PyUnicode_DecodeUTF8
[19] conn_connect: DateStyle ISO, MDY
[19] connection_setup: good connection object at 0x7f87c53e3240, refcnt = 1
```

### DIFF
--- a/Dockerfile.django-alpine
+++ b/Dockerfile.django-alpine
@@ -16,13 +16,15 @@ RUN \
     bind-tools \
     mysql-client \
     mariadb-dev \
-    postgresql14-client \
+    postgresql16-client \
     xmlsec \
     git \
     util-linux \
     curl-dev \
     openssl \
     libffi-dev \
+    python3-dev \
+    libpq-dev \
     && \
     rm -rf /var/cache/apk/* && \
   true
@@ -49,7 +51,7 @@ RUN \
     xmlsec \
     git \
     util-linux \
-    postgresql14-client \
+    postgresql16-client \
     curl-dev \
     openssl \
     # needed for integration-tests
--- a/Dockerfile.nginx-alpine
+++ b/Dockerfile.nginx-alpine
@@ -16,13 +16,15 @@ RUN \
     bind-tools \
     mysql-client \
     mariadb-dev \
-    postgresql14-client \
+    postgresql16-client \
     xmlsec \
     git \
     util-linux \
     curl-dev \
     openssl \
     libffi-dev \
+    python3-dev \
+    libpq-dev \
     && \
     rm -rf /var/cache/apk/* && \
   true
--- a/requirements.txt
+++ b/requirements.txt
@@ -35,7 +35,7 @@ Markdown==3.6
 mysqlclient==2.1.1
 openpyxl==3.1.5
 Pillow==10.4.0  # required by django-imagekit
-psycopg[binary]==3.2.1
+psycopg[c]==3.2.1
 cryptography==42.0.8
 python-dateutil==2.9.0.post0
 pytz==2024.1


Dockerfile.django-alpine
```diff
@@ -16,13 +16,15 @@ RUN \
     bind-tools \
     mysql-client \
     mariadb-dev \
-    postgresql14-client \
+    postgresql16-client \
```
mtesauro, line 19: What's the thinking on rev'ing this to postgresql16? Some distros (RHEL 8 comes to mind) don't have support for Postgres less then 13 - which is still supported by the upstream project till Nov 2025 (see [here](https://www.postgresql.org/support/versioning/)).

While I'm running PostgreSQL 16, not all of the broader DefectDojo community has the option to run that version.

Dockerfile.nginx-alpine
```diff
@@ -16,13 +16,15 @@ RUN \
     bind-tools \
     mysql-client \
     mariadb-dev \
-    postgresql14-client \
+    postgresql16-client \
```
mtesauro, line 19: ditto here about postgresql16-client as above :point_up:

mtesauro (changes requested):

mtesauro:
@gietschess 

Note: There's a PR to move to psycopg3 that's closer to being merged as well #10348

Dockerfile.django-alpine
gietschess: hi @mtesauro ,
in the default django docker image which is based in debian we're installing the postgres-client in version 15. the defectdojo helm chart postgres database dependency uses version 15. since the postgres client is backward compatible with older versions (see the [Notes section](https://www.postgresql.org/docs/current/app-psql.html)) there should be no issue using a older postgres database instance.

gietschess:
> @gietschess
> 
> Note: There's a PR to move to psycopg3 that's closer to being merged as well #10348

with this pr we would still use the [not recommended prebuild binary version](https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary) of psycopg.

i don't really mind about the postgres clients version, just wanted to keep it up to date as alpine informs about using a old client version when installing version 14:
```
* Setting postgresql14 as the default version
*
* You are using 'postgresql14'. It's recommended to upgrade to the latest
* major version provided by package 'postgresql16'.
* Use command 'pg_versions' to switch between versions.
*
```

kiblik:
Original tests have not been failing. Is there some way to improve tests to detect SegFault earlier?

gietschess:
hi @kiblik ,
i couldn't reproduce the segfault with the bitnami postgres database from the helmchart and the alpine docker image but with a external postgres database.
my best guess would be that the segfault results due to some encryption configuration of the database but i'm not completly sure about that.

Dockerfile.django-alpine
mtesauro: OK. That works for me - I just don't want to optimize for one distro and break others if we don't absolutely have to.

mtesauro:
@gietschess 

> > Note: There's a PR to move to psycopg3 that's closer to being merged as well #10348
> 
> with this pr we would still use the [not recommended prebuild binary version](https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary) of psycopg.

Yes, but if we're moving to psychopg3 then maybe do a PR for the non-binary version of psychopg3 instead of the non-binary version of psychopg2. At least that's what I was thinking.

At one point in time we switched to the binary version of psychopg for a specific reason but that was quite some time ago so I don't have a problem with using the non-binary/source version of that module.

I'd prefer not to use non-binary 2 over binary 3 of that module when non-binary 3 is a valid option.

> 
> i don't really mind about the postgres clients version, just wanted to keep it up to date as alpine informs about using a old client version when installing version 14:

The one place where this can cause problems is for iron installs - we've planning on deprecating that but it's not "official' yet.  I just don't want to optimize for one distro and break others if we don't absolutely have to - from a quick look at building psychopg3, only libpq5. Just checked with the most "interesting" distro (RHEL 8) and pip has no issues building psychopg3 assuming you have the needed Postgres dev packages installed.

So I'll make my comments as resolved as far as the PG 16 is concerned.

gietschess:
alright, i've tested with the prebuild psycopg3 binary alpine image which also leads into segmentation faults.
with the self build ones both debian and alpine looked fine.

mtesauro:
> alright, i've tested with the prebuild psycopg3 binary alpine image which also leads into segmentation faults.
> with the self build ones both debian and alpine looked fine.

OK. Perfect. Feel free to change this PR to use the source/you-build-it psycopg3 or close it and open a new one if that's easier for you.

I'm happy to take a PR to move to a built version of psycopg3 instead of the  binary version that's currently in place after #10348 was merged.

mtesauro (approved):
Just realized you already changed to psycopg[c] from psycopg[binary]

Approved

Maffooch (approved):

cneill (approved):

grendel513 (approved):

hajali-amine:
Do we know when this will be released? We're facing the same issue with OVH managed PG databases 👀

mtesauro:
> Do we know when this will be released? We're facing the same issue with OVH managed PG databases 👀

@hajali-amine This went into the dev branch so will be part of the 2.37.0 (August) release. Minor version releases happen on the 1st Monday of the month or for August 5th.

If you need it sooner than that, you can always change requirements.txt like the PR does and build your own docker images using the dockerfiles in this repo.

Issue #10490, opened 2024-07-02
Helm: problem with psql in 2.36.0


kiblik:
Problematic combination: Django `4.2` and PSQL `11`

Logs:

```python
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/usr/local/lib/python3.11/site-packages/django/utils/asyncio.py", line 26, in inner
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/db/backends/base/base.py", line 330, in cursor
    return self._cursor()
           ^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/db/backends/base/base.py", line 306, in _cursor
    self.ensure_connection()
  File "/usr/local/lib/python3.11/site-packages/django/utils/asyncio.py", line 26, in inner
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/db/backends/base/base.py", line 289, in ensure_connection
    self.connect()
  File "/usr/local/lib/python3.11/site-packages/django/utils/asyncio.py", line 26, in inner
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/db/backends/base/base.py", line 272, in connect
    self.init_connection_state()
  File "/usr/local/lib/python3.11/site-packages/django/db/backends/postgresql/base.py", line 309, in init_connection_state
    super().init_connection_state()
  File "/usr/local/lib/python3.11/site-packages/django/db/backends/base/base.py", line 239, in init_connection_state
    self.check_database_version_supported()
  File "/usr/local/lib/python3.11/site-packages/django/db/backends/base/base.py", line 214, in check_database_version_supported
    raise NotSupportedError(
django.db.utils.NotSupportedError: PostgreSQL 12 or later is required (found 11.22).
```

Issue is under investigation

mtesauro:
@kiblik 
PostgreSQL 11 is pretty old - in fact it's no longer supported as of November 9, 2023 by the PostgreSQL project -https://www.postgresql.org/support/versioning/

Do we have PostgreSQL 11 somewhere in our Helm charts? If so, we really shouldn't.

[...]

PR #10493, opened 2024-07-02
fix(exit in bash): Fix handling exit in initializer

kiblik:
The original `exit` (e.g. from #9002) worked correctly in `sh`.
However, by adding `shellcheck`, #9147 changed `sh` to `bash` which handles these situations differently.
Linter introduced an error that nobody noticed. 
Issue discovered during investigation of #10490


Maffooch:

PR #10496, opened 2024-07-02
feat(k8s-test): Try login and obtain API token

kiblik:
Simple "give me login page" is not enough. It should be possible to log in to the working application.

This PR add test for:
- simple login
- simple API test - get token

Added based on #10490


Maffooch:

PR #10497, opened 2024-07-02
feat(db-checker): Extension of "db reachable"

kiblik:
Extend `wait_for_database_to_be_reachable`. Not only for simple operation but check that DB is compatible.

Added based on https://github.com/DefectDojo/django-DefectDojo/issues/10490


Maffooch:

PR #10505, opened 2024-07-03
fix: Dockerfile warnings

fcecagno:
**Description**

When building the latest code, I got a few warnings from Docker:

```
5 warnings found (use --debug to expand):
- FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 8)
- FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 9)
- FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 35)
- FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 144)
- LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 40)
```

Docker version:
```
$ docker --version
Docker version 27.0.1, build 7fafd33
```

This PR doesn't change any functionality, just handles the warnings.


kiblik:
Some linter would be helpful. Would this one help? https://github.com/marketplace/actions/docker-lint If so, it might be added.

Maffooch (approved):

mtesauro (approved):
Approved

grendel513 (approved):

PR #10509, opened 2024-07-03
fix(doc): Disable markup.highlight.guessSyntax + enable mermaid

kiblik:
I found [`mermaid`](https://mermaid.js.org/) quite handy during the preparation of documentation for #7311.

Unfortunately, when `mermaid` is enabled `guessSyntax` has to be disabled otherwise graphs are not rendered (because `guessSyntax` marks `mermaid` snippets as `language-fallback` and they are not rendered at all).

I added languages to places where they have been probably guessed in the past. Now they do not need to be guessed.


Maffooch (approved):

cneill (approved):

mtesauro (approved):
Approved

hblankenship (approved):

PR #10533, opened 2024-07-08
fix(api-notif): Fix order of validators

kiblik:
During the implementation of #7311, I noticed specific edge cases that the implemented `validate` function did not handle correctly.

If there is an existing System Notification (user=None, product=None, template=false), it is not possible to add a Notification Template. It failed with `Notification for user and product already exists` even though it wasn't true. So I add a `template` flag to the checking query. However, I also needed to reorder checkers because "existing template" would fail with `Notification for user and product already exists` so a reorder was needed.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -2951,6 +2951,7 @@ class NotificationsSerializer(serializers.ModelSerializer):
     def validate(self, data):
         user = None
         product = None
+        template = False
 
         if self.instance is not None:
             user = self.instance.user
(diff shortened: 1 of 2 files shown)


mtesauro:
Closing and re-opening to try to make Flake8 test happy

mtesauro:
@kiblik Not sure what's up with flake8-your-pr since we have a PR that Maffooch did in bugfix to remove the need for that test now that we have ruff running.

Anyway, don't worry about making it happy for this PR. We'll sort out why in other PRs :+1:

mtesauro (approved):
Approved

cneill (approved):

Maffooch (approved):

grendel513 (approved):

PR #10545, opened 2024-07-09
Bulk Edit: Add note when pushing finding to jira

Maffooch:
When attempting to use bulk edit to add notes while also pushing changes to jira, the notes are not pushed as well. This PR will add an additional step to the `push_to_jira` logic to also consider the new note added by bulk editt

[sc-6732]


cneill (approved):

mtesauro (approved):
Approved

grendel513 (approved):

blakeaowens (approved):

PR #10553, opened 2024-07-10
Fix(django): Upgrade of 4.2.14

kiblik:
While waiting for the acceptance of #10409, it might be a good idea to upgrade Django at least to `4.2.14` to mitigate known issues: https://docs.djangoproject.com/en/5.0/releases/4.2.14/

This change is not offered by `dependabot` because it is monitoring only the latest versions (in running configuration).


mtesauro (approved):
Approved

Maffooch (approved):

cneill (approved):

grendel513 (approved):

PR #10555, opened 2024-07-10
Endpoint get or create: Do not raise warning when there is an existing endpoint

Maffooch:
When a single endpoints is found in the endpoint get or create helper method, we do not want to raise a warning


cneill (approved):

mtesauro (approved):
Approved

grendel513 (approved):

blakeaowens (approved):

PR #10562, opened 2024-07-12
Test Types: Return support for disabling test types via the `active` flag

Maffooch:
Return the ability to remove a given test type from being used, but without deleting the test type. This change has the following impacts for test types with a False `active` status:
- Import forms will not list inactive test types
- Creating/editing a test will not allow for an inactive test type to be selected

There is a slight caveat to this such that the API swagger schema does not appear to be calling the function to generate the list of test types each time the schema is loaded. The test type list is updated on uwsgi reload. My best guess is the schema is being cached on the server somehow. 

[sc-6869]

### DIFF
--- a/dojo/tools/factory.py
+++ b/dojo/tools/factory.py
@@ -48,7 +48,12 @@ def get_parser(scan_type):
 
 
 def get_inactive_test_types():
-    return Test_Type.objects.filter(active=False).values_list("name", flat=True)
+    try:
+        return list(Test_Type.objects.filter(active=False).values_list("name", flat=True))
+    except Exception:
+        # This exception is reached in the event of loading fixtures in to an empty database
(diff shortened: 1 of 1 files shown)


kiblik:
I'm not 100% sure, this will work in multi-Django deployment (k8s, number of pods `>1`). `signal` will probably trigger reload of `PARSERS` only in the container in which `Test_type` has been disabled.

Maffooch:
That is a good point. Maybe it really would be better to totally move away from the "in memory" list of parsers. I'll see what I can do

mtesauro (approved):
Approved

hblankenship (approved):

grendel513 (approved):

cneill (approved):

PR #10565, opened 2024-07-13
remove coverage from requirements.txt

manuel-sommer:
it looks like coverage is not used anymore

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,6 @@ asteval==1.0.1
 bleach==6.1.0
 bleach[css]
 celery==5.4.0
-coverage==7.6.0
 defusedxml==0.7.1
 django_celery_results==2.5.1
 django-auditlog==2.3.0


manuel-sommer:
reopening because of merge conflict failure

manuel-sommer:
I can't see a correlation between the failed tests and the PR @Maffooch. Could you take a look here?

mtesauro:
@manuel-sommer So, coverage isn't *directly* used by DefectDojo but it's used in some of our developer docs which is why we've kept it in requirements.txt

I'm not opposed to removing it since you'd like to think Python developers could easily `pip install` it if they wanted to check coverage for DefectDojo. We just talked about that module a while ago and kept it for the reason above.

So it's optionally 'used' by DefectDojo developers but not 'used' by the app itself.

@Maffooch Do you have an opinion on keeping or removing the coverage module?

Maffooch:
I vote we remove this. It will make docker images slightly smaller as well

If developers would like to test their unit test coverage, install the package is as easy as 
```
pip install coverage
```

mtesauro:
> I vote we remove this. It will make docker images slightly smaller as well
> 
> If developers would like to test their unit test coverage, install the package is as easy as
> 
> ```
> pip install coverage
> ```

Works for me!

cneill (approved):

Maffooch (approved):

grendel513 (approved):

Issue #10171, opened 2024-05-09
Reauth Gitlab doesn't work v.2.34.2 (and lower)


G1P0:
Hi. I use DefectDojo without SSL in my local environment for educational purpose. 

There's bug: OAuth 2.0 through Gitlab doesn't work when you try to "Login with Gitlab" for the second time. First time - it's done, it's ok. 
But for the second time (e.g. I've added new repositories in local gitlab and I need to sync them) - there's 500 Internal Server Error.

1. Add new app in Gitlab for DD
2. Add gitlab's env in DD's docker-compose (key, secret, api_url, aouth_enabled, auto_import, scope)
3. Click on "Login with Gitlab" -> approve auth on local gitlab -> it's done, everything works
4. Logout from Gitlab user 
5. Login with Gitlab again -> approve auth on local gitlab -> 500 Internal Server Error 

- [ X ] Docker Compose

 Operating System: Ubuntu Server 22.04
 Bug's DefectDojo is somewhere between 2.33.0 and 2.34.2
 Worked Defectdojo 2.33.0 

Logs from uwsgi: 
```
with self.db.wrap_database_errors:
  File "/usr/local/lib/python3.11/site-packages/django/db/utils.py", line 91, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/usr/local/lib/python3.11/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django.db.utils.IntegrityError: duplicate key value violates unique constraint "social_auth_usersocialauth_provider_uid_e6b5e668_uniq"
DETAIL:  Key (provider, uid)=(gitlab, 1) already exists.

[pid: 1|app: -|req: -/-] 192.168.56.1 (-) {48 vars in 1418 bytes} [Thu May  9 11:02:44 2024] GET /complete/gitlab/?code=*token* Id => generated 12833 bytes in 217 msecs (HTTP/1.1 200) 7 headers in 223 bytes (1 switches on core 0)
```
In v. 2.33.0 everything worked fine. 
e.g. logs with the reproduced steps:
```
# I logout after successful Gitlab login 
logout user: **gitlab-user** via ip: 192.168.56.1
GET /logout => generated 0 bytes in 27 msecs (HTTP/1.1 302) 10 headers in 549 bytes (1 switches on core 1)

[...]

PR #10566, opened 2024-07-15
Fix Reauth Gitlab doesn't work

Spanish93:
Fix for issue : https://github.com/DefectDojo/django-DefectDojo/issues/10171


mtesauro:
@Spanish93 Thanks for the PR however we only take PRs against the dev or bugfix branch.  Dependabot already has a PR for this in #10568 so we can go with that one.

PR #10582, opened 2024-07-17
/import helptext correction: endpoint_to_add

paulOsinski:
The endpoint_to_add option on POST /import is incorrectly described in the API docs:
Screenshot 2024-07-17 at 5 11 02 PM

Helptext says that this field allows you to add a new Endpoint, but this is inaccurate. Instead, this method looks for an existing Endpoint ID in the database and associates it with the incoming Findings.

I am changing the help text so that it is described accurately. I am also reflecting the help text to/reimport because this method works in the same way on that endpoint.

- Tested /import endpoint to ensure feature works as described
- Tested /reimport endpoint to ensure feature works as described
[sc-6963]

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -2064,7 +2064,7 @@ class ImportScanSerializer(serializers.Serializer):
         queryset=Endpoint.objects.all(),
         required=False,
         default=None,
-        help_text="The IP address, host name or full URL. It must be valid",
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

hblankenship (approved):

cneill (approved):

mtesauro (approved):
Approved

PR #10594, opened 2024-07-19
Aqua: Improve exception handling for reports without vulnerabilities :bug:

### LINKED ISSUE #10585: Import Aqua Scan with sensitive datas
**Bug description**
We are integrating Aqua Scan to DefectDojo. The AquaScanner scans for vulnerabilities, but also for sensitive datas, and those appear on the scan report. When I Import the scan manually in the Defect-dojo UI, I have an error 'NoneType'. Also, when this happens, the scan results is still added to the engagement, but with 0 findings in it (which is not the case).

**Steps to reproduce**
1. Scan an image with the aquascanner with sensitive datas in it (I scanned Juice-shop for example). Take the json file.
2. Import the Aqua scan results to defect-dojo
3. I have this error in the UI
image
 
I looked in the logs, and it looks like the Aqua parser is iterating over each vulnerabilities of each resources, but the sensitive datas don't have that key, they have sensitive_data-items instead. 
image

image

**Expected behavior**
The Aqua parser can handle an Aqua Scan having sensitive datas in it.

I also think that when that kind of error occurs, it should be better to not add a test with 0 findings in defect dojo as this is misleading.

manuel-sommer:
Catch cases where an aqua report may not have any vulnerabilities. In this case, the aqua parser should not through an exception, but simply pass on that section of the report

closes #10585

### DIFF
--- a/dojo/tools/aqua/parser.py
+++ b/dojo/tools/aqua/parser.py
@@ -24,8 +24,9 @@ class AquaParser:
 
             for node in vulnerabilityTree:
                 resource = node.get("resource")
-                vulnerabilities = node.get("vulnerabilities")
-
+                vulnerabilities = node.get("vulnerabilities", [])
+                if vulnerabilities is None:
+                    vulnerabilities = []
                 for vuln in vulnerabilities:
                     item = get_item(resource, vuln, test)
                     unique_key = resource.get("cpe") + vuln.get("name", "None")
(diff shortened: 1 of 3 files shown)


cneill (approved):

grendel513 (approved):

mtesauro (approved):
Approved

Issue #10611, opened 2024-07-23
Add an other Aqua Scan Format


kzzz1:
I have an Aqua Scan Report coming from Azure Devops job. The file has slightly different json structure, having outter square brackets and 2 more lines at the beginning (image_name and results). The rest is the same, I think the part we want is in the 'results' json key. That file gives me 0 findings when I import it like this. Is it possible to add that new format? 

![image](https://github.com/user-attachments/assets/4eb75dcd-9b60-41a0-aa8a-e799b103a9db)

manuel-sommer:
Could you add an example file?

kzzz1:
sure
[AquaSecurity-test-local-202407191.json](https://github.com/user-attachments/files/16349480/AquaSecurity-test-local-202407191.json)

manuel-sommer:
See PR @kzzz1

PR #10616, opened 2024-07-23
:bug: extend aqua format issue #10611

manuel-sommer:
#10611

### DIFF
--- a/dojo/tools/aqua/parser.py
+++ b/dojo/tools/aqua/parser.py
@@ -19,9 +19,19 @@ class AquaParser:
 
     def get_items(self, tree, test):
         items = {}
-        if "resources" in tree:
+        if isinstance(tree, list):
+            vulnerabilityTree = tree[0]["results"]["resources"]
+            for node in vulnerabilityTree:
+                resource = node.get("resource")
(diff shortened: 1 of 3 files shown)


dojo/tools/aqua/parser.py
```diff
@@ -19,9 +19,19 @@ def get_findings(self, json_output, test):
 
     def get_items(self, tree, test):
         items = {}
-        if "resources" in tree:
+        if isinstance(tree, list):
+            vulnerabilityTree = tree[0]["results"]["resources"]
```
cneill, line 23: Instead of duplicating all the code related to walking the tree, why not just set the `vulnerabilityTree` variable based on the format we received? A brief comment explaining why this is necessary also wouldn't hurt. Finally, there should be a length check on `tree` to make sure the reference to `[0]` won't cause an index out-of-bounds error on an empty report.

cneill (approved):
Looks good, thank you!

mtesauro (approved):
Approved

blakeaowens (approved):

hblankenship (approved):

PR #10617, opened 2024-07-23
Listing Tables: Add toggle switch in system settings

Maffooch:
For a long time, it has been a point of confusion to have the java script based filtering and sorting option on a given listing page. The filtering/sorting only applies to the items rendered on the page (by default this is 25) rather than all items in the database.  

To alleviate the confusion, there will now be a toggle switch in the system settings that allows users to disable the "page based" filtering and sorting mechanism. To maintain the expected functionalities that users have been used to seeing, this new toggle will default to enabled

[sc-6997]


mtesauro (approved):
Approved

blakeaowens (approved):

cneill (commented):
Just a small suggestion for the help text, otherwise looks good

cneill (approved):

PR #10623, opened 2024-07-24
Option Compression: Accommodate unsaved objects a bit better

Maffooch:
When compressing importer options, we need to accommodate any unsaved objects a bit better

[sc-7013]


grendel513 (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #10636, opened 2024-07-26
Finding notes cascading deletes

dogboat:
**Description**

This patch adds 'cascading deletes' to Notes and NoteHistory entries for all models that have a Notes relationship (i.e., Engagement, Test, Finding, Risk_Acceptance, Tool_Product_Settings, Cred_User, and Benchmark_Product). It's implemented via pre-delete signals so that this cleanup will take place even when delete() is performed on a QuerySet rather than an individual model.

[sc-3671]

**Test results**

Manual testing done to test deletion of Notes attached to the above-mentioned models.


mtesauro (approved):
Approved

blakeaowens (approved):

hblankenship (approved):

Maffooch (approved):

PR #10639, opened 2024-07-26
Engagement: Add missing permission check to view an Engagement

Maffooch:
Returns the permission check when viewing an engagement after it was omitted during the conversion to CBV


cneill (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

hblankenship (approved):

PR #10641, opened 2024-07-27
Refresh Helm Chart Lock File: The removal

Maffooch:
The refresh helm chart lock file action uses the `pull_request_target` trigger, which can lead to leaking secret. Because the helm chart lock file is updated on each modification to the chart.yml file by renovate/dependabot, the easiest solution is to remove this action.


mtesauro (approved):
Approved

blakeaowens (approved):

hblankenship (approved):

grendel513 (approved):

PR #10645, opened 2024-07-29
Importer: Correct logic bug for empty scan reports

Maffooch:
When importing an empty scan report through the import endpoint, it is possible for two tests to be created during a single request


mtesauro (approved):
Approved

cneill (approved):

blakeaowens (approved):

hblankenship (approved):

PR #10689, opened 2024-08-06
Change ingress netpol

C4tWithShell:
**Description**

Change ingress network policy template to allow to set peers as well as ports

### DIFF
--- a/helm/defectdojo/templates/network-policy.yaml
+++ b/helm/defectdojo/templates/network-policy.yaml
@@ -13,14 +13,9 @@ spec:
   podSelector:
     matchLabels:
       app.kubernetes.io/instance: {{ .Release.Name }}
+  {{- if .Value.networkPolicy.ingress}}
   ingress:
-    - from:
-        - podSelector:
-            matchLabels:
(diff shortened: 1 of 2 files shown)


kiblik (changes requested):
I suppose this kind of change might break running deployments. I have no problem make the definition more flexible but can you probably move 
```yaml
    - from:
        - podSelector:
            matchLabels:
```
to default setting/value?

C4tWithShell:
> I suppose this kind of change might break running deployments. I have no problem make the definition more flexible but can you probably move
> 
> ```yaml
>     - from:
>         - podSelector:
>             matchLabels:
>               app.kubernetes.io/instance: {{ .Release.Name }}
> ```
> 
> to default setting/value?

Right. I changed logic. So it will not affect running deployments

kiblik (approved):

mtesauro (approved):
Approved

cneill (approved):

grendel513 (approved):

Maffooch (approved):

PR #10612, opened 2024-07-23
helm: helm chart enhancements

mikesindieiev:
**Description**

1. Helm chart follows the same variable name style across the `values.yaml` file
2. Setting UWSGI enableDebug to true will automatically set DD_DEBUG environment variable to 'True'
3. Standard mysql configuration is added. Bitnami uses old collation settings by default. 

**Existing DefectDojo database will continue to use old collation**

PR #10692, opened 2024-08-06
Remove DebugPy

Maffooch:
With the recent fiasco around missing pypi images for debugpy that broke out build/testing automation, an opportunity was created to evaluate if we wanted to support this behavior anymore


cneill (approved):

mtesauro (approved):
Approved

kiblik:
Out of curiosity, may I know what was the fiasco? I haven't noticed any.

I use `debug` mode in docker-compose quite often during development. Not exactly `debugpy` but environmental variables used in `docker-compose.override.debug.yml`.
If you are removing `docker-compose.override.debug.yml`, may I ask to set `PYTHONWARNINGS` to `error` and `DD_DEBUG` to `'True'` in all containers in `docker-compose.override.dev.yml` as it was set in `docker-compose.override.debug.yml`? Because `dev` will become the main development environment now.

mtesauro:
@kiblik 

> Out of curiosity, may I know what was the fiasco? I haven't noticed any.

Perfectly reasonable to ask why - no problem.

The thing that happened earlier this week (Monday) and has happened several times in the past is:
(1) debugpy releases an update - can be a minor version bump (1.8.4 in the most recent cast) or release an update only for one architecture (arm64)
(2) Dependabot notices the update and does a PR, we merge it to dev
(3) debugpy yanks the release and the version disappears (this is what happened to 1.8.4). Sometimes they release a new one, sometimes the fall back to the old version. In Monday's case, 1.8.5 came out late on Tuesday (something like that)

Between 2 and 3 a dev tool not used directly by DefectDojo "the app"  breaks building our containers and a bunch of GH Actions.

Since it was a dev thing that was breaking container building in GHA, we decided to yank it and have people manually install it if they used it.

I'm very willing to explore other options as long as they don't break GH Actions and accommodate people who don't use debugpy. In my and @Maffooch 's case, we didn't mind setting up debugpy manually when we needed it but I can see how that would be painful if you used it regularly.

Got any suggestions?

kiblik:
> @kiblik
> 
> > Out of curiosity, may I know what was the fiasco? I haven't noticed any.
> 

Thank you for the explanation

> Got any suggestions?

I support the idea of dropping it and if somebody needs some deep troubleshooting, it is happening usually locally so anybody can add it ad-hoc.
Same idea (with dropping it) came to my mind when I added [ruff `T10`](https://docs.astral.sh/ruff/rules/#flake8-debugger-t10). But I just "fix" it with `noqa` at that moment.

Maffooch:
> If you are removing docker-compose.override.debug.yml, may I ask to set PYTHONWARNINGS to error and DD_DEBUG to 'True' in all containers in docker-compose.override.dev.yml as it was set in docker-compose.override.debug.yml? Because dev will become the main development environment now.

Happy to do so, and great call 😄

grendel513 (approved):

kiblik:
As I see changes from #10612, I just realized, that 
https://github.com/DefectDojo/django-DefectDojo/blob/51cea97bd6bfdf11be3b6809d7f3e58c865303a3/readme-docs/KUBERNETES.md?plain=1#L165-L173
might be removed as well.

Maffooch:
Good catch. Removed

hblankenship (approved):

fopina:
just got past this one and [!devcon](https://github.com/DefectDojo/django-DefectDojo/pull/11216) would have helped restore a debug environment without having yet another compose override 😄

PR #10696, opened 2024-08-07
Rest Framework: Remove Parallel + Shuffle

Maffooch:
The rest framework tests have been very flakey not that they are actually producing failures for GHAs (this part is a good thing)

To increase stability for these tests, we should revert the `--prallel` and `--shuffle` flags

[sc-7127]

### DIFF
--- a/docker/entrypoint-unit-tests-devDocker.sh
+++ b/docker/entrypoint-unit-tests-devDocker.sh
@@ -54,10 +54,11 @@ EOF
 echo "Unit Tests"
 echo "------------------------------------------------------------"
 
-python3 manage.py test unittests -v 3 --keepdb --no-input --failfast --shuffle --parallel --exclude-tag="non-parallel" || {
+# Removing parallel and shufful for now to maintain stability
+python3 manage.py test unittests -v 3 --keepdb --no-input --exclude-tag="non-parallel" || {
     exit 1; 
 }
-python3 manage.py test unittests -v 3 --keepdb --no-input --failfast --shuffle --tag="non-parallel" || {
+python3 manage.py test unittests -v 3 --keepdb --no-input --tag="non-parallel" || {
     exit 1; 
 }
 
--- a/docker/entrypoint-unit-tests-devDocker.sh
+++ b/docker/entrypoint-unit-tests-devDocker.sh
@@ -54,7 +54,7 @@ EOF
 echo "Unit Tests"
 echo "------------------------------------------------------------"
 
-# Removing parallel and shufful for now to maintain stability
+# Removing parallel and shuffle for now to maintain stability
 python3 manage.py test unittests -v 3 --keepdb --no-input --exclude-tag="non-parallel" || {
     exit 1; 
 }
--- a/docker/entrypoint-unit-tests.sh
+++ b/docker/entrypoint-unit-tests.sh
@@ -80,9 +80,10 @@ python3 manage.py migrate
 echo "Unit Tests"
 echo "------------------------------------------------------------"
 
-python3 manage.py test unittests -v 3 --keepdb --no-input --failfast --shuffle --parallel --exclude-tag="non-parallel" || {
+# Removing parallel and shufful for now to maintain stability
+python3 manage.py test unittests -v 3 --keepdb --no-input --exclude-tag="non-parallel" || {
     exit 1; 
 }
-python3 manage.py test unittests -v 3 --keepdb --no-input --failfast --shuffle --tag="non-parallel" || {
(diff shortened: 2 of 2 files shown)


Maffooch:
Tests worked first time.. I'm going to run them again to make sure this is not a false positive

Maffooch:
Twice could be coincidental.. If third luck is  a charm, then I will be happy

Maffooch:
Looks like we are back to stability 😄

docker/entrypoint-unit-tests-devDocker.sh
```diff
@@ -54,10 +54,11 @@ EOF
 echo "Unit Tests"
 echo "------------------------------------------------------------"
 
-python3 manage.py test unittests -v 3 --keepdb --no-input --failfast --shuffle --parallel --exclude-tag="non-parallel" || {
+# Removing parallel and shufful for now to maintain stability
```
cneill, line 57:
```suggestion
# Removing parallel and shuffle for now to maintain stability
```

docker/entrypoint-unit-tests.sh
```diff
@@ -80,9 +80,10 @@ python3 manage.py migrate
 echo "Unit Tests"
 echo "------------------------------------------------------------"
 
-python3 manage.py test unittests -v 3 --keepdb --no-input --failfast --shuffle --parallel --exclude-tag="non-parallel" || {
+# Removing parallel and shufful for now to maintain stability
```
cneill, line 83:
```suggestion
# Removing parallel and shuffle for now to maintain stability
```

docker/entrypoint-unit-tests-devDocker.sh
```diff
@@ -54,10 +54,11 @@ EOF
...
+python3 manage.py test unittests -v 3 --keepdb --no-input --exclude-tag="non-parallel" || {
```
cneill, line 58: Should we resume running these "non-parallel" tests again, at least while we have `--parallel` disabled?

```suggestion
python3 manage.py test unittests -v 3 --keepdb --no-input || {
```

docker/entrypoint-unit-tests-devDocker.sh
```diff
@@ -54,10 +54,11 @@ EOF
...
     exit 1; 
 }
-python3 manage.py test unittests -v 3 --keepdb --no-input --failfast --shuffle --tag="non-parallel" || {
+python3 manage.py test unittests -v 3 --keepdb --no-input --tag="non-parallel" || {
```
cneill, line 61:
```suggestion
python3 manage.py test unittests -v 3 --keepdb --no-input || {
```

docker/entrypoint-unit-tests.sh
```diff
@@ -80,9 +80,10 @@ python3 manage.py migrate
...
+python3 manage.py test unittests -v 3 --keepdb --no-input --exclude-tag="non-parallel" || {
```
cneill, line 84:
```suggestion
python3 manage.py test unittests -v 3 --keepdb --no-input || {
```

docker/entrypoint-unit-tests.sh
```diff
@@ -80,9 +80,10 @@ python3 manage.py migrate
...
     exit 1; 
 }
-python3 manage.py test unittests -v 3 --keepdb --no-input --failfast --shuffle --tag="non-parallel" || {
+python3 manage.py test unittests -v 3 --keepdb --no-input --tag="non-parallel" || {
```
cneill, line 87:
```suggestion
python3 manage.py test unittests -v 3 --keepdb --no-input || {
```

cneill (commented):
Should these also be removing `--failfast`? I'm okay with it given that we've recently improved performance by eliminating MySQL/RabbitMQ from the matrix of tests, and seeing *all* the tests that failed is useful. Just want to be sure it was intentional

Maffooch:
Keeping the tags part since we don't want to run the same unit test command twice. Eventually, we will figure out the parallel business, but today is not that day

Maffooch:
> Should these also be removing --failfast? I'm okay with it given that we've recently improved performance by eliminating MySQL/RabbitMQ from the matrix of tests, and seeing all the tests that failed is useful. Just want to be sure it was intentional

Yes it’s intentional. Since those tests will not take ~8.5 minutes rather than ~3, I think providing users with as much context for failure could reduce the amount of revs (and consumed workers) it takes for contributors to get passing tests

cneill (approved):

mtesauro (approved):
Approved

grendel513 (approved):

hblankenship (approved):

PR #10700, opened 2024-08-07
Ruff: Add and fix TCH

kiblik:
Add and fix https://docs.astral.sh/ruff/rules/#flake8-type-checking-tch
Specially: https://docs.astral.sh/ruff/rules/typing-only-standard-library-import/


mtesauro (approved):
Approved

cneill (approved):

blakeaowens (approved):

PR #10706, opened 2024-08-07
Ruff: Add and fix PLW15

kiblik:
Add Ruff rule PLW15* https://docs.astral.sh/ruff/rules/#warning-w_1
Fix PLW1510 and PLW1514


mtesauro (approved):
Approved

docker/install_chrome_dependencies.py
```diff
@@ -27,7 +27,7 @@ def ldd(file_path):
     # For simplicity, I'm assuming if we get an error, the code is non-zero.
     try:
         result = subprocess.run(
-            ["ldd", file_path], capture_output=True, text=True,
+            ["ldd", file_path], capture_output=True, text=True, check=False,
```
cneill, line 30: Since `subprocess.CalledProcessError` appears to only be raised when `check=True`, we should do that here.

> If check is true, and the process exits with a non-zero exit code, a [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) exception will be raised. Attributes of that exception hold the arguments, the exit code, and stdout and stderr if they were captured. ([ref](https://docs.python.org/3/library/subprocess.html#subprocess.run))

I guess we should leave `check=False` on the other line since the exception would be unhandled? I'm not really sure why we're running `ldd` twice here in the first place

cneill (changes requested):

Maffooch (approved):
Approving as the `check=False` is the default behavior and technically does not change any behavior to specify it explicitly. Handling the exception is bonus points 😄

cneill (approved):
Chatted with @Maffooch and @mtesauro about this, we'll deal with the `check=True` thing in a later PR if we need to.

grendel513 (approved):

PR #10712, opened 2024-08-07
Ruff: Add and fix RUF015

kiblik:
Add and fix RUF015 rule https://docs.astral.sh/ruff/rules/unnecessary-iterable-allocation-for-first-element/


Maffooch:

PR #10723, opened 2024-08-08
Appcheck web application scanner parser

dogboat:
**Description**
This patch adds a parser for AppCheck Web Application Security Scanner. The scanner itself appears to use a variety of underlying engines to produce data, so it implements a few 'engines' as plugins to handle the specific differences in data where observed.

**Test results**
Tests added for the main parser, as well as the supporting engines.

**Documentation**
Documentation has been added.

[sc-6055]


mtesauro (approved):
Approved

grendel513 (approved):

dojo/tools/appcheck_web_application_scanner/parser.py
```diff
@@ -0,0 +1,87 @@
...
+import logging
+import os
+from typing import Any
+
+from dojo.tools.appcheck_web_application_scanner.engines.base import BaseEngineParser
+
+LOGGER = logging.getLogger(__name__)
+
+
+def load_scanning_engine_parsers() -> dict[str, BaseEngineParser]:
```
cneill, line 13: Is there a particular reason to do this dynamically rather than just importing the known engines explicitly? Might encourage people to contribute additional engines (if any) rather than just keeping them to themselves 😉

cneill (commented):
I notice that there are a lot of `[[markup]]` strings showing up in various places in these findings. Probably better to filter that out? I'm not exactly sure what format that is intended to denote, but while it looks like it gets bullets and some other things right, it is also trying to render links with `[[<url>]]`, which isn't valid Markdown and results in the links *not* being clickable. Tables look particularly messy, though I'm not sure there's much we can reasonably do about that.

Screenshot 2024-08-09 at 19 28 36
Screenshot 2024-08-09 at 19 29 23
Screenshot 2024-08-09 at 19 30 30
Screenshot 2024-08-09 at 19 32 31

dojo/tools/appcheck_web_application_scanner/parser.py
dogboat: Forward-planning laziness. 😂 Excellent point about encouraging contributions, I've fixed it. Thanks!

dogboat:
> I notice that there are a lot of `[[markup]]` strings showing up in various places in these findings. Probably better to filter that out? I'm not exactly sure what format that is intended to denote, but while it looks like it gets bullets and some other things right, it is also trying to render links with `[[<url>]]`, which isn't valid Markdown and results in the links _not_ being clickable. Tables look particularly messy, though I'm not sure there's much we can reasonably do about that.

Good point; I took a stab at fixing this by removing "[[markup]]", "[[", and "]]" from the title, description (per se as well as any notes/details appended to it), and solution. Not sure if that was too literal a reading of this. Was that what you were hoping for, or were you wanting all 'markup' removed? (A very halfhearted attempt at research suggests maybe this is like wiki markup? Or maybe it's AppCheck's own proprietary spin...)

hblankenship (approved):

cneill (approved):
This looks good, thanks!

Maffooch (approved):
I've spent some time looking through this implementation and it's a really neat approach to mapping fields from an arbitrary place in a JSON blob to a DefectDojo Finding attribute. However, we lose a lot of flexibility and granularity with this approach. For example, a DefectDojo finding has a "References" attribute that the following screenshot (also note that bullets are not properly aligned with the data) of the description would be better suited under. The same could also be said about other sections in the "Details" section of a vulnerability such as Impact, CVEs, External Sources, etc.

I may be incorrect from my read of these engines, but It does not appear there is an easy way to identify where data is coming from within the report, or move that data freely as everything is dynamically set

image


This is not a request for change in any way, but is something to keep in mind when developing your next parser 😄  Maintainability and flexibility is extremely important for open source projects as "code owners" are often a more fluid thing (if they exist at all). The best advice I can offer here is to check out other parsers for inspiration to see how a given parser is interacting with data from submitted reports

PR #10737, opened 2024-08-09
fix(k8s): run k8s test on the latest version of k8s and minikube

kiblik:
Run k8s on the latest version of k8s and minikube, as described in the comment.

### DIFF
--- a/.github/workflows/k8s-tests.yml
+++ b/.github/workflows/k8s-tests.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Setup Minikube
         uses: manusa/actions-setup-minikube@v2.11.0
         with:
-          minikube version: 'v1.31.2'
+          minikube version: 'v1.33.1'
           kubernetes version: ${{ matrix.k8s }}
           driver: docker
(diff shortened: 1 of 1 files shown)


.github/workflows/k8s-tests.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           # are tested (https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#available-versions)
           - databases: pgsql
             brokers: redis
-            k8s: 'v1.26.11'
+            k8s: 'v1.30.3'
```
cneill, line 31: I notice minikube is on [`v1.33.1`](https://github.com/kubernetes/minikube/releases/tag/v1.33.1) now. We should probably try to bump that as well (line 40).

kiblik: I agree. Done

mtesauro (approved):
Approved

cneill (approved):
Thanks!

grendel513 (approved):

PR #10738, opened 2024-08-09
Inline finding images on reports

dogboat:
**Description**

This patch extends file_images template to allow for inline-image generation. When called with format="INFLINE," instead of generating picture tokens and using those to retrieve images, images will be encoded as base64 and embedded directly into the report. It updates the Finding and Endpoint listing widgets to use inline images.

Looking for feedback on this implementation; the PR would target release 2.37.2 or later.

[sc-6857]


Maffooch (approved):

grendel513 (approved):

mtesauro (approved):
Approved

cneill (approved):

PR #10740, opened 2024-08-10
fix(helm): Break circular dependency

### LINKED ISSUE #10727: DB-migration-checker container blocks migration job from running in Helm upgrade (2.36.6 to 2.37.0, Chart version 1.6.143 to 1.6.144)
**Be informative**
We have DefectDojo deployed on a Kubernetes platform. We mostly use the Helm chart and values provided in the standard Helm chart and Values files in this repository. 
Our current deployment is on Chart version 1.6.143 and we're trying to upgrade to 1.6.144. This kept failing and it's probably because of a wrong order of operations in the Helm deployment.

When upgrading to 2.37.0 (Chart version 1.6.144) the created pods now run an init container called 'db-migration-checker'. This container checks if the migration has succeeded or not, and eventually fails if it doesn't detect a migration.
The pod waits until the init container succeeds before starting the other containers. If it doesn't succeed, the pod is terminated and the upgrade fails.

The migration happens in a job called 'initializer'. This job only runs after a succesful installation or upgrade because of these Helm hooks:
```initializer:
  jobAnnotations: {
    helm.sh/hook: "post-install,post-upgrade"
  }
```
On https://github.com/DefectDojo/django-DefectDojo/blob/master/helm/defectdojo/values.yaml#L302

We use a PostgreSQL database if that matters.

**Bug description**
The 'db-migration-checker' init container checks for a migration that only happens until after the upgrade is completed. 
Because the migration never happens, the init container itself fails, which causes the upgrade to fail and thus the migration job never runs. 

The 'db-migration-checker' needs the migration job to succeed, but the migration job doesn't run until the 'db-migration-checker' container (and the rest of the upgrade) succeeds, which never happens. This is a circular dependency that cannot work in the current order of operations.
Upgrading to 2.37.0 when (mostly) using the standard Helm charts and values is not possible at the moment.

#10454 probably introduced this block on the upgrade, causing this circular dependency.

**Steps to reproduce**
1. Have an existing DefectDojo deployment on 2.36.6 (or other pre-2.37.0) deployed via standard Helm chart (1.6.143 or earlier)
2. Use Helm to upgrade to 2.37.0 (Chart version 1.6.144), make sure to keep dbMigrationChecker enabled
3. The init containers 'db-migration-checker' fail
4. The upgrade fails
5. The migration job never starts

**Expected behavior**
The db-migration-checker doesn't block the upgrade, or the migrations job needs to run while the upgrade is underway.

- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version 2.37.0 (from 2.36.6)

Our workaround was to disable the dbMigrationChecker:
```
dbMigrationChecker:
  enabled: false
``` 
This caused the upgrade to succeed. We do think checking the migration status before further deploying the application is a good idea, so we would like this circular dependency to get resolved.

kiblik:
Fix #10727

### DIFF
--- a/helm/defectdojo/templates/initializer-job.yaml
+++ b/helm/defectdojo/templates/initializer-job.yaml
@@ -48,6 +48,24 @@ spec:
           path: {{ .hostPath }}
           {{- end }}
       {{- end }}
+      initContainers:
+      - name: wait-for-db
+        command:
+        - '/bin/bash'
+        - '-c'
+        - '/wait-for-it.sh ${DD_DATABASE_HOST:-postgres}:${DD_DATABASE_PORT:-5432} -t 30 -s -- /bin/echo Database is up'
+        image: '{{ template "django.uwsgi.repository" . }}:{{ .Values.tag }}'
+        imagePullPolicy: {{ .Values.imagePullPolicy }}
+        {{- if .Values.securityContext.enabled }}
+        securityContext:
+          {{- toYaml .Values.securityContext.djangoSecurityContext | nindent 10 }}
+        {{- end }}
+        envFrom:
+        - configMapRef:
+            name: {{ $fullName }}
+        - secretRef:
+            name: {{ $fullName }}
+            optional: true
       containers:
       {{- if .Values.cloudsql.enabled  }}
       - name: cloudsql-proxy
--- a/helm/defectdojo/values.yaml
+++ b/helm/defectdojo/values.yaml
@@ -298,9 +298,7 @@ django:
 
 initializer:
   run: true
-  jobAnnotations: {
-    helm.sh/hook: "post-install,post-upgrade"
-  }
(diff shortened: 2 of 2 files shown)


helm/defectdojo/templates/initializer-job.yaml
```diff
@@ -48,6 +48,35 @@ spec:
           path: {{ .hostPath }}
           {{- end }}
       {{- end }}
+      initContainers:
+      - name: whatForDB
```
mtesauro, line 52: Apparently upper case makes things fail:
image

mtesauro (approved):
Approved

Maffooch (approved):

grendel513 (approved):

cneill (approved):

PR #10760, opened 2024-08-14
restore relative date filter for findings

hblankenship:
[sc-7182]

The ability to filter findings on relative dates (7 days, 30 days, etc) was removed in favor of the specific date range provided by Date Discovered. This restores the ability to filter by a relative date while keeping the specific date.


grendel513 (approved):

dojo/filters.py
```diff
@@ -1544,6 +1544,7 @@ def filter(self, qs, value):
 class FindingFilterHelper(FilterSet):
     title = CharFilter(lookup_expr="icontains")
     date = DateFromToRangeFilter(field_name='date', label="Date Discovered")
+    date_recent = DateRangeFilter(field_name="date", label="Relative Date")
```
Maffooch, line 1547:
```suggestion
    date = DateRangeFilter(field_name="date", label="Date Discovered")
```
Reverting the `date` to being the `DateRangeFilter` will allow us to avoid having to update any dashboard tile changes, and potentially other places that reference the `date` field as an integer (for a choice in the drop down specified in the `DateRangeFilter` class). I cannot make the full suggestion in Github to remove the current `date` filter since it is not in the diff, but the following line needs to be removed:
```
date = DateFromToRangeFilter(field_name='date', label="Date Discovered")
```

The same functionality post removal of the above will be preserved due to the existence of the `on/before/after` filters

Maffooch (changes requested):

Maffooch (approved):

mtesauro (approved):
Approved

cneill (approved):

PR #10786, opened 2024-08-20
feat(api-token): Add ability to use API tokens but not disable "api-token-auth"

kiblik:
`DD_API_TOKENS_ENABLED` is useful for disabling all API tokens.

From now, `DD_API_TOKEN_AUTH_ENDPOINT_ENABLED` might be used when Tokens are needed but only "user+pass -> token" endpoint needs to be disabled.


mtesauro (approved):
Approved

docker-compose.override.dev.yml
```diff
@@ -9,6 +9,7 @@ services:
       DD_ADMIN_USER: "${DD_ADMIN_USER:-admin}"
       DD_ADMIN_PASSWORD: "${DD_ADMIN_PASSWORD:-admin}"
       DD_EMAIL_URL: "smtp://mailhog:1025"
+      DD_API_TOKEN_AUTH_ENDPOINT_ENABLED: "False"
```
cneill, line 12: Is there a specific reason to disable this in the `dev` override config and not others? Seems like a development environment would be pretty low-risk.

kiblik: I'm sorry. It is leftover from testing. I will remove it.

kiblik: I'm sorry. It is leftover from testing. I will remove it.

grendel513 (approved):

blakeaowens (approved):

PR #10789, opened 2024-08-20
Integration Tests: Check for exceptions at build time

Maffooch:
In the event of an exception in the `docker/install_chrome_dependencies.py` file at build time, the exception should not be swallowed, and instead raised to prevent unit tests from churning needlessly

[sc-7194]


docker/install_chrome_dependencies.py
```diff
@@ -5,36 +5,28 @@
...
-    except subprocess.CalledProcessError:
-        stdout = ""
-        code = 1
-    return stdout, code
+    with suppress(subprocess.CalledProcessError):
+        return run_command(["ldd", file_path])
+    return "", 1
 
 
 raw_deps = ldd("/opt/chrome/chrome")
```
cneill, line 32: Doesn't this also need to check the return code, or does `ldd` return a nonzero status code in cases that aren't really errors for us here? Just thinking we may not want to ignore *all* `ldd` errors.

Maffooch: I believe the intention is to return an empty string such we skip the logic of identifying missing packages. We could just allow it to error out instead though If that would be preferable

cneill: It [appears](https://stackoverflow.com/questions/1256145/ld-linux-verify-exit-codes) that `ldd` exits with a nonzero status code when there's some kind of issue reading the executable (or if it's not a valid executable at all). Unfortunately that StackOverflow answer is all I can find - the [`ldd` man page](https://www.man7.org/linux/man-pages/man1/ldd.1.html) doesn't say anything about exit status codes. I suppose we can wait until the executable fails to run since that might give us a more descriptive error

cneill (approved):

mtesauro (approved):
Approved

Maffooch:
This test was ran when we knew integration tests would fail, and the error was still not caught earlier.. It seems like more attention is needed on this one

hblankenship (approved):

Maffooch:
This is the type of error I would except to see when searching for dependencies fails
image

I will now pull the latest dev images. I know those will build correctly, so if the tests are successful with this changes with solid containers, than I believe we will be good here

Maffooch:
Looks like I will need to spend more time on this solution...

Maffooch:
@cneill I swapped in the dynamic chrome version fetching and the builds still failed. The interesting part is that the script is working sorta as expected? `ldd` is not failing this time, but some dependencies are still not appearing to be pulled in
```
#18 [build  9/17] RUN   missing_chrome_deps=$(python install_chrome_dependencies.py) &&   apt-get -y install $missing_chrome_deps
#18 0.101 Reading package lists...
#18 0.507 Building dependency tree...
#18 0.614 Reading state information...
#18 0.731 0 upgraded, 0 newly installed, 0 to remove and 13 not upgraded.
#18 DONE 0.7s

#19 [build 10/17] RUN apt-get install -y libxi6 libgconf-2-4 jq libjq1 libonig5 libxkbcommon0 libxss1 libglib2.0-0 libnss3   libfontconfig1 libatk-bridge2.0-0 libatspi2.0-0 libgtk-3-0 libpango-1.0-0 libgdk-pixbuf2.0-0 libxcomposite1   libxcursor1 libxdamage1 libxtst6 libappindicator3-1 libasound2 libatk1.0-0 libc6 libcairo2 libcups2 libxfixes3   libdbus-1-3 libexpat1 libgcc1 libnspr4 libgbm1 libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libxcb1 libxext6   libxrandr2 libxrender1 gconf-service ca-certificates fonts-liberation libappindicator1 lsb-release xdg-utils
```

That seems out of the scope of this PR. Does the current implementation of check the error codes of `ldd` satisfy your request?

grendel513 (approved):

cneill (approved):
Looks good! We can circle back on the other issues with installing Chrome in the future

PR #10802, opened 2024-08-23
Bump selenium

cneill:
**Description**

This moves the dependencies currently installed in the integration tests Dockerfile into a dedicated `requirements.txt` and bumps Selenium from 4.9.0 to 4.23.0.


Maffooch:

PR #10804, opened 2024-08-23
Appcheck null byte fix

dogboat:
**Description**

This patch fixes a bug in the AppCheck parser where it attempted to save a string containing a null byte to the database, causing a crash. This is accomplished with two adjustments:

1) HTTP/2 request/response pairs are now extracted from the source data; this was where the null byte originated. Since HTTP/2 req/res pair entries are formatted differently in the source data than HTTP/1 pairs are, these were being automatically appended to the Finding's description attribute instead, null byte and all. These are no longer appended to the description as they are pulled out.
2) Non-printable characters are filtered from various (string) attributes and replaced with string-representable versions. The attributes this applies to include the description, which was the catch-all attribute for various source-data entries including a "details" object that (sometimes) includes the req/res pairs. As mentioned above, because HTTP/2 req/res pairs were not extracted from the "details" object, they were appended -- null byte and all -- to the description, causing the save failure. Going forward, all things added to this attribute are filtered, which means that this shouldn't happen in the future if the format changes or some new protocol (HTTP/3) is represented. The tradeoff, of course, is that we might not immediately notice if we're not properly capturing req/res pairs... but at least it won't crash. :)

**Test results**

New unittests added. Manual testing performed with supplied data.

[sc-7273]


grendel513 (approved):

cneill (approved):

mtesauro (approved):
Approved

hblankenship (approved):

PR #10812, opened 2024-08-26
feat(unittest): Not avoid PLW1514

kiblik:
It might be better to fix based on recommendations - not avoid it.

### DIFF
--- a/unittests/tools/test_appcheck_web_application_scanner_parser.py
+++ b/unittests/tools/test_appcheck_web_application_scanner_parser.py
@@ -222,7 +222,7 @@ class TestAppCheckWebApplicationScannerParser(TestCase):
             self.assertEqual(3, len(findings))
 
     def test_appcheck_web_application_scanner_parser_http2(self):
-        with open("unittests/scans/appcheck_web_application_scanner/appcheck_web_application_scanner_http2.json") as testfile:  # noqa: PLW1514
+        with open("unittests/scans/appcheck_web_application_scanner/appcheck_web_application_scanner_http2.json", encoding="utf-8") as testfile:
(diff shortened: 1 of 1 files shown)


Maffooch:
I elected to avoid the encoding due the testing of removal of the `\00` char a few lines into the test. I wasn't sure if the attempting to encode that null char would break something, or if we wanted to the encode that the file at all

@dogboat what do you think would be the best approach here?

dogboat:
I think this should be fine. Encodings are super weird but I tested out reading the file explicitly as utf-8 locally and it loads the chars as expected; unsaved requests get the null byte as expected, and it's stripped out/replaced where it should be. We could add something like `self.assertTrue("\x00" in finding.unsaved_response)` to each of the checks if we want to make sure that stays preserved, dunno why I didn't do that previously.

kiblik:
> I think this should be fine. Encodings are super weird but I tested out reading the file explicitly as utf-8 locally and it loads the chars as expected; unsaved requests get the null byte as expected, and it's stripped out/replaced where it should be. We could add something like `self.assertTrue("\x00" in finding.unsaved_response)` to each of the checks if we want to make sure that stays preserved, dunno why I didn't do that previously.

I would prefer `self.assertIn("\x00", finding.unsaved_response)` (I tried to offer it in #10817 as well). But why not?

Will you prepare it or should I? As part of this PR or will you open new one?

Maffooch (approved):
If there `AssertIn` change has been made already as part of #10817, then I thing this one is good to go

cneill (approved):

blakeaowens (approved):

mtesauro (approved):
Approved

PR #10817, opened 2024-08-27
feat(unittests): Try to avoid assertTrue/False

kiblik:
Python supports many useful checkers in unittests which (if a test does not pass) are able to provide useful information about the context (why it is not going through, on which position).

It might be good to prefer these `assert*`. And it is not possible to add context as an additional parameter for faster troubleshooting.


cneill (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #10818, opened 2024-08-27
fix(helm): Longer wait for DB

kiblik:
During a couple of last PRs, I noticed that sometimes k8s are failing. Reason: The first run of the initializer fails while waiting on DB. The second run pass successfully.

I suppose 30 seconds might not be enough for DB, 5min should be an acceptable compromise.


cneill (approved):

mtesauro (approved):
Approved

Maffooch (approved):

grendel513 (approved):

Issue #7235, opened 2022-12-05
The Initializer container Job is stuck and doesn't complete the Job because of the CloudSQL Proxy


bruno561:
**Be informative**
When using Cloud SQL as a managed database, the Job related to Initializer isn't completed because your Sidecar is still running. Sidecar is cloudsql-proxy.

**Bug description**
As described in the informational section, the Initializer Job fails when using CloudSQL. NotReady status is displayed.

Image of the logs of the container named Initializer. Logs stop at this point:
![image](https://user-images.githubusercontent.com/65369082/205642316-2a74f35e-f5f4-4770-8c0e-34b9d32ab0fc.png)

Image of Pod related to Job with status of NotReady.
![image](https://user-images.githubusercontent.com/65369082/205642600-35a3a7ef-b82b-4e37-abdc-2c0dbadb89ce.png)

**Steps to reproduce**
Basically, just use Cloud SQL in the Values file.
```
  # Google CloudSQL support in GKE via gce-proxy
  cloudsql:
    # To use CloudSQL in GKE set 'enable: true'
    enabled: true
```

**Expected behavior**
Have some way that when the initializer container finishes the job the cloudsql-proxy container is deleted, in order to finish the Job.

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
- GKE (Google Kubernetes Engine) version 1.23.12-gke.100
- https://github.com/DefectDojo/django-DefectDojo/tree/2.16.2/helm/defectdojo

**Logs** 
![image](https://user-images.githubusercontent.com/65369082/205642316-2a74f35e-f5f4-4770-8c0e-34b9d32ab0fc.png)
The cloudsql-proxy container hangs at this point:
![image](https://user-images.githubusercontent.com/65369082/205643890-18100a29-56b9-4017-9473-df17b52dfb4a.png)

tellierd:
Has anyone worked on a fix for this? I'm having the very same issue. Thx

mtesauro:
@tellierd TBH, Helm isn't an [officially supported deployment method](https://github.com/DefectDojo/django-DefectDojo/issues/7235#issuecomment-1721522661) for DefectDojo so we rely on the community to submit PRs for the Helm chart in this repo.

There are far to many deployment choices for k8s/Helm for an open source project to able to test them all.

jndeverteuil:
[...]

PR #10824, opened 2024-08-28
fix(helm): set cloudsql-proxy as sidecar container to allow initializer and dbmigration to run

jndeverteuil:

**Description**

This PR is to fix an issue when using `cloudsql-proxy` as per the following discussion: https://github.com/DefectDojo/django-DefectDojo/issues/7235

> When using Cloud SQL as a managed database, the Job related to Initializer isn't completed because your Sidecar is still running. Sidecar is cloudsql-proxy.

By setting the `cloudsql-proxy` as a special `initContainer` of type`sidecar` (see [relevant kubernetes doc](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/)), we allow the `wait-for-db` initContainer to reach the database.

> Kubernetes implements sidecar containers as a special case of [init containers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/); sidecar containers remain running after Pod startup.

> These restartable sidecar containers are independent from other init containers and from the main application container(s) within the same pod.

> If an init container is created with its restartPolicy set to Always, it will start and remain running during the entire life of the Pod.

> If you define a Job that uses sidecar using Kubernetes-style init containers, the sidecar container in each Pod does not prevent the Job from completing after the main container has finished.

image

image

Without this, `wait-for-db` is never allowed to reach the database since `cloudsql-proxy` is set as a regular `container` which are running only after `initcontainer` are finished.


helm/defectdojo/templates/celery-beat-deployment.yaml
```diff
@@ -68,15 +68,11 @@ spec:
       {{- end }}
       {{- if .Values.dbMigrationChecker.enabled }}
```
tidusete, line 69: I know that the dbMigrationChecker is enabled by default, but shouldn't be outside this conditional? I think using CloudSQL is not related with the fact that you must have enabled dbMigrationChecker
I tested your MR using an external CloudSQL instance and it only works if dbMigrationChecker is enabled.

helm/defectdojo/templates/celery-worker-deployment.yaml
```diff
@@ -66,15 +66,11 @@ spec:
       {{- end }}
       {{- if .Values.dbMigrationChecker.enabled }}
```
tidusete, line 67: Same problem here

helm/defectdojo/templates/django-deployment.yaml
```diff
@@ -84,15 +84,11 @@ spec:
       {{- end }}
       {{- if .Values.dbMigrationChecker.enabled }}
```
tidusete, line 85: Same problem here

helm/defectdojo/templates/celery-beat-deployment.yaml
kiblik: I agree with out in this point

helm/defectdojo/templates/django-deployment.yaml
kiblik, line 85:
```suggestion
      {{- if or .Values.dbMigrationChecker.enabled .Values.cloudsql.enabled }}
```

helm/defectdojo/templates/django-deployment.yaml
```diff
@@ -108,6 +104,11 @@ spec:
         - "-ip_address_types=PRIVATE"
         {{- end }}
       {{- end }}
+      {{$data := dict "fullName" $fullName }}
+      {{- $newContext := merge . (dict "fullName" $fullName) }}
+      {{- include "dbMigrationChecker" $newContext | nindent 6 }}
+      {{- end }}
```
kiblik, line 110:
```suggestion
      {{- end }}
      {{- end }}
```

kiblik (changes requested):
I suppose this should help.

helm/defectdojo/templates/celery-beat-deployment.yaml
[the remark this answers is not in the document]
tidusete: Exactly

tidusete (commented):
I agree on all what @kiblik suggest

helm/defectdojo/templates/django-deployment.yaml
jndeverteuil: Nice we had the same conclusion, I just hadn't seen the suggestion here otherwise I would've implemented it instead of creating my commit, but same exact line used

helm/defectdojo/templates/celery-beat-deployment.yaml
jndeverteuil: Thanks for the feedback, I solved it in the new commit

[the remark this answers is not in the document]
jndeverteuil: Thank you, implemented!

kiblik (approved):

helm/defectdojo/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "2.40.0-dev"
 description: A Helm chart for Kubernetes to install DefectDojo
 name: defectdojo
-version: 1.6.155-dev
+version: 1.6.156-dev
```
Maffooch, line 5:
```suggestion
version: 1.6.155-dev
```
This only gets revved on release

tidusete (approved):
lgtm

tidusete:
Is there any issue on it? I'm waiting for this fix to end up on a release :pray:

mtesauro (approved):
Approved

tidusete:
Is there any problem with the MR? It looks like it's still Open

mtesauro:
@tidusete 
> Is there any problem with the MR? It looks like it's still Open

PRs require 4 approvals from DefectDojo maintainers and this one is 2 short currently.

Maffooch (approved):

grendel513 (approved):

PR #10829, opened 2024-08-30
Parser for AWS Inspector2 findings

siniysv:
A parser for AWS Inspector2 ("new" Inspector) findings exported with `aws inspector2 list-findings` in JSON format.

**Checklist**

[sc-8014]


dojo/tools/aws_inspector2/parser.py
```diff
@@ -0,0 +1,281 @@
+import json
+from datetime import UTC, datetime
+
+from dojo.models import Endpoint, Finding
+
+
+class AWSInspector2Parser(object):
```
manuel-sommer, line 7: A lot of AWS users use AWS Security Hub to retrieve all findings. Could you also advance this parser: https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/awssecurityhub/inspector.py ?

Are the returned AWS Security Hub findings of Inspector2 equal to the returned findings of Inspector2 or are they differently formatted?

dojo/settings/settings.dist.py
```diff
@@ -1273,6 +1273,7 @@ def saml2_attrib_map_format(dict):
     "Kiuwan SCA Scan": ["description", "severity", "component_name", "component_version", "cwe"],
     "Rapplex Scan": ["title", "endpoints", "severity"],
     "AppCheck Web Application Scanner": ["title", "severity"],
+    "AWS Inspector2 Scan": ["title"],
```
manuel-sommer, line 1276: Either add more fields as the title is probably not unique enough, or only use uniqueidfromtool. Furthermore, you forgot to edit the settings.dist.py hash sum https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/settings/.settings.dist.py.sha256sum

manuel-sommer (changes requested):

dojo/tools/aws_inspector2/parser.py
mtesauro: It's definitely preferred to have a single parser that understands multiple formats vs multiple parsers - if nothing else, it's easier for people to use who might not know the difference between Inspector JSON versions.

siniysv: thank you for your attention, I'll try to answer below.
Security Hub collects findings from inspector and a few other AWS services, then it converts the findings into one common format (ASFF - AWS Security Finding Format https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-findings-format.html) and provides API to retrieve the findings in that format. AWS Inspector provides own API with own finding format. It is not necessary to use Security Hub in order to use Inspector since those are 2 different AWS services. Answering to your @manuel-sommer  question - yes, they are differently formatted.

I would extend the existing parser If I had a way to easily detect what finding format is used in an uploaded json, although the name of the parser would be misleading since AWS Security Hub and AWS Inspector are two different services. 

If you believe DefectDojo needs only Security Hub parser because it is indeed an AWS service that can collect findings from other services (full list here https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-internal-providers.html), and there is no need to support standalone AWS service report import, I'm totally okay with this approach. Thank you.

mtesauro: Understood about AWS Security Hub vs AWS Inspector.

I would make sense in this case to have a parser just for AWS Inspector not that you've provided more details - TBH, I can't keep up with all the AWS whatever services.

You will need get the ruff linter happy (as well as the other tests) before the PR will be reviewed and merged (assuming we get 4 approvals).

I'm going to re-kick the tests now - if you want to do that yourself, you can close & open the PR to get them to restart assuming this isn't your first PR against DefectDojo.

dojo/fixtures/test_type.json
```diff
@@ -47,5 +47,12 @@
         },
         "model": "dojo.test_type",
         "pk": 7
+    },
+    {
+        "fields": {
+            "name": "AWS Inspector2"
+        },
+        "model": "dojo.test_type",
+        "pk": 8
```
Maffooch, line 56:
```suggestion
```
This file is no longer modified when adding new parsers

dojo/tools/aws_inspector2/parser.py
```diff
@@ -0,0 +1,281 @@
...
+            resource_infos = raw_finding.get("resources", {})
+            impact = []
+            endpoints = []
+            for resource_info in resource_infos:
+                resource_type = resource_info.get("type", None)
+                resource_id = resource_info.get("id", "N/A")
+                resource_details = resource_info.get("details", {})
+                resource_region = resource_info.get("region", "N/A")
+                endpoint_host = f"{resource_type} - {resource_id}"
+                if resource_type == "AWS_EC2_INSTANCE":
```
Maffooch, line 186: It would make the code much easier to read by converting individual if cases into a function to process each resource type. This is

Maffooch (changes requested):

grendel513 (approved):

Maffooch (approved):

mtesauro (approved):
Approved

dojo/settings/settings.dist.py
```diff
@@ -1283,6 +1283,7 @@ def saml2_attrib_map_format(dict):
     "Kiuwan SCA Scan": ["description", "severity", "component_name", "component_version", "cwe"],
     "Rapplex Scan": ["title", "endpoints", "severity"],
     "AppCheck Web Application Scanner": ["title", "severity"],
+    "AWS Inspector2 Scan": ["title", "severity"],
```
manuel-sommer, line 1286: Hi @Maffooch ,
I am not sure if title and severity is sufficient here as Inspector could scan multiple (different) container with the same CVE. Thus, I would recommend to also use the description or information provided in "resources". (e.g. "imageHash": "sha256:c667ad018b538adadb5bd1db75836c8561b23bc7d31d9cd8941fa74f5df93802",)

manuel-sommer (changes requested):

dojo/settings/settings.dist.py
Maffooch: I can do `title, severity, description`. In most cases, the `unique_id_from_tool` will be used instead

hblankenship (approved):

PR #10843, opened 2024-09-02
fix(edit-prod): Add assigned PT to queryset

kiblik:
Until now, when a user has no permission to add Prod or edit ProdType (a.k.a `get_authorized_product_types(Permissions.Product_Type_Add_Product)` is empty) but the same user still has permission to edit product (Maintainer for that specific product), generated edit form for Product contained empty list for `Product Types`. So user wasn't able to edit this product.

Screenshot 2024-09-02 at 16 36 47

From now on, it does not matter which other permissions a user has, the list of product types contains at least ProdType which is already assigned - so it is possible to submit the form.

Screenshot 2024-09-02 at 16 37 01


mtesauro (approved):
Approved

Maffooch (approved):

cneill (approved):

blakeaowens (approved):

PR #10850, opened 2024-09-03
feat(members): List global role members in Prod and ProdType

kiblik:
Until now, UI listed only regular members of Prod or ProdType. From now on, owners of Prods are transparently informed that also `Global_Role` members also have access to their data.

In ProdType:
image

In Prod:
image

### DIFF
--- a/dojo/product/queries.py
+++ b/dojo/product/queries.py
@@ -13,6 +13,7 @@ from dojo.models import (
     App_Analysis,
     DojoMeta,
     Engagement_Presets,
+    Global_Role,
     Languages,
     Product,
     Product_API_Scan_Configuration,
@@ -71,8 +72,15 @@ def get_authorized_members_for_product(product, permission):
 
     if user.is_superuser or user_has_permission(user, product, permission):
         return Product_Member.objects.filter(product=product).order_by("user__first_name", "user__last_name").select_related("role", "user")
-    else:
-        return None
+    return Product_Member.objects.none()
+
+
+def get_authorized_global_members_for_product(product, permission):
+    user = get_current_user()
+
+    if user.is_superuser or user_has_permission(user, product, permission):
+        return Global_Role.objects.filter(group=None, role__isnull=False).order_by("user__first_name", "user__last_name").select_related("role", "user")
+    return Global_Role.objects.none()
 
 
 def get_authorized_groups_for_product(product, permission):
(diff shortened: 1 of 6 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

dojo/product/queries.py
```diff
@@ -75,6 +76,15 @@ def get_authorized_members_for_product(product, permission):
         return None
 
 
+def get_authorized_global_members_for_product(product, permission):
+    user = get_current_user()
+
+    if user.is_superuser or user_has_permission(user, product, permission):
+        return Global_Role.objects.filter(group=None).order_by("user__first_name", "user__last_name").select_related("role", "user")
+    else:
+        return None
```
cneill, line 85: It looks like this includes users who have the global role "None" that do not actually have access to a Product? 

Screenshot 2024-09-13 at 18 02 23

Screenshot 2024-09-13 at 18 03 57

Screenshot 2024-09-13 at 18 04 10

Also, shouldn't we return something like an empty list here instead of `None`? I believe [this conditional](https://github.com/DefectDojo/django-DefectDojo/pull/10850/files#diff-3488d38cb788c93c2606a9f191d4c58f417c1b3bd432b8a83909b2c62cfdc277R298) in the template:

```
{% if product_groups or product_type_groups or global_product_groups %}
```

Could result in a "NoneType object is not iterable" error when trying to loop over it [here](https://github.com/DefectDojo/django-DefectDojo/pull/10850/files#diff-3488d38cb788c93c2606a9f191d4c58f417c1b3bd432b8a83909b2c62cfdc277R351﻿) if `global_product_groups` is `None` but one of the other 2 values is non-`None`:

```
{% for type_group in global_product_groups %}
```

kiblik: I forgot about these cases. Thank you.
I thought that the unsettling Global role, record was removed.

I would prefer `...objects.none()` instead. To keep the same data-type as the return type.
In `if`s it is evaluated as `False` and there is no problem with iterations:
```python
>>> bool(Product_Type_Member.objects.none())
False
>>> for i in Product_Type_Member.objects.none():
...     print(i)
...
>>>
```

blakeaowens (approved):

grendel513 (approved):

PR #10889, opened 2024-09-10
Product API scan config: Display view scan configs button for all products

Maffooch:
Viewing a product's API scan config is permissible if the product can be viewed. However, the UI would only display the "View API Scan Config" button if the user had permission to at least edit a product. This PR corrects the permission checks in the UI to display the button when a user has permission to view a product

[sc-7458]


mtesauro (approved):
Approved

cneill (approved):

grendel513 (approved):

blakeaowens (approved):

PR #10890, opened 2024-09-10
Update view_finding.html

37b:
### Description

View Finding fields are all wrapped in `<pre>` tags, breaking markdown formatting. This PR just removes those tags and allows markdown formatting to be parsed correctly.

### Test results

No changes to tests

### Documentation

None

### DIFF
--- a/dojo/templates/dojo/view_finding.html
+++ b/dojo/templates/dojo/view_finding.html
@@ -920,7 +920,7 @@
                         class="glyphicon glyphicon-chevron-up"></i></a></span></h4>
             </div>
             <div id="vuln_desc" class="panel-body finding-description collapse in">
-              <pre>{{ finding.description|markdown_render|default_if_none:"" }}</pre>
+              {{ finding.description|markdown_render|default_if_none:"" }}
             </div>
         </div>
 
@@ -968,7 +968,7 @@
                         class="glyphicon glyphicon-chevron-{% if finding.mitigation %}up{%else%}down{%endif%}"></i></a></span></h4>
             </div>
             <div id="vuln_mitigation" class="panel-body collapse {% if finding.mitigation %}in{%endif%}">
-                <pre>{{ finding.mitigation|markdown_render|default_if_none:"" }}</pre>
+                {{ finding.mitigation|markdown_render|default_if_none:"" }}
             </div>
         </div>
         {% if finding.burprawrequestresponse_set.all %}
@@ -986,7 +986,7 @@
                                         <i class="glyphicon glyphicon-chevron-down"></i></a></span></h4>
                             </div>
                             <div id="vuln_req_{{ forloop.counter }}" class="panel-body collapse">
-                                <pre>{{ req_resp.get_request }}</pre>
+                                {{ req_resp.get_request }}
                             </div>
                         </div>
                         <div class="panel panel-default">
@@ -997,7 +997,7 @@
                             </div>
 
                             <div id="vuln_res_{{ forloop.counter }}" class="panel-body collapse">
-                                <pre>{{ req_resp.get_response }}</pre>
+                                {{ req_resp.get_response }}
                             </div>
                         </div>
                     {% endfor %}
@@ -1010,7 +1010,7 @@
                         class="glyphicon glyphicon-chevron-{% if finding.impact %}up{%else%}down{%endif%}"></i></a></span></h4>
             </div>
             <div id="vuln_impact" class="panel-body collapse {% if finding.impact %}in{%endif%}">
-                <pre>{{ finding.impact|markdown_render|default_if_none:"" }}</pre>
+                {{ finding.impact|markdown_render|default_if_none:"" }}
             </div>
         </div>
 
@@ -1020,7 +1020,7 @@
                         class="glyphicon glyphicon-chevron-{% if finding.steps_to_reproduce %}up{%else%}down{%endif%}"></i></a></span></h4>
             </div>
             <div id="vuln_reproduce" class="panel-body collapse {% if finding.steps_to_reproduce %}in{%endif%}">
-                <pre>{{ finding.steps_to_reproduce|markdown_render|default_if_none:"" }}</pre>
+                {{ finding.steps_to_reproduce|markdown_render|default_if_none:"" }}
             </div>
         </div>
 
@@ -1030,7 +1030,7 @@
                         class="glyphicon glyphicon-chevron-{% if finding.severity_justification %}up{%else%}down{%endif%}"></i></a></span></h4>
             </div>
             <div id="sev_just" class="panel-body collapse {% if finding.severity_justification %}in{%endif%}">
-                <pre>{{ finding.severity_justification|markdown_render|default_if_none:"" }}</pre>
+                {{ finding.severity_justification|markdown_render|default_if_none:"" }}
             </div>
         </div>
 
@@ -1040,7 +1040,7 @@
                         class="glyphicon glyphicon-chevron-{% if finding.references %}up{%else%}down{%endif%}"></i></a></span></h4>
             </div>
             <div id="vuln_refs" class="panel-body collapse {% if finding.references %}in{%endif%}">
-                <pre>{{ finding.get_references_with_links|markdown_render|default_if_none:"" }}</pre>
+                {{ finding.get_references_with_links|markdown_render|default_if_none:"" }}
             </div>
         </div>
 


Maffooch:
@37b I'm curious if this could be a browser issue. Markdown appears to be rendering fine on version 2.38.1 of dojo

Chrome:
image

Safari:
image

37b:
> @37b I'm curious if this could be a browser issue. Markdown appears to be rendering fine on version 2.38.1 of dojo
> 
> Chrome: image
> 
> Safari: image

@Maffooch 

Did you notice those are all monospaced fonts? It should render like this, but the `<pre>` tags are colliding with the markdown renderer.

Screenshot 2024-09-10 at 20 12 21


Instead it renders like this...which is ugly and mangles some markdown formatting.

Screenshot 2024-09-10 at 20 14 07

Jon-the-2nd:
It seems this is a CSS issue rather than a markdown to HTML issue. The HTML code that is generated reflects the initial markdown but then the CSS makes it all look the same again.

I found that tweaking the dojo.css file a little bit goes a long way in making the fields on the 'finding' page more readable. Here's what I added at the bottom of dojo.css:

```
.panel-body {
      background-color: #eaf2f5;
  }
  
  .panel-body h1,h2,h3,h4,h5,h6,p,pre {
      font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
  }
  
  .panel-body hr {
      border: 1px solid #bcdce8;
  }
  
  .panel-body blockquote {
      margin: 0;
      border-left: 5px solid #bcdce8;
      padding: 0 10px;
  }
  
  .panel-body blockquote>p {
      font-size: 13px;
  
  }
  
  .panel-body pre {
      line-height: 1;
  }
  
  .panel-body p {
      margin: 0;
  }
  
  .codehilite {
      background-color: #ccd0d3 !important;
      border: 1px solid #8f8686 !important;
      padding: 10px !important;
      margin: 0 !important;
      color: #d96966;
  }
  
  .panel-body code {
      color: #d96966;
      line-height: 1.2;
  }
```
Now this may have an impact on other pages in the application, but I haven't gotten around to that yet.

### Additional issues

I've run into two additional but related problems:

1. The 'Request' and 'Response' fields don't seem to support markdown at all, yet they have the same WYSIWYG editor. This is quite confusing when manually adding a finding as you can spend some time adding markdown to the text input, then when you save it it doesn't render as expected.
2. The behaviour when creating a new finding (Add New Finding from the Findings menu) is slightly different. In the editor the 'Code' button is missing and a 'Preview' button is added instead. Now it seems that both buttons have keyboard shortcuts which can still be used even if the button isn't displayed, but for instance in the 'Edit Finding' page when typing text in a field and hitting Ctrl-P will display the 'preview', however if you go to another field and hit Ctrl-P again it open a 'print' dialog rather than disabling the preview.
3. The 'preview' mode has different styles applied than what is actually rendered when saving the finding.

Behaviour seems to be the same on DefectDojo versions 2.37.0 and 2.38.1

37b:
> It seems this is a CSS issue rather than a markdown to HTML issue. The HTML code that is generated reflects the initial markdown but then the CSS makes it all look the same again.
> 
> I found that tweaking the dojo.css file a little bit goes a long way in making the fields on the 'finding' page more readable. Here's what I added at the bottom of dojo.css:
> 
> ```
> .panel-body {
>       background-color: #eaf2f5;
>   }
>   
>   .panel-body h1,h2,h3,h4,h5,h6,p,pre {
>       font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
>   }
>   
>   .panel-body hr {
>       border: 1px solid #bcdce8;
>   }
>   
>   .panel-body blockquote {
>       margin: 0;
>       border-left: 5px solid #bcdce8;
>       padding: 0 10px;
>   }
>   
>   .panel-body blockquote>p {
>       font-size: 13px;
>   
>   }
>   
>   .panel-body pre {
>       line-height: 1;
>   }
>   
>   .panel-body p {
>       margin: 0;
>   }
>   
>   .codehilite {
>       background-color: #ccd0d3 !important;
>       border: 1px solid #8f8686 !important;
>       padding: 10px !important;
>       margin: 0 !important;
>       color: #d96966;
>   }
>   
>   .panel-body code {
>       color: #d96966;
>       line-height: 1.2;
>   }
> ```
> 
> Now this may have an impact on other pages in the application, but I haven't gotten around to that yet.
> 
> ### Additional issues
> I've run into two additional but related problems:
> 
> 1. The 'Request' and 'Response' fields don't seem to support markdown at all, yet they have the same WYSIWYG editor. This is quite confusing when manually adding a finding as you can spend some time adding markdown to the text input, then when you save it it doesn't render as expected.
> 2. The behaviour when creating a new finding (Add New Finding from the Findings menu) is slightly different. In the editor the 'Code' button is missing and a 'Preview' button is added instead. Now it seems that both buttons have keyboard shortcuts which can still be used even if the button isn't displayed, but for instance in the 'Edit Finding' page when typing text in a field and hitting Ctrl-P will display the 'preview', however if you go to another field and hit Ctrl-P again it open a 'print' dialog rather than disabling the preview.
> 3. The 'preview' mode has different styles applied than what is actually rendered when saving the finding.
> 
> Behaviour seems to be the same on DefectDojo versions 2.37.0 and 2.38.1

All I did was remove the preformatted text tags and all looks fine now. You can try it in dev tools to see it without rebuilding. Even if there are alternate methods to improve the visuals, it doesn't make sense to wrap everything in `<pre>` tags.

Jon-the-2nd:
> All I did was remove the preformatted text tags and all looks fine now. You can try it in dev tools to see it without rebuilding. Even if there are alternate methods to improve the visuals, it doesn't make sense to wrap everything in `\<pre\>` tags.

It does look better indeed without the pre tags. I think it's still a bit confusing that there's a difference between the way it is rendered by the 'preview' button and the way it does when the finding is saved though. 

What about the request/response fields, aren't these supposed to support markdown too? The thing is if you remove the pre tags there, everything just gets wrapped and become unreadable.

37b:
> > All I did was remove the preformatted text tags and all looks fine now. You can try it in dev tools to see it without rebuilding. Even if there are alternate methods to improve the visuals, it doesn't make sense to wrap everything in `\<pre\>` tags.
> 
> It does look better indeed without the pre tags. I think it's still a bit confusing that there's a difference between the way it is rendered by the 'preview' button and the way it does when the finding is saved though.
> 
> What about the request/response fields, aren't these supposed to support markdown too? The thing is if you remove the pre tags there, everything just gets wrapped and become unreadable.

@Jon-the-2nd 

I'm not following. What is the preview button you're referring to?

This pr removes `<pre>` from all the relevant fields in the dom so markdown formatting works as intended everywhere.

If the request/response pairs should be rendered in monospace fonts, then we should use either markdown for that, or just leave the `<pre>` tags on these fields, but that would mean if there _were_ markdown it may fail to render correctly. Given these fields are strictly "code" I think that's a fair trade off.

Are there other fields that should only ever be monospaced? 

@Maffooch thoughts?

Maffooch:
Ahh okay I see. It looks the issue is that there is not a differentiation between 
- `code snippets`
- not code snippets

That is is the biggest difference I can see between the font

The monospace font used here has been around since I got involved in the project 😅 It's just normal dojo to me

37b:
> It seems this is a CSS issue rather than a markdown to HTML issue. The HTML code that is generated reflects the initial markdown but then the CSS makes it all look the same again.
> 
> I found that tweaking the dojo.css file a little bit goes a long way in making the fields on the 'finding' page more readable. Here's what I added at the bottom of dojo.css:
> 
> ```
> .panel-body {
>       background-color: #eaf2f5;
>   }
>   
>   .panel-body h1,h2,h3,h4,h5,h6,p,pre {
>       font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
>   }
>   
>   .panel-body hr {
>       border: 1px solid #bcdce8;
>   }
>   
>   .panel-body blockquote {
>       margin: 0;
>       border-left: 5px solid #bcdce8;
>       padding: 0 10px;
>   }
>   
>   .panel-body blockquote>p {
>       font-size: 13px;
>   
>   }
>   
>   .panel-body pre {
>       line-height: 1;
>   }
>   
>   .panel-body p {
>       margin: 0;
>   }
>   
>   .codehilite {
>       background-color: #ccd0d3 !important;
>       border: 1px solid #8f8686 !important;
>       padding: 10px !important;
>       margin: 0 !important;
>       color: #d96966;
>   }
>   
>   .panel-body code {
>       color: #d96966;
>       line-height: 1.2;
>   }
> ```
> 
> Now this may have an impact on other pages in the application, but I haven't gotten around to that yet.
> 
> ### Additional issues
> I've run into two additional but related problems:
> 
> 1. The 'Request' and 'Response' fields don't seem to support markdown at all, yet they have the same WYSIWYG editor. This is quite confusing when manually adding a finding as you can spend some time adding markdown to the text input, then when you save it it doesn't render as expected.
> 2. The behaviour when creating a new finding (Add New Finding from the Findings menu) is slightly different. In the editor the 'Code' button is missing and a 'Preview' button is added instead. Now it seems that both buttons have keyboard shortcuts which can still be used even if the button isn't displayed, but for instance in the 'Edit Finding' page when typing text in a field and hitting Ctrl-P will display the 'preview', however if you go to another field and hit Ctrl-P again it open a 'print' dialog rather than disabling the preview.
> 3. The 'preview' mode has different styles applied than what is actually rendered when saving the finding.
> 
> Behaviour seems to be the same on DefectDojo versions 2.37.0 and 2.38.1

Sorry, I am rereading this and getting a better understanding. I think adding markdown support in the  request/response is easy enough, but personally I don't think it should be used in these sections as they should be purely for request/response pairs and nothing else. Instead, the form should remove the editor buttons or have some other indicator that these fields do not support formatting.

I can't remember the last time I created a finding manually, but I suspect the reason the preview looks different to the saved finding is because of the `<pre>` tags, which are not considered when the JS fires. This PR should fix that bit at least.

Jon-the-2nd:
@37b, thanks for clarifying. One question though about the request/response pairs: that notion of pairs and the fact that when viewing it says Request #1 and Response #1 makes it appear like it's possible to add multiple such pairs, but I can't seem to find a way to do that. Would you happen to know how that works?

mtesauro:
@Jon-the-2nd 
Generally tool parsers are the ones that create multiple request/response pairs. Burpsuite is an example of this.

Typically, manual testers who are manually adding findings only include a single request/response pair which is why the UI was created with only a single pair available.

I _believe_ but haven't checked that additional pairs can be added via the API.

37b:
> @Jon-the-2nd Generally tool parsers are the ones that create multiple request/response pairs. Burpsuite is an example of this.
> 
> Typically, manual testers who are manually adding findings only include a single request/response pair which is why the UI was created with only a single pair available.
> 
> I _believe_ but haven't checked that additional pairs can be added via the API.

@mtesauro 
Do you have an opinion on improving the typography and/or making the distinction between code samples or pre-formatted text and standard text?

Maffooch:
Something that we have not considered is that the `pre` tags are used for maintaining spacing. The fear is that tool output in DefectDojo could appear radically different following this change. We don't know what assumptions are made by parsers, and checking all 180+ of them is a big lift

PR #10897, opened 2024-09-11
Manage not defined metadata in mitigations and add assumptions comments

arivra:
**Description**

Threat Composer has different behavior for _metadata_ depending on whether it is a threat or a mitigation. In the case of the threat, the 'metadata' key always exists even if no data has been included. However, in mitigation that key may not exist. A fix has been included to validate the 'metadata' key in the mitigation, including an empty list if it does not exist. In addition, comments of the assumptions have been included.

### DIFF
--- a/dojo/tools/threat_composer/parser.py
+++ b/dojo/tools/threat_composer/parser.py
@@ -70,12 +70,12 @@ class ThreatComposerParser:
 
             if "threatAction" in threat:
                 title = threat["threatAction"]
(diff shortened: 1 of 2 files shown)


cneill (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

Maffooch (approved):

PR #10902, opened 2024-09-12
feat(disclaimers): Split disclaimers

kiblik:
Until now, all disclaimers have been the same which is not the best.

Now you can choose between:
- `disclaimer_notifications`
- `disclaimer_reports`

There is one more to inform users that they should not include any personal information in notes:
- `disclaimer_notes`

Plus if there is an internal policy to have a disclaimer in all reports, it is possible to use `disclaimer_reports_forced` to not allow users to be excluded it.

E.g.:
image

image


dojo/db_migrations/0220_system_settings_disclaimer_notif.py
Maffooch, line 1: This could be consolidated into the other migration

dojo/db_migrations/0220_system_settings_disclaimer_notif.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dojo', '0219_system_settings_disclaimer_notif'),
+    ]
+
+    operations = [
+        migrations.RunPython(copy_notif_field),
```
Maffooch, line 20: It would be ideal to have a "reverse" operation for this as well if possible. Even if is just a [noop](https://docs.djangoproject.com/en/5.1/ref/migration-operations/#django.db.migrations.operations.RunPython.noop)

kiblik: Good idea. Thank you. It would also be good to handle these in other (future) cases.

dojo/db_migrations/0220_system_settings_disclaimer_notif.py
kiblik: I'm not sure I understand this comment

Maffooch: migrations 0220 and 0219 could be consolidated into the same plugin

kiblik: When (for any reason) `RunPython` fails, migration is not marked as done. You can try to rerun migrations when you fix the issue (it might be any issue). The problem is if `RunPython` is part of other migration steps (like `RenameField`, `AddField`, ...) that have been performed successfully. You can try to rerun migration but it will fail because you are not able to run the same `RenameField` or `AddField`.
`RunPython` is the most sensitive step failure.
Because of this, we agreed that in this project, we will always put `RunPython` to the separated migration.
Same as:
- https://github.com/DefectDojo/django-DefectDojo/pull/6192
- https://github.com/DefectDojo/django-DefectDojo/pull/6053
- https://github.com/DefectDojo/django-DefectDojo/pull/5953

dojo/db_migrations/0219_system_settings_disclaimer_notif.py
```diff
@@ -0,0 +1,38 @@
...
+
+    dependencies = [
+        ('dojo', '0218_system_settings_enforce_verified_status_and_more'),
+    ]
+
+    operations = [
+        migrations.RenameField(
+            model_name='system_settings',
+            old_name='disclaimer',
+            new_name='disclaimer_notifications',
```
valentijnscholten, line 16: I don't know how common this is, but it might affects users setting the disclaimer via the api. A little note in the upgrade notes? It also affects people who are using customized notification templates.

dojo/db_migrations/0220_system_settings_disclaimer_notif.py
Maffooch: Ahh okay that makes sense! I'll keep this in mind going forward (for review and my own migrations)

Thanks for bringing the receipts 😄

Maffooch (approved):

mtesauro (approved):
Approved

grendel513 (approved):

blakeaowens (approved):

PR #10905, opened 2024-09-13
Wiz Parser: Add SCA parser and fortify old format

Maffooch:
Wiz has a SCA format that does not appear to be supported yet. In order to support it best, I modified the original parser to be a bit safer in the event a new format comes along

[sc-7534]


mtesauro (approved):
Approved

dojo/tools/wiz/parser.py
```diff
@@ -1,91 +1,209 @@
...
+            return "Info"
+        return severity
+
+    def _convert_status(self, row: dict) -> dict:
+        """Convert the "FindingStatus" column to a dict of Finding statuses.
+
+        - Open-> Active = True
+        - Other statuses that may exist...
+        """
+        if (status := row.get("FindingStatus")) is not None:
```
cneill, line 176: Since `{"active": True}` is returned in both cases here, is this just to accommodate possible future conditions beyond `Open`?

dojo/tools/wiz/parser.py
```diff
@@ -1,91 +1,209 @@
...
+            if status == "Open":
+                return {"active": True}
+        # Return the default status of active
+        return {"active": True}
+
+
+class WizParser(
+    WizParserByTitle,
+    WizParserByDetailedName,
+):
```
cneill, line 186: It's not hurting anything as-is, but we probably don't need inheritance here?

```suggestion
class WizParser:
```

cneill (approved):
Some minor questions, but I don't consider either one a blocker

dojo/tools/wiz/parser.py
Maffooch: Correct. I tried to capture that with the bullet in the doctoring: `Other statuses that may exist..`

My report only had `Open` status, but I would imagine there are others

dojo/tools/wiz/parser.py
Maffooch: Yeah, not strictly necessary. Will accept the suggestions. Some of the tests are stuck anyways

grendel513 (approved):

blakeaowens (approved):

PR #10917, opened 2024-09-16
Fix metrics aggregation

dogboat:
**Description**

This patch fixes an issue with metrics aggregation by explicitly specifying the order_by() clause on the aggregate_counts_by_period() method. Not explicitly specifying it caused additional aggregation when the passed-in QuerySet had an ordering.

**Test results**

Tests need to be adjusted to check for this case but I'll add a followup soon.

### DIFF
--- a/dojo/metrics/utils.py
+++ b/dojo/metrics/utils.py
@@ -500,7 +500,7 @@ def aggregate_counts_by_period(
         )
         desired_values += ("closed",)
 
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

dogboat:
[sc-7543]

grendel513 (approved):

cneill (approved):

mtesauro (approved):
Approved

PR #10928, opened 2024-09-17
GHA: Remove Docker Caches

Maffooch:
The docker caches in this repo are filling up pretty often. This is in part due to so many unique PRs coming through and using the caches to store their docker layers. It is so awesome to see such engagement, but the cache issues are causing sporadic failures when the cache gets full. To mitigate these cache fills, a cron has been running to periodically empty the caches. This begs the question: "if we are clearing the cache so often, are we really benefitting from having It"

[sc-7396]


mtesauro (approved):
Approved

cneill (approved):

blakeaowens (approved):

grendel513 (approved):

PR #10931, opened 2024-09-18
Prefetching: Add swagger docs for models already supporting prefetching

Maffooch:
- The /api/v2/global_role endpoint does not support prefetching, but it really should
- I felt sorta icky copy pasting the whole drf-spectacular schema extension, so I added a function to generalize it, then add it to all of the viewsets that already supported prefetch via the `PrefetchDojoModelViewSet` mixin

[sc-7579]

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -297,7 +297,7 @@ class GlobalRoleViewSet(
 
 
 # Authorization: object-based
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

cneill (approved):

blakeaowens (approved):

grendel513 (approved):

PR #10932, opened 2024-09-18
Manage Images: Do not display thumbnail for PDF

Maffooch:
When attaching files to a finding, images will have a thumbnail generated for quick glance viewing. Currently PDFs are part of the whitelist file extensions that allow a thumbnail to be generated, but that does not work, so we get a "cannot find image" icon

image

[sc-7547]


mtesauro (approved):
Approved

cneill (approved):

grendel513 (approved):

PR #10933, opened 2024-09-18
Prefetch Serialization: Add a preference during mapping

Maffooch:
When prefetching via the API, the serializer to user is dynamically fetched. There are cases where there is more than one `ModelSerializer` on defined, and the wrong one is selected. This PR adds a dict to specify the preferred serializer to user a per model basis

The motivation for this PR is when prefetching a finding with files. Before this PR, a 500 error would be generated

[sc-6061]

### DIFF
--- a/dojo/api_v2/prefetch/prefetcher.py
+++ b/dojo/api_v2/prefetch/prefetcher.py
@@ -3,10 +3,16 @@ import sys
 
 from rest_framework.serializers import ModelSerializer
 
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

cneill (approved):

blakeaowens (approved):

grendel513 (approved):

PR #10934, opened 2024-09-18
Risk Exceptions: Add/Remove notes when finding is added/removed from risk exception

Maffooch:
When managing risk exceptions, it would very nice to have record on the finding when the finding is added/removed from a risk exception object, or when a simple risk exception is set via the boolean flag. This can achieved but adding a note on the finding

This PR does not address:
- Setting/Unsetting simple risk exception in the edit finding form
- Setting/Unsetting simple risk exception vie the API
- Setting/Unsetting simple risk exception during bulk edit

[sc-7544]

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -1623,7 +1623,7 @@ def process_resolution_from_jira(finding, resolution_id, resolution_name, assign
                     owner=finding.reporter,
                 )
                 finding.test.engagement.risk_acceptance.add(ra)
-                ra_helper.add_findings_to_risk_acceptance(None, ra, [finding])
+                ra_helper.add_findings_to_risk_acceptance(User.objects.get_or_create(username="JIRA")[0], ra, [finding])
                 status_changed = True
         elif jira_instance and resolution_name in jira_instance.false_positive_resolutions:
             if not finding.false_p:
@@ -1633,7 +1633,7 @@ def process_resolution_from_jira(finding, resolution_id, resolution_name, assign
                 finding.mitigated = None
                 finding.is_mitigated = False
                 finding.false_p = True
-                ra_helper.risk_unaccept(None, finding)
+                ra_helper.risk_unaccept(User.objects.get_or_create(username="JIRA")[0], finding)
                 status_changed = True
         else:
             # Mitigated by default as before
@@ -1645,7 +1645,7 @@ def process_resolution_from_jira(finding, resolution_id, resolution_name, assign
                 finding.mitigated_by, _created = User.objects.get_or_create(username="JIRA")
                 finding.endpoints.clear()
                 finding.false_p = False
-                ra_helper.risk_unaccept(None, finding)
+                ra_helper.risk_unaccept(User.objects.get_or_create(username="JIRA")[0], finding)
                 status_changed = True
     else:
         if not finding.active:
@@ -1655,7 +1655,7 @@ def process_resolution_from_jira(finding, resolution_id, resolution_name, assign
             finding.mitigated = None
             finding.is_mitigated = False
             finding.false_p = False
-            ra_helper.risk_unaccept(None, finding)
+            ra_helper.risk_unaccept(User.objects.get_or_create(username="JIRA")[0], finding)
             status_changed = True
 
     # for findings in a group, there is no jira_issue attached to the finding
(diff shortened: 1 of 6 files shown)


dojo/risk_acceptance/helper.py
```diff
@@ -112,9 +113,20 @@ def remove_finding_from_risk_acceptance(risk_acceptance, finding):
     finding.save(dedupe_option=False)
     # best effort jira integration, no status changes
     post_jira_comments(risk_acceptance, [finding], unaccepted_message_creator)
+    # Add a note to reflect that the finding was removed from the risk acceptance
+    if user is not None:
+        finding.notes.add(Notes.objects.create(
+            entry=(
+                f"{Dojo_User.generate_full_name(user)} removed this finding from the risk acceptance: "
```
mtesauro, line 120: Can we also include the user ID in parenthesis after the full user name?

Users can change their names so this isn't always a definitive value - user ID won't change so is the safest way to identify who did the action though not user friendly. Having both is optimal.

This way I can't change my DefectDojo user data to "Cody Maffucci" and get you in trouble :wink:

mtesauro (changes requested):

dojo/risk_acceptance/helper.py
Maffooch: Sure, can do. Please note that the if the user object changes in any way, the note written on the finding will not change. The note itself can be edited, but the modification will be auditlogged on the note object

mtesauro (approved):
Approved

dojo/jira_link/helper.py
```diff
@@ -1623,7 +1623,7 @@ def process_resolution_from_jira(finding, resolution_id, resolution_name, assign
                     owner=finding.reporter,
                 )
                 finding.test.engagement.risk_acceptance.add(ra)
-                ra_helper.add_findings_to_risk_acceptance(ra, [finding])
+                ra_helper.add_findings_to_risk_acceptance(None, ra, [finding])
```
cneill, line 1626: Should this maybe include the `JIRA` user rather than `None`, as with the `mitigated_by` field in the `else` block [below](https://github.com/DefectDojo/django-DefectDojo/pull/10934/files#diff-ce808f9fd6f829add9bca414ba21800d5b7b333803aaeaff4ff600085c1409f1R1645)? The helper methods don't post a note if `None` is supplied as the user.

dojo/jira_link/helper.py
```diff
@@ -1633,7 +1633,7 @@ def process_resolution_from_jira(finding, resolution_id, resolution_name, assign
                 finding.mitigated = None
                 finding.is_mitigated = False
                 finding.false_p = True
-                ra_helper.risk_unaccept(finding)
+                ra_helper.risk_unaccept(None, finding)
```
cneill, line 1636: Same question as above

dojo/jira_link/helper.py
```diff
@@ -1645,7 +1645,7 @@ def process_resolution_from_jira(finding, resolution_id, resolution_name, assign
                 finding.mitigated_by, _created = User.objects.get_or_create(username="JIRA")
                 finding.endpoints.clear()
                 finding.false_p = False
-                ra_helper.risk_unaccept(finding)
+                ra_helper.risk_unaccept(None, finding)
```
cneill, line 1648: Same question as above

dojo/jira_link/helper.py
```diff
@@ -1655,7 +1655,7 @@ def process_resolution_from_jira(finding, resolution_id, resolution_name, assign
             finding.mitigated = None
             finding.is_mitigated = False
             finding.false_p = False
-            ra_helper.risk_unaccept(finding)
+            ra_helper.risk_unaccept(None, finding)
```
cneill, line 1658: Same as above

cneill (commented):
Just wondering if we should create a note when JIRA updates produce changes to risk acceptances

dojo/jira_link/helper.py
Maffooch: A note requires a DefectDojo User object that exists in the database in order to be created. We cannot always guarantee there is user that matches the one from jira. 

I suppose we could copy the functionality used in the webhook of creating a dedicated jira service account: https://github.com/DefectDojo/django-DefectDojo/blob/e4fbfe0f2c48c10dcfcd6422b907050a91b22fca/dojo/jira_link/views.py#L254

cneill (approved):

blakeaowens (approved):

grendel513 (approved):

PR #10938, opened 2024-09-19
User: Make email required at all times, password required for new users

Maffooch:
Creating local users is not the best experience. To accommodate a flow that supports sending temporary passwords to user, and forcing them to login again, we need to make the password field and email field required attributes for new users

[sc-7616]

### DIFF
--- a/unittests/test_rest_framework.py
+++ b/unittests/test_rest_framework.py
@@ -1709,7 +1709,9 @@ class UsersTest(BaseClass.BaseClassTest):
         self.assertEqual(self.endpoint_model.objects.count(), length + 1)
 
     def test_create_user_with_non_configuration_permissions(self):
-        payload = self.payload.copy()
+        payload = self.payload.copy() | {
+            "password": "testTEST1234!@#$",
+        }
         payload["configuration_permissions"] = [25, 26]  # these permissions exist but user can not assign them becaause they are not "configuration_permissions"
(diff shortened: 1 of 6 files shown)


mtesauro (approved):
Approved

unittests/test_rest_framework.py
```diff
@@ -1699,6 +1699,15 @@ def __init__(self, *args, **kwargs):
...
+    def test_create(self):
+        payload = self.payload.copy() | {
+            "password": "testTEST1234!@#$",
+        }
+        length = self.endpoint_model.objects.count()
+        response = self.client.post(self.url, payload)
+        self.assertEqual(201, response.status_code, response.content[:1000])
+        self.assertEqual(self.endpoint_model.objects.count(), length + 1)
+
     def test_create_user_with_non_configuration_permissions(self):
```
cneill, line 1711: I guess this will fail with both `object does not exist` and a message about the user missing a password, but seems like we should supply the password just to make it clearer what's *supposed* to fail here.

cneill (commented):
Just a small comment about one of the tests

unittests/test_rest_framework.py
Maffooch: sure thing

cneill (approved):

blakeaowens (approved):

grendel513 (approved):

PR #10945, opened 2024-09-21
Parsers: Specify lists rather than `dict.values()`

Maffooch:
A few parsers do not pass a list of findings, but rather an another type of iterable. This PR will correct these parsers to make the return type more explicit


mtesauro (approved):
Approved

grendel513 (approved):

cneill (approved):

blakeaowens (approved):

PR #10984, opened 2024-10-01
Update AWS SecurityHub ASFF Parser to include Impact

testaccount90009:

**Description**
I am submitting this PR to add 'Impact' to the Findings which contains the joined AWS ARNs.  Previously I edited this parser to add them to the Description and while that does accomplish what I set out to do, after using it in the wild for some time now - it is clear that attributing the resource_arns to their own field would be best.  As it sits now, the \n in the full_description breaks the Jira table formatting when pushing a Finding to Jira.

This PR aims to solve the formatting issue, by using a current field that is empty (Impact) to contain the joined resource AWS ARNs.

I am open to removing the description that sets the full_description work and the join on the AWS ARNs, if that is necessary.

**Test results**

This PR contains a commit that updates the respective unit test.

**Extra information**

- enhancement
- performance
- feature
- bugfix
- maintenance (a.k.a chores)
- dependencies


mtesauro (approved):
Approved

dojo/tools/asff/parser.py
```diff
@@ -81,6 +81,7 @@ def get_findings(self, file, test):
                 severity=self.get_severity(item.get("Severity")),
                 active=active,
                 unique_id_from_tool=item.get("Id"),
+                impact=resource_arn_strings,
```
Maffooch, line 84: Is there a possibility that this variable is not defined?

testaccount90009: Which one?  I have tested in both my dev stack and prod stacks with importing one of our ASFF json files with these proposed changes and the Impact field is populated appropriately.  The resource_arn_strings is defined and joined in the description as the full_description.  

This change of adding impact=resource_arn_strings just associates the joined list of AWS ARNs into the impact field.

testaccount90009: I can remove the resource_arn_strings being added to the full_description to make that look better, if needed.  With the impact being set as the resource_arn_strings, the same data is already duplicated in the full_description.  I could always revert full_description back to the raw description, since with this changes the 'impact' field will contain those values instead.

Maffooch: I was looking at this statement:
```
if resource_arns:
    resource_arn_strings = ", ".join(resource_arns)
    full_description = f"**AWS resource ARN:** {resource_arn_strings}\n\n{control_description}"
else:
    full_description = control_description
```

If the `else` block is triggered, then the `resource_arn_strings` variable does not appear to be defined. When trying to to assign it to the impact of the finding, an error may occur

testaccount90009: @Maffooch Ah great catch, sorry I am a novice at programming - you raise a good point.  I'll make another commit to add to the else statement in the event that there are no resource_arns.

Although technically if a vulnerability is coming out of AWS Security Hub and the ASFF format, it should **always** contain a resource_arn.  I don't think a vulnerability could exist in theory without an affected resource, but maybe it's possible - so I'll add a case for that.

Much appreciated for your help.

testaccount90009: @Maffooch I believe the recent push with commits should solve what you bring up.

blakeaowens (approved):

Maffooch (approved):

hblankenship (approved):

PR #10995, opened 2024-10-03
Fix tag creation bug in FindingTemplateSerializer (letters instead of full tags)

StephanPillhofer:
Fixes a bug that caused tags in finding templates created via the API to be split into single-letter tags instead of preserving the intended full tags.

Example of correctly and wrongly created tags for the tag "string":

grafik


mtesauro (approved):
Approved

Maffooch (approved):

grendel513 (approved):

cneill (approved):

PR #11001, opened 2024-10-04
Mend SCA imports contain locations which are similar to filePaths for SAST scans

testaccount90009:
This code will use the 'locations' for SCA scan outputs to do the same thing that's done for SAST 'filePaths'.  Since a Finding report will either be from SAST or SCA, it is unlikely that a collision will happen, since those findings are inherently different from Mend.  Since the filepaths is already being joined for the SAST implementation, if it is indeed SCA results instead, the same thing will happen except now with the appropriate locations of the library and vulnerability.

Note: this is not from Mend Platform or the CLI Agent output, but rather the Mend SCA portal.  There is a new Platform API that combines both SAST and SCA vulnerabilities, so a new parser at some point for that would be good, and then it's possible to rename this to 'Legacy' for the Mend parser, since the 'Platform' should be the new.

**Test results**

I've added a scrubbed sample file and unit test.

**Documentation**

I can update the respective documentation with regards to these changes.


testaccount90009:
The force push was to rebase my branch against upstream/dev, vs upstream/master, as I accidentally made my branch originally off of the master branch and not dev.  Apologies.

requirements.txt
```diff
@@ -21,7 +21,7 @@ django-slack==5.19.0
 git+https://github.com/DefectDojo/django-tagging@develop#egg=django-tagging
 django-watson==1.6.3
 django-prometheus==2.3.1
-Django==5.0.8
+Django==5.1.1
```
mtesauro, line 24: There shouldn't be  an update to Django in this parser change - please remove.

testaccount90009: Thank you for the catch - I am assuming this was something from a different commit when I did a rebase against dev.  It should only be the 3 files - nothing in this one.  I've rebased it against the latest dev branch.

mtesauro: No worries

mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

grendel513 (approved):

hblankenship (approved):

PR #11015, opened 2024-10-07
docs(helm): add documentation about installation and external databas…

leofvo:
Hello, 

After installing the helm chart and set up an external database connection, I feel like it's a bit tricky, and we could make it more straight forward with a little documentation. 

So I wrote the documentation from my own experience installing the helm chart and set up an external database.

Feel free to give feedback.

PR #11016, opened 2024-10-07
fix(helm): add missing env config on job

leofvo:
The job isn't working well when using an external database because the init container checking if the database is accessible isn't taking the same env values as the container that is initializing the database config

I propose to fix this little issue by simply uniformizing the env provided on the initContainer and the container.

The fix is minor, if you need more information, feel free to ask.

### DIFF
--- a/helm/defectdojo/templates/initializer-job.yaml
+++ b/helm/defectdojo/templates/initializer-job.yaml
@@ -67,8 +67,8 @@ spec:
             name: {{ $fullName }}
             optional: true
         env:
-        {{- if .Values.extraEnv }}
-        {{- toYaml .Values.extraEnv | nindent 8 }}
+        {{- with .Values.extraEnv }}
+        {{- toYaml . | nindent 8 }}
         {{- end }}
       containers:
       {{- if .Values.cloudsql.enabled  }}
@@ -122,8 +122,8 @@ spec:
               name: {{ .Values.postgresqlha.postgresql.existingSecret }}
               key: postgresql-postgres-password
               {{- end }}
-        {{- if .Values.extraEnv }}
-        {{- toYaml .Values.extraEnv | nindent 8 }}
+        {{- with .Values.extraEnv }}
+        {{- toYaml . | nindent 8 }}
         {{- end }}
         resources:
           {{- toYaml .Values.initializer.resources | nindent 10 }}
--- a/helm/defectdojo/templates/initializer-job.yaml
+++ b/helm/defectdojo/templates/initializer-job.yaml
@@ -66,6 +66,20 @@ spec:
         - secretRef:
             name: {{ $fullName }}
             optional: true
+        env:
+        - name: DD_DATABASE_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              {{- if eq .Values.database "postgresql" }}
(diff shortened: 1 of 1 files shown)


kiblik:
Same as https://github.com/DefectDojo/django-DefectDojo/pull/11015#issuecomment-2399681231

kiblik:
I'm not sure, I understand the reason here. `/wait-for-it.sh` (the only command in `wait-for-db` container) is testing only TCP connection. It is not trying to log in to the DB. Why is `DD_DATABASE_PASSWORD` needed here?

mtesauro:
@LeoFVO kiblik is correct here - we only take PRs against the dev or bugfix branches. This seems like a dev branch type of PR IMHO.

LeoFVO:
> I'm not sure, I understand the reason here. `/wait-for-it.sh` (the only command in `wait-for-db` container) is testing only TCP connection. It is not trying to log in to the DB. Why is `DD_DATABASE_PASSWORD` needed here?

Alright, we could only add `extraEnv` here, password isn't useful :)

LeoFVO:
I removed the unused database password and pointed to dev branch.
Feel free to squash if required.

LeoFVO:
I just noticed that the job doesn't have helm hook (https://helm.sh/docs/topics/charts_hooks/).
I think we should consider adding:
```yml
    "helm.sh/hook": pre-install, pre-upgrade
    "helm.sh/hook-delete-policy": hook-succeeded
```

This will make the job run before installing or upgrading deployment for example.
The actual chart make the deployment loop waiting migration to run, and in case of use of argocd for example, the job is constantly launched (looping infinitly) due to the TTL attached. 
The job is deleted after 60 sec, but restarted by argocd because missing in application. Adding Helm hooks will solve this issues in bonus.

kiblik:
I agree that right now, the setup might not be the best for Argo. I would avoid `pre-install, pre-upgrade`. It would create circular dependency and deployment might not be successful because the database needs to be started before the initializer.

I would maybe go with
- define that if `.Values.initializer.keepSeconds` is 0 or undefined, `ttlSecondsAfterFinished` needs to be removed
- drop date from initializer job - to avoid starting different jobs even if it is not needed
- guideline for Argo users (info on how to set values to be able to run start DD stack) - e.g. set `.Values.initializer.keepSeconds` to zero

leofvo:
Agree, this has to be done in this PR or another one ?

kiblik:
> Agree, this has to be done in this PR or another one ?

The change is not that large that it would "have to" be in separated PR. But smaller PRs are usually reviewed and accepted quicker.

leofvo:
> > Agree, this has to be done in this PR or another one ?
> 
> The change is not that large that it would "have to" be in separated PR. But smaller PRs are usually reviewed and accepted quicker.

Okay, i'll check to make theses changes in another PR than. Can this one be validated ?

kiblik:
Ohh I see, in the context of #11015, it makes sense.

kiblik (dismissed):

leofvo:
Done

mtesauro (approved):
Approved

hblankenship (approved):

Maffooch (approved):

Issue #7532, opened 2023-02-01
set the celery worker group to 1001


nobletrout:
**Bug description**
The group ID for celery worker and beat is 1337 in the image.
This collides with istio's magic user IDs that rewrite iptables rules specific to that user ID.

**Steps to reproduce**
1. deploy istio
2. run celery workers
3. they don't connect to a db also in k8s

**Expected behavior**
It should work

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
- need istio deployed in mTLS strict mode
- need postgres server deployed in another namespace

The easiest solution here is to specify a different GID/UID that doesn't collide with this system. A securityContext can also be used but not might be the easiest to deploy in a controlled environment.

manuel-sommer:
could you make a PR for that @nobletrout ?

AdamJoelNichols:
Hi @manuel-sommer - hope it's ok, I created a PR for this change - please let me know if it's in line with what you expected - thanks!
https://github.com/DefectDojo/django-DefectDojo/pull/11052

mjnagel:
Definitely +1 to this as it causes unexpected behavior when deploying with Istio that is hard to track down. We were able to work around this in the helm chart by changing the security context to run as a different group, but I think it would make sense to switch to 1001 which is _relatively_ widely adopted for images, and already what the user id is set to. I'm not familiar with the base image(s) being used and if that would clash with them though.

valentijnscholten:
I think we have to follow the conlusion from the maintainers/owners in #11052 and close this here.

PR #11052, opened 2024-10-11
Update Dockerfile.django* to use gid 4242

AdamJoelNichols:
The gid set for the image is currently 1337.
This collides with istio, a popular k8s traffic
management, telemetry, and security tool, which
also uses gid 1337 to rewrite iptables so that
it can do its magic.

This results in DefectDojo's initializer being
unable to connect to postgresql.

Original bug issue:
[github.com/DefectDojo/django-DefectDojo/issues/7532
](https://github.com/DefectDojo/django-DefectDojo/issues/7532)
**Test results**

No test updates required, I don't believe.  I did not see any that validated or made use of the current gid.

Manual testing:
Brought up DefectDojo within a cluster that makes use of istio.
Fails to connect to postgresql before gid change.
Succeeds in its connection after gid update.

### DIFF
--- a/Dockerfile.django-alpine
+++ b/Dockerfile.django-alpine
@@ -34,7 +34,7 @@ RUN CPUCOUNT=1 pip3 wheel --wheel-dir=/tmp/wheels -r ./requirements.txt
 FROM base AS django-alpine
 WORKDIR /app
 ARG uid=1001
-ARG gid=1337
+ARG gid=4242
 ARG appuser=defectdojo
 ENV appuser=${appuser}
 RUN \
--- a/Dockerfile.django-debian
+++ b/Dockerfile.django-debian
@@ -33,7 +33,7 @@ RUN CPUCOUNT=1 pip3 wheel --wheel-dir=/tmp/wheels -r ./requirements.txt
 FROM base AS django
 WORKDIR /app
 ARG uid=1001
-ARG gid=1337
+ARG gid=4242
 ARG appuser=defectdojo
 ENV appuser=${appuser}
 RUN \


jbraswell:
This looks like a relatively benign change that will make the images compatible with Istio. Any chance this could be merged soon?

kiblik:
Just 2 questions:
1. Why exactly 4242? Is there some best practice with using UIDs/GIDs? How we can be sure it will not overlap with another tool in the future?
2. Are we sure there will be no impact on existing deployments? E.g. uploaded files stored in media dir has been created with original GID - as UID has higher priority in permission evaluation this might not be the issue but is it possible to check deeper same/similar potential issues?

mtesauro:
I agree with @kiblik  on his points.

While changing the GID or UID may work for you, it's impossible to know if others using DefectDojo have deployments that rely on the currently set UID & GID so I'm not eager to change them without a pretty compelling reason and with some sort of heads up/warning to the current user base so we don't provide bad "surprises" to them the next time they upgrade DefectDojo.

You do have the dockerfiles and everything needed to adjust the GID for your specific needs which is one of the great parts of Open Source. Yes, you'd end up running a slightly custom install of DefectDojo but building and storing your own container images is not a bad thing nor an excessive burden.

AdamJoelNichols:
@kiblik 
1. 4242 is somewhat arbitrary - though I suppose readers of Hitchhiker's Guide may disagree.  The main reason for the change is that the current GID, 1337 ("leet") is somewhat less arbitrary and therefore more prone to collision.  I'm open to whatever number you all think is best, this PR is mainly a suggestion to not use 1337.  
2. I cannot be sure.  @mtesauro is correct.  We are currently making use of DD as a custom install - and happy to have the option.  In some major upgrade though, I would ask you to consider moving away from 1337, again, as this number is more prone to collision due to its imbued meaning.

Maffooch:
@AdamJoelNichols thank you for bringing this to our attention! In the [Q4 community update](https://github.com/DefectDojo/django-DefectDojo/discussions/11199), we are prioritizing the structure and simplicity of the docker files. We will keep your suggestions in mind when doing that work 😄

Issue #10207, opened 2024-05-14
Exception on invalid environment


kiblik:
**Be informative**
DD raises an exception (HTTP-500 - internal server error) if a user uses a name of an non-existing environment in (re)imports

**Bug description**

`ImportScanSerializer.set_context` and `ReImportScanSerializer.set_context` uses
```python
        context["environment"] = Development_Environment.objects.get(
            name=data.get("environment", "Development")
        )
```
which is able to handle not defined `environment` but does not handle non-existenting one

**Steps to reproduce**
```bash
curl -X 'POST' \
  'http://xxx:8080/api/v2/reimport-scan/' \
...
  -F 'environment=non_existing_enviro' \
...
```

**Expected behavior**
There are 2 options
- HTTP-400 validation error
- Create environment - but `set_context` is happening outside of AutoCreate context

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [x] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version (see footer) or commit message: [2024-05-14 12:04:46 -0500] 1b86b6684: Optionally Enforce SLA Remediation Days (#10179) [ (HEAD, upstream/dev)]

**Logs** 
```python

uwsgi-1         | [14/May/2024 19:26:39] ERROR [dojo.api_v2.exception_handler:43] Development_Environment matching query does not exist.
uwsgi-1         | Traceback (most recent call last):
uwsgi-1         |   File "/usr/local/lib/python3.11/site-packages/rest_framework/views.py", line 506, in dispatch
uwsgi-1         |     response = handler(request, *args, **kwargs)
uwsgi-1         |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
uwsgi-1         |   File "/usr/local/lib/python3.11/site-packages/rest_framework/mixins.py", line 19, in create
uwsgi-1         |     self.perform_create(serializer)
uwsgi-1         |   File "/app/dojo/api_v2/views.py", line 2792, in perform_create
uwsgi-1         |     serializer.save(push_to_jira=push_to_jira)
uwsgi-1         |   File "/app/dojo/api_v2/serializers.py", line 2538, in save
uwsgi-1         |     context = self.set_context(data)
uwsgi-1         |               ^^^^^^^^^^^^^^^^^^^^^^
uwsgi-1         |   File "/app/dojo/api_v2/serializers.py", line 2415, in set_context
[...]

PR #11053, opened 2024-10-11
Fix for issue #10207  non-existent env import

hblankenship:
When a user passes a currently non-existent Development_Environment into import or re-import, and auto_create_context is true, it will now be created if it does not already exist instead of failing with a 500 server error.

Comments also added to indicate the expectation that the call to get_or_create succeeds.

[sc-7806]

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -2175,9 +2175,12 @@ class ImportScanSerializer(serializers.Serializer):
         context = dict(data)
         # update some vars
         context["scan"] = data.pop("file", None)
-        context["environment"] = Development_Environment.objects.get(
+        # The assumption below is that a Development environment will be gotten or created, whether of the one provided or the Development environment
+        # because the Development environment always exists...
+        # If this fails, there are bigger problems.
+        context["environment"] = Development_Environment.objects.get_or_create(
             name=data.get("environment", "Development"),
-        )
+        )[0]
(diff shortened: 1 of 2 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

dojo/api_v2/serializers.py
```diff
@@ -2454,9 +2462,18 @@ def set_context(
         context = dict(data)
         # update some vars
         context["scan"] = data.get("file", None)
-        context["environment"] = Development_Environment.objects.get(
-            name=data.get("environment", "Development"),
-        )
+
```
grendel513, line 2465: Now that there is more going on here, should we make this a function so that we dont have copy past code? For maintainability it would be good to have only one place where this identical code resides.

rossops (approved):

dojo/api_v2/serializers.py
hblankenship: This was a fairly large lift with regard to changes in that file so I will request re-reviews.

mtesauro (approved):
Approved

Maffooch (approved):
I pulled this down and tested locally. Great job!

blakeaowens (approved):

cneill (approved):

grendel513 (approved):

PR #11054, opened 2024-10-12
feat: separate nginx and django image tags

corang:
Having the tags combined keeps people from being able to use different versions of images. In my case I'm trying to use hardened containers from a government source that doesn't tag the nginx image in line with the django image so I can't deploy the chart with the hardened images.


kiblik:
This tag is updated during the preparation of the helm chart.
https://github.com/DefectDojo/django-DefectDojo/blob/340cbb2db00c189f6308bb769825c2d78b8b71ef/.github/workflows/release-x-manual-helm-chart.yml#L67
This space will need to be adjusted accordingly as well.

mtesauro:
@corang We don't accept PRs against master. Please make this PR against the dev branch.

corang:
@mtesauro Fixed!
@kiblik  Thanks for the heads up, think I changed it to what's appropriate

kiblik:
Thank you @corang,
"Unfortunately", I just noticed some other things.

1. [In the past](https://github.com/DefectDojo/django-DefectDojo/pull/10612), we have been consolidating naming conversation in the HELM chart based on [official recommendations](https://helm.sh/docs/chart_best_practices/values/#naming-conventions). I would like to ask you to follow it (skip `-`, use camelCase)
1. `.tag` is used on much more places ([celery-worker](https://github.com/DefectDojo/django-DefectDojo/blob/master/helm/defectdojo/templates/celery-worker-deployment.yaml), [celery-beat](https://github.com/DefectDojo/django-DefectDojo/blob/master/helm/defectdojo/templates/celery-beat-deployment.yaml) and maybe more?). Please, can you take a real deep dive into it?
1. Similar to 1., this is change which redefines variables which somebody might use somewhere (redefine based on own needs). So this change needs to be announced in the Release notes.
1. (Maybe it is bigger than this PR) If we are changing the way how tags are used in the HELM, it might be a good idea to start using a more common convention for it. I would go with adding `bitnami/common` as a dependency and use [the "ImageRoot" syntax](https://github.com/bitnami/charts/tree/main/bitnami/common/#imageroot) for it (similar to [this](https://github.com/bitnami/charts/blob/1478868d7784fc7a82fefc0302aa2ebed2dfa624/bitnami/postgresql/templates/primary/statefulset.yaml#L187) and [this](https://github.com/bitnami/charts/blob/1478868d7784fc7a82fefc0302aa2ebed2dfa624/bitnami/postgresql/templates/_helpers.tpl#L48)). If would be helpful also for tools like Renovate.

Maffooch:

kiblik:
JFYI, this issue is addressed in #13370. From now on, different images/tags can be used

PR #11055, opened 2024-10-12
Add scm type to import-scan api method with auto_create_context

dmarushkin:
Hi guys! 

I use private gitlab for my codebase, then I try to load any report into defectdojo via api import-scan with auto_create_context option it does not set scm-type for new project and file links for finding in UI do not render well.

So due to [docs](https://defectdojo.github.io/django-DefectDojo/integrations/source-code-repositories/) I should capcure response for loaded scan, parse created project_id and send another request to api/v2/metadata/ with payload {"project": <captured_project_id>, "name": "scm-type", "value":"gitlab"}. That's not convenient at all )

Please consider this change with additional field product_scm_type in api/v2/import-scan api method. If you set auto_create_context, project and product_scm_type, it will create meta field "scm-type" for the project with project itself and UI file links will be rendered fine acording with used scm.


Maffooch:
Hi @dmarushkin thank you for your contribution! Though I like the approach you took, keeping the auto create context relevant to the necessities that the test and findings need is crucial in making this feature maintainable and easy to understand for newcomers.

dmarushkin:
Actualy I looked for way to set default scm-type for all new projects in system settings and the way to set it for each particular project in meta-vars looks like counter-intuitive...  

What do you think about adding scm-type as another row in system settings as drop down list of supported options with default value? In finding render first choise is vaule from metavars, second is the one from settings.

Maffooch:
I think it may be better to stick with the work around that you proposed earlier for this particular use case

>So due to [docs](https://defectdojo.github.io/django-DefectDojo/integrations/source-code-repositories/) I should capcure response for loaded scan, parse created project_id and send another request to api/v2/metadata/ with payload {"project": <captured_project_id>, "name": "scm-type", "value":"gitlab"}.

dmarushkin:
Ok, not a problem, thanks for your time here )

PR #11086, opened 2024-10-16
SLA Config: Add new config that does not enforce SLA

Maffooch:
It would be nice ship dojo with an SLA config that easily omits SLA configs on a per product basis

[sc-7952]

### DIFF
--- a/dojo/fixtures/sla_configurations.json
+++ b/dojo/fixtures/sla_configurations.json
@@ -18,7 +18,6 @@
     },
     {
         "model": "dojo.sla_configuration",
-        "pk": 100,
         "fields": {
             "name": "No SLA Enforced",
             "description": "No SLA is enforced for a product which uses this SLA configuration.",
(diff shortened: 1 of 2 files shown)


mtesauro (approved):
Approved

cneill (approved):

dojo/fixtures/sla_configurations.json
```diff
@@ -0,0 +1,36 @@
...
+            "medium": 90,
+            "enforce_medium": true,
+            "low": 120,
+            "enforce_low": true,
+            "async_updating": false
+        }
+    },
+    {
+        "model": "dojo.sla_configuration",
+        "pk": 100,
```
blakeaowens, line 21: I think you can leave this out and Django will automatically assign the next available ID when the fixture is loaded.

Hopefully `100` is safe if we decide to keep it... ;)

blakeaowens (changes requested):

grendel513 (approved):

blakeaowens (approved):

PR #11087, opened 2024-10-16
Fixing bug in Excel Exporter

venegator:
**Description**

This PR is aimed to fix an bug when exporting excel reports for findings that has the finding_group field set, causing the Excel file to have displaced columns to the left starting from the finding_group one. This is caused because when trying to add that key from the Finding Ojbject, it is raising an Exception, and it should display the string `"Value not supported"`, line 1002.

The problem is that when this situation happens, it is not increasing the column number value and when the next iteration is processed, the next key is overwriting the values of the column mentioned above, causing this way the displacement of the rest of the columns.

**Checklist**

### DIFF
--- a/dojo/reports/views.py
+++ b/dojo/reports/views.py
@@ -948,6 +948,7 @@ class ExcelExportView(View):
                     except Exception as exc:
                         logger.error("Error in attribute: " + str(exc))
                         cell = worksheet.cell(row=row_num, column=col_num, value=key)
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

cneill (approved):

PR #11151, opened 2024-10-29
:tada: add tenable plugin to reference #11127

### LINKED ISSUE #11127: Add Tenable Plugin Info to References
Tenable Vulnerability Scanner use so called Plugins to check a given host for the existence of Vulnerabilities.
Plugins can be thought of as a Script telling the scanner "Connect to port X, check for the Version string Y and if found do Z" These Plugins do have a unique Id and the plugin content can change over time. In every scan export, the Plugin Id, Published Date and Modified Date are contained. 

After importing the Scan information, the Information which Plugin Id and the date values of the Plugin are discarded so there is no possibility to check from Defect Dojo which Plugin Version was used to find the Vulnerability. This information could be valuable when investigating to find out exactly why and how the scanner detected a given vulnerability

The Scan Plugin Information should be added to the References section of a given Vulnerability when a Tenable scan is imported. Format wise this could be appended as a new line like:
<Tenable Plugin Information: ID: 12345, Published: 01/01/24, Modified: 01/01/24>

manuel-sommer:
closes #11127

### DIFF
--- a/dojo/tools/tenable/csv_format.py
+++ b/dojo/tools/tenable/csv_format.py
@@ -103,6 +103,9 @@ class TenableCSVParser:
             mitigation = str(row.get("Solution", row.get("definition.solution", row.get("Steps to Remediate", "N/A"))))
             impact = row.get("Description", row.get("definition.description", "N/A"))
             references = row.get("See Also", row.get("definition.see_also", "N/A"))
+            references += "\nTenable Plugin ID: " + row.get("Plugin", "N/A")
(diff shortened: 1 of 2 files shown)


Marterich (approved):

mtesauro (approved):
Approved

cneill (approved):

Maffooch (approved):

grendel513 (approved):

PR #11108, opened 2024-10-21
feat: add the extra labels capabilities for chart

JGodin-C2C:
**Description**

Add extra labels for the  deployment

**Test results**

This is not directly related to the defectdojo development but a deployment-wise improvement

**Documentation**

N/A

**Checklist**

This checklist is for your information.

- [x] Make sure to rebase your PR against the very latest `dev`.
- [x] Features/Changes should be submitted against the `dev`.
- [x] Give a meaningful name to your PR, as it may end up being used in the release notes.
- [x] If this is a new feature and not a bug fix, you've included the proper documentation in the docs at https://github.com/DefectDojo/django-DefectDojo/tree/dev/docs as part of this PR.

PR #11168, opened 2024-10-31
Harmonize helm

JGodin-C2C:
**Description**

Harmonize the different labels in the helm chart. 

See https://github.com/DefectDojo/django-DefectDojo/pull/11108


helm/defectdojo/templates/django-ingress.yaml
```diff
@@ -14,10 +14,13 @@ metadata:
...
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     helm.sh/chart: {{ include "defectdojo.chart" . }}
+    {{- with .Values.extraLabels }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
 {{- if or .Values.django.ingress.annotations .Values.gke.useGKEIngress }}
   annotations:
 {{- with .Values.django.ingress.annotations }}
-{{ toYaml . | indent 4 }}
+  {{ toYaml . | indent 4 }}
```
cneill, line 23:
```suggestion
  {{- toYaml . | nindent 4 }}
```

helm/defectdojo/templates/initializer-job.yaml
```diff
@@ -22,8 +25,11 @@ spec:
         defectdojo.org/component: initializer
         app.kubernetes.io/name: {{ include "defectdojo.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
-        {{- if .Values.initializer.labels }}
-        {{- toYaml .Values.initializer.labels | nindent 8 }}
+        {{- with .Values.extraLabels }}
+          {{- toYaml .Values.extraLabels | nindent 8 }}
```
cneill, line 29:
```suggestion
          {{- toYaml . | nindent 8 }}
```

helm/defectdojo/templates/network-policy.yaml
```diff
@@ -9,26 +9,29 @@ metadata:
...
     app.kubernetes.io/name: {{ include "defectdojo.name" . }}
+    {{- with .Values.extraLabels }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
 spec:
   podSelector:
     matchLabels:
       app.kubernetes.io/instance: {{ .Release.Name }}
-  {{- if .Values.networkPolicy.ingress}}
+  {{- with .Values.networkPolicy.ingress }}
```
cneill, line 19: I don't think `with` will work in this case since there is an `else` on line 22 that depends on this `if`

```suggestion
  {{- if .Values.networkPolicy.ingress }}
```

helm/defectdojo/templates/network-policy.yaml
```diff
@@ -9,26 +9,29 @@ metadata:
...
   ingress:
-  {{- toYaml .Values.networkPolicy.ingress | nindent 4 }}
+  {{- toYaml . | nindent 4 }}
```
cneill, line 21:
```suggestion
  {{- toYaml .Values.networkPolicy.ingress | nindent 4 }}
```

cneill (changes requested):
Thanks for tidying these other blocks up. Just a few tweaks required here

helm/defectdojo/templates/configmap.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     helm.sh/chart: {{ include "defectdojo.chart" . }}
+    {{- with .Values.extraLabels }}
+      {{- toYaml .Values.extraLabels | trim | nindent 4 }}
```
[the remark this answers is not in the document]
JGodin-C2C: I also remove the trim as it is also absent in other files.

cneill (approved):
Looks good! Thanks for your work on this

cneill:
We'll need to get some other reviewers to approve it, but I think it's good to go. Any other thoughts before we merge @dsever ?

mtesauro (approved):
Approved

kiblik (approved):
As an improvement, I do not see any issue here. I tested this change locally on a small setup as well.

1. I would probably use "common helper" for the future
    - but it might be done in the next PR - now there is no reason to block this PR; it is a "nice to have"-feature 
2. And I would add these labels to `gke-managed-certificate.yaml` as well
    - but this would need to be tested by people using this functionality. I do not have experience with it and I'm unable to test possible side effects (maybe there are not any but I can not be sure).

Maffooch (approved):

PR #11181, opened 2024-11-03
Simpler nginx dockerfile

fopina:
Instead of complex node installation on top of the build base image, this uses separate stage just for yarn and then collectstatic copies the components from it.

I've pinned the same node and yarn versions that were used - but it can always use just `node:20.11` to pick up on latest patches

**NOTE** I've opened another PR with an alternative that I believe to be even better but a bigger change - https://github.com/DefectDojo/django-DefectDojo/pull/11182

### DIFF
--- a/Dockerfile.nginx-alpine
+++ b/Dockerfile.nginx-alpine
@@ -31,113 +31,33 @@ COPY requirements.txt ./
 # https://github.com/unbit/uwsgi/issues/1318#issuecomment-542238096
 RUN CPUCOUNT=1 pip3 wheel --wheel-dir=/tmp/wheels -r ./requirements.txt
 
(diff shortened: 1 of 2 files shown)


kiblik (approved):

valentijnscholten:
@fopina superseded by #11669, I didn't see your PR, let us know if you still want to suggest more changes. Maybe in a new PR?

fopina:
@valentijnscholten I think that PR only does one small thing this one did, not the main goal of it which was "reduce complexity".
Not really related, just a conflict IMO.

But it has been here for very long without any feedback (including now your misunderstanding of the other superseding)
Feel free to close if it's not relevant 

Thanks

valentijnscholten:
Sorry, I linked the wrong PR. This one simplifies the node/yarn install, but slightly different from your approach: #11673

fopina:
Nice

PR #11182, opened 2024-11-03
Dockerfile: merge nginx into django

fopina:
https://github.com/DefectDojo/django-DefectDojo/pull/11181 is a "lighter" simplification but I believe this one is much better

By merging nginx as a target of Dockerfile.django:

* No risk of the `build` stage going out of sync (and wasting precious cache)
* We can actually re-use the `django` stage for collectstatic, saving one stage

> I've only updated the dockerfiles and docker-compose. No docs or any mentions of the files were updated. I can do so, if this PR gets a 👍


kiblik:
I like the idea of simplification.

By the way, different images are not mentioned only in the docker-compose file. They are pushed to [docker-hub](https://github.com/DefectDojo/django-DefectDojo/blob/fb442d2defd9ba4aa9c5ece79cde8e1362f33ffb/.github/workflows/release-x-manual-docker-containers.yml#L30) and used in helm charts. Maybe even more. Can you check other places as well, please?

fopina:
Hi @kiblik yeah, I did add that note to the description, that many places were missing but I didn’t want to waste the time if the idea of reusing same dockerfile was not welcome.
Can I assume it is then?

kiblik:
Sorry, I missed last paragraph of description.

I'm not core maintainer of project. My voice might be supportive but I have no formal power over decision.

dsever:
I don't think we should go to that direction, as you going away from the cloud native principles.

fopina:
@dsever honestly curious as I'm not familiar with their principles, would you be able to quote or link which principle it goes against?

mtesauro:
@fopina  One reason to keep Nginx as a separate container is that it allows users of DefectDojo to choose other options if they want to front DefectDojo with something else. This could be something like Caddy or maybe a LB from their cloud provider.

Yes, I understand that the static assets are built into the Nginx container and someone who chooses to use something different to front DefectDojo would need to also handle that situation but that would mean they'd only need to:
1. Create a new container (maybe) to replace nginx + the static assets
2. Use all the other upstream containers from DefectDojo without modification

So, if for no other reason, I like the fact that keeping Nginx separate gives the community flexibility to change out Nginx without having to modify the other containers needed to run DefectDojo.

mtesauro:
@fopina  Please see [the update](https://github.com/DefectDojo/django-DefectDojo/discussions/11199) I posted earlier today - the project is about to do a fairly significant change/upgrade to the containers for this project. Given that's about to happen, merging two dockerfiles into one doesn't really add much value IMHO. 

It does break our existing GHA tests.

I realize we'll have to update those once the new containers/dockerfiles are in place but why do that work twice when this PR doesn't change any real functionality - rather it create a different *style* of dockerfile.

I'd also prefer to slow roll changes to our containers given the size of our user base. (10+ million pulls). For new hardened container images, we can produce a new image for a period of time to allow people to transition to the new images before we deprecate and eventually stop building the older images.

I appreciate and can understand the desire to potentially optimize 2 dockerfiles into a single file but given the potential side effects, the optimization isn't better than the costs.

fopina:
Understood and agreed.

Impact (known or unknown) was an expected argument against the change and the reason I didn't fully complete it (updating docs and references).
Also why I mention the other PR I created.

I was just confused by which principle this would be against so I would avoid PRs like that again, but I see now it was a misunderstanding.

Thanks, I'll close the PR

mtesauro:
@fopina Thanks for being understanding.

FWIW, I'm trying to find time today to do a proper review the other PR you did - #11181 but at a quick glace, that one looks nice. :+1:

PR #11192, opened 2024-11-05
Ruff: Add and fix S108

kiblik:
Add S108 https://docs.astral.sh/ruff/rules/hardcoded-temp-file/ and fix it.

- In unittests, they are compared to files from findings - they are not used for writing of content - so it is safe
- `on_exception_log_kwarg` from `dojo/decorators.py` is used nowhere in the project. It looks like it is a leftover from previous testing.

### DIFF
--- a/dojo/decorators.py
+++ b/dojo/decorators.py
@@ -144,22 +144,6 @@ def get_parameter_froms_args_kwargs(args, kwargs, parameter):
     return model_or_id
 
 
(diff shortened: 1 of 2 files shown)


cneill (approved):

mtesauro (approved):
Approved

Maffooch (approved):

grendel513 (approved):

PR #11193, opened 2024-11-05
Ruff: Add and "fix" S106

kiblik:
Add S106 https://docs.astral.sh/ruff/rules/hardcoded-password-func-arg/ and mute it.

For testing the user form, it is fine to hardcode the password


cneill (approved):

mtesauro (approved):
Approved

grendel513 (approved):

PR #11198, opened 2024-11-05
Ruff: Add and fix S113

kiblik:
Add RUFF rule S113 https://docs.astral.sh/ruff/rules/request-without-timeout/

Regular parts have been easy to identify. Issues in sessions have [not been marked yet](https://github.com/astral-sh/ruff/issues/14102) but I fixed them.


dojo/settings/settings.dist.py
```diff
@@ -306,6 +306,9 @@
     DD_NOTIFICATIONS_SYSTEM_LEVEL_TRUMP=(list, ["user_mentioned", "review_requested"]),
     # When enabled, force the password field to be required for creating/updating users
     DD_REQUIRE_PASSWORD_ON_USER=(bool, True),
+    # For HTTP requests, how long connection is open before timeout
+    # This settings apply only on requests performed by "requests" lib used in Dojo code (if some included lib is using "requests" as well, this does not apply there)
+    DD_REQUESTS_TIMEOUT=(int, 5),
```
cneill, line 311: I think we'll want to be somewhat conservative here. There are all kinds of requests affected by this, and while an *infinite* timeout is bad, we don't want to force users to constantly fiddle with this setting. I'm just guessing here with a 60 second timeout, maybe others have a better sense for a reasonable default.

```suggestion
    DD_REQUESTS_TIMEOUT=(int, 60),
```

kiblik: I have no strong opinion about the exact value but if a full response is not coming in 60 seconds sth is not right in my opinion.

cneill: After chatting with @Maffooch I think a decent compromise is 30 (updated my suggestion)

mtesauro: I agree that a 30 second timeout is reasonable - especially since people can reduce that if they choose.

cneill (approved):

mtesauro (approved):
Approved

Maffooch (approved):

grendel513 (approved):

PR #11203, opened 2024-11-06
FileUpload Base64 extension fix

hblankenship:
[sc-5645]

Previously, when adding Files to a test, the API does not validate whether the title supplied is associated with a valid file extension. This fix validates the title of the 'files' object for a supported extension, and then return an error message if the file extension is not supported or does not exist. 

This fix applies to the generic json parser and the finding/files, test/files and engagement/files endpoints.


dojo/models.py
```diff
@@ -741,6 +742,28 @@ def get_accessible_url(self, obj, obj_id):
...
+        else:
+            file_name = self.title
+        if Path(file_name).suffix.lower() not in valid_extensions:
+            if accepted_extensions := f"{', '.join(valid_extensions)}":
+                msg = (
+                    _("Unsupported extension. Supported extensions are as follows: %s") % accepted_extensions
+                )
+            else:
+                msg = (
+                    _("File uploads are prohibited due to the list of acceptable file extensions being empty")
```
mtesauro, line 763: Interesting and a good catch. I'm not sure I would have thought of this case

mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

cneill (approved):

PR #11205, opened 2024-11-06
GHA Artifacts: Update to v4

Maffooch:
With the upcoming deprecations of version 3 of `actions/upload-artifact` and `actions/download-artifact`, we must get to v4 by following the [migration guide](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md)

[sc-8285]

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -45,7 +45,11 @@ jobs:
 
       # load docker images from build jobs
       - name: Load images from artifacts
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
+        with:
+          path: built-docker-image
+          pattern: built-docker-image-*
+          merge-multiple: true
 
       - name: Load docker images
         timeout-minutes: 10
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -54,8 +54,8 @@ jobs:
       - name: Load docker images
         timeout-minutes: 10
         run: |-
-             docker load -i nginx/nginx-${{ matrix.os }}_img
-             docker load -i django/django-${{ matrix.os }}_img
+             docker load -i nginx-${{ matrix.os }}_img
+             docker load -i django-${{ matrix.os }}_img
              docker load -i integration-tests/integration-tests-debian_img
              docker images
 
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -54,9 +54,9 @@ jobs:
       - name: Load docker images
         timeout-minutes: 10
         run: |-
-             docker load -i nginx-${{ matrix.os }}_img
-             docker load -i django-${{ matrix.os }}_img
-             docker load -i integration-tests/integration-tests-debian_img
+             docker load -i built-docker-image/nginx-${{ matrix.os }}_img
+             docker load -i built-docker-image/django-${{ matrix.os }}_img
+             docker load -i built-docker-image/integration-tests-debian_img
              docker images
 
       - name: Set integration-test mode
--- a/.github/workflows/k8s-tests.yml
+++ b/.github/workflows/k8s-tests.yml
@@ -58,8 +58,8 @@ jobs:
         timeout-minutes: 10
         run: |-
              eval $(minikube docker-env)
-             docker load -i nginx/nginx-${{ matrix.os }}_img
-             docker load -i django/django-${{ matrix.os }}_img
+             docker load -i nginx-${{ matrix.os }}_img
+             docker load -i django-${{ matrix.os }}_img
              docker images
 
       - name: Configure HELM repos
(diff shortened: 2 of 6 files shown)


mtesauro (approved):
Approved

blakeaowens (approved):

.github/workflows/integration-tests.yml
```diff
@@ -45,14 +45,18 @@ jobs:
...
 
       - name: Load docker images
         timeout-minutes: 10
         run: |-
-             docker load -i nginx/nginx-${{ matrix.os }}_img
-             docker load -i django/django-${{ matrix.os }}_img
-             docker load -i integration-tests/integration-tests-debian_img
+             docker load -i built-docker-image/nginx-${{ matrix.os }}_img
+             docker load -i built-docker-image/django-${{ matrix.os }}_img
+             docker load -i built-docker-image/integration-tests-debian_img
```
cneill, line 59: Will this actually trim the `built-docker-image-` prefix from these image names, or do we need something like this instead? Wasn't clear to me from the migration guide...

```suggestion
             docker load -i built-docker-image/built-docker-image-nginx-${{ matrix.os }}_img
             docker load -i built-docker-image/built-docker-image-django-${{ matrix.os }}_img
             docker load -i built-docker-image/built-docker-image-integration-tests-debian_img
```

.github/workflows/k8s-tests.yml
```diff
@@ -48,14 +48,18 @@ jobs:
...
+          merge-multiple: true
 
       - name: Load docker images
         timeout-minutes: 10
         run: |-
              eval $(minikube docker-env)
-             docker load -i nginx/nginx-${{ matrix.os }}_img
-             docker load -i django/django-${{ matrix.os }}_img
+             docker load -i built-docker-image/nginx-${{ matrix.os }}_img
+             docker load -i built-docker-image/django-${{ matrix.os }}_img
```
cneill, line 62: Same question as above..

```suggestion
             docker load -i built-docker-image/built-docker-image-nginx-${{ matrix.os }}_img
             docker load -i built-docker-image/built-docker-image-django-${{ matrix.os }}_img
```

.github/workflows/rest-framework-tests.yml
```diff
@@ -20,13 +20,17 @@ jobs:
...
+          pattern: built-docker-image-*
+          merge-multiple: true
 
       - name: Load docker images
         timeout-minutes: 10
         run: |-
-             docker load -i nginx/nginx-${{ matrix.os }}_img
-             docker load -i django/django-${{ matrix.os }}_img
+             docker load -i built-docker-image/nginx-${{ matrix.os }}_img
+             docker load -i built-docker-image/django-${{ matrix.os }}_img
```
cneill, line 33: Same as above..

```suggestion
             docker load -i built-docker-image/built-docker-image-nginx-${{ matrix.os }}_img
             docker load -i built-docker-image/built-docker-image-django-${{ matrix.os }}_img
```

cneill (commented):
I'm not entirely sure of the behavior of `merge-multiple` based on the migration documentation, but unless I'm reading something wrong it seems like we need to add the `built-docker-image-` prefix to the images we try to download+load into Docker

.github/workflows/integration-tests.yml
Maffooch: I don't believe the prefix is trimmed. Do we need the prefix to be trimmed?

Maffooch:
> I'm not entirely sure of the behavior of merge-multiple based on the migration documentation, but unless I'm reading something wrong it seems like we need to add the built-docker-image- prefix to the images we try to download+load into Docker

Why would we need the prefix? To make the names of the images more descriptive?

> I also wonder if we need to [specify the overwrite key on upload](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#overwriting-an-artifact) since we're not including something like the workflow run ID/etc in the artifact name.

I am not sure about this one. With the upcoming brownouts next week, I think it would be preferable to get this merged in, and then investigate how the overwrite flag will add value

Also please keep in mind, if the docker images are able to be downloaded so that the unit tests can run, then this implementation is functioning as expected. Please see previous failed implementations in the commit history to see what all went into this 😄

grendel513 (approved):

.github/workflows/integration-tests.yml
Maffooch: Also I didn't realize when I left the last comment, but the names of the images themselves do not have a `built-docker-image-` prefix. The name of the image stored is captured in the `path` entry:
```
path: ${{ matrix.docker-image }}-${{ matrix.os }}_img
```

This value is unchanged in this PR

rossops (approved):

Maffooch:
@cneill we can discuss how this can be iterated on after the release today 😄

.github/workflows/integration-tests.yml
cneill: Ahh, I think I mixed up the [`with` (artifact name)](https://github.com/Maffooch/django-DefectDojo/blob/ce0763271668b042b51489a6b730d700bbcd33ee/.github/workflows/build-docker-images-for-testing.yml#L54) and the [`path`](https://github.com/Maffooch/django-DefectDojo/blob/ce0763271668b042b51489a6b730d700bbcd33ee/.github/workflows/build-docker-images-for-testing.yml#L55)

Issue #11206, opened 2024-11-06
Acunetix Parser - month and day inversed in date


ptitkosmos:
**Bug description**
The detection date is wrongly parsed resulting in a swap between month and day in this date. E.g. 12th February 2024 becomes 2nd December 2024

**Steps to reproduce**
1. Create a Acunetix json file or use the attached file with the FirstSeenDate field filled with a revertable date : e.g. "FirstSeenDate": "12/02/2024 04:29 PM"
2. Import the file
3. See that the detection date is reverted

**Expected behavior**
The date should be well parsed.

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Defectdojo version
2.39.2

**Sample scan files**
If applicable, add sample scan files to help reproduce your problem.
[Uploading 3af6a0aa-1d48-43ad-851f-54329e152545.json…]()

**Screenshots**
![image](https://github.com/user-attachments/assets/968d5a9e-1834-468a-8616-fb86d6a31477)
![image](https://github.com/user-attachments/assets/4202d2c1-8f69-49bb-afdc-53de9fd10f4d)

ptitkosmos:
@manuel-sommer Thanks 😊

PR #11207, opened 2024-11-07
:bug: fix Acunetix date #11206

manuel-sommer:
#11206


cneill (commented):
I checked some of our example scan files to be sure this wasn't a user-configurable thing that would vary based on the user's timezone/preferences, and it at least appears that this is consistent. Unfortunately, changing this behavior didn't trigger any unit test failures, meaning we're not evaluating it at all today. Would you mind adding some test cases, @manuel-sommer ?

cneill (approved):
Thanks! 😄

Maffooch (approved):

blakeaowens (approved):

PR #11212, opened 2024-11-07
🐛 Fix Defender broken Endpoint

manuel-sommer:
Field ComputerDNS name is more a UserInfo
ComputerDNS name contains spaces which breaks the regex in models.py

### DIFF
--- a/dojo/tools/ms_defender/parser.py
+++ b/dojo/tools/ms_defender/parser.py
@@ -131,7 +131,7 @@ class MSDefenderParser:
         self.findings.append(finding)
         finding.unsaved_endpoints = []
         if machine["computerDnsName"] is not None:
-            finding.unsaved_endpoints.append(Endpoint(host=str(machine["computerDnsName"])))
+            finding.unsaved_endpoints.append(Endpoint(host=str(machine["computerDnsName"]).replace(" ","")))
         if machine["lastIpAddress"] is not None:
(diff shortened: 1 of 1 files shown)


kiblik:
I'm not sure about putting DNS name to UserInfo. It is quite well defined, what should be there

> An optional userinfo subcomponent followed by an at symbol (@), that may consist of a [user name](https://en.wikipedia.org/wiki/User_(computing)) and an optional [password](https://en.wikipedia.org/wiki/Password) preceded by a colon (:). Use of the format username:password in the userinfo subcomponent is deprecated for security reasons. Applications should not render as clear text any data after the first colon (:) found within a userinfo subcomponent unless the data after the colon is the empty string (indicating no password).

https://en.wikipedia.org/wiki/URL#Syntax

kiblik (approved):

cneill (commented):
Would you mind adding one or more unit test cases that evaluate the behavior you're seeing in the wild @manuel-sommer ? I'm surprised space, `(`, and `)` characters are ever showing up in real "DNS names". I wonder if these are really NetBIOS names or something, and if we should perhaps distinguish them from real DNS entries somehow rather than forcing them to conform to DNS structure here by just replacing those characters.

mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

cneill (approved):

manuel-sommer:
Could you merge this @mtesauro?

PR #11220, opened 2024-11-07
Burp Enterprise: Support newer format

Maffooch:
- Simplify/Solidify BurpE parser to work with newer formats
- Add support for CAPEC plus multiple CWE via vulnerability IDs
- Clean up finding format

[sc-8300]


blakeaowens (approved):

mtesauro (approved):
Approved

grendel513 (approved):

fopinappb:
Hi @Maffooch I see in this refactor the commonly "dedupe within parser" was left out.

Is this the new approach, leave it to the post-save deduplication?

That would make sense to me to reduce complexity and chance of inconsistency

Maffooch:
Hi @fopinappb thanks for the questions! The decision to dedupe within the parser or not is often a nuanced internal discussion. I have found that it is entirely dependent on how the report is structured. With Burp Enterprise specifically, the reports are structured with the title of a finding, and then all endpoints underneath it to reflect the many instances of a given vulnerability.

The changes made in this PR still do the dedupe within the parser, but it is not done in the same way as many other parsers. This is due to the table of contents at the beginning of a burp enterprise report. This table of contents is the best place to determine what endpoints are set on a give vulnerability. The parser works by [first scraping the table of contents, and aggregating results there based on finding title](https://github.com/DefectDojo/django-DefectDojo/blob/e80b7d99b6f2dd474f7cdbb43b2adcb41fff71b6/dojo/tools/burp_enterprise/parser.py#L37-L83). Once that is complete, more details about the vulnerability itself are added to the aggregated value later on in the [create_findings](https://github.com/DefectDojo/django-DefectDojo/blob/e80b7d99b6f2dd474f7cdbb43b2adcb41fff71b6/dojo/tools/burp_enterprise/parser.py#L227-L230) function

> I see hashcode fields were not changed and continue to NOT include description which is the only place with the full path for the finding. This will make issues of same category on different paths/parameters to be considered duplicates

Do you have an example of a report with query parameters that can be used to determine how to proceed? I would imagine the table of contents on the report would list paths with different query parameters as unique items, but that may not be the case

fopina:
Thanks for taking the time to comment @Maffooch 

I think my main question/misunderstanding is probably addressed in the other issue I opened for burp parser too, so I'll continue there for focusing!

PR #11229, opened 2024-11-10
:hammer: rework kubescape parser

manuel-sommer:
Fix multiple Kubescape parser issues:
- Add vulnid
- Remove passed rules https://hub.armosec.io/docs/scan-status
- Add vulnid to VULNERABILITY_URLS


mtesauro (approved):
Approved

Maffooch (approved):

dojo/settings/settings.dist.py
```diff
@@ -1744,6 +1744,7 @@ def saml2_attrib_map_format(dict):
     "ELSA": "https://linux.oracle.com/errata/&&.html",  # e.g. https://linux.oracle.com/errata/ELSA-2024-12714.html
     "ELBA": "https://linux.oracle.com/errata/&&.html",  # e.g. https://linux.oracle.com/errata/ELBA-2024-7457.html
     "RXSA": "https://errata.rockylinux.org/",  # e.g. https://errata.rockylinux.org/RXSA-2024:4928
+    "C-": "https://hub.armosec.io/docs/",  # e.g. https://hub.armosec.io/docs/c-0085
```
cneill, line 1747: It seems somewhat unlikely that a `C-` prefix will be completely unique to ARMO... We may need to think in the future about how to better categorize different vulnerability IDs. It's getting increasingly convoluted to handle all these different patterns in this way with display tags.

No action needed on this PR, mostly leaving this as a note for myself and to get others' feedback.

dojo/tools/kubescape/parser.py
```diff
@@ -59,65 +59,62 @@ def get_findings(self, filename, test):
...
+                            description += "**Category:** " + category + "\n"
+                        elif controlSummary is not None and "category" in controlSummary and "name" in controlSummary["category"]:
+                            category = controlSummary["category"]["name"]
+                            description += "**Category:** " + category + "\n"
+
+                        steps_to_reproduce = "The following rules have failed :" + "\n"
+                        steps_to_reproduce += "\t**Rules:** " + str(json.dumps(control["rules"], indent=4)) + "\n"
+                        steps_to_reproduce += "Resource object may contain evidence:" + "\n"
+                        steps_to_reproduce += "\t**Resource object:** " + str(json.dumps(resource["object"], indent=4))
+                        if rule["status"] != "passed":
```
cneill, line 104: It would be clearer/more efficient to handle this on line 63 instead of waiting until the end and indenting this block:

```python
if rule["status"] == "passed":
    continue
```

dojo/tools/kubescape/parser.py
```diff
@@ -59,65 +59,62 @@ def get_findings(self, filename, test):
...
+                                steps_to_reproduce=steps_to_reproduce,
+                                severity=severity,
+                                component_name=resourceid,
+                                static_finding=True,
+                                dynamic_finding=False,
+                        )
+                            findings.append(find)
+                            if controlID is not None:
+                                find.unsaved_vulnerability_ids = []
+                                find.unsaved_vulnerability_ids.append(controlID)
```
cneill, line 119: Generally clearer to finish defining the Finding before appending it to the `findings` list.

```suggestion
                            )
                            if controlID is not None:
                                find.unsaved_vulnerability_ids = []
                                find.unsaved_vulnerability_ids.append(controlID)
                            findings.append(find)
```

dojo/tools/kubescape/parser.py
```diff
@@ -59,65 +59,62 @@ def get_findings(self, filename, test):
...
+                            find = Finding(
+                                title=textwrap.shorten(title, 150),
+                                test=test,
+                                description=description,
+                                mitigation=mitigation,
+                                steps_to_reproduce=steps_to_reproduce,
+                                severity=severity,
+                                component_name=resourceid,
+                                static_finding=True,
+                                dynamic_finding=False,
```
cneill, line 114: While it's fine to add this vulnerability ID type as one that turns into a link in the UI, it shouldn't be removed from `references`. The UI link won't be usable by anyone pulling findings via the API.

cneill (changes requested):
Adding a link for the `C-` prefix is fine, but the link should still be included in `references`.

dojo/settings/settings.dist.py
manuel-sommer: Yeah, you are right. This `C-` prefix is used to not mix `C` with `CVE`.

dojo/tools/kubescape/parser.py
manuel-sommer: Yeah, good point :-)

grendel513 (approved):

cneill (approved):

PR #11256, opened 2024-11-13
Ruff: Add and fix S110 (+ merge all S1 rules)

kiblik:
Add and fix https://docs.astral.sh/ruff/rules/try-except-pass/


Maffooch (changes requested):
Instead of passing, could we get by on logging the exception? That seems safer than removing the exception handling altogether

kiblik:
> Instead of passing, could we get by on logging the exception? That seems safer than removing the exception handling altogether

In general, I do not consider "suppressing exceptions and replacing them with writing to error log" as the best approach. Admins usually check logs only if there is wrong behavior or if they have some smart rules in their central monitoring. Simple ignoring of errors is not the best.

I made rewrites and I suppose I solved most of the cases which might be problematic.

- `dojo/benchmark/views.py`: If benchmarks are not created, we should not silently ignore them. If they were only logged, it might bring frustration to users who do not receive any feedback, and just ignore the user’s input.
- `dojo/cred/views.py`: I rewrite to a method that has the same outcome but without exception
- `dojo/finding/helper.py`: Also ”no-exception” rewrite 
- `dojo/middleware.py`: Here catching makes sense but I decreased it to reasonable scope
- `dojo/product/views.py`: Might be a good idea to test this but if there is an issue in the form, the user should be informed, not ignored
- `dojo/templatetags/display_tags.py`: I believe I made a rewrite that avoids all reasonable situations 
- `dojo/tools/gitlab_api_fuzzing/parser.py`: Also ”no-exception” rewrite 
- `dojo/tools/h1/parser.py`: Also ”no-exception” rewrite 
- `dojo/tools/kiuwan/parser.py`: Also ”no-exception” rewrite 
- `dojo/tools/veracode/json_parser.py`: Also ”no-exception” rewrite 
- `tests/Import_scanner_test.py`: This is a test and it is not failing - so there is probably no problematic situation
- `tests/base_test_class.py`: Also ”no-exception” rewrite (using of existing function)

But I’m happy for any feedback.

kiblik:
@Maffooch, how do you see this PR? Do you accept my adjustments and reasonings?

blakeaowens (approved):

hblankenship (approved):

cneill (approved):

PR #11257, opened 2024-11-13
feat(helm): Allow to keep initializer if requested

kiblik:
If the user needed to keep the `initializer` in the stack, there was no way (only set an unreasonable large number).

Now, any value that is not a positive integer will keep the `initializer` there.

Tested with:
- `--set 'initializer.keepSeconds=[]'` - stayed
- `--set 'initializer.keepSeconds=null'` - stayed
- `--set 'initializer.keepSeconds=x'` - stayed
- `--set 'initializer.keepSeconds=-1'` - stayed
- `--set 'initializer.keepSeconds=0'` - stayed
- `--set 'initializer.keepSeconds=1'` - removed after 1s

Context:

### DIFF
--- a/helm/defectdojo/templates/initializer-job.yaml
+++ b/helm/defectdojo/templates/initializer-job.yaml
@@ -15,7 +15,9 @@ metadata:
     {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
+  {{- if and (int .Values.initializer.keepSeconds) (gt (int .Values.initializer.keepSeconds) 0) }}
   ttlSecondsAfterFinished: {{ .Values.initializer.keepSeconds }}
+  {{- end }}
   template:
     metadata:
       labels:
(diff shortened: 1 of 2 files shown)


mtesauro (approved):
Approved

Maffooch (approved):
@kiblik I am not very familiar with helm, but what would be the use case of keeping the init container around longer?

kiblik:
> @kiblik I am not very familiar with helm, but what would be the use case of keeping the init container around longer?

It is not connected to HELM itself. Quite often I see cases when migration was not executed during the upgrade or it fails for some reason. Keeping a container (hopefully with logs) should help with debugging - in case logs are needed as evidence. Yes, log collection to some other place is highly recommended in best practice but I suppose this is easier for users who miss this kind of solution. As far as I know, finished jobs do not consume additional resources (CPU, memory).

Plus, right now user is forced to accept the setting (that the job will be removed). Yes, values can be adjusted to much higher but it is not possible to avoid removal (unless you set it to unreasonable value).

This PR does not change the previous default behavior.

Maffooch:
Ahh that makes perfect sense

cneill (approved):

blakeaowens (approved):

PR #11269, opened 2024-11-15
AnchoreCTL Policies: Additional checks for severity in description

hblankenship:
[sc-1018]

Modified the way severity is used in the anchorectl_policies parser. Now, if the description contains the severity as the first word, it will use that. Otherwise, it drops back to using the status to determine severity.

Finally, in the case of the 'go' status, severity is set to Low by default.


Maffooch (approved):

cneill (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #11273, opened 2024-11-16
SAML2 group mapping

fopina:
**Description**

While SAML2 *authentication* works well as it is, it's not very useful for *authorization*.

This PR introduces the ability to map SAML2 IdP groups (or some other list attribute) to Dojo groups.
And its configuration is similar to Azure AD group matching:
* Take all groups from IdP that match a configurable regexp
* Take all groups from the user on Dojo that match the same regexp
* Take all existing groups in Dojo that match the same regexp

**Testing**

> Auth0 was used to test the setup as it support SAML2

* Create new application, `Regular Web Application`
* Under application addons, enable `SAML2 web-app`
* Use settings
```json
{
 "mappings": {
   "user_id":     "urn:oid:0.9.2342.19200300.100.1.1",
   "email":       "urn:oid:1.2.840.113549.1.9.1.1",
   "given_name":  "urn:oid:2.5.4.3",
   "family_name": "urn:oid:2.5.4.4",
   "groups":      "http://schemas.xmlsoap.org/claims/Group"
 },
 "passthroughClaimsWithNoMapping": false
}
```
* Add `http://localhost:8080/saml/acs/` as callback url and click `Enable`

As there are no groups in Auth0 and roles are not shared via SAML assertions, we can push them (roles) into a custom assertion (based on [this](https://community.auth0.com/t/send-roles-as-part-of-saml-assertion-when-auth0-is-the-idp/109864) and [this](https://auth0.com/docs/authenticate/protocols/saml/saml-configuration/customize-saml-assertions#example-include-user_metadata-attributes-in-an-assertion) links

* Go to Actions choose Trigger
* Choose postlogin trigger
```javascript
/**
* Handler that will be called during the execution of a PostLogin flow.
*
* @param {Event} event - Details about the user and the context in which they are logging in.
* @param {PostLoginAPI} api - Interface whose methods can be used to change the behavior of the login.
*/
exports.onExecutePostLogin = async (event, api) => {
  var roles = event.authorization.roles;
	api.samlResponse.setAttribute('urn:oid:2.5.4.33', roles);
};
```

* Back to Dojo, set the following extra configuration (via env):

```env
DD_SAML2_ENABLED: 'true'
DD_SAML2_METADATA_AUTO_CONF_URL: <METADATA URL FROM AUTH0 SAML2 ADDON>
DD_SITE_URL: http://localhost:8080
DD_SOCIAL_LOGIN_AUTO_REDIRECT: "true"
DD_SOCIAL_AUTH_SHOW_LOGIN_FORM: "false"
DD_SAML2_CREATE_USER: "true"
DD_SAML2_ATTRIBUTES_MAP: email=email,uid=username,cn=first_name,sn=last_name
DD_SAML2_GROUPS_FILTER='.*'
DD_SAML2_GROUPS_ATTRIBUTE=roleOccupant
```

* Now create some roles and add them to the test user in Auth0.
* Create Dojo groups with the exame same name
* Login using that test user and the user should be part of the mapped Dojo groups
  * You can verify in logs that those groups were deteted


dojo/backends.py
```diff
@@ -0,0 +1,82 @@
...
+
+class Saml2Backend(_Saml2Backend):
+
+    """Subclass to handle adding SAML2 groups as DefectDojo/Django groups to a user"""
+
+    @cached_property
+    def group_re(self):
+        if not settings.SAML2_ENABLED or not settings.SAML2_GROUPS_ATTRIBUTE or not settings.SAML2_GROUPS_FILTER:
+            return None
+        return re.compile(settings.SAML2_GROUPS_FILTER)
```
kiblik, line 22:
```suggestion
        if settings.SAML2_ENABLED and settings.SAML2_GROUPS_ATTRIBUTE and settings.SAML2_GROUPS_FILTER:
            return re.compile(settings.SAML2_GROUPS_FILTER)
        return None
```
same functionality but easier to read

dojo/backends.py
```diff
@@ -0,0 +1,82 @@
...
+
+        This does NOT create the groups if they do not exist. They have to be created in the UI
+        This is not a big issue and it works around an existing bug with dojo/group/utils.py::group_post_save_handler (user does not yet exist and he is forcefully added to the new group - boom)
+        """
+        user = super()._update_user(user, attributes, attribute_mapping, force_save=force_save)
+        if self.group_re is None:
+            return user
+
+        # list of all existing "SAML2-mapped" groups
+        all_saml_groups = {group.name: group for group in Dojo_Group.objects.all() if self.group_re.match(group.name)}
```
kiblik, line 44:
```suggestion
        all_saml_groups = {group.name: group for group in Dojo_Group.objects.filter(social_provider=Dojo_Group.SAML) if self.group_re.match(group.name)}
```

dojo/settings/settings.dist.py
```diff
@@ -177,6 +177,11 @@
         "Lastname": "last_name",
     }),
     DD_SAML2_ALLOW_UNKNOWN_ATTRIBUTE=(bool, False),
+    # similar to DD_SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_GROUPS_FILTER, regular expression for which SAML2 groups to map to Dojo groups (with same name)
+    # Groups need to already exist in Dojo. And if value is not set, no group processing is done
+    DD_SAML2_GROUPS_FILTER=(str, ""),
```
kiblik, line 182: Can you call it `_READERS` as it adds a Reader role? In the future 
```suggestion
    DD_SAML2_GROUPS_FILTER_READERS=(str, ""),
```
Or even more future-proof (if new roles would be created): dict with mapping to roles
```suggestion
    DD_SAML2_GROUPS_FILTER=(dict, {"readers": ""}),
```
More info, about how dict works with `environ`
- https://django-environ.readthedocs.io/en/latest/types.html#environ-env-dict
- https://django-environ.readthedocs.io/en/latest/tips.html#complex-dict-format

kiblik (commented):
I like this change. I was considering similar functionality in the past, and I'm happy that somebody has done it now.

A couple of comments:
- Thanks for the nice explanation in the description of PR, can you put it in the documentation as well? https://documentation.defectdojo.com/integrations/social-authentication/
- If you have an implementation that can filter-out the right list of groups, why not create them? I'm not saying it has to be default, but opt-in.
- Other social provider's groups are markered (Remote, Azure). Can you add ”SAML” as a social provider? https://github.com/DefectDojo/django-DefectDojo/blob/090b2c7a77a67f02064b158559e5fcdc665c4291/dojo/models.py#L257

Thank you

fopina:
Thank you @kiblik 

Quick feedback on number 2: my initial implementation was to create the groups on demand (not optional) but there's the save signal on groups that uses user for that request and fails (with NoneType exception as it is unexpected) because there's no user (at that time yet as not done authenticating).

I decided to remove creation as that also allows more granular control on the groups.

If I make it optional too, it would allow both things, but what's the right fix on the owner? Should I modify the signal too to allow a group without owner?

kiblik:
> If I make it optional too, it would allow both things, but what's the right fix on the owner? Should I modify the signal too to allow a group without owner?

Is it the owner of a group? I'm not able to find this attribute. Can you send the link to which signal was failing, please?

fopina:
I believe it was this one https://github.com/fopina/django-DefectDojo/blob/268583954fb74635fc708c49d8826e1534cd8f5d/dojo/group/utils.py#L37
Not 100% sure as I’ve been using this in my fork for a few months now 

Easy to fix in several ways, not sure which one is preferred

kiblik:
> Easy to fix in several ways, not sure which one is preferred

It is hard to say about preference but [impersonate](https://django-crum.readthedocs.io/en/latest/#impersonate-user-none) came to my mind first.

fopina:
I meant less technically but more business oriented: which user should own the group? Should anyone?

dojo/backends.py
fopina: I guess this makes sense thinking of the AzureAD pipeline yet it "forces" groups to have been created by this (instead of added manually to match IdP groups), right?

dojo/settings/settings.dist.py
fopina: At first sight, I agree this would make sense, but does it really? When would anyone not be just reader in a group where the membership is expected to be managed by IdP?

Noone should be able to add/remove people from the group in Dojo, they should do it in IdP, otherwise that person will eventually be removed from the group (upon login)

Maybe I'm missing some use case in my mind, please let me know if so. Otherwise this would just increase complexity of the configuration without any added benefit

fopina:
@kiblik my bad, AzureAD pipeline [does create the groups](https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/pipeline.py#L113) but one of the required settings skips that owner user in the signal.

I guess I can simply do the same then, I'll update the PR next week 👍

Bagautdino (approved):

mtesauro:
Closing as stale

PR #11276, opened 2024-11-17
:bug: fix trivyoperator tags

manuel-sommer:
Some TrivyOperator tags were empty

### DIFF
--- a/dojo/tools/trivy_operator/vulnerability_handler.py
+++ b/dojo/tools/trivy_operator/vulnerability_handler.py
@@ -83,9 +83,8 @@ class TrivyVulnerabilityHandler:
                 dynamic_finding=False,
                 service=service,
                 file_path=file_path,
+                tags = [tag for tag in finding_tags if tag != ""]
             )
-            if finding_tags != "":
-                finding.tags = finding_tags
             if vuln_id:
                 finding.unsaved_vulnerability_ids = [UniformTrivyVulnID().return_uniformed_vulnid(vuln_id)]
             findings.append(finding)
--- a/dojo/tools/trivy_operator/vulnerability_handler.py
+++ b/dojo/tools/trivy_operator/vulnerability_handler.py
@@ -83,7 +83,7 @@ class TrivyVulnerabilityHandler:
                 dynamic_finding=False,
                 service=service,
                 file_path=file_path,
-                tags = [tag for tag in finding_tags if tag != ""]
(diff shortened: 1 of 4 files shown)


mtesauro (approved):
Approved

dojo/tools/trivy_operator/vulnerability_handler.py
```diff
@@ -83,8 +83,9 @@ def handle_vulns(self, labels, vulnerabilities, test):
                 dynamic_finding=False,
                 service=service,
                 file_path=file_path,
-                tags=finding_tags,
             )
```
cneill, line 86: Since `finding_tags` is defined as a list on line 35, it'll never be `""`. You could do something like this instead:

```python
tags=[tag for tag in finding_tags if tag != ""]
```

manuel-sommer:
@cneill , no resource_namespace shouldn't be a list here. 
Before the fix an example:
grafik
After the fix, the same example:
grafik

cneill:
Hm. I can't replicate this behavior with the exposed secrets scan report [here](https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/scans/trivy_operator/exposedsecretreport_many.json). I get a single `default` tag either way. I didn't realize a single string was a valid input for the `tags` field. All the other parsers provide lists of tags (e.g. [Dependency Check](https://github.com/DefectDojo/django-DefectDojo/blob/8b20fa0a3f55691f7707dedab50b46bde90d5c8b/dojo/tools/dependency_check/parser.py#L325), [Threat Composer](https://github.com/DefectDojo/django-DefectDojo/blob/8b20fa0a3f55691f7707dedab50b46bde90d5c8b/dojo/tools/threat_composer/parser.py#L87), [Meterian](https://github.com/DefectDojo/django-DefectDojo/blob/8b20fa0a3f55691f7707dedab50b46bde90d5c8b/dojo/tools/meterian/parser.py#L72)...).

Does your scan file differ in some way from the one in `unittests`?

cneill (approved):

hblankenship (approved):

Maffooch (approved):

Maffooch:
@manuel-sommer I am not sure if this ready for merge or not. Let me know if you are ready for this one to go in 😄

manuel-sommer:
In case I find another bug, I will submit a PR regarding trivy operator. I am actively testing it for soonish future use.

PR #11277, opened 2024-11-17
Fix nuclei deduplication

fopina:
**Description**

Many nuclei templates have several _matchers_, such as [tech-detect](https://github.com/projectdiscovery/nuclei-templates/blob/main/http/technologies/tech-detect.yaml)
If there are several matchers, each is reported as a separate finding.

The current parser puts matcher under `component_name` yet the hash fields do not include `component_name` resulting in findings being closed as duplicates when they're not (such as multiple technologies for the mentioned template)

The deduplication done within the nuclei parser itself considers matcher as key already which was not aligned with hash fields, as those didn't consider `component_name`.

## alternative

Another option could be appending matcher to `title` instead of changing the hash code fields. This would allow seeing the matcher in the findings list (that does not show components) but we would lose the ability to filter by component.


cneill (approved):

Maffooch (approved):

Maffooch:
@fopina this PR says there are conflicts, but idk what they are though... Pretty strange

fopina:
@Maffooch I received the notification yet it's still missing approvals and I was considering just waiting for all approvals before resolving them

Yet, as I came here to comment, I also resolved them 👍  (it was the settings shasum)

mtesauro (approved):
Approved

blakeaowens (approved):

PR #11298, opened 2024-11-20
dissallow already linked issue

### LINKED ISSUE #9930: Finding Jira Mapping API allows assignment of a single Jira Issue to multiple findings
**Bug Description**:

When using the `jira_finding_mappings` API endpoint in DefectDojo, it is possible to update a finding's Jira mapping with a Jira issue that is already assigned to another finding. This action does not raise any errors, but it results in unexpected behavior. Specifically, after updating the mapping, the integration between Jira and DefectDojo only works with the original finding. For instance, if the Jira issue is closed, only the original issue will be affected, not the new assigned finding.

Additionally, attempting to map an already existing Jira issue through the UI results in an "already linked" error, indicating that the UI correctly prevents this action.

**Steps to reproduce**
1. Use the PUT `jira_finding_mappings` API endpoint to update a finding's Jira mapping with a Jira issue already assigned to another finding
2. Observe that no errors are raised
3. Notice that the integration between Jira and DefectDojo only affects the original finding, not the new assigned finding

**Expected behavior**
The API jira_finding_mappings endpoint should raise an error or prevent updating a finding's Jira mapping with a Jira issue that is already assigned to another finding. 

- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version v. 2.31.4

hblankenship:
[sc-5525]

Fixes #9930 

When using the jira_finding_mappings API endpoint, trying to update a finding's Jira mapping with a Jira issue that is already assigned to another finding will now raise a validation error.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1287,6 +1287,12 @@ class JIRAIssueSerializer(serializers.ModelSerializer):
             msg = "Either engagement or finding or finding_group has to be set."
             raise serializers.ValidationError(msg)
 
+        if finding:
+            linked_finding = jira_helper.jira_already_linked(finding, data.get("jira_key"), data.get("jira_id"))
+            if linked_finding:
+                msg = "JIRA issue " + data.get("jira_key") + " already linked to " + reverse("view_finding", args=(linked_finding,))
+                raise serializers.ValidationError(msg)
+
         return data
 
 
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -1426,6 +1426,17 @@ def add_simple_jira_comment(jira_instance, jira_issue, comment):
         return False
 
 
+def jira_already_linked(finding, jira_issue_key, jira_id):
+    jira_issues = JIRA_Issue.objects.filter(jira_id=jira_id, jira_key=jira_issue_key).exclude(engagement__isnull=False)
+    jira_issues = jira_issues.exclude(finding=finding)
+
+    result = 0
+    if len(jira_issues) > 0:
+        result = jira_issues[0].finding_id
+
+    return result
+
+
 def finding_link_jira(request, finding, new_jira_issue_key):
     logger.debug("linking existing jira issue %s for finding %i", new_jira_issue_key, finding.id)
 


dojo/jira_link/helper.py
```diff
@@ -1426,6 +1426,17 @@ def add_simple_jira_comment(jira_instance, jira_issue, comment):
...
 
+def jira_already_linked(finding, jira_issue_key, jira_id):
+    jira_issues = JIRA_Issue.objects.filter(jira_id=jira_id, jira_key=jira_issue_key).exclude(engagement__isnull=False)
+    jira_issues = jira_issues.exclude(finding=finding)
+
+    result = 0
+    if len(jira_issues) > 0:
+        result = jira_issues[0].finding_id
+
+    return result
```
Maffooch, line 1437:
```suggestion
def jira_already_linked(finding, jira_issue_key, jira_id) -> Finding | None:
    jira_issues = JIRA_Issue.objects.filter(jira_id=jira_id, jira_key=jira_issue_key).exclude(engagement__isnull=False)
    jira_issues = jira_issues.exclude(finding=finding)

    return jira_issues.first()
```

dojo/api_v2/serializers.py
```diff
@@ -1287,6 +1287,12 @@ def validate(self, data):
             msg = "Either engagement or finding or finding_group has to be set."
             raise serializers.ValidationError(msg)
 
+        if finding:
+            linked_finding = jira_helper.jira_already_linked(finding, data.get("jira_key"), data.get("jira_id"))
+            if linked_finding:
+                msg = "JIRA issue " + data.get("jira_key") + " already linked to " + reverse("view_finding", args=(linked_finding,))
+                raise serializers.ValidationError(msg)
```
Maffooch, line 1294:
```suggestion
            if (linked_finding := jira_helper.jira_already_linked(finding, data.get("jira_key"), data.get("jira_id"))) is not None:
                msg = "JIRA issue " + data.get("jira_key") + " already linked to " + reverse("view_finding", args=(linked_finding.id,))
                raise serializers.ValidationError(msg)
```

Maffooch (changes requested):
Updated the function to return the whole finding rather than just the ID. It could be useful in the future

Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #11299, opened 2024-11-20
Settings SHA: The Removal

Maffooch:
Following the introduction of the settings.dist.py integrity check, it has made contributing to DefectDojo more challenging. As PRs are merged with settings changes for parsers, vulnerability IDs, and new environment variables, everyone else with similar changes will now have conflicts. At times this could impact many people at once, and even multiple times within a singe PR. 

Thought the integrity check is a great idea in enforcing that settings.dist.py not be touched, the tradeoff for ease of development is a turnoff for contributors

[sc-8583]


kiblik:
Q: If `HASHCODE_FIELDS_PER_SCANNER`, `HASHCODE_ALLOWS_NULL_CWE` and `DEDUPLICATION_ALGORITHM_PER_PARSER` would be excluded from `settings.dist.py` (e.g. they would be part of parsers `dojo/X/parser.py`: `XParser.get_deduplication_algo()` + dynamically load them), can you consider to keep hash-checker? It would bring also cleaner labeling of PRs (right now simple change of hashcode fields adds PR label as "change of settings of the whole system" but it is more change of some parser's behavior)

mtesauro:
> Q: If `HASHCODE_FIELDS_PER_SCANNER`, `HASHCODE_ALLOWS_NULL_CWE` and `DEDUPLICATION_ALGORITHM_PER_PARSER` would be excluded from `settings.dist.py` (e.g. they would be part of parsers `dojo/X/parser.py`: `XParser.get_deduplication_algo()` + dynamically load them), can you consider to keep hash-checker? It would bring also cleaner labeling of PRs (right now simple change of hashcode fields adds PR label as "change of settings of the whole system" but it is more change of some parser's behavior)

@kiblik Taking a step back, I'm curious the benefit you though came from the current sha256sum of settings.dist.py. To me, it was a signal to users that they shouldn't be editing settings.dist.py but using local_settings.py for their deploy customizations.

Your reply focuses more on hash codes so I'm wondering if I missed something about the original intent of adding that sha256sum check.

kiblik:
> > Q: If `HASHCODE_FIELDS_PER_SCANNER`, `HASHCODE_ALLOWS_NULL_CWE` and `DEDUPLICATION_ALGORITHM_PER_PARSER` would be excluded from `settings.dist.py` (e.g. they would be part of parsers `dojo/X/parser.py`: `XParser.get_deduplication_algo()` + dynamically load them), can you consider to keep hash-checker? It would bring also cleaner labeling of PRs (right now simple change of hashcode fields adds PR label as "change of settings of the whole system" but it is more change of some parser's behavior)
> 
> @kiblik Taking a step back, I'm curious the benefit you though came from the current sha256sum of settings.dist.py. To me, it was a signal to users that they shouldn't be editing settings.dist.py but using local_settings.py for their deploy customizations.
> 
> Your reply focuses more on hash codes so I'm wondering if I missed something about the original intent of adding that sha256sum check.

Thank you @mtesauro. Good point, I haven't described my story behind it.
From my observation (but please correct me if I'm wrong), most of the edits of `settings.dist.py` are connected to changes in parser parameters (`HASHCODE_FIELDS_PER_SCANNER`, `HASHCODE_ALLOWS_NULL_CWE`, `DEDUPLICATION_ALGORITHM_PER_PARSER`). So they are also the biggest reason for merge conflicts.
If these fields were excluded from `settings.dist.py`, the number of friction points would decrease and the user (developers) experience would increase.
I had an idea of the distribution of the mentioned fields to parsers in the back of my head for a longer period of time but I have not communicated/implemented it until now. I thought it would "fix" this problematic factor.
But if you think that the ratio of changes is different and the number of "other" changes is significantly higher (PRs will still affect each other a lot) I would like to ask whether there is no other way to achieve the original goal (inform the user that direct editing of `settings.dist.py` is not recommended way of configuring this application).

manuel-sommer:
What if we just add this information to the documentation?
If things go wrong in personal deployments due to changes of settings.dist.py, they did not read the documentation. Then, we can just refere to the documentation in issues / discussions.

Maffooch:
There is already a pretty good entry in the documentation warning about the changing `settings.dist.py` with respect to updates: https://documentation.defectdojo.com/getting_started/configuration/

What scares me about moving deduplication settings out of settings.py is that I fear they may not be configurable without rebuilding images.  If that were not an issue, and the task was still to move ~85 dedupe settings to the parser level, it makes me wonder: Is all of that work (and potential migration woes/instability on update) worth it to keep an integrity check?

Maffooch:
It is sorta humorous to me that removing the integrity check is causing conflicts around the integrity check 😅

cneill:
> I would like to ask whether there is no other way to achieve the original goal (inform the user that direct editing of `settings.dist.py` is not recommended way of configuring this application).

It seems like the only way to make this less painful would be to auto-generate the checksum at PR merge and/or release time, rather than asking developers to manage it. This would prevent merge conflicts while preserving the original intent, which I think has some value. Today, auto-generation [only happens as part of automated merges](https://github.com/DefectDojo/django-DefectDojo/pull/11299/files#diff-d0c02aeb8aa3b8577e0fbda547ffdf2f647ea6b25179af74db42db9994289a6e), e.g. dev -> master, but it could in principle be extended to all PR merges that affect the settings file.

This would require a flag (e.g. `DD_DISABLE_SETTINGS_CHECKSUM`) that would _only_ disable the "did you change the settings file" check when running unit tests in CI/CD and when developing locally. The original version used `DEBUG` as an escape valve for this, but that setting has other implications beyond checking the settings hash, so we don't necessarily want to enable it in CI/CD.

cneill (approved):
I'm not sure we need to announce the removal of this check in our release notes, but just wanted to note that we included it in the release notes [in the original PR](https://github.com/DefectDojo/django-DefectDojo/pull/10212/files#diff-4ab5f51f1f66da6d33fa2ddeaa768df6939f85dc4302b965c8da9571ba51ff9c)

mtesauro:
Sorry, I'm late to this party, have had some non-DefectDojo stuff keeping me from engaging at my normal level - it seems to me that we should do thing in phases  / iteratively.

1. We start with merging this PR as is to remove the check and see how things go
2. We give the removal till the Jan minor version release to see how thing go
3. Re-evaluate what if any next steps are needed.

The irony of having this PR have a merge conflict for the very thing we're wanting to remove isn't lost on me. :face_with_head_bandage: 

FWIW, I'm going to go ahead and approve this with the understanding that @Maffooch can sort out the merge conflict before it's actually merged.

mtesauro (approved):
Approved

grendel513 (approved):

valentijnscholten:
DD could help admins by displaying the effective settings/environment on a page. With some care around secrets / obfuscation. Or/additionally they could be displayed on start-up. Hopefully this will help admins before they come to #slack or @github.

Was it ever considered to only apply the check on official releases, or builds made from `master` and `bugfix` branches? Leaving out the SHA from the settings.dist.py could disable the check. The release or merge process could insert the hash for merges to or releases from master/bugfix.

PR #11300, opened 2024-11-20
SLA config: allow to use it on all levels

kiblik:
Original PR: #10025


Maffooch:
@kiblik Something I did not realize until taking another look at this new PR, there could be some tricky accommodations here when considering the possibility to update the SLA config on an engagement, while a product is running it's process of updating the SLA config. 

Without doing a deep dive on any other tripwires, the amount of complexity that would be needed to accommodate these extra SLA config levels make me question the value that they will provide.

kiblik:
> @kiblik Something I did not realize until taking another look at this new PR, there could be some tricky accommodations here when considering the possibility to update the SLA config on an engagement, while a product is running it's process of updating the SLA config.
> 
> Without doing a deep dive on any other tripwires, the amount of complexity that would be needed to accommodate these extra SLA config levels make me question the value that they will provide.

I planned to run the absolutely same validation + restriction as it is already implemented on `product` level: https://github.com/DefectDojo/django-DefectDojo/blob/6d847ab5394421bc207ed1a1f91867706bb69d3d/dojo/api_v2/serializers.py#L2019-L2026
With reference to `product.sla_configuration`. So updates will be possible to perform only on the "whole product" level. I didn't plan to implement this kind of lock for Engs or Tests. Yes, it might look like a waste of resources (change of `sla_config` on the `test` level will trigger recalculation for the whole product) but I wasn't able to find a more elegant solution for now.

mtesauro:
Sorry for the delayed response. 

I'm a bit concerned about the complexity this adds generally beyond what Cody raised :point_up: 

If we're documenting or explaining to people how DefectDojo works, we're going to have to say we have SLAs at Product, unless, they're at the Engagement, or unless it's at the test level. We'd also have to add SLA UI to both the Engagement and Test level (as well as API) to fully make this discoverable by DefectDojo users. We already have a rather cluttered UI with many UI elements that are used by a small portion of the community and it appears this would only add to that UI deficit. I'd rather spend cycles on updating the UI to a more modern toolkit/library TBH.

I'm also concerned about how a change of SLA at the test level will cause recalculation for the entire product. We're seeing people with 9700+ products doing 26K tests per day so for a group of our users the extra calculation could be debilitating.

I guess what I'm getting at is that the benefits of having SLA at Engagement & Test aren't worth the extra complexity and computation. 

Personally, I've only ever used SLAs at the Product level and the more I think about it, I begin to wonder if needing SLAs at something other then Product is more of a data modeling problem then a DefectDojo feature problem.  I'd rather have more products then have the cognitive load of remember what Engagement or Test had what SLA - that or having to look to the UI to tell me that.

Maybe I'm not describing it all that well here but in my gut this feels like a poor choice / trade-off between more features vs complexity.

Maffooch:
@kiblik closing this one for now

kiblik:
@mtesauro,
- I fully agree that UI is getting a bit overcomplicated over time. But I believe that with correct "hints"/"help notes", this might be acceptable. In API, there are just 2 additional fields to existing models. I do not consider this a radical increase in the complexity of readiness.
  - I also agree that this kind of functionality (as many others that are in DD) should be well documented and explain how, when, and why to use them. I can improve the documentation part. This should be done via proper examples. Some users might find themself in them. I was already trying to describe some of the scenarios here: https://github.com/DefectDojo/django-DefectDojo/pull/10025#issuecomment-2085385670 (others are lower)
- Regarding calculation:
  - If findings are added, it is necessary to calculate them even now. If the test or the eng is already created with the correct parameters at the beginning, it is not needed to recalculate the whole product.
  - If you choose the right SLA configuration during the creation of the test or the eng, there is no reason to reason to retrigger it.
  - If you still consider this a factor that radically decreases performance, we might define it as an opt-out (or even an opt-it) feature.
- I agree that shaping everybody's needs into one data-structure is challenging. It is the reason why I agree with the existence of Prod-Types or with freedom of how users use Engagements (they are trying to create an abstract layer helping to fit everybody's setup). However I still believe that there are reasons to use SLAs on lower levels.
  - In our case, we have different SLA policy for exposed endpoints and all others. It is because exposed endpoints are more critical and they are easier to compromise. So we run separated scans for exposed and separated scans for internal parts. But all of these parts are still part of one product.
  - Somebody else might have a similar rule for DAST (where you can discover the vulnerability of blackbox) vs SAST (where vulnerabilities might be much more critical but vulnerabilities discoverable by script kiddie on public endpoint have to be solved faster than findings from internal SAST scan). And all findings are still in one product.

Issue #11254, opened 2024-11-13
Graph 'Reported Severity by Month' says '0' instead of the dates. Also in official DefectDojo demo environment.


cgnl:
The 'Reported Finding Severity by Month' graph says 0 instead of the dates.
This happens in my local DefectDojo environment, but also on the demo environment on [https://demo.defectdojo.org/dashboard](https://demo.defectdojo.org/dashboard).
I also reported this on Slack.

**Steps to reproduce**
1. Go to (https://demo.defectdojo.org/dashboard)
2. View the graph 'Reported Finding Severity by Month'
3. It says '00:00:00' instead of the dates.

**Expected behavior**
A correctly displayed graph with the dates and the severity count.

**Deployment method** *(select with an `X`)*
- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Ubuntu 24.04.1 LTS
 - DefectDojo version (see footer) or commit message: 2.40.1

**Screenshots**
<img width="1469" alt="Scherm­afbeelding 2024-11-13 om 14 08 12" src="https://github.com/user-attachments/assets/65b5d338-2553-4a26-9445-744d58174e71">

cgnl:
Solved.

PR #11304, opened 2024-11-21
Update Reported Finding Severity by Month on the dashboard to be by month instead of day.

hblankenship:
[sc-8650]

Addresses #11254 

The graph on the dashboard for "Reported Finding Severity by Month" is currently displaying data in days, rather than months. And when there is only 1 data point (all findings are from the same day as it was in the demo case in the issue above) then the graph does not display correctly.

This changes the minimum data point to 1 and specifies that the data is by month. This should display accurate, if not visually appealing, data when there is only 1 month's worth of data. As time progresses, the chart will look better and will display up to 6 months of Reported Finding Severity by Month.


mtesauro (approved):
Approved

Maffooch (approved):

cneill (approved):

PR #11306, opened 2024-11-21
Disallow multiple single-use notes on a single object 

hblankenship:
[sc-6050]

Returns a BAD_REQUEST status and an error message when attempting to POST a note with a Note Type of 'is_single' - if another Note with that Note Type already exists on the given object (finding, engagement, test).


mtesauro (approved):
Approved

Maffooch (approved):

cneill (approved):

PR #11313, opened 2024-11-22
Add a filter for Findings for Has Any JIRA (grouped or single)

hblankenship:
[sc-4131]

Added a filter for Has Any JIRA which will filter the findings on having group jira or jira_issue. This will also allow a dashboard tile to exist that shows all findings with any type of JIRA issue.

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -235,6 +235,39 @@ class FindingSLAFilter(ChoiceFilter):
         return self.options[value][1](self, qs, self.field_name)
 
 
+class FindingHasJIRAFilter(ChoiceFilter):
+    def any(self, qs, name):
+        return qs.filter(Q(jira_issue=None) | Q(finding_group__jira_issue=None))
+
+    def has_jira_not_grouped(self, qs, name):
+        return qs.filter(Q(jira_issue=True) & Q(finding_group__jira_issue=False))
+
+    def has_jira_only_grouped(self, qs, name):
+        return qs.filter(Q(jira_issue=False) | Q(finding_group__jira_issue=True))
+
+    def has_any_jira(self, qs, name):
+        return qs.filter(Q(jira_issue=True) | Q(finding_group__jira_issue=True))
+
+    options = {
+        None: (_("Has No JIRA"), any),
+        0: (_("Has JIRA Non-grouped"), has_jira_not_grouped),
+        1: (_("Has JIRA Grouped Only"), has_jira_only_grouped),
+        2: (_("Has Any JIRA"), has_any_jira),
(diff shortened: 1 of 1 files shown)


dojo/filters.py
```diff
@@ -1610,6 +1640,7 @@ class FindingFilterHelper(FilterSet):
                 lookup_expr="isnull",
                 exclude=True,
                 label="Has Group JIRA")
+            has_any_jira = FindingHasJIRAFilter(label="Has Any JIRA")
```
Maffooch, line 1643:
```suggestion
        has_any_jira = FindingHasJIRAFilter(label="Has Any JIRA")
```
This should be present if groups are enabled or not, since regular jira links are a part of this

Maffooch (changes requested):

Maffooch (approved):

dojo/filters.py
```diff
@@ -235,6 +235,35 @@ def filter(self, qs, value):
...
 
+class FindingHasJIRAFilter(ChoiceFilter):
+    def no_jira(self, qs, name):
+        return qs.filter(Q(jira_issue=None) & Q(finding_group__jira_issue=None))
+
+    def any_jira(self, qs, name):
+        return qs.filter(~Q(jira_issue=None) | ~Q(finding_group__jira_issue=None))
+
+    def all_items(self, qs, name):
+        return qs.filter(Q(jira_issue=None) | Q(finding_group__jira_issue=None))
```
cneill, line 246: This will only return Findings that are either missing `jira_issue` or missing `finding_group__jira_issue`. Shouldn't it simply return everything, unfiltered?

Maffooch: This function is only called when the input supplied does not match what is expected, and low key sorta fails. I believe the thought process is along the lines of "if something unexpected happens, removing all findings with jira links would be a good indicator. This is just me speculating though. Would need @hblankenship to confirm

On the other side, returning all results (as if the filter was never applied) is a totally valid approach as well

mtesauro (approved):
Approved

Under the assumption that the pending comment from @cneill is resolved

blakeaowens (approved):

dojo/filters.py
hblankenship: I apologize for not getting to this earlier. Testing shows that the option returns the same thing not applying the filter returns. The reason it returns them all, despite what it looks like, is because jira_issue is None for finding_group__jira_issue findings and finding_group__jira_issue is None for jira_issue items.  I have modified it to just return without filtering to alleviate the confusion.

cneill (approved):

PR #11331, opened 2024-11-26
Ruff: Enable and fix RUF010

kiblik:
Remove "RUF010" from ignored and fix it.
https://docs.astral.sh/ruff/rules/explicit-f-string-type-conversion/

I see no reason to use `str(...)` in combination with `f"{...}`.

Ruff recommended `f{...!s}` but I believe this is easier and sufficient.


Maffooch (approved):

mtesauro (approved):
Approved

hblankenship (approved):

grendel513 (approved):

PR #11364, opened 2024-12-03
feat(GHA): Add SHA pinning

kiblik:
Official documentation recommends SHA pinning: https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

Also, renovate [supports this syntax](https://docs.renovatebot.com/modules/manager/github-actions/).

### DIFF
--- a/.github/workflows/build-docker-images-for-testing.yml
+++ b/.github/workflows/build-docker-images-for-testing.yml
@@ -19,7 +19,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           persist-credentials: false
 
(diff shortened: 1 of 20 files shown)


mtesauro:
This makes total sense to me :+1:

mtesauro (approved):
Approved

blakeaowens (approved):

grendel513 (approved):

cneill (approved):
From Renovate's documentation, it's not clear to me if it will automatically bump e.g. `v4.4.3` to `v4.4.4` or if it stays pinned to that specific tag forever, only updating the SHA underpinning that tag changes. If it's the latter, it might make sense to pin to e.g. `# v4` instead of `# v4.4.3`.

Since this is definitely an improvement either way, I'm okay to approve and we can keep an eye on it to see how it behaves.

kiblik:
@cneill, it looks like versions are not locked https://github.com/DefectDojo/django-DefectDojo/pull/11410/
So we are good :)

cneill:
> @cneill, it looks like versions are not locked #11410 So we are good :)

Excellent!

Issue #6748, opened 2022-08-22
Your credentials aren't allowed for Google OAuth Authentication


franzramadhan:
**Slack us first!**
I have raised this issue in the slack - https://owasp.slack.com/archives/C2P5BA8MN/p1661154750234779

**Be informative**
I am trying to setup OAuth2 authentication using Google by following [this wiki](https://defectdojo.github.io/django-DefectDojo/integrations/social-authentication/#google)
But I got `AuthForbidden at /complete/google-oauth2/` error after clicking the Login with Google button.
Following environment variables have been set inside the uwsgi container.

```
DD_SOCIAL_AUTH_GOOGLE_OAUTH2_SECRET=G<redacted>
DD_SOCIAL_AUTH_GOOGLE_OAUTH2_KEY=72<redacted>-<redacted>.apps.googleusercontent.com
DD_SOCIAL_AUTH_GOOGLE_OAUTH2_ENABLED=true
```

<img width="717" alt="Screen Shot 2022-08-22 at 20 16 41" src="https://user-images.githubusercontent.com/1482172/185930555-52f3cc2b-7f97-420a-a0a7-25441e48a5e9.png">

**Bug description**
Traceback

```
Environment:


Request Method: GET
Request URL: https://<defectdojo-public-url>/complete/google-oauth2/?state=zcdf7miQNPvWCuIR6HVSz1BzYLSNo7Xe&code=4%2F0AdQt8qh4uuehZ82fvSdcfs2dWRS-GuZr5TsN-TFVOh0bdUhChlFazlkSDeNsKWaD_2BhHQ&scope=email+profile+openid+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fuserinfo.email+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fuserinfo.profile&authuser=0&hd=<mydomain.com>&prompt=none

Django Version: 3.2.14
Python Version: 3.8.13
Installed Applications:
('django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.sites',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'polymorphic',
 'django.contrib.admin',
 'django.contrib.humanize',
 'gunicorn',
 'auditlog',
 'dojo',
 'watson',
 'tagging',
 'imagekit',
 'multiselectfield',
 'rest_framework',
 'rest_framework.authtoken',
 'dbbackup',
[...]

PR #11372, opened 2024-12-04
fix(oauth2):  google oauth2 whitelisting.

JGodin-C2C:
# Fix google oauth2 whitelisting.

The variable is a tuple and should be managed as such

**Description**

Fixes: #6748 

***What***

Google oauth using environment variables as configuration (k8s for instance) could not get the correct configuration.
The google whitelisting is an array and was red as a string. 
It was hence not working the way it was intended to. 

***How***

I used the same method used as the allowed_hosts to pars arrays and updated the documentation to explain how to use the configuration keys in a helm deployments. 
  

**Checklist**


blakeaowens (approved):

mtesauro (approved):
Approved

Maffooch (approved):

cneill (approved):

PR #11432, opened 2024-12-17
summarizing multiple similar findings into problems

LeoOMaia:
**Description**
- Feature:
We created a "Problems" tab that disambiguates similar findings based on the script_id that detected them. This allows us to consolidate most findings into a single problem, enabling the vulnerability analyst to more accurately identify all types of issues without duplication.
- Bug fix implemented by this PR:
A bug in the OpenVAS XML parser has been fixed, where it was not correctly identifying the script_id and was always returning None.

**Test results**

We tested the creation and association of findings to a specific problem according to the JSON file we provided, which already identifies similar findings by grouping them based on their script_id. We also tested that after creating the findings and problems, deleting a finding would update the problem by reducing the number of associated findings, and if all findings related to a problem were deleted, the problem would be automatically removed. Additionally, we verified the logic where if all findings become inactive, the problem status changes from open to closed, and if at least one finding remains active, the problem stays open.


dojo/problem/helper.py
```diff
@@ -0,0 +1,137 @@
...
+import requests
+from requests.packages.urllib3.exceptions import InsecureRequestWarning
+
+from dojo.models import Problem, Finding
+
+import logging
+logger = logging.getLogger(__name__)
+
+CONFIG_FILE = os.path.join(os.path.dirname(__file__), 'config.json')
+CACHED_JSON_FILE = os.path.join('/app/media', 'cached_disambiguator.json')
```
cunha, line 12: Is `/app/media` always available? This seems like something that would exist in a container, but not on a classic install. Maybe there is a configuration variable we can use to know where to store the cached file.

dojo/problem/helper.py
```diff
@@ -0,0 +1,137 @@
...
+
+def save_json_to_cache(data):
+    logger.info('Saving disambiguator JSON to cache')
+    with open(CACHED_JSON_FILE, 'w') as f:
+        json.dump(data, f)
+
+def load_json():
+    try:
+        # Disable SSL warnings
+        requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
```
cunha, line 60: SSL warnings should be enabled by default, and ignoring SSL errors should require explicit configuration by the user.

dojo/problem/helper.py
```diff
@@ -0,0 +1,137 @@
...
+                return data
+
+        return {}
+
+    except (requests.RequestException, ValueError, json.JSONDecodeError) as e:
+        logger.error('Error loading disambiguator JSON: %s', e)
+        return {}
+
+def extract_script_id(full_id):
+    parts = full_id.split('____')
```
cunha, line 83: Why do we need this split? This is apparently called on the data in the mapping file, which should have the data in a clear structure such that we can avoid this type of post-processing.

dojo/problem/helper.py
```diff
@@ -0,0 +1,137 @@
...
+    return parts[0] if len(parts) == 2 else None
+
+def find_or_create_problem(finding):
+    data = load_json()
```
cunha, line 87: This is apparently loading the JSON file every time the function is called, which seems to happen for each findings. To make things more efficient, we should load the data once, and then just access it for each finding.

dojo/problem/helper.py
```diff
@@ -0,0 +1,137 @@
...
+    script_id = finding.vuln_id_from_tool
+
+    valid_ids_mapping = {
+        key: [extract_script_id(full_id) for full_id in script_ids if extract_script_id(full_id)]
+        for key, script_ids in data.items()
+    }
+
+    for key, valid_ids in valid_ids_mapping.items():
```
cunha, line 95: This performs a O(N) search; the mappings can be stored in a dict mapping script-id to problem-id, which would allow for O(1) access.

dojo/problem/helper.py
```diff
@@ -0,0 +1,137 @@
...
+            problem = _get_or_update_problem(valid_ids, finding, script_id)
+            if problem:
+                return problem
+
+    # if the script_id is not in the mapping, create a new one
+    return _get_or_create_problem_by_script_id(script_id, finding)
+
+def _get_or_update_problem(valid_ids, finding, script_id):
+    for valid_id in valid_ids:
+        related_finding = Finding.objects.filter(vuln_id_from_tool=valid_id).first()
```
cunha, line 106: This is performing a nested O(N*M) search; it seems like we should just do `Problem.objects.filter(problem_id=problem_id)` to get the problem directly from the database (efficiently given an index exists on the `problem_id` field). The `problem_id` can likely be the `key` in the previous function.

dojo/problem/tasks.py
```diff
@@ -0,0 +1,37 @@
...
+from dojo.decorators import dojo_async_task
+from dojo.problem.helper import CONFIG_FILE, validate_json, download_json, save_json_to_cache
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+@dojo_async_task
+@app.task
+def daily_cache_update(**kwargs):
```
cunha, line 16: It's worth it trying to reuse the `load_json()` function here to avoid code duplication.

dojo/settings/settings.dist.py
```diff
@@ -1109,7 +1109,7 @@ def saml2_attrib_map_format(dict):
 if len(env("DD_CELERY_BROKER_TRANSPORT_OPTIONS")) > 0:
     CELERY_BROKER_TRANSPORT_OPTIONS = json.loads(env("DD_CELERY_BROKER_TRANSPORT_OPTIONS"))
 
-CELERY_IMPORTS = ("dojo.tools.tool_issue_updater", )
+CELERY_IMPORTS = ("dojo.tools.tool_issue_updater", "dojo.problem.tasks")
```
cunha, line 1112: Maybe rename the file to something like `dojo.problem.update_mappings` to make what it does clearer.

dojo/tools/nmap/parser.py
```diff
@@ -126,11 +128,15 @@ def get_findings(self, file, test):
                         severity=severity,
                         mitigation="N/A",
                         impact="No impact provided",
+                        vuln_id_from_tool=script_id,
```
cunha, line 131: This could be in a separate request, as it's a clear improvement over the existing code that doesn't depend on the addition of the `Problem` functionality.

dojo/tools/nmap/parser.py
```diff
@@ -96,6 +97,7 @@ def get_findings(self, file, test):
                             "**Extra Info:** {}\n".format(port_element.find("service").attrib["extrainfo"])
                         )
                     description += service_info
+                script_id = "NoID"
```
cunha, line 100: Why is this preferrable to just setting `vuln_id_from_tool = None` (the default) if there is no ID?

dojo/tools/openvas/xml_parser.py
```diff
@@ -27,7 +29,8 @@ def get_findings(self, filename, test):
                     title = title + "_" + finding.text
                     description.append(f"**Port**: {finding.text}")
                 if finding.tag == "nvt":
-                    description.append(f"**NVT**: {finding.text}")
+                    script_id = finding.get("oid")
```
cunha, line 32: Could be in a separate PR too.

dojo/tools/openvas/xml_parser.py
```diff
@@ -38,11 +41,16 @@ def get_findings(self, filename, test):
 
             finding = Finding(
                 title=str(title),
+                test=test,
```
cunha, line 44: Why do we need this?

cunha (changes requested):
The most important comments are changing some of the algorithms to have better asymptotic performance.

dojo/tools/openvas/xml_parser.py
LeoOMaia: when we try to use finding.save(), always appear `An exception error occurred during the report import: Finding has no test.`, seeing others parser, all of them, used this `test` and helped to this problem

dojo/tools/nmap/parser.py
LeoOMaia: I created a new PR #11454

dojo/tools/nmap/parser.py
LeoOMaia: You are right

dojo/problem/helper.py
```diff
@@ -0,0 +1,137 @@
...
+    'Low': 2,
+    'Info': 1
+}
+
+def load_config():
+    with open(CONFIG_FILE, 'r') as f:
+        return json.load(f)
+
+def validate_json(data):
+    if not isinstance(data, dict):
```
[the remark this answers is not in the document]
LeoOMaia: I checked and nothing of Pydantic or jsochema or other similar things. But have a lot of `isinstance`.

dojo/problem/helper.py
LeoOMaia: I chose to use `/app/media` because this path is defined in `docker-compose` for the `uwsgi` and `celeryworker` containers using the following volume configuration:
```
volumes:
        - type: bind
          source: ./docker/extra_settings
          target: /app/docker/extra_settings
        - "defectdojo_media:${DD_MEDIA_ROOT:-/app/media}"
```
but really have a problem to choose where is the json if is defined DD_MEDIA_ROOT. I think in change to use:
```
MEDIA_ROOT = os.getenv('DD_MEDIA_ROOT', '/app/media')
CACHED_JSON_FILE = os.path.join(MEDIA_ROOT, 'cached_disambiguator.json')
```

dojo/tools/openvas/xml_parser.py
cunha: Curious as this would break importing data; but in any case, it might also be best submitted in a separate PR as a bugfix.

dojo/settings/settings.dist.py
```diff
@@ -1744,6 +1748,15 @@ def saml2_attrib_map_format(dict):
...
 SILENCED_SYSTEM_CHECKS = ["django_jsonfield_backport.W001"]
 
+# Problem utilizes mappings from a JSON file to disambiguate Findings.
+# The JSON file should have the following structure:
+# {
+#     "problem_id_1": ["script_id_1", "script_id_2", "script_id_3"],
+#     "problem_id_2": ["script_id_4", "script_id_5"],
+#     "problem_id_3": ["script_id_6"]
+# }
+PROBLEM_MAPPINGS_JSON_URL = "https://homepages.dcc.ufmg.br/~leonardooliveira/disambiguator.json"
```
cunha, line 1758: Maybe move this to the comment, explain it is a mapping that aggregates equivalent/similar `Findings` from Nmap, OpenVAS, Metasploit and Nuclei. Have an accompanying `README.md` file explaining how the mapping is built. Set the default configuration to `None` to make it opt-in.

Also, if this variable is set to `None`, it would be best to simply hide the "Problems" button in Dojo's left toolbar. (So it doesn't cause any changes to deployments unless the mapping is configured.)

dojo/templatetags/get_different_total_script_id.py
```diff
@@ -0,0 +1,9 @@
+from dojo.models import Finding
+
+from django import template
+
+register = template.Library()
+
+@register.filter
+def count_unique_vulns(findings):
```
cunha, line 8: `...unique_vulns` seems confusing, as it is counting the number of distinct script IDs. Maybe call it `count_distinct_script_ids()`.

dojo/problem/helper.py
```diff
@@ -0,0 +1,118 @@
...
+    script_to_problem_mapping = {
+        script_id: key
+        for key, script_ids in mappings_json_findings.items()
+        for script_id in script_ids
+    }
+    return script_to_problem_mapping
+
+def load_json(check_cash=True):
+    try:
+        if check_cash:
```
cunha, line 64: "cash" means "money", consider renaming to `check_cache` to avoid confusion.

dojo/problem/helper.py
```diff
@@ -0,0 +1,118 @@
...
+
+def load_cached_json():
+    try:
+        if os.path.exists(CACHED_JSON_FILE):
+            with open(CACHED_JSON_FILE, 'r') as f:
+                data = json.load(f)
+                if validate_json(data):
+                    return data
+    except (ValueError, json.JSONDecodeError):
+        pass
```
cunha, line 46: We should log something here to notify the user of the error (assuming Dojo has some logging going on).

dojo/problem/helper.py
```diff
@@ -0,0 +1,118 @@
...
+    for key, value in data.items():
+        if not isinstance(key, str) or not isinstance(value, list):
+            return False
+        if not all(isinstance(item, str) for item in value):
+            return False
+    return True
+
+def download_json(json_url):
+    response = requests.get(json_url, timeout=5, verify=False)
+    response.raise_for_status()
```
cunha, line 35: We should check that the URL `is not None` before attempting to download. If it `is not None` and we get an error, we should log the error and **not** crash the server; we should just continue using an empty mapping.

cunha (changes requested):
Some additional suggestions in the comments.

dojo/settings/settings.dist.py
```diff
@@ -1744,6 +1748,11 @@ def saml2_attrib_map_format(dict):
 # see https://github.com/laymonage/django-jsonfield-backport
 SILENCED_SYSTEM_CHECKS = ["django_jsonfield_backport.W001"]
 
+# By default, this mapping is not configured (set to None). If configured, it allows
+# the "Problems" button to appear in Dojo's left toolbar.
+# You can check more information in https://homepages.dcc.ufmg.br/~leonardooliveira/defectdojo/README.md
+PROBLEM_MAPPINGS_JSON_URL = None
```
cunha, line 1754:
```suggestion
# A finding-to-problem mapping covering Nmap, OpenVAS and Nuclei is available in
# <link to JSON file>
PROBLEM_MAPPINGS_JSON_URL = None
```

dojo/templates/base.html
```diff
@@ -3,6 +3,7 @@
 {% load authorization_tags %}
 {% load i18n %}
 {% load static %}
+{% load problem_tags %}
```
cunha, line 6: How about we move the `check_problem_enabled` tag into one of the existing files? Maybe `display_tags` or `navigation_tags`?

cunha (approved):
Minor comments inside.

dojo/templatetags/navigation_tags.py
```diff
@@ -137,3 +138,7 @@ def create_page_nav(page_idx):
 @register.filter
 def can_add_product(user):
     return get_authorized_product_types(Permissions.Product_Type_Add_Product).count() > 0
+
+@register.filter
+def check_problems_enabled(user):
+    return settings.PROBLEM_MAPPINGS_JSON_URL
```
cunha, line 144:
```suggestion
    return settings.PROBLEM_MAPPINGS_JSON_URL is not None
```

dojo/templates/base.html
```diff
@@ -325,6 +325,33 @@
                                             </ul>
                                             <!-- /.nav-second-level -->
                                         </li>
+                                        {% if request.user|check_problems_enabled %}
```
cunha, line 328: If possible, we should drop the `request.user` parameter, likely by changing `check_problems_enabled` to be a simple tag instead of a `filter`.

dojo/settings/settings.dist.py
```diff
@@ -1744,6 +1748,13 @@ def saml2_attrib_map_format(dict):
 # see https://github.com/laymonage/django-jsonfield-backport
 SILENCED_SYSTEM_CHECKS = ["django_jsonfield_backport.W001"]
 
+# By default, this mapping is not configured (set to None). If configured, it allows
+# the "Problems" button to appear in Dojo's left toolbar.
+# You can check more information in https://homepages.dcc.ufmg.br/~leonardooliveira/defectdojo/README.md
+# A finding-to-problem mapping covering Nmap, OpenVAS and Nuclei is available in
+# <link to JSON file>
```
cunha, line 1755: I actually meant that we should have a link to the existing JSON file here so people can copy-paste it in if they want.

cunha (approved):
Minor comments inside.

dojo/templates/base.html
LeoOMaia: It seems that the `simple_tag` cannot return a variable that you can use with if, and the filter only accepts something that can be used if a parameter exists. If it doesn't pass, it's always interpreted as if the function doesn't exist.
I'm doing something similar to the logic in the `can_add_product` function in `navigation_tags`. The function is receiving the parameter but not using it, and it's still working.

cunha (approved):

Maffooch:
Hi @LeoOMaia this is a neat idea, but we are unable to take this contribution due to the added complexity of the finding model. In the future, please create an issue in GitHub, or make a post in slack about new features before writing any code.

PR #11446, opened 2024-12-20
Add settings

FirePanda169:
**Description**

Add settings:
* DD_DEFAULT_SCM_TYPE - default value for scm_type
* DD_JIRA_CONNECT_METHOD - import path for jira connect method

DD_DEFAULT_SCM_TYPE makes it a little easier to work with scm-type, since most people use only one scm-type

**Test results**
...

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -3197,7 +3197,7 @@ class Finding(models.Model):
                 st = dojo_meta.value.strip()
                 if st:
                     return st.lower()
-        return ""
+        return settings.DEFAULT_SCM_TYPE
 
     def scm_public_prepare_base_link(self, uri):
         # scm public (https://scm-domain.org) url template for browse is:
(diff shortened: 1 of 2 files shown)


Maffooch:
HI @FirePanda169 I am curious about how much we would gain from short circuiting the SCM search times. If the performance gain is negligible, I am unsure if these settings will be more helpful than they would be confusing

FirePanda169:
I'll check what the difference in response time will be

FirePanda169:
Hi @Maffooch I checked the response time on loading 10 findings. The time is almost the same. But I don't have records in the DojoMeta model. I don't have a chance to check more thoroughly yet.

Maffooch:
@FirePanda169 thank you for doing the quick validation! Once you test with some DojoMeta that would trigger the code paths that these new settings are influencing, let me know what the performance savings are

Maffooch:
Hi @FirePanda169 thank you for indulging my curiosities on the performance savings from this PR. I think where the PR stands right now, we are good close

PR #11447, opened 2024-12-20
Allow None Option for Active/Verified on Import/Reimport to Mirror UI Options

### LINKED ISSUE #11235: When importing a Github scan via API the Status of the finding is wrong
**Bug description**
When importing a scan from Github dependabot or Github code scanning via DefectDojo API findings  that have the state **fixed**, **closed** or **dismissed** are imported as **Active** when it should be imported as **MITIGATED**, however when doing it through the UI it works fine (a **fixed** finding appears as **MITIGATED**)

**Steps to reproduce**
1. Get a github  dependabot or code-scanning json report
2. Import it via API using the **import-scan** endpoint
3. The finding that is in state **fixed** appears as **ACTIVE** on DefectDojo

**Expected behavior**
A finding that has the state **FIXED**, **closed** or **dismissed** should be imported as **MITIGATED**

- [x ] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Ubuntu 20.04
 - DefectDojo version v. 2.38.2

**Sample scan files**
[GHAS-example.json](https://github.com/user-attachments/files/17700028/GHAS-example.json)

hblankenship:
The help on the Active/Verified options for the Import and Reimport Scan APIs is misleading. The help says 'Override the active setting from the tool' and 'Override the verified setting from the tool' which suggests that True means 'Yes, do that', and False would mean 'No, don't override the tool'. However, viewed that way, there is no way to set the Active/Verified, just a way to say 'override' it with "something". The UI allows for a third option: "None" which means do NOT override the tool instead of Override the Tool with This Setting. This PR allows the API to also have a 'None' option and updates the help text to match what is given in the UI.

fixes #11235 

[sc-9259]

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -2087,10 +2087,10 @@ class CommonImportScanSerializer(serializers.Serializer):
         help_text="Minimum severity level to be imported",
     )
     active = serializers.BooleanField(
-        help_text="Override the active setting from the tool.", required=False,
+        help_text="Force findings to be active/inactive or default to the original tool (None)", required=False,
     )
     verified = serializers.BooleanField(
-        help_text="Override the verified setting from the tool.", required=False,
+        help_text="Force findings to be verified/not verified or default to the original tool (None)", required=False,
     )
 
     # TODO: why do we allow only existing endpoints?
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

cneill (approved):

mtesauro (approved):
Approved

grendel513 (approved):

PR #11454, opened 2024-12-22
Fixing openvas parser and including script_id for openvas and nmap

LeoOMaia:
i'm including the script_id from tool to consider in the parser of `nmap` and `openvas` fixing a problem of get the id from openvas parser who get the id by the param of the tag.

### DIFF
--- a/dojo/tools/openvas/xml_parser.py
+++ b/dojo/tools/openvas/xml_parser.py
@@ -16,6 +16,7 @@ class OpenVASXMLParser:
         report = root.find("report")
         results = report.find("results")
         for result in results:
+            script_id = None
             for finding in result:
                 if finding.tag == "name":
                     title = finding.text
@@ -27,7 +28,8 @@ class OpenVASXMLParser:
                     title = title + "_" + finding.text
                     description.append(f"**Port**: {finding.text}")
                 if finding.tag == "nvt":
-                    description.append(f"**NVT**: {finding.text}")
+                    script_id = finding.get("oid")
+                    description.append(f"**NVT**: {script_id}")
                 if finding.tag == "severity":
                     severity = self.convert_cvss_score(finding.text)
                     description.append(f"**Severity**: {finding.text}")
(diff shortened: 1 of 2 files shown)


dojo/tools/openvas/xml_parser.py
```diff
@@ -27,7 +28,8 @@ def get_findings(self, filename, test):
                     title = title + "_" + finding.text
                     description.append(f"**Port**: {finding.text}")
                 if finding.tag == "nvt":
-                    description.append(f"**NVT**: {finding.text}")
+                    script_id = finding.get("oid")
+                    description.append(f"**NVT**: {script_id}")
```
cunha, line 32: This will prevent us from showing `finding.text`. If there is any information in the tag's content (`finding.text`), then we could instead print a linebreak and the content (in addition to the `script_id`) to maintain backwards compatibility.

dojo/tools/openvas/xml_parser.py
```diff
@@ -38,10 +40,12 @@ def get_findings(self, filename, test):
 
             finding = Finding(
                 title=str(title),
+                test=test,
```
cunha, line 43: Seems like a bugfix, likely best in a separate PR.

cunha (changes requested):
Looks good. Suggestions:

* Check whether we should print `finding.text` when `tag == "nvt"`
* Move the `test=test` bugfix to a separate PR

dojo/tools/openvas/xml_parser.py
LeoOMaia: here #11464

cunha (approved):

dojo/tools/openvas/xml_parser.py
```diff
@@ -27,7 +28,9 @@ def get_findings(self, filename, test):
                     title = title + "_" + finding.text
                     description.append(f"**Port**: {finding.text}")
                 if finding.tag == "nvt":
-                    description.append(f"**NVT**: {finding.text}")
+                    script_id = finding.get("oid") or finding.text
```
Maffooch, line 31: This may have some implication for hash code generation since the description is used by default. Please remove this section, and keep the `script_id` to the `vuln_id_from_tool` field

Maffooch (changes requested):

dogboat (approved):

mtesauro (approved):
Approved

grendel513 (approved):

Maffooch (approved):

PR #11464, opened 2024-12-23
including test for Finding in xml parser

LeoOMaia:
We are facing difficulties saving a finding to later perform the disambiguation of findings into problems, which will summarize a group of findings into a single problem. There are already several parsers from other tools that pass the variable test as a parameter to the Finding class. I believe there should be no issues with doing something like this.
More information here #11432

### DIFF
--- a/dojo/tools/openvas/xml_parser.py
+++ b/dojo/tools/openvas/xml_parser.py
@@ -38,6 +38,7 @@ class OpenVASXMLParser:
 
             finding = Finding(
                 title=str(title),
+                test=test,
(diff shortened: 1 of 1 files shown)


cunha (approved):

cunha:
Do write some explanation in the PR opening comment so project maintainers have some context about the change.

Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

dogboat (approved):

PR #11472, opened 2024-12-26
Mitigated On/Before/After now use DateTimeFilter

hblankenship:
Fixes filtering for  mitigation for the on/before/after fields

[sc-7759]

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -21,6 +21,7 @@ from django_filters import (
     CharFilter,
     DateFilter,
     DateFromToRangeFilter,
+    DateTimeFilter,
     FilterSet,
     ModelChoiceFilter,
     ModelMultipleChoiceFilter,
@@ -1456,9 +1457,9 @@ class ApiFindingFilter(DojoFilter):
     jira_change = DateRangeFilter(field_name="jira_issue__jira_change")
     last_reviewed = DateRangeFilter()
     mitigated = DateRangeFilter()
-    mitigated_on = DateFilter(field_name="mitigated", lookup_expr="exact")
-    mitigated_before = DateFilter(field_name="mitigated", lookup_expr="lt")
-    mitigated_after = DateFilter(field_name="mitigated", lookup_expr="gt")
+    mitigated_on = DateTimeFilter(field_name="mitigated", lookup_expr="exact")
+    mitigated_before = DateTimeFilter(field_name="mitigated", lookup_expr="lt")
+    mitigated_after = DateTimeFilter(field_name="mitigated", lookup_expr="gt")
     # NumberInFilter
     cwe = NumberInFilter(field_name="cwe", lookup_expr="in")
     defect_review_requested_by = NumberInFilter(field_name="defect_review_requested_by", lookup_expr="in")
@@ -1587,9 +1588,9 @@ class FindingFilterHelper(FilterSet):
     duplicate = ReportBooleanFilter()
     is_mitigated = ReportBooleanFilter()
     mitigated = DateRangeFilter(field_name="mitigated", label="Mitigated Date")
-    mitigated_on = DateFilter(field_name="mitigated", lookup_expr="exact", label="Mitigated On")
-    mitigated_before = DateFilter(field_name="mitigated", lookup_expr="lt", label="Mitigated Before")
-    mitigated_after = DateFilter(field_name="mitigated", lookup_expr="gt", label="Mitigated After")
+    mitigated_on = DateTimeFilter(field_name="mitigated", lookup_expr="exact", label="Mitigated On")
+    mitigated_before = DateTimeFilter(field_name="mitigated", lookup_expr="lt", label="Mitigated Before")
+    mitigated_after = DateTimeFilter(field_name="mitigated", lookup_expr="gt", label="Mitigated After")
     planned_remediation_date = DateRangeOmniFilter()
     planned_remediation_version = CharFilter(lookup_expr="icontains", label=_("Planned remediation version"))
     file_path = CharFilter(lookup_expr="icontains")
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -1550,7 +1550,7 @@ class ApiFindingFilter(DojoFilter):
             value = value.replace(hour=23, minute=59, second=59)
 
         return queryset.filter(mitigated__gt=value)
-    
+
     def filter_mitigated_on(self, queryset, name, value):
         if value.hour == 0 and value.minute == 0 and value.second == 0:
             # we have a simple date without a time, lets get a range from this morning to tonight at 23:59:59:999
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -1459,7 +1459,7 @@ class ApiFindingFilter(DojoFilter):
     mitigated = DateRangeFilter()
     mitigated_on = DateTimeFilter(field_name="mitigated", lookup_expr="exact")
     mitigated_before = DateTimeFilter(field_name="mitigated", lookup_expr="lt")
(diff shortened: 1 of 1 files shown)


dojo/filters.py
```diff
@@ -1456,9 +1457,9 @@ class ApiFindingFilter(DojoFilter):
     jira_change = DateRangeFilter(field_name="jira_issue__jira_change")
     last_reviewed = DateRangeFilter()
     mitigated = DateRangeFilter()
-    mitigated_on = DateFilter(field_name="mitigated", lookup_expr="exact")
-    mitigated_before = DateFilter(field_name="mitigated", lookup_expr="lt")
-    mitigated_after = DateFilter(field_name="mitigated", lookup_expr="gt")
+    mitigated_on = DateTimeFilter(field_name="mitigated", lookup_expr="exact")
+    mitigated_before = DateTimeFilter(field_name="mitigated", lookup_expr="lt")
+    mitigated_after = DateTimeFilter(field_name="mitigated", lookup_expr="gt")
```
Maffooch, line 1462: If a user were to be reliant on passing a date to this filter, would the API accept the absence of the timestamp? For example, if the query was `mitigated_on=2025-01-02` would a validation error be thrown?

hblankenship: A validation error does not get thrown if you use 2025-01-02 without the timestamp. Should it?

Maffooch: No, a validation error should not be raised. I was asking to make sure that this change would not cause any issues for users that were currently using dates

Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

cneill (changes requested):
This doesn't seem to be operating how I would expect... I spun up a test instance, created some findings, and marked them mitigated. The UI uses a date-picker widget that won't allow you to enter a timestamp, so I couldn't test that.

With the API, I have 3 mitigated findings with these `mitigated` values:

```
❯ curl -H "Authorization: Token ${DD_API_KEY}"  "http://localhost:8080/api/v2/findings/" | jq '.results[] | select(.mitigated != null) | .mitigated'

"2025-01-13T18:08:05.911321Z"
"2025-01-13T18:08:05.922679Z"
"2025-01-13T18:08:05.930498Z"
```

Testing `mitigated_on` with a simple date:
```
❯ curl -H "Authorization: Token ${DD_API_KEY}"  "http://localhost:8080/api/v2/findings/?mitigated_on=2025-01-13"
{"count":0,"next":null,"previous":null,"results":[],"prefetch":{}}%
```

It wants an exact timestamp:
```
❯ curl -H "Authorization: Token ${DD_API_KEY}"  "http://localhost:8080/api/v2/findings/?mitigated_on=2025-01-13T18:08:05.911321Z"
{"count":1,"next":null,"previous":null,"results":[{"id":1, ... }]}
```

The `mitigated_after` and `mitigated_before` filters *sort of* work as expected with simple dates, but I assume it is just plugging in a `00:00:00` for the time, so for example this returns my 3 findings that were mitigated today, 2025-01-13, where  `mitigated_after` would only have included things *after* today (23:59:59) in the past.

```
❯ curl -H "Authorization: Token ${DD_API_KEY}"  "http://localhost:8080/api/v2/findings/?mitigated_after=2025-01-13"
{"count":3,"next":null,"previous":null,"results":[{"id":1, ... }]}
```

I think this needs some additional tweaking, and could really use some unit tests.

hblankenship:
> This doesn't seem to be operating how I would expect... I spun up a test instance, created some findings, and marked them mitigated. The UI uses a date-picker widget that won't allow you to enter a timestamp, so I couldn't test that.
> 
> With the API, I have 3 mitigated findings with these `mitigated` values:
> 
> ```
> ❯ curl -H "Authorization: Token ${DD_API_KEY}"  "http://localhost:8080/api/v2/findings/" | jq '.results[] | select(.mitigated != null) | .mitigated'
> 
> "2025-01-13T18:08:05.911321Z"
> "2025-01-13T18:08:05.922679Z"
> "2025-01-13T18:08:05.930498Z"
> ```
> 
> Testing `mitigated_on` with a simple date:
> 
> ```
> ❯ curl -H "Authorization: Token ${DD_API_KEY}"  "http://localhost:8080/api/v2/findings/?mitigated_on=2025-01-13"
> {"count":0,"next":null,"previous":null,"results":[],"prefetch":{}}%
> ```
> 
> It wants an exact timestamp:
> 
> ```
> ❯ curl -H "Authorization: Token ${DD_API_KEY}"  "http://localhost:8080/api/v2/findings/?mitigated_on=2025-01-13T18:08:05.911321Z"
> {"count":1,"next":null,"previous":null,"results":[{"id":1, ... }]}
> ```
> 
> The `mitigated_after` and `mitigated_before` filters _sort of_ work as expected with simple dates, but I assume it is just plugging in a `00:00:00` for the time, so for example this returns my 3 findings that were mitigated today, 2025-01-13, where `mitigated_after` would only have included things _after_ today (23:59:59) in the past.
> 
> ```
> ❯ curl -H "Authorization: Token ${DD_API_KEY}"  "http://localhost:8080/api/v2/findings/?mitigated_after=2025-01-13"
> {"count":3,"next":null,"previous":null,"results":[{"id":1, ... }]}
> ```
> 
> I think this needs some additional tweaking, and could really use some unit tests.

With simple dates being given, it does indeed use 00:00:00 as the time. The request was to make it use DateTimeFilter which means it expects a full datetime. Per the Story: 

```The UI and API use a date filter for the mitigated field rather than a datetime filter. The datetime filter should be used instead

So when you said 'after' and gave 2025-01-13, the assumption is that you meant 2025-01-13T00:00:00. That said, I think we can alter the after to add a 23:59:59 timestamp if it is not added instead of the 00:00:00.
In a similar manner, mitigated_on will NOT work without a specific time. DateTimeFilters being what they are. Happy to entertain thoughts on this one.

File: dojo/filters.py [hunk 2]
```diff
@@ -1544,6 +1545,12 @@ class Meta:
         exclude = ["url", "thread_id", "notes", "files",
                    "line", "cve"]
 
+    def filter_mitigated_after(self, queryset, name, value):
+        if value.hour == 0 and value.minute == 0 and value.second == 0:
+            value = value.replace(hour=23, minute=59, second=59)
```
cneill, line 1550: This logic is a little trickier when it comes to the API, because the user *could* intentionally query for "after 2025-01-01 00:00:00"... But I think this approach strikes the right balance - it seems much more likely that we'll receive dates without timestamps than dates with explicit 00:00:00 timestamps, and this would mirror the UI's behavior as well. In any case, the user can always search for "after 2024-12-31" to achieve the same thing if that's truly what they want.

dojo/filters.py
```diff
@@ -1456,9 +1457,9 @@ class ApiFindingFilter(DojoFilter):
     jira_change = DateRangeFilter(field_name="jira_issue__jira_change")
     last_reviewed = DateRangeFilter()
     mitigated = DateRangeFilter()
-    mitigated_on = DateFilter(field_name="mitigated", lookup_expr="exact")
-    mitigated_before = DateFilter(field_name="mitigated", lookup_expr="lt")
-    mitigated_after = DateFilter(field_name="mitigated", lookup_expr="gt")
+    mitigated_on = DateTimeFilter(field_name="mitigated", lookup_expr="exact")
```
cneill, line 1460: It seems very unlikely to me that a user will ever actually want to list Findings mitigated on an exact timestamp down to the microsecond, even if this would theoretically give them that option. I think searching the range `[ <date> 00:00:00, <date+1> 00:00:00 )` is the right approach here too, regardless of whether the user specified a timestamp.

cneill (approved):
Looks good! 😄

PR #11493, opened 2025-01-02
feat(GHA): Pin azure/setup-helm

kiblik:
Add `pinact` which is able to detect unpinned GHA


.github/workflows/gha-pin.yml
```diff
@@ -0,0 +1,17 @@
+name: Ruff Linter
```
Maffooch, line 1: This really threw me for a loop
```suggestion
name: SHA Pin Check
```

Maffooch (changes requested):

kiblik:
I dropped `pinact-action` because it wasn't possible to use it without pinning another pinning: https://github.com/suzuki-shunsuke/pinact-action/issues/505

The pin is not missing that often. Keeping another config would just increase overhead.

mtesauro (approved):
Approved

hblankenship (approved):

dogboat (approved):

Maffooch (approved):

Issue #11480, opened 2024-12-30
Missing "Fingerprint" Field in Semgrep JSON Report Due to Login Requirement, Causes Immediate Closure of Uploaded Findings


farsheedify:
Description:
Semgrep recently removed the "Fingerprint" field from its reports. If you do not log in to Semgrep Cloud, the value for this field is set to "requires login". This causes all findings to be closed immediately after upload, except for one. This issue should be handled similarly to cases where the "Fingerprint" field is absent in the uploaded report.

Steps to Reproduce:

1. Upload the provided sample file for a product in Defectdojo.
2. Observe that the "Fingerprint" field in the findings is set to "requires login" due to not logging into Semgrep Cloud within our DevSecOps framework.
3. Notice that all findings, except one, are closed immediately after the upload completes.

Expected Behavior:
The system should handle the "requires login" value in the "Fingerprint" field in the same way it handles the absence of this field, preventing the premature closure of findings.

i suspect this can be similar to https://github.com/DefectDojo/django-DefectDojo/issues/11470, however, I am not sure.
[Semgrep_w-o_Fingerprints.json](https://github.com/user-attachments/files/18276313/Semgrep_w-o_Fingerprints.json)

valentijnscholten:
Looking at the Semgrep CLI this is indeed a valid use-case, running a scan without logging in. Would you be able to make a PR to detect this "requires login" string? I notice it's also present in the `lines` field, where it should be ignored as well even though it doesn't affect deduplication / reimport.

farsheedify:
@valentijnscholten I made the following PR:
https://github.com/DefectDojo/django-DefectDojo/pull/11495

PR #11495, opened 2025-01-03
Handling "requires login" in "fingerprint" and "lines" fields of Semgrep JSON Report (issue #11480)

farsheedify:
**Description**

This PR addresses the handling of the "requires login" value in the fingerprint field of Semgrep reports https://github.com/DefectDojo/django-DefectDojo/issues/11480. The change ensures that when the "requires login" value is encountered, it is treated as if it is not set, allowing the deduplication process to fall back to legacy deduplication based on hash codes.

Additionally, a new JSON report for Semgrep has been added to the unittests folder, and a function has been implemented in the test suite to verify this change. While I have not run the unit tests yet, I have tested the parser locally within my DefectDojo environment using the provided test JSON report. The results were compared with the Demo instance of DefectDojo to confirm the resolution of the bug.

**Test results**

The test suite has been extended with a new function to cover the changes in this PR.
I have not run the unit tests yet, but I have manually tested the parser with the newly added JSON report in my local instance of DefectDojo.
Results from both my local environment and the Demo instance of DefectDojo are attached to compare the resolution of the bug. The sample JSON report is also attached.
[fingerprint_test.json](https://github.com/user-attachments/files/18295359/fingerprint_test.json)

unittest-local
unittest-demo


valentijnscholten (changes requested):
Thanks for the PR!

I think the `lines` section can contain the same string that should be ignored. Could you add the same logic there?

Also looks like you might need to rebase onto `bugfix` as there is now a superfluous commit not related to this PR.

farsheedify:
@valentijnscholten 
Sure, 
just to clarify, are you saying that if either the "fingerprint" or "line" contains "requires login," we should consider the unique ID to be null?

farsheedify:
The "requires login" for the "lines" field is now handled, and as a result, the snippet becomes null, as demonstrated in these two images.

snippet-demoDojo
snippet-localDojo

valentijnscholten:
Thanks, let's wait for a maintainer to allow the tests to run in GitHub.

paraddise (approved):
Looks good

blakeaowens (approved):

mtesauro (approved):
Approved

dogboat (approved):

PR #11503, opened 2025-01-04
Ruff: Fix PTH118, merge PTH11

kiblik:
Fix [PTH118](https://docs.astral.sh/ruff/rules/os-path-join/), merge `PTH11*` rules


hblankenship (approved):

blakeaowens (approved):

Maffooch (approved):
I was a little shocked by the syntax here at first, but I think it is actually more readable over all. Great work here

dogboat (approved):

PR #11504, opened 2025-01-04
Simplify the use of custom NGINX SSL certificates with Docker Compose

Bernasv:
Instead of manually copying custom user certificates into the image for SSL usage, the arguments `NGINX_CRT_PATH` and `NGINX_KEY_PATH` have been added to simplify the process. These arguments enable automatic handling of certificates when `GENERATE_TLS_CERTIFICATE` is set to false. Providing the paths to the certificate and key through these arguments simplifies the setup, or they can be left empty to use the default certificate generation.


Maffooch:
Hi @Bernasv thank you for your contribution! I think the majority of folks using DefectDojo use the prebuilt images, and this change does not offer any benefits there. This seems like a change that would better kept locally rather than upstream.

PR #11505, opened 2025-01-05
Match all unchanged findings on reimport

dziewxc:
This is a proposed fix of https://github.com/DefectDojo/django-DefectDojo/issues/3958
Currently when reimporting, only the first matching vulnerability is added to the list of unchanged items. So if there are 2 or more vulnerabilities with the same hash, only the first one will be considered unchanged. Rest of them will get mitigated. With this change all the matching vulnerabilities will be considered unchanged.

I tested several scenarios:
1. Report with duplicated vulns A and B was added and then reimported without changes: no findings were mitigated
2. After first reimport, other one was sent with only vulnerability A in the report: no findings were affected
3. Then the report without any of those findings was sent with reimport: all the findings got mitigated
4. After that I sent again report with finding A and all the duplicates were reactivated

I think this is how it should work, but I may miss some edge case.
About the dynamic finding - I wrote it to update only from the last one, I'm not sure what is it updating exactly but I suspect it shouldn't be called 2 times, the last one is enough.

### DIFF
--- a/dojo/importers/default_reimporter.py
+++ b/dojo/importers/default_reimporter.py
@@ -204,11 +204,11 @@ class DefaultReImporter(BaseImporter, DefaultReImporterOptions):
             deduplicationLogger.debug(f"found {len(matched_findings)} findings matching with current new finding")
             # Determine how to proceed based on whether matches were found or not
             if matched_findings:
-                existing_finding = matched_findings[0]
-                finding, force_continue = self.process_matched_finding(
-                    unsaved_finding,
-                    existing_finding,
-                )
+                for existing_finding in matched_findings:
+                    finding, force_continue = self.process_matched_finding(
+                        unsaved_finding,
+                        existing_finding,
+                    )
                 # Determine if we should skip the rest of the loop
                 if force_continue:
                     continue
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Thank you for the PR. One complication of this approach is that the endpoints are no longer updated for the first finding. This makes sense as we don't know which endpoints belong to which finding. Would there be something smart we can do here? Like maybe look at the endpoints (or other field(s)) to see if we can deduce an exact match?

Unsure about if we would need to inform the user of not being able to process the endpoints.

dziewxc:
> Thank you for the PR. One complication of this approach is that the endpoints are no longer updated for the first finding. This makes sense as we don't know which endpoints belong to which finding. Would there be something smart we can do here? Like maybe look at the endpoints (or other field(s)) to see if we can deduce an exact match?

Not sure to be honest. I assumed it was random before (the first finding was taken), so now I just changed it to the last one. But I don't use endpoints, so it's not obvious for me what is the expected behaviour here.

dziewxc:
> The endpoints are what make this exceptionally tricky. This problem will likely require a very involved and thought out solution. Closing this one for now

Do you think that someone would review the PR if I would figure out the way to handle the endpoints? We want to use the tool in our company, but Trivy sends us a lot of duplicates in reports. It makes tracking those vulnerabilities impossible because they got closed on second reimport.

PR #11542, opened 2025-01-10
Kubescape: Reduce the size of steps to reproduce

maxi-bee:
- This is a change on the kubescape parser
- It removes the resource objects (a whole manifest) from "steps to reproduce" as it is often so long (thousands of lines) that conflicts with the default Jira configurations of maximum accepted length of fields, resulting on the failing to create a Jira ticket via the Defect Dojo integration
- Note that also, there is arguably little value on storing this very large objects on the database (for duplicate and original findings)
- Potentially, the Jira integration should validate that, but that isn't possibly the case

### DIFF
--- a/dojo/tools/kubescape/parser.py
+++ b/dojo/tools/kubescape/parser.py
@@ -98,8 +98,6 @@ class KubescapeParser:
 
                         steps_to_reproduce = "The following rules have failed :" + "\n"
                         steps_to_reproduce += "\t**Rules:** " + str(json.dumps(control["rules"], indent=4)) + "\n"
-                        steps_to_reproduce += "Resource object may contain evidence:" + "\n"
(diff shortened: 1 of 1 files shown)


a-ruff (approved):
LGTM

cneill (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

maxi-bee:
Hey folks. Once this is merged to dev when is this expected to hit a release?

mtesauro:
DefectDojo does minor releases every month that takes the dev branch and merges that to master/main. Those happen generally on the first Monday of the month.

The releases between the minor releases are based on the bugfix branch and happen every Monday between minor releases.

In the case of this PR, it will be released in 2.43.0 on Feb 3rd.

PR #11545, opened 2025-01-10
Request Review does not apply RBAC in an expected way

hblankenship:
When a user is added to a group with the global role Owner those users should be able to be requested to review any Finding that exists. They should also be able to be added as leads on tests and engagements. 

[sc-9708]


mtesauro (approved):
Approved

Maffooch (approved):

cneill (approved):

blakeaowens (approved):

PR #11546, opened 2025-01-10
Prefetching multiple endpoints should return all prefetch models

hblankenship:
When attempting a prefetch with multiple parameters, the response would only return one of the prefetch parameters. 

[sc-6144]


hblankenship:
Apparently an interesting case where the code actually works but the unittests fail....

Maffooch:
Moving out of draft to have tests run

mtesauro (approved):
Approved

dogboat (approved):

Maffooch (approved):

PR #11548, opened 2025-01-10
Verified Status Toggle: Add Granularity

Maffooch:
Adding some granularity to the enforce verified status to allow users to make more granular changes

image

[sc-9724]


mtesauro (approved):
Approved

grendel513 (approved):

blakeaowens (approved):

unittests/test_jira_import_and_pushing_api.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, *args, **kwargs):
         DojoVCRAPITestCase.__init__(self, *args, **kwargs)
 
     def assert_cassette_played(self):
-        if True:  # set to True when committing. set to False when recording new test cassettes
+        if False:  # set to True when committing. set to False when recording new test cassettes
```
hblankenship, line 47: Should this be False?

hblankenship (changes requested):
Check my previous comment.

unittests/test_jira_import_and_pushing_api.py
Maffooch: Good catch! it should be set to `True`
```suggestion
        if True:  # set to True when committing. set to False when recording new test cassettes
```

hblankenship (approved):

PR #11566, opened 2025-01-14
Reimport Legacy Reimport: Bump logging from debug to warning

Maffooch:
When a legacy reimport algo is encountered, the logging shouldn be a bit more verbose


mtesauro (approved):
Approved

cneill (approved):

dogboat (approved):

grendel513 (approved):

PR #11586, opened 2025-01-16
feat(ui): Add option to define custom css and js

kiblik:
Custom styles (in CSS) are beneficial if you are running DD in the company and you would like to add some corporate colors.

### DIFF
--- a/dojo/context_processors.py
+++ b/dojo/context_processors.py
@@ -26,6 +26,8 @@ def globalize_vars(request):
         "DOCUMENTATION_URL": settings.DOCUMENTATION_URL,
         "API_TOKENS_ENABLED": settings.API_TOKENS_ENABLED,
         "API_TOKEN_AUTH_ENDPOINT_ENABLED": settings.API_TOKEN_AUTH_ENDPOINT_ENABLED,
+        "CUSTOM_CSS": settings.CUSTOM_CSS,
+        "CUSTOM_JS": settings.CUSTOM_JS,
     }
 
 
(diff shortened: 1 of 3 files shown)


kiblik:
@mtesauro or @Maffooch any feedback regarding this proposal?

zet-mar:
From my point of view, it could be beneficial not just for our company but also for all the others who have defined company design. 
Currently, it is possible to solve this but by "ugly hack," which will be solved by this implementation. Can you review it and push it forward to production 🙏 
Thank you in advance.

kiblik:
> It could be useful. I wonder if we should add a comment about the risks of including/injecting javascript?

Thank you for the review. Where should I place it? To release notes? Or to the `settings.dist.py` file, next to the definition?

valentijnscholten:
> Or to the settings.dist.py file, next to the definition?

I vote for this.

Maffooch:
Hi @kiblik,

First off, thank you so much for opening this PR — and for everything you’ve been contributing. Your work has made a huge difference, and we’re really lucky to have you involved.

We gave this change some serious thought. It’s a really interesting idea, and we can absolutely see why having more flexibility would be helpful. That said, we’re a little hesitant to allow injecting custom JS and CSS directly, mainly because of a few concerns around brand consistency, potential security risks, and the challenges it could create for supporting users. We want to make sure the experience stays stable and predictable for everyone using the project.

That said — if there’s a specific problem you’re trying to solve with this, we’d love to work with you to find another way! Maybe there’s a safer or more scoped approach we can take.

Thanks again for everything you’re doing — we truly appreciate it. Looking forward to hearing your thoughts!

PR #11592, opened 2025-01-17
Django AuditLog: Upgrade to 3.x

Maffooch:
django-auditlog has a new version that will make filtering and other things more efficient

[sc-5527]


unittests/test_copy_model.py
```diff
@@ -1,9 +1,36 @@
+from django.db import models
+
 from dojo.models import Endpoint, Endpoint_Status, Engagement, Finding, Product, Test, User
 
 from .dojo_test_case import DojoTestCase
 
 
-class TestCopyFindingModel(DojoTestCase):
+class EqualityHelpers:
+    def compare_querysets(self, queryset_1, queryset_2):
```
kiblik, line 9: What about to use https://docs.djangoproject.com/en/5.1/topics/testing/tools/#django.test.TransactionTestCase.assertQuerySetEqual
I have never used that but based on the name, it might be compatible.

mtesauro (approved):
Approved

blakeaowens (approved):

unittests/test_copy_model.py
Maffooch: Great find! This was a drop in replacement when I ran tests locally'

Thanks for the suggestion 😄

hblankenship (approved):

dogboat (approved):

Maffooch:
These tests run and pass locally... Closing and reopening

PR #11593, opened 2025-01-17
Add GHA updater for keep sample data at a reasonable date

Maffooch:
The data in the sample data fixture is super old, and needs to be updated. This PR adds a new binary to slide the dates of all objects forward based on the most recent date in the fixture vs today. So for example, the latest date in this fixture file was in October 2021. This is a little over three years ago, so the binary will slide all dates in the file roughly three years forward

The GHA is set to run quarterly and open up a new PR to be reviewed with the changes outlined above

[sc-9788]

### DIFF
--- a/.github/workflows/update-sample-data.yml
+++ b/.github/workflows/update-sample-data.yml
@@ -1,4 +1,4 @@
-name: Run Binary and Create PR
+name: Update Sample Data
 
 env:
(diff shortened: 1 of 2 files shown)


mtesauro (approved):
Approved

dogboat (approved):
Neat!

hblankenship (approved):

grendel513 (approved):

PR #11594, opened 2025-01-17
View Alerts: Sanitize and mark safe

Maffooch:
Alerts are being HTML escaped in places where they should not be. Rather on relying on the default HTML escaping, we should use the markdown bleaching templatetag to allow for more options in object titles. Here is a before and after of an innocuous alert vs a dangerous one (no xss triggered here). 

Before: 
Screenshot 2025-01-17 at 2 23 43 PM

After:
Screenshot 2025-01-17 at 2 23 56 PM

[sc-9944]

### DIFF
--- a/dojo/templates/dojo/alerts.html
+++ b/dojo/templates/dojo/alerts.html
@@ -1,6 +1,7 @@
 {% extends "base.html" %}
 {% load i18n %}
 {% load static %}
+{% load display_tags %}
 {% block content %}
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

dogboat (approved):

hblankenship (approved):

grendel513 (approved):

PR #11565, opened 2025-01-14
:tada: advance NoseyParker to support version 0.22.0

manuel-sommer:
#11535

Issue #11535, opened 2025-01-10
Need support for Noseyparker latest version in DefectDojo


Himan10:
Hey Team,

We have been using DefectDojo in our organisation for quiet a few time and we recently made a change in the tools that we use for secret scanning in github/gitlab repositories, previously we were using gitleaks to collect the secrets and we replaced it with Noseyparker. However, with the versions we are using of defectdojo i.e., v 2.42.0 and below, we are not able to upload the secret scanning report obtained from noseyparker recent latest version i.e., 0.21.0 and above. We are receiving an error that says

`An exception error occurred during the report import: ['Invalid Nosey Parker data, make sure to use Nosey Parker v0.16.0']`

We have a requirement in our organisation to use latest version of Noseyparker and Defectdojo both but due to this issue, we can't meet the compliance requirements of our organisation. 

I've also compared the finding stats of both the version of Noseyparker i.e., 0.16 VS 0.21. I ran both the versions of noseyparker on the same repository and here's the generated finding summary. 

Running the scan using noseyparker v16
<img width="1792" alt="Screenshot 2025-01-10 at 11 36 30 AM" src="https://github.com/user-attachments/assets/7e023d18-21a0-4d98-b746-9408c8007c68" />

Running the scan using noseyparker v21
<img width="1792" alt="Screenshot 2025-01-10 at 12 05 19 PM" src="https://github.com/user-attachments/assets/e04941e7-fb71-4061-a6cd-b1a66ca31288" />

As shown in the above two images, both the version have different set of findings and the latest version often groups a specific category of findings. Also, as mentioned in the documentation of Noseyparker latest version, there have been some changes in the JSON output report i.e., 

```
The JSON output format from report has changed slightly (https://github.com/praetorian-inc/noseyparker/pull/236).

Now, the JSON representation of provenance entries from extensible enumerators (i.e., scan --enumerator=FILE, introduced in v0.20.0) includes an additional "payload" field around the actual provenance content. For example, an extended provenance entry that previously would look like this:

{"kind": "extended", "filename": "input.txt"}
is now represented like this:

{"kind": "extended", "payload": {"filename": "input.txt"}}
```
along with some changes in the representation of report in human format and datastore schema.

Could you please look into this issue and let us know if there'll be support for latest noseyparker versions in the upcoming defectdojo versions or if there's a way we could do this on our own, for example: by writing a parser or change the format of latest noseyparker report.

manuel-sommer:
[...]

PR #11615, opened 2025-01-22
Noseyparker version 16 and 22 support for git history and without git history scans

Himan10:
**ISSUE:** https://github.com/DefectDojo/django-DefectDojo/issues/11535

@manuel-sommer also has pushed some changes to a different PR regarding the support for version 22, I've added a comment in this PR as well with regards to issues I've encountered with his code. The reason to create a new PR for this because I was not able to push changes to manuel's PR, so I've attached his PR as well in case if anyone wants to check out the previous changes 

**Manuel's PR:** https://github.com/DefectDojo/django-DefectDojo/pull/11565

**Reason:** 
> I've identified some issues with the changes you've pushed to this PR. In Noseyparker, we have a flag to scan repositories without including their Git history (the flag --git-history=none is used for this). The structure of the reports generated by Noseyparker with --git-history=none is different from those generated without this flag. When running a normal scan with Noseyparker (which includes the Git history), the scan generally includes commit metadata. However, this commit metadata is not present in scans where --git-history=none is used.

> Previously, we were unable to import Noseyparker scans, where the Git history of the repository was excluded, into DefectDojo. I've made some changes to your PR to support importing both the versions of the Noseyparker report, one with Git history and one without. I've tested this code locally with the latest version of DefectDojo and v2.33.0 (release mode).

### DIFF
--- a/dojo/tools/noseyparker/parser.py
+++ b/dojo/tools/noseyparker/parser.py
@@ -17,85 +17,158 @@ class NoseyParkerParser:
 
     def get_description_for_scan_types(self, scan_type):
         return "Nosey Parker report file can be imported in JSON Lines format (option --jsonl). " \
-               "Supports v0.16.0 of https://github.com/praetorian-inc/noseyparker"
+               "Supports v0.16.0 and v0.22.0 of https://github.com/praetorian-inc/noseyparker"
 
     def get_findings(self, file, test):
         """
         Returns findings from jsonlines file and uses filter
         to skip findings and determine severity
         """
-        dupes = {}
-
+        self.dupes = {}
         # Turn JSONL file into DataFrame
         if file is None:
             return None
         if file.name.lower().endswith(".jsonl"):
             # Process JSON lines into Dict
             data = [json.loads(line) for line in file]
-
             # Check for empty file
             if len(data[0]) == 0:
                 return []
-
             # Parse through each secret in each JSON line
             for line in data:
                 # Set rule to the current secret type (e.g. AWS S3 Bucket)
-                try:
-                    rule_name = line["rule_name"]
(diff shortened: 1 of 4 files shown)


dojo/tools/noseyparker/parser.py
```diff
@@ -17,85 +17,158 @@ def get_label_for_scan_types(self, scan_type):
 
     def get_description_for_scan_types(self, scan_type):
         return "Nosey Parker report file can be imported in JSON Lines format (option --jsonl). " \
-               "Supports v0.16.0 of https://github.com/praetorian-inc/noseyparker"
+               "Supports v0.16.0 and v0.22.0 of https://github.com/praetorian-inc/noseyparker"
```
manuel-sommer, line 20: Hi @Himan10 ,
I suggest to close this PR and just review mine. This PR also targets branch master which is not the right target.

manuel-sommer (changes requested):

dojo/tools/noseyparker/parser.py
Himan10: okay but could you give me permissions to push changes to your PR ? It looks like I don't have permissions. I reviewed your changes tho but it was failing with scans where `--git-history` was set to none.

manuel-sommer: just add a comment in the code inside the PR and submit the review. (Do a code review)

mtesauro: @manuel-sommer :+1: on keeping the original #11565 open and closing this one

manuel-sommer:
This can be closed @mtesauro The functionality is covered in my PR

mtesauro:
Closing - work will happen in #11565

PR #11622, opened 2025-01-23
Accessibility ergonomy filtering

### LINKED ISSUE #11621: Accessibility and ergonomy of the filtering section
**Bug description**

Users will find it difficult to find and distinguish between filters in the "Open Findings" dialog because they are not arranged on a uniform vertical grid.

Users will assume that filters are active because the "Clear Filters" link is also implemented as an interactive element if no filters have been selected before.

**Steps to reproduce**
1. Go to any section with filtering, for example all findings filtering.
2. See what is described above.

**Expected behavior**

The link "Clear Filters" should only be active if filters have actually been selected. Otherwise, the link should be grayed out.
The elements should be arranged on a uniform vertical grid.

Image

littlesvensson:
Fixes #11621

**Description**

Used bootstrap classes and some JS logic for filtering. The fields are now aligned so its easier to orientate and keep responsivity at the same time. Clear button is disabled in case no filter is set in the fields.

Before:
Screenshot 2025-01-23 at 11 29 22

After:
Screenshot 2025-01-23 at 11 23 12

smaller screen
Screenshot 2025-01-23 at 11 23 38

smaller screen with at least one filter option 
Screenshot 2025-01-23 at 11 23 50


mtesauro:
@littlesvensson You can't see me but I'm doing a happy dance. :man_dancing: 

One thing, please make this PR against the dev branch rather than master. We don't accept source PRs against master.

littlesvensson:
> One thing, please make this PR against the dev branch rather than master. We don't accept source PRs against master.

thanks @mtesauro for feedback. Yes sorry, that was only accidentally to master 🙈  , original thought was to the bugfix as it fixes one. I will reopen the PR to a different branch once I ll update the unit tests.

mtesauro:
@littlesvensson 
> > One thing, please make this PR against the dev branch rather than master. We don't accept source PRs against master.
> 
> thanks @mtesauro for feedback. Yes sorry, that was only accidentally to master 🙈 , original thought was to the bugfix as it fixes one. I will reopen the PR to a different branch once I ll update the unit tests.

Thanks for moving to the dev branch.

While I LOVE this change, we try to keep  bugfix kinda small (though loads of that is qualitative judgements) I asked for Dev since this is as much as improvement as it's a bugfix so felt 'better' as a change for the next minor version.

mtesauro:
> @mtesauro Ahhh ok, thank you for the explanation, I will be more cautious about branch choices next time 🙈 Should I still move it? As I see its already approved from some of you in the bugfix one.

@littlesvensson If we have some approvals already, it's not worth changing the branch. It's just a general rule we do our best to follow but like all rules, there's the occasional exception.

I'd rather have you send the time you want to contribute on better things than switching a branch on a PR that already has approvals. :smile:

PR #11634, opened 2025-01-24
Accessibility ergonomy filtering

littlesvensson:
Fixes https://github.com/DefectDojo/django-DefectDojo/issues/11621

Description

Used bootstrap classes and some JS logic for filtering. The fields are now aligned so its easier to orientate and keep responsivity at the same time. Clear button is disabled in case no filter is set in the fields.

Before:
Screenshot 2025-01-23 at 11 29 22

After:
Screenshot 2025-01-23 at 11 23 12

smaller screen
Screenshot 2025-01-23 at 11 23 38

smaller screen with at least one filter option
Screenshot 2025-01-23 at 11 23 50


kiblik (approved):

mtesauro (approved):
SO VERY Approved

blakeaowens (approved):

PR #11637, opened 2025-01-24
Import History: Make the absence of action more clear

Maffooch:
In the event an empty report is submitted, and there are no findings to be closed, the test_import_finding_action_set will be empty. When viewing import history objects, there is no indication as to what occurred. 

image

[sc-9738]


mtesauro (approved):
Approved

hblankenship (approved):

dogboat (approved):

blakeaowens (approved):

PR #11643, opened 2025-01-24
Ruff: Add and fix TRY300

kiblik:
Add rule [TRY300](https://docs.astral.sh/ruff/rules/try-consider-else/) and fix it. The original recommendation is to put it to `else` but in my opinion, in all these cases, it might go outside of the `try` part at all.

There were also 2 ruff-autofixes in `dojo/jira_link/helper.py` (because of "RET505: Unnecessary `else` after `return` statement")


Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

hblankenship (approved):

PR #11647, opened 2025-01-25
Ruff: Add and fix N999

kiblik:
Add rule [invalid-module-name (N999)](https://docs.astral.sh/ruff/rules/invalid-module-name/#invalid-module-name-n999) and fix following violations:

```
dojo/settings/settings.dist.py:1:1: N999 Invalid module name: 'settings.dist'
tests/Import_scanner_test.py:1:1: N999 Invalid module name: 'Import_scanner_test'
```

This is a bit of radical change as it is renaming a globally used and known setting file. I will be happy for any feedback.


Maffooch:
@kiblik this one feels a little too radical form my taste. I think we should ignore this rule for now, and revisit in the future, if applicable

Maffooch (changes requested):
Reposting this in case it got missed before: 

https://github.com/DefectDojo/django-DefectDojo/pull/11647#issuecomment-2622959912

kiblik:
> @kiblik this one feels a little too radical form my taste. I think we should ignore this rule for now, and revisit in the future, if applicable

I fully agree that it looks "radical". But not sure how it might change in the future.

Can I at least rename `Import_scanner_test` and add `noqa` to `settings.dist`

Maffooch:
There is likely a good chance that we do not revisit this one. Renaming `settings.dist.py` is a big move that would impact docs, historical knowledge, potential automation, etc.

For now, I think just ignoring the rule in the ruff toml would be the best approach forward here

kiblik:
In that case, I will rename only integration test and set noqa for settings.dist.py

Maffooch:
@kiblik I think you suggestion is sound - I missed your last message in my inbox

Maffooch (approved):

mtesauro (approved):
Approved

dogboat (approved):

mtesauro:
@kiblik This one is good to go except for the merge conflicts. :+1:

PR #11649, opened 2025-01-25
feat(bash-script): Drop all `dc-` scripts

kiblik:
Since DD does not support MySQL and RabbitMQ, `dc-` scripts do not have additional values (they existed to simplify usage of `--profile` and `--env-file` parameters). Now they are more limiting as they support only one additional parameters. 

As only one DB backed and one MB backed is used, `--no-deps` is no longer needed.

Scripts for unittests and integration tests are still useful (from my perspective). But they have been renamed to decrease confusion as to why they are named that way.


mtesauro:
Thanks for this PR - it was on my list of things to simplify in DefectDojo so it's as easy to use as is feasible and, like you noticed, not having MySQL & RabbitMQ support means those `dc-` things are no longer necessary :+1:

mtesauro (approved):
Approved

hblankenship (approved):

Maffooch (approved):

Maffooch:
@kiblik I added back the `--no-deps` flag to unit tests in #11592. For some reason, this was breaking things for that PR, though, tests would pass locally - Not sure what the deal was

PR #11650, opened 2025-01-25
fix(exit in bash): Fix handling exit in initializer

kiblik:
Reopen #10493

PR #11651, opened 2025-01-25
feat(db-checker): Extension of "db reachable"

kiblik:
Reopen #10497

Extend `wait_for_database_to_be_reachable`. Not only for simple operation but check that DB is compatible.

Added based on https://github.com/DefectDojo/django-DefectDojo/issues/10490

This PR also contains changes from #11650 because they are needed for proper testing of this functionality.

### DIFF
--- a/.github/workflows/k8s-tests.yml
+++ b/.github/workflows/k8s-tests.yml
@@ -85,7 +85,7 @@ jobs:
                ./helm/defectdojo \
                --set django.ingress.enabled=true \
                --set imagePullPolicy=Never \
-               --set initializer.keepSeconds="x" \
+               --set initializer.keepSeconds="-1" \
                ${{ env[matrix.databases] }} \
                ${{ env[matrix.brokers] }} \
                --set createSecret=true \
--- a/.github/workflows/k8s-tests.yml
+++ b/.github/workflows/k8s-tests.yml
@@ -85,6 +85,7 @@ jobs:
                ./helm/defectdojo \
                --set django.ingress.enabled=true \
                --set imagePullPolicy=Never \
+               --set initializer.keepSeconds="x" \
                ${{ env[matrix.databases] }} \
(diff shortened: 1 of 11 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

.github/workflows/k8s-tests.yml
```diff
@@ -85,6 +85,7 @@ jobs:
                ./helm/defectdojo \
                --set django.ingress.enabled=true \
                --set imagePullPolicy=Never \
+               --set initializer.keepSeconds="x" \
```
cneill, line 88: Is this value intentional? Docs suggest this `ttlSecondsAfterFinished` should be an integer - I'm not sure what it would do with a value of "x"?

https://github.com/DefectDojo/django-DefectDojo/blob/3f21fb35724788cb87cb3169b781c84e051b6d34/helm/defectdojo/templates/initializer-job.yaml#L22

Screenshot 2025-02-03 at 17 50 12

kiblik: You would be right if the value were passed to `ttlSecondsAfterFinished`. However thanks to the previous line whole `ttlSecondsAfterFinished` is dropped if you put their value other than a positive integer. This functionality was added in #11257
https://github.com/DefectDojo/django-DefectDojo/blob/3f21fb35724788cb87cb3169b781c84e051b6d34/helm/defectdojo/templates/initializer-job.yaml#L21

I might change it to `keep me`, `0`, `-1`, ... or anything more descriptive here. I'm open to any recommendation.

cneill: Ah, I should have read up one line! :joy: I think -1 is a pretty common pattern for "ignore this".

cneill (approved):

blakeaowens (approved):

PR #11654, opened 2025-01-26
fix(dockerfile): rename django stage in alpine

kiblik:
Until know, it wasn't possible to build `alpine` locally and run it:
```bash
$ export DEFECT_DOJO_OS=alpine
$ ./dc-build.sh       
Checking docker compose version
Supported docker compose version
Building docker compose
[+] Building 0.1s (1/1) FINISHED                                                                  docker:default
 => [uwsgi internal] load build definition from Dockerfile.django-alpine                                    0.0s
 => => transferring dockerfile: 4.34kB                                                                      0.0s
failed to solve: target stage "django" could not be found
```

The same naming convention for both images is needed.


Maffooch (approved):

mtesauro (approved):
Approved

cneill (approved):

blakeaowens (approved):

PR #11669, opened 2025-01-27
make alpine builds work with docker compose

valentijnscholten:
The docker compose expects `django` as target image, not `django-alpine`.

The release keys for Node where outdated, fixed as well.


kiblik:
We are on the same page 😀
https://github.com/DefectDojo/django-DefectDojo/pull/11654

valentijnscholten:
Happy to close mine, but I added another fix that is needed when you're not buildng on `x64`.

Maffooch (approved):
I think both PRs can go in

mtesauro (approved):
Approved

cneill (approved):

blakeaowens (approved):

PR #11673, opened 2025-01-27
Support builds for different architectures including arm64

valentijnscholten:
**Description**

Currently Defect Dojo base images are pinned to the `manifest digest` of `lindex/amd64` images.
This prevents users or systems from building on/for other architectures such as `linux/arm64`.

This PR changes this into using `index digests`. These digests do not point to a single `linux/amd64` image, but to the collection of images for all architectures.

I couldn't find a good page where this is explained, so I asked ChatGPT:

> In Docker, especially when dealing with multi-architecture (multi-arch) images, it's essential to understand the distinction between an image's manifest digest and the index (or manifest list) digest.
> 
> Manifest Digest: Each Docker image has a manifest that describes its properties, including its layers, size, and configuration. The digest is a unique identifier (a SHA256 hash) of this manifest. When you pull an image using its digest, you're ensured to retrieve the exact same image every time, regardless of tags. 
> STACK OVERFLOW
> 
> Index (Manifest List) Digest: For multi-arch images, Docker uses a manifest list (also known as an image index) to reference multiple image manifests, each tailored for a specific architecture or operating system. The manifest list itself has a digest, which serves as a pointer to the appropriate image manifest based on the client's platform. Pulling an image using the manifest list's digest allows Docker to automatically select the correct image variant for your system. 
> GOOGLE CLOUD
> 
> In summary, while a manifest digest uniquely identifies a specific image variant, the index (manifest list) digest references a collection of image variants suitable for different platforms, facilitating seamless deployment across diverse environments.

**Test results**

AMD64:
The result is that when building on `linux/amd64`, the result is exactly the same as before. Docker will automatically pick the `linux/amd64` digest from the index digest. This will be the exact same digest we previously put into the `Dockerfile` / `docker-compose.yml` files.

ARM64:
When on `linux/arm64`, docker will now choose `linux/arm64` images for `alpine`, `debian`, `postgres`, etc. And will build `django` and `nginx` images for `linux/arm64`.

This allows users to build Defect Dojo on arm64 platforms, for example to run on the [cheaper AWS Graviton instances](https://about.gitlab.com/blog/2021/08/05/achieving-23-cost-savings-and-36-performance-gain-using-gitlab-and-gitlab-runner-on-arm-neoverse-based-aws-graviton2-processor/).
It might also help users/developers on MacOS, not sure as I don't have a Mac myself.

Builds and smoke test done on AWS t4g instance.

**Next steps**
Next step would be for us to also build arm64 images and publish them on Docker Hub.
Currently our builds are running in GitHub actions on AMD64 runners. We could build these arm64 images using Qemu.
I've noticed that these builds are not always succeeding, mainly because compilation of Python wheels like uwsgi are failing.
I will look into running an arm64 build on a native arm64 GitHub runner image, but this requires some fiddling to get the index digest to be created correctly.

Let's start with this PR to help people get started on arm64.

**Additional info**
This PR also simplifies the `node` install process for the `alpine` builds. I am not aware of any need to build `node` from source, we can just `add` the `alpine` package.


Maffooch (approved):
I tested this locally and found good results

mazmar:
+1 on this, we are running pure ARM kubernetes now and cannot deploy defectdojo.

valentijnscholten:
@mazmar we'll try to publish arm64 images for releases. For now it could be interesting to see if you can build and deploy with the changes in this PR. It's based off the ’dev’ branch, so it's close to what will be 2.43.0 on Monday.

mtesauro (approved):
Approved

blakeaowens (approved):

dogboat (approved):

PR #11685, opened 2025-01-28
feat(db-migration-checker): Add successful message

kiblik:
There was a missing successful message for `db-migration-checker`. It might be a bit confusing if the last message of the container is `Database is not migrated to the latest state yet` or if there is a message at all.


valentijnscholten (approved):

Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #11689, opened 2025-01-29
Text elements must have sufficient color contrast against the background.

oussama-taoufiq:
**Error Title**
Text elements must have sufficient color contrast against the background
**Error Severity**
Serious
**Status**
Fail
**Accessibility Issue**
[Description of issue] The 2 instance of Colour contrast as shown in the screenshots : The visual presentation of text and images of text does not have a contrast ratio of at least 4.5:1

**Impact on users**

- People with low vision often have difficulty reading text that does not contrast with its background. This can be exacerbated if the person has a color vision deficiency that lowers the contrast even further. Providing a minimum luminance contrast ratio between the text and its background can make the text more readable even if the person does not see the full range of colors. It also works for the rare individuals who see no color.

[Pattern] Throughout the application dashboard.
[Steps to reproduce]
With the application called Colour Contrast Analyser, select the foreground colour and the background colour of the 2 mentioned in the screenshot and check the contrast ratio

**Remediation**
[Recommendation] Please have the contrast ratio of the 2 instance and across the site of at least 4.5:1

[Additional Resources]
https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html

**Screenshot**
image

************************************************
**Screenshot**
image

### DIFF
--- a/dojo/static/dojo/css/dojo.css
+++ b/dojo/static/dojo/css/dojo.css
@@ -272,7 +272,7 @@ body #header form#api_selector .input input#input_apiKey {
 }
 
 .breadcrumb a.active {
-    color: #777 !important;
+    color: #546474 !important;
 }
 
 .badge.badge-count {
@@ -1353,7 +1353,7 @@ div.custom-search-form {
 }
 
 .panel-default > .panel-heading {
-    color: white;
+    color: #252525;
     font-size: 16px;
     background-color: rgba(32, 166, 216, 0.75);
     border-color: #A1C0D7;
--- a/dojo/static/dojo/css/dojo.css
+++ b/dojo/static/dojo/css/dojo.css
@@ -446,11 +446,6 @@ form ul {
     padding-left: 0px
 }
 
-form ul li a#reset-password,
-form ul li a#forgot-username {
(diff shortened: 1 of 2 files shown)


mtesauro (approved):
Approved

Maffooch (changes requested):
The following class would need be updated as well: https://github.com/DefectDojo/django-DefectDojo/blob/85acabd8f7f00e2068bb7bbf89591974dfead6bf/dojo/static/dojo/css/dojo.css#L1357C1-L1360C2

Maybe it is because I have seen the white text for so long, but the black makes this look totally different 😅 @blakeaowens what are your thoughts on this?
image

oussama-taoufiq:
@Maffooch Should I make the changes or wait for @blakeaowens's response?

Maffooch:
> @Maffooch Should I make the changes or wait for @blakeaowens's response?

It would probably be best to wait for @blakeaowens  here

Maffooch:
Thank you for your contribution! We have been toying with the idea of  updating the idea of DefectDojo, but have not landed on any concrete ideas for a color scheme. Until that time, I think it might be best to wait on any major changes for the time being.

PR #11701, opened 2025-01-31
api spec: SLA days remaining can be null

### LINKED ISSUE #10793: APIv2 Serializer - sla_days_remaining in FindingSerializer should have_null set to True
Be informative
I use OpenAPI spec to generate me a nice client without too much effort but the generated models (or the spec in Django) might be wrong

Bug description
I use https://openapi-generator.tech/ to generate me a Client + Models from Swgger Json "/api/v2/oa3/schema/?format=json"
So far all good.

However when I try to create report for engagment I got validation issues becasue some findings (Severity = INFO) have empty SLA, so null value for `sla_days_remaining`

so this line (https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/api_v2/serializers.py#L1656) should be patch

`sla_days_remaining = serializers.IntegerField(read_only=True)` to
sla_days_remaining = serializers.IntegerField(read_only=True, allow_null=True)

And/Or give the ability to set SLA for Info findings.

valentijnscholten:
fixes #10793

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1719,7 +1719,7 @@ class FindingSerializer(TaggitSerializer, serializers.ModelSerializer):
     )
     push_to_jira = serializers.BooleanField(default=False)
     age = serializers.IntegerField(read_only=True)
-    sla_days_remaining = serializers.IntegerField(read_only=True)
+    sla_days_remaining = serializers.IntegerField(read_only=True, allow_null=True)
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

hblankenship (approved):

PR #11732, opened 2025-02-04
Hash Codes: Ensure Vulnerability IDs are used in calculations when added outside of import

Maffooch:
When creating findings outside of an import/reimport through the UI and API, vulnerability IDs are not being set on the finding model before the finding is saved. This leads to inaccurate hash codes being calculated, and duplicates not being matched successfully. This PR addresses those issues

[sc-10032]


mtesauro (approved):
Approved

hblankenship (approved):

dogboat (approved):

blakeaowens (approved):

PR #11738, opened 2025-02-05
jira push error reasons should not be propagated to all channels

### LINKED ISSUE #11575: Notifications regarding problems with Jira should ignore `WARNING` level events
DefectDojo version: 2.41.3

We did set `Jira minimum severity` to `High` in System settings, enabled MS Teams notifications and in System Notification Settings enabled notifications regarding JIRA problems:

image

And the problem is, we receive tons of messages about `Finding below the minimum JIRA severity threshold`:

image

But this is not an error, nor a problem with integration with Jira (that would be, e.g. lack of permissions or expired token). In `celery-worker` this is a `WARNING` level.

```
[13/Jan/2025 12:33:00] WARNING [dojo.jira_link.helper:779] <class 'dojo.models.Finding'>: CVE-2024-50602 Libexpat1 2.4.7-1ubuntu0.4 cannot be pushed to JIRA: Finding below the minimum JIRA severity threshold (High)..
```

IMO only `ERROR` level events should be sent (just as it says in the description in the UI).

valentijnscholten:
**Description**

Historically the current notification system always propagates all messages to all channels configured for the notification type / source.
Some messages however are not relevant for all channels, for example reasons why findings cannot be pushed to JIRA.
But we do need to display these alerts in the UI because that's the only place where users can see feedback of these errors/warnings.
This PR introduces an `alert_only` flag to indicate that notifications should only be an alert.
It does reuse all the existing `NotificationManager` logic around System vs Personal notifications, templating and recipients.

This PR also fixes a bug where the "See All Alerts / Clear All Alerts" buttons where not shown due to a Javascript error.

Fixes #11575

**Test results**
Creating a unit test for this is currently not possible, unless I create a whole bunch of code to Mock more things.
I'm not sure if that's worth it for this small corner case situation.
I did test the different scenario's, and can still create JIRA issues: https://defectdojo.atlassian.net/browse/DOJOTEST-24

### DIFF
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -2983,7 +2983,7 @@ def finding_bulk_update_all(request, pid=None):
                         ) = jira_helper.can_be_pushed_to_jira(group)
                         if not can_be_pushed_to_jira:
                             error_counts[error_message] += 1
-                            jira_helper.log_jira_alert(error_message, group)
+                            jira_helper.log_jira_cannot_be_pushed_reason(error_message, group)
                         else:
                             logger.debug(
                                 "pushing to jira from finding.finding_bulk_update_all()",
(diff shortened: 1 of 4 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

hblankenship (approved):

dogboat (approved):

PR #11739, opened 2025-02-05
Duplicate Delete errors: catch IntegrityErrors (A)

valentijnscholten:
**Description**
Fixes and see: #6217

This PR catches integrity errors that might happen if the background dedupe delete job deletes findings before the import process is fully completed. In the absence of transactions, for now we implement this "fix".

**Test results**
Manual testing only as creating a unit test / integration test is not feasible within a reasonable timeframe.

PR #11740, opened 2025-02-05
Duplicate Delete Errors: Inline import history + tags (B)

valentijnscholten:
**Description**
Fixes and see: #6217

This PR replaces the bulk creation of the import history at the end of the (re)import with inline statements to create these records one by one. This prevents `IntegrityErrors` from happening during race conditions with the background duplicate delete job.

**Test results**
Manual testing only as creating a unit test / integration test is not feasible within a reasonable timeframe.


valentijnscholten:
Closed, see #11739

PR #11741, opened 2025-02-05
Duplicate Delete Errors: postpone dedupe and jira (C)

valentijnscholten:
**Description**
Fixes and see: #6217

This PR moves dedupe and pushing to JIRA to the end of the import, after all findings have been created. The goal is to prevent race conditions with the background duplicate delete job.

**Test results**
Manual testing only as creating a unit test / integration test is not feasible within a reasonable timeframe.


valentijnscholten:
Closed, see #11739

PR #11743, opened 2025-02-05
API Tags: Add filter for `AND` expressions

Maffooch:
When using the `tags` filter on a finding, the operation applied with multiple tags is an `OR`, but an `AND` expression would be desired. To avoid creating some breaking changes, this PR adds new filters with the filters for `*tags__and` in all the places where tag filters are currently used

[sc-10134]

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -137,6 +137,10 @@ class CharFieldInFilter(filters.BaseInFilter, filters.CharFilter):
 
 class CharFieldFilterANDExpression(CharFieldInFilter):
     def filter(self, queryset, value):
+        # Catch the case where a value if not supplied
+        if not value:
+            return queryset
+        # Do the filtering
         objects = value.split(",")
         return (
             queryset.filter(**{f"{self.field_name}__in": objects})
@@ -1217,15 +1221,15 @@ class ApiEngagementFilter(DojoFilter):
     tags = CharFieldInFilter(
         field_name="tags__name",
         lookup_expr="in",
-        help_text="Comma separated list of exact tags")
+        help_text="Comma separated list of exact tags (uses OR for multiple values)")
     tags__and = CharFieldFilterANDExpression(
         field_name="tags__name",
         help_text="Comma separated list of exact tags to match with an AND expression")
     product__tags = CharFieldInFilter(
         field_name="product__tags__name",
         lookup_expr="in",
-        help_text="Comma separated list of exact tags present on product")
-    product__tags__and = CharFieldInFilter(
+        help_text="Comma separated list of exact tags present on product (uses OR for multiple values)")
+    product__tags__and = CharFieldFilterANDExpression(
         field_name="product__tags__name",
         help_text="Comma separated list of exact tags to match with an AND expression present on product")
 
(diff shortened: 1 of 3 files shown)


mtesauro (approved):
Approved

dojo/filters.py
```diff
@@ -135,6 +135,20 @@ def __init__(self, *args, **kwargs):
...
+    def filter(self, queryset, value):
+        # Catch the case where a value if not supplied
+        if not value:
+            return queryset
+        # Do the filtering
+        objects = value.split(",")
+        return (
+            queryset.filter(**{f"{self.field_name}__in": objects})
+            .annotate(object_count=Count(self.field_name))
+            .filter(object_count=len(objects))
```
dogboat, line 148: Are we worried about duplicates at all? E.g. if a user submits `tags__and=a,a,b` this won't match anything. Do we need to dedupe, or is that a weird case anyway (why would they do that?).

dojo/filters.py
```diff
@@ -135,6 +135,20 @@ def __init__(self, *args, **kwargs):
...
 
+class CharFieldFilterANDExpression(CharFieldInFilter):
+    def filter(self, queryset, value):
+        # Catch the case where a value if not supplied
+        if not value:
+            return queryset
+        # Do the filtering
+        objects = value.split(",")
+        return (
+            queryset.filter(**{f"{self.field_name}__in": objects})
```
dogboat, line 146: In playing around, I noticed this is currently case-sensitive, but only for the __and versions. Do we care about that? Seems like the init() of DojoFilter does some form magic for tag fields that results in case-insensitive outcomes for the 'or' version, but this isn't exactly for the old UI. Well, anyway, just thought I'd ask.

dogboat (changes requested):
This is great! No real requests, just two questions.

hblankenship (approved):

blakeaowens (approved):

dojo/filters.py
Maffooch: Tags should be forced to lowercasee by default, but the library we are using does not respect the settings it provides. We have had to manually force tags to lowercase in some instances. Having case sensitivity would likely be a good move for folks that are in a "in between" stage of having lower case and mixed case tags

dojo/filters.py
Maffooch: It sounds like you proved the case to prune duplicates in the test you performed. I can work on a fix

dojo/filters.py
```diff
@@ -135,6 +135,20 @@ def __init__(self, *args, **kwargs):
         super(CharFilter, self).__init__(*args, **kwargs)
 
 
+class CharFieldFilterANDExpression(CharFieldInFilter):
+    def filter(self, queryset, value):
+        # Catch the case where a value if not supplied
+        if not value:
+            return queryset
+        # Do the filtering
+        objects = value.split(",")
```
Maffooch, line 144:
```suggestion
        objects = set(value.split(","))
```
Removing duplicate values per feedback from @dogboat

dojo/filters.py
Maffooch: Please see the latest comment that addresses this feedback

dogboat (approved):

PR #11750, opened 2025-02-06
Ruff: Merge safe rules (A, FIX, PLW01)

kiblik:
After many fixes, we can safely merge some rules without touching the code. I move `FIX002` so we can merge all `FIX` rules.


Maffooch (approved):
🎉

mtesauro (approved):
Approved

dogboat (approved):

PR #11769, opened 2025-02-07
Fix aria-label in search box

kevin-vuong99:
**Description**

The "Search" text area has no labels corresponding to it. 
image

**Accessibility Issue**

[Impact on users] If a form control does not have a properly associated text label, the function or purpose of that form control may not be presented to screen reader users. Form labels also provide visible descriptions and larger clickable targets for form control.

[Steps to reproduce] While using the WAVE addon for FireFox, run it on the page and you will get this error.

**Remediation**

The aria-label had a typo which was fixed.


mtesauro (approved):
Approved

blakeaowens (approved):

PR #11784, opened 2025-02-10
Fix async fincding import

FirePanda169:
**Description**

Adding missing decorator for findings import function

**Test results**

...

### DIFF
--- a/dojo/importers/default_importer.py
+++ b/dojo/importers/default_importer.py
@@ -5,6 +5,8 @@ from django.core.serializers import deserialize, serialize
 from django.db.models.query_utils import Q
 from django.urls import reverse
 
+from dojo.celery import app
+from dojo.decorators import dojo_async_task
 import dojo.finding.helper as finding_helper
 import dojo.jira_link.helper as jira_helper
 from dojo.importers.base_importer import BaseImporter, Parser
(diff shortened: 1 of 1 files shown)


FirePanda169:
There is also a suggestion for optimization. Use asynchronous import only if more than 1 task will be created. For one chunk, it will definitely be useful.
Perhaps make the threshold 2 * ASYNC_FINDING_IMPORT_CHUNK_SIZE.
What do you think?

`if settings.ASYNC_FINDING_IMPORT:`
rewrite to
`if settings.ASYNC_FINDING_IMPORT and settings.ASYNC_FINDING_IMPORT_CHUNK_SIZE < len(parsed_findings):`

FirePanda169:
it's sad because importing a lot of findings becomes a problem without a solution.
I tried to come up with something to use `Finding.objects.bulk_create(...)`. But the rewritten save method doesn't let me do it easily.

valentijnscholten:
This function is not intended to be async. To do Async import you can set `DD_ASYNC_FINDING_IMPORT` to `True`. But that feature has always been experimental and is now depracated.

Maffooch:
@FirePanda169 after talking amongst the other moderators, we have decided to move forward with deprecating this feature. Please make plans to disable async import

PR #11809, opened 2025-02-12
fix / cleanup paths in integration tests

valentijnscholten:
Remove unused variables and fix the error(s) in integration tests caused by Ruff linting PRs.

```
integration-tests-1  | ======================================================================
integration-tests-1  | ERROR: test_import_same_engagement_tests (__main__.CloseOldDedupeTest.test_import_same_engagement_tests)
integration-tests-1  | ----------------------------------------------------------------------
integration-tests-1  | Traceback (most recent call last):
integration-tests-1  |   File "/app/tests/base_test_class.py", line 24, in wrapper
integration-tests-1  |     return func(self, *args, **kwargs)
integration-tests-1  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
integration-tests-1  |   File "/app/tests/close_old_findings_dedupe_test.py", line 141, in test_import_same_engagement_tests
integration-tests-1  |     driver.find_element(By.ID, "id_file").send_keys(self.relative_path + "/dedupe_scans/dedupe_endpoint_1.xml")
integration-tests-1  |                                                     ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
integration-tests-1  | TypeError: unsupported operand type(s) for +: 'PosixPath' and 'str'
```

### DIFF
--- a/docker/entrypoint-integration-tests.sh
+++ b/docker/entrypoint-integration-tests.sh
@@ -251,7 +251,7 @@ else
     else
         fail "$test"
     fi
-    
+
     test="False Positive History tests"
     echo "Running: $test"
     if python3 tests/false_positive_history_test.py ; then
--- a/docs/content/en/open_source/upgrading/2.44.md
+++ b/docs/content/en/open_source/upgrading/2.44.md
@@ -14,6 +14,6 @@ The Burp parser now has a custom deduplication configuration to make deduplicati
 This command has various command line arguments to tweak its behavior, for example to trigger a run of the deduplication process.
 See [dedupe.py](https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/management/commands/dedupe.py) for more information.
 
---- 
+---
 
 Check the [Release Notes](https://github.com/DefectDojo/django-DefectDojo/releases/tag/2.44.0) for the contents of the release.
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -1585,6 +1585,7 @@ def get_foreign_keys():
 
 
 def csv_export(request):
+    logger.debug("starting csv export")
     engagements, test_counts = get_engagements(request)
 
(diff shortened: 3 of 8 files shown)


valentijnscholten:
@manuel-sommer @kiblik Could you take a look? I just converted it to `str` to make it work, but maybe the official end goal for Ruff is to have some magic done without converting to `str` but using `/`?

kiblik:
For the processing of paths in Python, it is recommended to use `Path` and related operations (like `/` instead of `+`).
And change it to string as the latest operation: `.send_keys(str(Path(...) / "abc.def" ))`
But you need to be careful, first character of the string after `/` should be non-`/` (sounds a bit complicated). Example is the best:
`.send_keys(self.relative_path + "/dedupe_scans/dedupe_endpoint_1.xml")`
should not be 
`.send_keys(self.relative_path / "/dedupe_scans/dedupe_endpoint_1.xml")`
but
`.send_keys(self.relative_path / "dedupe_scans" / "dedupe_endpoint_1.xml")`

valentijnscholten:
And will that return a str or a Path? Will send_keys accept it?

Would you be willing to make one more Ruff PR to make it look like you/Ruff wants it to be? I have a bunch of other things first to work on.

valentijnscholten:
I made some changes to make it work. The path in `import_scanner_test` is changed into a `str` as the code is doing some [:5] stuff on it (but this test is disabled anyway).

Maffooch (approved):

kiblik (changes requested):
Some adjustments to use `Path` properly.

mtesauro (approved):
Approved

valentijnscholten:
Thanks @kiblik

dogboat (approved):

manuel-sommer (approved):

PR #11810, opened 2025-02-12
[Experiment] Switch to Chromium for Integration tests

valentijnscholten:
I ran into this when working on the `arm64` builds. Chrome is not built for arm64 so we cannot download/install Chrome + Chromedriver in our docker images. Easy solution would be to switch to Chromium, which is built for arm64.

But some of the tests fail, just leaving this PR here as possible future improvement or when we decide it's worth pursuing/investing time in.


valentijnscholten:
Not planned for now.

PR #11812, opened 2025-02-12
Allow tab to "Show Password" within Login Page

kevin-vuong99:
**Description**

If a user is using only the keyboard, they are not able to tab on "Show Password" within the login page.

**Impact on users**

People who are blind (who cannot use devices such as mice that require eye-hand coordination).
People with low vision (who may have trouble finding or tracking a pointer indicator on screen).
Some people with hand tremors find using a mouse very difficult and therefore usually use a keyboard.

**Steps to reproduce**

When you start tabbing within the login page, from the top (URL), Focus goes to Username and then password, and then Login, bypassing show Password.

**Remediation**

Have the "Show Password" keyboard operable by allowing tabbing on the element to be gained.

Image below shows "Show Password" being tabbed over:
image

Instead of just using the div tag and text, I've added a button since it supports tabbing with a keyboard. I added a CSS id selector and rules to match the previous look of the page. The bold tags for "Show Password" were also moved to the CSS rule block.

### DIFF
--- a/dojo/static/dojo/css/dojo.css
+++ b/dojo/static/dojo/css/dojo.css
@@ -1832,6 +1832,14 @@ small {
     min-width:200px;
 }
 
+#toggleBox {
+    background-color: transparent!important;
+    border: none!important;
+    font-weight: bold!important;
+    position: absolute!important;
+    left: 0!important;
+}
+
 .EasyMDEContainer .table {
     width: unset!important;
 }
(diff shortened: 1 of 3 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

hblankenship (approved):

blakeaowens (approved):

Issue #11814, opened 2025-02-13
Broken endpoints on AWSSecHub Parsers Guardduty & Inspector


quirinziessler:
When importing AWS Securityhub Findings from Guardduty & Inspector, the hosts are being saved as invalid strings.

Example: `AwsEc2Instance arn:aws:ec2:eu-central-1:12345678910:instance/i-12aa12aa12aa12aa`
A correct hostname would be: `AwsEc2Instance_arn_aws_ec2_eu-central-1_12345678910_instance_i-12aa12aa12aa12aa`

Solution: Adding a underline between component_name & in the resource ID replacing ":" and "/" by underlines. I will submit a PR for this in a few minutes.

![Image](https://github.com/user-attachments/assets/e9e67f68-3db1-4d68-9c94-34c776227cf8)

![Image](https://github.com/user-attachments/assets/eeb20dcf-5c3f-4676-8815-2f298132a399)

manuel-sommer:
I guess this can be closed @quirinziessler

quirinziessler:
> I guess this can be closed [@quirinziessler](https://github.com/quirinziessler)

Let's wait for the release. Until then the issue still affects all users.

quirinziessler:
Fixed by last release.

PR #11815, opened 2025-02-13
:bug: fix AWS Endpoints broken

quirinziessler:
fix for #11814


Maffooch:
@quirinziessler is the main issue here that endpoints cannot be edited in the UI after they are broken? If so, I am not sure if fixing this parser bug would be worth creating duplicate endpoints for all users using these parsers. For example, if users had 100 endpoints with ARNs containing colons, now they will have another 100 endpoint with underscores instead

quirinziessler:
@Maffooch the main issue is that within imports errors are thrown and running into an HTTP 500 in the worst case (depending on finding amount). Furthermore the endpoints can not be migrated.

Would a migration script solve your concerns?

quirinziessler:
@Maffooch any news / decision here?

quirinziessler:
> Hi @quirinziessler yes, I think a migration would definitely help users out. Thanks!

done @Maffooch

dojo/tools/aws_inspector2/parser.py
```diff
@@ -188,7 +188,7 @@ def process_endpoints(self, finding: Finding, raw_finding: dict) -> Finding:
             resource_type = resource_info.get("type", None)
             resource_id = resource_info.get("id", "N/A")
             resource_details = resource_info.get("details", {})
-            endpoint_host = f"{resource_type} - {resource_id}"
+            endpoint_host = f"{resource_type}_{resource_id}".replace(":", "_").replace("/", "_")
```
manuel-sommer, line 191: Could you make an info about this also please in the documentation?

manuel-sommer (changes requested):

manuel-sommer:
Would be awesome if we could get this on the road until next release @mtesauro and @Maffooch

dojo/db_migrations/0222_aws_sechub_update_endpoints.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 5.0.8 on 2024-09-12 18:22
+
+from django.db import migrations
+
+
+def aws_sechub_update_endpoints(apps, schema_editor):
+    endpoint_model = apps.get_model('dojo', 'Endpoint')
+    endpoints = endpoint_model.objects.filter(host__contains="arn:aws")
```
Maffooch, line 8: This should filter against findings with a test type of only the AWS things you have modified in this PR

quirinziessler: @Maffooch How shall I do so? The endpoints don't have a property test type as far as I see.

Maffooch: That is true. However,  I think it would be preferable to have a partial migration of endpoints that were explicitly intended to change versus a full migration of endpoints that should not have changed. 

For example, let's say Trivy generates ARNs with colons and slashes. If those endpoints were updated, but the Trivy parser was still producing ARN endpoints with colons and spaces, there would be a lot of "duplicate" endpoints

I would do something like this to filter for endpoints from the AWS tools that you are updating:
```python
Endpoint.objects.filter(finding__test__test_type__name__in=["AWS test type 1", "AWS test type 2", ...])
```

quirinziessler:
My rebase got messed up. Will create a fresh PR.

quirinziessler:
New PR: #11902

PR #11830, opened 2025-02-16
arm64 builds: unit tests

valentijnscholten:
**Description**

This PR takes another step towards publishing arm64 images (continuing on from #11673)
Most tutorials for multi platform builds just add a few lines to install Qemy and add `-platforms: linux/amd64,linux/arm64` and are done. Unfortunately our nice python wheels such as `uwsgi` don't get build successfully under Qemu.
So we have run the builds natively on Githubs  `arm64` runners.

Steps taken (before I knew that native runners was the only way):
- Build locally for `arm64` with Qemu -> FAILED
- Try to switch our builds to install upstream binaries for `uwsgi` so we don't have to build them -> FAILED
- Build on a AWS t4g instance -> FAILED (With PyOpenSSL error)
- Build on a GitHub ARM runner -> FAILED (With PyOpenSSL error)

At first the PyOpenSSL error seemed related to ARM64 as I had no other reports.
So I thought it was a good idea to run the unit tests also on `arm64` runners to make sure they work.
Afterwards I found out the PyOpenSSL error was also present on `amd64` and I fixed that.
Now that we have unit tests working in GitHub actions anyway, I think it's good to have them running.
At least for a while during the next couple of releases to see if anything pops up on these `arm64` builds.

The integration test are not run for `arm64`. These tests are fully based on Chrome for which there is no `arm64` build.
I tried to switch the integration tests to Chromium, but that would need some more work: #11810 and might not be worth it.

**Test results**
GitHub usually doesn't take workflow definitions from forks, so I created a branch in the Defect Dojo repository.
GitHub might still not pick up the workflow changes as it may look at `master`. Let's see if we have to wait until the next monthly release.

**Next steps (in another PR)** 
- Publish the arm64 images during our releases


Maffooch (approved):

mtesauro (approved):
Approved

Maffooch:
@valentijnscholten I would also like a 25% refund on my monitor 😂

hblankenship (approved):

dogboat (approved):
25% coupon applied.

PR #11831, opened 2025-02-16
jira issue type migration quickfix

### LINKED ISSUE #11821: Make extra JIRA Issue Type work without database schema migrations (DD_JIRA_EXTRA_ISSUE_TYPES )
Currently changing the list of available JIRA Issue types via `DD_JIRA_EXTRA_ISSUE_TYPES` will trigger database schema migration in Django. This is not ideal, especially since #11638. 

This least if configuration data and should not be tied to the database schema. I suggest to put the issues types in their own table where they can be managed via the UI/API. A foreign key can be used to guard integrity.
This change may involve a (one-time?) data migration that reads the DD_JIRA_EXTRA_ISSUE_TYPES and puts them into the new table.

Adding back in the `makemigrations` will sometimes solve it, but will leave users with a migration chain that diverges from the official builds.

**Additional context**
A workaround for users wanting to add extra JIRA issue types is to:
- set `DD_JIRA_EXTRA_ISSUE_TYPES` as documented
- also add the extra issue type(s) to the list in [0182_alter_jira_instance_default_issue_type](https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/db_migrations/0182_alter_jira_instance_default_issue_type.py) and to `0027_jira_issue_type_settings.py`. This requires a rebuild of the `django` container / services using that, i.e. `uwsgi`, `celerybeat`, `celeryworker`, `initializer`.

Raised on Slack recently: https://owasp.slack.com/archives/C2P5BA8MN/p1739483830256399

valentijnscholten:
Fixes #11821 (for now the easy way).

A proper fix may take some time, this fix will allow users to add extra JIRA Issue Types again via `DD_JIRA_EXTRA_ISSUE_TYPES`

Usually you shouldn't edit existing migration files. But in this case these "choices" have no effect on the actual database model. Django ORM doesn't propagate these choices to the database. It performs the validation in Python code.


Maffooch (approved):

mtesauro (approved):
Approved

hblankenship (approved):

dogboat (approved):

PR #11840, opened 2025-02-17
fix(ui): Increase readability in "Members" list

kiblik:
Until the table was showing only the name of the ProdType but there was a chance for misunderstanding that is the name of ProdType and permissions are inherited. Now it is explicitly mentioned next to name of ProdType


Maffooch (approved):

mtesauro (approved):
Approved

hblankenship (approved):

dogboat (approved):

PR #11856, opened 2025-02-19
Added docstring for the burp parser

Jino-T:
**Description**

I am a Defect Dojo intern and I wrote a docstring for the burp parser to describe the fields it returns and to also describe the fields used for dedupe.


Maffooch (commented):
Love this! My advice for this parser (and others in the future) would be to the move these function to the top of the file so that they are more easily found by folks in the future

PR #11860, opened 2025-02-20
feat(helm): Add secret annotations

al-cheb:
**Description**

Allow adding annotations to secret resources. 

E.g.: A Kubernetes mutating webhook that makes direct secret injection using annotations - https://bank-vaults.dev/docs/mutating-webhook/annotations/

### DIFF
--- a/helm/defectdojo/templates/secret.yaml
+++ b/helm/defectdojo/templates/secret.yaml
@@ -16,6 +16,12 @@ metadata:
     helm.sh/resource-policy: keep
     helm.sh/hook: "pre-install"
     helm.sh/hook-delete-policy: "before-hook-creation"
+    {{- with .Values.secretAnnotations }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+    {{- with .Values.annotations }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
 type: Opaque
 data:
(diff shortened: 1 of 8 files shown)


helm/defectdojo/values.yaml
```diff
@@ -131,6 +131,9 @@ monitoring:
 
 annotations: {}
 
+# Add annotations for secret resources
+secretAnnotations: {}
```
kiblik, line 135: I would change it to the following, to have it more future-proof (in case we will need to customize secrets more):
```suggestion
secrets:
  # Add annotations for secret resources
  annotations: {}
```

helm/defectdojo/templates/secret.yaml
```diff
@@ -16,6 +16,12 @@ metadata:
     helm.sh/resource-policy: keep
     helm.sh/hook: "pre-install"
     helm.sh/hook-delete-policy: "before-hook-creation"
+    {{- with .Values.secretAnnotations }}
```
kiblik, line 19:
```suggestion
    {{- with .Values.secrets.annotations }}
```
See note in `values.yaml`

kiblik (approved):

Maffooch (approved):

cneill (approved):

PR #11717, opened 2025-02-02
Ruff: Add and fix PLR6201

kiblik:
Add rule [literal-membership (PLR6201)](https://docs.astral.sh/ruff/rules/literal-membership/#literal-membership-plr6201) and fix it

PR #11862, opened 2025-02-20
Surveys: Correct Question 404

Maffooch:
When editing a survey question, a 404 is presented for a valid object. At some point, the content type for Questions changed to `Defect Dojo` (the verbose name of the app) rather than `dojo` (the common name)

There is only one place where the name of the content type is accessed, so adding some backward compatible checks corrected the issue

[sc-10195]

### DIFF
--- a/dojo/survey/views.py
+++ b/dojo/survey/views.py
@@ -539,17 +539,17 @@ def edit_question(request, qid):
                 extra_tags="alert-info")
     type = str(ContentType.objects.get_for_model(question))
 
-    if type == "dojo | text question":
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

cneill (approved):

hblankenship (changes requested):
Would it make sense to switch this to the set notation {} as per kiblik's recently approved PR: https://github.com/DefectDojo/django-DefectDojo/pull/11717

Maffooch:
> Would it make sense to switch this to the set notation {} as per kiblik's recently approved PR: #11717

Good catch! That would have resulted in some conflicts at release time

PR #11891, opened 2025-02-24
:arrow_up: Bump urllib3 to 1.26.20

manuel-sommer:
https://www.cvedetails.com/cve/CVE-2024-37891/ 
" Users are advised to update to either version 1.26.19 or version 2.2.2"

https://github.com/DefectDojo/django-DefectDojo/pull/11455

PR #11892, opened 2025-02-24
:tada: implement trivy scanner

manuel-sommer:
CVE-2024-37891 --> https://github.com/DefectDojo/django-DefectDojo/pull/11891
CVE-2024-6484 --> https://github.com/DefectDojo/django-DefectDojo/actions/runs/13506024016/job/37735642044?pr=11892 --> components/yarn.lock (bootstrap 3.4.1)

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -12,7 +12,7 @@ jobs:
       uses: actions/checkout@v4
 
     - name: Run Trivy vulnerability scanner in fs mode
-      uses: aquasecurity/trivy-action@0.28.0
+      uses: aquasecurity/trivy-action@0.29.0
       with:
         scan-type: 'fs'
         scan-ref: '.'
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -1,0 +1,19 @@
+name: Trivy
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
+jobs:
+  build:
+    name: Vulnerability Scanner
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Run Trivy vulnerability scanner in fs mode
+      uses: aquasecurity/trivy-action@0.28.0
+      with:
+        scan-type: 'fs'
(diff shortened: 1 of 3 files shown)


manuel-sommer:
Feedback welcome @valentijnscholten @Maffooch @mtesauro

manuel-sommer:
Rebased because of https://github.com/DefectDojo/django-DefectDojo/pull/11891

valentijnscholten:
I understand that it could be useful to scan applications with Trivy for vulnerabilities. What exact use case  / process would we follow here and what goal do we want to achieve? Does it generate a report that is visible on the PR or Security Tab? Will it show a difference between existing vulnerabilities and new ones introduced by the PR? Or is it just a tool that generates some output that may or may not assist us in making decisions?

mtesauro:
I wonder if this would be something better to run outside of PRs.

Here's why.

I'm a new contributor who writes a new parser for DefectDojo. It just so happens that libFoo.js has a vulnerability discovered right before I do my PR and it's used to create the charts in DefectDojo's metrics pages. I fail this test which has NOTHING to do with the code I'm trying to contribute.

I'm just worried we're going to cause pain to our community for things that didn't do and don't have control over. 

Honestly, I'd rather write a cron job that pulls the containers we have in Docker Hub and runs Trivy and whatever else against those containers and pushes the results to a DefectDojo install. It might surprise you but DefectDojo Inc actually runs DefectDojo internally to manage our scan data. I'd be happy to add this to our automation and create PRs to update things that show up.

The key difference is I'm choosing to take on that effort vs it getting dropped in the lap of a community contributor.

Does that make sense?

manuel-sommer:
Thank you for the clarification @mtesauro. It does not surprise me that you use also DefectDojo yourself. 
I did not look into it, if this it technically possible, but would it make sense to run trivy ci checks only on PRs from renovate and dependabot to get instant feedback?

Maffooch:
> would it make sense to run trivy ci checks only on PRs from renovate and dependabot to get instant feedback?

Talking internally with Matt, we believe this would likely still be more churn than we would benefit from - even when running Trivy only on package updates from renovate/dependabot

PR #11898, opened 2025-02-25
ms_defender: Improve reliability and error handling

### LINKED ISSUE #11896: MS Defender parser doesn't add endpoints
**Slack us first!**
The easiest and fastest way to help you is via Slack. There's a free and easy signup to join our #defectdojo channel in the OWASP Slack workspace: [Get Access.](https://owasp.org/slack/invite)
If you're confident you've found a bug, or are allergic to Slack, you can submit an issue anyway.

**Be informative**
Please enter as much information as possible, otherwise we can't provide support. If possible upgrade to the latest release or dev version and try again.

**Bug description**
When trying to import recent API outputs from MS Defender, zipping them as documented [here](https://docs.defectdojo.com/en/connecting_your_tools/parsers/file/ms_defender/), findings are imported but no endpoints are created

**Steps to reproduce**
1. Go to [Demo](https://demo.defectdojo.org/)
2. Import the attached zip as "MS Defender" import type
3. Check for imported findings
4. Notice no endpoints created

**Expected behavior**
Endpoints should be created using machines.json contents

Doesn't matter

**Environment information**
Doesn't matter

[defender.zip](https://github.com/user-attachments/files/18960957/defender.zip)

valentijnscholten:
Fixes #11896 

Exceptions such as KeyErrors were silently swallowed leading to endpoints not being created for findings.

This PR:
- Logs the exception
- Greatly reduces the chances of exceptions by using `.get` to retrieve values from the dict/json object
- Added test case with a report that was triggering this because of a missing `defenderAvStatus` field.

I tried to remove the `str()` wrappers everywhere, but it seems they are needed to handle the "" vs null vs None cases with these json reports. So I left them for now.


manuel-sommer (approved):
looks good to me

mtesauro (approved):
Approved

hblankenship (approved):

dogboat (approved):

cneill (approved):

PR #11923, opened 2025-03-01
SLA Calculations: Add unit tests to capture current behaviour

valentijnscholten:
Adds some unit tests to capture current behaviour:
- to understand the calculation that does some magic with days remaining
- to make sure we don't regress in the future if we simplify the calculation, or add business days support.


Maffooch (approved):

blakeaowens (approved):

dogboat (approved):

mtesauro (approved):
Approved

PR #11942, opened 2025-03-04
docker entrypoints: use bash everywhere

valentijnscholten:
Some entrypoint scripts where using `sh` which doesn't recognize `if [[...` used in places like `wait_for_database_to_be_reachable`

Fixes these errors on startup:
```
/entrypoint-uwsgi.sh: 21: [[: not found
```

Credits to @9alexx3 on Slack for suggesting this fix.

### DIFF
--- a/docker/entrypoint-unit-tests-devDocker.sh
+++ b/docker/entrypoint-unit-tests-devDocker.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # Run available unittests with a setup for local dev:
 # - Make migrations and apply any needed changes
 # - Leave container up after running tests to allow debugging, rerunning tests, etc.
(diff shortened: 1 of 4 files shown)


kiblik:
I would like to know how it is possible that the tests passed.

valentijnscholten:
I don't know all the intricate details of how `sh` and `bash` handle errors with or without `set -e` and within or outside of subshells and within or outside of functions.

Looking at this post: https://stackoverflow.com/questions/4072984/how-do-i-get-the-effect-and-usefulness-of-set-e-inside-a-shell-function I don't think I want to know :-)

mtesauro (approved):
Approved

Maffooch (approved):

cneill (approved):

dogboat (approved):

PR #11950, opened 2025-03-05
Ruff: Add and autofix B009

kiblik:
Add ruff rule [get-attr-with-constant (B009)](https://docs.astral.sh/ruff/rules/get-attr-with-constant/#get-attr-with-constant-b009) and use autofix


Maffooch (approved):

mtesauro (approved):
Approved

hblankenship (approved):

PR #11952, opened 2025-03-05
Ruff: Add and fix B905

kiblik:
Add rule [zip-without-explicit-strict (B905)](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/#zip-without-explicit-strict-b905) and fix it


Maffooch (approved):

mtesauro (approved):
Approved

hblankenship (approved):

dogboat (approved):

PR #11953, opened 2025-03-05
Ruff: Add and autofix B010

kiblik:
Add [set-attr-with-constant (B010)](https://docs.astral.sh/ruff/rules/set-attr-with-constant/#set-attr-with-constant-b010) and autofix it


Maffooch (approved):

mtesauro (approved):
Approved

dogboat (approved):

PR #11964, opened 2025-03-06
dedupe managament command: ensure ordering

### LINKED ISSUE #11948: Deduplication dedupe the oldest vulnerability
**Be informative**
I am running DefectDojo on Docker via Portainer. I set a env var to duplicate my test 'Pen Test' using Unique Id From Tool (that I calculate using a self-made python library to use API). My Unique Id is calculated by: Endpoint (host,port,protocol), IP Address and Vulnerability Title

**Bug description**
 Duplication seems work, but DefectDojo set as "Duplicate" the oldest vulnerability (with a lower id and date) instead of the newest.

**Expected behavior**
The newest vulnerability is marked as "Duplicated"

- [x ] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Debian GNU/Linux 12 (bookworm)
 - Docker Compose version: 1
 - DefectDojo version (see footer): v. 2.43.3 ( release mode )
 - 
**Screenshots**
In this example, they have same UNIQUE ID FROM TOOLS but some differences about Severity, Date and descriptions. They are in 2 different Engagements.

Image

How can I fix this?

valentijnscholten:
**Description**
Fixes #11948 
For deduplication it's important to start with the newest findings first to make sure the old(est) existing finding is selected as the original. This is already working correctly inside Defect Dojo, but not in the `dedupe` management command. This PR fixes that.

### DIFF
--- a/dojo/management/commands/dedupe.py
+++ b/dojo/management/commands/dedupe.py
@@ -61,6 +61,9 @@ class Command(BaseCommand):
             findings = Finding.objects.all().filter(id__gt=0)
             logger.info("######## Will process the full database with %d findings ########", findings.count())
 
+        # process in reverse order to ensure the newer findings are marked as duplicate of older findings
+        findings.order_by("-id")
+
         # Phase 1: update hash_codes without deduplicating
         if not dedupe_only:
             logger.info("######## Start Updating Hashcodes (foreground) ########")


Maffooch (approved):

mtesauro (approved):
Approved

hblankenship (approved):

dojo/management/commands/dedupe.py
```diff
@@ -61,6 +61,9 @@ def handle(self, *args, **options):
             findings = Finding.objects.all().filter(id__gt=0)
             logger.info("######## Will process the full database with %d findings ########", findings.count())
 
+        # process in reverse order to ensure the newer findings are marked as duplicate of older findings
+        findings.order_by("-id")
```
dogboat, line 65: Don't we need to reassign this back to the `findings` var?

dogboat (changes requested):

dojo/management/commands/dedupe.py
valentijnscholten: hmmm yes we should. But the user running into the problem confirmed the fix is working.
But I also notice that further down we already have "desc" ordering for dedupe.
I wonder if it is just a timing issue as the dedupe by default happens in the background.
Let me check what the queries are without the changes in this PR.

valentijnscholten: Ordering is already DESC without the changes in this PR, so no need for the PR. I'll check with the reporer.

valentijnscholten:
Closing this as the ordering already is in DESC order, no need for the PR (see above).

PR #11965, opened 2025-03-06
arm64: Publish arm64 builds for releases

valentijnscholten:
**Description**
This PR (Finally) adds `arm64` builds to our releases. Users can now happily deploy on AWS `t4g`or GCP `C4A` based clusters.

The PR is a little more complex because some Python wheels just wouldn't compile under `Qemu`. So we are building the images on a native `arm64` runner. After that we have to merge the container digests to create a multiplatform index difest for Docker Hub. In the end it's better to have a native build anyway!

Changes in this PR:
- Add an outer strategy loop to build for both àmd64` and `arm64`. By not combining it in the innter strategy loop we run the inner strategy on different runners via a parameter.
- Change the parameters to be `linux/amd64` and `linux/arm64` which are more recognizable.
- Separate the the repo names in GITHUB_ORG and DOCKER_ORG to avoid confusion.
- Simplify difference in steps between Debian and Alpine.
- Remove some unneeded copying of environment variables from GITHUB_ENV into env

Please note tags can no longer be set on the build and push step, but are set on the index digest after/while merging the container manifests.

**Test results**
I did testing in my own fork and deployed the builds on a `t4g` in AWS. You can try yourself by using:

- `valentijnscholten/defectdojo-nginx:2.44.0-dev-valentijn`
- `valentijnscholten/defectdojo-django:2.44.0-dev-valentijn`

One way to test this PR at release time is by NOT merging it yet, but selecting the `arm64-unit-tests` branch as the source for the workflow definitions. This way the release will still be performed as normal, but it will use the new workflows. If something goes wrong, we can rerun/continue using the old workflows from `master` as usual. 

**Documentation**
The release process steps are still the same.

I am happy to perform the next release and/or be on standby during the next release to fix any issues that may arise.


Maffooch (approved):

hblankenship (approved):

dogboat (approved):

mtesauro (approved):
Approved

PR #11986, opened 2025-03-10
update permissions documentation links to reflect correct paths

blakeaowens:
The links to existing permissions/groups documentation (e.g. [https://documentation.defectdojo.com/usage/permissions/](https://documentation.defectdojo.com/usage/permissions/)) no longer exist.

This pull request updates documentation links for Product Type and Product membership, as well as Group Membership in the User, Group, Product Type, and Product views.

[sc-10454]

### DIFF
--- a/dojo/templates/dojo/view_group.html
+++ b/dojo/templates/dojo/view_group.html
@@ -45,7 +45,7 @@
                 <div class="clearfix">
                     <h4 class="pull-left">Members of this Group</h4>
                     &nbsp;
-                    <a href="https://documentation.defectdojo.com/usage/permissions/#groups" target="_blank">
+                    <a href="https://docs.defectdojo.com/en/customize_dojo/user_management/create_user_group/#manage-a-groups-users" target="_blank">
                         <i class="fa-solid fa-circle-question"></i></a>
                     {% if group|has_object_permission:"Group_Manage_Members" %}
                     <div class="dropdown pull-right">
@@ -118,7 +118,7 @@
                 <div class="clearfix">
                     <h4 class="pull-left">Product Types this Group can access</h4>
                     &nbsp;
-                    <a href="https://documentation.defectdojo.com/usage/permissions/" target="_blank">
+                    <a href="https://docs.defectdojo.com/en/customize_dojo/user_management/create_user_group/#add-product-roles-or-product-type-roles-for-a-group" target="_blank">
                         <i class="fa-solid fa-circle-question"></i></a>
                     {% if request.user.is_superuser %}
                     <div class="dropdown pull-right">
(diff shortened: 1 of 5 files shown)


dogboat (approved):

mtesauro (approved):
Approved

Maffooch (approved):

dojo/templates/dojo/view_group.html
hblankenship, line 1: Should the two links that go to the same location be combined? I initially found it confusing that I was in the same place until I realized it was under one umbrella. 

"Product Types and Products this Group can access"

dojo/templates/dojo/view_group.html
```diff
@@ -118,7 +118,7 @@ <h4 class="pull-left">Members of this Group</h4>
                 <div class="clearfix">
                     <h4 class="pull-left">Product Types this Group can access</h4>
                     &nbsp;
-                    <a href="https://documentation.defectdojo.com/usage/permissions/" target="_blank">
+                    <a href="https://docs.defectdojo.com/en/customize_dojo/user_management/create_user_group/#add-product-roles-or-product-type-roles-for-a-group" target="_blank">
```
hblankenship, line 121: Should the two links that go to the same location be combined? I initially found it confusing that I was in the same place until I realized it was under one umbrella.

"Product Types and Products this Group can access"

dojo/templates/dojo/view_user.html
```diff
@@ -106,7 +106,7 @@ <h4 class="pull-left">{% trans "Contact Information" %}</h4>
                 <div class="clearfix">
                     <h4 class="pull-left">{% trans "Product Types this User can access" %}</h4>
                     &nbsp;
-                    <a href="https://documentation.defectdojo.com/usage/permissions/" target="_blank">
+                    <a href="https://docs.defectdojo.com/en/customize_dojo/user_management/about_perms_and_roles/#productproduct-type-membership--roles" target="_blank">
```
hblankenship, line 109: Same here?

dojo/templates/dojo/view_group.html
blakeaowens: No, they should not. They are two separate places in the UI to list different types of group relationships. One is listing Product Type Groups, and the other is listing Product Groups. The link does take you to the same location in the documentation for both sections, but that is because the documentation describes Product Roles and Product Type Roles for Groups in one place.

Screenshot 2025-03-13 at 11 39 13 AM

dojo/templates/dojo/view_user.html
blakeaowens: Same for here, nothing can be combined. There is one documentation link that explains Product and Product Type membership for Users, but two separate places in the UI to add Product and/or Product Type membership to a user (which cannot be combined).

hblankenship (approved):

PR #12002, opened 2025-03-12
Ruff: Add TD007 and N803

manuel-sommer:


kiblik:
Rule [description](https://docs.astral.sh/ruff/rules/invalid-argument-name/) says 
> Checks for argument names that do not follow the snake_case convention.

I would also follow `snake_case` if possible because simple lowering of `camelCase` to `flatcase` makes it harder to read.

mtesauro (approved):
Approved

dogboat (approved):

kiblik (approved):

PR #12010, opened 2025-03-14
fix(helm-metrics): Flag format for promExporter changed

kiblik:
For `nginx-prometheus-exporter` from `v1.0.0`, the flag format was changed. Old is still not deprecated, but there is already a warning:
```
the flag format is deprecated and will be removed in a future release, please use the new format: --nginx.scrape-uri
```


Maffooch (approved):

mtesauro (approved):
Approved

hblankenship (approved):

PR #12014, opened 2025-03-14
DOCKER.md: use docker compose everywhere

valentijnscholten:
Some example commands where using `docker ...` instead of `docker compose`.

### DIFF
--- a/readme-docs/DOCKER.md
+++ b/readme-docs/DOCKER.md
@@ -302,10 +302,9 @@ docker compose up
 ```
 Then
 ```
-docker ps
-#find the name of the uwsgi container from the above command
-docker exec -ti [container-name] bash
+docker exec -it uwsgi /bin/bash
 ```
+You're now inside the container.
 Rerun all the tests:
 
 ```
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

readme-docs/DOCKER.md
```diff
@@ -286,7 +285,7 @@ If you want to enter the container to run more tests or a single test case, leav
 docker/setEnv.sh dev
 docker compose up
 ```
-Then 
+Then
 ```
 docker ps
```
dogboat, line 290: Curious if this one (and the `docker exec` below) need to be changed as well. Approving this so if not, just ignore me. :)

dogboat (approved):

readme-docs/DOCKER.md
valentijnscholten: good point. wasn't even aware these could be done with compose as well. i've added the compose way but also left the plain docker way as that will give you a image id which might be helpful. does that make sense?

hblankenship (approved):

PR #12047, opened 2025-03-18
Fix the accessibility issue in the login page: Labels or instructions not provided when content requires user input

oussama-taoufiq:
Page / Screen Title
Defect Dojo login
Page URL / Screen ID
https://demo.defectdojo.org
Error Title
Labels or instructions not provided when content requires user input
Error Severity
Serious
Status
Fail
Accessibility Issue
[Description of issue] The required fields such as "Username" and "Password" are mandatory in order to login. The only visual indicator that a user assume that its required, are the "*".

[Impact on users] User might not know that the fields are required.

[Pattern] On the login page, the 2 label for the input field.

[Steps to reproduce]
Please see screenshot

[Sample of code] <label class="col-sm-2 control-label " for="id_username">Username<sup>*</sup></label>

Remediation
[Recommendation] Please have the word (required) within the label of the input and put the "*" and required in red.

[Example of Compliant Code] <div class="wb-frmvld"> <form action="#" method="get" id="validation-example"> ... <div class="form-group"> <label for="fname1" class="required"><span class="field-name">First name</span> <strong class="required">(required)</strong></label> <input class="form-control" id="fname1" name="fname1" type="text" autocomplete="given-name" required="required" data-rule-minlength="2" /> </div>

[Additional Resources] https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G184 https://wet-boew.github.io/v4.0-ci/demos/formvalid/formvalid-en.html

Impacted Users
Users of Assistive Technologies (AT)
Keyboard-only users

Screenshot

image

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -1,0 +1,4923 @@
+{
+  "name": "doks",
+  "version": "0.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "doks",
+      "version": "0.0.0",
+      "license": "MIT",
+      "dependencies": {
+        "@tabler/icons": "^3.12.0",
+        "@thulite/doks-core": "^1.8.0",
+        "@thulite/images": "^3.3.0",
+        "@thulite/inline-svg": "^1.1.0",
+        "@thulite/seo": "^2.4.1",
+        "thulite": "^2.5.0"
+      },
+      "devDependencies": {
+        "prettier": "^3.3.3",
+        "vite": "^6.0.9"
+      },
+      "engines": {
+        "node": ">=20.11.0"
+      }
+    },
+    "node_modules/@ampproject/remapping": {
+      "version": "2.3.0",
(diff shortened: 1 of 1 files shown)


blakeaowens:
Hello @oussama-taoufiq, would you please migrate this PR to use the `aria-required` attribute instead of the `(required)` text? Documentation on `aria-required` can be [found here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-required).

mtesauro:
> Hello @oussama-taoufiq, would you please migrate this PR to use the `aria-required` attribute instead of the `(required)` text? Documentation on `aria-required` can be [found here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-required).

+1,000 on using aria-required

oussama-taoufiq:
Hello @blakeaowens, @mtesauro, could you tell me where the username and password input are injected into the login page to be able to add the area-required attribute?

Maffooch:
The login for is rendered by the `form_fields.html` snippet. This change would be in the same place that you added the `(required)` label

PR #12048, opened 2025-03-18
Jira Webhook: Prevent finding group findings from being reopened

Maffooch:
When interacting with jira tickets that attached finding groups, we should no be opening any findings on the DefectDojo side because jira has no way of knowing if a finding really should be reopened or not

[sc-10319]


mtesauro (approved):
Approved

dogboat (approved):

blakeaowens (approved):

cneill (approved):

PR #12056, opened 2025-03-20
Generic JSON: Explicitly process tags like other tools

Maffooch:
I noticed that the generic json parser did not make use of the `unsaved_tags` field like other parsers to. We should be consistent here

[sc-10594]


mtesauro (approved):
Approved

dogboat (approved):

blakeaowens (approved):

cneill (approved):

PR #12061, opened 2025-03-20
sso docs: make environment variables vs local_settings more explicit

valentijnscholten:
Recently this text was improved. I suggest a further improvement. Mainly to make it more clear that if users opt to put the environment variables inside local_settings.py, they need to remove the `DD_` prefix.

If this turns out to be still not clear for users I think we will have to remove any references to `local_settings.py` here and just point to the open source configuration page which explains this in more detail.


paulOsinski:
Looks good, thank you Val

Maffooch (approved):

mtesauro (approved):
Approved

dogboat (approved):

hblankenship (approved):

PR #12106, opened 2025-03-25
Import Memory Handling: Do not maintain parsed findings long term

Maffooch:
When processing many imports at once, it was noticed that memory of the redis container was maintaining a very high level. This is because the findings parsed from the submitted report were being saved on the Importer class, and in scope for the entire duration of the process. We should not maintain the list of parsed findings after they have been consumed 

[sc-10595]


hblankenship (approved):

dogboat (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #12113, opened 2025-03-27
:tada: Add slackware security advisory to vulnid

manuel-sommer:
It depends, but some use "SSA:" and some use "SSA-" to define the slackware security advisory. This PR covers both.


Maffooch (approved):
Nice job catching that there are two ways to achieve the same thing

dogboat (approved):

hblankenship (approved):

mtesauro (approved):
Approved

PR #12119, opened 2025-03-27
Re-Enable features.md search

paulOsinski:
This feature summary page is useful in search until we get more "Tags" documentation written, so it should not be excluded from search.


hblankenship (approved):

paulOsinski:
Broke it, let me try again

Issue #11833, opened 2025-02-17
SLA Business days calculation not working - DD_SLA_BUSINESS_DAYS


Nsilva07:
DD_SLA_BUSINESS_DAYS should be used to calculate SLA using business days instead of calendar days. 

From dojo/settings/settings.dist.py: 
Use business day's to calculate SLA's and age instead of calendar days
DD_SLA_BUSINESS_DAYS=(bool, False)

**Steps to reproduce**

Define the variable DD_SLA_BUSINESS_DAYS=True using your preferred method. 
Define the SLAs in the application.
Add a new finding. 
Verify the SLA calculation after adding a vulnerability, it will be using calendar days. 

**Expected behavior**
SLA calculation should consider Business days only. 

**Deployment method** *(select with an `X`)*
- [x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version: 2.42.3

**Logs** 
Docker compose logs do not contain errors or any information related with the variable DD_SLA_BUSINESS_DAYS.

Docker compose exec uwsgi env | grep DD_SLA_BUSINESS_DAYS returns DD_SLA_BUSINESS_DAYS=True, indicating that the variable is defined but DefectDojo is not applying it. 

Tested by adding the variable in docker-compose.yml, extra_settings.py, local_settings.py and settings.dist.py.

valentijnscholten:
How are you concluding it doesn't work correctly. Is it that the  `sla_expiration_date` fields are incorrect? Do you have some screenshots?

The feature was implemented in #4260, but it looks like it only applies to the calculation of the findings `age`. At first glance it looks like maybe the business days feature wasn't taken into account when the (later) `sla_expiration_date` field was added in #9473 and #9494.

Could you also take a look @blakeaowens before I dive in?

Nsilva07:
@valentijnscholten thats simple, because i have defined DD_SLA_BUSINESS_DAYS=True using multiple ways, confirmed that the variable is applied via docker compose exec uwsgi env, and in the interface the SLA expiration date always reflect calendar days(new findings after changing the variable, and after editing current findings), in the same way when a finding is pushed to jira the due date reflects calendar days and not business days.

valentijnscholten:
[...]

PR #12131, opened 2025-03-29
Remove non-working DD_SLA_BUSINESS_DAYS feature to avoid confusion

valentijnscholten:
As reported in #11833 the SLA Business days feature is nog longer working. There are currently no plans to reintroduce it. This PR removes left over code and configuration settings to avoid confusion.
If the feature returns, it will probably be part of the SLA Configuration settings so it can re-use the existing recalculation logic on configuration changes.
Please follow #11833 for future developments.

### DIFF
--- a/docs/content/en/open_source/upgrading/2.46.md
+++ b/docs/content/en/open_source/upgrading/2.46.md
@@ -5,7 +5,8 @@ weight: -20250407
 description: Tag Formatting Changes + Import Payload Changes
 ---
 
-### Tag Formatting Update
+
+## Tag Formatting Update
 
 Tags can no longer contain the following characters:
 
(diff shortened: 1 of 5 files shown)


Maffooch (changes requested):
The PR looks good overall. I think this should wait until the 2.46.0 release given how close we are to 2.45.0. It would be preferable to allow as much soak time as we can

Maffooch:
@valentijnscholten please see conflicts on this one and prep for the 2.46.0 release

blakeaowens (approved):

Maffooch (approved):

mtesauro (approved):
Approved

cneill (approved):

PR #12133, opened 2025-03-29
sla_config: use mass update for recalculation

valentijnscholten:
This PR optimizes the updating of the `sla_expiration_date` field. A change to an SLA Configuration or Product can trigger a recalculation of this field for all affected findings. 

Before this PR this took ~8.5 minutes for 20k findings. This is on a i9 16 core with an NVM SSD. 
This PR used the `mass_model_updater` function that has proven to work more efficiently. Now the job completes in ~30s time.

Before this PR, every finding would be post processed in the background celery work. This isn't needed as a change to the SLA expiration doesn't affect dedupe or pushing to JIRA. Also the product grade would be calculated for each affected finding (so multiple times per product). This is not needed. This background work took over 30 minutes for 20k (I killed the instance after 30 minutes).

The PR also removes some code duplication.

I can imagine this change is also beneficial for Pro users where larger instances are more common, as well as more intensive usage of the instances.

### DIFF
--- a/dojo/sla_config/helpers.py
+++ b/dojo/sla_config/helpers.py
@@ -40,9 +40,6 @@ def update_sla_expiration_dates_sla_config_sync(sla_config, products, severities
 
     mass_model_updater(Finding, findings, lambda f: f.set_sla_expiration_date(), fields=["sla_expiration_date"])
 
-    # for f in findings:
-    #     f.save_no_options()
-
     # reset the async updating flag to false for all products using this sla config
     for product in products:
         product.async_updating = False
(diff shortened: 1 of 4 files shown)


dojo/sla_config/helpers.py
```diff
@@ -3,25 +3,53 @@
...
+            "test__engagement__product",
+            "test__engagement__product__sla_configuration",
+    )
+
+    findings = findings.order_by("id").only("id", "sla_start_date", "date", "severity", "test")
+
+    mass_model_updater(Finding, findings, lambda f: f.set_sla_expiration_date(), fields=["sla_expiration_date"])
+
+    # for f in findings:
+    #     f.save_no_options()
```
blakeaowens, line 44:
```suggestion
```

Maffooch (approved):

hblankenship (approved):

dogboat (approved):

mtesauro (approved):
Approved

PR #12137, opened 2025-03-30
releases: publish nightly builds of dev

valentijnscholten:
This PR uses the existing building blocks in our workflows to push a nightly build of the `dev` branch.

This is useful to user who want to test fixes or new features only available in `dev`. Or users who just want to run of `dev` which is very stable for Defect Dojo.

This PR also changes `github.event.inputs.xxxx` to `inputs.xxx`. This cost me a lot of time as the old input uses values provided in the UI when starting the workflow ignoring any values provided by the workflow files themselves.

I also tried to publish a nightly `bugfix` release. But the `bugfix` branch doesn't get its own version number. We have to decide if it's useful to publish these builds since we have a weekly release cadence for `bugfix`.

valentijnscholten_defectdojo-django general _ Docker Hub


hblankenship (approved):

mtesauro (approved):
Approved

Issue #12065, opened 2025-03-21
XML report from Webinspect 24.2 can not import


stanleylin75:
the XML report from Webinspect 24.2 can not import it

the error message is "An exception error occurred during the report import: ["This doesn't seem to be a valid Webinspect xml file."]"

any suggestion?

and is it possiable to support the X.FPR

stanleylin75:
[2024020107747_888_36750_20240222120043.zip](https://github.com/user-attachments/files/19413088/2024020107747_888_36750_20240222120043.zip)

manuel-sommer:
The problem is that this provided file has a structure that differs from the other Microfocus Webinspect. Thus, this has to be implemented. You can weit a little and I might find time to work on this, or you do a PR yourself.

stanleylin75:
i will wait, thanks you.

manuel-sommer:
Please review @stanleylin75

manuel-sommer:
Hi @stanleylin75 ,
did the way the vuln scanner outputs the scan result change in version 24.2?

manuel-sommer:
Can you close this @mtesauro ?

PR #12155, opened 2025-04-01
:tada: Implement Fortify Webinspect new report format

manuel-sommer:
#12065


valentijnscholten:
At first glance the parsing code looks quite different. Is this really just a new report format of the same scan type? If it is, will there be any impact on deduplication/reimport?

manuel-sommer:
The report format differs from the one before. I did not specifically test deduplication here.

mtesauro (approved):
Approved

dojo/tools/fortify/xml_parser.py
```diff
@@ -1,12 +1,69 @@
...
         root = fortify_scan.getroot()
+        if root.tag == "Scan":
+            return self.xml_structure_24_2(root, test)
+        if root.tag == "ReportDefinition":
+            return self.xml_structure_before_24_2(root, test)
+        raise ValueError
+
+    def xml_structure_24_2(self, root, test):
+        items = []
+        for Issues in root.findall("Issues"):
```
valentijnscholten, line 18: Shouldn't this be cased like `issues`.

dojo/tools/fortify/xml_parser.py
```diff
@@ -1,12 +1,69 @@
...
+            for Issue in Issues.iter("Issue"):
```
valentijnscholten, line 19: rename to `issue`?

valentijnscholten: etc

valentijnscholten:
Couple of remarks:
- To me it seems like a different scan result not just a different format. Might be better to split if off and/or make the docs more clear on how this report can be generated
- Can you rename the sample scan to be more descriptive of what it is, i.e. webinspect_4_2_many_findings.xml 
- The naming of the variables seems different to what we usually do, see my remarks inline.

manuel-sommer:
> Couple of remarks:
> 

- [ ]  To me it seems like a different scan result not just a different format. Might be better to split if off and/or make the docs more clear on how this report can be generated
- [x] Can you rename the sample scan to be more descriptive of what it is, i.e. webinspect_4_2_many_findings.xml
- [x] The naming of the variables seems different to what we usually do, see my remarks inline.

dojo/tools/fortify/xml_parser.py
```diff
@@ -1,12 +1,69 @@
...
+            return self.xml_structure_24_2(root, test)
+        if root.tag == "ReportDefinition":
+            return self.xml_structure_before_24_2(root, test)
+        raise ValueError
+
+    def xml_structure_24_2(self, root, test):
+        items = []
+        for issues in root.findall("Issues"):
+            for issue in issues.iter("Issue"):
+                CheckTypeID = issue.find("CheckTypeID").text
```
valentijnscholten, line 20:
```suggestion
                check_type_id= issue.find("CheckTypeID").text
```

dojo/tools/fortify/xml_parser.py
```diff
@@ -1,12 +1,69 @@
...
+                EngineType = issue.find("EngineType").text
```
valentijnscholten, line 21:
```suggestion
                engine_type = issue.find("EngineType").text
```

valentijnscholten (changes requested):
As per my commit suggestions, shouldn't the variables used in the new parser method be named like we do everywhere else?

valentijnscholten (approved):

Maffooch (approved):

blakeaowens (approved):

PR #12160, opened 2025-04-02
Product Revenue: Do no allow negative revenue

Maffooch:
It does not make sense for revenue to be a negative number. Because the revenue field is a decimal field, we must add a validator to disallow values less than 0

[sc-10753]


mtesauro (approved):
Approved

hblankenship (approved):

dogboat (approved):

blakeaowens (approved):

PR #12168, opened 2025-04-03
wiz: handle mitigated timestamps

### LINKED ISSUE #12152: Extend Wiz parser to handle RESOLVED status and parse Resolved Time.
When working with a Wiz CSV, RESOLVED entries should be created in DefectDojo as Closed findings, with a Mitigated Date that equals their Resolved Time.

See unit test from Wiz, line 11:  Status is RESOLVED, and although Resolved Time is not set in this file, if it was, it would be recorded as a timestamp that looks like this `2020-01-01 08:21:41.14242 +0000 UTC`

[multiple_findings.csv](https://github.com/user-attachments/files/19542158/multiple_findings.csv)

valentijnscholten:
Fixes #12152 

Handle `Resolved Time` from Wiz scans to populate the `mitigated` field.

Looks like `Wiz` is using some kind of almost ISO-8061 format for this column: `2025-04-03 20:20:00.43042 +0000 UTC`


Maffooch (approved):

blakeaowens (approved):

cneill (approved):

mtesauro (approved):
Approved

PR #12169, opened 2025-04-04
Ruff: Add S324 rule

manuel-sommer:


Maffooch (changes requested):
Something like this might be more robust for future contributors in the `ruff.toml` file under the `[lint.per-file-ignores]` section:
```
"dojo/tools/**" = [
    "S324", # https://docs.astral.sh/ruff/rules/hashlib-insecure-hash-function/
]
```

Maffooch (approved):

kiblik:
Question is. Do we want to do it easy (as it is proposed right now) or hard (as Ruff page recommends) :)

I'm okay with the current setup. The recommended solution is to add `usedforsecurity=False` to all `md5` calls (in case you are not able, simply replace it with a different hash function).

dogboat (approved):

manuel-sommer:
Let's keep it this way please

Maffooch:
> The recommended solution is to add `usedforsecurity=False` to all `md5` calls (in case you are not able, simply replace it with a different hash function).

Great suggestion! @manuel-sommer do you have bandwidth to add this?

manuel-sommer:
done @Maffooch

kiblik (approved):

hblankenship (approved):

PR #12178, opened 2025-04-06
feat(perf): Speed-up loading by using smaller resources (js,css)

kiblik:
No rocket science, just use `.min.js` and `.min.css` if exist


Maffooch (approved):

valentijnscholten (approved):

valentijnscholten:
nice find, approved.

hblankenship (approved):

PR #12193, opened 2025-04-07
Import/Reimport Stats: Change name of left untouched

Maffooch:
Following a successful import or reimport, a JSON blob for `statistics` is generated to provide the differential of finding activity. There was a section in the `delta` JSON blob that referred to a key labeled `left untouched`. This value does not comply with REST norms, and has been renamed to `left_untouched`.

[sc-6122]


cneill (approved):

mtesauro (approved):
Approved

hblankenship (approved):

valentijnscholten:
I'm late to the party @Maffooch, but I think we should go with just "untouched". to avoid dealing with spaces/underscores etc.

Maffooch:
@mtesauro what are your thoughts on Val's comment? I am not opposed to going with just `untouched` - it would actually simplify this PR a bit to do so

mtesauro:
I am good with just "untouched"

Maffooch:
@valentijnscholten thanks for raising that! We will go with `untouched`

dojo/db_migrations/0226_import_history_left_untouched_rename.py
```diff
@@ -0,0 +1,42 @@
+# Generated by Django 5.0.8 on 2024-09-12 18:22
+
+from django.db import migrations, models
+
+
+original_left_untouched_value = "left untouched"
+updated_left_untouched_value = "left_untouched"
```
mtesauro, line 7: Let's make this "untouched" :wink:

mtesauro (changes requested):

mtesauro (approved):
Approved

PR #12216, opened 2025-04-11
Fix: Prevent redirect to logout after login (security issue)

4NK1T:
## Description

**Issue:**
When users specify the `next` parameter in the login URL (e.g., `/login?next=logout`), the application redirects them directly after a successful login. This leads to unintended behavior, specifically immediate logout upon login, as users are redirected to the logout page.

**Proposed Solution:**
- Implemented validation logic in the `get_redirect_url()` method to ensure the `next` parameter cannot redirect users to unauthorized locations, specifically paths starting with `logout`.
- Ensured the redirect URL is properly validated against the allowed hosts.

## Test results

- Tested locally; confirmed `/login?next=logout` no longer redirects users to logout immediately after login.
- Normal redirection functionality after login remains unaffected.

## Checklist

### DIFF
--- a/dojo/user/views.py
+++ b/dojo/user/views.py
@@ -26,6 +26,8 @@ from django.utils.http import urlencode
 from django.utils.timezone import now
 from django.utils.translation import gettext as _
 from rest_framework.authtoken.models import Token
+from django.utils.http import url_has_allowed_host_and_scheme
+from urllib.parse import urlparse
 
 from dojo.authorization.authorization_decorators import user_is_configuration_authorized
 from dojo.authorization.roles_permissions import Permissions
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Thanks, could you modify the description to describe the issue that is being fixed? Please note that security vulnerabilities should be reported via https://github.com/DefectDojo/django-DefectDojo/security/policy

Maffooch:
Hi @4NK1T 
Thank you for your contribution! We did some brief testing on this reported behavior, and while it is strange, we cannot determine any security implications that come with this behavior. Can you please provide some additional information as to why you believe this to be a security issue?

mtesauro:
I'm with @Maffooch on this one - I can't see a way this is anything more then perhaps unexpected behavior.

I tried starting with trying going directly to https://demo.defectdojo.org/login?next=/logout - as if someone tricked me to click on that link:
1. Click on link
2. See the login page and login
3. I get redirected to the login page with a couple of notifications :shrug: 
image

Ok. Maybe I'm confused if I didn't pay attention to the URL but the attacker has gained nothing

What happens if I click on that link after I'm already logged in? I get another browser tab with an odd looking login page. Yeah, the UI is a bit weird but that's about it. I'm not logged out but, honestly, should I be? I've been tricked into going to the login page with an add GET next parameter. Sure, maybe confuses a user but again, the attacker gains nothing. I could enter my creds in the new odd looking login page and I'm logged out like above so... :shrug: 

image

I'm super curious where the security issue is here. I can't see anything an attacker could gain beyond perhaps confusing some DefectDojo users.

4NK1T:
This is not a major security issue i raised it coz i thought it could be exploited

Thiss is about preventing abuse of the next parameter on login. Currently, it's possible for a user to be redirected to /logout immediately after logging in either through a crafted URL or a phishing attempt that includes ?next=logout.

While this might seem minor, it can be leveraged for social engineering or to create confusion for users the current implementation doesn’t validate the next parameter, which could open the door to other unsafe redirect paths if extended in the future.

valentijnscholten:
I don't see any tangible risk here as described by @mtesauro above so I'm closing this. We're happy to reconsider if there's new arguments that require a an allowlist/blocklist on the `next` parameter.

PR #12247, opened 2025-04-15
nessus: parse more fields

valentijnscholten:
Users reported on Slack that the Nessus CSV report can contain fields that are currently not parsed by DefectDojo:

image (5)
Reference: https://owasp.slack.com/archives/C2P5BA8MN/p1743609798217049

This PR parses some more fields in the csv parse:
- epss_score
- more references
- vpr score
- stig score
- canvas
- metasploit exploitable
- core impact
- canvas
- plugin information

To me it's not clear what some of these fields, but probably anyone using Nessus will understand. So it's good to have these values in the description, references or severity_justification field.

We decided for now not to parse the CWE field that is present in the XREF fields as it would influence the hash code calculation which we would not be able to get aligned with older findings that did not have the cwe field parsed and stored.

Ruff required me to update all test cases.

### DIFF
--- a/dojo/tools/tenable/csv_format.py
+++ b/dojo/tools/tenable/csv_format.py
@@ -101,12 +101,44 @@ class TenableCSVParser:
                 raw_severity = int_severity
             # convert the severity to something dojo likes
             severity = self._convert_severity(raw_severity)
+
+            epss_score = None
+            epss_score_string = row.get("EPSS Score") if "EPSS Score" in row else None
+            # example seen so far are "1234" for an actual score of "0.1234" so let's prepend that instead of "risky" divisions
+            if epss_score_string:
+                if "0." not in epss_score_string:
+                    epss_score_string = "0." + epss_score_string
+                epss_score = float(epss_score_string)
+
             # Other text fields
             description = row.get("Synopsis", row.get("definition.synopsis", "N/A"))
+
+            severity_justification = f"Severity: {severity}\n"
+            for field in (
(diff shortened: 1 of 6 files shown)


dojo/tools/tenable/xml_format.py
Maffooch, line 1: Since the CWE value is already present in XML, it should not be changed

dojo/tools/tenable/csv_format.py
```diff
@@ -123,9 +156,12 @@ def get_findings(self, filename: str, test: Test):
                     test=test,
                     description=description,
                     severity=severity,
+                    cwe=cwe,
```
Maffooch, line 159:
```suggestion
```

Maffooch (changes requested):

dojo/tools/tenable/csv_format.py
valentijnscholten: I think 0 is the default in models.py. But this line was a left over anyway to be commented out.

dojo/tools/tenable/xml_format.py
valentijnscholten: it's not updated, I commented that out. I now commented out the helper function as well. I'm hesitant to remove it as we may decide to use it in the future as CWE is a useful field.

mtesauro (approved):
Approved

hblankenship (approved):

dogboat (approved):

PR #12249, opened 2025-04-15
Reimport: Special statuses should be respected from reports

Maffooch:
When performing a reimport, the status of a finding from the report that is either false positive, out of scope, or risk accepted is being discarded. This is due to a flaw in the  way findings are calculated for the `close_old_findings` function.

Let's say we have a finding in the database that is active. In the next reimport, the same finding is being reported form the tool as false positive. Reimport will do the following:
1.  Store the list of original findings
2. Parse the findings
3. Find a match between this new false positive finding, and the existing active finding
4. Set the active finding to false positive, and save the finding to the database
5. Determine the findings to be closed by subtracting the findings from the report from the original list of findings from step 1.
    - This is just the original active finding (though the record is stale since it was updated in step 4.)
6. Close the active finding by setting the mitigated status
7. Find out that the false positive status was discarded

This situation is caused from attempting to close a stale record, that may or may not already be closed, and therefore overwrite any other statuses. The solution here is simple: fetch the latest status on the finding before deciding if it needs to be closed. 

[sc-10908]

### DIFF
--- a/dojo/tools/checkmarx_one/parser.py
+++ b/dojo/tools/checkmarx_one/parser.py
@@ -71,7 +71,6 @@ class CheckmarxOneParser:
                     "description": (
                         f"{query.get('description')}\n\n"
                         f"**Category**: {query.get('category')}\n"),
-                    "verified": query.get("state") != "TO_VERIFY",
                     "test": test,
                 }
                 # Iterate over the individual issues
@@ -81,6 +80,8 @@ class CheckmarxOneParser:
                         date = self._parse_date(instance.get("firstDetectionDate"))
                     else:
                         date = self._parse_date(instance.get("lastDetectionDate"))
+                    instance_details = self.determine_state(instance)
+                    instance_details.update(base_finding_details)
                     # Create the finding object
                     finding = Finding(
                         severity=instance.get("severity").title(),
@@ -174,14 +175,15 @@ class CheckmarxOneParser:
                     date = self._parse_date(instance.get("firstFoundDate"))
                 else:
                     date = self._parse_date(instance.get("foundDate"))
+                instance_details = self.determine_state(instance)
+                instance_details.update(base_finding_details)
                 # Create the finding object
                 finding = Finding(
                     severity=instance.get("severity").title(),
                     date=date,
                     file_path=instance.get("destinationFileName"),
                     line=instance.get("destinationLine"),
-                    verified=instance.get("state") != "TO_VERIFY",
-                    **base_finding_details,
+                    **instance_details,
                 )
                 # Add some details to the description
                 if node_snippet := get_node_snippet(instance.get("nodes", [])):
@@ -219,11 +221,9 @@ class CheckmarxOneParser:
                 + "**uri**: " + locations_uri + "\n"
                 + "**startLine**: " + str(locations_startLine) + "\n"
                 + "**endLine**: " + str(locations_endLine) + "\n",
-                false_p=False,
-                duplicate=False,
-                out_of_scope=False,
                 static_finding=True,
                 dynamic_finding=False,
+                **self.determine_state(result),
             )
             findings.append(finding)
         return findings
@@ -273,6 +273,7 @@ class CheckmarxOneParser:
             test=test,
             static_finding=True,
             unique_id_from_tool=unique_id_from_tool,
+            **self.determine_state(vulnerability),
         )
 
     def get_results_kics(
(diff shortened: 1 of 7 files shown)


dogboat (approved):

unittests/tools/test_checkmarx_one_parser.py
```diff
@@ -153,3 +153,15 @@ def test_sca_finding(finding):
...
+
+    def test_checkmarx_one_false_positive_status(self):
+        with (get_unit_tests_scans_path("checkmarx_one") / "one-open-one-false-positive.json").open(encoding="utf-8") as testfile:
+            parser = CheckmarxOneParser()
+            findings = parser.get_findings(testfile, Test())
+            self.assertEqual(2, len(findings))
+            # check the first first finding is false positive
+            self.assertEqual(True, findings[0].false_p)
+            self.assertEqual(False, findings[0].active)
+            # check the first second finding is not false positive
```
cneill, line 165:
```suggestion
            # check the first finding is false positive
            self.assertEqual(True, findings[0].false_p)
            self.assertEqual(False, findings[0].active)
            # check the second finding is not false positive
```

dojo/tools/checkmarx_one/parser.py
```diff
@@ -81,6 +80,8 @@ def parse_iac_vulnerabilities(
                         date = self._parse_date(instance.get("firstDetectionDate"))
                     else:
                         date = self._parse_date(instance.get("lastDetectionDate"))
+                    instance_details = self.determine_state(instance)
+                    instance_details.update(base_finding_details)
```
cneill, line 84: Should `instance_details` be referenced on line 94 instead of `base_finding_details`, as is done on line 186? (Can't make a suggestion since that line's not in the diff)

https://github.com/DefectDojo/django-DefectDojo/blob/84b90bea7d413dc2ea3bd842f966a2d967d04f58/dojo/tools/checkmarx_one/parser.py#L86-L95

https://github.com/DefectDojo/django-DefectDojo/blob/84b90bea7d413dc2ea3bd842f966a2d967d04f58/dojo/tools/checkmarx_one/parser.py#L181-L187

Maffooch: Good catch, you are right here

valentijnscholten:
When a report has findings that are mitigated since the previous import, a similar problem will occur. I don't think this PR solves that yet. A work around for that was done in #12204.

I think the root cause is not that old data is being passed into `close_old_findings`. The problem is that findings are that are mitigated (including false positive, risk accepted or out of scope) are passed into `close_old_findings`. 

The `close_old_findings` method is meant to close findings that _are no longer in the report_. To achieve that, reimport keeps track of:

- new findings -> `new_items`
- reactivated findings -> `reactivated_items`
- untouched findings -> `unchanged_items`
- false_p/risk_accepted/out of scope findings -> not tracked <-- this is the root cause
- mitigated findings -> not tracked <-- similar root cause

The result is that too many findings are passed into `close_old_findings`:

https://github.com/DefectDojo/django-DefectDojo/blob/3fd00f0290ab0fb35e34174d9816b7f34d86cd5e/dojo/importers/default_reimporter.py#L238

Although the refresh solution may work for now, it might not work in the future if something changes in the reimport. 

The refresh implementation also results in incorrect counts for the `updated_count` field and for the values passed into notification handler:

https://github.com/DefectDojo/django-DefectDojo/blob/3fd00f0290ab0fb35e34174d9816b7f34d86cd5e/dojo/importers/default_reimporter.py#L128-L138

I think tracking mitigated findings and "special status findings" should be added so we can stick to the original design that `close_old_findings` only processes findings that are no longer in the report hat is being imported.

i.e. 

`        self.to_mitigate = (set(self.original_items) - set(self.reactivated_items) - set(self.unchanged_items)) - set(self.mitigated_items)`

This would expose another error in the reimporter where findings are marked as false positive/risk accepted/out of scope without setting the `mitigated` and `is_mitigated_fields`: https://github.com/DefectDojo/django-DefectDojo/blob/3fd00f0290ab0fb35e34174d9816b7f34d86cd5e/dojo/importers/default_reimporter.py#L606-L617

Alternative the mitigated findings and "special status" findings can be tracked separately from each other.

valentijnscholten:
Turns out that by tracking the mitigated findings, we see that in some unit tests a findings get mitigated _and_ reactivated within the same import. I guess for now it's OK to stick to the refresh solution as that solves the reported problem and doesn't seem to make things worse.

valentijnscholten (approved):

cneill (approved):

mtesauro (approved):
Approved

PR #12258, opened 2025-04-17
WIP: track mitigated status better in reimport

valentijnscholten:
WIP alternative to #12249 

possibly `self.mitigate_finding` should be used instead of setting the mitigation related fields manually.


valentijnscholten:
The unit tests are failing because in a single reimport first an existing active inding gets mitigated but later that same existing finding gets matched against another finding in the reimport and gets reactivated again. This make the finding part of both `mitigated_items` and `reactivated_items` resulting in two import history items. this sound incorrect. But the code without the changes in this PR is also incorrect as it records the finding as reactivated while it already was active before the import.

PR #12291, opened 2025-04-22
Reimport: Special statuses should be respected from reports

Maffooch:
Replacement of #12249 

When performing a reimport, the status of a finding from the report that is either false positive, out of scope, or risk accepted is being discarded. This is due to a flaw in the  way findings are calculated for the `close_old_findings` function.

Let's say we have a finding in the database that is active. In the next reimport, the same finding is being reported form the tool as false positive. Reimport will do the following:
1.  Store the list of original findings
2. Parse the findings
3. Find a match between this new false positive finding, and the existing active finding
4. Set the active finding to false positive, and save the finding to the database
5. Determine the findings to be closed by subtracting the findings from the report from the original list of findings from step 1.
    - This is just the original active finding (though the record is stale since it was updated in step 4.)
6. Close the active finding by setting the mitigated status
7. Find out that the false positive status was discarded

This situation is caused from attempting to close a stale record, that may or may not already be closed, and therefore overwrite any other statuses. The solution here is simple: fetch the latest status on the finding before deciding if it needs to be closed. 

[sc-10908]

### DIFF
--- a/dojo/importers/default_reimporter.py
+++ b/dojo/importers/default_reimporter.py
@@ -264,6 +264,15 @@ class DefaultReImporter(BaseImporter, DefaultReImporterOptions):
         # Determine if pushing to jira or if the finding groups are enabled
         mitigated_findings = []
         for finding in findings:
+            # Get any status changes that could have occurred earlier in the process
+            # for special statuses only.
+            # An example of such is a finding being reported as false positive, and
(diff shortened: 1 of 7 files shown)


dogboat (approved):

cneill (approved):

mtesauro (approved):
Approved

PR #11924, opened 2025-03-01
SLA Calculations 2/2: Simplify logic

valentijnscholten:
If #11923 gets merged, we have unit tests to capture the SLA calculation behaviour.

So we can now simplify the logic to make it easier to understand, maintain and (possibly) add business days support.

PR #12301, opened 2025-04-23
sla: parse finding.date implicitly

### LINKED ISSUE #12299: Multiple parser fails when importing in 2.45.2
When parsing Edgescan file (v. 2.45.2), an error occurs:

Image

You have to uncomment the following to circumvent the error:
Image

valentijnscholten:
Fixes #12299 introduced in #11924

The old code had parsing in place for parsers that didn't set the `finding.date` correctly. These parsers provided a `str` instead of `date`. This PR ensures this implicit parsing is performed with the simplified SLA logic.

This implicit parsing was tucked away in the `_age()` method: https://github.com/DefectDojo/django-DefectDojo/blob/8e7cc01142c5f8d7b31b85e686cbd28d341ad2ca/dojo/models.py#L3004-L3007

 The unit tests of parsers don't trigger this SLA logic, we'll look to add more checks to the parser unit tests in a future PR.

All affected parsers are now working again:

image

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -10,6 +10,7 @@ from decimal import Decimal
 from pathlib import Path
 from uuid import uuid4
 
+import dateutil
 import hyperlink
 import tagulous.admin
 from auditlog.registry import auditlog
@@ -3058,9 +3059,8 @@ class Finding(models.Model):
 
         # some parsers provide date as a `str` instead of a `date` in which case we need to parse it #12299 on GitHub
         sla_start_date = self.get_sla_start_date()
-        from dateutil.parser import parse
         if sla_start_date and isinstance(sla_start_date, str):
-            sla_start_date = parse(sla_start_date).date()
+            sla_start_date = dateutil.parser.parse(sla_start_date).date()
 
         sla_period, enforce_period = self.get_sla_period()
         if sla_period is not None and enforce_period:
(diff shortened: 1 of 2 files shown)


dojo/models.py
```diff
@@ -3053,9 +3056,15 @@ def set_sla_expiration_date(self):
         if not system_settings.enable_finding_sla:
             return
 
+        # some parsers provide date as a `str` instead of a `date` in which case we need to parse it #12299 on GitHub
+        sla_start_date = self.get_sla_start_date()
+        from dateutil.parser import parse
```
Maffooch, line 3061: This should be at the top of the file

dojo/models.py
Maffooch, line 1: A quick test of saving a finding with varying formats (string, date, datetime) in the date field would be ideal here

Maffooch (changes requested):

dojo/models.py
valentijnscholten: Good point. I assumed it was like this to avoid circular references/imports. Changed it.

dojo/models.py
valentijnscholten: Done

Maffooch (approved):

manuel-sommer (approved):

dogboat (approved):

hblankenship (approved):

PR #12302, opened 2025-04-23
Add input validation (branch to release num) for the release gha

rossops:
**Description**

- Adds a dropdown choice to the GHA that kicks off the OS release. Currently this is a free form text field which could lead to shenanigans. 
- Adds input validation based on branch and release number. We expect dev releases last octet to end in "0" and bugfix releases to be 1 or greater. 

**Test results**

Validated working and non-working combinations of branch to release number. 
- dev with 2.33.1 (failed validation, as expected)
- dev with 2.33.0 (pass)
- bugfix with 2.33.0 (fail)
- bugfix with 2.33.1 (pass)
- either branch with letters, non-alphanum chars, etc. (fail).

### DIFF
--- a/.github/workflows/release-1-create-pr.yml
+++ b/.github/workflows/release-1-create-pr.yml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Validate proper bugfix branch release_number format is being used
-        if: ${{ github.event.inputs.from_branch == 'bugfix' }}
+        if: ${{ inputs.from_branch == 'bugfix' }}
         run: |
           # Expect the last octet in release_number to not be 0
           echo "${{ github.event.inputs.release_number }}" | grep "^[0-9]*\.[0-9]*\.[1-9]$"
--- a/.github/workflows/release-1-create-pr.yml
+++ b/.github/workflows/release-1-create-pr.yml
@@ -11,6 +11,11 @@ on:
       from_branch:
         description: "Select branch to release from ('release/x.y.z'. If `dev` is entered, a new release branch will be created from `dev`)"
         required: true
+        type: choice
+        default: 'bugfix'
+        options:
+          - bugfix
+          - dev
       release_number:
         description: "Release version (x.y.z format)"
         required: true
@@ -19,6 +24,18 @@ jobs:
   create_pr:
     runs-on: ubuntu-latest
     steps:
+      - name: Validate proper bugfix branch release_number format is being used
+        if: ${{ github.event.inputs.from_branch == 'bugfix' }}
+        run: |
+          # Expect the last octet in release_number to not be 0
+          echo "${{ github.event.inputs.release_number }}" | grep "^[0-9]*\.[0-9]*\.[1-9]$"
+
+      - name: Validate proper dev branch release_number format is being used
+        if: ${{ github.event.inputs.from_branch == 'dev' }}
+        run: |
+          # Expect the last octet in release_number to not be 1-9
+          echo "${{ github.event.inputs.release_number }}" | grep "^[0-9]*\.[0-9]*\.0$"
+
       - id: Set-GitHub-org
         run: echo "GITHUB_ORG=${GITHUB_REPOSITORY%%/*}" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


blakeaowens (approved):

.github/workflows/release-1-create-pr.yml
```diff
@@ -19,6 +24,18 @@ jobs:
   create_pr:
     runs-on: ubuntu-latest
     steps:
+      - name: Validate proper bugfix branch release_number format is being used
+        if: ${{ github.event.inputs.from_branch == 'bugfix' }}
+        run: |
+          # Expect the last octet in release_number to not be 0
+          echo "${{ github.event.inputs.release_number }}" | grep "^[0-9]*\.[0-9]*\.[1-9]$"
```
valentijnscholten, line 31:
```suggestion
          echo "${{ inputs.release_number }}" | grep "^[0-9]*\.[0-9]*\.[1-9]$"
```

.github/workflows/release-1-create-pr.yml
```diff
@@ -19,6 +24,18 @@ jobs:
...
+
+      - name: Validate proper dev branch release_number format is being used
+        if: ${{ github.event.inputs.from_branch == 'dev' }}
```
valentijnscholten, line 34:
```suggestion
        if: ${{ inputs.from_branch == 'dev' }}
```

.github/workflows/release-1-create-pr.yml
```diff
@@ -11,6 +11,11 @@ on:
       from_branch:
         description: "Select branch to release from ('release/x.y.z'. If `dev` is entered, a new release branch will be created from `dev`)"
```
valentijnscholten, line 12:
```suggestion
        description: "Select branch to release from ('release/x.y.z'. If `dev` or `bugfix` is entered, a new release branch will be created from that branch)"
```

valentijnscholten (changes requested):
posted some comments/suggestions

valentijnscholten:
Further down the workflow there are cases that check for the `from_branch` starting with `release/`. That case is no longer supported now with the `choice` input. Should it be removed? We used it in the past where we were manually prepping (bugfix) releases. I think with the current way of working we no longer need to support that case.

On the other hand it would be nice to be able to do out-of-band releases from a hotfix branch if needed. Maybe add `hotfix` to the dropdown? Or am I making this PR to broad now :-)

.github/workflows/release-1-create-pr.yml
```diff
@@ -11,6 +11,11 @@ on:
...
         required: true
+        type: choice
+        default: 'bugfix'
```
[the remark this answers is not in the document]
rossops: I went back and forth on that myself. I landed on bugfix as default because thats what we do the most of. However, given the added input validation and usage of a choice field, I could go either way. I'll let @Maffooch make the deciding vote :)

rossops:
@valentijnscholten good catch!

> Further down the workflow there are cases that check for the `from_branch` starting with `release/`. That case is no longer supported now with the `choice` input. Should it be removed? We used it in the past where we were manually prepping (bugfix) releases. I think with the current way of working we no longer need to support that case.

Assuming you mean [these?](https://github.com/DefectDojo/django-DefectDojo/blob/8e7cc01142c5f8d7b31b85e686cbd28d341ad2ca/.github/workflows/release-1-create-pr.yml#L30-L38) 

IMO its no longer needed, but before I change this, @Maffooch is there some history or something i'm not considering here as to why we would re-release a release in this workflow? 

> On the other hand it would be nice to be able to do out-of-band releases from a hotfix branch if needed. Maybe add `hotfix` to the dropdown? Or am I making this PR to broad now :-)

Might not be a bad idea. I checked the release history going back a few years and I dont see any indication of a hotfix style release (w.x.y.z vs x.y.z), but maybe there's an edge case here for out of band releases.  I'll defer to our resident historian @Maffooch.

valentijnscholten:
The release/xxx branche option was not to re-release releases but to prepare the next release. This was before we had the bugfix branch. I think that option (release/xxx) can be removed. And a hotfix option is not really needed right now I guess, so we keep it simple :-)

rossops:
I tightened up the description on the from_branch as well to better reflect the process used.

Maffooch (approved):
At this point after doing bugfixes for a couple years now, those `release/` branches have just been noise. If we ever have any need for an out of band hot fix, we usually just go from the bugfix branch.

For the state of this PR, this is a big step in the right direction! I think we could merge this now, and then iterate again if needed

valentijnscholten (approved):

hblankenship (approved):

PR #12303, opened 2025-04-23
fix(api): Enable to set `recommendation` and `decision` in `risk_acceptance`

kiblik:
# Issue

Until now, the API endpoint `POST /api/v2/risk_acceptance/` (which also applies to `PUT` and `PATCH`) has not accepted `recommendation` and `decision`. As a result, DD used the default value (fix for recommendation and accept for decision). There was no error message.
For `GET` requests, DD translated the chosen value (one of `A`, `V`, `M`, `F`, `T`) to human-readable form.
It sounds nice, but not if it means that fields are not writable.
Screenshot 2025-04-23 at 16 43 24
Screenshot 2025-04-23 at 16 45 00

# Solution
This PR dropped `SerializerMethodField`s with `get_...` methods, which locked these fields to read-only mode. Now, during the creation of RiskAcc, the user is able to define `recommendation` and `decision`.
Screenshot 2025-04-23 at 16 59 22

# But
The only "disadvantage" is that it is now unable to see a human-readable form anymore.
Screenshot 2025-04-23 at 16 58 59

# No buts
APIs are not for humans but for machines, so simple values should not be the issue.
Translation is still possible based on the description in the OpenAPI spec:
```bash
$ curl 'http://localhost:8080/api/v2/oa3/schema/?format=json' 2>/dev/null | jq '.paths."/api/v2/risk_acceptance/".get.parameters[] | select(.name=="recommendation").description' -r
Recommendation from the security team.

* `A` - Accept (The risk is acknowledged, yet remains)
* `V` - Avoid (Do not engage with whatever creates the risk)
* `M` - Mitigate (The risk still exists, yet compensating controls make it less of a threat)
* `F` - Fix (The risk is eradicated)
* `T` - Transfer (The risk is transferred to a 3rd party)
```
or
```bash
$ curl -X 'OPTIONS' 'http://localhost:8080/api/v2/risk_acceptance/' -H 'accept: application/json' -H 'Authorization: Token xxx' 2>/dev/null | jq .actions.POST.recommendation.choices
[
  {
    "value": "A",
    "display_name": "Accept (The risk is acknowledged, yet remains)"
  },
  {
    "value": "V",
    "display_name": "Avoid (Do not engage with whatever creates the risk)"
  },
  {
    "value": "M",
    "display_name": "Mitigate (The risk still exists, yet compensating controls make it less of a threat)"
  },
  {
    "value": "F",
    "display_name": "Fix (The risk is eradicated)"
  },
  {
    "value": "T",
    "display_name": "Transfer (The risk is transferred to a 3rd party)"
  }
]
```

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1525,8 +1525,6 @@ class TestImportSerializer(serializers.ModelSerializer):
 
 
 class RiskAcceptanceSerializer(serializers.ModelSerializer):
-    recommendation = serializers.SerializerMethodField()
-    decision = serializers.SerializerMethodField()
     path = serializers.SerializerMethodField()
 
     def create(self, validated_data):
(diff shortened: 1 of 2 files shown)


blakeaowens (approved):

mtesauro (approved):
Approved

Maffooch (approved):

dogboat (approved):

PR #12332, opened 2025-04-28
False Positive Status: Update docs

Maffooch:


mtesauro (approved):
Approved

valentijnscholten:
Is the False Positive History feature (that also works with import) still a thing? I think that's why the original text was there.

dogboat (approved):

hblankenship (approved):

valentijnscholten:
> > Is the False Positive History feature (that also works with import) still a thing? I think that's why the original text was there.
> 
> Yes, False Positive History is still a thing, but it was not actually mentioned in the original text. It may be worth making a separate write up for false positive history. Is that something you would be willing to take on?

I have never used it and I am not sure I am aware of the original use case / goal it is trying to achieve. Happy to leave it for now.

valentijnscholten:
> > Is the False Positive History feature (that also works with import) still a thing? I think that's why the original text was there.
> 
> Yes, False Positive History is still a thing, but it was not actually mentioned in the original text. It may be worth making a separate write up for false positive history. Is that something you would be willing to take on?

I have never used it and I am not sure I am aware of the original use case / goal it is trying to achieve. Happy to leave it for now.

Maffooch:
Let's leave it for now and revisit in the future

valentijnscholten (approved):

PR #12333, opened 2025-04-29
Jira webhook comment duplicate patch

Maffooch:
This PR updates a few things to catch some corner cases:
- Remove the check for the same author when looking for existing comments
  - There is [a check further up in the webhook comment processor to skip comments from jira made by the same user that created the jira instance in dojo](https://github.com/DefectDojo/django-DefectDojo/blob/477583eb674a6a93516a2b794cd9889cb6d7ab81/dojo/jira_link/views.py#L234-L238). This is great for all smaller teams, but not larger ones
  - In cases where this is not true, like when a service account is connecting to jira, only the case is considered when the webhook could fire twice (see description in #9513), the content of the message must be checked as well. This will not cause a regression as the content is still a safe check.
- Check for note contents that reflect a comment made in DefectDojo, as well as a comment that would be coming from Jira
  - This additional check will prevent a loop of: 
       1. comment left in dojo and pushed to jira with user prefix
       2. jira comment created and send webhook response
       3. dojo processes webhook responses, and creates comment on finding
       
[sc-10999]

### DIFF
--- a/dojo/jira_link/views.py
+++ b/dojo/jira_link/views.py
@@ -255,10 +255,6 @@ def check_for_and_create_comment(parsed_json):
     # Set the fields for the notes
     author, _ = User.objects.get_or_create(username="JIRA")
     entry = f"({commenter_display_name} ({commenter})): {comment_text}"
-    print("\n\n")
-    print(f"\t entry: {entry}")
-    print(f"\t comment_text_without_defectdojo_user: {comment_text_without_defectdojo_user}")
-    print("\n\n")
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

dogboat (approved):

valentijnscholten (approved):
It's a little hard to oversee all the possible variations without unit cases. At the same it looks like this is a safe change as it only looks at comments created in the last 30 seconds, reducing the chance of falsely dropping a commdn and not storing it.

PR #12346, opened 2025-04-29
Store fingerprint from bearer in unique_id_from_tool

wolframite:
**Description**

I'm storing the unique fingerprint from bearer which is always present in the scan file in the field unique_id_from_tool. I also changed the deduplication algorithm setting in dojo/settings/settings.dist.py from DEDUPE_ALGO_HASH_CODE to DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE

**Test results**

There is no existing unit test to extend.

**Documentation**

Updating the documentation should not be necessary for this tiny change

**Checklist**

**Extra information**

This is my first PR for defect dojo, please let me know if I made a mistake. I'll fix it right away.

### DIFF
--- a/dojo/tools/bearer_cli/parser.py
+++ b/dojo/tools/bearer_cli/parser.py
@@ -46,6 +46,7 @@ class BearerCLIParser:
                     sast_source_line=bearerfinding["source"]["start"],
                     sast_source_file_path=bearerfinding["filename"],
                     vuln_id_from_tool=bearerfinding["id"],
+                    # the fingerprint is not constant over time, but because it's not used for dedupe it's safe and useful to set it
                     unique_id_from_tool=bearerfinding["fingerprint"],
                 )
 
--- a/dojo/tools/bearer_cli/parser.py
+++ b/dojo/tools/bearer_cli/parser.py
@@ -46,6 +46,7 @@ class BearerCLIParser:
                     sast_source_line=bearerfinding["source"]["start"],
                     sast_source_file_path=bearerfinding["filename"],
                     vuln_id_from_tool=bearerfinding["id"],
+                    unique_id_from_tool=bearerfinding["fingerprint"],
                 )
 
                 items.append(finding)
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -1584,6 +1584,7 @@ DEDUPLICATION_ALGORITHM_PER_PARSER = {
     "MobSF Scorecard Scan": DEDUPE_ALGO_HASH_CODE,
     "OSV Scan": DEDUPE_ALGO_HASH_CODE,
     "Nosey Parker Scan": DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE,
+    # The bearer fingerprint is not unique across multiple scans, so it shouldn't be used for deduplication (https://github.com/DefectDojo/django-DefectDojo/pull/12346#issuecomment-2841561634)
     "Bearer CLI": DEDUPE_ALGO_HASH_CODE,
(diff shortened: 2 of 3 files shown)


valentijnscholten:
Thanks for your PR. Could you add some asserts to the unit tests on this field?

I also notice the `old_fingerprint` field in the reports. Do you know how that works? Just need to make sure the `fingerprint` is unique/constant for each finding.

wolframite:
Checking the [source](https://github.com/Bearer/bearer/blob/64859dcae9e1617cc425cf53eb4723c5c7144773/pkg/report/output/security/security.go#L231-L235) where the fingerprints are generated it looks like they're only unique within one scan. If the code changes and a weakness is fixed, then the fingerprints will be re-used.

The fingerprint is the md5 hash of the filename and the rule id suffixed with the index of findings of this specific file. The old fingerprint has the same hash, but is suffixed with the index of all findings. I assume that this was just the old way of calculating the fingerprint. 

This is a bug in bearer, because if you ignore the fingerprint with the index 2 and then fix the weakness with the index 0, the weakness which previously had index 3 will be ignored now instead.

This sounds like we should not use the fingerprint to track findings across multiple scans. That means I should revert deduplication setting back to DEDUPE_ALGO_HASH_CODE right?

Maffooch:
> This sounds like we should not use the fingerprint to track findings across multiple scans. That means I should revert deduplication setting back to DEDUPE_ALGO_HASH_CODE right?

Yep, you've got the right idea 😉 if the current hash code fields of title and severity are fitting the need we may be okay here

wolframite:
If I remember correctly the title is the rule id + file name + line number, so this would be "uniquer".

Maffooch (approved):
This should be strictly additive without change to dedupe settings for existing folks, so this seems safe

dojo/tools/bearer_cli/parser.py
```diff
@@ -46,6 +46,7 @@ def get_findings(self, file, test):
                     sast_source_line=bearerfinding["source"]["start"],
                     sast_source_file_path=bearerfinding["filename"],
                     vuln_id_from_tool=bearerfinding["id"],
+                    unique_id_from_tool=bearerfinding["fingerprint"],
```
valentijnscholten, line 49: It is safe now as it is. To prevent someone(tm) from adding the `unique_id_from_tool` to the hash code config should we add a little comment here and in `settings.dist.py`? @Maffooch

Maffooch: It doesn't hurt to be more verbose! @wolframite cab you please summarize your analysis of the fingerprint field in a comment where @valentijnscholten pointed out?

wolframite: Sure, no problem. I added the comment, please let me know if you're ok with the wording

valentijnscholten: Thanks. I just realized there's one more thing... Could you update the testcase(s) to assert on the fingerprint being set into `unique_id_from_tool`?

valentijnscholten, line 49:
```suggestion
                    # the fingerprint is not constant over time, but because it's not used for dedupe it's safe and useful to set it
                    unique_id_from_tool=bearerfinding["fingerprint"],
```

wolframite: Huh, I don't know how I overlooked the test for bearer, I was pretty sure there wasn't one. It's updated it now. Do you want me to fixup all changes and rebase with dev?

wolframite: Also thank you @valentijnscholten and @Maffooch for guiding me along my first ticket!

valentijnscholten (approved):

mtesauro (approved):
Approved

hblankenship (approved):

PR #12347, opened 2025-04-30
Generic Parser: Support Test Type Meta

Maffooch:
Generic Finding Import is capable of importing data from the report that influences the name of the test type and some other fields on the test. This behavior could be expanded to also influence the `static_tool` and `dynamic_tool` fields as well.

The following things were added to accomplish this:
- Expand the `ParserTest` class  to accommodate the new fields
- Rework the dynamic test parser in the importers to allow for fields to be set during import, and overwritten during reimport if fields are supplied
- Unit tests at the parser and import/reimport level to cover to above
- Docs to explain how these fields are used

[sc-11026]


Maffooch:
I did not realize I based this PR from an older branch I used in another PR 🥴

Issue #10219, opened 2024-05-17
DD Silently accepts `test_type` mismatch


kiblik:
**Be informative**
DD supports the creation of custom `test_type` if the parser supports `get_tests`. The report can contain custom `test_type` which is useful during deduplication. `test_type` is set during the first import.

**Bug description**
If a user reimports the file that contains a different `test_type`, DD silently accepts this mismatch. But it should not.

**Steps to reproduce**
1. Import report 1 as generic
1. Import report 2 as generic
1. See that `test_type` is still the same

**Expected behavior**
I suppose there are 2 possible ways (I can imagine that both would be implemented and users can choose which is better for their deployment; but I would default with _fail_):
- **Fail**: When DD detects inconsistency between the original test type and the test type coming from a new report, it should raise some error (before starting deduplication and writings to DB)
- **Adopt**: Update `test_type`. Before deduplication starts.

Open question: Right now DD supporters only one test type for report. Is this going to change in the future? How should reimport behave in that case? 

**Environment information**
 - DefectDojo version: `[2024-05-16 12:17:39 -0500] f44c975c1: Bump python-gitlab from 4.4.0 to 4.5.0 (#10209) [ (HEAD, upstream/dev)]`

**Screenshots**

<img width="389" alt="image" src="https://github.com/DefectDojo/django-DefectDojo/assets/5609770/23f80a4b-3873-4fca-9814-1738224a26cc">

**Sample scan files**
- [generic-report-test-type-1.json](https://github.com/DefectDojo/django-DefectDojo/files/15347114/generic-report-test-type-1.json)
- [generic-report-test-type-2.json](https://github.com/DefectDojo/django-DefectDojo/files/15347122/generic-report-test-type-2.json)

Maffooch:
This is a question that has been in my mind recently. I think the largest consideration here would have to be around deduplication settings. In the case of updating the test type name from import1 to import2, there maybe be different deduplication settings between the two test types, and lead to some weird behavior. Primary reason for this is that the hash codes for findings under test type 1 would not be updated when suddenly the findings are now under test type 2.

For that reason, I believe that only leaves one option that you had proposed:
> Fail: When DD detects inconsistency between the original test type and the test type coming from a new report, it should raise some error (before starting deduplication and writings to DB)

I believe  this could be a simple check for when the test already exists: Does the calculated test type name from the internal test match the test type name that currently exists in the system? If not, raise an exception

valentijnscholten:
[...]

PR #12348, opened 2025-04-30
Generic Parser: Support Test Type Meta

Maffooch:
Generic Finding Import is capable of importing data from the report that influences the name of the test type and some other fields on the test. This behavior could be expanded to also influence the `static_tool` and `dynamic_tool` fields as well.

The following things were added to accomplish this:
- Expand the `ParserTest` class  to accommodate the new fields
- Rework the dynamic test parser in the importers to allow for fields to be set during import, and overwritten during reimport if fields are supplied
- Unit tests at the parser and import/reimport level to cover to above
- Docs to explain how these fields are used

[sc-11026]

### DIFF
--- a/docs/content/en/connecting_your_tools/parsers/generic_findings_import.md
+++ b/docs/content/en/connecting_your_tools/parsers/generic_findings_import.md
@@ -9,6 +9,7 @@ You can use Generic Findings Import as a method to ingest JSON or CSV files into
 Files uploaded using Generic Findings Import must conform to the accepted format with respect to CSV column headers / JSON attributes.
 
 These attributes are supported for CSV:
+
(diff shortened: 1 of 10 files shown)


blakeaowens (approved):

hblankenship (approved):

kiblik:
It looks that this PR allow to change test name (in `update_test_from_internal_test`). But it does nothing with test type.
Am I reading this correctly?
So #10219 would still be in place. Is there any chance to solve it as well?

Maffooch:
@kiblik good catch! Overwriting the name was actually not supported before this PR, so it is would be best to maintain that in this PR. 

As for the test type name not being overwritten, I have some thoughts that I will share on the issue to keep things all in the same placee

PR #12354, opened 2025-04-30
fix(GHA): Avoid some actions in forks

kiblik:
Disable execution of specific GHA in forks.
They would just fail, and it just generates noise.


Maffooch (approved):

dogboat (approved):

hblankenship (approved):

mtesauro (approved):
Approved

PR #12355, opened 2025-04-30
Urllib3 upgrade + Rerecord JIra responses

Maffooch:
Initially, we saw that tests were failing in #12218 on the jira tests, so I had asserted that the issue was with the jira client based on the stack traces there found in the test. As it turns out, testing the jira instance creation and finding push flows manually on urllib 2.4.0 was a success, so that meant the unit tests may be the issue. Sure enough, rerecording the tests allowed for successful tests execution 

Big thanks to @dogboat for doing the testing to figure out the jira client is not the issue here!

[sc-11031]

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -38,7 +38,7 @@ pytz==2025.1
 redis==5.2.1
 requests==2.32.3
 sqlalchemy==2.0.40  # Required by Celery broker transport
-urllib3==1.26.20
+urllib3==2.4.0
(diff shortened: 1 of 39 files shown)


mtesauro (approved):
Approved

valentijnscholten (approved):

dogboat (approved):

hblankenship (approved):

PR #12361, opened 2025-05-01
feat(RiskAcc-Eng): Clean RiskAcc-Eng mapping + Allow correct handling by API

kiblik:
This PR is trying to fix "not the best" implementation of mapping between Engagements and RiskAcc.
Until now, it was set as `ManyToManyField` in `Engagements`, and from `RiskAcc`, it was just using the first possible match. This implementation is eventually a one-to-many relation, which should be solved via `ForeignKey` in `RiskAcc`.


kiblik:
@mtesauro & @Maffooch, this PR will need more work. But I prepared an initial commit to open discussion: Was there any specific reason why this was implemented as `ManyToManyField`? It just consumes (a bit) more resources and performance, + adds more operations. Have I missed sth?

valentijnscholten:
I think the initial idea was that a Risk Acceptance could apply to more than just one engagement. For example accept the risk of a CVE that occurs in multiple branches. I think via the UI you cannot actually do this, but it might be possible via the API?
There's a risk we kill someone's current use case, or our own future use cases. All in all changing it to a 1-to-many relationship doesn't really provide huge savings or performance gains, so I think it should be left as is.

kiblik:
@mtesauro & @Maffooch, can we have your opinion on which way we should move forward? Thank you.

Maffooch:
Hi @kiblik, thank you for pausing on this feature before continuing any further. Risk Exceptions are one of those features that is used a little differently by everyone. As Val pointed out, we don't want to break any current use cases, or point ourselves into a corner for future use cases. Minimizing changes to the risk exception object would be the best move we could make at this time

kiblik:
> Hi @kiblik, thank you for pausing on this feature before continuing any further. Risk Exceptions are one of those features that is used a little differently by everyone. As Val pointed out, we don't want to break any current use cases, or point ourselves into a corner for future use cases. Minimizing changes to the risk exception object would be the best move we could make at this time

Thank you @Maffooch. I will prepare the implementation, which will:
- Behave the same as you can experience in UI
- Enable full management via API (it is not possible right now)
- Simplify data structure
- Keep the door open to change it to "less strict" implementation in the future

vini-ppro:
Hi all, I am the author of the bug that triggered this discussion.

@valentijnscholten - as for your question "it might be possible via the API?" I tried that, but I got an error that is not possible to have findings from different engagements under the same risk acceptance, so there is a validation somewhere.

As for the general use case for Risk Acceptances. It is quite common for us to accept the risk for a certain CVE across the whole business. In our view, **it should be possible to have a single Risk Acceptance comprising many findings across many engagements**. I would like to think this is a very common requirement for many companies

Because it is not possible to have a single Risk Acceptance across engagements, we had to implement workarounds. We created a script that creates one Risk Acceptance per finding. It could happen that we create more than 1k Risk Acceptances for a single CVE, this is not ideal, we may clog the DB at some point.

kiblik:
> Hi all, I am the author of the bug that triggered this discussion.
> 
> @valentijnscholten - as for your question "it might be possible via the API?" I tried that, but I got an error that is not possible to have findings from different engagements under the same risk acceptance, so there is a validation somewhere.
> 
> As for the general use case for Risk Acceptances. It is quite common for us to accept the risk for a certain CVE across the whole business. In our view, **it should be possible to have a single Risk Acceptance comprising many findings across many engagements**. I would like to think this is a very common requirement for many companies
> 
> Because it is not possible to have a single Risk Acceptance across engagements, we had to implement workarounds. We created a script that creates one Risk Acceptance per finding. It could happen that we create more than 1k Risk Acceptances for a single CVE, this is not ideal, we may clog the DB at some point.

I hope you are not the author of the bug, but only the author of the bug report 😄 

You are writing at the best time. I almost finished this PR (only one validation statement and a fix of unittests are missing). I wrote it in a way that reflects current behaviour (allows mapping RiskAcc only to engagements), but correctly handled in case of API processing.
I went to this road because _"we don't want to break any current use cases, or point ourselves into a corner for future use cases. Minimizing changes to the risk exception object would be the best move we could make at this time"_.

On the other hand, I agree that if there is no specific use case (and I do not know it), I suppose it might be a much better idea to move RiskAcc from the Engagement level to the Product level. Cross-product level would not be the best because I'm not sure how we would handle permissions there (e.g., Can I see all RAs? Because if I want to add my findings to some pre-created RA, I need to be able to see it).

@Maffooch, @valentijnscholten, @mtesauro, can I ask you for reconsideration? Change from the Engagement level to the Product level would not break any correct implementation; it would just open the door for a more flexible definition/handling.

Arthur-DTAG:
We also have this use case:
> ... **it should be possible to have a single Risk Acceptance comprising many findings across many engagements**...

E.g. you track multiple versions of a product in different engagement (within the same product or multiple). Then a new CVE gets reported. You investigate and see that it has no effect to you. Therefor We would like to accept it for all installed Versions (so multiple engagements) as single Risk Acceptance since it has been evaluated together. Depending on the Application/System, this can affect many CVE (e.g. because it is not exposed).

kiblik:
> @Maffooch, @valentijnscholten, @mtesauro, can I ask you for reconsideration? Change from the Engagement level to the Product level would not break any correct implementation; it would just open the door for a more flexible definition/handling.

@Maffooch, @valentijnscholten, @mtesauro, can I know your opinion?

twwd:
@Maffooch why was this closed? The bug #12644 still exists.

valentijnscholten:
Looks like this PR tried to not only refactor the Risk Acceptance database model, but also fix the linked bug. We will need a new separate PR for the bug.

Issue #12362, opened 2025-05-01
AnchoreCTL Policies Report Parser error


Sopuru:
AnchoreCTL Policies Report Parser
I downloaded policy report via anchorectl using the recommended -o json flag, tried importing that into DefectDojo but it came back with an error. Can this parser be updated to accepted policy generate via anchorectl? 

See attached example of the data used and error message i got:

Sample File
[Sample Policy report attached](https://github.com/user-attachments/files/20000758/ctlpol.json)

![Error](https://github.com/user-attachments/assets/9c8d4063-4a67-496b-bc2e-fd111dc33c84)

valentijnscholten:
That attached report format doesn't match the format expected by the parser: https://docs.defectdojo.com/en/connecting_your_tools/parsers/file/anchorectl_policies/

Has the format changed, or is the attached report from another anchor tool maybe? cc @fhoeborn 

There are other Anchore formats, but none of them look like the attached report:
- https://docs.defectdojo.com/en/connecting_your_tools/parsers/file/anchore_enterprise/
- https://docs.defectdojo.com/en/connecting_your_tools/parsers/file/anchore_engine/
- https://docs.defectdojo.com/en/connecting_your_tools/parsers/file/anchore_grype/
- https://docs.defectdojo.com/en/connecting_your_tools/parsers/file/anchorectl_vulns/

I raised #12375 to improve the error handling/feedback to the user.

fhoeborn:
I haven't been in touch with Anchore / AnchoreCTL since a while - looking at the release notes of anchoreCTL there have been some releases in between (with major jumps) - so I guess the format changed

PR #12375, opened 2025-05-04
anchorectl: add format check

valentijnscholten:
Based on #12362 adding some format checks to make it more clear when the report format doesn't match the parsers expectation. Also removed the silent swallowing of exceptions to avoid the non-informative "No findings were added/updated/closed/reactivated as the report is empty or the findings in Defect Dojo are identical to those in the uploaded report" when an exception occurs.


mtesauro (approved):
Approved

Maffooch (approved):

dogboat (approved):

hblankenship (approved):

PR #12400, opened 2025-05-07
tags: prevent validation from removing tags

### LINKED ISSUE #12394: Tags are not saved
**Bug description**
When editing or creating Product, Findings, etc. on version 2.46, Tag field is no longer saved. 

**Steps to reproduce**
1. Create or edit a product, finding, engagement, etc.
2. Add one or more tags
3. Save

**Expected behavior**
On saving, tags should be saved also. This happens on my side and also on demo platform

valentijnscholten:
Fixes #12394 where any save action via the UI will clear the `tags` field on the saved object.

Any `clean_<fieldname>` method in Django must return the cleaned value of the field. If nothing is returned, the field will become empty. https://docs.djangoproject.com/en/5.2/ref/forms/validation/

Confirmed validation is working OK:
image

I tried to add a simple smoke test for the tags field in the UI tests, but it's not trivial as the tags field is dynamically generated on the client-side.

REST API validation and modification of objects is working fine (has test cases)

### DIFF
--- a/dojo/forms.py
+++ b/dojo/forms.py
@@ -341,6 +341,7 @@ class ProductForm(forms.ModelForm):
 
     def clean_tags(self):
         tag_validator(self.cleaned_data.get("tags"))
+        return self.cleaned_data.get("tags")
 
 
 class DeleteProductForm(forms.ModelForm):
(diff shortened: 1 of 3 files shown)


mtesauro (approved):
Approved

dogboat (approved):

blakeaowens (approved):

hblankenship (approved):

Issue #12397, opened 2025-05-07
nuclei incorrect deduplication


egorbeliy:
**Bug description**
Nuclei findings with different Endpoints but same other finding fields (component name, cve, etc)  marked as duplicates after being imported

**Steps to reproduce**

1. do scan of amazon.com and google.com using nuclei
`nuclei -list targets.txt -j -s low -o output.json`

[targets.txt](https://github.com/user-attachments/files/20085334/targets.txt)
sample of output has been attached

2. import results to defectdojo
```
curl -s -X POST "$DEFECTDOJO_URL/api/v2/import-scan/" \
  -H "Authorization: Token $DEFECTDOJO_TOKEN" \
  -F "scan_date=$(date +%F)" \
  -F "minimum_severity=Low" \
  -F "scan_type=Nuclei Scan" \
  -F "file=@$NUCLEI_OUTPUT;type=application/json" \
  -F "engagement_name=Passive scan" \
  -F "product_name=Passive scan" \
  -F "verified=false" \
  -F "active=true" \
  -F "auto_create_context=true" \
  -F "auto_create=true" \
  -F "product_type_name=Research and Development" \
  -F "close_old_findings=false" \
  -F "push_to_jira=false"
```
3. google.com findings are active, amazon.com findings being closed as duplicate
<img width="1566" alt="Image" src="https://github.com/user-attachments/assets/36e77bd8-0724-4a7b-b32b-e44512693bad" />

3. 
**Expected behavior**
Amazon findings aren't duplicates, 2 findings with 2 endpoints in each should be created

**Environment information**
 - Defectdojo version - v. 2.44.2
 - Nuclei version - v3.4.2

manuel-sommer:
Could you check the PR @egorbeliy ?

egorbeliy:
i guess looks good

PR #12405, opened 2025-05-07
:bug: fix Nuclei deduplication #12397

manuel-sommer:
#12397

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -1313,7 +1313,7 @@ HASHCODE_FIELDS_PER_SCANNER = {
     "NeuVector (compliance)": ["title", "vuln_id_from_tool", "description"],
     "Wpscan": ["title", "description", "severity"],
     "Popeye Scan": ["title", "description"],
-    "Nuclei Scan": ["title", "cwe", "severity", "component_name"],
+    "Nuclei Scan": ["title", "cwe", "severity", "component_name", "endpoints"],
     "KubeHunter Scan": ["title", "description"],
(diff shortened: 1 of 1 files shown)


valentijnscholten (commented):
Can you put a one liner in the description to make it clear what is fixed?

Also could you explain why it is needed to add the `endpoints` to the hash codes. I haven't looked at the source code, but my understanding is that endpoints are always considered, even when they are not explicitly part of the hash code fields:

https://github.com/DefectDojo/django-DefectDojo/blob/96a7cb4c5cb3564f301c579b7e0444aba6d7b8db/dojo/settings/settings.dist.py#L1450-L1459

Maybe the parser must be updated to mark the findings as dynamic?

Maffooch:
>Maybe the parser must be updated to mark the findings as dynamic?

I was thinking something similar. Some parsers aggregate endpoints by their finding. I think that should be implemented here as well. Tenable is great example of this use case

manuel-sommer:
Hm, I am not really sure how to proceed here. You are right @valentijnscholten, I wasn't aware that endpoints are always considered. Furthermore, the Finding is dynamic by default: https://github.com/DefectDojo/django-DefectDojo/blob/924c2c88645ad555193f9122c5eba2dc7bd5c65c/dojo/models.py#L2542

egorbeliy:
So, it means that if the findings have different endpoints, they shouldn't be closed as duplicates. But based on my testing and the screenshot I provided, they are still being closed as duplicates

Maffooch:
Closing as stale

PR #12406, opened 2025-05-07
Managed Files: Sanitized file name before downloading

Maffooch:
When downloading files with quotes in the title stored in DefectDojo, some browsers truncate the name of file, while others replace the problematic characters with a hyphen. I think the replacement mechanism is the most sound one as it leaves the file in a workable state still

[sc-11081]


valentijnscholten (approved):

mtesauro (approved):
Approved

PR #12408, opened 2025-05-07
Checkmarx one doc update

skywalke34:
**Description**
Checkmarx one documentation update

**Test results**
no tests

**Documentation**
Documentation update only.

### DIFF
--- a/docs/content/en/connecting_your_tools/parsers/file/checkmarx_one.md
+++ b/docs/content/en/connecting_your_tools/parsers/file/checkmarx_one.md
@@ -33,54 +33,54 @@ The parser can handle several variations of the Checkmarx One JSON output format
 
 ### Standard Format Field Mapping Details
 
-| Data Field # | Checkmarx Data Field | DefectDojo Finding Field | Parser Line # | Notes |
-|--------------|----------------------|--------------------------|---------------|-------|
-| 1 | type | unsaved_tags | 361-368 | Added as a tag identifying the finding type (sast, kics, sca, etc.) |
-| 2 | id | unique_id_from_tool | 342, 392, 409 | Primary unique identifier for the finding |
-| 3 | similarityId | unique_id_from_tool | 342, 392, 409 | Used as fallback if id not present |
-| 4 | status | - | 374 | Used for state determination but not directly mapped |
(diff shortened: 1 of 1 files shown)


valentijnscholten (commented):
I think it can be very helpful to document which fields are parsed and how. Could eventually become part of the "how to write a parser" guide?

One thing I am wondering about is the line numbers. These could potentially change slightly on small changes or non-functional changes like linting or refactorings. Does it bring a lot of value to put these line numbers in the docs? My initial thoughts that when people use "CTRL+F" on the parser file they'll quickly find the section for a specific field?

Maffooch (approved):
Comment from Val is spot on
> One thing I am wondering about is the line numbers. These could potentially change slightly on small changes or non-functional changes like linting or refactorings. Does it bring a lot of value to put these line numbers in the docs? My initial thoughts that when people use "CTRL+F" on the parser file they'll quickly find the section for a specific field?

I wouldn't consider this a blocker, but something to by mindful about in the future

skywalke34:
> I think it can be very helpful to document which fields are parsed and how. Could eventually become part of the "how to write a parser" guide?
> 
> One thing I am wondering about is the line numbers. These could potentially change slightly on small changes or non-functional changes like linting or refactorings. Does it bring a lot of value to put these line numbers in the docs? My initial thoughts that when people use "CTRL+F" on the parser file they'll quickly find the section for a specific field?

I too have questioned and am uncertain of the value of including line numbers vs trying to identify the function, and agree your thoughts are completely valid.  The reason I decided to go ahead and try to include line numbers - even if a single function from lines X to Y have all the parsing logic or line numbers are changed from linting, etc - is for future analysis by an AI LLM.  I am learning there is a wide range of approaches each individual parser can take to process the same fields into the same finding object.  My thinking is if the documentation contains enough detail about which data fields are being parsed, including line numbers in the parser.py code, it should be possible to perform AI analysis comparing all the parsers, the fields they parse, and how they are parsing them - which may help us come up with a more standardized approach to coach parser developers how to parse specific fields, best practices, etc. I could just be taking it "too far"! :)  ¯\_(ツ)_/¯

skywalke34:
Carefully removed references to line numbers.

hblankenship (approved):

mtesauro (approved):
Approved

valentijnscholten (approved):

PR #12409, opened 2025-05-07
Updated Nexpose XML (Rapid7) Parser Documentation

skywalke34:
**Description**
Update Nexpose XML (Rapid7) Parser documentation including Field Mapping Details, Special Processing Notes, Conversion details, and links.

**Test results**
No tests changed

**Documentation**
Documentation update only.

### DIFF
--- a/docs/content/en/connecting_your_tools/parsers/file/nexpose.md
+++ b/docs/content/en/connecting_your_tools/parsers/file/nexpose.md
@@ -2,7 +2,109 @@
 title: "Nexpose XML 2.0 (Rapid7)"
 toc_hide: true
 ---
-Use the full XML export template from Nexpose.
+
+# Nexpose XML 2.0 (Rapid7) Parser Documentation
+
+## Overview
+
+The Nexpose XML 2.0 (Rapid7) parser for DefectDojo supports importing vulnerability scan results from Rapid7's Nexpose vulnerability scanner. This parser processes XML reports containing detailed information about vulnerabilities detected across hosts and services in your environment.
+
+## Supported File Types
+
+The Nexpose parser accepts XML 2.0 file format. To generate this file:
+
+1. In the Nexpose console, go to the Reports section
+2. Create a new report or edit an existing one
+3. Select "XML Export" as the report format
+4. Ensure you select "XML 2.0" as the version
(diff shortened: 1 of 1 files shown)


Maffooch (approved):
Approving for now. Great suggestion from @kiblik 

@skywalke34 please take a look at that comment

dogboat (approved):

valentijnscholten (approved):

hblankenship (approved):

skywalke34:
Rapid7 Insight VM generates reports both from build-in report templates or custom SQL query reports.  As I do not have access to a Rapid7 environment, if I include information on Insight VM, is it possible to know which standard report templates (i.e. "Top Remediation Report") that can be imported and parsed without error?  Should I also include information that custom reports are not supported?

kiblik:
> > If we are doing this kind of rewrite, it might be a good idea to add information about https://docs.rapid7.com/insightvm/ InsightVM is a Nexpose instance connected to the cloud. The steps for obtaining the report are the same. Also, the report is the same.
> 
> Rapid7 Insight VM generates reports both from build-in report templates or custom SQL query reports. As I do not have access to a Rapid7 environment, if I include information on Insight VM, is it possible to know which standard report templates (i.e. "Top Remediation Report") that can be imported and parsed without error? Should I also include information that custom reports are not supported?

As far as I know, Nexpose and InsightVM use (almost) the same Console UI, and they share most of the documentation.
DD support _XML Export 2.0_ report generated by Nexpose (successfully tested). This should be true about InsightVM as well.
_Top Remediations_ is a different kind of report. The output is PDF.
_XML Export 2.0_ is information which should user look for in _Template_ section (see pictures in https://docs.rapid7.com/insightvm/creating-a-basic-report)

Screenshot 2025-05-09 at 17 02 46
Screenshot 2025-05-09 at 17 10 14

- https://docs.rapid7.com/nexpose/working-with-report-formats/ 
- https://docs.rapid7.com/insightvm/working-with-report-formats/

Maffooch:
> As far as I know, Nexpose and InsightVM use (almost) the same Console UI

Unless we are certain of this, I think we should leave off the insightVM note

valentijnscholten (approved):

PR #1294, opened 2019-06-25
Endpoint remedy

Maffooch:
In reference to issue #1027 

Endpoints can be remediated or deleted individually or in bulk. Reports will now reflect differences between remediated and vulnerable endpoints by separating endpoints into different sections.  

While tweaking reports, I followed advice given [here](https://github.com/DefectDojo/django-DefectDojo/pull/1266#issuecomment-504676672) and corrected the behavior of private notes. The notes section of a report will only appear if there is at least one public note. The section will not appear at all with exclusive private notes or no notes at all.

- [x] Your code is flake8 compliant 
- [x] If this is a new feature and not a bug fix, you've included the proper documentation in the ReadTheDocs documentation folder. https://github.com/DefectDojo/Documentation/tree/master/docs or provide feature documentation in the PR.
- [x] Model changes must include the necessary migrations in the dojo/dd_migrations folder.
- [ ] Add applicable tests to the unit tests.

PR #12434, opened 2025-05-12
Tags: Add support for comma separation for multipart forms (import/reimport)

Maffooch:
When attempting to attach multiple tags during a reimport through the swagger docs page, the curl request that is generated places all submitted tags in the same form field in a comma separated list. In the recent changes from #1294, commas are not longer allowed. This PR allowed for tag enforcement in lists, but not when a string is passed. Unfortunately correcting the generated curl request is not really feasible, so we must return to approach of splitting all tags from the API by commas.

Slack Context: https://defectdojo-workspace.slack.com/archives/C2P5BA8MN/p1747055360228049 
[sc-11116]

### DIFF
--- a/unittests/test_tags.py
+++ b/unittests/test_tags.py
@@ -163,10 +163,6 @@ class TagTests(DojoAPITestCase):
     def test_finding_patch_remove_tags_non_existent(self):
         return self.test_finding_put_remove_tags_non_existent()
 
-    def test_finding_create_tags_with_commas(self):
-        tags = ["one,two"]
-        self.create_finding_with_tags(tags, expected_status_code=400)
-
     def test_finding_create_tags_with_spaces(self):
         tags = ["one two"]
         self.create_finding_with_tags(tags, expected_status_code=400)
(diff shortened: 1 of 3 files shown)


unittests/test_tags.py
```diff
@@ -212,6 +208,25 @@ def test_import_and_reimport_with_tags(self):
         for tag in tags:
             self.assertIn(tag, response["tags"])
 
+    def test_import_reimport_multipart_tags(self):
```
valentijnscholten, line 211: The test only tests `import`, 
```suggestion
    def test_import_multipart_tags(self):
```

valentijnscholten: But is this test really testing something different than: 

https://github.com/DefectDojo/django-DefectDojo/blob/05a74d5279e1d910b0f6bb4f6c1300a23f255e61/unittests/test_tags.py#L192-L213 ?

valentijnscholten (changes requested):

unittests/test_tags.py
Maffooch: Yep, it is testing tags with a comma in one of the tags where the tag came from a string rather than a list. This is a good indicator that the tags were supplied from a multipart form

valentijnscholten: Ah yes.

valentijnscholten (approved):

hblankenship (approved):

dogboat (approved):

mtesauro (approved):
Approved

PR #12435, opened 2025-05-12
excel export: enhance handling of finding groups, better logging

valentijnscholten:
Enhance logging for errors during Excel exports. Or rather they are warnings, so we log them at `warning` level from now on.

Inspired by #11911

Added later: Some more testing lead me to what I believe is a copy-and-paste error where `get_foreign_keys()` was not called. 

So the PR now also fixes the conversion problem and outputs finding group values correctly:

image

### DIFF
--- a/dojo/reports/views.py
+++ b/dojo/reports/views.py
@@ -944,6 +944,7 @@ class ExcelExportView(View):
 
         row_num = 1
         for finding in findings:
+            logger.debug(f"processing finding: {finding.id}")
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

Maffooch (approved):

hblankenship (approved):

dogboat (approved):

mtesauro (approved):
Approved

PR #12439, opened 2025-05-13
UI Pagination: Reduce the options to more reasonable numbers

Maffooch:
In most web applications, the dropdown options for page size rarely exceed 50, but DefectDojo had an option for 500, and even "All" results. When working in an instance with many users, fetching all results in a single page could cause slower response times for other users using the application at the same time. To accommodate, the highest page in the drop down will be 250

Note: This change does not enforce a page size limit of any kind, but simply removes the advertisement of the larger page sizes

[sc-11117]


hblankenship (approved):

dogboat (approved):

valentijnscholten (approved):

mtesauro (approved):
Approved

Issue #8264, opened 2023-06-16
CVSS Regex in models.py is not good


HelpMe-AC:
Hi Team !

I'm writing a new API lib to import data from specific tool.
When calling "Finding" model to import vulnerability in DefectDojo, an error occurs as the "cvssv3_regex" is not consistent with cvss library used by DefectDojo.

The regex used is at line 2009 : cvssv3_regex = RegexValidator(regex=r'^AV:[NALP]|AC:[LH]|PR:[UNLH]|UI:[NR]|S:[UC]|[CIA]:[NLH]', message="CVSS must be entered in format: 'AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:H'").

The 'cvss' lib is able to handle CVSSv2 and CVSSv3. For this to work, we have to prefix the CVSS vector with "CVSS:3.0/" which is not accepted by the regex in defectdojo/dojo/models.py.
Also the slash is not in the regex, so need to add it at the end of the regex.

I suggest this improvement : "cvssv3_regex = RegexValidator(regex=r'^CVSS:3.0\\/AV:[NALP]|AC:[LH]|PR:[UNLH]|UI:[NR]|S:[UC]|[CIA]:[NLH]|\\/', message="CVSS must be entered in format: 'AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:H'")".

So importing an existing CVSSv3 vector from a vuln tool is not possible and CVSSv3 score will be not calculated.

**Steps to reproduce**
Import a CVSSv3 vector by using 'Finding' model of DefectDojo.

**Deployment method** *(select with an `X`)*
- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

damiencarol:
wer shouyld use the package `cvss`

valentijnscholten:
@HelpMe-AC Thank you for reporting. We agree that the RegEx validator is not the right approach. But we're also seeing that in most cases it is possible to store the CVSSv3 vector in Defect Dojo. With or without prefix and with or without trailing slashes.
What we do see is that the `cvssv3_score` is only calculated if the `cvssv3` field contains a vector that is considered valid by the `cvss` library, which means it must have a prefix.
Is that what you also see? Your report above reads as if you are unable to store any vectors with a prefix in Defect Dojo. We have some test cases for this coming in #12457 
I'm not trying to argue, just tyring to fully understand the report. 

[...]

PR #12440, opened 2025-05-13
Improve cvssv3 validation

valentijnscholten:
Although it's not fully clear "what the problem is" reported in #8264 , we want to stop using the regex for validation and use the `cvss` library in a validator. This PR updates the validator on the model and updates the test cases to reflect the new behaviour. The new behaviour is better:

- When saving via API or UI invalid vectors are recjted. (2 vectors, v3 vectors without a prefix, v3 vectors with a trailing slash, v4 vectors, ...)
- When `Finding.save()` is called, no validation is perfrmed by Django. So we have make sure the parsers only store valid vectors.

I updated the parsers that did not yet follow the guidelines in `docs/content/en/open_source/contributing/how-to-write-a-parser.md`. I considered:
- adding a `set_data_from_cvss_string(...)` to the Finding model
- adding code to `Finding.save()` to validate/parse the vector string

But in the end I went for a helper method in `dojo.utils` to parse the CVSS vector. This way we have the parsing logic in one central place and the parsers can decide which of the 3 parsed fields they want to use on the model (`vector`, `score`, `severity`). The logic is also failsafe, invalid vectors result in empty cvss fields but won't cause the import to fail.

Remaining questions:
- Do we want to keep the logic in place in `Finding.save()` that sets (overwrites) the `cvssv3_score` if a valid vector is set? (I think for now YES)
- What to we do when `Finding.save()` encounters an invalid vector in `cvssv3`. Should we clear the vector before saving? (I think for now NO)
- Do we want to make sure all parsers use the same/new parse function? (I think for now NO as long as they parse the vector correctly)

Risks:
- The parsers are fail-safe, but users submitting invalid CVSSv3 vectors through the API may run into HTTP_400 validation errors. But I'm not sure if that is a bad thing?

Feedback welcome. I also think this is a nice preparation for when we want to support CVSS4 vectors.


docs/content/en/open_source/contributing/how-to-write-a-parser.md
```diff
@@ -168,14 +168,22 @@ Good example:
...
+```
 
-It's easy to use and will make the parser aligned with the rest of the code.
+If you need more manual processing, you can parse the `CVSS` vector directly.
 
 Example of use:
 
 ```python
-from cvss.cvss3 import CVSS3
+from dojo.utils import cvss.cvss3 import CVSS3
```
Maffooch, line 186: Not sure which import should be the correct one

dojo/models.py
```diff
@@ -2699,11 +2699,15 @@ def save(self, dedupe_option=True, rules_option=True, product_grading_option=Tru
...
+
+                cvss_data = parse_cvss_data(self.cvssv3)
+                if cvss_data:
+                    self.cvssv3 = cvss_data.get("vector")
+                    self.cvssv3_score = cvss_data.get("score")
+
             except Exception as ex:
-                logger.error("Can't compute cvssv3 score for finding id %i. Invalid cvssv3 vector found: '%s'. Exception: %s", self.id, self.cvssv3, ex)
+                logger.warning("Can't compute cvssv3 score for finding id %i. Invalid cvssv3 vector found: '%s'. Exception: %s.", self.id, self.cvssv3, ex)
+                # should we set self.cvssv3 to None here to avoid storing invalid vectors? it would also remove invalid vectors on existing findings...
```
Maffooch, line 2710: I think we should remove invalid vectors, but only if the id is `None` before the finding is saved. That allows us to prevent the issue from spreading even further without implications for existing data

dojo/tools/npm_audit_7_plus/parser.py
```diff
@@ -166,7 +166,11 @@ def get_item(item_node, tree, test):
         dojo_finding.cwe = cwe
 
     if (cvssv3 is not None) and (len(cvssv3) > 0):
-        dojo_finding.cvssv3 = cvssv3
+        from dojo.utils import parse_cvss_data
```
Maffooch, line 169: This should be a the tope of the file

dojo/validators.py
```diff
@@ -0,0 +1,53 @@
...
+    if len(cvss_vectors) > 0:
+        vector_obj = cvss_vectors[0]
+
+        if isinstance(vector_obj, CVSS3):
+            # all is good
+            return
+
+        if isinstance(vector_obj, CVSS4):
+            # CVSS4 is not supported yet by the parse_cvss_from_text function, but let's prepare for it anyway: https://github.com/RedHatProductSecurity/cvss/issues/53
+            msg = "Unsupported CVSS(4) version detected."
```
Maffooch, line 41: Nice job prepping for this in the future

dojo/db_migrations/0229_alter_finding_cvssv3_alter_finding_template_cvssv3.py
Maffooch, line 1: This will need to be synced up as well

Maffooch (changes requested):
This is looking really good - only a couple questions/nits

docs/content/en/open_source/contributing/how-to-write-a-parser.md
valentijnscholten: addressed this

dojo/models.py
valentijnscholten: addressed this, but maybe we should just throw a ValidationError? That is what will happen in the future if we implement better validation support for Findings in general via `clean`.

dojo/db_migrations/0229_alter_finding_cvssv3_alter_finding_template_cvssv3.py
valentijnscholten: done and rebased branch

dojo/models.py
Maffooch: actually, I think the validator will do this already, as its called before the save method is, right?

valentijnscholten: I believe only when setting a CVSS3 vector through the UI or API, but not when a (re)import is done.

Maffooch (approved):
Let's get this in dev for a bit and see where we land

mtesauro (approved):
Approved

valentijnscholten:
I will resolve the conflicts after 2.47.3 is release and merged back into dev. There might be more conflicts by then and I don't want to burden the release process with merge conflicts by merging this PR to dev before 2.47.3.

blakeaowens (approved):

PR #12446, opened 2025-05-14
Wizcli improvements

OsamaMahmood:
**Description**

1. Core Functionality:
  - Implemented a robust parsing system for different types of Wiz CLI scan results:
    - Library vulnerabilities
    - Secrets detection
    - OS package vulnerabilities
    - Infrastructure as Code (IaC) rule matches
 
2. Deduplication Improvements:
  - Enhanced the `_generate_unique_id` method to ensure consistent finding deduplication using `unique_id_from_tool`:
    - Now uses sorted components for stable hash generation
    - Properly handles None values and whitespace
    - Fixed docstring formatting and code style issues
    - Components are now consistently ordered regardless of input order
  
**Checklist**

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -1493,6 +1493,9 @@ DEDUPLICATION_ALGORITHM_PER_PARSER = {
     "Kubescape JSON Importer": DEDUPE_ALGO_HASH_CODE,
     "Kiuwan SCA Scan": DEDUPE_ALGO_HASH_CODE,
     "Rapplex Scan": DEDUPE_ALGO_HASH_CODE,
+    "Wizcli Img Scan": DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL,
+    "Wizcli Dir Scan": DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL,
+    "Wizcli IAC Scan": DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL,
 }
 
 # Override the hardcoded settings here via the env var
--- a/dojo/tools/wizcli_common_parsers/parsers.py
+++ b/dojo/tools/wizcli_common_parsers/parsers.py
@@ -32,19 +32,106 @@ class WizcliParsers:
         match = re.search(r"(https?://[^\s)]+)", text)
         return match.group(1) if match else None
 
-                    finding_description = (
-                        f"**Library Name**: {lib_name}\n"
-                        f"**Library Version**: {lib_version}\n"
-                        f"**Library Path**: {lib_path}\n"
-                        f"**Vulnerability Name**: {vuln_name}\n"
-                        f"**Fixed Version**: {fixed_version}\n"
-                        f"**Source**: {source}\n"
-                        f"**Description**: {description}\n"
-                        f"**Score**: {score}\n"
-                        f"**Exploitability Score**: {exploitability_score}\n"
-                        f"**Has Exploit**: {has_exploit}\n"
-                        f"**Has CISA KEV Exploit**: {has_cisa_kev_exploit}\n"
-                    )
+    @staticmethod
+    def _generate_unique_id(components: list) -> str:
+        """
+        Generates a stable unique ID for findings.
+
+        Args:
+            components: List of components to use for ID generation
+
+        """
+        # Filter out None and empty values
+        filtered_components = [str(c).strip() for c in components if c is not None and str(c).strip()]
+
+        # Sort components for consistent order regardless of input order
+        filtered_components = sorted(filtered_components)
+
+        id_string = "|".join(filtered_components)
+        hash_object = hashlib.sha256(id_string.encode("utf-8"))
+        return hash_object.hexdigest()
+
+    @staticmethod
+    def parse_libraries(libraries_data, test):
+        """
+        Parses library vulnerability data into granular DefectDojo findings.
+        Creates one finding per unique vulnerability (CVE/ID) per library instance (name/version/path).
+        """
+        findings_list = []
+        if not libraries_data:
+            return findings_list
+
+        for lib_item in libraries_data:
+            lib_name = lib_item.get("name", "N/A")
+            lib_version = lib_item.get("version", "N/A")
+            lib_path = lib_item.get("path", "N/A")
+            lib_line = lib_item.get("startLine")
+
+            vulnerabilities_in_lib_instance = lib_item.get("vulnerabilities", [])
+            if not vulnerabilities_in_lib_instance:
+                continue
+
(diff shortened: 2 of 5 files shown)


valentijnscholten (changes requested):
Thank you @OsamaMahmood for your extensive PR. We do have some feedback:

- Could you look at updating the tests/samples scans to reflect the updates to the parsers?
- Could you look using the hash code configuration for deduplication?

I just raised #12463 to clarify the use of the `unique_id_from_tool` field. It's intended/accepted use is to contain value present in the report that can be used to recognize the finding inside the tool. And for strong and exact deduplication. 
We will discuss internally if/how we can accomodate values computed by the parser that might be useful for deduplication.

OsamaMahmood:
> Could you look using the hash code configuration for deduplication?

Hi @valentijnscholten i have updated the `setting.dist.py` to use Hash Code for de-duplication the results are same what i am getting by using `unique_id_from_tool`, findings are only getting de-duplicated in  `Wizcli Dir Scan` and `Wizcli Img Scan` but for `Wizcli IAC scan` its still the same even if i upload the same report it closes some findings even though its the same report.

dojo/tools/wizcli_common_parsers/parsers.py
```diff
@@ -1,190 +1,441 @@
...
+                    mitigation=mitigation,
+                    file_path=lib_path,
+                    line=lib_line if lib_line is not None else 0,
+                    component_name=lib_name,
+                    component_version=lib_version,
                     static_finding=True,
                     dynamic_finding=False,
-                    mitigation=None,
-                    test=test,
+                    unique_id_from_tool=unique_id,
```
valentijnscholten: This field can only contain values from the report, not generated/constructed/computed values.

dojo/settings/settings.dist.py
```diff
@@ -1591,6 +1591,9 @@ def saml2_attrib_map_format(din):
     "Kubescape JSON Importer": DEDUPE_ALGO_HASH_CODE,
     "Kiuwan SCA Scan": DEDUPE_ALGO_HASH_CODE,
     "Rapplex Scan": DEDUPE_ALGO_HASH_CODE,
+    "Wizcli Img Scan": DEDUPE_ALGO_HASH_CODE,
+    "Wizcli Dir Scan": DEDUPE_ALGO_HASH_CODE,
+    "Wizcli IAC Scan": DEDUPE_ALGO_HASH_CODE,
```
valentijnscholten: You probably also want to set the `HASHCODE_FIELDS_PER_SCANNER` entries for these parser to finetune the hash code algorithm and deduplication results.

dojo/tools/wizcli_common_parsers/parsers.py
OsamaMahmood: done

dojo/settings/settings.dist.py
OsamaMahmood: updated

dojo/settings/settings.dist.py
```diff
@@ -1345,6 +1345,9 @@ def saml2_attrib_map_format(din):
     "Red Hat Satellite": ["description", "severity"],
     "Qualys Hacker Guardian Scan": ["title", "severity", "description"],
     "Cyberwatch scan (Galeax)": ["title", "description", "severity"],
+    "Wizcli Img Scan": ["title", "description", "file_path", "line", "component_name", "component_version"],
+    "Wizcli Dir Scan": ["title", "description", "file_path", "line", "component_name", "component_version"],
+    "Wizcli IAC Scan": ["title", "description", "file_path", "line", "component_name"],
```
valentijnscholten: The description contains values that can sometimes change over time. Is it really needed for dedupe?

valentijnscholten (changes requested):
Thanks two more things:
- see my comment on the hash code fields
- because the dedupe config has changed AND the title is not set differently, this needs some docs in the upgrade notes for 2.47.3. 

Can you add instructions on how to recalculate the hash codes (see other releases to get a starting point). And line that states dedupe can mismatch between findings imported by the new parser versus the old parser (because of the change in values for the title field).

dojo/settings/settings.dist.py
OsamaMahmood: thanks @valentijnscholten for highlighting this fixed it

OsamaMahmood:
hi @valentijnscholten all changes done requesting for review

docs/content/en/open_source/upgrading/2.47.md
```diff
@@ -12,4 +12,15 @@ There are no special instructions for upgrading to 2.47.x. Check the [Release No
 
 ## Removal of Asynchronous Import
 
-Please note that asynchronous import has been removed as it was announced in 2.46. If you haven't migrated from this feature yet, we recommend doing before upgrading to 2.47.0
\ No newline at end of file
+Please note that asynchronous import has been removed as it was announced in 2.46. If you haven't migrated from this feature yet, we recommend doing before upgrading to 2.47.0
+
+## Wizcli Parsers Hashcode Changes
+**Hash Code changes** Wizcli parsers have been updated to populate more fields in release 2.47.3. Some of these fields are part of the hash code calculation. To recalculate the hash code please execute the following command:
```
valentijnscholten: `Wizcli parsers have been updated to populate more fields. Some of these fields are part of the hash code calculation. To recalculate the hash code please execute the commands below. Please note that due to the changes in the parser deduplication might not always happen against findings that were already present in Defect Dojo.`

valentijnscholten: I don't know why but Github doesn't offer me the "Suggest Changes" option here.

valentijnscholten:
@Maffooch @mtesauro Is there already a guideline/agreement on how to handle changes in parsers that affect deduplication? Sometimes recalculating the hash codes is enough. But in this case the title field is changing. Existing findings in Defect Dojo will have a different title and hash code as a result of that. There is note in the upgrade notes.

valentijnscholten:
@OsamaMahmood We have discussed this PR internally and we have some concerns around the changes in the deduplication logic/settings. It will result in new findings being created which are not deduplicated against existing findings during import/reimport. When users are using `close_old_findings`, the existing findings are closed. 

I though of some ways we could avoid this "breaking" change:
1) Update the PR to maintain the same deduplication logic/result. I don't think this is possible since some extra fields are set that are usefule such as `component_name` and `component_version`.
2) Keep the old parsers and three new parsers with suffix " v2".
3) Keep the older pasers and merge the new parsers in this PR into one "Wiz Autodetect" parser that supports all three formats and autodetects the format.

Is there any option of those you think that could work?

Maffooch:
Closing as stale - please feel free to reopen if closing now is a mistake

PR #12448, opened 2025-05-14
defender: fix no vulnerabilities check

valentijnscholten:
The check to bail out early expects a `vulnerabilities/` folder entry to always be present in the zip file if there are files in that folder. We've seen that this is not always the case as some zips do not contain entries for folders, only entries for actual files.
This PR changes the "bail out check" to work well in both scenario's.

Raised by user on Slack: https://owasp.slack.com/archives/C2P5BA8MN/p1747126211915059

### DIFF
--- a/dojo/tools/ms_defender/parser.py
+++ b/dojo/tools/ms_defender/parser.py
@@ -34,9 +34,8 @@ class MSDefenderParser:
                 input_zip = zipfile.ZipFile(file.name, "r")
             else:
                 input_zip = zipfile.ZipFile(file, "r")
+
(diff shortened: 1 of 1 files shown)


valentijnscholten:
@manuel-sommer Could you review if with this PR the code still achieves the goal of your earlier PR?

Maffooch (approved):

hblankenship (approved):

dogboat (approved):

manuel-sommer (approved):
thank you for the ping, works as expected.

mtesauro (approved):
Approved

PR #12463, opened 2025-05-16
unique_id_from_tool: clarify values and usage

valentijnscholten:
This PR clarifies this:

When using the `unique_id_from_tool` or `vuln_id_from_tool` fields in the hash code configuration, it's important that these are uqniue for the finding and constant over time across subsequent scans. If this is not the case, the values can still be useful to set on the finding model without using them for deduplication.

This is based on #12346 and #12446


Maffooch (approved):

mtesauro (approved):
Approved

dogboat (approved):

hblankenship (approved):

Issue #12411, opened 2025-05-08
Nmap report parsing enhancement


egorbeliy:
**Is your feature request related to a problem? Please describe**

i do nmap scan for google and amazon, then i'm trying to import nmap results in defectdojo.
`nmap -Pn -T4 -iL targets.txt -oX "$NMAP_OUTPUT"`
[targets.txt](https://github.com/user-attachments/files/20100767/targets.txt)
[nmap-results.xml.txt](https://github.com/user-attachments/files/20100732/nmap-results.xml.txt)

```
curl -s -X POST "$DEFECTDOJO_URL/api/v2/import-scan/" \
  -H "Authorization: Token $DEFECTDOJO_TOKEN" \
  -F "scan_date=$(date +%F)" \
  -F "minimum_severity=Info" \
  -F "scan_type=Nmap Scan" \
  -F "file=@$NMAP_OUTPUT" \
  -F "engagement_name=Passive scan" \
  -F "product_name=Passive scan" \
  -F "verified=false" \
  -F "active=true" \
  -F "auto_create_context=true" \
  -F "auto_create=true" \
  -F "product_type_name=Research and Development" \
  -F "close_old_findings=false" \
  -F "push_to_jira=false"
```
i check description and couldn't find all hostnames from the nmap xml report file:
```
<hostname name="amazon.com" type="user"/>
<hostname name="google.com" type="user"/>
```
I see only PTR for google 
<hostname name="par10s10-in-f238.1e100.net" type="PTR"/>
and ip address for amazon
<address addr="54.239.28.85" addrtype="ipv4"/>

<img width="511" alt="Image" src="https://github.com/user-attachments/assets/bc145f06-ce54-4fe3-8693-0eeb288a013e" />

**Describe the solution you'd like**
I guess it would be great to have all "arrays" of hostnames in Description.
It can help you to know what host you scanned if they have FQNDs.
Nmap scan results has been attached as well.

valentijnscholten:
Good suggestion, are you able to do a PR?

PR #12466, opened 2025-05-16
:bug: Nmap parser: Add url info to description #12411

manuel-sommer:
#12411


dojo/tools/nmap/parser.py
```diff
@@ -46,8 +46,15 @@ def get_findings(self, file, test):
             if fqdn is not None:
                 host_info += f"**FQDN:** {fqdn}\n"
 
-            host_info += "\n\n"
+            url = (
+                host.find("hostnames/hostname[@type='user']").attrib["name"]
+                if host.find("hostnames/hostname[@type='user']") is not None
+                else None
+            )
```
valentijnscholten, line 53: The existing code has hostnames of type PTR. This PR adds type USER. But what if there are more types now or in the future? Shouldn't the parser just add all hostnames regardless of type?

Maffooch: This is a good idea if it is feasible

valentijnscholten:
Thank you. It might be nitpicking, but should we just list all of the urls and use the type to classify them?
For example just show **PTR**: <value> instead of fqdn? The code seems to assume the PTR record is the FQDN, but it might not be. If the user scanned "google.com" the fqdn in Defect Dojo is shown as ar10s10-in-f238.1e100.net.
This might confuse users. They would need to know the implementation details of the parser to know where the value comes from. It might be easier and more clear if the description for that google.com scan looked like this:

```
**IP Address:** 172.217.18.238
**PTR:** par10s10-in-f238.1e100.net
**User:** google.com
**custom:** lalala
```

valentijnscholten (commented):
as per my comment

Maffooch (commented):
as per Vals comment (Adding this to make the review reminder bot happy :-))

manuel-sommer:
Done, sorry, was quite busy

valentijnscholten (approved):
no worries, happy with all your contributions. approved :-D

mtesauro (approved):
Approved

Maffooch (approved):

Issue #12468, opened 2025-05-16
Account for missing CWE values in the HCL AppScan on Cloud SAST parser


xpert98:
HCL AppScan on Cloud SAST results may include issues that do not include a CWE child element.  The existing parser breaks completely when the CWE element is missing and no results from the XML file are imported.

Adding **cwe = 0** to django-DefectDojo/dojo/tools/hcl_asoc_sast/parser.py will fix the issue, such as in the following snippet:

```python
if report is not None:
    for finding in report:
        title = ""
        description = ""
        cwe = 0 #this is the fix
        for item in finding:
            match item.tag:
```

I can submit a PR for this unless it is more efficient for you to add that line.

manuel-sommer:
Could you close this please @xpert98

PR #12469, opened 2025-05-16
:bug: fix missing CWE in HCL Appscan #12468

manuel-sommer:
#12468

### DIFF
--- a/dojo/tools/hcl_appscan/parser.py
+++ b/dojo/tools/hcl_appscan/parser.py
@@ -38,6 +38,7 @@ class HCLAppScanParser:
             for finding in report:
                 title = ""
                 description = ""
+                cwe = 0
                 for item in finding:
                     match item.tag:
                         case "severity":


valentijnscholten:
@Maffooch I remember you mentioning somewhere default for `cwe` should be `0`. Or was it `None`? We should put that into the parser writing guide :-)

Maffooch:
The [default on the model is `0` when the `cwe` is not supplied](https://github.com/DefectDojo/django-DefectDojo/blob/7a7c7b70d70f922303b7d97d62590f685ddb3b8d/dojo/models.py#L2321-L2323). For parsers specifically, so long as there is consistency within the parser of what is supplied, I think it's okay to be flexible. We just don't want a shakeup between existing data and new data If it can be avoided 😄

hblankenship (approved):

valentijnscholten (approved):

mtesauro (approved):
Approved

Issue #12447, opened 2025-05-14
Include CVSS score when using OpenVAS csv parser


jostaub:
**Is your feature request related to a problem? Please describe**
The OpenVAS CSV parser does not include the CVSS score, even when it is part of the CSV file. However, I would like the CVSS score to always be included in Jira tickets. For this to happen, the score would need to be included in the finding.

**Describe the solution you'd like**
Include the CVSS score in the DefectDojo finding, as the Trivy parser does.

**Describe alternatives you've considered**
It could be confusing that OpenVAS scores may not necessarily be Version 3.1, yet the UI will show the score as 3.1 even if it is not. This problem could be solved if #12445 or a similar issue is accepted. This improvement could be made optional by specifying a env. variable.

**Additional context**
I would upstream the improvement if accepted since i wrote a patch for the parser for my own instance.

valentijnscholten:
Released in 2.47.0.

PR #12472, opened 2025-05-17
Include CVSS score in finding when using OpenVAS csv parser

jostaub:
**Description**

Adresses #12447

**Test results**

Ran `python manage.py test unittests.tools.test_openvas_parser --keepdb` tests which all passed.

### DIFF
--- a/dojo/tools/openvas/csv_parser.py
+++ b/dojo/tools/openvas/csv_parser.py
@@ -265,9 +265,9 @@ class OpenVASCSVParser:
         impact_column_strategy.successor = references_column_strategy
         mitigation_column_strategy.successor = impact_column_strategy
         description_column_strategy.successor = mitigation_column_strategy
-        severity_column_strategy.successor = description_column_strategy
-        cvss_score_column_strategy.successor = severity_column_strategy
-        ip_column_strategy.successor = cvss_score_column_strategy
+        cvss_score_column_strategy.successor = description_column_strategy
+        severity_column_strategy.successor = cvss_score_column_strategy
+        ip_column_strategy.successor = severity_column_strategy
         hostname_column_strategy.successor = ip_column_strategy
         cwe_column_strategy.successor = hostname_column_strategy
         title_column_strategy.successor = cwe_column_strategy
--- a/dojo/tools/openvas/csv_parser.py
+++ b/dojo/tools/openvas/csv_parser.py
@@ -151,6 +151,18 @@ class SeverityColumnMappingStrategy(ColumnMappingStrategy):
             finding.severity = "Info"
 
 
+class CvssColumnMappingStrategy(ColumnMappingStrategy):
+    def __init__(self):
+        self.mapped_column = "cvss"
+        super().__init__()
+
+    def map_column_value(self, finding, column_value):
+        # skip empty values
+        if not column_value:
+            return
+        finding.cvssv3_score = float(column_value)
+
+
 class DescriptionColumnMappingStrategy(ColumnMappingStrategy):
     def __init__(self):
         self.mapped_column = "summary"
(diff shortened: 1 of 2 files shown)


dojo/tools/openvas/csv_parser.py
```diff
@@ -231,6 +243,7 @@ def create_chain(self):
         ip_column_strategy = IpColumnMappingStrategy()
         hostname_column_strategy = HostnameColumnMappingStrategy()
         severity_column_strategy = SeverityColumnMappingStrategy()
+        cvss_column_strategy = CvssColumnMappingStrategy()
```
valentijnscholten, line 246:
```suggestion
        cvssv_score_column_strategy = CvssColumnMappingStrategy()
```

dojo/tools/openvas/csv_parser.py
```diff
@@ -253,7 +266,8 @@ def create_chain(self):
         mitigation_column_strategy.successor = impact_column_strategy
         description_column_strategy.successor = mitigation_column_strategy
         severity_column_strategy.successor = description_column_strategy
-        ip_column_strategy.successor = severity_column_strategy
+        cvss_column_strategy.successor = severity_column_strategy
```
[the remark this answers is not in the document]
jostaub: I added a new commit to address this.

dojo/tools/openvas/csv_parser.py
jostaub: I added a new commit to address this.

dojo/tools/openvas/csv_parser.py
```diff
@@ -253,7 +266,8 @@ def create_chain(self):
         mitigation_column_strategy.successor = impact_column_strategy
         description_column_strategy.successor = mitigation_column_strategy
         severity_column_strategy.successor = description_column_strategy
-        ip_column_strategy.successor = severity_column_strategy
+        cvss_score_column_strategy.successor = severity_column_strategy
+        ip_column_strategy.successor = cvss_score_column_strategy
```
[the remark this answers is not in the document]
jostaub: I'm not entirely sure I understand the intention behind this change. As it stands, it seems like this would effectively disable the cvss_score_column_strategy. Is the goal to make the cvss_score_column_strategy a successor of the severity_column_strategy?

dojo/tools/openvas/csv_parser.py
```diff
@@ -252,7 +265,8 @@ def create_chain(self):
         impact_column_strategy.successor = references_column_strategy
         mitigation_column_strategy.successor = impact_column_strategy
         description_column_strategy.successor = mitigation_column_strategy
-        severity_column_strategy.successor = description_column_strategy
+        cvss_score_column_strategy.successor = description_column_strategy
+        severity_column_strategy.successor = cvss_score_column_strategy
```
Maffooch, line 269: I am confused on this one. Is this the correct mapping? This parser is written in a way very different than any other 😅

jostaub:
@Maffooch / @valentijnscholten 
Here’s how I understand it:

The successor field is used to build a linked list—let’s call it the strategy list—which contains the parsing strategies for all columns. This list is created once by the main loop and then reused for each row in the CSV.

For each row, the code iterates over every column and tries to find a suitable parsing strategy for the current colum by traversing the strategy list via the successor field. This traversal continues until it finds a strategy with a matching column name or until no further successor exists. Once a match is found, the corresponding strategy parses the field, and the result is stored in the current finding. The loop then proceeds to the next column.

Essentially, the order of strategies and their successors only determines the sequence in which strategies are evaluated against the current column name—it doesn’t affect the result, provided a matching strategy is present.

So for example if a strategy's successor points to itself, it would create an infinite loop during traversal.
Where the parsed value is saved in the finding is defined in the Column Strategy Object.

dojo/tools/openvas/csv_parser.py
jostaub: (i added a larger comment to address this to the MR, please look there)

valentijnscholten:
I am more trying to understand why it works like this, what problem it solves and if it's needed.

Maffooch (approved):
This is a goofy parser, but I don't think that fact should get in the way of this PR to add supper for the cvss score field. Tests look good, so I think we should move this forward. We can think about a refactor another day

valentijnscholten (approved):

manuel-sommer:
> @manuel-sommer Can you tell a bit about the successor field? A quick glance at the code makes me think every column gets processed already even without the successor construct.

I have no clue, I just restuctured the code an splitted Openvas into csv and xml here: https://github.com/DefectDojo/django-DefectDojo/pull/9658

valentijnscholten:
Ok, I misread the commit history.

mtesauro (approved):
Approved

dogboat (approved):

Issue #12454, opened 2025-05-15
Finding_group not affected by minimum severity on Jira


9alexx3:
**Bug description**
As discussed via [Slack](https://owasp.slack.com/archives/C2P5BA8MN/p1747229904147839):
When you set a minimum severity for Jira and perform a scan with the arguments `group_by=finding_title` and `create_finding_groups_for_all_findings=true`. The current behavior is to create issues for each group regardless of their severity previously configured in `System Settings`.

**Steps to reproduce**.
1. Have a minimum value of Jira issue.
![Image](https://github.com/user-attachments/assets/0b2a1037-065c-4aa4-bf46-ceeb9dd30530)

2. Enable Finding Groups.
![Image](https://github.com/user-attachments/assets/7a3a1fb4-79be-47ae-8a2a-711bb9a1e4fb)

3. Upload a scan with `group_by=finding_title` and `create_finding_groups_for_all_findings=true` arguments.
4. You will get all the groups made with a Jira incident.

![Image](https://github.com/user-attachments/assets/a68f376f-6229-4920-b3e2-30083b8b80cf)

4.1 If you enter a finding you will see that it is associated but you will see the symbol that its severity is not enough to settle in Jira.

![Image](https://github.com/user-attachments/assets/ee566874-7bba-41ed-be30-103b72bb24a3)

![Image](https://github.com/user-attachments/assets/bbd56b07-dc9a-413e-97c1-95832072b539)

**Expected behavior**
It should create a Jira issue for each group of vulnerabilities where the severity is the one set or higher in `System Settings`.

**Deployment method** *(select with an `X`)*
- [X] Docker Compose

**Environment information**
 - Operating System: Rocky Linux 9
 - Docker Compose: v2.32.4
 - DefectDojo version: v. 2.46.1

valentijnscholten:
Thanks for reporting. We can add a check and make sure finding groups are not pushed if all findings in the group are below the threshold.

Wat would your expectation be if some of the findings in the group are below the threshold, but some are above it? Should Defect Dojo push the group with all findings or only those who are above the threshold?

9alexx3:
I believe it would be more appropriate to push only the elements above the threshold, as this corresponds to the expected behavior when no grouping is applied.

valentijnscholten:
Fixed in 2.48.0

PR #12473, opened 2025-05-18
jira helper: clarify some code

valentijnscholten:
While working on #12454 I noticed some code in the `jira_helper` needed some updates to make it more clear and in line with what it's doing after several refactorings in the past.


valentijnscholten:
looks like for finding that status is actually a list but for a finding group it's a string...

PR #12475, opened 2025-05-18
Finding Groups: Respect minimum severity and active/verified rules when pushing to JIRA

### LINKED ISSUE #12454: Finding_group not affected by minimum severity on Jira

valentijnscholten:
Fixes #12454 and [sc-11185]

In Defect Dojo single JIRA issues are only pushed of they are "qualified" to be pushed to JIRA:

- They must be `Active`
- They must be `Verified` (if Verified status is enforced globally or for the JIRA integration)
- They must be above the `jira_minimum threshold`

This PR now enforces these same rules for Finding Groups.

At first I tried to model this the same as for findings where the `status()` method inside the model calculates the status. But I didn't want to clutter the `models.py` with JIRA related logic, so I put everything in `dojo/jira_link/helper.py` as much as possible

Changes to the Finding Group JIRA issue:
- Description is updated. At first I filtered out any findings not matching the active/verified/severity criteria. But this may get confusing if findings get resolved in Defect Dojo they just disappear from JIRA until the JIRA ticket is empty. Same for changes in Severity or Verified fields in findings. The JIRA Group Issue now shows two tables. The first contains findings matching the criteria ("qualified findings"), the second findings that do not match the criteria ("non-qualified findings"). This way JIRA users will always have an overview of the full group. 

- Status is based on qualified findings
- Priority is based on qualified findings
- SLA Deadline and due date is based on qualified findings.

Screenshot 2025-05-20 213332

Unit tests have been updated, but I can imagine this PR would be a good one to test manually in the UI.


valentijnscholten:
Per request I've also added syncing of the JIRA Priority field for finding groups based on the current list of qualified findings in the group.

Maffooch (commented):
For folks that do not update their jira templates, will anything break here?

valentijnscholten:
> For folks that do not update their jira templates, will anything break here?

Depends on what they've put in their templates :-)

But in general this PR doesn't remove any methods on the finding_group model or jira helper, so I don't expect (m)any things breaking. There is a note in the upgrade notes to make people aware.

Maffooch (changes requested):
Finally made some time to test this in depth. Everything looks great!

The only thing to add would be to apply the same changes to the limited groups template as well. It should only equate to an extra hundred characters or so

valentijnscholten:
Good point, updated the limited template: 

image

valentijnscholten:
I will resolve the conflicts after 2.47.3 is release and merged back into dev. There might be more conflicts by then and I don't want to burden the release process with merge conflicts by merging this PR to dev before 2.47.3.

Maffooch:
That is _exactly_ how I envisioned that! Will approve on Monday after the release

Maffooch (approved):

mtesauro (approved):
Approved

dogboat (approved):

hblankenship (approved):

PR #12476, opened 2025-05-19
new parser ReversingLabs Spectra Assure

rl-maartenb:
**Description**
New parser for ReversingLabs Spectra Assure 

**Test results**
unit tests exist

- data in unittests/scans/reversinglabs_spectraassure/*
- test in: unittests/tools/test_reversinglabs_spectraassure_parser.py

**Documentation**
Documentation added in 

- docs/content/en/connecting_your_tools/parsers/file/reversinglabs_spectraassure.md


valentijnscholten:
Thank you for the PR @rl-maartenb !

Could you take a look at the commit messages? They seem to have "Your Name" as author, which looks like there's something off?

Could you also take a look at the additional info on the `unique_id_from_tool`  field in https://github.com/DefectDojo/django-DefectDojo/pull/12463? I wonder if the `component: ` and `dependency: ` prefix are really necessary?

rl-maartenb:
Regarding: `Could you take a look at the commit messages? They seem to have "Your Name" as author, which looks like there's something off?`
yes that is an error on my side the proper .gitconfig for this tree was not used, also as i struggled to get the pr  to pass the ruff linter i added some messy commits, all with the wrong email and identity, i will revoke the pr and mak a fresh one with the proper email, next week as i am in the Netherlands again on a short holiday starting tomorrow. 

Regarding: `Could you also take a look at the additional info on the unique_id_from_tool field in https://github.com/DefectDojo/django-DefectDojo/pull/12463? I wonder if the component:  and dependency:  prefix are really necessary?`
Your feedbak is very helpfull, i will rework the unique id so that it actually become repeatable (the same) on every scan, currently it uses out internal uuid but that is different for each scan.

rl-maartenb:
new pr is at: https://github.com/DefectDojo/django-DefectDojo/pull/12579

Issue #12378, opened 2025-05-05
OpenVas re-import does not recognize same result


LinoSino:
**Bug Description**

When importing or re-importing the same OpenVAS scan (or scans containing the same vulnerability), some findings are not recognized as already existing in the same test.

When I import the attached openvas.csv file for the first time, it correctly creates a finding. However, when I re-import the same file, it creates a duplicate finding instead of recognizing that it already exists.

The re-import was tested both manually via the GUI and through an API call.

**Steps to reproduce**
1. Create a new test and import the attached CSV file.
2. Re-import the same CSV file into the same test.
3. A second, duplicate finding is created.

**Expected behaviour**
The re-import should recognize existing findings based on the generated hash. This works for most findings, but not for all. An example of a finding that fails to be recognized is included in the attached CSV file.

**Deployment method** *(select with an `X`)*
- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: [Ubuntu 24.04.2]
 - Docker Compose version v2.35.1
 - DefectDojo Community  v. 2.45.3

**Sample scan files**
[openvas.csv](https://github.com/user-attachments/files/20032908/openvas.csv)

**Screenshots**
Uploaded scan twice
<img width="803" alt="Image" src="https://github.com/user-attachments/assets/83268a62-1289-41b9-aa76-f97940abf33e" />

**Additional context**
These are the hash_codes it generated:
second import:
Hash_code changed from "None" to "b1dccf748cfd23192d684069f99a237f0ed9201cfda912c29c9ddb89d5678eab"
first import:
Hash_code changed from "None" to "b1dccf748cfd23192d684069f99a237f0ed9201cfda912c29c9ddb89d5678eab"

valentijnscholten:
Could this be the scenario from #3958 ?

LinoSino:
I'm not entirely sure, as in my example I only have a single finding that I reimport so there sould not be an issue with just checking the first finding. In the file I uploaded, there's just one example finding where this occurs and there exist more but not every OpenVas finding behaves this way just some. Even with deduplication enabled (not just the reimport check), it's not detected as a duplicate.

As a further test, I modified the OpenVas CSV parser to use only the endpoint and Vulnerability-Name for checking, just to rule out anything strange with the description, but that didn’t work either.

jostaub:
The problems seems related to: #1163 and #2687

valentijnscholten:
[...]

PR #12487, opened 2025-05-20
legacy reimport: make matching on title case-insensitive

valentijnscholten:
In some scenario's the reimporter will fall back to the legacy dedupe algorithm that looks at the `title` and `severity` of existing findings. Although this is not a preferred situation, it happens when no dedupe hash field configuration is present for the parser.

At the time of writing this is the case for the OpenVAS parser. This can lead to false negative matching behavior as reported in #12378. This is because Defect Dojo uses `titlecase` to format the title of a finding before saving. So existing findings will can have a different `title` from the findings being processed by the reimporter because the latter are not saved yet.  Since Postgres is performing case sensitive comparisons by default, we have to explicitly make the comparison case insensitive with this PR.

There are no tests cases covering this I believe. Adding a test case is doable, for example using an OpenVAS report. However we  will probably add hash code config soon for OpenVAS rendering the test case useless. I think we should accept that we don't have explicit test cases for this "corner" or "legacy" scenario. I do think it's beneficial to merge this fix as it might be something first time users / PoC users run into for certain less used scan formats.

### DIFF
--- a/dojo/importers/default_reimporter.py
+++ b/dojo/importers/default_reimporter.py
@@ -404,7 +404,7 @@ class DefaultReImporter(BaseImporter, DefaultReImporterOptions):
             # If you have use cases going through this section, you're advised to create a deduplication configuration for your parser
             logger.warning("Legacy reimport. In case of issue, you're advised to create a deduplication configuration in order not to go through this section")
             return Finding.objects.filter(
-                    title=unsaved_finding.title,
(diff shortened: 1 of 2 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

hblankenship (approved):

PR #12501, opened 2025-05-23
:hammer: Merge the MobSF scanner

manuel-sommer:
It is more userfriendly to have only one MobSF scan type to choose. Before parsing MobSF files, you have to analyse which parser is the right one.


valentijnscholten:
I notice 3 MobSF entries in `settings.dist.py`

dojo/settings/settings.dist.py
```diff
@@ -1326,7 +1325,7 @@ def saml2_attrib_map_format(din):
     "HCLAppScan XML": ["title", "description"],
     "HCL AppScan on Cloud SAST XML": ["title", "file_path", "line", "severity"],
     "KICS Scan": ["file_path", "line", "severity", "description", "title"],
-    "MobSF Scan": ["title", "description", "severity"],
+    "MobSF Scan": ["title", "severity", "cwe", "file_path", "description"],
```
valentijnscholten, line 1328: I think this change means a note should be added to the upgrade notes to recalculate the hash codes?

manuel-sommer: I will add this once it is clear when this will be released.

manuel-sommer: @Maffooch fyi: When do you plan to release this? Then, I can add the info.

valentijnscholten: 2.47.0 looks quite busy, so I'll connect it to 2.47.1

manuel-sommer: Target branch is dev, because of the database migration. So lets say 2.48.0, right?

dojo/settings/settings.dist.py
Maffooch, line 1: Dedupe fields appear to be pretty different. Are the report formats different enough that they may benefit from having separate sets of deduplication hash code fields?

manuel-sommer: The file_path makes sense in my point of view as Mobsfscan Scan can report multiple "files" which are differentiated in the file_path, but we can remove cwe as file_path overrules the cwe here.

manuel-sommer:
> I looked at the mobsfscan github repo and website. It looks like there are two ways to generate reports. One via the CLI, the other via the API (of the SAAS platform?). And we used to have two parsers for them and this PR merges them into 1 parser?
> 
> If this is correct, could you add a little of this info to the docs and/or description of the PR?
> 
> Could you also add instructions to the upgrade notes about the removal of the `Mobsfscan Scan` scan type and the new value to be used (`MobSF Scan`)

The thing is that I checked out MobSF and I wanted to import it to Defectdojo, but was confused about two parsers being listed. Thus, I am not really familiar with MobSF, but that PR helps beginners.
I will add upgrade instructions.

valentijnscholten:
Hi @manuel-sommer, we have some concerns around the disappearance of the `Mobsfscan Scan` `scan_type` which might break integrations. What do you think about not migrating the existing tests, but adding the old `scan_type` to the new parser similar to #12604

manuel-sommer:
> Hi @manuel-sommer, we have some concerns around the disappearance of the `Mobsfscan Scan` `scan_type` which might break integrations. What do you think about not migrating the existing tests, but adding the old `scan_type` to the new parser similar to #12604

Done @valentijnscholten

valentijnscholten (changes requested):
Thanks for sticking with us. On final request, could you add something to the upgrade notes after all and to the parser docs? Something like _"MobSF Scan" has been merged into the "MobSF" parser. The "MobSf Scan" scan_type has been retained to keep deduplication working for existing Tests, but users are encouraged to move to the "MobSF" scan_type._

valentijnscholten (approved):

manuel-sommer:
Do you think we can merge this for release 2.50.0?

manuel-sommer:
Can we get this on the road please @mtesauro ?

manuel-sommer:
@valentijnscholten Can we merge this even though the unittests are failing as the docs unittests have nothing to do with this PR?

valentijnscholten (changes requested):
Can you move the upgrade notes to the correct file? Can you try to rebase to see if that fixes the docs build error. The error seems persistent.

mtesauro:
@manuel-sommer 
> Can you move the upgrade notes to the correct file? Can you try to rebase to see if that fixes the docs build error. The error seems persistent.

I agree with moving the upgrade notes as Val mentioned.

We're also looking into that doc error - seems like one of the dependencies that builds the doc hasn't upgraded to the version of hugo we're using. We're trying to verify that currently.

manuel-sommer:
Done @valentijnscholten . Let's see if that fixes the broken docs.

manuel-sommer:
ok, problem persists. Can we merge it anyway?

valentijnscholten (approved):

PR #12503, opened 2025-05-24
Bugfix: fix gap between component header and filter body

jostaub:
**Description**
If you show the filter options, there is a gap between the component panel header and the filter content. This gap is not present in similar views, such as Open Findings. This issue occurs because the `is-filters` `<div>` is not a child of the `panel panel-default` `<div>` like it is in the other views.

image

This PR moves the filters on both HTML component pages into the correct `<div>`.

### DIFF
--- a/dojo/templates/dojo/components.html
+++ b/dojo/templates/dojo/components.html
@@ -16,9 +16,9 @@
                     </div>
                 </h3>
             </div>
-        </div>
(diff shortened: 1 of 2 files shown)


valentijnscholten (approved):

mtesauro (approved):
Approved

dogboat (approved):

PR #12504, opened 2025-05-24
Fix helm chart for nightly-dev builds

valentijnscholten:
The nightly-dev builds currently end up incorrectly in the helm chart due to the `-dev` suffix not being detected.
The change in this PR works OK for version numbers with and without this suffix.

To make the heml chart downloads work, the `nightly-dev` release must be a published release. Otherwise github changes the artifact download urls to something like `https://github.com/DefectDojo/django-DefectDojo/releases/download/untagged-8327861234712346718234671236748129392146712394123412367841678234/<helm-chart.tgz>`

Added to `2.47.1` as I won't be available during `2.47.0`.

### DIFF
--- a/.github/workflows/release-x-manual-helm-chart.yml
+++ b/.github/workflows/release-x-manual-helm-chart.yml
@@ -22,6 +22,21 @@ on:
         description: 'Release number'
         required: true
 
+      make_draft:
+        type: boolean
+        description: 'Mark as draft release?'
+        default: true
+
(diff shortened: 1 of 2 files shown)


mtesauro (approved):
Approved

Maffooch (approved):

dogboat (approved):

blakeaowens (approved):

PR #12509, opened 2025-05-26
remove google sheets leftovers

valentijnscholten:
Remove some leftovers from the Google Sheets feature that was removed in `2.21.0`


dojo/management/commands/initialize_permissions.py
Maffooch, line 1: Seems like this whole file could be deleted now

Maffooch (approved):
Small nit -  not blocking

dojo/management/commands/initialize_permissions.py
valentijnscholten: Yeah I wasn't sure if I should remove it or leave it in place for if we need it in the future

dogboat (approved):

blakeaowens (approved):

PR #12524, opened 2025-05-28
fix(django): Upgrade to 5.2

kiblik:
Upgrade to the latest TLS (support of 5.1 ended December 3, 2025).


valentijnscholten:
@kiblik The fix for `django-tagoulus` may not arrive in time. And we need a fix based on 2.1.0 as 2.1.1 has a bug showing tags twice. I went ahead and create a fix in my fork and added it to the PR here.

valentijnscholten:
Published this as we want to get it reviewed and merged before the xmas holiday starts.

requirements.txt
```diff
@@ -50,7 +50,8 @@ django-crum==0.7.9
 JSON-log-formatter==1.1.1
 django-split-settings==1.3.2
 # do not upgrade to 2.1.1 - https://github.com/DefectDojo/django-DefectDojo/issues/12918
-django-tagulous==2.1.0
+# use fork with django 5.2 fixes, but based on 2.1.0
```
kiblik, line 53: @valentijnscholten, is there any chance of having it based on `2.1.1`?

kiblik: But the rest LGTM

kiblik: Sorry, I just noticed a comment in one line above.

kiblik:
I'm not able to add an approval (because PR/commits is mine), but it looks good to me.

mtesauro (approved):
Approved

PR #12528, opened 2025-05-29
Dojo Meta: Migrate to `filterset_class` + Add case Insensitive filters

Maffooch:
Adds filters to the `metadata` endpoint that can filter on `name` and `value` in a case insensitive way

[sc-11313]


blakeaowens (approved):

mtesauro (approved):
Approved

PR #12579, opened 2025-06-10
ReversingLabs SpectraAssure rl-json parser for DefectDojo

rl-maartenb:
Feature: ReversingLabs SpectraAssure rl-json parser for DefectDojo

Documentation and unittests provded according to the plugin documentation

This adds a new parser for the rl-json format produced by the SpectraAssure cli or portal.


rl-maartenb:
after bot complain added 'title: ...' to the doc file

rl-maartenb:
fixed the readme after the checker complained

rl-maartenb:
my local hugo extended test now shows no errors

dojo/tools/reversinglabs_spectraassure/parser.py
```diff
@@ -0,0 +1,117 @@
...
+import logging
+from typing import Any
+
+from dojo.models import Finding
+from dojo.tools.reversinglabs_spectraassure.rlJsonInfo import RlJsonInfo
+from dojo.tools.reversinglabs_spectraassure.rlJsonInfo.cve_info_node import CveInfoNode
+
+logger = logging.getLogger(__name__)
+
+WHAT = "ReversingLabs Spectra Assure"
```
valentijnscholten: Can you give this a clearer name?

dojo/tools/reversinglabs_spectraassure/rlJsonInfo/__init__.py
```diff
@@ -0,0 +1,594 @@
...
+
+        cin.scan_date = datetime.datetime.fromisoformat(self._rest["timestamp"]).date()
+        cin.scan_tool = self.SCAN_TOOL_NAME
+        cin.scan_tool_version = self._rest.get("version", "no_scan_tool_version_specified")
+        cin.cvss_version = int(this_cve.get("cvss", {}).get("version", "0"))
+        score = float(this_cve.get("cvss", {}).get("baseScore", "0.0"))
+        cin.score = score
+        cin.score_severity = self._score_to_severity(score=score)
+
+        # TODO: tags
```
valentijnscholten: Is this still a TODO?

dojo/tools/reversinglabs_spectraassure/rlJsonInfo/__init__.py
```diff
@@ -0,0 +1,594 @@
...
+        dep_purl = dependency.get("purl", "")
+        dep_name = dependency.get("product", "")
+        dep_version = dependency.get("version", "")
+
+        # if we have a dependency purl then purl, otherwise component product + version
+        tail = dep_purl
+        if len(tail) == 0:
+            tail = f"{dep_name}@{dep_version}"
+
+        cin.unique_id_from_tool = f"{cin.component_file_sha256}:{cve}:{tail}"
```
valentijnscholten: the `unique_id_from_tool` value should be something that is in the report, not a computed/constructed value. Can you look at this?

dojo/tools/reversinglabs_spectraassure/rlJsonInfo/__init__.py
rl-maartenb: sorry my mistake, tags have been added, i will remove the todo line

dojo/tools/reversinglabs_spectraassure/parser.py
rl-maartenb: I expect you mean on the left side of the '=', yes will fix that

dojo/tools/reversinglabs_spectraassure/rlJsonInfo/__init__.py
rl-maartenb: This one is more tricky.

In order for unique-id from tool to be valuable for detecting duplicates across future uploads we have to distinguish between what we call components (unpacked-files), vulnerabilities (cve) and component-dependencies in our report.

We detect vulnerabilities (cve) directly on components (unpacked-files) and we detect vulnerabilities (cve) on dependencies of components.
Components (unpacked files) are files in our report and so have a unique sha256. Dependencies may occur on multiple files(components) in the scan (as a result of unpacking, zip files, msi installers and such).
Dependencies usually have a package_url, or if they dont they have a version and name or product string.
So to uniquely identify one vulnerability (cve) on one item (component or dependency) we need to combine items.
- for vulnerabilities only on components (without dependencies) we need the sha256 and the cve to uniquely point to the file and the cve.
- for vulnerabilities on dependencies of compnents we need the sha256 of the components the package-url of the dependency and the cve of the vulneramility.
Once we have that we can fully use deduplication on future scans and imports into DefectDojo.

The report file internally uses 'uuid' to condense the report items but they are only unique inside this one report-file so have no meaning for deduplication. As such we do not have `one single uniqe id` in our report that uniquely identifies one DefectDojo-Finding. 

A good example is the provided multi language installer in 'unittests/scans/reversinglabs_spectraassure/HxDSetup_2.5.0.exe-report.rl.json' 
as the installer supports 18 different languages we see 18 embedded files. each file has dependencies and many depdendencies have vulnerabilities we report.

I hope this helps to explain why we have to `construct` a unique id.
i would guess the alternative would be not to privide a unique-id.

dojo/tools/reversinglabs_spectraassure/parser.py
rl-maartenb: done

dojo/tools/reversinglabs_spectraassure/rlJsonInfo/__init__.py
rl-maartenb: done

dojo/tools/reversinglabs_spectraassure/rlJsonInfo/__init__.py
rl-maartenb: needs review

Maffooch: > i would guess the alternative would be not to privide a unique-id.

I suspect this may be the case. The `unique_id_from_tool` is intended to track the same vulnerability fro report to report. Ideally, this is handled by the vendor, but in cases where it is not, DefectDojo will essentially construct something similar you are doing, but not at the parser level. Instead, those fields will be captured in the [deduplication settings through the hash code fields](https://docs.defectdojo.com/en/open_source/archived_docs/usage/features/#deduplication-algorithms). It is essentially the same thing you are doing, just with DefectDojo fields after the findings are created by the importer/reimporter.

Maffooch (changes requested):
The structure of this parser is quite a bit different than other parsers that are found in DefectDojo. This may make it bit more difficult to maintain going forward

The [trivy parser](https://github.com/DefectDojo/django-DefectDojo/blob/b3b30e8791c5e750093d79d6e04ed5008d89c831/dojo/tools/trivy/parser.py#L45) could be a good example to mimic

dojo/tools/reversinglabs_spectraassure/rlJsonInfo/__init__.py
rl-maartenb: uniq_id from tool removed

rl-maartenb:
Will have a look at the trivy parser but i already see that it will not match our data, we use a significant amount or references to other items in our reports. 

I see that most data from other reports produce a relativly flat list of results, however we produce compacted dicts with cross references. 

The actual parsing happens in:
- https://github.com/rl-maartenb/django-DefectDojo/blob/dev/dojo/tools/reversinglabs_spectraassure/rlJsonInfo/__init__.py

I can naturally move that code into parser.py if that would suit you better.

In general our report requires some significant study to see how the network of references work before doing any maintenance on the parser as it is easy to overlook things or make the wrong assuptions.

Maffooch:
> In general our report requires some significant study to see how the network of references work before doing any maintenance on the parser as it is easy to overlook things or make the wrong assuptions.

Wow, you weren't kidding! 2400 lines for a single vulnerability is quite a bit of data.

This confirms my fear that if, for reason, you're not around to maintain this parser if something changes in the report format producee by Reversing Labs, making changes will be a heavy task. I suppose the request for simplification is more  important than I realized. 

Rather than asking you to reengineer the parsing logic, I think we can get away with documenting the parser. How does it sound to heavily comment the code in both files, and then write a short README in the same directory as the `parser.py` file that explains why things are split and which classes are responsible for what?

valentijnscholten:
I'd like to request to make the variable names more clear as well. I had trouble following the code with variables like r, rr, cin.

rl-maartenb:
@valentijnscholten: will do

@Maffooch: good idea, will do

rl-maartenb:
both done,  short names replaced with more meaningful ones, the actual parser in `rlJsonInfo` now explains better what we use currently from the rl-json report and how things are chained inside the report.

rl-maartenb:
not sure if this is related to my changes 
Unit tests, Attempt #2: Some jobs were not successful
https://github.com/DefectDojo/django-DefectDojo/actions/runs/15733336564/job/44360465014

```
Run helm install \
  helm install \
    --timeout 800s \
    --wait \
    --wait-for-jobs \
    defectdojo \
    ./helm/defectdojo \
    --set django.ingress.enabled=true \
    --set imagePullPolicy=Never \
    --set initializer.keepSeconds="-1" \
     --set postgresql.enabled=true --set createPostgresqlSecret=true  \
     --set redis.enabled=true --set celery.broker=redis --set createRedisSecret=true  \
    --set createSecret=true \
    --set tag=debian
  shell: /usr/bin/bash -e {0}
  env:
    DD_HOSTNAME: defectdojo.default.minikube.local
    HELM_REDIS_BROKER_SETTINGS:  --set redis.enabled=true --set celery.broker=redis --set createRedisSecret=true 
    HELM_PG_DATABASE_SETTINGS:  --set postgresql.enabled=true --set createPostgresqlSecret=true 
    MINIKUBE_HOME: /home/runner/work/_temp
    pgsql:  --set postgresql.enabled=true --set createPostgresqlSecret=true 
    redis:  --set redis.enabled=true --set celery.broker=redis --set createRedisSecret=true 
walk.go:7[5](https://github.com/DefectDojo/django-DefectDojo/actions/runs/15733336564/job/44360465014#step:9:5): found symbolic link in path: /home/runner/work/django-DefectDojo/django-DefectDojo/helm/defectdojo/README.md resolves to /home/runner/work/django-DefectDojo/django-DefectDojo/readme-docs/KUBERNETES.md. Contents of linked file included and used
Error: INSTALLATION FAILED: context deadline exceeded
Error: Process completed with exit code 1.
```

mtesauro (approved):
Approved

PR #12584, opened 2025-06-10
feat(docker): Depends_on based on initializer

kiblik:
This change blocks the start of `uwsgi`, `celerybeat`, and `celeryworker` until `initializer` is not `completed_successfully`. This is useful for:
- avoiding errors in `uwsgi`, `celerybeat`, and `celeryworker` during start-up when the db is not ready
- make it easier to identify errors with `initializer` (they might get easily lost within other logs).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,8 @@ services:
       dockerfile: "Dockerfile.nginx-${DEFECT_DOJO_OS:-debian}"
     image: "defectdojo/defectdojo-nginx:${NGINX_VERSION:-latest}"
     depends_on:
-      - uwsgi
+      uwsgi:
+        condition: service_started
     environment:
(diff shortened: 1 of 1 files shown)


kiblik:
> Cool! What happens when there is an error in the `initializer`? I guess it just fails with an error and the other containers are never started?

Exactly (well, except psql and redis - they do not use dependency, so they would still be running)

For testing, I changed `models.py`, did not prepare migrations, `initializer` failed, and the whole process of startup stopped (docker did not touch `uwsgi`, `nginx` or `celery*`). This was the last message before coming back to bash.
```
service "initializer" didn't complete successfully: exit 1
```

Maffooch (approved):

mtesauro:
Excellent work here :rocket:

mtesauro (approved):
Approved

Issue #12399, opened 2025-05-07
Issue on Closed Findings Metrics


X0x1RG9f:
**Bug description**
Closed Findings metrics (https://demo.defectdojo.org/metrics?date=5&view=dashboard) does not correspond to the real closed findings. After a short analysis, it seems that the list of closed findings of 2025 (for example) only calculates findings closed on 2025 AND opened on 2025.

**Steps to reproduce**
1. Open a finding with a date in 2022
2. Close the same finding today
3. Navigate to the metric page, the closed finding won't be appearing as closed this year. 

**Expected behavior**
Closed findings metrics should not depends on opening date, but only on closed date. If finding is closed today, it should be included in the stats even if opened 4 years ago.

PR #12595, opened 2025-06-11
simple metrics: count closed findings not opened in current month

valentijnscholten:
Fixes #12534 and #12399

Simple metrics was only showing findings closed in the selected month when they were also opened in the selected month.

I t could be that was intentional at the time it was implemented, but I believe it makes more sense to count all findings that were closed in that month regardless of when they were opened.


Maffooch (approved):

dogboat (approved):

mtesauro (approved):
Approved

PR #12599, opened 2025-06-13
Simple metrics closed per month query improvement

valentijnscholten:
Small improvement after #12595 to make sure only the right findings are counted / no corner cases with `mitigated==null` happening


mtesauro (approved):
Approved

Maffooch (approved):

PR #10894, opened 2024-09-10
Add support for Invicti parser through Netsparker

Maffooch:
[Netsparker has now become Invicti](https://www.invicti.com/blog/news/netsparker-is-now-invicti-signaling-a-new-era-for-modern-appsec/), so we should support Invicti by extending the Netsparker parser. If there are modifications to the incivicti format going forward, it will be much easier to make a change to an existing parser.

[sc-2526]

PR #12604, opened 2025-06-13
Burp Enterprise renamed to Burp DAST

valentijnscholten:
Burp Entrprise got renamed to Burp Suite DAST.

This PR covers this by adding a new `scan_type` to the existing parser. And renaming some files.

This way we don't have to copy-and-paste and maintain duplicate docs and unit tests, so slightly different from #10894 

The docs page for `burp_suite_dast` also mentions "Burp Enterprise" (the old name).

I adjusted the choices for the `scan_type` a little bit to allow us to add some extra text to indicate the parser was renamed.
Not sure if this is desired as it might make people think they _need_ to migrate over, but that would break deduplication/reimport because of the difference in `scan_type` value.

Some screenshots:

Screenshot 2025-06-13 221601

Screenshot 2025-06-13 222057

Screenshot 2025-06-13 221619

Screenshot 2025-06-13 221958

[sc-11366]


hblankenship (approved):

dogboat (approved):

mtesauro (approved):
Approved

Issue #12575, opened 2025-06-09
Bad performace. Engagement page takes ~50 s when tests have many findings — replace DISTINCT counts with Subquery counters in `prefetch_for_view_tests`


DenysMoskalenko:
### **Bug description**
Opening the “View Engagement” page for an engagement that contains tests with **≈ 20 000 findings and > 1 000 re-imports** takes **~50 s**.  
`prefetch_for_view_tests()` counts findings with `Count(..., distinct=True)` after joining `dojo_finding` and `dojo_test_import`, which explodes to ~25 M rows before grouping.

---

### **Steps to reproduce**
1. Import or create an engagement that has a test with ~20 000 findings (e.g. large SARIF or Gitleaks scan) and many re-imports.  
2. Navigate to **Product → Engagements → View Engagement** (`/engagement/<id>`).  
3. Observe 45–50 s page load; PostgreSQL shows heavy temp-file usage.

---

### **Expected behavior**
The page should render in under a second regardless of findings volume.

---

### **Deployment method**
- [X] Docker Compose
- [x] Kubernetes
- [ ] GoDojo

---

### **Environment information**
* OS / container base : Debian 12 (official images)  
* Docker compose : `v2.24.4`  
* DefectDojo commit : `[2025-06-08] 6e5a3d0` (current `dev`)  
* PostgreSQL : 14

---

### **Proposed fix**

Replace the six `Count(..., distinct=True)` annotations in  
`dojo/views/engagement/views.py::prefetch_for_view_tests` with index-only **Subquery** counters so each is evaluated once per `test_id`.  
Execution time on the same dataset drops from **≈ 50 s → ≈ 200 ms**.

```python
from django.db.models import (
    Count, IntegerField, OuterRef, Subquery, Value, QuerySet
)
from django.db.models.functions import Coalesce
from dojo.models import Finding, Test_Import   # adjust import path if needed

def _count_subquery(qs):
    """Return a Subquery that yields one aggregated count per test_id."""
    return Subquery(
        model_qs.values("test_id").annotate(c=Count("*")).values("c")[:1],  # one row per test_id
        output_field=IntegerField(),
    )

def prefetch_for_view_tests(tests: QuerySet) -> QuerySet:
    if not isinstance(tests, QuerySet):
        logger.warning("unable to prefetch because query was already executed")
        return tests

    qs = tests.select_related("lead", "test_type").prefetch_related("tags", "notes")

[...]

PR #12606, opened 2025-06-14
Optimize `view_engineer` to use DB-side aggregation and cut load time

DenysMoskalenko:
> Initially inspired by [#12575](https://github.com/DefectDojo/django-DefectDojo/issues/12575)  

---

#### ✨  Key Points
* **Full rewrite of `view_engineer`**
  * replaces Python loops with SQL `COUNT` / `CASE` / `Subquery`
  * helper functions: `_age_buckets`, `_product_stats`, `_augment_series_with_accepted`
  * preserves the 5-minute `@cache_page` (But now it **can be removed** and I guess - should be removed)
* **One SQL round-trip per metric**
  * month/week open / closed / accepted
  * product severity tables
  * age buckets for high/critical
* **Targeted ORM tuning**
  * lean `select_related` & `prefetch_related`
  * dead imports & paths removed
* **No template or migration changes**

---

#### 🚀  Performance on our dataset
**I cannot run it on the our production dataset for now, but I excpect huge improvement**

| Env            | Before | After | Δ |
|----------------|--------|-------|---|
| dev | **12.7s** | **32 ms** | –99.7 % |
| prod | **≈ 190s** | - | - |

<details>
<summary>Screenshots</summary>

| Before | After |
|:--:|:--:|
| Screenshot 2025-06-14 at 23 31 00 | Screenshot 2025-06-14 at 23 30 29|

</details>

### DIFF
--- a/dojo/metrics/views.py
+++ b/dojo/metrics/views.py
@@ -3,15 +3,12 @@ import collections
 import logging
 import operator
 from calendar import monthrange
-from collections import OrderedDict
 from datetime import date, datetime, timedelta
-from functools import reduce
-from operator import itemgetter
 
 from dateutil.relativedelta import relativedelta
 from django.contrib import messages
 from django.core.exceptions import PermissionDenied
-from django.db.models import Case, Count, IntegerField, Q, Sum, Value, When
+from django.db.models import Case, Count, F, IntegerField, Q, Sum, Value, When
 from django.http import HttpResponseRedirect
 from django.shortcuts import get_object_or_404, render
 from django.urls import reverse
@@ -35,7 +32,7 @@ from dojo.metrics.utils import (
     identify_view,
     severity_count,
 )
-from dojo.models import Dojo_User, Engagement, Finding, Product, Product_Type, Risk_Acceptance, Test
+from dojo.models import Dojo_User, Finding, Product, Product_Type, Risk_Acceptance
 from dojo.product.queries import get_authorized_products
 from dojo.product_type.queries import get_authorized_product_types
 from dojo.utils import (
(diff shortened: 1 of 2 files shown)


valentijnscholten:
Thanks again for a nice optimization PR. If the queries are now much faster, the caching can probably be removed.

Since these changes are a little more involved. Is there anyway we can verify the results of the queries are still the same/correct. Would it be feasible to add a little unit test similar to https://github.com/DefectDojo/django-DefectDojo/blob/bugfix/unittests/test_metrics_queries.py ?

valentijnscholten:
Something I also wondered if there is a better/clearer name to the `stuff` named variables. To make it more clear what they are.

dojo/metrics/views.py
```diff
@@ -671,307 +668,247 @@ def engineer_metrics(request):
...
+    open_findings = (
+        reporter_findings.filter(mitigated__isnull=True)
+        .select_related("test__engagement__product__prod_type", "reporter")
+        .prefetch_related("risk_acceptance")
+    )
+
+    # --------------------
+    # Month & week buckets
+    month_start = datetime(now.year, now.month, 1, tzinfo=tz)
+    month_end = month_start.replace(day=monthrange(now.year, now.month)[1])
```
dogboat: I think this will miss range checks on datetime fields that occur on the last day. `month_end` terminates at 00:00 of the last day, so datetimes on the last day after that won't be included. The checks for `mitigated__range` and `risk_acceptance__created__range` below are affected by this.

I know this was already present in the existing code, so I'm very sorry to ask you to look at it (🙇), but would it be possible to fix that as part of this? One possible solution is, capture the first day of the next month instead as the "end" date and use `gte`/`lt` filters instead of `range` (i.e., `gte=month_start, lt=end`).

dojo/metrics/views.py
```diff
@@ -671,307 +668,247 @@ def engineer_metrics(request):
...
+            medium=Count("id", filter=Q(severity="Medium")),
+            low=Count("id", filter=Q(severity="Low")),
+            total=Count("id"),
+        )
+    )
+    by_id = {c["pid"]: c for c in counts}
+    top10 = sorted(by_id.items(), key=lambda kv: kv[1]["total"], reverse=True)[:10]
+
+    def row(prod_id):
+        prod = next(p for p in products if p.id == prod_id)
```
dogboat: Nit: doesn't this create new generator and iterate over it each call? Why not create a dict immediately above this (mapping prod.id to prod) and use that to perform lookups?

dojo/metrics/views.py
```diff
@@ -671,307 +668,247 @@ def engineer_metrics(request):
...
+            start = datetime.strptime(bucket[0].strip(), "%b %Y").replace(tzinfo=tz)
+            end = start.replace(day=monthrange(start.year, start.month)[1])
+        else:  # "week"
+            wk_a, wk_b = (d.strip() for d in bucket[0].split("-"))
+            year = timezone.now().year
+            start = datetime.strptime(f"{wk_a} {year}", "%b %d %Y").replace(tzinfo=tz)
+            end = datetime.strptime(f"{wk_b} {year}", "%b %d %Y").replace(tzinfo=tz)
+
+        accepted = (
+            Finding.objects.filter(risk_acceptance__owner=owner, risk_acceptance__created__range=[start, end])
```
dogboat: Same comment as above re: `range`.

dogboat (changes requested):
Nice work!

I also agree with @valentijnscholten about tests and variable names if possible.

valentijnscholten:
@DenysMoskalenko Thank you again for your contributions. I've moved this to the 2.49.0 while we wait for conflicts resolution.

DenysMoskalenko:
> @DenysMoskalenko Thank you again for your contributions. I've moved this to the 2.49.0 while we wait for conflicts resolution.

Thanks, I will fix the conflicts.
Also I really want to add a unittests you asked, just have no time right now(

Maffooch (approved):
@DenysMoskalenko great work here! I'm approving for now, but happy to wait on merging until you're happy with it 😄

valentijnscholten (approved):

DenysMoskalenko:
@valentijnscholten Hi)
I have added some unit tests as you asked, sry if they are not so good, i am not a big Django developer(

PR #12607, opened 2025-06-15
API: prevent duplicate saves of taggable entities or when pushing to JIRA

valentijnscholten:
When creating or updating findings (or other entities that have a `tags` field), the `Finding.save()` method was called multiple times. Ultimately the `finding` was saved correctly, but this lead to unnecessary use of resources and could in the future lead to problems if/when the Defect Dojo could be relying more on Django signals such as `post_save()`. 

The above was caused by the way the `tags` fields were implemented long ago. I tried to tweak the `TaggitSerializer` to prevent the duplicate saves from happening, but this lead nowhere. Then I realized I could just tweak our `TagListSerializer` to make it work without the "extra" `TaggitSerialiser`. 

I also looked at the `from tagulous.contrib.drf import TagSerializer`, but that doesn't support the single string with multiple comma separated tags that we re-added support for in #12434

Duplicate saves could also be triggered when the finding needed to be pushed to JIRA. This PR also resolves that.

Added some unit tests to make sure all looks well.

I had to rerecord one JIRA related tests around finding groups. Usually this is a red flag, but looking at the test case and asserts I think all is still good. Maybe the old codebase did some duplicagte/unnecessary/idempotent calls to JIRA?

[sc-10686]

### DIFF
--- a/dojo/api_v2/exception_handler.py
+++ b/dojo/api_v2/exception_handler.py
@@ -46,7 +46,7 @@ def custom_exception_handler(exc, context):
         # There is no standard error response, so we assume an unexpected
         # exception. It is logged but no details are given to the user,
         # to avoid leaking internal technical information.
-        logger.error(exc)
+        logger.error(exc, exc_info=True)  # noqa: LOG014
         response = Response()
         response.status_code = HTTP_500_INTERNAL_SERVER_ERROR
         response.data = {}
@@ -67,6 +67,6 @@ def custom_exception_handler(exc, context):
     else:
         # HTTP status code 500 or higher are technical errors.
         # They get logged and we don't change the response.
-        logger.error(exc)
+        logger.error(exc, exc_info=True)  # noqa: LOG014
 
     return response
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -208,6 +208,8 @@ class TagListSerializerField(serializers.ListField):
         self.pretty_print = pretty_print
 
     def to_internal_value(self, data):
+        logger.debug(f"to_internal_value called with data: {data}")
+        # logger.debug("Stacktrace:\n%s", "".join(traceback.format_stack()))
         if isinstance(data, list) and data == [""] and self.allow_empty:
             return []
         if isinstance(data, six.string_types):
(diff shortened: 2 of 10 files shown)


dojo/api_v2/exception_handler.py
Maffooch, line 1: I have been manually adding this back in a pinch for so long, but have always forgot to commit it 🥴 thank you for doing so!

dojo/api_v2/serializers.py
```diff
@@ -1932,18 +1886,14 @@ def create(self, validated_data):
...
-        )
-        # If we need to push to JIRA, an extra save call is needed.
-        # TODO: try to combine create and save, but for now I'm just fixing a
-        # bug and don't want to change to much
-        if push_to_jira or new_finding:
-            new_finding.save(push_to_jira=push_to_jira)
-        # This final call will save the finding again and return it
-        return self._save_tags(new_finding, to_be_tagged)
+            # can we avoid this extra save? the cve has already been set above in validated_data. but there are no tests for this
+            # on finding update nothing is done # with vulnerability_ids?
```
Maffooch, line 1893: Seems like the extra save could be removed since th vulnerability IDs have. dedicated function now

dojo/api_v2/serializers.py
Maffooch, line 1: Please clean up those extra comments when you are done with all the manual testing/validation

dojo/api_v2/views.py
```diff
@@ -989,13 +989,13 @@ def tags(self, request, pk=None):
                 all_tags = serializers.TagSerializer({"tags": all_tags}).data[
                     "tags"
                 ]
+                for tag in new_tags.validated_data["tags"]:
+                    for sub_tag in tagulous.utils.parse_tags(tag):
```
Maffooch, line 993: curious why this was needed. Does the `tagulous.utils.parse_tags` function not work as expected when passed a list vs individual strings?

dojo/models.py
```diff
@@ -2678,7 +2678,7 @@ def __str__(self):
     def save(self, dedupe_option=True, rules_option=True, product_grading_option=True,  # noqa: FBT002
              issue_updater_option=True, push_to_jira=False, user=None, *args, **kwargs):  # noqa: FBT002 - this is bit hard to fix nice have this universally fixed
         logger.debug("Start saving finding of id " + str(self.id) + " dedupe_option:" + str(dedupe_option) + " (self.pk is %s)", "None" if self.pk is None else "not None")
-
+        # logger.debug("Stacktrace:\n%s", "".join(traceback.format_stack()))
```
Maffooch:
```suggestion
```

Maffooch (changes requested):
This is looking really promising! Excellent work here

dojo/api_v2/views.py
valentijnscholten: IIRC it parses the string repesentation of the list ending up with something like:
- ["tag1"
- "tag2"
- "tag3"]
But it could be that it was needed because I also changed the `TagListSerializerField` to not render the tags as one string. That changes allows us to just let the ModelSerializer serialize the tags without having to pop them and process them ourselves.

dojo/api_v2/serializers.py
valentijnscholten: that extra functions doesn't actually save anything on the finding itself, I think we have to add a small test to ensure a CVE is stored on the finding itself. Unless that field is slated for deprecation?

valentijnscholten: I added some tests to make sure vulnerability ids are stored and the cve is populated

Maffooch (approved):
Excellent work here! These are my favorite kinds of changes to see here 🤗

hblankenship (approved):

dogboat (approved):

mtesauro (approved):
Approved

Issue #12610, opened 2025-06-16
JIRA: finding_text field not showing in issue description


9alexx3:
**Bug description**
The `finding_text` field does not display in the Jira issue description, even when it is defined in its instance (jira/id/edit).

**Steps to reproduce**

1. Get a valid Jira configuration with a Finding Text.

![Image](https://github.com/user-attachments/assets/6ca5dbe8-1362-4504-84b5-a2e09c9ac69a)

2. Push to Jira.
3. The expected text is not being displayed in the Jira issue description (jira_limited and jira_full tested).

 **Expected behavior**
The Finding_text field should be displayed in the Jira issue description even if it is at the end of the issue.

**Deployment method**
- [X] Docker Compose

**Environment information**
 - Operating System: Rocky Linux 9
 - Docker Compose version:  v2.32.4
 - DefectDojo version: v. 2.47.1

**Additional context**
After some research, I was able to fix the problem and I will create a PR for this.

PR #12612, opened 2025-06-16
Fix Jira FindingText

9alexx3:
**Description**

[Fixes 12610](https://github.com/DefectDojo/django-DefectDojo/issues/12610). Add the `finding_text` attribute from the Jira instance to the issue description field and update the templates to properly display it.

**Test results**

image
If no text is configured, no message is displayed.

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -637,13 +637,13 @@ def jira_description(obj):
         kwargs["finding"] = obj
         jira_instance = get_jira_instance(obj)
         if jira_instance:
-            loggin.debug("Adding findind_text to the jira description")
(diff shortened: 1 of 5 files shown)


dojo/jira_link/helper.py
Maffooch, line 1: It would be preferable to update update the `jira_description` to accept arbitrary `**kwargs` that could then be passed directly to the context of the `render_to_string` function. This would avoid the need of hitting the database again to get the jira instance

Maffooch (changes requested):

dojo/jira_link/helper.py
valentijnscholten: We should careful not to put objects in there with sensitive data like password/api keys.

PR #12624, opened 2025-06-18
Mayhem SARIF support (new parser)

xansec:
**Description**

This supports parsing Mayhem-generated SARIF reports. In general, the existing SARIF support _should_ work, however, there are some idiosyncrasies as Mayhem is a DAST tool, where the output fields are used in a somewhat non-idiomatic way. Rather than creating a patch change for the existing SARIF report support, I propose this new Mayhem specific support, so that we can keep up with the changes to the SARIF output without burdening or imposing on the existing SARIF parser.

**Test results**

Tests have been included in `dojo/unittests` 

**Documentation**

Documentation has been added under `dojo/docs/content/en/connecting_your_tools/parsers/file/mayhem.md`

**Checklist**

- [X] If this is a new feature and not a bug fix, you've included the proper documentation in the docs at https://github.com/DefectDojo/django-DefectDojo/tree/dev/docs as part of this PR.

### DIFF
--- a/unittests/tools/test_mayhem_parser.py
+++ b/unittests/tools/test_mayhem_parser.py
@@ -10,8 +10,8 @@ class TestMayhemParser(DojoTestCase):
         if finding.cwe:
             self.assertIsInstance(finding.cwe, int)
         self.assertFalse(finding.static_finding)  # Mayhem is DAST!
-        self.assertTrue(True, finding.dynamic_finding)  # Mayhem is DAST!
-        self.assertIsInstance(finding.description, str) 
+        self.assertTrue(finding.dynamic_finding)  # Mayhem is DAST!
+        self.assertIsInstance(finding.description, str)
         self.assertEqual(1, finding.reporter_id)
 
     def test_mcode_many_report(self):


xansec:
I believe the issue raised by the security bot are by design - the SARIF included various examples of exploits, and credentials are known demo credentials for public demo endpoints.

valentijnscholten:
Thank you for raising the PR. I just wanted to double check if there's really no way around copying and pasting the existing parser.  It's not a lot of code, but I can imagine if SARIF gets more populat we might get more parsers that need slightly different handling of things. Have you considered instead to inherit/extend the existing parser class and overwrite some methods? When I do a "lazy" compare of the SARIF parser with the Mayhem parsers the difference seem very limited, mainly some markdown handling and title cleanup + cwe handling?

xansec:
Good catch - I went ahead and extended the SarifParser class instead of copying code. There were a couple of issues with how SarifParser uses member functions vs package functions, so I ended up having to copy code anyways (otherwise polymorphism wasn't working and reverting back to the original SarifParser functions), but I tried to keep this to a minimum. Let me know if there's anything else.

valentijnscholten:
Looks like we need to make some changes to the Sarif parser to make it more extensible. That might be too much to try and achieve in this PR. If you could ensure the Mayhem unit tests are testing all the Mayhem specific fields/data, that will help in the future when we make the Sarif parser more extensible.

valentijnscholten:
@xansec What do you think about the unit tests? I believe the cwe handling is different from the default SARIF parser, so it's good to have that present in the unit test assertions. What else is different/extended in the Mayhem parser and would be good to assert on in the unit tests?

xansec:
I have added more comprehensive unit tests for both CWEs as well as a few other fields in a finding. This should cover most cases.

Maffooch (approved):

blakeaowens (approved):

PR #12639, opened 2025-06-19
Import EPSS data from Anchore Grype scans

bwt-sloanj:
**Description**

Grype added EPSS metrics in v0.92.0, see:
https://github.com/anchore/grype/releases/tag/v0.92.0

Adds EPSS support with a test based on the same busybox image used in anchore_grype/no_vuln.json.

Also removes the hardcoded CWE value as Grype does not support CWE data.

**Test results**

Adds `unittests.tools.test_anchore_grype_parser.TestAnchoreGrypeParser.test_grype_epss_values`.

Modifies `unittests.tools.test_anchore_grype_parser.TestAnchoreGrypeParser.test_check_all_fields` to reflect CWE change.

Fixes typo in `unittests.tools.test_anchore_grype_parser.TestAnchoreGrypeParser.test_grype_parser_with_one_critical_vuln_has_one_findings`.

**Documentation**

Updates documentation with information on scanning SBOMs and newer style of getting Grype JSON output written to file.

**Checklist**

- [x] Your code is ruff compliant.

### DIFF
--- a/dojo/tools/anchore_grype/parser.py
+++ b/dojo/tools/anchore_grype/parser.py
@@ -22,8 +22,8 @@ class AnchoreGrypeParser:
 
     def get_description_for_scan_types(self, scan_type):
         return (
-            "A vulnerability scanner for container images and filesystems. JSON report generated with '-o json' "
-            "format"
+            "A vulnerability scanner for container images, filesystems, and SBOMs. "
+            "JSON report generated with '--output=json' format."
         )
 
     def get_findings(self, file, test):
@@ -41,11 +41,13 @@ class AnchoreGrypeParser:
             if "fix" in vulnerability:
                 vuln_fix_versions = vulnerability["fix"].get("versions")
             vuln_cvss = vulnerability.get("cvss")
+            vuln_epss = vulnerability.get("epss")
 
             rel_datasource = None
             rel_urls = None
             rel_description = None
             rel_cvss = None
+            rel_epss = None
             vulnerability_ids = None
             related_vulnerabilities = item.get("relatedVulnerabilities")
             if related_vulnerabilities:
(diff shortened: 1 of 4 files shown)


dojo/tools/anchore_grype/parser.py
```diff
@@ -168,8 +175,9 @@ def get_findings(self, file, test):
                 dupes[dupe_key] = Finding(
                     title=finding_title.replace("\x00", ""),
                     description=finding_description.replace("\x00", ""),
-                    cwe=1352,
```
Maffooch, line 171: Why remove the CWE value here?

bwt-sloanj: Happy to reverse it, but I removed it because it was the only parser that was using that as a default CWE for Findings. When looking at the results of a scan in the UI I feel it misleadingly implies that Grype supplied that value and having the same value for every single finding on the page it becomes visual noise to me. So... I thought I'd take the chance to remove it.

bwt-sloanj: The link it gives in DD also seems to discourage its use for individual vulnerabilities: https://cwe.mitre.org/data/definitions/1352.html

Maffooch: I always get a little twitch when touching the CWE because it is one of the fields in the default dedupe settings. For Anchore Grype, it looks like that is not the case: https://github.com/DefectDojo/django-DefectDojo/blob/ae1f44c655cca0faf27fcd1ddc9d34a5486ef253/dojo/settings/settings.dist.py#L1248

I think it could be safe to remove this field after all!

bwt-sloanj: Ah gotcha - if you prefer I could separate that into a separate commit so it's easier to trace and revert? Or a separate PR to make it explicit?

valentijnscholten (approved):

dojo/tools/anchore_grype/parser.py
bwt-sloanj: Can this be resolved?

Maffooch (approved):

mtesauro (approved):
Approved

PR #12650, opened 2025-06-21
Async Delete: Correct instances of multiple audit log entries for delete

Maffooch:
Correcting the following:
```python
  File "/usr/local/lib/python3.11/site-packages/celery/app/trace.py", line 453, in trace_task
    R = retval = fun(*args, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/celery/app/trace.py", line 736, in __protected_call__
    return self.run(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/utils.py", line 2331, in delete_chunk
    obj.delete()
  File "/app/dojo/models.py", line 1614, in delete
  File "/usr/local/lib/python3.11/site-packages/django/db/models/base.py", line 1273, in delete
    return collector.delete()
           ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/db/models/deletion.py", line 508, in delete
    signals.post_delete.send(
  File "/usr/local/lib/python3.11/site-packages/django/dispatch/dispatcher.py", line 189, in send
    response = receiver(signal=self, sender=sender, **named)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/engagement/signals.py", line 43, in engagement_post_delete
    le = LogEntry.objects.get(
         ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/db/models/query.py", line 652, in get
    raise self.model.MultipleObjectsReturned(
auditlog.models.LogEntry.MultipleObjectsReturned: get() returned more than one LogEntry -- it returned 2!
```

### DIFF
--- a/dojo/endpoint/signals.py
+++ b/dojo/endpoint/signals.py
@@ -19,7 +19,7 @@ def endpoint_post_delete(sender, instance, using, origin, **kwargs):
                 action=LogEntry.Action.DELETE,
                 content_type=ContentType.objects.get(app_label="dojo", model="endpoint"),
                 object_id=instance.id,
-            ).order_by('-id').first():
+            ).order_by("-id").first():
                 description = _('The endpoint "%(name)s" was deleted by %(user)s') % {
                                 "name": str(instance), "user": le.actor}
         create_notification(event="endpoint_deleted",  # template does not exists, it will default to "other" but this event name needs to stay because of unit testing
--- a/dojo/endpoint/signals.py
+++ b/dojo/endpoint/signals.py
@@ -13,16 +13,15 @@ from dojo.notifications.helper import create_notification
 @receiver(post_delete, sender=Endpoint)
 def endpoint_post_delete(sender, instance, using, origin, **kwargs):
     if instance == origin:
+        description = _('The endpoint "%(name)s" was deleted') % {"name": str(instance)}
         if settings.ENABLE_AUDITLOG:
-            le = LogEntry.objects.get(
+            if le := LogEntry.objects.filter(
                 action=LogEntry.Action.DELETE,
                 content_type=ContentType.objects.get(app_label="dojo", model="endpoint"),
                 object_id=instance.id,
-            )
-            description = _('The endpoint "%(name)s" was deleted by %(user)s') % {
+            ).order_by('-id').first():
+                description = _('The endpoint "%(name)s" was deleted by %(user)s') % {
                                 "name": str(instance), "user": le.actor}
-        else:
-            description = _('The endpoint "%(name)s" was deleted') % {"name": str(instance)}
         create_notification(event="endpoint_deleted",  # template does not exists, it will default to "other" but this event name needs to stay because of unit testing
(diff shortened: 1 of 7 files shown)


mtesauro (approved):
Approved

dojo/utils.py
```diff
@@ -2333,6 +2335,15 @@ def delete_chunk(self, objects, **kwargs):
                 logger.debug("ASYNC_DELETE: object has already been deleted elsewhere. Skipping")
                 # The id must be None
                 # The object has already been deleted elsewhere
+            except LogEntry.MultipleObjectsReturned:
+                # Delete the log entrys first, then delete
+                LogEntry.objects.filter(
+                    content_type=ContentType.objects.get_for_model(obj.__class__),
+                    object_pk=str(obj.pk),
+                    action=LogEntry.Action.DELETE,
```
valentijnscholten, line 2343: I don' t have any context, but why is it enough to only delete LogEntry objects with DELETE action? Is it that these are added during deletion and we have to remove them again/in time?

Maffooch: Correct, notifications are doing a `.get` which will fail unless there is exactly one object. I think I misread the exception, because I was under the impression that the exception was being raised from the audit log lib. It turns out, the exception is being raised from notification related activities. I am going to make those a bit more resilient as well. Thank you for raising this!

dojo/endpoint/signals.py
```diff
@@ -13,16 +13,15 @@
...
                 action=LogEntry.Action.DELETE,
                 content_type=ContentType.objects.get(app_label="dojo", model="endpoint"),
                 object_id=instance.id,
-            )
-            description = _('The endpoint "%(name)s" was deleted by %(user)s') % {
+            ).order_by("-id").first():
+                description = _('The endpoint "%(name)s" was deleted by %(user)s') % {
                                 "name": str(instance), "user": le.actor}
-        else:
-            description = _('The endpoint "%(name)s" was deleted') % {"name": str(instance)}
```
valentijnscholten, line 25: Is this removing the description in case audit logging is not enabled?

Maffooch: Nope, just moved it up a few lines to fit this model:
```python
description = "something generic"
if auditlog enabled:
    if auditlog object is fetched
        description = "more detailed description"
return description
```

valentijnscholten: Ah yes of course

valentijnscholten (approved):

hblankenship (approved):

dogboat (approved):

PR #12669, opened 2025-06-21
Feature/asvs 5.0 benchmark

ivhorodko:
### 🐞 Bug fix  
* **Symptom:** The Benchmark-Type dropdown on the product page always showed a single entry, **“OWASP ASVS v3.1”**, even though **v4.0.1** (and other types) were already loaded in the DB.  
* **Root cause:** The view didn’t pass the full list of enabled `Benchmark_Type` records to the template.  
* **Fix:** `dojo/product/views.py` now fetches `Benchmark_Type.objects.filter(enabled=True)` and exposes it in the context → the dropdown lists *all* enabled benchmarks.

### ✨ Feature / Enhancement
* **Added OWASP ASVS 5.0**:
Includes the latest ASVS 5.0 as a selectable benchmark, with new categories and initial requirements.
* **Why**:
Previously, only the old ASVS 3.1 was visible in the dropdown, even though 4.0.1 was already in the database. ASVS 3.1 is quite outdated, so it’s nice to have more up-to-date standards available.
* **What changed**:
The dropdown now shows all available ASVS versions (including 5.0), so you can map findings to the most current requirements.

---

## Screenshots  🖼️

| Before | After |
| ------ | ----- |
| image | image |

Additional screenshots of the new ASVS 5.0 benchmark page:  
image

---

## References  🔗

Place any external links here (PDF, CSV, official page, etc.):

* [Official ASVS page](https://owasp.org/www-project-application-security-verification-standard/)
* [Official OWASP ASVS 5.0 PDF](https://github.com/OWASP/ASVS/raw/v5.0.0/5.0/OWASP_Application_Security_Verification_Standard_5.0.0_en.pdf)
* [CSV source used for import](https://github.com/OWASP/ASVS/releases/download/v5.0.0_release/OWASP_Application_Security_Verification_Standard_5.0.0_en.csv)

---

## Test results

1. Loaded fixtures on a fresh database (`manage.py loaddata …`).  
2. Verified the dropdown now lists **ASVS 3.1, ASVS 4.0.1, ASVS 5.0**.  
3. Created a dummy product → mapped findings to new v5.0 requirements without issues.

### DIFF
--- a/dojo/product/views.py
+++ b/dojo/product/views.py
@@ -71,6 +71,7 @@ from dojo.forms import (
 from dojo.models import (
     App_Analysis,
     Benchmark_Product_Summary,
+    Benchmark_Type,
     BurpRawRequestResponse,
     DojoMeta,
     Endpoint,
@@ -90,7 +91,6 @@ from dojo.models import (
     System_Settings,
     Test,
     Test_Type,
-    Benchmark_Type,
 )
 from dojo.product.queries import (
     get_authorized_global_groups_for_product,
--- a/dojo/product/views.py
+++ b/dojo/product/views.py
@@ -148,7 +148,7 @@ def product(request):
     prod_list.object_list = prefetch_for_product(prod_list.object_list)
 
     # Get benchmark types for the template
-    benchmark_type = Benchmark_Type.objects.filter(enabled=True).order_by("name")
+    benchmark_types = Benchmark_Type.objects.filter(enabled=True).order_by("name")
 
     add_breadcrumb(title=_("Product List"), top_level=not len(request.GET), request=request)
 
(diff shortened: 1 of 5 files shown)


dojo/product/views.py
```diff
@@ -146,13 +147,17 @@ def product(request):
     # perform annotation/prefetching by replacing the queryset in the page with an annotated/prefetched queryset.
     prod_list.object_list = prefetch_for_product(prod_list.object_list)
 
+    # Get benchmark types for the template
+    benchmark_type = Benchmark_Type.objects.filter(enabled=True).order_by("name")
```
valentijnscholten: Shouldn't this be plural?

ivhorodko: Good point, changed to benchmark_types for clarity. Was following the old naming for consistency, but plural is better here. Updated in both view and template. Thanks!

mtesauro (approved):
Approved

hblankenship (approved):

Issue #6907, opened 2022-09-27
Wrong keys in several requests causing Status 500 Errors


ilomax:
**Bug description**
I have been receiving 500 responses while performing different actions in DefectDojo. The two cases I have noticed for now are when logging out of my account, and upon deleting a group in the administration panel.

**Steps to reproduce**
- **Logout:**
1. Log in with AzureAD SSO.
2. Just clicking the _"Logout"_ button in the profile dropdown
The error occurs whether the account is new or already existing.

- **Group deletion:**
1. Go to `/admin/auth/group/`
2. Select a group or click on it to inspect it
4. Select the "Delete selected groups" action or click the "DELETE" button on the group's page

**Expected behavior**
In the first case, the user should be disconnected and, I'm guessing, shown the login page.
In the second one, the group should be deleted.

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
- Version: `2.14.3` (also in `2.12.0`)

**Logs** 

- **Logout:**
```
│| uwsgi [27/Sep/2022 08:39:07] ERROR [django.request:224] Internal Server Error: /login
││ uwsgi Traceback (most recent call last):
││ uwsgi   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
││ uwsgi     response = get_response(request)
│| uwsgi   File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
││ uwsgi     response = wrapped_callback(request, *callback_args, **callback_kwargs)
│| uwsgi   File "/app/./dojo/decorators.py", line 186, in _wrapped
││ uwsgi     return fn(request, *args, **kw)
││ uwsgi   File "/app/./dojo/decorators.py", line 186, in _wrapped
││ uwsgi     return fn(request, *args, **kw)
│| uwsgi   File "/app/./dojo/user/views.py", line 115, in login_view
││ uwsgi     urlencode({'next': request.GET.get('next')})))
││ uwsgi   File "/usr/local/lib/python3.8/site-packages/django/utils/http.py", line 113, in urlencode
││ uwsgi     raise TypeError(
│| uwsgi TypeError: Cannot encode None for key 'next' in a query string. Did you mean to pass an empty string or omit the value?
```

- **Group deletion:**
```
[...]

PR #12677, opened 2025-06-23
login next param: set default for sso redirects

valentijnscholten:
In certain SSO scenario's the user can end up with a `/login` url without any `next` parameter. This would result in a ` NoneType`  error. This PR fixes this by sending users to `/dashboard`  by default. Inspired by #6907


Maffooch (approved):

mtesauro (approved):
Approved

PR #12686, opened 2025-06-24
trivy: map status field

valentijnscholten:
Trivy has a "status" field for triaging which is currently unmapped:
"status": "affected" or "not_affected" or "fix_deferred" or "will_not_fix"

This would be useful information to handle.

see: https://trivy.dev/latest/docs/configuration/filtering/

"Affected": the package is affected by the flaw on this platform
"Not affected": the package, which ships on this platform, is not affected by the flaw
"Fix deferred": the package is affected by the flaw on this platform, and may be fixed in the future
"Under investigation": it is currently unknown whether the flaw affects the package on this platform, and is under investigation
"Will not fix": the package is affected by the flaw on this platform, but there is no current intention to fix it (primarily for flaws that are of Low or Moderate impact that pose no significant risk to customers)​

This PR stores the Trivy `status` value in the `Impact` field.

It also tries to map the `status`  to Defect Dojo status flags. This is not trivial as the `status` of the vulnerability in the Trivy report is set by the OS/Package vendor. If they state `will_not_fix` it doesn't automatically mean Defect Dojo needs to close the finding on import.

The ` trivy.md`  file in this PR contains a table with status flag mappings.

[sc-11484]

### DIFF
--- a/docs/content/en/connecting_your_tools/parsers/file/trivy.md
+++ b/docs/content/en/connecting_your_tools/parsers/file/trivy.md
@@ -4,7 +4,7 @@ toc_hide: true
 ---
 JSON report of [trivy scanner](https://github.com/aquasecurity/trivy).
 
-The status field in Trivy is mapped to the Defect Dojo status flags in the following way:
+The [status](https://trivy.dev/latest/docs/configuration/filtering/) field in Trivy is mapped to the Defect Dojo status flags in the following way:
 
 | Trivy Status         | Active | Verified | Mitigated | False Positive | Remarks                                                                                                         |
 |----------------------|--------|----------|-----------|---------------|-----------------------------------------------------------------------------------------------------------------|
@@ -19,6 +19,7 @@ The status field in Trivy is mapped to the Defect Dojo status flags in the follo
 
 The status field contains the status as assigned by the OS/Package vendor such as Red Hat, Debian, etc.
 As s Defect Dojo user you still have to asses the appropiate action in your product context.
+If you want to exclude certain status from being imported into Defect Dojo, please [filter them in the export from Trivy](https://trivy.dev/latest/docs/configuration/filtering/)
 
 ### Sample Scan Data
 Sample Trivy scans can be found [here](https://github.com/DefectDojo/django-DefectDojo/tree/master/unittests/scans/trivy).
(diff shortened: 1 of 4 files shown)


docs/content/en/connecting_your_tools/parsers/file/trivy.md
```diff
@@ -4,5 +4,22 @@ toc_hide: true
...
+| under_investigation  | True   | False    | False     |               | no status flag in Defect Dojo to capture this, but verified is False                                            |
+| will_not_fix         | True   | True     | False     |               | no different from affected as Defect Dojo doesn't have a flag to capture will_not_fix by OS/Package Vendor; we can't set active to False as the user needs to risk accept this finding |
+| fix_deferred         | True   | True     | False     |               | no different from affected as Defect Dojo doesn't have a flag to capture will_not_fix by OS/Package Vendor; we can't set active to False as the user needs to (temporarily) risk accept this finding |
+| end_of_life          | True   | True     | False     |               | no different from affected as Defect Dojo doesn't have a flag to capture will_not_fix by OS/Package Vendor; we can't set active to False as the user needs to (temporarily) risk accept
+
+The status field contains the status as assigned by the OS/Package vendor such as Red Hat, Debian, etc.
+As s Defect Dojo user you still have to asses the appropiate action in your product context.
```
paulOsinski: Can you change line 21 to
`It is recommended to assess the appropriate action in your Product's context.`

paulOsinski:
I agree with most of these except where `not_affected` = `False Positive`, because different users use the False Positive label for different workflows and I don't want to make assumptions.  I would just leave this as `Active` and then the user can check the Impact section to decide what they want to do from there.

docs/content/en/connecting_your_tools/parsers/file/trivy.md
valentijnscholten: done

valentijnscholten:
> I agree with most of these except where `not_affected` = `False Positive`, because different users use the False Positive label for different workflows and I don't want to make assumptions. I would just leave this as `Active` and then the user can check the Impact section to decide what they want to do from there.

adjusted

hblankenship (approved):

Issue #7961, opened 2023-04-12
Moving from bitnami to Docker OCI as Helm source


dsever:
Bitnami is moving helm content to docker OCI as primary source

https://blog.bitnami.com/2023/04/httpsblog.bitnami.com202304bitnami-helm-charts-now-oci.html

valentijnscholten:
We've moved away from Bitnami.

Issue #11039, opened 2024-10-10
Initializer job name in Helm chart uses date which prevent application in ArgoCD from being in Sync


Brawdunoir:
Hi,

I’m using the Helm chart version 1.6.153

**Is your feature request related to a problem? Please describe**
Because of the helper `initializer.jobname`, a new job could be created each minute.
This results in a permanent OutOfSync status of the defectdojo application within Argo CD, with a missing job ressource.

**Describe the solution you'd like**
One solution could be to use [Helm hooks](https://helm.sh/docs/topics/charts_hooks/) like `pre-install` and/or `pre-upgrade` and always use the same job name (thanks to hook deletion policies).

**Additional context**
Helper causing the problem:
```yaml
{{- define "initializer.jobname" -}}
{{ .Release.Name }}-initializer-{{- printf "%s" now | date "2006-01-02-15-04" -}}
{{- end -}}
```
Argo CD dashboard:
![{7743EE4A-399F-49EA-B326-B34A18A33623}](https://github.com/user-attachments/assets/1bec21c3-4bde-46ee-a439-80110f9f3721)

kiblik:
Thank you, @Brawdunoir, for pointing this out. I had planned to look closer at my similar/related issues, and you probably found the reason.

Unfortunately, `pre-` hooks would not work well for new installations. The job would be triggered before the database would be started. The database would not be started until the job was finished (as far as I know - we have been solving a similar issue with `post-` hooks).

Still not sure what is the best solution but "same name" sounds like a suitable solution.

I would like to check it deeper.

Sidenote: this is a side-effect of how Argo handles charts. As far as I know, there is no issue for "regular" helm deployments right now.

Brawdunoir:
Thanks for the quick response! I have quite the experience of building charts and got into this job issue lately but without your chart dependencies constraint (we use Helmfile to deploy/order multiple charts).

Reusing the same job name without hooks will not work because a Job has a lot of immutable fields (including pod labels and container images). That's the reason behind the hook deletion policies.

It would be interesting to understand how other charts like Bitnami's are copping with this issue.

I suspect that because (in my knowledge) a proper mechanism from Helm itself does not exist to handle such situations, you could only rely on an init-container on your job.
Your application will still be launched, without prior initialization of the database, but hey, that's life with Kubernetes, and your application should handle such cases programmatically (with checks and retries). And after maybe a few restarts because of readiness, at the end, everything should stabilize.

[...]

PR #12691, opened 2025-06-25
feat: improve Helm chart

fernandezcuesta:
- Add extraInitContainers to celery+django deployments.
- Add extraEnv to all deployments
- Remove existing volume logic in favor of agnostic extraVolumes and extraVolumeMounts
- Add the ability to deploy secrets as regular (non-hooked) resources due to [issues found with ArgoCD](https://github.com/argoproj/argo-cd/issues/13429)
- Make some secret references optional (secret might not be created)
- Fix optional secret mounts + reference
- Update bitnami chart reference (OCI)
- Bump up redis chart

---

Reasoning behind the logic change for volumes, with the existing behaviour we cannot mount projected volumes (e.g. a secret mount where we want the key names being renamed) or per-container volumeMounts (e.g. nginx emptyDir when readOnlyRootFs is enforced).


valentijnscholten:
@fernandezcuesta Thanks for the PR, we'll ask some helm/k8s experienced devs to look at it. In the mean time could you look at the conflicts? Also this seems to be a bigger change, could you look at basing it against the `dev` branch

fernandezcuesta:
@valentijnscholten done, thanks for having a look at it! I also changed the base to `dev` branch.

helm/defectdojo/Chart.lock
```diff
@@ -1,9 +1,9 @@
 dependencies:
 - name: postgresql
-  repository: https://charts.bitnami.com/bitnami
-  version: 16.7.0
+  repository: oci://registry-1.docker.io/bitnamicharts
+  version: 16.7.13
```
kiblik: TBH, this is a bit confusing for me. I'm using this chart, and `16.7.13` is automatically selected in the final template even though it is not pinned here. I want to look at this more.

helm/defectdojo/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "2.47.3"
 description: A Helm chart for Kubernetes to install DefectDojo
 name: defectdojo
-version: 1.6.193
+version: 1.6.194-dev
```
kiblik: This is not needed. Version is updated automatically
https://github.com/DefectDojo/django-DefectDojo/blob/a40696889da9c0521c525af1060ff73d262a9a2b/.github/workflows/release-1-create-pr.yml#L71-L81

helm/defectdojo/Chart.yaml
```diff
@@ -11,9 +11,9 @@ maintainers:
 dependencies:
   - name: postgresql
     version: ~16.7.0
-    repository: "https://charts.bitnami.com/bitnami"
+    repository: oci://registry-1.docker.io/bitnamicharts
     condition: postgresql.enabled
   - name: redis
-    version: ~19.6.0
-    repository: "https://charts.bitnami.com/bitnami"
+    version: ~21.2.5
```
kiblik: This is a bit problematic. See https://github.com/DefectDojo/django-DefectDojo/pull/12393#pullrequestreview-2819835169

However, I can imagine bumping the helm chart but pinning the docker image tag. But I would recommend separated PR for this.

helm/defectdojo/Chart.yaml
```diff
@@ -11,9 +11,9 @@ maintainers:
 dependencies:
   - name: postgresql
     version: ~16.7.0
-    repository: "https://charts.bitnami.com/bitnami"
+    repository: oci://registry-1.docker.io/bitnamicharts
```
kiblik: This is a nice step. I like it.

helm/defectdojo/values.yaml
```diff
@@ -512,15 +492,6 @@ redis:
 #   DD_SOCIAL_AUTH_AUTH0_SECRET: 'xxx'
 extraConfigs: {}
 
-# To add (or override) extra variables which need to be pulled from another configMap, you can
-# use extraEnv. For example:
-# extraEnv:
```
kiblik: Full drop might be a bit problematic for people who are already using this. I'm not against migrating to your format (it is moew cleaner). But
- It is necessary to mention it in the release notes - https://github.com/DefectDojo/django-DefectDojo/blob/dev/docs/content/en/open_source/upgrading/2.48.md
- It might be a good idea to use the deprecation period (add support for `.Values.*.*.extraEnv` now, explain in release notes, create new PR which will remove `.Values.extraEnv`, merge it in 2 or 3 months)

helm/defectdojo/Chart.yaml
fernandezcuesta: Ok will undo this. Thanks for raising the flag, I guess this is due to new Redis versions not being OSS anymore? (Rings me a bell that ArgoCD had to roll it back not so long ago due to CNCF constraints)

helm/defectdojo/values.yaml
fernandezcuesta: Fair enough. I'm not familiar with the release process for this project so I'll happily follow your directions. In any case what I'm used to is cutting a breaking change release of the chart by not only bumping major but also making it clear in the release notes as you mention.

helm/defectdojo/Chart.lock
fernandezcuesta: That was autogenerated after a `helm dependency update` IIRC but can roll it back

helm/defectdojo/values.yaml
fernandezcuesta: btw I'm thinking about keeping it as is (as you mention) and rename it as commonEnv later on, wdyt?

helm/defectdojo/Chart.yaml
fernandezcuesta: restored

helm/defectdojo/Chart.yaml
fernandezcuesta: restored

helm/defectdojo/values.yaml
fernandezcuesta: restored

rossops (approved):

mtesauro (approved):
Approved

helm/defectdojo/templates/celery-beat-deployment.yaml
```diff
@@ -71,22 +69,16 @@ spec:
...
-          secretName: {{ .name }}
-          {{- else if (eq .type "hostPath") }}
-          type: {{ .pathType | default "Directory" }}
-          path: {{ .hostPath }}
-          {{- end }}
+      {{- with .Values.celery.beat.extraVolumes }}
+        {{- . | toYaml | nindent 6 }}
       {{- end }}
-      {{- if or .Values.dbMigrationChecker.enabled .Values.cloudsql.enabled }}
+      {{- if coalesce .Values.dbMigrationChecker.enabled .Values.cloudsql.enabled .Values.celery.beat.extraInitContainers }}
```
kiblik, line 75: Nice usage of `coalesce`. I have not seen it in this context yet.

helm/defectdojo/values.yaml
```diff
@@ -12,7 +12,10 @@ createPostgresqlSecret: false
 # - enabled, enables tracking configuration changes based on SHA256
 # trackConfig: disabled
 
-# extraLabels: {}
+# Avoid using pre-install hooks, which might cause issues with ArgoCD
```
kiblik, line 15: Argo works for me. I just needed to add
```yaml
      argocd.argoproj.io/hook: "Sync"
```
Plus, I added waves for smoother deployment.
Btw, #11039
But I agree that it might be a good idea to allow disabled hooks at all.

helm/defectdojo/templates/configmap.yaml
```diff
@@ -28,7 +28,7 @@ data:
   DD_CELERY_BROKER_USER: ''
   DD_CELERY_BROKER_HOST: {{ if eq .Values.celery.broker "redis" }}{{ template "redis.hostname" . }}{{ end }}
   DD_CELERY_BROKER_PORT: '{{ if eq .Values.celery.broker "redis" }}{{- if ( hasKey .Values.redis "master" ) -}}{{ .Values.redis.master.service.ports.redis }}{{ else }}6379{{ end }}{{- end -}}'
-  DD_CELERY_BROKER_PARAMS: '{{ if eq .Values.celery.broker "redis" }}{{- if .Values.redis.transportEncryption.enabled -}}{{ .Values.redis.transportEncryption.params | default "ssl_cert_reqs=optional" }}{{ end }}{{ end }}'
+  DD_CELERY_BROKER_PARAMS: '{{ if eq .Values.celery.broker "redis" }}{{- if .Values.redis.tls.enabled -}}{{ .Values.celery.brokerParams | default "ssl_cert_reqs=optional" }}{{ end }}{{ end }}'
```
kiblik: I'm not against moving params to `.Values.celery.brokerParams`, but they are becoming more general. They should not be used only if `.Values.redis.tls.enabled == true`

helm/defectdojo/templates/initializer-job.yaml
```diff
@@ -130,14 +130,15 @@ spec:
         - configMapRef:
             name: {{ $fullName }}
         - secretRef:
-            name: {{ $fullName }}
+            name: {{ $fullName }}-extrasecrets
+            optional: true
```
kiblik, line 151: I understand why this might look like a typo, but it is not. This definition is right.

`{{ $fullName }}` contains `DD_ADMIN_PASSWORD`, `DD_CREDENTIAL_AES_256_KEY` and `DD_SECRET_KEY` which are necesseary for init of application.
Values from `{{ $fullName }}-extrasecrets` are usually not needed for initialization. But we might add them.

```suggestion
            name: {{ $fullName }}
        - secretRef:
            name: {{ $fullName }}-extrasecrets
            optional: true
```

or feel free to move values from `{{ $fullName }}` to
```yaml
        - name: DD_XXX
          valueFrom:
            secretKeyRef:
              name: {{ $fullName }}
              key: DD_XXX
              optional: true
```

It might make more sense to do it the other way around (drop separate loading of `DD_CREDENTIAL_AES_256_KEY` and `DD_SECRET_KEY` from all deployments and use there):
```yaml
        - secretRef:
            name: {{ $fullName }}
```
On the other hand, `DD_ADMIN_PASSWORD` is not needed in deployments (only in the initializer). So it might be against the least privilege principle. 

Btw, it is probably the reason why the test is failing. `DD_ADMIN_PASSWORD` was not provided to the initializer, so it used a random password.

helm/defectdojo/values.yaml
```diff
@@ -358,52 +358,27 @@ initializer:
...
-  # - `path` - defines where file should be exposed
-  # - `subPath` - extracts only particular file from secret or configMap
-  # - `pathType` - only for hostPath, can be one of the "DirectoryOrCreate", "Directory" (default), "FileOrCreate",
-  #   "File", "Socket", "CharDevice", "BlockDevice"
-  # - `hostPath` - only for hostPath, file or directory from local host
-  # @type: array<map>
+  # Additional environment variables injected to the initializer job pods.
+  extraEnv: []
+  # Array of additional volume mount points for the initializer job pods.
+  extraVolumeMounts: []
```
kiblik, line 383: You probably wanted to edit also 
```yaml
      {{- range .Values.initializer.extraVolumes }}
      - name: userconfig-{{ .name }}
        {{ .type }}:
          {{- if (eq .type "configMap") }}
          name: {{ .name }}
          {{- else if (eq .type "secret") }}
          secretName: {{ .name }}
          {{- else if (eq .type "hostPath") }}
          type: {{ .pathType | default "Directory" }}
          path: {{ .hostPath }}
          {{- end }}
      {{- end }}
```

and

```yaml
        {{- range .Values.initializer.extraVolumes }}
        - name: userconfig-{{ .name }}
          readOnly: true
          mountPath: {{ .path }}
          subPath: {{ .subPath }}
        {{- end }}
```
in `helm/defectdojo/templates/initializer-job.yaml`

helm/defectdojo/templates/tests/unit-tests.yaml
```diff
@@ -51,11 +51,13 @@ spec:
             secretKeyRef:
               name: {{ $fullName }}
               key: DD_SECRET_KEY
+              optional: true
         - name: DD_CREDENTIAL_AES_256_KEY
           valueFrom:
             secretKeyRef:
               name: {{ $fullName }}
               key: DD_CREDENTIAL_AES_256_KEY
+              optional: true
```
kiblik, line 60: I just realised, it might be a good idea to use the same approach (with `valueFrom.secretKeyRef` for `DD_ADMIN_PASSWORD`) instead of the command `ADMIN_PASS=$(kubectl get secret/defectdojo -o jsonpath='{.data.DD_ADMIN_PASSWORD}' | base64 -d)`

helm/defectdojo/templates/django-deployment.yaml
```diff
@@ -12,20 +12,18 @@ metadata:
...
 spec:
   replicas: {{ .Values.django.replicas }}
   {{- with .Values.django.strategy }}
   strategy:
     {{- toYaml . | nindent 4 }}
   {{- end }}
-  {{- if .Values.revisionHistoryLimit }}
-  revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
+  {{- with .Values.revisionHistoryLimit }}
+  revisionHistoryLimit: {{ . }}
```
kiblik, line 26: `revisionHistoryLimit` is in multiple files. Can you replace it in all of them, please?

```
./templates/celery-worker-deployment.yaml
./templates/django-deployment.yaml
./templates/celery-beat-deployment.yaml
```

helm/defectdojo/templates/django-deployment.yaml
```diff
@@ -61,9 +59,9 @@ spec:
       {{- end }}
     spec:
       serviceAccountName: {{ include "defectdojo.serviceAccountName" . }}
-      {{- if .Values.imagePullSecrets }}
+      {{- with .Values.imagePullSecrets }}
```
kiblik, line 62: Same as `revisionHistoryLimit` comment

helm/defectdojo/values.yaml
```diff
@@ -147,11 +150,32 @@ celery:
   broker: redis
   # To use an external celery broker, set the hostname here
   brokerHost: ""
+  # Parameters attached to the broker URL, defaults to "ssl_cert_reqs=optional" if redis.tls.enabled
+  brokerParams: ""
   logLevel: INFO
   annotations: {}
```
kiblik, line 153: This might be dropped. It is not used anymore.
```suggestion
```

helm/defectdojo/values.yaml
```diff
@@ -147,11 +150,32 @@ celery:
...
+    extraEnv: []
+    # A list of additional initContainers to run before celery beat containers.
+    extraInitContainers: []
+    # Array of additional volume mount points for the celery beat containers.
+    extraVolumeMounts: []
+    # A list of extra volumes to mount
+    # @type: array<map>
+    extraVolumes: []
+    # Enable liveness probe for Celery beat container.
+    livenessProbe: {}
```
kiblik, line 168: Also `readinessProbe` and `startupProbe` are mentioned in the template, but only `livenessProbe` is here. Can you add others as well, please?

helm/defectdojo/templates/celery-worker-deployment.yaml
```diff
@@ -69,20 +67,14 @@ spec:
...
-        {{ .type }}:
-          {{- if (eq .type "configMap") }}
-          name: {{ .name }}
-          {{- else if (eq .type "secret") }}
-          secretName: {{ .name }}
-          {{- else if (eq .type "hostPath") }}
-          type: {{ .pathType | default "Directory" }}
-          path: {{ .hostPath }}
-          {{- end }}
+      {{- with .Values.celery.beat.extraVolumes }}
```
kiblik: Probably copy-paste issue
```suggestion
      {{- with .Values.celery.worker.extraVolumes }}
```

helm/defectdojo/templates/celery-beat-deployment.yaml
```diff
@@ -11,14 +11,12 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     helm.sh/chart: {{ include "defectdojo.chart" . }}
     {{- with .Values.extraLabels }}
-        {{- toYaml . | nindent 4 }}
+      {{- toYaml . | nindent 4 }}
     {{- end }}
-{{- if .Values.celery.annotations }}
+  {{- with .Values.celery.worker.annotations }}
```
kiblik: Some copy-paste typo probably
```suggestion
  {{- with .Values.celery.beat.annotations }}
```
Plus, this might be the issue in some cases. Until now, `.Values.celery.worker.annotations` was used to annotate pods created from this 
deployment. Annotating deployment with exactly the same annotations might be problematic for somebody.
In the initializer, there are `annotations` and `jobAnnotations`. What about a similar approach?

helm/defectdojo/values.yaml
```diff
@@ -165,6 +189,25 @@ celery:
...
+    extraEnv: []
+    # A list of additional initContainers to run before celery worker containers.
+    extraInitContainers: []
+    # Array of additional volume mount points for the celery worker containers.
+    extraVolumeMounts: []
+    # A list of extra volumes to mount.
+    # @type: array<map>
+    extraVolumes: []
+    # Enable liveness probe for Celery worker containers.
+    livenessProbe: {}
```
kiblik, line 207: Same as beat

helm/defectdojo/values.yaml
```diff
@@ -478,26 +453,31 @@ gke:
...
-    enabled: false
-    params: ''
   auth:
     existingSecret: defectdojo-redis-specific
     existingSecretPasswordKey: redis-password
     password: ""
   architecture: standalone
-  # To use an external Redis instance, set enabled to false and uncomment
-  # the line below:
-  # redisServer: myrediscluster
```
kiblik, line 494: Why dropping this?

helm/defectdojo/templates/celery-worker-deployment.yaml
```diff
@@ -13,12 +13,10 @@ metadata:
     {{- with .Values.extraLabels }}
       {{- toYaml . | nindent 4 }}
     {{- end }}
-{{- if .Values.celery.annotations }}
+  {{- with .Values.celery.worker.annotations }}
```
kiblik: Same comment as beat

kiblik (changes requested):
Sorry for the delay, I had been out for a couple of weeks, plus your PR is changing quite a lot of variables.

This is changing quite critical settings. It should be mentioned in the [release notes](https://github.com/DefectDojo/django-DefectDojo/tree/master/docs/content/en/open_source/upgrading).

But thank you for this work. I was planning to add some adjustments that you used, but never had enough time.

kiblik:
Btw, this fix #7961

helm/defectdojo/values.yaml
fernandezcuesta: I wasn't able to make it work with that hook, not entirely sure this was a local issue or not though.

helm/defectdojo/templates/configmap.yaml
fernandezcuesta: OK I see your point, changing that!

helm/defectdojo/values.yaml
fernandezcuesta: I guess it's safer to leave it as a "common celery annotations" field, like something you expect in both beat/worker deployments+pods.

helm/defectdojo/values.yaml
fernandezcuesta: Moved to the comment in L473:

```
 # To use an external Redis instance, switch enabled to false and set the address in .Values.celery.brokerHost
```

helm/defectdojo/templates/tests/unit-tests.yaml
fernandezcuesta: I'm a bit lost here, I can inject DD_ADMIN_PASSWORD to the unit-test pod, but how that it works? I mean, k8s-tests is doing the "Simple API check" against the Django service and it needs the credentials for that.

fernandezcuesta:
@kiblik Thanks for your detailed review and suggestions. I tried to go through all, but have some doubts about one of them (see inline comments).

helm/defectdojo/Chart.lock
```diff
@@ -1,9 +1,9 @@
 dependencies:
 - name: postgresql
-  repository: https://charts.bitnami.com/bitnami
-  version: 16.7.0
+  repository: oci://registry-1.docker.io/bitnamicharts
```
kiblik: This will not be needed soon (#13039), but I would keep it in the context of this PR

helm/defectdojo/templates/configmap.yaml
kiblik: I like this solution

helm/defectdojo/values.yaml
kiblik: Hmmm... I'm just checking `_helpers.tpl` and I'm thinking what is the right approach here 🤔.

https://github.com/DefectDojo/django-DefectDojo/blob/941dd6cfb4f609367719ef26ee25b0f880f0c5cc/helm/defectdojo/templates/_helpers.tpl#L45-L67

- Postgres and Redis are both essential parts for DD.
- They might be used as part of the stack, or both can be used _aaS_ or any combination between
- Postgres is used by all Django components (uwsgi, celery, initializer), same for Redis (well, except initializer, but definitely not only celery)
- The following two statements should use the same conventions (they do not follow right now)
```
{{- printf "%s" ( .Values.postgresql.postgresServer | default "127.0.0.1" ) -}}
{{- printf "%s" (.Values.celery.brokerHost | default .Values.redis.redisServer) -}}
```

helm/defectdojo/templates/tests/unit-tests.yaml
kiblik: I'm sorry, I confused `helm/defectdojo/templates/tests/unit-tests.yaml` with `.github/workflows/k8s-tests.yml`. My comment is irrelevant.

kiblik (changes requested):
Many of the issues have been solved; some small ones are still open.
However, I'm still missing some info in the [release notes](https://github.com/DefectDojo/django-DefectDojo/blob/dev/docs/content/en/open_source/upgrading/2.50.md)
Mainly parts that have potential side-effects (rename `annotations` ot `podAnnotations` in celery)

helm/defectdojo/values.yaml
fernandezcuesta: OK now I see your point. I did yet another change to the external db (pg/redis) values which I propose to take out from the umbrella chart values to avoid potential breaks in the upstream chart (e.g. `redis` helm chart doing strict schema validation). As per Release notes (which were missing, sorry!):

- `redis.redisServer` → `redisServer` (prevents potential schema conflicts with upstream chart)
- `redis.transportEncryption.params` → `redisParams` (prevents potential schema conflicts with upstream chart)
- `postgresql.postgresServer` → `postgresServer` (prevents potential schema conflicts with upstream chart)

docs/content/en/open_source/upgrading/2.50.md
```diff
@@ -2,10 +2,54 @@
...
+- **Enhanced probe configuration for Celery**: Added support for customizing liveness, readiness, and startup probes in both Celery beat and worker deployments.
+- **Enhanced environment variable management**: All deployments now include `extraEnv` support for adding custom environment variables. For backwards compatibility, `.Values.extraEnv` can be used to inject common environment variables to all workloads.
+
+### Other changes
+
+- **Celery pod annotations**: Now we can add annotations to Celery beat/worker pods separately.
+- **Flexible secret deployment**: Added the capability to deploy secrets as regular (non-hooked) resources to address compatibility issues encountered with CI/CD tools (such as ArgoCD).
+- **Optional secret references**: Some secret references are now optional, allowing the chart to function even when certain secrets are not created.
+- **Fixed secret mounting**: Resolved issues with optional secret mounts and references.
+- **Updated Bitnami chart reference**: Migrated to OCI (Open Container Initiative) format for the Bitnami chart dependency.
```
kiblik:
```suggestion
- **Updated Bitnami chart reference**: Migrated to OCI (Open Container Initiative) format for the chart dependency.
```
This part might need to be edited based on #13063 
CC @rossops

helm/defectdojo/templates/_helpers.tpl
```diff
@@ -53,15 +53,16 @@ Create the name of the service account to use
 {{-   printf "%s-%s" .Release.Name "postgresql" | trunc 63 | trimSuffix "-" -}}
 {{-  end -}}
 {{- else -}}
-{{-  printf "%s" ( .Values.postgresql.postgresServer | default "127.0.0.1" ) -}}
+{{- .Values.postgresServer | default "127.0.0.1" | quote -}}
```
kiblik, line 56: CC @lchastel
JFYI, your #12965 has not been released yet, but it will probably change the format before it is officially released. Please see https://github.com/DefectDojo/django-DefectDojo/pull/12691#discussion_r2303088727 for full context.

helm/defectdojo/values.yaml
kiblik: I really like this separation. I was planning to add `values.schema.json` (see #12984) to avoid wrong values. And I wasn't sure what to do with the overlap with subchart values. As soon as we separate them, it is much easier to adopt them.

kiblik (changes requested):
Just a small issue - I would drop the Bitnami reference from the release notes (see the context)

And TBH, I would probably change the chart version to `1.7.1` (it needs to be `1.7.1-dev`, the suffix will be removed during releasing).
Because this is a radical change, and it should be noticeable in the version number as well. It not only adds value, but it might also affect existing deployments.

Huge thanks, this PR is a large cleanup and makes the chart more flexible and easier to read and use.

kiblik:
I really hoped this will be part of 2.50.0
As it wasn't merged before release, changelog has to be moved to `2.51.md` file.

fernandezcuesta:
> I really hoped this will be part of 2.50.0 As it wasn't merged before release, changelog has to be moved to `2.51.md` file.

Done!

Maffooch (approved):
Apologies for missing this one folks 😅

valentijnscholten:
@kiblik I merged this as it looks it's still valid and doesn't conflict with the vendoring of the bitnami charts we did last week?

PR #12703, opened 2025-06-26
integration tests: sync suite between GHA and entrypoint

valentijnscholten:
I noticed the integration test suite in GHA and the container entrypoint were slightly out of sync.


mtesauro (approved):
Approved

dogboat (approved):

Issue #10268, opened 2024-05-24
Finding Group Filter has too many options


maxwai:
**Bug description**

When using the filter on the general "All Findings" Page (or other page where all Findings regardless of Product/Engagement are shown), all known groups are correctly shown in the List.
But when using the filter while on an Engagement Test, I also get all known groups. But I would expect to only see the Groups that belong to this Engagement Test since the Findings shown can only belong to the Groups of the current Engagement Test.

**Steps to reproduce**
1. Create multiple Tests/Engagements
2. Create groups in each Test
3. Try to filter by finding group inside a Test -> all finding groups, also from other Tests, are showing up.

**Expected behavior**

The Finding Group filter should only show finding groups that are available in the given context:

- all if viewing all findings
- only groups of Tests belonging to the product if viewing product findings
- only groups of the current Test if viewing findings in a Test

**Deployment method** *(select with an `X`)*
- [x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: not relevant
 - DefectDojo version: v2.33.6 but isn't fixed in v2.34.4 according to release notes

valentijnscholten:
Turns out the Finding Filter class doesn't know which Engagement or Test the user is looking it. Only the Product context is known. I raised to #12711 to limit the options in the dropdown to the Finding Groups inside the same Product. Hope this helps.

maxwai:
While this helps already, in the end the Finding Filter would need to be changed to also be aware of the engagement since it is highly possible to have groups with the same name between engagements of the same product.

AbhimanyuGit2507:
Hi @maxwai I would like to work on this issue if its still open, could you assign it to me?

valentijnscholten:
@AbhimanyuGit2507 feel free to submit a PR

maxwai:
I'm not a collaborator so I can't

AbhimanyuGit2507:
> I'm not a collaborator so I can't

Oh sorry for the misunderstanding, I'm actually new to the repo hence the mistake...

PR #12711, opened 2025-06-27
finding groups: filter by product if applicable

valentijnscholten:
Partly fixes: #10268

This PR limits the list of Finding Group options in the Finding Filter to those groups that belong to the Product the user is viewing.
This also applies to when viewing findings by Test or by Enagement. In that case Finding Groups from the whole product will be shown. This is because currently the FindingFilter class is not aware of which Engagement or Test is being viewed. It only has the Product as context via `self.pid`. I didn't want to refactor too many things and I think limiting the list to the Product context is already a good improvement.


mtesauro (approved):
Approved

hblankenship (approved):

PR #12712, opened 2025-06-27
remove actual_time and estimated_time fields

valentijnscholten:
Removes `Test` fields that aren't used/referenced anywhere:
- `actual_time`
- `estimated_time`


mtesauro:
Wow, those are some old model fields.

When they were added we were considering tracking the accuracy of estimating manual testing vs the actual time it took the manual test to complete.

As you noticed, that idea was never fully implemented. Thanks for the clean-up work.

mtesauro (approved):
Approved

Maffooch (approved):

PR #12714, opened 2025-06-28
dev: hot reloading improvements celery/html/tpl

valentijnscholten:
Improvements to hot-reloading/autoreload:
- fix path to `/app/dojo/settings/settings.py`
- also reload on `.html` and `.tpl` changes
- enable autoreload on the `celeryworker`

For `celery` it could be done without `watchmedo`, but that would require installting pypi package `celery[dev]` which would end up also in production. I'm not sure of the impact of that, so I avoided it.

### DIFF
--- a/Dockerfile.django-alpine
+++ b/Dockerfile.django-alpine
@@ -67,6 +67,7 @@ RUN export PYCURL_SSL_LIBRARY=openssl && \
 COPY \
   docker/entrypoint-celery-beat.sh \
   docker/entrypoint-celery-worker.sh \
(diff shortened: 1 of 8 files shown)


mtesauro:
>For celery it could be done without watchmedo, but that would require installting pypi package celery[dev] which would end up also in production. I'm not sure of the impact of that, so I avoided it.

I agree with this :point_up:  thinking

mtesauro (approved):
Approved

Maffooch (approved):
This is legit life changing

dogboat (approved):

PR #12716, opened 2025-06-29
unittests: import query/task count capture

valentijnscholten:
This PR ads a unit test for import/reimport that captures the number of queries executed and the number of dojo async tasks created/generated.

By capturing this we help ourselves in the future when PRs affected the number of queries. Sometimes this is intended and OK. Sometimes it means there's an unforeseen side effect that lowers performance.

I have some performance related PRs coming up and it will be nice to see the numbers go down.

I choose the StackHawk parser because its samples have multiple endpoints per finding making it a good case to also cover the endpoint part of the import/reimport process.


Maffooch (approved):

hblankenship (approved):

mtesauro (approved):
Approved

PR #12732, opened 2025-07-02
jira: truncate description if max length exceeded

### LINKED ISSUE #6668: Error pushing to Jira: description limit is not checked on Dojo side
**Bug description**
Dojo doesn't check description max length before export to jira. As a result error recieved for findings with too long description.
```
<class ‘dojo.models.Finding’>: Elasticsearch Information Disclosure, The entered text is too long. It exceeded the maximum number of characters of 32,767.
```

**Expected behavior**
Trim description before push to Jira, or add as attachment file.

- [X ] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

valentijnscholten:
Fixes #6668 and [sc-11551]

JIRA has a max length of `32768` for the `description` field. This PR truncates accordingly and provides a hint to check the full description in JIRA (and links to the finding) in JIRA.

Example with 50.000 'A' characters added to the description as a test of what it looks like:

image

### DIFF
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -648,7 +648,7 @@ def jira_description(obj, **kwargs):
 
     description = render_to_string(template, kwargs)
     defect_dojo_obj_url = get_full_url(obj.get_absolute_url())
-    max_length = getattr(settings, "JIRA_DESCRIPTION_MAX_LENGTH", 32768)
+    max_length = getattr(settings, "JIRA_DESCRIPTION_MAX_LENGTH", 32767)
     suffix = f"\n\nIssue Description Too Long: See [DefectDojo|{defect_dojo_obj_url}] for full description."
     if len(description) > max_length:
         # suffix can be longer after rendering do to urlenocoding, so we take twice the length of the suffix as a buffer
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -226,7 +226,7 @@ env = environ.FileAwareEnv(
     DD_MAX_REQRESP_FROM_API=(int, -1),
     DD_MAX_AUTOCOMPLETE_WORDS=(int, 20000),
     DD_JIRA_SSL_VERIFY=(bool, True),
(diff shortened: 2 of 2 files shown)


mtesauro (approved):
Approved

dojo/jira_link/helper.py
```diff
@@ -646,6 +647,12 @@ def jira_description(obj, **kwargs):
         kwargs["finding_group"] = obj
 
     description = render_to_string(template, kwargs)
+    defect_dojo_obj_url = get_full_url(obj.get_absolute_url())
+    max_length = getattr(settings, "JIRA_DESCRIPTION_MAX_LENGTH", 32768)
```
hblankenship: I was curious where the 32768 value came from. I found this in some JIRA issues where people were requesting no limit for this field:
https://jira.atlassian.com/browse/JRACLOUD-59124

What I find here is that the maximum length is technically 32767. Is our number one too many (assuming the environment variable is unset)?

```suggestion
    max_length = getattr(settings, "JIRA_DESCRIPTION_MAX_LENGTH", 32767)
```

dojo/settings/settings.dist.py
```diff
@@ -226,6 +226,7 @@
     DD_MAX_REQRESP_FROM_API=(int, -1),
     DD_MAX_AUTOCOMPLETE_WORDS=(int, 20000),
     DD_JIRA_SSL_VERIFY=(bool, True),
+    DD_JIRA_DESCRIPTION_MAX_LENGTH=(int, 32768),
```
hblankenship: Same here as above. Is this also elsewhere in the code? Hmm...(short answer is no after searching).

```suggestion
    DD_JIRA_DESCRIPTION_MAX_LENGTH=(int, 32767),
```

hblankenship (changes requested):

dojo/jira_link/helper.py
valentijnscholten: Nice find, must have been my muscle memory from my MSDOS days.

PR #12735, opened 2025-07-02
add option to group findings by `vuln_id_from_tool`

LeoOMaia:
based  on #12684
In this PR we're adding grouping by the vuln_id_from_tool field, as suggested in the discussion, as a viable static option for the grouping dropdown.

### DIFF
--- a/dojo/finding/helper.py
+++ b/dojo/finding/helper.py
@@ -255,6 +255,9 @@ def get_group_by_group_name(finding, finding_group_by_option):
             group_name = f"Filepath {finding.file_path}"
     elif finding_group_by_option == "finding_title":
         group_name = finding.title
+    elif finding_group_by_option == "vuln_id_from_tool":
+        if finding.vuln_id_from_tool:
+            group_name = f"Vulnerability ID {finding.vuln_id_from_tool}" if finding.vuln_id_from_tool else "None"
     else:
         msg = f"Invalid group_by option {finding_group_by_option}"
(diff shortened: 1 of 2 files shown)


valentijnscholten (changes requested):
I think this is a good addition, but can you correct the names to match `vuln_id_from_tool` instead of `script_id`?

dojo/finding/helper.py
```diff
@@ -255,6 +255,9 @@ def get_group_by_group_name(finding, finding_group_by_option):
             group_name = f"Filepath {finding.file_path}"
     elif finding_group_by_option == "finding_title":
         group_name = finding.title
+    elif finding_group_by_option == "vuln_id_from_tool":
+        if finding.vuln_id_from_tool:
+            group_name = f"Vulnerability ID {finding.vuln_id_from_tool}"
```
valentijnscholten: What happens if there is no `vuld_id_from_tool`? I think in that case the `group_name` should be set to a str value of `"None"`

LeoOMaia: makes sense, I've changed now

valentijnscholten:
Could you rebase it to `bugfix`? I yes, then it could be included in `2.48.1`.  If not, it will be in `2.49.0` :-)

LeoOMaia:
I created the #12744 to include in 2.48.1
I had some trouble during the rebase, so i did another branch

PR #12736, opened 2025-07-02
endpoint metrics test: ignore order

valentijnscholten:
Since there's no explicit ordering in the queries, the order sometimes mismatches what the test expects.
Since there's no explicit ordering in the queries, we have to assume the order doesn't matter, so we compare while ignoring the order.


mtesauro (approved):
Approved

hblankenship (approved):

dogboat (approved):

PR #11401, opened 2024-12-10
Update JIRA for Finding Group When Risk Acceptance Expires

hblankenship:
[sc-9132]

When a risk exception expires, and reopens all of the associated findings, those changes should be reflected in jira.

PR #12737, opened 2025-07-02
risk acceptance expiration: keep link with findings

valentijnscholten:
Partially reverts #11401 to not remove findings from a Risk Acceptance when it expires.

The change to `process_resolution_from_jira` is just a code readability improvement.

The change to `view_eng.html` makes the number of findings cell in the Risk Acceptance table clickable.

The main reason for #11401 in january was to make sure expired risk acceptances were getting reflected in JIRA, so this PR adds some calls for that.

The PR also triggers a JIRA sync after a simple risk accept.

The PR also makes some changes to move the `save()` call out of the jira helper class.

[sc-11552]


dojo/jira_link/helper.py
```diff
@@ -141,6 +141,18 @@ def _safely_get_obj_status_for_jira(obj: Finding | Finding_Group, *, isenforced:
     return status or ["Inactive"]
 
 
+def is_keep_in_sync_with_jira(finding):
+    keep_in_sync_enabled = False
+    # Check if there is a jira issue that needs to be updated
+    jira_issue_exists = finding.has_jira_issue or (finding.finding_group and finding.finding_group.has_jira_issue)
+    # Only push if the finding is not in a group
```
hblankenship: Is this comment accurate? The test above suggests that if in finding group and the group has a jira issue then it will push?

valentijnscholten: I think that's a comment left over from your PR :-) But the code now works slightly different so I'll remove the comment as it is confusing indeed. Good find!

blakeaowens (approved):

PR #12738, opened 2025-07-03
Ruff: Preparation for TRY301

manuel-sommer:


dojo/tools/blackduck_component_risk/importer.py
```diff
@@ -42,10 +42,10 @@ def _process_zipfile(self, report: Path) -> (dict, dict, dict):
         """
         components = {}
         source = {}
-        try:
-            with zipfile.ZipFile(str(report)) as zipf:
-                c_file = False
-                s_file = False
+        with zipfile.ZipFile(str(report)) as zipf:
```
valentijnscholten: what happens if an exception occurs here as it will no longer be caught

Maffooch (approved):

dogboat (approved):

PR #12744, opened 2025-07-03
include vuln_id_from_tool in group_by

LeoOMaia:
based on #12684 and #12735 
In this PR we're adding grouping by the vuln_id_from_tool field, as suggested in the discussion, as a viable static option for the grouping dropdown.


valentijnscholten (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #12751, opened 2025-07-06
Add CVSS4 support

### LINKED ISSUE #12445: Feature request: CVSS v4 scores/vector support
Hi,

Currently, DefectDojo only "supports" CVSS Version 3.x. By "support," I mean that the fields in the Findings model and the integrated CVSS score calculator are specifically named and structured around CVSS v3. In the past, there have already been separate issues discussing support for different CVSS versions (e.g., #9378 and #12365). Now that CVSS 4.0 is gaining broader adoption (e.g., the EUVD uses it), this topic is becoming more relevant.  I think it would be practical to have a version-agnostic way of storing score and vector data, since this would allow users to freely choose which CVSS version to use without modifying the DefectDojo data model or having contradictory information in field names.

A previous, now-closed approach (see #9380) proposed adding new CVSS-specific fields to the Findings model. In my opinion, this kind of solution risks cluttering the model over time as more versions emerge.

**Proposed Solution**:
I propose a refactor of how CVSS data is stored in the Findings model. Instead of embedding the CVSS version into the field names, DefectDojo should generalize the field structure and store only the version agnostic core data:

- cvss_vector instead of cvssv3
- cvss_score instead of cvssv3_score

Additionally, I suggest introducing a new field: cvss_version, which would explicitly store the CVSS version used. Ideally, this field wouldn’t be necessary—since the version is usually part of the vector—but some data sources (e.g., OpenVAS CSV reports) do not include the vector, only a numeric score. In such cases, being able to specify the version used to generate the score could be useful (especially if there would be an "Unknown" version choice). Another benefit of having a separate cvss_version field is that it allows users to explicitly choose which CVSS version to use when manually adding a finding. However, one could argue that this could be handled purely as a UI feature, with the CVSS version inferred from the score prefix—since using the prefix is part of the specification. That said, this approach could introduce complications if the vector needs to be processed further.

**Advantages**:
Ideally, this approach enables the easy addition of future CVSS versions (simply add a new value to the allowed values like in the severity field), as long as the underlying paradigm of score and vector remains consistent. Furthermore, it allows users to freely choose a CVSS version while maintaining clean and consistent variable naming in both the UI and source code.

**Disadvantages**:
This change may break compatibility for applications relying on the current DefectDojo API. However, it should be feasible to map cvssv3_score and cvssv3 to the new generalized fields (cvss_score, cvss_vector) and provide a deprecation warning, with full removal in a future major release.

I’d would be happy to hear/read the thoughts of the community and maintainers on this approach. I’m willing to implement the change myself if it receives sufficient approval (I believe it may touch files currently affected by the feature freeze).

valentijnscholten:
Fixes #12445

- add `cvssv4` and `cvss4_score` fields to `Finding` model
- add score calculation/sanitization similar to what was done for `cvssv3` field
- update UI to show both vectors when present
- update `auditjs` parser + tests
- update `parse_cvss_data` to accomodate more use cases
- add links to external calculators to relevant forms
- update how to write a parser guide
- add system_settings to allow hiding of cvssv3 fields or cvssv4 fields

vectors

image

~QUESTIONS:~
~- Is it right to have the fields separated so we can support v3 anv v4 in parallel?~
~- Or should we converge this into `cvss`, `cvss_score` and `cvss_version` and prefer v4 over v3 if both are provided? This could be a problem for companies still tied to v3 in their compliance asessments?~
~- We could do both. Leave the PR as is, but also add these "merged" fields where we take the values from v4 if present else from v3 (or None otherwise)~
~- Do we want to embed a JS calculator again? The First/RedHat calculator is Vue based. The metaeffekt calculator is standalone JS:~
~- https://www.first.org/cvss/calculator/4-0~
~- https://www.metaeffekt.com/security/cvss/calculator/~


Maffooch:
>QUESTIONS:
>- Is it right to have the fields separated so we can support v3 anv v4 in parallel?
>- Or should we converge this into `cvss`, `cvss_score` and `cvss_version` and prefer v4 over v3 if both are provided? This could be a problem for companies still tied to v3 in their compliance asessments?
>- We could do both. Leave the PR as is, but also add these "merged" fields where we take the values from v4 if present else from v3 (or None otherwise)

I think it would best to keep the two fields separate for now. As you mentioned, there are likely some companies that are reliant on v3 for compliance purposes. We also want to avoid potentially breaking changes in existing automation.

> Add UI calculator

While I don't love the idea of embedding another big snippet of html on the view finding page, it would be consistent with the current behavior of v3. Being consistent would be ideal, but I am not sure it is worth the cost. Another option would be to link out to that universal calculator for both calculators. @mtesauro what do you think?

valentijnscholten:
I also considered linking out, but some companies might not like some external services being accessed. We could also link out to our own copy of that UI calculator hosted from within Defect Dojo. We could get rid of the old ugly one and just open the new one in a window. We could start with that and then see if there is a need for more integration like auto transfer for vectors between screens and whatnot. That would be the easiest for the Pro UI as well.

mtesauro:
> > QUESTIONS:
> > 
> > * Is it right to have the fields separated so we can support v3 anv v4 in parallel?
> > * Or should we converge this into `cvss`, `cvss_score` and `cvss_version` and prefer v4 over v3 if both are provided? This could be a problem for companies still tied to v3 in their compliance asessments?
> > * We could do both. Leave the PR as is, but also add these "merged" fields where we take the values from v4 if present else from v3 (or None otherwise)
> 
> I think it would best to keep the two fields separate for now. As you mentioned, there are likely some companies that are reliant on v3 for compliance purposes. We also want to avoid potentially breaking changes in existing automation.

I agree with this as well

> > Add UI calculator
> 
> While I don't love the idea of embedding another big snippet of html on the view finding page, it would be consistent with the current behavior of v3. Being consistent would be ideal, but I am not sure it is worth the cost. Another option would be to link out to that universal calculator for both calculators. @mtesauro what do you think?

It's probably best to include whatever we what displayed by default in DefectDojo in the source code aka shipped in our containers as we cannot rely on outbound connectivity for any default feature of DefectDojo. 
So, I see the options as:
1. Follow the same process as CVSSv3 and include the code for the calculator in our containers/source
2. Default to no CVSSv4 calculator and have people opt-in to having it call out to an external resource. This resource would default to First's calculator but allow people to update to a different URL should they choose or need to because of network restrictions, policy, whatever.

Other thoughts on the CVSS calculators:
* It makes sense to me to have both v3 and v4 be optionally displayed. As much we have plenty of System Settings, this makes sense to me especially as we start to deprecate v3. This is something we should do as we adopt v4.
* In an ideal world, both calculators would be overlays/pop-out UIs that don't busy up the edit finding UI any more then they currently are but that's more of an 'icing on the cake' and maybe we bake the v4 cake first. That nicer UI could be some follow-on work that someone in the community with good UI foo can tackle.

docs/content/en/open_source/contributing/how-to-write-a-parser.md
```diff
@@ -166,43 +166,52 @@ Good example:
...
+
+    cvss_vector = <get CVSS3 or CVSS4 vector from the report>
+    cvss_data = parse_cvss_data(cvss_vector)
+    if cvss_data:
+        finding.severity = cvss_data["severity"]
+        finding.cvssv3 = cvss_data["cvssv3"]
+        finding.cvssv4 = cvss_data["cvssv4"]
+        # we don't set any score fields as those will be overwritten by Defect Dojo
 ```
+Not all values have to be used as scan reports usuyall provide their own value for `severity`.
```
Maffooch:
```suggestion
Not all values have to be used as scan reports usually provide their own value for `severity`.
```

dojo/models.py
Maffooch, line 1: @blakeaowens for your awareness of new fields

dojo/forms.py
```diff
@@ -3740,3 +3815,30 @@ def set_permission(self, codename):
...
+    enable_cvss4 = get_system_setting("enable_cvss4_display", True)
+
+    # Hide CVSS3 fields if disabled
+    if not enable_cvss3:
+        if "cvssv3" in form_instance.fields:
+            del form_instance.fields["cvssv3"]
+        if "cvssv3_score" in form_instance.fields:
+            del form_instance.fields["cvssv3_score"]
+        if "cvss_info" in form_instance.fields:
+            del form_instance.fields["cvss_info"]
```
dogboat: Do we want to remove this here? It's removed below if both 3 and 4 are disabled, but we'd want it to show if just 3 is disabled right?

dojo/models.py
```diff
@@ -2712,20 +2733,32 @@ def save(self, dedupe_option=True, rules_option=True, product_grading_option=Tru
...
+                    self.cvssv3 = cvss_data.get("cvssv3")
+                    self.cvssv3_score = cvss_data.get("cvssv3_score")
 
             except Exception as ex:
                 logger.warning("Can't compute cvssv3 score for finding id %i. Invalid cvssv3 vector found: '%s'. Exception: %s.", self.id, self.cvssv3, ex)
                 # remove invalid cvssv3 vector for new findings, or should we just throw a ValidationError?
                 if self.pk is None:
                     self.cvssv3 = None
 
+        # behaviour for CVVS4 is slightly different. Extracting thsi into a method would lead to probabl hard to read code
```
dogboat: nit: typos in comment
```suggestion
        # behaviour for CVVS4 is slightly different. Extracting this into a method would lead to probably hard to read code
```

dojo/forms.py
```diff
@@ -1152,7 +1178,7 @@ class AddFindingForm(forms.ModelForm):
             "invalid_choice": EFFORT_FOR_FIXING_INVALID_CHOICE})
 
     # the only reliable way without hacking internal fields to get predicatble ordering is to make it explicit
-    field_order = ("title", "date", "cwe", "vulnerability_ids", "severity", "cvssv3", "description", "mitigation", "impact", "request", "response", "steps_to_reproduce",
+    field_order = ("title", "date", "cwe", "vulnerability_ids", "severity", "cvssv3", "cvssv3_Score", "cvssv4", "cvssv4_score", "description", "mitigation", "impact", "request", "response", "steps_to_reproduce",
```
dogboat: turbo nit: "Score" is capitalized in `cvssv3_Score`
```suggestion
    field_order = ("title", "date", "cwe", "vulnerability_ids", "severity", "cvssv3", "cvssv3_score", "cvssv4", "cvssv4_score", "description", "mitigation", "impact", "request", "response", "steps_to_reproduce",
```

unittests/test_metrics_queries.py
```diff
@@ -21,23 +21,23 @@ def add(*args, **kwargs):
...
...eview": False, "defect_review_requested_by_id": 1, "is_mitigated": False, "thread_id": 11, "mitigated": None, "mitigated_by_id": None, "reporter_id": 1, "numerical_severity": "S0", "last_reviewed": None, "last_reviewed_by_id": None, "param": None, "payload": None, "hash_code": "5d368a051fdec959e08315a32ef633ba5711bed6e8e75319ddee2cab4d4608c7", "line": None, "file_path": "", "component_name": None, "component_version": None, "static_finding": False, "dynamic_finding": False, "created": datetime(2017, 12, 1, 0, 0, tzinfo=UTC), "scanner_confidence": None, "sonarqube_issue_id": None, "unique_id_from_tool": None, "vuln_id_from_tool": None, "sast_source_object": None, "sast_sink_object": None, "sast_source_line": None, "sast_source_file_path": None, "nb_occurences": None, "publish_date": None, "service": None, "planned_remediation_date": None, "planned_remediation_version": None, "effort_for_fixing": None, "test__engagement__product__prod_type__member": False, "test__engagement__product__member": True, "test__engagement__product__prod_type__authorized_group": False, "test__engagement__product__authorized_group": False, "known_exploited": False, "ransomware_used": False, "kev_date": None}
```
dogboat, line 24: Hahahah sorry about this... 😥 Thanks for making it better...

dojo/forms.py
valentijnscholten:
```suggestion
```

valentijnscholten: Good catch Mr Boat. I could have sworn I tested this. Code updated!

dogboat (approved):
Awesome, thank you!

80998:
We need CVSSv4 support ASAP, my team has been putting our CVSSv4 vector into the CVSSv3 field which was broken with this new validation update. So now we cannot put in CVSSv4 scores into the fields now since we pulled the latest DD update. We will have to update again migrate all the scores when the CVSSv4 field DD update comes.

valentijnscholten:
It will be in the august 2.49.0 release on the first monday.

80998:
@valentijnscholten That's excellent, thank you very much 🙏

80998:
You guys have the fastest responses ever, thank you so much for that !

PR #12753, opened 2025-07-07
Bug Fix: improve Kiuwan SCA parser to support multi component findings

mwager:
This PR enhances the Kiuwan SCA Parser by modifying the logic to create one finding per component, instead of taking only the first component from the components array.

Motivation

In the current implementation, for a given CVE from a Kiuwan SCA scan, only the first component listed is used to create a finding. However, many CVEs in Kiuwan are related to multiple components. This leads to loss of detail and incomplete representation of risks in DefectDojo.

**Test results**

- Manual test with real-world Kiuwan JSON input shows that all findings are now preserved and properly associated with the correct component.

### DIFF
--- a/dojo/tools/kiuwan_sca/parser.py
+++ b/dojo/tools/kiuwan_sca/parser.py
@@ -1,7 +1,10 @@
 import hashlib
 import json
+import logging
 
 from dojo.models import Finding
+
+logger = logging.getLogger(__name__)
 
 __author__ = "mwager"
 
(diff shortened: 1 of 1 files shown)


mwager:
TODO: _Ideally you extend the test suite in `tests/` and `dojo/unittests` to cover the changed in this PR._

I will soon check how to update & run unit tests, any info/docs on it appreciated :)

valentijnscholten:
Thanks for looking into this, information on running unit tests: https://github.com/DefectDojo/django-DefectDojo/blob/master/readme-docs/DOCKER.md#run-the-tests-with-docker-compose

dojo/tools/kiuwan_sca/parser.py
```diff
@@ -37,50 +40,57 @@ def get_findings(self, filename, test):
...
-                dupes[key] = finding
+            components = row.get("components", [])
+            if not components:
+                logger.warning("WARNING: Insights Finding from kiuwan does not have a related component - Skipping.")
+                continue
+
+            # We want one unique finding in DD for each component affected:
+            for component in components:
+                finding = Finding(test=test)
+                finding.unique_id_from_tool = f"{row['cve']}|{component.get('artifact')}|{component.get('version')}"
```
Maffooch: Constructing the `unique_id_from_tool` is not something that is generally done. The intention is that the field is created from the tool. 

This parser in particular uses the `DEDUPE_ALGO_HASH_CODE`, so I think the `unique_id_from_tool` should be omitted. This could be an issue for folks that updated the default hash code fields to use `unique_id_from_tool`, so we should also include something in the release notes to indicate the change

Here is the current dedupe settings:
https://github.com/DefectDojo/django-DefectDojo/blob/d17c6dbbc5fa469119ca4206f7ec40963e0900c7/dojo/settings/settings.dist.py#L1336

mwager: Actually Kiuwan provides a unique id, this was just a test on our side. I will run one more test later this week and maybe change this back. This PR should be about the components issue only.

mwager: Pushed changes back to old logic

Maffooch:
@mwager how are things going here?

mwager:
@Maffooch Sorry, I forgot to update the tests. Done from my side now!

dojo/tools/kiuwan_sca/parser.py
```diff
@@ -37,50 +40,53 @@ def get_findings(self, filename, test):
...
-                    + "|"
-                    + str(finding.cwe)
-                ).encode("utf-8"),
-            ).hexdigest()
-
-            if key not in dupes:
-                dupes[key] = finding
+            components = row.get("components", [])
+            if not components:
+                logger.warning("WARNING: Insights Finding from kiuwan does not have a related component - Skipping.")
```
valentijnscholten: is this really a WARNING level type of message? It sounds like something that is common to occur in which case DEBUG is a better level?

dojo/tools/kiuwan_sca/parser.py
```diff
@@ -37,50 +40,53 @@ def get_findings(self, filename, test):
...
+
+            # We want one unique finding in DD for each component affected:
+            for component in components:
+                finding = Finding(test=test)
+                finding.unique_id_from_tool = str(row["id"])
+                finding.cve = row["cve"]
+                finding.description = row["description"]
+                finding.severity = self.SEVERITY[row["securityRisk"]]
+
+                finding.component_name = component.get("artifact", "Unknown")
```
valentijnscholten: the common approach across Defect Dojo I believe is to net set the field if it's not present.

dojo/tools/kiuwan_sca/parser.py
```diff
@@ -37,50 +40,53 @@ def get_findings(self, filename, test):
...
-                dupes[key] = finding
+            components = row.get("components", [])
+            if not components:
+                logger.warning("WARNING: Insights Finding from kiuwan does not have a related component - Skipping.")
+                continue
+
+            # We want one unique finding in DD for each component affected:
+            for component in components:
+                finding = Finding(test=test)
+                finding.unique_id_from_tool = str(row["id"])
```
valentijnscholten: Just checking to make sure this is really a unique id that is stable across multiple scans? Is there any documentation on the Kiuwan format(s) (couldn't easily find them)?

dojo/tools/kiuwan_sca/parser.py
mwager: Hey @valentijnscholten IMO it is better a warning because a supply chain finding without component info wouldnt make much sense, no? But If you prefer I can change it back.

dojo/tools/kiuwan_sca/parser.py
mwager: @valentijnscholten It was already using this field before this PR - I just set it to be a string now. We are using it for long time and there is always a unique id coming from kiuwan API. If it is the same finding it is the same id again yes.

See here: https://static.kiuwan.com/rest-api/kiuwan-rest-api.html

grafik

dojo/tools/kiuwan_sca/parser.py
mwager: @valentijnscholten  updated

mwager:
@Maffooch @valentijnscholten Does it look good to you? :)

dojo/tools/kiuwan_sca/parser.py
valentijnscholten: I think WARNING is meant to be for sysadmins doing log analysis/monitoring so I prefer debug.

dojo/tools/kiuwan_sca/parser.py
valentijnscholten: ok

dojo/tools/kiuwan_sca/parser.py
mwager: @valentijnscholten  ok updated!

mwager:
@valentijnscholten should be good to go now?

cc @shodanwashere

valentijnscholten:
We've realized that the PR will result in violations of the (logical) "unique" constraint for the `unique_id_from_tool` field. We're discussing how we could handle this scenario now and in the future. Possible compromise for this parser could be to store the row id in the `vuln_id_from_tool` field? The `unique_id_from_tool` is not used in dedupe currently for this parser.

mwager:
@shodanwashere Could you have a look and discuss this with the team? Thx!

shodanwashere:
@valentijnscholten true, since two components could have the exact same vuln, the IDs would no longer be unique. does the `unique_id_from_tool` have some constraint preventing it from being null? otherwise, we could go with having the id on the `vuln_id_from_tool` field.

valentijnscholten:
@shodanwashere Using the `vuln_id_from_tool` field is acceptable in this case.

mwager:
@valentijnscholten does it look good now? 

cc @shodanwashere

shodanwashere:
i'll assume that's what we're intending to do. @valentijnscholten will it work like how @mwager has implemented?

valentijnscholten (approved):

mtesauro (approved):
Approved

Maffooch (approved):

Jino-T (approved):

PR #12785, opened 2025-07-15
Make KEV data visible on findings listing

dogboat:
This PR displays KEV-related data (known exploited/used in ransomware/kev added date) stored on the Finding model to the Finding listing pages (all/open/closed/risk accepted). It also adds basic filtering against those fields.

Listing page view:
Screenshot 2025-07-15 at 1 17 32 PM

Filter view:
Screenshot 2025-07-15 at 1 07 35 PM

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -1797,6 +1797,7 @@ class FindingFilterHelper(FilterSet):
         self.form.fields["on"].widget = date_input_widget
         self.form.fields["before"].widget = date_input_widget
         self.form.fields["after"].widget = date_input_widget
+        self.form.fields["kev_date"].widget = date_input_widget
         self.form.fields["mitigated_on"].widget = date_input_widget
         self.form.fields["mitigated_before"].widget = date_input_widget
         self.form.fields["mitigated_after"].widget = date_input_widget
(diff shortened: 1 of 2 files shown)


dogboat:
[sc-11506]

valentijnscholten (commented):
Is it sufficient to only be able to filter by the exact KEV date? I can imagine users want to do a weekly search to find out "which findings where added to KEV since last week"? 
I also realize adding before/after fields may clutter the UI more.

dogboat:
> Is it sufficient to only be able to filter by the exact KEV date? I can imagine users want to do a weekly search to find out "which findings where added to KEV since last week"? I also realize adding before/after fields may clutter the UI more.

Yeah good point! I've added before/after filters for it. Does add a bit more to the UI, but it's already got a lot there so maybe it's not too much? Let me know what you think, and thanks for pointing this out, I think these are very useful.

Screenshot 2025-07-16 at 10 03 38 AM

Maffooch (approved):

valentijnscholten (approved):

80998:
Yes very good idea, I want to know if they are on the KEV also, excellent update coming ☝️🥹

blakeaowens (approved):

mtesauro (approved):
Approved

PR #12788, opened 2025-07-15
reuse Sarif base parser for snyk and mayhem parsers

valentijnscholten:
Over time two Sarif based parsers were added because they needed some custom parser logic to fully parse their scan reports:

- Mayhem
- Snyk Code

This resulted in some code duplication. This PR changes the Sarif base parser to be extensible and uses this to simplify the Snyk Code and Mayhem parsers. I didn't go as far as to create extension hooks for each an every field. This PR now has some examples that can be used in future PRs if more fields need to be extendible.

Initially I had the CWE parsing logic for the Mayhem `taxa` based CWEs also in an extenstion hook. But then I found out that the "taxa" format is actually somewhat official Sarif formatting. Also there was already code in the base Sarif parser looking for CWEs in various places. So the PR adds the "taxa way" to the base Sarif parser. This way it's also beneficial for any other non-Mayem report that might use this format.

Additional things in this PR:
- Reduce code duplication between the `snyk` parser and `snyk_code` parser.
- Enhance test cases for the `snyk_code` parser

[sc-11544]

### DIFF
--- a/docs/content/en/connecting_your_tools/parsers/file/snyk_code.md
+++ b/docs/content/en/connecting_your_tools/parsers/file/snyk_code.md
@@ -2,8 +2,7 @@
 title: "Snyk Code"
 toc_hide: true
 ---
-Snyk output file (snyk test \--json \> snyk.json) can be imported in
-JSON format. Only SCA (Software Composition Analysis) report is supported (SAST report not supported yet).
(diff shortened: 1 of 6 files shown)


paulOsinski:
looks good to me / does what I need it to, anyway!

dogboat (approved):

blakeaowens (approved):

PR #12792, opened 2025-07-16
pytz removal

valentijnscholten:
`Django` 5.x no longer supports `pytz`, so we need to remove it.

Looking at various articles and information it looks like a fairly simple replacement with `zoneinfo`: 

- https://engineering.backmarket.com/moving-forward-in-pythons-time-and-time-zones-aa3ccbd4569a
- https://pytz-deprecation-shim.readthedocs.io/en/latest/index.html

These do mention various corner cases but I don't think we're affected by those.

I added this to 2.50.0 as 2.49.0 might be too soon as probable some testing needs to be done also with Pro Cloud instances and the extra/new metrics in Pro?

This replaces #12101


blakeaowens (approved):

mtesauro (approved):
Approved

Issue #12633, opened 2025-06-19
Add "has fix" filter for CVE finding


tienmtran:
## :warning: Note on feature completeness :warning:

We are narrowing the scope of acceptable enhancements to DefectDojo. Learn more here:
https://github.com/DefectDojo/django-DefectDojo/blob/master/readme-docs/CONTRIBUTING.md

**Is your feature request related to a problem? Please describe**
No

**Describe the solution you'd like**
I would like to have the "has fix" in filter or parser to quick check for which CVEs fix available. Some CVE scanner allow to ignore cve which dont have fix yet in their output result but I want to have full visibility of that. I prefer DD can filter out for that case.

**Additional context**
I use trivy to scan container and upload it to DD

valentijnscholten:
Do you have more examples of which parsers/scanners provide a `fixable` flag? I have found:
- Trivy
- Aqua

tienmtran:
I see grype, blackduck, jfrog,... also.
It's not really `fixable` flag in input. But when parsing for `mitigation` field this could be none
So if we can do at filter level or at parser level to add a tag for the fixable information

manuel-sommer:
Also Harbor has a fixable information:
https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/scans/harbor_vulnerability/harbor-5-vuln.json 
Maybe also RedHatSatellite in the "installable" field. 
I could help make a PR if this feature would be merged. What are your thoughts @valentijnscholten ?

tienmtran:
Hi, any update? Would we able to add this into Dojo's feature?

manuel-sommer:
Do we want to add this feature @valentijnscholten and @mtesauro ?

Kasyap7:
I am using Anchore Engine scan reports, and it would be helpful to include a has_fix tag in these reports as well.

- If the Mitigation field specifies an actionable fix (e.g., “Upgrade to package xyz”), the has_fix tag should be added.
- If the Mitigation field states “Upgrade to package none” or simply “None,” the has_fix tag should not be included, as no fix is available.

This enhancement would make it easier to filter findings based on fix availability.

manuel-sommer:
How to proceed here? Do we want to add an "update findings functionality" in the reimport or add fix_available to the deduplication setting?

valentijnscholten:
> Do we want to add an "update findings functionality" in the reimport

This.

manuel-sommer:
[...]

PR #12793, opened 2025-07-16
:tada: Add a 'fix_available' field to the findings #12633

manuel-sommer:
see #12633


blakeaowens (approved):

manuel-sommer:
Just a question: Do I also have to make a migrations to add a default value to all findings?

dogboat (approved):

mtesauro (approved):
Approved

valentijnscholten:
@manuel-sommer In my experience this field can have three values "Yes", "No" or "Unknown" (or "NotSet" or mabe we could use "None" or "Null" for this. I see that the default has been set to `True`, which I don't think is good assumption in some cases.  This is a Django default used when records are inserted, but it's not something the database is aware of.

mtesauro:
> @manuel-sommer In my experience this field can have three values "Yes", "No" or "Unknown" (or "NotSet" or mabe we could use "None" or "Null" for this. I see that the default has been set to `True`, which I don't think is good assumption in some cases. This is a Django default used when records are inserted, but it's not something the database is aware of.

Oh interesting point.
I was thinking of it strictly in a binary fashion e.g. If has a fix then true, otherwise false.

I do see your point that the availability of a fix may be unknown especially when using something from a 3rd party where "unknown" would have some definite value.

manuel-sommer:
How to proceed here @mtesauro and @valentijnscholten ?

- Do we need another db migration and if yes, do we want to set all findings to "true" or "false"?
- And: Maybe we should include the fix_available fieldto be choosen as a deduplication option for parsers.

Maffooch:
> Do we need another db migration and if yes, do we want to set all findings to "true" or "false"?

Yeah it wouldn't hurt to have another migration to allow for the field to be unset, with `None` being the default.

> And: Maybe we should include the fix_available fieldto be choosen as a deduplication option for parsers.

For parser that could report this kind of data, I think dedupe could possibly make a mess of things. For example, any change in the value of `fix_available` during an import would cause the original finding to be closed for reimport, and the new on to be set to a duplicate for import. Seems like it would be best to leave this out of dedupe

manuel-sommer:
> > Do we need another db migration and if yes, do we want to set all findings to "true" or "false"?
>
> Yeah it wouldn't hurt to have another migration to allow for the field to be unset, with `None` being the default.

The default value after each import is `True` as the value is a boolean. 

> > And: Maybe we should include the fix_available fieldto be choosen as a deduplication option for parsers.
>
> For parser that could report this kind of data, I think dedupe could possibly make a mess of things. For example, any change in the value of `fix_available` during an import would cause the original finding to be closed for reimport, and the new on to be set to a duplicate for import. Seems like it would be best to leave this out of dedupe

But isn't this the behavior that we would like to support? I personally would really welcome this. The default value is "True", so most parsers (e.g. SAST tools etc.) aren't affected. A team can focus on findings that are fixable and a finding which is not fixable is closed automatically and a new finding will appear with new SLAs once the fix is available. This helps teams in a way that they don't have to analyse findings if they do now support a fix or not. Or is there a way to update an existing finding once a fix becomes available? If no, please let's support this setting, don't change the standard behavior, but make it available through https://docs.defectdojo.com/en/open_source/installation/configuration/#local_settingspy

valentijnscholten:
I don't think we should allow users to add the `fix_available` field to the hash code config. I don't think many people fully understand the implications and can't imagine many use cases where you would want a new finding to be created just because a fix has become available. I do think reimport should update this field if it changes in value across scans. What about the deduplication process? Let's say it finds a duplicate which has a fix available but the existing original finding doesn't?

Regarding the default I'm still torn between "let's leave it at True" by default as that's the most common scenario. On the other hand there might be many findings by parsers that do not support the `fix_available` field where `True` might be incorrect as default. For example TLS protocol version or cipher suites not supported by (virtual) appliances. Packages that gone end-of-life or are abandoned. I wonder if we should make the effort to reflect the real world and make the field an enum that can be "Yes", "No",  "Unspecified". Or that we should make the effort to see if we can "abuse" the `None` value for a Boolean field. But that feels like a recipe for some kind of error/mismatch/regression in the future.

manuel-sommer:
> I don't think we should allow users to add the `fix_available` field to the hash code config. I don't think many people fully understand the implications and can't imagine many use cases where you would want a new finding to be created just because a fix has become available. I do think reimport should update this field if it changes in value across scans. What about the deduplication process? Let's say it finds a duplicate which has a fix available but the existing original finding doesn't?
> 
I did not look at the code, but as far as I know, this logic within the reimport is not yet supported, right?  e.g. look at https://github.com/DefectDojo/django-DefectDojo/issues/6774

> Regarding the default I'm still torn between "let's leave it at True" by default as that's the most common scenario. On the other hand there might be many findings by parsers that do not support the `fix_available` field where `True` might be incorrect as default. For example TLS protocol version or cipher suites not supported by (virtual) appliances. Packages that gone end-of-life or are abandoned. I wonder if we should make the effort to reflect the real world and make the field an enum that can be "Yes", "No", "Unspecified". Or that we should make the effort to see if we can "abuse" the `None` value for a Boolean field. But that feels like a recipe for some kind of error/mismatch/regression in the future.

I made a PR here, please review: https://github.com/DefectDojo/django-DefectDojo/pull/12817

PR #12797, opened 2025-07-16
feat: Added cve_list field on finding model

BernnardoSBO:
## Description
This pull request is the first step toward implementing the new feature discussed [here](https://github.com/DefectDojo/django-DefectDojo/discussions/11796)  . It introduces a new field for a CVE list, which establishes a more accurate many-to-many relationship between the Finding model and CVEs. This field leverages the Tagulous library, which simplifies the implementation by abstracting away the need to create a dedicated model for CVEs. This abstraction is precisely the 'syntactic sugar' that streamlines the management and filtering of these tags.

Looking ahead, this new field will enable future enhancements regarding the cross-referencing of CVEs and their associated metadata. Furthermore, this PR lays the groundwork for subsequent changes we plan to propose.

To ensure the reliability of this implementation, comprehensive unit tests have been added to cover the new functionality. 

## What was done

- Modified openvas xml parser to use new cve_list field
- Modified openvas xml parser to use the mitigation field
- Added filtering support over new field
- Modified finding UI to include the cve_list
- Performed the migration needed

## Test Results
- Unit tests implemented over new cve_list field in Finding's model

## Documentation
- Given that the enhancement comprises only a field addition to the Finding model, no documentation was needed


Maffooch:
Hi @BernnardoSBO, this feature already exists through the [`Vulnerability_ID` field](https://github.com/DefectDojo/django-DefectDojo/blob/d17c6dbbc5fa469119ca4206f7ec40963e0900c7/dojo/models.py#L3414-L3423) to handle all sorts of [security advisory types](https://github.com/DefectDojo/django-DefectDojo/blob/d17c6dbbc5fa469119ca4206f7ec40963e0900c7/dojo/settings/settings.dist.py#L1809-L1870)

PR #12803, opened 2025-07-18
:bug: rebase dev branch db_migrations

manuel-sommer:
Hi @mtesauro, @valentijnscholten and @Maffooch 
could you please merge this fast as this is a hotfix for the dev branch?

PR #12807, opened 2025-07-18
static group view

LeoOMaia:
I developed a new tab so that people can better see the information of the `Finding_Group` that is created during the scan import. This is acording to #12684.
As a next step we plan to allow include an option to dynamically create and visualize groups by properties of `Finding`s.

Bellow the image of Findings tab:
Screenshot from 2025-07-18 12-40-52
And the image of the reduced result with Finding_Group tab:
Screenshot from 2025-07-18 12-37-30

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -1999,6 +1999,41 @@ class FindingFilter(FindingFilterHelper, FindingTagFilter):
         self.form.fields["reviewers"].queryset = self.form.fields["reporter"].queryset
 
 
+class FindingGroupsFilter(FilterSet):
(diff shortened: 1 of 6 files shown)


LeoOMaia:
Hey guys. looks like in the dev branch in the last 2 commits was created 2 entries with the same number in the db migrations.

mtesauro:
> Hey guys. looks like in the dev branch in the last 2 commits was created 2 entries with the same number in the db migrations.

That's being worked in #12803 so give us a bit to get that sorted out

LeoOMaia:
Ok, when it's resolved, I'll resubmit the PR with new migrations.

PR #12811, opened 2025-07-20
importers: clean tags before saving

valentijnscholten:
Some parsers such as the `sysdig` parsers were inserting spaces into `tags`. This violates the validation rules introduced recently.

This PR:
- Changes Sysdig parser to not insert spaces
- Changes `import` and `reimport` to replace invalid characters in tags with '`_`'  using a new `clean_tags` function
- Changes Sysdig parser to use the `clean_tags` function (as an example)

In general we want to improve data validation for import/reimport which we'll be looking at next. For now we wanted to do a specific fix for the sysdig parsers and general tags usage. But we didn't want to completely block on any invalid characters in tags as they might appear in place where the user has little or no control. 

For other fields there is the same risk. So validating all fields and raising errors or skipping invalid findings is for a next PR in a monthly release where we can test various cases and provide informtion in the release notes/upgrade notes.


mtesauro (approved):
Approved

Maffooch (approved):

rossops (approved):

Jino-T (approved):

PR #12814, opened 2025-07-20
static group view

LeoOMaia:
I developed a new tab so that people can better see the information of the Finding_Group that is created during the scan import. This is acording to #12684.
As a next step we plan to allow include an option to dynamically create and visualize groups by properties of Findings.

Bellow the image of Findings tab:
Screenshot from 2025-07-18 12-40-52
And the image of the reduced result with Finding_Group tab:
Screenshot from 2025-07-18 12-37-30

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -1999,6 +1999,41 @@ class FindingFilter(FindingFilterHelper, FindingTagFilter):
         self.form.fields["reviewers"].queryset = self.form.fields["reporter"].queryset
 
 
+class FindingGroupsFilter(FilterSet):
+    name = CharFilter(method="filter_name", label="Name")
+    severity = ChoiceFilter(
+        choices=[
+            ("Low", "Low"),
+            ("Medium", "Medium"),
+            ("High", "High"),
+            ("Critical", "Critical"),
+        ],
+        method="filter_min_severity",
+        label="Min Severity",
+    )
+    engagement = ModelMultipleChoiceFilter(queryset=Engagement.objects.none(), label="Engagement")
+    product = ModelMultipleChoiceFilter(queryset=Product.objects.none(), label="Product")
+
+    class Meta:
+        model = Finding
+        fields = ["name", "severity", "engagement", "product"]
+
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop("user", None)
+        self.pid = kwargs.pop("pid", None)
+        super().__init__(*args, **kwargs)
+        self.set_related_object_fields()
+
+    def set_related_object_fields(self):
+        if self.pid is not None:
+            self.form.fields["engagement"].queryset = Engagement.objects.filter(product_id=self.pid)
+            if "product" in self.form.fields:
+                del self.form.fields["product"]
+        else:
+            self.form.fields["product"].queryset = get_authorized_products(Permissions.Product_View)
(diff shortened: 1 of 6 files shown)


dojo/finding_group/views.py
```diff
@@ -204,3 +212,118 @@ def push_to_jira(request, fgid):
...
+        page_size = int(request.GET.get("page_size", 25))
+        paginator = Paginator(queryset, page_size)
+        page_number = request.GET.get("page")
+        return paginator.get_page(page_number)
+
+    def get(self, request: HttpRequest) -> HttpResponse:
+        global_role = Global_Role.objects.filter(user=request.user).first()
+        user_groups = Dojo_Group.objects.filter(users=request.user)
+        products = Product.objects.filter(Q(members=request.user) | Q(authorization_groups__in=user_groups)).distinct()
+        if request.user.is_superuser or (global_role and global_role.role):
```
valentijnscholten: I would expect some calls to `get_authorized_products` or something similar here (i.e. not building a queryfilter yourself)

valentijnscholten: or maybe there should be an `get_authorized_finding_groups` method added somewhere to abstract his logic away.

dojo/filters.py
```diff
@@ -2016,6 +2016,41 @@ def set_related_object_fields(self, *args: list, **kwargs: dict):
         self.form.fields["reviewers"].queryset = self.form.fields["reporter"].queryset
 
 
+class FindingGroupsFilter(FilterSet):
+    name = CharFilter(method="filter_name", label="Name")
```
valentijnscholten: should this be some case insensitive like `name = CharFilter(lookup_expr="icontains", label="Name")`

valentijnscholten (commented):
I've placed some comments. Let me also check if the rest of the team also thinks this is a useful addition.

LeoOMaia:
Okay, but I think it's better to just leave the title "Finding Group" page. I made it similar to the "Finding" page with the All/Open/Closed options.

LeoOMaia:
One thing I noticed while implementing the code, which I think is worth mentioning: when we perform two scan imports, each with one finding that would be grouped into the same Finding_Group, DefectDojo is creating two groups with the same name instead of just including the finding from the second scan in the first group created, doubling the space occupied in the database. Is it supposed to be like this?

valentijnscholten:
Converted to draft while waiting for completion.

dojo/finding_group/views.py
LeoOMaia: I changed to use the `get_authorized_products`

dojo/filters.py
LeoOMaia: yeah, swapped to use lookup_expr and in addition removed the `method` in the variable bellow name

valentijnscholten:
> One thing I noticed while implementing the code, which I think is worth mentioning: when we perform two scan imports, each with one finding that would be grouped into the same Finding_Group, DefectDojo is creating two groups with the same name instead of just including the finding from the second scan in the first group created, doubling the space occupied in the database. Is it supposed to be like this?

Finding_Groups are currently bound to a single test. So if you have two separate imports and separate Tests, there will be two finding groups created.

valentijnscholten (approved):

mtesauro (approved):
Approved

PR #12817, opened 2025-07-21
:bug: Fix db_migration fix_available field to allow None

manuel-sommer:
[#12793](https://github.com/DefectDojo/django-DefectDojo/pull/12793#issuecomment-3095418494)

### DIFF
--- a/dojo/db_migrations/0238_finding_fix_available.py
+++ b/dojo/db_migrations/0238_finding_fix_available.py
@@ -13,6 +13,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='finding',
             name='fix_available',
-            field=models.BooleanField(default=True, help_text='Denotes if there is a fix available for this flaw.', verbose_name='Fix Available'),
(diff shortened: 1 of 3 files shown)


manuel-sommer:
@valentijnscholten fyi

manuel-sommer:
@Maffooch and @mtesauro: Do we want to change this migration according to the comments in https://github.com/DefectDojo/django-DefectDojo/pull/12793? If yes, we should merge it before 2.49.0

Maffooch (approved):
I think this would be the best way to go for now - let tools set the value if the tool provides that info. Otherwise, it can be managed without some heavy assumptions

manuel-sommer:
So, do I understand this right @Maffooch - we want to add this field also to deduplication, to let the tool set the info in order to provide updates if a finding changes?

mtesauro (approved):
Approved

valentijnscholten (approved):
Let's do the filters in a follow up so that we can merge this to get the migrations in the right place for mondays release.

manuel-sommer:
I made the filter already in this [PR](https://github.com/DefectDojo/django-DefectDojo/pull/12793) as far as I understand you right.

PR #12826, opened 2025-07-21
Trivy: Use CVSS scores from other vendors where applicable

Maffooch:
Trivy scans reports the source of where the severity field comes from. We use this field to associate a CVSS score with the finding. In cases where the severity source does not have a CVSS vector, it would be beneficial to still include a vector from the first match in the following list:
 - Detected Severity Source
 - NVD
 - GHSA
 - Redhat
 - Bitnami

In the sample scan files, there was a CVSS score that did not match the CVSS vector, so I updated it

### DIFF
--- a/dojo/tools/trivy/parser.py
+++ b/dojo/tools/trivy/parser.py
@@ -17,6 +17,13 @@ TRIVY_SEVERITIES = {
     "UNKNOWN": "Info",
 }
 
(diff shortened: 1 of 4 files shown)


blakeaowens (approved):

mtesauro (approved):
Approved

valentijnscholten (approved):

PR #12827, opened 2025-07-21
product api: optimize list of finding ids

valentijnscholten:
The `products` API endpoint returns the list of ids of all active findings for each product.

This is slow in larger instances or large page sizes.

This PR optimizes the query a little bit. For a 100K finding instance the response time went from ~6s to ~0.9s. Of course the mileage may vary depending on available resources and dataset specifics.

I tried fursther optimizing is using prefetching. But prefetching across multiple relationship levels in Django ORM isn't possible in one ago and requires some aggregation/counting in python making it slower again.

[sc-11611]


blakeaowens (approved):

dogboat (approved):

mtesauro (approved):
Approved

PR #12828, opened 2025-07-21
Dependency Check: Support CVSS v3

Maffooch:
Dependency Check includes some CVSS v3 fields to construct the vector manually. We should attempt to use it where we can


blakeaowens (approved):

mtesauro (approved):
Approved

valentijnscholten (approved):

PR #12842, opened 2025-07-23
close old findings: make test cases test default behaviour

valentijnscholten:
As a follow up to #12837 we update the import/reimport testcases.

There are/were a lot of tests cases, but these were all settings values for `close_old_findings` explicitly. This PR changes the test cases to rely on the defaults used in the API and only set the value when we do not want to use/test the default value.
There's two additional test cases added as well.

For the UI tests in `test_import_reimport.py` we cannot apply this mechanism as leaving out a checkbox on an html form will result in an interpretation as `False` in Django (and other html/http framework).

The tests in `test_importers_closeold.py` are not using the API so we cannot apply the same principle here.


Maffooch (approved):

mtesauro (approved):
Approved

PR #12848, opened 2025-07-23
semgrep pro parser

valentijnscholten:
Adds support for Semgrep Pro reports to be parsed.

The documentation on Semgrep Pro is not fully clear on what kind of different reports there are. And we only have limited samples. This PR implements parsing [issues ](https://semgrep.dev/api/v1/docs/#tag/Finding/operation/semgrep_app.core_exp.findings.handlers.issue.openapi_list_recent_issues). These can be SCA and Code findings. The sample contains Code findings, so those are tested by the unit tests. According to the docs the SCA issues should look a lot similar and should work as well.

Please let us know if you have any sample data that can help us improve Semgrep Pro parsing in future PRs.

[sc-11639]


mtesauro (approved):
Approved

Maffooch (approved):

paulOsinski (approved):

PR #12862, opened 2025-07-27
perf2: skip post processing if not needed

valentijnscholten:
There'll be a sequence of import performance related PRs. They are separated into small PRs to simplify review and troubleshooting. As they all use `test_importers_performance`, they'll build on eachother. So they'll have to be merged by natural ordering.

The `finding.save()` method creates an async task to postprocess the finding even if it's not needed. This is mainly of importance if product grading is disabled. In that case post processing is not needed, so we shouldn't create a celery task.
The celery task would exit quite fast, but still it would incur cpu cycles, redis cycles and at least 1 database query.
Let's save some trees and possibly some cost.


mtesauro (approved):
Approved

Maffooch (approved):

dogboat (approved):

blakeaowens (approved):

PR #12863, opened 2025-07-27
perf3: importers should not save findings again

valentijnscholten:
At the end of the import or reimport process there was an call to `finding.save()` mainly to trigger post processing such as dedupe, pushing to JIRA and product grading.
But at that point all finding data was already saved in earlier steps of the import process. So instead of doing a save, we call the post processing directly. This saves a database UPDATE statement which is expensive and can trigger further processing/queries such as audit logging. Some may argue the code becomes less clear or harder to understand. Since this is the primary process of Defect Dojo and the main performance concern it is an acceptable downside.

### DIFF
--- a/dojo/importers/default_importer.py
+++ b/dojo/importers/default_importer.py
@@ -198,7 +198,9 @@ class DefaultImporter(BaseImporter, DefaultImporterOptions):
 
             # Force parsers to use unsaved_tags (stored in below after saving)
             unsaved_finding.tags = None
-            unsaved_finding.save(dedupe_option=False)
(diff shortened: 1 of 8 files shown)


valentijnscholten:
This looked like a promising optimization possibility. However hash_code generation logic is depending on related objects like endpoints and vulnerability ids being present. This means there's not really a way around have two calls to save.
The PR here tried to opimize it to only set UPDATE the hash_code field, but still lots of tests are failing. 
So for now we have to leave things as is as it seems to become too complicated.

PR #12867, opened 2025-07-28
Async Delete: Handle exceptions for duplicate requests

Maffooch:
There are cases of async delete where the same resource is deleted by separate requests. For example, deleting an engagement followed by that engagement's product will produce exceptions in the product tasks where the engagement does not exist. In this case, we know the object is already deleted, so we can simply ignore the error

### DIFF
--- a/dojo/test/signals.py
+++ b/dojo/test/signals.py
@@ -8,30 +8,32 @@ from django.dispatch import receiver
 from django.urls import reverse
 from django.utils.translation import gettext as _
 
-from dojo.models import Finding, Test
+from dojo.models import Finding, Test, Engagement, Product
 from dojo.notes.helper import delete_related_notes
 from dojo.notifications.helper import create_notification
 
 
 @receiver(post_delete, sender=Test)
 def test_post_delete(sender, instance, using, origin, **kwargs):
-    if instance == origin:
-        description = _('The test "%(name)s" was deleted') % {"name": str(instance)}
-        if settings.ENABLE_AUDITLOG:
-            if le := LogEntry.objects.filter(
-                action=LogEntry.Action.DELETE,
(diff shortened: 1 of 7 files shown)


dojo/utils.py
```diff
@@ -2367,7 +2367,8 @@ def crawl(self, obj, model_list, **kwargs):
             model = model_info[0]
             model_query = model_info[1]
             filter_dict = {model_query: obj}
-            objects_to_delete = model.objects.filter(**filter_dict)
+            # Oly fetch the IDs since we will make a list of IDs in the following function call
```
blakeaowens:
```suggestion
            # Only fetch the IDs since we will make a list of IDs in the following function call
```

dojo/test/signals.py
```diff
@@ -8,30 +8,32 @@
...
+            description = _('The test "%(name)s" was deleted') % {"name": str(instance)}
+            if settings.ENABLE_AUDITLOG:
+                if le := LogEntry.objects.filter(
+                    action=LogEntry.Action.DELETE,
+                    content_type=ContentType.objects.get(app_label="dojo", model="test"),
+                    object_id=instance.id,
+                ).order_by("-id").first():
+                    description = _('The test "%(name)s" was deleted by %(user)s') % {
+                                    "name": str(instance), "user": le.actor}
+            create_notification(event="test_deleted",  # template does not exists, it will default to "other" but this event name needs to stay because of unit testing
```
blakeaowens:
```suggestion
            create_notification(event="test_deleted",  # Template does not exist, it will default to "other" but this event name needs to stay because of unit testing
```

blakeaowens (changes requested):

valentijnscholten (approved):

dogboat (approved):

blakeaowens (approved):

PR #12891, opened 2025-07-31
Support for whitelisted file extensions

Maffooch:
It is possible to import/reimport any file extension. The UI already enforces a handful of extensions that are allowed, but the API does not validate in the same way 

https://github.com/DefectDojo/django-DefectDojo/blob/2748b8d05450800a36b9282f5bd917bf6e673046/dojo/forms.py#L649-L653

A validation error will look like this:
```json
"file": [
    "File extension “exe” is not allowed. Allowed extensions are: xml, csv, nessus, json, jsonl, html, js, zip, xlsx, txt, sarif."
  ],
```


dogboat (approved):

blakeaowens (approved):

mtesauro (approved):
Approved

rossops (approved):

PR #12893, opened 2025-08-01
fix datatable search box location with style override

blakeaowens:
[sc-11702]

This pull requests fixes an issue introduced to the DataTable search functionality that placed the table search `div` on a new line (styled to `display: block;` rather than `display: inline-block;`).

Before:
Screenshot 2025-08-01 at 12 33 28 AM

After:
Screenshot 2025-08-01 at 12 33 13 AM

### DIFF
--- a/dojo/static/dojo/css/dojo.css
+++ b/dojo/static/dojo/css/dojo.css
@@ -1439,6 +1439,10 @@ div.dt-buttons {
     margin: 10px!important;
 }
 
+div.dt-search {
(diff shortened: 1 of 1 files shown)


valentijnscholten (approved):
Funny I tried this change in my browser dev tools and the box wouldn't movechange. Probably PEBKAC ;-)

dogboat (approved):

mtesauro (approved):
Approved

PR #12898, opened 2025-08-01
Import performance 3a: reduce data in `finding.save()` call

valentijnscholten:
This is the third performance PR in a series of small improvements.

The import process is always doing two `finding.save()` calls. These are expensive as they result in a database UPDATE statement. But they also trigger various post processing methods and signals result in more database queries, for example by `django-auditlog`.

I've created two PRs that can help here, we need to choose which wone we want to merge:

- Option 3A (this PR). Keep the two `save()` calls but make the second one much smaller by only updating the fields that have been changed (`hash_code`). To achieve this we have to separate the processing of the `cve` field from the other vulnerability ids. (Remember the `cve` field will be deprecated and removed at some point in the future).

- Option 3B: Calculate hash_code and CVE before the first `save()` call. After that we don't need another save because all other processing only sets related fields like vulnerability ids. (Remember the `cve` field will be deprecated and removed at some point in the future). This logic uses the exact same code as `reimport` is using to calculate the hash_code, so this is proven logic.

I prefer option 3B (the other PR).

### DIFF
--- a/dojo/templatetags/multiply.py
+++ b/dojo/templatetags/multiply.py
@@ -4,5 +4,5 @@ register = template.Library()
 
 
 @register.filter
-def multiply(value, arg):
+def multiply(value, arg):  # noqa: FURB118
     return value * arg
--- a/dojo/templatetags/multiply.py
+++ b/dojo/templatetags/multiply.py
@@ -4,5 +4,5 @@ register = template.Library()
 
 
 @register.filter
-def multiply(value, arg):  # noqa: FURB118
+def multiply(value, arg):
     return value * arg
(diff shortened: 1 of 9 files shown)


dojo/templatetags/multiply.py
```diff
@@ -4,5 +4,5 @@
 
 
 @register.filter
-def multiply(value, arg):
+def multiply(value, arg):  # noqa: FURB118
```
Maffooch:
```suggestion
def multiply(value, arg):
```

Maffooch:
Closing in favor of #12900

PR #12901, opened 2025-08-01
performance do_not_reactivate: adding a note doesn't need a finding save

valentijnscholten:
When adding a note to a finding to state that it was not reactivated, we don't need to call save on the finding itself.
The same approach was taken in the past for close_old_findings.


Maffooch (approved):

blakeaowens (approved):

mtesauro (approved):
Approved

PR #12903, opened 2025-08-02
new snyk_issue_api parser for `code` issues

valentijnscholten:
Next to the `snyk_code` parser that parses results from the CLI command, this new parser is able to parser results from the Snyk Issue API.

This API returns findings in a completely different format, so this is a new parser.
We only have a sample for issues of type `code`, so that's what supported for now.

The markdown doc has information about the mapping of fields.

Assumptions:
- `id` field is suitable as `unique_finding_from_tool`

[sc-11654]

### DIFF
--- a/docs/content/en/connecting_your_tools/parsers/file/snyk_issue_api.md
+++ b/docs/content/en/connecting_your_tools/parsers/file/snyk_issue_api.md
@@ -35,7 +35,7 @@ The parser maps fields from the Snyk Issue API response to DefectDojo's Finding 
 | static_finding | true | Always set to true |
 | dynamic_finding | false | Always set to false |
 | out_of_scope | attributes.ignored | Set to true if issue is ignored |
-| fix_available | coordinates[].is_fixable_* | True if any fixability flag is true |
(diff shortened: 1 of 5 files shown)


valentijnscholten:
@manuel-sommer This report provides some `is_fixable_xxx` flags. But for all SAST findings I guess there's a fix because people can fix the code. What shall we do with that. I guess only set the flag if parser provide it otherwise leave it `None` even if the finding is a SAST finding?

manuel-sommer:
> @manuel-sommer This report provides some `is_fixable_xxx` flags. But for all SAST findings I guess there's a fix because people can fix the code. What shall we do with that. I guess only set the flag if parser provide it otherwise leave it `None` even if the finding is a SAST finding?

Yeah, I would generally mark SAST findings as `is_fixable = True`, but we haven't finished the disucssion yet how to proceed with this field. Either an updatefunctionality to edit existing findings via reimport is possible or via deduplication.

dogboat (approved):

mtesauro (approved):
Approved

PR #12905, opened 2025-08-02
unittest: run in parallel

valentijnscholten:
let's see if it works in 2025


valentijnscholten:
Still errors even if I run those problematic test without `--parallel`. Weird but not worth pursuing as it looks like there's no speed-up to be had from running in parallel anyway.

PR #12914, opened 2025-08-04
Perf4 chord grade

valentijnscholten:


dojo/importers/default_reimporter.py
```diff
@@ -286,13 +351,21 @@ def close_old_findings(
...
         if self.findings_groups_enabled and self.push_to_jira:
             for finding_group in {finding.finding_group for finding in findings if finding.finding_group is not None}:
                 jira_helper.push_to_jira(finding_group)
 
+        # Calculate grade once after all findings have been closed
+        if mitigated_findings:
+            product = self.test.engagement.product
+            system_settings = System_Settings.objects.get()
+            if system_settings.enable_product_grade:
+                calculate_grade(product)
```
Maffooch: This block of code is written four times total. Can it moved to the base importer?

dojo/importers/default_importer.py
```diff
@@ -225,9 +238,48 @@ def process_findings(
...
+
+                    logger.debug(f"Launched chord with {len(post_processing_task_signatures)} tasks (batch #{current_batch_number}, size: {len(post_processing_task_signatures)})")
+
+                    # Reset for next batch (only if not final)
+                    post_processing_task_signatures = []
+                    if not is_final_finding:
+                        current_batch_number += 1
+            else:
+                # Execute task immediately for synchronous processing
+                post_processing_task_signature()
```
Maffooch, line 261: This block is written twice total. Can it be moved to the base importer as well?

dojo/models.py
```diff
@@ -2251,13 +2251,16 @@ def hash_code_allows_null_cwe(self):
...
         super().delete(*args, **kwargs)
-        with suppress(Test.DoesNotExist, Engagement.DoesNotExist, Product.DoesNotExist):
-            # Suppressing a potential issue created from async delete removing
-            # related objects in a separate task
-            calculate_grade(self.engagement.product)
+        if product_grading_option:
+            with suppress(Test.DoesNotExist, Engagement.DoesNotExist, Product.DoesNotExist):
+                # Suppressing a potential issue created from async delete removing
+                # related objects in a separate task
+                system_settings = System_Settings.objects.get()
```
Maffooch: _slight_ optimization
```suggestion
                system_settings = System_Settings.objects.only("enable_product_grade").get()
```

dojo/models.py
```diff
@@ -2856,15 +2859,18 @@ def copy(self, test=None):
...
         super().delete(*args, **kwargs)
-        with suppress(Finding.DoesNotExist, Test.DoesNotExist, Engagement.DoesNotExist, Product.DoesNotExist):
-            # Suppressing a potential issue created from async delete removing
-            # related objects in a separate task
-            calculate_grade(self.test.engagement.product)
+        if product_grading_option:
+            with suppress(Finding.DoesNotExist, Test.DoesNotExist, Engagement.DoesNotExist, Product.DoesNotExist):
+                # Suppressing a potential issue created from async delete removing
+                # related objects in a separate task
+                system_settings = System_Settings.objects.get()
```
Maffooch: _slight_ optimization
```suggestion
                system_settings = System_Settings.objects.only("enable_product_grade").get()
```

dojo/models.py
Maffooch: Product grade calculation is also performed on the engagement delete, but the check for system settings was not applied there

Maffooch (changes requested):
I need to learn more about these celery chords! They seem incredible

dojo/models.py
valentijnscholten: I don't think it will as it will either break or bypass the per request cache.

dojo/importers/default_reimporter.py
valentijnscholten: moved it

dojo/models.py
valentijnscholten: added for engagement delete, nice find.

dojo/importers/default_importer.py
valentijnscholten: done. I've left the if statement outside of the new method as it would otherwise look like some magic black box

mtesauro (approved):
Approved

PR #12922, opened 2025-08-04
:tada: Advance reimport to update fix_available field #12633

manuel-sommer:
- I advanced fix_available to be updated within each reimport: https://github.com/DefectDojo/django-DefectDojo/issues/12633#issuecomment-3152115864
- I added field fix_version to populate a dedicated fixed version in the header of a finding. This field can also be used to be updated equal to fix_available within the reimport. An example screenshot can be found here:
grafik
- I added anchore grype to populate these fields 

https://github.com/DefectDojo/django-DefectDojo/issues/12633#issuecomment-3223523700


manuel-sommer:
@valentijnscholten is this the right direction for updating fix_available in reimport?

valentijnscholten:
Yes

manuel-sommer:
Shall I target this against bugfix, or do we want a longer testphase and keep it against dev?

valentijnscholten:
The main reason fields are not updated during reimport is that there's a risk it overwrites fields that were changed by the user.  At least that's what I think the reason is. And this could be an issue here are the user may have set the `fix_available` field to `None` or `False` explicity. On the other hand we want to make that if a fix has become available we update the finding. What are your thoughts on this @Maffooch @mtesauro

manuel-sommer:
I mean that is what is requested also in the linked issue explicitly.

valentijnscholten:
We've discussed this and since this is sort of a "status" field or "meta" field it's OK to let `reimport` update it. Two remarks:
- I think we would just always need to update it, so also the case where it goes from `True` to `False` (or to `None`)
- I think it would have to be updated for any existing finding, so all branches of `process_matched_finding`.

Maffooch:
> We've discussed this and since this is sort of a "status" field or "meta" field it's OK to let `reimport` update it. Two remarks:
> 
> * I think we would just always need to update it, so also the case where it goes from `True` to `False` (or to `None`)
> * I think it would have to be updated for any existing finding, so all branches of `process_matched_finding`.

I agree with this approach

manuel-sommer:
Please review @valentijnscholten

manuel-sommer:
Another point @valentijnscholten :
I can also remove the mitigation field from the deduplication settings in general and update it as well. Also take a look here: 
https://github.com/DefectDojo/django-DefectDojo/pull/13053
https://github.com/DefectDojo/django-DefectDojo/pull/13055
With these two PRs, all mitigation fields in use would be removed from HASHCODE_FIELDS_PER_SCANNER in settings.dist.py
This enables us to update the mitigation field also as soon as a fix_available changes. These two fields often correlate.

valentijnscholten:
Maybe start with just the `fix_available` field first as we are all in agreement that is good change to make.

manuel-sommer:
Could we release this for the next release on Monday @mtesauro ?

dojo/importers/default_reimporter.py
```diff
@@ -431,6 +431,11 @@ def process_matched_mitigated_finding(
         to cover circumstances where mitigation timestamps are different, and
         decide which one to honor
         """
+        if existing_finding.fix_available != unsaved_finding.fix_available:
+            existing_finding.fix_available = unsaved_finding.fix_available
+            if existing_finding.fix_version != unsaved_finding.fix_version:
+                existing_finding.fix_version = unsaved_finding.fix_version
+
```
valentijnscholten, line 489: Why are thse if-statements here, shouldn't it be safe to just assign the value from the unsaved_finding?

dojo/importers/default_reimporter.py
```diff
@@ -537,6 +542,10 @@ def process_matched_active_finding(
         # First check that the existing finding is definitely not mitigated
         if not (existing_finding.mitigated and existing_finding.is_mitigated):
             logger.debug("Reimported item matches a finding that is currently open.")
+            if existing_finding.fix_available != unsaved_finding.fix_available:
+                existing_finding.fix_available = unsaved_finding.fix_available
+                if existing_finding.fix_version != unsaved_finding.fix_version:
+                    existing_finding.fix_version = unsaved_finding.fix_version
```
valentijnscholten: Same

dojo/templates/dojo/view_finding.html
```diff
@@ -556,6 +556,9 @@ <h3 class="pull-left finding-title">
                 {% if finding.component_version %}
                 <th>Component Version</th>
                 {% endif %}
+                {% if finding.fix_version %}
+                <th>Fixed Version</th>
+                {% endif %}
```
valentijnscholten, line 562: I think we should add the fix_available field as well? Maybe combine it into the same cell, i.e. "Fix available": Yes (4.13.4)

dojo/models.py
```diff
@@ -2416,6 +2416,11 @@ class Finding(models.Model):
                                         default=None,
                                         verbose_name=_("Fix Available"),
                                         help_text=_("Denotes if there is a fix available for this flaw."))
+    fix_version = models.CharField(null=True,
+                                         blank=True,
+                                         max_length=100,
+                                         verbose_name=_("Fix version"),
+                                         help_text=_("Version of the affected component where the flaw will be fixed."))
```
valentijnscholten:
```suggestion
                                         help_text=_("Version of the affected component in which the flaw is fixed."))
```

valentijnscholten: we also have to be aware that this can be multiple versions i.e. 1.2.1 and 2.4.1 if multiple release lines are supported.

dojo/importers/default_reimporter.py
manuel-sommer: good point

manuel-sommer: I removed the second if statement. I keep the first one as then the finding might not need to be touched at all.

dojo/importers/default_reimporter.py
manuel-sommer: See above

dojo/templates/dojo/view_finding.html
manuel-sommer: grafik
 looks fine to me

valentijnscholten: But where is the fix_available field shown?

manuel-sommer:
> Can you please write some unit tests for the change of state? Ideally every change made in the reimporter class should have coverage

done @Maffooch

manuel-sommer:
> some comments posted.

done @valentijnscholten

manuel-sommer:
Could you help to retrigger the unittests @mtesauro ?

valentijnscholten (commented):
can we show the fix_available fields somewhere, at least when fix_version is empty?

manuel-sommer:
> can we show the fix_available fields somewhere, at least when fix_version is empty?

grafik

manuel-sommer:
Ready to be reviewed besides the failing checks. Could you retrigger the pipelines @mtesauro ?

valentijnscholten:
They're running now but are you sure you've committed all the migrations needed?

manuel-sommer:
> They're running now but are you sure you've committed all the migrations needed?

Yeah, I made an error while rebasing and had two migrations point to the same db migration. Since that rebase, these unit test errors persist.

dojo/db_migrations/0243_finding_fix_version.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('dojo', '0242_file_upload_cleanup'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='finding',
+            name='fix_version',
+            field=models.CharField(blank=True, help_text='Version of the affected component where the flaw will be fixed.', max_length=100, null=True, verbose_name='Fix version'),
```
valentijnscholten:
```suggestion
            field=models.CharField(blank=True, help_text='Version of the affected component in which the flaw is fixed.', max_length=100, null=True, verbose_name='Fix version'),
```

dojo/templates/dojo/view_finding.html
manuel-sommer: grafik
I pushed another commit

valentijnscholten: The fix_available might be interesting to show also when it's False or Unknown (None). But I will approve now so it can be merged.

valentijnscholten (approved):

manuel-sommer:
> Can you please write some unit tests for the change of state? Ideally every change made in the reimporter class should have coverage

done

valentijnscholten:
> > They're running now but are you sure you've committed all the migrations needed?
> 
> Yeah, I made an error while rebasing and had two migrations point to the same db migration. Since that rebase, these unit test errors persist.

I synced the migration with the text change made earlier, now they are green.

manuel-sommer:
Ah, ok. Thank you. :-)

manuel-sommer:
A question here @Maffooch and @valentijnscholten : Shall I rename "fix_version" to "fix_description"?
Then, also static findings can be covered in a better way.

valentijnscholten:
I'm afraid I'm not following, could you give some more context?

manuel-sommer:
Dynamic Finding:
fix_version = 2.3.4

Static Finding:
fix_version = Do this to remediate the static finding.

Thus, fix_version is not really a version fix in static finding, or we rather point on the mitigation field in this regard or leave it open.

valentijnscholten:
The `fix_version` should contain 1 or more versions in which the issue is fixed. If it is a code change or configuration change or something else which is not about updating to a newer version we should use the `mitigation` field (or sometimes it's part of the `description`)

manuel-sommer:
> The `fix_version` should contain 1 or more versions in which the issue is fixed. If it is a code change or configuration change or something else which is not about updating to a newer version we should use the `mitigation` field (or sometimes it's part of the `description`)

Then, @valentijnscholten I would suggest to remove it as only a small number of vuln scanners post multiple fixed versions. Also, we can put that information to mitigation

valentijnscholten:
What do you want the remove? I think the `fix_version` field is useful to have as can be a column on reports/exports/dashboards.

manuel-sommer:
Ok, sounds legid

manuel-sommer:
fyi @kiblik

manuel-sommer:
Would be awesome if we could get this on the road for this release @mtesauro

valentijnscholten:
@manuel-sommer Looks good. Can you update the docs/content/en/connecting_your_tools/import_scan_files/using_reimport.md doc file to also mention the fix_version field.

(And rebase the migration)

manuel-sommer:
> @manuel-sommer Looks good. Can you update the docs/content/en/connecting_your_tools/import_scan_files/using_reimport.md doc file to also mention the fix_version field.
> 
> (And rebase the migration)

done @valentijnscholten

manuel-sommer:
could you

valentijnscholten:
@manuel-sommer I think this PR is a nice addition. Do you need any help to get across?

manuel-sommer:
@valentijnscholten , as there were multiple rebases, the db migration causes these failures. I just rebased again and adapted the db migration, everything should be fine.

manuel-sommer:
> @manuel-sommer Can you regenerated the migrations? This is needed due to the addition of django-pghistory.

just regenerated a new migration @valentijnscholten

manuel-sommer:
I would love to see this being released as I opened the PR already in August.

manuel-sommer:
> Approved. Do you think a small note in 2.52.md should be added to make users aware? It's the first time we let reimport override any non-status fields.

done @valentijnscholten

docs/content/en/open_source/upgrading/2.52.md
```diff
@@ -92,5 +92,8 @@ There are other instructions for upgrading to 2.52.x. Check the [Release Notes](
 
 Mobsfscan Scan" has been merged into the "MobSF Scan" parser. The "Mobsfscan Scan" scan_type has been retained to keep deduplication working for existing Tests, but users are encouraged to move to the "MobSF Scan" scan_type.
 
+## Reimport updates fields fix_available and fix_version
+If any incoming Findings match Findings that already exist, the incoming Finding is checked if the fields `fix_available` and `fix_version` differ and are updated if yes.
```
valentijnscholten: I don't know why but github doesn't allow me to do code suggestions sometimes.
Can you do

_Reimport will update existing findings `fix_available` and `fix_version` fields based on the incoming scan report._

valentijnscholten: but leave the heading

manuel-sommer:
Updated the upgrading notes to 2.53

Just checking in to see if there's a timeline for the review. I'd really appreciate it if this could be merged soon – it would also save me from having to maintain the upgrade notes and db migrations every month. @mtesauro

Maffooch (approved):

blakeaowens (approved):

PR #12930, opened 2025-08-05
ADD: Alternative command to change password

ThiagoCruzBr:
Alternative command that can be used into automated change password process.

### DIFF
--- a/readme-docs/DOCKER.md
+++ b/readme-docs/DOCKER.md
@@ -152,6 +152,11 @@ Make sure you write down the first password generated as you'll need it when re-
 docker compose exec -it uwsgi ./manage.py changepassword admin
 ```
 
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Thank you for the PR. I am hesitant though whether it provides any benefit over the much simpler already documented `docker compose exec -it uwsgi ./manage.py changepassword admin` ?

ThiagoCruzBr:
@valentijnscholten 
I see. The simpler command is interactive, so my command suggestion is more applicable to automated process like Ansible playbook.

valentijnscholten:
That's a fair point. If you base the PR against `bugfix` and change it to use `docker compose exec..` we might be able to get it approved and merged.

PR #12916, opened 2025-08-04
quickFix: invalid config in "Feat(nginx): Add support for IPv6"

kiblik:
This reverts commit e9d987206cd747b76df89605e9bf8108e8abdfc4.

PR #12938, opened 2025-08-06
Enable ipv6 in nginx (if available)

kiblik:
#12916 offered a quick fix by reverting previous implementation.
This PR properly fixes #12710 - correctly handles setups where IPv6 is not enabled


valentijnscholten:
Thanks. Updated the title a bit as those end up in the release notes :-)

blakeaowens (approved):

PR #12939, opened 2025-08-06
Display Tags: Do not rely on the request object being present

Maffooch:
In cases of an exception, I was receiving an error page from nginx rather than getting the 5xx response hosted by django. This was caused from a key error of the `request` object in the context of a given view


valentijnscholten (approved):

mtesauro (approved):
Approved

PR #12954, opened 2025-08-08
Docs: Restore `package-lock.json`

Maffooch:
the new `docsearch` dependencies got removed from the `package-lock.json` file during the release due to conflict resolution. When developing locally, the command `npm run dev` is used, which reads from `package.json` and then updates the `package-lock.json` file. However, when docs are deployed for production, the `package-lock.json` file is used to avoid pulling in any upgrades.

The search function is busted because it is trying to reference packages that are not installed


dogboat (approved):

Maffooch:
Going to merge with two approvals since this is technically a dependency upgrade

PR #12959, opened 2025-08-10
Documentation: Guide to testing hugo pipeline locally

Maffooch:
There was a snag in deploying some recent features features to production due to the way Hugo bundles and minifies its assets. This means we have a situation where some features could work during local development, but not during a production deployment. To remedy this this PR include two thing:
- Patches to correct some issue with a feature not working in production
- A guide to instruct users to test both deployment methods before submitting PRs


mtesauro (approved):
Approved

rossops (approved):

PR #12974, opened 2025-08-12
Ruff: Add and fix DTZ002 & DTZ011 + reorg some timezone components

kiblik:
Add and fix:
- [call-datetime-today (DTZ002)](https://docs.astral.sh/ruff/rules/call-datetime-today/#call-datetime-today-dtz002)
- [call-date-today (DTZ011)](https://docs.astral.sh/ruff/rules/call-date-today/#call-date-today-dtz011)


kiblik:
> I am curious how this behaves with the `DD_TIME_ZONE` env var

Thanks for pointing it out. I forgot about this env var. In my opinion, it is quite confusing to have multiple places for setting the time zone.

During deeper analysis, I found some leftovers (addressed in #12995) that might simplify this PR.

I was also trying to find the reason for the existence of `time_zone` in `system_setting`.
Here it is #313 (wow, 3-digit PR, was some quite some time ago 😄).

In my opinion, the idea was nice and I support it in some way (to have more settings in `system_setting` so you would be able to change it without restarting of application). Unfortunately, with `time_zone`, it is a little bit complicated. I suppose that the native Django approach (using `TIME_ZONE`, loaded from env var `DD_TIME_ZONE`) is used in multiple places of application. The same setting of time zone setting should be used across the whole application. It should not be selectively loaded in different parts differently. And it is much harder for maintenance to refer to `system_setting` everywhere (it is easy to forget about it).

I'm willing to drop all `get_system_setting("time_zone")` and use only the Django native approach. If you agree.

Maffooch:
> I'm willing to drop all get_system_setting("time_zone") and use only the Django native approach. If you agree.

I do agree with this approach. It would be really nice to have the system setting working everywhere, but that would require a bit of a migration that could have some consequences if it didn't go perfectly. I'm definitely open to doing this in the future, but am not sure now is best time

PR #12976, opened 2025-08-12
File Path Access: Prevent exception for non existent paths

Maffooch:
Prevent exception for non existent paths


mtesauro (approved):
Approved

blakeaowens (approved):

rossops (approved):

PR #12984, opened 2025-08-13
feat(helm): Improve docs, add schema

kiblik:
- README in the form of a symlink was dropped. It was generating some ugly warnings until now
- `readme-docs/KUBERNETES.md` was replaced with a simple description where the information can be found
- No content from `README.md`/`KUBERNETES.md` was removed, only
  - moved to another location (`README.md.gotmpl`)
  - extended by autogenerated help (list of available variables)
- add hint for string/nil values
- add `values.schema.json` - this is useful for values validation before manifests are sent to the k8s API
- add native linter
- add generator/validator of documentation (if you updated `values.yaml` but did not mention it in the documentation, GHA will fail)

There is still a huge place for improvement (render comments from `values.yaml` to descriptions, add `$ref` for better validation of k8s resources, mention all available variables, add `no-additional-properties`, ...). Still, it is a good second step (the first was definitely #12691) for improving the helm chart.


kiblik:
@fernandezcuesta, I was planning to autogenerate `values.schema.json` + documentation using
- https://github.com/losisin/helm-values-schema-json
- https://github.com/norwoodj/helm-docs

However, I bumped into the following issue:
```
- at '/postgresServer': got null, want string
- at '/imagePullSecrets': got null, want string
- at '/redisServer': got null, want string
```
I know that I would be able to forcefully patch related places with `type: [null, string]`, but it does not look/sound good.
Do you have any recommendations for a better way to generate schema and docs automatically? I wasn't able to fix it with any helper/annotation in the `values.yaml` file using the mentioned tools.

fernandezcuesta:
Hi, I guess it's gonna be easier to set those 3 to an empty string rather than `nil` also to be consistent with the rest of default values.
On the other hand, it's weird that annotations aren't doing the job (it should be pretty much the same as [this example](https://github.com/losisin/helm-values-schema-json/tree/main/docs#multiple-types))

kiblik:
> Hi, I guess it's gonna be easier to set those 3 to an empty string rather than `nil` also to be consistent with the rest of default values. On the other hand, it's weird that annotations aren't doing the job (it should be pretty much the same as [this example](https://github.com/losisin/helm-values-schema-json/tree/main/docs#multiple-types))

Ok, it looks like I made so silly typo when I was testing it before because now it works perfectly.

PR #12996, opened 2025-08-15
Bump: Python & nginx

kiblik:
For some reason, Renovate and Dependabot are not recognising new versions of images, so we have been left out a couple versions behind.

For now, blocked by #12998


kiblik:
This PR was separated to smaller PRs

PR #12998, opened 2025-08-15
feat(docker): Drop nginx debian

kiblik:
Both Django and Nginx have been available for a long time in Debian and Alpine alternatives. However, Nginx has never been the true Debian image. Yes, Debian was used in the building phase, but only for running `collectstatic`. But final image is based on Alpine every time.
https://github.com/DefectDojo/django-DefectDojo/blob/2b63f9e37e29090dd08bdb22e2a94084402767d5/Dockerfile.nginx-debian#L76

This PR drops support for building the nginx-debian image and supports only nginx-alpine.

If there are no complaints about this idea in general, I will add related information to the release notes as well (I just do not want to waste time on writing it until it is clear that this PR will pass).


.github/workflows/release-x-manual-docker-containers.yml
```diff
@@ -36,6 +36,9 @@ jobs:
       matrix:
           docker-image: [django, nginx]
           os: [alpine, debian]
+          exclude:
+              - docker-image: nginx
+                os: debian
```
mtesauro, line 41: This potentially appears to be removing all Debian images from being pushed to Docker Hub or am I misreading this.
I've never messed with this GH Action - at first I thought it was just the nginx image but then I wasn't sure so wanted to ask first.

valentijnscholten: I think this is interpreted as an "AND" and will only exclude `nginx-debian`.

valentijnscholten (approved):
Approved is in itself it looks OK. However I think we should consider droppding Debian alltogether, i.e. also for the `django` image. Although I am a fan of `Debian`, the go-to base image seems to be `alpine` for most projects. The images will be  ~20% smaller as well. And we can simplify our workflows, plus reduce the maintenance load on version update PRs.

.github/workflows/release-x-manual-docker-containers.yml
kiblik: @valentijnscholten is right here.

kiblik:
> Approved is in itself it looks OK. However I think we should consider droppding Debian alltogether, i.e. also for the `django` image. Although I am a fan of `Debian`, the go-to base image seems to be `alpine` for most projects. The images will be ~20% smaller as well. And we can simplify our workflows, plus reduce the maintenance load on version update PRs.

Fully agree with this. I'm happy to help with this, but I would do it step by step (now nginx, next django).
And I agree. `Alpine` is a more Docker-preferred solution. It is much smaller, and on average, there are fewer vulnerabilities.

mtesauro:
@valentijnscholten @kiblik  I've been working on wolfi based images to provide hardened containers for the project. I'd prefer to keep the Debian images around until we have a hardened replacement that uses glibc TBH.

The reason I'm not a fan of Alpine is the choice of musl and the strange bugs that can slip in when most distros use glibc. glibc has much better QA / QE / testing coverage due to it's much wider usage. [Here's](https://martinheinz.dev/blog/92) an example of these odd and painful bugs - or [here](https://dev.to/kakisoft/dockeralpine-why-you-should-avoid-alpine-linux-44he) or [here](https://pythonspeed.com/articles/alpine-docker-python/).

Networks keep getting faster and disk storage is ridiculously cheap so trading off better tested underlying OS vs image size feels like a pretty good trade to me.

mtesauro (approved):
Approved

Maffooch (approved):

Maffooch:
@rossops leaving the last approval to you so that you're aware of this change

valentijnscholten:
In think the situation around Alpine had improved over the years. But I don't have a strong opinion on which distro to use. But I would like to switch to one distro only at some point 😄

rossops (approved):

mtesauro:
> In think the situation around Alpine had improved over the years. But I don't have a strong opinion on which distro to use. But I would like to switch to one distro only at some point 😄

I agree with the idea of a single distro for our containers. Give me bit to finish the hardened containers and I think everyone will be happy as they should have few to no vulns and be as small as feasible.

PR #13001, opened 2025-08-17
Dynamic groups

LeoOMaia:
This is acording to #12684 and #12814.
I developed a new tab to group findings in accord in the mode (Vuln ID from tool/Title/CVE) the user wants to group dynamically.

Advantages of using dynamic grouping instead of static groups:
- Flexibility for the user to choose how the groups should be organized.
- Ability to group all existing findings as well as new ones generated by the system, without requiring imports.
- Support for grouping findings across multiple products simultaneously. The backend automatically filters which findings should be considered for each user. This means one user may already have all findings in a group analyzed, while for another the group may still be under review and remain open.
- Independent from the PostgreSQL database, making it easy to extend with new features for findings without affecting the rest of the system.

Disadvantages:
- Requires Redis to store groupings.
- Any changes to findings trigger a reconstruction of the groups when a user accesses the tab. (Note: if findings are changed but the user does not open the tab, the groups stored in Redis will not be created or updated. More details can be found in the README located in the directory where the views for dynamic and static groups are defined.)

Below is how it will be shown for the user to choose the group:
Screenshot from 2025-08-17 16-50-40

The mode selected by the user (default is empty and displays nothing):
Screenshot from 2025-08-17 16-53-55

Below is an example with the static and dynamic groups tab:
- Static (contains more groups):
Screenshot from 2025-08-17 16-31-28

- Dynamic:
Screenshot from 2025-08-17 16-33-27

And another example where a user has 1 and 2 products, and another has only 2 products where all the findings are mitigated:
- User1:
Screenshot from 2025-08-17 16-38-07

- User2:
Screenshot from 2025-08-17 16-45-06

Below is the findings page in a dynamic group:
Screenshot from 2025-08-17 17-01-50

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -2088,15 +2088,15 @@ class DynamicFindingGroupsFindingsFilter(FilterSet):
         ],
         label="Severity",
     )
-    script_id = CharFilter(lookup_expr="icontains", label="Script ID")
+    vuln_id_from_tool = CharFilter(lookup_expr="icontains", label="Vulnerability Id From Tool")
     reporter = ModelMultipleChoiceFilter(queryset=Dojo_User.objects.none(), label="Reporter")
-    status = ChoiceFilter(choices=[("Yes", "Yes"), ("No", "No")], label="Active")
+    active = ChoiceFilter(choices=[("Yes", "Yes"), ("No", "No")], label="Active")
     engagement = ModelMultipleChoiceFilter(queryset=Engagement.objects.none(), label="Engagement")
     product = ModelMultipleChoiceFilter(queryset=Product.objects.none(), label="Product")
 
     class Meta:
         model = Finding
-        fields = ["name", "severity", "script_id", "reporter", "status", "engagement", "product"]
+        fields = ["name", "severity", "vuln_id_from_tool", "reporter", "active", "engagement", "product"]
 
     def __init__(self, *args, **kwargs):
         self.user = kwargs.pop("user", None)
--- a/dojo/finding_group/redis.py
+++ b/dojo/finding_group/redis.py
@@ -10,23 +10,16 @@ from typing import Self
 
 import redis
 from django.conf import settings
-from django.utils.safestring import mark_safe
+from django.utils.functional import cached_property
 
 from dojo.models import Finding
 
 logger = logging.getLogger(__name__)
 
 DD_TEST = os.getenv("DD_TEST", "False").lower() == "true"
-SEVERITY_ORDER = {
-    "Critical": 5,
-    "High": 4,
-    "Medium": 3,
-    "Low": 2,
-    "Info": 1,
-}
 USER_MODES_KEY = "finding_groups_user_modes"
-SYSTEM_CHANGE = "finding_groups_last_finding_change"
-LAST_UPDATE = "finding_groups_last_update"
+LAST_FINDING_CHANGE = "finding_groups_last_finding_change"
+LAST_FINDING_UPDATE = "finding_groups_last_update"
 
 
 class GroupMode(StrEnum):
@@ -41,7 +34,6 @@ class DynamicFindingGroups:
     name: str = ""
     severity: str = "Info"
     main_finding_id: int | None = None
-    sla_finding_id: int | None = None
     finding_ids: set[int] = field(default_factory=set)
 
     def to_dict(self) -> dict:
@@ -63,12 +55,9 @@ class DynamicFindingGroups:
         return None
 
     def update_sev_sla(self, finding: Finding) -> None:
-        if SEVERITY_ORDER[finding.severity] > SEVERITY_ORDER[self.severity]:
+        if Finding.get_number_severity(finding.severity) > Finding.get_number_severity(self.severity):
             self.severity = finding.severity
             self.main_finding_id = finding.id
-        if finding.active and finding.sla_days_remaining():
-            if not self.sla_finding_id or finding.sla_days_remaining() < Finding.objects.get(id=self.sla_finding_id).sla_days_remaining():
-                self.sla_finding_id = finding.id
 
     def add(self, finding: Finding) -> None:
         self.update_sev_sla(finding)
@@ -77,7 +66,6 @@ class DynamicFindingGroups:
     # This method is used when we filter findings in a finding group
     def reconfig_finding_group(self) -> None:
         self.severity = "Info"
-        self.sla_finding_id = None
         findings = Finding.objects.filter(id__in=self.finding_ids)
         for finding in findings:
             self.update_sev_sla(finding)
@@ -85,13 +73,16 @@ class DynamicFindingGroups:
     @staticmethod
     def get_group_names(finding: Finding, mode: GroupMode) -> list[str] | None:
         if mode == GroupMode.VULN_ID_FROM_TOOL:
-            return [finding.vuln_id_from_tool]
+            if finding.vuln_id_from_tool:
+                return [finding.vuln_id_from_tool]
         if mode == GroupMode.TITLE:
-            return [finding.title]
+            if finding.title:
+                return [finding.title]
         if mode == GroupMode.CVE:
-            cves = list(
-                finding.vulnerability_id_set.values_list("vulnerability_id", flat=True),
-            )
+            cves = [
+                cve for cve in finding.vulnerability_id_set.values_list("vulnerability_id", flat=True)
+                if cve
+            ]
             if cves:
                 return cves
         return None
@@ -110,14 +101,14 @@ class DynamicFindingGroups:
             logger.info("Redis is not used in test environment, skipping.")
             return
         redis_client = get_redis_client()
-        redis_client.set(SYSTEM_CHANGE, datetime.now().isoformat())
+        redis_client.set(LAST_FINDING_CHANGE, datetime.now().isoformat())
 
     @staticmethod
     def set_last_update(mode: GroupMode, timestamp: datetime | None = None) -> None:
         if timestamp is None:
             return
         redis_client = get_redis_client()
-        redis_client.hset(LAST_UPDATE, mode.value, timestamp.isoformat())
+        redis_client.hset(LAST_FINDING_UPDATE, mode.value, timestamp.isoformat())
 
     @staticmethod
     def add_finding(finding: Finding, mode: GroupMode) -> None:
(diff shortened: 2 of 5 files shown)


dojo/filters.py
```diff
@@ -2043,6 +2043,78 @@ def set_related_object_fields(self):
...
+    severity = MultipleChoiceFilter(
+        choices=[
+            ("Low", "Low"),
+            ("Medium", "Medium"),
+            ("High", "High"),
+            ("Critical", "Critical"),
+        ],
+        label="Severity",
+    )
+    script_id = CharFilter(lookup_expr="icontains", label="Script ID")
```
valentijnscholten: Can we rename this to `vuln_id_from_tool` to avoid confusion as the script_id field doesn't exist in the Finding model?

dojo/filters.py
```diff
@@ -2043,6 +2043,78 @@ def set_related_object_fields(self):
...
+    reporter = ModelMultipleChoiceFilter(queryset=Dojo_User.objects.none(), label="Reporter")
+    status = ChoiceFilter(choices=[("Yes", "Yes"), ("No", "No")], label="Active")
```
valentijnscholten: Can we rename this to `active` or `is_active` to avoid confusion as the status field doesn't exist in the Finding model?

dojo/finding_group/redis.py
```diff
@@ -0,0 +1,256 @@
...
+DD_TEST = os.getenv("DD_TEST", "False").lower() == "true"
+SEVERITY_ORDER = {
+    "Critical": 5,
+    "High": 4,
+    "Medium": 3,
+    "Low": 2,
+    "Info": 1,
+}
+USER_MODES_KEY = "finding_groups_user_modes"
+SYSTEM_CHANGE = "finding_groups_last_finding_change"
```
valentijnscholten: Coudl this have a more clear name like LAST_FINDING_CHANGE or LAST_FINDING_UPDATE

dojo/finding_group/redis.py
```diff
@@ -0,0 +1,256 @@
...
+    @staticmethod
+    def load_from_id(finding_group_id: str, fg_key: str) -> Self | None:
+        redis_client = get_redis_client()
+        finding_group_data = redis_client.hget(fg_key, finding_group_id)
+        if finding_group_data:
+            return DynamicFindingGroups.from_dict(json.loads(finding_group_data))
+        return None
+
+    def update_sev_sla(self, finding: Finding) -> None:
+        if SEVERITY_ORDER[finding.severity] > SEVERITY_ORDER[self.severity]:
```
valentijnscholten: Could this also work without another mapping by using the numerical_severity as we do elsewhere?

dojo/finding_group/redis.py
```diff
@@ -0,0 +1,256 @@
...
+            days_remaining = "No SLA"
+        else:
+            status = "age-blue"
+            status_text = "No active finding"
+            days_remaining = "Concluded"
+        title = (
+            f'<a class="has-popover" data-toggle="tooltip" data-placement="bottom" title="" href="#" data-content="{status_text}">'
+            f'<span class="label severity {status}">{days_remaining}</span></a>'
+        )
+        return mark_safe(title)
```
valentijnscholten: I wonder if this could be made more clean by maybe putting it in a display tag to avoid mixing logic and display too much? I also wonder about the overlap there is with the Finding_Group model which also has logic to determine SLA data (and severity?) for a list of findings.

dojo/finding_group/redis.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+@lru_cache(maxsize=1)
+def get_redis_client() -> redis.Redis:
+    host = getattr(settings, "REDIS_HOST", "redis")
+    port = getattr(settings, "REDIS_PORT", 6379)
+    return redis.Redis(host=host, port=port, decode_responses=True)
```
valentijnscholten, line 157: I think this needs to reuse existing config in settings.dist.py?

dojo/finding_group/redis.py
```diff
@@ -0,0 +1,256 @@
...
+
+DD_TEST = os.getenv("DD_TEST", "False").lower() == "true"
+SEVERITY_ORDER = {
+    "Critical": 5,
+    "High": 4,
+    "Medium": 3,
+    "Low": 2,
+    "Info": 1,
+}
+USER_MODES_KEY = "finding_groups_user_modes"
```
valentijnscholten, line 20: I was confused by the word "USER" in this name. If I look at the code it seems to be grouping mode of the group. Is there are more clear name for this constant? Or is it that the grouping mode is stored _per user_?

dojo/finding_group/views_dynamic.py
```diff
@@ -0,0 +1,244 @@
...
+    def get_finding_groups(self, request: HttpRequest, products=None):
+        user_fids, active_fids = self.get_findings(products)
+        list_finding_group = []
+        for finding_group in self.finding_groups_map.values():
+            finding_group.finding_ids = set(finding_group.finding_ids) & user_fids
+            if self.filter_finding_group(finding_group, request):
+                if finding_group.finding_ids & active_fids:
+                    list_finding_group.append(finding_group)
+        return self.order_field(request, list_finding_group)
+
```
valentijnscholten: Can you explain what's happening here?

dojo/finding_group/redis.py
```diff
@@ -0,0 +1,256 @@
...
+
+import redis
+from django.conf import settings
+from django.utils.safestring import mark_safe
+
+from dojo.models import Finding
+
+logger = logging.getLogger(__name__)
+
+DD_TEST = os.getenv("DD_TEST", "False").lower() == "true"
```
valentijnscholten, line 19: It feels weird to have this flag/variable. Can we avoid it or at least use the same pattern as our other env variables via settings.dist.py?

LeoOMaia:
I understand why this happened: I didn’t handle the case where vuln_id_from_tool is None.
Since [None] is still a valid list, the check
```
if not finding_groups:
    return
```
does not filter it out.
In the function:
```
def get_group_names(finding: Finding, mode: GroupMode) -> list[str] | None:
    if mode == GroupMode.VULN_ID_FROM_TOOL:
        return [finding.vuln_id_from_tool]
    if mode == GroupMode.TITLE:
        return [finding.title]
    if mode == GroupMode.CVE:
        cves = list(
            finding.vulnerability_id_set.values_list("vulnerability_id", flat=True),
        )
        if cves:
            return cves
    return None
```
i will fix that, sorry.

dojo/finding_group/redis.py
LeoOMaia: It's per user. Once a user selects a mode, the next time they access the tab it will default to their previously chosen mode, without affecting other users who may have a different mode set.

dojo/finding_group/redis.py
LeoOMaia: I didn’t find anything definitive in the settings that confirms the default host and port for Redis. Because of that, I decided to handle it directly in the code. This way, if someone wants to change the host or port, they just need to define those two variables in the project settings. It works, but I’m not sure this is the best practice. Would you recommend handling it differently?

dojo/finding_group/views_dynamic.py
LeoOMaia: 1. It retrieves two sets of finding IDs:
`user_fids`: all findings the user is allowed to see (possibly restricted by products).
`active_fids`: the subset of findings that are currently active.

2. It iterates through all finding groups in self.finding_groups_map.

3. For each group, it intersects (`&`) the group’s finding IDs with user_fids, so the group only keeps findings the user has access to.

4. It applies additional filters through `filter_finding_group`, based on the request.

5. If the group still has at least one active finding (`finding_group.finding_ids & active_fids`), it is added to the result list to show in `ListOpenDynamicFindingGroups`.

6. Finally, it orders the groups using `order_field` and returns them.

dojo/finding_group/redis.py
LeoOMaia: I created this variable in `docker-compose.override.unit_tests_cicd.yml` to avoid using Redis during CI, because Redis is not created and there is no variable to indicate when we are running tests in CI. Do you recommend something to avoid this variable?

dojo/finding_group/redis.py
LeoOMaia: I will change to use `Finding.get_number_severity`

dojo/finding_group/views_dynamic.py
valentijnscholten: Can you add some comments or an overall explanation of how the grouping and caching works?
Are you doing the filtering/grouping/ordering in python code?
Are you caching the results per user?

LeoOMaia: I'm doing it in Python code, and I'm not caching the results per user. I use the finding_group created for the title mode, for example, stored in Redis, which is used by everyone who chose this mode, and I filter the findings and groups that the user can't see or wants to filter.

valentijnscholten:
When I look at your screenshots, the list of Dynamic Finding Groups looks a lot like pages in Defect Dojo where we display lists of entities with some statistics. For example the View Engagement page lists all the Tests in the Engagement and the number of findings, etc.

This basically a way to view at findings that are grouped by a certain field, in this case `test_id`. This is all done by the database using Django ORM / SQL queries. This is quite performant, at least up to 100K findings.

This PR wants to group findings by other fields such as `vuln_id_from_tool`, `cve`, `title`, etc.
Could you explain why a different implementation approach is needed here?

For example if I do an SQL query to group findings by `title`, this is very fast in my 170K instance (0.6s):

```
SELECT 
    f.title,
    COUNT(*) as total_findings,
    COUNT(CASE WHEN f.active = true THEN 1 END) as active_findings,
    MAX(f.numerical_severity) as max_numerical_severity,
    COUNT(DISTINCT t.engagement_id) as affected_engagements,
    COUNT(DISTINCT e.product_id) as affected_products,
    STRING_AGG(DISTINCT p.name, ', ') as product_names
FROM 
    dojo_finding f
    JOIN dojo_test t ON f.test_id = t.id
    JOIN dojo_engagement e ON t.engagement_id = e.id
    JOIN dojo_product p ON e.product_id = p.id
GROUP BY 
    f.title
ORDER BY 
    total_findings DESC;
```

This gives me similar results to the screenshots in this PR:

image

And that is including some joines to list the affected products, which would be a nice feature.

I might be forgetting something, but my initial approach would be to implement the dynamic grouping in the same way.

What are your thoughts on this @LeoOMaia 

Also did another test today to look at the static finding group pages, but running into this error:

```
  File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
               ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/views/generic/base.py", line 105, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/views/generic/base.py", line 144, in dispatch
    return handler(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/finding_group/views.py", line 302, in get
    paginated_finding_groups = self.paginate_queryset(finding_groups, request)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/dojo/finding_group/views.py", line 288, in paginate_queryset
    paginator = Paginator(queryset, page_size)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/core/paginator.py", line 46, in __init__
    self._check_object_list_is_ordered()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/core/paginator.py", line 142, in _check_object_list_is_ordered
    warnings.warn(
    ^

Exception Type: UnorderedObjectListWarning at /finding_group/open
Exception Value: Pagination may yield inconsistent results with an unordered object_list: <class 'dojo.models.Finding_Group'> QuerySet.
```

LeoOMaia:
I think your idea would be much better and would make it possible to remove the Redis storage. Regarding the message you received, was it an error or just a warning? I checked, and indeed I didn’t define a default ordering for when the user doesn’t request any sorting. I’m opening a PR to add a default ordering.

You’re right that grouping by title and vuln_id_from_tool would work correctly with the query you shared, and I agree this is a good approach for those fields. For CVE, however, the logic would need to be different. A single finding can have multiple CVEs, so the idea was to create groups for each CVE, which means the same finding could belong to multiple groups. Because of that, the query would have to be adapted, but I believe it’s still possible to achieve with a modified version of your example.

I’m currently looking into how I can adjust my implementation to better align with the query-based approach you suggested.

LeoOMaia:
Thank you for your comments! I’ll convert this PR into a draft so I can carefully review my code and remove Redis following your suggestion. It might take a little time, but once I’m done, I’ll reopen it.

Issue #12834, opened 2025-07-22
Upcoming changes to the Bitnami catalog (effective August 28th, 2025)


sjedlick:
With the note https://github.com/bitnami/charts/issues/35164, Bitnami announces the end of the Bitnami public catalog. Since DefectDojo also uses some Bitnami charts (Redis, Postgres) that currently point to the public catalog, the question arises as to whether Defectdojo will use its own charts for Redis and Postgres in the future?

Thank you
Swen

c-goosen:
This will affect a couple of things in the Kuberentes and docker and docker-compose setup. 
Quick search:

- docker.io/bitnami/redis
- docker.io/bitnami/postgres


Bitnami has accounted for legacy migrations with this, but not all tags are copied over yet: https://hub.docker.com/r/bitnamilegacy/postgresql/tags

This might cause major issues in defect-dojo upgrades of earlier versions.

c-goosen:
Bitnami is moving to this setup -> https://hub.docker.com/u/bitnamisecure, but this would not allow targeting specific postgresql versions.

niklasweimann:
Could you please clarify the strategy for dependencies?

My understanding is that defectdojo will switch to the `bitnamilegacy` repository and then awaiting a drop-in replacement. If this is accurate, does it imply that we won't receive any updates for these dependencies until the replacement is available?

valentijnscholten:
What would be your suggestion?

We're still looking at the options. We expected some alternatives may pop-up, it looks like Cloudpirates is setting up an alternative: https://www.cloudpirates.io/knowledge/blog/bitnami-helm-charts-was-sich-ab-august-2025-aendert-und-was-du-jetzt-tun-solltest-
Also Redis I think provides their own helm chart.

For the short term we may either switch to the bitnami legacy repo to at least have versioned charts. Or we may use the bitnami secure repo which only has _latest_ tags. 

The helm chart is a community supported deployment method so help is welcome.

yo-l1982:
Some alternatives for redis.
https://artifacthub.io/packages/helm/dandydev-charts/redis-ha 
Is used by ArgoCD and seems to be a stable chart(But HA).
Also dragonflydb is a drop in replacement for redis that is interesting.
https://www.dragonflydb.io/docs/getting-started/kubernetes

Sadly not much postgres other than bitnami.

d-mankowski-synerise:
> Sadly not much postgres other than bitnami.

[...]

PR #13006, opened 2025-08-18
feat(helm): Try to pin non-bitnami images

kiblik:
Because of #12834, I'm testing a scenario: Are we able to have version pinning without official bitnami images?


rossops:

PR #13015, opened 2025-08-19
feat(helm): Use bitnamisecure images + pin to latest digest

kiblik:
TBD


rossops:

PR #13033, opened 2025-08-22
BlackDuck: Support import in plaintext or bytes

Maffooch:
I found that while doing some importing using python request, black duck zip files had to be imported with `rb` mode in the files parameter:
```python
[("file", open(file_path, "rb"))]
```
However, when attempting to process CSV files in this way, I would constantly get an exception for two reasons:
 - The `temporary_file` function does not work well with bytes
 - Following the check if a file is zipped, the file pointer is not reset, so there are no bytes to read, which lead to a None type exception

To mitigate this, we need to support byte imports are a first class citizen. I checked the other black duck parsers as well to get them up to speed


paulOsinski (approved):

mtesauro (approved):
Approved

PR #13037, opened 2025-08-24
cvss4: remove no longer needed custom parsing

valentijnscholten:
With `cvss` 3.6 being released and used by us, we no longer need to maintain the copied in code to parse CVSS4. 
We can rely on the upstream `cvss` lib.


mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

PR #13042, opened 2025-08-25
fix(eng/failure_redirect): Fix rendering when coming from product site

kiblik:
Django raised `AttributeError: 'NoneType' object has no attribute 'id'` when a user visited `DD/product/<id>/import_scan_results` and triggered `failure_redirect`.

### DIFF
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -1055,9 +1055,14 @@ class ImportScanResultsView(View):
     ) -> HttpResponseRedirect:
         """Redirect the user to a place that indicates a failed import"""
         ErrorPageProductAnnouncement(request=request)
+        if obj := context.get("engagement"):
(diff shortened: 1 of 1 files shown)


valentijnscholten (commented):
The `engagement_or_product` thing looks a bit weird. I know it's pre-existing, but wouldn't it be cleaner to use `product` and `engagement`?

kiblik:
> The `engagement_or_product` thing looks a bit weird. I know it's pre-existing, but wouldn't it be cleaner to use `product` and `engagement`?

Well, it is comming from https://github.com/DefectDojo/django-DefectDojo/blob/e28f39e7aa1bbb9df636c383d73547d817961821/dojo/engagement/views.py#L721-L741

But I can separate functions

mtesauro (approved):
Approved

PR #13045, opened 2025-08-25
fix(eng/test-validate_forms): Show error if form is not valid

kiblik:
If there was any mistake in the forms, UI did not provide any feedback.
If `DD_CREATE_CLOUD_BANNER=True`, it showed _Pro comes with support. ..._ but there was no information what can be wrong (not even in logs).
`View` had been validating forms, but it has not provided any feedback.
Now it createsa  regular error message.

### DIFF
--- a/dojo/engagement/views.py
+++ b/dojo/engagement/views.py
@@ -878,13 +878,11 @@ class ImportScanResultsView(View):
         level are bubbled up to the user first before we process too much
         """
         form_validation_list = []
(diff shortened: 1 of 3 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

valentijnscholten (approved):

PR #13053, opened 2025-08-26
:tada: Add fix_available to Harbor

manuel-sommer:
https://github.com/DefectDojo/django-DefectDojo/issues/12633#issuecomment-2989905282

I removed also the mitigation from dedeuplication as this was only there to close the old finding in case the mitigation becomes available.

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -1301,7 +1301,7 @@ HASHCODE_FIELDS_PER_SCANNER = {
     "JFrog Xray Scan": ["title", "description", "component_name", "component_version"],
     "CycloneDX Scan": ["vuln_id_from_tool", "component_name", "component_version"],
     "SSLyze Scan (JSON)": ["title", "description"],
-    "Harbor Vulnerability Scan": ["title", "mitigation"],
+    "Harbor Vulnerability Scan": ["title"],
     "Rusty Hog Scan": ["file_path", "payload"],
     "StackHawk HawkScan": ["vuln_id_from_tool", "component_name", "component_version"],
     "Hydra Scan": ["title", "description"],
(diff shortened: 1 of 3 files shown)


dojo/settings/settings.dist.py
```diff
@@ -1301,7 +1301,7 @@ def saml2_attrib_map_format(din):
     "JFrog Xray Scan": ["title", "description", "component_name", "component_version"],
     "CycloneDX Scan": ["vuln_id_from_tool", "component_name", "component_version"],
     "SSLyze Scan (JSON)": ["title", "description"],
-    "Harbor Vulnerability Scan": ["title", "mitigation"],
```
Maffooch: This would create some churn for folks that didn't know about this change. Please revert this change

manuel-sommer: @Maffooch: I would suggest to create a dbmigration or recalculate the hashcode and remove the mitigation from deduplication. The explanation can be found here: https://github.com/DefectDojo/django-DefectDojo/pull/12922#issuecomment-3222906747 
This enables us to update the mitigation field in all reimports as soon as fix_available changes. These two fields often correlate. 
The only affected vulnerability scanners are Harbor and KrakenD Audit

valentijnscholten: It would be good to think about some guidelines for parsers on how to handle dynamic data. I remember some PRs I did on parsers where I moved some "dynamic" data from the description to other fields like mitigation. Not sure if that is the best approach, but maybe it could work if we have some rule like "description should be static over time so it can be used in hash_code calculation, other fields like mitigation can be dynamic and cannot (or should not) be used in hash_code calculations".

manuel-sommer: yeah, but I guess this can be a followup. Let's first merge this PR

Maffooch (approved):

blakeaowens (approved):

PR #13054, opened 2025-08-26
fix(docker): Docker Debian upgrade to Trixie

kiblik:
Renovate / dependabot are probably not able to recognise higher version of Debian (they are using names, not semver).


Maffooch (approved):

mtesauro (approved):
Approved

I wonder if the paint is dry on the Trixie release from Debian ;-)

dogboat (approved):

PR #13055, opened 2025-08-26
:tada: Add fix_available to KrakenDAudit

manuel-sommer:
https://github.com/DefectDojo/django-DefectDojo/issues/12633#issuecomment-2989905282

I removed also the mitigation from dedeuplication as this was only there to close the old finding in case the mitigation becomes available.

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -1343,7 +1343,7 @@ HASHCODE_FIELDS_PER_SCANNER = {
     "Invicti Scan": ["title", "description", "severity"],
     "Checkmarx CxFlow SAST": ["vuln_id_from_tool", "file_path", "line"],
     "HackerOne Cases": ["title", "severity"],
-    "KrakenD Audit Scan": ["description", "severity"],
+    "KrakenD Audit Scan": ["description", "mitigation", "severity"],
     "Red Hat Satellite": ["description", "severity"],
     "Qualys Hacker Guardian Scan": ["title", "severity", "description"],
     "Cyberwatch scan (Galeax)": ["title", "description", "severity"],


dojo/settings/settings.dist.py
```diff
@@ -1343,7 +1343,7 @@ def saml2_attrib_map_format(din):
     "Invicti Scan": ["title", "description", "severity"],
     "Checkmarx CxFlow SAST": ["vuln_id_from_tool", "file_path", "line"],
     "HackerOne Cases": ["title", "severity"],
-    "KrakenD Audit Scan": ["description", "mitigation", "severity"],
```
Maffooch: This would create some churn for folks that didn't know about this change. Please revert this change

manuel-sommer: @Maffooch: I would suggest to create a dbmigration or recalculate the hashcode and remove the mitigation from deduplication. The explanation can be found here: https://github.com/DefectDojo/django-DefectDojo/pull/12922#issuecomment-3222906747 
This enables us to update the mitigation field in all reimports as soon as fix_available changes. These two fields often correlate. 
The only affected vulnerability scanners are Harbor and KrakenD Audit

Maffooch: have you considered the use case of someone updating dedupe fields via the `local_settings.py` file or environment variables? We have found that making these sorts of changes through a migration doesn't often go so smoothly. I would feel more comfortable not changing any dedupe settings and knowing user data will not be impacted. Especially since the mitigation field is not changing in this PR

A follow up question: do we want to update the mitigation field in all reimports? I would think no as we cannot make an accurate guess as to whether the user had changed the mitigation of a finding.

Maffooch (approved):

PR #13057, opened 2025-08-26
:tada: Add fix_available to Trivy

manuel-sommer:
https://github.com/DefectDojo/django-DefectDojo/issues/12633#issuecomment-2989100044

### DIFF
--- a/dojo/tools/trivy/parser.py
+++ b/dojo/tools/trivy/parser.py
@@ -289,6 +289,9 @@ class TrivyParser:
                 package_version = vuln.get("InstalledVersion", "")
                 references = "\n".join(vuln.get("References", []))
                 mitigation = vuln.get("FixedVersion", "")
+                fix_available = True
+                if mitigation == "":
+                    fix_available = False
                 impact = vuln.get("Status", "")
                 status_fields = self.convert_trivy_status(vuln.get("Status", ""))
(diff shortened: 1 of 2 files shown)


mtesauro (approved):
Approved

valentijnscholten:
I like this idea of populating the `fix_available` field in ways like this. Would it be helpful/needed to add to each parser doc how the field is populated? I remember a PR that was documenting for some parsers which fields from the report were parsed and how they were mapped. We could use AI to complete this for all parsers and then keep it up-to-date with parser changes?
Or even without that we could start adding a line to the docs of the parsers that set `fix_available` to indicate how the parsers decides on the value of the field? This is especially valuable if the value is derived from other fields or based on text matching of another field.

valentijnscholten (commented):
Should these fix_available PRs have a migration to set the field for existing findings?
Should they be combined into one PR to make it manageable for thsi release?

manuel-sommer:
I guess they don't need a migration as reimport with this PR https://github.com/DefectDojo/django-DefectDojo/pull/12922 will take care of this earlier or later
I intentionally splitted these into multiple MRs to make review easier / faster. This would provide a chance for some parsers to get merged if other parsers would be delayed because of review recommendations.

valentijnscholten (changes requested):
Lots of people are using `import` and will have lots of existing findings. But I'm OK to  keep it simple for now.
But I do think we need to add a line to the docs.

PR #13061, opened 2025-08-26
:lipstick: Restructure Kubehunter json files to make it readable

manuel-sommer:


mtesauro (approved):
Approved

valentijnscholten (approved):
Approved.
Question: Should we make this the default that we make sample files readable/pretty-printed? Until now I assumed we didn't want to do that as strictly speaking we are modifying the sample file.

mtesauro:
> Approved. Question: Should we make this the default that we make sample files readable/pretty-printed? Until now I assumed we didn't want to do that as strictly speaking we are modifying the sample file.

I think it's in the "Nice to have bucket" meaning, I wouldn't refuse a PR that doesn't pretty print a sample file and I would approve a PR like this one that makes them pretty.

PR #13066, opened 2025-08-26
feat(form-import): DRY File Extension

kiblik:
Apply **DRY** ([Don't repeat yourself](https://en.wikipedia.org/wiki/Don't_repeat_yourself)) principle on the use of File Extension.

Because it is quite easy to forget to add it to other places.

### DIFF
--- a/dojo/forms.py
+++ b/dojo/forms.py
@@ -549,9 +549,7 @@ class ImportScanForm(forms.Form):
     tags = TagField(required=False, help_text="Add tags that help describe this scan.  "
                     "Choose from the list or add new tags. Press Enter key to add.")
     file = forms.FileField(
-        widget=forms.widgets.FileInput(
-            attrs={"accept": ".xml, .csv, .nessus, .json, .jsonl, .html, .js, .zip, .xlsx, .txt, .sarif"},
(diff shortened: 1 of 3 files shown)


Maffooch:
The reason I went away from the DRY principle in this case is because of this comment on the the `DD_FILE_UPLOAD_TYPES` variable:
> List of acceptable file types that can be uploaded to a given object via arbitrary file upload

it felt wrong to me to conflate the two purposes. Maybe we should have two variables? One for file uploads to engagements/tests/findings, and one for import/reimport

kiblik:
> The reason I went away from the DRY principle in this case is because of this comment on the the `DD_FILE_UPLOAD_TYPES` variable:
> 
> > List of acceptable file types that can be uploaded to a given object via arbitrary file upload
> 
> it felt wrong to me to conflate the two purposes. Maybe we should have two variables? One for file uploads to engagements/tests/findings, and one for import/reimport

You are right. Is it better now?

Maffooch (approved):
Yep, I think this is much better!

dogboat (approved):

blakeaowens (approved):

PR #13079, opened 2025-08-29
Password Reset: Disable for SSO users

Maffooch:
It is possible to set password reset for users that were authorized through SSO mechanisms. This has potential to lock a user into a flow of not being able to reset a password that does not exist


blakeaowens (approved):

valentijnscholten (approved):

mtesauro (approved):
Approved

Issue #12872, opened 2025-07-29
Severe Performance Degradation with Multiple Users on Findings and Dashboard


Hamudah:
## Problem description  
We are experiencing severe performance issues in our OS DefectDojo instance (latest release). Even with only three parallel users, the system becomes noticeably slow. One example reported by a user:

- Navigating to a finding (e.g., _All Products → product x → Sort findings High → Finding takes **at least 8 seconds** per request.
- If five users open different findings in parallel, the load time for a single user increases to **30–60 seconds** or more.
- This delay is reproducible and affects other areas like the **dashboard**, where initial load time is ~8s, but under load (parallel access) also increases to **30s+**.

📷 Screenshots:
- ![Finding load]<img width="3822" height="1255" alt="Image" src="https://github.com/user-attachments/assets/7244cd3b-84eb-4a88-b25c-45a95c8e2ece" />
- ![Finding load under parallel access]() <img width="3827" height="1444" alt="Image" src="https://github.com/user-attachments/assets/eb56b560-cb05-4e32-9110-76ae353f11d4" />
- ![Dashboard normal]()<img width="3781" height="1013" alt="Image" src="https://github.com/user-attachments/assets/9ccd2924-ba7e-4dd1-8611-a429a0ffd85e" />
- ![Dashboard degraded]()<img width="3720" height="1071" alt="Image" src="https://github.com/user-attachments/assets/844d4365-2a6b-4d9b-8417-de8b938f0e42" />

The issue gives the impression of global performance degradation when individual actions are executed (e.g., deleting reports, creating engagements).

---

## Steps to reproduce
1. Open a finding as described above.
2. Observe the time it takes (~4-8s).
3. Repeat this from 5 browser sessions or accounts in parallel.
4. Load times increase drastically (up to 60s per action).

---

## Expected behavior  
Findings and the dashboard should load in a reasonable time (ideally <2-4s), regardless of moderate parallel access.

## Additional context  
We noticed that displaying findings with lots of history or similar findings seems to slow things down more. We suspect possible performance bottlenecks in DB queries, ORM joins or similar.

## Our questions:
1. **What performance optimization measures would you recommend?**  
   (e.g., database indexing, caching layers, configuration tuning)

2. **Would scaling vertically (more CPU/RAM) help mitigate these issues**, or is the current bottleneck rather **architectural/software-related**?

3. Are there known performance limitations or improvements planned regarding how findings and dashboards are loaded and rendered?

[...]

PR #13080, opened 2025-08-30
uwsgi: default to 4 processes x 4 threads

valentijnscholten:
The default of having only 2 processes with 2 threads each can easily lead to very poor performance, see #12872, #12950

There's documentation on this on how to tune a production instance, but this limit can already be quite a limitation in non-production use or doing a PoC install. The CPU usage, memory usage, IOPS is sometimes still low with only 4 threads which makes it not immediately clear why the performance is low or requests are timing out.

This PR raised the defaults to 4 processed with 4 threads each. It also moves the performance related docs page closer to the installation & configuration section.


mtesauro (approved):
Approved

dogboat (approved):

blakeaowens (approved):

PR #13081, opened 2025-08-30
Default ordering to FindingGroup

LeoOMaia:
As discussed in #13001

### DIFF
--- a/dojo/finding_group/views.py
+++ b/dojo/finding_group/views.py
@@ -236,8 +236,8 @@ class ListFindingGroups(View):
             order_field_param = order_field_param[1:] if reverse_order else order_field_param
             if order_field_param in {"name", "creator", "findings_count", "sla_deadline"}:
                 prefix = "-" if reverse_order else ""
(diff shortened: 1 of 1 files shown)


valentijnscholten:
What does this PR do? I think it fixes the ordering of findings in finding groups if a non-default ordering is selected?

LeoOMaia:
The warning error that was being generated in the GroupFindings view basically occurs because I didn’t define a default ordering for the view when no specific ordering is selected. I’m just setting a default to avoid the warning.

valentijnscholten:
Is it similar to #13119 or are both PRs needed?

LeoOMaia:
I reviewed it and the idea seems similar to setting a default ordering, but the way the person is doing it looks wrong. If someone selects to order by a field, for example `name`, in the end the `paginate_queryset` will overwrite it by ordering with `id` i think.

valentijnscholten:
Can you rebase your PR and remove their change so we can merge your fix?

LeoOMaia:
I created a new PR to the `bug` branch. This current PR is targeting `dev`, so we can close this branch, is that okay?

valentijnscholten:
Yes

PR #13083, opened 2025-08-31
System settings: correct page title/breadcrumbs

valentijnscholten:
Small textual change to help people find the right browser tab and not confuse Application Settings (which don't exist in OS) and System Settings (which are great).

The page title had "Application Settings" when viewing the "System settings", this PR corrects that.


mtesauro (approved):
Approved

Maffooch (approved):

Jino-T (approved):

PR #13090, opened 2025-09-02
fix: uncontrolled data used in path expression

opsysdebug:
fix this issue, we need to prevent the user from being able to specify arbitrary or dangerous file extensions via the upload. The best way to address this without breaking existing functionality is to extract the extension, sanitize it (e.g., restrict it to a defined set such as `.zip`, `.pdf`, etc., or otherwise ensure only alphanumeric/expected extensions are used), and ensure it does not contain any path separators or special characters. An alternative is to discard the user-supplied extension and use a fixed, safe one, or check that it matches a whitelist.

Implement this by:  
- Importing `werkzeug.utils.secure_filename` (or, if not allowed, do minimal extension sanitization inline).
- Validating the extracted extension, e.g., by matching against a regular expression that allows only safe characters (e.g., `^[a-zA-Z0-9.]+$` and not starting with a dot-dot or containing slashes).
- Optionally filtering extensions to a whitelist (e.g., `.zip`, `.pdf`).
- If the extension is not valid, either reject the upload or set a safe default (such as `.bin`).

Modify only the relevant lines in `handle_uploaded_threat` in `dojo/utils.py`, adding any necessary imports above.

**Checklist**

### DIFF
--- a/dojo/utils.py
+++ b/dojo/utils.py
@@ -8,6 +8,7 @@ import os
 import pathlib
 import re
 from calendar import monthrange
+from werkzeug.utils import secure_filename
 from collections.abc import Callable
 from datetime import date, datetime, timedelta
 from math import pi, sqrt
@@ -1363,8 +1364,14 @@ def get_page_items_and_count(request, items, page_size, prefix="", *, do_count=T
 
 
 def handle_uploaded_threat(f, eng):
-    path = Path(f.name)
+    # Sanitize the file extension
+    original_name = secure_filename(f.name)
+    path = Path(original_name)
     extension = path.suffix
+    # Optional: limit to a whitelist of allowed extensions
+    allowed_extensions = {'.zip', '.pdf', '.docx', '.xlsx', '.png', '.jpg', '.jpeg', '.svg', '.txt'}
+    if extension.lower() not in allowed_extensions:
+        extension = '.bin'
     # Check if threat folder exist.
     threat_dir = Path(settings.MEDIA_ROOT) / "threat"
     if not threat_dir.is_dir():


dojo/utils.py
```diff
@@ -1363,8 +1364,14 @@ def get_page_items_and_count(request, items, page_size, prefix="", *, do_count=T
...
 def handle_uploaded_threat(f, eng):
-    path = Path(f.name)
+    # Sanitize the file extension
+    original_name = secure_filename(f.name)
+    path = Path(original_name)
     extension = path.suffix
+    # Optional: limit to a whitelist of allowed extensions
+    allowed_extensions = {'.zip', '.pdf', '.docx', '.xlsx', '.png', '.jpg', '.jpeg', '.svg', '.txt'}
+    if extension.lower() not in allowed_extensions:
+        extension = '.bin'
```
odaysec, line 1374: Here’s an explanation of the code changes made in lines 1367 to 1374 of dojo/utils.py in PR #13090:

**Before the change:**
```python
path = Path(f.name)
extension = path.suffix
```
- The code took the uploaded file’s name directly and extracted its extension, without any security checks.

**After the change:**
```python
# Sanitize the file extension
original_name = secure_filename(f.name)
path = Path(original_name)
extension = path.suffix
# Optional: limit to a whitelist of allowed extensions
allowed_extensions = {'.zip', '.pdf', '.docx', '.xlsx', '.png', '.jpg', '.jpeg', '.svg', '.txt'}
if extension.lower() not in allowed_extensions:
    extension = '.bin'
```
**What’s new and why?**
- The filename is "sanitized" using `secure_filename` (from werkzeug), which removes unsafe characters and path traversal attempts.
- The file extension is extracted from the sanitized name.
- A whitelist of allowed extensions is checked: only safe types (like .pdf, .png, .txt, etc.) are accepted.
- If the extension is not allowed, it is replaced with `.bin` (a generic, safe extension).

**Purpose:**  
These changes prevent users from uploading files with potentially dangerous names or extensions, reducing the risk of security vulnerabilities like arbitrary file execution or path traversal.

dojo/utils.py
```diff
@@ -1363,8 +1364,14 @@ def get_page_items_and_count(request, items, page_size, prefix="", *, do_count=T
 
 
 def handle_uploaded_threat(f, eng):
```
manuel-sommer, line 1366: Hi @opsysdebug :

- Please also take a look at https://github.com/DefectDojo/django-DefectDojo/blob/e28f39e7aa1bbb9df636c383d73547d817961821/dojo/settings/settings.dist.py#L287
- Please also point your PR against dev, not master.

manuel-sommer (changes requested):

Maffooch:
Converting to draft until this is completed

> Hi @opsysdebug :
> 
> * Please also take a look at https://github.com/DefectDojo/django-DefectDojo/blob/e28f39e7aa1bbb9df636c383d73547d817961821/dojo/settings/settings.dist.py#L287
> * Please also point your PR against dev, not master.

dojo/utils.py
odaysec: Hi @manuel-sommer 

The selected line is the function definition:

```python
def handle_uploaded_threat(f, eng):
```

This function is responsible for handling uploaded files (likely related to threats) within the application. The parameters are:
- `f`: The uploaded file object.
- `eng`: Likely an engine/context parameter for processing.

The recent changes within this function improve file handling security by sanitizing the filename, checking the file extension against a whitelist, and defaulting to `.bin` if the extension isn’t allowed. This helps mitigate risks associated with unsafe file uploads.

PR #13095, opened 2025-09-02
docker: remove dev dependencies from release images

valentijnscholten:
I wasn't going to spend time on this, but then Cursor AI got it right almost on the first go.

Some dependencies are only needed for development, we shouldn't install those in the release ("production")  images.

### DIFF
--- a/Dockerfile.django-alpine
+++ b/Dockerfile.django-alpine
@@ -31,7 +31,7 @@ COPY requirements.txt ./
 # https://github.com/unbit/uwsgi/issues/1318#issuecomment-542238096
 RUN CPUCOUNT=1 pip3 wheel --wheel-dir=/tmp/wheels -r ./requirements.txt
 
-FROM base AS django
+FROM base AS release
 WORKDIR /app
 ARG uid=1001
 ARG gid=1337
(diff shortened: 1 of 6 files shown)


mtesauro (approved):
Approved

mtesauro:
Did you consider doing the same for for the Debian dockerfile?
https://github.com/DefectDojo/django-DefectDojo/blob/master/Dockerfile.django-debian

valentijnscholten:
> Did you consider doing the same for for the Debian dockerfile? https://github.com/DefectDojo/django-DefectDojo/blob/master/Dockerfile.django-debian

Yes, it's part of the PR or did I miss a part?

mtesauro:
> > Did you consider doing the same for for the Debian dockerfile? https://github.com/DefectDojo/django-DefectDojo/blob/master/Dockerfile.django-debian
> 
> Yes, it's part of the PR or did I miss a part?

@valentijnscholten My bad. Been dealing with a bunch of family things lately and I guess I was tired when I looked at this before. I see that you did both so never mind my comment :point_up:

blakeaowens (approved):

PR #13114, opened 2025-09-04
Update the  Qualys Hacker Guardian parser to be compatible with larger CSV files

Jino-T:
[sc-11785]


Maffooch:
This should go against bugfix instead

PR #13125, opened 2025-09-06
fix(helm): Drop bitnami repo reference

kiblik:
As we are not using bitnami anymore, adding it as Helm repo is not needed.

### DIFF
--- a/.github/workflows/release-x-manual-helm-chart.yml
+++ b/.github/workflows/release-x-manual-helm-chart.yml
@@ -66,7 +66,6 @@ jobs:
 
       - name: Configure HELM repos
         run: |-
-             helm repo add bitnami https://charts.bitnami.com/bitnami
              helm dependency list ./helm/defectdojo
              helm dependency update ./helm/defectdojo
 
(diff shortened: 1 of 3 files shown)


kiblik:
There is still one reference in the Helm README, but it is referring to a different use case (test of external DB)
https://github.com/DefectDojo/django-DefectDojo/blob/52ef305bd1eda9da1aa3910da95d8c21776ce16c/readme-docs/KUBERNETES.md?plain=1#L408-L416
And two references in `values.yaml`, but there is no better replacement for documentation at the moment.
https://github.com/DefectDojo/django-DefectDojo/blob/d41ceaf996782f1ed884f390a9a59e21ff9a1bea/helm/defectdojo/values.yaml#L392
https://github.com/DefectDojo/django-DefectDojo/blob/d41ceaf996782f1ed884f390a9a59e21ff9a1bea/helm/defectdojo/values.yaml#L464

valentijnscholten:
Should we mention somewhere that we have vendored the bitnami charts for now?

kiblik:
> Should we mention somewhere that we have vendored the bitnami charts for now?

Not sure if it is needed. How do you see it, @rossops?

mtesauro (approved):
Approved

PR #13155, opened 2025-09-10
v3 labels

dogboat:
This patch introduces v3 text copy. "Product" is renamed to "Asset" and "Product Type" is renamed to "Organization" throughout. Additionally, new URL routes are added to reflect the change (i.e., `/asset`). A new system setting is added to toggle between the two versions, defaulting to the new copy.

All copy that is altered is concentrated in `labels.py`, which acts as a sort of centralized directory. A context processor is added allowing templates to reference these values through the `labels` variable. Application code can similarly access this data through use of the `get_labels()` method. URL routing is handled by a middleware that checks which version of copy is selected and sets the urlconf accordingly. To that end, a few packages' urls.py have been restructured to support this.


dogboat:
[sc-11741]

dojo/v3_migration/utils.py
```diff
@@ -0,0 +1,43 @@
...
+
+from django.shortcuts import redirect
+from django.urls import reverse
+
+from dojo.models import System_Settings
+
+
+def v3_migration_enabled():
+    """Returns whether v3 migration is enabled."""
+    return System_Settings.objects.get().enable_v3_migration
```
Maffooch: We can get away fetching just one column here
```suggestion
    return System_Settings.objects.only("enable_v3_migration").get().enable_v3_migration
```

dojo/filters.py
```diff
@@ -512,7 +519,7 @@ class FindingTagFilter(DojoFilter):
         field_name="test__engagement__product__tags__name",
         to_field_name="name",
         queryset=Product.tags.tag_model.objects.all().order_by("name"),
-        help_text="Filter Products by the selected tags")
+        help_text=labels.ASSET_FILTERS_TAGS_FILTER_HELP)
```
Maffooch, line 523: Label needs to be updated here

dojo/filters.py
```diff
@@ -537,13 +544,17 @@ class FindingTagFilter(DojoFilter):
...
-        help_text="Search for tags on a Product that contain a given pattern, and exclude them",
+        help_text=labels.ASSET_FILTERS_ASSETS_WITHOUT_TAGS_HELP,
         exclude=True)
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.form.fields["test__engagement__product__tags"].help_text = labels.ASSET_FILTERS_TAGS_FILTER_HELP
+        self.form.fields["not_test__engagement__product__tags"].label = labels.ASSET_FILTERS_ASSETS_WITHOUT_TAGS_LABEL
+        self.form.fields[
+            "not_test__engagement__product__tags"].help_text = labels.ASSET_FILTERS_ASSETS_WITHOUT_TAGS_HELP
```
Maffooch: Label here too

dojo/product/signals.py
```diff
@@ -8,34 +8,36 @@
...
-                            title=instance.name,
-                            product=instance,
-                            url=reverse("view_product", args=(instance.id,)),
-                            url_api=reverse("product-detail", args=(instance.id,)),
-                        )
+                                title=instance.name,
+                                product=instance,
+                                url=reverse("view_product", args=(instance.id,)),
+                                url_api=reverse("product-detail", args=(instance.id,)),
+                            )
```
Maffooch: Correcting indention oops
```suggestion
                            title=instance.name,
                            product=instance,
                            url=reverse("view_product", args=(instance.id,)),
                            url_api=reverse("product-detail", args=(instance.id,)),
                        )
```

dojo/settings/settings.dist.py
```diff
@@ -484,7 +484,7 @@ def generate_url(scheme, double_slashes, user, password, host, port, path, param
 # 'axes.backends.AxesModelBackend',
 # ]
 
-ROOT_URLCONF = "dojo.urls"
+ROOT_URLCONF = "dojo.v3_migration.urls"
```
Maffooch: The `dojo.urls` file already imports all of the v3 paths, what is the reasoning for changing the root here?

dojo/settings/settings.dist.py
```diff
@@ -326,6 +326,8 @@
     # For HTTP requests, how long connection is open before timeout
     # This settings apply only on requests performed by "requests" lib used in Dojo code (if some included lib is using "requests" as well, this does not apply there)
     DD_REQUESTS_TIMEOUT=(int, 30),
+    # Dictates if v3 org/asset relabeling (+url routing) will be enabled
+    DD_ENABLE_V3_ORGANIZATION_ASSET_RELABEL=(bool, True),
```
Maffooch: Let's start with this off for now, and then we can cut over in the near future 
```suggestion
    DD_ENABLE_V3_ORGANIZATION_ASSET_RELABEL=(bool, False),
```

Maffooch (approved):

valentijnscholten (approved):

PR #13157, opened 2025-09-11
Ruff: Add PLC1901

manuel-sommer:


manuel-sommer:
Would be awesome if we could merge this soonish @mtesauro because of the number of touched files, I don't have to rebase often then.

mtesauro (approved):
Approved

PR #13169, opened 2025-09-12
Auditlog: Add django-pghistory as audit log (optional for now)

### LINKED ISSUE #10835: Audit logs grows indefinetly (other suggested solutions didn't work for me)
The `auditlog_logentry.ibd` grows forever (as already reported in #9153 ).

However, I tried both the solutions suggested (setting the `DD_AUDITLOG_FLUSH_RETENTION_PERIOD` environment variable didn't have any effect and running the `docker exec -it django-defectdojo-uwsgi-1 ./manage.py auditlogflush --yes --before-date 2023-01-01` gets stuck).

My DefectDojo instance is running in docker deployment and version is v. 2.36.4 ( release mode ).

How can we delete old audit logs? Also, if we don't care about audit logs, can we just periodically delete the entire `auditlog_logentry.ibd` as a workaround ? (of course if this won't break anything..).

Any help is appreciated, thanks!

valentijnscholten:
### Summary
As part of evaluating and improving import/reimport performance we have selected [django-pghistory](https://django-pghistory.readthedocs.io/) as the new auditlog for Defect Dojo.

It is entirely based on Postgres triggers to maintain a history/audit log. This makes it run faster and offers new features such as reverting a record to an old version, adding richer context to processes and events, finding events on related records, etc.
This PR doesn't introduce new features yet, it just adds `django-pghistory`  as replacement for `django-auditlog`.

This PR allows still for using `django-auditlog` as auditlog. There's also some builds published to Docker Hub: https://hub.docker.com/r/defectdojo/defectdojo-django/tags?name=pghistory

In a future release `dhango-pghistory` will be the auditlog implementation and `django-auditlog`  will no longer be available for auditlogging. 

### HowTo

To switch an instance over to `django-pghistory`, some steps are needed:
- set `DD_AUDITLOG_TYPE` to `django-pghistory` 
- run `docker compose exec -it uwsgi ./manage.py pghistory backfill` to load the initial snapshot of data into `django-pghistory`.  (this is NOT a data migration from `django-auditlog`, but an initial snapshot from the current data in Defect Dojo).

In the future both will be part of a new release which will perform the backfill automatically. 
Once switched over, you cannot switch back (unless you know what you're doing).

The action history pages will always display the data from both `django-pghistory`  and ` django-auditlog`. As these data formats are completely different, there are two tables on the action history page.

### Some notes about the implementation
- `django-pghistory`  works with database triggers these are created as part of a migration regardless of which audit log is configured.
- the entrypoint script will enabled/disable the triggers based on the chosen auditlog type. this can not be done in models.py, apps.py or other Django lifecycle events as it was always to early or too late.
- models are registered with `django-auditlog` depending on the chosen auditlog type in settings.

image

image

### Performance
I did a couple of test runs to compare both audit log types using the [JFrog Unified Very Many](https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/scans/jfrog_xray_unified/very_many_vulns.json) sample scan that contains ~14000 findings. These tests runs show a 20-30% speed improvement on my laptop running docker compose. In a (production) environment with an external database the difference might be bigger due to the increased latency. `django-pghistory` runs inside the database so there are a lot less network roundtrips needed.

### Scaling
`djang-pghistory` documents two settings or design decisions that affect performance: https://django-pghistory.readthedocs.io/en/3.4.3/performance/

- Row level vs Statement level triggers.
Defect Dojo doesn't do (m)any bulk inserts/updates on the tracked models. So there's no benefit now to switch to Statement level triggers. 
If needed we can switch in the future using a one-time schema-only migration.

- Denormalizing Context
Context is stored in its own table and events have a foreign key relation to this table. If the Context is intensively used and updated during processes/requests and lots of parallel requests are happening that trigger pghistory events, there could be some contention on that Context table. To "solve"  this we could choose to store the context in a column in the Event tables. I don't think the typical use-case in Defect Dojo would really benefit from this, it could be detrimental as it would require more storage and filtering/extracting data from the context might be slower.
If needed we can switch in the future using a one-time schema+data migration.

- Indices
By default `djang-pghistory` doesn't create much indices, only a couple of ones we don't need. Changes made:
- Remove indices on foreign key fields in the event tables (i.e. user_id, found_by, etc). We don't query on those.
- Add index on fields that we query on like `pgh_created`, `pgh_label`.
- Add index on `obj_id` field used in joins/queries
- Add index on fields from the json context, i.e. `user`, `remote_addr`, `url`

### Flushing the Auditlog
This PR also optimize the existing auditlog flushing (fixes #10835):
- Use a `timestamp` instead of a `date` for filtering the entries to be deleted so we can use the database index
- Batch the deletes in batches to avoid deleting millions of records at once
- Add a management command to flush the audit log

###  Next steps
When testing is succesfull and we want to make `django-pghistory` the default, we need to do the following in a follow up PR

- Add a migration that performs the backfill. Leave the management command in place as it provides a way for users to perform the backfill manually. The migration code will skip records that already have an `initial_import` record
- Remove registering models for `django-auditlog`
- Remove registering models for `django-auditlog` from MIDDLEWARE
- Remove registering models for `django-auditlog` installed apps (not sure, it might need to remain to be able to load LogEntry model)

### DIFF
--- a/dojo/product/signals.py
+++ b/dojo/product/signals.py
@@ -13,6 +13,7 @@ from dojo.labels import get_labels
 from dojo.models import Product
 from dojo.notifications.helper import create_notification
 from dojo.pghistory_models import DojoEvents
+from dojo.utils import get_current_user
 
 labels = get_labels()
 
@@ -61,9 +62,14 @@ def product_post_delete(sender, instance, **kwargs):
                 ).order_by("-id").first():
                     user = le.actor
 
+            if not user:
+                current_user = get_current_user()
+                user = current_user
+
             # Update description with user if found
             if user:
-                description = labels.ASSET_DELETE_WITH_NAME_WITH_USER_SUCCESS_MESSAGE % {"name": instance.name, "user": le.actor}
+                description = labels.ASSET_DELETE_WITH_NAME_WITH_USER_SUCCESS_MESSAGE % {"name": instance.name, "user": user}
+
         create_notification(event="product_deleted",  # template does not exists, it will default to "other" but this event name needs to stay because of unit testing
                             title=_("Deletion of %(name)s") % {"name": instance.name},
                             description=description,
--- a/dojo/product_type/signals.py
+++ b/dojo/product_type/signals.py
@@ -70,7 +70,8 @@ def product_type_post_delete(sender, instance, **kwargs):
 
             # Update description with user if found
             if user:
-                description = labels.ORG_DELETE_WITH_NAME_SUCCESS_MESSAGE % {"name": instance.name}
+                description = labels.ORG_DELETE_WITH_NAME_WITH_USER_SUCCESS_MESSAGE % {"name": instance.name, "user": user}
+
         create_notification(event="product_type_deleted",  # template does not exists, it will default to "other" but this event name needs to stay because of unit testing
                             title=_("Deletion of %(name)s") % {"name": instance.name},
                             description=description,
--- a/unittests/test_importers_performance.py
+++ b/unittests/test_importers_performance.py
@@ -198,9 +198,9 @@ class TestDojoImporterPerformance(DojoTestCase):
         configure_pghistory_triggers()
 
         self.import_reimport_performance(
-            expected_num_queries1=646,
+            expected_num_queries1=559,
             expected_num_async_tasks1=10,
-            expected_num_queries2=599,
+            expected_num_queries2=491,
             expected_num_async_tasks2=22,
             expected_num_queries3=284,
             expected_num_async_tasks3=20,
@@ -244,9 +244,9 @@ class TestDojoImporterPerformance(DojoTestCase):
         testuser.usercontactinfo.save()
 
         self.import_reimport_performance(
-            expected_num_queries1=646,
+            expected_num_queries1=559,
             expected_num_async_tasks1=10,
-            expected_num_queries2=604,
+            expected_num_queries2=496,
             expected_num_async_tasks2=22,
             expected_num_queries3=289,
             expected_num_async_tasks3=20,
@@ -293,9 +293,9 @@ class TestDojoImporterPerformance(DojoTestCase):
         self.system_settings(enable_product_grade=True)
 
         self.import_reimport_performance(
-            expected_num_queries1=647,
+            expected_num_queries1=560,
             expected_num_async_tasks1=11,
-            expected_num_queries2=605,
+            expected_num_queries2=497,
             expected_num_async_tasks2=23,
             expected_num_queries3=290,
             expected_num_async_tasks3=21,


docker-compose.yml
```diff
@@ -53,6 +53,7 @@ services:
       DD_SECRET_KEY: "${DD_SECRET_KEY:-hhZCp@D28z!n@NED*yB!ROMt+WzsY*iq}"
       DD_CREDENTIAL_AES_256_KEY: "${DD_CREDENTIAL_AES_256_KEY:-&91a*agLqesc*0DJ+2*bAbsUZfR*4nLw}"
       DD_DATABASE_READINESS_TIMEOUT: "${DD_DATABASE_READINESS_TIMEOUT:-30}"
+      DD_AUDITLOG_TYPE: "${DD_AUDITLOG_TYPE:-django-pghistory}"
```
Maffooch: We should default to just audit-log for now, and then convert later on
```suggestion
      DD_AUDITLOG_TYPE: "${DD_AUDITLOG_TYPE:-django-auditlog}"
```

docker-compose.yml
```diff
@@ -74,6 +75,7 @@ services:
...
```
Maffooch: We should default to just audit-log for now, and then convert later on
```suggestion
      DD_AUDITLOG_TYPE: "${DD_AUDITLOG_TYPE:-django-auditlog}"
```

docker-compose.yml
```diff
@@ -94,6 +96,7 @@ services:
...
```
Maffooch: We should default to just audit-log for now, and then convert later on
```suggestion
      DD_AUDITLOG_TYPE: "${DD_AUDITLOG_TYPE:-django-auditlog}"
```

docker-compose.yml
```diff
@@ -115,6 +118,7 @@ services:
...
     volumes:
```
Maffooch: We should default to just audit-log for now, and then convert later on
```suggestion
      DD_AUDITLOG_TYPE: "${DD_AUDITLOG_TYPE:-django-auditlog}"
    volumes:
```

valentijnscholten:
> We should default to just audit-log for now, and then convert later on

That was the idea, but as mentioned in the description above it's non-trivial to make everything work in a code base where both `django-auditlog` and `django-pghistory` are available and we want to have a working toggle. Flipping the toggle to default to `django-auditlog` would break the unit test around delete notifications. 

I spent some more time on this and found out it's caused by our `post_delete` signal now running *before& the `django-auditlog` signal running. So we cannot retrieve the `actor` information from the auditlog event. I am not sure if that approach was even a valid one from the beginning, but that's how it's done in the signal. I now changed the signal to just use the `get_current_user` method, which is also used by our AuditLogMiddleware under the hood. The test now passes.

However, I expect the `get_current_user` to return `None` for when `ASYNC_OBJECT_DELETE` is enabled. The options that I see are:

1) Accept that when using ASYNC_OBJECT_DELETE the notifications sent by Defect Dojo do not contain the user/actor who deleted the object
2) Change our `DojoAsyncTask` annotation to propagate the current user into the celery tasks. I tried this, but it's a bigger change probably and would affect the call chain of tasks. I think this would be too big a change to embed in this auditlog PR. 
3) Remove `django-auditlog` from this branch and just make the switch to `django-pghistory`. We would need do add a migration and we would be good to go a new release or for instances to use this build. We could keep the branch in sync with `master` for a while until we add the PR to a release.

When using `django-pghistory` the problem goes away because the auditlog is created in-line by postgres so the actor/user information will be available when our `post_delete` signal is called and the info is retrieved from the pghistory entries as it was before.

I think option 2 is something to look at later in a separate PR. I think we will need it at some point if we push more work into the celeryworkers. Maybe Pro is already doing this?

Option 1 is the easiest and most flexible. There are no other signals depending on presence of auditlog entries. Still there's a chance of something not working around auditlogs. The chance is small since we have unit tests.
Option 3 is the safest as it would simplify the codebase a lot.

We could to option 1 for now so we can easily test on the Demo instances and maybe some dev/test/staging instances for Pro. Once everything looks good we could transform the branch or a new PR to implement option 3.

I've updated the GHA unit tests to run the test suite against both auditlog types.

Maffooch:
> We could to option 1 for now so we can easily test on the Demo instances and maybe some dev/test/staging instances for Pro. Once everything looks good we could transform the branch or a new PR to implement option 3.

This sounds very reasonable to me. Thank you for providing a very detailed to write up to make the conclusion very easy to grok

dogboat (approved):

PR #13190, opened 2025-09-16
fix(helm): Update based following #13146

kiblik:
#12984 was created and not rebased after #13146
So now the docs need to be updated


valentijnscholten (approved):

Maffooch (approved):

kiblik:
Not needed anymore, fixed in #13177

PR #13191, opened 2025-09-16
feat(helm): Trigger update of Helm docs when version is changed

kiblik:
When the version is updated, HELM docs become outdated and need to be adjusted.


mtesauro (approved):
Approved

rossops (approved):

PR #13211, opened 2025-09-17
make close_old_findings tooltip clearer when service is not set

### LINKED ISSUE #13163: `close_old_findings` without service does not close all findings
https://github.com/DefectDojo/django-DefectDojo/blob/228d0d5f6809c94e61e06a42abc55a85551ad32f/dojo/forms.py#L562

> If service has been set, only the findings for this service will be closed.

I believe this behavior is not accurate, but I'm not sure which one is the intended behavior - the documented one or the real one.

My understanding of this is:
* if service is set during import, only findings matching service (and not in import) will be closed
* if service is NOT set during import, ANY findings (not in import) will be closed

I haven't checked code, but that is not exactly what happened:
* All findings have a service defined
* I import *without* a service
* Findings not present in the report are *not* closed

I believe this happens because `service` is part of the "hash code always fields", so matching will always consider `service`. If service is empty, it will be matching findings without service instead of *ignoring* service.

My question being:
* Is my interpretation of tooltip correct? 
* If so, is tooltip wrong or behavior wrong?

> Note: I don't have any use case where I want to import "blank" service and have it close all findings, it was a mistake I did in an import and noticed the inconsistent behavior. I believe fixing tooltip is the easiest option as the fixing behavior has more work and impact.

### update

Considering the code in https://github.com/DefectDojo/django-DefectDojo/blob/228d0d5f6809c94e61e06a42abc55a85551ad32f/dojo/importers/default_importer.py#L305 it looks explicit that the behavior is desired, so I guess tooltip could be updated

Maybe 🤷 ?

> If service is not set, only findings without a service will be closed.

Also, reimporter has it differently (does not catch the `NULL` case that might be relevant in many DB systems and/or configurations), maybe they should reuse a method `get_old_findings()` to avoid getting the criteria out of sync between both:
https://github.com/DefectDojo/django-DefectDojo/blob/228d0d5f6809c94e61e06a42abc55a85551ad32f/dojo/importers/default_reimporter.py#L171

I can make a PR, just looking for feedback first 👍

fopina:
**Description**

This fixes #13163 

> Old findings no longer present in the new report get closed as mitigated when importing. If service has been set, only the findings for this service will be closed. 

The current tooltip misled me into thinking that `when service is NOT set, all findings are considered regardless of service`, yet it is not the behavior: when service is not set, only findings without a service will be closed.

I thought it could be just me and my english but at least [ChatGPT agrees](https://chatgpt.com/share/68cae627-0c30-800c-99ac-e07e6e922176) 😄 

Tried to come up with a smaller change but in the end I think just making it explicit does not make it too big anyway.

Screenshot 2025-09-17 at 23 00 55

**Checklist**

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -2314,14 +2314,16 @@ class ImportScanSerializer(CommonImportScanSerializer):
         required=False,
         default=False,
         help_text="Old findings no longer present in the new report get closed as mitigated when importing. "
-                    "If service has been set, only the findings for this service will be closed. "
+                    "If service has been set, only the findings for this service will be closed; "
+                    "if no service is set, only findings without a service will be closed. "
                     "This only affects findings within the same engagement.",
     )
     close_old_findings_product_scope = serializers.BooleanField(
         required=False,
         default=False,
         help_text="Old findings no longer present in the new report get closed as mitigated when importing. "
-                    "If service has been set, only the findings for this service will be closed. "
+                    "If service has been set, only the findings for this service will be closed; "
+                    "if no service is set, only findings without a service will be closed. "
                     "This only affects findings within the same product."
                     "By default, it is false meaning that only old findings of the same type in the engagement are in scope.",
     )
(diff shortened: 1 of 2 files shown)


Jino-T (approved):

paulOsinski (approved):

paulOsinski:
Nice work - I will work on adding this to our docs as well.

Issue #13161, opened 2025-09-12
Drop django-tagging as dependency


fopina:
https://github.com/DefectDojo/django-DefectDojo/blob/228d0d5f6809c94e61e06a42abc55a85551ad32f/requirements.txt#L21

Can some custom migration or check be added to cover cases of people that still need this (old version), pointing them towards a middle-step upgrade first?

In some CIs internet access is limited and there are internal proxies to popular package managers (such as pypi), flagging/failing `git+http` dependencies.

I'm not familiar with the reason why this is kept: based on comment I can only assume that django-tagging was replaced with tagulous and some migration is moving the data.
If that's the case, that migration could be updated to just throw an exception (keeping the same name) directing owner to a previous version before running that one (if there is any data to migrate, to cover for new setups).

I'm happy to push a PR if that's the case, so it can be considered. Or to push a PR on the django-tagging fork to publish it to pypi under some "forked" package name, if that's preferred (though cleanup would sound better to me)

valentijnscholten:
This is on our radar and could be something we do for v3. If we want to do it sooner we could:
- add to the upgrade notes of v2.0.0. that users need to upgrade to v1.11.0 first to make sure tags are migrated already
- remove the data migration logic and requirements entry in the next monthly release.

Let me check with the team first before you make a PR.

valentijnscholten:
@fopina We would welcome a PR for this 👍

fopina:
Tried to come up with some strict testing but it's impossible to install Dojo v1.10 at this point - too many dependency issues...
I've created #13216 - let me know what you think!

PR #13216, opened 2025-09-18
Drop django-tagging as dependency

fopina:
**Description**

This PR addresses https://github.com/DefectDojo/django-DefectDojo/issues/13161

django-tagging has been replaced with tagulous long time ago, in #3333, including a migration (0066) to copy the tags over. All references later removed in #4419 (migration 0093)
Both of these are merged even before v2.0.0.

I believe this dependency can be safely removed at this point.

This PR removes it from `requirements.txt` and removes the data-migration from 0066:
* New setups do not even run 0066 as there is a squashed 1-90 (post-v2)
* The squashed 1 to 90 does not include data migration
* The fields are removed in 93, which applies whether squashed is used or not

Release notes v2 updated to highlight the new extra step, for those who haven't updated Dojo in over 5 years (if anyone).

**Test results**

Only test assertion done is that unit tests execute properly, including running all migrations on a clean database.
I've also temporarily removed the squashed 1-90 migration to make sure 66 was executed without any issues and it is.

I've tried adding some logic to the migration 0066 as in to check whether tags had been migrated or not (and fail otherwise), but I've been unable to "go back" to a version actively using django-tagging.

I have honestly tried in [this branch](https://github.com/fopina/django-DefectDojo/tree/drop-django-tagging/old-branch) of Dojo v1.11 and fought for quite some time with invalid debian packages and dependency issues (with conflicting requirements due to new versions and lack of transitive dependency pinning).

I ended up realizing that migration 66 (or any other) does not actually delete "tagging" data, so the check I had imagined wouldn't help anyway so I gave up trying this realistic (yet unlikely) scenario

**Documentation**

[release notes for 2.0](docs/content/en/open_source/upgrading/2.0.md) updated

**Checklist**

### DIFF
--- a/docker/sample_data/initial_dojo_data.json
+++ b/docker/sample_data/initial_dojo_data.json
@@ -233,22 +233,6 @@
   },
   {
     "fields": {
-      "model": "tag",
-      "app_label": "tagging"
-    },
-    "model": "contenttypes.contenttype",
-    "pk": 39
-  },
-  {
-    "fields": {
-      "model": "taggeditem",
-      "app_label": "tagging"
-    },
-    "model": "contenttypes.contenttype",
-    "pk": 40
-  },
-  {
-    "fields": {
       "expire_date": "2016-09-10T12:40:25.483Z",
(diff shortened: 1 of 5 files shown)


blakeaowens (approved):

valentijnscholten (approved):
Thanks for the PR.

The reason why this data migration (and others) are not present in the squashed migration is:

https://github.com/DefectDojo/django-DefectDojo/blob/68821a8166253869918412fb045b9d6e3a4d9437/dojo/db_migrations/0001_squashed_0090_index_duplicate_finding.py#L20-L37

I think the idea of the squash was that at some point we can remove individual migrations 0001-0090. But dropping `django-tagging` is a good start.

fopina:
> New setups do not even run 0066 as there is a squashed 1-90 (post-v2)

@valentijnscholten I meant here to highlight the low impact of the change, not to question the decision to not include it 👍 no point having data migrations if there is no data 😄 

rebased 🤞

valentijnscholten (approved):

PR #13217, opened 2025-09-18
fix dojo_testdata fixture

valentijnscholten:
During #13182 I ran into some obscure fixture data loading errors. Turns out that by using a Django TransactionTestCase, it can leave data behind or put the (test) database into a mode where it won't accept any invalid fixtures.
Turns out that the `dojo_testdata.json` fixture we're using is missing some data and has some out-of-sync contenttypes.
With the changes in this PR the fixture loads fine again. I decided to submit it as a PR on its own.

### DIFF
--- a/dojo/fixtures/dojo_testdata.json
+++ b/dojo/fixtures/dojo_testdata.json
@@ -1,5 +1,78 @@
 [
   {
+    "model": "dojo.role",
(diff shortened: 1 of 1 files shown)


valentijnscholten:
giving up for now.

PR #13220, opened 2025-09-18
do not build/start unused services in unit test docker compose files

fopina:
**Description**

I've often used `deploy.replicas: 0` in compose override files to "disable" some of the services. Yet, that just prevents them from starting, they're still built.

Recently, I've seen "unused profile" being used for the same purpose, with the added benefit that this way services images are not built either.

This PR applies that to the unit test compose overrides.

**Test results**

`docker compose build --no-cache` (with unit_test override set) takes around 120/140 seconds in current file. In this branch, it takes 100/110 seconds (and doesn't start unused services either).

**Checklist**

### DIFF
--- a/docker-compose.override.unit_tests.yml
+++ b/docker-compose.override.unit_tests.yml
@@ -1,10 +1,6 @@
 ---
 services:
-  nginx:
-    image: busybox:1.37.0-musl@sha256:254e6134b1bf813b34e920bc4235864a54079057d51ae6db9a4f2328f261c2ad
-    entrypoint: ['echo', 'skipping', 'nginx']
-    volumes:
-      - defectdojo_media_unit_tests:/usr/share/nginx/html/media
+  nginx: !reset
   uwsgi:
     build:
       target: django-unittests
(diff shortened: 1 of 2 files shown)


valentijnscholten:
Can you rebase/merge?

Is this "not used" profile something built in to docker compose? Can't seem to find anything about it? Or is the idea that if you do not specify a profile, any service with a profile will not be built/started?

fopina:
> Or is the idea that if you do not specify a profile, any service with a profile will not be built/started?

This one!
But "if you do not specify one of those profiles, service will not be built/started", just to include the case "if you specify a profile but it does not match"

fopina:
Actually just tried `initializer: !reset {}` instead and also works.

Would you prefer? Looks less hacky

valentijnscholten:
That looks cleaner indeed.

What does this part do:

```
    depends_on: !override
      postgres:
        condition: service_started
```
Is it to get rid of the `depends_on` for `redis` and `initializer`?

fopina:
It does look much cleaner. Should've thought of it from the start when I read about it together with `override` 🤦 

And yes, to remove dependency on non-existing services.

Default compose override behavior is to merge dictionaires. `!reset` just cleans it up, `!override` replaces it with new definition.

reset removes the yaml entry, override replaces with new one (instead of merging)

Issue #13218, opened 2025-09-18
blackduck_api using deprecated datetime.utcnow()


valentijnscholten:
This warning is printed in the logs as ERROR, but the test suite continues, but fails eventually. 

```
uwsgi-1  | [18/Sep/2025 18:13:46] ERROR [dojo.tools.factory:130] failed to load api_blackduck
uwsgi-1  | Traceback (most recent call last):
uwsgi-1  |   File "/app/dojo/tools/factory.py", line 119, in <module>
uwsgi-1  |     module = import_module(f"dojo.tools.{module_name}.parser")
uwsgi-1  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
uwsgi-1  |   File "/usr/local/lib/python3.12/importlib/__init__.py", line 90, in import_module
uwsgi-1  |     return _bootstrap._gcd_import(name[level:], package, level)
uwsgi-1  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
uwsgi-1  |   File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
uwsgi-1  |   File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
uwsgi-1  |   File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
uwsgi-1  |   File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
uwsgi-1  |   File "<frozen importlib._bootstrap_external>", line 999, in exec_module
uwsgi-1  |   File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
uwsgi-1  |   File "/app/dojo/tools/api_blackduck/parser.py", line 5, in <module>
uwsgi-1  |     from .importer import BlackduckApiImporter
uwsgi-1  |   File "/app/dojo/tools/api_blackduck/importer.py", line 5, in <module>
uwsgi-1  |     from .api_client import BlackduckAPI
uwsgi-1  |   File "/app/dojo/tools/api_blackduck/api_client.py", line 1, in <module>
uwsgi-1  |     from blackduck import Client
uwsgi-1  |   File "/usr/local/lib/python3.12/site-packages/blackduck/__init__.py", line 2, in <module>
uwsgi-1  |     from .HubRestApi import HubInstance
[...]

PR #13225, opened 2025-09-19
Unit Tests: Ignore deprecation Warning from BlackDuck pypi package

Maffooch:
Unit tests are failing due to a deprecation warning in the black duck pypi package. The library does not appear to be very active these days, so ignoring the warning feels like the quickest path to getting our tests back on track

### DIFF
--- a/dojo/tools/api_blackduck/api_client.py
+++ b/dojo/tools/api_blackduck/api_client.py
@@ -1,3 +1,8 @@
+import warnings
+
+# Ignore DeprecationWarning from the blackduck library because we cannot do anything about it
+# until they fix it in their library
+warnings.filterwarnings("ignore", category=DeprecationWarning)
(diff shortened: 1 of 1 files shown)


blakeaowens (approved):

dogboat (approved):

valentijnscholten (approved):
fixes #13218

fopina:
Last 2 days I've been repushing same PRs as it would pass sometimes 😆 , thanks!

But is it ok to suppress deprecation warnings entirely (for whole dojo)? Or should it be changed to suppress only these calls?

Maffooch:
Oh, I didn't realize that is how the warning system worked. Would this work?
```python
with warnings.catch_warnings(action="ignore", category=DeprecationWarning):
    from blackduck import Client  # noqa: E402
```

PR #13227, opened 2025-09-19
debug toolbar: disable by default

valentijnscholten:
The debug toolbar setup currently interferes with running unit tests when in `dev` mode. I'm going to revisit the debug toolbar setup/integration. For now the most practical setting is allow it to be enabled, but disable it by default. Also in the `dev` docker "environment".

Ideally we do not package the static file for the debug toolbar, but this seems hard to accomplish without duplicating the Dockerfile or duplicating many lines within  the Dockerfile. For I think we can accept this as the django image is what will actually run at run-time and that will NOT have the debug-toolbar in the production images.


mtesauro (approved):
Approved

Issue #12918, opened 2025-08-04
tags shown twice inside tags filter fields (on hold)


valentijnscholten:
Reproducible on https://demo.defectdojo.org

1. Goto 
2. Select some tags from the dropdown
3. Observe they are shown correctly (and only once)

<img width="531" height="309" alt="Image" src="https://github.com/user-attachments/assets/973a268b-5df7-4e98-9f1c-97e49d284e53" />

4. Click "Apply filters"
5. Observer page refresh, open filter section again, observer tags are now duplicated/shown twice

<img width="449" height="281" alt="Image" src="https://github.com/user-attachments/assets/1d546bed-3c4c-4f19-991d-970edc03696b" />

<img width="426" height="354" alt="Image" src="https://github.com/user-attachments/assets/aa690228-12b9-44a3-bf4b-d6264b1611c6" />

There's some javascript running to initialize these Selectv4 fields. Looks like it might be running twice as also the entries in the dropdown are duplicated.

Please note this a different issues from what is reported in #12601. But it might actually have the same root cause

fopinappb:
Same as @valentijnscholten, I've been bugged by the duplicates in product/finding filter screen and this _looks_ like same issue of #13228 but it did not cover it.

Trying to look into that PR, it's re-defining `tags` form fields as `tagulous.form.TagField` instead of leaving the default, but unsure how to apply that to the filter fields.

Also, it seems weird that a model `tagulous.models.TagField` does not become `tagulous.form.TagField` by default

I'll try to look into it

fopina:
Slight update: it does become a form.TagField by default and the fix in that PR actually breaks autocomplete (I've updated there).
I'll continue to look for another way to fix the duplicated rendering

fopinappb:
https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/templates/dojo/filter_js_snippet.html#L18

this line instantly renders it duplicated. According to tagulous docs, this line should not be needed as tagulous already does it, but if we comment it, it's not rendered either.
Once applied, rendered twice.

Might be some other hook that applies select2 again but I haven't found it yet, sharing in case that rings any bell 😄

valentijnscholten:
The explicit initalization is needed because we defer it in the select2 setup: https://django-tagulous.readthedocs.io/en/latest/forms.html#select2-version-4

I tried various front-end changes to avoid what seems like a duplicate initialization, but have not found the solution yet.

[...]

PR #13228, opened 2025-09-19
Tags: Remove duplicates in edit forms

Maffooch:
When editing objects with tags, each entry was being duplicated:

image

Adding a specific definition of the `TagField` in the form classes for each object has something to do with the way the form is rendered automatically by django


mtesauro (approved):
Approved

valentijnscholten:
WowI think this fixes #12918 which has been bugging me for months!

valentijnscholten:
I do still see it happening in filter forms:
image

rossops (approved):

fopina:
I was looking into why "redefining" tags as a `TagField` (when it was already) caused a different behavior and the difference is that with this change field `autocomplete_tags` is set to `None`
That prevents duplication but it also prevents tags from being listed (unless we're looking at the product that adds it)

Steps:
* Go to https://demo.defectdojo.org/product/3/edit and add tag `whatever` to it (save)
* Go to https://demo.defectdojo.org/product/4/edit and click on the tag field - before this PR, it would display tags already in use somewhere else (`whatever`), now it's empty list (`No results found`)

PR #13231, opened 2025-09-21
k8s-tests: bump k8s and minukube

valentijnscholten:
k8s-tests: bump k8s and minukube


Maffooch:
Merging with two since this is a dependency update

PR #13241, opened 2025-09-22
Unit Testing: Do no run in debug mode in order to reduce logging

Maffooch:
Viewing a failed test in the GHA logs is not very helpful without rendering the raw view because there are 50k+ lines of debug logs. Reducing the logging verbosity would make the tests quite a bit faster as well as finding failures easier in the GHA UI

### DIFF
--- a/docker-compose.override.unit_tests_cicd.yml
+++ b/docker-compose.override.unit_tests_cicd.yml
@@ -14,7 +14,7 @@ services:
       - "defectdojo_media_unit_tests:${DD_MEDIA_ROOT:-/app/media}"
     environment:
       PYTHONWARNINGS: error  # We are strict about Warnings during testing
-      DD_DEBUG: 'True'
+      DD_DEBUG: 'False'
       DD_LOG_LEVEL: 'ERROR'
       DD_TEST_DATABASE_NAME: ${DD_TEST_DATABASE_NAME:-test_defectdojo}
       DD_DATABASE_NAME: ${DD_TEST_DATABASE_NAME:-test_defectdojo}


valentijnscholten:
When I look at the output of tests the biggest generators of output are the vcr related tests. Maybe a rerecording will reduce the output? There's also some dedupe and false positives history tests which put out lots of logging. Maybe those log statements can be commented out.

Maffooch:
Yep - they are debug logging the entire request 🥴 I would prefer to keep those logs in case someone needs them. not sure why the tests are ignoring the log level. Will make more time for this soon

valentijnscholten (commented):
Should we put a few lines in the docker readme on how to get DEBUG log output during unit tests?

Maffooch:
sure thing

Maffooch:
> Should we put a few lines in the docker readme on how to get DEBUG log output during unit tests?

Turns out we have two docker compose override files: one for [GHA](https://github.com/DefectDojo/django-DefectDojo/blob/40d1b8798efe1d8bbc0302ec8ba7e00070f6479b/docker-compose.override.unit_tests_cicd.yml) and one for [locally running tests](https://github.com/DefectDojo/django-DefectDojo/blob/03ce8aba7affb67c26a7160e457141e8cf16fb34/docker-compose.override.unit_tests.yml) that is still running in debug mode. The [override specified in the docker docs](https://github.com/DefectDojo/django-DefectDojo/blob/ebdde07c3c5e496e273de2607f15f95fa487b9f6/readme-docs/DOCKER.md#L295-L302) is the latter option with debug mode enabled by default. I think we are fine to proceed with this PR as is

valentijnscholten:
> sure thing

Sure Thing as in "Yes, I (Cody) will do that" or as in "I (Cody) happily wait for someone to do that". 

I can try to deduce it from the PR, but not sure if both changes are needed and what their side effects are.

valentijnscholten (approved):

PR #13256, opened 2025-09-24
Added Line Field for SonarQube Scan and SonarQube Scan Detailed Scan Types

Jino-T:
[sc-11829]

I added the line field everywhere relevant except for [this location](https://github.com/Jino-T/django-DefectDojo/blob/7c0f3f0252c7a76d5c5863ca186360c7c3856d2d/dojo/tools/sonarqube/soprasteria_helper.py#L77). I trusted the comment left here. Let me know if I should overrule this.


Maffooch (dismissed):
I got an exception from [this file](https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/scans/sonarqube/findings_over_api.json) because the line number was a string when it should be an integer

mtesauro (approved):
Approved

Maffooch (approved):

rossops (approved):

PR #13260, opened 2025-09-25
Fix: Apply unsaved tags to endpoints during import

Lexorn-ADT:
Tags were not being applied to endpoints when importing in "add_endpoints_to_unsaved_finding" (endpoint_manager.py) due to missing code.

### DIFF
--- a/dojo/importers/endpoint_manager.py
+++ b/dojo/importers/endpoint_manager.py
@@ -41,6 +41,9 @@ class EndpointManager:
                     query=endpoint.query,
                     fragment=endpoint.fragment,
                     product=finding.test.engagement.product)
+                if hasattr(endpoint, 'unsaved_tags') and endpoint.unsaved_tags:
+                    logger.debug(f"Applying tags to endpoint {ep.host}: {endpoint.unsaved_tags}")
+                    ep.tags.add(*[tag.lower() for tag in endpoint.unsaved_tags])
             except (MultipleObjectsReturned):
                 msg = (
                     f"Endpoints in your database are broken. "
--- a/dojo/importers/endpoint_manager.py
+++ b/dojo/importers/endpoint_manager.py
@@ -43,7 +43,7 @@ class EndpointManager:
                     product=finding.test.engagement.product)
                 if hasattr(endpoint, 'unsaved_tags') and endpoint.unsaved_tags:
                     logger.debug(f"Applying tags to endpoint {ep.host}: {endpoint.unsaved_tags}")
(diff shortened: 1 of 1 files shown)


Maffooch (commented):
I had never considered this use case before. Are there parsers today the attempt to implement this behavior? If not, do you have a good example of where endpoint tags can be ingested at import/reimport time?

Lexorn-ADT:
Hi! This feature seems very similar to the "unsaved tags" functionality for findings, which works as expected. However, my use case might be niche? Let me explain below.
I developed a custom parser that processes vulnerabilities as findings, with devices represented as endpoints. The parser handles .json files where each object represents a vulnerability. I extract specific keys as tags and add them to unsaved_tags. Each object also includes an array of active endpoints, each with its own list of tags. The issue is that while unsaved_tags are supported for both findings and endpoints, they only function correctly for findings during an API import. The UI import works fine for both.
As a workaround, I considered using hostname or userinfo, but since IP addresses can be dynamic, this approach creates new endpoints instead of updating existing ones during re-imports. After examining the code, I discovered that while unsaved_tags for endpoints are stored during API imports, they aren't processed further. By adding two lines of code, I enabled endpoint tag support, and now it works as intended.
Regarding the provided parsers, I haven't found any that utilize endpoint tags (though I haven't reviewed all of them). Could this bug explain their absence? Some parsers do use tags but only for findings.

Code examples:
works in git version:
```
if unsaved_tags:
    finding.unsaved_tags = unsaved_tags
```

Doesn't work in git version:
```
if isinstance(endpoint_tags, list):
    endpoint.unsaved_tags = [t.strip() for t in endpoint_tags if t.strip()]
```

Maffooch:
I took a look around the repo, and the use of `unsaved_tags` on the Endpoint model is not used by anything else in DefectDojo. It looks like the `unsaved_tags` attribute is initialized on the finding before anything happens, so the same may need to be done on the Endpoint model:
https://github.com/DefectDojo/django-DefectDojo/blob/8a4731057ddc2f927d0be807a258b03e9132d103/dojo/models.py#L2724-L2732

In order to accept this PR, unit tests that exercise this code, and check that tags are being created as expected would be required

valentijnscholten:
I think it's a nice plan to support the unsaved_tags on endpoints. Please note the tags need to be cleaned before being added, similar to: 
https://github.com/DefectDojo/django-DefectDojo/blob/b8eee86069059c522e03036f90e0b76ea4e524ca/dojo/importers/default_importer.py#L214

Maffooch:
Marking as this as draft until it is ready for review again

Maffooch:
Closing as stale

PR #13261, opened 2025-09-25
feat(finding list): Add planned_remediation_version

kiblik:
We are missing `planned_remediation_version` in the listing of findings

### DIFF
--- a/dojo/templates/dojo/findings_list_snippet.html
+++ b/dojo/templates/dojo/findings_list_snippet.html
@@ -399,7 +399,10 @@
                                         {% trans "Service" %}
                                     </th>
                                     <th scope="col"> 
-                                        {% trans "Planned Remediation" %}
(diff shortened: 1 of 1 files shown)


manuel-sommer:
This sounds like being almost a duplicate to https://github.com/DefectDojo/django-DefectDojo/pull/12922

kiblik:
> This sounds like being almost a duplicate to #12922

🤔 looks like. In that case, it might be a good idea to rename `planned_remediation_version` to `fix_version` and not create it as a new field.

valentijnscholten:
I don't think these are the same fields.

fix_version is tied to the component, i.e. a library. planned_remediation is tied to the product/asset/service/application and is independent from the versions of any libraries.

Maffooch (approved):

mtesauro (approved):
Approved

Jino-T (approved):

PR #13271, opened 2025-09-26
Similar Findings: Restrict to Product Scope

Maffooch:
The similar finding table allows for manually setting duplicates of findings that deduplication did not fully catch. This is a neat feature, but it also allowed for findings to be a duplicate of a finding from a different product entirely. This not something a naturally occurring circumstance in deduplication, so we should not allow it in similar findings either

### DIFF
--- a/dojo/templates/dojo/view_finding.html
+++ b/dojo/templates/dojo/view_finding.html
@@ -726,7 +726,7 @@
             <div class="panel-heading">
                 <h4 class="has-filters">Similar Findings ({{ similar_findings.paginator.count }})
                     <i class="fa-solid fa-circle-question has-popover" data-trigger="hover"
-                       data-content="Opening this panel shows findings in this Product that are not exact duplicates,
+                       data-content="Opening this panel shows findings in this Asset that are not exact duplicates,
                         but have similar values for Vulnerability Id, CWE, file_path, line number, unique_id_from_tool.
                         It has a filter panel where filtering can be made less or more strict. The resulting findings can be view, marked as duplicate or original.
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

dojo/templates/dojo/view_finding.html
```diff
@@ -728,8 +728,7 @@ <h4 class="has-filters">Similar Findings ({{ similar_findings.paginator.count }}
                     <i class="fa-solid fa-circle-question has-popover" data-trigger="hover"
                        data-content="Opening this panel shows findings that are not exact duplicates,
```
valentijnscholten:
```suggestion
                       data-content="Opening this panel shows findings in this Product that are not exact duplicates,
```

blakeaowens (approved):

dojo/templates/dojo/view_finding.html
```diff
@@ -726,10 +726,9 @@ <h4>Duplicate Cluster ({{ finding|finding_duplicate_cluster_size }})<span class=
             <div class="panel-heading">
                 <h4 class="has-filters">Similar Findings ({{ similar_findings.paginator.count }})
                     <i class="fa-solid fa-circle-question has-popover" data-trigger="hover"
-                       data-content="Opening this panel shows findings that are not exact duplicates,
+                       data-content="Opening this panel shows findings in this Product that are not exact duplicates,
```
Maffooch:
```suggestion
                       data-content="Opening this panel shows findings in this Asset that are not exact duplicates,
```

PR #13285, opened 2025-09-30
performance optimization: add tags in bulk

valentijnscholten:
Tags are a first class citizin in Defect Dojo, so good performance is important.

Currently the `django-tagulous` library doesn't provide a way to "bulk add tags to a collection of model instances".
We've raised a (feature request}[https://github.com/radiac/django-tagulous/issues/190] and (a starter/draft PR)[https://github.com/radiac/django-tagulous/pull/191] upstream, but are unsure if and when this would be merged (or not accepted).

Currently bulk adding of tags happen during:

- imports/reimports
- bulk edit
- product tag inheritance [1]

This results in *lots* of database queries. If we skip the two tags in the performance unit test, we save between ~100 and ~180 queries per run. And that's for a scan report with only 16 findings.

```
valentijn:~/dd$ ./run-unittest.sh --test-case unittests.test_importers_performance 2>&1 | grep Assertion
AssertionError: 592 != 679 : 592 queries executed, 679 expected
AssertionError: 498 != 606 : 498 queries executed, 606 expected
AssertionError: 592 != 679 : 592 queries executed, 679 expected
AssertionError: 503 != 611 : 503 queries executed, 611 expected
AssertionError: 597 != 684 : 597 queries executed, 684 expected
AssertionError: 509 != 617 : 509 queries executed, 617 expected
```

In this context I feel it's justifiable that we add a "bulk add tag(s) to models" method to Defect Dojo.
This PR introduces that method and performs tagging in batches of 1000 findings. It needs 3 queries per batch.
The current upstream implementation needs 3 queries per finding, i.e. 3000 queries per batch.

The PR adds test cases for various scerario's. 
The PR doesn't use (m)any `django-tagulous` internals, mainly Django ORM methods.

I considered creating a fork of `django-tagulous`, but that would add a big maintenance burden.

With the (~10k findings sample file)[https://github.com/DefectDojo/django-DefectDojo/blob/bugfix/unittests/scans/jfrog_xray_unified/very_many_vulns.json] the import time without dedupe gets reducede from ~372s to ~190s (on my laptop). **So almost 50% faster.**

[1] The bulk method is used in import, reimport and bulk edit. Product Tag Inheritance has not been touched yet, it deserves a PR on its own once we feel the bulk tag add is the way forward.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -120,7 +120,7 @@ services:
           source: ./docker/extra_settings
           target: /app/docker/extra_settings
   postgres:
-    image: postgres:18.0-alpine@sha256:9636ae7feacd8d630303eede7f95cd0f472d514e7864422c6aa8ea07b2171df8
+    image: postgres:18.0-alpine@sha256:fadc2b63607986d054ccb7302ca6fedc7bb9b6586d03f2110ae3fbc1ce806693
     environment:
       POSTGRES_DB: ${DD_DATABASE_NAME:-defectdojo}
       POSTGRES_USER: ${DD_DATABASE_USER:-defectdojo}
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,7 @@ Markdown==3.9
 openpyxl==3.1.5
 Pillow==11.3.0  # required by django-imagekit
 psycopg[c]==3.2.10
-cryptography==46.0.1
+cryptography==46.0.2
 python-dateutil==2.9.0.post0
 redis==6.4.0
 requests==2.32.5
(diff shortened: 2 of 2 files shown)


mtesauro (approved):
Approved

dojo/management/commands/import_unittest_scan.py
Maffooch, line 1: @valentijnscholten can you please rename references of "product" to "asset" and "product type" to "organization"

valentijnscholten: I will do this in a separate PR after the performance prs have been merged as multiple of them include this exact same file

Maffooch: Understood - thank you! Will get this merged

PR #13286, opened 2025-09-30
GitHub Secrets Detection Report Parser

Logicmn:
**Description**

Hello again! This PR adds a new parser for GitHub secrets detection uploads.

**Test results**

Added unit tests for new parser. Results:
```
test_file_path_and_line_assignment (unittests.tools.test_github_secrets_detection_report_parser.TestGithubSecretsDetectionReportParser.test_file_path_and_line_assignment)
Test file path and line number extraction ... ok
test_parse_file_invalid_format_raises (unittests.tools.test_github_secrets_detection_report_parser.TestGithubSecretsDetectionReportParser.test_parse_file_invalid_format_raises)
Non-list JSON should raise ... ok
test_parse_file_with_multiple_vulns_has_multiple_findings (unittests.tools.test_github_secrets_detection_report_parser.TestGithubSecretsDetectionReportParser.test_parse_file_with_multiple_vulns_has_multiple_findings)
Multiple entries produce corresponding findings ... ok
test_parse_file_with_no_vuln_has_no_findings (unittests.tools.test_github_secrets_detection_report_parser.TestGithubSecretsDetectionReportParser.test_parse_file_with_no_vuln_has_no_findings)
Empty list should yield no findings ... ok
test_parse_file_with_one_vuln_parsed_correctly (unittests.tools.test_github_secrets_detection_report_parser.TestGithubSecretsDetectionReportParser.test_parse_file_with_one_vuln_parsed_correctly)
Single secret alert entry parsed correctly ... ok
test_severity_assignment (unittests.tools.test_github_secrets_detection_report_parser.TestGithubSecretsDetectionReportParser.test_severity_assignment)
Test severity assignment logic ... ok
test_url_setting (unittests.tools.test_github_secrets_detection_report_parser.TestGithubSecretsDetectionReportParser.test_url_setting)
Test URL assignment from GitHub alert ... ok

Ran 7 tests in 0.054s
```

Ref links:
- GitHub secret scanning API reference: https://docs.github.com/en/rest/secret-scanning/secret-scanning?apiVersion=2022-11-28


dojo/tools/github_secrets_detection_report/parser.py
```diff
@@ -0,0 +1,146 @@
...
+
+            # Determine severity based on state and other factors
+            if state == "resolved":
+                severity = "Info"
+            elif validity == "active" and publicly_leaked:
+                severity = "Critical"
+            elif validity == "active":
+                severity = "High"
+            else:
+                severity = "Medium"
```
valentijnscholten, line 122: I understand why you're doing this, but I am wondering if this will result in desired/predictable behaviour. On reimport it can happen that the severity changes because an alert is resolved or has ben leaked publicy since the previous import. If the severity changes, reimport and the deduplication process will see this is a new finding because the severity is part of the hash_code config. Is it really needed to have the severity in the hash_code config? To me it almost never feels like the right thing to do.

Logicmn: That's a great catch. I see two options for remedying:

1. Statically set the severity to `medium` or `high` always.
2. Remove severity from the dedupe hash_code, and continue to set severity dynamically.

Do you have a preference for either of these options? I'm leaning towards the second option, as I think dynamic severity would be neat, but I want to make sure I'm not missing any other undesired behavior and/or edge cases.

valentijnscholten: I think the severity should be removed from the hash_code config. And then the dynamic severity logic I think is "new", but I think it's safe to see how well it works.

Logicmn: Sounds good - done.

Maffooch (approved):

mtesauro (approved):
Approved

PR #13297, opened 2025-10-02
bugfix search results: enable bulk adding of tags and notes

### LINKED ISSUE #8410: Bulk edit of notes and tags is not possible in search results
**Bug description**
In the search results it isn't possible to set notes or tags via bulk edit menu.

**Steps to reproduce**
1. Enter search term into global search bar (top right corner)
2. Select multiple findings in the results
3. Open bulk edit menu
4. Input fields for notes and tags are missing

**Expected behavior**
I would expect the same possibility to set tags like in other finding listings (and mentioned in https://defectdojo.github.io/django-DefectDojo/usage/features/#adding-and-removing).
If the feature is not available here, the labels shouldn't be displayer. (But I really would like to use this feature in search results).

- [x] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version 2.24.0

**Screenshots**

image

valentijnscholten:
Fixes #8410


mtesauro (approved):
Approved

Maffooch (approved):

Jino-T (approved):

blakeaowens (approved):

PR #13300, opened 2025-10-02
findings list: support ordering by more fields

### LINKED ISSUE #6303: Sort vulnerabilities by overdue SLA
There is currently no way to sort vulnerabilities by overdue SLA. This functionality is useful when attempting to prioritize patching of vulnerabilities.

As a user of DefectDojo, I want to be able to sort the open vulnerabilities by overdue SLA so that I can prioritize my patching strategy.

An alternative could be filtering specifically overdue vulnerabilities. That would likely require implementing a custom filter on the finding object

**Additional context**
I brought this up in [slack](https://owasp.slack.com/archives/C2P5BA8MN/p1651869465999809) a few days ago

### LINKED ISSUE #13296: Jira configration using new feature (service account)
With the new feature of Jira - Service Account - it's now possible to use those accounts to communicate with Jira, avoiding creating fake users in our AD. The service accounts are special as they require to use a specific api url. The configuration works perfectly, but the link in the finding is "broken".

Image

Image

Have 2 fields, one for Jira Api URL and one for Jira domain URL.

None

**Additional context**
Documentation about new jira feature:
https://support.atlassian.com/user-management/docs/understand-service-accounts/

valentijnscholten:
Fixes #6303

Allows ordering by SLA Expiration Date and other fields. It does get a bit crowded now with all the arrows, especially since more columns now have a client-side ordering pair of arrows and a server-side pair. What if we just remove the client-side ordering altogether?

UPDATE:
The PR was extended, it now does:
- add server-side sorting for the SLA age column and other usefule columns
- remove client-side sorting for these columns
- fixes the bulk edit tag field not being initialized when `filter_string_matching` is enabled. #13311


valentijnscholten:
@mtesauro Did you see my question?

> What if we just remove the client-side ordering altogether?

mtesauro:
> @mtesauro Did you see my question?
> 
> > What if we just remove the client-side ordering altogether?

ROFL. I thought you were going to do that actually - either in a future commit here or a different PR. I guess I read it as a statement rather then a question. My bad.

The problem we've always had with client-side ordering is that it only knows about what's displayed on the screen which is rarely all the findings. So it really doesn't work unless you have all findings in the table and those pages can get HUGE so that's not great either.

Seems like server-side ordering would solve for both issues I mentioned IMHO.

Maffooch:
I would be down to use this opportunity to get rid of client side sorting. We already have a toggle for this, so maybe we default it to disabled, and then phase it out
https://github.com/DefectDojo/django-DefectDojo/blob/fc44e8b5008e642fd992ae640346f60143f300a8/dojo/models.py#L590-L594

Maffooch (approved):
Also approving since the server side ordering is a good start

Jino-T (approved):

PR #13303, opened 2025-10-02
enhance/normalize EDITABLE_MITIGATED_DATA handling

### LINKED ISSUE #6078: EDITABLE_MITIGATED_DATA not work
Hello teams,
I have installed 2.8.0 docker version to be able to evaluate it and it seems to me that it is not possible via API to set a closing date of a finding (upd: also in 2.6.2).

P.S. I have inserting EDITABLE_MITIGATED_DATA = True in the local_settings.py file, then in the environment section of uwsgi container of docker-compose.yml 

best,
Sal

valentijnscholten:
Fixes #6078

I thought it would be nice to allow the Finding API to accept `mitigated` and `mitigated_by` values for super users and when `settings.EDITABLE_MITIGATED_DATA` was `True`.

This PR adds that. It also normalizes the check in all place.

Two notes:
- The Close Finding UI form did not allow setting these fields at all. Now it does if the setting is enabled and the user is a superuser.
- The Close Finding API allowed non-superusers to modify the `mitigated` timestamp field if the `EDITABLE_MITIGATED_DATA`  was enabled. Not sure if this was intentional, but this PR now guards that field by the same central helper method. So only superusers can set if and `EDITABLE_MITIGATED_DATA` must be enabled. Is this oké?
- The Close Finding UI and API have this missing note types functionality. I don't fully understand it, but this PR now allows superusers to create findings that are mitigated and do not have any notes. Is this oké?

It looked like a simple change at first, but these inconsistencies made it hard to find out what is/was the intended behaviour. Do we want to allow creation and updating of findings as mitigated without doing the close finding api call?

Users can do this in the Edit Finding feature of the UI, so I think it should be supported in the API as well.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1772,6 +1772,18 @@ class FindingSerializer(serializers.ModelSerializer):
         return instance
 
     def validate(self, data):
+        # Enforce mitigated metadata editability (single validate method)
+        attempting_to_set_mitigated = any(field in data for field in ["mitigated", "mitigated_by"])
+        user = getattr(self.context.get("request", None), "user", None)
+        if attempting_to_set_mitigated and not finding_helper.can_edit_mitigated_data(user):
+            errors = {}
+            if "mitigated" in data:
+                errors["mitigated"] = ["Editing mitigated timestamp is disabled (EDITABLE_MITIGATED_DATA=false)"]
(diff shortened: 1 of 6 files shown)


Maffooch:
I think this is a fine approach - one tests as passing, I'll approve

Maffooch (approved):

Jino-T (approved):

Issue #13273, opened 2025-09-29
Remove "Finding review requested" alert when review has been finalized


savdbroek:
**Is your feature request related to a problem? Please describe**
A clear and concise description of what the problem is.
When working through Peer Review Request alerts it is confusing that the notification doesn't disappear when the Peer Review gets finalized. Especially when there are multiple or many of these. There is no way to quickly know which alert is for which peer review. Now one first has to navigate to See All Alerts and then manually find the Peer review they just finished before deleting the alert. It gets move confusing when some Peer reviews can be immediately updated whilst others might not be ready for it yet.

**Describe the solution you'd like**
As a member of a team that handles Peer Reviews I'd like the alerts for Peer Reviews to be removed when it has been finalized. 

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions or features you have considered.
Optionally a unique ID could be shown in the alert to be able to tell them apart. Like the engagement ID they are from.

**Additional context**

<img width="793" height="984" alt="Image" src="https://github.com/user-attachments/assets/e498426f-fced-4e65-9997-37cebe8bfeb7" />

valentijnscholten:
There it no link between the alerts and the finding it relates to. This is similar to when Windows creates an Alert about a new email in Outlook. Once you've read the email, the notification remains.

However if you click on "See All Alerts" you'll see the title of the finding and the title is clickable and will bring you to the finding that needs (or needed) review.

<img width="1888" height="1070" alt="Image" src="https://github.com/user-attachments/assets/45a9d8d2-11bc-4a67-968e-c28a8c62b29e" />

I have also raised a PR that adds some more info to the title of the Alert, similar to the other alerts in Defect Dojo.

<img width="757" height="164" alt="Image" src="https://github.com/user-attachments/assets/1a86343a-104c-4d7d-a4a9-ad9954e00e1e" />

<img width="1887" height="913" alt="Image" src="https://github.com/user-attachments/assets/75f7162f-a1d2-407b-b5aa-0558dda5e17a" />

Does this help?

savdbroek:
Definitely an improvement thanks! We will start using that. 

Just wondering, is it technically possible to add a delete button to the alert box per alert (my screenshot)? Combined with the more info on hover this should make the workflow a lot simpler.

The struggle with the See All Alerts page is that the alerts seem to be in a different order than in the alert box.

valentijnscholten:
[...]

PR #13314, opened 2025-10-03
make alert order explicit

valentijnscholten:
In #13273 it's reported that the ordering of alerts can be different depending on the context. Let's make it explicit.


Maffooch (approved):

mtesauro (approved):
Approved

Jino-T (approved):

PR #13321, opened 2025-10-04
Update default audit log type to django-auditlog

Maffooch:
Change the default audit logging system from django-pghistory to django-auditlog.


mtesauro (approved):
Approved

valentijnscholten (approved):
oops.. good catch!

rossops (approved):

PR #13324, opened 2025-10-04
Refactor vulnerability ID and endpoint retrieval in Finding model

Maffooch:
Streamline the retrieval process for vulnerability IDs and endpoints in the Finding model by separating logic for unsaved and saved states, improving code clarity and maintainability.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -2966,7 +2966,7 @@ class Finding(models.Model):
                 return "".join(sorted(vulnerability_id_str_list))
             return ""
 
-        return _get_unsaved_vulnearbility_ids(self) or _get_saved_vulnerability_ids(self)
+        return _get_saved_vulnerability_ids(self) or _get_unsaved_vulnearbility_ids(self) 
 
     # Get endpoints to use for hash_code computation
     # (This sometimes reports "None")
@@ -2994,7 +2994,7 @@ class Finding(models.Model):
                 return "".join(sorted(endpoint_str_list))
             return ""
 
-        return _get_unsaved_endpoints(self) or _get_saved_endpoints(self)
+        return _get_saved_endpoints(self) or _get_unsaved_endpoints(self)
 
     # Compute the hash_code from the fields to hash
     def hash_fields(self, fields_to_hash):
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -2945,53 +2945,56 @@ class Finding(models.Model):
 
     # Get vulnerability_ids to use for hash_code computation
     def get_vulnerability_ids(self):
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Do we now run the risk that unsaved items are returned for findings that were already saved and have saved items but the unsaved ones were not cleared? In the old code there was a `if self.id is None` check to only return unsaved items for new findings that hadn't been saved yet.

manuel-sommer (commented):
Just a suggestion:
Maybe you want to also remove field cve. See https://github.com/DefectDojo/django-DefectDojo/pull/9791#pullrequestreview-1958009612

Maffooch:
If the check were to look for saved objects first, and then fall back to the unsaved versions, would that address the concern?

Maffooch:
> Just a suggestion:
Maybe you want to also remove field cve. See https://github.com/DefectDojo/django-DefectDojo/pull/9791#pullrequestreview-1958009612

This will be a much larger effort for another PR

valentijnscholten (approved):
The unsaved methode has a typo.

dojo/models.py
```diff
@@ -2945,53 +2945,56 @@ def compute_hash_code_legacy(self):
 
     # Get vulnerability_ids to use for hash_code computation
     def get_vulnerability_ids(self):
-        vulnerability_id_str = ""
-        if self.id is None:
-            if self.unsaved_vulnerability_ids:
+
+        def _get_unsaved_vulnearbility_ids(finding) -> str:
```
Maffooch:
```suggestion
        def _get_unsaved_vulnerability_ids(finding) -> str:
```

dojo/models.py
```diff
@@ -2945,53 +2945,56 @@ def compute_hash_code_legacy(self):
...
+            if finding.id is not None:
+                vulnerability_ids = Vulnerability_Id.objects.filter(finding=finding)
+                deduplicationLogger.debug("get_vulnerability_ids after the finding was saved. Vulnerability references count: " + str(vulnerability_ids.count()))
+                # convert list of vulnerability_ids to the list of their canonical representation
+                vulnerability_id_str_list = [str(vulnerability_id) for vulnerability_id in vulnerability_ids.all()]
+                # sort vulnerability_ids strings
+                return "".join(sorted(vulnerability_id_str_list))
+            return ""
+
+        return _get_saved_vulnerability_ids(self) or _get_unsaved_vulnearbility_ids(self)
```
Maffooch:
```suggestion
        return _get_saved_vulnerability_ids(self) or _get_unsaved_vulnerability_ids(self)
```

PR #13328, opened 2025-10-05
SAML2 group mapping

fopina:
**Description**

> This is the second take of https://github.com/DefectDojo/django-DefectDojo/pull/11273/

While SAML2 *authentication* works well as it is, it's not very useful for *authorization*.

This PR introduces the ability to map SAML2 IdP groups (or some other list attribute) to Dojo groups.
And its configuration is similar to Azure AD group matching:
* `DD_SAML2_GROUPS_ATTRIBUTE` defines which attribute holds the group list - and if unset (default), group processing is disabled (just authentication, current behavior), similar to `DD_SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_GET_GROUPS`
* Take all groups from IdP that match the configurable regexp  `DD_SAML2_GROUPS_FILTER` - and if unset, every group will match, similar to `DD_SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_GROUPS_FILTER`
* Take all groups from the user on Dojo that match the same regexp
  * Add the missing ones from previous step
  * Remove the ones no longer present in previous

**Note**

I've updated the condition to bypass assign current user as owner of the new group in dojo/group/utils.py
Initially I simply added an extra condition `not settings.SAML2_GROUPS_ATTRIBUTE`: similar to AzureAD, skip it if SAML2 group matching was enabled.  
But then realized that condition (even the existing one) is not ideal, as groups can still be manually created via the UI (and without any "social_provider"), even when AzureAD (or other) is enabled.  
I changed the condition to simply check if `social_provider` is set for 2 reasons:
* technically assured it comes from those steps as `social_provider` is not exposed via UI
* makes business sense (IMO) that a group from a social provider does not have a owner user, as it's owned by someone else outside DefectDojo

**Testing**

> Auth0 was used to test the setup as it support SAML2

* Create new application, `Regular Web Application`
* Under application addons, enable `SAML2 web-app`
  * Use settings
  ```json
  {
   "mappings": {
     "user_id":     "urn:oid:0.9.2342.19200300.100.1.1",
     "email":       "urn:oid:1.2.840.113549.1.9.1.1",
     "given_name":  "urn:oid:2.5.4.3",
     "family_name": "urn:oid:2.5.4.4",
     "groups":      "http://schemas.xmlsoap.org/claims/Group"
   },
   "passthroughClaimsWithNoMapping": false
  }
  ```
* Add `http://localhost:8080/saml/acs/` as callback url and click `Enable`

As there are no groups in Auth0 and roles are not shared via SAML assertions, we can push them (roles) into a custom assertion (based on [this](https://community.auth0.com/t/send-roles-as-part-of-saml-assertion-when-auth0-is-the-idp/109864) and [this](https://auth0.com/docs/authenticate/protocols/saml/saml-configuration/customize-saml-assertions#example-include-user_metadata-attributes-in-an-assertion) links

* Go to Actions choose Trigger
* Choose postlogin trigger
```javascript
/**
* Handler that will be called during the execution of a PostLogin flow.
*
* @param {Event} event - Details about the user and the context in which they are logging in.
* @param {PostLoginAPI} api - Interface whose methods can be used to change the behavior of the login.
*/
exports.onExecutePostLogin = async (event, api) => {
  var roles = event.authorization.roles;
	api.samlResponse.setAttribute('urn:oid:2.5.4.33', roles);
};
```

* Back to Dojo, set the following extra configuration (add to `docker-compose.yml`, under `uwsgi` service):

```yaml
DD_SAML2_ENABLED: "true"
DD_SAML2_METADATA_AUTO_CONF_URL: <METADATA URL FROM AUTH0 SAML2 ADDON>
DD_SITE_URL: http://localhost:8080
DD_SOCIAL_LOGIN_AUTO_REDIRECT: "true"
DD_SOCIAL_AUTH_SHOW_LOGIN_FORM: "false"
DD_SAML2_CREATE_USER: "true"
DD_SAML2_ATTRIBUTES_MAP: email=email,uid=username,cn=first_name,sn=last_name
DD_SAML2_GROUPS_FILTER: ".*"
DD_SAML2_GROUPS_ATTRIBUTE: roleOccupant
```

* Now create some roles (eg: `DojoUser`) and add them to the test user in Auth0.
* Create Dojo groups with the exame same name (eg: `DojoUser`)
* Login using that test user and the user should be part of the mapped Dojo groups
  * You can verify in logs that those groups were deteted


valentijnscholten:
Converted to draft as you mentioned it's draft :-)

fopina:
Sorry about that @valentijnscholten, I meant to make it draft and must have missed it in the end. It's ready now!

@kiblik as you were reviewing the previous "take", let me know what you think of it like this - closer to AzureAD for consistency, I hope :)

fopina:
@Maffooch any feedback on this one? I hope adding group support to more auth backends is a welcome change 🙏

fabiogermann:
Is there any feedback on this one - it would be a great addition! Is there anything missing?

fopina:
There's quite a lot to review and comment in this project, it's normal

However it'd be nice to avoid the PR needing rebases 😃 

@Maffooch @valentijnscholten  Is there lack of agreement for this change internally or it just hasn't made to discussion? If the first, I'd be happy to know the reason, maybe there's some other suggestion

devGregA:
Hi @fopina, sorry for the delay here. Thank you for your pull request and contributions to Dojo.

@mtesauro and @Maffooch pinged me on this one a while ago, so totally my fault for the delay. After considering for quite some time, I’m of the opinion  a feature like this isn’t something we should be maintaining in OS for several reasons:

1. We can’t reasonably maintain tests and functionality around this feature. One area DefectDojo shines and we regularly get constant positive feedback on is the feature set we have works as advertised. Maintaining test coverage, accounts around this feature, etc, would be a major burden for our project maintainers.

2. Maintaining this feature would essentially be subsidizing enterprises that can afford to support DefectDojo and be a Pro customer. One of the things we’re always trying to balance is OS vs Pro. If we don’t generate enough biz, we can’t maintain open source. If a company is large enough to use SSO with groups, it seems reasonable they should be on Pro, which in turn helps us to maintain open source, rather than making it more difficult to maintain as outlined in 1.

fopina:
Hi Greg I appreciate the transparency!
Thanks

fabiogermann:
Hi @devGregA ,

Thank you for the transparency — it’s genuinely appreciated.

I’d like to offer an additional perspective. In many organizations, overall IT maturity tends to outpace the maturity of their security operations. As a result, applications or solutions that do not provide features such as SCIM or automated provisioning are often overlooked or deprioritized during evaluation. While there may be an assumption that organizations will eventually upgrade to a “Pro” tier, the reality is that budget constraints frequently limit that progression.

While the expectation may be that organizations should adopt the Pro tier from the outset, the reality is that budget constraints often make that difficult, even when the added capabilities would clearly provide value.

A strong example of a product that successfully enables this kind of gradual, maturity-aligned progression is SonarQube, which allows organizations to start small and scale capabilities as their needs and security maturity evolve.

PR #13329, opened 2025-10-05
fix(gha): Run Release-Nightly only once a day

kiblik:
Until now, `Release-Nightly: Build & Push DEV` has been triggered multiple times between 5:00 and 5:59 UTC. But we needed it only once.

image
https://github.com/DefectDojo/django-DefectDojo/actions/workflows/release-nightly-dev.yml


mtesauro (approved):
Approved

PR #13331, opened 2025-10-05
docker compose: switch to Valkey as message broker

### LINKED ISSUE #13323: Migrate from `redis` to `valkey`
Since the license change at Redis the fork ValKey has become widely popular and is backed by industry giants such as AWS. AWS is advising to use ValKey over Redis and is using lower prices for ValKey compared to Redis.

Defect Dojo will migrate over to ValKey, we're currently investigating the migration paths. At first sight it looks like the existing Docker volume can just be used by the ValKey container.

- https://valkey.io/topics/migration/

valentijnscholten:
Fixes #13323 

Since the license change at Redis the fork ValKey has become widely popular and is backed by industry giants such as AWS. AWS is advising to use ValKey over Redis and is using lower prices for ValKey compared to Redis.

This PR migrates the docker compose setup to use Valkey instead of Redis.

The simplest approach is chosen:
- re-use the existing Redis Docker volume

### Justification

Since Valkey 7.2 is a fork of Redis 7.2 it is a drop-in replacement. I've tested the by generating a large backlog of celery task before making the changes in this PR. Shuttting down, applying the changes and bringin Defect Dojo up resulted in celeryworkers resuming work and processing the backlog.

I've checked the Dockerfiles for both [Redis ](https://github.com/redis/docker-library-redis/blob/7d9ee3a9c8e9b5085cf0e354100a5aaf9be6a744/7.2/alpine/Dockerfile) and [Valkey](https://github.com/valkey-io/valkey-container/blob/mainline/7.2/alpine/Dockerfile). They show that they're laregely the same:

- both use `/data` to store their data
- both use id=999, gid=1000 for file ownership

The Valkey docs mention alternative migration paths such as taking snapshots and restoring these: 
- https://valkey.io/topics/migration/

The re-use of the existing volume is much simpler and fits with the statement that Valkey 7.2 is compatible with Redis 7.2

In theory we could go to Valkey 8 or 9, but somehow it feels like the safer option to stick with 7.2 for now.

### Client libraries

I tried migrating from the `redis` Python library to the `valkey` Python library, but ran into some issues with Celery. Research shows it's recommended to stick to the `redis` library unitl all other frameworks such as Celery have adopted first-class support of the Valkey platform.

###  Volume names still contains `redis`

By re-using the existing docker volume, we are stuck with the name `defectdojo_redis`. We can't rename volumes, so we just have to live with this. Or.....

EDIT: We've chosen to retain the volume and accept for now that it is named `defectdojo_redis`

~~1) Its possible to change the name of the volume to `defectdojo_valkey` if and only if users perform some manual steps during the upgrade. This would look something like:~~

~~```~~
~~`docker compose down`~~
~~`docker volume create XXXX_defectdojo_valkey`~~
~~`docker run --rm -u 0:0 -v oldvol:/from -v XXXX_defectdojo_valkey:/to alpine sh -c 'tar -C /from -cpf - . | tar -C /to -xpf -'`~~
~~`docker compose up -d`~~
~~`docker volume rm XXXX_defectdojo_redis`~~
~~```~~
~~The XXX is the prefix given to volume names by docker compose, usually based on the folder Defect Dojo is in.~~
~~In theory we could create a script that reads the exact volume name and have a `migrate-redis-to-valkey.sh` script that does everything transparently.~~

~~2) That script could (in theory) also be run by the `initializer`. Something like this could be added to the entrypoint:~~

~~```~~
~~# Optional one-time migration from old Redis volume (/redis-data) to new Valkey volume (/valkey-data)~~
~~# Runs before DD_INITIALIZE check so it executes even if initialization is skipped~~
~~if [ "${DD_MIGRATE_REDIS_VOLUME:-1}" != "0" ] && [ -d /redis-data ] && [ -d /valkey-data ]; then~~
~~  if [ ! -f /valkey-data/.dojo_redis_migrated ] \~~
~~     && [ -z "$(ls -A /valkey-data 2>/dev/null || true)" ] \~~
~~     && [ -n "$(ls -A /redis-data 2>/dev/null || true)" ]; then~~
~~    echo "Migrating data from old Redis volume (/redis-data) to new Valkey volume (/valkey-data)..."~~
~~    tar -C /redis-data -cpf - . | tar -C /valkey-data -xpf -~~
~~    touch /valkey-data/.dojo_redis_migrated~~
~~  else~~
~~    echo "Skip migration: destination not empty, source empty, or already migrated"~~
~~  fi~~
~~fi~~
~~```~~

~~3) Or we could add a one-time init service to the docker-compose file. Something like:~~

~~```~~
~~services:~~
~~  valkey-init-migrate:~~
~~    image: alpine:3.19~~
~~    command: >~~
~~      sh -c 'set -eu;~~
~~             if [ -z "$(ls -A /to 2>/dev/null)" ] && [ -n "$(ls -A /from 2>/dev/null)" ]; then~~
~~               tar -C /from -cpf - . | tar -C /to -xpf -;~~
~~             else~~
~~               echo "Skip copy: /to not empty or /from empty";~~
~~             fi'~~
~~    volumes:~~
~~      - oldredisdata:/from:ro~~
~~      - valkeydata:/to~~
~~    restart: "no"~~

~~  valkey:~~
~~    image: valkey/valkey:7.2-alpine~~
~~    depends_on:~~
~~      valkey-init-migrate:~~
~~        condition: service_completed_successfully~~
~~    volumes:~~
~~      - valkeydata:/data~~
~~    # command: ["valkey-server","--appendonly","yes"]  # if you used AOF~~

~~volumes:~~
~~  oldredisdata:~~
~~    external: true         # existing Redis volume name~~
~~    name: oldredisdata~~
~~  valkeydata:~~
~~    # new volume Compose will create (your new name)~~
~~```~~

~~I say in theory everywhere because all these "automated" migrations seem risky to me. Simply reusing the existing `defectdojo_redis` volume is fail-safe. Let me know your thoughts or alternative (simpler?) approaches to get rid of the `redis` in the volume name.~~

~~**UPDATE:**~~
~~- Easiest solution for us would be to instruct users to let all Celery tasks finish before shutting down the celeryworker(s). In that case we can start with a clean sheet with a nicely named volume (and Valkey 9).~~
~~```~~

### DIFF
--- a/docs/content/en/open_source/upgrading/2.52.md
+++ b/docs/content/en/open_source/upgrading/2.52.md
@@ -5,7 +5,6 @@ weight: -20251006
 description: Replaced Redis with Valkey & Helm chart changes & MobSF parser merge
 ---
 
-## Valkey
 ## Fix UI overwriting service field from parsers
 
 The web form in the UI by default sends an empty string, which ended up overwriting the service value provided by parsers.
@@ -24,7 +23,7 @@ See [PR 13517](https://github.com/DefectDojo/django-DefectDojo/pull/13517) for m
 A bug was fixed in the `UNIQUE_ID_OR_HASH_CODE` algorithm where it stopped processing candidate findings with equal `unique_id_from_tool` or `hash_code` value.
 Strictly speaking this is not a breaking change, but we wanted to make you aware that you can see more (better) more deduplicatation for parsers using this algorithm.
 
-## Merge of MobSF parsers
+## Valkey in `docker compose`
 
 Since the license change at Redis the fork ValKey has become widely popular and is backed by industry giants such as AWS. AWS is advising to use ValKey over Redis and is using lower prices for ValKey compared to Redis.
 


mtesauro:
For the record, I'm all for this approach. Keeping an :eye: on this for it to leave Draft status :wink:

9alexx3:
Hello @valentijnscholten 

This migration fixes the [Redis latest vulnerability](https://redis.io/blog/security-advisory-cve-2025-49844/) (CVE-2025-49844) that was fixed in v8.2.2?

valentijnscholten:
Yes it does: https://github.com/valkey-io/valkey/releases/tag/7.2.11, but we'll be updating redis to 7.2.12 before we're moving to valkey as we still have some testing/adjustments to do to the helm chart.

valentijnscholten:
Looks there's some contradicting info floating around as according to Redis themselves the CVE is fixes in 7.2.11, which we are already running: https://github.com/redis/redis/releases/tag/7.2.11

mtesauro:
> Looks there's some contradicting info floating around as according to Redis themselves the CVE is fixes in 7.2.11, which we are already running: https://github.com/redis/redis/releases/tag/7.2.11

This was my understanding aka DefectDojo was not vulnerable when this was reported due to the version we had in the lastest version.

So, keep those DefectDojo instances up to the latest, greatest version :+1:

docs/content/en/open_source/upgrading/2.52.md
```diff
@@ -1,10 +1,38 @@
...
+
+If you want to be 110% sure no tasks will be lost you could perform the upgrade in two steps:
+
+1) Stop nginx, uwsgi, celerybeat to prevent new tasks from being created:
+
+`docker compose down nginx, uwsgi, celerybeat`
+
+2) Observe the logs of the celeryworker(s) and wait until all tasks are finished:
+
+`docker compose logs celeryworker`
```
kiblik: As an alternative to this, I recommend observing the length of the queue.
```
$ docker compose exec redis redis-cli llen celery
(integer) 8
```
It is recommended to wait until the command does not return `0`.

docs/content/en/open_source/upgrading/2.52.md
```diff
@@ -1,10 +1,38 @@
 ---
 title: 'Upgrading to DefectDojo Version 2.52.x'
 toc_hide: true
-weight: -20251006
-description: Helm chart changes.
+weight: -20250804
```
kiblik:
```suggestion
weight: -20251006
```
It is not good idea to change this value. It would put this page back in history (considering the order of pages).

docs/content/en/open_source/upgrading/2.52.md
valentijnscholten: thanks

docs/content/en/open_source/upgrading/2.52.md
```diff
@@ -2,12 +2,37 @@
...
+## Valkey
 
-Mobsfscan Scan" has been merged into the "MobSF Scan" parser. The "Mobsfscan Scan" scan_type has been retained to keep deduplication working for existing Tests, but users are encouraged to move to the "MobSF Scan" scan_type.
+Since the license change at Redis the fork ValKey has become widely popular and is backed by industry giants such as AWS. AWS is advising to use ValKey over Redis and is using lower prices for ValKey compared to Redis.
+
+Defect Dojo 2.52 now uses ValKey as a message broker. Teh existing redit volume can be used by Valkey, so this is just a drop in replacement.
+
+If you want to know more or have a setup where you cannot just re-use the existing volume, please visit https://valkey.io/topics/migration/.
+
+When you shutdown Defect Dojo to perform the upgrade, the celery tasks that are in the queue are storede to disk. After the upgrade, the celery workers will process these tasks as normal.
```
valentijnscholten:
```suggestion
When you shutdown Defect Dojo to perform the upgrade, the celery tasks that are in the queue are stored to disk. After the upgrade, the celery workers will process these tasks as normal.
```

kiblik (approved):

rossops (approved):

PR #13368, opened 2025-10-07
fix(helm): Index contained outdated annotations

kiblik:
During the last release cycle, GHA which is responsible for fixing Helm annotations, had not been in the `master` branch yet, so annotations had not been added to charts. This should fix index for now.

Yes, tests are failing, but it is only because they are defined based on the old definition, and related imports are not available anymore.

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -2,11 +2,13 @@ apiVersion: v1
 entries:
   defectdojo:
   - annotations:
-      artifacthub.io/changes: ""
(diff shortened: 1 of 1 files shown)


valentijnscholten:
@kiblik Looks like the helm-charts branch needs an update from master to get the action definitions updated?

Maffooch:
Closing as stale - please reopen if this is still needed/desired

PR #13370, opened 2025-10-07
feat(helm): Split image locations+tags, allow digest pinning

kiblik:
`tag` and `repositoryPrefix` fields have been deprecated. Currently, image tags used in containers are derived by default from the `appVersion` defined in the Chart.  
This behavior can be overridden by setting the `tag` value in `images.django` and `images.nginx`.  
If fine-tuning is necessary, each container’s image value can also be customized individually (`celery.beat.image`, `celery.worker.image`, `django.nginx.image`, `django.uwsgi.image`, `initializer.image`, and `dbMigrationChecker.image`).  
Digest pinning is now supported as well.

### DIFF
--- a/.github/workflows/k8s-tests.yml
+++ b/.github/workflows/k8s-tests.yml
@@ -69,6 +69,8 @@ jobs:
                defectdojo \
                ./helm/defectdojo \
                --set django.ingress.enabled=true \
+               --set images.django.image.tag=latest \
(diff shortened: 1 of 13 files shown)


kiblik:
@mtesauro / @Maffooch, GHA is failing because of some unrelated issue. Can we merge this PR as it is?

Issue #13338, opened 2025-10-05
Acunetix parser generates broken endpoints when `<StartURL>` includes a port number


Irfan-Mohd:
### **Bug description**
When importing Acunetix scan results, DefectDojo’s Acunetix parser does not correctly handle `<StartURL>` values that contain a hostname and port number but no scheme. This results in malformed endpoints such as:

`example.co.id:// 🚩`

If the hostname is long, the parser instead triggers a `500` error with the message:

`value too long for type character varying(20)`

The issue occurs because the parser misinterprets the hostname as the scheme. When it detects a` :`, it prepends `//`, which causes `hyperlink.parse()` to incorrectly parse URLs with ports. in this case: `example.co.id:8080` is interpreted as `example.co.id://`

### **Steps to reproduce**
**Steps to reproduce the behavior:**
1. In DefectDojo, create a Test and **Import Scan** &rarr; Acunetix Scan &rarr; upload Sample XML file.
2. Open any finding from the imported report.
3. In the `Vulnerable Endpoints / Systems` section, observe the malformed endpoint.
**Note:** This issue occurs if the endpoint has less than 20 characters before the port number. such as `example.co.id:8080`

### **Expected behavior**
 
Endpoints with port number should be parsed correctly:

example.co.id:8080

### **Deployment method**
- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

### **Environment information**
 - **DefectDojo version:** v. 2.50.4 ( 45273925 ) 

### **Sample scan files**
[XML_http_example_co_id_.xml](https://github.com/user-attachments/files/22712534/XML_http_example_co_id_.xml)

### **Screenshots**

<img width="1842" height="822" alt="Image" src="https://github.com/user-attachments/assets/77f8cd0a-f29b-4305-94a8-705076698fae" />

if the endpoint has more than 20 characters before the port number, 

<img width="1371" height="611" alt="Image" src="https://github.com/user-attachments/assets/9262e4aa-c2ad-47a4-a385-f1a4b6ef63a0" />

manuel-sommer:
I guess we can close this @Irfan-Mohd

PR #13371, opened 2025-10-07
fix: handle broken endpoints when <StartURL> includes a port number in Acunetix XML parser

Irfan-Mohd:
Fixes https://github.com/DefectDojo/django-DefectDojo/issues/13338


valentijnscholten (approved):
Thank you!

PR #13372, opened 2025-10-07
add unit tests to test importer deduplication

valentijnscholten:
There's a (very) comprehensive test suite to test the deduplication logic for all dedupe algorithm types in https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/test_deduplication_logic.py.
This tests deduplication logic by creating findings, triggering dedupe and assess the results.

What we also need is unit tests that perform a full import process and assess the dedeuplication logic of the importer. Currently this is identical to what happens when creating findings on their own. But this might change in a future PR where we optimize deduplication for batches of findings.

This PR adds these unit tests so we capture the current behaviour and can safely go for that batch dedupe PR.

We also capture the query counts and async counts. Existing counts have been raised slightly because we enabled dedupe.

### DIFF
--- a/unittests/test_importers_performance.py
+++ b/unittests/test_importers_performance.py
@@ -46,6 +46,7 @@ class TestDojoImporterPerformance(DojoTestCase):
         self.system_settings(enable_webhooks_notifications=False)
         self.system_settings(enable_product_grade=False)
         self.system_settings(enable_github=False)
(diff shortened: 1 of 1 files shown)


mtesauro:
Love this. Can't wait to see it come out of "Draft"

PR #13384, opened 2025-10-08
feat(nginx): Make nginx Non-root & Read-only-friendly

### LINKED ISSUE #11031: DefectDojo containers only function when group is 0 (root)
**Bug description**
When running the defectdojo-nginx and defectdojo-django (celery-worker) containers in a secured cluster (root not allowed) they fail to start

**Steps to reproduce**
1. Deploy helm chart on kubernetes with podSecurityContext and containerSecurityContext for all pods/containers set to

```yaml
securityContext:
    runAsUser: 1001
    runAsGroup: 1001
    runAsNonRoot: true
```

**Expected behavior**
Helm chart deploys successfully and defectdojo comes up healthy

**Deployment method**
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: k3d 1.29.6+k3s2
 - DefectDojo version : 2.38.4

**Logs** 


celery-worker:
```
uid=1001(defectdojo) gid=1001 groups=1001                                                                                                                  
[09/Oct/2024 18:51:34] INFO [dojo.models:4572] enabling audit logging                                                                                      
Waiting for database to be reachable                                                                                                                       
[09/Oct/2024 18:51:36] INFO [dojo.models:4572] enabling audit logging                                                                                      
System check identified no issues (0 silenced).                                                                                                            
[09/Oct/2024 18:51:39] INFO [dojo.models:4572] enabling audit logging                                                                                      
/usr/local/lib/python3.11/site-packages/celery/platforms.py:799: SecurityWarning: An entry for the specified gid or egid was not found.                    
We're assuming this is a potential security issue.                                                                                                         
                                                                                                                                                           
  warnings.warn(SecurityWarning(ASSUMING_ROOT))                                                                                                            
Running a worker with superuser privileges when the                                                                                                        
worker accepts messages serialized with pickle is a very bad idea!                                                                                         
                                                                                                                                                           
If you really want to continue then you have to set the C_FORCE_ROOT                                                                                       
environment variable (but please think about this before you do).                                                                                          
                                                                                                                                                           
User information: uid=1001 euid=1001 gid=1001 egid=1001
```

defectdojo-nginx
```
sed: can't create temp file '/etc/nginx/nginx.confXXXXXX': Permission denied                                                                               
Nginx metrics are enabled                                                                                                                                  
sed: can't create temp file '/etc/nginx/nginx.confXXXXXX': Permission denied                                                                               
/entrypoint-nginx.sh: line 30: can't create /etc/nginx/.htpasswd: Permission denied                                                                        
Basic auth is on for user ...                                                                                                                              
2024/10/09 18:51:19 [emerg] 1#1: mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)                                                     
nginx: [emerg] mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
```

I'm not sure whats going on with the celery worker but the defectdojo-nginx container has the majority of its filesystem owned by root

kiblik:
Fix #11031


Maffooch:
@mtesauro does this fit into your plans for image hardening?

mtesauro:
> @mtesauro does this fit into your plans for image hardening?

TBH, I was going to start with the DefectDojo/Django containers first as those are the containers where we have the most code / do the most modifications.

Changing the UID to under the typical 1000+ is interesting but I don't believe it will cause issues for k8s and it also seems that OpenSift has stopped wanting specific UIDs based on [this](https://www.redhat.com/en/blog/a-guide-to-openshift-and-uids) so this shouldn't hurt those using k8s or compose currently as far as I can tell.

mtesauro:
I can test this PR out when it's not a draft as well

kiblik:
> I can test this PR out when it's not a draft as well

The current issue with this PR is that as soon as the process tries to modify configs
https://github.com/DefectDojo/django-DefectDojo/blob/1b338e259718b47dadc23da669f031339f811f73/docker/entrypoint-nginx.sh#L24
https://github.com/DefectDojo/django-DefectDojo/blob/1b338e259718b47dadc23da669f031339f811f73/docker/entrypoint-nginx.sh#L28
https://github.com/DefectDojo/django-DefectDojo/blob/1b338e259718b47dadc23da669f031339f811f73/docker/entrypoint-nginx.sh#L33
https://github.com/DefectDojo/django-DefectDojo/blob/1b338e259718b47dadc23da669f031339f811f73/docker/entrypoint-nginx.sh#L36
execution fails (because FS is read-only).

I'm considering creating a separate PR that will be responsible for generating the final config, which will be on a read-write location. It would combine multiple partial definitions of the `nginx.conf` file. Also, I would like to merge it with `nginx_TLS.conf` because most of the file is the same.

PR #13387, opened 2025-10-09
Fix AWS Inspector2 parser CVSS mapping

skywalke34:
## Summary
- Fixed AWS Inspector2 parser to extract CVSS from `packageVulnerabilityDetails.cvss[]` instead of `inspectorScoreDetails.adjustedCvss`
- This ensures the original CVE CVSS score from the vulnerability database (e.g., NVD) is used rather than AWS Inspector's adjusted score

## Changes
- Modified `get_cvss_details()` method in `dojo/tools/aws_inspector2/parser.py`
- Added comments explaining the change

## Test plan
- Existing unit tests pass
- Parser correctly extracts CVSS vector from vulnerability database

Authored by T. Walker - DefectDojo

### DIFF
--- a/dojo/tools/aws_inspector2/parser.py
+++ b/dojo/tools/aws_inspector2/parser.py
@@ -100,13 +100,20 @@ class AWSInspector2Parser:
 
         return finding
 
+# Extracting CVSS from packageVulnerabilityDetails.cvss[] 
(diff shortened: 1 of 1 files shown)


skywalke34:
CVSS score is calculated from vector and is not imported as it will be overwritten anyway.  Change not necessary.

PR #13403, opened 2025-10-12
fix(renovate): Drop Bitnami

kiblik:
As we are not using officially bitnami, it does not need to be checked by renovate.


mtesauro:
@kiblik Happy to merge this one the conflict is fixed.

PR #13405, opened 2025-10-12
feat(renovate): Add validator

kiblik:
Renovate can accept the old format of config, but it is better to stick with the latest standard.
Plus, add a validator for config.


Maffooch (approved):

kiblik:
This change is not needed anymore.

kiblik:
Changes had been added in #13406

PR #13408, opened 2025-10-13
feat(helm): Use Valkey

### LINKED ISSUE #13323: Migrate from `redis` to `valkey`

kiblik:
Fix #13323

#13331 was replaced only in `docker-compose`. This PR replaces it in the HELM chart as well.
Official Valkey HELM charts would require too many customisations; we needed something more developed. https://artifacthub.io/packages/helm/cloudpirates-valkey/valkey supported many fields that are compatible with the previous Bitnami stack.

There are multiple places that are mentioned in release notes.

### DIFF
--- a/helm/defectdojo/README.md
+++ b/helm/defectdojo/README.md
@@ -749,6 +749,7 @@ A Helm chart for Kubernetes to install DefectDojo
 | postgresql.primary.podSecurityContext.fsGroup | int | `1001` | fsGroup specification below is not applied if enabled=false. enabled=false is the required setting for OpenShift "restricted SCC" to work successfully. |
 | postgresql.volumePermissions.containerSecurityContext | object | `{"runAsUser":1001}` | if using restricted SCC set runAsUser: "auto" and if running under anyuid SCC - runAsUser needs to match the line above |
 | redisParams | string | `""` | Parameters attached to the redis connection string, defaults to "ssl_cert_reqs=optional" if `redisScheme` is `rediss` |
+| redisPort | int | `6379` | Define the protocol to use with the external Redis instance |
 | redisScheme | string | `"redis"` | Define the protocol to use with the external Redis instance |
 | redisServer | string | `nil` | To use an external Redis instance, set `redis.enabled` to false and set the address here: |
 | revisionHistoryLimit | int | `10` | Allow overriding of revisionHistoryLimit across all deployments. |
--- a/helm/defectdojo/templates/_helpers.tpl
+++ b/helm/defectdojo/templates/_helpers.tpl
@@ -96,6 +96,17 @@ redis
 {{- end -}}
 
 {{- /*
+  Determine the default port to use for Redis.
+*/}}
+{{- define "redis.port" -}}
+{{- if .Values.valkey.enabled -}}
+{{- .Values.valkey.service.port -}}
+{{- else -}}
+{{- .Values.redisPort -}}
+{{- end -}}
+{{- end -}}
+
+{{- /*
   Builds the repository names for use with local or private registries
 */}}
 {{- define "celery.beat.image" -}}
--- a/helm/defectdojo/templates/configmap.yaml
+++ b/helm/defectdojo/templates/configmap.yaml
@@ -30,8 +30,8 @@ data:
   DD_CELERY_BROKER_SCHEME: {{ template "redis.scheme" . }}
   DD_CELERY_BROKER_USER: ''
   DD_CELERY_BROKER_HOST: {{ template "redis.hostname" . }}
-  DD_CELERY_BROKER_PORT: '{{ .Values.valkey.service.port }}'
-  DD_CELERY_BROKER_PARAMS: '{{ .Values.redisParams | default $defaultBrokerParams }}'
+  DD_CELERY_BROKER_PORT: '{{ template "redis.port" . }}'
+  DD_CELERY_BROKER_PARAMS: '{{ template "redis.params" . }}'
   DD_CELERY_BROKER_PATH: '{{ .Values.celery.path | default "//" }}'
   DD_CELERY_LOG_LEVEL: {{ .Values.celery.logLevel }}
   DD_CELERY_WORKER_POOL_TYPE: {{ .Values.celery.worker.appSettings.poolType | default "solo" }}
--- a/helm/defectdojo/values.yaml
+++ b/helm/defectdojo/values.yaml
@@ -696,9 +696,6 @@ extraEnv: []
 # ```
 localsettingspy: ""
 
-#
-# External database support.
-#
 # -- Parameters attached to the valkey connection string, defaults to "ssl_cert_reqs=optional" if `valkey.tls.enabled`
 valkeyParams: ""
 #
@@ -709,6 +706,8 @@ valkeyParams: ""
 redisServer: ~
 # -- Parameters attached to the redis connection string, defaults to "ssl_cert_reqs=optional" if `redisScheme` is `rediss`
 redisParams: ""
+# -- Define the protocol to use with the external Redis instance
+redisPort: 6379
 # -- Define the protocol to use with the external Redis instance
 redisScheme: redis
 #
--- a/docs/content/en/open_source/upgrading/2.53.md
+++ b/docs/content/en/open_source/upgrading/2.53.md
@@ -22,7 +22,7 @@ Added Helm chart support for Celery and Django deployments for Horizontal Pod Au
 ##### Renamed values 
 
 HELM values had been changed to the following:
-- `createRedisSecret` → `createValkeySecret` (`createRedisSecret` is still in place to be able to correctly show error message, but it will be removed in the future)
+- `createRedisSecret` → `createValkeySecret`
 - `redis.enabled` → `valkey.enabled`
 - `redis.auth.existingSecret` → `valkey.auth.existingSecret`, plus value is pointing to secret `defectdojo-valkey-specific` now
 - `redis.auth.existingSecretPasswordKey` → `valkey.auth.existingSecretPasswordKey`, plus value is pointing to secret `valkey-password` now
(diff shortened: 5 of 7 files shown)


Maffooch:
Hi @kiblik the valkey migration is slated for release in the upcoming minor release on November 3rd. The [docker compose PR](https://github.com/DefectDojo/django-DefectDojo/pull/13331) looks ready to roll after your feedback was incorporated. Is there anything left to be implemented in this PR for the helm chart?

kiblik:
So, my original idea was to replace Bitnami Redis with the [official Valkey Helm chart](https://github.com/valkey-io/valkey-helm). Unfortunately, this is not fully possible because authN issue.
The official Valkey Helm chart does not support the approach: "This is a password which I generated/prepared for you. Use it". Valkey needs the `users.acl` file with the full definition of users.
This issue is partially solved in the [open PR](https://github.com/valkey-io/valkey-helm/pull/14), but still not fully support "This is a prepared password. Use it."
So we might 
- use Valkey chart from Bitnami. But charts and/or Docker images would have to be mirrored (not a good and maintainable solution) 
- use [CloudPirates chart](https://github.com/CloudPirates-io/helm-charts/tree/main/charts/valkey) (comes from the original Bitnami definition, but they keep it nicely maintained)
- use deployment without authN (we might depend on network isolation using NetworkPolicy)
- ???

kiblik:
I hope migration will be easier, and I did not have enough time to come back to it until now.

kiblik:
I still need some time to write the migration steps + description of PR + see https://github.com/CloudPirates-io/helm-charts/issues/552
But I suppose the review can start.

kiblik:
@fernandezcuesta, I see that you have been adding support for `sentinel` and `tls` for Redis. Can you please test the changes in this PR?

kiblik:
This PR is fully ready for review.

helm/defectdojo/README.md
```diff
@@ -359,16 +361,13 @@ You will still need to set a host value as well.
 
 If you want to use a redis-sentinel setup as the Celery broker, you will need to set the following.
 
-1. Set redis.scheme to "sentinel" in values.yaml
+1. Set valkey.scheme to "sentinel" in values.yaml
 2. Set two additional extraEnv vars specifying the sentinel master name and port in values.yaml
 
 ```yaml
-celery:
-  broker: 'redis'
```
fernandezcuesta, line 367: I guess this means no other broker will ever be supported?

helm/defectdojo/README.md
```diff
@@ -583,8 +581,9 @@ A Helm chart for Kubernetes to install DefectDojo
 | cloudsql.use_private_ip | bool | `false` | whether to use a private IP to connect to the database |
 | cloudsql.verbose | bool | `true` | By default, the proxy has verbose logging. Set this to false to make it less verbose |
 | createPostgresqlSecret | bool | `false` | create postgresql secret in defectdojo chart, outside of postgresql chart |
-| createRedisSecret | bool | `false` | create redis secret in defectdojo chart, outside of redis chart |
+| createRedisSecret | bool | `false` | this value is deprecated and will be removed in the future |
```
fernandezcuesta: would be awesome to have this translated into the schema; did a quick check on the helm plugin and doesn't look to be supported though

helm/defectdojo/templates/_helpers.tpl
```diff
@@ -58,29 +58,25 @@
...
-{{- if .Values.redis.tls.enabled -}}
+{{- if .Values.valkey.tls.enabled -}}
 {{- printf "rediss" -}}
-{{- else if .Values.redis.sentinel.enabled -}}
+{{- else if .Values.valkey.sentinel.enabled -}}
 {{- printf "sentinel" -}}
 {{- else -}}
 {{- printf "redis" -}}
 {{- end -}}
 {{- end -}}
```
fernandezcuesta, line 96:
```suggestion
{{- /*
  Determine the default params to use for Redis.
*/}}
{{- define "redis.params" -}}
{{- $redisScheme := template "redis.scheme" . -}}
{{- $defaultBrokerParams := ternary "ssl_cert_reqs=optional" "" (eq "rediss" $redisScheme -}}
{{- if .Values.valkey.enabled -}}
{{- default $defaultBrokerParams .Values.valkeyParams -}}
{{- else -}}
{{- default $defaultBrokerParams .Values.redisParams -}}
{{- end -}}

{{- /*
  Determine the protocol to use for Redis.
*/}}
{{- define "redis.scheme" -}}
{{- if .Values.valkey.enabled -}}
{{- if .Values.valkey.tls.enabled -}}
rediss
{{- else if .Values.valkey.sentinel.enabled -}}
sentinel
{{- else -}}
redis
{{- end -}}
{{- else -}}
{{- .Values.redisScheme -}}
{{- end -}}
{{- end -}}
```

helm/defectdojo/templates/configmap.yaml
```diff
@@ -1,5 +1,5 @@
 {{- $fullName := include "defectdojo.fullname" . -}}
-{{- $defaultBrokerParams := ternary "ssl_cert_reqs=optional" "" .Values.redis.tls.enabled -}}
+{{- $defaultBrokerParams := ternary "ssl_cert_reqs=optional" "" .Values.valkey.tls.enabled -}}
```
fernandezcuesta:
```suggestion
{{- $redisScheme := template "redis.scheme" . -}}
{{- $defaultBrokerParams := ternary "ssl_cert_reqs=optional" "" (eq "rediss" $redisScheme -}}
```

helm/defectdojo/values.yaml
```diff
@@ -676,7 +674,7 @@ localsettingspy: ""
 # @schema type:[string, null]
 # -- To use an external Redis instance, set `redis.enabled` to false and set the address here:
 redisServer: ~
-# -- Parameters attached to the redis connection string, defaults to "ssl_cert_reqs=optional" if `redis.tls.enabled`
+# -- Parameters attached to the redis connection string, defaults to "ssl_cert_reqs=optional" if `valkey.tls.enabled`
 redisParams: ""
```
fernandezcuesta, line 708:
```suggestion

# -- Parameters attached to the valkey connection string, defaults to "ssl_cert_reqs=optional" if `valkey.tls.enabled`
valkeyParams: ""

#
# External database support.
#
# @schema type:[string, null]
# -- To use an external Redis instance, set `redis.enabled` to false and set the address here:
redisServer: ~
# -- Parameters attached to the redis connection string, defaults to "ssl_cert_reqs=optional" if `redisScheme` is `rediss`
redisParams: ""
# -- Define the protocol to use with the external Redis instance
redisScheme: redis
```

fernandezcuesta: Ideally the scheme and params should be isolated when using an external redis or when opting in for Valkey.

 ⚠️ didn't tested myself though, but hope the logic is clear

fernandezcuesta (commented):
@kiblik added some logic and yet another parameter to better isolate the 2 options (external Redis vs Valkey)

helm/defectdojo/README.md
mtesauro: We currently don't have plans to support multiple brokers FWIW

helm/defectdojo/README.md
kiblik: Yes, the `deprecated` flag is not supported on the key level. Only for whole charts.

However, schema (as we use it) is just validating types (so you do not put a string in a dict or an int in a boolean - we do not use `no-additional-properties`, yet). It means that I can drop this field (from `values.yaml` and schema). Schema validation will not complain (undefined fields are not validated), and the error message will still be shown.

helm/defectdojo/values.yaml
kiblik: I like this approach. As it hasn't been tested, I will need more time to do that. I will check it next week.

sgissi:
> So, my original idea was to replace Bitnami Redis with the [official Valkey Helm chart](https://github.com/valkey-io/valkey-helm). Unfortunately, this is not fully possible because authN issue. The official Valkey Helm chart does not support the approach: "This is a password which I generated/prepared for you. Use it". Valkey needs the `users.acl` file with the full definition of users. This issue is partially solved in the [open PR](https://github.com/valkey-io/valkey-helm/pull/14), but still not fully support "This is a prepared password. Use it." So we might

Hi, I'm one of the Valkey Helm chart maintainers. We have implemented both inline and secret-based passwords in the chart. Note that it currently only supports standalone deployments with replicas coming soon.

kiblik:
@fernandezcuesta, couple more adjustments were needed (replace `template` with `include` or missing `.` in `template`), but the base was pretty good. Btw, I also added `redisPort` (just in case).

I do not have access to `rediss` or `sentinel` or any other external instance so I have not been able to test all scenarios. But `helm template ... --show-only templates/configmap.yaml | grep 'DD_CELERY_BROKER'` was adjusting values based on my expectations. So I hope we are good.

@fernandezcuesta, feel free to have one more look (resolve open conversations).

@Maffooch, @mtesauro, @rossops, this PR is ready for review.

kiblik:
> > So, my original idea was to replace Bitnami Redis with the [official Valkey Helm chart](https://github.com/valkey-io/valkey-helm). Unfortunately, this is not fully possible because authN issue. The official Valkey Helm chart does not support the approach: "This is a password which I generated/prepared for you. Use it". Valkey needs the `users.acl` file with the full definition of users. This issue is partially solved in the [open PR](https://github.com/valkey-io/valkey-helm/pull/14), but still not fully support "This is a prepared password. Use it." So we might
> 
> Hi, I'm one of the Valkey Helm chart maintainers. We have implemented both inline and secret-based passwords in the chart. Note that it currently only supports standalone deployments, with replicas coming soon.

@sgissi, thank you for your input. We decided to go with `CloudPirates` for a smoother transition.

rossops (approved):

mtesauro (approved):
Approved

Nice job! :rocket:

PR #13416, opened 2025-10-13
feat(session): Single user session

kiblik:
Some security policies require allowing users to have only one active session.
This PR allows activating this behaviour. Default behaviour is still the same (multiple sessions allowed).


valentijnscholten:
Just wondering what happens when a user uses the API. Will there be a session created as well?

kiblik:
> Just wondering what happens when a user uses the API. Will there be a session created as well?

This is a valid question. I do not know at this moment. I will test it.

kiblik:
> Just wondering what happens when a user uses the API. Will there be a session created as well?

I executed a test where the user used a token to access the API. He had not been logged off. He was only logged off when I logged in as the same user in a private session.
So this works well.

valentijnscholten (approved):

PR #13424, opened 2025-10-14
fix(helm): re-add annotation hint

kiblik:
Hint from #13397 was removed because of automation in the release process. Placing a hint outside of `annotation:` might help.


mtesauro (approved):
Approved

rossops (approved):

PR #13435, opened 2025-10-16
Ruff: PT - simplify rules

kiblik:
We are using a different style of tests (official Django tests), so it does not make sense to try to fix it.


mtesauro (approved):
Approved

Jino-T (approved):

PR #13437, opened 2025-10-16
Ruff: Fix N805

kiblik:
Fix N805 by making methods static or implementing them correctly.


mtesauro (approved):
Approved

valentijnscholten:
@kiblik would it make sense to double check the (force) change password still works?

mtesauro:
> @kiblik would it make sense to double check the (force) change password still works?

Great point @valentijnscholten I really doubt we have a test for that flow.

kiblik:
> @kiblik would it make sense to double check the (force) change password still works?

Well, I had not been able to test `enable_force_password_reset` in `dojo_ratelimit` (doing local DOS attack is possible but not sure if I need to go that far with testing).
But I tried to change value `force_password_reset` manually and `disable_force_password_reset` in `change_password` was working as expected (after sucessful change of password, value was cleaned).

PR #13250, opened 2025-09-23
Fix duplicated tags (everywhere)

fopina:
**Description**

Fixes https://github.com/DefectDojo/django-DefectDojo/issues/12918 - more details on [this comment](https://github.com/DefectDojo/django-DefectDojo/issues/12918#issuecomment-3325779302)

Reverts https://github.com/DefectDojo/django-DefectDojo/pull/12672/ as [the only change](https://github.com/radiac/django-tagulous/commit/d401e70cb62b873211ed5f85c7fd7f0ea1a8c577) in tagulous 2.1.1 seems to be introducing the issue: available tags added as options (already) then selected tags re-added as options (new "fix" in 2.1.1), duplicating tags.

Revert https://github.com/DefectDojo/django-DefectDojo/pull/13228 as no longer needed after this downgrade and also because that PR [broke available tag listing](https://github.com/DefectDojo/django-DefectDojo/pull/13228#issuecomment-3324597913)

~~This should also unblock https://github.com/DefectDojo/django-DefectDojo/pull/12601 as issue wasn't actually caused by it but by the tagulous upgrade that was done right before...~~
Update: tested bumping version to did it in this PR as well

**Test results**

* Create product A with tag A
* Create product B with tag B
* Edit product A, tag A is NOT duplicated
* Click tag field, tag B also shows as option (this was broken by https://github.com/DefectDojo/django-DefectDojo/pull/13228)
* Go to product list and filter by tag A and "apply filter"
* Verify tag A shows only once after reload

**Checklist**

This checklist is for your information.

- [ ] Make sure to rebase your PR against the very latest `dev`.
- [ ] Features/Changes should be submitted against the `dev`.
- [x] Bugfixes should be submitted against the `bugfix` branch.
- [x] Give a meaningful name to your PR, as it may end up being used in the release notes.
- [x] Your code is flake8 compliant.
- [x] Your code is python 3.11 compliant.
- [ ] If this is a new feature and not a bug fix, you've included the proper documentation in the docs at https://github.com/DefectDojo/django-DefectDojo/tree/dev/docs as part of this PR.
- [ ] Model changes must include the necessary migrations in the dojo/db_migrations folder.
- [ ] Add applicable tests to the unit tests.
- [ ] Add the proper label to categorize your PR.

PR #13440, opened 2025-10-16
Downgrade django-tagulous to 2.1.0

### LINKED ISSUE #12918: tags shown twice inside tags filter fields (on hold)

valentijnscholten:
Reapply changes from https://github.com/DefectDojo/django-DefectDojo/pull/13250

Fixes #12918

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -48,7 +48,8 @@ packageurl-python==0.17.5
 django-crum==0.7.9
 JSON-log-formatter==1.1.1
 django-split-settings==1.3.2
-django-tagulous==2.1.1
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

Maffooch (approved):
Merging since this is a dep update

PR #13441, opened 2025-10-16
Downgrade django-tagulous to 2.1.0

### LINKED ISSUE #12918: tags shown twice inside tags filter fields (on hold)

valentijnscholten:
Reapply changes from https://github.com/DefectDojo/django-DefectDojo/pull/13250

This is a twin PR to https://github.com/DefectDojo/django-DefectDojo/pull/13440 that goes into bugfix.

Adding this PR to go into `dev` to make sure it won't get lost.

Fixes #12918

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -48,7 +48,8 @@ packageurl-python==0.17.5
 django-crum==0.7.9
 JSON-log-formatter==1.1.1
 django-split-settings==1.3.2
-django-tagulous==2.1.1
+# do not upgrade to 2.1.1 - https://github.com/DefectDojo/django-DefectDojo/issues/12918
+django-tagulous==2.1.0
 PyJWT==2.10.1
 cvss==3.6
 django-fieldsignals==0.7.0


Maffooch (approved):
Merging since this is a dep update

fopina:
Thanks for the heads up on this @valentijnscholten 

Looking at the PR you mentioned, it looks like the change was left out, do you know if it was git merging or if it was manual conflict resolution?
I can imagine conflict resolution during this bugfix>dev merges might be quite messy.
Wondering if this has happened before and how often, whether it would make sense to think of something to prevent it 

Such as adopting these "twin PRs" as you did now, as process. It's what we do at work as well (every bug fix PR is opened on minor and major branches by the author)

valentijnscholten:
Lately we have been doing some bigger PRs and bigger release which makes these "merge master xxx into yyyy" more complex. We've decided now that these "operational" PRs should also undergo the review process to reduce the risk of regressions/mistakes. We do have quite an extensive test suite which has helped as well.

fopina:
That's true, tests introduced are less likely to be lost in conflict resolution and they'd pick up on the misses
Bad for the UI-only fixes 😄 
Tough PRs to review though!

Maybe there could be these twin PRs, but done via automation? When PR is done on `bugfix`, workflow cherrypicks the merge commit into a new PR for `dev` and assigns the author of the original PR for review?
With everything fresh in mind, quick to review the small change or resolve (few) conflicts there and takes that inital effort off you guys (maintainers). It could even be "pre-approved", so only the author needs to say ay and it gets merged immediately (and only if they change anything / resolve a conflict then maintainer approval required)

valentijnscholten:
I considered implementing nightly automated merges/PR-creation from bugfix to dev, but haven't discussed yet with the team.

PR #13442, opened 2025-10-16
tag based filtering: avoid duplicate rows in results

### LINKED ISSUE #13429: UI Filter Open Findings - Bug appear twice in the findings list
**Bug description**

If an active bug has two tags, for example "nonprod" and "prod", and the filter  "Finding Tag Contains" -> "prod" is selected (Open Findings) in the UI search, the same bug will appear twice in the results list.

Filter String Matching Optimization=active in the Settings.

valentijnscholten:
Because the `tags` field is a M2M relationship behind the scenes, this can result in duplicate rows in the result due to the nature of SQL JOINs.

This PR resolves this by applying DISTINCT when filtering on tag names.

```
                            # distinct has a performance impact, so only apply it if needed.
                            # we considered Postgress' DISTINCT ON, but it would enforce ordering by id
                            # we considered changing to an EXISTS subquery, but it would make
                            # our code dependant on the some of the django-tagulous internal.
```

Fixes #13429

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -147,10 +147,6 @@ class CharFieldFilterANDExpression(CharFieldInFilter):
         )
 
 
-class TagExistsIContainsFilter(CharFilter):
-    pass
-
-
 class FindingStatusFilter(ChoiceFilter):
     def any(self, qs, name):
         return qs
@@ -360,21 +356,30 @@ class DojoFilter(FilterSet):
 
     def filter_queryset(self, queryset):
         qs = super().filter_queryset(queryset)
-        try:
-            if hasattr(self, "form") and hasattr(self.form, "cleaned_data"):
-                for name, f in self.filters.items():
-                    field_name = getattr(f, "field_name", "") or ""
-                    # filtering on tag names would result duplicate rows, one for each matching tag
-                    if "tags__name" in field_name:
-                        value = self.form.cleaned_data.get(name, None)
-                        if value not in (None, "", [], (), {}):
-                            # distinct has a performance impact, so only apply it if needed.
-                            # we considered Postgress' DISTINCT ON, but it would enforce ordering by id
-                            # we considered changing to an EXISTS subquery, but it would make
-                            # our code dependant on the some of the django-tagulous internal
+        if hasattr(self, "form") and hasattr(self.form, "cleaned_data"):
+            for name, f in self.filters.items():
+                field_name = getattr(f, "field_name", "") or ""
+                # Only apply distinct for tag lookups that can duplicate base rows
+                if "tags__name" in field_name:
+                    value = self.form.cleaned_data.get(name, None)
+                    if value not in (None, "", [], (), {}):
+                        lookup_expr = getattr(f, "lookup_expr", None)
+                        is_exclude = getattr(f, "exclude", False)
+                        needs_distinct = (
+                            is_exclude
+                            or lookup_expr in {
+                                "in",
+                                "contains",
+                                "icontains",
+                                "startswith",
+                                "istartswith",
+                                "endswith",
+                                "iendswith",
+                            }
+                        )
+                        # exact/iexact typically won't duplicate rows
+                        if needs_distinct:
                             return qs.distinct()
-        except Exception:
-            return qs.distinct()
         return qs
 
 


Maffooch:
@fopina can you please take a look at this one? It may pique your interest 😄

dojo/filters.py
```diff
@@ -147,6 +147,10 @@ def filter(self, queryset, value):
         )
 
 
+class TagExistsIContainsFilter(CharFilter):
```
fopina: Is this filter picked up automatically by something or was it some experiment and forgotten?

dojo/filters.py
```diff
@@ -354,6 +358,25 @@ def __init__(self, *args, **kwargs):
...
                     self.form.fields[field].label = "Not " + self.form.fields[field].label
 
+    def filter_queryset(self, queryset):
+        qs = super().filter_queryset(queryset)
+        try:
+            if hasattr(self, "form") and hasattr(self.form, "cleaned_data"):
+                for name, f in self.filters.items():
+                    field_name = getattr(f, "field_name", "") or ""
+                    # filtering on tag names would result duplicate rows, one for each matching tag
+                    if "tags__name" in field_name:
```
fopina: Both `tags` and `tag` filters use `tags__name` field but only `tag` does a contains.
The exact match of `tags` should never yield duplicates as `tags__name` is unique, right?

Maybe a more generic would be `if filter lookup_expr is icontains and exclude is False and field is an m2m then use distinct`?
Not sure how to check for the latter (field is an m2m), need to inspect filter object to see if it is already bound to the actual model field (if so, it would be clean otherwise a bit hacky)
Of course restricting the distinct to "icontains" would only make sense for joins on *unique* fields. If it's a join on a non-unique field, it could yield duplicate results even with iexact matches...

Leaving some food for thought but maybe the simple condition (to cover all) would be `if not exclude and field is m2m`. It would apply to more cases then needed, compared to previous suggestion (eg: exact matches on unique relations) but the same as current condition, extended to any field

Not only covers other existing m2m fields (if there's any, didn't check), but also prepares for future fields that might drop into the models/filters *and* looks "better" (in sense it doesn't have "tags__name" hardcoded)

WDYT?

dojo/filters.py
```diff
@@ -354,6 +358,25 @@ def __init__(self, *args, **kwargs):
...
+                        value = self.form.cleaned_data.get(name, None)
+                        if value not in (None, "", [], (), {}):
+                            # distinct has a performance impact, so only apply it if needed.
+                            # we considered Postgress' DISTINCT ON, but it would enforce ordering by id
+                            # we considered changing to an EXISTS subquery, but it would make
```
fopina: It's a shame for DISTINCT ON 😄 it's the price to pay for the performance boost!

The `EXISTS` option however sounds appealing! What internals would it be pinned to? Just the `.tags.through`, right?

dojo/filters.py
```diff
@@ -354,6 +358,25 @@ def __init__(self, *args, **kwargs):
...
+                            # our code dependant on the some of the django-tagulous internal
+                            return qs.distinct()
+        except Exception:
```
fopina: What exceptions are expected here? Shouldn't we avoid these broad excepts? Especially as it's also returning a .distinct() (in case that would be the source for unexpected exceptions)

blakeaowens (approved):

dojo/filters.py
valentijnscholten: removed the leftover.

dojo/filters.py
valentijnscholten: just some stuff around mapping the fields through the m2m model. nothing crazy but I feel the current approach is easier to understand for everyone and has less chances or corner cases where some magical combination of filters/ordering might break if we rewrite the tag filter to an EXISTs query.

dojo/filters.py
valentijnscholten: removed the leftover

dojo/filters.py
valentijnscholten: I've narrowed the condition which triggers the distinct.

valentijnscholten:
@fopina Thanks for reviewing. To be honest initially I just planned on slapping on the `distinct()` always. But decided at least make it a little smarter. But I guess I should have upped my game before showing it to the Django guru's out there :-D . The bug has been around for 5 years before we got the first report of it, so I think the PR is now fine to deal with it.

fopina:
Far from guru sorry if I made myself try to sound like one..! I’d likely just slap the distinct() for everything and only notice when production list of findings continuously timed out

I guess tag contains is something rare to use as the whole point is to pinpoint them (even a list), makes sense to go unnoticed

thanks!

valentijnscholten:
I was just joking. I think it's good to have people around to keep me sharp.

Maffooch:
@fopina FWIW I am/was impressed with your insight 😄

PR #13460, opened 2025-10-18
Show unique id from tool together with hash_code in title elements

valentijnscholten:
To help with understanding deduplication, the `hash_code` was already shown when hovering over the ID in the view finding screen. This PR adds the `unique_id_from_tool` to help the user understand deduplication. Both are also shown when hovering over the status in the list of findings.


mtesauro (approved):
Approved

Issue #13461, opened 2025-10-19
`service` field cleared by UI import


valentijnscholten:
When uploading a Trivy report via the UI the `service` field set by the parser gets overwritten with '""' because it was not specified in the UI.

This is important because the `service` field is taken into account always for deduplication. 

Possible solutions:
- Implement a similar mechanism as for `active` and `verified` to let the user choose to overwrite (clear) the `service` field or use the default from the parser?
- If left empty, always use the value from the parser.
- Alternative: Do not allow parsers to set the `service` field. But there's possibly value in case of scan reports that contain multiple services.

yashvrathod:
Hi, I’d like to work on fixing this issue with the `service` field being cleared when uploading Trivy reports via the UI.

PR #13466, opened 2025-10-19
Fix service field overwrite issue #13461

Layyzyy:
**Description**

This PR fixes issue #13461 where the service field set by parsers (like Trivy) gets overwritten with an empty string when the field is not specified in the UI.

**Root Cause**
The problem occurred in `dojo/importers/default_importer.py` where the code checked `if self.service is not None:` before overwriting the finding's service field. When the UI form left the service field empty, it passed an empty string ("") to the importer, which is not None, so the condition was True and the parser's value was overwritten with the empty string.

**Solution**
Modified the condition in two locations to check both:
- `if self.service is not None and self.service != "":`

This ensures that:
1. If service is None (not provided), the parser's value is preserved
2. If service is an empty string (left empty in UI), the parser's value is preserved
3. If service is a non-empty string (explicitly set in UI), it overwrites the parser's value

**Changes**
- Modified `dojo/importers/default_importer.py` at lines 207 and 346
- Applied the fix to both `process_findings` and `close_old_findings` methods

**Test Results**
The fix addresses the core issue where empty service fields in the UI were overwriting parser values. This is important because the service field is used for deduplication.

**Impact**
This change ensures that when uploading reports via the UI (e.g., Trivy reports), leaving the service field empty will preserve the value set by the parser, rather than overwriting it with an empty string.

**Checklist**
- [x] Give a meaningful name to the PR
- [x] Code changes are minimal and focused
- [x] Bugfix submitted against appropriate branch

### DIFF
--- a/dojo/importers/default_importer.py
+++ b/dojo/importers/default_importer.py
@@ -204,7 +204,7 @@ class DefaultImporter(BaseImporter, DefaultImporterOptions):
             # scan_date was provided, override value from parser
             if self.scan_date_override:
                 unsaved_finding.date = self.scan_date.date()
-            if self.service is not None:
+            if self.service is not None and self.service != "":
                 unsaved_finding.service = self.service
 
             # Force parsers to use unsaved_tags (stored in below after saving)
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Thank you for the PR. We'll have to decide what the best solution is. As you can see the current fix in this PR breaks existing behaviour and affectts imports coming in from the API where the user might have sent the _empty string_ on purpose.

Do you have ideas about these cases and a better fix? 

A completely different option could be to not allow parsers to set the `service` field to avoid these kinds of confusions.

valentijnscholten:
@Layyzyy Thank you for the PR, but we chose another approach: #13517

PR #13489, opened 2025-10-21
Advance OIDC to enable groups mapping

manuel-sommer:
- Implement OIDC groups to be parsed. 
- Added Unittests for update_oidc_groups and update_azure_groups

### DIFF
--- a/.github/workflows/release-1-create-pr.yml
+++ b/.github/workflows/release-1-create-pr.yml
@@ -98,7 +98,7 @@ jobs:
           chart-search-root: "helm/defectdojo"
 
       - name: Push version changes
-        uses: stefanzweifel/git-auto-commit-action@28e16e81777b558cc906c8750092100bbb34c5e3 # v7.0.0
+        uses: stefanzweifel/git-auto-commit-action@04702edda442b2e678b25b537cec683a1493fcb9 # v7.1.0
         with:
           commit_user_name: "${{ env.GIT_USERNAME }}"
           commit_user_email: "${{ env.GIT_EMAIL }}"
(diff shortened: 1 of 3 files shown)


manuel-sommer:
@valentijnscholten could you please take a look here?

runderwoodcr14:
This feature is really needed, I think will make a great leap towards Cloud Native using CNCF landscape

quirinziessler:
Is there any update on this? When can I expect being this being part of the main versions? Checking out the branch it works without issues. Would appreciate this getting merged soon and really making use of DD's IdP integration.

devGregA:
Hi @manuel-sommer, sorry for the delay here. Thank you for your pull request and contributions to Dojo.

@mtesauro and @Maffooch pinged me on this one a while ago, so totally my fault for the delay. After considering for quite some time, I’m of the opinion  a feature like this isn’t something we should be maintaining in OS for several reasons:

1. We can’t reasonably maintain tests and functionality around this feature. One area DefectDojo shines and we regularly get constant positive feedback on is the feature set we have works as advertised. Maintaining test coverage, accounts around this feature, etc, would be a major burden for our project maintainers.

2. Maintaining this feature would essentially be subsidizing enterprises that can afford to support DefectDojo and be a Pro customer. One of the things we’re always trying to balance is OS vs Pro. If we don’t generate enough biz, we can’t maintain open source. If a company is large enough to use SSO with groups, it seems reasonable they should be on Pro, which in turn helps us to maintain open source, rather than making it more difficult to maintain as outlined in 1.

manuel-sommer:
Hi @devGregA , 
would this PR be accepted if I remove the update_oidc_groups method? So, just the OIDC migration?

PR #13491, opened 2025-10-21
Deduplicate findings in batches

valentijnscholten:
Traditionally Defect Dojo has been deduplicating (new) findings one-by-one. This works well for small imports and has the benefit of an easy to understand codebase and test suite.

For larger imports however the performance is bad and resource usage is (very) high. A 1000+ finding import can cause a celery worker to spend _minutes_ on deduplication.

This PR changes the deduplication process for import and reimport to be done in batches. This biggest benefit is that there now will be 1 database query per batch (1000 findings), instead of 1 query per finding (1000 queries).

During the development of the PR I realized:
- more test cases were needed: [13499](https://github.com/DefectDojo/django-DefectDojo/pull/13499), [13464](https://github.com/DefectDojo/django-DefectDojo/pull/13464), [13463](https://github.com/DefectDojo/django-DefectDojo/pull/13463), [13372](https://github.com/DefectDojo/django-DefectDojo/pull/13372).
- some deduplication bugs needed fixing: [13513](https://github.com/DefectDojo/django-DefectDojo/pull/13513)

Although batching dedupe sounds like a simple PR, the caveat is that with the one-by-one deduplication the result of the deduplication of the first finding in the report can have an affect on the deduplication result of the next findings (if there are duplicates inside the same report). This should be a corner case and usually means the deduplication configuration need some fine tuning. Nevertheless we wanted to make not to cause unexpected/different behavior here. The new tests should cover this.

The PR splits the deduplication process in three parts:
1) Finding possible candidates
2) Match the (new) finding against the candidates
3) Act upon it if a match is found

One of the reasons for doing this is that we want to use the exact same matching logic for the `reimport` process. Currently that has an _almost_ identical matching algorithm, but with minor unintentional differences. Once this PR here has proven itself, we will adjust the reimport process. Next to the "reimport matching" the reimport process also deduplicates new findings. This part is already using the batchwise deduplication in this PR.

A quick test with the `jfrog_xray_unified/very_many_vulns.json` samples scan (10k findings) shwo the obvious huge improvement in deduplication time. Please note that we're not only doing this for performance, but also to reduce the resources (cloud cost) needed to run Defect Dojo.

initial import (no duplicates):
| branch | import time | dedupe time | total time | 
|--------|:-----------:|:-----------:|:-----------:|
| dev | ~200s | ~400s | ~600s |
| dedupe-batching | ~190s | _~12s_ | ~200s |

second import into the same product (all duplicates):
initial import (no duplicates):
| branch | import time | dedupe time | total time | 
|--------|:-----------:|:-----------:|:-----------:|
| dev | ~200s | ~400s | ~600s |
| dedupe-batching | ~190s | _~180s_ | ~370s |

Imagine what this can do for `reimport` performance if we switch that to batch mode.

### DIFF
--- a/dojo/finding/deduplication.py
+++ b/dojo/finding/deduplication.py
@@ -529,7 +529,6 @@ def _dedupe_batch_legacy(findings):
 
 def dedupe_batch_of_findings(findings, *args, **kwargs):
     """Batch deduplicate a list of findings. The findings are assumed to be in the same test."""
-
     # Pro has customer implementation which will call the Pro dedupe methods, but also the normal OS dedupe methods.
     from dojo.utils import get_custom_method  # noqa: PLC0415 -- circular import
     if batch_dedupe_method := get_custom_method("FINDING_DEDUPE_BATCH_METHOD"):


Maffooch (approved):

mtesauro (approved):
Approved

dojo/finding/deduplication.py
```diff
@@ -0,0 +1,561 @@
...
+
+    enabled = System_Settings.objects.get().enable_deduplication
+
+    if enabled:
+        # sort findings by id to ensure deduplication is deterministic/reproducible
+        findings = sorted(findings, key=attrgetter("id"))
+
+        from dojo.utils import get_custom_method  # noqa: PLC0415 -- circular import
+        if batch_dedupe_method := get_custom_method("FINDING_DEDUPE_BATCH_METHOD"):
+            batch_dedupe_method(findings, *args, **kwargs)
```
dogboat: Sorry if I'm not interpreting this correctly -- do we need to return after this, if a custom method is found and used?

valentijnscholten: Very good point. I think we need to `return` as the Pro method will perform all dedupe, including the OS algorithms. Will update the PR and run tests again.

valentijnscholten: Converted to draft while I make the finishing touches.

valentijnscholten: @dogboat adjust the code to jump to the custom implementation upon function entry and return afterwards. all tests passing also in pro with the fix from https://github.com/DefectDojo/django-DefectDojo/pull/13680 also merged here. Can you double check?

PR #13502, opened 2025-10-22
reimplement favicons, update theme

paulOsinski:
Note: this PR requires that we upgrade our Hugo to **[v0.152.1](https://github.com/gohugoio/hugo/releases/tag/v0.152.1)**.  These are breaking changes and our site will not build without that upgrade.

This upgrades our documentation to use v1.8 of Doks, our theme, which I followed according to their documentation: https://getdoks.org/docs/upgrade/guides/doks-v1.8/

I can confirm this means Favicons work, and the site builds as long as you're using the latest version of Hugo.  Favicons do not go in the `../static/favicons` folder anymore, they live in `../assets` and one of the plugins here will automatically create the required favicons when you run in production.

[sc-11968]

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Setup Hugo
         uses: peaceiris/actions-hugo@75d2e84710de30f6ff7268e08f310b60ef14033f # v3.0.0
         with:
-          hugo-version: '0.140.1' # renovate: datasource=github-releases depName=gohugoio/hugo versioning=loose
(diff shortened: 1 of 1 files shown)


Maffooch (changes requested):
Will need to update Hugo version in a few places:
- https://github.com/DefectDojo/django-DefectDojo/blob/e31214aac5067810abf094860ee172b59f785dc3/.github/workflows/validate_docs_build.yml#L15-L16
- https://github.com/DefectDojo/django-DefectDojo/blob/e31214aac5067810abf094860ee172b59f785dc3/.github/workflows/gh-pages.yml#L18-L19

mtesauro (approved):
Approved

PR #13505, opened 2025-10-23
Highlight weekends in the calendar..(Highlight Saturdays and Sundays in the calendar)

Wanderer65:
This PR adds a visual highlight to weekends in the calendar view. 
Saturdays and Sundays now have a light yellow background, making it easier for users to distinguish weekends at a glance.

- Updated `calendar.html` to include `dayRender` logic for weekend highlighting
- Used FullCalendar’s `dayRender` callback for applying styles
- Works for both Engagements and Tests calendar views

No changes to backend logic; purely frontend enhancement.

### DIFF
--- a/components/package.json
+++ b/components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "defectdojo",
-  "version": "2.51.2",
+  "version": "2.52.0-dev",
   "license" : "BSD-3-Clause",
   "private": true,
   "dependencies": {
--- a/helm/defectdojo/Chart.yaml
+++ b/helm/defectdojo/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: "2.51.2"
+appVersion: "2.52.0-dev"
 description: A Helm chart for Kubernetes to install DefectDojo
 name: defectdojo
-version: 1.7.2
+version: 1.7.3-dev
 icon: https://defectdojo.com/hubfs/DefectDojo_favicon.png
 maintainers:
   - name: madchap
@@ -33,11 +33,5 @@ dependencies:
 #   - kind: security
 #     description: Critical bug
 annotations:
-  artifacthub.io/prerelease: "false"
-  artifacthub.io/changes: |
-    - kind: fixed
-      description: Drop initialDelaySeconds if eq. zero
-    - kind: added
-      description: Add support for automountServiceAccountToken
-    - kind: changed
-      description: Bump DefectDojo to 2.51.2
+  artifacthub.io/prerelease: "true"
+  artifacthub.io/changes: ""
--- a/helm/defectdojo/README.md
+++ b/helm/defectdojo/README.md
@@ -495,7 +495,7 @@ kubectl delete pvc data-defectdojo-redis-0 data-defectdojo-postgresql-0
 
 # General information about chart values
 
-![Version: 1.7.2](https://img.shields.io/badge/Version-1.7.2-informational?style=flat-square) ![AppVersion: 2.51.2](https://img.shields.io/badge/AppVersion-2.51.2-informational?style=flat-square)
+![Version: 1.7.3-dev](https://img.shields.io/badge/Version-1.7.3--dev-informational?style=flat-square) ![AppVersion: 2.52.0-dev](https://img.shields.io/badge/AppVersion-2.52.0--dev-informational?style=flat-square)
 
 A Helm chart for Kubernetes to install DefectDojo
 


dojo/templates/dojo/calendar.html
```diff
@@ -1,7 +1,12 @@
-{% extends 'base.html' %}
 {% load static %}
 {% load i18n %}
 
+{% block head_extra %}
+    {{ block.super }}
+    <!-- fullCalendar CSS from CDN -->
+    <link href="https://cdn.jsdelivr.net/npm/fullcalendar@3.10.5/dist/fullcalendar.min.css" rel="stylesheet" />
```
valentijnscholten: can you switch to locally bundled js libs please?

dojo/templates/dojo/calendar.html
```diff
@@ -32,59 +37,74 @@
...
     <br/><br/>
 {% endblock %}
+
 {% block postscript %}
     {{ block.super }}
+    <!-- jQuery and fullCalendar JS from CDN -->
+    <script src="https://code.jquery.com/jquery-3.6.0.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/moment@2.29.4/moment.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/fullcalendar@3.10.5/dist/fullcalendar.min.js"></script>
+
```
valentijnscholten: can you switch to locally bundled js libs please?

dojo/templates/dojo/calendar.html
```diff
@@ -32,59 +37,74 @@
...
             $('#lead').val([{% for lead in leads %} '{{ lead }}', {% endfor %}]);
+
             $('#calendar').fullCalendar({
                 header: {
                     left: 'prev,next today',
                     center: 'title',
                     right: 'month,basicWeek,basicDay'
                 },
                 editable: false,
-                eventLimit: true, // allow "more" link when too many events
```
valentijnscholten, line 54: can you retain the comment please

dojo/templates/dojo/calendar.html
```diff
@@ -1,49 +1,32 @@
-{% extends 'base.html' %}
```
valentijnscholten, line 1: why is this changed/removed?

dojo/templates/dojo/calendar.html
```diff
@@ -1,49 +1,32 @@
...
 {% load static %}
 {% load i18n %}
 
-{% block content %}
+{% block head_extra %}
```
valentijnscholten, line 5: why is this changed/removed?

dojo/templates/dojo/calendar.html
```diff
@@ -1,49 +1,32 @@
...
-                </div>
-                <div style="display: inline-block;">
-                    <input class="btn btn-primary" type="submit" value="Apply" />
-                </div>
-            </div>
-        </div>
-    </form>
-    <br/><br/>
-    <div id="calendar"></div>
-    <br/><br/>
```
valentijnscholten, line 33: why is this changed/removed?

valentijnscholten (changes requested):
Some comments and can you base the PR against the `bugfix` branch as per the PR instructions?

Kepvion:
Hi @valentijnscholten , I checked the remote repository and there is no bugfix branch, only master. Should I base the PR on master instead, or is there another branch I should use?

valentijnscholten:
This branch: https://github.com/DefectDojo/django-DefectDojo/tree/bugfix. In your local clone you should be able to do something like:

`git checkout bugfix`

Kepvion:
I checked the remote repository, but there doesn’t appear to be a bugfix branch ..

valentijnscholten:
How did you create your fork? Looks like it only has the `master` branch. For now it looks OK as bugfix and master are not too different. I've changed the base branch here on the PR.

valentijnscholten:
But it looks like your vendoring the libraries? The goal is that any new packages must be installed via yarn.

Kepvion:
Hi @valentijnscholten , thanks for the feedback! I understand that instead of vendoring the libraries, I should install them via yarn. Should I update the PR to use yarn-managed FullCalendar, or is it okay to keep it vendored for this change?

valentijnscholten:
You need to make the PR in line with the existing codebase which has all dependencies in npm/yarn.

valentijnscholten:
Hello @Wanderer65, Do you need more gelp/guidance?

Kepvion:
Well,I am trying check if that is good?? Mayeb aftet that I might need help 😅

Kepvion:
@valentijnscholten yeah please Guidance?

valentijnscholten:
@Wanderer65 At first glance it looked you wanted to add a new npm/yarn package `fullcalendar`. But that's already in place in the project. Can you not include it in the pages needed similar to what is done in `base.html`:

`<script src="{% static "fullcalendar/dist/fullcalendar.min.js" %}"></script>`

If it's in `base.html` already, it should be available almost anywhere in Defect Dojo already. I see that `calendar.html` extends `base.html`, so the `js` and `css` should already be loaded without doing any link or script tags.
Maybe they are needed to load moment. And jquery, but that should also already be loaded.
But moment and jquery are not present in the fullcalendar package that is included with Defect Dojo.
Maybe they are only available in newer versions of fullcalendar? Looks like the latest is v6 and Defect Dojo is v3.

You could attempt to upgrade it and make it work, but I'm not sure if the calendar is that high on the list of priorities for the Defect Dojo team.

You could run `yarn` inside the `components` directory to add new packages, etc. You should commit the resulting package-lock.json and yarn.lock files.

Maffooch:
Closing as stale

PR #13509, opened 2025-10-23
fix(renovate): Clean records

kiblik:
- track `node-version` is not needed. Renovate knows this. See https://github.com/DefectDojo/django-DefectDojo/pull/13508/files
- tune `matchStrings`


mtesauro (approved):
Approved

PR #13514, opened 2025-10-23
scan_added_empty.tpl: fix symlink problem

### LINKED ISSUE #13493: scan_added_empty is a symbolic link to nothing
### Discussed in https://github.com/DefectDojo/django-DefectDojo/discussions/13486

<div type='discussions-op-text'>

<sup>Originally posted by **David-Haendler** October 21, 2025</sup>
Hi!

I think I might have found a small bug, but I'm not sure. 
It seems to me that 
dojo/templates/notifications/alert/scan_added_empty.tpl
is a symbolic link that points to nothing. This seems like a bad practice, unless it has some other purpose I can't figure out?</div>

valentijnscholten:
Fixes #13493 scan_added_empty had turned into a symlink to nothing. Fixed by adding original content.

### DIFF
--- a/dojo/templates/notifications/alert/scan_added_empty.tpl
+++ b/dojo/templates/notifications/alert/scan_added_empty.tpl
@@ -1,1 +1,1 @@
-{% include "notifications/alert/scan_added.tpl" %}
+{% include notifications/alert/scan_added.tpl %}


Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #13521, opened 2025-10-23
endpoint import optimize

valentijnscholten:
The way endpoints are handled during import has a large impact on performance.
The fastest way would be to handle them in batches, but that's not a trivial change.
There are some low hanging fruits that we can implement. 
This PR and already reduces the number of queries significantly.
The Atomic wrapper has been removed as it doesn't do anything in this scenario with only one select and one write.


valentijnscholten:
wdyt @kiblik ("endpoint guru")

fopina:
Quickly looking at the change, looks like a good place for (shameless self promo) https://github.com/fopina/django-bulk-update-or-create

valentijnscholten:
Interesting. My search for a solution didn't bring me your library. I did think about writing similar logic, but decided not to because A) the endpoint model doesn't have a single primary key field B) there probably will be some big changes coming around endpoints early next year which could make the investment right now useless. But good to know this library, it might be handy elsewhere or later on after the big endpoint change.

PR #13523, opened 2025-10-24
:bug: add middleware to handle social auth provider unavailability gracefully

manuel-sommer:
grafik

--> 
grafik

grafik
grafik


valentijnscholten:
Is there not a better/cleaner way to do this, for example by subclassing the social auth middleware?
It feels like a lot of extra code and configuration to handle a case which is probably not even that common.
I think name, path_prefix, check_path are not needed. The rest can be loaded straight from settings?
I also wonder if the standard login is always available as fallback.

manuel-sommer:
> Is there not a better/cleaner way to do this, for example by subclassing the social auth middleware? It feels like a lot of extra code and configuration to handle a case which is probably not even that common. I think name, path_prefix, check_path are not needed. The rest can be loaded straight from settings? I also wonder if the standard login is always available as fallback.

I adapted the code. Also, the standard login is always available as fallback.

valentijnscholten:
Thanks for the efforts @manuel-sommer , but did you consider other approaches? I am not yet convinced this middleware solution is the best way to implement this. When I look at the Django Social Auth app I see some pointers around customizing the middleware and exception handling:

- https://github.com/omab/django-social-auth/blob/699571a3b6d84aa2c25ffce59daa4c7f3559a0e4/example/example/middleware.py
- https://github.com/omab/edx-platform/blob/f098633cee8b941d04e1beb1ff3b3acd5a3511e4/common/djangoapps/third_party_auth/middleware.py
- https://github.com/omab/django-social-auth/blob/699571a3b6d84aa2c25ffce59daa4c7f3559a0e4/doc/configuration.rst#exceptions-middleware

Could you take a look and/or convince us that the proposed solution in this PR is the best fit?

manuel-sommer:
> Thanks for the efforts @manuel-sommer , but did you consider other approaches? I am not yet convinced this middleware solution is the best way to implement this. When I look at the Django Social Auth app I see some pointers around customizing the middleware and exception handling:
> 
>     * https://github.com/omab/django-social-auth/blob/699571a3b6d84aa2c25ffce59daa4c7f3559a0e4/example/example/middleware.py
> 
>     * https://github.com/omab/edx-platform/blob/f098633cee8b941d04e1beb1ff3b3acd5a3511e4/common/djangoapps/third_party_auth/middleware.py
> 
>     * https://github.com/omab/django-social-auth/blob/699571a3b6d84aa2c25ffce59daa4c7f3559a0e4/doc/configuration.rst#exceptions-middleware
> 
> 
> Could you take a look and/or convince us that the proposed solution in this PR is the best fit?

Thank you for the guidance here @valentijnscholten. Learned something new :-) 
How about the PR now?

valentijnscholten:
Looks nice and clean @manuel-sommer  :-) I don't have a way to test this. Dis you manage to test these specific types or errors? We need to make sure we don't break the SSO for the Pro clients :-)

manuel-sommer:
@valentijnscholten, I implemented a unittest, but don't have the chance to test all different possibilities in a real environment.

dojo/middleware.py
```diff
@@ -75,6 +80,20 @@ def __call__(self, request):
         return self.get_response(request)
 
 
+class CustomSocialAuthExceptionMiddleware(SocialAuthExceptionMiddleware):
+    def process_exception(self, request, exception):
+        if isinstance(exception, requests.exceptions.RequestException):
+            messages.error(request, "Login via social authentication is temporarily unavailable. Please use the standard login below.")
+            return redirect("/login")
```
kiblik: If `DD_SOCIAL_LOGIN_AUTO_REDIRECT` is enabled, there is a high chance that the user can get stuck in an infinite loop of redirects. I would recommend replacing all `redirect("/login")` with `redirect("/login?force_login_form")`.

dojo/middleware.py
```diff
@@ -75,6 +80,20 @@ def __call__(self, request):
         return self.get_response(request)
 
 
+class CustomSocialAuthExceptionMiddleware(SocialAuthExceptionMiddleware):
+    def process_exception(self, request, exception):
+        if isinstance(exception, requests.exceptions.RequestException):
+            messages.error(request, "Login via social authentication is temporarily unavailable. Please use the standard login below.")
```
kiblik: Standard login might not be enabled (`DD_SOCIAL_AUTH_SHOW_LOGIN_FORM=False`). It might be good to say, _"Please use the standard login below."_

mtesauro (approved):
Approved

blakeaowens (approved):

mtesauro:
@kiblik Are you good with this? Wanted to ask before we merged it since you had requested changes that haven't been dismissed.

Maffooch:
@kiblik I'm gonna go ahead and merge this one so that the changes can be used in https://github.com/DefectDojo/django-DefectDojo/pull/13572

PR #13528, opened 2025-10-24
[docs] moving Parser Docs to new index

paulOsinski:
This PR moves the parser documentation to a new URL index `https://docs.defectdojo.com/en/connecting_your_tools/parsers/` => `https://docs.defectdojo.com/supported_tools` to clean things up.

Primarily intended to make the documentation more navigable, because our list of parsers overflows the screen when it's expanded from the main documentation.  This moves things into a more walled-off 'appendix' like section while keeping the navigation the same.

Primarily affects the nav - left side is the new version, right side is the old (where Parser docs are mashed in with everything else).
Screenshot 2025-10-24 at 3 50 06 PM

- merged with "hugoupdate" branch so that I could preview the changes.
[sc-11972]

### DIFF
--- a/unittests/test_parsers.py
+++ b/unittests/test_parsers.py
@@ -29,7 +29,7 @@ class TestParsers(DojoTestCase):
                 "sysdig_common",  # common classes for sysdig parsers
             }:
                 with self.subTest(parser=parser_dir.name, category="docs"):
-                    doc_file = Path(basedir) / "docs" / "content" / "en" / "connecting_your_tools" / "parsers" / category / f"{doc_name}.md"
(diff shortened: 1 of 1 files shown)


Jino-T (approved):

paulOsinski:
@valentijnscholten I believe I've sorted this out [here](https://github.com/DefectDojo/django-DefectDojo/pull/13528/commits/a8b60f61b51cb90735e6d65ff2ebc33693bdbf3d) but I haven't been able to run the unittests locally yet.

valentijnscholten:
I was about to quickly fix it, but there are 38 hits when I search for `en/connecting_your_tools` or `connecting_your_tools` . Probably most of them need to be changed?

paulOsinski:
@valentijnscholten seems to be working now 🤞

PR #13542, opened 2025-10-27
Added Ability to Edit found_by value in API

Jino-T:
[sc-11934]

Tested and functional in API & Pro UI

**Not able to remove found_by value associated with finding's test.** 

- Ex: If finding was imported with Semgrep parser, you are unable to remove Semgrep from the found_by field. 
- A separate portion of code must check to make sure that this value is always present in found_by. 
- Let me know if you want me to look into this further and allow this functionality.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1779,16 +1779,14 @@ class FindingSerializer(serializers.ModelSerializer):
 
         # Get found_by from validated_data
         found_by = validated_data.pop("found_by", None)
-
         # Handle updates to found_by data
         if found_by:
-            instance.found_by.clear()
-            instance.found_by.add(*found_by)
+            instance.found_by.set(found_by)
         # If there is no argument entered for found_by, the user would like to clear out the values on the Finding's found_by field
         # Findings still maintain original found_by value associated with their test
-        else:
+        # In the event the user does not supply the found_by field at all, we do not modify it
+        elif isinstance(found_by, list) and len(found_by) == 0:
             instance.found_by.clear()
-
         instance = super().update(
             instance, validated_data,
         )


valentijnscholten:
I wonder how this field is used/edited? In the code it looks like it's only used to maintain a list of test types that reported this finding via the set of duplicates?

dojo/api_v2/serializers.py
```diff
@@ -1774,6 +1777,18 @@ def update(self, instance, validated_data):
...
+
+        # Handle updates to found_by data
+        if found_by:
+            instance.found_by.clear()
+            instance.found_by.add(*found_by)
+        # If there is no argument entered for found_by, the user would like to clear out the values on the Finding's found_by field
+        # Findings still maintain original found_by value associated with their test
+        else:
+            instance.found_by.clear()
+
```
Maffooch:
```suggestion
        # Get found_by from validated_data
        found_by = validated_data.pop("found_by", None)
        # Handle updates to found_by data
        if found_by:
            instance.found_by.set(found_by)
        # If there is no argument entered for found_by, the user would like to clear out the values on the Finding's found_by field
        # Findings still maintain original found_by value associated with their test
        # In the event the user does not supply the found_by field at all, we do not modify it
        elif isinstance(found_by, list) and len(found_by) == 0:
            instance.found_by.clear()
```

Updated the following:
- Explicit check for when `found_by` is not found in the payload
- Use `set` rather then `add` in the event a user replaces a list of test types with a totally different set

Maffooch (changes requested):

Jino-T:
Good idea Cody. I tested your suggestion and it works in the API.

Maffooch:
Not sure if three approvals is quite enough...

PR #13548, opened 2025-10-27
Fix recipient handling in create_notification method

dorkdiaries9:
### Summary
Adds a defensive check in `create_notification()` to skip processing when `recipients` is provided but empty.

### Details
- Prevents unnecessary calls to `_process_recipients()`
- Adds debug log for traceability

### Testing
Validated that:
- Notifications continue working normally for valid recipient lists
- Debug log appears and processing stops for empty lists

### Type of Change
- [x] Bugfix (non-breaking change)

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -9,18 +9,18 @@
       "version": "1.8.0",
       "license": "MIT",
       "dependencies": {
-        "@docsearch/css": "^4.2.0",
-        "@docsearch/js": "^4.2.0",
-        "@tabler/icons": "^3.34.1",
-        "@thulite/doks-core": "^1.8.3",
-        "@thulite/images": "^3.3.1",
-        "@thulite/inline-svg": "^1.2.0",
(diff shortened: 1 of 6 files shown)


valentijnscholten:
@dorkdiaries9 Thanks for the PR. Does it fix an actual bug or is it more defensive/optimization. I haven't looked at the code, but I wonder where this notification for 0 recipients is triggered.

dorkdiaries9:
> Thanks for the PR. Does it fix an actual bug or is it more defensive/optimization. I haven't looked at the code, but I wonder where this notification for 0 recipients is triggered.

Thanks for reviewing!
This change prevents unnecessary notification processing when the recipients list is empty.
Earlier, the method continued execution even if recipients was an empty list, which could lead to redundant logs or unintended calls to _process_recipients.
So, this is a small bug fix to ensure that we only proceed when there are valid recipients.

valentijnscholten (approved):

PR #13563, opened 2025-10-29
feat: Add dry_run support for reimport-scan operations

Siryu6:
## Description

This PR implements a **dry_run feature for reimport-scan operations** that allows users to preview what changes would occur during a reimport without making any database modifications.

**Key Features:**
- Optional `dry_run` boolean parameter (defaults to `false`) on reimport-scan API endpoints
- Returns detailed statistics about findings that would be created, reactivated, closed, or left untouched
- Respects all existing reimport parameters (`close_old_findings`, `do_not_reactivate`, severity filters, etc.)
- Non-breaking: all existing functionality remains unchanged

**Use Case:**
This feature is particularly useful for CI/CD pipelines where teams want to validate scan results and preview changes before merging to production environments.

**API Response Example:**
```json
{
  "test_id": 123,
  "dry_run": true,
  "changes_preview": {
    "would_create": 5,
    "would_reactivate": 2,
    "would_close": 3,
    "would_leave_untouched": 10,
    "total_changes": 10
  },
  "findings_details": {
    "new_findings": [...],
    "reactivated_findings": [...],
    "closed_findings": [...],
    "unchanged_findings": [...]
  }
}
```

## Documentation

API documentation is included in the serializer field descriptions (`help_text`).

## Files Changed

- `dojo/api_v2/serializers.py` - API integration for dry_run parameter
- `dojo/importers/default_reimporter.py` - Core dry_run_reimport() logic
- `dojo/importers/default_importer.py` - Return signature update
- `dojo/importers/options.py` - dry_run parameter validation
- `dojo/fixtures/dojo_testdata.json` - Fixed missing SLA_Configuration
- `dojo/engagement/views.py` - View layer dry_run support
- `dojo/test/views.py` - View layer dry_run support
- `unittests/dojo_test_case.py` - Test helper methods
- `unittests/test_import_reimport.py` - Minor test adjustments
- `unittests/test_import_reimport_dry_run.py` - New comprehensive test suite

**Labels to add:**
- `enhancement`
- `feature`


valentijnscholten (changes requested):
Thank you for the PR. We'll take a look and see if the feature is something that we can add without adding too much maintenance burden. Although we have lifted the feature freeze, it's still advised to double check before starting a PR.

Some initial thoughts:
- More than half of the changes in the PR seem to be cosmetic. Would it be possible to revert these cosmetic changes so it becomes easier to digest and review?
- The dry_run now seems to be a code path seperate from the "real" reimport. Have you considered using the rel reimport code path and add some if-statements to make a dry_run mode? Two code paths increase the risk of differences/mismatches.
- The reimport matching logic can be influenced by finding that are created during reimport. For example if finding 100 and 101 in the report are new and both have the same hash_code, the finding 101 will not be imported as it is matched against finding 100. In dry_run mode this match won't happen, so the statistics might be different. I haven't verified this, but this is what I am thinking.
- The reimport process might change in the near future where it will perform batch based matching and/or batch processing of the report. I am not sure if it will be possible to maintain a 100% complete dry_run feature.

Siryu6:
> More than half of the changes in the PR seem to be cosmetic. Would it be possible to revert these cosmetic changes so it becomes easier to digest and review?

Yeah my bad, that's my IDE 😅 I'll revert that. 

> The dry_run now seems to be a code path seperate from the "real" reimport. Have you considered using the rel reimport code path and add some if-statements to make a dry_run mode? Two code paths increase the risk of differences/mismatches.

I first did that but it was lot of if conditions, didn't like it and decided to separate it, if you prefer this i can switch to that. 

> The reimport matching logic can be influenced by finding that are created during reimport. For example if finding 100 and 101 in the report are new and both have the same hash_code, the finding 101 will not be imported as it is matched against finding 100. In dry_run mode this match won't happen, so the statistics might be different. I haven't verified this, but this is what I am thinking.

I'll verify its properly implemented and fix it if it's not the case. 

> The reimport process might change in the near future where it will perform batch based matching and/or batch processing of the report. I am not sure if it will be possible to maintain a 100% complete dry_run feature.

I guess that's a good reason to switch to if statements 😅

Siryu6:
Thank you for the detailed feedback! I've addressed all three concerns:
   
   1. **Cosmetic changes**: Reverted in commit c5135d8c7
   2. **Duplicate code path**: Refactored to use single path with conditionals in commit 366458289
   3. **Finding matching within report**: Implemented in-memory deduplication in commit 23a9b498f
   
The dry_run mode now uses the same code path as regular reimport and accurately simulates deduplication, including matches within the same scan report.

valentijnscholten:
Still lots of formatting changes. We're going to have to put this on hold until after the reimport changes we've planned for Q1.

Maffooch:
Marking as draft until we are ready for this one

Siryu6:
Hi @valentijnscholten 👋 

Looking for update about "reimport changes planned for Q1" 😅 
I still think this feature would be nice and be glad to adapt this to your changes whenever it's ready 😄

Maffooch:
Hi @Siryu6! We've decided to hold on this effort here indefinitely

PR #13567, opened 2025-10-29
chore: update notify-pr-reviewers-action to always use the latest version

Maffooch:
The original introduction of this action did not pin to a specific commit hash or version intentionally. This PR restores that decision


mtesauro (approved):
Approved.

This will fix a PR reviewer tool we wrote so thanks for catching this whoops @Maffooch

Maffooch:
Merging as this is a dep update with 2 approvals

PR #13572, opened 2025-10-30
:bug: Robustify create_user to handle None value

manuel-sommer:
grafik

### DIFF
--- a/dojo/pipeline.py
+++ b/dojo/pipeline.py
@@ -183,10 +183,6 @@ def sanitize_username(username):
 def create_user(strategy, details, backend, user=None, *args, **kwargs):
     if not settings.SOCIAL_AUTH_CREATE_USER:
         return None
-    username = details.get("username")
-    if not username:
-        username = details.get("email")
-    if not username:
-        username = details.get("fullname")
+    username = details.get(settings.SOCIAL_AUTH_CREATE_USER_MAPPING)
     details["username"] = sanitize_username(username)
     return social_core.pipeline.user.create_user(strategy, details, backend, user, args, kwargs)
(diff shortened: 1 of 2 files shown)


manuel-sommer:
> Are you sure you only want to log this as warning? Error and gracefully stopping the login process would be better imo.

Thank you @quirinziessler for your review

Maffooch:
Would it be possible to raise this error in the UI by using [the middleware introduced in your other PR](https://github.com/DefectDojo/django-DefectDojo/pull/13523)? Not sure if this pipeline function would be too late in the process or not

manuel-sommer:
Ok, merge the other PR and I will check if I can add it there with rebasing this PR.

manuel-sommer:
I don't want to add an additional functionality after everyone has approved and right before it gets merged.

manuel-sommer:
Done @Maffooch, it works solid, you can review

Maffooch (approved):
@manuel-sommer nice work 😄

mtesauro (approved):
Approved

dojo/pipeline.py
```diff
@@ -183,5 +183,10 @@ def sanitize_username(username):
 def create_user(strategy, details, backend, user=None, *args, **kwargs):
     if not settings.SOCIAL_AUTH_CREATE_USER:
         return None
-    details["username"] = sanitize_username(details.get("username"))
+    username = details.get("username")
+    if not username:
+        username = details.get("email")
+    if not username:
+        username = details.get("fullname")
+    details["username"] = sanitize_username(username)
```
kiblik: I don't know if it's the best idea. If `username` is missing, it might mean that _social_ is not configured well. This might:
- point to unexpected behaviour (you do not expect by default that some other fields is used as username)
- create a security hole: if changing of email or fullname is fully in the hands of the user (in IDP), the user might impersonate another user.

I'm not against this change in general. It just should not be on by default.

manuel-sommer: Hm, good point.
Maybe we should make it configurable through settings.dist.py?

quirinziessler (approved):

rossops (approved):

Issue #13564, opened 2025-10-29
n0s1 scanner (confluence) and more?


BoBeR182:
**Scanner Name**
Name of the scanner, brief description of the scanner and link.
n0s1 scanner, scans for secrets in Confluence and more, https://github.com/spark1security/n0s1 

Note: this is the scan output for a confluence scan, there are many scan type options

**Sample File**
```json
[...]

PR #13580, opened 2025-10-31
:tada: implement n0s1 scanner #13564

manuel-sommer:
#13564

### DIFF
--- a/Dockerfile.integration-tests-debian
+++ b/Dockerfile.integration-tests-debian
@@ -1,7 +1,7 @@
 
 # code: language=Dockerfile
 
-FROM openapitools/openapi-generator-cli:v7.16.0@sha256:e56372add5e038753fb91aa1bbb470724ef58382fdfc35082bf1b3e079ce353c AS openapitools
+FROM openapitools/openapi-generator-cli:v7.17.0@sha256:868b97eb4e5080d2cdfd5b3eeaa4d52e4bbb7c56f14e234b08b0b0bc4f38a78f AS openapitools
 # currently only supports x64, no arm yet due to chrome and selenium dependencies
 FROM python:3.13.7-slim-trixie@sha256:5f55cdf0c5d9dc1a415637a5ccc4a9e18663ad203673173b8cda8f8dcacef689 AS build
 WORKDIR /app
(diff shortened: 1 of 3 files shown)


mtesauro (approved):
Approved

dojo/tools/n0s1/parser.py
```diff
@@ -0,0 +1,71 @@
...
+                description += f"**Tags:** {', '.join(merged_regex['tags'])}\n"
+
+            find = Finding(
+                title=title,
+                test=test,
+                description=description,
+                severity="High",  # You can adjust this based on your logic
+                dynamic_finding=True,
+                static_finding=False,
+                unique_id_from_tool=finding_data.get("id", finding_id),
```
valentijnscholten: Looking at the source of n0s1 this id seems to be pretty unique (it's a hash of the url of the location where the secret was found + the redacted secret itself). Should we use it for deduplication?

manuel-sommer:
> This is a parser that can return different type of scans. Can you look at implementing the dynamic test_type feature similar to rusty hog?
> 
> https://github.com/DefectDojo/django-DefectDojo/blob/1ba1122954b6fa214862e23345373f75434cdaa1/dojo/tools/rusty_hog/parser.py#L33-L70

Hi @valentijnscholten , before I implement this, I would need multiple different output files. I was just doing the PR, not using it to test different scenarios (for this I don't have time right now). 
How about continuing with this and as soon as someone asks for other scenarios to cover, we can advance this PR with their provided scan results?

valentijnscholten:
It's only about adding logic to set the test_type name / parser name. Bot about having all scanners supported.

unittests/tools/test_n0s1_parser.py
```diff
@@ -0,0 +1,31 @@
...
+            parser = N0s1Parser()
+            test_type = Test_Type(name="n0s1 Scanner")
+            test = Test(test_type=test_type)
+            findings = parser.get_findings(testfile, test)
+            self.assertEqual(17, len(findings))
+            finding = findings[0]
+            self.assertEqual(finding.title, "AWS")
+            self.assertIsNotNone(finding.description)
+            self.assertTrue(finding.dynamic_finding)
+            self.assertEqual(test.test_type.name, "n0s1 Scanner")
```
valentijnscholten, line 21: I would expect something like confluence in here?

manuel-sommer: @valentijnscholten , I compared with rustyhog, the functionality is fine, works as expected. I added a unittest also for rustyhog here to underline it.

valentijnscholten: OK, I always thought there was some dynamic test_type generated But it looks like it's just the title field that expects "Confluence" to be in there. str(test) ==

Image

PR #13587, opened 2025-11-01
auditlog: switch to pghistory (for real)

valentijnscholten:
In #13169 we introducted `django-pghistory` as an alternative auditlog solution. The goal has always been to move to `django-pghistory`, which is what this PR does now that we have had some time to test it.

This PR:
- Removes `DD_AUDITLOG_TYPE`
- Removes any code related to tracking changes with `django-auditlog`
- Keeps existing records
- Keeps displaying existing records
- Removes any unit tests / workflow runs that were specific to `django-auditlog`

I tried removing the `django-auditlog` as a dependency, but this has some complications. We may have to postpone this a bit to not have everything in one release.

### DIFF
--- a/.github/workflows/test-helm-chart.yml
+++ b/.github/workflows/test-helm-chart.yml
@@ -150,7 +150,7 @@ jobs:
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
       - name: Generate values schema json
-        uses: losisin/helm-values-schema-json-action@660c441a4a507436a294fc55227e1df54aca5407 # v2.3.1
+        uses: losisin/helm-values-schema-json-action@f3517c55537e26953c8a11be7549ea908990130d # v2.3.2
         with:
           fail-on-diff: true
           working-directory: "helm/defectdojo"
--- a/.github/workflows/update-sample-data.yml
+++ b/.github/workflows/update-sample-data.yml
@@ -43,7 +43,7 @@ jobs:
           git push --set-upstream origin $(git rev-parse --abbrev-ref HEAD)
 
       - name: Create Pull Request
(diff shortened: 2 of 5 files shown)


valentijnscholten:
Converted back  to draft as we need to align the merge/release with Pro.

dojo/settings/settings.dist.py
```diff
@@ -2085,12 +2078,8 @@ def show_toolbar(request):
...
     crum_index = middleware_list.index("crum.CurrentRequestUserMiddleware")
 
-    if AUDITLOG_TYPE == "django-auditlog":
-        # Insert AuditlogMiddleware before CurrentRequestUserMiddleware
-        middleware_list.insert(crum_index, "dojo.middleware.AuditlogMiddleware")
-    elif AUDITLOG_TYPE == "django-pghistory":
-        # Insert pghistory HistoryMiddleware before CurrentRequestUserMiddleware
-        middleware_list.insert(crum_index, "dojo.middleware.PgHistoryMiddleware")
+    # Insert pghistory HistoryMiddleware before CurrentRequestUserMiddleware
+    middleware_list.insert(crum_index, "dojo.middleware.PgHistoryMiddleware")
```
Maffooch: Can this moved the `MIDDLEWARE` block?

Maffooch (approved):

dojo/settings/settings.dist.py
valentijnscholten: @Maffooch you mean like this https://github.com/DefectDojo/django-DefectDojo/pull/13587/commits/8878a0f1934bb2975f5815f2a02afc5dc01fd915

mtesauro (approved):
Approved

blakeaowens (approved):

Issue #13593, opened 2025-11-02
Calender: Incorrect End Date Displayed for Engagements


Fatte1254:
## :warning: Note on feature completeness :warning:

We are narrowing the scope of acceptable enhancements to DefectDojo. Learn more here:
https://github.com/DefectDojo/django-DefectDojo/blob/master/readme-docs/CONTRIBUTING.md

**Is your feature request related to a problem? Please describe**
Problem:
When creating a test or engagement with a date range (e.g., from 03 November to 10 November), the calendar view incorrectly displays the end date only up to 09 November instead of the selected 10 November.

**Describe the solution you'd like**
Solution:
Adjust the calendar logic to correctly include the end date in the displayed range, ensuring that engagements scheduled until 10 November are shown as lasting through the entire day on 10 November.

valentijnscholten:
Confirmed in demo instance.

manuel-sommer:
Please close this @Fatte1254

PR #13595, opened 2025-11-03
:bug: calender: Fix incorrect end dates for engagements and tests #13593

manuel-sommer:
#13593


mtesauro (approved):
Approved

Maffooch (approved):

valentijnscholten (commented):
Why do we need to add 1 day?

manuel-sommer:
> Why do we need to add 1 day?

Because of: https://github.com/DefectDojo/django-DefectDojo/issues/13593

manuel-sommer:
> I know. But why is the solution to add 1 day. Or do we do that everywhere?

To be honest, I didn't look deeper into this.

mtesauro:
@valentijnscholten 
> Why do we need to add 1 day?

In truth, I wondered that as well but (my bad) assumed it was only for the corner case, not in general.

manuel-sommer:
So, I spend a little time investigating:
start = datetime(2025, 11, 3)
end = datetime(2025, 11, 10)

--> time range: 03 Nov 2025 00:00 → 10 Nov 2025 00:00

So, start date is inclusive, end date is exclusive. That's why we have to add 1 day in this special case as we don't have a specific time slot, but a time range.

valentijnscholten (approved):
Ok thanks. I was wondering earlier if it could be something like a "off by 1 hour" timezone thing. But this makes sense now.

PR #13614, opened 2025-11-04
Adding SOCIAL_AUTH_REDIRECT_IS_HTTPS, to enable use of HTTPS protocol when redirecting after login using social auth.

marcelhorner:
**Description**

When setting up DD_SECURE_SSL_REDIRECT to False and Azure auth, after login the redirect uses HTTP protocol by default. To enable DefectDojo to redirect with HTTPS protocol after login, which is required by Azure, we need to set SOCIAL_AUTH_REDIRECT_IS_HTTPS to True on Django framework.

This change adds the DD_SOCIAL_AUTH_REDIRECT_IS_HTTPS to to defectdojo\dojo\settings\settings.dist.py, making it possible to set SOCIAL_AUTH_REDIRECT_IS_HTTPS on Django framework. 

**Test results**

I have a DefectDojo installation on a Kubernetes cluster. It was installed with helm and helm value file has DD_SECURE_SSL_REDIRECT: "False" config. After setting up auth on Azure I got an error about redirect not matching with Azure config.

After adding SOCIAL_AUTH_REDIRECT_IS_HTTPS  var on defectdojo\dojo\settings\settings.dist.py and setting it to True on helm value file Azure auth proccess worked.

### DIFF
--- a/docs/content/en/customize_dojo/user_management/configure_sso.md
+++ b/docs/content/en/customize_dojo/user_management/configure_sso.md
@@ -511,9 +511,7 @@ If during the login process you get the following error: *The
 in the client app settings.* and the `redirect_uri` HTTP
 GET parameter starts with `http://` instead of
 `https://` you need to add
-`SOCIAL_AUTH_REDIRECT_IS_HTTPS = True` to Docker environment variables, or to your `local_settings.py` file.
-You can also setup SOCIAL_AUTH_REDIRECT_IS_HTTPS by adding the DD_SOCIAL_AUTH_REDIRECT_IS_HTTPS var to extraConfigs
-section of your Helm value yaml file, when deploying with Helm.
+`DD_SOCIAL_AUTH_REDIRECT_IS_HTTPS = True` to Docker Compose environment variables, or `SOCIAL_AUTH_REDIRECT_IS_HTTPS` to your `local_settings.py` file.
 
 2. Restart DefectDojo, and 'Login With Okta' should appear on the login screen.
 


valentijnscholten (changes requested):
Thanks! Looks good. Could you add something like ", or set the `DD_SOCIAL_AUTH_REDIRECT_IS_HTTPS "
to:
https://github.com/DefectDojo/django-DefectDojo/blob/8bc37388ff13dc212f1233d526748a98af1c3221/docs/content/en/customize_dojo/user_management/configure_sso.md?plain=1#L509-L514

marcelhorner:
> Thanks! Looks good. Could you add something like ", or set the `DD_SOCIAL_AUTH_REDIRECT_IS_HTTPS " to:
> 
> https://github.com/DefectDojo/django-DefectDojo/blob/8bc37388ff13dc212f1233d526748a98af1c3221/docs/content/en/customize_dojo/user_management/configure_sso.md?plain=1#L509-L514

Done.

docs/content/en/customize_dojo/user_management/configure_sso.md
```diff
@@ -512,6 +512,8 @@ in the client app settings.* and the `redirect_uri` HTTP
 GET parameter starts with `http://` instead of
 `https://` you need to add
 `SOCIAL_AUTH_REDIRECT_IS_HTTPS = True` to Docker environment variables, or to your `local_settings.py` file.
+You can also setup SOCIAL_AUTH_REDIRECT_IS_HTTPS by adding the DD_SOCIAL_AUTH_REDIRECT_IS_HTTPS var to extraConfigs
+section of your Helm value yaml file, when deploying with Helm.
```
valentijnscholten:
```suggestion
`DD_SOCIAL_AUTH_REDIRECT_IS_HTTPS = True` to Docker Compose environment variables, or `SOCIAL_AUTH_REDIRECT_IS_HTTPS` to your `local_settings.py` file.
```

mtesauro (approved):
Approved

docs/content/en/customize_dojo/user_management/configure_sso.md
```diff
@@ -511,7 +511,7 @@ If during the login process you get the following error: *The
 in the client app settings.* and the `redirect_uri` HTTP
 GET parameter starts with `http://` instead of
 `https://` you need to add
```
manuel-sommer, line 513: the wording does not fit to the setting.
_and the `redirect_uri` HTTP
GET parameter starts with `https://` instead of
`http://` you need to add_ (because the DD_SOCIAL_AUTH_REDIRECT_IS_HTTPS=(bool, False))

And shouldn't we make this setting always to True just because DefectDojo is a security tool?

Maffooch: > And shouldn't we make this setting always to True just because DefectDojo is a security tool?

No as the default is false from the lib. We want to avoid potentially breaking SSO for some folks

manuel-sommer: sounds legit

PR #13210, opened 2025-09-17
fix: add missing resources, securityContext and env entries

fernandezcuesta:
**Description**

This is a complement of #12691 which addressed some Helm chart refactoring, but failed to allow securityContext as well as mounted volumes and resources for all containers and initContainers.

For example, when PSS policies are enforced, we may need to specify `readOnlyRootFilesystem: false` and mount tmpfs volues in all (init)containers. This PR closes the gaps left from #12691.

**Test results**

Tested in a production instance from a fork.

**Documentation**
n/a

PR #13619, opened 2025-11-04
fix(helm): Fix PVC templating after #13210

kiblik:
After #13210, charts with enabled PVC were complaining
```
executing "defectdojo/templates/media-pvc.yaml" at <.Values.extraAnnotations>: nil pointer evaluating interface {}.extraAnnotations
```

JFIY @fernandezcuesta, we missed some details. I hope this is the only missed detail.
More complex tests of HELM charts would be useful.

### DIFF
--- a/helm/defectdojo/Chart.yaml
+++ b/helm/defectdojo/Chart.yaml
@@ -37,3 +37,5 @@ annotations:
   artifacthub.io/changes: |
     - kind: fixed
       description: Broken rendering of media PVC
+    - kind: fixed
+      description: Typo in description of digests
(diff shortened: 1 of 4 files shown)


fernandezcuesta:
Oops my bad, sorry for that. I'm more convinced to always use $ regardless of whether inside a with statement or not....

kiblik:
> Oops my bad, sorry for that. I'm more convinced to always use $ regardless of whether inside a with statement or not....

If we would perform another "large rewrite", I would probably go with $ everywhere. But for now, this fix is enough in my opinion.

mtesauro (approved):
Approved

valentijnscholten (approved):

kiblik:
Based on https://github.com/DefectDojo/django-DefectDojo/pull/13620#issuecomment-3493245903, I merged changes into this PR.
There is no other HELM-related PR that would have this kind of easy fix. Others need more time for review.

blakeaowens (approved):

PR #13630, opened 2025-11-05
SLA Calculations: Remove product grade calculation and consolidate task handlers

Maffooch:
When reviewing the SLA calculation functions, I was confused by the two helper functions and consolidated them. Additionally, I noticed the product grade calculation may be wasted compute. The SLA of a finding is not considered as part of the product grade, so no reason to the recalculate

### DIFF
--- a/dojo/sla_config/helpers.py
+++ b/dojo/sla_config/helpers.py
@@ -44,13 +44,8 @@ def update_sla_expiration_dates_sla_config_sync(sla_config: SLA_Configuration, p
     mass_model_updater(Finding, findings, lambda f: f.set_sla_expiration_date(), fields=["sla_expiration_date"])
 
     # reset the async updating flag to false for all products using this sla config
-    # TODO: It would be faster to use update. Especially since we likely dont want signals to fire here
+    # use update as we don't want save() and signals to be triggered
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

dojo/sla_config/helpers.py
```diff
@@ -2,49 +2,55 @@
...
-        product.async_updating = False
-        super(Product, product).save()
-        calculate_grade(product)
+    # TODO: It would be faster to use update. Especially since we likely dont want signals to fire here
+    products.update(async_updating=False)
+    # for product in products:
+    #     product.async_updating = False
+    #     super(Product, product).save()
+    #     # TODO: I cannot see a reason why we need to calculate the grade here again
+    #     calculate_grade(product)
```
valentijnscholten:
```suggestion
    # use update as we don't want save() and signals to be triggered
    products.update(async_updating=False)
```

valentijnscholten: BTW I think in the old code the call to `super(Product, product).save()` also bypasses the `save()` method and signals. But using `update()` looks cleaner.

valentijnscholten (approved):

PR #13650, opened 2025-11-07
Improve tag handling in importers and add tests for tag imports

Maffooch:
Enhance the tag processing logic in the DefaultImporter to correctly handle both lists and strings of tags. Introduce unit tests to verify the import functionality for reports containing tags.


dojo/importers/default_importer.py
```diff
@@ -226,7 +226,11 @@ def process_findings(
             # Process any endpoints on the endpoint, or added on the form
             self.process_endpoints(finding, self.endpoints_to_add)
             # Parsers must use unsaved_tags to store tags, so we can clean them
-            finding.tags = clean_tags(finding.unsaved_tags)
+            cleaned_tags = clean_tags(finding.unsaved_tags)
+            if isinstance(cleaned_tags, list):
+                finding.tags.set(cleaned_tags)
+            elif isinstance(cleaned_tags, str):
+                finding.tags.set([cleaned_tags])
```
valentijnscholten, line 233: reimport has similar code that needs to be changed:
https://github.com/valentijnscholten/django-DefectDojo/blob/1ba1122954b6fa214862e23345373f75434cdaa1/dojo/importers/default_reimporter.py#L691-L693

unittests/test_tags.py
```diff
@@ -285,6 +285,22 @@ def test_import_multipart_tags(self):
...
+                for tag in desired_tags:
+                    self.assertIn(tag, finding["tags"])
+
+        # Import a report with findings that have tags
+        import0 = self.import_scan_with_params(self.generic_sample_with_tags_filename, scan_type="Generic Findings Import")
+        test_id = import0["test"]
+        response = self.get_test_findings_api(test_id)
+        findings = response["results"]
+        # Make sure we have what we are looking for
+        assert_tags_in_findings(findings, 2, ["security", "network"])
```
valentijnscholten, line 303: should there also be a test for the non list but str case?

Maffooch: I not sure how to test the string case tbh the import/reimport flow only passes a list

valentijnscholten: What was the reason to add the str case?

Maffooch: The function signature says the possible return types are a list and a string, so I thought it would be best to accommodate both cases 
https://github.com/DefectDojo/django-DefectDojo/blob/d4e751346fefd4042765a35daf23f8553e278e1c/dojo/validators.py#L35

valentijnscholten: ah, yes.

valentijnscholten (approved):
Approved. What is the reason this PR fixes the/a problem? Is it that `finding.tags = clean_tags(finding.unsaved_tags)` doesn't work but `finding.tags.set( clean_tags(finding.unsaved_tags))` does?

mtesauro (approved):
Approved

Maffooch:
> What is the reason this PR fixes the/a problem? Is it that finding.tags = clean_tags(finding.unsaved_tags) doesn't work but finding.tags.set( clean_tags(finding.unsaved_tags)) does?

Yes. In previous versions, the tags were set before the finding was saved for the first time, so the tags would be applied normally. In today's version, the tags are set after the finding is saved, but not in the same way the other m2m relationships are managed. Since we only save the finding once in the importer now, we can't just set the tags attribute any longer. We have to use add/set in the m2m manager

rossops (approved):

Jino-T (approved):

PR #13677, opened 2025-11-11
fix(helm): merge extraAnnotations with init job annotations

qlimenoque:
## Description

Merges global `extraAnnotations` with initializer job's `podAnnotations` to ensure consistent annotation handling across all DefectDojo resources.

### Changes
- Renamed `initializer.annotations` to `initializer.podAnnotations` for clarity
- Updated initializer job template to merge `extraAnnotations` with `podAnnotations`
- Both annotation sets are now properly applied to the pod template
- Bumped chart version from 1.8.1 to 1.9.0

### ⚠️ Breaking Change

This change is **not backward compatible**:
- `initializer.annotations` has been renamed to `initializer.podAnnotations`
- `initializer.jobAnnotations` affects the Job spec metadata
- `initializer.podAnnotations` affects the Pod template metadata within the Job

Users must update their `values.yaml` to use `initializer.podAnnotations` instead of `initializer.annotations`.

### Migration

If you were using:
```yaml
initializer:
  annotations:
    foo: bar
```

Update to:
```yaml
initializer:
  podAnnotations:
    foo: bar
```

### DIFF
--- a/helm/defectdojo/values.schema.json
+++ b/helm/defectdojo/values.schema.json
@@ -914,9 +914,6 @@
                 "affinity": {
                     "type": "object"
                 },
-                "annotations": {
-                    "type": "object"
-                },
                 "automountServiceAccountToken": {
                     "type": "boolean"
                 },
@@ -965,6 +962,9 @@
                     "type": "object"
                 },
                 "nodeSelector": {
+                    "type": "object"
+                },
+                "podAnnotations": {
                     "type": "object"
                 },
                 "podSecurityContext": {


kiblik:
Can you fix failed GHA, please? Update of schema, docs and release notes are needed.

valentijnscholten:
@qlimenoque And can you base the PR against the `dev` branch.

kiblik:
> Can you fix failed GHA, please? Update of schema, docs and release notes are needed.

Plus https://github.com/DefectDojo/django-DefectDojo/blob/dev/docs/content/en/open_source/upgrading/2.53.md

kiblik:
@qlimenoque, keys in `helm/defectdojo/values.schema.json` are alphabetically ordered, so a simple rename is not enough. Can you move it to the right place? 
Or feel free to use the method described in related GHA:
https://github.com/DefectDojo/django-DefectDojo/blob/b9836f2fffb0588e45b3b6fbffa06bb643543b04/.github/workflows/test-helm-chart.yml#L143-L146

qlimenoque:
@kiblik thanks for pointing out the instruction!
I actually found the Helm schema plugin myself, but I used `helm schema` without the `--use-helm-docs` flag, which kept deleting the description field from the schema. I assumed the project used a different tool and ended up just renaming it 😀

It might be helpful to include this instruction in the **README.md** inside the Helm chart directory
It’s much harder to spot when it’s only mentioned in a CI step

PR #13679, opened 2025-11-11
log a line when custom hash method is used

valentijnscholten:
add some logging to make it explicit and clear when debugging hash code issues.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -3485,15 +3485,16 @@ class Finding(models.Model):
     def set_hash_code(self, dedupe_option):
         from dojo.utils import get_custom_method  # noqa: PLC0415 circular import
         if hash_method := get_custom_method("FINDING_HASH_METHOD"):
+            deduplicationLogger.debug("Using custom hash method")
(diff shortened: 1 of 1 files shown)


dojo/models.py
```diff
@@ -3494,15 +3485,16 @@ def violates_sla(self):
     def set_hash_code(self, dedupe_option):
         from dojo.utils import get_custom_method  # noqa: PLC0415 circular import
         if hash_method := get_custom_method("FINDING_HASH_METHOD"):
+            deduplicationLogger.debug("Using custom hash method")
```
dogboat, line 3488: Since this is for debug, is there any value in including the method name (either from the given setting or by stringifying the resolved method)?

Jino-T (approved):

dojo/models.py
mtesauro: @valentijnscholten Before I approve and merged this I wanted to give you a chance to comment here.

blakeaowens (approved):

Issue #13681, opened 2025-11-11
Github SAST Scan duplicate findings due to hashcode issue


Logicmn:
**Bug description**
Github SAST Scan uses `line` as field its hashcode. These alerts in GitHub change lines often, causing duplicate findings to be created for the same alert when re-uploading a scan.

```
"Github SAST Scan": ["vuln_id_from_tool", "severity", "file_path", "line"]
```
[Link to code](https://github.com/DefectDojo/django-DefectDojo/blob/b9836f2fffb0588e45b3b6fbffa06bb643543b04/dojo/settings/settings.dist.py#L1355).

**Steps to reproduce**
1. Import a test with findings using Github SAST Scan scan type
2. Change the line number on one of the findings
3. Re-upload the test/scan
4. Observe duplicate finding

**Expected behavior**
The line field should not factor into the hash code, and the alert should only be uploaded once.

Logicmn:
I was trying to brainstorm a better determination of uniqueness for these findings, but I have been struggling because there is no unique alert ID provided from GitHub. If we simply remove `line` from the hashcode, I am not sure it will be unique enough. For example, in the case of two alerts in the same file, of the same severity, and the same vuln_id (e.g. `codeql/js/ssrf`), these would be merged into one finding in DefectDojo, if `line` was removed. But perhaps that would be OK? I'm curious what others think...

manuel-sommer:
The same issue occurs within all kinds of sast scanners, e.g. bandit or rusty hog.

Logicmn:
Makes sense, it's a tough problem to tackle. I can't speak for other scanners, but I have proposed a solution for Github SAST in the following draft PR: https://github.com/DefectDojo/django-DefectDojo/pull/13683

If we can solve this duplication issue, it would remove the headache of manually cleaning up dupes, and increase integrity of Github SAST Scan findings in DefectDojo.

manuel-sommer:
What you can do is: closing the old findings (there is a setting in the api call when you import / reimport the file).

Logicmn:
We already set `close_old_findings` to true, this does not solve the problem. When we create a Jira, it is linked to the original finding. When re-importing, the original finding with the linked Jira will be closed, and a new finding will be created. This causes confusion, and manually rectifying this can take time and effort.

devGregA:
[...]

PR #13683, opened 2025-11-11
Fix Duplicate Findings in Github SAST Scan by Improving Hash Code

Logicmn:
**Description**

This is an update to the fields used to generate the hash code for Github SAST Scan findings. As explained in [this issue](https://github.com/DefectDojo/django-DefectDojo/issues/13681), currently the `line` field is used in the hash code. Since these alerts in change lines in GitHub often, it creates duplicate findings for the same alert when re-uploading a scan.

To fix this, I propose replacing the `line` field with the `url` field in the hash code. The format of the `url` field is:
```https://github.com/OWASP/test-repository/security/code-scanning/35```

This includes the organization, the repository, as well as the alert id. Because the alert id is only unique at the repository level, including only the alert id would not be enough; The organization and repository name is necessary as well. The `url` field conveniently contains all three. The `url` field is already populated by the parser, which pulls it from the `html_url` field in the GitHub alert. It can be expected that this field will always exist.


valentijnscholten:
The `url` field currently is not allowed as hash_code field. It could be interesting to try out if it works well. Can you add it to `HASHCODE_ALLOWED_FIELDS` (same file) and do some testing? You may have to recalculate your hash_codes using the `dedupe` management command.

@Maffooch Another example of where it might be helpful if we would allow parser to calculate a hash based on fields in the report that are not stored in defect dojo.

PR #13689, opened 2025-11-12
docs: Update the API pull documentation

sNiXx:
**Description**

Updating a few minor inconsistencies and errors, but also adding more details about custom trust configuration.

**Test results**

N/A

**Documentation**

This is a docs-only PR

### DIFF
--- a/docs/content/supported_tools/parsers/api/sonarqube.md
+++ b/docs/content/supported_tools/parsers/api/sonarqube.md
@@ -2,20 +2,24 @@
 title: "SonarQube API Import"
 toc_hide: true
 ---
-All parsers which using API have common basic configuration step but with different values. Please, [read these steps](../) at first.
+All parsers that use API pull have common basic configuration steps, but with different values. Please, [read these steps](../) first.
 
-In `Tool Configuration`, select `Tool Type` to "SonarQube" and `Authentication Type` "API Key".
-Note the url must be in the format of `https://<sonarqube_host>/api`
+## Tool Configuration
+
+In `Tool Configuration`, select `Tool Type` "SonarQube" and `Authentication Type` "API Key".
+The URL must be in the format of `https://<sonarqube_host>/api`
 Paste your SonarQube API token in the "API Key" field.
-By default the tool will import vulnerabilities issues
-and security hotspots only, but additional filters can be setup using the 
-Extras field separated by commas (e.g. `BUG,VULNERABILITY,CODE_SMELL`). When using
-SonarCloud, you must also specify the Organization ID in the Extras field as follows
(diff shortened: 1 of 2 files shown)


docs/content/supported_tools/parsers/api/sonarqube.md
```diff
@@ -24,23 +28,29 @@ In "Add API Scan Configuration"
...
 of the SonarQube project will be used.
 
-**Note:**: If `https` is used for the SonarQube, the certificate must be
-trusted by the DefectDojo instance.
+## Custom Trust
+
+If you are connecting to SonarQube via HTTPS, the issuer of the certificate that is presented by
+SonarQube must be trusted.
+
+One way of achieving this is by defining the `REQUESTS_CA_BUNDLE` environment variable to point
```
paulOsinski, line 54: > `REQUESTS_CA_BUNDLE` environment variable

this is an env var within SonarQube?  Just want to specify

paulOsinski (dismissed):

docs/content/supported_tools/parsers/api/sonarqube.md
sNiXx: No, this appears to be an env variable of the python `requests` library  (see [here](https://requests.readthedocs.io/en/latest/user/advanced/)). Although its usage in DefectDojo is not really documented besides in one sentence [here](https://github.com/DefectDojo/django-DefectDojo/blob/b9836f2fffb0588e45b3b6fbffa06bb643543b04/docs/content/en/customize_dojo/user_management/configure_sso.md?plain=1#L655) and in the helm instructions (e.g. [here](https://github.com/DefectDojo/django-DefectDojo/blob/b9836f2fffb0588e45b3b6fbffa06bb643543b04/helm/defectdojo/values.yaml#L449))

As far as I understand, `REQUESTS_CA_BUNDLE` should ideally point to a file (or a directory) that includes custom trusted CA certificate(s) as well as the other trusted CA certificates already present (i.e. it overrides the default trust store).

I would propose to reflect this in the documentation by referencing the official `requests` docs.

Another alternative could be to bind /etc/ssl/certs to a read-only mounted volume,  add the custom trusted CA certificates with Ubuntu's `update-ca-certificates` command and then set `REQUESTS_CA_BUNDLE` to /etc/ssl/certs/ca-certificates.crt

Writing these lines, I wonder, if it would make sense to have a dedicated page for this?

Jino-T (approved):

docs/content/supported_tools/parsers/api/sonarqube.md
valentijnscholten: @sNiXx That would be very helpful (a page about REQUESTS_CA_BUNDLE). But can be a follow up PR as it's not blocking this.

PR #13694, opened 2025-11-12
feat(renovate): Wait 2 days to use latest k8s

kiblik:
When a new k8s is released, tests with Minikube are failing because it is not fully ready to use the latest version. If Renovate waits 2 days, nothing critical will happen, and the review process will be much smoother (no need to troubleshoot why tests are failing).


mtesauro (approved):
Approved

PR #13613, opened 2025-11-04
fix(helm/dependabot/renovate): Fix broken automatic update

kiblik:
Finish #13520
Fix #13530 & #13612
GHA is failing because action wasn't initialized correctly (https://github.com/DefectDojo/django-DefectDojo/actions/runs/19070470968/job/54486644356#step:4:21) - it needs to be checked out on the branch.

PR #13695, opened 2025-11-12
fix(helm/renovate/dependabot): Commit changes & fix condition format

kiblik:
Extend #13613 and #13520
Fix #13668 and #13666

This part is a bit hard for testing and writing it right on the first try. I hope this will finally work correctly.

### DIFF
--- a/.github/workflows/test-helm-chart.yml
+++ b/.github/workflows/test-helm-chart.yml
@@ -115,6 +115,8 @@ jobs:
         if: startsWith(github.head_ref, 'renovate/') || startsWith(github.head_ref, 'dependabot/')
         run: |
           yq -i '.annotations."artifacthub.io/changes" += "- kind: changed\n  description: ${{ github.event.pull_request.title }}\n"' helm/defectdojo/Chart.yaml
+          git add helm/defectdojo/Chart.yaml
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

valentijnscholten (approved):

kiblik:
@mtesauro & @Maffooch, I just found out that even after this PR, we still might be facing the issue of not triggering GHA, because if a new offered change is breaking sth, checks will pass (because unittests would not be triggered).

https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#triggering-a-workflow-from-a-workflow
> When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's GITHUB_TOKEN, a new workflow will not run even when the repository contains a workflow configured to run when push events occur. For more information, see [Use GITHUB_TOKEN for authentication in workflows](https://docs.github.com/en/actions/security-guides/automatic-token-authentication).

> If you do want to trigger a workflow from within a workflow run, you can use a GitHub App installation access token or a personal access token instead of GITHUB_TOKEN to trigger events that require a token.

PR #13732, opened 2025-11-18
Add notification when finding is created via the API

Juu:
This PR adds a notification trigger when a finding is created via the API, like this was already implemented for a finding creation in WebUI.

This was explicitly noted in the UI code with a comment:
`TODO: but same should be implemented for API endpoint`

### DIFF
--- a/unittests/test_notifications.py
+++ b/unittests/test_notifications.py
@@ -456,7 +456,7 @@ class TestNotificationTriggersApi(APITestCase):
             f'Finding "Finding With Auto-Assigned Reporter Notification" was added by {self.admin}',
         )
         self.assertEqual(call_args[1]["icon"], "exclamation-triangle")
-        
+
         # Verify the finding was created successfully
         created_id = response.data.get("id")
         self.assertIsNotNone(created_id)


Maffooch (approved):

dojo/api_v2/serializers.py
```diff
@@ -1949,6 +1950,16 @@ def create(self, validated_data):
         if push_to_jira:
             jira_helper.push_to_jira(new_finding)
 
+        # Create a notification
+        create_notification(
+            event="finding_added",
+            title=_("Addition of %s") % new_finding.title,
+            finding=new_finding,
+            description=_('Finding "%s" was added by %s') % (new_finding.title, new_finding.reporter),
```
kiblik, line 1958: Are we sure that `new_finding.reporter` is set every time?

kiblik (commented):
Can you write some tests as well, please?

dojo/api_v2/serializers.py
Juu: I believe it is, as if reporter is not provided in the API request, it falls back in `FindingCreateSerializer.validate` to the authenticated user making the API request.

https://github.com/DefectDojo/django-DefectDojo/blob/42acb6172bbc026b5501537b33f030849dba51b9/dojo/api_v2/serializers.py#L1981-L1983

Juu: I added some tests (using copilot, tbh)

valentijnscholten (changes requested):
Thank you for the PR. Can you ask copilot to add the tests as new methods in the `TestNotificationTriggersApi` class in `test_notifications.py`?

PR #13738, opened 2025-11-19
Add python-logstash for ELK logging configuration

guipal:
**Description**

Currently, logs are being sent tothe console using StreamLogger. We would like to send the logs to our ELK stack so we can better investigate on them. To configure it for Django apps there is this [answer from elk support](https://discuss.elastic.co/t/configure-elasticsearch-on-django-app/326320). Nevertheless, after following the proposal we got the following exception starting up the application
`ModuleNotFoundError: No module named ‘logstash’`

**Resolution**

Add python-logstash library to the requirements when we build the image.


valentijnscholten:
Can you not just send the docker logs to logstash instead?

Maffooch:
Closing for now. Please see https://github.com/DefectDojo/django-DefectDojo/pull/13738#pullrequestreview-3490134075

PR #13754, opened 2025-11-21
fix(init): Initialzer has to fail if migration is not done

kiblik:
Since #13169, we are avoiding failure of `docker/entrypoint-initializer.sh`, which is understandable in the related context; however, it is not the best behavior. If migration is missing, we should `exit` with some error code; otherwise, it is easy to miss bugs in implementation.


Maffooch (approved):

valentijnscholten (commented):
I think it has always been a warning. Then I made it stricter at some point. Then during this PR I think I realized users are locked out if they make changes but can no longer start Defect Dojo anymore to run `makemigrations`. Now would probably be a good time to add to the DOCKER.md how users (developers) can make new migrations. I think we still need to do something like starting uwsgi and postgress with --no-deps.

kiblik:
> I think it has always been a warning. Then I made it stricter at some point. Then during this PR I think I realized users are locked out if they make changes but can no longer start Defect Dojo anymore to run `makemigrations`. Now would probably be a good time to add to the DOCKER.md how users (developers) can make new migrations. I think we still need to do something like starting uwsgi and postgress with --no-deps.

I'm not sure about starting uwsgi. It easily create confusion, why service is started but failing with some unrelated error. User will see that UI is not able to access not existing field, not that he did not perform migration.
Plus, warning about missing migration will get lost somewhere in log history. Process should fail as soon as possible.

What about extending `docker/entrypoint-initializer.sh`: if you set `DD_DEBUG`, I will generate migration for you.

valentijnscholten:
We need to provide instructions to users/developers on how to create a migration. It's not clear currently.
Generating automatically has proven to be problematic in the past because users just ignore it and then come to Slack if there upgrade fails. It also possibly creates multiple migrations that users will have to merge when they are working on a PR.

rossops (approved):

Maffooch:
> Then during this PR I think I realized users are locked out if they make changes but can no longer start Defect Dojo anymore to run makemigrations.

I have been bit by this before, and my solution was to back out my model changes, start the app, reapply changes, and then make the new migrations. I'm not quite sure how to capture that in documentation

valentijnscholten:
What I do is
```
docker compose down
docker compose up uwsgi postgres --no-deps
docker compose exec uwsgi bash -c "python manage.py makemigrations"
docker compose down
docker compose up
```

Maffooch:
^ that is better advice! It should be included in the docs somewhere before we can move forward on this one

Maffooch:
Converting to draft for now

PR #13756, opened 2025-11-22
fix: enable uwsgi DD_UWSGI_EXTRA_ARGS passthrough

Bump-Action:
### **Description**
Prevent resetting `DD_UWSGI_EXTRA_ARGS` variable in uwsgi entry point

### **What problem does it solve?**
Allows you to passthrough your startup parameters to the uwsgi binary

### DIFF
--- a/docker/entrypoint-uwsgi.sh
+++ b/docker/entrypoint-uwsgi.sh
@@ -27,9 +27,9 @@ python3 manage.py check
 
 DD_UWSGI_LOGFORMAT_DEFAULT='[pid: %(pid)|app: -|req: -/-] %(addr) (%(dd_user)) {%(vars) vars in %(pktsize) bytes} [%(ctime)] %(method) %(uri) => generated %(rsize) bytes in %(msecs) msecs (%(proto) %(status)) %(headers) headers in %(hsize) bytes (%(switches) switches on core %(core))'
 
-EXTRA_ARGS="${EXTRA_ARGS:-}"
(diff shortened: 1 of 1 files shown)


valentijnscholten (approved):

docker/entrypoint-uwsgi.sh
```diff
@@ -27,7 +27,7 @@ python3 manage.py check
 
 DD_UWSGI_LOGFORMAT_DEFAULT='[pid: %(pid)|app: -|req: -/-] %(addr) (%(dd_user)) {%(vars) vars in %(pktsize) bytes} [%(ctime)] %(method) %(uri) => generated %(rsize) bytes in %(msecs) msecs (%(proto) %(status)) %(headers) headers in %(hsize) bytes (%(switches) switches on core %(core))'
 
-EXTRA_ARGS=""
+EXTRA_ARGS="${EXTRA_ARGS:-}"
```
kiblik: Isn't it better to call it `DD_UWSGI_EXTRA_ARGS`? As soon as `EXTRA_ARGS` appears in some configuration, it might be confusing: Extra args of what?

valentijnscholten: agreed. I thought EXTRA_ARGS was some well-known way to startup uwsgi, but it's DD specific.

mtesauro: +1

Bump-Action:
> @Bump-Action Thank you for the PR. Could you change it to be `DD_UWSGI_EXTRA_ARGS`?

Done

Maffooch (approved):

PR #13760, opened 2025-11-24
:tada: Add 'fix_available' field to zora parser

manuel-sommer:


dojo/tools/zora/parser.py
```diff
@@ -43,8 +43,6 @@ def get_findings(self, content, test: Test) -> list[Finding]:
             description += f"**Image**: {row.get('image')}\n"
             description += f"**ID**: {row.get('id')}\n"
             description += f"**Details**: {row.get('description')}\n"
-            if row.get("fixVersion"):
-                description += f"**Fix Version**: {row.get('fixVersion')}\n"
```
mtesauro, line 47: Does it hurt to have the text of "Fix Version" in the description if fix_available is also set?

You'd think that the first question asked after seeing "fix_available = true" would be "What version is fixed".

manuel-sommer: omg, I forgot to add this info to fix_version

Issue #13761, opened 2025-11-24
Proposed Feature: Test Type Selection Toggle


manuel-sommer:
To enhance the end-user experience, introduce a system setting that allows administrators to enable or disable specific test types.

Default Behavior: All test types are enabled by default.
When Toggle is Active: Administrators can select which test types are available. Only the selected test types can be imported and used for filtering.

This feature would provide flexibility and ensure that users only interact with relevant test types, improving efficiency and reducing complexity.

What is your opinion on that?

manuel-sommer:
What is your opinion on this @valentijnscholten?

kiblik:
I remember there was some setting about excluding parsers
https://github.com/DefectDojo/django-DefectDojo/blob/69536d96a244606e0217f277fb7e3e54997b909a/dojo/tools/factory.py#L40
But I'm not sure about the whole story.

manuel-sommer:
https://github.com/DefectDojo/django-DefectDojo/pull/5084

Seems to work, but not completely. E.g. when a parser is disabled, then, you can still find the parser in the filter options.

valentijnscholten:
#7386 seems related.

manuel-sommer:
I am closing this as this is a duplicate

Issue #7386, opened 2023-01-09
Disable from Findings Import unused scan types


giveen:
Is there a way for me to disable unused scan types in important scan results?

![image](https://user-images.githubusercontent.com/1180939/211397435-461fdb15-6cab-447f-8f73-29e37048296d.png)

I only use Nessus and get tired of having to search for it each time.

dsever:
Not at this moment out of the box

giveen:
> 

Anyway I can mess with code in the Django file or will that just screw things up?

valentijnscholten:
There is a setting to disable parser by regex, but it looks like they still show up in the UI. An error is raised when an import is attempted for a disabled parser. I will discuss what the reasoning behind this is as it might be better to just hide the disabled parsers.

https://github.com/DefectDojo/django-DefectDojo/blob/452739255adc514cedabb64d2db73803ef1be934/dojo/settings/settings.dist.py#L250

mtesauro:
FWIW, I think it is preferable to ensure that disabled parsers do NOT show up in the UI - that would seem to the the expectation when they are disabled.

As long as I can see the disabled ones in the test types listing and re-enabled them if the situation changes, I think we're good.

valentijnscholten:
So where do you want do see them and where would they need to be hidden? The problem with hiding them is that users might think Defect Dojo doesn't support that type of scan report. Maybe we show them everywhere, but put the active ones on top and then the inactive ones at the bottom with "(intactive)" as suffix. Both list sorted alphabetically.

mtesauro:
Ahh, I missed a critical point above.

I was assuming a user **chose** to mark a parser as inactive. I've seen users ask do this when there's multiple parsers that are named similarly but the company only uses one of them e.g. the multiple Qualys ones.

The idea would be to keep the other users from getting confused between multiple Qualys ones, they mark the unused ones as "inactive' and those no longer show up in the list. However, the "Test Types" page linked from the left-nav always shows all of them active & inactive.

For new/default installs, we should always show all the parsers. I was just thinking giving people the ability to 'hide' some of them may make sense in some use cases so simplify the UI.

manuel-sommer:
Can we close this?

PR #13767, opened 2025-11-24
:bug: Remove unselected parsers from filters and test types

manuel-sommer:
https://github.com/DefectDojo/django-DefectDojo/issues/13761
https://github.com/DefectDojo/django-DefectDojo/issues/7386

grafik

### DIFF
--- a/.github/workflows/build-docker-images-for-testing.yml
+++ b/.github/workflows/build-docker-images-for-testing.yml
@@ -40,7 +40,7 @@ jobs:
           echo $GITHUB_ENV
 
       - name: Checkout
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           persist-credentials: false
 
--- a/.github/workflows/fetch-oas.yml
+++ b/.github/workflows/fetch-oas.yml
@@ -22,7 +22,7 @@ jobs:
         file-type: [yaml, json]
     steps:
       - name: Checkout
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           ref: release/${{ env.release_version }}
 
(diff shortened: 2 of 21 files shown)


valentijnscholten:
Do you have screenshots of other places where the list is shown?

Import form:

image

API spec:

image

manuel-sommer:
> I wonder if we have conflicting/overlapping settings here. If you go to https://demo.defectdojo.org/test_type you can already mark test types (scanners) as active or inactive. I haven't looked at the code yet, but do we need two different ways to activate/deactivate scanners? The solution via an env variable is less flexible as it requires sysadmin involvement and restarts to change. wdyt @Maffooch

You are right @valentijnscholten, looks like a duplicate. Then, the better option would be to remove the env way.

valentijnscholten:
Maybe the intention was to have some sort of "company wide exclusion" via the env variable and then allow security teams to enable/disable on top of that? I don't know.

manuel-sommer:
> Maybe the intention was to have some sort of "company wide exclusion" via the env variable and then allow security teams to enable/disable on top of that? I don't know.

If we keep both, we should add documentation.

Maffooch:
I think the active flag on individual test types is sufficient. There are also some helpers to exclude inactive scanners for import/reimport forms: https://github.com/DefectDojo/django-DefectDojo/blob/a85bbba64d4e882d3a2cbf5858f9427c27ba70df/dojo/tools/factory.py#L50-L68

manuel-sommer:
FYI: I will remove the functionality (PARSER_EXCLUDE in settings) within a PR against dev as this is a breaking change.
In this PR I will move forward to fix the inactive part to remove them from the filters etc.

valentijnscholten:
Might be better to do all of it in one PR in dev. If we split it up people might get confused if the parser exclude setting is still there. Wdyt?

Maffooch:
> Might be better to do all of it in one PR in dev. If we split it up people might get confused if the parser exclude setting is still there. Wdyt?

I agree with this thinking

manuel-sommer:
Please review @valentijnscholten 
I rebased the PR onto dev, removed DD_PARSER_EXCLUDE, adjusted the filter to exclude the inactive test types and exclude them also within the import.

Maffooch (approved):

mtesauro (approved):
Approved

manuel-sommer:
It would be awesome if we can merge this the upcoming days as I am reachable limited soonish due to vacation.

PR #13785, opened 2025-11-29
Fixes #13555 AND logic not present when filtering Findings via two or more tags

### LINKED ISSUE #13555: UI: AND logic not present when filtering Findings via two or more tags
We add various tags to our many findings on DefectDojo, two important tags are: `pci:true` and `pentest:true`.

Therefore, if we want to find all pentest related findings in the scope of PCI-DSS, we would add these tags in the tags filter:

Image

However, as a result, we receive many more findings than expected. Apparently, DefectDojo is listing all findings with either of these tags (OR), instead of only findings exclusively containing the two tags (AND).

To reproduce this, simply try to filter by two or more tags and you see that the finding list increases rather than decreasing when new tags are added.

It seems logic to me that the more tags we add to a search, more specific the results should be. Am I missing something?

PoojasPatel013:
Closes #13555

I have applied logic to filters.py, where OR logic is present. Do a review and address any required revisions. Glad if I am even a little helpful in building this amazing app!!

I have kept the existing code as is. And added AND logic with label Tags (AND)
**Test results**

Have performed manual verification using the UI with a local development instance:

Screenshot 2025-11-29 163131 

**Setup:** Created findings with various tag combinations (e.g., `pci`, `pentest`, and `cloud`).
**Action:** Tested the new `Tags (AND)` filter with `pci` and `pentest`.. 
**Result:** The new filter works as expected without disrupting the existing functionality.

**Documentation**
N/A (UI enhancement).

### DIFF
--- a/helm/defectdojo/README.md
+++ b/helm/defectdojo/README.md
@@ -512,7 +512,7 @@ The HELM schema will be generated for you.
 
 # General information about chart values
 
-![Version: 1.8.4-dev](https://img.shields.io/badge/Version-1.8.2--dev-informational?style=flat-square) ![AppVersion: 2.53.0-dev](https://img.shields.io/badge/AppVersion-2.53.0--dev-informational?style=flat-square)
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Thanks, please base the PR against `bugfix` or `dev` as per the guidelines.

PR #13800, opened 2025-12-01
[Docs] Changes Group.Read.All to GroupMember.Read.All in docs

dcfSec:
**Description**

This PR changes the recommended Entra Id API permission from Group.Read.All to GroupMember.Read.All. The reason for this is that with Group.* permission the application can also access MS Teams APIs (see [here](https://graphpermissions.merill.net/permission/Group.Read.All). For least privileges it is better to use the GroupMember.Read.All permission. This permission still allows to read basic group properties and group member (see [here](https://graphpermissions.merill.net/permission/GroupMember.Read.All)) but prevents the application from reading e.g., Teams chats or shared files.

- Group.Read.All: "Allows the app to list groups, and to read their properties and all group memberships on behalf of the signed-in user. Also allows the app to read calendar, conversations, files, and other group content for all groups the signed-in user can access."
- GroupMember.Read.All: "Allows the app to list groups, read basic group properties and read membership of all groups the signed-in user has access to."

### DIFF
--- a/docs/content/en/customize_dojo/user_management/configure_sso.md
+++ b/docs/content/en/customize_dojo/user_management/configure_sso.md
@@ -133,7 +133,7 @@ The Azure AD token need to be configured to include Group IDs. Without this step
 To update the format of the token, add a [Group Claim](https://learn.microsoft.com/en-us/entra/identity/hybrid/connect/how-to-connect-fed-group-claims) that applies to whatever Group type you are using.
 If unsure of what type that is, select `All Groups`. Do not activate `Emit groups as role claims` within the Azure AD "Token configuration" page.
 
-Application API permissions need to be updated with the `Group.Read.All` permission so that groups can be read on behalf of the user that has successfully signed in.
+Application API permissions need to be updated with the `GroupMember.Read.All` permission so that groups can be read on behalf of the user that has successfully signed in.
 
 ##### Group Cleaning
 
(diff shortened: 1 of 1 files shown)


Maffooch (changes requested):
Rather than asserting the usage of `GroupMember.Read.All` over `Group.Read.All` I think it would be preferable to add a note explaining why `GroupMember.Read.All` can be used in place of  `Group.Read.All`

dcfSec:
> Rather than asserting the usage of `GroupMember.Read.All` over `Group.Read.All` I think it would be preferable to add a note explaining why `GroupMember.Read.All` can be used in place of `Group.Read.All`

I think it is important that people _should_ use the lower privileged permissions and not only _can_ use them. Also Microsoft themself recommends the GroupMember permission as least-privileged permissions (See [here](https://learn.microsoft.com/en-us/graph/api/group-get?view=graph-rest-1.0&tabs=http#permissions) for the used endpoint.

I'll think how to better phrase that, if you wish to leave the Group permission in the documentation.

mtesauro:
@dcfSec  While I also agree that people _should_ use the least privileged option (and we should highlight that as the preferred option), I also believe the docs should reflect all the options, noting which are better than others rather than excluding less optimal options.

I've been doing this security stuff long enough to know sometimes circumstances forces the best of the bad options vs the truly best option from time to time.

I'd rather have our users informed and able to make the choice that best fits their situation which they'll know best.

PR #13801, opened 2025-12-01
Release 2.53.0: Merge Bugfix into Dev

rossops:


rossops:
@kiblik Can you take a look at the failing lint helm chart test for this PR? https://github.com/DefectDojo/django-DefectDojo/actions/runs/19829442254/job/56811361081?pr=13801

If im interpreting it correctly, it looks like it fails if there are no differences between the two merging branches for the chart, which would be a valid condition.

PR #7456, opened 2023-01-19
Avoid "STATICFILES_DIRS" warning

kiblik:
Avoid this warning:
```python
django-defectdojo-initializer-1   | System check identified some issues:
django-defectdojo-initializer-1   |
django-defectdojo-initializer-1   | WARNINGS:
django-defectdojo-initializer-1   | ?: (staticfiles.W004) The directory '/app/components/node_modules' in the STATICFILES_DIRS setting does not exist.
```

PR #13829, opened 2025-12-05
fix(node_modules): Avoid staticfiles.W004

kiblik:
This was already fixed in #7456. I'm not sure why it was reverted.


Maffooch (approved):

mtesauro (approved):
Approved

PR #13831, opened 2025-12-05
Prowler Scan Parser

Jino-T:
[sc-10823]

This is a parser for Prowler scan. It supports json and csv for the 4 cloud types:
- AWS
- GCP
- Azure
- Kubernetes

### DIFF
--- a/dojo/tools/prowler/parser.py
+++ b/dojo/tools/prowler/parser.py
@@ -16,8 +16,9 @@ class ProwlerParser:
         return "Prowler report file can be imported in JSON format or in CSV format."
 
     def get_findings(self, filename, test):
-        if str(filename).lower().endswith(".csv"):
+        name = getattr(filename, "name", str(filename)).lower()
+        if name.endswith(".csv"):
             return ProwlerParserCSV().get_findings(filename, test)
-        if str(filename).lower().endswith(".json"):
+        if name.endswith(".json"):
             return ProwlerParserJSON().get_findings(filename, test)
-        return None
+        return []
--- a/dojo/tools/prowler/parser_json.py
+++ b/dojo/tools/prowler/parser_json.py
@@ -120,12 +120,12 @@ class ProwlerParserJSON:
 
     def add_cloud_type_metadata(self, node: dict, cloudtype: str, description: str) -> str:
         # Add metadata for GCP, AWS, and Azure
-        if cloudtype in {"GCP", "AWS", "Azure"}:
+        if cloudtype in {"GCP", "AWS", "AZURE"}:
             description += "\n\n" + "**" + cloudtype + " Region** : " + node.get("cloud", {}).get("region", "N/A")
             return description
 
         # Add metadata for Kubernetes
-        if cloudtype == "Kubernetes":
+        if cloudtype == "KUBERNETES":
             for resource in node.get("resources", []):
                 pod = resource.get("data", {}).get("metadata", {}).get("name")
                 namespace = resource.get("data", {}).get("metadata", {}).get("namespace")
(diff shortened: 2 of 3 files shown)


dojo/tools/prowler/parser.py
```diff
@@ -0,0 +1,23 @@
+from dojo.tools.prowler.parser_csv import ProwlerParserCSV
+from dojo.tools.prowler.parser_json import ProwlerParserJSON
+
+
+class ProwlerParser:
```
manuel-sommer, line 5: Hi @Jino-T ,
did you also take a look at 
https://github.com/DefectDojo/django-DefectDojo/tree/master/dojo/tools/aws_prowler and https://github.com/DefectDojo/django-DefectDojo/tree/master/dojo/tools/aws_prowler_v3plus ?
Maybe you want to unify them?

Jino-T: Hi Manuel, 
Thank you for the feedback. This was something that we had originally considered, however, these existing parsers are specifically tailored for AWS findings. We decided that having a general parser that standardized deduplication across AWS, Azure, GCP, and Kubernetes findings would be best.

dojo/tools/prowler/parser_json.py
```diff
@@ -0,0 +1,137 @@
...
+            mitigation = (
+                "**Remediation Description** : "
+                + node.get("remediation", {}).get("desc", "N/A")
+                + "\n\n"
+                + "**Remediation References** : "
+                + ", ".join(node.get("remediation", {}).get("references", []))
+            )
+            impact = node.get("risk_details", "")
+            compliance = node.get("unmapped", {}).get("compliance", {})
+            references = "**Related URL** : " + node.get("unmapped", {}).get("related_url")
```
valentijnscholten: if related_url is None this might raise an exception?

dojo/tools/prowler/parser_csv.py
```diff
@@ -0,0 +1,124 @@
...
+
+        dupes = {}
+        for row in csvarray:
+            # Skip vulnerability if the status is "PASS", continue parsing is status is "FAIL" or "MANUAL"
+            if row.get("STATUS") == "PASS":
+                continue
+
+            provider = row.get("PROVIDER", "N/A")
+            # Set provider name to the correct capitalization format
+            provider = provider.capitalize() if provider in {"kubernetes", "azure"} else provider.upper()
```
valentijnscholten: would it be better to enforce the same value always so upper() on all of them? The description and title are part of the dedupe config so if there's a report that has `KuberNetes` as value it will mismatch with a scanner that reports it as `kubernetes` (which gets converted to `Kubernetes`).

dojo/tools/prowler/parser.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def get_description_for_scan_types(self, scan_type):
+        return "Prowler report file can be imported in JSON format or in CSV format."
+
+    def get_findings(self, filename, test):
+        if ".csv" in str(filename).lower():
+            return ProwlerParserCSV().get_findings(filename, test)
+        if ".json" in str(filename).lower():
+            return ProwlerParserJSON().get_findings(filename, test)
+        return None
```
valentijnscholten: Would it be better to make this more strict, i.e. `endsWith('.csv')?
Otherwise some "hacker" might try and upload `my_report.csv.json.xml.bmp.zip.rar` :-)

dojo/tools/prowler/parser_json.py
Jino-T: Nice catch! My code change works if related_url = "" or if there is no related_url attribute.

dojo/tools/prowler/parser_csv.py
Jino-T: That is a good idea. I was trying to keep the format that the company uses but dedupe is definitely more important than that.

dojo/tools/prowler/parser.py
Jino-T: Those hackers are gonna have to try harder after my code change :sunglasses:

mtesauro (approved):
Approved

PR #13833, opened 2025-12-05
fix(GHA): Correction of #13722

kiblik:
These GHAs are hard for local testing. AI recommended me to use a different approach.
Let's hope this will work.


mtesauro (approved):
Approved

PR #13841, opened 2025-12-08
docs: add new page on custom trust

sNiXx:
**Description**

As discussed in #13689, I have added a dedicated page for the custom trust topic. I am not sure if this is the best place to put it, as it may be also required for SAML etc.? Also, will the link I created from the `api/_index.md` to the custom_trust page work like this? Feedback would be much appreciated.

**Test results**

N/A

**Documentation**

This is a docs-only PR

### DIFF
--- a/docs/content/supported_tools/parsers/api/_index.md
+++ b/docs/content/supported_tools/parsers/api/_index.md
@@ -29,3 +29,34 @@ Follow these steps to set up API importing:
 4.  After this is done, you can import the findings on the `Product` page through
     `Findings -> Import Scan Results`. As the `Scan type`, select the related type
     (the `API Scan Configuration` created above) and click `Import`.
+
+## Custom Trust
+
+In some cases, you may want to connect to a tool that uses a certificate from a certification authority (CA) that is not
+in the default trust store (e.g. a company-internal CA), which requires that you add custom trust to an existing trust
+store or replace the existing trust store with your own.
+
+### Using a Custom-Built DefectDojo Image
+
+When you are building your own container image for `django-DefectDojo`, you can simply add the certificates you would
+like to include as custom trust to the `docker/certs` path (see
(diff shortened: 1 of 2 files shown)


mtesauro:
@paulOsinski Can you provide some feedback on this one?

docs/content/supported_tools/parsers/custom_trust.md
```diff
@@ -0,0 +1,34 @@
+---
```
paulOsinski: rather than create a new article, can this be added as a section to the _index.md page?

sNiXx: That is definitely possible. However, I was raising this question before: 

> I am not sure if this is the best place to put it, as it may be also required for SAML etc.? 

I am specifically talking about the `REQUESTS_CA_BUNDLE` env variable being mentioned [here](https://github.com/DefectDojo/django-DefectDojo/blob/b9836f2fffb0588e45b3b6fbffa06bb643543b04/docs/content/en/customize_dojo/user_management/configure_sso.md?plain=1#L655).

So, it would make sense to put it somewhere more "generic", so that it can be also referenced from the SSO page and any other page that may benefit from a link to the custom trust section.

sNiXx: @paulOsinski if you can provide an answer to my question, I can finalize this PR.

paulOsinski: Hi @sNiXx sorry for the delay.  For now, I would prefer we just add this as a section to the _index.md article.  We're continually updating documentation so it might end up somewhere else in the future.

sNiXx: Thank you for the response :-) I just moved it to _index.md, as requested.

paulOsinski (approved):

docs/content/supported_tools/parsers/custom_trust.md
paulOsinski: Thanks @sNiXx !

valentijnscholten (approved):

rossops (approved):

PR #13842, opened 2025-12-08
fix(django): Upgrade to 6.0

kiblik:
TBD


kiblik:
Waiting for https://github.com/django-commons/django-prometheus/pull/489

valentijnscholten:
Although it's interesting to prepare for Django 6.0 compatibility, I think we should sit on this for a bit. Django 5.2 is an LTS release with a long support horizon, longer than 6.0. If we make it work for 6.0 but it breaks in 6.1 and we have to wait for some library to fix it we may get in trouble. 

TL;DR; Let's re-assess after the summer when 6.1 is available.

PR #13849, opened 2025-12-08
Increasing timeouts for unit tests

rossops:
Bump the timeout on the unit test jobs from 15 (default) to 20. 15 was causing some tests to fail.


valentijnscholten:
I think the timeout that is hit is at https://github.com/valentijnscholten/django-DefectDojo/blob/1ba1122954b6fa214862e23345373f75434cdaa1/.github/workflows/rest-framework-tests.yml#L61 so that would need updating me thinks

Maffooch:
Converting to draft until ^ is accommodated

blakeaowens (approved):

valentijnscholten (approved):

PR #13866, opened 2025-12-11
fix(HELM): Use renovate-compatible format

kiblik:
Renovate was not able to recognise that there is a new version HELM chart. Syntax was valid but renovate has some bug there (https://github.com/renovatebot/renovate/discussions/30678).
Now it works, and a "more strict" version pinning is in place.

This is one of the reasons why HELM related GHA were failing.


mtesauro (approved):
Approved

blakeaowens (approved):

PR #13885, opened 2025-12-12
api tokens: allow admins to reset user tokens

valentijnscholten:
Allow superusers to rotate api tokens for users, keep track of last token reset and password reset.
A notfication for the affected user is generated.
[sc-12150]


blakeaowens (approved):

mtesauro (approved):
Approved

PR #13889, opened 2025-12-13
reimport: match findings in batches

valentijnscholten:
Reimport processes all findings in the report. For each finding it tries to find a match in the existing findings already present in the Test. This matching used to be done a a per finding basis. So 1000 findings results in a 1000 database queries. This PR batches this based on the code we also use for deduplication. Candidates for matches are retrieved in batches for a 1000 incoming findings at once. This also allows for prefetching vulnerability_ids, endpoints and endpoin statuses. This makes reimport a lot more effecient.

An improvement of ~77.7% on reimport of identical reports:
```
dev: Successfully reimported 'jfrog_xray_unified/very_many_vulns.json' into test ID 1008 (took 116.76 seconds)
batch: Successfully reimported 'jfrog_xray_unified/very_many_vulns.json' into test ID 1008 (took 25.78 seconds)
```

If the report being reimported has some new or mitigated findings, of course there will still be some time used to process these changes making the real world gains a little lower.

The PR also includes some other optimizations/fixes:
- Only reconcile vulnerability ids if they have changed
- No longer delete existing vulnerability ids twice
- Separate vulnerability_id handling for new and existing findings to allow optimization
- Some low hanging bulk update optimizations for endpoint statuses
- Fix a bug where a change in `cve` was not stored during reimport

This PR affects Pro.

### DIFF
--- a/scripts/update_performance_test_counts.py
+++ b/scripts/update_performance_test_counts.py
@@ -56,8 +56,62 @@ class TestCount:
         )
 
 
+def extract_test_methods(test_class: str) -> list[str]:
+    """Extract all test method names from the test class."""
+    if not TEST_FILE.exists():
+        msg = f"Test file not found: {TEST_FILE}"
+        raise FileNotFoundError(msg)
+
+    content = TEST_FILE.read_text()
+
+    # Find the test class definition
+    class_pattern = re.compile(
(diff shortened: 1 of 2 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

dogboat (approved):

PR #13891, opened 2025-12-13
reimport: optimize vulnerability_id processing

valentijnscholten:
Reimport was overwriting vulnerability_ids for every finding in the report. Since this is a OneToMany field this is expensive.
This PR prefetches the existing vulnerability_ids and only updates them if the report being reimported has different vulnerability_ids for this finding. Also it gets rid of the duplicate delete query for existing findings.

on a reimport of the 13k jfrog sample report this saves almost 50% of time (117s -> 63s)

```
Successfully reimported 'jfrog_xray_unified/very_many_vulns.json' into test ID 1008 (took 116.76 seconds)
Successfully reimported 'jfrog_xray_unified/very_many_vulns.json' into test ID 1008 (took 62.76 seconds)
```

The pr also fixes a bug where a change in finding.cve was not persisted. 
Also added some test cases.

### DIFF
--- a/dojo/importers/default_reimporter.py
+++ b/dojo/importers/default_reimporter.py
@@ -250,14 +250,19 @@ class DefaultReImporter(BaseImporter, DefaultReImporterOptions):
                         unsaved_finding,
                         self.user,
                     )
+                # Existing finding - use reconcile_vulnerability_ids
+                is_new_finding = False
             else:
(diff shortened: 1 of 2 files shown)


valentijnscholten:
It's too time consuming to implement this separate from the batch reimport pr, so I'm closing this.
Changes are included in https://github.com/DefectDojo/django-DefectDojo/pull/13889

PR #13898, opened 2025-12-15
docs: edits to deduplication-tuning docs

LeongBryan:
**Description**

Minor edits to the documentation:
- added missing backticks
- fixed image path to point to correct asset


Jino-T (approved):

valentijnscholten (approved):

mtesauro:
@LeongBryan  Not sure why but that Docs Github action needs to pass before we merge this. That GHA follows the same process we use to generate and publish the docs site so if that GHA fails, generating the docs site on the next release would also fail.

LeongBryan:
Hi @mtesauro, understood. My bad I think on changing the image path. The simplest action was to close this and use a different PR with the single minor change of adding the missing backticks: https://github.com/DefectDojo/django-DefectDojo/pull/13924. 

Sorry for the trouble :)

PR #13916, opened 2025-12-16
push_to_jira: fix pushing to JIRA during import/reimport in asynchronous mode

valentijnscholten:
Findings weren't pushed to JIRA during import/reimport when postprocessing was happening in the background (default behaviour, unless `block_execution` was set to `True`). Bug was introduced in 2.53.0.

This was not caught by our unit tests as these run in synchronous mode.

During debugging I added some logging helpful for troubleshooting, let's leave that in place for now via this PR.


Maffooch (approved):

mtesauro (approved):
Approved

dogboat (dismissed):

Jino-T (dismissed):

mtesauro (approved):
Approved

dogboat (approved):

Issue #13766, opened 2025-11-24
Sorting by SLA does not work in some cases since 77d6bdd0 change


pna-nca:
**Bug description**
When a product has an SLA profile assigned which defines certain expiration days for certain vulnerabilities, SLA counter starts. Some findings may expire, some may not. It is important to have those which close to expiration sorted on top.

This worked previously (although, on client side).

After change the 77d6bdd0 , sorting the list of open findings in a product view ( defectdojo.host/product/{ID}/finding/open?o=sla_age_days ) does not work anymore.

However:
- in some cases it works;
- clicking SLA column changes the order of findings, but the pattern is unclear.

**Steps to reproduce**
Unfortunately, the main problem is to define the situation when the sorting breaks.

In our case, we have some products where it works, and some where it does not.

Can you advice on what to look for to provide you with more information? Which product or finding properties are essential for SLA sorting?

**Expected behavior**
List of open findings in product view can be sorted by SLA counter.

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
DefectDojo: 2.52.2
Helm chart: 1.8.2

**Logs** 
N/A

**Sample scan files**
N/A

**Screenshots**
Sorting by SLA:
<img width="432" height="1038" alt="Image" src="https://github.com/user-attachments/assets/638ded69-3f9e-4a3d-9673-7b825e770572" />

**Additional context** (optional)
N/A

valentijnscholten:
My first instinct would be to look at the database query being executed for this list. Are you able to do some sort of logging or inspection? There are also some guides on how to configure Django to log (all) queries: https://www.geeksforgeeks.org/python/how-to-log-all-sql-queries-in-django/

And what is the url when sorting? It will contain the sort order as a parameter.

pna-nca:
> My first instinct would be to look at the database query being executed for this list. Are you able to do some sort of logging or inspection? There are also some guides on how to configure Django to log (all) queries: https://www.geeksforgeeks.org/python/how-to-log-all-sql-queries-in-django/

I tried this approach.
The article you linked is not sufficient, in case of DefectDojo one needs to additionally set DD_DEBUG to True.
It would also be nice to have this kind logging controlled fully via environment variables, without the need of rebuilding container images...

[...]

PR #13918, opened 2025-12-17
make ordering by sla_age more reliable

valentijnscholten:
See #13766 

Remove conflicting order by and make the annotation "None/Null safe".

### DIFF
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -256,6 +256,11 @@ class BaseListFindings:
         return findings
 
     def filter_findings_by_form(self, request: HttpRequest, findings: QuerySet[Finding]):
+        # Apply default ordering if no ordering parameter is provided
+        # This maintains backward compatibility with the previous behavior
+        if not request.GET.get("o"):
+            findings = findings.order_by(self.get_order_by())
+
         # Set up the args for the form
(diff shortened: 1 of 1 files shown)


dojo/finding/views.py
```diff
@@ -256,6 +256,11 @@ def filter_findings_by_filter_name(self, findings: QuerySet[Finding]):
         return findings
 
     def filter_findings_by_form(self, request: HttpRequest, findings: QuerySet[Finding]):
+        # Apply default ordering if no ordering parameter is provided
+        # This maintains backward compatibility with the previous behavior
+        if not request.GET.get("o"):
+            findings = findings.order_by(self.get_order_by())
+
```
kiblik, line 263: Isn't it better to put ordering to the model via the `Meta` class?
https://docs.djangoproject.com/en/6.0/topics/db/models/#meta-options

valentijnscholten: it is set on Meta. this line just ensures it's used if no ordering is selected in the filtering. without it, it will generate a query without the ordering from Meta

kiblik: Sorry, I did not understand the issue from the previous reading correctly.

PR #13922, opened 2025-12-17
 Make Twistlock Parser use discoveredDate for Date

Jino-T:
[sc-12290]

### DIFF
--- a/dojo/tools/twistlock/parser.py
+++ b/dojo/tools/twistlock/parser.py
@@ -7,6 +7,7 @@ import textwrap
 from datetime import datetime
 
 import dateutil.parser
+from django.utils import timezone
 
 from dojo.models import Finding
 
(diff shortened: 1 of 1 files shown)


dojo/tools/twistlock/parser.py
```diff
@@ -235,6 +237,10 @@ def get_item(vulnerability, test, image_metadata=""):
     status = vulnerability.get("status", "There seems to be no fix yet. Please check description field.")
     cvssv3_score = vulnerability.get("cvss")
     riskFactors = vulnerability.get("riskFactors", "No risk factors.")
+    try:
+        date = str(dateutil.parser.parse(vulnerability.get("discoveredDate")).date())
+    except (ValueError, TypeError, dateutil.parser.ParserError):
+        date = datetime.now()
```
Maffooch:
```suggestion
        date = timezone.now()
```
using timezone aware dates will suppress some warnings from the logs

dojo/tools/twistlock/parser.py
```diff
@@ -6,6 +6,8 @@
 import textwrap
 from datetime import datetime
 
+import dateutil.parser
```
Maffooch, line 9:
```suggestion
import dateutil.parser
from django.utils import timezone
```
I would be surprised if ruff is happy with the import order with this new line. be prepared to have to run ruff again

Maffooch (changes requested):

valentijnscholten (approved):

PR #13946, opened 2025-12-20
finding template refactor

### LINKED ISSUE #10619: Tags cannot be deleted after saving per finding.
Tags cannot be deleted after saving per vulnerability.  I have tried many times to modify my tags, but after clicking finish, the previous tags are still there and it will not be deleted. Another person had another problem with tags (#10618 issue), it seems that there is a problem in the new update of this part.

valentijnscholten:
Fixes #10619 

Breaking changes / feature removal:
- As communicated in the [2025Q1 community update](https://github.com/DefectDojo/django-DefectDojo/discussions/12153) the automated matching of Finding Templates based on `CWE` and/or `title` has now been removed.

Enhancements to the Finding_Template model:
- added all relevant fields from the Finding model.
- added endpoints
- simplify vulnerability_id handling
- setting tags from template now works


mtesauro:
I got DryRun Security and our Docs Dry Run mixed up. I don't see why that docs GHA is failing though.

valentijnscholten:
@mtesauro Me neither, it seems to go OOM on generating an image. But I didn't touch anything related to that. But it does fail consistently on this PR. Who knows more about this docs thing? May be we need to increase the memory setting but we would have to do that on master possibly or at least dev to get GH to pick those up.

blakeaowens (approved):

PR #13951, opened 2025-12-22
docs: remove info about optional service key

sNiXx:
**Description**

"Service key 1" for the SonarQube Tool Configuration is said to be optional. However, I tested adding a configuration without it and got the following error:
image showing an error when adding a SonarQube Tool Configuration

Therefore, I would suggest to remove this sentence. Or, if such behavior is desirable, add the missing functionality that uses the product name (DefectDojo) as the project key (SonarQube) if "Service key 1" is left empty.

**Test results**

N/A

**Documentation**

This is a docs-only PR

### DIFF
--- a/docs/content/supported_tools/parsers/api/sonarqube.md
+++ b/docs/content/supported_tools/parsers/api/sonarqube.md
@@ -22,11 +22,8 @@ separate the items in the "Extras" field by a vertical bar (e.g.
 In `Add API Scan Configuration`
 -   `Service key 1` must
     be the SonarQube project key, which can be found by navigating to a specific project and
-    selecting the value from the url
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Thanks for letting us know. 

Looking at the code there is a fallback in place:

https://github.com/DefectDojo/django-DefectDojo/blob/f322a049f1b9ad9daa9ef52fe63cc41806c8c0fe/dojo/tools/api_sonarqube/importer.py#L113-L125.

Might be better to report this as is bug report with the stacktrace from the logs so that it can be fixed/removed.

sNiXx:
> Might be better to report this as is bug report with the stacktrace from the logs so that it can be fixed/removed.

Good idea, just did that: #13959

Thank you for also highlighting the code section :+1:

PR #13955, opened 2025-12-22
github action fetch openapi spec must wait for dojo to be up

valentijnscholten:
don't fetch spec before dojo is fully started.


valentijnscholten:
Merged with a approval as it's release ops related.

PR #13960, opened 2025-12-22
release process fix for 2.53.4: also start valkey is it's now required by the entrypoitn scripts

valentijnscholten:


Maffooch:
Merging with one approver since this is a release ops related

valentijnscholten:
Needed because of #13931

PR #13963, opened 2025-12-22
Fix CycloneDX XML import failing when vulnerability description is missing

### LINKED ISSUE #10249: Import of CycloneDX Scan fails if vulnerability description is missing
**Bug description**
When uploading a CycloneDX report to DefectDojo (2.34.4) an error is throw over a missing description fields in the vulnerability section. As the documentation of CycloneDX does not mandate the description field to be present in the report this seems like a bug, because DefectDojo requires this field in the report.

`16/May/2024 13:22:55] ERROR [dojo.api_v2.exception_handler:36] null value in column "description" of relation "dojo_finding" violates not-null constraint`

**Steps to reproduce**
1. Go to DefectDojo
2. Go to an engagement
3. Upload the sample file attached (petclinic-bom.xml)
4. See error:

**Expected behavior**
I would expect that the CycloneDX report would upload successfully if the 'desription' field is not mandatory.

- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Official DefectDojo docker image from DockerHub
 - DefectDojo version 2.34.4

**Logs** 
`16/May/2024 13:22:55] ERROR [dojo.api_v2.exception_handler:36] null value in column "description" of relation "dojo_finding" violates not-null constraint`

**Sample scan files**
[petclinic-bom.xml.md](https://github.com/DefectDojo/django-DefectDojo/files/15402334/petclinic-bom.xml.md)

**Screenshots**
image

- The issue was flagged previously in https://github.com/DefectDojo/django-DefectDojo/issues/9277. It was supposedly fixed in DefectDojo 2.31.0
- Earlier this year there has been a refactor of the XML parser. It seems the fix was only applied to the legacy xml function. The legacy function get triggered on CycloneDX1.0, while the new function get triggered on newer version. The new function does not set a default Description is the description in the uploaded report is missing.
- manage_vulnerability_legacy() ([Legacy function](https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/cyclonedx/xml_parser.py#L99-L109))
- manage_vulnerabilty_xml() ([New function](https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/cyclonedx/xml_parser.py#L190))

valentijnscholten:
Fixes #10249

## Description
When uploading a CycloneDX report to DefectDojo, an error was thrown when the vulnerability description field was missing. The CycloneDX specification does not mandate the description field to be present, but DefectDojo requires it in the database.

## Root Cause
The `_manage_vulnerability_xml()` function (used for CycloneDX 1.4+) did not set a default value for missing descriptions, while `manage_vulnerability_legacy()` (used for CycloneDX 1.0) already had this fix applied in version 2.31.0.

## Solution
Added a default description when both `description` and `detail` fields are missing, similar to the legacy function. The default description includes the vulnerability ID and severity.

## Testing
- Verified the fix handles missing descriptions correctly
- No linting errors introduced
- Follows the same pattern as the legacy function

### DIFF
--- a/dojo/tools/cyclonedx/xml_parser.py
+++ b/dojo/tools/cyclonedx/xml_parser.py
@@ -194,6 +194,15 @@ class CycloneDXXMLParser:
             "b:ratings/b:rating/b:severity", namespaces=ns,
         )
         severity = Cyclonedxhelper().fix_severity(severity)
+        # by the schema, only id is mandatory, even the severity and description are
+        # optional
+        if not description:
+            description = "\n".join(
+                [
+                    f"**Id:** {vuln_id}",
+                    f"**Severity:** {severity}",
+                ],
+            )
         references = ""
         for advisory in vulnerability.findall(
             "b:advisories/b:advisory", namespaces=ns,


mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

Jino-T (approved):

PR #13969, opened 2025-12-23
Fix: Apply tags to findings/endpoints when TRACK_IMPORT_HISTORY is disabled

### LINKED ISSUE #13312: tags not applied to findings/endpoints if TRACK_IMPORT_HISTORY is disabled
If `TRACK_IMPORT_HISTORY ` (default True) is disabled, the tags are not applied to findings and endpoints when the import is done.

https://github.com/DefectDojo/django-DefectDojo/blob/452739255adc514cedabb64d2db73803ef1be934/dojo/importers/base_importer.py#L344

Possible fixes:
- apply tags regardless. this will require some work as the current logic depends on the import history being present
- remove the `TRACK_IMPORT_HISTORY` flag

valentijnscholten:
Fixes #13312

## Problem
When `TRACK_IMPORT_HISTORY` is disabled, tags were not being applied to findings and endpoints during import because the tag application logic was inside `update_import_history()` which returned early when the setting was disabled.

## Solution
Refactored the code to:
- Extract tag application into a dedicated `apply_import_tags()` method
- Call `apply_import_tags()` from importers after `update_import_history()`
- Remove tag application logic from `update_import_history()`

This ensures tags are applied regardless of the `TRACK_IMPORT_HISTORY` setting while maintaining separation of concerns and eliminating code duplication.

## Changes
- `dojo/importers/base_importer.py`: Added `apply_import_tags()` method and removed tag logic from `update_import_history()`
- `dojo/importers/default_importer.py`: Added call to `apply_import_tags()` after `update_import_history()`
- `dojo/importers/default_reimporter.py`: Added call to `apply_import_tags()` after `update_import_history()`

### DIFF
--- a/dojo/importers/base_importer.py
+++ b/dojo/importers/base_importer.py
@@ -1,6 +1,7 @@
 import base64
 import logging
 import time
+from collections.abc import Iterable
 
 from celery import chord, group
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


mtesauro (approved):
Approved

blakeaowens (approved):

PR #13975, opened 2025-12-24
Add test_type mismatch validation during reimport (#10219)

### LINKED ISSUE #10219: DD Silently accepts `test_type` mismatch

valentijnscholten:
## Description

This PR fixes issue #10219 where DefectDojo silently accepts test_type mismatches during reimport.

## Changes

- Added validation in `consolidate_dynamic_tests` method to detect test_type mismatches during reimport
- Raises `ValidationError` with descriptive message when test_type doesn't match existing test
- Validation occurs before any findings are processed or deduplication starts
- Added comprehensive test cases covering:
  - Initial import (should create new test successfully)
  - Reimport with matching test_type (should succeed)
  - Reimport with different test_type (should fail with ValidationError)

## Testing

All test cases pass:
- `test_import_generic_with_custom_test_type` - Initial import creates test correctly
- `test_reimport_generic_with_matching_test_type` - Reimport with matching test_type succeeds
- `test_reimport_generic_with_different_test_type` - Reimport with different test_type raises ValidationError

## Related Issue

Fixes #10219

### DIFF
--- a/docs/content/en/working_with_findings/organizing_engagements_tests/product_hierarchy.md
+++ b/docs/content/en/working_with_findings/organizing_engagements_tests/product_hierarchy.md
@@ -25,9 +25,9 @@ Product Types can have Role\-Based Access Control rules applied, which limit tea
 
 #### What can a Product Type represent?
 
-* If a particular software project has many distinct deployments or versions, it may be worth creating a single Product Type which covers the scope of the entire project, and having each version exist as individual Products.  
+* If a particular software project has many distinct deployments or versions, it may be worth creating a single Product Type which covers the scope of the entire project, and having each version exist as individual Products.
 ​
-* You also might consider using Product Types to represent stages in your software development process: one Product Type for 'In Development', one Product Type for 'In Production', etc.  
+* You also might consider using Product Types to represent stages in your software development process: one Product Type for 'In Development', one Product Type for 'In Production', etc.
 ​
 * Ultimately, it's your decision how you wish to organize your Products, and what you Product Type to represent. Your DefectDojo hierarchy may need to change to fit your security teams' needs.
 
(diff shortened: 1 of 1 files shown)


valentijnscholten:
Every time I look at this feature I have to find out how it works. To help myself and other I added a short piece of documentation, let me know if that makes sense.

Maffooch (approved):

Jino-T (approved):

PR #13984, opened 2025-12-26
remove dojo_model_to/from_id decorator

valentijnscholten:
- The custom decorators that we have on Celery tasks interfere with some (advanced) celery functionality. 
- The impact of removal is limited and we can now choose explicitly whether to pass on model `id`'s or `instance`s. 
- It also reduced the difference between running `block_execution==False` (default)  or `block_execution==True`.
- `CELERY_PASS_MODEL_BY_ID` was always True as turning it to False would break everything as models with Tagulous fields cannot be serialized

### DIFF
--- a/unittests/test_deduplication_logic.py
+++ b/unittests/test_deduplication_logic.py
@@ -1075,6 +1075,7 @@ class TestDuplicationLogic(DojoTestCase):
         finding_new.is_mitigated = False
         finding_new.unique_id_from_tool = original_unique_id
         finding_new.save()
+        finding_new.refresh_from_db()
 
         # The new finding should be marked as duplicate of the active finding,
         # not the mitigated one (even though mitigated has lower ID)


Maffooch (approved):

dojo/settings/settings.dist.py
```diff
@@ -89,8 +89,7 @@
     DD_CELERY_RESULT_EXPIRES=(int, 86400),
     DD_CELERY_BEAT_SCHEDULE_FILENAME=(str, root("dojo.celery.beat.db")),
     DD_CELERY_TASK_SERIALIZER=(str, "pickle"),
-    DD_CELERY_PASS_MODEL_BY_ID=(str, True),
-    DD_CELERY_LOG_LEVEL=(str, "INFO"),
+        DD_CELERY_LOG_LEVEL=(str, "INFO"),
```
Maffooch:
```suggestion
    DD_CELERY_LOG_LEVEL=(str, "INFO"),
```

PR #13986, opened 2025-12-26
refactor dojo async task base task

valentijnscholten:
The custom decorators that we have on Celery tasks interfere with some (advanced) celery functionality like signatures. This PR refactors this to have a clean base task that passes on context, but does not interfere with celery mechanisms.

The logic to decide whether or not the task is to be called asyncrhonously is now in a dispatch method.

Pro needs similar changes via https://github.com/DefectDojo-Inc/dojo-pro/pull/973, these must be merged "at the same time".

### DIFF
--- a/dojo/importers/location_manager.py
+++ b/dojo/importers/location_manager.py
@@ -6,7 +6,7 @@ from django.db.models import QuerySet
 from django.utils import timezone
 
 from dojo.celery import app
-from dojo.decorators import dojo_async_task
+from dojo.celery_dispatch import dojo_dispatch_task
 from dojo.location.models import AbstractLocation, LocationFindingReference
 from dojo.location.status import FindingLocationStatus
 from dojo.models import (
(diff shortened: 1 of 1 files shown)


valentijnscholten:
I have had to include the commits from https://github.com/DefectDojo/django-DefectDojo/pull/14090 in this PR to make sure the downstream Pro codebase still works.

valentijnscholten:
i've also had to include the commits from https://github.com/DefectDojo/django-DefectDojo/pull/14109 to make the downstream  Pro codebase work.

valentijnscholten:
And then I had to include the commits from https://github.com/DefectDojo/django-DefectDojo/pull/14080 to fix other errors in Pro.

dogboat:
This looks great but will probably cause conflicts with the [locations PR](https://github.com/DefectDojo/django-DefectDojo/pull/14198); at the least, some of the code there will need to be updated to take advantage of these improvements. Maybe we should hold off on merging this one until the next release? @mtesauro

mtesauro (changes requested):
I would like to hold on this until after Locations lands as getting Locations in place allows other code changes to move forward while avoiding inevitable conflicts.

valentijnscholten:
I don't really mind if we merge this one before monday or shortly after. But we have to be aware that any async tasks related to endpoints should probably be converted into synchronous code or be called with `sync=True` because Locations are considered for deduplication.

valentijnscholten:
Closed in favor of https://github.com/DefectDojo/django-DefectDojo/pull/14240

PR #13994, opened 2025-12-29
Use system pycurl packages instead of pip installation

valentijnscholten:
## Description

Remove pycurl from `requirements.txt` and use OS-provided system packages instead to eliminate the need for compilation during Docker builds.

### 1. pycurl is NOT directly used by DefectDojo

```bash
$ grep -r "import.*pycurl\|from.*pycurl" dojo/
# No results - pycurl is never imported in DefectDojo code
```

pycurl is only a **transitive dependency** for Celery's SQS broker transport. It's only needed when `DD_CELERY_BROKER_SCHEME=sqs` is configured.

### 2. No binary wheels for Linux

pycurl does not provide pre-compiled `manylinux` or `musllinux` wheels, requiring compilation from source during every Docker build. This requires:
- `curl-dev` / `libcurl4-openssl-dev` headers
- `gcc` and build tools
- `PYCURL_SSL_LIBRARY=openssl` environment variable

### 3. System packages are available and well-maintained

| Platform | Package | Version |
|----------|---------|---------|
| Debian Trixie | `python3-pycurl` | 7.45.6 |
| Alpine 3.22 | `py3-curl` | 7.45.6 |

Both provide the same version, ensuring consistency across image variants.

## Impact on Users

- **SQS users**: No change - pycurl is still available via system packages
- **Non-SQS users**: No change - pycurl was never used anyway
- **Local development**: Developers not using Docker using SQS locally need to install pycurl manually (`pip install pycurl` or system package)


valentijnscholten:
Now that is looks like we will keep building our own pscyopg and uwsgi packages, moving pycurl to a system package has limited benefits. I created #14004 instead to use the recommened way to install celery + sqs dependencies.

Jino-T (approved):

valentijnscholten:
Closing as we're going with https://github.com/DefectDojo/django-DefectDojo/pull/14004

PR #13999, opened 2025-12-29
Use system uWSGI packages instead of pip installation

valentijnscholten:
Builds on https://github.com/DefectDojo/django-DefectDojo/pull/13993 and https://github.com/DefectDojo/django-DefectDojo/pull/13993 so merge those first.

## Description

Replace pip-installed uWSGI with OS-provided system packages to eliminate compilation during Docker builds. This removes the need for build toolchains (gcc, build-essential, python3-dev) in Docker images.

### 1. Eliminates Compilation

uWSGI requires compilation from source when installed via pip, which:
- Requires build toolchain (gcc, build-essential, python3-dev, libffi-dev)
- Takes significant time during Docker builds
- Increases image size with build dependencies

### 2. System Packages Are Well-Maintained

| Platform | Package | Version | Maintainer |
|----------|---------|---------|------------|
| Debian Trixie | `uwsgi` + `uwsgi-plugin-python3` | 2.0.28 | Debian maintainers |
| Alpine 3.22 | `uwsgi` + `uwsgi-python3` | 2.0.30 | Alpine maintainers |

These are maintained by official distribution teams and receive security backports.

### 3. Version Considerations

- **Debian**: 2.0.28 (3 patch versions behind pip 2.0.31)
- **Alpine**: 2.0.30 (1 patch version behind pip 2.0.31)

The downside is that we no longer get Renovate / Dependabot PRs for updates of uwsgi.
The patch-level differences are typically minor bug fixes. Security patches are backported to system packages within their release cycle. I can't remember any time that these small patches broke Defect Dojo. In general it might be better to rely on official binaries instead of compiling our own possibly with suboptimal settings or small changes depending on where/how they are built. But I'm open for discussion/happy to close if we don't get consensus.

### DIFF
--- a/Dockerfile.django-alpine
+++ b/Dockerfile.django-alpine
@@ -55,9 +55,17 @@ RUN \
 	--no-cache-dir \
 	--no-index \
   --find-links=/tmp/wheels \
-  -r ./requirements.txt && \
-  # Install uwsgi via pip for Alpine (system package is compiled for Python 3.12, but we use Python 3.13)
-  pip3 install --no-cache-dir uwsgi==2.0.28
+  -r ./requirements.txt
+
+# Install uwsgi via pip for Alpine (system package is compiled for Python 3.12, but we use Python 3.13)
+# uwsgi needs build dependencies to compile
+RUN apk add --no-cache --virtual .uwsgi-build-deps \
+    gcc \
+    musl-dev \
+    linux-headers \
+    python3-dev \
+    && pip3 install --no-cache-dir uwsgi==2.0.28 \
+    && apk del .uwsgi-build-deps
 
 COPY \
   docker/entrypoint-celery-beat.sh \


valentijnscholten:
Turns out it doesn't work on Alpine because alpine comes with Python 3.12 by default, so the uwsgi plugin apks is compiled against Python 3.12 while we use Python 3.13. In Debian it works (for now) as Trixie uses Python 3.13 out of the box.
I fixed it a little bit dirty in this PR to have the Alpine builds fall back to pip for uwsgi installation.
This feels a bit hacky. And once we want to go to Python 3.14 we will have the same problem on Debian.

Two options:

1) Forget about this PR and keep using pip to install uwsgi and accept it takes a long time to build
2) Forget about having our own python version and just use the default that comes with Alpine or Debian.

I think option 2 is not really what we want, so this leaves option 1.

PR #14004, opened 2025-12-30
use celery[sqs] bundle

valentijnscholten:
The official docs suggest to install `celery[sqs]` to allow for AWS SQS support instead of manually installing pycurl, boto2 and whatever else maybe needed in the future.

I think we should go with this route and close https://github.com/DefectDojo/django-DefectDojo/pull/13994


mtesauro:
Merging as this is effectively a change in dependencies like Dependabot & Renovate - it actually make that more simple.

PR #14009, opened 2025-12-31
Ruff: use Python 3.14 compatible code

kiblik:
Using 3.14, we should also follow Ruff's recommendation for this version.


valentijnscholten:
merged into the renovate branch which will get reviewed before merge into dev.

PR #14011, opened 2025-12-31
dedupe reopen: continue to try all match candidates

### LINKED ISSUE #14010: Regression: deduplication stops when an inactive/migitated candidate has been found
In some scenario's there are multiple existing findings that are a candidate for deduplication.

In the pre- 2.53.0 all of these candidates where considered. Since 2.53.0 only the first candidate is considered. Usually this is not a problem as almost all matching logic is done by the candidated queries or matching function. However the `set_duplicate` method can raise expcetions in certain corner cases not covered by the candidated selection/matching:

https://github.com/DefectDojo/django-DefectDojo/blob/87ff93ad596e30dc905d0c33147148865d528a5f/dojo/finding/deduplication.py#L128-L166

This behaviour was introduced in https://github.com/DefectDojo/django-DefectDojo/pull/9558

After this exception the finding is left active non-duplicate. What should happen is that the next matching candidate is tried with `set_duplicate`. That would restore the old behaviour. 

This PR also removes the no longer used `set_duplicate_reopen` function which is no longer used since https://github.com/DefectDojo/django-DefectDojo/pull/9558.

Reported via Slack after an upgrade from 2.50.0 to 2.53.1: https://owasp.slack.com/archives/C2P5BA8MN/p1767021148268659?thread_ts=1767017147.331459&cid=C2P5BA8MN

valentijnscholten:
Fixes #14010

This PR does not affect Pro which has its own PR.

### DIFF
--- a/dojo/finding/deduplication.py
+++ b/dojo/finding/deduplication.py
@@ -345,7 +345,7 @@ def _is_candidate_older(new_finding, candidate):
     return is_older
 
 
-def match_hash_candidate(new_finding, candidates_by_hash) -> Iterator[Finding]:
+def get_matches_from_hash_candidates(new_finding, candidates_by_hash) -> Iterator[Finding]:
     if new_finding.hash_code is None:
         return
     possible_matches = candidates_by_hash.get(new_finding.hash_code, [])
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

PR #14013, opened 2025-12-31
Add permission classes and refine queryset in BurpRawRequestResponseViewSet

Maffooch:
Enhance the BurpRawRequestResponseViewSet by adding permission classes to restrict access and refining the queryset to filter results based on user permissions. This change improves security and ensures users only see authorized findings.


mtesauro (approved):
Approved

PR #14017, opened 2025-12-31
Vulnerability IDs: Do not allow users to import empty strings

Jino-T:
[sc-12317]


dojo/finding/helper.py
```diff
@@ -801,6 +808,8 @@ def save_vulnerability_ids_template(finding_template, vulnerability_ids):
     # Remove old vulnerability ids
     Vulnerability_Id_Template.objects.filter(finding_template=finding_template).delete()
 
+    # Remove undisired vulnerability ids
+    sanitize_vulnerability_ids(vulnerability_ids)
```
Maffooch:
```suggestion
```

PR #14024, opened 2026-01-02
[docs] create sitemap at root on Hugo deploy

paulOsinski:
Hugo is not creating a sitemap at the root of our docs (which should be at https://docs.defectdojo.com/sitemap.xml)

This PR fixes that issue and will create an XML sitemap in our root folder, listing all active URLs.


mtesauro (approved):
Approved

rossops (approved):

PR #14058, opened 2026-01-07
Re order Jira Alert Description

Jino-T:
[sc-12181]

The goal is to have the error appear above the finding details because oftentimes the error will get trimmed from the alert.

Before:
tmp4

After:
tmp3


mtesauro (approved):
Approved

PR #7889, opened 2023-03-27
Remove Rules Framework and Google Sheets Integration

Maffooch:
Removing these features as they have been marked for removal for a long time
[sc-711]

PR #14075, opened 2026-01-12
Product Grade: Configuration Removal

Maffooch:
Configuration of the product grade calculation has been removed

### DIFF
--- a/unittests/test_importers_performance.py
+++ b/unittests/test_importers_performance.py
@@ -310,9 +310,9 @@ class TestDojoImporterPerformanceSmall(TestDojoImporterPerformanceBase):
         self.system_settings(enable_product_grade=True)
 
         self._import_reimport_performance(
-            expected_num_queries1=315,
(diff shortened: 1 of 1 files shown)


Jino-T (approved):

kiblik:
Why this? There are users using different evaluation methods. There is no way to customize it now.

Maffooch:
The method used to customize the product grading was very prone to system breaking mistakes, and should have been removed in #7889 along with the old rules engine that operated in a similar manner. I plan to post a message in the community slack channel later today

X0x1RG9f:
Hi @Maffooch, thanks for the wonderful work !

A little bit worried / questioning about this merge, does it means in the future there will be no possibility to have grades at all ? To have grades configured dynamically?

Maffooch:
@X0x1RG9f product grade are still being calculated, just with the default function that came with dojo out the box. The only change here is that the function is no longer configurable

PR #14079, opened 2026-01-12
Remove unused asteval dependency

valentijnscholten:
- Remove `asteval==1.0.8` from `requirements.txt`
- Remove asteval license notice from `NOTICE` file


Maffooch (approved):
dep related, so we could merge with two

PR #14080, opened 2026-01-12
Asset/Organizations Endpoints: Patches, permission checking, and API tests

Maffooch:
Enhance the `AssetSerializer` to allow null values and set defaults for certain fields. Refactor authorization functions to include type hints, improving code clarity and maintainability. Update permission checks to support multiple primary key attributes in post requests.

[sc-12453]

### DIFF
--- a/dojo/api_v2/permissions.py
+++ b/dojo/api_v2/permissions.py
@@ -684,6 +684,7 @@ class UserHasProductTypeGroupPermission(permissions.BasePermission):
             Permissions.Product_Type_Group_Delete,
         )
 
+
 class UserHasOrganizationGroupPermission(permissions.BasePermission):
     def has_permission(self, request, view):
         return check_post_permission(


valentijnscholten (commented):
Thinking out loud: Could this be abused by providing both product and asset parameters. Product with a value where the user has access but asset value they don't have access?

Maffooch:
That's a good question! I will test and find out. Worst case is I will duplicate the permission classes

Maffooch:
Answer is yes and no. The permission check can be bypassed with the method described, but the resulting queryset does not have the object of the "other" permission. For example, I submit a post to the asset endpoint against a `prod_type` my user does have access to, but the serializer will attempt to add the new product/asset to ID of the `organization`

Payload:
```json
{
  "name": "sec",
  "description": "string",
  "prod_type": 1,
  "organization": 2
}
```
Response:
```json
{
  "organization": [
    "Invalid pk \"2\" - object does not exist."
  ],
  "message": "{'organization': [ErrorDetail(string='Invalid pk \"2\" - object does not exist.', code='does_not_exist')]}"
}
```

I don't really love the idea of being protected by just the queryset filtering, so I'm gonna duplicate the permission classes instead to err on the side of caution. Thank you for raising this!

PR #14107, opened 2026-01-16
Import/Reimport: Push to jira when findings is not grouped

Maffooch:
When pushing to jira with the `group_by` parameter, if the group critera selected does not assign a finding to a new/existing group, the finding will not be pushed to jira. This is because the logic to push to jira will skip an inidividual finding in favor of pushing a group. However, we can get in a situation where we do not end up with a finding group, and the result will be a finding that was not pushed to jira, but should have been.

The solution taken here is to track if a finding was a candidate for a new/exitisng group, and then use that result to determine the decision to delay the push to jira logic for a group, or just push the single finding

[sc-12485]


valentijnscholten:
@Maffooch If I remember correctly my original interntion was that all findingd would be grouped, even if some needed fields are not present. From that point of view I would think a group with title "None" or "Unknown" or "Ungrouped" or "Unmapped" or whatever would be preferably. The goal then would be to prevent the "mixed state" where we have findings that have a group issue and findings that have an individual issue.
However it seems that this mixed state can occur anyway via other ways so we should probably support it, which is what this PR does. Had you also considered the above?

Issue #13899, opened 2025-12-15
UWSGI container high memory usage on Kubernetes with 2.53.1


goat-ssh:
**Bug description**
We've upgraded v2.50.4 -> 2.53.1. Our usage pattern have not changed and the container has quite some spikes

**Expected behavior**
Memory usage is stable

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [x] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version 2.53.1
 - k8s 1.33 AKS

**Screenshots**

<img width="1723" height="857" alt="Image" src="https://github.com/user-attachments/assets/e32652f5-0ac6-49b7-b4e6-e336d0b52079" />

**Additional context** (optional)

helm values (chart 1.9.1):
```
celery:
  beat:
    resources:
      limits:
        cpu: 500m
      requests:
        cpu: 100m
  worker:
    resources:
      limits:
        cpu: 500m
      requests:
        cpu: 100m
createPostgresqlSecret: true
createSecret: true
createValkeySecret: true
django:
  ingress:
    annotations:
      nginx.ingress.kubernetes.io/proxy-body-size: 50M
    ingressClassName: nginx
  nginx:
    resources:
      limits:
        cpu: 200m
      requests:
        cpu: 100m
  uwsgi:
    enableDebug: true
    resources:
      limits:
        cpu: 500m
        memory: 1Gi
      requests:
        cpu: 100m
global:
  postgresql:
    auth:
      password: ddddd
host: dojo.xyz
```

valentijnscholten:
Quite a lot has changed between 2.50.x and 2.53.x. Do you have any idea which version introduced it?

The upgrade notes lists the biggest changes:
- https://docs.defectdojo.com/en/open_source/upgrading/2.51/
- https://docs.defectdojo.com/en/open_source/upgrading/2.52/
- https://docs.defectdojo.com/en/open_source/upgrading/2.53/
- and https://github.com/DefectDojo/django-DefectDojo/pull/13914
- python 3.12 -> 3.13

Is there anything in there that rings a bell or is typically used a lot in your instance?

Biggest changes in the uwsgi container code in these releases have been about improving performance. But some of them could lead to a little more memory use. The graph however doesn't show "massive" memory use. More like a tradtional saw tooth pattern I know from Java applications.  We don't get many memory issue reports and I have never looked into how to debug them.

[...]

PR #14117, opened 2026-01-18
fix bleach memory leak & simplify git commit hash checker

valentijnscholten:
While investigating #13899 I found and fixed:
- bleach was used in a way where for every request it was adding things to the allow list which made the allowlist grow over time (confirmed small memory leak)
- the git commit hash is being checked and displayed on *every page* using a subprocess that literally executes git. there were some log lines intermittently about processes and files not being closed properly. some memory analysis showed the top 3 allocations where all from the git python module. this PR stops executing git and just looks at the `.git` folder to get the commit hash (possible memory leak)

Although I don't think these two issues solve the reported "ever increasing memory usage", it will definitely help.

I also looked at including some memory diagnostics functionality using `tracemalloc`. But this slows down the application a lot so it's not suited to enable or even bundle in the production images.


mtesauro (approved):
Approved

PR #14124, opened 2026-01-19
Fix Content-Type header bugs in file downloads and MIME type handling

### LINKED ISSUE #14118: The API response header format for /api/v2/findings/{id}/files/download/{file_id}/ is incorrect.
When you use the `/api/v2/findings/{id}/files/download/{file_id}/` API (including when clicking to download an image on the find viewing page), the response packet returns the image as a TXT file because the `Content-Type: ('image/png', None)` header in the response packet is formatted incorrectly.
Image
This is because in line 2659 of dojo\utils.py, content_type=f"{mimetypes.guess_type(file_path)}, mimetypes.guess_type() returns (type, encoding), so it should be changed to content_type=f"{mimetypes.guess_type(file_path)[0]} (the same problem was found in line 733 of api_v2\views.py).

valentijnscholten:
## Summary

Fixes #14118

This PR fixes multiple bugs related to MIME type handling in file downloads across the application:

- **Fixed tuple-as-string bug**: `mimetypes.guess_type()` returns a tuple `(type, encoding)`, but it was being used directly in f-strings, resulting in invalid `Content-Type` headers like `"('image/png', None)"` instead of `"image/png"`
- **Added fallback for unknown file types**: When MIME type cannot be determined (returns `None`), now falls back to `"application/octet-stream"` 
- **Fixed JSON content type**: Changed incorrect `content_type="json"` to proper `"application/json"`
- **Fixed potential crash**: Prevented `AttributeError` in template tag when attempting to call `.startswith()` on `None`

All file downloads now properly set `Content-Type` headers with appropriate fallbacks for unknown file types.

### DIFF
--- a/dojo/api_v2/views.py
+++ b/dojo/api_v2/views.py
@@ -751,7 +751,7 @@ class RiskAcceptanceViewSet(
         # send file
         response = FileResponse(
             file_handle,
-            content_type=f"{mimetypes.guess_type(str(file_path))}",
+            content_type=mimetypes.guess_type(str(file_path))[0] or "application/octet-stream",
             status=status.HTTP_200_OK,
         )
         response["Content-Length"] = file_object.size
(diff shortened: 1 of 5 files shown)


mtesauro (approved):
Approved

Issue #14136, opened 2026-01-21
Trivy parser doesn't parse vulnerability_ids and file_path for misconfigurations


DeD1rk:
**Slack us first!**

I tried to join slack but the invite is expired.

**Bug description**

When importing a trivy report, the findings have little detail. For example, the ID/AVDID doesn't appear in the `vulnerability ID from tool` field. The 'target' (typically a path for me) also doesn't show up in any structured field, only in the description. I assume that this also makes accurate deduplication difficult.

**Steps to reproduce**
1. Create a trivy config scan from some OpenTofu/Terraform configurations.
2. Import it.

**Expected behavior**
The findings have the vulnerability ID set and the 'Target' file path in the 'File path', 'SAST Source File Path' or some other field.

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version (see footer) or commit message: v2.54.1 (helm chart 1.9.7)

**Logs** 
n/a

**Sample scan files**
If applicable, add sample scan files to help reproduce your problem.

[trivy-parser-sample.json](https://github.com/user-attachments/files/24763143/trivy-parser-sample.json)

**Screenshots**

Description:
```
**Target:** ../../aws-credentials/main.tf
**Type:** Terraform Security Check

CIS recommends that you apply IAM policies directly to groups and roles but not users. Assigning privileges at the group or role level reduces the complexity of access management as the number of users grow. Reducing access management complexity might in turn reduce opportunity for a principal to inadvertently receive or retain excessive privileges.

One or more policies are attached directly to a user
Number	Content
20	resource "aws_iam_user" "this" {
21	  for_each = var.users
22	
23	  name = "cluster-${var.cluster_name}-${each.key}"
24	}
```

But the structured fields:

<img width="1416" height="896" alt="Image" src="https://github.com/user-attachments/assets/738d0bdd-c824-4166-9a01-b089f77f7236" />

**Additional context** (optional)
Seems related to #9234 and https://github.com/DefectDojo/django-DefectDojo/issues/7728#issuecomment-1453485105

manuel-sommer:
see fix @DeD1rk

DeD1rk:
Nice! And very fast, thanks!

DeD1rk:
Oh it wasn't merged yet of course, oops

manuel-sommer:
done, feel free to close it

PR #14139, opened 2026-01-21
:tada: add Trivy misconfiguration fields #14136

manuel-sommer:
#14136

### DIFF
--- a/docs/content/en/open_source/upgrading/2.54.3.md
+++ b/docs/content/en/open_source/upgrading/2.54.3.md
@@ -1,0 +1,9 @@
+---
+title: 'Upgrading to DefectDojo Version 2.54.3'
+toc_hide: true
+weight: -20250602
+description: Trivy parser deduplication
+---
+
+## Trivy parser deduplication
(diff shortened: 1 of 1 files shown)


valentijnscholten:
This affects deduplication, but there's not much we can do about it.
- Future imports/findings will have *better* deduplication
- Existing findings might not be matched in future imports because the currently existing finding do not have the extra vulnerability_id / file_path values.
Put a not in the upgrade notes to make users aware?

mtesauro:
> This affects deduplication, but there's not much we can do about it.
> 
>     * Future imports/findings will have _better_ deduplication
> 
>     * Existing findings might not be matched in future imports because the currently existing finding do not have the extra vulnerability_id / file_path values.
>       Put a not in the upgrade notes to make users aware?

Yeah, that's probably the best we can do. I do agree that this is better for future imports / dedups.

mtesauro (approved):
Approved

manuel-sommer:
Done, added release notes @valentijnscholten

PR #14140, opened 2026-01-21
Fix risk acceptance API to link to engagement and add validations and permission check

### LINKED ISSUE #12644: Risk Acceptances created via the API do not show within the Risk Acceptance panel of an Engagement
We have an automation to create Risk Acceptances via the API (i.e using endpoint `api/v2/risk_acceptance/`). For that we are using the following payload:

```
{
    "name": "Test Risk Acceptance",
    "recommendation": "A",
    "recommendation_details": "This is just a test risk acceptance",
    "decision": "A",
    "decision_details": "This is just a test risk acceptance",
    "accepted_by": "example@example.com",
    "expiration_date": "2025-06-30T09:49:52.017Z",
    "reactivate_expired": true,
    "restart_sla_expired": false,
    "owner": 8,
    "accepted_findings": [
        234564
    ]
}
```

By DefectDojo design, Risk Acceptances are part of Engagement. In the Engagement view, there's a section that list all Risk Acceptances that are part of that Engagement. However, when a Risk Acceptance is created via the API, such Risk Acceptance will not be listed under the list of Risk Acceptances for that Engagement.

On the other hand, when Risk Acceptance is created via the UI, by clicking the + button, see screenshot below, it works as expected and the Risk Acceptance is listed.

Image

**Can we please have the same behavior as the UI when creating Risk Acceptances via the API?**

This issue probably arises because the API uses the default DRF behavior for creating Risk Acceptances, relying on the serializer without custom logic, while the UI method [add_risk_acceptance](https://github.com/DefectDojo/django-DefectDojo/blob/a81cf78d109f3d39bef73c7c1e2626234ed0574c/dojo/engagement/views.py#L1192) includes additional logic for handling related objects

valentijnscholten:
## Summary

Fixes #12644

This PR addresses several critical issues with the risk acceptance API:

1. **Risk acceptances created via API now appear in engagement panel**
   - Added `engagement.risk_acceptance.add(instance)` in the `create()` method
   - Fixes the main bug where API-created risk acceptances were not linked to engagements

2. **Added validation for `enable_full_risk_acceptance` product setting**
   - API now respects the product-level setting before creating instances
   - Validates in `validate()` method to fail early (no need to create then delete)

3. **Added protection against engagement switching**
   - Prevents moving risk acceptances between engagements via PATCH/PUT
   - Validates even when risk acceptance has no findings (edge case)
   - Uses `self.instance.accepted_findings.all()` instead of filtering for better performance

4. **Comprehensive API tests**
   - Added `test_risk_acceptance_api.py` with 7 test cases
   - Covers all edge cases and validation scenarios
   - All tests passing

5. **API now checks that user has Risk_Acceptance permission and not just Finding.Edit permission**

This PR doesn't break Pro and doesn't need any changes in Pro as Pro has its own separate model and relationship to track links between Risk Aceeptance and Engagement.

### DIFF
--- a/dojo/api_v2/serializers.py
+++ b/dojo/api_v2/serializers.py
@@ -1557,6 +1557,7 @@ class RiskAcceptanceSerializer(serializers.ModelSerializer):
         ra_helper.add_findings_to_risk_acceptance(user, instance, instance.accepted_findings.all())
 
         # Add risk acceptance to engagement
+        # This is fine as Pro has its own model + relationshop to track links with engagements.
         if instance.accepted_findings.exists():
             engagement = instance.accepted_findings.first().test.engagement
             engagement.risk_acceptance.add(instance)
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

PR #14155, opened 2026-01-23
Refactor engagement and risk acceptance permissions

Maffooch:
Introduce a base class for related object permissions to streamline permission checks for engagements, risk acceptances, and findings. Update existing permission classes to utilize the new structure, enhancing code maintainability and clarity.

### DIFF
--- a/dojo/fixtures/dojo_testdata.json
+++ b/dojo/fixtures/dojo_testdata.json
@@ -2823,7 +2823,7 @@
     }
   },
   {
-    "pk": "184770c4c3256aba904297610fbb4da3fa15ba36",
+    "pk": "184770c4c3256aba904297610fbb4da3fa15ba37",
     "model": "authtoken.token",
     "fields": {
       "user": 7,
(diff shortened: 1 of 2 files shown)


dojo/api_v2/views.py
```diff
@@ -736,7 +737,7 @@ def get_queryset(self):
         },
     )
     @action(detail=True, methods=["get"])
-    def download_proof(self, request, pk=None):
+    def download_proof(self, request, pk=None, permission_classes=(IsAuthenticated, permissions.UserHasRiskAcceptanceRelatedObjectPermission)):
```
valentijnscholten: Is there are reason why the permission_classes are passed as a parameter instead of in the decorator? Are they picked up like this?

Maffooch: Good catch, this is supposed to be in the `action` decorator

Maffooch:
> See comment. Also as far as I can see there are no tests covering this new mechanism? Would be good to at least have one to prove it's working. Would this be something you could add and then release in 2.55.0 in the dev branch?

Agreed - the passing of existing DRF tests felt like enough, but I will create some more tests to cover these new classes explicitly

Maffooch:
@valentijnscholten I have added new tests to cover the API endpoints that use these new base classes. This approach felt like a two-for-one deal to get more endpoint coverage while exercising the base classes do what the intend

PR #14177, opened 2026-01-27
Locations more models

dogboat:
Remember when I said the previous PR had all the model changes for URLs/Locations? I lied! Ha! I didn't fool Cody, he knew this would happen.

This PR brings the (maybe actually) final changes to the various models for Locations. Primarily, Locations get an `inherited_tags` field and some of the methods for dealing with those; and URLs have been tweaked a bit to parse more in line with Endpoints for a few of the cases found in tests.

I've included registering Locations/URLs with the audit logging mechanism as well as the post-delete URL signal, as those felt sorta model-y to me as well. There's also a few linter-related changes so this wouldn't fail the PR checks.


dojo/location/models.py
```diff
@@ -151,6 +158,32 @@ def associate_with_product(
...
+            LocationFindingReference.objects.filter(
+                location=self,
+                finding=finding,
+            ).delete()
+            # If there are no findings left associated with this location, remove the association with the product too
+            if not LocationFindingReference.objects.filter(
+                location=self,
+                finding__test__engagement__product=finding.test.engagement.product,
+            ).exists():
+                self.disassociate_from_product(finding.test.engagement.product)
```
dogboat: Perhaps unwanted, since it's possible to associated a product with a location without a finding, and this could potentially blow such intended things away. I included it because it mirrored how `associate_with_finding()` works, but I wouldn't be at all surprised if y'all say remove this.

mtesauro (approved):
Approved

dojo/location/models.py
Maffooch: The disassociation itself is great (and I was hoping would be managed like if the finding was deleted for example) but auto disassociation with the product sorta defeats the purpose of the project. We are seeking to know "Endpoint X is in scope of the product, and I can prove it is not vulnerable because there are no findings attached to it"

Maffooch:
```suggestion
```
I suppose I could add a "point and click" removal method 😄

dogboat: Hahaha you make my life easy, thank you!

Maffooch (approved):

Jino-T (approved):

blakeaowens (approved):

PR #14178, opened 2026-01-27
updated ui tests: dedupe_test and product_test

dogboat:
This PR includes some updates to UI/integration tests to work with V3.

The first, to dedupe_test, isn't strictly necessary but more a quality-of-life-for-my-local-dev thing. I noticed when running several of the tests in sequence, sometimes the false positive history setting gets enabled, causing the subsequent dedupe test that attempts to enable dedupe to fail as these setting conflict. The edit to that test deselects false positive history/retroactive application of it as necessary when enabling dedupe to prevent that conflict. Again, not necessary if you don't want it.

The edits to product_test are more necessary. One of the changes made to the UI for inputting URLs in V3 is that there is no longer a single freeform textbox that can take multiple entries. The test has been updated to detect whether the single input exists and use it (for V2 Endpoints) or use the new UI if not. The product custom fields UI has also been changed (though the change is the same for both V2 and V3), so the tests for add/edit custom fields have been correspondingly updated.

### DIFF
--- a/tests/dedupe_test.py
+++ b/tests/dedupe_test.py
@@ -55,6 +55,12 @@ class DedupeTest(BaseTestCase):
         driver.get(self.base_url + "system_settings")
         if not driver.find_element(By.ID, "id_enable_deduplication").is_selected():
             driver.find_element(By.XPATH, '//*[@id="id_enable_deduplication"]').click()
+            # Disable false positive history if enabled (conflicts with dedupe)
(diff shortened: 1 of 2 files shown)


dogboat:
The failing linter issues have been resolved in the [other PR](https://github.com/DefectDojo/django-DefectDojo/pull/14177), we can wait for that to be merged in and I'll update accordingly.

dogboat:
Closing this one, changes here will be part of the mondo pr instead.

PR #14179, opened 2026-01-27
updates to github workflows

dogboat:
This PR updates some of the github workflows to run all tests (UI/integration and unittests) with locations both enabled and disabled.

Well, "all." Some will be skipped. Herein, the endpoint_test integration test is explicitly skipped when locations are enabled as the functionality it tests no longer exists. Additionally, subsequent PRs will include changes to unittests that skip some functionality, but it'll be handled within the class or as a decorator on it.

Big thanks to Val for having given me a pattern to follow in his pghistory work, and to Cody for directing me to that.

### DIFF
--- a/.github/workflows/rest-framework-tests.yml
+++ b/.github/workflows/rest-framework-tests.yml
@@ -6,12 +6,14 @@ on:
         platform:
           type: string
           default: "linux/amd64"
+        v3_feature_locations:
+          type: boolean
+          default: false
 
 jobs:
   unit_tests:
     name: Rest Framework Unit Tests
     runs-on: ${{ inputs.platform ==  'linux/arm64' && 'ubuntu-24.04-arm' || 'ubuntu-latest' }}
-
     strategy:
       matrix:
         os: [alpine, debian]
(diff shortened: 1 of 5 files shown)


.github/workflows/rest-framework-tests.yml
```diff
@@ -53,10 +55,11 @@ jobs:
 
       # no celery or initializer needed for unit tests
       - name: Unit tests
-        timeout-minutes: 20
+        timeout-minutes: 25
```
dogboat, line 58: I don't _think_ this change is necessary anymore. We can drop it down again if you'd like and change it back if necessary.

dogboat:
The failing linter issues have been resolved in the [other PR](https://github.com/DefectDojo/django-DefectDojo/pull/14177), we can wait for that to be merged in and I'll update accordingly.

.github/workflows/rest-framework-tests.yml
Maffooch: Does not hurt to keep it imo

dogboat:
Closing this one, changes here will be part of the mondo pr instead.

.github/workflows/rest-framework-tests.yml
mtesauro: Agreed.

PR #14180, opened 2026-01-27
feat(async): Show number of tasks waiting in queue

kiblik:
For easier troubleshooting, this PR adds the ability to see how many tasks are waiting for processing.

image

image

image


mtesauro (approved):
Approved

valentijnscholten:
Nice. Recently I was looking into this as well, to provide some more info.

PR #14183, opened 2026-01-27
fix(async): watchmedo was installed incorrectly

kiblik:
The celery Worker was not working in dev/debug mode because the package was not installed correctly.


valentijnscholten:
Hmm I wonder what happened. Maybe it broke on an update to v5 cause I'm sure it worked before :-) Thanks!

PR #14192, opened 2026-01-28
Include Trufflehog verified secret info in report

jamesgol:
Trufflehog will verify certain secrets if it has a detector and they are valid.  Here's a trivial patch to include the verified info in the title and description.

### DIFF
--- a/dojo/tools/trufflehog/parser.py
+++ b/dojo/tools/trufflehog/parser.py
@@ -144,8 +144,6 @@ class TruffleHogParser:
             description += f"**Reason:** {detector_name}\n"
             description += f"**Path:** {file}\n"
             description += f"**Contents:** {redacted_info}\n"
-            if verified:
-                description += f"**Verified:** {verified}\n"
 
             if structured_data:
                 description += f"**Structured Data:**\n{self.walk_dict(structured_data)}\n"


mtesauro:
@jamesgol I've kicked off the GHAs and assuming they go green, approvals will start happening.

Thanks for your contribution! :rocket:

valentijnscholten:
Thanks. Changing the title will affected the deduplication via the hash_code calculation, not sure how to handle this.

jamesgol:
If the title change is an issue we could just drop that and only have it in the description.  I'm trying to come up with a way to make this info available in defect dojo and not have to parse the file again to figure it out.

It looks like some of the other parsers also set the Finding verified flag in certain cases, maybe this would be reasonable here too?

valentijnscholten:
I was about to suggest that, but was not yet sure as people might be using the flag already. But in think that's a good idea to implement. We've also had cases where we did allow changes that affect the deduplication, but implemented as a V2 of the parser.

dojo/tools/trufflehog/parser.py
```diff
@@ -144,6 +144,8 @@ def get_findings_v3(self, data, test):
             description += f"**Reason:** {detector_name}\n"
             description += f"**Path:** {file}\n"
             description += f"**Contents:** {redacted_info}\n"
+            if verified:
+                description += f"**Verified:** {verified}\n"
```
Maffooch:
```suggestion
```
Description is also part of the hash code fields for trufflehog. I think setting the verified flag is the most comprehensive solution

jamesgol: Ahh, didn't realize that.  Haven't been too deep into the internals to know how everything works under the hood.

Maffooch (approved):

mtesauro (approved):
Approved

PR #14195, opened 2026-01-28
feat(async): Set priorities to async tasks

kiblik:
This PR is adding support for priority queues. More important tasks will be processed sooner and vice versa.

The status of each queue is listed in the system settings overview:
image

Queue with id `0` is automatically marked without a suffix (all other are using naming convention `celery:1`, `celery:2`...). Thanks to this fact, we are keeping backward compatibility (if there are any tasks in the queue during migration to the new version, we will still process those tasks after the upgrade).

It is a bit harder to write tests for this. But I tested manually (stop worker, play around, start worker, observe logs) and it was working as expected.

The role of each queue is described in `settings.dist.py` (and reflected in the overview in system settings). If task do not have a priority assigned, priority is set to the default value: `3` 

I'm open to feedback if priority of any task should be changed.

### DIFF
--- a/dojo/finding/helper.py
+++ b/dojo/finding/helper.py
@@ -396,7 +396,7 @@ def add_findings_to_auto_group(name, findings, group_by, *, create_finding_group
                     finding_group.findings.add(*findings)
 
 
-@app.task
+@app.task(priority=3)
 def post_process_finding_save(finding_id, dedupe_option=True, rules_option=True, product_grading_option=True,  # noqa: FBT002
              issue_updater_option=True, push_to_jira=False, user=None, *args, **kwargs):  # noqa: FBT002 - this is bit hard to fix nice have this universally fixed
     finding = get_object_or_none(Finding, id=finding_id)
@@ -459,7 +459,7 @@ def post_process_finding_save_internal(finding, dedupe_option=True, rules_option
             jira_helper.push_to_jira(finding.finding_group)
 
 
-@app.task
+@app.task(priority=4)
 def post_process_findings_batch(
     finding_ids,
     *args,
--- a/dojo/jira_link/helper.py
+++ b/dojo/jira_link/helper.py
@@ -778,7 +778,7 @@ def push_to_jira(obj, *args, **kwargs):
 
 
 # we need thre separate celery tasks due to the decorators we're using to map to/from ids
-@app.task(priority=3)
+@app.task(priority=4)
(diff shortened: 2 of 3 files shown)


dojo/settings/settings.dist.py
```diff
@@ -1245,6 +1246,31 @@ def saml2_attrib_map_format(din):
...
+        CELERY_BROKER_TRANSPORT_OPTIONS["queue_order_strategy"] = "priority"
+
+    if "priority_steps" not in CELERY_BROKER_TRANSPORT_OPTIONS:
+        # There are 4 queues. Lower number = higher priority
+        # 0 - live checks
+        # 1 - keep system consistent (e.g. activation/deavtivation webhooks)
+        # 2 - regular tasks
+        # 3 - notifications and jira
+        # 4 - garbidge collectors
+        CELERY_BROKER_TRANSPORT_OPTIONS["priority_steps"] = list(range(4))
```
valentijnscholten:
```suggestion
        CELERY_BROKER_TRANSPORT_OPTIONS["priority_steps"] = list(range(5))
```

kiblik: 🤦.  Yes. Thanks

kiblik: 🤦.  Yes. Thanks

valentijnscholten:
I've converted it to draft as we can only test/merge this after 2.55.2 has been released and merged back into `dev`.

valentijnscholten:
@kiblik I resolved conflicts and published the PR so we can merge it and test it with Pro.

The PR does not affect Pro in the sense that Pro will keep working and the Pro celery tasks will get priority 3 which is fine. But we can/should do a Pro PR together or after this one to move some of the Pro celery tasks to other priority levels.

valentijnscholten (approved):

valentijnscholten:
@kiblik At first I thought this was a nice way to implement separate queues without having to actually create separate queues in Valkey. However with this approach there's a risk of starvation on the prio 4/5 tasks if an instance is extremely busy (dupe_delete for example). 

It also may make the task ordering unexpected/unpredictable making it harder to analyse performance problems. What do you think? Since we've never had any reports on the need for priorities or separate queues I have become hesitant to merge this. 

A compromise could be to have only two priorities 0 (critical) stuff that is needed to keep the instance alive/healthy (1) everything else.

Maffooch:
This PR got sorta broken... If you would still want this feature added, I ask that you sync your fork, and recreate the PR

PR #14198, opened 2026-01-29
locations: everything else

dogboat:
This PR introduces Locations as a replacement for Endpoints. To migrate to using locations, first the `migrate_endpoints_to_locations` management command should be run, and then the `V3_FEATURE_LOCATIONS` setting should be set to True. This should be considered an irreversible process.

Once enabled, things should generally feel like business as usual, with a few changes to the UI and API. Most notably, the current `/api/v2/endpoints` and `/api/v2/endpoint_status` endpoints will return read-only versions of the data.

Some behind-the-scenes work has been done to migrate to the new terminology vis-à-vis variables, method names, etc., but a good amount of work remains to be done there; some renames cascade into the plumbing of Dojo's testing framework, and this PR is huge enough already.

Tests have been updated to run both against Endpoints and Locations.

Even though this feature took only one commit (amazing, right‽) a lot of the work done here actually belongs to Cody. For anything good, give him the credit; anything bad is probably my fault.


mtesauro (approved):
Approved

blakeaowens (approved):

dojo/tools/cyberwatch_galeax/parser.py
```diff
@@ -205,7 +207,12 @@ def build_findings_for_cve(self, cve_code, c_data, test):
         if not products:
             mitigated_date = timezone.now()
             mitigation = f"Fixed At: {mitigated_date}"
-            endpoints = [Endpoint(host=e) for e in c_data["no_product_endpoints"]]
+            if settings.V3_FEATURE_LOCATIONS:
+                # TODO: Implement Locations
+                locations = [URL(host=e) for e in c_data["no_product_locations"]]
+            else:
+                # TODO: Delete this after the move to Locations
+                locations = [Endpoint(host=e) for e in c_data["no_product_locations"]]
```
valentijnscholten, line 214: Are these TODOs or DONEs?

dojo/tools/wapiti/parser.py
```diff
@@ -93,7 +100,12 @@ def get_findings(self, file, test):
                 # check if dupes are present.
                 if dupe_key in dupes:
                     find = dupes[dupe_key]
-                    find.unsaved_endpoints.extend(finding.unsaved_endpoints)
+                    if settings.V3_FEATURE_LOCATIONS:
+                        # TODO: Implement Locations
+                        find.unsaved_locations.extend(finding.unsaved_locations)
+                    else:
+                        # TODO: Delete this after the move to Locations
```
valentijnscholten, line 106: Are these TODOs or DONEs?

unittests/test_importers_performance.py
```diff
@@ -227,6 +228,8 @@ def _import_reimport_performance(
                             test, _, _len_new_findings, _len_closed_findings, _, _, _ = reimporter.process_scan(scan)
 
 
+# TODO: Implement Locations
```
valentijnscholten, line 231: Anything left TODO here?

valentijnscholten (commented):
The old code has the problem that the Endpoint table doesn't have a unique constraint. This leads to kind of complicated and sometimes expensive logic just to create endpoints during imports. There are situations in which duplicate can or could be created. There's also a warning in the code to log if duplicates are found. In the new code this warning seems to have been removed.

I would say that now is the time to fix the URL model to have a uniqueconstraint. I believe concatting all the fields would not work because they all have a high max length. I think a hash of all fields in a new "url_hash" field could work to make everything around inserts straightforward again. Updates might still be problematic if someone causes a URL to become identical to another URL. But I belivee this problem is already present right now. The update problem is also something that could be fixed later. The insert problem should be done right now as the migration also provides a way to cleanup/deduplicate.

dojo/importers/location_manager.py
```diff
@@ -0,0 +1,164 @@
...
+        saved_url, _ = URL.objects.get_or_create(
+            protocol=unsaved_url.protocol,
+            user_info=unsaved_url.user_info,
+            host=unsaved_url.host,
+            port=unsaved_url.port,
+            path=unsaved_url.path,
+            query=unsaved_url.query,
+            fragment=unsaved_url.fragment,
+            host_validation_failure=unsaved_url.host_validation_failure,
+        )
```
valentijnscholten: This could lead to `MultipleObjectReturned` if there are duplicate entries.

dojo/finding/deduplication.py
```diff
@@ -208,12 +208,48 @@ def are_urls_equal(url1, url2, fields):
...
+
+def are_location_urls_equal(url1, url2, fields):
+    deduplicationLogger.debug("Check if url %s and url %s are equal in terms of %s.", url1, url2, fields)
+    for field in fields:
+        if (field == "scheme" and url1.protocol != url2.protocol) or (field == "host" and url1.host != url2.host):
+            return False
+        if (field == "port" and url1.port != url2.port) or (field == "path" and url1.path != url2.path) or (
+                field == "query" and url1.query != url2.query) or (
+                field == "fragment" and url1.fragment != url2.fragment) or (
+                field == "userinfo" and url1.userinfo != url2.userinfo) or (field == "user" and url1.user != url2.user):
```
valentijnscholten: I think the field is called `user_info` not `userinfo`

valentijnscholten (commented):
I think the below also needs looking in to:
- The `settings.dist.py` still contains only settings for `endpoints` (hashcode/equality related). Should these be renamed/aliased for their `Location` counterparts? And/or docs/comments updated?
- dojo/jira/helper.py contains some code that uses/update endpoints which will no longer work when enabling locations
- dojo/tasks.py displays endpoint count which will no longer be accurate when enabling locations
- view_test.html, view_finding.html reference test_import.import_settings.endpoint instead of test_import.import_settings.locations
- display_tags.py reference import_settings.get("endpoints", get("endpoint")) but should reference locations

dojo/tools/cyberwatch_galeax/parser.py
dogboat: D'oh, first is done (and removed now, thanks for pointing it out), second is a message for future us to catch endpoint-related code to delete when we fully remove endpoints.

dojo/tools/wapiti/parser.py
dogboat: Same as above -- first is done (removed now), second is for future us.

unittests/test_importers_performance.py
dogboat: I skipped this test for Locations, it's not implemented for those; part of the reason being that I didn't apply the performance changes you'd been working on to locations, which I imagine we'll do here soon. At that point I would add test coverage for locations performance, so this is still a TODO. If you'd like me to, I could add the test now, but the numbers might make everybody sad hahaha.

dojo/finding/deduplication.py
dogboat: Nice catch, fixed!

dogboat:
> The old code has the problem that the Endpoint table doesn't have a unique constraint. This leads to kind of complicated and sometimes expensive logic just to create endpoints during imports. There are situations in which duplicate can or could be created. There's also a warning in the code to log if duplicates are found. In the new code this warning seems to have been removed.
> 
> I would say that now is the time to fix the URL model to have a uniqueconstraint. I believe concatting all the fields would not work because they all have a high max length. I think a hash of all fields in a new "url_hash" field could work to make everything around inserts straightforward again. Updates might still be problematic if someone causes a URL to become identical to another URL. But I belivee this problem is already present right now. The update problem is also something that could be fixed later. The insert problem should be done right now as the migration also provides a way to cleanup/deduplicate.

I've added this as you've described: URLs (well, their hashes) are unique now. For updates I just throw a "hey this exists" message if there's a collision.

dojo/importers/location_manager.py
dogboat: This should be fixed now that we have unique URLs.

dogboat:
>     * The `settings.dist.py` still contains only settings for `endpoints` (hashcode/equality related). Should these be renamed/aliased for their `Location` counterparts? And/or docs/comments updated?>

We wanted to keep the same settings names defined so users don't have to tweak those at all when migrating (compatibility). Behind the scenes, it uses the new locations stuff instead when appropriate.

It would certainly be nice to change all the terminology fully, and maybe we could somehow do it ("if v3 and hash_field_name in {'endpoints', 'locations'} do x" with warnings about the old one as we go forward?), but for now this is simpler.
 
>     * dojo/jira/helper.py contains some code that uses/update endpoints which will no longer work when enabling locations
> 
>     * dojo/tasks.py displays endpoint count which will no longer be accurate when enabling locations
> 
>     * view_test.html, view_finding.html reference test_import.import_settings.endpoint instead of test_import.import_settings.locations
> 
>     * display_tags.py reference import_settings.get("endpoints", get("endpoint")) but should reference locations

Argh, sorry! Thanks! Fixed!

valentijnscholten:
> We wanted to keep the same settings names defined so users don't have to tweak those at all when migrating (compatibility). Behind the scenes, it uses the new locations stuff instead when appropriate.
> 
> It would certainly be nice to change all the terminology fully, and maybe we could somehow do it ("if v3 and hash_field_name in {'endpoints', 'locations'} do x" with warnings about the old one as we go forward?), but for now this is simpler.
Maybe add some (words to the)comment to make it clear it also applies to Locations (or URLs)?

unittests/test_importers_performance.py
valentijnscholten: If you add the bad numbers now you can feel good when you forward port the performance improvements and check the numbers again :-D 
If you want me to extend the performance test I can do that.

dogboat:
> > Maybe add some (words to the)comment to make it clear it also applies to Locations (or URLs)?

Good idea, done!

Maffooch (approved):

rossops (approved):

PR #13562, opened 2025-10-29
Add OpenReports import support

mfyll:
### Add OpenReports JSON Import Parser
**Description**

Adds support for importing scans in [OpenReports.io](http://openreports.io/) format. Currently only [statnett operator](https://github.com/statnett/image-scanner-operator) outputs trivy scans in this format, but being an open standard backed by e.g kyverno, more tools could adopt it.

More information can be found here: [OpenReports.io](https://openreports.io/)

Steps to get the relevant file:

```bash
kubectl get reports -ojson -A > reports.json
```

**Features:**

- Parses single reports and Kubernetes List objects
- Extracts CVE IDs, component versions, and fix availability
- Maps severity levels and creates proper service identifiers
- Handles vulnerability metadata and tagging

**Test results**

Added unit tests covering empty results, single reports, list format, and parser metadata. All tests pass with sample JSON files included.

I have also tested this by importing the resulting JSON file in the DefectDojo GUI and this works. 

**Documentation**

Parser follows standard DefectDojo interface with inline documentation and sample files.

Checklist

- [x]  Rebased against latest dev
- [x]  Submitted against dev branch
- [x]  Flake8 and Python 3.12 compliant
- [x]  Unit tests added

PR #14206, opened 2026-01-29
remove duplicated openreports parser doc

fopina:
**Description**

https://github.com/DefectDojo/django-DefectDojo/pull/13528/ moved parser docs and https://github.com/DefectDojo/django-DefectDojo/pull/13562 was created in between

It seems during conflict resolution it ended up with .md doc in both old and new location

This PR just removes the old one as it is not used at all


fopina:
@mfyll content in both old and new is different - the new one is being used, should it be the other one?

Maffooch (approved):

PR #14240, opened 2026-02-04
refactor dojo async task base task (bugfix branch)

valentijnscholten:
## Summary

This PR ports the async task refactor changes from #13986 to the bugfix branch.

The custom decorators that we have on Celery tasks interfere with some (advanced) celery functionality like signatures. This PR refactors this to have a clean base task that passes on context, but does not interfere with celery mechanisms.

The logic to decide whether or not the task is to be called asynchronously is now in a dispatch method.

### Key Changes:
- Removes `@dojo_async_task` decorator pattern
- Introduces `dojo_dispatch_task` function for dispatching Celery tasks
- Creates clean base task (`PgHistoryTask`) that preserves pghistory context
- Updates all async task callers to use the new dispatch pattern

Pro needs similar changes via DefectDojo-Inc/dojo-pro#2124

### DIFF
--- a/dojo/finding_group/views.py
+++ b/dojo/finding_group/views.py
@@ -16,7 +16,6 @@ import dojo.jira_link.helper as jira_helper
 from dojo.authorization.authorization import user_has_permission_or_403
 from dojo.authorization.authorization_decorators import user_is_authorized
 from dojo.authorization.roles_permissions import Permissions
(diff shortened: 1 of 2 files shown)


mtesauro (approved):
Approved

Jino-T (approved):

PR #14256, opened 2026-02-05
Add test to ensure duplicate findings are deleted in the proper order

Jino-T:
Add test to ensure that when a maximum duplicate value is surpassed, the oldest finding (based on date attribute on finding) is deleted. We want to avoid duplicate findings being deleted in order of finding id.

[sc-12742]

### DIFF
--- a/unittests/test_duplication_loops.py
+++ b/unittests/test_duplication_loops.py
@@ -403,7 +403,7 @@ class TestDuplicationLoops(DojoTestCase):
             Finding.objects.filter(id=self.finding_b.id).exists(),
             "The newest duplicate (Finding B) should still exist.",
         )
-        # Finding A should now only have 2 duplicate in its set
+        # Finding A should now only have 1 duplicate in its set
         self.finding_a.refresh_from_db()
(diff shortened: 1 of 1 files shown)


unittests/test_duplication_loops.py
```diff
@@ -374,6 +375,38 @@ def test_list_relations_for_three_reverse(self):
...
+        self.assertFalse(
+            Finding.objects.filter(id=self.finding_c.id).exists(),
+            "The oldest duplicate (Finding c) should have been deleted.",
+        )
+        # Finding b (2024) should exist
+        self.assertTrue(
+            Finding.objects.filter(id=self.finding_b.id).exists(),
+            "The newest duplicate (Finding B) should still exist.",
+        )
+        # Finding A should now only have 2 duplicate in its set
```
Maffooch:
```suggestion
        # Finding A should now only have 1 duplicate in its set
```

Maffooch (approved):

mtesauro (approved):
Approved

PR #14259, opened 2026-02-06
Auto Create Context: Fetch all objects for correct jira project associations

Maffooch:
When using auto create context, the previous implementation was only fetching products successfully if they already existed, but did not fetch an engagement or test. Because of this, jira settings on an engagement that were not inherited from a product were never considered when determining if the the import/reimport should be pushed to jira. 

Until this case occured, it appeared as if auto create context was functional because existing engagements/tests were being fetched at the serializer level after the auto create context process occurred at the viewset level

### DIFF
--- a/unittests/test_jira_import_and_pushing_api.py
+++ b/unittests/test_jira_import_and_pushing_api.py
@@ -1117,7 +1117,7 @@ class JIRAImportAndPushTestApi(DojoVCRAPITestCase):
     # this test and deduplication will prevent the second import from creating any findings at all,
     # which means no jira issues will be created on the second import, which is what we need to assert
     # that the jira project is being fetched correctly and push_to_jira is being set to True in the import settings
-    @toggle_system_setting_boolean("enable_deduplication", False)
+    @toggle_system_setting_boolean("enable_deduplication", False)  # noqa: FBT003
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

unittests/test_jira_import_and_pushing_api.py
```diff
@@ -1079,3 +1079,88 @@ def create_engagement_epic(self, engagement):
...
+        This test is responsible for ensuring that all related objects in auto context are fetched appropriately.
+        To test this, we will first set up a jira instance with a project configured at the engagement level only.
+        It is not really important that we test that findings are pushed to jira here, but we can assert that the
+        import history import settings reflect that the viewset was given a "True" value for push_to_jira,
+        which is only possible if the engagement's jira project was correctly fetched before the serializer was invoked.
+        """
+        import_reimport_config = self._test_setup_jira_project_for_engagement()
+        # Not run the import again
+        import1 = self.reimport_scan_with_params(
+            import_reimport_config.pop("test_id", None),
```
valentijnscholten, line 1155: i don't think there is a test_id for this scenario
```suggestion
            None
```

rossops (approved):

PR #14266, opened 2026-02-07
Expose `has_any_jira_issue` filter for findings

### LINKED ISSUE #12670: Findings with a group JIRA issue are returned in search for Findings with `has_jira == False`
The finding list pages in Defect Dojo have a filter to search for Findings that have or do not have a JIRA issue.

For findings that are in a group and that group has a JIRA issue, this filter currently doesn't work. The filter only checks if the finding itself has JIRA issue.

Question is should we:
1. Update the filter to match any JIRA issue, either for the Finding or the group that finding is in
2. Add an extra filter `has_group_issue` to filter for JIRA issues linked to the group the Finding is in
3. Both
4. Do nothing

http://localhost:8080/finding/open?has_jira_issue=false

Image

valentijnscholten:
## Summary
- The `has_jira_issue` filter on the finding list page only checks if the finding itself has a JIRA issue. For findings in a group where the group has a JIRA issue, this filter doesn't match.
- A `FindingHasJIRAFilter` class that checks both the finding and the finding group already existed (`has_any_jira`) but was never included in `get_finding_filterset_fields()`, so it was invisible in the UI and API.
- This PR renames it to `has_any_jira_issue`, adds descriptive `help_text`, and exposes it in the finding filter fields when both JIRA and finding groups are enabled.

Closes #12670

### DIFF
--- a/dojo/filters.py
+++ b/dojo/filters.py
@@ -518,6 +518,7 @@ def get_finding_filterset_fields(*, metrics=False, similar=False, filter_string_
         if get_system_setting("enable_jira"):
             fields.extend([
                 "has_jira_group_issue",
+                "has_any_jira_issue",
             ])
 
     return fields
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

Jino-T (approved):

PR #14269, opened 2026-02-08
Fix Qualys parser to prevent merging findings on different ports

ArnaavSinghSandhu:
Description
Fix: Qualys Infrastructure Scan merging findings on different ports.

Currently, the Qualys parser collapses multiple findings of the same QID into a single finding if they occur on the same host, even if they are on different ports (e.g., a vulnerability found on both port 80 and port 443). This results in data loss during the import process.

This PR implements the following:

Updated parse_finding in dojo/tools/qualys/parser.py to correctly extract the <PORT> tag from the Qualys XML.

Correctly maps these ports to the unsaved_endpoints and unsaved_locations attributes of the Finding object.

Ensures that findings remain distinct in DefectDojo by recognizing the unique port/endpoint combination.

Test results
I have verified this fix using the following steps:

New Unit Test: Added test_parse_file_with_multiple_ports_for_same_qid to unittests/tools/test_qualys_parser.py.

Sample Data: Created unittests/scans/qualys/test_qualys.xml which contains a single QID assigned to two different ports on the same host.

Execution: Ran the unit test suite inside the Docker environment.

Result: Ran 1 test in 0.009s. OK. The parser now correctly identifies 2 distinct findings instead of 1.
Screenshot 2026-02-08 at 10 15 38 PM

Documentation
No documentation changes are required as this is a bug fix for existing parser logic to bring it in line with expected DefectDojo behavior.

Checklist
[x] Bugfixes should be submitted against the bugfix branch.

[x] Give a meaningful name to your PR.

[x] Your code is flake8 compliant.

[x] Add applicable tests to the unit tests.

### DIFF
--- a/dojo/tools/qualys/parser.py
+++ b/dojo/tools/qualys/parser.py
@@ -78,7 +78,6 @@ class QualysParser:
         return [
             "title",
             "severity",
-            "endpoints",
         ]
 
     def get_scan_types(self):


dojo/tools/qualys/parser.py
```diff
@@ -78,6 +78,7 @@ def get_dedupe_fields(self) -> list[str]:
         return [
             "title",
             "severity",
+            "endpoints",
```
Maffooch:
```suggestion
```
The dedupelication fields are not being updated here (and they should not be) so this should be removed

ArnaavSinghSandhu:
Hi @Maffooch, thanks for the review. You're right—the branch has drifted too far and picked up unrelated commits during the rebase attempt. I'm going to close this PR now and open a fresh one against the dev branch with the requested change (removing 'endpoints' from dedupe fields) to keep the history clean

PR #14273, opened 2026-02-09
serialize the location id as id in url body

dogboat:
This PR updates AbstractedLocationSerializer to return the the ID of the Location object rather than the object itself (e.g., the URL).

URLs have a 1-1 with Location; the ID of the Location is used when accessing individual URLs via the API (`/url`) to keep consistency with the quick-lookup `/location` API endpoint. However, the body of the `/url` endpoint returns the ID of the URL object itself in the `.id` attribute, and the ID of the Location object in a `.location` field. This is a little weird (sorry!), since we never use the URL's ID anywhere else from a user's perspective and I'm sure it'll give someone an unnecessary headache down the road.

Thus, this PR updates the AbstractedLocationSerializer to return the Location's ID under the `.id` attribute; similarly, URLSerializer is updated to exclude the `.location` attribute as it's now redundant. At the same time, I've excluded the `.hash` field; it's used internally for uniqueness checks, there's no real reason to return it at this time.

### DIFF
--- a/dojo/location/api/serializers.py
+++ b/dojo/location/api/serializers.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from rest_framework.relations import PrimaryKeyRelatedField
 from rest_framework.serializers import CharField
 
 from dojo.api_helpers.serializers import BaseModelSerializer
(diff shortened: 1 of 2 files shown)


Jino-T (approved):

valentijnscholten:
@Maffooch @dogboat If this needs to be in 2.55.3 can you rebase against bugfix?

dogboat:
> @Maffooch @dogboat If this needs to be in 2.55.3 can you rebase against bugfix?

Done!

PR #14277, opened 2026-02-09
Groups: Support config permissions in API

Maffooch:
The API did not support users only having config permissions to view/add groups


Jino-T (approved):

mtesauro:
re-kicking GHA tests

PR #14285, opened 2026-02-10
Add deprecation notice for questionnaire API endpoints

Maffooch:
Introduce a deprecation notice for questionnaire-related APIendpoints, specifying an end-of-life date of June 1, 2026. This change aims to inform users about the upcoming removal of these endpoints.

[sc-12734]

### DIFF
--- a/docs/content/releases/os_upgrading/2.56.md
+++ b/docs/content/releases/os_upgrading/2.56.md
@@ -7,7 +7,7 @@ description: Deprecation of Questionnaire API Endpoints
 
 ## Deprecation: Questionnaire API Endpoints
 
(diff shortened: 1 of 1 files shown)


valentijnscholten (commented):
Is it just the endpoints that will be removed or the questionnaire feature itself? Possibly some notes/announcement in the upgrade notes is needed?

mtesauro:
> Is it just the endpoints that will be removed or the questionnaire feature itself? Possibly some notes/announcement in the upgrade notes is needed?

My understanding was the API endpoints would be replaced with better / more ergonomic ones and the feature itself would remain in DefectDojo.

@Maffooch Can keep me honest here.

Maffooch:
Yes, just the API endpoints are being removed. They are 95% read only, and are just causing confusion

I'll add breaking changes to 2.56.0 coming March, and another set for 2.59.0 coming in June

PR #14288, opened 2026-02-10
fix: update Qualys parser to support port-based findings

ArnaavSinghSandhu:
This PR replaces #14269 

Changes:

Rebased: The branch is now fully up to date with the latest dev branch.

Parser Update: Improved port extraction logic in the Qualys parser.

Deduplication: As requested by @Maffooch, I have removed endpoints and port from the get_dedupe_fields list to maintain the existing deduplication behavior while still improving the data extraction.

Formatting: Ran ruff to ensure all linting and formatting follow the project's style guide.

### DIFF
--- a/dojo/tools/qualys/parser.py
+++ b/dojo/tools/qualys/parser.py
@@ -353,13 +353,18 @@ def parse_finding(host, tree):
         if temp.get("CVSS_value") is not None:
             finding.cvssv3_score = temp.get("CVSS_value")
         finding.verified = True
+        endpoint_port = None
+        if port and str(port).isdigit():
+            endpoint_port = int(port)
         # manage endpoint/location
         if settings.V3_FEATURE_LOCATIONS:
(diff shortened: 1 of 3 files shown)


Maffooch:
@ArnaavSinghSandhu something I had not considered sooner is that what happens to existing qualys findings after this parser change? Would the endpoint from previous imports still continue to be matched, or would they be marked as mitigated, and new endpoints with ports be created?

ArnaavSinghSandhu:
That's a good point, @Maffooch. Since the endpoint string is changing from just the IP to IP:Port, DefectDojo will indeed treat them as new EndPoint objects.

On the next import:

The old Endpoints (IP only) will likely be marked as mitigated because they are no longer present in the scan file in that exact format.

The new Endpoints (IP:Port) will be created.

However, because we removed endpoints from the get_dedupe_fields (as per your request), the Findings themselves will not duplicate. They will simply point to the new, more accurate endpoints.

Maffooch:
This PR got sorta broken... If you would still this feature added, I ask that you sync your fork, and recreate the PR

PR #14309, opened 2026-02-12
Support sync kwarg in process_findings for inline post-processing

valentijnscholten:
## Summary
- Thread `sync` kwarg from `process_findings()` through to `dojo_dispatch_task()` in both `DefaultImporter` and `DefaultReImporter`, so callers can force `post_process_findings_batch` to run inline.
- Pop `sync` from kwargs in `sync_process_findings()` to avoid duplicate keyword argument errors.


mtesauro (approved):
Approved

PR #14343, opened 2026-02-18
fix for ms defender parser: use endpoint instead of url when not v3

dogboat:
Bug from the transition to V3. MSDefender parser changed to use a URL instead of Endpoint when V3 locations are disabled, which they should not.


Maffooch (approved):

mtesauro (approved):
Approved

valentijnscholten (approved):

Jino-T (approved):

Issue #14345, opened 2026-02-19
Depenency Track Parser doesn't set unique_id_from_tool


AndreVirtimo:
Findings from Dependency Track import will be recreated (old is closed, new one is created) when new vulnerability_ids are added by DT. This can not be fixed by setting deduplication algorithm to DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE because the parser sets the DT uuid to vuln_id_from_tool. But it has to be unique_id_from_tool.

See also discussion on Slack https://defectdojocommunity.slack.com/archives/C0A4DBKANDS/p1771402183329359

**Steps to reproduce**
1. Import Dependency Track finding with only a CVE vuln id
2. Reimport the same finding with an additional alias with a GHSA id

**Expected behavior**
The existing finding should get an update and stay open.

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [x] Kubernetes
- [ ] GoDojo

**Environment information**
 - DefectDojo version v. 2.55.1 

**Screenshots**

E-Mail:

<img width="661" height="480" alt="Image" src="https://github.com/user-attachments/assets/17b69353-d341-4bc2-bfb9-6894145e5533" />

New and old finding:

<img width="2430" height="994" alt="Image" src="https://github.com/user-attachments/assets/fe57e968-ea7f-4fd9-b220-04851fca8227" />

PR #14346, opened 2026-02-19
Dependency Track parser: Store DT uuid into unique_id_from_tool instead of vuln_id_from_tool

AndreVirtimo:
Store DT uuid into unique_id_from_tool instead of vuln_id_from_tool
change default deduplication algorithm to DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE

Fixing #14345

### DIFF
--- a/dojo/tools/dependency_track/parser.py
+++ b/dojo/tools/dependency_track/parser.py
@@ -198,6 +198,7 @@ class DependencyTrackParser:
             vulnerability_description += "\nVulnerability Description: {description}".format(description=dependency_track_finding["vulnerability"]["description"])
         if "uuid" in dependency_track_finding["vulnerability"] and dependency_track_finding["vulnerability"]["uuid"] is not None:
             unique_id_from_tool = dependency_track_finding["vulnerability"]["uuid"]
+            vuln_id_from_tool = dependency_track_finding["vulnerability"]["uuid"]
 
         # Get severity according to Dependency Track and convert it to a severity DefectDojo understands
         dependency_track_severity = dependency_track_finding["vulnerability"]["severity"]
@@ -230,6 +231,7 @@ class DependencyTrackParser:
             component_version=component_version,
             file_path=file_path,
             unique_id_from_tool=unique_id_from_tool,
+            vuln_id_from_tool=vuln_id_from_tool,
             static_finding=True,
             dynamic_finding=False)
 
--- a/unittests/tools/test_dependency_track_parser.py
+++ b/unittests/tools/test_dependency_track_parser.py
@@ -42,6 +42,7 @@ class TestDependencyTrackParser(DojoTestCase):
             self.assertEqual(1, len(findings[2].unsaved_vulnerability_ids))
             self.assertEqual("CVE-2016-2097", findings[2].unsaved_vulnerability_ids[0])
             self.assertEqual("900991f6-335a-49cb-9bf6-87b545f960ce", findings[2].unique_id_from_tool)
+            self.assertEqual("900991f6-335a-49cb-9bf6-87b545f960ce", findings[2].vuln_id_from_tool)
             self.assertTrue(findings[2].false_p)
             self.assertTrue(findings[2].is_mitigated)
             self.assertFalse(findings[2].active)
@@ -76,6 +77,7 @@ class TestDependencyTrackParser(DojoTestCase):
             self.assertEqual(12, len(findings))
             self.assertTrue(all(item.file_path is not None for item in findings))
             self.assertTrue(all(item.unique_id_from_tool is not None for item in findings))
+            self.assertTrue(all(item.vuln_id_from_tool is not None for item in findings))
             self.assertIn("CVE-2022-42004", findings[0].unsaved_vulnerability_ids)
             self.assertIn("DSA-5283-1", findings[0].unsaved_vulnerability_ids)
             self.assertIn("GHSA-rgv9-q543-rqg4", findings[0].unsaved_vulnerability_ids)
@@ -97,6 +99,7 @@ class TestDependencyTrackParser(DojoTestCase):
         self.assertEqual(12, len(findings))
         self.assertTrue(all(item.file_path is not None for item in findings))
         self.assertTrue(all(item.unique_id_from_tool is not None for item in findings))
+        self.assertTrue(all(item.vuln_id_from_tool is not None for item in findings))
         self.assertIn("CVE-2022-42004", findings[0].unsaved_vulnerability_ids)
         self.assertEqual(8.3, findings[0].cvssv3_score)
 


dojo/tools/dependency_track/parser.py
```diff
@@ -197,7 +197,7 @@ def _convert_dependency_track_finding_to_dojo_finding(self, dependency_track_fin
         if "description" in dependency_track_finding["vulnerability"] and dependency_track_finding["vulnerability"]["description"] is not None:
             vulnerability_description += "\nVulnerability Description: {description}".format(description=dependency_track_finding["vulnerability"]["description"])
         if "uuid" in dependency_track_finding["vulnerability"] and dependency_track_finding["vulnerability"]["uuid"] is not None:
-            vuln_id_from_tool = dependency_track_finding["vulnerability"]["uuid"]
+            unique_id_from_tool = dependency_track_finding["vulnerability"]["uuid"]
```
Maffooch, line 200:
```suggestion
            unique_id_from_tool = dependency_track_finding["vulnerability"]["uuid"]
            vuln_id_from_tool = dependency_track_finding["vulnerability"]["uuid"]
```

dojo/tools/dependency_track/parser.py
```diff
@@ -229,7 +229,7 @@ def _convert_dependency_track_finding_to_dojo_finding(self, dependency_track_fin
             component_name=component_name,
             component_version=component_version,
             file_path=file_path,
-            vuln_id_from_tool=vuln_id_from_tool,
+            unique_id_from_tool=unique_id_from_tool,
```
Maffooch, line 233:
```suggestion
            unique_id_from_tool=unique_id_from_tool,
            vuln_id_from_tool=vuln_id_from_tool,
```

unittests/tools/test_dependency_track_parser.py
```diff
@@ -41,6 +41,7 @@ def test_dependency_track_parser_has_many_findings(self):
             self.assertIsNone(findings[1].unsaved_vulnerability_ids)
             self.assertEqual(1, len(findings[2].unsaved_vulnerability_ids))
             self.assertEqual("CVE-2016-2097", findings[2].unsaved_vulnerability_ids[0])
+            self.assertEqual("900991f6-335a-49cb-9bf6-87b545f960ce", findings[2].unique_id_from_tool)
```
Maffooch, line 44:
```suggestion
            self.assertEqual("900991f6-335a-49cb-9bf6-87b545f960ce", findings[2].unique_id_from_tool)
            self.assertEqual("900991f6-335a-49cb-9bf6-87b545f960ce", findings[2].vuln_id_from_tool)
```

unittests/tools/test_dependency_track_parser.py
```diff
@@ -74,8 +75,10 @@ def test_dependency_track_parser_findings_with_alias(self):
 
             self.assertEqual(12, len(findings))
             self.assertTrue(all(item.file_path is not None for item in findings))
-            self.assertTrue(all(item.vuln_id_from_tool is not None for item in findings))
+            self.assertTrue(all(item.unique_id_from_tool is not None for item in findings))
```
Maffooch, line 79:
```suggestion
            self.assertTrue(all(item.unique_id_from_tool is not None for item in findings))
            self.assertTrue(all(item.vuln_id_from_tool is not None for item in findings))
```

unittests/tools/test_dependency_track_parser.py
```diff
@@ -93,7 +96,7 @@ def test_dependency_track_parser_findings_with_cvssV3_score(self):
             findings = parser.get_findings(testfile, Test())
         self.assertEqual(12, len(findings))
         self.assertTrue(all(item.file_path is not None for item in findings))
-        self.assertTrue(all(item.vuln_id_from_tool is not None for item in findings))
+        self.assertTrue(all(item.unique_id_from_tool is not None for item in findings))
```
Maffooch, line 101:
```suggestion
        self.assertTrue(all(item.unique_id_from_tool is not None for item in findings))
        self.assertTrue(all(item.vuln_id_from_tool is not None for item in findings))
```

Maffooch (changes requested):
This is a good change, but I think we should also maintain the value of `vuln_id_from_tool` to accommodate folks who have customized the dedupe settings in their `local_settings.py` files. If the `vuln_id_from_tool` field is suddenly empty, any existing DT findings would not be matched again

AndreVirtimo:
> This is a good change, but I think we should also maintain the value of `vuln_id_from_tool` to accommodate folks who have customized the dedupe settings in their `local_settings.py` files. If the `vuln_id_from_tool` field is suddenly empty, any existing DT findings would not be matched again

I don't agree. There is not deduplication algorithm which uses vuln_id_from_tool. The field description for vuln_id_from_tool is "Non-unique technical id from the source tool associated with the vulnerability type." which does not fit to the uuid from DT.

valentijnscholten:
> I don't agree. There is not deduplication algorithm which uses vuln_id_from_tool. The field description for vuln_id_from_tool is "Non-unique technical id from the source tool associated with the vulnerability type." which does not fit to the uuid from DT.

Users can customize the deduplication behaviour via the settings in settings.dist.py. If users have chosen `vuln_id_from_tool` in there, their hash_code calculations will change which might break their dedupe for newly imported reports.

In general I don't like `vuln_id_from_tool` AND `unique_id_from_tool` being populated with the same unique id, but maybe in this case it's better to do it to avoid a painful upgrade for users using `vuln_id_from_tool`. Alternatively we can do a "v2 style" approach as we did with for example OpenVAS: https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/tools/openvas/parser.py

valentijnscholten:
@AndreVirtimo Do you have aliases enabled in your DT instance? The example vulnerabilities you provided on Slack seem to be aliases of eachother. Usually DT exports this in the report that is being sent to Dojo and Dojo will import all aliases as vulnerability_ids. This way both vulnerabilities should result in the same hash_code and become duplicates and work OK in reimports.

AndreVirtimo:
> @AndreVirtimo Do you have aliases enabled in your DT instance? The example vulnerabilities you provided on Slack seem to be aliases of eachother. Usually DT exports this in the report that is being sent to Dojo and Dojo will import all aliases as vulnerability_ids. This way both vulnerabilities should result in the same hash_code and become duplicates and work OK in reimports.

We have NVD and GITHUB as sources. The alias option for GITHUB is active.

The finding from NVD was imported first, after 4 days the same finding came from GITHUB and the GHSA id was added to the list of vulnerability_ids. This also alters the hash.

AndreVirtimo:
> Users can customize the deduplication behaviour via the settings in settings.dist.py. If users have chosen `vuln_id_from_tool` in there, their hash_code calculations will change which might break their dedupe for newly imported reports.

If backward compatibility is more important than a strict usage of the fields, than we should keep `vuln_id_from_tool`. Maybe this could be changed with DD 3.

Maffooch:
With parsers, backward compatibility is paramount. From a data integrity perspective, a situation where new findings are created due to default hash code changes often will restarts SLAs. This is not acceptable for many orgs

valentijnscholten:
in that case having vulnerability_ids in the hash_code config is a risk for some parsers including dependency track. But that's unrelated to this PR.

AndreVirtimo:
Ok. I have accepted your suggestions to keep the old `vuln_id_from_tool`.

Jino-T (approved):

PR #14350, opened 2026-02-19
fix: dedupe management command FieldError with only("id") and select_related

valentijnscholten:
## Summary

- **Fix `FieldError` in async dedupe path**: `findings.only("id")` raised a `FieldError` when `select_related` traversal was still active on the queryset. Clearing both `select_related(None)` and `prefetch_related(None)` before calling `only("id")` resolves this, making the async dispatch path work regardless of how the queryset was originally constructed.
- **Fix `--dedupe_batch_mode` flag**: Changed from `action="store_true"` to `action=argparse.BooleanOptionalAction` so the flag can be explicitly disabled with `--no-dedupe_batch_mode` (previously it could only be enabled, not disabled at the CLI).


mtesauro (approved):
Approved

PR #14362, opened 2026-02-21
Fix PGDATA path to make postgres data durable (fixes #14358)

### LINKED ISSUE #14358: Database data is lost after 'docker compose down' due to PGDATA path mismatch
**Describe the bug**
Currently, in the default docker-compose.yml, the PostgreSQL volume is mounted to /var/lib/postgresql/data. However, the PostgreSQL image used in the project (or its specific version: postgres:18.1-alpine@sha256:4eb15de8e7b692c02427a2df278d18eb89422a534e428efb6d43c968250334d4) stores data in a different internal directory (e.g., /var/lib/postgresql/18/docker).
As a result, the persistent volume remains empty, while all database files are stored in the container's ephemeral writable layer. When docker compose down is executed, the container is removed, and all data is lost.

**To Reproduce**
1. Run docker compose up -d.
2. Complete the setup and add any data to DefectDojo.
3. Run docker compose down (without the -v flag).
4. Run docker compose up -d again.
5. All previously entered data is gone, and the database is re-initialized.

**Expected behavior**
Database records should persist between container restarts/removals because a volume is defined.

**Technical Details**
The output of SHOW data_directory; inside the postgres container shows:
data_directory: /var/lib/postgresql/18/docker (or similar non-standard path).
But the docker-compose.yml mounts the volume to:
/var/lib/postgresql/data.

Image

**Suggested Fix**
Explicitly set the PGDATA environment variable in the postgres service to match the volume mount point:

```
  postgres:
    environment:
      PGDATA: /var/lib/postgresql/data
    volumes:
      - defectdojo_postgres:/var/lib/postgresql/data
```

valentijnscholten:
## Summary

- Set PGDATA environment variable to /var/lib/postgresql/data in postgres service
- PostgreSQL 18 changed default data path to /var/lib/postgresql/18/docker, causing data to be stored in container ephemeral layer instead of the volume
- Fixes #14358

### DIFF
--- a/docs/content/releases/os_upgrading/2.51.md
+++ b/docs/content/releases/os_upgrading/2.51.md
@@ -77,7 +77,7 @@ If you already have a valid backup of the postgres 16 database, you can start at
 
 _Note: If you are using a bound volume, the path has changed for Postgres18. It is now `/var/lib/postgresql/` instead of `/var/lib/postgresql/data`. Failure to change the path may result in errors about failure to create a shim task. See the discussion in [docker-library/postgres](https://github.com/docker-library/postgres/issues/1370)._
 
-> **Warning: Data loss after `docker compose down` (fixed in 2.55.4)**
+> **Postgres Data storage**
 >
 > PostgreSQL 18 changed its default `PGDATA` path from `/var/lib/postgresql/data` to `/var/lib/postgresql/18/docker`. Because the Docker volume was mounted at `/var/lib/postgresql/data`, data was written to the container's ephemeral layer instead of the volume.
 >


mtesauro:
I saw that issue come in and wondered what changed.

Thanks for chasing that down @valentijnscholten

DarkR0ast:
@Maffooch 

The release note explains the PGDATA path change, but I don’t think it fully covers the upgrade case for users who already ran PostgreSQL 18 with the old compose file.

In that scenario, the volume may already contain a valid cluster under /var/lib/postgresql/data/18/docker, while the volume root contains 18/ and data -> .. After switching to `PGDATA=/var/lib/postgresql/data`, Postgres can fail with:

`initdb: error: directory "/var/lib/postgresql/data" exists but is not empty`

That's what happend for me.

So I think the docs should explicitly mention this migration case and advise users to either temporarily point PGDATA to the existing cluster path or migrate the cluster before switching, otherwise upgrades may fail even though data still exists.

valentijnscholten:
@DarkR0ast Can you create a PR with suggested improvements? I don't know what steps a user needs to take.

DarkR0ast:
@valentijnscholten I submitted a PR with an extended documentation. https://github.com/DefectDojo/django-DefectDojo/pull/14561

PR #14377, opened 2026-02-23
Fix webp issues with Dev deployments

paulOsinski:
`npm run dev` has been causing issues specifically surrounding webp conversion, leading to OOM errors.  This fix bypasses webp conversion for local Hugo builds of our docs and simply serves the PNG instead.

I've also shrunk quite a few image files while testing - sort of irrelevant but I figure it makes our repo a bit slimmer ¯\_(ツ)_/¯

I can drop that commit if it's not meaningful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,25 +40,19 @@ and reset every day. Do not put sensitive data in the demo. An easy way to test 
 ## Quick Start for Docker Compose
 
 ```sh
-# Clone the project
-git clone https://github.com/DefectDojo/django-DefectDojo
-cd django-DefectDojo
-
-# Check if your installed toolkit is compatible
-./docker/docker-compose-check.sh
-
-# Building Docker images
-docker compose build
-
-# Run the application
-# (see https://github.com/DefectDojo/django-DefectDojo/blob/dev/readme-docs/DOCKER.md for more info)
-docker compose up -d
-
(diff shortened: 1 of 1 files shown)


devGregA:
Won't this make our images lower res? Maybe a happy medium, but just wanted to note a potential negative side effect.

mtesauro:
> Won't this make our images lower res? Maybe a happy medium, but just wanted to note a potential negative side effect.

We could also change the process and handle conversion to webp as a one-time async process in it's own PR so it's not part of a normal build. Once they're converted, there's no reason not to just keep the webp in the repo going forward and loose the former png if webp ends up being a smaller but same resolution file.

paulOsinski:
Ok, I'm going to drop the image resizing commit because I'm realizing a few of these images are quite blurry after the batch job was run.

@mtesauro the only issue with converting to webp at the repo level is maintenance, because this webp conversion process generates many resolutions of the same file we could end up with a large amount of webps.  Hugo serves a specific webp depending on the page resolution, to handle mobile, big screens, etc.

Currently Hugo recreates all of these webps at build time, but we could adopt a caching approach so that we're not re-converting every time.  I haven't seen any issues in Prod yet, though, I only get this OOM when I try to run the build process on my laptop.

mtesauro (approved):
Approved

blakeaowens (approved):

PR #14391, opened 2026-02-25
Also update defect_dojo_sample_data_locations.json in sample data workflow

valentijnscholten:
## Summary

- The `update-sample-data` workflow previously only updated `defect_dojo_sample_data.json`
- Added the same update step for `defect_dojo_sample_data_locations.json`, which is used when the v3 feature locations flag is enabled


Maffooch (approved):

blakeaowens (approved):

fopina:
@valentijnscholten I didn’t have this one when doing #14336 

Does it also work for it?

valentijnscholten:
Yes

PR #14395, opened 2026-02-26
LocationData for parsers

dogboat:
This PR updates how parser/importer interaction works for locations in V3. Rather than having parsers return lists of concrete model types, they return LocationData objects, a dataclass that can be changed separate from models. This also allows us to start tracking other data that might be useful in the future that is not tied directly to a model class; this PR introduces dependency info gathering to that end. Parsers have been updated accordingly.

A few other changes are present to locations-related models. The `hash` field on URL has been renamed to `identity_hash` and some fields have been added to Location(Finding|Product)Reference models to allow us to track data pertinent to the association between a Location and Finding/Product.


Maffooch (approved):
Very clean implementation!

mtesauro (approved):
Approved

valentijnscholten (approved):

blakeaowens (approved):

PR #14396, opened 2026-02-26
fix: don't close old findings when reimport auto-creates a new test

### LINKED ISSUE #14363: First call to reimport closes findings outside test
As per: https://defectdojocommunity.slack.com/archives/C0A4DBKANDS/p1771564613056809

We have a lot of tests which are based on Dependency Check reports. All imports are done by Jenkins pipelines. Yesterday we set up a new pipeline and the first import of an empty DC report closed a bunch of findings a test which was not addressed by the import. The second run closed again some findings but this time it was a test which was not affected by the first import.
I can't see the import in the history of the wrong test, but I can see them in the reimport history on the affected findings within the tests.

This was the initial reimport call (the report has no findings).
```
curl --fail --location --request POST https://.../api/v2/reimport-scan/ --header Authorization: Token **** --form 
minimum_severity="Info" --form active="true" --form verified="true" --form scan_type="Dependency Check Scan" --form 
test_title="be-collaboration - Dependency Check Scan" --form product_type_name="Portal" --form product_name="BPC" --form engagement_name="support/4.2.x" --form auto_create_context="true" --form skip_duplicates="true" --form 
deduplication_on_engagement="true" --form close_old_findings="true" --form branch_tag="support/4.2.x" --form 
commit_hash="d2bcb5c8d3385df8ca5b3c3f6d652d07bd7c1489" --form build_id="1" --form 
group_by="component_name+component_version" --form file=@".../workspace/target/dependency-check-report.xml"
````

This issue sounds like #14353 but that's about an explicity initial Import which is expected behaviour.

valentijnscholten:
## Summary

- When `reimport-scan` is called with `auto_create_context=True` and the test does not yet exist, the endpoint falls back to `DefaultImporter` to create the test.
- If `close_old_findings=True` is set (the reimport default), `DefaultImporter.close_old_findings()` queries across the engagement/product scope for findings of the same test type that don't match the newly imported hash codes — and closes them. On an empty report this closes **all** matching findings in other tests.
- This goes against the docs and intended behaviour of the reimport endpoint.
- Fixed by passing `close_old_findings=False` to the importer in this path. A brand-new test has no prior findings to compare against, so closing old findings has no valid semantic here.
- Alternative fix by creating an empty test first is not immediately feasible as for dynamic parsers we don't know the `TestType` up front.

Fixes #14363.

### DIFF
--- a/unittests/test_import_reimport.py
+++ b/unittests/test_import_reimport.py
@@ -2513,6 +2513,63 @@ class ImportReimportTestAPI(DojoAPITestCase, ImportReimportMixin):
         date = findings["results"][0]["date"]
         self.assertEqual(date, "2006-12-26")
 
+    def test_reimport_auto_create_does_not_close_findings_in_existing_test(self):
+        """
+        Regression test for #14363: when reimport with auto_create_context=True creates
+        a brand new test, close_old_findings must not close findings from other tests in
+        the same engagement scope.
+
+        The serializer now forces close_old_findings=False when calling DefaultImporter
+        in this path. Without the fix, all 4 findings from the pre-existing test would be
+        incorrectly closed.
+        """
+        product_type, _ = Product_Type.objects.get_or_create(name="PT CloseOld AutoCreate")
+        product, _ = Product.objects.get_or_create(
+            name="P CloseOld AutoCreate",
+            description="test",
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

PR #14402, opened 2026-02-26
Reimport: Do not reactivate endpoint statuses with special statuses

Maffooch:
## Problem:
Endpoints previously marked as false positive, out of scope, or risk accepted could still be reactivated or associated with findings during reimport. This caused:
- Unintended endpoint reactivation
- Incorrect association with existing findings
- Inconsistent endpoint state handling

## Solution:
This update changes the reimport behavior to exclude endpoints with special statuses. Specifically:
- Endpoints marked as false positive, out of scope, or risk accepted will not be reactivated.
- Findings now only consider actionable endpoints.
- Endpoint lifecycle management is now more predictable and accurate.

## Behavior Changes:
- Reimport no longer affects endpoints with excluded statuses.
- Findings associations remain accurate and only include relevant endpoints.
- Reduces noise and prevents accidental state changes for intentionally excluded endpoints.

## Impact:
- Improves data integrity and accuracy in endpoint management.
- Ensures the system respects endpoint status semantics.
- Reduces confusion and potential errors during reimport operations.

### DIFF
--- a/dojo/finding/deduplication.py
+++ b/dojo/finding/deduplication.py
@@ -8,7 +8,7 @@ from django.db.models import Prefetch
 from django.db.models.query_utils import Q
 
 from dojo.celery import app
-from dojo.models import Finding, System_Settings
+from dojo.models import Endpoint_Status, Finding, System_Settings
 
 logger = logging.getLogger(__name__)
 deduplicationLogger = logging.getLogger("dojo.specific-loggers.deduplication")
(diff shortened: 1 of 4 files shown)


valentijnscholten:
We should find a way to do this without 40 extra database queries.

valentijnscholten:
> We should find a way to do this without 40 extra database queries.

Done, it's now 1 query less than `dev` instead of 40 more :-)

Maffooch:
@valentijnscholten you're my hero - thanks for taking this into a better direction!

blakeaowens (approved):

PR #14404, opened 2026-02-27
Run import performance test also for v3 Locations

valentijnscholten:
## Summary

### DIFF
--- a/dojo/base_models/base.py
+++ b/dojo/base_models/base.py
@@ -42,7 +42,7 @@ class BaseModelWithoutTimeMeta(Model):
 
         abstract = True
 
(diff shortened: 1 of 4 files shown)


valentijnscholten:
The tests pass locally, but not GHA. Maybe let's wait for the locations code to settle down before merging this?

valentijnscholten:
Fixed it by making the new BaseModel respect the settings at runtime. The new full_clean stuff causes 100s of extra queries...

valentijnscholten:
set to draft as the counts have changed due to changes made on `dev`.

valentijnscholten:
Superseded by #14500, which uses a branch directly in the upstream repo so CI workflows can be tested.

PR #14408, opened 2026-02-27
Refactor zip handling with safe_open_zip and safe_read_all_zip

Maffooch:
Consolidate zip file handling by introducing `safe_open_zip` and `safe_read_all_zip` functions. Update existing importers to utilize these new functions for safer zip file operations.


mtesauro:
> Could we use https://github.com/barseghyanartur/safezip instead?

FWIW, I kinda like not having YAD - yet another dependency

Maffooch:
> FWIW, I kinda like not having YAD - yet another dependency

Yeah agreed

valentijnscholten (approved):

PR #14411, opened 2026-02-27
fix the way bulk update endpoints in finding view works in v3

dogboat:
This PR updates the 'bulk update endpoints' functionality in the Finding view to work when V3/Locations are enabled. When using V3, the bulk edit options change to being radio buttons instead of checksbox since only one status can be set at a time in the new model.

### DIFF
--- a/dojo/templates/dojo/snippets/endpoints.html
+++ b/dojo/templates/dojo/snippets/endpoints.html
@@ -147,7 +147,7 @@
                 <div class="col-md-12">
                     <div class="panel panel-default table-responsive">
                         <div class="panel-heading">
-                            <h4>Vulnerable Endpoints / Systems ({{ finding.active_endpoint_count }}) </h4>
+                            <h4>Vulnerable Endpoints / Systems ({{ finding.active_endpoint_count }})
                                 <span class="pull-right"><a data-toggle="collapse" href="#vuln_endpoints"><i
                                         class="glyphicon glyphicon-chevron-up"></i></a></span>
                             </h4>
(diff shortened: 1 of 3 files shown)


dojo/templates/dojo/snippets/endpoints.html
```diff
@@ -147,7 +147,7 @@ <h6> Location </h6>
                 <div class="col-md-12">
                     <div class="panel panel-default table-responsive">
                         <div class="panel-heading">
-                            <h4>Vulnerable Endpoints / Systems ({{ finding.active_endpoint_count }}) </h4>
+                            <h4>Vulnerable Endpoints / Systems ({{ finding.active_endpoint_count }})
```
dogboat, line 150: Unrelated to the actual fix at hand but I noticed the extra hd close tag here was causing the chevron to render oddly.

dojo/templates/dojo/view_finding.html
```diff
@@ -758,48 +758,6 @@ <h4 class="has-filters">Similar Findings ({{ similar_findings.paginator.count }}
         <input type="hidden" name="return_url" value="{{ request.get_full_path }}" />
     </form>
 
-    <div class="hidden" style="padding-bottom: 5px;" id="bulk_edit_menu">
```
dogboat, line 761: This whole chunk was really just moved below so the bulk edit button appears closer to the endpoints lists.

valentijnscholten (approved):

mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

PR #14421, opened 2026-03-02
error message when viewing non-URL

dogboat:
This PR updates the URL ("endpoint") view to return a more helpful error rather than just crashing when viewing a non-URL.

### DIFF
--- a/dojo/url/ui/views.py
+++ b/dojo/url/ui/views.py
@@ -8,7 +8,7 @@ from django.contrib.admin.utils import NestedObjects
 from django.core.exceptions import PermissionDenied, ValidationError
 from django.core.management import call_command
 from django.db import DEFAULT_DB_ALIAS
-from django.http import HttpRequest, HttpResponseRedirect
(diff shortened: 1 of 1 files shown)


dojo/url/ui/views.py
```diff
@@ -100,6 +100,14 @@ def process_endpoint_view(request: HttpRequest, location_id: int, *, host_view=F
 
     """
     location = get_object_or_404(Location, id=location_id)
+    if location.location_type != URL.get_location_type():
+        messages.add_message(
+            request,
+            messages.ERROR,
+            "Viewing this object is only available in the Pro new UI.",
```
Maffooch:
```suggestion
            "Viewing this object is only available in the Pro UI.",
```

Maffooch (approved):

blakeaowens (approved):

valentijnscholten (approved):

Jino-T (approved):

PR #14434, opened 2026-03-02
minor: use django.conf.settings over dojo.settings everywhere

fopina:
**Description**

I was having issues when setting DJANGO_MODULE_SETTINGS to a custom module and noticed `dojo.settings` was imported directly in some places

* wsgi.py
  * I can't find anything using this file (only `dojo/wsgi.py`)
  * typical django project does not have it either - only `mainapp/wsgi.py` and it is currently here
  * deleted
* dojo/location/models.py - setting read from dojo.settings instead of django.conf.settings
  * replaced use with django.conf.settings
  * also moved to `app.ready()` to avoid triggering LazySettings evaluation at model import time
* widgets.py - also replace dojo.settings direct use


valentijnscholten:
Let's verify it doesn't break Pro.

valentijnscholten:
https://github.com/DefectDojo-Inc/dojo-pro/actions/runs/22684307629

fopinappb:
@valentijnscholten for those without access to pro, was that a good or a bad run? 😄

Maffooch (approved):

dojo/apps.py
```diff
@@ -16,6 +17,12 @@ class DojoAppConfig(AppConfig):
     verbose_name = "Defect Dojo"
 
     def ready(self):
+        if settings.ENABLE_AUDITLOG:
+            from auditlog.registry import auditlog  # noqa: PLC0415 raised: AppRegistryNotReady
+
+            from dojo.location.models import Location  # noqa: PLC0415 raised: AppRegistryNotReady
+
+            auditlog.register(Location)
```
dogboat: Question (for everybody) -- we're using pghistory now right? So we could actually just get rid of this entirely maybe? (By which I mean: leave location/models.py updated as it is, and just don't touch this file at all.)

dogboat (changes requested):

dojo/apps.py
dogboat: And yes, registering Location with auditlog is my fault. ;-)

valentijnscholten: I think it should go indeed

fopina: Should I just remove apps.py changes then, leaving the rest?

valentijnscholten: Yes

fopina: PR updated 👍

dogboat (approved):
Thank you!

valentijnscholten (approved):
Thank you @fopina and @fopinappb

valentijnscholten:
> @valentijnscholten for those without access to pro, was that a good or a bad run? 😄

It looks OK

blakeaowens (approved):

PR #14442, opened 2026-03-04
replace bleach with nh3 for HTML sanitization

valentijnscholten:
Bleach has been deprecated since Feb '23. Although it still works, a security product like Defect Dojo needs to switch to the goto replacement: `nh3` which is also faster.

## Summary

- Replaces the deprecated (archived) `bleach` library with `nh3`, its actively maintained, Rust-backed successor
- Drops `bleach[css]` from `requirements.txt` and adds `nh3`
- Updates all call sites in `dojo/utils.py`, `dojo/templatetags/display_tags.py`, and `dojo/templatetags/get_banner.py`
- Replaces `bleach.ALLOWED_TAGS` / `bleach.ALLOWED_ATTRIBUTES` with explicit constants (`_NH3_ALLOWED_TAGS`, `_NH3_ALLOWED_ATTRIBUTES`) shared between the two template tag modules

**Note on `style` attribute:** `bleach` supported CSS property-level filtering via `CSSSanitizer` (e.g. allowing only `color` and `font-weight`). `nh3` has no equivalent — it cannot filter individual CSS properties, only entire attributes. To avoid allowing arbitrary CSS injection, the `style` attribute is no longer permitted on any element. In practice this means inline styling (e.g. colored text in the login banner, background-color on images in markdown) will be stripped rather than sanitized. This is the safer trade-off.

**Note on disallowed tag handling:** `bleach` escaped disallowed tags (e.g. `<script>` → `&lt;script&gt;`), making them visible as literal text. `nh3` strips them entirely, keeping only the text content. This is the correct behavior for a sanitizer.


mtesauro (approved):
Approved

docs/content/en/open_source/upgrading/2.57.md
```diff
@@ -2,6 +2,13 @@
...
+Check the [Release Notes](https://github.com/DefectDojo/django-DefectDojo/releases/tag/2.57.0) for the contents of the release.
+
+## HTML sanitization: bleach replaced by nh3
+
+The `bleach` library has been replaced by [`nh3`](https://nh3.readthedocs.io/) for HTML sanitization. This is a drop-in replacement in most cases, but there are two minor behavioral changes to be aware of:
+
+- **`style` attributes are no longer allowed.** `bleach` supported CSS property-level filtering (e.g. allowing only `color` or `font-weight`). `nh3` has no equivalent, so `style` attributes are stripped entirely to avoid allowing arbitrary CSS injection. Content that previously relied on inline styles (e.g. colored text in the login banner, background-color on markdown images) will lose that styling.
+- **Disallowed tags are stripped rather than escaped.** Previously, a tag like `<script>` would be rendered as the literal text `&lt;script&gt;`. Now the tag is removed entirely and only its text content is kept. This is the correct behavior for a sanitizer.
```
Maffooch, line 14: Since this is a security tool, wont this render XSS findings stored in dojo totally incomplete? It wont be be clear what the original payload is if we strip out the tags entirely. It feels better to URL encode them for our specific use case

Maffooch (changes requested):
Just a question - I'm not dying on this hill 😄

PR #14449, opened 2026-03-05
perf(fp-history): batch false positive history processing

valentijnscholten:
## Summary

- **N+1 query fix**: replaced per-finding `match_finding_to_existing_findings()` calls with a single product-scoped batch query (`_fetch_fp_candidates_for_batch`) shared across all findings in the batch
- **Bulk update**: replaced all per-finding `save()` / `save_no_options()` calls in false positive history paths with `QuerySet.update()`, which bypasses Django signals identically to the previous calls (all update sites carry a comment explaining this)
- **`post_process_findings_batch`** (import/reimport): now calls `do_false_positive_history_batch()` instead of a per-finding loop — one DB query instead of N
- **`_bulk_update_finding_status_and_severity`** (bulk edit): findings grouped by `(product, dedup_alg)` and processed with a single batch call per group; retroactive reactivation also batched
- **Dead-code fix**: `process_false_positive_history` in the single-finding edit view had the condition `finding.false_p and not finding.false_p` (always `False`) because `form.save(commit=False)` with `instance=finding` mutates the object in place. Fixed by capturing `old_false_p = finding.false_p` before the form save and passing it as a keyword argument
- **Algorithm dispatch unified**: extracted `_fp_candidates_qs()` as the single source of truth for hash_code / unique_id_from_tool / unique_id_or_hash / legacy query building, shared by both `match_finding_to_existing_findings` (returns lazy QS for chaining) and `_fetch_fp_candidates_for_batch` (evaluates into a keyed dict)
- **Moved to `deduplication.py`**: all FP history helpers relocated from `dojo/utils.py` to `dojo/finding/deduplication.py` alongside the equivalent dedupe helpers; import sites in `helper.py`, `views.py`, and tests updated accordingly
- **4 new unit tests**: batch single-query behaviour, retroactive batch FP marking, retroactive reactivation (previously unreachable), and the no-reactivation guard
- **Query counts**: added some asserts on query counts to make sure we don't regress to N+1 in the future. Didn't go the full monty as with the import/reimport performance test as FP History is much less used.

Needs a Pro PR to cater for the moved/renamed methods.


mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

dogboat (approved):

PR #14457, opened 2026-03-06
show social provider label in groups

fopina:
**Description**

When groups are synchronized with a social provider (that supports it), group.social_provider is set.
There is no way to set it manually nor any other scenario that sets it.
There's no visual indiciation in the UI of a group that was created (and is maitained) by a social provider integration and a group that was created manually (exists only in Dojo / local).

This is missed as social groups should not be modified as (most) changes will be overwritten and it's good to be able to spot them besides the name prefix of the configuration.

**Test results**

* Logged in with a social provider to populate groups
* Logged in as admin and created a group (via UI)
* Check list of groups: verify provider groups have a label with its name and local group has no label
* Check details of group with and without provider: verify same as previous step

### DIFF
--- a/dojo/templates/dojo/groups.html
+++ b/dojo/templates/dojo/groups.html
@@ -83,7 +83,7 @@
                                         </li>
                                     </ul>
                                 </td>
-                                <td><a href="{% url 'view_group' g.id %}">{{ g.name }}</a></td>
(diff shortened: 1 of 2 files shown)


valentijnscholten (approved):

fopina:
Btw I've added this as a label rather than a new column/form field as I guess it's not very meaningful for most people (as it wasn't asked before), so it doesn't take any visual space if social groups are not used

Maffooch (approved):

mtesauro (approved):
Approved

dogboat (approved):

PR #14458, opened 2026-03-06
perf: batch duplicate marking in batch deduplication

valentijnscholten:
## Summary

Deduplication matching happens in batches. But the "marking" of findings as duplicates still happened one by one. For duplicate heavy instances this is a major slow down and resource hog. This PR batches the "marking" of duplicates to reduce the complexity from O(2N) to O(1).

- `set_duplicate` now accepts a `save=False` keyword argument; when false, field changes are applied to the model instance in memory without hitting the database
- Added `_flush_duplicate_changes` helper that bulk-updates all modified duplicate findings in a single `Finding.objects.bulk_update` call — one round-trip regardless of batch size, no saves on original findings
- All four `_dedupe_batch_*` algorithm functions (`hash_code`, `unique_id`, `uid_or_hash`, `legacy`) now collect duplicate pairs and flush via `_flush_duplicate_changes` instead of saving per-pair
- Transitive duplicates (findings that previously pointed to `new_finding` as their original, now re-pointed to the true original) are captured through `set_duplicate`'s return value and included in the bulk update
- Added `Finding.DEDUPLICATION_FIELDS` constant (allowlist) used with `.only()` in `get_finding_models_for_deduplication` to skip loading unused columns for the findings being processed
- Added `Finding.DEDUPLICATION_DEFERRED_FIELDS` constant (denylist of large text columns) used with `.defer()` in `build_candidate_scope_queryset` to skip loading unused columns for the candidate pool
- Performance test query count updated: second import with duplicates drops from 236 → 183 queries (−53)


Maffooch (approved):

mtesauro (approved):
Approved

dojo/finding/deduplication.py
```diff
@@ -654,21 +692,39 @@ def get_matches_from_legacy_candidates(new_finding, candidates_by_title, candida
...
 
 
+def _flush_duplicate_changes(modified_new_findings):
+    """
+    Persist duplicate field changes collected during a batch deduplication run.
+
+    Bulk-updates all modified new findings in one round-trip instead of one
+    save() call per finding.  Uses bulk_update (no signals) which is consistent
+    with the original code that called super(Finding, ...).save(skip_validation=True),
+    bypassing Finding.save() in both cases.
```
dogboat, line 702: If I'm understanding this comment correctly, it implies is that the new code doesn't send signals (true, it does not), just as the old code does not. However, I believe pre/post save signals are sent by `Model#save_base()`, which is still called. At this point in time it might not really matter (maybe to Pro?); just something to be wary of going forward.

If the point is simply that all the `Finding#save()` logic is skipped, yes, ignore me. :o)

dogboat (approved):

dojo/finding/deduplication.py
valentijnscholten: This is a leftover from a mistake Claude made, I will remove/adjust the comment. There is a companion PR for Pro and your comment reminds me to triple check that we don't miss any signal. I don't think we will because all findings import will get their priority recalculated anyway at the end of the import.

blakeaowens (approved):

PR #14459, opened 2026-03-06
use tags.add() instead of tags.set() on reimport

paulOsinski:
This fixes an issue introduced in 2.54.2 by commit [9c4f51e70f](https://github.com/DefectDojo/django-DefectDojo/commit/9c4f51e70f) ("tags from parser: fix parsers, add tests and fallback"). That commit correctly fixed parsers (Trivy, SARIF, SonarQube, etc.) to use unsaved_tags instead of finding.tags directly.

But a side effect, the tags.set() line in finding_post_processing — which previously never fired for those parsers because unsaved_tags was always empty — now fires on every reimport, **replacing all tags on the matched existing finding**.

This causes problems with manually created Finding Tags being overwritten on reimport.  Instead I recommend calling finding.tags.add() which does not replace any tags on the Finding that may exist previously.  IMO is something that should be handled by PATCHing or bulk-editing the Findings, not reimporting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -149,3 +149,6 @@ docs/.devcontainer/Dockerfile
 docs/LICENSE
 docs/.hugo_build.lock
 .cursor-rules
+
+# claude etc
+MEMORY.md
--- a/dojo/importers/default_importer.py
+++ b/dojo/importers/default_importer.py
@@ -243,9 +243,9 @@ class DefaultImporter(BaseImporter, DefaultImporterOptions):
             # Parsers must use unsaved_tags to store tags, so we can clean them
             cleaned_tags = clean_tags(finding.unsaved_tags)
             if isinstance(cleaned_tags, list):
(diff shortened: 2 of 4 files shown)


valentijnscholten:
Not sure if that commit actually caused a regression as before that commit the tags were also overwritten during reimport: https://github.com/DefectDojo/django-DefectDojo/blob/a8d9140580e86453e5aa7da8c4c87ccbb1c3cfae/dojo/importers/default_reimporter.py#L634-L635

Or maybe this never happened because parser were setting the tags field directly themselves?

Good to have  a test case now which captures the expectations, even though there might be cases where users actually expect tags to be removed?

rossops (approved):

mtesauro (approved):
Approved

Maffooch (approved):

valentijnscholten (approved):

PR #14461, opened 2026-03-07
feat: add Remove from Finding bulk action on View Finding page

valentijnscholten:
## Summary

While working on #14660 I noticed the "delete" icon was missing from the endpoint bulk edit menu on the View Finding page.

- Add a **Remove from Finding** trash button next to the Bulk Edit dropdown on the View Finding page, allowing users to disassociate selected endpoints from a finding without deleting the endpoint objects themselves
- Non-V3 path (`endpoint_status_bulk_update`): when `remove_from_finding` is posted, delete the matching `Endpoint_Status` records for the selected endpoints
- V3/Locations path (`finding_location_bulk_update`): when `remove_from_finding` is posted, delete the matching `LocationFindingReference` records
- Both views already require `Finding_Edit` permission via `@user_is_authorized`; no additional permission check is needed
- The JS appends selected endpoint IDs and the `remove_from_finding` flag to the existing `bulk_change_form` before submitting, reusing the same form and URL as the bulk status update
- Add unit tests covering single removal, partial removal, multi-endpoint removal, `Endpoint_Status` cleanup, no-op without the flag, and redirect behaviour

image

### DIFF
--- a/dojo/templates/dojo/view_finding.html
+++ b/dojo/templates/dojo/view_finding.html
@@ -929,10 +929,8 @@
                 </ul>
             </div>
             <div class="btn-group mr-2" role="group">
-                <button type="button" class="btn btn-sm btn-danger" data-toggle="tooltip" data-placement="bottom" title="Remove selected endpoints from this finding" aria-label="Remove selected endpoints from this finding">
-                    <a class="white-color remove-from-finding" href="#">
-                        <i class="fa-solid fa-trash"></i>
-                    </a>
+                <button type="button" class="btn btn-sm btn-primary remove-from-finding" data-toggle="tooltip" data-placement="bottom" title="Remove selected endpoints from this finding" aria-label="Remove selected endpoints from this finding">
+                    <i class="fa-solid fa-trash"></i>
                 </button>
             </div>
         </div>
@@ -1475,7 +1473,7 @@
                 });
             });
 
-            $('a.remove-from-finding').on('click', function(e) {
+            $('button.remove-from-finding').on('click', function(e) {
                 e.preventDefault();
                 if (confirm('Remove selected endpoints from this finding?')) {
                     $('input[type=checkbox].select_one:checked').each(function(){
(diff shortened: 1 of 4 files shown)


dojo/templates/dojo/view_finding.html
```diff
@@ -928,6 +928,13 @@ <h4 class="has-filters">
                     </li>
                 </ul>
             </div>
+            <div class="btn-group mr-2" role="group">
+                <button type="button" class="btn btn-sm btn-danger" data-toggle="tooltip" data-placement="bottom" title="Remove selected endpoints from this finding" aria-label="Remove selected endpoints from this finding">
```
dogboat: This is what it looks like to me, on both Chrome and Firefox (MacOS):

Image

...which is weird, because you posted a screenshot that looks fantastic. Is it just me?

dojo/templates/dojo/view_finding.html
```diff
@@ -1468,6 +1475,18 @@ <h4>Credential
                 });
             });
 
+            $('a.remove-from-finding').on('click', function(e) {
```
dogboat: Any chance we could have the click handler be on the entire button and not just the anchor? As it is now, I have to click the anchor content, which is a bit smaller than the button itself. When I was testing I hit the button but not the icon, which didn't trigger the action.

dogboat (changes requested):
Handy! This works, I just had some questions about the way the UI renders. Could be just my system tho.

valentijnscholten:
The screenshot is old and now I have an even newer one where the trash button looks the same as in other places, red on a grey background. Click handler adjusted, can't believe I didn't do that myself as I hate buttons not being clickable.

image

From another place in dojo the trash icon:

image

dogboat (approved):
Cool, thank you!

Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

mtesauro:
Closing and re-opening to kick the tests again

PR #14462, opened 2026-03-07
feat: allow sorting endpoints by active findings count

### LINKED ISSUE #10540: Orderning by Active Findings?
Hey guys,

I started using DefectDojo recently. It has been a blast!
But I am trying to make a report on the "most" vulnerable hosts based on active findings they have.
Cant find any filter or grid who can make me order by active findings, only name or "product".

Something like this is possible?

image

valentijnscholten:
## Summary

- Add ability to sort the endpoint list and host list by active findings count in the UI (ascending and descending)
- Annotate the endpoint queryset with `active_finding_count` before filtering so the sort operates at the database level
- Add `active_finding_count` to `EndpointFilterHelper` and `ApiEndpointFilter` ordering fields, and expose it in `EndpointSerializer`
- V3 Locations path: add `active_findings` ordering to `URLFilter` (UI) and `V3EndpointCompatibleFilterSet` (API compat); expose `active_finding_count` in `V3EndpointCompatibleSerializer`; the annotation is already provided by `annotate_location_counts_and_status` and `annotate_host_contents`
- Update `endpoints.html` and `url/list.html` templates to render the column header as a sortable link and use the annotated count directly from the queryset
- Add Selenium integration tests to `endpoint_extended_test.py` covering ascending and descending sort for both endpoint list and host list views; tests automatically select the correct sort field name based on `DD_V3_FEATURE_LOCATIONS`

Closes #10540

### DIFF
--- a/dojo/templates/dojo/url/list.html
+++ b/dojo/templates/dojo/url/list.html
@@ -97,7 +97,7 @@
                                 <th>Endpoint</th>
                             {% endif %}
                             {% if not product_tab %}<th class="text-center" nowrap="nowrap">Active (Total) Products</th>{% endif %}
-                            <th class="text-center" nowrap="nowrap">Active (Total) Findings</th>
+                            <th class="text-center" nowrap="nowrap">{% dojo_sort request 'Active (Total) Findings' 'active_findings' %}</th>
                             <th class="text-center" nowrap="nowrap">Overall Status</th>
                         </tr>
                         {% for location in locations %}
--- a/dojo/templates/dojo/url/list.html
+++ b/dojo/templates/dojo/url/list.html
@@ -97,7 +97,7 @@
                                 <th>Endpoint</th>
                             {% endif %}
                             {% if not product_tab %}<th class="text-center" nowrap="nowrap">Active (Total) Products</th>{% endif %}
-                            <th class="text-center" nowrap="nowrap">{% dojo_sort request 'Active (Total) Findings' 'active_findings' %}</th>
+                            <th class="text-center" nowrap="nowrap">{% dojo_sort request 'Active (Total) Findings' 'active_findings' field='ordering' %}</th>
                             <th class="text-center" nowrap="nowrap">Overall Status</th>
                         </tr>
                         {% for location in locations %}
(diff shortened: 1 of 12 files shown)


mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

dojo/templates/dojo/url/list.html
```diff
@@ -97,7 +97,7 @@ <h3 class="has-filters">
                                 <th>Endpoint</th>
                             {% endif %}
                             {% if not product_tab %}<th class="text-center" nowrap="nowrap">Active (Total) Products</th>{% endif %}
-                            <th class="text-center" nowrap="nowrap">Active (Total) Findings</th>
+                            <th class="text-center" nowrap="nowrap">{% dojo_sort request 'Active (Total) Findings' 'active_findings' %}</th>
```
dogboat: This one didn't work for me; from what I can tell, while the table does request the correct URL per dojo_sort, the `URLFilter` used by the view that renders this snippet expects the `ordering` parameter instead of `o`. In other words, clicking the listing here doesn't reorder, but changing the queryparam to e.g. `ordering=-active_findings` does.

dogboat (changes requested):
Do we also want this for `dojo.url.api.filters.URLFilter`? It's the filterset used by the viewset (`dojo.url.api.views.URLViewSet`) that handles the `/api/v2/url/` endpoint when V3 Locations are enabled. (...Not to be confused with the URLFilter you've already updated.)

On the other hand it's not widely used yet so maybe we should wait to see what's actually needed.

Edit: re: the above, it might be better to add the field to the AbstractedLocationFilter and LocationFilter filtersets, but that's starting to veer off into something more than what this is attempting to solve; maybe ignoring the above is best for now. Sorry for the confusion.

dojo/templates/dojo/url/list.html
valentijnscholten: didn't get a notification about the edited comment, but i pushed some changes as the idea was that it would also work for Locations/URLS. But I don't a have working dataset with that locally. Can you review the latest changes?

dogboat (approved):
Awesome, thank you! And sorry for the confusion!

Issue #9735, opened 2024-03-13
Closed Findings metric uses finding open date instead of closed date.


lappsec:
**Bug description**
In the Metrics dashboard, closed findings are shown/grouped according to the date the finding was initial discovered, not the date that it was closed/mitigated. So if a finding was discovered 2023/07/01 and closed on 2024/03/13, the "Closed Findings" metrics (both line chart and bar chart) are empty when choosing a date range of say 90 days, even though the finding was closed within that range. If the date range is changed to include the discovered/created date, the finding is included in the "Closed Findings" metrics. 

This makes the Closed Findings metric inaccurate and mostly useless, as you can't see how many were closed within a given period unless the finding was created AND reported within the searched period. 

**Steps to reproduce**
1. Within a product, create a finding and set the Date to 2023/07/01
2. Close the finding.
3. In the product, go to Metrics. The default view won't show any closed findings in that period. 
4. Edit the filter and under Date, choose something like 90 days, which would definitely include the date (today) that the finding was closed. It will still show 0 closed findings. 
5. Edit the Date filter to include the creation date period (last year or any date). The closed finding metric will then be populated. 

**Expected behavior**
The "Closed Findings" metric should include findings that were closed within the given Date filter/timeframe, instead of being based on the finding creation date. 

**Deployment method** *(select with an `X`)*
- [X ] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: [e.g. Ubuntu 18.04]
 - DefectDojo version (see footer) or commit message: [use `git show -s --format="[%ci] %h: %s [%d]"`]

Issue #2187, opened 2020-04-13
Product metrics now showing closed (mitigated) findings


AlexanderTyutin:
During implementing continuous vulnerability scan I was using APIv2 option "close_old_findings". And here in the main dashboard I see 4 closed findings. If I click on "Closed in Last 7 Days" I can see products related to these closed findings.

![image](https://user-images.githubusercontent.com/49285004/79100921-33609c00-7d70-11ea-899c-b470c01a364e.png)

But if I go to products metrics then no one of 4 products has any closed findings there.

![image](https://user-images.githubusercontent.com/49285004/79101200-b4b82e80-7d70-11ea-8d47-3c16c2671ecf.png)

![image](https://user-images.githubusercontent.com/49285004/79101206-ba157900-7d70-11ea-8c0a-4f6d4b22c561.png)

![image](https://user-images.githubusercontent.com/49285004/79101221-c13c8700-7d70-11ea-95be-1a7bce5a04bf.png)

![image](https://user-images.githubusercontent.com/49285004/79101239-c6013b00-7d70-11ea-82cc-3c1ab1210641.png)

When I click to closed findings I see a list of them. And when I click on any finding I see the finding record and the product which is related to the finding:

![image](https://user-images.githubusercontent.com/49285004/79101281-e0d3af80-7d70-11ea-8d30-6ec96ee92012.png)

valentijnscholten:
Changed the title to make it better aligned with what the problem is, it's not only for auto closed findings.

valentijnscholten:
We need to do a re-definition of all statuses and update/check our codebase, this will happen in #2215.

PR #14464, opened 2026-03-08
bugfix(metrics): closed findings counter always shows 0 when no new findings are imported 

vvpoglazov:
The **Closed Findings** counter on the Product Metrics page shows `0` even when closed findings exist and are correctly displayed when clicking the link.

| Metrics page (counter) | Findings list (after clicking) |
|---|---|
| 0 Closed Findings | 1 finding shown |

Root cause: `finding_queries()` builds `end_date` from the latest finding discovery date (as midnight `00:00:00`) and uses it as the upper bound for `mitigated__range`. This is semantically incorrect
 - a finding can be closed at any time after discovery, including days or weeks after the last scan. Any finding closed after that midnight cutoff is silently excluded from the counter.

The bug "self-healed" only when a new finding with a later discovery date was imported - making it particularly hard to notice. In environments with stable finding profiles (re-imports without new findings), the counter stayed at `0` permanently.

Related historical issues: #9735, #2187.

Fix: replace `end_date` with `timezone.now()` as the upper bound so any finding closed up to the current moment is counted. Applied in both `dojo/product/views.py` and `dojo/metrics/utils.py`.

**Test results**

Added regression tests in `unittests/test_product_metrics_closed_count.py`:
- finding closed on the same day as discovery date (after midnight) is counted
- finding discovered in the past but closed recently is counted
- all findings closed at any time on the same day are counted

Updated `test_closed_findings_filtered_by_mitigated_date` in `unittests/test_metrics_queries.py` to reflect correct behavior: findings closed after the discovery date range but before `now()` should appear in closed metrics.

**Documentation**

**Checklist**

- [x] Rebased against latest `bugfix`
- [x] Bugfix submitted against `bugfix` branch
- [x] Code is ruff compliant
- [x] Code is Python 3.13 compliant
- [x] Unit tests added

**Unit tests**

image

### DIFF
--- a/dojo/metrics/utils.py
+++ b/dojo/metrics/utils.py
@@ -76,11 +76,12 @@ def finding_queries(
 
     # Filter by the date ranges supplied
     all_findings_within_date_range = all_authorized_findings.filter(date__range=[start_date, end_date])
-    # Get the list of closed findings filtered by mitigated date (not discovery date)
-    # This ensures findings closed within the date range are included even if discovered outside it
+    # Filter closed findings by mitigated date (not discovery date).
+    # Use timezone.now() as the upper bound so findings closed after the latest
+    # discovery date are not excluded from the counter.
     closed_filtered_findings = all_authorized_findings.filter(
         CLOSED_FINDINGS_QUERY,
-        mitigated__range=[start_date, end_date],
+        mitigated__range=[start_date, timezone.now()],
         mitigated__isnull=False,
     )
     accepted_filtered_findings = all_findings_within_date_range.filter(ACCEPTED_FINDINGS_QUERY)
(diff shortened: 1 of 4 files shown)


Maffooch (changes requested):
Would this change support a use case of trying to answer the question "how many findings were closed in Q4 2025" ?

vvpoglazov:
Hello @Maffooch, yes, I believe this fix supports that use case.

Users can filter by mitigated date using either the `Mitigated Date` dropdown or the `Mitigated Before` / `Mitigated After` fields. 

Before this fix, the closed findings counter showed incorrect results (often 0) regardless of which mitigated filter was applied, because the upper bound for `mitigated__range` was derived from the latest finding discovery date - not from the mitigated date itself.

With this fix, the counter correctly reflects the filtered result. I'll test this locally and attach screenshots.

vvpoglazov:
I've created findings closed at different points in time:

image

The **Mitigated Date** filter now correctly shows the closed findings counter: 

| Mitigated Date filter | Expected | Result |
|---|---|---|
| Any date | 9 | image |
| Past year | 8 | image |
| Past 90 days | 6 | image |
| Past 30 days | 4 | image |
| Past 7 days | 2 | image |

**Q: How many findings were closed in Q4 2025?**

Using **Mitigated After: 2025-10-01** and **Mitigated Before: 2025-12-31**:

image

Maffooch (approved):
Excellent testing! Thanks for the PR

valentijnscholten (approved):
Thanks!

mtesauro (approved):
Approved

blakeaowens (approved):

PR #14483, opened 2026-03-10
prefetching locations when building dedupe candidate scope queryset

dogboat:
This PR updates the `build_candidate_scope_queryset()` function to handle the prefetch list differently when V3 locations are enabled to prevent crashes when updating locations on a finding. Previously, changing locations when editing a finding would cause a 500 when V3 locations are enabled; now it does not.

### DIFF
--- a/unittests/test_deduplication_logic.py
+++ b/unittests/test_deduplication_logic.py
@@ -9,7 +9,7 @@ from django.conf import settings
 from django.core import serializers
 from django.utils import timezone
 
-from dojo.finding.deduplication import set_duplicate
+from dojo.finding.deduplication import build_candidate_scope_queryset, set_duplicate
 from dojo.importers.default_importer import DefaultImporter
 from dojo.models import (
     Development_Environment,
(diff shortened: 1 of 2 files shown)


valentijnscholten (commented):
The PR description seems to suggest there is a bug that is fixed by this PR. Should the PR add a testcase to prove the bug is fixed and we don't regress in the future?

dogboat:
> The PR description seems to suggest there is a bug that is fixed by this PR. Should the PR add a testcase to prove the bug is fixed and we don't regress in the future?

Good call, added a test for it.

Maffooch (approved):

mtesauro (approved):
Approved

unittests/test_deduplication_logic.py
```diff
@@ -1892,3 +1897,20 @@ def enable_dedupe(self, *, enable=True):
...
+
+    @skip_unless_v3
+    def test_build_candidate_scope_queryset_does_not_crash_when_locations_enabled_but_endpoint_exists(self):
+        # Create an Endpoint and associate it with a Finding
+        with Endpoint.allow_endpoint_init():
+            endpoint = Endpoint.objects.create(host="test-host.com")
+        test = Test.objects.get(id=3)
+        finding = Finding.objects.filter(test=test).first()
+        finding.endpoints.add(endpoint)
+        # This explodes if it tries to prefetch Endpoints when they are disabled
```
valentijnscholten:
```suggestion
        # This used to explode when it tried to prefetch Endpoints when they are disabled, now fixed
```

valentijnscholten (approved):

blakeaowens (approved):

PR #14485, opened 2026-03-10
remove libtiff install from Dockerfiles

dogboat:
This PR removes libtiff from the alpine/debian Dockerfiles. libtiff has some vulnerabilities; we don't allow tiff uploads anyway. Currently libtiff is installed as a dependency for pillow, but now [their documentation](https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html#tiff) suggests it's not necessary.

### DIFF
--- a/Dockerfile.django-alpine
+++ b/Dockerfile.django-alpine
@@ -45,7 +45,6 @@ RUN \
   apk add --no-cache \
     openjpeg \
     jpeg \
-    tiff \
     bind-tools \
     xmlsec \
     git \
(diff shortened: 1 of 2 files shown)


dogboat:
[sc-12881]

Maffooch (approved):

valentijnscholten (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #14478, opened 2026-03-09
fix: risk acceptance proof download throws 500

valentijnscholten:
## Summary

- Fix risk acceptance proof download returning a 500 error (fixes #14467)
- The file path on line 1583 of `dojo/engagement/views.py` was wrapped in quotes (`"risk_acceptance.path.name"`), making it a string literal instead of an attribute access — so Django tried to open a nonexistent file
- Removed the quotes so the actual uploaded file path is resolved
- Rewrote the risk acceptance selenium tests which were silently broken — they always passed but never actually created a risk acceptance:
  - Navigate to Ad Hoc Engagement (where the test finding lives) instead of the wrong engagement
  - Use `execute_script` to make the bootstrap-select hidden native `<select>` visible so Selenium can interact with it
  - Upload a proof file during risk acceptance creation
  - Added `test_download_risk_acceptance_proof` to verify no 500 error on download
  - Tightened assertions so failures are no longer masked by loose success checks

PR #14486, opened 2026-03-10
feat: run single integration tests from dev mode

valentijnscholten:
## Summary

Working on #14478 I realized I wanted a quick way to run integration tests without having to switch via `docker/setEnv.sh`. 
This PR adds an "optional" integration_tests service to the `dev` override setup which allows us to run a single integraiton tests without having to switch or even restart containers.

- Add `integration-tests` service with a Docker Compose profile (`integration`) to the dev override, allowing a single integration test to be run against the running dev stack without switching environments via `setEnv.sh`
- Add `run-integration-test-dev.sh` helper script for convenience: `./run-integration-test-dev.sh tests/finding_test.py`
- Ignore the harmless `Cross-Origin-Opener-Policy` Chrome console warning in integration tests (occurs over plain HTTP, has no security impact)
- Configure `logging.basicConfig` in the integration test base class so log output (e.g. skipped console errors) is actually visible
- Document the new workflow in `readme-docs/DOCKER.md`


Maffooch (approved):

mtesauro (approved):
Approved

dogboat (approved):
It works for me (MacOS/aarch64) with two small tweaks; noting here for anybody else that might find it helpful.
* Had to grant executable on the script
* Had to explicitly specify `platform: "linux/amd64"` on the `integration-tests` service. I think this is because chromium doesn't work on aarch64 but boots up anyway, then just dies with a cryptic error code.

Approving since it's only because of my stupid platform.

blakeaowens (approved):

PR #14488, opened 2026-03-10
Notifications: Clean up duplicate system notification entries

Maffooch:
There is a code path that allows for two system notifications to be written to the database. I imagine it is a race condition, so it is difficult to pin it down exactly. There are two mitigations in place now
- Add atomic context manager when creating the system notifications row
- Delete the newer version of the table in the event the more than one exists in the db


valentijnscholten (approved):

dogboat (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #14489, opened 2026-03-11
perf(importers): batch endpoint creation and status updates during import/reimport

valentijnscholten:
## Summary
The biggest performance hit during import/reimport is caused by endpoint handling. This is still done a per finding and per endpoint (+ endpoint status) basis resulting in many small DB queries.
This PR extends the endpoint manager to become stateful and keep track of new endpoints and new/changed endpoint statuses during an import. After every batch (of 1000 findings) it will persist the gathered data to the database, just in time for the deduplication batch to start in a celery task.

- Replace per-finding `endpoint_get_or_create()` calls (1-2 DB queries each) with a stateful `EndpointManager` that accumulates endpoints and statuses across all findings in a batch and flushes them in bulk at batch boundaries
- A scan with 1000 findings × 5 endpoints (200 unique) goes from ~6200 DB queries to ~3-5 queries
- `EndpointManager` now and holds internal accumulators: `_endpoints_to_create` (dict, deduplicates by normalized key), `_statuses_to_create`, `_statuses_to_mitigate`, `_statuses_to_reactivate`
- New public API: `record_endpoint()`, `record_status_for_create()`, `record_statuses_to_mitigate()`, `record_statuses_to_reactivate()`, `persist()`
- `update_endpoint_status()` now accumulates mitigate/reactivate lists instead of dispatching one Celery task per finding; all bulk updates fire in `persist()` at the batch boundary
- `_get_or_create_endpoints()` fetches all existing product endpoints in one query (using the `idx_ep_product_lower_host` index), bulk-creates missing ones inside `transaction.atomic()`, and returns a key→Endpoint map for `Endpoint_Status` creation
- Removed old methods that no longer have callers: `add_endpoints_to_unsaved_finding`, `chunk_endpoints_and_disperse`, `chunk_endpoints_and_reactivate`, `chunk_endpoints_and_mitigate`, `mitigate_endpoint_status`, `reactivate_endpoint_status`
- Added unit tests for `_make_endpoint_unique_tuple` normalization (protocol/host case, default port collapsing, None/empty handling)
- Updated performance test fixture to match new stateful manager interface
- Remove the `status_finding_non_special` optimization attribute as it was a risky optimization


valentijnscholten:
This needs an update with the changes that fix the bug already fixed by https://github.com/DefectDojo/django-DefectDojo/pull/14569

valentijnscholten:
> This needs an update with the changes that fix the bug already fixed by #14569

Done

Maffooch (approved):

blakeaowens (approved):

dogboat (approved):

Jino-T (approved):

PR #14494, opened 2026-03-11
api: load jira custom_fields as json

paulOsinski:
Currently when passing `custom_fields` to `/jira_product_configurations` those entries are stored in the database as an escaped block of text rather than JSON.

This method uses json.loads() to parse the data sent over and will raise a 400 if the json is invalid.


Maffooch (approved):

blakeaowens (approved):

mtesauro (approved):
Approved

dogboat (approved):

PR #14500, opened 2026-03-12
Locations V3: add import performance test

valentijnscholten:
## Summary

Supersedes #14404.


valentijnscholten:
Superseded by #14501 which targets the bugfix branch.

PR #14501, opened 2026-03-12
Locations V3: add import performance test and autocorrect counts

valentijnscholten:
## Summary

Add import performance tests for when V3 LOCATIONS FEATURE is enabled. Same approach as the existing performance test, i.e. capturing/asserting query counts and celery task counts.

~~Additionally this PR introduces a new GitHub workflow that automatically corrects any counts that are wrong. ~~

~~This makes it easier for developers to keep the counts in sync with the codebase. The commit done by this workflow will NOT trigger a new workflow run, so not another 20minute wait time for the tests to complete.~~
~~If the counts are corrected a comment is placed on the PR to make the author (and reviewers) aware~~

The automation didn't play nice with github actions. The commit worked fine, but it would not trigger a new workflow run which left the PR without any checks and no way to trigger the checks again via the UI. Not even via close-reopen. The only solution would be to create PAT with write access and use that in the workflow. That would be a bit too much.
Experiment closed. The performance test still runs in its own workflow so we have clearer message/logging.

Supersedes #14404.


valentijnscholten:
Example log output:
image

Maffooch (approved):
This is super cool!

rossops (approved):

blakeaowens (approved):

mtesauro (approved):
Approved

Issue #14496, opened 2026-03-12
Fresh install via docker compose fails with django-linear-migrations error on dev env


manuel-sommer:
**Error**
docker compose build breaks up after:
```
Step 14/31 : RUN   --mount=from=build,src=/tmp/wheels,target=/tmp/wheels   --mount=from=build,src=/app/requirements.txt,target=/app/requirements.txt   pip3 install 	--no-cache-dir 	--no-index   --find-links=/tmp/wheels   -r ./requirements.txt
[+] build 0/1
 ⠙ Image defectdojo/defectdojo-django:latest Building                                                                       182.2s
the --mount option requires BuildKit. Refer to https://docs.docker.com/go/buildkit/ to learn how to build images with BuildKit enabled
```
docker compose up shows error:
```
initializer-1       | Traceback (most recent call last):
initializer-1       |   File "/app/manage.py", line 10, in <module>
initializer-1       |     execute_from_command_line(sys.argv)
initializer-1       |     ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
initializer-1       |   File "/usr/local/lib/python3.13/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
initializer-1       |     utility.execute()
initializer-1       |     ~~~~~~~~~~~~~~~^^
initializer-1       |   File "/usr/local/lib/python3.13/site-packages/django/core/management/__init__.py", line 416, in execute
initializer-1       |     django.setup()
initializer-1       |     ~~~~~~~~~~~~^^
initializer-1       |   File "/usr/local/lib/python3.13/site-packages/django/__init__.py", line 24, in setup
initializer-1       |     apps.populate(settings.INSTALLED_APPS)
initializer-1       |     ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
initializer-1       |   File "/usr/local/lib/python3.13/site-packages/django/apps/registry.py", line 91, in populate
initializer-1       |     app_config = AppConfig.create(entry)
initializer-1       |   File "/usr/local/lib/python3.13/site-packages/django/apps/config.py", line 193, in create
initializer-1       |     import_module(entry)
initializer-1       |     ~~~~~~~~~~~~~^^^^^^^
initializer-1       |   File "/usr/local/lib/python3.13/importlib/__init__.py", line 88, in import_module
initializer-1       |     return _bootstrap._gcd_import(name[level:], package, level)
[...]

PR #14503, opened 2026-03-12
docs: document BuildKit as a prerequisite for Docker Compose builds

valentijnscholten:
## Summary

- Adds BuildKit as an explicit prerequisite in `readme-docs/DOCKER.md`
- Notes that BuildKit is enabled by default since Docker 23.0, and documents the `DOCKER_BUILDKIT=1` workaround for older versions
- Fixes the root cause of the confusion reported in #14496, where a missing BuildKit error during build led to a confusing `ModuleNotFoundError` at runtime


Maffooch (approved):

mtesauro (approved):
Approved

I was honestly surprised that non-buildkit is still happening but I probably shouldn't have been.

rossops (approved):

blakeaowens (approved):

PR #14515, opened 2026-03-13
chore: normalize line endings to LF (CRLF -> LF)

valentijnscholten:
## Summary

Some files are violating the `.gitattributes` we recently set. This one-time PR normalizes this. After that git will protect us via .gitattributes.

- Normalize 134 files from CRLF to LF per .gitattributes
- Fixes phantom "modified" files when working across Windows/WSL
- No functional changes, line endings only


Maffooch (approved):

mtesauro (approved):
Approved

Gotta love that we can't even settle on a line ending across the OSes 🤷

mtesauro:
@valentijnscholten I suppose we can remove the "Migration" label from this since it's only changing the line endings rather than changing how any migration actually works. Correct?

valentijnscholten:
Removed all the labels.

rossops (approved):

blakeaowens (approved):

PR #14521, opened 2026-03-14
Add import history cleanup task and max_import_history setting fixes …

### LINKED ISSUE #13776: Clean up of import history
We have a job running every 5 minutes which creates a vulnerability report which in turn gets reimported into DefectDojo via `/reimport`. The findings get correctly deduplicated and deleted with the 'Delete duplications' option, but the import history seems to grow infinitely. 

The screenshots show the large number of imports of three tests. The size of these tables seems to grow significantly.

```
                     table_full_name                      |    size                                                                                                                                                                                                                                                                                                                                                                      
----------------------------------------------------------+------------
 public.watson_searchentry                                | 19 GB
 public.dojo_test_import_finding_action                   | 19 GB
```

<img width="201" height="175" alt="Image" src="https://github.com/user-attachments/assets/df9e6937-4972-4f4a-80f9-683f58ca7834" />

The tests in question have quite a significant number of findings, that also contributes to the size.

It should be able to limit the import history, similar to the number of duplicate findings. Or not store empty import history.

The TRACK_IMPORT_HISTORY option, however that option will maybe be removed and has other side effects: https://github.com/DefectDojo/django-DefectDojo/issues/13312.

**Additional context**
Using version `2.51.2`.
Slack thread: https://owasp.slack.com/archives/C2P5BA8MN/p1764060461585449.

From the slack conversation, to lower the size of the above tables (followed with a vacuum in case of postgres):
* `dojo_test_import_finding_action` can be safely deleted
* Run `./manage.py buildwatson` to rebuild `watson_searchentry`

tejas0077:
## Description
When running frequent reimports, the dojo_test_import_finding_action table 
grows infinitely causing significant database bloat (reported cases of 19GB+). 

This PR adds a configurable max_import_history setting similar to the existing 
max_dupes feature to automatically clean up old import history records.

Changes made:
- Added max_import_history field to System_Settings model
- Added DD_IMPORT_HISTORY_MAX_PER_OBJECT setting to settings.dist.py
- Added async_import_history_cleanup celery task in tasks.py
- Added migration 0262_system_settings_max_import_history.py

Closes #13776

## Test results
Manually verified the new field appears correctly in the System_Settings model. 
The cleanup task follows the same pattern as the existing async_dupe_delete task.

## Documentation

## Checklist
- Make sure to rebase your PR against the very latest dev.
- Features/Changes should be submitted against the dev branch.
- Model changes include the necessary migrations in dojo/db_migrations folder.
- Added the proper labels to categorize this PR.

### DIFF
--- a/dojo/tasks.py
+++ b/dojo/tasks.py
@@ -221,6 +221,7 @@ def _async_import_history_cleanup_impl():
         max_history = system_settings.max_import_history
         max_per_run = settings.IMPORT_HISTORY_MAX_PER_OBJECT
     except System_Settings.DoesNotExist:
+        logger.error('System_Settings not found, skipping import history cleanup')
         return
 
     if max_history is None:
--- a/dojo/tasks.py
+++ b/dojo/tasks.py
@@ -207,7 +207,44 @@ def jira_status_reconciliation_task(*args, **kwargs):
         return jira_status_reconciliation(*args, **kwargs)
 
 
-@app.task
+
+@app.task(bind=True)
+def async_import_history_cleanup(*args, **kwargs):
+    with pghistory.context(source="import_history_cleanup_task"):
+        _async_import_history_cleanup_impl()
+
+
+def _async_import_history_cleanup_impl():
+    """Delete oldest Test_Import records when a test exceeds max_import_history."""
+    try:
+        system_settings = System_Settings.objects.get()
+        max_history = system_settings.max_import_history
+        max_per_run = settings.IMPORT_HISTORY_MAX_PER_OBJECT
+    except System_Settings.DoesNotExist:
+        return
+
+    if max_history is None:
+        logger.info("skipping import history cleanup: max_import_history not configured")
+        return
+
+    logger.info("cleaning up import history (max per test: %s, max deletes per run: %s)", max_history, max_per_run)
+
+    tests_with_excess = Test_Import.objects \
+        .values("test") \
+        .annotate(import_count=Count("id")) \
+        .filter(import_count__gt=max_history)[:max_per_run]
+
+    total_deleted_count = 0
+    for entry in tests_with_excess:
+        test_id = entry["test"]
+        imports = Test_Import.objects.filter(test_id=test_id).order_by("created")
+        excess_count = entry["import_count"] - max_history
+        for test_import in imports[:excess_count]:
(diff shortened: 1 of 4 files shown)


dojo/tasks.py
```diff
@@ -207,7 +207,44 @@ def jira_status_reconciliation_task(*args, **kwargs):
...
+
+
+def _async_import_history_cleanup_impl():
+    """Delete oldest Test_Import records when a test exceeds max_import_history."""
+    try:
+        system_settings = System_Settings.objects.get()
+        max_history = system_settings.max_import_history
+        max_per_run = settings.IMPORT_HISTORY_MAX_PER_OBJECT
+    except System_Settings.DoesNotExist:
+        return
```
valentijnscholten, line 225: please log something or just let the exception bobble up

valentijnscholten (changes requested):
Did you test this as there is no trigger configured for this job to run?
Can you add a test case to make sure the right entries are deleted (oldest first, no more than needed, ...)

tejas0077:
Hi @valentijnscholten, thank you for the review!

I have addressed the first point by adding proper logging in the System_Settings.DoesNotExist exception handler.

Regarding the trigger for the job — you are correct that there is no periodic task configured yet. Should I add a Celery beat schedule similar to the existing async_dupe_delete task? Looking at the codebase, async_dupe_delete is triggered via the system settings UI. Should max_import_history follow the same pattern?

Regarding test cases — I will add unit tests to verify:
- Oldest entries are deleted first
- No more than the configured limit are deleted per run
- Entries within the limit are not deleted

Could you point me to an existing test for async_dupe_delete so I can follow the same pattern?

Maffooch (changes requested):
Import histories serve as an audit log for what occurred during an import/reimport. Rather than just dropped after a certain number of events, It would be better to make this cleanup time based (i.e. delete records older than one year)

tejas0077:
Thanks @Maffooch! That makes sense — treating import history as an audit log with time-based retention is a better approach. I'll rework the cleanup task to delete records older than a configurable number of days (defaulting to 365 days / 1 year) instead of using a count-based limit. Should the setting be called `import_history_retention_days` or do you have a preferred name for it?

Maffooch:
closing as stale

PR #14527, opened 2026-03-15
Fix missing username in uWSGI logs when using API Token authenticatio…

### LINKED ISSUE #13751: Missing username in log files when using API Token
**Slack us first!**
[Done!](https://owasp.slack.com/archives/C2P5BA8MN/p1763647860488289)

**Be informative**
Regarding the logs of the uwsgi container, it correctly shows the username of the user making the request on the web server, but doesn't for the API calls.

When using the web interface (to access web pages), the uwsgi container logs the following:

```text
uwsgi-1  | [pid: 41|app: -|req: -/-] 172.18.0.1 (USERNAME) {64 vars in 1258 bytes} [XXX] GET /alerts/count => generated 15 bytes in 57 msecs (HTTP/1.1 200) 7 headers in 212 bytes (1 switches on core 1)
```

However, when making a request to the API using an API token, the log is as follow:

```text
uwsgi-1  | [pid: 41|app: -|req: -/-] 172.18.0.1 (-) {44 vars in 681 bytes} [XXX] GET /api/v2/findings/3473444/ => generated 2403 bytes in 311 msecs (HTTP/1.1 200) 8 headers in 261 bytes (1 switches on core 2)
```

Notice the dash instead of the username of the user who generated the API token and performed the request.

This appears only when using the Token provided by the `/api/key-v2` page.
If the user uses the swagger UI to make the API request, then the username correctly appears.

After discussion on [Slack](https://owasp.slack.com/archives/C2P5BA8MN/p1763652353923489?thread_ts=1763647860.488289&cid=C2P5BA8MN):

> So issue is connected to AuthN method, not to url.

> This may need custom middleware to get loggeed: https://stackoverflow.com/questions/26240832/django-and-middleware-which-uses-request-user-is-always-anonymous

**Bug description**
A clear and concise description of what the bug is. For errors include at least the exact error message you are seeing (including traceback).

**Steps to reproduce**
1. Go to `/api/key-v2` page using web interface
2. Copy the given python code example (using Authorization Token in the headers)
3. Execute the script
4. Look at the uwsgi container logs and notice that the username doesn't appear in the logged request

**Expected behavior**
I expect the username to appear in the logs.

The logs are used for traceability and having no indication of the user that performed a request is a flaw that prevent forensic analysis or event simply traceability.

**Deployment method**
- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Operating System: Debian 12
 - Docker Compose: v2.39.4
 - DefectDojo version: tested on v2.51.3 and v2.52.1

**Logs** 
```text
uwsgi-1  | [pid: 41|app: -|req: -/-] 172.18.0.1 (USERNAME) {64 vars in 1258 bytes} [XXX] GET /alerts/count => generated 15 bytes in 57 msecs (HTTP/1.1 200) 7 headers in 212 bytes (1 switches on core 1)
uwsgi-1  | [pid: 41|app: -|req: -/-] 172.18.0.1 (-) {44 vars in 681 bytes} [XXX] GET /api/v2/findings/3473444/ => generated 2403 bytes in 311 msecs (HTTP/1.1 200) 8 headers in 261 bytes (1 switches on core 2)
```

**Sample scan files**
N/A

**Screenshots**
N/A

N/A

tejas0077:
## Description
When using API Token authentication, uWSGI logs show a dash (-) instead 
of the username, making it impossible to trace which user made which API 
request. This breaks audit trails and forensic analysis.

Web interface requests correctly log the username, but API token requests do not.

Fix: Added ApiTokenUsernameLoggingMiddleware that sets REMOTE_USER in the 
request metadata after authentication is complete, so uWSGI can log the 
correct username regardless of the authentication method used.

Fixes #13751

## Test results
Manually traced the middleware execution. The middleware runs after 
AuthenticationMiddleware so the user is always authenticated before 
we attempt to set REMOTE_USER.

## Documentation

### DIFF
--- a/dojo/middleware.py
+++ b/dojo/middleware.py
@@ -338,3 +338,22 @@ class AsyncSearchContextMiddleware(SearchContextMiddleware):
             for i, batch in enumerate(batches, 1):
                 logger.debug(f"AsyncSearchContextMiddleware: Triggering batch {i}/{len(batches)} for {model_name}: {len(batch)} instances")
                 update_watson_search_index_for_model(model_name, batch)
+
+
+class ApiTokenUsernameLoggingMiddleware:
+    """
+    Middleware to set REMOTE_USER in uWSGI logs when using API Token authentication.
+    When using API tokens, uWSGI logs show '-' instead of the username.
+    This middleware sets the REMOTE_USER environ variable so uWSGI can log it correctly.
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        # After the request is processed, the user is authenticated
+        if request.user and request.user.is_authenticated:
+            # Set REMOTE_USER so uWSGI logs the username correctly
+            request.META["REMOTE_USER"] = request.user.username
+        return response
--- a/dojo/middleware.py
+++ b/dojo/middleware.py
@@ -339,21 +339,3 @@ class AsyncSearchContextMiddleware(SearchContextMiddleware):
                 logger.debug(f"AsyncSearchContextMiddleware: Triggering batch {i}/{len(batches)} for {model_name}: {len(batch)} instances")
                 update_watson_search_index_for_model(model_name, batch)
 
-
-class ApiTokenUsernameLoggingMiddleware:
-    """
-    Middleware to set REMOTE_USER in uWSGI logs when using API Token authentication.
-    When using API tokens, uWSGI logs show '-' instead of the username.
-    This middleware sets the REMOTE_USER environ variable so uWSGI can log it correctly.
-    """
-
-    def __init__(self, get_response):
-        self.get_response = get_response
-
-    def __call__(self, request):
-        response = self.get_response(request)
-        # After the request is processed, the user is authenticated
-        if request.user and request.user.is_authenticated:
-            # Set REMOTE_USER so uWSGI logs the username correctly
-            request.META["REMOTE_USER"] = request.user.username
-        return response
--- a/dojo/middleware.py
+++ b/dojo/middleware.py
@@ -76,6 +76,7 @@ class LoginRequiredMiddleware:
                 uwsgi = __import__("uwsgi", globals(), locals(), ["set_logvar"], 0)
                 # this populates dd_user log var, so can appear in the uwsgi logs
                 uwsgi.set_logvar("dd_user", str(request.user))
+                request.META["REMOTE_USER"] = str(request.user)
         return response
 
 
--- a/dojo/middleware.py
+++ b/dojo/middleware.py
@@ -310,3 +310,25 @@ class AsyncSearchContextMiddleware(SearchContextMiddleware):
             for i, batch in enumerate(batches, 1):
                 logger.debug(f"AsyncSearchContextMiddleware: Triggering batch {i}/{len(batches)} for {model_name}: {len(batch)} instances")
                 dojo_dispatch_task(update_watson_search_index_for_model, model_name, batch)
+
+
+from rest_framework.authentication import TokenAuthentication as DRFTokenAuthentication
+
+
+class DojoTokenAuthentication(DRFTokenAuthentication):
+    """Custom token authentication that logs the username to uWSGI."""
+
+    def authenticate(self, request):
+        result = super().authenticate(request)
+
+        if result is not None:
+            user, _token = result
+            username = str(user)
+
+            request.META["REMOTE_USER"] = username
+
+            with suppress(ModuleNotFoundError):
+                uwsgi = __import__("uwsgi", globals(), locals(), ["set_logvar"], 0)
+                uwsgi.set_logvar("dd_user", username)
+
+        return result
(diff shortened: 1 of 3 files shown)


Maffooch (changes requested):
This is partially implemented here: https://github.com/DefectDojo/django-DefectDojo/blob/c89ac6087b553301e95214eb59d31a657641b8f6/dojo/remote_user.py

Please update this middleware instead of adding a new one

tejas0077:
Thanks @Maffooch! I'll update the existing `dojo/remote_user.py` middleware instead of adding a new one. Will push the changes shortly.

tejas0077:
Thanks @Maffooch! I've moved the logic into the existing `RemoteUserMiddleware` in `remote_user.py` using `process_response`, and removed the separate `ApiTokenUsernameLoggingMiddleware` from `middleware.py` and `settings.dist.py`. Please take a look!

valentijnscholten:
> This is partially implemented here: https://github.com/DefectDojo/django-DefectDojo/blob/c89ac6087b553301e95214eb59d31a657641b8f6/dojo/remote_user.py
> 
> Please update this middleware instead of adding a new one

@Maffooch I don't see how REMOTE_USER authentication is related to apikey authentication?

tejas0077:
Hi @valentijnscholten, you raise a valid point.  RemoteUserMiddleware only runs when `AUTH_REMOTEUSER_ENABLED=True`, so placing the logic there won't fix the issue for API token authentication.

Looking at the codebase again, I see there's already an existing middleware in `dojo/middleware.py` that sets `dd_user` via `uwsgi.set_logvar()`. Would it make more sense to extend that existing middleware to also set `REMOTE_USER` for all authenticated requests, regardless of the auth method used?

@Maffooch could you clarify where the right place would be to handle this for API token auth specifically?

Maffooch:
@valentijnscholten is right — `remote_user.py` is specifically for proxy-based REMOTE_USER authentication and only runs when `AUTH_REMOTEUSER_ENABLED=True`, so it is the wrong place for this.

@tejas0077 the right place is `LoginRequiredMiddleware` in `dojo/middleware.py`. It already handles setting `dd_user` via `uwsgi.set_logvar()` for all authenticated users (lines 73-78). The fix should add `request.META["REMOTE_USER"] = str(request.user)` right next to the existing `uwsgi.set_logvar` call, so uwsgi's `%u` format also picks up the username for API token auth requests.

Please revert the changes to `remote_user.py` and update `LoginRequiredMiddleware` instead.

valentijnscholten:
> The fix should add request.META["REMOTE_USER"] = str(request.user)

I'm not following. What has REMOTE_USER todo with tokens?

tejas0077:
Hi @Maffooch and @valentijnscholten, I've updated the fix based on your feedback. Reverted the changes to remote_user.py and instead added request.META["REMOTE_USER"] = str(request.user) in LoginRequiredMiddleware in dojo/middleware.py, right next to the existing uwsgi.set_logvar call. This ensures the username is logged correctly for API token auth requests without touching the remote user authentication flow.

valentijnscholten:
Discussed this with @Maffooch and the way I (or now we) see this is that we should extend the upstream `rest_framework.authentication.TokenAuthentication` middelware to log the username. i.e. add a new middleware to `middleware.py` and use that instead of the upstream middleware at https://github.com/DefectDojo/django-DefectDojo/blob/b6e0a2a7a973fe5f8694f212def8f8e1cf7b11da/dojo/settings/settings.dist.py#L912-L913
@tejas0077 Does that make sense to you and could you look at updating the PR?

tejas0077:
Fixed the Ruff linting issues. The PR is now ready for review.

Summary of changes:
1. Created DojoTokenAuthentication class that extends DRF's TokenAuthentication
2. Sets REMOTE_USER and dd_user after successful token authentication
3. Updated dojo/settings/settings.dist.py to use DojoTokenAuthentication instead of rest_framework.authentication.TokenAuthentication

This ensures API token requests log the correct username in uWSGI logs.

tejas0077:
> Discussed this with @Maffooch and the way I (or now we) see this is that we should extend the upstream `rest_framework.authentication.TokenAuthentication` middelware to log the username. i.e. add a new middleware to `middleware.py` and use that instead of the upstream middleware at
> 
> https://github.com/DefectDojo/django-DefectDojo/blob/b6e0a2a7a973fe5f8694f212def8f8e1cf7b11da/dojo/settings/settings.dist.py#L912-L913
> 
> 
> @tejas0077 Does that make sense to you and could you look at updating the PR?

@valentijnscholten @Maffooch Done. I've implemented the approach you discussed:

1. Extended rest_framework.authentication.TokenAuthentication with DojoTokenAuthentication in dojo/middleware.py
2. This custom class sets REMOTE_USER and dd_user after successful token authentication
3. Updated dojo/settings/settings.dist.py to use DojoTokenAuthentication instead of the upstream TokenAuthentication

The fix now logs the username correctly for API token requests in uWSGI logs, while keeping the change isolated to token authentication.

dojo/middleware.py
```diff
@@ -71,6 +72,7 @@ def __call__(self, request):
                 uwsgi = __import__("uwsgi", globals(), locals(), ["set_logvar"], 0)
                 # this populates dd_user log var, so can appear in the uwsgi logs
                 uwsgi.set_logvar("dd_user", str(request.user))
+                request.META["REMOTE_USER"] = str(request.user)
```
valentijnscholten, line 75: Not sure how to make it more clear but there should not be anything in this PR that does anything with `REMOTE_USER`.

valentijnscholten (changes requested):

Maffooch:
closing as stale

PR #14528, opened 2026-03-15
Fix Qualys parser collapsing findings with same QID but different por…

### LINKED ISSUE #13682: Collapses Qualys Scan Findings with Same QID but Different Ports and Endpoints
Be informative
When importing Qualys scan reports, multiple rows with the same QID and title but different ports and endpoints are merged by DefectDojo. This causes the overall count of vulnerabilities to be mismatched since distinct port and endpoint vulnerabilities get combined into one finding.

Problem description
DefectDojo collapses findings with the same QID and title regardless of differences in ports and endpoints from Qualys reports, leading to inaccurate vulnerability counts. We need a way to configure import/deduplication settings to keep these separate by port and endpoint.

Steps to reproduce

Import a Qualys scan report with multiple vulnerabilities that have the same QID but appear on different ports (e.g., ports 80, 5985, 9999) and endpoints.

Observe that DefectDojo collapses these into a single finding, losing port and endpoint-level granularity.

Check the count of findings and note that it is less than expected.

Expected behavior
Each finding with the same QID but unique ports and endpoints should be treated as separate vulnerabilities in DefectDojo, maintaining accurate counts and distinct entries.

Additional context (optional)
I understand DefectDojo has deduplication algorithms that merge findings for easier management, but in this case, the merging is too aggressive because both ports and endpoints are not considered uniquely. Guidance on configuring deduplication to include port and endpoint differentiation during import would be highly appreciated.

tejas0077:
Description
When importing Qualys scan reports, findings with the same QID but
different ports were being collapsed into a single finding, causing
inaccurate vulnerability counts and loss of port-level granularity.
Root cause: the finding title only used QID and vulnerability name,
so findings with the same QID on different ports (e.g. 80, 5985, 9999)
got deduplicated into one finding.
Fix: port is now added directly to the Endpoint (and LocationData for V3) when present. Each QID+port combination gets its own endpoint on the finding. Finding titles and deduplication are completely unchanged.
Fixes #13682
Test results
Manually traced the parser logic. Port is already extracted from the
XML as temp["port_status"] and is now passed to the Endpoint object when present.
Documentation
Checklist

Bugfix submitted against the bugfix branch.
Meaningful PR name given.
Proper label added.


valentijnscholten:
This will need some consideration/assurance as this will completely break deduplication with existing findings. Posted a comment on #13682

tejas0077:
Hi @valentijnscholten, thank you for the feedback!

You raise a valid point. Changing the title format will break 
deduplication with existing findings since the hash code is 
calculated from the title.

A safer approach would be to keep the title unchanged but instead 
use the port in the hash code calculation for Qualys specifically, 
by adding port to the HASHCODE_FIELDS_PER_SCANNER setting:

"Qualys Scan": ["title", "severity", "vulnerability_ids", "cwe", "port"]

This way:
- Existing finding titles remain unchanged
- Deduplication correctly separates findings by port
- No breaking change to existing data

Would you prefer this approach instead? I can update the PR accordingly.

Maffooch (changes requested):
Qualys is a super popular parser, and breaking deduplication would have far reaching impacts. Port is not a field that could be used for deduplication since port is on the endpoint model rather than the finding model.

Something that could be more palatable is to add the port to the endpoint, and then add multiple endpoints to the finding

tejas0077:
Thanks @Maffooch, that makes sense! I'll rework the fix to keep a single finding per QID but attach multiple endpoints with the respective ports. That way port-level granularity is preserved without touching titles or breaking deduplication. Will update the PR shortly.

tejas0077:
Hi @Maffooch, I've reworked the fix as suggested. Instead of modifying the title, the port is now added directly to the `Endpoint` (and `LocationData` for V3) when present. This way each QID+port combination gets its own endpoint on the finding, port-level granularity is preserved, and existing titles/deduplication are completely unchanged. Please take a look!

valentijnscholten:
The title is still being modified, I believe this has to be removed.

tejas0077:
Hi @valentijnscholten, thanks for catching that! I've removed the port from the finding title. The title is now back to the original format `QID-XXXX | Vulnerability Name` and the port is only added to the Endpoint. Please take a look!

Maffooch:
Please add some unit tests for this to prevent regressions, add some updates to the release notes, and then I think this should be good!

tejas0077:
Hi @Maffooch, I have addressed both your review requests:

Added unit tests -created a test XML file with the same QID (12345) on three different ports (80, 443, 8080) and added a test that verifies each port gets its own separate endpoint, the finding title remains unchanged as QID-12345 | Test Vulnerability, and all 3 ports are correctly captured.
Added release notes -created docs/content/releases/os_upgrading/2.57.md with a note about the Qualys parser fix referencing issue #13682.

Please take a look!

valentijnscholten:
@tejas0077 Can you rebase to get rid of the conflicts?

tejas0077:
Hi @valentijnscholten, I've rebased the branch onto the latest bugfix branch. Conflicts are resolved. Please take a look!

Maffooch:
closing as stale

PR #14549, opened 2026-03-18
Parse Twistlock packagePath so that we can record where the CVE is found

coheigea:
**Description**

Parse Twistlock packagePath so that we can record where the CVE is found

**Test results**

Test added.


valentijnscholten (approved):

mtesauro (approved):
Approved

Maffooch (approved):

Jino-T (approved):

PR #14551, opened 2026-03-18
Drop System_Settings "credentials" field

dogboat:
This PR removes the `credentials` field from the `System_Settings` model. This field was originally added as part of a feature that has since been removed, and is no longer used.

[sc-12905]


mtesauro (approved):
Approved

Maffooch (approved):

Jino-T (approved):

valentijnscholten (approved):
Approving. There will probably be a conflict with the other PR with a migration, but the Django Linear Migration library stuff should make it result in a PR conflict and force us (you) to rebase the migration :-)

PR #14553, opened 2026-03-19
Fix import-languages 500 errors and optimize DB performance

Maffooch:
## Summary

The `/api/v2/import-languages/` endpoint was producing HTTP 500 errors due to database integrity issues. This PR fixes the root causes and optimizes the endpoint's database access pattern.

### 1. Missing unique constraint on `Language_Type.language`
The `Language_Type` model's `language` field had no unique constraint, allowing duplicate records to accumulate over time. This caused:
- **`MultipleObjectsReturned` exceptions** when code assumed a single Language_Type per language name (the `import_github_languages` management command already had a workaround for this)
- **Unpredictable FK references** when building the language type lookup dict — if duplicates existed, the dict arbitrarily kept one, which could differ from what existing `Languages` records referenced
- **Race conditions** where concurrent requests could both create the same `Language_Type`, since there was no DB-level uniqueness enforcement

### 2. Non-atomic delete + insert on `Languages`
The existing code deleted all `Languages` records for a product, then bulk-created new ones. If the `bulk_create` failed (e.g., `IntegrityError` from the `unique_together` constraint on `(language, product)`), the product's language data was already gone — leaving the database in a broken state.

### 3. Generic `Exception` for parse errors
Invalid JSON input raised a bare `Exception("Invalid format")`, which Django REST Framework could not handle gracefully — resulting in a 500 instead of a 400 validation error.

### Migration (`0262_language_type_unique_language.py`)
A single migration that first deduplicates existing data, then adds the unique constraint:

**Step 1 — Data deduplication (RunPython):**
- Groups `Language_Type` records by `language` name and identifies duplicates
- Keeps the lowest-ID record as canonical for each duplicate group
- Reassigns all `Languages` FK references from duplicate `Language_Type` records to the canonical one
- Handles secondary `unique_together` violations on `Languages` that arise from FK reassignment (keeps lowest-ID, deletes rest)
- Deletes the now-unreferenced duplicate `Language_Type` records
- Logs a summary of all deduplication actions

**Step 2 — Schema change (AlterField):**
- Adds `unique=True` to `Language_Type.language`, which creates a unique **index** on the column
- This both prevents future duplicates and speeds up all lookups by language name (index seek vs. full table scan)

### Serializer Rewrite (`ImportLanguagesSerializer.save()`)

**Reliability fixes:**
| Before | After |
|--------|-------|
| `raise Exception("Invalid format")` → 500 | `raise serializers.ValidationError(...)` → 400 |
| No input deduplication | `dict.fromkeys()` deduplicates language names |
| `bulk_create()` without conflict handling — race condition | `bulk_create(ignore_conflicts=True)` — concurrent-safe |
| DELETE ALL + INSERT ALL not wrapped in transaction | `transaction.atomic()` wraps upsert + delete |
| No `IntegrityError` handling | Caught and re-raised as `serializers.ValidationError` |

**Performance optimizations:**
| Before | After |
|--------|-------|
| 3 Language_Type queries (filter existing → bulk_create new → re-filter new) | 2 queries (bulk_create with ignore_conflicts → single filter all) |
| DELETE ALL + INSERT ALL on Languages (high write amplification) | UPSERT via `bulk_create(update_conflicts=True)` + targeted DELETE of stale rows only |
| No index on `Language_Type.language` (full table scan) | Unique index enables index seeks |

The UPSERT approach means that for repeated imports of similar CLOC files, unchanged languages are updated in-place rather than deleted and re-inserted — significantly reducing write amplification and I/O.

### Management Command Cleanup (`import_github_languages.py`)
- Removed `MultipleObjectsReturned` exception handling — no longer possible with the unique constraint
- Fixed minor issue: `element.get("color", 0)` → `color` (variable was already extracted above)

### New Tests
- `test_create_with_invalid_json` — verifies invalid file content returns HTTP 400 (not 500)
- `test_create_idempotent` — verifies importing the same file twice succeeds with correct results

## Files Changed
- `dojo/models.py` — Added `unique=True` to `Language_Type.language`
- `dojo/api_v2/serializers.py` — Rewrote `ImportLanguagesSerializer.save()`, added `transaction` import
- `dojo/management/commands/import_github_languages.py` — Removed dead exception handling
- `dojo/db_migrations/0262_language_type_unique_language.py` — New migration (data dedup + schema change)
- `dojo/db_migrations/max_migration.txt` — Updated to latest migration
- `unittests/test_rest_framework.py` — Added 2 new test methods

## Test plan
- [ ] Existing `ImportLanguagesTest` tests pass (covers create and update flows)
- [ ] New `test_create_with_invalid_json` passes (400 response for bad input)
- [ ] New `test_create_idempotent` passes (double import produces correct state)
- [ ] Migration applies cleanly on a database with existing Language_Type duplicates
- [ ] Manual test: upload valid CLOC JSON via `/api/v2/import-languages/`
- [ ] Manual test: concurrent imports for different products do not interfere


mtesauro (approved):
Approved

valentijnscholten (approved):

Jino-T (approved):

valentijnscholten:
Nvm

dogboat (approved):

PR #14562, opened 2026-03-20
Fix deterministic ordering for async_dupe_delete when duplicate dates tie

valentijnscholten:
## Summary

Excess duplicates are deleted in chronological order, oldest are deleted, newest are retained. In cases where many duplicates had the same `date` the order of deletion was `undefined`. This PR as explicit fallback ordering to `id`.

- Order excess duplicate findings by `date`, then `id`, so deletion removes the oldest duplicates first when many rows share the same `date` (e.g. scanner-provided dates).
- Aligns `async_dupe_delete` with documented behavior: delete the oldest duplicate findings until the configured maximum is satisfied.

[sc-12922]

### DIFF
--- a/dojo/tasks.py
+++ b/dojo/tasks.py
@@ -138,8 +138,11 @@ def _async_dupe_delete_impl():
         originals_with_too_many_duplicates = Finding.objects.filter(id__in=originals_with_too_many_duplicates_ids).order_by("id")
 
         # prefetch to make it faster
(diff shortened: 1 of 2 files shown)


Maffooch (approved):

paulOsinski (approved):
Thank you!

mtesauro (approved):
Approved

Jino-T (approved):

PR #14564, opened 2026-03-21
Optimize prepare_duplicates_for_delete

valentijnscholten:
## Summary

While working on optimizing the (hard) delete of findings, I found this optimization that can/should be its own PR. 

- Replace per-original O(n×m) loop in `prepare_duplicates_for_delete()` with a single bulk UPDATE for inside-scope duplicate reset, significantly reducing query count for large deletion operations
- Add 11 unit tests covering all duplicate deletion scenarios (inside-scope reset, outside-scope reconfiguration, mixed clusters, cascade delete setting, status/found_by inheritance)
- Add WARN-level logging to `fix_loop_duplicates()` for production visibility into duplicate loop frequency
- Add explanatory comment on `removeLoop()` documenting optimization opportunity

### DIFF
--- a/dojo/finding/helper.py
+++ b/dojo/finding/helper.py
@@ -633,6 +633,8 @@ def prepare_duplicates_for_delete(test=None, engagement=None):
         logger.warning("nothing to prepare as test and engagement are None")
         return
 
+    # should not be needed in normal healthy instances.
+    # but in that case it's a cheap count query and we might as well run it to be safe
     fix_loop_duplicates()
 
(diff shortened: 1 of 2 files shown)


valentijnscholten:
This is not needed if we decide to merge https://github.com/DefectDojo/django-DefectDojo/pull/14566

Maffooch:
Feels like #14566 would be a better approach. Will move this PR to the next milestone in case others have different opinions

PR #14566, opened 2026-03-21
perf: replace per-object async delete with SQL cascade walker

valentijnscholten:
## Summary

Replaces the per-object chunked deletion strategy with a single-pass SQL cascade walker (`cascade_delete`), eliminating the need to fan out deletion work across multiple Celery workers. This auto detects relationships and respect the SET_NULL or DO_NOTHING or CASCADE_DELETE settings of the relationship.

The old delete code had many inefficienies, especially around duplicates. 
- When a product was deleted, the delete was split up over each engagement. A finding B in engagement B that was duplicate of finding A in engagement A would be recalculated to be a duplciate of finding C in engagement C. But then engagement C would be deleted, so finding B had to be recalculated again, etc, etc, 
- Django collector used for most things related to finding models.
- Every finding was handled by its own resulting in 1+N queries problems.
- Updates were done using `finding.save()` were this was not needed

**Key architectural changes:**

- **SQL cascade walker** replaces Django's ORM `Collector` for bulk deletion. Walks `_meta.related_objects` and issues bottom-up `DELETE` statements directly, bypassing per-object signals and Python-level overhead.
- **Findings deleted separately** from the parent object tree. M2M through tables and self-referential FKs (duplicate_finding) are handled explicitly before cascade, then the parent object's remaining children (Tests, Engagements, Endpoints) are cleaned up in a single `cascade_delete` call.
- **Duplicate cluster preparation** now scoped to the full deletion target (product/product_type) instead of per-engagement iteration, avoiding unnecessary reconfiguration of findings that are about to be deleted.
- **Outside-scope duplicates** deleted before main scope to prevent FK violations during chunked finding deletion.
- **`reconfigure_duplicate_cluster`** uses `.update()` instead of `Finding.save()`, avoiding Django signal storms (prioritization, dedup, audit) during deletion.
- A notification is still emitted for the top-level object that was deleted, same as the old async delete code.

### Performance (3 × 1373 findings, 3 engagements, many duplicates, 1 product, JFrog Xray Api Scan)

| Scenario | Duration | Celery workers |
|---|---|---|
| Old code | ~42s | 6 |
| New code | ~3.4s | 1 |

So it's 13 times quicker and uses 6 times less celeryworkers. 

### Performance (3 × 1050 findings, 3 engagements, many duplicates, 1 product, Acunetix 360 Scan)

| Scenario | Duration | Celery workers |
|---|---|---|
| Old code | ~50s | 8 |
| New code | ~3.5s | 1 |

So it's 14 times quicker and uses 8 times less celeryworkers. If you add Notes, Files, the gains become even larger.

This should have huge effect on instances that do a lot of deletions.

Inspired by: https://dev.to/redhap/efficient-django-delete-cascade-43i5

### DIFF
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -302,7 +302,7 @@ env = environ.FileAwareEnv(
     # Initial behaviour in Defect Dojo was to delete all duplicates when an original was deleted
     # New behaviour is to leave the duplicates in place, but set the oldest of duplicates as new original
     # Set to True to revert to the old behaviour where all duplicates are deleted
-    DD_DUPLICATE_CLUSTER_CASCADE_DELETE=(bool, False),
+    DD_DUPLICATE_CLUSTER_CASCADE_DELETE=(bool, True),
     # Enable Rate Limiting for the login page
     DD_RATE_LIMITER_ENABLED=(bool, False),
     # Examples include 5/m 100/h and more https://django-ratelimit.readthedocs.io/en/stable/rates.html#simple-rates
--- a/dojo/settings/settings.dist.py
+++ b/dojo/settings/settings.dist.py
@@ -302,7 +302,7 @@ env = environ.FileAwareEnv(
     # Initial behaviour in Defect Dojo was to delete all duplicates when an original was deleted
     # New behaviour is to leave the duplicates in place, but set the oldest of duplicates as new original
     # Set to True to revert to the old behaviour where all duplicates are deleted
-    DD_DUPLICATE_CLUSTER_CASCADE_DELETE=(str, False),
+    DD_DUPLICATE_CLUSTER_CASCADE_DELETE=(bool, False),
     # Enable Rate Limiting for the login page
     DD_RATE_LIMITER_ENABLED=(bool, False),
     # Examples include 5/m 100/h and more https://django-ratelimit.readthedocs.io/en/stable/rates.html#simple-rates
--- a/dojo/finding/helper.py
+++ b/dojo/finding/helper.py
@@ -616,7 +616,8 @@ def reconfigure_duplicate_cluster(original, cluster_outside):
 
 
 def prepare_duplicates_for_delete(obj):
-    """Prepare duplicate clusters before deleting a Test, Engagement, Product, or Product_Type.
+    """
+    Prepare duplicate clusters before deleting a Test, Engagement, Product, or Product_Type.
 
     Resets inside-scope duplicate FKs and reconfigures outside-scope clusters
     so that cascade_delete won't hit FK violations on the self-referential
@@ -664,7 +665,7 @@ def prepare_duplicates_for_delete(obj):
         .iterator(chunk_size=500)
     )
 
-    for chunk_ids in batched(originals_ids, 500):
+    for chunk_ids in batched(originals_ids, 500, strict=False):
         for original in Finding.objects.filter(id__in=chunk_ids).prefetch_related("original_finding"):
             # Inside-scope duplicates were already unlinked by the bulk UPDATE above,
             # so original_finding.all() now only contains outside-scope duplicates.
--- a/dojo/finding/helper.py
+++ b/dojo/finding/helper.py
@@ -31,6 +31,7 @@ from dojo.models import (
     Endpoint,
     Endpoint_Status,
     Engagement,
+    FileUpload,
     Finding,
     Finding_Group,
     JIRA_Instance,
@@ -696,6 +697,61 @@ def engagement_pre_delete(sender, instance, **kwargs):
 @receiver(post_delete, sender=Engagement)
 def engagement_post_delete(sender, instance, **kwargs):
     logger.debug("engagement post_delete, sender: %s instance: %s", to_str_typed(sender), to_str_typed(instance))
+
+
+def bulk_clear_finding_m2m(finding_qs):
+    """
+    Bulk-clear M2M through tables for a queryset of findings.
+
+    Must be called BEFORE cascade_delete since M2M through tables
+    are not discovered by _meta.related_objects.
+
+    Special handling for FileUpload: deletes via ORM so the custom
+    FileUpload.delete() fires and removes files from disk storage.
+    """
+    finding_ids = finding_qs.values_list("id", flat=True)
+
+    # Collect FileUpload IDs before deleting through table entries
+    file_ids = list(
+        Finding.files.through.objects.filter(
+            finding_id__in=finding_ids,
+        ).values_list("fileupload_id", flat=True),
+    )
+
+    # Collect Note IDs before deleting through table entries
+    note_ids = list(
+        Finding.notes.through.objects.filter(
+            finding_id__in=finding_ids,
+        ).values_list("notes_id", flat=True),
+    )
+
+    # Auto-discover and delete all M2M through tables
+    for m2m_field in Finding._meta.many_to_many:
+        through_model = m2m_field.remote_field.through
+        # Find the FK column that points to Finding
+        fk_column = None
+        for field in through_model._meta.get_fields():
+            if hasattr(field, "related_model") and field.related_model is Finding:
+                fk_column = field.column
+                break
+        if fk_column:
+            count, _ = through_model.objects.filter(
+                **{f"{fk_column}__in": finding_ids},
+            ).delete()
+            if count:
+                logger.debug(
+                    "bulk_clear_finding_m2m: deleted %d rows from %s",
+                    count, through_model._meta.db_table,
+                )
+
+    # Delete FileUpload objects via ORM so custom delete() removes files from disk
+    if file_ids:
+        for file_upload in FileUpload.objects.filter(id__in=file_ids).iterator():
+            file_upload.delete()
+
+    # Delete orphaned Notes
+    if note_ids:
+        Notes.objects.filter(id__in=note_ids).delete()
 
 
 def fix_loop_duplicates():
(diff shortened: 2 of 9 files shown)


valentijnscholten:
While working this PR it felt a little complicated.

But in a way it actually makes things simpler as we don't have to cascade deletes from products to engagements to tests, we don't have to do the chunking, don't have to worry about duplicate clusters across engagements interfering with each other, etc.

Maffooch (commented):
## Database Performance Review (4.5M findings, 17K products perspective)

Great PR — the architectural shift from per-object Celery fan-out to a single-pass SQL cascade walker is a significant improvement. The 13-14x speedup benchmarks are compelling. Below are concerns specifically about **worst-case behavior at scale** that could cause system outages on large instances.

---

### 1. CRITICAL — `bulk_delete_findings` materializes ALL finding IDs into memory before chunking

**`dojo/finding/helper.py` — `bulk_delete_findings()`**

```python
finding_ids = list(finding_qs.values_list("id", flat=True).order_by("id"))
```

A Product_Type with 4.5M findings → Python list of 4.5M integers. The underlying query (`SELECT id FROM dojo_finding WHERE ... ORDER BY id`) on 4.5M rows is a sequential scan + sort that holds a DB snapshot open during materialization.

**Suggestion:** Use `.iterator(chunk_size=...)` with `itertools.batched` (already imported and used in `prepare_duplicates_for_delete` for `originals_ids`) to stream IDs without materializing the full list.

---

### 2. CRITICAL — Step 5 cascade wraps the entire top-level object deletion in a single `transaction.atomic()`

**`dojo/utils.py` — `async_delete_task()`**

```python
with transaction.atomic():
    cascade_delete(type(obj), pk_query, skip_relations={Finding})
```

For a Product_Type with 17K products → millions of Endpoint/Endpoint_Status/Test/Engagement rows deleted in **one transaction**. This will:
- Hold row-level locks on every deleted row until COMMIT
- Generate WAL entries potentially in the gigabytes
- Block concurrent queries touching these rows (API requests, dedup workers, imports)
- If the Celery worker is OOM-killed mid-transaction, PostgreSQL rollback takes longer than the forward operation

**Suggestion:** Chunk Step 5 similarly to findings — e.g., `cascade_delete` per-Product (when deleting a Product_Type) or per-Engagement (when deleting a Product), each in its own `transaction.atomic()`.

---

### 3. HIGH — No `statement_timeout` or deadlock protection on raw SQL

**`dojo/utils_cascade_delete.py` — `execute_compiled_sql()`**

The old code had explicit deadlock detection with exponential backoff retry. The new code removes this entirely. Raw SQL DELETEs on large tables can deadlock with concurrent dedup workers, import tasks, or API updates — and will run indefinitely without a timeout.

**Suggestion:** Add `SET LOCAL statement_timeout = '300s'` (or configurable) at the start of each `transaction.atomic()` block. Consider at least logging deadlock errors if retry logic is intentionally omitted.

---

### 4. HIGH — `outside_dupes_qs` generates the scope subquery twice + `.exists()` then `.count()` is two queries

**`dojo/utils.py` — `async_delete_task()`, Step 3**

```python
outside_dupes_qs = (
    Finding.objects.filter(duplicate_finding_id__in=finding_qs.values_list("id", flat=True))
    .exclude(id__in=finding_qs.values_list("id", flat=True))
)
if outside_dupes_qs.exists():
    logger.info("... %d outside-scope duplicates ...", outside_dupes_qs.count())
```

Two issues:
1. `finding_qs.values_list(...)` appears twice — the planner may evaluate it twice
2. `.exists()` then `.count()` is two separate queries

**Suggestion:** Assign the subquery to a variable for reuse, and replace `.exists()` + `.count()` with just `.count()`:
```python
outside_count = outside_dupes_qs.count()
if outside_count:
    logger.info("... %d outside-scope duplicates ...", outside_count)
```

---

### 5. HIGH — `FileUpload` objects deleted one-by-one in a loop

**`dojo/finding/helper.py` — `bulk_clear_finding_m2m()`**

```python
for file_upload in FileUpload.objects.filter(id__in=file_ids).iterator():
    file_upload.delete()
```

O(N) individual DELETE queries. If a product has thousands of file uploads, this is a bottleneck.

**Suggestion:** Batch the DB deletes (e.g., `FileUpload.objects.filter(id__in=chunk).delete()`) and handle disk file cleanup separately, or at minimum document that per-object delete is required for the custom `FileUpload.delete()` disk cleanup.

---

### 6. MEDIUM — `cascade_delete` recursion builds nested subqueries 4-5 levels deep

**`dojo/utils_cascade_delete.py`**

Each recursion level nests the parent's PK query in another `Subquery`. For Product_Type → Product → Engagement → Test → Endpoint_Status, the DELETE at level 4 becomes 4 nested subselects. PostgreSQL generally handles this OK, but at 4.5M findings scale, hash joins may spill to disk.

**Suggestion:** Test with `EXPLAIN ANALYZE` on a representative dataset. Consider materializing intermediate ID sets if the planner produces suboptimal plans.

---

### 7. MEDIUM — `fix_loop_duplicates()` runs on EVERY delete — not a "cheap count query" at 4.5M findings

**`dojo/finding/helper.py` — `prepare_duplicates_for_delete()`**

```python
loop_qs = Finding.objects.filter(
    duplicate_finding__isnull=False,
    duplicate_finding__duplicate_finding__isnull=False
)
```

This is a self-join (Finding → Finding → Finding) counted across the entire table. On 4.5M findings this is expensive. The comment says "cheap count query" but a double self-join is not cheap at this scale. It also runs on every Test/Engagement delete via signals, not just top-level.

**Suggestion:** Gate behind a feature flag, move to a maintenance command, or at minimum only run it once at the top of `async_delete_task` rather than per-signal.

---

### 8. MEDIUM — `pre_bulk_delete_findings` signal passes un-bounded queryset

**`dojo/finding/helper.py` — `bulk_delete_findings()`**

Any signal receiver that materializes the `finding_qs` (e.g., for notifications or audit) could OOM on a 4.5M-finding queryset.

**Suggestion:** Document that receivers MUST NOT materialize the queryset, or consider sending the signal per-chunk.

---

### 9. LOW — `bulk_remove_all_tags` called twice for Findings

`bulk_clear_finding_m2m()` calls `bulk_remove_all_tags(Finding, ...)`, then `cascade_delete(Finding, ...)` calls it again. Second call is a no-op but still runs the aggregation query.

**Suggestion:** Add a `skip_m2m=True` parameter to `cascade_delete` or deduplicate the calls.

---

### 10. LOW — No progress visibility for long-running deletes

A 4.5M-finding delete could run 10+ minutes. If the Celery worker dies mid-operation, there's no way to know what was partially deleted.

**Suggestion:** Log the total expected scope at the start (finding count, engagement count, etc.) and consider a "deleting" status flag on the object.

---

Overall this is a solid improvement. The main risks at extreme scale are the single large transaction in Step 5 (#2) and full ID materialization (#1) — fixing those two would make this much safer for large deployments.

valentijnscholten:
Thanks for the thorough review @Maffooch! Addressed in 44caf6db2a:

**1 (CRITICAL) — ID materialization**: Replaced `list()` with `iterator()` + `itertools.batched` to stream finding IDs without materializing the full list into memory.

**2 (CRITICAL) — Single transaction in Step 5**: Kept as-is. By the time Step 5 runs, all Findings and Endpoint_Status rows are already deleted — only lightweight rows (Tests, Engagements, Endpoints) remain, so the transaction is fast even for large Product_Types. Added a comment explaining this.

**3 (HIGH) — No statement_timeout/deadlock protection**: Added `SET LOCAL statement_timeout = '300s'` to every SQL execution in `cascade_delete`, and `OperationalError` logging that surfaces deadlock/timeout details.

**4 (HIGH) — Double subquery + exists/count**: Extracted `scope_ids` variable for reuse and replaced `.exists()` + `.count()` with a single `.count()`.

**5 (HIGH) — FileUpload per-object delete**: Added comment explaining the intentional per-object delete (custom `FileUpload.delete()` removes files from disk). File attachments are uncommon so the overhead is negligible.

**6 (MEDIUM) — Nested subqueries**: Tested at ~1M findings with good performance. The new code is still orders of magnitude faster than the old fan-out approach. No changes needed.

**7 (MEDIUM) — fix_loop_duplicates on every delete**: Added `scope_qs` parameter to scope the double self-join to the deletion set instead of the full table. The celerybeat task still runs unscoped.

**8 (MEDIUM) — Unbounded queryset in signal**: Added documentation to the signal definition that receivers MUST NOT materialize the full queryset (use `.filter()`, `.iterator()`, or aggregation).

**9 (LOW) — bulk_remove_all_tags called twice**: Added `skip_m2m_for` parameter to `cascade_delete` — `bulk_delete_findings` now passes `skip_m2m_for={Finding}` to skip redundant M2M cleanup.

**10 (LOW) — Progress visibility**: Not addressed in this PR — out of scope for now.

Maffooch (approved):

dojo/settings/settings.dist.py
```diff
@@ -302,7 +302,7 @@
     # Initial behaviour in Defect Dojo was to delete all duplicates when an original was deleted
     # New behaviour is to leave the duplicates in place, but set the oldest of duplicates as new original
     # Set to True to revert to the old behaviour where all duplicates are deleted
-    DD_DUPLICATE_CLUSTER_CASCADE_DELETE=(str, False),
+    DD_DUPLICATE_CLUSTER_CASCADE_DELETE=(bool, False),
```
Maffooch:
```suggestion
    DD_DUPLICATE_CLUSTER_CASCADE_DELETE=(bool, True),
```
Carry over from https://github.com/DefectDojo/django-DefectDojo/pull/14565

blakeaowens (approved):

dogboat (approved):

Jino-T (approved):

PR #14591, opened 2026-03-25
fix(github_sast): set unique_id_from_tool for dedup

### LINKED ISSUE #13681: Github SAST Scan duplicate findings due to hashcode issue

samiat4911:
**Description**

Fixes #13681.
The `Github SAST Scan` parser was previously missing a `unique_id_from_tool` mapping and its deduplication algorithm fallback (`DEDUPE_ALGO_HASH_CODE`) included the `line` field. Because GitHub alerts frequently change lines during natural development, this caused duplicate findings to be created for the same alert on reimport.

This PR fixes the issue by:
1. Extracting the stable alert `number` field from the GitHub JSON and mapping it to `unique_id_from_tool` in the `GithubSASTParser`.
2. Setting the deduplication configuration for `Github SAST Scan` to use `DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL` in `settings.dist.py`.

```python
# dojo/tools/github_sast/parser.py
finding = Finding(
    ...
    vuln_id_from_tool=rule_id,
    unique_id_from_tool=str(vuln.get("number")),
)
```
**Test results**

Extended the test suite in `unittests/tools/test_github_sast_parser.py` to ensure the new mapping is successfully parsed.

```python
# unittests/tools/test_github_sast_parser.py
self.assertEqual("35", finding.unique_id_from_tool)
```
- Ran local `pytest` manually for `unittests.tools.test_github_sast_parser`. All tests passed.
- Verified parser extraction logic via a standalone script across multiple fixtures (`github_sast_one_vul.json`, `github_sast_many_vul.json`).

**Documentation**

No documentation updates are necessary for this bug fix as it aligns the parser with expected DefectDojo deduplication behavior without changing user-facing configurations.

**Checklist**


Maffooch:
Hi @samiat4911 I always get a little nervous with deduplication changes.. Have you tested this change against existing data to ensure that deduplication can still be performed without any issues. My flow is usually like this:
- Start form dev branch branch
- Import some data
- reimport the same file (see no findings were touched)
- Checkout feature branch
- reimport the same file (see no findings were touched)

If I make it to the end of that flow, I feel confident that the change is not impacting

samiat4911:
> Hi @samiat4911 I always get a little nervous with deduplication changes.. Have you tested this change against existing data to ensure that deduplication can still be performed without any issues. My flow is usually like this:
> 
> * Start form dev branch branch
> * Import some data
> * reimport the same file (see no findings were touched)
> * Checkout feature branch
> * reimport the same file (see no findings were touched)
> 
> If I make it to the end of that flow, I feel confident that the change is not impacting

Hi @Maffooch, great point about deduplication,

I've updated the PR to use `DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL_OR_HASH_CODE` instead.

This ensures backward compatibility: existing findings without a `unique_id_from_tool` will safely fall back to the old hash code matching, preventing any duplicates on upgrade. New imports will then use the stable GitHub alert `number`.

Maffooch:
Thank you for verifying! @valentijnscholten are you okay with this approach?

dojo/tools/github_sast/parser.py
```diff
@@ -71,6 +71,7 @@ def get_findings(self, filename, test):
                 static_finding=True,
                 dynamic_finding=False,
                 vuln_id_from_tool=rule_id,
+                unique_id_from_tool=str(vuln.get("number")),
```
valentijnscholten: Can you defend against the get returning None?

valentijnscholten (changes requested):

dojo/tools/github_sast/parser.py
samiat4911: Ah, good catch I didn't realize str(None) would evaluate to the literal string 'None'. I just pushed a quick fix so it safely returns an actual None if the number isn't present.

valentijnscholten (approved):

mtesauro (approved):
Approved

Maffooch (approved):

Jino-T (approved):

PR #13214, opened 2025-09-18
OpenVAS parser improvments

jostaub:
This PR should address the problems described in:
- #12378 
- #12849

Here an overview of the most important changes:
- Use of a hash code algorithm for de duplication (bases of problems described in the linked issues)
- Increased consistency in parsing between the XML and CSV parsers.
- Combined findings where the only differences are in fields that cannot be rehashed due to inconsistent values between scans (e.g. fields containing timestamps or packet IDs). This prevents duplicates if the vulnerability is found multiple times on the same endpoint.
- Increased parser value coverage
- Heuristic for fix_available detection
- Updated mapping to DefectDojo fields compared to version 1.
- Improved docs
- Simmilar XML and CSV implementation
- Added Testing files (this is the reason for the large line count)

In #12378 we discussed why it is a new parser version instead of an update of the existing parser. The concrete fields (that where the reason mentioned in the issue) are no longer the same since i changed the dedpulication algo since then. However the base reason is still the same (hashing fields not present in original parser).

For a long time, this PR did not deduplicate on endpoints. However, this resulted in duplication markings during reimport in the default engagement configuration, which is why this functionality was removed (for now).

PR #14592, opened 2026-03-25
fix: Add file_path based detail mode for Anchore Grype parser

Kasyap7:
Modified dupe_key in `Anchore Grype detailed` parser to include file_path, preventing same CVE found in different binaries from being merged into a single finding 

- **Default mode unchanged**: aggregates findings per CVE + component name + version using DEDUPE_ALGO_HASH_CODE
- **Detailed mode**: creates a separate Finding per file path using DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL with unique_id_from_tool composed as `vuln_id|component_name|component_version|file_path`
- **Update settings.dist.py**: add "Anchore Grype detailed" to HASHCODE_ALLOWS_NULL_CWE and DEDUPLICATION_ALGORITHM_PER_PARSER
- Add unit tests for both modes and test fixture same_cve_different_paths.json
- Update parser documentation with field mapping table

Reference: https://github.com/DefectDojo/django-DefectDojo/discussions/14573


Maffooch (dismissed):

mkutsybala (approved):
I intended to make similar changes.

anna-shcherbak:
I also had the same issue, where after import Grype report to DefectDojo, many findings went missing due to the same vuln id, component, and its version, but different paths.
And this poses a big security risk when my team doesn't see all the vulnerabilities in the code, due to this bug.
Hope this will be fixed as soon as possible.

valentijnscholten:
@Maffooch As usual "fixing" a parser sometimes means side effects for users with existing findings. In this case it cannot be fixed by a `dedupe` run as the findings are/were already deduplicated in the parser.

Options that I see:
1) Accept that the current parser is wrong and the new parser is right and people will get their existing findings closed/mismatched and new findings created.
2) Introduce a V2 parser similar to what we did with [Openvas](https://github.com/DefectDojo/django-DefectDojo/pull/13214)
3) Introduce a "detailed" mode similar to using the mechanism built-in to our parser contract: https://github.com/DefectDojo/django-DefectDojo/blob/ccb402112b210936975d04dc1fc29a593ae46ea6/dojo/tools/checkmarx/parser.py#L91-L97
This detail mode would not collapse findings for different `file_path`s into one finding.

My vote is for number 3.

Jino-T (approved):

mkutsybala (commented):
Probably would be more explicit and pythonic to fix it like this:
`dupe_key = (vuln_id, artifact_name, artifact_version, file_path)`

Maffooch:
@valentijnscholten wouldn't this PR simply allow for more findings to be created per import/reimport? I don't think that would have implication on close old findings or deduplication, but I could be missing something there. I understand that we cannot go back in time to create findings that were not created in the past, but I don believe that is the case with option 3 either. If we take option 3, we will have the same outcome of new findings being created when compared to the existing anchore grype parser, but users will need to move their CI to the new test type, and then be unable to deduplicate against older anchore grype findings since that would a different test type.

valentijnscholten:
The PR as-is would generate new findings. But they will all have the same hash_code and be marked as duplicates. 
To avoid that, the hash_code field config would need to be changed as well: https://github.com/DefectDojo/django-DefectDojo/discussions/14573#discussioncomment-16321105.
And that would lead to the situation I described that existing findings without `file_path` would not longer be matched in reimports.

Maffooch:
Ohh I understand now! Thank you linking the comment on the issue. In that case, I agree that option 3 is the least impactful to user data

blakeaowens (approved):

Kasyap7:
Added `Anchore Grype detailed` logic to grype parse along with testcases

dojo/tools/anchore_grype/parser.py
```diff
@@ -185,7 +195,10 @@ def get_findings(self, file, test):
                 if finding_epss_score is None and rel_vuln_id:
                     finding_epss_score, finding_epss_percentile = self.get_epss_values(vuln_id, vuln_epss)
 
-            dupe_key = finding_title
+            if self.mode == "detailed":
+                dupe_key = f"{vuln_id}|{artifact_name}|{artifact_version}|{file_path}"
+            else:
+                dupe_key = finding_title
```
valentijnscholten:
```suggestion
            if self.mode == "detailed":
                dupe_key = f"{vuln_id}|{artifact_name}|{artifact_version}|{file_path}"
            else:
                dupe_key = f"{vuln_id}|{artifact_name}|{artifact_version}"
```

Might be better to make it explicit and readable.

valentijnscholten (changes requested):
Thanks again for the PR. Would you be willing to add some info to `/docs/content/supported_tools/parsers/file/anchore_grype.md`.

valentijnscholten (approved):

blakeaowens (approved):

Jino-T (approved):

Maffooch (approved):

PR #14593, opened 2026-03-25
Standardize CI tests on Debian AMD64 and document supported image variants

Maffooch:
## Summary
- Add a **Docker Image Variants** section to the installation docs documenting that Debian AMD64 is the officially supported and tested configuration, while Alpine and ARM64 are community-supported
- Reduce the CI test matrix to only build and test the Debian AMD64 Django image, cutting integration test jobs roughly in half
- Switch performance tests from alpine to debian since django-alpine is no longer built for testing

## Test plan
- [ ] Verify CI workflows trigger correctly on a PR and only build `django-debian`, `nginx-alpine`, and `integration-tests-debian` for `linux/amd64`
- [ ] Confirm integration tests, REST framework tests, and performance tests all pass using the Debian image
- [ ] Review the installation docs render correctly with the new Docker Image Variants table

### DIFF
--- a/.github/workflows/build-docker-images-for-testing.yml
+++ b/.github/workflows/build-docker-images-for-testing.yml
@@ -24,6 +24,8 @@ jobs:
         os: [alpine, debian]
         platform: ["${{ inputs.platform }}"]
         exclude:
+            - docker-image: django
+              os: alpine
             - docker-image: nginx
               os: debian
             - docker-image: integration-tests
--- a/.github/workflows/build-docker-images-for-testing.yml
+++ b/.github/workflows/build-docker-images-for-testing.yml
@@ -24,8 +24,6 @@ jobs:
         os: [alpine, debian]
         platform: ["${{ inputs.platform }}"]
         exclude:
-            - docker-image: django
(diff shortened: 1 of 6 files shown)


.github/workflows/build-docker-images-for-testing.yml
```diff
@@ -24,6 +24,8 @@ jobs:
         os: [alpine, debian]
         platform: ["${{ inputs.platform }}"]
         exclude:
+            - docker-image: django
+              os: alpine
```
valentijnscholten:
```suggestion
```

I would prefer that (for now) we keep publishing the alpine images, and this PR seems to agree as the release workflows are not changed. 

But to make sure we do not run into a broken build on release day I think we should keep building the alpine images as part of the test suite.

valentijnscholten (changes requested):

paulOsinski (approved):

valentijnscholten (approved):

blakeaowens (approved):

mtesauro (approved):
Approved

PR #14610, opened 2026-03-30
update invicti parser to use FirstSeenDate

paulOsinski:
Invicti parser will now apply FirstScanDate, if present, to the Finding's `date` object.  This change does not affect deduplication or anything destructive, we're simply parsing new data that we weren't before.


dojo/tools/netsparker/parser.py
```diff
@@ -62,6 +68,7 @@ def get_findings(self, filename, test):
             dupe_key = title
             request = item["HttpRequest"].get("Content", None)
             response = item["HttpResponse"].get("Content", None)
+            finding_date = _parse_date(item.get("FirstSeenDate")) or scan_date
```
Maffooch: This should reflect the `USE_FIRST_SEEN` [flag](https://github.com/DefectDojo/django-DefectDojo/blob/3699b7611c899516ffc25062ae76bf3483e265a1/dojo/settings/settings.dist.py#L349-L350)

Maffooch (changes requested):

paulOsinski:
@Maffooch I have added the flag gating for `USE_FIRST_SEEN`.  If you're wondering about the force-push, I made a change to the Checkmarx parser but realized it was out of scope so I dropped it from the PR.

Maffooch (approved):

valentijnscholten (approved):
Looks good, but ideally we update the comment on the `USE_FIRST_SEEN` flag in `settings.dist.py` as it currently only mentions Qualys. Maybe also add a line to the docs for both parsers supporting this flag? Not blocking for merge.

rossops (approved):

mtesauro (approved):
Approved

PR #14612, opened 2026-03-30
when deleting a URL via API, perform_delete should call delete on the…

dogboat:
This PR updates the `URL` API endpoint to correct a bug discovered by @blakeaowens; it implements his solution.

Previously, deleting a URL object through the API deleted the URL object itself, leaving a dangling AbstractLocation entry. The fix is simple enough: override the URLViewSet's `perform_destroy()` method to call `.delete()` on the instance's AbstractLocation reference instead. This will cascade to the URL.


mtesauro (approved):
Approved

Maffooch (approved):

valentijnscholten (approved):

Jino-T (approved):

PR #14616, opened 2026-03-31
AWS Inspector 2 Line number bug + other changes

Jino-T:
[sc-12948]

- There was a bug with line number where if the scan didn't have the relevant value the import would crash because the default value was the string "N/A" instead of none.
- Additionally the following fields were added for Package Vulnerability type findings:
  1. component_name                                                                           
  2. component_version                                                                        
  3. file_path                                                                                
  4. references                                                                               
  5. publish_date                                                                             
  6. cvssv3_score

### DIFF
--- a/dojo/tools/aws_inspector2/parser.py
+++ b/dojo/tools/aws_inspector2/parser.py
@@ -137,6 +137,14 @@ class AWSInspector2Parser:
             finding.component_name = vulnerable_packages[0].get("name")
             finding.component_version = vulnerable_packages[0].get("version")
             finding.file_path = vulnerable_packages[0].get("filePath")
+            if settings.V3_FEATURE_LOCATIONS and finding.component_name:
+                finding.unsaved_locations.append(
+                    LocationData.dependency(
+                        name=finding.component_name,
+                        version=finding.component_version or "",
+                        file_path=finding.file_path or "",
+                    ),
+                )
         # reference URLs from the advisory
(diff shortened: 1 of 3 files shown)


dojo/tools/aws_inspector2/parser.py
```diff
@@ -149,8 +169,8 @@ def get_code_vulnerability(self, finding: Finding, raw_finding: dict) -> Finding
         file_path_info = raw_finding.get("filePath", {})
         file_name = file_path_info.get("fileName", "N/A")
         file_path = file_path_info.get("filePath", "N/A")
-        start_line = file_path_info.get("startLine", "N/A")
-        end_line = file_path_info.get("endLine", "N/A")
+        start_line = file_path_info.get("startLine", None)
+        end_line = file_path_info.get("endLine", None)
```
Maffooch, line 181: This is a fine change to make here, but further down in the description building section, the start and end lines need to be "N/A" in the description when those values are `None` to maintain current deduplication behaviors

Maffooch (dismissed):

dojo/tools/aws_inspector2/parser.py
Jino-T: Good thinking. I added logic to handle this. Nice catch

Maffooch (approved):

valentijnscholten (approved):

rossops (approved):

dogboat (approved):

PR #14625, opened 2026-04-01
Migration endpoints to locations fix

dogboat:
This PR updates the `endpoints_to_locations` Django management command in two ways. First, it allows the command to be run when V3/Locations is enabled. Second, it logs its migration progress so users won't think it's hung.


Maffooch (approved):

valentijnscholten (approved):

Jino-T (approved):

mtesauro (approved):
Approved

PR #14626, opened 2026-04-01
fix(parsers): use unsaved_tags instead of tags= in Finding constructor

valentijnscholten:
## Summary

While working performance improvements I noticed the JFrog xray parser looked quite slow compared to other parsers.
Turns out that setting the `tags` field while constructing `Finding` models (in-memory) made it very slow.

The 14k findings report took anywhere between 60 and 90 second just to parse. By switching to `unsaved_tags` it's now ~9seconds. Some other parsers needed the same fix and are part of this PR.

- Passing `tags=` directly to the `Finding()` constructor triggers expensive tagulous tag processing for every finding object created during parsing
- Changed 5 affected parsers to use `finding.unsaved_tags = tags` after construction instead, which lets the import pipeline handle tags efficiently
- Affected parsers: jfrog_xray_unified, dependency_check, cargo_audit, anchore_grype, threat_composer

Benchmark on JFrog Xray Unified with 14,219 findings: **99s → 7.97s (12x faster)**


valentijnscholten:
cleaned up the format only changes which are not relevant here.

mtesauro (approved):
Approved

Maffooch (approved):

rossops (approved):

Jino-T (approved):

PR #14627, opened 2026-04-01
fix(reimport): do not update finding tags on reimport for matched findings

valentijnscholten:
## Summary

The behaviour of the Finding Tags in the reimport has been going through a bit of rough patch. At some point it was broken and it was fixed via #14459 . But that PR resulted in tags being added to existing tags, possibly leading to "too many" tags or "obsolete" tags. To avoid any confusion we have now decided to make the tags field behave like all other fields. It's set during import and not updated afterwards. 

We are considering flags on the reimport request to trigger (certain) fields to be updated during reimport, but for now this PR is the desired behaviour.

- On reimport, tags from the report were being appended to matched/existing findings via `tags.add()`, causing tags to accumulate across reimports instead of being left unchanged. This was a regression introduced in #14459 which changed `tags.set()` to `tags.add()`.
- This fix skips tag updates for matched findings during reimport, consistent with how other finding fields (title, severity, description, etc.) are handled — only new findings get tags from the report.
- Updated `test_import_report_with_tags` to reflect the corrected behavior.

### DIFF
--- a/dojo/importers/default_reimporter.py
+++ b/dojo/importers/default_reimporter.py
@@ -409,6 +409,7 @@ class DefaultReImporter(BaseImporter, DefaultReImporterOptions):
                     finding = self.finding_post_processing(
                         finding,
                         unsaved_finding,
+                        is_matched_finding=bool(matched_findings),
                     )
                     # all data is already saved on the finding, we only need to trigger post processing in batches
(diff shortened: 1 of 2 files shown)


mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

dogboat (approved):

PR #14630, opened 2026-04-02
fix: clear reverse M2M through tables before cascade deletion

valentijnscholten:
## Summary

The optimized Async Delete from #14566 correctly cascaded M2M relationships, but not _reverse_ M2M relationships. I could have sworn I tested Finding Groups, but to be sure this PR has test cases for these M2M relationships.

- Fixes FK violations during async object deletion caused by reverse M2M through tables not being cleared before finding/engagement deletion.
- Both `bulk_clear_finding_m2m` and `cascade_delete_related_objects` only discovered forward M2M fields (`model._meta.many_to_many`). Reverse M2M fields defined on other models (`Finding_Group.findings`, `Risk_Acceptance.accepted_findings`, `EnhancedRiskAcceptance.affected_engagements`) were missed.
- Uses `get_fields()` which returns both forward `ManyToManyField` (via `field.remote_field.through`) and reverse `ManyToManyRel` (via `field.through`) to discover all through tables.

## Context

Follow-up to #14566. The error observed in production:

```
psycopg.errors.ForeignKeyViolation: update or delete on table "dojo_finding" violates foreign key constraint
"dojo_finding_group_f_finding_id_ID_fk_dojo_find" on table "dojo_finding_group_findings"
```

Triggered from `async_delete_task` → `bulk_delete_findings` when deleting a Product with findings in Finding Groups.

### DIFF
--- a/dojo/finding/helper.py
+++ b/dojo/finding/helper.py
@@ -726,11 +726,18 @@ def bulk_clear_finding_m2m(finding_qs):
     # Remove tags with proper count maintenance
     bulk_remove_all_tags(Finding, finding_ids)
 
-    # Auto-discover and delete remaining (non-tag) M2M through tables
-    for m2m_field in Finding._meta.many_to_many:
-        if hasattr(m2m_field, "tag_options"):
+    # Auto-discover and delete M2M through tables — both forward (Finding._meta.many_to_many)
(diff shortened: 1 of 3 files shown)


mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

rossops (approved):

PR #14634, opened 2026-04-03
Add permission checks for moving engagements between products

Jino-T:
- When editing an engagement and changing its product (moving it to a different product), the system now
   verifies the user has `Engagement_Edit` permission on the destination product                          
  - This applies to both the API (PUT/PATCH on `/api/v2/engagements/{id}/`) and the UI (edit engagement
  form)                                                                                                   
  - Adds 8 new tests covering authorized moves, unauthorized moves, same-product updates, and partial
  updates without a product field


Maffooch (approved):

paulOsinski (approved):

mtesauro (approved):
Approved

valentijnscholten (approved):

PR #14666, opened 2026-04-08
fix css overflow issue - reports

paulOsinski:
## Summary
PDF reports have a text wrapping bug where multi-line finding fields (description, mitigation, impact, steps to reproduce, etc.) overflow past page margins and get cut off when printed to PDF. This is especially common with findings imported from tools like BugCrowd CSV, which embed `<pre>` tags in the field data.

## Root cause (two issues):

- Missing overflow-wrap: The pre CSS in `report_base.html` had word-break: normal but no overflow-wrap: break-word, so long unbroken strings (URLs, tokens, encoded data) would not wrap at the container boundary.

- Nested `<pre>` tags: All 7 PDF report templates wrapped markdown_render output in `<pre>` tags. When imported finding data already contained `<pre>` tags (e.g. `<pre data-language="plain">`), this produced nested `<pre><pre>...</pre></pre>` elements. The inner `<pre>` could pick up default browser styles (overflow: auto, scrollbars) that override the report CSS, causing content to clip or scroll instead of wrapping.

## Fix:

- Replaced `<pre>` wrappers around all markdown_render calls with `<div class="report-field">` across all 7 PDF report templates — eliminates the nesting problem
- Added .report-field and .report-field pre CSS rules with overflow-wrap: break-word — ensures long strings wrap at margins, even when the data itself contains `<pre>` tags
- Left `<pre class="raw_request">` tags untouched — those correctly wrap raw request/response data that isn't markdown-rendered
- Add a unit test


paulOsinski:
re: dryrun.  Likely a false positive

XSS alerts (8 red/orange): The scanner flagged `markdown_render` because it "could not find the implementation" and assumed it might emit unsanitized HTML. The implementation is in `dojo/templatetags/display_tags.py (line 85) `and it's already sanitized through two layers:

- `markdown.markdown() `— converts markdown to HTML
- `bleach.clean()` — sanitizes the output against an explicit allowlist (markdown_tags, markdown_attrs, markdown_styles) with a CSSSanitizer

This is the same `markdown_render` filter that was already in use on these exact same lines before our change. We didn't introduce it, we only changed the wrapper element.

Maffooch (approved):

valentijnscholten (commented):
I wonder if we really need these test cases? If we make tests for everything then soon we will have a very long running test suite. wdyt?

mtesauro (approved):
Approved

blakeaowens (approved):

Maffooch:
> I wonder if we really need these test cases? If we make tests for everything then soon we will have a very long running test suite. wdyt?

ever heard of regression testing 😉

valentijnscholten:
But this introduces multiple tests just to see if there's is one html pre tag not nested somewhere as it may introduce a small layout rendering thing. I think this is too small a thing to have so many tests.

rossops (approved):

PR #14667, opened 2026-04-08
feat/helm: add basic support for Gateway API HTTPRoute for routing

ariyonaty:
### Add Gateway API HTTPRoute support for Django service

Adds an optional HTTPRoute resource (Gateway API v1) as an alternative to the existing Ingress for exposing the Django service. This aligns with the Kubernetes ecosystem's move toward the Gateway API as the successor to Ingress.

### Changes

- `templates/django-httproute.yaml` : New template gated by django.httpRoute.enabled. 
- `values.yaml` : Added django.httpRoute section (enabled: false by default) with annotations and parentRefs fields.
- `values.schema.json` : Schema definition for the new django.httpRoute object.

### Usage

```yaml
django:
  httpRoute:
    enabled: true
    annotations: {}
    parentRefs:
    - name: my-gateway
      namespace: default
```

### Notes

- Disabled by default — no impact on existing deployments.
- Can be used alongside or instead of the existing Ingress resource.
- parentRefs entries default group to gateway.networking.k8s.io and kind to Gateway; only name is required.

### DIFF
--- a/helm/defectdojo/templates/django-httproute.yaml
+++ b/helm/defectdojo/templates/django-httproute.yaml
@@ -1,0 +1,53 @@
+{{- if .Values.django.httpRoute.enabled -}}
+{{- $fullName := include "defectdojo.fullname" . -}}
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  {{- if or .Values.extraAnnotations .Values.django.httpRoute.annotations }}
+  annotations:
+  {{- range $key, $value := .Values.extraAnnotations }}
+    {{ $key }}: {{ quote $value }}
+  {{- end }}
+  {{- range $key, $value := .Values.django.httpRoute.annotations }}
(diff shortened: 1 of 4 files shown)


helm/defectdojo/templates/django-httproute.yaml
```diff
@@ -0,0 +1,53 @@
...
+  hostnames:
+  - {{ .Values.host | quote }}
+  rules:
+  - matches:
+    - path:
+        type: PathPrefix
+        {{- if .Values.django.httpRoute.path }}
+        value: {{ .Values.django.httpRoute.path }}
+        {{- else }}
+        value: "/"
```
rossops, line 48: The template reads .Values.django.httpRoute.path, but path is neither in values.yaml nor in values.schema.json. Either add it explicitly (with default "/") or remove the conditional — currently it's a hidden undocumented option.

Maffooch:
closing as stale

Issue #14640, opened 2026-04-05
Close Old Findings doesn't work for parsers that set the service field


valentijnscholten:
Whenever parsers set the `service` field on findings, this breaks the `close_old_findings` functionality.

For example the StackHawk parser sets the `service` field based on the `application` name from the report. All findings get a `service` value, for example "Application A".
But the Reimport requests did not contain a `service` value. The Close Old Findings feature will only close findings that have the service value specified in the request. So any findings that have a non-empty value for `service` will not get closed. In this case no findings get closed as all of them have `Application A` from the original import of the report. 

Also if the application name changes the findings in the reimport report will no longer be matched against existing findings.

I think parser should never set the service fields as it is meant to be a meta field.

A quick search shows these affected parsers:

<img width="915" height="1053" alt="Image" src="https://github.com/user-attachments/assets/3e2a34a3-5203-46bb-90ab-9576c887f3d8" />

mtesauro:
I agree with that thinking.

I see service as something a user of DefectDojo would set vs something a parser should never set since it cannot know what any specific user wants for service.

manuel-sommer:
Interesting finding. Multiple contributions here were implemented from me. So if the right approach is to remove the service field, I can submit PRs to fix it in the parsers. Please give me feedback @valentijnscholten or @mtesauro before I submit a PR.
Also: Is it possible to cover this edge case (and maybe other fields) through a global unittest? Something like: "This field is used within a parser, but is not foreseen to be used within a parser as it breaks other functionalities"? Are there additional fields that affect close_old_findings or other functionalities?

valentijnscholten:
The challenge is that if we adjust the parsers to no longer set the service field, this will affect the hash_code calculation. So probably we would have to have a data migration that unsets the service field for existing findings. However there is no way to know if those service values came from the parser or from the import/reimport request parameter. Do you have any idea on what a sensible approach forward might be?

manuel-sommer:
1. Fix parsers: Update affected parsers to no longer set the service field at all.
2. To avoid breaking hash matching and reimport behavior, we can follow the same pattern used in previous migrations (for example [0208_merge_acunetix.py](https://github.com/DefectDojo/django-DefectDojo/blob/master/dojo/db_migrations/0208_merge_acunetix.py)):
- Identify findings created by affected parsers (based on test.test_type)
- Only touch findings where the service is set and the parser is known to have populated it
[...]

PR #14687, opened 2026-04-15
:bug: fix close_old_findings through service field in several parsers #14640

manuel-sommer:
#14640


manuel-sommer:
@valentijnscholten please take a look here

dojo/db_migrations/0264_clear_service_for_affected_parsers.py
Maffooch, line 1: If a service was supplied to these test types at import time, does that value override the value set by the parser? If so, could this migration potentially erase service fields that are from the user?

manuel-sommer: That's a question beyond my knowledge. Maybe @valentijnscholten has an opinion on this.

valentijnscholten: The importers overwrite any fields set by the parsers:

https://github.com/DefectDojo/django-DefectDojo/blob/9d661d72e9f92b8e6e3eef334afb7257dd968fb1/dojo/importers/default_importer.py#L216-L217

https://github.com/DefectDojo/django-DefectDojo/blob/9d661d72e9f92b8e6e3eef334afb7257dd968fb1/dojo/importers/default_reimporter.py#L337-L338

Maffooch: So it seems like this migration would erase data intentionally set by users...

manuel-sommer: Is there a possibility to distinguish between the two ways?

manuel-sommer:
any updates on this @valentijnscholten and @Maffooch ? I will resolve conflicts after the next review.

Maffooch:
I think this one will be too dangerous to walk back for existing data - closing

manuel-sommer:
Shall I do a PR that will not touch existing data (without a db migration)?

valentijnscholten:
I vote Yes as #14640 is still a thing that should be fixed in my opinion. The PR should probably then also get a check / raise a validation error if any future parser tries to set the service field.

PR #14701, opened 2026-04-15
perf: bulk-apply parser-supplied per-finding tags during import

valentijnscholten:
Tags are accumulated per batch and applied just before the `post_process_findings_batch` task is dispatched, so deduplication and rules tasks see the tags already written to the DB.

Both `default_importer` and `default_reimporter` use the same approach. For the reimporter, `finding_post_processing` accepts an optional `tag_accumulator` list; when supplied, tags are accumulated rather than applied inline (backward-compatible for any direct callers).

These queries are not covered (yet) by the performance test that we have as the Stackhawk parser doesn't set tags.

### DIFF
--- a/dojo/importers/default_importer.py
+++ b/dojo/importers/default_importer.py
@@ -19,8 +19,8 @@ from dojo.models import (
     Test_Import,
 )
 from dojo.notifications.helper import create_notification
-from dojo.utils import get_full_url, perform_product_grading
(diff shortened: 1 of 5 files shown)


mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

Jino-T (approved):

PR #14708, opened 2026-04-18
Add centralized banner system with OS messaging support

Maffooch:
- Introduces a new Open Source Messaging banner that fetches markdown content from a GCS bucket, renders it with bleach-sanitized HTML, and        
  displays it via the existing additional_banners template loop. Cached for 1 hour; fetch failures silently yield no banner.
- Replaces the DD_CREATE_CLOUD_BANNER / CREATE_CLOUD_BANNER setting and all cloud-banner-specific logic with a centralized additional_banners      
  context processor pattern. Both OS messages and product announcements now flow through the same rendering pipeline, each tagged with a source field
   (os, product_announcement).              
- Migrates ProductAnnouncementManager from django.contrib.messages to session-based banners (_product_banners), which are popped by the context    
  processor and rendered in the unified template loop.                                                                                               
- Removes the evaluate_pro_proposition Celery task/beat schedule and the create_announcement_banner initialization command logic.
- Simplifies the add_announcement_to_new_user signal by removing cloud-banner conditional checks.                                                  
- Adds 52 unit tests covering the OS message fetcher/parser and product announcement manager.

### DIFF
--- a/dojo/announcement/os_message.py
+++ b/dojo/announcement/os_message.py
@@ -7,7 +7,7 @@ from django.core.cache import cache
 
 logger = logging.getLogger(__name__)
 
-BUCKET_URL = "https://storage.googleapis.com/defectdojo-os-messages-dev/open_source_message.md"
+BUCKET_URL = "https://storage.googleapis.com/defectdojo-os-messages-prod/open_source_message.md"
 CACHE_SECONDS = 3600
 HTTP_TIMEOUT_SECONDS = 2
 CACHE_KEY = "os_message:v1"
--- a/dojo/announcement/os_message.py
+++ b/dojo/announcement/os_message.py
@@ -7,7 +7,7 @@ from django.core.cache import cache
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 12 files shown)


valentijnscholten (commented):
What is the reason for the server-side caching? Is it for client side performance or to save on outgoing traffic on the bucket? Wouldn't it be much easier to just let the browser load it straight from the bucket and configure the bucket with an appropriate `Cache-Control` header? This way it would also work if the server is inside some network that is not allowed to egress to GCS.

Maffooch:
There is some formatting server side before the raw content is sent to the browser. Seems like it would be less performant to do that computation in the browser on every request rather than once every few through the cache

valentijnscholten (approved):
Yeah I don't understand why so much processing is needed. The content is in your control so I would expect it to be tailer made and would only need a nh3 or dompurify sanitize as assurance for OS users/admins.

blakeaowens (approved):

rossops (approved):

mtesauro (approved):
Approved

PR #14714, opened 2026-04-20
Fix Contrast parser collapsing findings with the same rule name

Jino-T:
[sc-12973]

## Summary
- The `dupe_key` in the Contrast parser was hashed from `vuln_id_from_tool` (Rule Name), causing all findings of the same rule type (e.g. `ldap-injection`) to be collapsed into a single finding
- Changed the `dupe_key` to hash `unique_id_from_tool` (Vulnerability ID) instead, since each Contrast Vulnerability ID uniquely identifies a distinct finding instance
- The existing merge/aggregation block is retained as a safety net for the edge case of duplicate Vulnerability IDs in a report

## Test plan
- [x] `test_ldap_multiple_findings` — verifies that 3 findings with the same Rule Name but different Vulnerability IDs are returned as 3 distinct findings (regression test for the bug)
- [x] `test_duplicate_vuln_id_is_merged` — verifies the merge safety net still works when a Vulnerability ID appears twice in a report
- [x] `test_example_report` and `test_example2_report` — existing tests updated to reflect correct post-fix finding counts


Maffooch (approved):

dojo/tools/contrast/parser.py
```diff
@@ -90,7 +90,7 @@ def get_findings(self, filename, test):
                 )
 
             dupe_key = hashlib.sha256(
-                f"{finding.vuln_id_from_tool}".encode(),
+                f"{finding.unique_id_from_tool}".encode(),
```
valentijnscholten: we might as well just use `unique_id_from_tool` and not hash it?

valentijnscholten (approved):

blakeaowens (approved):

dojo/tools/contrast/parser.py
Jino-T: I agree, might as well save the extra work. Makes it easier to debug too

mtesauro (approved):
Approved

PR #14715, opened 2026-04-20
docs optimization: use defer instead of async

paulOsinski:
By using `defer` as the argument on our REO.js script we can have it run after the rest of the page is ready, so that we're not competing for load time in our docs.


Jino-T (approved):

mtesauro (approved):
Approved

valentijnscholten (changes requested):
Can you remove the helm chart changes?

paulOsinski:
done, thanks @valentijnscholten!

Maffooch (approved):

valentijnscholten (approved):

PR #14764, opened 2026-04-27
refactor: consolidate RBAC into dojo/authorization package

Maffooch:
## Description

Consolidate every RBAC / authorization concern into a single `dojo/authorization/` package. Before this PR, authorization code lived in seven different places: `dojo/models.py` (RBAC models), 14 per-app `queries.py` files (`get_authorized_*`), every view file (`@user_is_authorized` decorators), `dojo/api_v2/permissions.py`, `dojo/location/api/permissions.py`, and `dojo/templatetags/authorization_tags.py`. After this PR, it all lives in `dojo/authorization/`.

Net diff: 85 files changed, +3015 / −2113.

### What moves into `dojo/authorization/`

- **`models.py`** — 7 RBAC models (`Role`, `Global_Role`, `Dojo_Group_Member`, `Product_Member`, `Product_Group`, `Product_Type_Member`, `Product_Type_Group`) extracted from `dojo/models.py`. `app_label='dojo'` is preserved so no migrations are needed; ~47 import sites are updated.
- **`api_permissions.py`** — `dojo/api_v2/permissions.py` and `dojo/location/api/permissions.py` merged into one module; originals deleted; consumers and tests rewritten.
- **`template_filters.py`** — filter functions extracted from `dojo/templatetags/authorization_tags.py`. The templatetags module becomes a thin registration proxy.
- **`query_filters.py` + `query_registrations.py`** — registry pattern (~1.9k lines of RBAC filter logic) extracted from 14 per-app `queries.py` files. Each `get_authorized_*` becomes a thin wrapper that defers to the registry and falls back to unfiltered querysets when no RBAC backend is registered.
- **`url_permissions.py` + `middleware.py`** — map ~198 URL names to permission checks and enforce them via `AuthorizationMiddleware.process_view`. Removes `@user_is_authorized`, `@user_has_global_permission`, and `@user_is_configuration_authorized` from 26 view files.
- **`__init__.py`** — exports the public surface and triggers query-filter registration at app startup.

### Why

- **Single source of truth** for "is this allowed?" — checks live next to the models, registry, and middleware that enforce them.
- **No more per-view decorators** — one URL-permission map replaces ~200 decorator lines scattered across 26 view files.
- **Decouples `dojo.models` from authorization** — RBAC models live in their own module, removing circular-import pressure.
- **Pluggable** — `get_authorized_*` falls back to unfiltered querysets when no RBAC backend is registered, so non-RBAC deployments keep working.

### Behavioral changes

None intended. Authorization checks for the ~198 mapped URLs now run in middleware (`process_view`) instead of inside view bodies, but produce the same allow/deny outcome.

## Test results

- Full Django unit-test suite (Rest Framework × 4 platform combos) — green.
- K8s deployment smoke (kubernetes 1.33.11, 1.35.4) — green.
- Full Selenium UI matrix — green.
- New `unittests/test_permissions_audit.py` exercises the URL-permission map for completeness.


valentijnscholten (commented):
> No more per-view decorators — one URL-permission map replaces ~200 decorator lines scattered across 26 view files.

I don't think I like this. It requires jumping back and forth to the mapping to find what mapping is in place and it increases the chance someone forgets a mapping altogheter or leaves an old mapping behind. Why is this being done?

Maffooch:
The hope is to make permissions easier the manage in a single location. This will likely make reviews easier as well since we only have to look out for a single point of failure for authorization rather than having it spread out

blakeaowens (approved):

mtesauro (approved):
Approved

(assumes the merge conflict is fixed first)

Jino-T (approved):

dojo/authorization/middleware.py
```diff
@@ -0,0 +1,50 @@
...
+            return
+
+        resolver_match = request.resolver_match
+        if resolver_match is None:
+            return
+
+        url_name = resolver_match.url_name
+        checks = URL_PERMISSIONS.get(url_name)
+        if not checks:
+            return
```
valentijnscholten, line 31: Could we switch to deny by default, or are there many valid urls/views that have no checks? Deny by default would help a lot in catching future changes where a permission check is forgotten.

valentijnscholten (commented):
see above

PR #14767, opened 2026-04-27
refactor: consolidate notifications into dojo/notifications/ package

Maffooch:
## Summary
- Consolidates the entire notification domain — models, admin, services (`helper.py`), signals, tasks, settings, context processors, forms, UI views/urls/templates, and API serializers/viewsets/urls — under `dojo/notifications/`, matching the canonical `dojo/url/` shape from `CLAUDE.md`.
- Backward-compatible re-exports preserved at every original location (`dojo/models.py`, `dojo/forms.py`, `dojo/utils.py`, `dojo/tasks.py`, `dojo/context_processors.py`, `dojo/api_v2/{serializers,views}.py`, `dojo/notifications/helper.py`).
- 69 templates relocated into `dojo/notifications/templates/notifications/` (62 channel `.tpl` files keep identical lookup names; 7 UI `.html` files re-namespaced from `dojo/...html` to `notifications/...html`). A new TEMPLATES `DIRS` entry resolves both trees from the package.
- Notification env-var schema (`NOTIFICATIONS_ENV_DEFAULTS` + `populate_settings`) extracted into `dojo/notifications/settings.py`; `settings.dist.py` consumes it. Celery beat task paths updated.
- No new migrations — string FK references in the relocated models are migration-equivalent to the original class references.

## Test plan
- [x] `python manage.py check` — clean (4 silenced)
- [x] `python manage.py makemigrations --check --dry-run` — `No changes detected`
- [x] `python manage.py spectacular --fail-on-warn` — clean
- [x] Pytest: `test_notifications`, `test_apiv2_notifications`, `test_jira_webhook`, `test_cleanup_alerts`, `test_rest_framework` (1325 tests; all pass, 610 version-skipped)
- [ ] CI passes
- [ ] Smoke-test in dev: trigger a finding save and confirm Slack/email/webhook channels still fire end-to-end


valentijnscholten (commented):
What does this PR mean for users that have customized their templates?

Maffooch:
They will need to change their templates - I can add some breaking changes to the release notes

blakeaowens (approved):

valentijnscholten:
My question was about the tpl files and it looks like no action is needed for people who modified those?

Maffooch:
Yeah most folks will likely not be impacted. The tpl files are changing locations, so if there folks with custom template downstream, and then update to the new version, the default tpl files will be read rather than the custom ones. So the action required for users that have modified those templates will be to move them to the new location.

valentijnscholten:
For me it's not clear from the added upgrade notes that users need to move their customized templates (and where).
I would make that explicit. And then remove all the text about the html templates as that is not a typical use case / something we actively support.

valentijnscholten (approved):

Jino-T (approved):

mtesauro (approved):
Approved

PR #14789, opened 2026-04-30
Remove 'safe' filter from description output

Maffooch:
This pull request makes a minor change to the `dojo/templates/notifications/alert/other.tpl` template by removing the `safe` filter from the `description` variable, which will prevent HTML in the description from being rendered as markup.


Jino-T (approved):

valentijnscholten (approved):

mtesauro (commented):
Approved

mtesauro (approved):
Approved

blakeaowens (approved):

PR #14797, opened 2026-05-03
perf(dupe-delete): use bulk_delete_findings + correlated subquery in async_dupe_delete

valentijnscholten:
## Summary
Usually the number of findings deleted in the asyn delete job is small, but let's use the optimized cascade_delete method anyway to make it efficient.

- Replaces the per-row `Finding.delete()` loop in `async_dupe_delete` with `bulk_delete_findings`, which uses raw SQL cascade (`cascade_delete_related_objects`) instead of Django's per-object Collector.
- Moves excess-duplicate selection fully into the DB via a correlated subquery that counts newer siblings per original (later date, or same date but higher id). A duplicate is excess when `newer_cnt >= max_dupes`, eliminating the Python-side per-original counting loop.
- Adds `select_related("test__engagement__product")` + `only("id", "test_id")` to the collection pass to avoid N+1 product lookups.
- The deleted findings are pure duplicates (`original_finding` is empty), so skipping `finding_delete()`'s duplicate-cluster reconfiguration is safe.

### DIFF
--- a/dojo/tasks.py
+++ b/dojo/tasks.py
@@ -7,12 +7,13 @@ from django.apps import apps
 from django.conf import settings
 from django.core.exceptions import SuspiciousOperation
 from django.core.management import call_command
-from django.db.models import Count, Prefetch
+from django.db.models import Count, IntegerField, OuterRef, Q, Subquery
+from django.db.models.functions import Coalesce
 
 from dojo.auditlog import run_flush_auditlog
 from dojo.celery import app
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

blakeaowens (approved):

Jino-T (approved):

mtesauro (approved):
Approved

PR #14806, opened 2026-05-04
endpoint: optimize __eq__ via product_id

valentijnscholten:
The `Endpoint` equality method compares the full `self.product` fields. This can trigger a query to get the product details and sometimes results in a 1+N query pattern, for example during previewing of deletes.

This PR switches to comparing `self.product_id`.

The equivalent Location models don't have an equality method.

### DIFF
--- a/dojo/models.py
+++ b/dojo/models.py
@@ -1799,14 +1799,11 @@ class Endpoint(models.Model):
 
     def __eq__(self, other):
         if isinstance(other, Endpoint):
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

Jino-T (approved):

dogboat (approved):

mtesauro (approved):
Approved

PR #14808, opened 2026-05-04
Remove gitpython

dogboat:
This PR removes gitpython as a requirement for Dojo, allowing us to remove a library we no longer need.

The only usage of gitpython was in a test file that was disabled for all our automatic test checks. The references to this test file (in comments) were removed as well.

Finally, for the debian Dockerfile, the git binary is removed after it is used to for building Python requirements.

[sc-13059]


Maffooch (approved):

mtesauro (approved):
Approved

Jino-T (approved):

blakeaowens (approved):

mtesauro (dismissed):
FYI: Need to do a bit more testing of this.

valentijnscholten (commented):
I'm fine with it but don't want to trump Matts request for testing :-)

Jino-T (approved):

mtesauro (approved):
Approved

PR #14810, opened 2026-05-05
cascade delete: prepare preview_only parameter

valentijnscholten:
Add a preview_only parameter that let's cascade_delete count objects instead/before deleting them.


Maffooch (approved):

blakeaowens (approved):

Jino-T (approved):

mtesauro (approved):
Approved

PR #14811, opened 2026-05-05
test: pin query-count baselines for tag inheritance hot paths

valentijnscholten:
## Summary

- Adds `unittests/test_tag_inheritance_perf.py` with `assertNumQueries` baselines on the six hottest tag inheritance paths.
- Numbers are pinned against current `dev` behavior so subsequent optimization PRs show up as concrete query-count reductions instead of relying on manual benchmarking.
- The class is intentionally temporary — pins move down as the redesign work lands, and the file can be deleted (or the assertions relaxed to upper bounds) once the targets are met.

## Hot paths covered

| Scenario | Pinned baseline |
|---|---:|
| Product tag add → 100 findings | 4758 |
| Product tag remove → 100 findings | 4540 |
| Create 100 findings under inheritance | 4025 |
| Create 1 finding under inheritance | 64 |
| Finding remove inherited tag (sticky re-add) | 44 |
| Finding add user tag (sticky path) | 17 |

Each test sets up its fixture outside the query-counting block, exercises one operation under `assertNumQueries`, and asserts a positive correctness check so query-count tightening cannot smuggle in a behavior bug.

## Notes

- Class uses `@override_settings(CELERY_TASK_ALWAYS_EAGER=True, CELERY_TASK_EAGER_PROPAGATES=True)` so dispatched Celery tasks execute synchronously in the test connection and their queries are captured. Product tag tests still call `propagate_tags_on_product_sync(product)` explicitly because the `m2m_changed` handler dispatches with `countdown=5` and `dojo_dispatch_task` skips foreground execution when no current user is set in tests.
- Existing tag inheritance suite (`unittests.test_tag_inheritance`, 36 tests) and bulk tag suite (`unittests.test_tag_utils_bulk`, 29 tests) continue to pass unchanged.
- No production code is modified in this PR.

### DIFF
--- a/unittests/test_tag_inheritance_perf.py
+++ b/unittests/test_tag_inheritance_perf.py
@@ -1,0 +1,253 @@
+"""
+Query-count baselines for tag inheritance hot paths.
+
+These tests pin the *current* number of SQL queries DefectDojo issues for each
+hot path under tag inheritance. They are tripwires: future redesign work
+(see /home/valentijn/.claude/plans/tag-inheritance-redesign.md) will reduce
+these numbers, and any regression that pushes them back up will fail loudly.
+
+Each test:
+  - sets up its fixture *outside* the query-counting block
+  - exercises one operation under ``assertNumQueries`` with the pinned baseline
+  - asserts a positive correctness check so query tightening cannot smuggle in
+    a behavior bug.
+"""
+from __future__ import annotations
+
+import logging
+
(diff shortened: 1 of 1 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

Jino-T (approved):

blakeaowens (approved):

PR #14812, opened 2026-05-05
perf(tags): bulk-propagate inherited tags + gate child post_save on create

valentijnscholten:
This is going to be multi stage optimization to keep the PRs a little smaller to ease review.

This PR basically adds another batch tag method to apply inherited tags more efficiently.

## Summary

- Replaces the per-row `obj.save()` loop in `dojo/product/helpers.py:propagate_tags_on_product_sync` with bulk SQL through the existing tag-utils helpers. For every child model (Engagement / Test / Finding / Endpoint / Location), reads current `inherited_tags` in one query, computes the per-child diff against the Product's tags, and applies adds/removes via `bulk_add_tag_mapping` and the new `bulk_remove_tags_from_instances` helper. Both `tags` and `inherited_tags` fields are kept in sync.
- Adds `bulk_remove_tags_from_instances` to `dojo/tag_utils.py`, symmetric to `bulk_add_tags_to_instances`. Decrements Tagulous tag counts to avoid drift, deletes through-model rows in one DELETE per tag-batch, invalidates prefetch caches.
- Gates the per-child `inherit_tags_on_instance` post_save handler on `created=True`. The previous behavior fired on every save (create OR update), repeatedly re-applying inherited tags to children whose tag state had not changed. Sticky enforcement on user-driven tag edits is unchanged (still handled by `make_inherited_tags_sticky` on `m2m_changed`).
- Lowers two pinned query-count baselines in `unittests/test_tag_inheritance_perf.py` to reflect the reductions.

## Query-count impact

| Hot path | Before | After | Reduction |
|---|---:|---:|---:|
| Product tag add -> 100 findings | 4758 | 91 | ~52x |
| Product tag remove -> 100 findings | 4540 | 53 | ~85x |
| Create 1 finding under inheritance | 64 | 64 | (Phase B target) |
| Create 100 findings under inheritance | 4025 | 4025 | (Phase B target) |
| Finding add user tag (sticky) | 17 | 17 | (Phase B target) |
| Finding remove inherited tag (sticky re-add) | 44 | 44 | (Phase B target) |

Sticky and child-creation paths are unchanged in this PR. Phase B targets those (centralized inheritance module + drop the duplicate `inherited_tags` TagField).

## Verification

- `unittests.test_tag_inheritance` - 36 tests pass (5 skipped for v3 feature flag)
- `unittests.test_tag_utils_bulk` - 29 tests pass
- `unittests.test_tag_inheritance_perf` - 6 tests pass with the lowered pins

## Behavior preserved

- Inheritance flags (system + per-product) gate propagation as before.
- `inherited_tags` field semantics unchanged - still a duplicate `TagField` kept in lock-step with `tags`. Phase B replaces it with a JSON column.
- Sticky enforcement (re-add inherited tag if a user removes it from a child) still runs through the existing `make_inherited_tags_sticky` m2m_changed handler.
- Locations with multiple linked products still receive the union of all related products' tags.

### DIFF
--- a/dojo/product/helpers.py
+++ b/dojo/product/helpers.py
@@ -1,5 +1,6 @@
 import contextlib
 import logging
+from collections import defaultdict
 
 from django.conf import settings
 from django.db.models import Q
@@ -7,6 +8,7 @@ from django.db.models import Q
 from dojo.celery import app
 from dojo.location.models import Location
 from dojo.models import Endpoint, Engagement, Finding, Product, Test
+from dojo.tag_utils import bulk_add_tag_mapping, bulk_remove_tags_from_instances
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


Maffooch (approved):

blakeaowens (approved):

Jino-T (approved):

mtesauro (approved):
Approved

PR #14827, opened 2026-05-06
perf(tags): centralize tag inheritance + replace signal disconnect with batch context manager (Phase B)

valentijnscholten:
> [!IMPORTANT]
> Depends on:
> - #14812

This PR lands the thread-safety fix. Instead of diconnectings signas we introduce a context manager that disables tag inheritance.

## Stage 1 (this commit) — thread-local batch context

Adds `dojo/tag_inheritance.py` with:
- `batch_mode()` context manager (thread-local depth counter)
- `is_in_batch_mode()` predicate

While inside a batch, `make_inherited_tags_sticky` (`m2m_changed`) early-returns. Calling code is responsible for applying inheritance in bulk.

Replaces the previous pattern in `dojo/importers/location_manager.py:556`:

```python
disconnected = signals.m2m_changed.disconnect(
    make_inherited_tags_sticky, sender=Location.tags.through,
)
try:
    ...
finally:
    if disconnected:
        signals.m2m_changed.connect(...)
```

`Signal.disconnect` mutates Django's process-global receiver list. While disconnected, every thread/greenlet in the same process loses sticky enforcement. Safe only under Celery `--pool=prefork` + single-threaded gunicorn; broken under `--pool=threads|gevent|eventlet`, gunicorn `--threads >1`, or ASGI threadpools. Also fragile on hard process exit mid-import (handler stays disconnected for the rest of the process lifetime).

The new context lives in `threading.local()`: each thread has its own depth counter, the signal handler stays globally connected, and the suppression decision is per-thread. No global mutation, no reconnect hazard.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -129,7 +129,7 @@ services:
     volumes:
       - defectdojo_postgres:/var/lib/postgresql/data
   valkey:
-    image: valkey/valkey:9.0.3-alpine@sha256:e1095c6c76ee982cb2d1e07edbb7fb2a53606630a1d810d5a47c9f646b708bf5
+    image: valkey/valkey:9.0.4-alpine@sha256:d1cc70645bbcef743615463a2fa4616e841407545e18f560aed0c49671a90147
     volumes:
       # we keep using the redis volume as renaming is not possible and copying data over
       # would require steps during downtime or complex commands in the intializer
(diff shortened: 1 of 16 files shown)


Maffooch (approved):

blakeaowens (approved):

mtesauro (approved):
Approved

dogboat (approved):
Thanks for fixing that!

PR #14838, opened 2026-05-08
Use a dedicated permission class for BurpRawRequestResponseViewSet

Maffooch:
## What

The top-level `/api/v2/request_response_pairs/` viewset previously reused `UserHasFindingRelatedObjectPermission`, a `BaseRelatedObjectPermission` subclass whose `has_permission` returns `True`. That class is shaped for `@action(detail=True, …)` endpoints where DRF resolves the parent `Finding` from the URL and `has_object_permission` runs against it. On a top-level `POST` there is no parent object resolved yet, so the create flow only ran `has_object_permission` against the not-yet-saved row and skipped any check on the client-supplied `finding` foreign key.

This PR introduces `UserHasBurpRawRequestResponsePermission`, mirroring the pattern already used by `UserHasFindingPermission`, `UserHasProductPermission`, and the other parent-keyed viewsets:

- `has_permission` calls `check_post_permission(request, Finding, "finding", Permissions.Finding_Edit)` on `POST`, which fetches the referenced finding and rejects with `403`/`404` unless the user has `Finding_Edit`. Missing `finding` is rejected with `ParseError`.
- `has_object_permission` dereferences `obj.finding` for retrieve/update/delete so list/detail/PUT/PATCH/DELETE behavior is unchanged.

The other 11 callers of `UserHasFindingRelatedObjectPermission` are all `@action(detail=True, …)` decorators where the URL-resolved finding makes the existing class behave correctly, so the base class is left as-is.

## Test plan

Added `unittests.test_rest_framework.RequestResponsePairsAuthzTest` covering:

1. Positive control — admin POST returns `201` and the row count increases.
2. A user with no product/product_type/global membership receives `404` for `GET /api/v2/findings/7/` and a 4xx for `POST /api/v2/request_response_pairs/` with that finding ID; the row count is unchanged.
3. POST without `finding` returns 4xx (the `ParseError` path in `check_post_permission`).

```bash
docker compose -f docker-compose.yml -f docker-compose.override.unit_tests.yml run --rm \
  -e DJANGO_SETTINGS_MODULE=dojo.settings.settings \
  --entrypoint '' uwsgi \
  python3 manage.py test \
  unittests.test_rest_framework.RequestResponsePairsAuthzTest -v2 --keepdb
```

### DIFF
--- a/dojo/api_v2/permissions.py
+++ b/dojo/api_v2/permissions.py
@@ -455,6 +455,23 @@ class UserHasFindingNotePermission(BaseRelatedObjectPermission):
     }
 
 
+class UserHasBurpRawRequestResponsePermission(permissions.BasePermission):
+    def has_permission(self, request, view):
+        return check_post_permission(
+            request, Finding, "finding", Permissions.Finding_Edit,
+        )
+
+    def has_object_permission(self, request, view, obj):
+        return check_object_permission(
+            request,
+            obj.finding,
+            Permissions.Finding_View,
+            Permissions.Finding_Edit,
+            Permissions.Finding_Edit,
+            Permissions.Finding_Edit,
+        )
+
+
 class UserHasImportPermission(permissions.BasePermission):
     def has_permission(self, request, view):
         # permission check takes place before validation, so we don't have access to serializer.validated_data()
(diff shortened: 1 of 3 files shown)


valentijnscholten (approved):

mtesauro (approved):
Approved

Jino-T (approved):

paulOsinski (approved):

PR #14839, opened 2026-05-08
feat(importers): apply import-time tags per batch before post-processing, do not tag old findings

valentijnscholten:
## Summary

- Import-time tags (`apply_tags_to_findings` / `apply_tags_to_endpoints`) were previously applied after all findings were processed and post-processing tasks (deduplication, rules) had already been dispatched — those tasks could not see the tags on the findings.
- Tags are now applied per batch inside `process_findings()`, immediately after parser tags and location persistence, before `dojo_dispatch_task()` — the same pattern already used by `bulk_apply_parser_tags`.
- Closed findings are intentionally excluded: they are absent from the current report and should not receive the import tags.
- Renames the reimporter's ambiguous `batch_findings` (match-batch slice of unsaved parser objects) to `unsaved_findings_batch`, freeing `batch_findings` for the new dedupe-batch accumulator and fixing a latent iterator-mutation bug where appending to the iterated list caused extra loop iterations.
- Consolidates `apply_import_tags()` into `apply_import_tags_for_batch()`, removing a redundant `Finding.objects.filter()` re-query and an unused `Iterable` import.
- Updates performance test query counts accordingly (each import loses 1 query from the removed post-loop `apply_import_tags` call).

### DIFF
--- a/dojo/importers/base_importer.py
+++ b/dojo/importers/base_importer.py
@@ -1,7 +1,6 @@
 import base64
 import logging
 import time
-from collections.abc import Iterable
 
 from django.conf import settings
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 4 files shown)


Maffooch (approved):

dogboat (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

PR #14841, opened 2026-05-08
:zap: speed up migrate_endpoints_to_locations (~14× fewer queries)

Maffooch:
This is a perf bugfix to an existing one-shot data migration; no new features or surface area, no new dependencies, single file changed. The migration is currently producing latently-incorrect data in addition to being unusably slow on real instances (16+ hours for ~650 endpoints / 22k findings — see Description), so I went ahead.

**Description**

`dojo/management/commands/migrate_endpoints_to_locations.py` is a one-shot migration that maps each `Endpoint` → `URL`/`Location`, each `Endpoint_Status` → `LocationFindingReference`, and each (endpoint→product, finding→product) pair → `LocationProductReference`. On a real instance with 650 endpoints and 22k findings, it had produced **80 locations in 2 hours** (~90 s/endpoint) and was projected at **~16 hours** total. Per-endpoint queries on a 50-endpoint / 1,000-finding local benchmark came in at **613**.

This PR keeps changes inside the management command itself (no edits to shared `Location` model methods) and reduces per-endpoint queries to **44** and per-endpoint wall-clock to **17 ms** on the same benchmark — a **13.9× query reduction / 14.1× wall-clock improvement**. Idempotent re-runs validated; strict accuracy checks pass.

The work was applied as five logically-distinct changes; each row of the table below was measured independently against a freshly-reset migration target so the gain attribution is honest:

| Step | Wall-clock | ms/ep | Queries/ep | Notes |
|---|---:|---:|---:|---|
| Baseline (instrumented, no logic change) | 12.00 s | 240 | 613 | 14 product-ref status warnings (pre-existing bug) |
| + `select_related`/`prefetch_related` | 10.63 s | 213 | 528 | Eliminates lazy joins on `tags`, `endpoint_meta`, `status_endpoint`, and `finding → test → engagement → product`/`mitigated_by` |
| + `tags.add(*names)` splat | 10.08 s | 202 | 507 | `tags` phase −61% |
| + `DojoMeta.bulk_create(ignore_conflicts=True)` | 10.24 s | 205 | 498 | `meta` phase −67% — `DojoMeta.unique_together = (location, name)` makes ignore_conflicts equivalent to the previous `get_or_create` |
| + `bulk_create` LFR/LPR + in-memory product status | **0.85 s** | **17** | **44** | Bypasses `BaseModel.save → full_clean()` validate_unique queries AND the `inherit_tags_on_linked_instance` post_save signal that fires `all_related_products` (a finding/test/engagement/product join) on every save |

Where the savings actually came from, surfaced by the `--query-count` flag on a single endpoint with 20 statuses:
- ~3 `validate_unique` queries × 2 saves per status = ~120 redundant queries / endpoint just from `full_clean`
- 1 `all_related_products` + 1 `system_settings` per save × ~40 saves per endpoint = ~80 redundant queries from the inherit-tags post_save signal
- 3 lazy joins (`finding.test.engagement.product`) × 20 statuses = 60 queries
- The `.exists() → .get() → update_or_create` triple in `associate_with_finding`/`associate_with_product` — 2 redundant queries per status

`bulk_create(ignore_conflicts=True)` for the reference rows skips all of those at once.

#### One Django gotcha worth flagging in review

`LocationFindingReference.created` is `auto_now_add=True`. The original migration set `created` to the source `Endpoint_Status.date` via a post-save UPDATE, which works because `auto_now_add` only fires on the initial INSERT (`add=True`). With `bulk_create` we don't get a post-save UPDATE — and Django's `SQLInsertCompiler.pre_save_val` *does* call `Field.pre_save(add=True)` even from `bulk_create`, so `auto_now_add` would overwrite our explicit value with `now()`. The PR uses a small `_suspend_auto_now_add` context manager that toggles `auto_now_add=False` on the field during the bulk write, then restores it. Single-process management command — no thread-safety concerns.

#### Behavioral diffs you should know about

1. **`LocationProductReference.status` is now correct** (semantic improvement). The previous code called `Location.associate_with_product(...)`, which short-circuits if the (location, product) pair already exists — never updating the status when a later Active finding comes in for the same product. With the seed's 70/30 Active/non-Active distribution, ~28% of product refs landed on `Mitigated` when the logical truth was `Active`. The new code derives product status in-memory across all of an endpoint's finding statuses (`Active` iff any of them is Active, else `Mitigated`) and `bulk_create`s the row. The strict mode of the local accuracy verifier explicitly catches this — baseline fails strict, this PR passes strict.

2. **Inherited product tags via `post_save` signals do NOT fire**. `bulk_create` skips signals, so `inherit_tags_on_linked_instance` (`dojo/tags_signals.py:65-68`) does not run during the migration. **For deployments where any product has `enable_product_tag_inheritance=True` (or the system-wide setting is on), inherited product tags will not be propagated onto migrated `Location`s by this command.** Two reasonable mitigations, both deferred to a follow-up:
   - Add a one-time post-bulk pass that walks the migrated Locations and calls `inherit_tags()` (cheap — one query per location, paid once instead of per-LFR).
   - Use the existing `_bulk_inherit_tags` helper at `dojo/importers/location_manager.py:486`, which is designed for exactly this scenario.

   I'd be happy to land either as a follow-up PR if you confirm your prod has tag inheritance enabled. The benchmark seed does not exercise this path.

#### Things intentionally NOT changed

- **`Location.associate_with_finding` / `associate_with_product` are untouched.** They still have the `.exists() → .get() → update_or_create` pattern and the first-write-wins behavior on product status. The migration just stops calling them on its hot path. Other callers (e.g., importers) are unaffected.
- **No DB migration.** Pure code change, no schema impact.
- **No `--start-from <id>` / `--skip-validation` flags.** Considered during planning, dropped to keep blast radius small. Easy follow-ups if the prod run reveals a need.
- **No verbose-per-endpoint logging mode.** The default `Migrated X/Y (z%) — N ep/sec — ETA …` line is enough for ops; `--benchmark` covers diagnostic depth.

**Test results**

Verified locally on the docker dev stack with `DD_V3_FEATURE_LOCATIONS=True` against a representative seed (5 products × 10 endpoints × 20 findings, with the four non-Active status flag distributions exercised: 70% Active / 15% Mitigated / 5% RA / 5% FP / 5% OOS):

- Strict accuracy verifier passes after the migration: counts (URLs, Locations, LFR, LPR, location-DojoMeta), per-LFR field equality (`status`, `created`, `audit_time`, `auditor`), endpoint→location tag subset, and DojoMeta `(location, name)` parity all agree with source data.
- Idempotent: running the migration a second time produces no new rows and no errors; verifier still passes.
- `python manage.py check` clean. `ruff check` clean.
- The verifier and seed scripts used during development are local-only scratch tools; they are not in this PR (intentionally — single-file diff).

I have not exercised this on a 22k-finding instance directly. The 50/1,000 numbers above are linear-scaling from a per-endpoint perspective; if anyone has a prod-shape instance handy, validation there would be welcome.

**Documentation**

No user-facing docs to update — this is a one-shot internal data migration command. The new `--benchmark` / `--query-count` / `--batch-size` / `--progress-every` flags are documented via `argparse` `help=` strings.

### DIFF
--- a/dojo/management/commands/migrate_endpoints_to_locations.py
+++ b/dojo/management/commands/migrate_endpoints_to_locations.py
@@ -2,6 +2,7 @@ import contextlib
 import datetime
 import logging
 import time
+from collections import defaultdict
 
 from django.core.management.base import BaseCommand
 from django.db import connection
@@ -10,7 +11,7 @@ from django.utils import timezone
 
 from dojo.location.models import Location, LocationFindingReference, LocationProductReference
 from dojo.location.status import FindingLocationStatus, ProductLocationStatus
-from dojo.models import DojoMeta, Endpoint, Endpoint_Status
+from dojo.models import DojoMeta, Endpoint, Endpoint_Status, Product
 from dojo.url.models import URL
 
 logger = logging.getLogger(__name__)
@@ -97,6 +98,29 @@ class Command(BaseCommand):
         if self.benchmark:
             self.timings[phase] += time.perf_counter() - t0
             self.counts[phase] += 1
+
+    # -- Tag inheritance bookkeeping -----------------------------------------
+
+    def _track_product_location(self, product: Product, location: Location) -> None:
+        """
+        Record a (product, location) pair for the post-migration tag inheritance pass.
+
+        The migration creates locations that may be linked to multiple products
+        (via the endpoint's own product and via each finding's product). We
+        collect every contributing product per location so the post-pass can
+        call ``LocationManager(product)._bulk_inherit_tags(locations)`` once
+        per product group — covering the case where a location is shared
+        across products with differing ``enable_product_tag_inheritance``
+        flags (the helper short-circuits via its own diff check on repeat
+        visits, so redundancy is safe).
+        """
+        if product is None or product.id is None:
+            return
+        if location is None or location.id is None:
+            return
+        self.locations_by_product_id[product.id].add(location.id)
+        self.product_obj_by_id.setdefault(product.id, product)
+        self.location_obj_by_id.setdefault(location.id, location)
 
     # -- Migration logic --------------------------------------------------
 
@@ -197,6 +221,10 @@ class Command(BaseCommand):
             if finding is None:
                 continue
             product = finding.test.engagement.product
+            # Track this contributing product for the post-migration tag
+            # inheritance pass (covers the case where a finding's product
+            # differs from endpoint.product).
+            self._track_product_location(product, location)
             status = self._convert_endpoint_status_to_string_status(endpoint_status)
             # Endpoint_Status.date is a Date; the original code persisted
             # the same midnight-aware datetime in a post-save UPDATE. We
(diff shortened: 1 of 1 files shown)


valentijnscholten (approved):
Sounds good. But wouldn't it make sense to also add the step to inherit the tags as it's cheap / easy?

blakeaowens (approved):

mtesauro (approved):
Approved

dogboat (approved):

PR #14868, opened 2026-05-13
Validate user-supplied URLs in tool API clients

Maffooch:
## Summary
Five tool API clients constructed their underlying HTTP session without validating the configured server URL. Three of them (`SonarQube`, `Edgescan`, `Vulners`) accept the URL from a `Tool_Configuration` set by an administrator; `Bugcrowd` and `Cobalt` use a hardcoded public endpoint. Misconfigured or unreachable URLs (typo'd hostnames, leftover internal addresses, copy-pasted private IPs) produced raw connection errors instead of a clean validation message, and there was no defense-in-depth on the session itself.

This change brings each client in line with the established `dojo/tools/risk_recon/api.py` pattern:

- `dojo/tools/api_sonarqube/api_client.py` — call `validate_url_for_ssrf` on `tool_config.url`; swap `requests.Session()` → `make_ssrf_safe_session()`.
- `dojo/tools/api_edgescan/api_client.py` — same; the per-call `requests.get(...)` is replaced with `self.session.get(...)` so the same safe session handles every request.
- `dojo/tools/api_vulners/api_client.py` — validate `tool_config.url` when supplied (the `vulners` library handles its own transport, so no session swap).
- `dojo/tools/api_bugcrowd/api_client.py` and `dojo/tools/api_cobalt/api_client.py` — URL is hardcoded, so just swap to `make_ssrf_safe_session()` for defense-in-depth.

`unittests/test_tool_api_clients_ssrf.py` covers each client. Two existing SonarQube fixtures (`unit_sonarqube_toolConfig1.json` / `toolConfig2.json`) used `http://localhost/` and had to move to a publicly routable URL to keep the existing importer tests working; the three assertion strings in `test_api_sonarqube_importer.py` that depended on the literal `http://localhost/` prefix and the dummy hostname in `test_api_sonarqube_parser.py` were updated accordingly.

## Test plan
- [ ] \`python manage.py test unittests.test_tool_api_clients_ssrf\`
- [ ] \`python manage.py test unittests.tools.test_api_sonarqube_importer unittests.tools.test_api_sonarqube_parser unittests.test_api_sonarqube_updater\`
- [ ] \`python manage.py test unittests.tools.test_api_edgescan_importer unittests.tools.test_api_cobalt_importer unittests.tools.test_api_bugcrowd_importer unittests.tools.test_api_vulners_parser\`
- [ ] \`python manage.py check\`
- [ ] \`ruff check . --fix\`

### DIFF
--- a/dojo/fixtures/unit_sonarqube_toolConfig1.json
+++ b/dojo/fixtures/unit_sonarqube_toolConfig1.json
@@ -5,7 +5,7 @@
     "fields": {
       "name": "SQ1",
       "description": null,
-      "url": "http://localhost/",
+      "url": "https://8.8.8.8/",
       "tool_type": 1,
       "authentication_type": "API",
       "extras": null,
--- a/dojo/fixtures/unit_sonarqube_toolConfig2.json
+++ b/dojo/fixtures/unit_sonarqube_toolConfig2.json
@@ -5,7 +5,7 @@
     "fields": {
       "name": "SQ2",
       "description": null,
-      "url": "http://localhost/",
+      "url": "https://8.8.8.8/",
       "tool_type": 1,
       "authentication_type": "API",
       "extras": null,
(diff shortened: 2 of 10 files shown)


valentijnscholten (commented):
commented via DM

PR #14881, opened 2026-05-15
perf(watson): prefetch relations + force async indexing

valentijnscholten:
## Summary

Watson index updates were (in OS) done in the background. Still they were slow and inefficient as there is no prefetching of relations. So Waton will happily execute N x M x X x ... queries. They were also accumulating all models of a long running request in memory. 

- **Watson indexer N+1 fix.** Watson's `SearchAdapter._resolve_field` walks `__`-separated relation paths via per-instance `getattr`, triggering one query per FK hop per object during async indexing. For `Finding` (`test__engagement__product__name` + `jira_issue__jira_key`) and `Vulnerability_Id` (`finding__test__engagement__product__name`) on a 1000-row batch this adds thousands of extra queries per task.
- **New helper `dojo/utils_watson_prefetch.py`** auto-derives `select_related` / `prefetch_related` paths from each adapter's `fields`/`store` by walking `model._meta`, then applies them in `update_watson_search_index_for_model`. On any error we log loudly and fall back to the plain queryset so indexing still completes.
- **Toggle:** ``DD_WATSON_INDEX_PREFETCH_ENABLED`` env / `WATSON_INDEX_PREFETCH_ENABLED` setting, default `True`.
- **`force_async=True` dispatch flag.** `dojo_dispatch_task` / `we_want_async` now accept `force_async=True`, which keeps a task in the background even when the user has `block_execution=True` or `sync=True` is also passed. Wired into the async watson indexer middleware — index updates are slow and never need to be synchronous from the user's perspective.
- flush intermediately after reaching treshold.

## Why

- Bulk imports / reimports trigger async watson indexing via `AsyncSearchContextMiddleware`. Even on the background path the indexing was slow because Watson re-fetched every relation per object. This change cuts indexing queries from O(objects × FK depth) to O(1) per batch on the FK chain.
- `force_async` decouples \"this task should run in the background\" from \"this user wants foreground execution\" — the watson indexer is a clear case where the caller knows better than the user preference.

## Notes

- A separate naming follow-up was discussed (`sync=True` → `force_sync=True` for symmetry with `force_async`), but deferred to a dedicated PR since it touches ~84 sites


mtesauro (approved):
Approved

Maffooch (approved):

Jino-T (approved):

blakeaowens (approved):

PR #14931, opened 2026-05-28
Include analysis.detail from Dependency Track FPF in finding description

webdevred:
## Show Dependency Track audit detail in finding description

When Dependency Track pushes findings via the Finding Packaging Format, analysts can attach free-text notes to each finding through the `analysis.detail` field. This field was already parsed and available in the FPF document but was silently ignored, so auditors working in DefectDojo had no way to see those notes without going back to Dependency Track.

This appends the audit detail to the finding description under an "Audit Detail:" label when present. Findings without audit detail are unaffected. The change requires Dependency Track 4.14.0 or later, which is when `analysis.detail` was added to the FPF output (see companion PR in the Dependency Track repo).

Related to https://github.com/DependencyTrack/dependency-track/pull/6181


dojo/tools/dependency_track/parser.py
```diff
@@ -154,6 +154,9 @@ def _convert_dependency_track_finding_to_dojo_finding(self, dependency_track_fin
         analysis = dependency_track_finding.get("analysis")
         is_false_positive = bool(analysis is not None and analysis.get("state") == "FALSE_POSITIVE")
 
+        if analysis is not None and analysis.get("detail"):
+            vulnerability_description = f"Audit Detail: {analysis['detail']}\n\n{vulnerability_description}"
```
valentijnscholten: Wouldn't it make more sense to put the audit detail after the description? Or maybe even in the impact field?

valentijnscholten (commented):
Can you rebase against the `bugfix` branch?

dojo/tools/dependency_track/parser.py
webdevred: Good point, happy to discuss. My reasoning for putting it first: the audit detail is analyst-added context specific to this finding in your environment, whereas the vulnerability description is generic upstream text. When triaging, the analyst note is usually the more immediately relevant piece. Open to changing it if you feel strongly though.

On the impact field: that field has a specific semantic meaning (consequence of exploitation), so putting analyst workflow notes there felt wrong.

Also, I'll rebase against `bugfix` as requested.

valentijnscholten: Still feels as if the audit note should be at the end

webdevred: Fair enough, I moved it to the end.

webdevred:
Both the v4.14 and the v5 PRs to DependencyTrack are merged into their respective branches :partying_face:

valentijnscholten (approved):

Maffooch (approved):
Confirmed that description is not part of default dedupe fields

Jino-T (approved):

mtesauro (approved):
Approved

Issue #1, opened 2015-03-13
Same name as commercial software


r3dey3:
Just wanted to let you know the name you chosen is the same as a commercial issue tracking software - http://www.seapine.com/testtrack/issue-management

cneill:
Hi r3dey3, thanks for bringing this to our attention. We're doing some brainstorming now to figure out how best to resolve this confusion.

devGregA:
![homer](https://cloud.githubusercontent.com/assets/4741312/6646419/b63abeae-c992-11e4-8010-6ebf9d2ea1a5.gif)

devGregA:
![homer2](https://cloud.githubusercontent.com/assets/4741312/6646439/d4eab66a-c992-11e4-8ad3-f01eef9a6fc4.jpg)

rgbkrk:
You should call it tRacker.

> /me bows out

:bow:

grendel513:
We are in the process of rebranding.  A new name has been chosen and will soon be reflected in all content as well as logos.

grendel513:
The name change and rebranding has been completed.  We encourage everyone to rebase as the v1.0 release has been tagged.

Issue #14890, opened 2026-05-19
Do we have a API feature to revoke DefectDojo API Token which are active


atharva1051:
## :warning: Note on feature completeness :warning:

We are narrowing the scope of acceptable enhancements to DefectDojo. Learn more here:
https://github.com/DefectDojo/django-DefectDojo/blob/master/readme-docs/CONTRIBUTING.md

**Is your feature request related to a problem? Please describe**
Currently, DefectDojo API tokens, once generated, appear to remain active indefinitely unless the associated user is deleted or the token is manually removed via the UI (if such an option exists). There is no readily available API endpoint to programmatically revoke or invalidate an active API token. This poses a security risk, as compromised or deprecated tokens cannot be immediately revoked via automation, requiring manual intervention which can be slow and error-prone, especially in large or automated environments.

**Describe the solution you'd like**
As a security administrator or an automation engineer, I want an API endpoint to revoke or invalidate a specific DefectDojo API token so that I can programmatically manage and secure API access, ensuring that compromised or no-longer-needed tokens can be immediately disabled without manual intervention. Also and expiry date for tokens created.

This could involve:
*   A `DELETE` endpoint for `/api/v2/api-tokens/{token_id}` or similar, requiring appropriate administrative permissions.
*   An endpoint to list active tokens for a user or globally (with proper authorization) to identify the `token_id` for revocation.

**Describe alternatives you've considered**
*   **Manual revocation via UI:** This is not scalable for automated processes or in emergency situations where immediate revocation is needed across multiple tokens.
*   **Deleting the user:** This is an extreme measure that impacts all of the user's activities and is not suitable for revoking a single token.
*   **Token expiration:** Tokens created do not have expiration date. Even this can be a good enhancement.

**Additional context**
Implementing an API for token revocation would significantly enhance the security posture of DefectDojo deployments by providing granular control over API access and enabling faster response to potential security incidents. This aligns with best practices for API security and token management.

svader0:
@mtesauro As we discussed, I'll be looking into this one :)

PR #14932, opened 2026-05-29
feat(api-tokens): API token expiry and revoke-by-key endpoint

svader0:
**This PR addresses issue/feature request #14890.** 

**NOTE**: This PR description has been updated away from the original commit, which was significantly different.

**Description**

Currently, DefectDojo has no mechanism for managing API tokens via the API. In case a token gets leaked or needs to be revoked, the only existing solution is to call `reset_api_token`, which simply rotates the token. Superusers also have no way to see what tokens exist, who owns them, or when they were issued. This PR is an attempt to close the gap a little.

This PR introduces a per-user token expiry. A superuser can set a `token_expiry` on a specific user via the existing `user_contact_infos` endpoint. Expired tokens are rejected at authentication time. In addition to this, a new instance-wide environment variable `DD_API_TOKEN_DEFAULT_EXPIRY_DAYS` (talk about a mouthful...) applies a default lifetime to all generated tokens. The default is 0, a.k.a. no expiry, so there should be no issues with existing deployments unless they choose to opt into the change.

This PR also introduces one new endpoint `POST /api/v2/api-tokens/revoke/`, where you supply a key via JSON payload, and it gets revoked. Currently, you can only rotate a user's key by user lookup, you can't eradicate a specific key itself. This would solve the use case where, "uh oh!", a token gets leaked, and I want to eradicate it immediately and programmatically without tracking down its owner and rotating it.

**Documentation**

  `docs/content/automation/api/api-v2-docs.md` updated with new endpoint, expiry behavior, and the `DD_API_TOKEN_DEFAULT_EXPIRY_DAYS` variable.

**Checklist**


valentijnscholten:
Thank you for the PR. At first glance there seem to be various functionalities in this PR. I can imagine having an expiry datetime/period for tokens is good to have and not to hard to maintain. But I'm unsure of the other items like allowing super users to manage other users tokens etc. Do you have good example of use cases for each feature introduced by the PR?

svader0:
Thanks for taking a look! I definitely agree about the expiry date/time period. After thinking about it for a while, it seems that the practical use-case of the token-management stuff was a little too overstated, so I made a few changes to slim it down.

**Token expiry**: Like you said, it's fairly easy implementation and maintenance, and having an expiration date on API keys is pretty standard behavior. If a key were to get leaked, having an expiration date will reduce risk. I think this should stay as-is.

**Token Management**: The thinking was that if a token was compromised, as a superuser you may want to revoke it immediately without issuing a replacement. I was originally following the train of logic set about by the issue I was responding to, where you'd want a superuser to be able to have a list of existing tokens and then be able to revoke them selectively. But I think that in trying to solve that problem, I accidentally made things a little convoluted here, so I went ahead and simplified the solution.

Now there is one new endpoint in this PR, `POST /api/v2/api-tokens/revoke/`, where you supply a key via JSON payload, and it gets revoked. Currently, you can only rotate a user's key by user lookup, you can't eradicate a specific key itself. With this implementation, we avoid the API bloat of needing to look up all existing tokens, filter, and revoke like I had previously. 
This would solve the use case where, "uh oh!", a token gets leaked, and I want to eradicate it immediately and programmatically without tracking down its owner and rotating it.

If you think these two additions are too distinct to be in one PR, I am happy to open up a second and split them up.

valentijnscholten:
So this PR is hard to review not because of the implementation, but because it needs some careful thinking about API design, features, security, etc. My feeling is that it cannot be merged as-is. I have asked Claude also to review it and that review below contains some of my concerns. My suggestion would be to switch to django-restknow (or another library if it exists) and do it right. Might be a good time right now while we're "in transition" to v3. I'm going to have to ask @Maffooch and @mtesauro to lean in here as well, and will discuss it with them.

valentijnscholten:
Thanks for working on this — token lifecycle management is a real gap. I reviewed the full diff; some of this is blocking-ish (correctness gaps), and there's a bigger-picture question at the end about foundation.

## Correctness

What's solid:
- `ExpiringTokenAuthentication` cleanly wraps DRF `TokenAuthentication` and is swapped in via `DEFAULT_AUTHENTICATION_CLASSES`.
- `RevokeApiTokenView` is gated by `IsSuperUserOrGlobalOwner`, and `user_contact_infos` is `IsSuperUser`-gated, so the "superuser-only" claims actually hold. `UserContactInfoSerializer` uses `fields = "__all__"`, so `token_expiry` is genuinely PATCH-able there.

Gaps:

1. **Default expiry is bypassed on 2 of the 3 token-creation paths.** It's only applied inside `reset_token_for_user`. But `api_v2_key` auto-creates a token on **GET** (`Token.objects.create(...)`), and `api-token-auth` (`obtain_auth_token`) does `get_or_create` — both with no expiry. So after a superuser **revokes** a leaked token, the owner can just reload `/api/key-v2` (or hit `api-token-auth`) and get a fresh **non-expiring** token, even with `DD_API_TOKEN_DEFAULT_EXPIRY_DAYS=90`. The "applies a default lifetime to all generated tokens" claim doesn't hold, and revoke isn't durable. All token creation should funnel through one place that applies the default.

2. **Expiry is decoupled from the token** (stored on `UserContactInfo`, not the `Token`). Rotation has to remember to update it, and `reset_token_for_user` clobbers any superuser-set per-user expiry back to the instance default (documented, but a confusing wart). Any token created outside that one function has no expiry — see #1.

3. **401 vs 403:** `AuthenticationFailed` normally yields 401; this returns 403 only because a non-token authenticator happens to be first in the list so `authenticate_header` is `None`. Works today but is order-dependent/fragile (and 401 is arguably more correct for an expired credential).

4. **Migration collision:** `0269_usercontactinfo_token_expiry` depends on `0268`, but dev already has `0269_normalize_blank_finding_components`. Needs renumbering to `0270` before it'll merge.

5. **No audit trail / owner notification** on revoke — for an incident-response feature that's a notable omission.

## Missing features
- The PR description still advertises a `/api/v2/api-tokens/` list/retrieve viewset for superuser-wide token visibility, but that's **not in the diff** (only `revoke/` + expiry). Worth updating the description.
- No per-token expiry and no multiple-tokens-per-user (both blocked by DRF's one-token-per-user model).
- Tokens remain stored in **plaintext** in the DB (DRF default); the revoke endpoint looks them up by raw key. Expiry without hashed storage leaves the larger exposure untouched.

## Foundation: should this be built on django-rest-knox?

This is the bigger question. [django-rest-knox](https://github.com/jazzband/django-rest-knox) provides natively almost everything being hand-rolled here: **per-token expiry** (on the token itself, no `UserContactInfo` decoupling), **multiple tokens per user**, **hashed token storage** (a DB leak doesn't expose usable tokens), and built-in logout/revoke endpoints + an enforcing auth class.

The catch is that adopting it is a **breaking, multi-release migration**, not a tweak to this PR: DefectDojo's whole ecosystem uses DRF's `Token <key>` format, so a switch changes token format/length and the `api-token-auth` flow and breaks every existing integration unless both auth backends run in parallel during a deprecation window with token re-issuance. That's an epic, well beyond this PR's scope.

So my suggestion is **not** to retrofit Knox here. Either:
- **(a)** land a corrected incremental version — funnel all token creation through one default-expiry path (fixes #1), drop the "all tokens" overclaim, add an audit entry, renumber the migration — accepting per-user/single-token semantics for now; or
- **(b)** if multi-token / hashed storage / per-token expiry is the actual goal, treat Knox as a dedicated foundation epic and supersede this PR.

The decoupled-expiry and bypass-path issues above are essentially symptoms of forcing per-token semantics onto DRF's single-plaintext-token model — which is exactly what Knox exists to replace. Happy to help scope the Knox migration if the team wants to go that route.

svader0:
Makes a lot of sense. You're right in that this is more of a shotgun-surgery type of fix; I wasn't quite sure how drastic we wanted to go. Happy to help with implementing a more robust solution depending on what the higher-ups decide.

Maffooch:
@valentijnscholten thanks for the careful consideration here. I think you're right that this would be a better API v3 ONLY change. There have been no expirations on API v2 for many years, so this would be a big shift.

I would feel better if this new mechanism only impacted v3 so that the correct expectation can be set from the get go

blakeaowens (approved):

Maffooch (approved):

devGregA (approved):

PR #14934, opened 2026-05-31
Preserve verified flag when promoting duplicate to new original

### LINKED ISSUE #14911: Deleting an engagement can promote duplicate findings without verified flag, preventing Jira creation
**Bug description**
Deleting an engagement that contains the current primary/original finding can promote or reactivate a duplicate finding from another engagement, but the promoted finding may remain unverified.

The resulting finding state is actionable but does not qualify for automatic Jira creation when Jira Project Settings use "Push All Issues", because DefectDojo only pushes findings that are both Active and Verified.

Observed promoted finding state after stale engagement deletion:

```json
{
  "active": true,
  "verified": false,
  "duplicate": false,
  "risk_accepted": false,
  "severity": "High"
}
```

The finding is no longer marked as a duplicate and is active, but `verified=false` prevents Jira issue creation. This leaves a vulnerability visible in DefectDojo without the expected Jira ticket/workflow.

**Steps to reproduce**

1. Configure a Product with Jira Project Settings and enable `Push All Issues`.
2. Import a scan into engagement A with settings that create Active and Verified findings.
3. Import or reimport a matching finding in engagement B so that DefectDojo marks the engagement B finding as a duplicate of the engagement A finding.
4. Delete engagement A, or delete the engagement containing the current primary/original finding.
5. Observe that DefectDojo promotes/reactivates the duplicate finding from engagement B.
6. Inspect the promoted finding via the UI or API.
7. Observe that the promoted finding is `active=true` and `duplicate=false`, but `verified=false`.
8. Observe that no Jira issue is created for the promoted finding even though Product Jira Project Settings are configured to push all eligible findings.

In our environment this was observed with recurring scanner imports and stale engagement cleanup. One API-observed example after engagement deletion had `active=true`, `verified=false`, `duplicate=false`, `risk_accepted=false`, and High severity.

**Expected behavior**
When DefectDojo promotes/reactivates a duplicate finding because the original finding's engagement is deleted, the promoted finding should stay eligible for the same downstream workflows as the finding it replaces.

At minimum, DefectDojo should either:

- preserve or set `verified=true` when the promoted finding represents an active verified vulnerability, especially when the original/scan import settings were verified, or
- re-run the normal Jira push eligibility flow after promotion, or
- document and expose an intentional behavior if promoted duplicates are deliberately left unverified.

The key expectation is that deleting an engagement should not silently create an active primary finding that is skipped by Jira automation solely because `verified` was left false during promotion.

- [ ] Docker Compose
- [X] Kubernetes
- [ ] GoDojo

**Environment information**
 - Docker Compose or Helm version (Output of `docker compose version` or `helm version`): Not available at time of filing
 - DefectDojo version (see footer) or commit message: v. 2.58.3

**Logs** 
No application traceback was observed. The issue appears to be a data/state transition problem during duplicate promotion after engagement deletion.

The downstream symptom is that Jira creation is skipped because the promoted finding is not Verified. DefectDojo's Jira documentation states that Product-level `Push All Issues` applies to findings that are Active and Verified:

https://docs.defectdojo.com/issue_tracking/jira/os__jira_guide/

**Sample scan files**
Not available yet. The issue was observed with recurring scanner imports and later stale engagement deletion.

**Screenshots**
Not included.

This appears related to duplicate lifecycle handling rather than scanner parsing. The problematic sequence is:

1. Finding A is the primary finding in engagement A.
2. Finding B in engagement B is a duplicate of finding A.
3. Engagement A is deleted.
4. Finding B is promoted/reactivated as the remaining active primary finding.
5. Finding B remains `verified=false`.
6. Jira automation does not create an issue for finding B.

A regression test would be useful for:

1. Create a verified active finding in one engagement.
2. Create a duplicate finding in another engagement.
3. Delete the engagement containing the original finding.
4. Assert that the promoted replacement finding is active, not duplicate, and preserves/receives the expected verified state or otherwise triggers Jira eligibility handling.

valentijnscholten:
## Summary

- When the original of a duplicate cluster is deleted (e.g. via engagement deletion), `reconfigure_duplicate_cluster` in `dojo/finding/helper.py` promotes the first remaining duplicate to the new primary. It already copies `active` and `is_mitigated` from the original, but **not** `verified`.
- The promoted finding therefore kept its own `verified=False`, which blocks Jira's "Push All Issues" (requires Active **and** Verified). Vulnerability stayed visible in DefectDojo with no Jira ticket.
- Fix: also copy `verified` from the original to the new primary.
- Regression tests in `unittests/test_prepare_duplicates_for_delete.py` cover both directions (original verified+active+not-mitigated, and original unverified+inactive+mitigated), with the duplicate starting in the opposite state for each field so every copy is observable.
- Fixes #14911

Out of scope for this PR (potential follow-ups):
- Relinking the existing `JIRA_Issue` from the deleted original to the new primary (instead of creating a new ticket). The current `JIRA_Issue.finding` is a `OneToOneField` with `on_delete=CASCADE`, so the row is gone before promotion runs — requires intercepting pre-delete.
- Auto-closing the Jira ticket when no replacement duplicate exists.
- `Finding_Group`-level Jira relinking.

### DIFF
--- a/dojo/finding/helper.py
+++ b/dojo/finding/helper.py
@@ -603,6 +603,7 @@ def reconfigure_duplicate_cluster(original, cluster_outside):
             duplicate=False,
             duplicate_finding=None,
             active=original.active,
+            verified=original.verified,
             is_mitigated=original.is_mitigated,
         )
         new_original.found_by.set(original.found_by.all())
--- a/unittests/test_prepare_duplicates_for_delete.py
+++ b/unittests/test_prepare_duplicates_for_delete.py
@@ -294,15 +294,56 @@ class TestPrepareDuplicatesForDelete(DojoTestCase):
         self.assertFalse(dupe_of_b.duplicate)
         self.assertIsNone(dupe_of_b.duplicate_finding)
 
-    def test_original_status_copied_to_new_original(self):
-        """New original inherits active/is_mitigated status from deleted original."""
+    def test_original_status_copied_to_new_original_active_verified(self):
+        """
+        New original inherits active/verified/is_mitigated from deleted original.
+
+        Positive case: original is an open, verified, not-mitigated finding.
+        Duplicate starts with the opposite of each field so every copy is observable.
+
+        Regression test for issue #14911: promoted duplicates kept verified=False
+        even when the original was verified, blocking Jira "Push All Issues".
+        """
+        original = self._create_finding(self.test1, "Original")
+        original.active = True
+        original.verified = True
+        original.is_mitigated = False
+        super(Finding, original).save(skip_validation=True)
+
+        outside_dupe = self._create_finding(self.test2, "Outside Dupe")
+        outside_dupe.is_mitigated = True
+        super(Finding, outside_dupe).save(skip_validation=True)
+        self._make_duplicate(outside_dupe, original)  # forces active=False, verified default False
+
+        with impersonate(self.testuser):
+            prepare_duplicates_for_delete(self.test1)
+
+        outside_dupe.refresh_from_db()
+        self.assertFalse(outside_dupe.duplicate)
+        self.assertTrue(outside_dupe.active)
+        self.assertTrue(outside_dupe.verified)
+        self.assertFalse(outside_dupe.is_mitigated)
+
+    def test_original_status_copied_to_new_original_inactive_mitigated(self):
+        """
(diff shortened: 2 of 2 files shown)


mtesauro (approved):
Approved

dogboat (approved):

Jino-T (approved):

blakeaowens (approved):

PR #14941, opened 2026-06-01
fix(dependency_check): fold related dependency paths into description instead of emitting separate findings

valentijnscholten:
## Summary

- The Dependency Check parser previously emitted one finding per `<relatedDependency>` in the report, in addition to the finding for the main vulnerable dependency. Because OWASP Dependency-Check attaches the vulnerability only to the main dependency in the XML and the related entries are metadata pointing to other files in the same logical component, this produced N findings sharing the same title, CVE, component name and component version — only the file path differed. Issues such as duplicated Spring Boot or ActiveMQ findings stem from this.
- This change folds related dependency file paths into the main finding's description (under a `**Related Filepaths:**` block) and stops emitting one finding per related entry. The `related` tag is no longer applied because no separate related findings are created.
- Adds an explanation of the five `DependencyBundlingAnalyzer` scenarios (`hashesMatch`, `isShadedJar`, `isWebJar`, CPE+base-path+vulns+filename sibling match, NPM same name+version) as a docstring on the new helper in the parser, in the parser documentation page, and in the 2.59.1 upgrade notes.
- Updates the `multiple_vulnerabilities_has_multiple_findings` unit test from 11 findings to 9 and adds assertions for the `**Related Filepaths:**` block on the affected finding.
- Adds `docs/content/releases/os_upgrading/2.59.1.md` with required-actions guidance (saved filters referencing the `related` tag will need updating; on the next reimport the previously created `related` findings will be closed as no longer present).

### Background

Dependency-Check's `DependencyBundlingAnalyzer.evaluateDependencies` merges co-grouped artifacts into a main dependency under five scenarios. Only scenario 1 (`hashesMatch`) represents the same vulnerable artifact at multiple deploy locations; scenarios 2-5 are different files representing one logical component. The XML attaches the vulnerability only to the main dependency in every case, so emitting per-related-dependency findings inflated a single CVE into N near-duplicates regardless of which scenario produced the bundling.

### Files

- `dojo/tools/dependency_check/parser.py` — drop the per-related-dependency loop in `get_findings`; add `build_related_dependencies_block` helper; append the block to the main finding's description.
- `unittests/tools/test_dependency_check_parser.py` — adjust expected count from 11 to 9; add `**Related Filepaths:**` assertions on the affected finding; shift remaining item indices.
- `docs/content/supported_tools/parsers/file/dependency_check.md` — document the new behavior and the five bundling scenarios.
- `docs/content/releases/os_upgrading/2.59.1.md` — upgrade notes for the parser behavior change.

### DIFF
--- a/docs/content/releases/os_upgrading/2.59.1.md
+++ b/docs/content/releases/os_upgrading/2.59.1.md
@@ -1,0 +1,32 @@
+---
+title: 'Upgrading to DefectDojo Version 2.59.1'
+toc_hide: true
+weight: -20260603
+description: Dependency Check parser no longer emits separate findings for related dependencies; related file paths are now listed in the main finding's description.
+---
+
+## Dependency Check parser: related dependencies folded into the main finding
+
+The Dependency Check parser previously created one finding per `<relatedDependency>` in the report, in addition to the finding for the main vulnerable dependency. Because OWASP Dependency-Check attaches the vulnerability only to the main dependency in the XML and the related entries are metadata pointing to other files in the same logical component, this produced N findings sharing the same title, CVE, component name and component version — only the file path differed. For projects with Spring Boot, ActiveMQ, or other libraries whose CPE matches many sibling artifacts this produced significant noise.
+
+Starting in 2.59.1, the parser emits exactly one finding per vulnerability per main dependency. The file paths of any related dependencies are surfaced in the finding description under a `**Related Filepaths:**` block.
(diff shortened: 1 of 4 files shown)


mtesauro (approved):
Approved

dogboat (approved):

Jino-T (approved):

blakeaowens (approved):

PR #14966, opened 2026-06-07
perf(importers): batch Vulnerability_Id inserts

valentijnscholten:
## Summary

- Fix `sanitize_vulnerability_ids` in `dojo/finding/helper.py` — was a silent no-op (reassigned local variable, return value never used by caller); now returns the filtered list
- Replace per-row `Vulnerability_Id.save()` loop in `save_vulnerability_ids` with `bulk_create`, reducing N INSERT queries to 1 per finding for all callers (importer and reimporter)
- Add cross-finding batch accumulation in `DefaultImporter`: `store_vulnerability_ids` appends `Vulnerability_Id` objects to `self.pending_vulnerability_ids`; `flush_vulnerability_ids()` bulk-inserts the whole batch at each batch boundary alongside `location_handler.persist()`
- Add cross-finding batch accumulation in `DefaultReImporter`: `reconcile_vulnerability_ids` appends changed findings to `self.pending_vuln_id_deletes` and new IDs to `self.pending_vulnerability_ids`; `flush_vulnerability_ids()` issues one `DELETE WHERE finding_id IN (...)` followed by one `bulk_create` per batch; findings with unchanged IDs hit the early-exit path and never touch either buffer

For a scan with 1000 findings × 5 CVEs each: ~5000 INSERT queries reduced to O(number of batches) `bulk_create` calls. Same reduction applies on reimport for findings whose vulnerability IDs changed.


mtesauro (approved):
Approved

Jino-T (approved):

dogboat (approved):

rossops (approved):

PR #14968, opened 2026-06-07
test(perf) on bugfix: re-enable import performance tests with recalibrated query counts

valentijnscholten:
## Summary

- Re-enables `TestDojoImporterPerformanceSmall` and `TestDojoImporterPerformanceSmallLocations` which were skipped pending re-baseline after the RBAC→legacy authorization migration
- Recalibrates expected query counts: legacy auth has lower per-action overhead (no role-permission lookups, simpler dispatch), so all counts decreased by 1–7 queries per step
- Removes the unused `unittest.skip` import

### DIFF
--- a/unittests/test_importers_performance.py
+++ b/unittests/test_importers_performance.py
@@ -20,7 +20,6 @@ Always verify updated counts by:
 
 import logging
 from contextlib import contextmanager
-from unittest import skip
 from unittest.mock import patch
 
 from crum import impersonate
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

rossops (approved):

blakeaowens (approved):

Jino-T (approved):

PR #14969, opened 2026-06-07
perf(importers): batch BurpRawRequestResponse inserts + re-enable perf tests

valentijnscholten:
## Summary

- Replace per-finding `save()` calls in `process_request_response_pairs` with `bulk_create` at batch boundaries, mirroring the existing `location_handler` pattern. Reduces DB round-trips proportionally to the number of findings with request/response data.
- Drop the no-op `clean()` calls — `BurpRawRequestResponse` has no custom `clean()` and inherits the no-op from `models.Model`.
- Accumulate objects in `self.pending_burp_rr` (initialized in `BaseImporter.__init__`); flush via `flush_burp_request_response()` at the same batch boundaries where `location_handler.persist()` is called in both the importer and reimporter.
- Re-enable `TestDojoImporterPerformanceSmall` and `TestDojoImporterPerformanceSmallLocations` with recalibrated query counts after the RBAC→legacy authorization migration (legacy auth has lower per-action overhead, reducing counts by 1–7 queries per step).

### DIFF
--- a/dojo/importers/base_importer.py
+++ b/dojo/importers/base_importer.py
@@ -80,6 +80,7 @@ class BaseImporter(ImporterOptions):
         ImporterOptions.__init__(self, *args, **kwargs)
         self.pending_vulnerability_ids: list[Vulnerability_Id] = []
         self.pending_vuln_id_deletes: list[int] = []
+        self.pending_burp_rr: list[BurpRawRequestResponse] = []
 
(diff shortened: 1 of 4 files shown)


mtesauro (approved):
Approved

dogboat (approved):

blakeaowens (approved):

rossops (approved):

PR #14984, opened 2026-06-09
feat(parsers): add PICUS Breach and Attack Simulation CSV parser

skywalke34:
## Description
New parser for [Picus Security](https://www.picussecurity.com/), a Breach and Attack Simulation (BAS) platform. Picus runs simulated attacks and reports whether existing security controls prevented, logged, and alerted on each simulated action.

The parser:
- Parses Picus result CSV exports under the `PICUS Scan` scan type (the same schema covers the Email, Endpoint, Network, and Web vector exports).
- Maps each attack action to a Finding: a simulated attack that was **Not Blocked** is imported as an active finding (an open control gap); a blocked attack is imported as inactive.
- Takes severity from `threatSeverity`, builds the title from `threatName - actionName` (truncated to 500 chars), and constructs a markdown-table description.
- Adds MITRE ATT&CK tactic/technique/sub-technique and the attack category as tags, and maps CVE/CWE when present.
- Sets `vuln_id_from_tool` from the native Picus `actionId` for deduplication.

## Test results
20 unit tests covering:
- No/one/many findings
- Severity mapping (Critical/High/Medium/Low) sourced from `threatSeverity`
- Active vs inactive logic (Not Blocked vs Blocked)
- `vuln_id_from_tool` extraction, CVE/CWE mapping, MITRE tags
- Title truncation and the markdown description

All tests pass under both `V3_FEATURE_LOCATIONS=False` and `True`. `ruff` is clean.

## Documentation
Parser documentation at `docs/content/supported_tools/parsers/file/picus.md`.

## Deduplication
Registered in `settings.dist.py` with `DEDUPE_ALGO_HASH_CODE` keyed on the single stable field `vuln_id_from_tool` (the native `actionId`). `actionId` is stable across simulation runs while `simulationRunId` changes per run, so keying on `actionId` alone lets re-imported runs match prior findings and update their status rather than creating duplicates.

## Checklist
- [x] Rebased against the very latest `dev`
- [x] Submitted against `dev` branch
- [x] Meaningful PR name
- [x] Code is flake8/ruff compliant
- [x] Code is Python 3.13 compliant
- [x] Documentation included
- [x] No model changes, no migrations needed
- [x] Unit tests included (20 tests)
- [x] Deduplication configured in `settings.dist.py`


devGregA (approved):

dojo/tools/picus/parser.py
```diff
@@ -0,0 +1,133 @@
...
+    def _build_mitigation(self, get):
+        prevention = get("threatPreventionResult")
+        if prevention == "Not Blocked":
+            return (
+                "The simulated attack was NOT blocked by existing preventive controls. "
+                "Review and tune the relevant security controls to block this technique."
+            )
+        if prevention == "Blocked":
+            return "The simulated attack was blocked by existing preventive controls."
+        return ""
```
valentijnscholten: Should this be `""` or maybe `None`?

skywalke34: Good catch. I've reworked `_build_mitigation` so it no longer falls through to a bare `return ""` (just pushed in 11d5bef). It now aggregates the prevent → log → alert control posture plus any available Picus mitigation/triage references (mitigation guidance, detection content, payload output, action logs, detection signature), emitting only the fields that are present. In the edge case where none of those exist it now returns `None` rather than `""`, so the field stays unset instead of being persisted as an empty string.

Maffooch (approved):

mtesauro (approved):
Approved

valentijnscholten (approved):

PR #14985, opened 2026-06-09
feat(helm): add Gateway API HTTPRoute and External Secrets Operator support

jonasschultheiss:
## Summary

- **HTTPRoute support**: Adds a `django-httproute.yaml` template as an alternative to the classic Ingress resource for users running Kubernetes Gateway API controllers (e.g. Envoy Gateway, Istio, Cilium). Configured via `django.httpRoute` values. Disabled by default.
- **External Secrets Operator (ESO) support**: Adds an `external-secret.yaml` template for syncing secrets from external stores (HashiCorp Vault, AWS Secrets Manager, Azure Key Vault, etc.) via the [External Secrets Operator](https://external-secrets.io). Configured via `externalSecret` values. Disabled by default.

## Motivation

**HTTPRoute**: The Kubernetes Gateway API (`gateway.networking.k8s.io/v1`) is now GA and increasingly adopted. This allows users who have moved to Gateway API to use DefectDojo natively without maintaining separate HTTPRoute manifests. Related: #14667

**External Secrets**: Many production deployments avoid storing secrets in Helm values or manually-created Kubernetes Secrets. ESO is the standard solution for syncing secrets from external vaults. Adding first-class support makes DefectDojo easier to deploy securely.

## Changes

| File | Change |
|---|---|
| `helm/defectdojo/templates/django-httproute.yaml` | New HTTPRoute template |
| `helm/defectdojo/templates/external-secret.yaml` | New ExternalSecret template |
| `helm/defectdojo/values.yaml` | Added `django.httpRoute` and `externalSecret` value blocks |

## HTTPRoute details

- Gated by `django.httpRoute.enabled` (default `false`)
- Supports `parentRefs`, `hostnames`, and flexible `rules` with `matches` and `filters`
- Automatically selects port 80 or 443 based on `django.nginx.tls.enabled`
- Follows DefectDojo's existing labeling and annotation patterns (`extraLabels`, `extraAnnotations`)
- Can be used alongside or instead of the existing Ingress

Example values:

```yaml
django:
  httpRoute:
    enabled: true
    parentRefs:
      - name: my-gateway
        namespace: default
    hostnames:
      - defectdojo.example.com
    rules:
      - matches:
          - path:
              type: PathPrefix
              value: /
```

## ExternalSecret details

- Gated by `externalSecret.enabled` (default `false`)
- Uses the stable `external-secrets.io/v1` API
- Supports any SecretStore/ClusterSecretStore provider
- `targetSecretName` controls the name of the Kubernetes Secret created by ESO

Example values:

```yaml
externalSecret:
  enabled: true
  refreshInterval: 30m
  secretStoreRef:
    name: my-vault-store
    kind: ClusterSecretStore
  targetSecretName: defectdojo
  data:
    - secretKey: DD_ADMIN_PASSWORD
      remoteRef:
        key: defectdojo/admin
        property: password
    - secretKey: DD_SECRET_KEY
      remoteRef:
        key: defectdojo/app
        property: secret-key
```

## Test plan

- [x] `helm lint helm/defectdojo` passes
- [x] `helm template` with httpRoute enabled renders a valid `gateway.networking.k8s.io/v1 HTTPRoute` with correct service name and port
- [x] `helm template` with `django.nginx.tls.enabled=true` correctly selects port 443
- [x] `helm template` with externalSecret enabled renders a valid `external-secrets.io/v1 ExternalSecret`
- [x] `helm template` with default values produces neither HTTPRoute nor ExternalSecret resources

### DIFF
--- a/helm/defectdojo/Chart.yaml
+++ b/helm/defectdojo/Chart.yaml
@@ -34,4 +34,8 @@ dependencies:
 #     description: Critical bug
 annotations:
   artifacthub.io/prerelease: "true"
-  artifacthub.io/changes: ""
+  artifacthub.io/changes: |
+    - kind: added
+      description: Add Gateway API HTTPRoute support as an alternative to Ingress
+    - kind: added
+      description: Add External Secrets Operator (ESO) support for external secret stores
--- a/helm/defectdojo/README.md
+++ b/helm/defectdojo/README.md
@@ -618,6 +618,7 @@ A Helm chart for Kubernetes to install DefectDojo
 | django.extraInitContainers | list | `[]` | A list of additional initContainers to run before the uwsgi and nginx containers. |
 | django.extraVolumeMounts | list | `[]` | Array of additional volume mount points common to all containers and initContainers. |
 | django.extraVolumes | list | `[]` | A list of extra volumes to mount. |
+| django.httpRoute | object | `{"annotations":{},"enabled":false,"hostnames":[],"labels":{},"parentRefs":[],"rules":[]}` | Kubernetes Gateway API HTTPRoute configuration (alternative to Ingress) |
 | django.ingress.activateTLS | bool | `true` |  |
 | django.ingress.annotations | object | `{}` | Restricts the type of ingress controller that can interact with our chart (nginx, traefik, ...) `kubernetes.io/ingress.class: nginx` Depending on the size and complexity of your scans, you might want to increase the default ingress timeouts if you see repeated 504 Gateway Timeouts `nginx.ingress.kubernetes.io/proxy-read-timeout: "1800"` `nginx.ingress.kubernetes.io/proxy-send-timeout: "1800"` |
 | django.ingress.enabled | bool | `true` |  |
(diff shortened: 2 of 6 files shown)


mtesauro:
@jonasschultheiss Please make PRs against the `dev` branch. We don't accept PRs against main/master.

For details, see: https://github.com/DefectDojo/django-DefectDojo/blob/master/readme-docs/CONTRIBUTING.md

jonasschultheiss:
@mtesauro i've updated the base branch

rossops:
The list-item dash - is attached to the matches block, not the rule itself:
```
    {{- range .Values.django.httpRoute.rules }}
    {{- with .matches }}
    - matches:          # ← the "-" only appears if .matches exists
      {{- toYaml . | nindent 8 }}
    {{- end }}
```
Gateway API makes matches optional (an omitted matches defaults to path-prefix /), so a catch-all rule — e.g. rules: [{filters: [...]}] or rules: [{}] — is valid input. But without .matches, no dash is emitted and rules renders as a map instead of a list:
```
spec:
  rules:
      backendRefs:        # ← no "-", so rules is now a mapping, not a sequence
        - group: ""
          ...
```
helm template accepts this silently, but kubectl apply rejects it against the HTTPRoute CRD (rules must be an array). With multiple match-less rules you'd also get duplicate backendRefs/filters keys. The fix is to emit the - unconditionally per rule and hang matches/filters/backendRefs off it, e.g.:
```
    {{- range .Values.django.httpRoute.rules }}
    - {{- with .matches }}
      matches:
        {{- toYaml . | nindent 8 }}
      {{- end }}
      {{- with .filters }}
      filters:
        {{- toYaml . | nindent 8 }}
      {{- end }}
      backendRefs:
        - group: ""
          ...
```

rossops (approved):

Maffooch:
Hi @jonasschultheiss once the shema tests are passing, this is fine to merge based on @rossops feedback. I'm not super versed in helm, so not sure what feedback to provide to get the failing check to pass

Maffooch:
Closing as stale

PR #14999, opened 2026-06-10
fix(ui): use brand color tokens instead of hardcoded hex in new UI

ksitton58:
**Description**

Several new-UI templates hardcoded brand colors that should reference the design
tokens in `components/tailwind.css`:

- `calendar.html` — active engagement/test events used the legacy Bootstrap primary
  blue (`#337ab7`) instead of the brand blue.
- `benchmark.html` — table link color, same legacy `#337ab7`.
- `base.html` — sidebar background (`#003864`) and sub-nav link color (`#82B0D9`)
  hardcoded the Fuji Blue brand hues.

These now use the matching `var(--color-dd-primary-*)` tokens (500 / 900 / 200). The
token values are identical to the hardcoded hex, so there is no visual change.

Intentionally left as-is:
- Sidebar's lighter link color (`#C6DDF2`): the matching shade (`dd-primary-100`) is
  not emitted in the compiled CSS, since Tailwind v4 only outputs theme variables
  that a generated utility references.
- PDF report templates also contain `#337ab7`, but CSS custom properties don't resolve
  in the PDF renderer.
- Generic white/black, neutral grays, and the custom `#002a4d` shade have no token.

FullCalendar 3.10.5 applies the event `color` verbatim as an inline style, so the
variable resolves at render time.

**Test results**

Verified in the running new UI: sidebar bg = `rgb(0,56,100)`, sub-nav link =
`rgb(130,176,217)`, calendar event = `rgb(23,121,197)` — each identical to the hex it
replaced. No Python changed, so no unit tests added.

**Documentation**

N/A.

**Checklist**

- [x] Meaningful PR name.

### DIFF
--- a/dojo/templates/base.html
+++ b/dojo/templates/base.html
@@ -88,7 +88,7 @@
             #dd-sidebar {
                 scrollbar-width: thin;
                 scrollbar-color: rgba(255,255,255,0.2) transparent;
-                background: #003864;  /* Fuji Blue Hue 04 — deepest brand blue */
+                background: var(--color-dd-primary-900);  /* Fuji Blue Hue 04 — deepest brand blue */
             }
             /* Override dojo.css 'a' color inside sidebar */
             #dd-sidebar a,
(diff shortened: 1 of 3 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

paulOsinski (approved):

Jino-T (approved):

PR #15001, opened 2026-06-11
fix(ui): add missing "solid" keyword to disclaimer border in new UI

ksitton58:
**Description**

Fixes an invisible border on the disclaimer warning box throughout the new (Tailwind) UI.

The disclaimer box set its border with `border:1px #003333` — width and color, but no
border-style keyword. Per the CSS spec, `border-style` defaults to `none`, so the border
never renders regardless of width/color. The box was clearly meant to have a border (it
sets both), but none displayed. Adding `solid` makes it render — this applies equally to
the on-screen views and the report/PDF output.

Fixed across `dojo/templates/`:
- Interactive views: `view_test.html`, `form_fields.html`, `findings_list_snippet.html`
- Report templates: `custom_html_report.html` and the `*_pdf_report.html` set (endpoint,
  engagement, finding, product, product_endpoint, product_type, test)

Out of scope (same bug, deliberately separate): the classic UI (`dojo/templates_classic/`,
being deprecated) and the email notification `.tpl` templates (different rendering engine).

**Test results**

One-property CSS correction, no logic change. Confirmed the shorthand was the cause:
`border:1px <color>` computes `border-style:none` / `border-width:0` (no border); adding
`solid` renders a 1px border. Verified the computed values in-browser. No Python changed,
so no unit tests added.

**Documentation**

N/A.

**Checklist**

- [x] Bugfix — submitted against the `bugfix` branch.
- [x] Meaningful PR name.

### DIFF
--- a/dojo/templates/dojo/custom_html_report.html
+++ b/dojo/templates/dojo/custom_html_report.html
@@ -5,7 +5,7 @@
     {{ block.super }}
     <div class="container" id="html_report">
         {% if include_disclaimer %}
-            <div style="background-color:#DADCE2; border:1px #003333; padding:.8em; ">
+            <div style="background-color:#DADCE2; border:1px solid #003333; padding:.8em; ">
                 <span style="font-size:16pt;  font-family: 'Cambria','times new roman','garamond',serif; color:#ff0000;">Disclaimer</span><br/>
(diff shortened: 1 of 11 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

paulOsinski (approved):

rossops (approved):

PR #14769, opened 2026-04-28
feat(parsers): add Xygeni JSON parser (SAST, SCA, Secrets)

lmrb-1968:
## Description

This PR adds a single first-party parser for Xygeni JSON reports under
`dojo/tools/xygeni/`. It dispatches on `metadata.scanType` and exposes three
scan types: `Xygeni SAST Scan`, `Xygeni SCA Scan`, `Xygeni Secrets Scan`. The
pattern mirrors `rusty_hog`, `anchore_grype`, `checkmarx`, `sonarqube`, and
`mobsf`.

[Xygeni](https://xygeni.io) is a Software Supply Chain Security platform that
emits a JSON report per scanner. The full pre-approval discussion (with
field-mapping tables and example JSON per kind) is at #14755.

Opened as a **draft** because pre-approval is still pending — happy to wait
for maintainer feedback before any further action. The implementation is
provided here so reviewers can evaluate the concrete shape if helpful.

## Test results

`unittests/tools/test_xygeni_parser.py` covers:

- empty report per kind (3 fixtures)
- many-finding report per kind against real anonymized fixtures
  (501 SAST + 50 SCA + 61 Secrets findings)
- dispatch on `metadata.scanType` for a synthetic minimal report
- error paths: missing `metadata.scanType`, unsupported scan type

Run via the project's docker-compose unit-tests setup against real Postgres:

```
docker compose -f docker-compose.yml -f docker-compose.override.unit_tests.yml run --rm \
  --entrypoint /bin/bash uwsgi -lc \
  '. /secret-file-loader.sh; . /reach_database.sh; cd /app;
   unset DD_DATABASE_URL DD_CELERY_BROKER_URL;
   wait_for_database_to_be_reachable;
   python3 manage.py migrate --no-input;
   python3 manage.py test unittests.tools.test_xygeni_parser --keepdb -v 2'
```

Result: `Ran 10 tests in 0.084s — OK`. `ruff check` is clean against
`dojo/tools/xygeni/` and `unittests/tools/test_xygeni_parser.py`.

## Documentation

Added at `docs/content/supported_tools/parsers/file/xygeni.md` — covers all
three scan types, the common `metadata` envelope, the per-kind payload shapes,
and links to the sample fixtures.

## Checklist

- [x] Rebased on the latest `dev`.
- [x] Targets `dev` (new parser).
- [x] Ruff compliant (line-length 120, target py313).
- [x] Python 3.13 compliant.
- [x] Documentation page added.
[...]

PR #15003, opened 2026-06-12
Fix Xygeni parser deduplicating repeated SAST/Secrets findings in the same file

lmrb-1968:

This is a bug fix for an existing parser (the Xygeni parser added in #14769), which falls under the "Acceptable examples" in CONTRIBUTING (bug fix for an existing parser), so no separate pre-approval issue is filed.

**Description**

The Xygeni parser deduplicated away legitimate findings when the same secret value or code pattern appeared more than once in a single file, so only one occurrence showed up after import.

Root cause: Xygeni reuses a single `uniqueHash` across every occurrence of the same value in a file (it hashes the value, not the location), while giving each occurrence a distinct `issueId` that encodes the file path and line. The SAST and Secrets scan types deduplicate on `unique_id_from_tool`, which was set to `uniqueHash` — so occurrences after the first were marked as duplicates and hidden.

Fix: for SAST and Secrets, key `unique_id_from_tool` on the per-occurrence `issueId` (falling back to `uniqueHash` if `issueId` is absent), and keep `uniqueHash` as `vuln_id_from_tool` so it still groups occurrences of the same value. Each occurrence is now its own Finding.

SCA is intentionally left unchanged: there `uniqueHash` (`CVE#package:version`) is unique per finding, while `issueId` (`SCA.CVE-…`) collides across packages, so `uniqueHash` remains the correct dedup key.

No model, settings, or migration changes.

**Test results**

Extended `unittests/tools/test_xygeni_parser.py`:
- Updated the existing SAST/Secrets/dispatch assertions to the new `unique_id_from_tool` (issueId) values and added `vuln_id_from_tool == uniqueHash` checks.
- Added `test_secrets_repeated_in_same_file_stay_distinct` — a secret repeated on lines 9 and 29 of one file now yields two findings with distinct `unique_id_from_tool`.
- Added `test_sast_repeated_detector_in_same_file_stays_distinct` — a detector flagged on four lines now yields four distinct findings.

Against the checked-in fixtures the parser now produces 61/61 distinct ids for Secrets and 501/501 for SAST (previously several collapsed). Ruff (0.15.15) passes on all changed files.

**Documentation**

Updated `docs/content/supported_tools/parsers/file/xygeni.md` — the Deduplication section now shows the per-scan-type `unique_id_from_tool` / `vuln_id_from_tool` mapping and explains why SAST/Secrets key on `issueId` while SCA keys on `uniqueHash`.

**Checklist**

- [x] Make sure to rebase your PR against the very latest `dev`. (Rebased onto the latest `bugfix`, the correct base for a bug fix.)
- [x] If this is a new feature and not a bug fix, you've included the proper documentation. (Not a new feature; docs updated anyway.)
- [x] Model changes must include the necessary migrations in the dojo/db_migrations folder. (No model changes.)
- [x] Add the proper label to categorize your PR. (`bugfix`)


valentijnscholten (commented):
Thanks for looking into this and the PR. The changes does mean that users will have to take a one time "hit" as existing findings will be closed when a reimport occurs where the unique_id_from_tool will now no longer match. Do you see any way around it? If not we'll need some notes in the upgrade notes for 2.59.1 (version number to be determined).
Another approach would be to create a V2 parser like we have done in other places. My feeling is for this parser it would be OK to take the one time hit.

lmrb-1968:
Thanks, @valentijnscholten. I don't see a clean way around the one-time hit. The old findings were stored with `unique_id_from_tool` set to `uniqueHash`, and there's no data on existing findings to recompute the correct issueId from, so any  migration would be guessing. Whatever the new key is, it won't line up with what's already in the database, so the first reimport after upgrading will close the old findings and create the corrected ones. After that, reimports match normally.
  
I agree the one-time hit is acceptable here rather than a V2 parser, since the old behavior was hiding real findings. I've added an upgrade note to `docs/content/releases/os_upgrading/2.59.2.md` , as milestone marked was 2.59.2, describing the reimport behavior (one-time close/recreate for SAST and Secrets tests, SCA unaffected). Happy to adjust the wording or move it if you'd prefer it land in a different version's notes.

Maffooch (approved):
I agree with @valentijnscholten

docs/content/releases/os_upgrading/3.0.100.md
Maffooch, line 1: I missed this part - can you please update to be against version 3.0.100

Maffooch (changes requested):

mtesauro (approved):
Approved

docs/content/releases/os_upgrading/3.0.100.md
lmrb-1968: Done in bffd8f8da3. Renamed the note to `3.0.100.md`, updated the title, the in-body version reference, and the Release Notes link to 3.0.100, and set the weight to `-20260622` so it sorts as the weekly patch after 3.0.0.

Maffooch (approved):

valentijnscholten (approved):

Jino-T (approved):

Issue #13777, opened 2025-11-26
Scan added webhook gets send before deduplication has occured


daanschipper:
**Bug description**
The flow of reimporting is as follows, however the processing of findings takes place asynchronous whereas the sending out of notifications takes places synchronously, resulting is the `scan_added` webhook to contain incorrect information about the new, mitigated, reactivated and untouched findings.

From https://github.com/DefectDojo/django-DefectDojo/blob/59841b42937b948eb85874eefe5f1fdebd24521b/dojo/importers/default_reimporter.py#L79:
```
The full step process of taking a scan report, and converting it to
        findings in the database. This entails the the following actions:
        - Verify the API scan configuration (if supplied)
        - Parser the findings
        - Process the findings
        - Update the timestamps on the test
        - Update/Create import history objects
        - Send out notifications
        - Update the test progress
```

**Steps to reproduce**
Set up a webhook and import the same (large) vulnerability report twice, observe the `scan_added` webhook to contain incorrect information.
```
curl -X POST "http://defectdojo.dev.rootline.com/api/v2/reimport-scan/" \
    -H "Authorization: Token ..." \
    -F "file=@large-vulnerability-report.json" \
    -F "scan_type=Trivy Operator Scan" \
    -F "product_type_name=Test Product Type" \
    -F "product_name=Test Product" \
    -F "engagement_name=Test Engagement" \
    -F "test_title=Test Test" \
    -F "auto_create_context=true" \
    -F "close_old_findings=true"
```

**Expected behavior**
The scan_added webhook should be send _after_ deduplication has taken place.

**Deployment method** *(select with an `X`)*
- [ ] Docker Compose
- [x] Kubernetes
- [ ] GoDojo

**Environment information**
- Helm version 1.7
- DefectDojo version 2.51.2

**Logs** 
Use `docker compose logs` (or similar, depending on your deployment method) to get the logs and add the relevant sections here showing the error occurring (if applicable).

**Sample scan files**
If applicable, add sample scan files to help reproduce your problem.

**Screenshots**

**Additional context** (optional)
Slack thread https://owasp.slack.com/archives/C2P5BA8MN/p1764075409115599

Already working on a pull request to schedule the webhook asynchronous.

ArnaavSinghSandhu:
[...]

PR #15007, opened 2026-06-14
feat: allow import/reimport to wait for deduplication to complete

valentijnscholten:
## Summary

While thinking about what a v3 import API could look like, I realized we're missing one piece of the puzzel in the current (re)import.

By default deduplication runs as an async celery task. The results are not awaited, so the (re)import process can finish before deduplication is complete.
Consequences are:
- The mail notification `scan_added` contains incorrect statistics
- The API response contains incorrect `delta` statitistics.

With the recent optimizations on both import and deduplication the chances of these incorrect statistics has gone down. However the most confusing case happens when the deduplication is partly finished. In that case the statistics will contain _some_ duplicates, but not all of them. Example report: #13777

Pro also suffers from this when not using the background-import. Also the pattern in this PR could be used for other Pro features like the current asynchronous prioritization.

The solution in this PR is to introduce a new `deduplication_execution_mode` enum that controls deduplication execution:

async (default): perform deduplication in the background, so not wait for the results.
sync (`block_execution==True`): perform deduplication in the foreground/in-line with the (re)import.
async_await: perform deduplication in the background *and wait for the results.*

The `async_await` is the new option and could be the new default in the future. The `block_execution` user profile variable is **retained** as the global "run all async tasks (notifications, JIRA, grading, ...) in the foreground" switch; it is independent of the new `deduplication_execution_mode` and acts as a fallback (`block_execution=True` → `sync`) when no explicit mode is set.
Data is migrated. Using `block_execution==True` could be used to achieve a similar result, but it would slow the import a lot more because deduplication would run in the foreground. It would also make everything synchronous inluding notifications, pushing to JIRA, etc causing further slowdowns.

The new `deduplication_execution_mode` can be set in the user profile, but also in the API requests.

In most scenario's the await setting will not lead to significantly longer running import, may only 1 or 2 seconds extra.
The exception is on very busy/overloaded instances where the sync deduplication jobs are queueing up.
For this reason the maximum wait time is 1 minute. After 1 minute the (re)import no longer waits and returns statistics as-is.
A response variable `deduplication_complete` will be `False` in this case.

## Details

- Add a new import/reimport **deduplication execution mode**, configurable on the user profile via `deduplication_execution_mode` (`async` / `async_wait` / `sync`) and overridable per request through a `deduplication_execution_mode` field on the import and reimport endpoints. The request value takes precedence over the profile, otherwise it falls back to `async`.
- `async_wait` (new): deduplication/post-processing is still dispatched to the background, but the request waits for deduplication to finish before responding, so `scan_added` notifications and the returned statistics reflect the deduplicated state. JIRA push, product grading and other non-deduplication tasks remain asynchronous and are **not** awaited. `async` (default, return immediately) and `sync` (run inline) round out the modes.
- The new field is **independent of the existing `block_execution` flag**, which is retained as the global "run all async tasks (notifications, JIRA, grading, ...) in the foreground" switch. A data migration seeds `deduplication_execution_mode='sync'` for users who had `block_execution=True`, and the resolver falls back to `block_execution → sync`, so import behavior is unchanged for them.
- In `async_wait` mode the post-processing batch is dispatched with a per-call `ignore_result=False` (threaded through `dojo_dispatch_task` to `apply_async`) so the request can join on it via `AsyncResult.get()` despite the global `CELERY_TASK_IGNORE_RESULT=True`. The `post_process_findings_batch` task itself stays a plain `@app.task`, so `async`/`sync` imports do not store result rows. The wait is bounded by the new `DD_DEDUPLICATION_ASYNC_WAIT_TIMEOUT` setting (default 60s) so a missing or stuck worker degrades to responding anyway rather than hanging.
- Add a `deduplication_complete` boolean to the import/reimport response indicating whether deduplication had finished by the time the response was produced (`true` for `sync` and for `async_wait` when it completed within the timeout; `false` for `async`).
- Make `scan_added` notifications deduplication-aware: once deduplication has completed, refresh the duplicate status of the affected findings from the database and split the new / reactivated / untouched lists into real and duplicate sub-lists, exclude duplicates from the headline count (so an all-duplicate import sends `scan_added_empty`), and surface the duplicate groups in the mail and webhook templates. In plain `async` mode the lists are left untouched, preserving historical behavior.
- The import/reimport response statistics are accurate after the await without further change, since they query findings and annotate the duplicate count directly.
- The browser import/reimport views resolve `deduplication_execution_mode` from the user's profile (the API resolves it in the serializer), so UI imports honor the profile setting instead of always defaulting to async. (No per-import selector is added to the UI form in this change.)
- Surface `deduplication_execution_mode` in the user profile form and the user detail view, and add 2.60 upgrade notes.


dojo/db_migrations/0269_usercontactinfo_deduplication_execution_mode.py
```diff
@@ -0,0 +1,16 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dojo', '0268_release_authorization_to_pro'),
```
Maffooch: **Blocking: migration number collision.** `dev` already ships `0269_normalize_blank_finding_components`, which also declares `dependencies = [('dojo', '0268_release_authorization_to_pro')]`. With this PR there are two `0269` migrations on the same parent → Django reports *"Conflicting migrations detected; multiple leaf nodes in the migration graph"* on `migrate` / `makemigrations --check`, and the merge fails (this is the `conflicts-detected` label).

**Fix:** renumber both files and re-chain them after the current leaf:
- `0269_usercontactinfo_deduplication_execution_mode` → `0270_...`, with `dependencies = [('dojo', '0269_normalize_blank_finding_components')]`
- `0270_seed_deduplication_execution_mode` → `0271_...`, with `dependencies = [('dojo', '0270_usercontactinfo_deduplication_execution_mode')]`

The seed migration itself is fine and idempotent.

dojo/importers/base_importer.py
```diff
@@ -80,6 +82,79 @@ def __init__(
...
+            self.deduplication_complete = True
+            return
+        timeout = getattr(settings, "DEDUPLICATION_ASYNC_WAIT_TIMEOUT", 60)
+        logger.debug("async_wait: waiting for %d post-processing task(s) (timeout=%ss)", len(results), timeout)
+        success = True
+        for result in results:
+            if result is None or not hasattr(result, "get"):
+                continue
+            try:
+                result.get(timeout=timeout, propagate=False)
```
Maffooch, line 154: **Non-blocking: the timeout is applied per batch, not per import.** This loop does `result.get(timeout=timeout)` once for *each* dispatched batch, so worst-case total wait is `N_batches × timeout`. The PR body, `2.60.md`, and the model `help_text` all say *"maximum wait time is 1 minute"* — a large import (many batches) with a stuck/overloaded worker can block the request for many minutes. It still degrades safely (never a true hang), so non-blocking, but the advertised bound is wrong.

**Mitigation:** track a single deadline so the whole join is bounded by one timeout:
```python
import time
deadline = time.monotonic() + timeout
for result in results:
    if result is None or not hasattr(result, "get"):
        continue
    remaining = deadline - time.monotonic()
    if remaining <= 0:
        success = False
        break
    try:
        result.get(timeout=remaining, propagate=False)
    except Exception as e:
        logger.warning("async_wait: error/timeout waiting for post-processing: %s", e)
        success = False
```

dojo/finding/helper.py
```diff
@@ -456,7 +456,13 @@ def post_process_finding_save_internal(finding, dedupe_option=True, rules_option
...
 
 
-@app.task
+# ignore_result=False so the 'async_wait' import execution mode can join on the
+# dispatched batch via AsyncResult.get() even when CELERY_TASK_IGNORE_RESULT=True.
+# NOTE: this override may not be strictly necessary — in the past .get()/await
+# appears to have worked with the global CELERY_TASK_IGNORE_RESULT=True and a
+# Redis broker. Needs verification against a real broker/worker setup; if join
+# works without it, this override can be removed to avoid storing extra results.
+@app.task(ignore_result=False)
```
Maffooch: **Non-blocking: this stores a result row for *every* batch of *every* import, in all modes.** The result is only consumed in `async_wait`, but flipping the task default to `ignore_result=False` makes plain `async` imports (the default) write rows to the `django-db` result backend, retained for `CELERY_RESULT_EXPIRES` (24h). On a busy instance that's a steady stream of otherwise-useless `celery_taskmeta` rows.

Re: the NOTE's open question — I verified live that the join **does** work with this override against a real worker + `django-db` backend + global `CELERY_TASK_IGNORE_RESULT=True`. The remaining concern is just the storage cost for the non-`async_wait` modes.

**Mitigation:** leave the task default alone and request the result only when you'll await it, scoping it to the dispatch in `post_processing_dispatch_kwargs` / the `apply_async` call for `async_wait`, e.g. `apply_async(..., ignore_result=False)`. (Whether `.get()` still works without the override under your broker is worth one quick check; if it does, the per-call override can be dropped entirely.)

dojo/importers/options.py
```diff
@@ -345,6 +349,25 @@ def validate_do_not_reactivate(
...
+            expected_types=[str],
+            required=False,
+            default=DEDUPLICATION_EXECUTION_MODE_ASYNC,
+            **kwargs,
+        )
+        if mode not in DEDUPLICATION_EXECUTION_MODES:
+            mode = DEDUPLICATION_EXECUTION_MODE_ASYNC
+        # An explicit force_sync from a non-serializer caller still wins.
+        if kwargs.get("force_sync"):
+            mode = DEDUPLICATION_EXECUTION_MODE_SYNC
```
Maffooch, line 368: **Non-blocking: an explicit `async_wait` is silently downgraded to `sync` when `force_sync=True` is also passed.** A caller supplying both `deduplication_execution_mode='async_wait'` and `force_sync=True` ends up in `sync` with no warning. The main API/UI paths don't combine them (the serializer/views resolve a clean mode without `force_sync`), so low risk — but it's a surprising precedence for internal/test callers.

**Mitigation:** only let `force_sync` promote to `sync` when no explicit mode was supplied, e.g. apply the `force_sync` override *before* defaulting, or skip it when `kwargs.get("deduplication_execution_mode")` is a valid mode. At minimum, document that `force_sync` wins over an explicit mode.

unittests/test_import_execution_mode.py
```diff
@@ -0,0 +1,246 @@
...
+    def test_invalid_mode_defaults_to_async(self):
+        self.assertEqual(DEDUPLICATION_EXECUTION_MODE_ASYNC, self._importer("nonsense").deduplication_execution_mode)
+
+    def test_external_force_sync_promotes_to_sync_mode(self):
+        importer = self._importer(DEDUPLICATION_EXECUTION_MODE_ASYNC, force_sync=True)
+        self.assertEqual(DEDUPLICATION_EXECUTION_MODE_SYNC, importer.deduplication_execution_mode)
+
+
+@versioned_fixtures
+@override_settings(CELERY_TASK_ALWAYS_EAGER=True)
```
Maffooch, line 127: **Non-blocking: no regression guard for the real cross-process join.** This suite (and the perf test, which mocks `AsyncResult.get`) runs under `CELERY_TASK_ALWAYS_EAGER=True`, which executes the dispatched task inline on the request's connection — so the actual worker→request join with `CELERY_TASK_IGNORE_RESULT=True` is never exercised by CI. I confirmed it works by hand in this review, but the behavior the `helper.py` NOTE worries about is exactly what isn't covered.

**Mitigation:** consider an opt-in integration test (real broker/worker, e.g. under the integration-test harness) that asserts `async_wait` blocks until dedup completes and returns `deduplication_complete=true`. The `override_settings`/eager docstring here is accurate about *why* eager is used — just call out that the cross-process path is verified elsewhere/manually.

Maffooch (commented):
**Review (code + local e2e).** Reviewed the diff and exercised it end-to-end against the running stack.

**Verification (all green):**
- `unittests.test_import_execution_mode` → 20 passed; `unittests.test_importers_performance` → 11 passed / 6 skipped (incl. the new `test_deduplication_performance_pghistory_async_wait`).
- `ruff --config ruff.toml` → clean on all changed non-migration files.
- **Live e2e against a real Celery worker + broker + `django-db` result backend (not eager):** `async_wait` import → `deduplication_complete=true` with correct stats; `async` → `false`; invalid mode → 400. So the cross-process `AsyncResult.get()` join genuinely works with the global `CELERY_TASK_IGNORE_RESULT=True` — confirming the open question in the `helper.py` NOTE (see inline comment there).
- Server-rendered UI: `view_user` shows the new row and the profile form dropdown persists the setting.

One blocking item (the migration collision, #1) and four non-blocking suggestions inline below. Nice feature — the duplicate-aware notification split is a clear improvement.

dojo/importers/base_importer.py
valentijnscholten: This is by design and worst case it waits 1 minute after the last batch of findings was created. This is fine.

dojo/importers/options.py
valentijnscholten: I'm pretty sure this is fine. Once some provides `force_async=True` we need to go sync!

dojo/finding/helper.py
valentijnscholten: removed the ignore_result=False

dojo/db_migrations/0269_usercontactinfo_deduplication_execution_mode.py
valentijnscholten: rebased

unittests/test_import_execution_mode.py
valentijnscholten: good idea. I thought it was overkill and hard to do, but it turned out to be easy so good addition.

dojo/importers/base_importer.py
```diff
@@ -81,6 +83,79 @@ def __init__(
...
```
Maffooch, line 154: **Blocker — this join is a silent no-op as submitted.** `post_process_findings_batch` is dispatched as a plain `@app.task` and the global `CELERY_TASK_IGNORE_RESULT=True` means its result is never stored. `AsyncResult.get()` here then returns immediately (it never raises), so `success` stays `True` and `deduplication_complete` is set `True` even though dedup has not finished. Verified live against a real worker: `async_wait` returns in the same time as plain `async` with 0/N duplicates actually marked.

Fix (revert of `9d50bd9`), in `dojo/finding/helper.py`:
```python
# ignore_result=False so async_wait can join via AsyncResult.get() even when
# CELERY_TASK_IGNORE_RESULT=True
@app.task(ignore_result=False)
def post_process_findings_batch(...):
```
With that restored, this `.get()` correctly blocks on the worker and all duplicates are marked before responding. (GitHub won't let me leave a committable suggestion on `helper.py` because it has no net diff in this PR.)

unittests/test_import_execution_mode.py
Maffooch, line 127: `CELERY_TASK_ALWAYS_EAGER=True` makes `.get()` return an inline `EagerResult`, so dedup is always actually done here and `deduplication_complete` is `True` regardless of whether the real join works. This is why the `ignore_result` regression slips through. Please add a coverage path that runs against a real (non-eager) worker and asserts duplicates are marked at response time, not just that the field is `True`.

unittests/test_importers_performance.py
```diff
@@ -551,6 +553,40 @@ def test_deduplication_performance_pghistory_no_async(self):
...
+        """
+        configure_audit_system()
+        configure_pghistory_triggers()
+
+        # Enable deduplication
+        self.system_settings(enable_deduplication=True)
+
+        # Simulate the background worker's post-processing having completed so the join
+        # returns instantly without executing dedup on the request's DB connection.
+        with patch("celery.result.AsyncResult.get", return_value=None):
```
Maffooch, line 581: `patch("celery.result.AsyncResult.get", return_value=None)` bypasses the exact mechanism that is broken in production (the real `.get()` join under `CELERY_TASK_IGNORE_RESULT=True`). Mocking it to return instantly means this test can never catch the silent no-op. Fine for a pure query-count perf measurement, but it should not be the only non-eager coverage of `async_wait`.

tests/dedupe_test.py
```diff
@@ -509,6 +509,93 @@ def test_check_service(self):
...
+        self.goto_active_engagements_overview(driver)
+        driver.find_element(By.PARTIAL_LINK_TEXT, "Dedupe Async Wait").click()
+        driver.find_element(By.PARTIAL_LINK_TEXT, "Async Wait Test 2").click()
+        driver.find_element(By.ID, "dropdownMenu1").click()
+        driver.find_element(By.LINK_TEXT, "Re-Upload Scan").click()
+        driver.find_element(By.ID, "id_file").send_keys(str(self.relative_path / "dedupe_scans" / "multiple_findings.xml"))
+        driver.find_elements(By.CSS_SELECTOR, "button.btn.btn-primary")[1].click()
+        self.assertTrue(self.is_success_message_present(text="a total of 2 findings"))
+        # No sleep/retry: async_wait must have blocked until dedupe finished.
+        self.check_nb_duplicates_no_wait(2)
```
Maffooch: This is the only test meant to prove the cross-process wait, but it can't fail when the feature is broken: between the import POST returning and `check_nb_duplicates_no_wait` loading the findings list there are several Selenium navigations — ample time for a 2-finding background dedup to finish, so it passes whether or not the join actually blocked. Consider asserting `deduplication_complete` from the API response and/or checking duplicates with zero intervening navigation. (Also note line 530 `if "(DUPE)" and "Duplicate" in concatRow:` is a no-op on the `"(DUPE)"` half — it only tests `"Duplicate" in concatRow`.)

Maffooch (changes requested):
## Request changes — `async_wait` is non-functional as submitted

Thanks for tackling this — the problem is real (#13777) and the `async`/`async_wait`/`sync` enum, the resolver precedence, the migration, and the UI/profile surface are all well built. But I verified against a **real broker/worker** stack (uwsgi + celeryworker + valkey + postgres, not eager) and the headline `async_wait` mode does not actually wait. It returns `deduplication_complete=true` while deduplication has **not** happened yet, so the notification/statistics it is meant to fix are still wrong.

### Evidence (Generic Findings Import, isolated dedup-on-engagement engagements, 400 findings, identical re-import)

| mode | response latency | `deduplication_complete` | duplicates marked at response time |
|------|------|------|------|
| `async` | ~1.18s | `false` | 0/400 (honest) |
| **`async_wait`** | **~1.14s** | **`true`** | **0/400 ❌** |
| `sync` | ~1.64s | `true` | 400/400 ✅ |

`async_wait` returns in the same time as plain `async` (faster than `sync`) and claims completion, yet none of the findings are deduplicated when the response is produced — the dedup task finishes a moment later.

### Root cause (proven)
Commit `9d50bd9` dropped `@app.task(ignore_result=False)` on `post_process_findings_batch`, with a message claiming the join works under the global `CELERY_TASK_IGNORE_RESULT=True`. It does not: with the result ignored, `AsyncResult.get()` returns immediately instead of blocking, so the join is a silent no-op but doesn't raise (so `success` stays `True`). Re-adding the override and re-running the identical test:

| mode | latency | `deduplication_complete` | dupes at response |
|------|------|------|------|
| `async_wait` (ignore_result=False) | ~2.18s | `true` | 400/400 ✅ |

With the override restored it correctly blocks on the worker round-trip and all duplicates are marked. The fix is essentially reverting `9d50bd9`. (Note the PR description still claims `ignore_result=False` is present — it contradicts the code.)

### Why CI is green anyway (test gap)
No automated test exercises the real broker + `ignore_result=True` path that production uses: the API unit tests run `CELERY_TASK_ALWAYS_EAGER=True` (eager `.get()` always returns a real result), the perf test mocks `AsyncResult.get`, and the Selenium test asserts after several page navigations (enough time for a 2-finding background dedup to finish) and never checks `deduplication_complete`. So the feature's core guarantee has no test that can fail when it's broken.

### Required before approval
1. Restore `@app.task(ignore_result=False)` on `post_process_findings_batch` (revert `9d50bd9`), or otherwise make the join genuinely wait.
2. Add a test on a real (non-eager, non-mocked) worker that asserts both `deduplication_complete=true` **and** that duplicates are marked at response time.
3. Update the PR description to match the code re: `ignore_result`.
4. (nit) Document that `async_wait` can hold a web worker up to `DD_DEDUPLICATION_ASYNC_WAIT_TIMEOUT=60s`, a worker-occupancy risk under load.

Full testing procedure is posted as a separate comment. Inline notes below.

---
*Verified locally with Playwright + docker compose against the PR head `eab42db`; sync mode, async mode, the profile dropdown/persistence, the user-detail display, and the UI import resolving the profile mode all work correctly — the blocker is `async_wait` only.*

Maffooch:
## Testing procedure used to verify this PR

All testing was done against the **PR head (`eab42db`)** running on a full local stack — `uwsgi`, `celeryworker`, `celerybeat`, `valkey` (broker), `postgres`, `mailhog` — i.e. a **real broker + worker**, not `CELERY_TASK_ALWAYS_EAGER`. This matters: the bug only appears with a real worker and the default `CELERY_TASK_IGNORE_RESULT=True`, which is exactly what eager/mocked tests hide.

### Environment setup
1. Checked out PR head, applied the two new migrations (`0270`, `0271`), restarted `celeryworker`/`uwsgi`/`celerybeat` so the worker ran PR code.
2. Enabled deduplication globally (`system_settings.enable_deduplication=true`, `delete_duplicates=false`).
3. Confirmed `DD_DEBUG=True` so the `async_wait: waiting for N post-processing task(s)` and worker `post_process_findings_batch` log lines were observable.

### Mock data (Generic Findings Import)
Generated synthetic Generic Findings Import JSON of varying sizes — 5, 50, 200, and 400 findings — each finding with a distinct `title`/`unique_id_from_tool`/`severity`. For the clean run, titles were namespaced per-engagement so dedup scope could not bleed across tests.

### Dimensions exercised
- **All three modes**: `async`, `async_wait`, `sync`.
- **Both transports**: API (`POST /api/v2/import-scan/` with `deduplication_execution_mode`) and the browser UI (Playwright), the latter relying on the profile setting (no per-import UI selector exists).
- **Sizes**: 5 / 50 / 200 / 400 to amplify the timing signal (larger dedup = longer real wait).
- **Two passes per engagement**: pass 1 establishes baseline findings; pass 2 re-imports the identical file so every finding should dedupe.

### What was measured per import
1. **Response latency** (wall-clock around the import POST).
2. **`deduplication_complete`** from the import/reimport response body.
3. **Duplicates actually marked at response time** — immediately after the response, `GET /api/v2/findings/?test=<id>&duplicate=true` count, compared against the expected total. This is the ground truth for "did dedup really finish before responding."

### Controlled run (400 findings, isolated dedup-on-engagement engagements, 3 repeats per mode)
| mode | latency (pass 2) | `deduplication_complete` | dupes marked at response | expected |
|------|------|------|------|------|
| `async` | ~1.18s | `false` | 0/400 | (no wait, honest) |
| `async_wait` | ~1.14s | `true` | **0/400** | 400 ❌ |
| `sync` | ~1.64s | `true` | 400/400 | 400 ✅ |

### Root-cause isolation
Re-added `@app.task(ignore_result=False)` to `post_process_findings_batch`, restarted the worker, and re-ran the identical 400-finding matrix:

| mode | latency | `deduplication_complete` | dupes at response |
|------|------|------|------|
| `async_wait` (ignore_result=False) | ~2.18s | `true` | **400/400** ✅ |

The single variable that flips `async_wait` from broken → correct is the `ignore_result=False` override removed in `9d50bd9`. (Reverted the patch afterward to leave the tree at PR state.)

### Corroborating log evidence
- uwsgi: `async_wait: waiting for 1 post-processing task(s) (timeout=60s)` confirmed the join path was entered (results list not empty) yet returned in <1s — so `.get()` short-circuited rather than waiting.
- worker: `post_process_findings_batch called: finding_ids_count=200/400 ...` confirmed the dedup task ran **after** the response had already returned.

### UI / surface verification (Playwright)
- Profile page: `deduplication_execution_mode` dropdown renders with `async` / `async_wait` / `sync`; selecting `async_wait` + save → "Profile updated successfully" and persisted on reload.
- User detail page: row "Deduplication execution mode: Async, wait for deduplication" displayed.
- UI import on an engagement (no per-import selector) resolved the profile mode — uwsgi logged `deduplication_execution_mode='async_wait'` for the browser request, confirming the `engagement/views.py` / `test/views.py` resolution path works.
- API validation: invalid mode → HTTP 400; `async` → `deduplication_complete=false`; `sync` → inline dedup with duplicates marked immediately.

### Summary of what passed vs failed
- ✅ `sync` mode, `async` mode, profile dropdown + persistence, user-detail display, UI import mode-resolution, invalid-mode rejection, resolver precedence.
- ❌ `async_wait` mode: reports `deduplication_complete=true` without actually waiting for deduplication (fixed by restoring `ignore_result=False`).

valentijnscholten:
So first you tell me to remove `ignore_result=False` as it was proven that it was not needed. And now you tell me to restore it :-)

valentijnscholten:
I will restore it and verify manually again.

dojo/importers/base_importer.py
valentijnscholten: Fixed in 93d46dda6b. Rather than reverting to a global `@app.task(ignore_result=False)` (which would store a result row for every batch of `async`/`sync` imports too, per your earlier non-blocking note on this), `async_wait` now passes a per-call `ignore_result=False` threaded through `dojo_dispatch_task` to `apply_async`; the task itself stays a plain `@app.task`. Verified live against a real worker: `async_wait` -> 400/400 duplicates marked at response time, `async` -> 0. Reverting the fix makes the new test fail with `0 != 400`.

unittests/test_import_execution_mode.py
valentijnscholten: Added in c3b89b6317: `ImportAsyncWaitApiTest` in `tests/dedupe_test.py` — a non-eager, non-mocked API test against the real docker-compose worker. It imports 400 findings twice into a dedup-on-engagement engagement and asserts `deduplication_complete=true` **and** 400/400 duplicates marked at response time (with `async` as a control: incomplete + fewer marked). Confirmed it fails (`0 != 400`) when the join fix is reverted, so it can actually catch the regression.

unittests/test_importers_performance.py
valentijnscholten: Agreed — left as a pure query-count perf measurement. The real cross-process coverage now lives in `ImportAsyncWaitApiTest` (c3b89b6317), so the mocked `.get()` is no longer the only non-eager coverage of `async_wait`.

tests/dedupe_test.py
valentijnscholten: Removed the Selenium async_wait test in c3b89b6317 — it could not fail when the join was broken (few findings + several navigations let the background dedupe finish first), and the `"(DUPE)"` half of the condition was a no-op. Replaced by the deterministic real-worker API test described above.

valentijnscholten:
@Maffooch I added an explicit test. It looks a bit ugly because the only way to do it is to add an artificial delay in the dedupe celery task. This means there's a bit of code now only used for testing. A bit ugly, but at the same time it's just an if-statement that doesn't do anything in non-test mode.

Maffooch (approved):

mtesauro (approved):
Approved

Jino-T (approved):

blakeaowens (approved):

PR #15009, opened 2026-06-15
Refactor removal of deprecated features while preserving database state

Maffooch:
Update the migration process to remove the Stub Findings and Credential Manager features from Django's state while retaining their underlying database tables. This approach allows for data preservation during downgrades, ensuring a smoother transition for users.


blakeaowens (approved):

mtesauro (approved):
Approved

dogboat (approved):

rossops (approved):

PR #15023, opened 2026-06-17
fix(findings): resolve single-location filter against Location model

skywalke34:
**Description**

Filtering the findings list by a single location via the legacy-named `?endpoints=<id>` query parameter returns **HTTP 404** (e.g. `/finding?endpoints=1&location_status=Active`).

Root cause: after the Endpoint → Location migration, the `dojo_endpoint` table is empty and findings are linked to locations via `LocationFindingReference`. `dojo/filters.py` was correctly updated so the `endpoints` filter resolves against `locations__location`, **but** `ListFindings.add_breadcrumbs` in `dojo/finding/views.py` still looked the id up against the legacy `Endpoint` model:

```python
endpoint = get_object_or_404(Endpoint, id=endpoint_id)
```

Since that table is empty, `get_object_or_404` raised `Http404` for every valid location id — so the links generated by the Endpoints page (`dojo/templates/dojo/url/list.html`, e.g. `?endpoints={{ location.id }}&location_status=Active`) 404'd the entire findings list before it could render.

Fix: resolve the id against `Location` instead, mirroring the existing `dojo.url.ui.views.process_endpoint_view`, which already does `get_object_or_404(Location, id=location_id)` for the very same id. Only the breadcrumb lookup was stale; the filter and the `view_endpoint` / `vulnerable_endpoints` reverse URLs already operate on locations.

**Test results**

Added `unittests/test_finding_list_location_breadcrumb.py` (uses the V3 `dojo_testdata_locations.json` fixture, which ships location pk=2 with an Active `LocationFindingReference`):

- filtering by a single location id returns **200** (regression guard for the reported 404)
- the breadcrumb sets `filter_name == "Vulnerable Endpoints"`
- a non-existent location id still returns **404** (preserves the `get_object_or_404` contract against the correct model)

All 3 pass; `ruff` (0.15.15) clean on both changed files.

**Documentation**

**Checklist**

- [x] Ruff compliant.
- [x] Added unit tests.


skywalke34:
Thanks @dryrunsecurity. For context, the flagged pattern pre-dated this PR (the prior code did `get_object_or_404(Endpoint, id=endpoint_id)` with the same lack of scoping) and mirrored the canonical `dojo.url.ui.views.process_endpoint_view`; the findings data is also already authorization-scoped via `get_authorized_findings("view")`, so the exposure was a read-only breadcrumb that could reveal a location's existence/URL.

Regardless, I've hardened it in 8433599: the breadcrumb lookup now resolves through `get_authorized_locations("view", user=request.user)`, returning 404 for both missing and unauthorized location ids. Added a regression test (`test_unauthorized_location_404s`) for the unauthorized case.

Maffooch (approved):
Thanks for the PR! I made the object fetch work without locations as well

dogboat (approved):

mtesauro (approved):
Approved

paulOsinski (approved):

PR #15039, opened 2026-06-17
feat(jira): support multiple components in project settings (SC-13173)

Maffooch:
## Summary
Implements [SC-13173](https://app.shortcut.com/defectdojo/story/13173) — allow assigning more than one Jira component from a project's settings.

The JIRA project `component` field is a single text value that was sent to Jira as one component named after the entire string (e.g. `Security,DevSecOps`) instead of separate components. Jira's REST API expects a list: `[{"name": "Security"}, {"name": "DevSecOps"}]`.

## Changes
- `prepare_jira_issue_fields` now splits the component value on commas, trims whitespace, and drops empty entries, producing one Jira component per name. A single value without commas still yields one component.
- `JIRAProjectForm` documents that commas separate multiple components (form-level help text; no migration).
- Updated the Component section in the OS and PRO Jira guides.

### Known limitation
A component whose name legitimately contains a comma would be split. This is an accepted trade-off given component names rarely contain commas.

## Testing
- `unittests.test_jira_helper.JIRAComponentFieldTest` — new (verified red→green): single value, comma split, whitespace trimming, empty-entry dropping, and field omission when no component is set.
- `test_jira_helper` → 11 OK, `test_jira_config_product` → 15 OK.

### DIFF
--- a/docs/content/issue_tracking/jira/OS__jira_guide.md
+++ b/docs/content/issue_tracking/jira/OS__jira_guide.md
@@ -159,7 +159,7 @@ Here is an example of a **jira\_full** Issue:
 
 #### Component
 
-If you manage your Jira Space using Components, you can assign the appropriate Component for DefectDojo here.
+If you manage your Jira Space using Components, you can assign the appropriate Component for DefectDojo here. To assign more than one Component, enter a comma-separated list (for example, `Security, DevSecOps`); each value is sent to Jira as a separate component.
 
 **Custom fields**
 
--- a/docs/content/issue_tracking/jira/PRO__jira_guide.md
+++ b/docs/content/issue_tracking/jira/PRO__jira_guide.md
@@ -126,7 +126,7 @@ Here is an example of a **jira\_full** Issue:
 
 #### Component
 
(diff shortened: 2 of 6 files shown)


valentijnscholten (commented):
what if some one wants to have or is used to be able to use:

Image

Maffooch:
> what if some one wants to have or is used to be able to use:
> 
> Image

That's the tough part about this one... We have to pick a separator, and a comma felt like the best choice for consistency, but I am absolutely open to other options

valentijnscholten:
Yeah I understand. I think having commas in a component name is not very common. So maybe a note in the upgrade notes is enough.

mtesauro (approved):
Approved

Maffooch:
Good call. I've added a note to the 2.60.x upgrade notes (`docs/content/en/open_source/upgrading/2.60.md`) documenting that components are comma-separated and that a name containing a comma will be split. Pushed in e0638e60e9.

valentijnscholten (approved):

Jino-T (approved):

paulOsinski (approved):

PR #15044, opened 2026-06-18
perf(migrations): bulk backfill in 0268 release_authorization_to_pro

Maffooch:
## Problem

A customer with **15k users / 17k products** reported that
`dojo/db_migrations/0268_release_authorization_to_pro.py` took *many hours* to run.

The slow part is `backfill_authorized_users`, which translates RBAC rows into the
legacy `authorized_users` M2M:

- **Direct members:** one `get_or_create` per `Product_Member` / `Product_Type_Member` row — a `SELECT` + conditional `INSERT` each.
- **Group flattening (the killer):** a nested loop running a fresh `Dojo_Group_Member` query for *every* `Product_Group` / `Product_Type_Group` row, then a `get_or_create` per member — `O(groups × members)` sequential round-trips.
- **No logging**, so a multi-hour run looked like a hang.

At customer scale this is ~1.9M sequential queries, which on a remote/managed Postgres
(with pghistory audit triggers per INSERT) reaches multiple hours.

## Fix

- One pass over `Dojo_Group_Member` into an in-memory `group_id → [user_id]` map, reused by the group-grant expansion and the `Global_Role` flag updates.
- Accumulate `(obj_id, user_id)` pairs into a `set` (in-memory dedup), then `bulk_create(batch_size=1000, ignore_conflicts=True)` per through table. `ignore_conflicts` + the unique constraint preserves the `get_or_create` idempotency.
- `Global_Role` → `is_superuser` / `is_staff` updates stay bulk `.update()` and reuse the map.
- `logger.info` for detection, per-batch progress, flag counts, and completion (matches the `0082` / `0201` migration pattern).

The migration's operations list is unchanged (`makemigrations --check` reports no changes) — only the `RunPython` callable internals changed.

## Benchmarks (dockerized Postgres, results identical at every scale)

| Scale (users / products) | OLD time | OLD queries | NEW time | NEW queries |
|--------------------------|----------|-------------|----------|-------------|
| 5k / 2.5k (~86k pairs)   | 54.10 s  | 175,438     | 2.92 s   | 92          |
| 15k / 17k (~954k pairs)  | ~10 min *(extrapolated)* | ~1.94M *(extrapolated)* | 36.45 s | 959 |

**~1,900–2,000× fewer queries** — the load-independent proof. The old path is
latency-bound on sequential round-trips; the new path issues a near-constant handful.

## Verification

- Functional test of the real edited function: direct + group-flattened pairs correct (with dedup), `is_superuser`/`is_staff` flags correct.
- Idempotent on a second run (no duplicates, no error).

### DIFF
--- a/dojo/db_migrations/0268_release_authorization_to_pro.py
+++ b/dojo/db_migrations/0268_release_authorization_to_pro.py
@@ -56,7 +56,37 @@ reads (no DDL on the eight shared tables); the new ``authorized_users``
 M2M and the dropped ``default_group`` columns issue real DDL.
 """
 
+import logging
+from collections import defaultdict
+
 from django.db import migrations, models
+
+logger = logging.getLogger(__name__)
+
+# Number of through rows per bulk_create INSERT. Mirrors the batching used by
+# the other large data migrations (0082, 0201).
+BATCH_SIZE = 1000
+
+
+def _bulk_insert_pairs(through_model, obj_field, pairs, label):
+    """Insert ``(obj_id, user_id)`` pairs into an authorized_users through table.
+
+    ``pairs`` is a set of ``(obj_id, user_id)`` tuples, so it is already
+    deduplicated within this run. ``ignore_conflicts=True`` makes a re-run (or
+    any pre-existing row) a no-op against the table's unique constraint —
+    preserving the idempotency the previous ``get_or_create`` provided.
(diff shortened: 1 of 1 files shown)


mtesauro (approved):
Approved

valentijnscholten (approved):

rossops (approved):

rossops (approved):

paulOsinski (approved):

PR #15058, opened 2026-06-22
Modernize Tool Config credential encryption to AES-256-GCM

Maffooch:
## Summary

Modernizes the encryption used for Tool Config credentials (and any other secret routed through `dojo_crypto_encrypt`/`prepare_for_view`) from AES-256-OFB to AES-256-GCM, in a fully backward-compatible way, **and bumps the cryptography stack now that the code is safe for it**:

- `cryptography` 46.0.7 → **49.0.0** (supersedes #15029)
- `pyopenssl` 26.2.0 → **26.3.0**, which requires cryptography ≥ 49 (supersedes #15032)

The legacy scheme used `cryptography.hazmat.primitives.ciphers.modes.OFB`. OFB has been moved to the `decrepit` module and is scheduled for removal from `primitives`, so the current usage emits a deprecation warning and would eventually break under the bump above — which is exactly why the code change and the bump land together here.

## What changed (`dojo/utils.py`)

- **New "AES.2" format** = AES-256-GCM, written by `dojo_crypto_encrypt()` as `AES.2:<nonce_hex>:<ct_hex>` (12-byte random nonce, authenticated, no padding).
- **`prepare_for_view()` reads both formats** — branches on the version prefix; `AES.2` → GCM, anything else → the legacy OFB path. Existing DB values keep decrypting unchanged.
- **Same key** — reuses the existing `get_db_key()` value (valid 32-byte AES-256 key for both schemes), so no key rotation or settings change.
- **Lazy upgrade** — `AES.1` values transparently re-encrypt to `AES.2` the next time a Tool Config is saved (decrypt-on-GET + encrypt-on-save). No data migration.
- **Legacy path future-proofed** — OFB is now imported from `cryptography.hazmat.decrepit.ciphers.modes` (with an `ImportError` fallback to `primitives`), silencing the deprecation warning and surviving the eventual removal.
- **Hardened error handling** — hex decoding moved inside the `try`; catches `InvalidTag`/`ValueError`/`IndexError` so a tampered or malformed secret returns `""` instead of raising.

The four callers (`tool_config/views.py`, `api_sonarqube/api_client.py`, `display_tags.py` `get_pwd`) go through these helpers and need no changes. DefectDojo does not import `pyopenssl` directly — it is a transitive/pinned dependency, so its bump carries no code impact.

## Testing

New unit tests in `unittests/test_utils.py` cover: AES.2 round-trips (ascii/unicode/long/empty), backward-compatible decryption of a legacy AES.1 value, the new `AES.2:` format prefix, and tampered/garbage input returning `""`.

Verified the edited helpers against cryptography **49.0.0** with `DeprecationWarning` promoted to errors: AES.2 round-trips, legacy AES.1 decrypts with no deprecation warnings, and tamper/garbage/None all degrade to `""`.

### DIFF
--- a/dojo/db_migrations/0270_reencrypt_tool_config_credentials_aes_gcm.py
+++ b/dojo/db_migrations/0270_reencrypt_tool_config_credentials_aes_gcm.py
@@ -1,6 +1,6 @@
 import logging
 
-from django.db import migrations
+from django.db import migrations, models
 
 logger = logging.getLogger(__name__)
 
@@ -89,5 +89,25 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
+        # Widen the encrypted credential columns first. AES-256-GCM appends a
+        # 12-byte nonce and a 16-byte authentication tag (rendered as hex in the
+        # "AES.2:<nonce>:<ct>" payload), so a value stored at the old max length
+        # under AES.1 would overflow the column when re-encrypted below. Each
+        # field is grown by 50% of its previous max_length to leave ample room.
+        migrations.AlterField(
+            model_name="tool_configuration",
+            name="password",
+            field=models.CharField(blank=True, max_length=900, null=True),
+        ),
+        migrations.AlterField(
+            model_name="tool_configuration",
+            name="ssh",
+            field=models.CharField(blank=True, max_length=9000, null=True),
+        ),
+        migrations.AlterField(
+            model_name="tool_configuration",
+            name="api_key",
+            field=models.CharField(blank=True, max_length=900, null=True, verbose_name="API Key"),
+        ),
         migrations.RunPython(reencrypt_tool_config_credentials, noop_reverse),
     ]
--- a/dojo/db_migrations/0270_reencrypt_tool_config_credentials_aes_gcm.py
+++ b/dojo/db_migrations/0270_reencrypt_tool_config_credentials_aes_gcm.py
@@ -1,0 +1,93 @@
+import logging
+
+from django.db import migrations
+
+logger = logging.getLogger(__name__)
+
+# Tool_Configuration fields that hold credentials encrypted via
+# dojo_crypto_encrypt()/prepare_for_view(). Each is re-encrypted from the legacy
+# "AES.1" (AES-256-OFB) scheme to the modern "AES.2" (AES-256-GCM) scheme.
+ENCRYPTED_FIELDS = ("password", "ssh", "api_key")
+
+# Legacy stored-format prefix written by the old prepare_for_save().
+LEGACY_PREFIX = "AES.1:"
+
+# Re-encrypt in bounded chunks so a large Tool_Configuration table never loads
+# every row into memory at once.
+BATCH_SIZE = 500
+
+
+def reencrypt_tool_config_credentials(apps, schema_editor):
+    """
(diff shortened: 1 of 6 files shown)


valentijnscholten:
No migration? At some point we'll have to I think?

Maffooch:
Sure I can add a migration. The `prepare_for_view` reads both formats, and `dojo_crypto_encrypt` writes with the new algo. A migration felt a little extra to me, but happy to add one

valentijnscholten:
As far as I can see it only reencrypts whenever the user changes or saves the value. Which may never happen for many entries.

mtesauro (approved):
Approved

valentijnscholten (commented):
I suggest to add something to the upgrade notes to make users aware of this in case they run into problems or are crypto fanatics.

valentijnscholten (approved):
Thanks. I think in the future when the old algorithms are deleted, we'll have a break of backwards compatibility where users first need to upgrade to the latest version that still can re-encrypt old values.

dojo/db_migrations/0270_reencrypt_tool_config_credentials_aes_gcm.py
```diff
@@ -0,0 +1,93 @@
...
+                    # stored value as-is instead of overwriting it with junk.
+                    logger.warning(
+                        "Skipping Tool_Configuration %s field %r: legacy value did not decrypt",
+                        row["id"], field,
+                    )
+                    continue
+                updates[field] = dojo_crypto_encrypt(decrypted)
+
+            if updates:
+                Tool_Configuration.objects.filter(id=row["id"]).update(**updates)
```
dogboat, line 73: There's a risk here that we'd overflow some of these fields, which will cause the migration to crash. The new encryption scheme includes additional data (auth tag, 16 bytes) that could prevent some encrypted fields that are at the max length from being upgraded.

To replicate, under the dev branch create a tool config with the following password:
```password:92WF889R18WRQP3XVPPHSPTCWRCXFVY7VCWPA4QIY7NUJN9GYSZ4DJZ13VLW35TGYRYJTJ1XQFLUERAUQ28PUAT3XGMRNRV4U4L81HZXCPIK7BCAEIU4TU8J17GV2HZBDE67KUR1CY52WDIPTRNQ1P9PG4GRJK2MIJZDUPR60ONIZA6RPOLYPKY8UU3YJFRBJPCVUK7ZFJTT6W87ON9WJHBFD1LJI9G7DF4F1UFUJXEF235NCJ1QYH022L1ND0BYKE1TFP```

Save it. (Well -- you'll have to go back and edit it and save it. It looks like encryption doesn't happen on create? Is that a side issue?). Confirm the password in the DB is prefixed with "AES.1."

Then switch to the new branch and run the migration. It will fail with an exception like `django.db.utils.DataError: value too long for type character varying(600)`.

 Probably the simplest solution would be to just add an extra ~50 chars to the appropriate model fields.

dogboat (changes requested):

dojo/db_migrations/0270_reencrypt_tool_config_credentials_aes_gcm.py
Maffooch: Good catch! Will update the migration to expand the field size as well

Jino-T (approved):

dogboat (approved):

PR #15066, opened 2026-06-23
feat(cache): L1 singleton settings cache + Celery-scoped authorization caching

valentijnscholten:
Enabler for Pro: Cache harmonization PR to allow Pro to build on the in memory cache used in OS which in some Pro contexts performs (much) better than the Redis based "L2" cache. 

Bonus features:
- System settings and authorized findings calls are now cached also in celery tasks
- Protection against accidental writes on System Settings objects that were not explicitly retrieved with `no_cache=True`

## What

Two related **in-process** caching mechanisms for hot, repeated reads — no Redis/L2 tier, no new dependencies.

### 1. Singleton settings cache (`dojo/caching.py`)

A per-thread **L1 read-through cache** for global, low-cardinality singleton config (e.g. `System_Settings`). Each request/task reads a singleton from the DB at most once instead of on every access. In multi-finding processing (imports, hash_code calc, dedupe) this repeated reading was a real cost.

- `dojo_settings_cache(key=...)` — read-through decorator over a per-thread, TTL'd L1 store. Resolves L1 → getter; `None` is never cached.
- Freshness across processes comes from **resetting L1 at every request and task boundary** (middleware + the Celery task base), not from a shared tier — so a reused worker/uwsgi thread never serves config cached during a prior request/task.
- Values stored as **plain dicts** (`model_to_cache_dict` / `cache_dict_to_model`), never pickled model instances.
- `System_Settings` is served through the cache via its manager; `objects.get(no_cache=True)` bypasses it for writes; busted on save via a `post_save` receiver.
- `DojoSytemSettingsMiddleware` → renamed **`DojoSettingsManagerMiddleware`**, slimmed to: reset L1 at request start + surface a DB-read error banner. `DojoAsyncTask` resets L1 at the start of every task.
- Configurable via `DD_SETTINGS_CACHE_L1_TTL` (default `30`s; `-1` disables). `DD_CACHE_URL` optionally points `django.core.cache` at Redis for *other* framework caching — the singleton cache never reads or writes it.

**Write-guard.** The cached `System_Settings` instance is a read-only snapshot rebuilt from the cached dict; saving it could clobber concurrent changes with stale values. `System_Settings.save()` raises `ReadOnlyCachedInstanceError` on a cache-derived instance, so writers must fetch a fresh row via `objects.get(no_cache=True)` (the edit view and test helpers already do). `cache_dict_to_model(..., read_only=True)` tags the snapshot.

### 2. Authorization caching in Celery (`cache_for_request_or_task`)

`@cache_for_request` is a no-op in Celery (no request), so per-user authorization filtering re-runs on every call inside a task — notably the rules engine, which runs authorized querysets under `impersonate(rule.owner)`.

- New `cache_for_request_or_task`: resolves the store as **request cache → task cache → no caching**. The task cache is a per-thread store active **only** inside a `DojoAsyncTask` (begin/end around each task), so management commands / shells never cache without a reset boundary.
- The cache key folds in `get_current_user().pk`, so results that resolve `user` downstream (keyed on `user=None`) cannot leak between users on a reused worker thread — or within a task that impersonates multiple owners.
- All `get_authorized_*` / `authorized_*_id_set` / `get_*_permissions` getters migrated from `@cache_for_request` to `@cache_for_request_or_task`. Behavior inside a request is unchanged (still the request cache); caching is now *added* inside tasks.

## Why no L2 (shared/Redis)

An earlier revision layered a shared Redis L2 under L1; it was dropped. The only correctness need — a Celery worker seeing current config promptly — is already met by resetting L1 each task (it re-reads the DB), so a shared L2 + cross-process invalidation isn't required. L2's cost (Redis dependency, pickled-vs-dict version-keying, cross-process invalidation, Redis-down fragility) outweighed its benefit for these tiny, hot singletons. Trade-off: one small indexed DB read per singleton per request/task.

## Tests

- `unittests/test_caching.py` — L1 memoization / reset / `None`-not-cached / invalidation; the `System_Settings` save-guard; `cache_for_request_or_task` task-scope caching, per-user isolation, and reset.
- `test_importers_performance.py` expected query counts realigned to the CI-measured values (the cache migrations don't touch importer paths; the shifts are upstream drift picked up on rebase).

## Out of scope

- No shared/cross-process cache for singleton or authorization data.
- An ACL-table cache remains future work.


Maffooch (approved):

mtesauro (approved):
Approved

Jino-T (approved):

blakeaowens (approved):

PR #15081, opened 2026-06-24
Add Promptfoo (LLM eval & red-teaming) parser

Dashtid:
## Description

Adds a parser for [promptfoo](https://promptfoo.dev), an open-source LLM evaluation and red-teaming tool, aligned with the AI-testing direction in discussion #13242.

The parser ingests the JSON results file written by `promptfoo eval -o results.json` (and `promptfoo redteam run -o results.json`). promptfoo's pass/fail semantics are inverted relative to most scanners, which is the central design point:

- A result with **`success: true`** means every assertion passed — for a red-team probe that means the target model **defended** the attack — so it is **not** a finding.
- A result with **`success: false`** is a failed assertion (for a red-team probe, the attack succeeded) and becomes a Finding.
- Results with `failureReason == 2` (a provider/eval error rather than an assertion failure) are skipped — the test could not run, so it is not a vulnerability.

Other decisions:

- **Severity** comes from the red-team `metadata.severity` (`critical`/`high`/`medium`/`low`). A plain `promptfoo eval` failure carries no severity metadata and defaults to **Medium**.
- **CWE** is mapped from the plugin / harm category as a deliberately coarse starter, verified against MITRE: SQL-injection -> CWE-89, shell/command-injection -> CWE-78, prompt-injection / prompt-extraction -> CWE-1427, PII / privacy -> CWE-200, default -> CWE-1426 (*Improper Validation of Generative AI Output*).
- Failures for the same plugin against the same target are **aggregated** into one Finding (`nb_occurences`, keeping the most severe rung).
- Registered for `hash_code` deduplication on `title` + `component_name`. `severity` and `description` are intentionally excluded: `description` holds the per-run attack input/output, and `severity` is an aggregate that shifts as the set of failed attempts changes — neither is stable enough for the dedup hash.

## Test results

Adds `unittests/tools/test_promptfoo_parser.py` (14 tests) covering: zero/one/many findings, the severity matrix, the CWE mapping (including the specific `*-injection` rules taking precedence over the broad rule), aggregation, the plain-eval fallback (severity/title/identity from the failed assertion, metric-over-type), skipping passed and errored results, the lenient input shapes (bare list, top-level `results` list), `shareableUrl` -> `references`, string-form providers, bytes + UTF-8 BOM + non-ASCII input, and rejection of non-JSON input.

The sample scan files under `unittests/scans/promptfoo/` are **real promptfoo v3 output** (`results.version == 3`).

## Documentation

Adds `docs/content/supported_tools/parsers/file/promptfoo.md`.

## Checklist

- [x] Submitted against `dev`
- [x] Ruff-compliant (`ruff.toml`, ruff 0.15.16)
- [x] Python 3.13 compliant
- [x] Documentation included
- [x] No model changes (no migration needed)
- [x] Unit tests added
- [ ] Labels (for maintainers): suggest `Import Scans` and `settings_changes` (touches `settings.dist.py` for deduplication)

### DIFF
--- a/dojo/tools/promptfoo/parser.py
+++ b/dojo/tools/promptfoo/parser.py
@@ -1,0 +1,243 @@
+import json
+import logging
+
+from dojo.models import Finding
+
+logger = logging.getLogger(__name__)
+
+# promptfoo red-team severity strings -> DefectDojo severity.
+SEVERITY_MAP = {
+    "critical": "Critical",
+    "high": "High",
+    "medium": "Medium",
+    "low": "Low",
+}
+# Severity for a failed result that carries no red-team severity (a plain `promptfoo eval`
+# assertion failure has no pluginId/severity metadata). Medium is a neutral middle rung.
+DEFAULT_SEVERITY = "Medium"
+
+# Ascending severity ranking, used only to keep the most severe rung when aggregating.
+SEVERITY_RANK = {"Info": 0, "Low": 1, "Medium": 2, "High": 3, "Critical": 4}
+
+# Starter plugin/category -> CWE mapping, matched by substring (most specific first) against
+# the red-team pluginId and harmCategory. Verified against MITRE CWE 4.x:
+#   CWE-89    Improper Neutralization of Special Elements used in an SQL Command
+#   CWE-78    Improper Neutralization of Special Elements used in an OS Command
+#   CWE-1427  Improper Neutralization of Input Used for LLM Prompting (prompt injection)
+#   CWE-200   Exposure of Sensitive Information to an Unauthorized Actor (PII / data leak)
+#   CWE-1426  Improper Validation of Generative AI Output (default / output safety)
+# promptfoo plugin ids look like "harmful:hate", "pii:direct", "indirect-prompt-injection",
+# "sql-injection". Order matters: the specific *-injection rules must precede the broad
+# "injection" rule. Intentionally coarse; refine as promptfoo's plugin taxonomy is mapped
+# more finely.
+PLUGIN_CWE_RULES = [
+    ("sql-injection", 89),
+    ("shell-injection", 78),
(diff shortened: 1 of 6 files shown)


dojo/tools/promptfoo/parser.py
```diff
@@ -0,0 +1,243 @@
...
+        # "results" list or a bare list as lenient fallbacks for hand-trimmed exports.
+        if isinstance(data, dict):
+            results = data.get("results")
+            if isinstance(results, dict) and isinstance(results.get("results"), list):
+                return results["results"]
+            if isinstance(results, list):
+                return results
+        elif isinstance(data, list):
+            return data
+        return []
```
valentijnscholten: should this throw an "Unknown format" exception or some other exception with a hint of why there are no findings imported?

valentijnscholten (approved):

dojo/tools/promptfoo/parser.py
mtesauro: FWIW, I think we should. I would sure go look at the logs if an import successfully created zero findings.

Dashtid: Split into two cases here. An unrecognized upload — JSON without a promptfoo `results.results` array — now raises a `ValueError` with a hint to attach the unmodified results JSON, so a wrong or garbled file fails loudly instead of importing nothing silently.

I deliberately kept a genuinely empty result set as a non-error: for promptfoo, zero findings is a legitimate and common outcome — every probe passing means the target defended all the attacks (the inverted success semantics this parser is built around). For the logs point, a recognized file that yields zero findings now emits an INFO line explaining why (N results parsed, all passed/errored), so a successful-but-empty import is explained rather than silently confusing.

Thanks both.

Maffooch (approved):

Issue #1811, opened 2020-01-25
Product list sorting / filtering issues


valentijnscholten:
**Bug description**
The Product list page has some issues with:
- (Default) sorting
- Filter panel

**Sorting**
1. Go to /product
2. Observe that initial page load shows product list sorted by product name in ascending order (OK)
3. Observe that after page load the javascript takes over and sort the list by product name in descending order (NOK)
4. Observe that hoovering over the arrow at the product name table header has tooltip "sort descending" (NOK)
5. Click on that anyway
6. Observe that the javascript sorts the list by product name in ascending order (OK/NOK)
7. Parallel to that the page is reloaded with server-side sorting in descending order replacing the product list completely (OK/NOK)

So all in all a conflict between client-side and server-side sorting I think. I think this was added at some point last 6 months but I am only seeing it now because I never updated the static files :-D
Somebody added it so hopefully also somebody know how this front-end stuff works :-)

**Filter panel**
1. Go to /product
2. Observe that initial page load does not show the filter panel (OK)
3. Click on the arrow at product name table header to sort 
4. Observe that the page reloads and now shows the filter panel (NOK)
5. Same happens if you click on page 2 or next page for example.

**Expected behavior**
I am not a fan of client-side sorting, especially in a paged situation. So my vote would be to remove the client-side sorting.
I think the filter panel should only be shown when explicitly chosen by the user as on other pages in DD.

**Deployment method** *(select with an `X`)*
- [X] setup.bash / legacy-setup.bash

**Environment information**
 - Operating System: Debian Stretch, Ubuntu 19.04
 - Latest dev branch (89aa6f9eeedb09889d8ef9d12372397b09335d41)

![image](https://user-images.githubusercontent.com/4426050/73125947-12787e00-3fad-11ea-851e-fc6399655685.png)

valentijnscholten:
Hmm it seems to be working when using `runserver`. But behind nginx the above happens, tried on two different machines. Debian and Ubuntu. Some kind of settings I'm missing?

valentijnscholten:
Hold on, might be caused by some side effects of merging somebody elses branches. Let me double check.

dougmorato:
I think you are be right..... I added the datatables and I have been noticing some weird sorting issues recently. I also believe the conflict is between server vs client sorting.

[...]

PR #15082, opened 2026-06-24
fix(filters): don't auto-open the filter panel when only sorting a column

skywalke34:
**Description**

List pages share a single `django-filter` form for both filtering **and** column sorting (the `OrderingFilter` `o` parameter). The filter panel's initial expanded state is driven by `{% if ...form.has_changed %}`, so sorting a column — which only adds `?o=...` to the URL — flips `has_changed()` to `True` and auto-opens the filter panel even though no filter was applied.

This is most visible on the **Asset (Product) list**: clicking the *Organization* or *Active (Verified) Findings* column header reloads the page and the filter panel springs open. (Originally reported in #1811 and since regressed.)

**Fix:** add a small `has_active_filters` template filter that mirrors `form.has_changed()` but ignores the ordering field (`o`), and use it for the panel-expand condition on all list pages, in **both UI trees**:
- `dojo/templates/` — new (Tailwind) UI — 17 templates.
- `dojo/templates_classic/` — classic (default) UI — 18 templates (the same list pages, plus both filter panels on `action_history.html`).

Sorting is no longer treated as filtering, so the panel stays closed unless a real filter value is set.

This is intentionally narrow and low-risk:
- `form.has_changed()` itself is **not** modified, so nothing relying on Django's `has_changed()` semantics is affected.
- `django-filter` does not use `has_changed()` for filtering (it uses `cleaned_data`), so query/sort results are unchanged.
- The only observable effect is the filter panel's initial open/closed state.

`view_finding.html` uses a separate, hand-managed `has_changed` flag for similar-findings, and the condition in `view_notification_webhooks.html` is commented out — both intentionally left unchanged in both trees.

**Test results**

Added `unittests/test_filter_tags.py` covering `has_active_filters`:
- no params / sort-only (`o=prod_type__name`, `o=-name`) → not active (panel stays closed)
- a real filter (`name=…`) / filter + sort → active (panel opens)

```
Ran 6 tests in 0.017s
OK
```

- Verified end-to-end through the template engine that `{% if form|has_active_filters %}` renders the panel CLOSED for sort-only requests and OPEN when a filter value is present.
- Verified live on the new (Tailwind) UI: sorting the Asset list by Organization no longer opens the panel.
- Compile-checked all 18 modified classic templates parse correctly with the new `{% load filter_tags %}` + `|has_active_filters`.

**Documentation**

No documentation changes needed — UI behavior fix, no settings or model changes.

**Known related issue (out of scope here):** on tables enhanced by DataTables, clicking a sortable header briefly client-sorts the visible rows before the server-side `?o=` reload replaces them (a momentary "flash"). That is the long-standing client-vs-server sort conflict also noted in #1811 and is intentionally **not** addressed in this PR.

**Checklist**

- [x] Meaningful PR name (may be used in release notes).
- [x] Code is Ruff compliant (see `ruff.toml`).
- [x] Code is Python 3.13 compliant.
- [x] No model changes / migrations required.
- [x] Added applicable unit tests.
- [x] Fix applied to both the new and classic UI template trees.
- [ ] Add the `bugfix` label.

### DIFF
--- a/dojo/templates/dojo/components.html
+++ b/dojo/templates/dojo/components.html
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% load filter_tags %}
 {% load humanize %}
 {% load display_tags %}
 {% load static %}
@@ -16,7 +17,7 @@
                     </div>
                 </h3>
             </div>
-            <div id="the-filters" class="is-filters panel-body collapse {% if filter.form.has_changed %}in{% endif %}">
+            <div id="the-filters" class="is-filters panel-body collapse {% if filter.form|has_active_filters %}in{% endif %}">
                 {% include "dojo/filter_snippet.html" with form=filter.form %}
             </div>
         </div>
--- a/dojo/templates/dojo/dev_env.html
+++ b/dojo/templates/dojo/dev_env.html
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% load filter_tags %}
 {% load navigation_tags %}
 {% load authorization_tags %}
 {% block content %}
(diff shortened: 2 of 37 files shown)


skywalke34:
I was not able to set the bugfix label.

Maffooch (approved):

mtesauro (approved):
Approved

paulOsinski (approved):

valentijnscholten (approved):

Issue #15033, opened 2026-06-17
Govulncheck JSON and SARIF Parsers Produce Inconsistent and Incorrect Results


rr1010:
**Slack us first!**
https://defectdojocommunity.slack.com/archives/C0A4DBKANDS/p1781593721154039

**Bug description**
The Govulncheck integration in DefectDojo appears to produce inconsistent and inaccurate results when importing JSON and SARIF reports. The findings do not align with the actual output of the govulncheck CLI, leading to incorrect vulnerability counts, severity classification, and missing dependency metadata.

* Native govulncheck ./... output reports 29 vulnerabilities (expected baseline).
* JSON import (-format json) results in 234 findings, all marked as Informational, which is incorrect.
* SARIF import (-format sarif) results in 71 findings, with 29 High severity findings matching the CLI count, but:
* Titles and descriptions are unclear or not developer-friendly.
* Dependency/component information is not populated in the Components section.

Output of command govulncheck ./...
```
govulncheck ./...
=== Symbol Results ===

Vulnerability #1: GO-2026-5039
    Arbitrary inputs are included in errors without any escaping in
    net/textproto
  More info: https://pkg.go.dev/vuln/GO-2026-5039
  Standard library
    Found in: net/textproto@go1.24.4
    Fixed in: net/textproto@go1.25.11
    Example traces found:
      #1: config/config.go:98:23: config.LoadConfig calls viper.Viper.ReadInConfig, which eventually calls textproto.Reader.ReadMIMEHeader

Vulnerability #2: GO-2026-5037
    Inefficient candidate hostname parsing in crypto/x509
  More info: https://pkg.go.dev/vuln/GO-2026-5037
  Standard library
    Found in: crypto/x509@go1.24.4
    Fixed in: crypto/x509@go1.25.11
    Example traces found:
      #1: config/config.go:98:23: config.LoadConfig calls viper.Viper.ReadInConfig, which eventually calls x509.Certificate.Verify
      #2: config/config.go:98:23: config.LoadConfig calls viper.Viper.ReadInConfig, which eventually calls x509.Certificate.VerifyHostname
      #3: repository/expense.go:89:91: repository.expenseTrackerRepository.ListExpenses calls x509.HostnameError.Error

Vulnerability #3: GO-2026-5026
    Invoking failure to reject ASCII-only Punycode-encoded labels in
    golang.org/x/net/idna
  More info: https://pkg.go.dev/vuln/GO-2026-5026
[...]

PR #15087, opened 2026-06-25
fix(govulncheck): reject SARIF reports with a clear error pointing to the SARIF scan type

valentijnscholten:
## What

The Govulncheck Scanner parsers only handle govulncheck's **native JSON** output
(`govulncheck -format json`). When a govulncheck **SARIF** report
(`govulncheck -format sarif`) is uploaded to a Govulncheck Scanner scan type:

- `Govulncheck Scanner` (V1) crashed with an opaque `KeyError: 'Vulns'`
- `Govulncheck Scanner V2` silently returned **0 findings**

Both detect a SARIF document (top-level `runs` key) and now raise a clear
`ValueError` telling the user to import govulncheck SARIF output via the generic
**SARIF** scan type instead.

A note was added to the Govulncheck docs page documenting this.

## Why not parse SARIF here?

The generic SARIF parser already imports govulncheck SARIF correctly (findings,
severities, descriptions). Duplicating SARIF handling in the govulncheck parser
to gain only `component_name`/CVE — and `component_version` isn't even present in
govulncheck SARIF — isn't worth the maintenance surface. The right fix is to fail
fast with guidance toward the correct scan type.

## Testing

- `unittests/tools/test_govulncheck_parser.py`: added `test_parse_sarif_is_rejected`
  to both the V1 and V2 test classes.
- New fixture `unittests/scans/govulncheck/issue_15033_sarif.json` (trimmed
  govulncheck SARIF, one result per severity level).

Refs #15033

### DIFF
--- a/docs/content/supported_tools/parsers/file/govulncheck.md
+++ b/docs/content/supported_tools/parsers/file/govulncheck.md
@@ -27,6 +27,14 @@ original parser:
 
 The original **Govulncheck Scanner** parser is unchanged and remains available.
 
+### SARIF format
+
+The Govulncheck Scanner parsers only accept govulncheck's native JSON output
+(`govulncheck -format json`). To import govulncheck's SARIF output
+(`govulncheck -format sarif`), use the generic **SARIF** scan type instead — not
+the Govulncheck Scanner scan type. Uploading a SARIF report to a Govulncheck
+Scanner parser fails with an error pointing you to the SARIF scan type.
+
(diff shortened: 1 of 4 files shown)


Maffooch (approved):

dogboat (approved):

mtesauro (approved):
Approved

paulOsinski (approved):

PR #14987, opened 2026-06-09
refactor(reorg): extract every type of class into modules [10/10]

valentijnscholten:
## Stacked PRs

This is one of a stacked series that reorganizes `dojo/` into self-contained `dojo/{module}/` packages (matching the `dojo/url/` reference). They are stacked — each PR is based on the previous one — so **review and merge bottom-up in this order**:

1. #14970 — product_type
2. #14971 — test
3. #14972 — engagement
4. #14973 — product
5. #14974 — finding  (+ CWE & BurpRawRequestResponse fold-in)
6. #14981 — peripheral: user + system_settings
7. #14982 — peripheral: endpoint + tool_type/tool_config/tool_product
8. #14983 — peripheral: survey + benchmark
9. #14986 — peripheral: notes/note_type + file_uploads + reports + risk_acceptance
10. **#14987 — peripheral: regulations + banner + announcement + development_environment + object  ◀ this PR**

Merge #14970 first; each following branch is rebased onto the updated base and merged in turn. This PR is the top of the stack (on top of #14986) — full implementation brief is the **Phase 10** section in `AGENTS.md`.

## Summary

Reorganizes the `regulations`, `banner`, `announcement`, `development_environment`, and `object` domains into self-contained `dojo/{module}/` packages with `ui/` (and, where an API exists, `api/`) sub-packages, matching the `dojo/url/` reference layout. Pure code-movement: no behaviour or migration changes; backward-compat re-exports keep external consumers working.

### regulations / development_environment / announcement (have API layers)
- Move `Regulation`, `Development_Environment`, `Announcement`, `UserAnnouncement` into their `{models,admin}.py` (string-ref FKs). The `ANNOUNCEMENT_STYLE_CHOICES` constant moves into `dojo/announcement/models.py` (re-exported from `dojo.models`).
- Move `RegulationSerializer`/`DevelopmentEnvironmentSerializer`/`AnnouncementSerializer` into each module's `api/serializer.py` (re-exported from `api_v2/serializers.py` for prefetcher discovery), and `RegulationsViewSet`/`DevelopmentEnvironmentViewSet`/`AnnouncementViewSet` into `api/views.py` with `add_*_urls`, preserving routes/basenames.
- Move the forms into each module's `ui/forms.py` and `views.py`/`urls.py` into `ui/`; `announcement/os_message.py` and `signals.py` stay.

### banner / object (no API layer)
[...]

PR #15089, opened 2026-06-26
feat: allow disabling and dismissing the open source message banner

devGregA:
The open source message banner (the "Upgrade to Pro" promo fetched from the
GCS bucket) currently can't be turned off. This adds two ways to do that.

1. `DD_OS_MESSAGE_ENABLED` (default `True`) — an instance-wide switch.
   When it's off, `get_os_banner()` returns early, so we don't fetch the
   message or render the banner at all.

2. A dismiss (×) button on the banner. Dismissing it is remembered per user
   by storing a hash of the current message on `UserContactInfo`. If the
   message later changes, the hash no longer matches and the banner shows
   again. This follows the same pattern as the existing announcement
   dismissal, minus the confirmation page.

Default behavior is unchanged: leave the env var unset and don't click ×,
and the banner behaves exactly as it does today.

Notes:
- New `os_message_dismissed_hash` field on `UserContactInfo` (migration 0270).
- The × posts to a small `dismiss_os_message` view (CSRF handled by the form);
  JS hides the banner immediately and falls back to a normal POST if JS is off.
  Handled in both the classic and v3 base templates.
- Settings doc / `template-env` updated and a note added to the 3.1 upgrade page.

Testing: extended `unittests/test_os_message.py` (env gate, dismiss-token,
context-processor gating, and the dismiss view). All passing locally.


devGregA:
@dryrunsecurity fp drs_d57257ef This isn't user input. `banner.message` is fetched from our own GCS bucket and already runs through `bleach.clean()` in `parse_os_message()` (allowlist of strong/em/a, `strip=True`) before it hits the template. The `|safe` also predates this PR; I just moved the line when I reordered the dismiss button.

devGregA:
@dryrunsecurity fp drs_f11c06a6 Same deal, this is the classic-UI copy of that same line.

mtesauro (approved):
Approved

valentijnscholten:
I am going to push a slightly different way of implementing this flag so that we don't have to create a new column and migration for every future flag. But happy to stick to the original implementation if needed/wanted.

valentijnscholten:
Nice feature. One design suggestion on the storage, which I've gone ahead and implemented on the branch (plus resolved the conflicts with dev).

### Use an extensible JSON field instead of a column per flag
The dismissal was stored in a dedicated `os_message_dismissed_hash` column on `UserContactInfo`. That field is `editable=False`, never queried/filtered, and is one of a whole *category* of "remember the user dismissed/saw X" flags that keep showing up. Adding a column (and a migration) per minor per-user toggle is a lot of schema churn — and every `UserContactInfo` migration is a fresh source of migration-number collisions in the ongoing bugfix→dev merges.

So I replaced it with a single extensible field:

```python
user_state_details = models.JSONField(default=dict, blank=True, editable=False, ...)
```

The dismissal hash now lives under a `OS_MESSAGE_DISMISSED_KEY` key inside it. Future ephemeral UI flags ("don't show again", "seen the X tour", etc.) become a new key — **no column, no migration**.

Deliberately scoped: this is only for open-ended, non-queried UI state. Behavioral/queryable/security fields (`block_execution`, `token_*`, etc.) should stay as real columns.

Caveats worth knowing:
- Concurrency: writing a key is read-modify-write on the blob; fine for low-frequency UI prefs, but if two keys are ever written concurrently use a Postgres `jsonb_set` partial update to avoid clobbering.
- Serializer/admin exposure is manual per key — fine here since it's `editable=False` internal state.

### Conflicts with dev resolved
Merged current dev in. Two conflicts, both from the #14987 reorg that just landed:
- `UserContactInfo` moved to `dojo/user/models.py` — added `user_state_details` there (took dev's re-export `models.py`).
- announcement views moved to `dojo/announcement/ui/views.py` — kept dev's new-module imports and added the `OS_MESSAGE_DISMISSED_KEY` / `UserContactInfo` imports the dismiss view needs.

Migration renamed to `0270_usercontactinfo_user_state_details` (adds the JSONField; no collision — dev has no 0270).

Maffooch (approved):

valentijnscholten (approved):

blakeaowens (approved):

PR #15091, opened 2026-06-26
feat(i18n): enable UI translations with a per-user language setting [1/3]

devGregA:
DefectDojo's templates and models are already full of translation markers
(`{% trans %}`, `gettext`), and there are even committed pt_BR and ru catalogs,
but the runtime was never wired up, so nothing actually translated. This turns
it on and adds a per-user language preference.

### What's here
- Enable Django i18n: add `LocaleMiddleware`, `LANGUAGES`, and `LOCALE_PATHS`.
- Per-user language: a `language` field on `UserContactInfo` (migration `0271`)
  and a selector on the profile page.
- Resolution is cookie-based. The preference lives in the DB so it follows a user
  across devices, and it's cached in the standard `django_language` cookie that
  `LocaleMiddleware` reads on each request. We only touch the DB to seed the
  cookie when it's missing (roughly once per browser session) and to refresh it
  when the user changes their language, so the steady state has no extra
  per-request query. API requests are skipped and keep honoring `Accept-Language`.
- Compile the `.mo` catalogs into the image at build time (debian and alpine);
  the `.mo` files stay gitignored.

### Languages
English, Portuguese (Brazil), and Russian are offered for now. The pt_BR and ru
catalogs already existed but were never compiled or shipped, so I reviewed them
for broken format placeholders, injected markup, and the wording on
destructive/security actions and found nothing off (and fixed 3 malformed pt_BR
entries that were preventing the catalog from compiling at all). Spanish,
Japanese, German, and French get added once their translations are done.

### Scope / non-goals
- Stored database values and serialized API responses stay in English regardless
  of the selected language; only displayed text changes. Localizing API responses
  and the OpenAPI schema is deliberately left to a follow-up so existing
  integrations aren't affected.
- This is the first of a few PRs. The next one wraps form labels, flash messages,
  and the remaining hardcoded template strings that aren't marked yet (which is
  why some form fields still render in English here).

Based on #15089 (OS message opt-out). If that hasn't merged yet, its commits will
show up in this diff.

### Testing
Verified on a local instance: switching a user's language renders the UI in
Portuguese/Russian and falls back to English. `compilemessages` succeeds for all
catalogs and ruff passes.


devGregA:
Superseded by #15622, which consolidates the whole internationalization effort into a single PR against current `dev` (this branch had drifted about 540 commits behind). Closing this one.

PR #15095, opened 2026-06-26
Ree/perf indexes

rossops:
Add performance indexes for large finding datasets

Five indexes on dojo_finding backing common finding-list, SLA, and severity queries that were slow for customers with large datasets:

* idx_finding_testid_date (test_id, date) -- per-test finding lists
  ordered/filtered by date.
* idx_finding_sla_open_cov (sla_expiration_date, test_id) WHERE NOT
  is_mitigated -- SLA-expiration scans over open findings only.
* idx_finding_open_active_by_severity (severity) WHERE active AND NOT
  is_mitigated -- severity filters on the open/active working set.
* idx_finding_sev_numsev_open_unverified (severity, numerical_severity
  DESC) WHERE active AND NOT verified -- unverified-triage lists ordered
  by severity.
* idx_finding_sla_breach_cov (test_id, sla_expiration_date, date)
  INCLUDE (id) WHERE NOT is_mitigated -- covering index for SLA-breach
  lookups, served index-only without heap fetches.

The partial WHERE clauses keep these indexes small by excluding mitigated/inactive rows, which dominate row counts on large tenants but are rarely scanned by these queries.

Built CONCURRENTLY (atomic = False) so the migration takes no ACCESS EXCLUSIVE lock on the large dojo_finding table.

### DIFF
--- a/dojo/db_migrations/0271_finding_perf_indexes.py
+++ b/dojo/db_migrations/0271_finding_perf_indexes.py
@@ -31,7 +31,7 @@ class Migration(migrations.Migration):
             model_name="finding",
             index=models.Index(
                 fields=["severity"],
-                name="idx_finding_open_active_by_severity",
+                name="idx_finding_open_active_sev",
                 condition=models.Q(active=True, is_mitigated=False),
             ),
         ),
@@ -39,7 +39,7 @@ class Migration(migrations.Migration):
             model_name="finding",
             index=models.Index(
                 fields=["severity", "-numerical_severity"],
-                name="idx_finding_sev_numsev_open_unverified",
+                name="idx_finding_sev_open_unver",
                 condition=models.Q(active=True, verified=False),
             ),
         ),
--- a/dojo/db_migrations/0271_finding_perf_indexes.py
+++ b/dojo/db_migrations/0271_finding_perf_indexes.py
@@ -1,0 +1,55 @@
+from django.contrib.postgres.operations import AddIndexConcurrently
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    # CREATE INDEX CONCURRENTLY cannot run inside a transaction block, and avoids
+    # an ACCESS EXCLUSIVE lock on the (large) dojo_finding table.
+    atomic = False
+
+    dependencies = [
+        ("dojo", "0270_finding_visibility_perf_indexes"),
+    ]
+
+    operations = [
+        AddIndexConcurrently(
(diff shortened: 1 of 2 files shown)


valentijnscholten (commented):
@rossops I'm a right to assume you have tested these, checked that query plan and found out they were used/useful? I am asking because all the LLMs always tell me these indexes are not needed if you already have single field indexes covering the queries. But I never took the time to ran all this on large/huge datasets and compare.

valentijnscholten:
Also adding 5 indexes for all findings might take some time on larger instances. Would it be better to have this on `dev` i.e. next minor release?

Maffooch (approved):
Would be better to get these indexes out sooner rather than later

dogboat (approved):

rossops:
@valentijnscholten 

> I'm a right to assume you have tested these, checked that query plan and found out they were used/useful? I am asking because all the LLMs always tell me these indexes are not needed if you already have single field indexes covering the queries. But I never took the time to ran all this on large/huge datasets and compare.

> Also adding 5 indexes for all findings might take some time on larger instances. Would it be better to have this on `dev` i.e. next minor release?

Yes, sorry i didnt include the EXPLAINS. The performance gain is fairly dramatic with large data sets. Less so with smaller ones. Basically:

- Reads: Net positive across finding-list, SLA, and severity-filter paths — the queries that were our >2000 ms offenders on large tenants. The covering SLA-breach index and the severity-ordered index were the two I empirically confirmed eliminate heap fetches / sorts respectively on the clone.
- Safe / cheap: Four of five are partial (WHERE NOT is_mitigated / active), excluding the mitigated+inactive rows that dominate row counts but are rarely scanned. So the disk/maintenance footprint is far smaller than five full indexes would suggest.
- Write cost: Every INSERT/UPDATE to dojo_finding now maintains up to 5 more indexes, but the partial predicates mean a write only touches an index if the row currently matches (e.g. mitigating a finding removes it from four of them). For DefectDojo's read-heavy, bulk-import-then-query workload this is a clearly favorable trade.
- Deploy risk: Low, CONCURRENTLY means no table lock. The only caveat is that concurrent builds are slower and can leave an INVALID index if interrupted (would need a manual REINDEX/drop-recreate). When I ran the index CREATES on a DB with 10M findings, it took less than 15 seconds for all of them combined.

Jino-T (approved):

mtesauro (approved):
Approved

PR #15096, opened 2026-06-26
perf(finding-api): drop blanket DISTINCT from FindingViewSet list

rossops:
FindingViewSet.get_queryset() ended with .distinct() as a blanket guard against row multiplication from to-many filters. get_authorized_findings filters by a scalar product-id IN (no multiplication) and the prefetches don't join, so the only sources of duplicate rows are a handful of to-many filters and ordering keys. A query-wide DISTINCT over the wide finding row forces an expensive sort/hash-aggregate on every list request; rewrite the multiplying filters as Exists() and aggregate the multivalued sort keys so the DISTINCT can be removed. This should also greatly reduce pressure on databases with large findings tables.

dojo/filters.py:
- ExistsRiskAcceptanceFilter: the "Expired" (WAS_ACCEPTED) branch uses Exists() instead of a join over the risk_acceptance m2m. Overrides filter() (not just was_accepted) because ReportRiskAcceptanceFilter.options captured the parent's was_accepted at class-definition time, so overriding the method alone is dead code.
- MultivaluedOrderingFilter: ordering by a to-many field (found_by/reviewers) joins the relation and duplicates the base row; order by Min(field) instead (GROUP BY pk), paid for only when the user actually sorts by one.
- ApiFindingFilter: endpoints/found_by/reviewers/finding_group rewritten as method-based Exists() via a _has() helper; risk_acceptance -> the Exists subclass; o -> MultivaluedOrderingFilter with found_by/reviewers aggregated. tag filters still rely on DojoFilter.qs's tag-conditional distinct.

dojo/api_v2/views.py: FindingViewSet.get_queryset() returns findings without the blanket .distinct().

This is the open-source counterpart to the DefectDojo Pro change and is backwards compatible with it: Pro's ApiEnhancedFindingFilter still shadows these filters with its own overrides, and its reference to ApiFindingFilter.base_filters["o"].param_map keeps resolving because MultivaluedOrderingFilter extends OrderingFilter.

Adds unittests/test_api_finding_filter_no_multiply.py: 6 regression tests, each pinning the intent both ways -- the naive join demonstrably multiplies the matching finding (count == 2) while the real filterset/ordering returns it exactly once and still excludes non-matching findings.


Maffooch (approved):

valentijnscholten (approved):

Jino-T (approved):

mtesauro (approved):
Approved

PR #15097, opened 2026-06-26
fix(tags): keep user-set tags when creating a finding under product tag inheritance

### LINKED ISSUE #15092: Tags inheritance on product block tags on new finding (only when finding is created)
Hey guys

Tags are not apply on finding (only when finding is created) when tags inheritance is set to true on the product.
When editing the newly created finding and set the tags again, it works.

Reproduction step available in the current demo.defectdojo.org (v.3.0.100)

1. Enable tag inheritance on a product
Image

Image
2. Create engagement and test (or use existing one)

3. Create a new finding, with a random tag
4. Save and see the random tag disappeared

Image

Image

Image

Image  

5. Edit the finding and add a new tag
Image

Image

valentijnscholten:
Fixes #15092

## Problem

When a product has **tag inheritance enabled**, creating a finding with its own tag via the UI ("add finding" flow) silently drops the user's tag — only the inherited product tags survive. Editing the finding afterwards and adding a tag works, which is what made the loss look specific to creation.

Reproduced exactly:

| flow | result |
|---|---|
| create with pre-save tag (`finding.tags = [...]; finding.save()`) | ❌ `['producttag']` — user tag gone |
| add tag after save (edit flow) | ✅ `['myrandomtag', 'producttag']` |
| inheritance **off** | ✅ user tag survives |

## Root cause — signal-handler ordering

The `post_save` dispatch order for `Finding` is:

1. imagekit → 2. watson → 3. **`dojo.tags.signals.inherit_tags_on_instance`** → 4. **tagulous's flush** (`PostSaveHandler`)

Tagulous holds tags assigned to an *unsaved* instance in an in-memory manager and only writes them to the through table from its own `post_save` handler, which is registered in **tagulous'** `AppConfig.ready()`. Because `"dojo"` precedes `"tagulous"` in `INSTALLED_APPS`, dojo's inheritance handler runs **first**: reading `instance.tags` on the freshly-saved row loads an empty, DB-backed manager and discards the pending pre-save tags before tagulous can persist them.

## Fix

Before computing/merging inherited tags, flush the instance's own tag fields (tagulous `manager.post_save_handler()`) so the user's tags are committed first. This makes inheritance independent of signal-handler ordering — findings created with *or* without their own tags now end up correct. 

- `dojo/tags/signals.py` — new `_flush_pending_tag_fields()` helper, called at the top of `auto_inherit_product_tags()`.
- `unittests/test_tag_inheritance.py` — two regression tests (create-with-own-tag, create-without-tag). The first fails without the fix and passes with it.

## Testing

- Reproduced the bug and validated the fix in a running stack (both cases correct).
- `test_tag_inheritance` (39), `test_tags` (26), `test_tag_inheritance_perf` (24) — all green.

### DIFF
--- a/dojo/tags/signals.py
+++ b/dojo/tags/signals.py
@@ -3,6 +3,7 @@ import logging
 
 from django.db.models import signals
 from django.dispatch import receiver
+from tagulous.models.fields import SingleTagField, TagField
 
 from dojo.celery_dispatch import dojo_dispatch_task
 from dojo.location.models import Location, LocationFindingReference, LocationProductReference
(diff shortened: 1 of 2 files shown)


Maffooch (approved):

Jino-T (approved):

mtesauro (approved):
Approved

paulOsinski (approved):

rossops (approved):

PR #15103, opened 2026-06-29
perf(finding): add sla_expiration_date index for global finding list

rossops:
## Summary

Adds a full (non-partial) btree on `dojo_finding.sla_expiration_date`.

The global finding list ordered by `sla_expiration_date` was **seq-scanning and sorting the entire authorized finding set** on every page load (the order column had no walkable index). This index lets the planner walk it and stop at the `LIMIT` instead.

The existing partial `idx_finding_sla_open_cov` (`WHERE NOT is_mitigated`) **cannot** serve this query because the query has no `is_mitigated` predicate — hence a plain full index.

## Performance (measured on a 10M-finding clone, under load test)

Query: global finding list `ORDER BY sla_expiration_date ASC LIMIT 25` with the per-user authorization filter.

| | Plan | Mean | Max | Total |
|---|---|---|---|---|
| **Before** | Parallel Seq Scan (3.4M rows) → hash join → sort | 6591 ms | 20732 ms | 1582 s |
| **After** | Index Scan on `idx_finding_sla_exp` → nested-loop probe → stop at 25 | **115 ms** | 4992 ms | 96 s |

~57× mean improvement; the query dropped from the **#1** offender to negligible. Index size ~68 MB on the clone.

## Notes

- Built `CONCURRENTLY` (`atomic = False`) → no `ACCESS EXCLUSIVE` lock on the large `dojo_finding` table.
- Follows the pattern established in #15095 (`0271_finding_perf_indexes`), which this depends on.
- Companion `models.Index` added to `Finding.Meta.indexes` to keep model/migration state in sync.


Maffooch (approved):

mtesauro (approved):
Approved

valentijnscholten (approved):

Jino-T (approved):

Maffooch:
I'm gonna rebase this to dev so that the migrations line up a bit easier for the release on Monday

PR #15106, opened 2026-06-29
feat(ui): smooth and tidy the filter category accordion in the new UI

ksitton58:
**Description**

Addresses maintainer feedback that the new-UI filter panel felt "funky" to expand/collapse and was hard to read.

The filter categories used native `<details>`: the chevron animated but the body snapped open/closed, the header bars jumped to new positions instead of sliding together, and the categories packed inconsistently two-per-row (the filter form is `display: flex; flex-wrap: wrap`).

Changes:
- Convert each category to a small Alpine toggle (a `<button>` with `aria-expanded` and `type="button"`, so it never submits the filter form). The body sits in an overflow-clipped grid that animates `grid-template-rows: 0fr` to `1fr`. Animating the real layout height makes the whole header stack slide together smoothly and collapse fully to 0. This works in all browsers (no `interpolate-size` dependency). Default-open groups server-render the open state, so there is no flash on load.
- Give each category `flex: 0 0 100%` so they stack as full-width rows instead of packing two-per-row, making the panel easy to scan.

No filter behavior changes: same fields, same form submission.

**Test results**

Verified on a local build at the Findings filter panel: categories stack as full-width rows; toggling slides the body and the header stack together (collapses to 0px, expands to content height); clicking a header does not submit the form (`type="button"`); `aria-expanded` reflects state. No Python changed, so no unit tests added.

**Documentation**

N/A.

**Checklist**

- [x] Meaningful PR name.

### DIFF
--- a/dojo/static/dojo/css/dojo.css
+++ b/dojo/static/dojo/css/dojo.css
@@ -1084,8 +1084,9 @@ span.endpoint_product {
     .filter-form-group > .col-lg-3 { grid-column: span 3; }
 }
 
-/* Filter category grouping (details/summary) */
+/* Filter category grouping (Alpine-toggled accordion) */
 .filter-category {
+    flex: 0 0 100%;
     border-bottom: 1px solid var(--color-border, #DCDCDC);
 }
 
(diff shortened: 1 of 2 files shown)


mtesauro (approved):
Approved

Maffooch (approved):

blakeaowens (approved):

valentijnscholten (approved):

PR #15110, opened 2026-06-30
chore: replace archived upload-release-asset with gh release upload

Maffooch:
Replaces the GitHub-archived `actions/upload-release-asset@v1.0.2` in `release-drafter.yml`, which runs on the deprecated **node12** runtime and has no Node 24 release (so it can't be version-bumped — it must be replaced).

The two upload steps are collapsed into a single `gh release upload` step (GitHub CLI is preinstalled on runners): no action runtime, nothing to pin, and immune to future node deprecation. `softprops/action-gh-release@v2` was rejected because it runs on node20 and would re-introduce a soon-deprecated runtime.

The job now surfaces release-drafter's `tag_name` output (used to target the draft release) instead of the now-unused `upload_url`. Artifact paths are unchanged; `--clobber` keeps re-runs idempotent.

Follow-up to the org-wide GitHub Actions cleanup (Node 20 runtime work).

Shortcut: [sc-13332](https://app.shortcut.com/defectdojo/story/13332)


mtesauro (approved):
Approved

dogboat (approved):

valentijnscholten (approved):

valentijnscholten:
I have now way of testing this. Did you manage to test it?

Maffooch:
> I have now way of testing this. Did you manage to test it?

I have used this mechanism in other repos

Jino-T (approved):

PR #15112, opened 2026-06-30
fix(mass_model_updater): never skip a deferred tracked field as "unchanged" [sc-13375]

Maffooch:
## Problem

`mass_model_updater(..., skip_unchanged=True)` (added in #15046) decides whether to write each row by snapshotting its tracked `fields` **before** calling `function`, reading `model.__dict__.get(field)` to avoid triggering a deferred-field query.

But when a tracked field is **deferred** — the caller's queryset used `.only()`/`.defer()` and omitted it — `__dict__` has no entry, so the snapshot reads `None`. If `function` then recomputes that field to `None`, the comparison `None == None` marks the row "unchanged" and the **write is silently dropped**, leaving the stale persisted value in the DB.

This bit DefectDojo Pro's risk-mode SLA recalculation: its queryset deferred `sla_expiration_date` while updating it, so clearing a finding's SLA to `None` (enforcement disabled) never persisted — the finding kept its old SLA. (Companion fix in dojo-pro adds the field to its `.only(...)`; this PR removes the footgun at the source.)

## Fix

Guard the no-op check with `model.get_deferred_fields()`: if any tracked field is deferred, skip the comparison and always write that row. No extra query; non-deferred rows keep the fast skip-unchanged path.

## Tests

Adds two regression tests to `unittests/test_mass_model_updater.py`:
- `test_writes_deferred_field_recomputed_to_none` — a deferred field recomputed to `None` must issue an UPDATE and persist (fails without this fix: 0 UPDATEs).
- `test_writes_deferred_field_recomputed_to_value` — deferred field updated to a value persists.

Full `TestMassModelUpdater` suite (11 tests) passes; reverting the guard makes the None test fail with `0 not greater than 0`.

## Scope

`skip_unchanged` exists only on `dev` (not in 3.0.200 / `bugfix` / `master`), so released versions are unaffected.

### DIFF
--- a/dojo/utils.py
+++ b/dojo/utils.py
@@ -1647,7 +1647,11 @@ def mass_model_updater(model_type, models, function, fields, page_size=1000, ord
     When ``fields`` is given:
       - skip_unchanged (default True): rows whose tracked ``fields`` were not changed by
         ``function`` are not written (compared against the values loaded from the page
-        query; deferred fields are read from ``__dict__`` so no extra query is issued).
+        query, read from ``__dict__`` so no extra query is issued). If any tracked field
+        is deferred (e.g. the queryset used ``.only()``/``.defer()`` and omitted it) the
+        no-op check is skipped for that row and it is always written, because ``__dict__``
+        has no loaded value to compare against and would otherwise read as ``None`` --
+        silently dropping a real change to ``None``.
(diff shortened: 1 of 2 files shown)


mtesauro (approved):
Approved

dogboat (approved):

Maffooch:
Closing in favor or #15114

PR #15114, opened 2026-06-30
fix(mass_model_updater): read tracked fields via attribute access for skip_unchanged

valentijnscholten:
## Summary

- `mass_model_updater(..., skip_unchanged=True)` (added in #15046) snapshotted each tracked field with `model.__dict__.get(f)` before calling `function`, to avoid triggering a deferred-field load when deciding whether a row changed.
- When a tracked field is **deferred** — the caller's queryset used `.only()`/`.defer()` and omitted it — `__dict__` has no entry, so the snapshot reads `None`. If `function` recomputes that field to `None`, the comparison `None == None` marks the row "unchanged" and the **write is silently dropped**, leaving a stale persisted value.
- Root cause is a caller that defers a field it then updates (e.g. a `.only(...)` that omits the very field passed in `fields=`). The previous `__dict__` approach turned that caller mistake into silent data loss instead of a visible cost.
- Fix: read the pre-`function` value of each tracked field with normal attribute access (`getattr`) instead of `__dict__`. A deferred tracked field is loaded and compared against its real persisted value, so a real change can never be mistaken for a no-op. The skip-unchanged optimization keeps working for correctly-loaded querysets; a deferred tracked field now degrades to a visible per-row load rather than silent incorrectness. Documented that callers needing the fast path must load the tracked fields.
- Adds two regression tests to `unittests/test_mass_model_updater.py`: a deferred tracked field recomputed to `None` must issue an UPDATE and persist; a deferred tracked field left unchanged must still skip the write.
- `skip_unchanged` exists only on `dev` (not in released versions), so released versions are unaffected.

Alternative to #15112, which guards the no-op check with `get_deferred_fields()` and always writes deferred rows. This PR instead removes the `__dict__` indirection that caused the bug, so the no-op check stays correct (and the optimization keeps working) for deferred-but-loaded rows.

### DIFF
--- a/dojo/utils.py
+++ b/dojo/utils.py
@@ -1646,8 +1646,10 @@ def mass_model_updater(model_type, models, function, fields, page_size=1000, ord
 
     When ``fields`` is given:
       - skip_unchanged (default True): rows whose tracked ``fields`` were not changed by
-        ``function`` are not written (compared against the values loaded from the page
-        query; deferred fields are read from ``__dict__`` so no extra query is issued).
+        ``function`` are not written. The pre-``function`` value of each tracked field is
+        read with normal attribute access, so callers must ensure the tracked ``fields``
+        are loaded by the queryset (i.e. not excluded via ``.only()``/``.defer()``);
+        otherwise accessing a deferred tracked field issues a per-row query.
       - writer (optional): a callable ``writer(model_type, batch, fields)`` used to persist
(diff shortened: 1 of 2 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

dogboat (approved):

Jino-T (approved):

PR #15116, opened 2026-06-30
feat(ui): animate collapse panels in the new UI

ksitton58:
**Description**

Collapse panels in the new UI (finding-detail sections like Description / Mitigation / Impact, and other `data-toggle="collapse"` panels) toggled open and closed instantly with no animation, which felt abrupt. This adds a smooth height slide, consistent with the filter accordion.

Changes:
- `index.js`: the shared collapse shim now animates the target's height and vertical padding between 0 and its natural size over 250ms, then settles to `auto` (open) or `display:none` via `.in` (closed). It respects `prefers-reduced-motion` (instant toggle), preserves `aria-expanded`, guards against re-clicks mid-animation, and handles both `border-box` and `content-box`. One change covers every collapse panel in the new UI, with no per-panel template changes.
- `base.html`: cache-bust `index.js` with `?v={% static_v %}` (matching how the CSS is loaded) so this JS change actually reaches users' browsers. First-party JS was not versioned before.

No Bootstrap is reintroduced. This enhances the new UI's own vanilla-JS shim that already replaced Bootstrap's collapse.

**Test results**

Verified on a local build (finding detail page): clicking a section animates its height smoothly from 0 to full and back (sampled per frame: steady 0 to 56px on open, 56 to 0px on close, no stall or jump), `aria-expanded` toggles correctly, and the form is not submitted. No Python changed, so no unit tests added.

**Documentation**

N/A.

**Checklist**

- [x] Meaningful PR name.

### DIFF
--- a/dojo/static/dojo/js/index.js
+++ b/dojo/static/dojo/js/index.js
@@ -69,6 +69,67 @@
    Handles [data-toggle="collapse"] by toggling .in on the target element.
    CSS in tailwind.css:  .collapse { display:none }  .collapse.in { display:block }
 */
+/* Animate a collapse target open/closed by sliding its height between 0 and
+   its natural size, then settling to auto (open) or display:none via .in
+   (closed). Falls back to an instant toggle when reduced motion is requested. */
+function animateCollapse(target, toggle) {
+    if (target._ddCollapsing) return;  // ignore clicks while mid-animation
+    var willOpen = !target.classList.contains('in');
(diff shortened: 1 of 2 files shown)


mtesauro (approved):
Approved

Maffooch (approved):

valentijnscholten (approved):
I usually like fast and no animations, but I guess I'm not your default user. Approved :-)

blakeaowens (approved):

PR #15137, opened 2026-07-01
Fix: right-aligned dropdown menus overflow off the right edge of the page

### LINKED ISSUE #15125: Bug: Action dropdown menus render off-screen in V3 UI
## Description

In the V3 UI, some expandable action menus render outside the visible viewport, making menu items partially or completely inaccessible.

This appears to happen on finding detail pages where the action dropdown is positioned near the right edge of the screen. Instead of opening within the viewport, the menu extends beyond the page boundary.

## Steps to Reproduce

1. Open a Finding in the V3 UI.
2. Click the action dropdown (☰) in the upper-right corner.
3. Observe the position of the expanded menu.

## Expected Behavior

The dropdown menu should automatically position itself so that it remains fully visible within the browser viewport.

## Actual Behavior

The menu renders off-screen, causing menu items to be clipped and difficult or impossible to access.

## Screenshot

The attached screenshot shows the action menu extending beyond the right edge of the page.

Image

## Impact

Users cannot reliably access all actions from the dropdown, especially on smaller screens or when the menu is located close to the viewport edge.

## Possible Cause

The dropdown positioning logic does not appear to account for available viewport space. The menu should "flip" or align to the left when there is insufficient room to render on the right, as is common behavior in Popper-based dropdowns.

stevewallone:
**Description**

Fixes #15125.

When `bootstrap.min.css` was dropped, the essential Bootstrap dropdown rules
were re-implemented by hand in `components/tailwind.css` (the block commented
"Bootstrap dropdown compat — since bootstrap.min.css is not loaded…"). That
compat block recreated `.dropdown-menu` but **omitted the `.dropdown-menu-right`
utility** (`right: 0; left: auto`).

As a result, every menu marked `class="dropdown-menu dropdown-menu-right"` lost
its right-alignment and fell back to left-alignment (`left: 0`), opening to the
right of its toggle. Where the toggle sits flush against the right edge of the
page — most visibly the **finding actions menu** on the View Finding page — the
menu overflows off the right edge, clipping its labels ("Edit Fin…", "Copy
Fin…") and forcing the whole page into horizontal scroll.

This affects **42 usages across 30+ templates** (findings, products,
engagements, tests, endpoints, users, metrics, …), not just the finding page.

The fix restores the single missing utility in the Bootstrap-compat block and
rebuilds `tailwind-out.css`:

```css
.dropdown-menu-right { right: 0; left: auto; }
```

The sibling `.dropdown-menu-left` variant has **0** usages in the templates, so
it is intentionally not added (keeping the change minimal).

**Test results**

No Python was touched, so there is no unit test to add — this is a pure
CSS/build change and DefectDojo has no CSS test harness. Verified manually in
the running app (`docker-compose`), before and after, on `/finding/<id>`:

- **Before:** menu right edge at x=1645 vs viewport 1435 → overflows; page
  `scrollWidth` 1645 (horizontal scroll present); labels clipped.
- **After:** menu right-aligns to the toggle, right edge x=1552 vs viewport
  1589 → fully on-page; `scrollWidth` == viewport (no horizontal scroll); all
  labels readable.

The rebuilt `tailwind-out.css` diff is exactly one inserted rule
(`.dropdown-menu-right{left:auto;right:0}`) — built with tailwindcss 4.2.4 to
match the version that produced the committed artifact, so no unrelated churn.
`makemigrations --check` is not applicable (no model changes).

**Documentation**

None — internal CSS regression, no user-facing docs reference this.

**Checklist**

- [x] Meaningful PR name for release notes.
- [ ] Unit tests — N/A (CSS-only; verified manually with before/after evidence).
- [x] Categorised as a bugfix — please add the `bugfix` label (fork PRs can't set labels).


Maffooch (approved):

mtesauro (approved):
Approved

blakeaowens (approved):

valentijnscholten (approved):

PR #15139, opened 2026-07-01
fix: guard legacy endpoint access in Make Template / Merge Findings under V3_FEATURE_LOCATIONS

### LINKED ISSUE #15123: Bug: Creating a Finding Template from an Existing Finding or Merging Findings fails when `V3_FEATURE_LOCATIONS` is enabled and Endpoint Data exists
## Description

Creating a new Finding Template from an existing Finding fails with an Internal Server Error when `V3_FEATURE_LOCATIONS` is enabled.

The `mktemplate` view attempts to access the deprecated `Endpoint` model (`finding.endpoints.all()`), which raises a `NotImplementedError` when the V3 locations feature flag is enabled.

## Steps to Reproduce

1. Enable `V3_FEATURE_LOCATIONS`.
2. Open an existing Finding.
3. Select **Create Template** (or navigate to `/finding/<id>/mktemplate`).
4. Observe the error.

## Expected Behavior

A Finding Template should be created successfully from the existing Finding, regardless of whether `V3_FEATURE_LOCATIONS` is enabled.

## Actual Behavior

An Internal Server Error is returned.

```
[29/Jun/2026 14:05:08] ERROR [django.request:253] Internal Server Error: /finding/37675/mktemplate
  File "/app/dojo/finding/views.py", line 1750, in mktemplate
    endpoint_urls = [str(ep) for ep in finding.endpoints.all()]
                                       ~~~~~~~~~~~~~~~~~~~~~^^
...
  File "/app/dojo/models.py", line 1685, in __init__
    raise NotImplementedError(msg)
NotImplementedError: Endpoint model is deprecated when V3_FEATURE_LOCATIONS is enabled
```

## Root Cause

The `mktemplate` view still references the deprecated `Endpoint` model:

```python
endpoint_urls = [str(ep) for ep in finding.endpoints.all()]
```

When `V3_FEATURE_LOCATIONS` is enabled, instantiating the `Endpoint` model raises a `NotImplementedError`, causing the request to fail.

## Suggested Fix

Update the template creation logic to use the V3 Locations model/API instead of the deprecated `finding.endpoints` relationship, or make the code compatible with both legacy Endpoints and V3 Locations during the migration period.

stevewallone:
Fixes #15123.

**Problem:** Creating a Finding Template from a finding ("Make Template") or merging findings returns a 500 when `V3_FEATURE_LOCATIONS` is enabled and the finding still carries legacy `Endpoint` rows. The locations migration deliberately keeps legacy endpoint rows as backup, so any finding created before the migration trips this; findings created after V3 are unaffected — which is why the crash is data-dependent.

**Root cause:** two sites in `dojo/finding/views.py` iterate `finding.endpoints.all()` without the `Endpoint.allow_endpoint_init()` escape hatch. Iterating the queryset hydrates `Endpoint` instances via `Model.from_db()` → `Endpoint.__init__` raises `NotImplementedError: Endpoint model is deprecated when V3_FEATURE_LOCATIONS is enabled`:

- `mktemplate` — `[str(ep) for ep in finding.endpoints.all()]`
- `merge_finding_product` — `finding_to_merge_into.endpoints.add(*finding.endpoints.all())`

**Fix:** wrap both sites in `with Endpoint.allow_endpoint_init():`, matching the existing convention at the other legacy endpoint call sites (`product/views.py`, `product_type/views.py`, `api_v2/views.py`, `api_v2/mixins.py`), each tagged `# TODO: Delete this after the move to Locations`. This is also the approach the reporter (@DarkR0ast) suggested.

**Tests:** `unittests/test_finding_template_merge_endpoints_v3.py` — regression tests modeled on `test_delete_with_endpoints_v3.py`: a finding carrying a legacy endpoint row under `V3_FEATURE_LOCATIONS=True`, exercising both views. Both error with the production `NotImplementedError` before the fix and pass after; endpoint data is verified copied (template `endpoints_text` / merged finding's endpoints). Also verified manually in a local instance against both UI flows.

**Note on scope:** this restores the pre-V3 merge/template behavior for legacy endpoint data (no more 500). Two adjacent gaps exist but look like design decisions beyond this issue: (1) legacy `finding.endpoints` rows are not displayed anywhere in the V3 UI (the finding page renders only `finding.locations`), so merged legacy endpoints are invisible under V3; (2) the merge and mktemplate views do not carry over V3 `finding.locations` at all, even though `Location.associate_with_finding()` exists. Happy to file/take a follow-up issue for a locations-aware merge if wanted.

### DIFF
--- a/dojo/finding/views.py
+++ b/dojo/finding/views.py
@@ -76,6 +76,7 @@ from dojo.models import (
     IMPORT_UNTOUCHED_FINDING,
     BurpRawRequestResponse,
     Dojo_User,
+    Endpoint,
     Endpoint_Status,
     Engagement,
     FileAccessToken,
@@ -1747,7 +1748,8 @@ def mktemplate(request, fid):
 
         # Copy endpoints if they exist
         if finding.endpoints.exists():
-            endpoint_urls = [str(ep) for ep in finding.endpoints.all()]
+            with Endpoint.allow_endpoint_init():  # TODO: Delete this after the move to Locations
+                endpoint_urls = [str(ep) for ep in finding.endpoints.all()]
             finding_helper.save_endpoints_template(template, endpoint_urls)
 
         messages.add_message(
@@ -2333,9 +2335,10 @@ def merge_finding_product(request, pid):
 
                         # if checked merge the endpoints
                         if form.cleaned_data["add_endpoints"]:
-                            finding_to_merge_into.endpoints.add(
-                                *finding.endpoints.all(),
-                            )
+                            with Endpoint.allow_endpoint_init():  # TODO: Delete this after the move to Locations
+                                finding_to_merge_into.endpoints.add(
+                                    *finding.endpoints.all(),
+                                )
 
                         # if checked merge the tags
                         if form.cleaned_data["tag_finding"]:
--- a/unittests/test_finding_template_merge_endpoints_v3.py
+++ b/unittests/test_finding_template_merge_endpoints_v3.py
@@ -1,0 +1,142 @@
+"""
+Regression tests for issue #15123: the V3_FEATURE_LOCATIONS crash when creating a
+Finding Template from a finding or merging findings.
+
+When ``V3_FEATURE_LOCATIONS`` is enabled the legacy ``Endpoint`` model is deprecated and
+``Endpoint.__init__`` raises ``NotImplementedError``. Findings created before the V3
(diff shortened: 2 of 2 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

valentijnscholten (approved):
Thanks @stevewallone for your PRs!

blakeaowens (approved):

PR #15147, opened 2026-07-02
fix(risk_acceptance): reinstate findings when expiration date updated via API

Jino-T:
[sc-13436]

## Summary

- `RiskAcceptanceSerializer.update()` never called `ra_helper.reinstate()` when `expiration_date` changed, unlike the legacy Django view (`engagement/views.py:1308-1310`) which does so correctly.
- This caused two related bugs:
  1. **Findings stay Active after extending an expired RA** — when a user edits the expiration date from a past date to a future date via the API (Edit Risk Acceptance form in the UI), `reinstate()` is never called, so findings remain `active=True / risk_accepted=False` instead of being set back to inactive/risk-accepted.
  2. **Findings stay Inactive on subsequent expiry cycles** — because `reinstate()` was never called, `expiration_date_handled` was never cleared. The Celery expiration task queries with `expiration_date_handled__isnull=True`, so the RA is permanently excluded from every future expiry run and findings are never reactivated when the RA expires again.

## Fix

Capture `old_expiration_date` before `super().update()`, then call `ra_helper.reinstate()` when the date changes — matching the logic that already exists in the legacy Django view.

```python
old_expiration_date = self.instance.expiration_date
instance = super().update(instance, validated_data)
...
if instance.expiration_date != old_expiration_date:
    ra_helper.reinstate(instance, old_expiration_date)
```

`reinstate()` handles both outcomes internally: it reinstates findings (only when `expiration_date_handled` is set) and unconditionally clears `expiration_date_handled` and `expiration_date_warned`, making the RA eligible for the next Celery expiry cycle.

## Test plan

- [ ] Edit a Risk Acceptance with an expired date to a future date via the API — verify findings become `active=False / risk_accepted=True`
- [ ] Confirm `expiration_date_handled` is `None` after the update — verify the Celery task will pick it up on the next expiry cycle
- [ ] Edit a fresh (never-expired) RA's date — verify findings are unchanged and `expiration_date_handled` stays `None`


mtesauro (approved):
Approved

Maffooch:

Validated end-to-end via the REST API against a running instance:

- **Reproduced the bug on `bugfix`:** created a Risk Acceptance in the expired + `expiration_date_handled` state with an active accepted finding, then PATCHed `expiration_date` to a future date via `/api/v2/risk_acceptance/{id}/`. The finding stayed `active=True / risk_accepted=False` and `expiration_date_handled` remained set — so the RA would be permanently skipped by the Celery expiry task (both bugs in the description confirmed).
- **On this branch:** the same update (`expiration_date` + `accepted_findings`) correctly calls `reinstate()` — the finding flips to `active=False / risk_accepted=True` and `expiration_date_handled` is cleared (RA becomes eligible for the next expiry cycle). Matches the legacy Django view.

One follow-up (pre-existing, **not** introduced here): a partial PATCH sending only `expiration_date` (omitting `accepted_findings`) strips the findings from the RA before reinstate runs, so reinstate has nothing to reactivate. Not a blocker for this fix, but worth a separate look.

valentijnscholten (changes requested):
Thanks — the fix itself is correct and faithfully mirrors the legacy UI view (`engagement/ui/views.py:1294,1302-1304`): capture `old_expiration_date` before `super().update()`, then call `ra_helper.reinstate()` when it changes. I verified `reinstate()` is safe for both paths — a never-expired RA just clears the already-`None` flags (findings untouched), and an expired one reinstates the findings and clears `expiration_date_handled`, so the Celery expiry task will pick it up again. Both documented bugs are addressed.

**Requesting one change before merge: please add a regression test.** This is a behavioral bugfix with a clear, mechanical reproduction, and it lives in exactly the kind of path (serializer `update()` + Celery-eligibility flags) that silently regresses. `unittests/test_risk_acceptance.py` is the natural home. Suggested coverage:

1. Edit an **expired** RA's `expiration_date` to a future date via the API → assert its accepted findings become `active=False` / `risk_accepted=True`, and `expiration_date_handled is None` afterwards (so the expiry task will re-process it).
2. Edit a **never-expired** RA's date → assert findings are unchanged and `expiration_date_handled` stays `None`.

Non-blocking observation (existing behavior, not introduced here): `reinstate()` overwrites `expiration_date` with `now() + risk_acceptance_form_default_days` (90) whenever `expiration_date_handled` was set — so an API client PUTting a specific future date on a previously-expired RA won't get that exact date back, it'll get now()+90. It matches the legacy UI (shared helper), so this PR is consistent, but it's a surprise for API consumers worth a follow-up/docstring at some point.

Minor/cosmetic: you read `self.instance.expiration_date` for the old value but `instance.expiration_date` for the new — same object, works; using `instance` for both is marginally cleaner.

valentijnscholten (approved):

dogboat (approved):

PR #15148, opened 2026-07-03
Add report API formats

### LINKED ISSUE #12018: Add type extensions to findings reports via API
This suggestion is partly related to [issue 12015](https://github.com/DefectDojo/django-DefectDojo/issues/12015) (has the same context).

To retrieve all findings and generate the report, I was planning to automate the process through the API using the `/generate_report/` endpoint (which is different from issue 12015). The problem is that this endpoint only returns data in JSON format, making it difficult to interpret.

When navigating through the classic UI, I can see that several format options are available, such as HTML Report, CSV, and Excel.

Image

My idea is to add an option (a <select> element in the UI) that allows users to choose the report format. There is a simple proof of concept (PoC) using the browser's Inspect tool.

Image2

samiat4911:
**Description**

Closes #12018.

This PR adds selectable output formats to the report generation API. The existing API behavior remains the default by keeping JSON as the default `report_type`, while callers can now request `HTML`, `CSV`, or `Excel`.

The request serializer now accepts a `report_type` field:

```python
class ReportGenerateOptionSerializer(serializers.Serializer):
    REPORT_TYPE_CHOICES = (
        ("JSON", "JSON"),
        ("HTML", "HTML"),
        ("CSV", "CSV"),
        ("Excel", "Excel"),
    )

    include_finding_notes = serializers.BooleanField(default=False)
    include_finding_images = serializers.BooleanField(default=False)
    include_executive_summary = serializers.BooleanField(default=False)
    include_table_of_contents = serializers.BooleanField(default=False)
    report_type = serializers.ChoiceField(
        choices=REPORT_TYPE_CHOICES,
        default="JSON",
        help_text="Format for the generated report.",
    )
```

The shared API report helper now formats the response based on the requested type:

```python
def report_generate_response(request, obj, options):
    report_type = options.get("report_type", "JSON")
    data = report_generate(request, obj, options)

    if report_type == "JSON":
        report = serializers.ReportGenerateSerializer(data)
        return Response(report.data)

    if report_type == "HTML":
        return render(request, data["report_template"], data)

    if report_type == "CSV":
        return CSVExportView().build_response(data["findings"])

    if report_type == "Excel":
        return ExcelExportView().build_response(data["findings"])

    msg = f"Unsupported report_type: {report_type}"
    raise ValidationError(msg)
```

The existing CSV and Excel UI export views were refactored slightly so their response builders can be reused by the API without changing the existing UI export routes:

```python
    def get(self, request):
        findings, _obj = get_findings(request)
        findings = prefetch_related_findings_for_report(findings)
        self.findings = findings
        findings = self.add_findings_data()
        return self.build_response(findings)

    def build_response(self, findings):
```

The four existing report generation API actions now pass the validated `report_type` into the shared response helper:

```python
            options["report_type"] = report_options.validated_data["report_type"]
```

During testing, API requests exposed that `report_url_resolver()` could raise `KeyError` when `HTTP_HOST` was not present in request metadata. This PR adds a safe fallback:

```python
        if "HTTP_HOST" not in request.META:
            return request.build_absolute_uri("/").rstrip("/")
```

**Test results**

Added targeted API tests covering:

- Default JSON report output.
- HTML report output.
- CSV report output.
- Excel report output.
- Invalid `report_type` validation.

Test class added:

```python
@versioned_fixtures
class ReportGenerateFormatAPITest(DojoAPITestCase):
    fixtures = ["dojo_testdata.json"]
```

**Documentation**

Please update any documentation when needed in the [documentation folder](https://github.com/DefectDojo/django-DefectDojo/tree/dev/docs))

**Checklist**

- [ ] Your code is Ruff compliant (see [ruff.toml](../ruff.toml)).


valentijnscholten (commented):
Should there be some filtering on authorized users/products here?

samiat4911:
hello @valentijnscholten yes, I think there should be, but my understanding is that this is already covered by the existing viewset querysets before the report formatting runs. The new `report_type` logic reuses the existing object/queryset passed into `report_generate_response()`:
 - Product/engagement/test reports call `get_object()`, so they are limited by each viewset's authorized queryset.
- Findings reports call `self.get_queryset()`, so they are limited to authorized findings.
- CSV/Excel/HTML then reuse the already-filtered `data["findings"]` queryset.

So I don’t think this PR needs an extra authorization filter in the formatter itself unless you prefer an additional defensive check there too.

samiat4911:
I'm open to any suggestions @valentijnscholten thank you :)

valentijnscholten:
Thanks @samiat4911 — you're right that the report endpoints inherit authorization from the viewsets: the finding report goes through `self.get_queryset()` (which applies `get_authorized_findings`), and product/engagement/test go through `self.get_object()`, so the data is already scoped before formatting. This PR only adds output formats on top of that, so no extra filter is needed in the formatter itself. 👍

One defense-in-depth note on the `Product_Type` branch of `report_generate()` though: it builds its data with raw managers rather than the authorized helpers —

- `findings = ... Finding.objects.filter(test__engagement__product__prod_type=product_type)`
- the new `report_context` uses `Product.objects.filter(...)`, `Engagement.objects.filter(...)`, `Test.objects.filter(...)`

That's safe today because entry is gated by `get_object()` on the Product_Type (so access cascades), but unlike the finding path it doesn't scope explicitly with `get_authorized_*`. Could you either switch those to `get_authorized_findings` / `get_authorized_products` (etc.), or add a short comment noting authorization is enforced upstream by `get_object()`? That way a future refactor or a new caller of this branch can't accidentally leak cross-tenant data.

valentijnscholten (approved):
Thanks — this looks good, and the authorization concern is fully addressed (the `Product_Type` branch now scopes via `get_authorized_findings/products/engagements/tests("view")` instead of raw `.objects`). Verified the rest holds up:

- `get_authorized_tests` is imported, so the product_type path is fine.
- `report_generate_response` reuses `CSVExportView/ExcelExportView.build_response()`, and those helpers don't reference `self.request`, so calling them on a bare instance is safe.
- `add_findings_data()` is a no-op passthrough, so the API path skipping it drops nothing.
- `report_type` defaults to `JSON`, keeping existing consumers backward compatible.
- Tests cover all four formats plus the unknown-type → 400 case.

Approving. Two optional, non-blocking improvements for a follow-up (or now if you like):

1. The trailing `raise ValidationError(f"Unsupported report_type: {report_type}")` in `report_generate_response` runs only *after* `report_generate()` has already done all the work, and it's effectively unreachable since the `ChoiceField` rejects invalid values with a 400 before we get here. Consider validating the report type up front (or dropping the guard) so we don't build the report just to discard it.

2. The CSV/Excel responses hardcode `filename=findings.csv` / `findings.xlsx` regardless of the report object. A name derived from the object (e.g. `product_<id>_findings.csv`) would be friendlier, but purely cosmetic.

Maffooch (approved):

PR #15149, opened 2026-07-03
fix(importers): stop doubling the (scan_type) suffix in dynamic Test Type names

Maffooch:
**Shortcut:** [sc-13441](https://app.shortcut.com/defectdojo/story/13441)

## Summary

Fixes a bug report: Generic Findings Import created doubled Test Type names such as `Prisma Cloud (Generic Findings Import) Scan (Generic Findings Import)`.

Dynamic-test-type parsers (Generic, SARIF, and ~9 others) build the `Test_Type` name by appending ` Scan ({scan_type})` to the report's `type` field. The composition was not idempotent: when `type` already ended with the ` ({scan_type})` suffix, it was appended again, producing the doubled name.

## What changed

- **`resolve_dynamic_test_type_name()`** (new): single, idempotent name resolver. A `type` already ending in ` ({scan_type})` is used verbatim; everything else keeps the intentional `{type} Scan ({scan_type})` format (so `Tool1` → `Tool1 Scan (Generic Findings Import)`, and SARIF is unchanged).
- **`legacy_dynamic_test_type_name()`** (new): reproduces the pre-patch name. The reimport validation accepts **either** the new or the legacy name, so **reimports into tests created with the old doubled name keep working** — no data migration, no new `Test type mismatch` errors.
- **Docs**: corrected `generic_findings_import.md`, `OS__tests.md`, `PRO__tests.md`, `file/generic.md`, and added the idempotency rule to `OS__asset_hierarchy.md`. The old docs described a `{Test Name} (Generic Findings Import)` format the code never produced, which is what set the customer's expectation.

## Testing

TDD (failing test first, confirmed red reproducing the doubled name, then green):

- New fixture `unittests/scans/generic/generic_test_type_suffix.json` + 3 regression tests: no-doubling on import, idempotent reimport, and pre-patch doubled-name reimport still works.
- Full importer module (43), SARIF parser (25), and deduplication (50) suites pass; ruff clean.
- HTTP e2e on a running instance: `type: "Prisma Cloud (Generic Findings Import)"` → `Prisma Cloud (Generic Findings Import)` (no doubling); `type: "Tool1E2E"` → `Tool1E2E Scan (Generic Findings Import)` (unchanged); reimport → HTTP 201, no mismatch.

### DIFF
--- a/docs/content/asset_modelling/OS_hierarchy/OS__asset_hierarchy.md
+++ b/docs/content/asset_modelling/OS_hierarchy/OS__asset_hierarchy.md
@@ -159,6 +159,7 @@ Examples include:
 **Report-Defined Test Type Naming Rules:**
 - If the report's `type` field equals the scan type → uses scan type directly (e.g., "Generic Findings Import")
 - If the report's `type` field differs → creates "{type} Scan ({scan_type})" format (e.g., "Tool1 Scan (Generic Findings Import)")
+- If the report's `type` field already ends with the " ({scan_type})" suffix → uses it verbatim, so the suffix is never doubled (e.g., "Tool1 (Generic Findings Import)" stays "Tool1 (Generic Findings Import)")
 - If no `type` field is provided → uses scan type directly
 
 **Important Considerations:**
--- a/docs/content/asset_modelling/engagements_tests/OS__tests.md
+++ b/docs/content/asset_modelling/engagements_tests/OS__tests.md
@@ -37,7 +37,7 @@ While each method differs primarily in how scan data is parsed and ingested, the
 
 When no native parser exists for a given tool, **Generic Findings Import** allows you to import findings using a standardized JSON or CSV schema, regardless of the original source. 
 
-DefectDojo parses the provided data, creates a new Test (or imports into an existing one), and attaches the Findings. A corresponding Test Type is also created in the format “{Test Name} (Generic Findings Import).”
+DefectDojo parses the provided data, creates a new Test (or imports into an existing one), and attaches the Findings. A corresponding Test Type is also created based on the report's optional `type` field: when `type` is omitted (or equals the scan type) the Test Type is “Generic Findings Import”; when `type` is provided it becomes “{type} Scan (Generic Findings Import)” (a `type` that already ends with the “(Generic Findings Import)” suffix is used verbatim).
 
 |  | **Native Parsers** | **Generic Findings Import** | 
 |----------|---------------|------------------------|
--- a/docs/content/asset_modelling/engagements_tests/PRO__tests.md
+++ b/docs/content/asset_modelling/engagements_tests/PRO__tests.md
@@ -35,7 +35,7 @@ While each method differs primarily in how scan data is parsed and ingested, the
 
 When no native parser exists for a given tool, [**Generic Findings Import**](/supported_tools/parsers/generic_findings_import) allows you to import findings using a standardized JSON or CSV schema, regardless of the original source. 
 
-DefectDojo parses the provided data, creates a new Test (or imports into an existing one), and attaches the Findings. A corresponding Test Type is also created in the format “`{Test Name}` (Generic Findings Import).”
+DefectDojo parses the provided data, creates a new Test (or imports into an existing one), and attaches the Findings. A corresponding Test Type is also created based on the report's optional `type` field: when `type` is omitted (or equals the scan type) the Test Type is “Generic Findings Import”; when `type` is provided it becomes “`{type}` Scan (Generic Findings Import)” (a `type` that already ends with the “(Generic Findings Import)” suffix is used verbatim).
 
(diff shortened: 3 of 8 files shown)


valentijnscholten (commented):
Does auto create context need to be aware of old and new formats?
Does deduplication need to be aware of this? Probably not as the get_or_create ensures that if an old format test_type exists it will not create a new one?

Maffooch:
**Fixed, tested, and approving (milestone 3.1.100).**

**Issue found in review:** the idempotency rewrite dropped the legacy guard that returns the scan type as-is when `f"{type} Scan" == scan_type`. That re-introduced the exact doubling this PR is meant to remove — for every *dynamic* parser whose `scan_type` already ends in ` Scan` (the report `type` is the parser ID, and `scan_type` = ID + ` Scan`). Reproduced end-to-end on a running instance:

- On the PR branch (before fix): importing a **Horusec Scan** created Test Type `Horusec Scan (Horusec Scan)`; on `bugfix` the same import produced the clean `Horusec Scan`. Same doubling would hit **AWS Security Hub**, **Rusty Hog** variants, etc.

**Fix:** restored the guard in `resolve_dynamic_test_type_name` (`if f"{raw_type} Scan" == self.scan_type: return self.scan_type`) before the final ` Scan (...)` fallback, and added a regression test that imports a Horusec scan and asserts `Horusec Scan`.

**Verified after fix:**
- New `test_import_scan_suffixed_dynamic_type_is_not_doubled` — Horusec import → `Horusec Scan` ✅
- Existing `test_import_generic_type_with_suffix_is_not_doubled` — the original Generic doubling fix still holds (`Prisma Cloud (Generic Findings Import)` used verbatim, not re-appended) ✅
- `test_reimport_generic_type_with_suffix_is_idempotent` ✅  (3/3 passing)

The original goal (no doubled `(scan_type)` suffix for Generic dynamic imports) is preserved, and the regression for ` Scan`-suffixed parsers is closed. Looks good now.

Maffooch:
**Comprehensive all-scan-types verification (per request) — no regressions found.**

To be certain the guard fix doesn't affect any other tool, I proved the new resolver is behavior-identical to the pre-PR (legacy) naming for **every** registered scan type:

- **All 232 registered scan types × a matrix of report `type` values** (empty, equal-to-scan-type, base-without-` Scan`, arbitrary, and pre-suffixed): the new `resolve_dynamic_test_type_name` output matched the legacy behavior in **every** case — **0 unexpected differences**. The only divergences (464 of them) are exactly the intended doubling-fixes, where the report `type` already ends with ` ({scan_type})` and the new code uses it verbatim instead of re-appending ` Scan (...)`.
- **Full `test_importers_importer` suite: 44/44 passing.**
- **Live API imports** on a running instance, confirming un-doubled Test Type names:
  - `Horusec Scan` → `Horusec Scan` (was `Horusec Scan (Horusec Scan)` before the guard was restored)
  - `AWS Security Hub Scan` → `AWS Security Hub Scan`
  - `Choctaw Hog Scan` (Rusty Hog) → `Choctaw Hog Scan`
  - Generic dynamic types still resolve correctly and the original doubling fix is preserved.

Only *dynamic* test-type parsers (those implementing `get_tests`) invoke this method at all; static parsers use `scan_type` directly and are unaffected. Confident this is correct across the board.

Maffooch:
> Does auto create context need to be aware of old and new formats?

Yes, and that's handled, but on the reimport path rather than in auto-create-context itself. auto_create_context only builds the product/engagement/test hierarchy; the old-vs-new name awareness lives in consolidate_dynamic_tests. On reimport we now accept either name:

expected_test_type_name = self.resolve_dynamic_test_type_name(test_raw.type)
legacy_test_type_name   = self.legacy_dynamic_test_type_name(test_raw.type)
if self.test.test_type.name not in {expected_test_type_name, legacy_test_type_name}:
    # mismatch error

So a test created before this fix (with the old doubled name) still matches its report on reimport — no Test type mismatch, no parallel test, no migration needed.

One edge worth calling out explicitly: for a fresh import (not a reimport) where a customer previously had a doubled test type, get_or_create_test_type will create a new Test Type under the clean name and leave the old doubled one in place. That's harmless — existing tests keep reimporting fine via the legacy acceptance above — but it does mean the two can coexist. I think that's the right trade-off vs. a data migration; happy to reconsider if you'd rather I backfill/rename.

> Does deduplication need to be aware of this? Probably not as the get_or_create ensures that if an old format test_type exists it will not create a new one?

Correct on both counts. Dedup is unaffected: dedup configuration and hashing are keyed on scan_type/parser, not on the composed Test_Type name, and scan_type is unchanged by this PR. And as you said, get_or_create_test_type means an existing (old-format) test type is reused rather than duplicated. So there's no dedup-side change required.

valentijnscholten (approved):

Jino-T (approved):

PR #15151, opened 2026-07-03
Expose effective dedupe matching policy on the Test API (2/3)

devGregA:
> Part **2/3** of a visibility series: #15150 · #15151 (this PR) · #15152
> No merge-order constraint — this PR is independent of the other two.

**Description**

Exposes the effective finding-matching policy on the Test API as two read-only fields:

- **`deduplication_algorithm`** — `legacy`, `unique_id_from_tool`, `hash_code`, or `unique_id_from_tool_or_hash_code`
- **`hash_code_fields`** — the finding fields hashed for this test's scan type (`null` when the scan type has no per-scanner configuration and legacy default fields apply)

Both are existing model properties on `Test` (`dojo/test/models.py`) that already resolve `DEDUPLICATION_ALGORITHM_PER_PARSER` / `HASHCODE_FIELDS_PER_SCANNER` — the importer and dedupe machinery use them on every import. This PR adds **no new logic**; it only surfaces what the system already computes.

Why: "which fields are compared exactly?" is a recurring support question — deduplication and reimport matching behave differently per scanner, and today the answer lives in `settings.dist.py` where users can't see it. With this change the effective policy is one API call away (and available for UIs to render on the test page).

**Test results**

New test module `unittests/test_apiv2_test_dedupe_policy.py`:

- the fields are present on `GET /api/v2/tests/{id}/` and mirror the per-scanner settings (asserted dynamically against the settings, so the test doesn't rot when configs change)
- the fields are ignored on write (read-only), while normal PATCHes still work

Verified locally against PostgreSQL via the unit-test compose image: the new module plus `unittests.test_rest_framework.TestsTest` (21 tests) — all passing. Ruff (0.15.20, repo config) passes.

**Documentation**

No workflow changes — additive read-only API fields with self-documenting `help_text` (visible in the OpenAPI schema). Happy to add an API docs note if maintainers prefer.

---

### DIFF
--- a/dojo/test/api/serializer.py
+++ b/dojo/test/api/serializer.py
@@ -18,10 +18,15 @@ class TestSerializer(serializers.ModelSerializer):
     # HASHCODE_FIELDS_PER_SCANNER). Read-only: surfaced so users can see
     # which algorithm and fields deduplication and reimport matching use,
     # instead of having to ask support or read settings.dist.py.
-    deduplication_algorithm = serializers.ReadOnlyField(
+    # Typed fields (not ReadOnlyField) so drf-spectacular can emit an exact schema.
+    deduplication_algorithm = serializers.CharField(
+        read_only=True,
         help_text="Algorithm used to match findings for deduplication and reimport "
                   "(legacy, unique_id_from_tool, hash_code, or unique_id_from_tool_or_hash_code).")
-    hash_code_fields = serializers.ReadOnlyField(
+    hash_code_fields = serializers.ListField(
+        child=serializers.CharField(),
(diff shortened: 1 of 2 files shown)


Maffooch:

Validated against a running instance: `unittests.test_apiv2_test_dedupe_policy` passes (**2/2**) on a clean test DB. The new fields on the Test API serializer are read-only and expose the *effective* dedupe matching policy without changing existing response behavior. Small, focused, well-tested.

mtesauro (approved):
Approved

Maffooch (approved):

PR #15165, opened 2026-07-07
perf(importers): add defer_product_grading option + harden watson index-flush hook

dogboat:
## Summary

Two small, independent improvements for large imports/reimports, plus follow-up
hardening on the async watson index-flush hook. Both are backward compatible —
opt-in flag / no change to the default path.

### `defer_product_grading` importer option

New `ImporterOptions` flag (default `False`), honored by **both the importer and
reimporter**. When enabled, product-grade recalculation is skipped during the run —
per dedupe batch, at end-of-process, and after closing old findings — so the caller
runs a single `perform_product_grading()` pass afterward instead.

On a large import/reimport the default path recalculates the product grade once per
dedupe batch (and again at the end), which is redundant; deferring collapses that to
one pass. The default (`False`) preserves existing behavior, so no current caller is
affected. The validator documents that it's all-or-nothing: with the flag set, the
caller owns the final grade.

### Watson index-flush hook hardening

- **`suppress_intermediate_flush()`** — a thread-local context manager that turns the
  intermediate-flush hook into a no-op. `watson.skip_index_update()` only marks its
  context invalid on `__exit__`, so while a skip block is still accumulating, the flush
  hook can't tell it's a skip context and would drain it — dispatching async index
  tasks for objects meant to be discarded. Wrapping a `skip_index_update()` block in
  this makes the skip actually hold.
- **Singleton guard** — the hook now only intermediate-flushes the global singleton
  context. The async reindex task builds its own `SearchContextManager` and is itself
  the drain target; without the guard it would dispatch a clone of itself and loop.
  Ad-hoc contexts index their own batch on `end()` as before.

### Minor

- Guards a reimporter debug log behind `logger.isEnabledFor(DEBUG)`: it previously
  built a `(id, hash_code)` list over every original finding on each reimport,
  regardless of log level.

### Tests

- `test_importers_importer.py` — importer honors the flag (per-batch dispatch off +
  no end grade), with a control proving the default path still grades.
- `test_importers_closeold.py` — the close-old grade is deferred, for both importer
  and reimporter.
- `test_watson_intermediate_flush.py` — ad-hoc contexts don't drain;
  `suppress_intermediate_flush()` no-ops and restores.

[sc-13510]


valentijnscholten (changes requested):
Funny how Opus generated the exact same code/naming on your PR as my PR.

In https://github.com/DefectDojo-Inc/dojo-pro/pull/1778 I have chosen a more elegant option. That one also adds a Celery context manager to make the async index batching work in celery tasks i.e. the background importer in pro.

Can I suggest to remove the watson parts of your PR?

devGregA:
Refreshed to current `dev` (`0724720371`, 114 commits of drift) — resolutions: this branch's `product_grading_option=not self.defer_product_grading` carried into dev's new per-`push_to_jira` partitioned dispatch loops (importer + reimporter); kept BOTH independent middleware additions (this branch's `suppress_intermediate_flush()` kill-switch and dev's `watson_search_context_for_task()`), with the suppression check ported into dev's instance-level `add_to_context_with_flush` (the old class-level singleton guard is obsolete under dev's install).

Verified: `test_importers_importer` + `test_importers_closeold` + `test_watson_intermediate_flush` — **61/61 OK** on the merged branch; plus a full live E2E of the chunked-connector-sync pairing this PR supports (DefectDojo-Inc/dojo-pro#1769 — evidence table there): 12,345-finding chunked import exact, close-old exactly-once, abort/discard/legacy paths all green, celeryworker bounded ~1.0GiB.

dojo/middleware.py
```diff
@@ -317,6 +317,31 @@ def _drain_search_context_to_async(objects, source):
...
+    drain) and the skip actually holds. Thread-local, so it only affects the wrapping thread.
+    """
+    previous = getattr(_intermediate_flush_state, "suppressed", False)
+    _intermediate_flush_state.suppressed = True
+    try:
+        yield
+    finally:
+        _intermediate_flush_state.suppressed = previous
+
+
```
valentijnscholten, line 344: this is no longer needed as it was fixed in 
https://github.com/DefectDojo/django-DefectDojo/pull/15187
https://github.com/DefectDojo/django-DefectDojo/pull/15172

dojo/middleware.py
```diff
@@ -377,6 +402,13 @@ def install_intermediate_flush_hook():
     original_add = search_context_manager.add_to_context  # bound method
 
     def add_to_context_with_flush(engine, obj):
+        # Bulk ops under skip_index_update() (e.g. a large chunked reimport) set this to make the
+        # hook a full no-op on this thread. skip_index_update() only invalidates its context at
+        # __exit__, so mid-accumulation the hook would otherwise drain (index) objects meant to be
+        # discarded -- defeating the skip. Returning before original_add means nothing is even
+        # accumulated, so the skip holds and the caller's own end-of-op reindex does the indexing.
+        if getattr(_intermediate_flush_state, "suppressed", False):
+            return
```
valentijnscholten, line 411: this is no longer needed as it was fixed in 
https://github.com/DefectDojo/django-DefectDojo/pull/15187
https://github.com/DefectDojo/django-DefectDojo/pull/15172

unittests/test_watson_intermediate_flush.py
```diff
@@ -138,3 +139,27 @@ def test_ad_hoc_context_manager_does_not_drain(self):
...
+            with suppress_intermediate_flush():
+                for p in self.products[:3]:  # past threshold, but suppressed
+                    search_context_manager.add_to_context(engine_marker, p)
+                drain.assert_not_called()
+                self.assertEqual(len(objects), 0, "suppressed hook must not even accumulate")
+
+            # Flag restored after the block: the hook drains normally again at threshold.
+            for p in self.products[:2]:
+                search_context_manager.add_to_context(engine_marker, p)
+            drain.assert_called_once()
```
valentijnscholten, line 165: this is no longer needed as it was fixed in 
https://github.com/DefectDojo/django-DefectDojo/pull/15187
https://github.com/DefectDojo/django-DefectDojo/pull/15172

valentijnscholten:
>New ImporterOptions flag (default False), honored by both the importer and
reimporter. When enabled, product-grade recalculation is skipped during the run —
per dedupe batch, at end-of-process, and after closing old findings — so the caller
runs a single perform_product_grading() pass afterward instead.

I wonder if we should make this the default only behaviour? I left these intermediate grading calls in place to have the product grade update progressively during an import. But not sure this is important/useful now that imports are a lot faster.

PR #15167, opened 2026-07-07
docs: SSO user local-login fallback for open source

Maffooch:
**Shortcut:** [sc-13517](https://app.shortcut.com/defectdojo/story/13517)

## Problem

After migrating to open-source DefectDojo 3.x, SSO (SAML/OIDC/OAuth) is a Pro-only feature and the SSO login options are removed. Users that were provisioned through SSO can no longer log in: they were never given a local password, and the UI and API refuse to set one because DefectDojo detects any account with no usable password as an SSO account and blocks the change. Deleting and recreating those users loses their history, permissions, and object ownership.

## Change

Adds an Open Source documentation page, **Re-enabling login for SSO users (Open Source)**, under Admin → User Management. It explains why the lockout happens (`has_usable_password()` is the sole SSO signal in OSS) and documents the backend workaround:

- Set a local password with `user.set_password("<temp>")` from the Django shell (`docker compose exec uwsgi ./manage.py shell`), which flips `has_usable_password()` to `True` and lifts the SSO guards.
- Set `UserContactInfo.force_password_reset = True` so the middleware forces the user to choose their own password on next login.

Includes a single-user snippet, a Kubernetes variant, and the forgot-password alternative. Cross-links the new page from the existing "SSO Users" section of the creating-new-users page.

> Note: the correct call is `set_password(...)`, not `set_usable_password()` — the latter is not a Django model method (5.1 only added a `usable_password` form field; this repo runs Django 5.2.14).

## Verification

Built the docs with Hugo (`hugo --environment development`): builds cleanly with no errors or warnings. Confirmed the new page renders at `/admin/user_management/os__sso_user_local_login_fallback/`, the cross-link resolves both directions, and the linked 3.0 upgrade-notes anchor exists.

### DIFF
--- a/docs/content/admin/user_management/OS__creating_new_users.md
+++ b/docs/content/admin/user_management/OS__creating_new_users.md
@@ -36,6 +36,8 @@ The admin who creates the account is responsible for delivering the initial cred
 
 If your instance is configured with [SSO](../configure_sso/), the workflow is different — users are typically created on first login from the Identity Provider, and you only need to grant them group membership or roles afterwards.
 
+If you have moved to open-source DefectDojo (where SSO is Pro-only) and existing SSO users can no longer log in, see [Re-enabling login for SSO users](../os__sso_user_local_login_fallback/).
+
 ## Recovering from a lost MFA token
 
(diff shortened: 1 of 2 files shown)


paulOsinski (approved):

valentijnscholten (approved):
approved after changes :-)

PR #15172, opened 2026-07-07
fix(watson): fix + batch full-text index updates on Celery worker tasks

valentijnscholten:
## Description

Celery workers serve no HTTP request, so `AsyncSearchContextMiddleware` never opens a watson `search_context`. Without an active context, django-watson's `post_save` receiver falls into its else-branch and indexes every saved object **synchronously** — one `watson_searchentry` DELETE + INSERT per object. Any bulk save that runs in a worker (e.g. a large async import) then re-indexes objects one at a time, flooding the DB with thousands of index writes.

This PR refactors the performance test to perform a real API call. This results in queries and celery tasks triggered by middleware is now also captured. This explains the small increase in counts in this PR. But this actually a decrease because of the watson fix bundled in this PR.

### Approach

Phased, to confirm the problem before fixing it (regression test added downstream in DefectDojo Pro, where the async-import worker path lives — see the paired Pro PR):

1. **Reproduce** — a 1373-finding import through the worker path issued **~1000+** `watson_searchentry` INSERTs (query-count explosion). The failing test asserted the batched expectation.
2. **Fix** — batch the worker-side index updates the same way the request path already does.

### Changes

- **`watson_search_context_for_task` (`dojo/middleware.py`)** — a context manager that opens a watson `search_context` for the duration of a Celery task so saves accumulate and drain in `WATSON_ASYNC_INDEX_UPDATE_BATCH_SIZE`-sized async batches on exit, mirroring `AsyncSearchContextMiddleware` for the request path. An empty context (task saved no indexed models) drains to nothing — a cheap no-op.
- **`CELERY_TASK_CONTEXT_MANAGERS` default (`dojo/settings/settings.dist.py`)** — wires the CM in; `PluggableContextTask` enters it around every task. Downstreams (e.g. Pro) extend this list rather than replacing it.
- **Instance-bound intermediate-flush hook (`dojo/middleware.py`, `dojo/tasks.py`)** — `install_intermediate_flush_hook()` now binds the flush wrapper to the module-global `search_context_manager` **instance** instead of monkeypatching `SearchContextManager` at class level. The terminal index task `update_watson_search_index_for_model` builds its own local `SearchContextManager` to index one already-bounded batch; with the class-level patch, a full 1000-PK batch re-triggered the flush from inside the task and re-dispatched a clone of itself — infinite recursion under eager celery, an infinite re-dispatch loop (nothing ever indexed) on a real worker. Binding to the singleton makes that unrepresentable: watson's `post_save` and the request/task paths all go through the module-global instance (which flushes), while private throwaway contexts keep the stock `add_to_context` and simply index their batch on `end()`. The same loop is fixed on the release line by the minimal identity guard in #15187 (extracted from #15165); this PR supersedes the guard on `dev` with the structural fix.

## Test results

Verified in the integrated (Pro) environment; a downstream regression test drives the worker path (`AsyncImporter`) end-to-end:

- 1373-finding JFrog import: `watson_searchentry` INSERTs **~1000+ → 14** (`ceil(1000/100) + ceil(373/100)`), with **1373/1373** findings still indexed exactly once (no findings dropped, no recursion).
- `unittests/test_watson_intermediate_flush.py` passes (5/5), including the new `test_ad_hoc_context_manager_does_not_drain` regression test: an ad-hoc `SearchContextManager` pushed past the flush threshold must not drain (it would re-dispatch a clone of itself and loop). The existing tests exercise the global manager, which still flushes.
- Loop verified fixed live: dispatching a full 1000-pk `update_watson_search_index_for_model` task on unguarded code self-replicates without bound (10 → 200+ tasks in ~3 min); on this branch the same dispatch executes exactly once and indexes its batch.

OSS-standalone CI will run the full `unittests/` suite for final confirmation.

## Labels

- performance, bugfix, settings_changes (new `CELERY_TASK_CONTEXT_MANAGERS` in `settings.dist.py`)

### DIFF
--- a/dojo/middleware.py
+++ b/dojo/middleware.py
@@ -325,41 +325,41 @@ def watson_search_context_for_task():
 
 def install_intermediate_flush_hook():
     """
-    Wrap `watson.search.search_context_manager.add_to_context` with a
-    size-based flush. Once the per-request set reaches
+    Wrap `add_to_context` on the module-global `watson.search.search_context_manager`
+    with a size-based flush. Once the shared accumulation set reaches
     `WATSON_ASYNC_INDEX_UPDATE_BATCH_SIZE`, drain it into async tasks
     and clear it in place. Bounds memory on long-running requests
     (large imports) and starts celery batches earlier instead of
     dispatching all at end-of-request.
 
+    The wrapper is bound to the singleton INSTANCE, not the class: only the shared
+    request/task accumulation context batches. A throwaway local SearchContextManager
+    — e.g. the one update_watson_search_index_for_model builds to index one
+    already-bounded batch — keeps the stock method and indexes its own batch on
+    end(). If local contexts flushed too, that index task would re-dispatch itself
+    for the same batch (infinite recursion under eager celery / re-dispatch loop on
+    a worker). watson's post_save always adds to the module-global instance, so the
+    singleton is the only place the flush is needed.
+
     Idempotent — safe to call multiple times.
     Setting WATSON_ASYNC_INDEX_UPDATE_BATCH_SIZE to 0 or below disables
     the hook at runtime.
     """
-    cls = search_context_manager.__class__
-    if getattr(cls, "_dd_intermediate_flush_installed", False):
+    if getattr(search_context_manager, "_dd_intermediate_flush_installed", False):
         return
 
-    original_add = cls.add_to_context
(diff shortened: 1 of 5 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

PR #15173, opened 2026-07-07
refactor(locations): consistent object lookups in endpoint views

devGregA:
### Description

The endpoint (Location) views under the V3 Locations feature each resolved their `Location` with an ad-hoc `get_object_or_404` (and one `Location.objects.get`). This routes them all through a single small helper that fetches the `Location` from the standard location queryset, so object retrieval in these views is consistent with the list and host views that already use `get_authorized_locations`.

No new endpoints; a straightforward tidy-up of how the existing views look up their object.

### Test

Adds `unittests/test_location_cross_product_authz.py` covering the endpoint view lookups.

### DIFF
--- a/dojo/url/ui/views.py
+++ b/dojo/url/ui/views.py
@@ -43,6 +43,20 @@ from dojo.utils import (
 logger = logging.getLogger(__name__)
 
 
+def _get_location_or_404(request, location_id, permission):
+    """
+    Resolve a Location for the endpoint views via the shared authorized queryset.
+
+    Keeps object retrieval in these views consistent with the list/host views, which
+    already scope Location lookups through ``get_authorized_locations``. A lookup that
+    falls outside the queryset returns 404, matching object retrieval elsewhere.
+    """
+    return get_object_or_404(
+        get_authorized_locations(permission, Location.objects.all(), request.user),
+        id=location_id,
+    )
+
+
 def view_endpoint(request: HttpRequest, location_id: int):
     return process_endpoint_view(request, location_id, host_view=False)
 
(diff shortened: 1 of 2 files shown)


Maffooch:
**Fixed, tested, and approving (milestone 3.1.100).**

**Issue found in review:** the 3 cross-product tests failed whenever they actually run (`V3_FEATURE_LOCATIONS=true`) — reproduced locally and matching CI's failing `test-rest-framework (…, true)` leg (the `false` leg only passes because the suite is `@skip_unless_v3`). They asserted **404**, but the requests come back **400**.

**Root cause:** these endpoint views are already guarded by `AuthorizationMiddleware` (`URL_PERMISSIONS` maps them to `("object", Location, …)`), which raises `PermissionDenied` for a cross-product location *before* the view runs. DefectDojo renders `PermissionDenied` via `dojo.views.custom_unauthorized_view`, which returns **HTTP 400** app-wide. So the denial is real, just at 400 — not the 404 the view-level `get_authorized_locations` lookup would produce (it never gets reached).

**Fix:** corrected the three cross-product assertions to the actual denial status (400), with a comment documenting why. The view-level `_get_location_or_404` scoping stays as **defense-in-depth** behind the middleware.

**Verified under `V3_FEATURE_LOCATIONS=true`:**
- `test_location_cross_product_authz` — **4/4 passing**.
- Security property holds: cross-product **view/edit/delete → 400 denied**, and the object is **unchanged** (edit) / **still exists** (delete); own-product **view → 200**.

Note: the cross-product protection is enforced by the middleware, so this PR's view change is defense-in-depth rather than the sole gate. If information-hiding via a genuine **404** (instead of 400) is desired, that's a separate, app-wide change to how the middleware/`custom_unauthorized_view` handles object-scope denials. The PR's goal — no cross-product access to locations via the endpoint views — is met. Looks good now.

Maffooch (approved):

blakeaowens (approved):

PR #15185, opened 2026-07-08
perf(importers): fetch only needed columns in close_old_findings

valentijnscholten:
Found during analysis of OOMs during imports.

`close_old_findings` iterated `findings.values()` — every column of every finding in the test, including `description` — just to collect `is_mitigated`, `hash_code` and `unique_id_from_tool` for the exclusion filter. On large imports that materializes full-row dicts for thousands of findings for no reason.

Narrow the `values()` call to the three columns the loop reads. No behavior change; covered by a regression test that pins both the requested columns and the closed-findings outcome.


Maffooch (approved):

mtesauro (approved):
Approved

PR #15186, opened 2026-07-08
fix(importers): dispatch post-processing with per-finding push_to_jira

valentijnscholten:
Found during analysis of OOMs during imports.

The importers compute `push_to_jira` **per finding** (grouped findings must not be pushed individually — they are pushed as a group after the loop), but the batched post-processing dispatch fires only at batch boundaries and reads whatever the **last** loop iteration left in the variable. One finding's grouping state was therefore applied to the entire batch:

- batch ends in a **grouped** finding → whole batch dispatched with `push_to_jira=False`, so its ungrouped findings never reach JIRA;
- batch ends in an **ungrouped** finding → whole batch dispatched with `push_to_jira=True`, including its grouped findings.

Only visible with finding groups enabled *and* `push_to_jira` on; with grouping disabled or push off the flag is constant across the batch.

Fix: track each finding's own flag (`(finding_id, push_to_jira)`) and partition the batch by flag at dispatch time. Uniform batches still dispatch once, so task counts are unchanged for the common case. Applies to both `DefaultImporter` and `DefaultReImporter`.

Regression tests cover both leak directions (last-finding-grouped and last-finding-ungrouped).

### DIFF
--- a/dojo/importers/default_importer.py
+++ b/dojo/importers/default_importer.py
@@ -182,8 +182,12 @@ class DefaultImporter(BaseImporter, DefaultImporterOptions):
         parsed_findings: list[Finding],
         **kwargs: dict,
     ) -> list[Finding]:
-        # Batched post-processing (no chord): dispatch a task per 1000 findings or on final finding
-        batch_finding_ids: list[int] = []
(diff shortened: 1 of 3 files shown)


Maffooch (approved):

mtesauro (approved):
Approved

PR #15170, opened 2026-07-07
Product updates and tests

dogboat:
This PR updates and enhances product capabilities.

[sc-13305]

PR #15194, opened 2026-07-08
Guard authorized_users on the product/product type API serializers

devGregA:
### Description

Changing who is authorized on a product or product type goes through the `*_Manage_Members` permission in the server-rendered membership views. This applies the same permission check when `authorized_users` is written through the API serializers, so membership changes require the same permission regardless of entry point.

Covered serializers:
- `ProductSerializer` (`/api/v2/products/`)
- `ProductTypeSerializer` (`/api/v2/product_types/`)
- `AssetSerializer` (`/api/v2/assets/`, V3)
- `OrganizationSerializer` (`/api/v2/organizations/`, V3)

A shared helper (`raise_if_unauthorized_authorized_users_change`) performs the check so behavior is identical across all four. Users with the manage-members permission (and superusers) are unaffected, and other field edits are unaffected.

### Test results

Adds `unittests/test_authorized_users_api_authz.py` covering the `authorized_users` write paths for products, product types, and the V3 asset/organization aliases.

### DIFF
--- a/dojo/asset/api/serializers.py
+++ b/dojo/asset/api/serializers.py
@@ -1,6 +1,8 @@
 from rest_framework import serializers
 
 from dojo.api_v2.serializers import ProductMetaSerializer, TagListSerializerField
+from dojo.authorization.authorization import raise_if_unauthorized_authorized_users_change
+from dojo.authorization.roles_permissions import Permissions
 from dojo.models import (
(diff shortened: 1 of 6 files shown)


devGregA:
Closing in favor of #15170, which is the tracked change for this area. Handing off additional notes separately.

Issue #14923, opened 2026-05-27
pass username strings instead of Dojo_User objects to create_notification in process_tag_notifications


donhluca:
## Description

Fix `user_mentioned` notifications silently failing when a user is 
@mentioned in a Finding or Test note.

## Root Cause

`process_tag_notifications()` passed `Dojo_User` objects as `recipients` 
to `create_notification()`, but `_process_recipients()` filters using 
`user__username__in=kwargs["recipients"]` which expects username strings. 
The type mismatch caused the queryset to return zero results → zero 
notifications delivered.

## Change

One-line fix in `process_tag_notifications()`: return username strings 
from the list comprehension instead of `Dojo_User` objects.

## Testing

Verified manually via Django shell:
- Before fix: `create_notification(recipients=[Dojo_User object])` → 0 alerts
- After fix: `create_notification(recipients=["username"])` → 1 alert created

Tested on DefectDojo 2.58.4, Docker Compose deployment.

PR #15196, opened 2026-07-09
fix(notifications): deliver @mention notifications and match full usernames

blakeaowens:
## Description

Notifying a user with `@username` in a note has two bugs in
`process_tag_notifications` (`dojo/notifications/helper.py`):

1. **Notifications were never delivered.** The function collects `Dojo_User`
   objects and passes them as `recipients` to `create_notification`. But
   `NotificationManager._process_recipients` resolves recipients with
   `Notifications.objects.filter(user__username__in=recipients)`. `username`
   is a `CharField`, so each `Dojo_User` is coerced to `str(user)` =
   `"First Last (username)"`, which never equals a stored username — zero
   recipients match, and no alert/email/Slack message is sent. This fix passes
   `user.username` strings, which is what the parameter expects (see the
   existing `recipients=["admin"]` cases in `unittests/test_notifications.py`).

2. **Usernames with `. - @ +` couldn't be fully mentioned.** Django's
   username validators allow those characters, but the mention regex only
   captured `\w+`, so `@jane.doe` resolved to `jane`. The capture is broadened
   to `[\w.@+-]+`; the leading `(?:\A|\s)@` anchor still prevents email
   addresses written in prose (e.g. `contact a@b.com`) from triggering a
   mention, and a trailing sentence period is stripped so `thanks @jane.`
   still resolves to `jane`.

## Behavior change

- `@mention` notifications now actually reach the mentioned user (previously a
  silent no-op).
- `@jane.doe`-style (and SSO/email-style) usernames now resolve to the full
  username instead of a truncated prefix — which also avoids accidentally
  notifying a different user who happens to own that prefix.

## Testing

- `unittests/test_notifications.py` passes (its `user_mentioned` cases already
  use `recipients=["admin"]`, matching the corrected caller).
- Added a regression test that a note containing `@<active-user>` creates an
  alert for that user (previously none was created).

### DIFF
--- a/dojo/notifications/helper.py
+++ b/dojo/notifications/helper.py
@@ -889,9 +889,14 @@ class NotificationManager(NotificationManagerHelpers):
 
 
 def process_tag_notifications(request, note, parent_url, parent_title):
-    regex = re.compile(r"(?:\A|\s)@(\w+)\b")
+    # Django usernames may contain "@ . + - _" in addition to word characters (see
+    # Django's username validators), so capture that whole set instead of only \w.
+    # The leading (?:\A|\s)@ anchor keeps email addresses written in prose from
+    # triggering, and trailing dots are stripped so a mention that ends a sentence
+    # ("thanks @alice.") still resolves to the username.
+    regex = re.compile(r"(?:\A|\s)@([\w.@+-]+)")
 
-    usernames_to_check = set(un.lower() for un in regex.findall(note.entry))  # noqa: C401
+    usernames_to_check = {un.lower().rstrip(".") for un in regex.findall(note.entry)}
 
     users_to_notify = [
         Dojo_User.objects.filter(username=username).get()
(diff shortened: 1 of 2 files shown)


Maffooch (changes requested):
Can you please add some regression tests for this to lock in behavior

valentijnscholten (commented):
Does this superseed https://github.com/DefectDojo/django-DefectDojo/pull/14927?

blakeaowens:
> Does this superseed #14927?

@valentijnscholten Yes, this supersedes #14927. Both fix the same root cause (#14923): `process_tag_notifications()` handed `Dojo_User` objects to `create_notification(recipients=…)`, but `_process_recipients()` filters on `user__username__in`, so the objects matched nothing and no notification went out.

On top of that delivery fix, this PR also broadens the mention regex from `@(\w+)` to `@([\w.@+-]+)` (plus a trailing-dot strip) so usernames containing `.`, `-`, `@`, or `+` resolve to the full username instead of a truncated prefix.

Maffooch (approved):
Thanks for the tests!

Jino-T (approved):

PR #15204, opened 2026-07-09
fix(auditlog): make pghistory context JSON-safe before Celery dispatch

Maffooch:
[sc-13581]

## Summary

When a Celery task is dispatched during a request, `dojo_dispatch_task` injects the active pghistory context into the task kwargs as `_pgh_context` via `get_serializable_pghistory_context()`. That function returned `ctx.metadata.copy()` verbatim, so a non-JSON-serializable value in the context metadata (e.g. a `user` set to a `User` object instead of its pk) made kombu raise `EncodeError: Object of type User is not JSON serializable` and abort the entire task dispatch.

Reported in production on `PATCH /api/v2/products/{pk}/` (the product `post_save` signal dispatches a task), 700+ events (Sentry DJANGO-42W8).

## Fix

Coerce the captured context metadata to JSON-safe values: reduce Django model instances to their pk and recurse into dict/list/tuple. Defensive and behavior-preserving for already-serializable metadata (the common case — the standard `HistoryMiddleware` stores the user as a pk).

## Tests

`unittests/test_pghistory_context_serialization.py`:
- model instance in metadata is coerced to pk
- the captured context is kombu-JSON-serializable (the exact dispatch operation that raised before)
- nested model instances (list/dict) are coerced
- already-serializable metadata is unchanged (control)

Verified fail-before / pass-after. Run:
```
SECURE_SSL_REDIRECT=False python manage.py test unittests.test_pghistory_context_serialization --keepdb
```

Fixes DJANGO-42W8

### DIFF
--- a/dojo/auditlog/helpers.py
+++ b/dojo/auditlog/helpers.py
@@ -166,18 +166,43 @@ def process_events_for_display(events):
 # ---------------------------------------------------------------------------
 
 
+def _json_safe_metadata_value(value):
+    """
+    Coerce a pghistory context metadata value into something the JSON task
+    serializer can encode.
+
+    The context is injected into Celery task kwargs as ``_pgh_context`` and
+    serialized by kombu's JSON encoder at dispatch time. A stray model instance
+    in the metadata (e.g. a ``user`` set to a ``User`` object instead of its pk)
+    raises ``EncodeError: Object of type User is not JSON serializable`` and
+    aborts the whole task dispatch. Reduce model instances to their pk and
+    recurse into dict/list/tuple so nested values are covered too.
+    """
+    from django.db.models import Model  # noqa: PLC0415 -- avoid import cycle at app load
+
+    if isinstance(value, Model):
+        return value.pk
+    if isinstance(value, dict):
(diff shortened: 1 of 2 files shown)


blakeaowens (approved):

Jino-T (approved):

dojo/auditlog/helpers.py
```diff
@@ -166,18 +166,43 @@ def process_events_for_display(events):
...
+    The context is injected into Celery task kwargs as ``_pgh_context`` and
+    serialized by kombu's JSON encoder at dispatch time. A stray model instance
+    in the metadata (e.g. a ``user`` set to a ``User`` object instead of its pk)
+    raises ``EncodeError: Object of type User is not JSON serializable`` and
+    aborts the whole task dispatch. Reduce model instances to their pk and
+    recurse into dict/list/tuple so nested values are covered too.
+    """
+    from django.db.models import Model  # noqa: PLC0415 -- avoid import cycle at app load
+
+    if isinstance(value, Model):
```
valentijnscholten, line 183: Do I read this right that it "only works" for Model instances and not for other objects (that might also be not json serializable?

valentijnscholten (commented):
e

PR #15212, opened 2026-07-10
fix(generic parser): prevent nested list when cve and vulnerability_ids are both present

narvadanami:
## Bug

When both `cve` and `vulnerability_ids` were present in generic JSON import, `.append()` was used instead of `.extend()`, causing a nested list:

`['CVE-2020-36234', ['GHSA-5mrr-rgp6-x4gr', 'OSV-2021-1234']]` instead of `['CVE-2020-36234', 'GHSA-5mrr-rgp6-x4gr', 'OSV-2021-1234']`.

Without `cve`, `vulnerability_ids` worked correctly via direct assignment. The bug only manifested when both fields were present simultaneously.

## Fix

- Changed `.append()` to `.extend()` when merging `cve` into existing `vulnerability_ids`
- Added `isinstance(str)` protection for string-type `vulnerability_ids` (prevents `.extend()` from iterating over characters)
- Updated documentation with `vulnerability_ids` example

## Testing

- Added `test_parse_json_with_cve_and_vulnerability_ids` covering:
  - `cve` + `vulnerability_ids` (list) → flat list of 3 items
  - `vulnerability_ids` as string → wrapped in list
- All 42 generic parser tests pass


narvadanami:
@Maffooch @mtesauro hi! Could you please take a look? All 149 tests passed successfully, and the fix resolves a nested list bug in the generic parser. Thanks!

valentijnscholten (approved):

Maffooch (approved):

PR #15111, opened 2026-06-30
docs: DefectDojo Pro Customizable Dashboards (dashboard_v2)

skywalke34:
## Description

Adds documentation for the new DefectDojo Pro **Customizable Dashboards** (`dashboard_v2`) — the per-user, widget-and-layout dashboard system — and reworks the Dashboards section to cleanly separate Open Source from Pro.

### What's new
- **Customizable Dashboards** (`PRO__custom_dashboards.md`) — concepts (layouts, widgets, the widget catalog, sharing) plus a full UI walkthrough with screenshots.
- **Automating Dashboards with the API** (`PRO__custom_dashboards_api.md`) — discover the widget catalog, create and update layouts, set a default, share with a team, and render widget data over the REST API, with a complete lifecycle script.
- **Building Dashboards with an LLM** (`PRO__custom_dashboards_llm.md`) — a self-contained prompt pattern for having an LLM design, create, and verify dashboards via the API.
- Reworked the Dashboards section landing (`_index.md`) into an Open Source vs. Pro comparison plus a "where to go next" guide.

### Cleanup
- Removed the obsolete `about_custom_dashboard_tiles.md` (the old singular Pro "tiles" dashboard, superseded by `dashboard_v2`).
- Trimmed `Introduction_dashboard.md` to focus on the OSS Main Dashboard.
- Pointed the top-nav "Metrics & Reports" link back at the OSS Main Dashboard page so it lands on a page with the full left side menu, consistent with every other top-nav section.
- Minor link/reference touch-ups in a few neighboring pages.

### Notes
- Customizable Dashboards are a **DefectDojo Pro (beta)** feature; the new pages are marked `audience: pro` and carry a beta-enablement note.
- Docs-only change. Previewed locally with `hugo server`.

PR #15008, opened 2026-06-14
docs: add DefectDojo Pro Report Builder guides (UI, API, LLM)

skywalke34:
Adds DefectDojo Pro Report Builder documentation, replacing the outdated report builder pages under `metrics_reports/reports/`:

- **Report Builder** — UI walkthrough (Themes, Blocks, Templates, Generated Reports) with screenshots
- **Automating Reports with the API** — create → run → poll → download, with a stdlib-only Python script and a verified filter/field reference
- **Building Reports with an LLM** — a paste-in prompt that designs, creates, runs, and downloads reports
- **Using the Report Builder (Open Source)** — build / run / retrieve; no saved templates or report history

New pages are gated to the **DefectDojo Pro** docs version via the `audience` flag; the open-source page stays on the Open Source version. Inbound links to the removed pages were repointed, and the Metrics & Reports prev/next ordering was corrected. Content and the API/field references were verified against the `dojo-pro` reporting code and a live tenant; screenshots were captured from the Report Builder UI.

PR #15216, opened 2026-07-10
docs: add copy-to-clipboard to Report Builder LLM prompt and API blocks

skywalke34:
## Description

The Report Builder LLM guide (added in #15008) presents a large, self-contained prompt that users are meant to copy in one shot and paste into Claude/ChatGPT — but that block had no copy-to-clipboard button, unlike the equivalent block in the Dashboards LLM guide (#15111).

**Root cause:** the docs theme (Doks / `@thulite/doks-core`) adds its copy button via `clipboard.js`, which targets every element with class `highlight`. Hugo only wraps a fenced code block in `<div class="highlight">` when the fence declares a language. Two Report Builder blocks used a bare ` ``` ` fence, so they rendered as plain `<pre>` with no copy button.

## Change

Tag the two bare fences as ` ```text ` (matching the working Dashboards LLM guide):

- `PRO__report_builder_llm.md` — the self-contained LLM prompt
- `PRO__report_builder_api.md` — the API base-URL example

Every other fence in the Report Builder docs already declared `bash`/`json`/`python`/`shell`, so they already had copy buttons. After this change, **all** API and LLM code/prompt blocks in the Report Builder docs are copy-to-clipboard capable.

## Verification

Built the site locally with Hugo (`hugo --gc`) and confirmed:

- `report-builder-llm/` went from 1 → **2** `.highlight` blocks; the prompt now renders as `<code class="language-text">` inside `<div class="highlight">`.
- `report-builder-api/` shows all fences wrapped in `.highlight`.

Docs-only change; no code paths affected.

Generated with Claude Code under the direction of T. Walker (skywalke34) - DefectDojo

### DIFF
--- a/docs/content/metrics_reports/reports/PRO__report_builder_api.md
+++ b/docs/content/metrics_reports/reports/PRO__report_builder_api.md
@@ -22,7 +22,7 @@ export DD_IMPORTER_DOJO_API_TOKEN="YOUR_API_TOKEN"
 
 The base URL for all calls is your instance plus `/api/v2`:
 
-```
+```text
 https://[YOUR-INSTANCE].cloud.defectdojo.com/api/v2
 ```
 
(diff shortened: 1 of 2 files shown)


Maffooch (approved):

paulOsinski:
Didn't even know you could do that.  Nice work!

paulOsinski (approved):

PR #15219, opened 2026-07-10
Fix Snyk Issue API reachability rendering; pin parser output surfaces

devGregA:
**Description**

Fixes the Snyk Issue API parser's reachability rendering. The parser collapsed Snyk's per-coordinate `reachability` values into a single boolean via `reachability != "not-applicable"`, so `no-path-found` — Snyk affirmatively reporting that **no** path to the vulnerable code exists — and even a completely missing value both rendered as `Reachable: Yes` in the impact field.

The impact text now renders the raw values Snyk reported (`Reachability: reachable` / `package` / `no-path-found` / `not-applicable`, deduped and sorted for determinism), and renders nothing when Snyk reports nothing.

This PR also adds output-stability assertions for existing parser output surfaces that users and downstream tooling rely on, so future changes to these formats are deliberate rather than incidental:

- govulncheck (V2): the `**Reachability:** <level> (<severity> severity)` description line and the `**Traces:**` block
- sysdig_reports: the `InUse:<bool>` tag format

**Test results**

- New fixture `unittests/scans/snyk_issue_api/snyk_sca_scan_api_reachability.json` exercises all four Snyk cases (`reachable`, `package`, `no-path-found`, and a coordinate with no reachability value) with assertions on the rendered impact text — including that the previously-misrendered `no-path-found` and missing-value cases no longer claim reachability.
- Updated the existing assertion that pinned the old misleading `Reachable: No` text.
- Added the stability assertions to the existing govulncheck V2 and sysdig 2025-format tests using existing fixtures.
- All touched files are Ruff-compliant; relying on CI for the full suite run.

**Documentation**

No documentation changes needed — output rendering fix and test hardening only.

### DIFF
--- a/dojo/tools/snyk_issue_api/parser.py
+++ b/dojo/tools/snyk_issue_api/parser.py
@@ -79,13 +79,14 @@ class SnykIssueApiParser:
         line = None  # Always None for SCA
         component_name = None
         component_version = None
-        reachable = False   # SCA only
+        reachability_values = []   # SCA only, raw values as reported by Snyk
         impact_locations = []
 
         for coordinate in coordinates:
             if not is_type_code:
-                if coordinate.get("reachability") != "not-applicable":
-                    reachable = True
+                reachability = coordinate.get("reachability")
+                if reachability:
+                    reachability_values.append(reachability)
 
             for representation in coordinate.get("representations", []):
                 if not is_type_code:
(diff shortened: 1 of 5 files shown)


devGregA:
Superseded by #15523, which carries this same parser fix on `bugfix`.

This branch was cut from `dev` on 2026-07-06 and had since accumulated ~19 unrelated `dev` commits, so it could not simply be retargeted at the patch line without dragging them along. #15523 is the same change cherry-picked onto current `bugfix`, plus a docstring fix for the repo's ruff config (which this branch would have failed).

Targeting `bugfix` because it is a correctness fix to an existing rendered field: Snyk's `no-path-found` currently renders as `Reachable: Yes`, the exact inverse of what Snyk reported.

PR #15234, opened 2026-07-13
fix(shell): silence the shell model auto-import banner (docker + manual runs)

Maffooch:
## What

Since Django 5.1, `manage.py shell` auto-imports every model on startup and prints a banner on **every** invocation:

```
36 objects could not be automatically imported:

  dojo.auditlog.services.ObjectsProductTags
  dojo.location.models.Tagulous_Location_tags
  pro.auditlog.RoleProxy
  ...

237 objects imported automatically (use -v 2 for details).
```

The "could not be imported" entries are dynamically-generated Tagulous tag models (`Tagulous_*_tags`) and auditlog proxy models — they exist on the app registry but aren't importable by their `"<module>.<name>"` path, so Django's auto-import can't resolve them. Nothing is actually wrong, but the list reads as an error.

This showed up in two places, both covered here:

### 1. Container startup scripts
`docker/reach_database.sh` (DB connectivity check) and `docker/entrypoint-first-boot.sh` (superuser creation) pipe short scripts into `manage.py shell`. They import exactly what they need, so they pass `--no-imports` (Django ≥ 5.2; the project pins 5.2.14) to skip auto-import entirely. The piped code still runs and exit codes are unchanged.

### 2. Manual `manage.py shell` runs
Developers running `shell` by hand won't pass `--no-imports`, so a `dojo/management/commands/shell.py` override filters `get_auto_imports()` down to paths that actually import. Real models still auto-import; the banner is clean. The filter is generic (it tries the import and keeps what succeeds), so it also covers Pro's proxy models without Pro needing its own override.

## Testing

On Django 5.2.14:
- `reach_database.sh`'s connection test with `--no-imports`: banner gone, `echo $?` = `0`.
- Manual `manage.py shell`: prints only `237 objects imported automatically` — no "could not be imported" list; `Finding` and other real models still auto-import.
- `unittests/test_shell_command.py` (new): asserts every path returned by `get_auto_imports()` imports, that real models are kept, and that every dropped path is genuinely non-importable. Green.
- `ruff --config ruff.toml` clean on the new files.

### DIFF
--- a/docker/entrypoint-first-boot.sh
+++ b/docker/entrypoint-first-boot.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 # called from entrypoint-initializer.sh when no admin user exists (first boot)
-cat <<EOD | python manage.py shell
+cat <<EOD | python manage.py shell --no-imports
 import os
 from django.contrib.auth.models import User
 User.objects.create_superuser(
--- a/docker/reach_database.sh
+++ b/docker/reach_database.sh
@@ -13,7 +13,7 @@ wait_for_database_to_be_reachable() {
             exit 1
         fi
     done
-    cat <<EOD | python manage.py shell
+    cat <<EOD | python manage.py shell --no-imports
 from django.db import connections
 connections['default'].cursor()
 EOD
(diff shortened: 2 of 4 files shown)


mtesauro (dismissed):
Not sure what's going on with those tests

Jino-T (approved):

blakeaowens (approved):

PR #15236, opened 2026-07-13
feat(search): add DD_WATSON_SEARCH_ENABLED toggle to gate watson indexing

blakeaowens:
## Add `DD_WATSON_SEARCH_ENABLED` toggle to gate django-watson

django-watson registers post-save/pre-delete search indexers on 9 models
(Finding is on the import hot path), and its only reader is the legacy
`/simple_search` page. There is currently no way to turn it off — only the
`DD_WATSON_ASYNC_INDEX_UPDATE_BATCH_SIZE` / `DD_WATSON_INDEX_PREFETCH_ENABLED`
tuning knobs.

This adds a master switch, `DD_WATSON_SEARCH_ENABLED` (default **True**, so OSS
behaviour is unchanged). When set to `False`:

- watson models are **not registered** (no post-save index writes),
- the `watson` app and `AsyncSearchContextMiddleware` are stripped from settings,
- `/simple_search` returns **410 Gone** instead of querying unregistered models,
- `installwatson` is skipped in first-boot (both the `complete_initialization`
  command and `entrypoint-first-boot.sh`).

This lets deployments that have replaced watson stop paying the indexing cost and reclaim the
`watson_searchentry` table, without affecting standard OSS installs.

### Changes
- `dojo/settings/settings.dist.py` — `DD_WATSON_SEARCH_ENABLED` env + settings gate
- `dojo/apps.py` — registration moved to `register_watson_models()`, called only when enabled
- `dojo/search/views.py` — 410 when disabled
- `dojo/management/commands/complete_initialization.py` + `docker/entrypoint-first-boot.sh` — gate `installwatson`
- `unittests/test_watson_search_disabled.py` — new test (410 disabled / 200 enabled)

### DIFF
--- a/docker/entrypoint-first-boot.sh
+++ b/docker/entrypoint-first-boot.sh
@@ -27,8 +27,13 @@ EOD
     python3 manage.py loaddata "${i%.*}"
   done
 
-  echo "Installing watson search index"
-  python3 manage.py installwatson
+  watson_enabled=$(echo "${DD_WATSON_SEARCH_ENABLED:-True}" | tr '[:upper:]' '[:lower:]')
+  if [ "$watson_enabled" != "false" ] && [ "$watson_enabled" != "0" ] && [ "$watson_enabled" != "off" ] && [ "$watson_enabled" != "no" ]; then
+    echo "Installing watson search index"
+    python3 manage.py installwatson
+  else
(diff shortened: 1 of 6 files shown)


blakeaowens:
[sc-13667] Tracked in Shortcut: https://app.shortcut.com/defectdojo/story/13667

Maffooch (approved):

Jino-T (approved):

PR #15296, opened 2026-07-20
Restrict configuration permission assignment to superusers in the user API

svader0:
Consistency and hardening change to the user API serializer. Assigning or changing a user's configuration permissions is now limited to superusers, matching how the same serializer already treats the other privilege-bearing user fields (staff and superuser status). Includes a regression test. No functional change for superusers, who manage these permissions today.


blakeaowens (approved):

devGregA (approved):
Reviewed the change and the added regression tests — looks good, and CI is green. Thanks Sam! ✅

valentijnscholten:
Review by GPT5.6 Sol passed.

Issue #15124, opened 2026-07-01
Bug: V3 endpoint migration is extremely slow and throughput degrades over time


DarkR0ast:
## Description

Running the V3 endpoint-to-location migration is significantly slower than expected. In addition to the low initial throughput, migration performance steadily degrades as more endpoints are processed.

For a dataset of approximately **65k endpoints**, the migration starts at around **3.4 endpoints/sec** and gradually drops to **0.3 endpoints/sec**, increasing the estimated completion time from roughly **5 hours** to **38+ hours**.

## Steps to Reproduce

1. Enable `V3_FEATURE_LOCATIONS`.
2. Run the endpoint migration (`migrate_endpoints_to_locations`).
3. Monitor the migration progress and throughput.

## Expected Behavior

Migration throughput should remain relatively stable throughout execution, with only minor fluctuations.

## Actual Behavior

Migration speed continuously decreases as the migration progresses.

Example output:

```text
Migrated 3,900/65,194 endpoints (6.0%) — 3.4 endpoints/sec — ETA 5h 4m
...
Migrated 4,350/65,194 endpoints (6.7%) — 2.5 endpoints/sec — ETA 6h 53m
...
Migrated 23,300/65,194 endpoints (35.7%) — 0.3 endpoints/sec — ETA 38h 29m
```

The slowdown appears cumulative rather than transient, suggesting the migration becomes less efficient as more records are processed.

## Additional Observations

During the migration, individual endpoint failures are logged and skipped, for example:

```text
ValidationError: {'protocol': ['icmp is not a supported protocol']}

ERROR [dojo.management.commands.migrate_endpoints_to_locations:475]
Failed to migrate endpoint id=104369; continuing
```

These validation failures are handled correctly and the migration continues, but they do not appear to explain the overall degradation in throughput.

## Impact

For larger DefectDojo instances, the migration may require tens of hours or longer to complete, making adoption of V3 Locations difficult in production environments.

## Questions

* Is this level of performance expected for the current implementation?
* Are there known bottlenecks in the migration (e.g. repeated queries, missing bulk operations, growing in-memory state, transaction handling, or index updates) that could explain the progressive slowdown?

<img width="666" height="252" alt="Image" src="https://github.com/user-attachments/assets/4fb7bbd3-f3a0-4d0f-bc89-db02bc18f2dd" />

[...]

PR #15311, opened 2026-07-22
perf(migration): batch endpoint tags and restore inheritance

AKSHATSPAR:

This is a bug fix for #15124, not a new enhancement.

**Description**

The reporter's completed migration, whose progress output initially listed 65,194 Endpoints, attributed 115,317.36 seconds (99.8% of its measured phase time) to tags and required several restarts. The command copies tags separately for each tagged Endpoint; in a controlled rerun of the exact PR base, those writes also increased Tagulous counters without creating new relationships.

This PR:

- batches Endpoint tag writes and deduplicates Locations that normalize to the same URL;
- makes each batch atomic, falling back to per-Endpoint retries on failure; and
- replaces the stale `LocationManager._bulk_inherit_tags` call removed by #14877, restoring first-pass Product-tag inheritance when enabled.

**Tests**

Added focused coverage for full and partial batches, duplicate Locations, clean reruns, inherited tags, atomic rollback, and per-Endpoint recovery.

Local validation: 3 migration tests passed, 29 bulk-tag tests passed, and 39 inheritance tests completed (37 passed, 2 skipped). Django system and migration checks and Ruff also passed.

**Controlled comparison**

I compared the exact PR base (`22e2519`) and patch in two fresh databases inside the same PostgreSQL 18.4 container. The workload contained 1,000 Endpoints, five direct tags per Endpoint (14 unique direct tags), and one inherited Product tag. Both runs used `--batch-size 100 --progress-every 50 --benchmark --query-count`.

| Metric | Exact base | Patched |
| --- | ---: | ---: |
| First-pass main-loop time | 9.68 s | 4.33 s |
| First-pass tag phase | 5.34 s | 0.20 s |
| First-pass queries / Endpoint (50-Endpoint chunks) | 41.0-42.5 | 19.1-20.9 |
| Rerun main-loop time | 8.75 s | 1.15 s |
| Rerun tag phase | 7.95 s | 0.23 s |
| Rerun queries / Endpoint (50-Endpoint chunks) | 42.0-42.3 | 7.1 |

The patched runs had no tag or inheritance mismatches. After the rerun, all 6,000 `Location.tags` relationships, 1,000 `Location.inherited_tags` relationships, and all 15 counters remained correct.

This controlled workload is not a reproduction of the reporter's deployment and does not establish the same improvement for every tag distribution or at the full 65,194-Endpoint scale. The patch prevents new counter drift during this command; it does not repair counters that were already incorrect.

**Documentation**

No documentation change is needed because the command interface and workflow are unchanged.

**Checklist**

- [x] Targets the `bugfix` branch.
- [x] Uses a release-note-ready title.
- [x] Adds focused unit coverage.
- [x] Does not change models or require a migration.
- [ ] Maintainer label: `performance` / `bugfix`.

### DIFF
--- a/dojo/management/commands/migrate_endpoints_to_locations.py
+++ b/dojo/management/commands/migrate_endpoints_to_locations.py
@@ -5,13 +5,14 @@ import time
 from collections import defaultdict
 
 from django.core.management.base import BaseCommand
-from django.db import connection
+from django.db import connection, transaction
 from django.db.models import Prefetch
 from django.utils import timezone
 
 from dojo.location.models import Location, LocationFindingReference, LocationProductReference
 from dojo.location.status import FindingLocationStatus, ProductLocationStatus
 from dojo.models import DojoMeta, Endpoint, Endpoint_Status, Product
+from dojo.tags.utils import bulk_add_tag_mapping
 from dojo.url.models import URL
 
 logger = logging.getLogger(__name__)
@@ -45,7 +46,7 @@ def _suspend_auto_now_add(model, field_name: str):
 PHASES = (
     "fetch_endpoint",   # iterator yields the next endpoint
     "url_create",       # URL.get_or_create_from_values + Location side-effect
-    "tags",             # endpoint tag copy onto the location
+    "tags",             # batched endpoint tag copy onto locations
     "meta",             # DojoMeta copy onto the location
     "finding_refs",     # LocationFindingReference creation per Endpoint_Status
     "product_refs",     # LocationProductReference creation
@@ -108,11 +109,11 @@ class Command(BaseCommand):
         The migration creates locations that may be linked to multiple products
         (via the endpoint's own product and via each finding's product). We
         collect every contributing product per location so the post-pass can
-        call ``LocationManager(product)._bulk_inherit_tags(locations)`` once
-        per product group — covering the case where a location is shared
-        across products with differing ``enable_product_tag_inheritance``
-        flags (the helper short-circuits via its own diff check on repeat
-        visits, so redundancy is safe).
+        call ``apply_inherited_tags_for_locations`` once per product group —
+        covering the case where a location is shared across products with
+        differing ``enable_product_tag_inheritance`` flags (the helper
+        short-circuits via its own diff check on repeat visits, so redundancy
+        is safe).
         """
         if product is None or product.id is None:
             return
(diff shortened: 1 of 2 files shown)


valentijnscholten:
**Verdict: solid PR.** I verified the core claims against the current `bugfix` tip and found no correctness bugs. CI is fully green (both unit-test matrix variants, k8s, UI, performance).

### Verified correctness claims

- **The inheritance fix is a real regression fix, not just a refactor.** `_bulk_inherit_tags` no longer exists anywhere — it was removed in #14877 (`968cc5d6fd`). On today's `bugfix`, `dojo/management/commands/migrate_endpoints_to_locations.py:359` raises `AttributeError` for *every* product group, silently swallowed by the per-product `except Exception`. The inheritance pass is currently completely broken; this PR restores it. That alone justifies merging.
- **The new call is semantically sound.** `apply_inherited_tags_for_locations(locations, product=product)` no-ops when the gating product's flag is off, and on flag-on visits rediscovers each location's full product set from the ref tables and diff-checks — so the command's once-per-contributing-product grouping handles shared locations across mixed-flag products correctly, exactly as the (updated) comments describe. The pass runs after the main loop, so all `LocationProductReference`/`LocationFindingReference` rows it depends on are committed.
- **The atomicity design closes a real gap.** `bulk_add_tag_mapping`'s internal atomic only wraps the through-row `bulk_create`; the counter `UPDATE` runs after it. Without the PR's outer `transaction.atomic()`, a failure between those steps would leave rows without counter updates. The rollback test (`write_then_fail_once`) specifically validates that committed helper work is rolled back — good test.
- **Rerun idempotency holds.** `bulk_add_tag_mapping` pre-fetches existing `(instance, tag)` pairs and only counts genuinely new rows, so cross-batch revisits, the per-endpoint fallback re-covering a shared location, and full command reruns don't drift counters. The tests assert this at the through-row and counter level.
- **Flush placement is right.** Flushing on `i % batch_size == 0` runs regardless of per-endpoint failure (the removed `continue` is compensated by `_record_endpoint_failure` dedup), and the final partial-batch flush happens before `successful = i - len(self.failed_endpoints)` is computed, so tag-phase failures are reflected in the summary.

### Test quality

Genuinely good. The dedup test's pigeonhole construction (4 shared-host endpoints across 3+2 batches, plus the "shared" tag on the unique host) makes the `shared_locations_written == 3` assertion order-independent, as its comment claims — I checked the arithmetic. The mock patches the command module's imported name, so the inheritance pass's own `bulk_add_tag_mapping` calls don't pollute `call_count`. Plain `TestCase` + `override_settings(V3_FEATURE_LOCATIONS=True)` matches existing convention (e.g. `test_tag_inheritance_perf.py`).

### Minor observations (none blocking)

- **"Failed" endpoints can still get tags written.** Tags are queued in `_endpoint_to_url`; if a later phase (meta/refs) fails, the endpoint is recorded as failed but its queued tags still flush. This mirrors the base behavior (tags were written immediately, before the phases that could fail), and reruns heal it — but a one-line comment near `_queue_location_tags` noting this would help future readers.
- **`--batch-size` help text is stale.** It still reads "Endpoint.objects.iterator() chunk size", but the option now also controls tag-flush cadence and the `bulk_create` batch size inside the helper. Worth a small wording update.
- **`_record_endpoint_failure` dedups `None` ids**: two distinct endpoints that both report `id=None` would collapse into one failure entry. Practically unreachable (iterator rows always have pks) — fine as is.
- **Benchmark bookkeeping**: the "tags" phase now accumulates both per-endpoint queueing time and flush time, and `counts["tags"]` mixes endpoint and flush increments. `counts` isn't used in the summary output, so this is cosmetic.
- **Pre-existing helper quirk, not introduced here**: if two keys in one `tag_to_instances` mapping normalize to the same tag case-insensitively (e.g. `"Foo"` and `"foo"`), `bulk_add_tag_mapping` would double-count the counter and rely on `ignore_conflicts`. Not reachable from this command since tag names come from a single (already case-normalized) Endpoint tag table — just noting since this PR widens the helper's callers.
- **Coverage gap (small)**: no test for a non-tag endpoint failure landing exactly on a batch boundary (the "flush independently of per-endpoint success" branch). The logic is simple enough that I wouldn't hold the PR for it.

### Security / performance

No security concerns — management command, ORM-only queries, no user-controlled input paths, no new dependencies. The claimed query reduction (~41 → ~19-21 queries/endpoint first pass, ~7 on rerun) is consistent with what the code change would produce; the benchmark methodology in the description is unusually rigorous.

valentijnscholten (approved):
As per Claude's analysis good to go. We can always improve further if needed, but this one is already a big performance improvement. 
And a bugfix if Claude is right.

Maffooch (approved):

PR #15354, opened 2026-07-24
Tighten field-level authorization in the user API serializer

svader0:
Non-superusers are already prevented from changing `is_staff`, `is_superuser`, and configuration permissions on user records through the API. The account's identity fields were not covered by an equivalent guard, so a user holding the user-management permission could change them on accounts other than their own.

This adds the missing check so those fields are superuser-only when the target is a different account, consistent with the guards already in place. Editing your own account and superuser edits are unaffected.

Includes regression tests covering the new guard and its negative/positive controls.


svader0:
Superseded by #15191, which merged the same change to bugfix first. Closing this one.

PR #15359, opened 2026-07-27
docs(rbac): document Custom RBAC Roles

blakeaowens:
## What

DefectDojo Pro's RBAC is no longer a fixed set of five roles. Customers can
create their own roles by choosing individual permissions, and the five
built-ins (Reader / Writer / Maintainer / Owner / API Importer) become locked,
clonable presets. This documents the new surface and corrects the existing
role docs, which describe the five roles as the entire model.

## Changes

**New guide:** `admin/user_management/PRO__custom_rbac_roles.md`
- Enabling Custom Roles from Settings > Feature Flags, and what stays
  available while it is off (the Roles list remains readable)
- Who can manage roles: superusers and the built-in Owner global role only
- The Roles table, and creating a role by cloning a built-in or starting empty
- All four permission groups: Object Permissions, Group & Member Permissions,
  Global Feature Permissions (global-role only), and the Additional
  Permissions checklist
- How to read the grid: the View/Manage gate rule, and what checkboxes,
  shaded cells, inherited-View markers, and check/cross icons mean
- Built-in immutability, deleting a role that is still assigned, and
  reassignment via the API
- Every place a custom role can be assigned, and why owner-tier is reserved
- How a custom global role earns a built-in tier by fully covering it, and
  why Owner can never be earned that way
- Role history, an API reference table, and the operational caveats (union
  semantics, dropdown limit, propagation delays)
- Three screenshots: the Roles table, the role form with the permission grid,
  and the read-only permission view

**Corrected framing in the existing role docs**
- Permissions in DefectDojo: the five roles are now introduced as built-in
  locked presets that can be cloned, and the multiple-roles section documents
  the union rule, which was not written down anywhere
- Action permission charts: the chart is the built-in defaults rather than
  the definition of the model
- User Management index: links the new page

## Notes

- Docs only, no application code.
- Base is `bugfix` so the docs land alongside the release that introduces
  custom roles.
- Written against the merged implementation rather than the feature PR
  description, which had drifted: the gate is a feature-registry flag rather
  than a model column, there are three permission tables rather than two, the
  standalone permission list changed, Risk Acceptance is now a normal object
  row, and delete-with-reassign is API-only rather than exposed in the UI.


blakeaowens:
[sc-13957] Tracked in Shortcut: https://app.shortcut.com/defectdojo/story/13957

Maffooch (approved):

devGregA (approved):

Issue #15239, opened 2026-07-14
Read a fixed/patched version from the Dependency-Track FPF and set it as Finding mitigation


webdevred:
**Is your feature request related to a problem? Please describe**

The Dependency-Track parser (`dojo/tools/dependency_track/parser.py`) reads the component and vulnerability fields from the Finding Packaging Format (FPF), but it never reads a fixed or patched version. When an advisory has a known fix, that value does not reach the imported finding: the mitigation says nothing about which version resolves the vulnerability, so a user has to open the advisory and look it up by hand.

**Describe the solution you'd like**

I have been using DefectDojo together with DependencyTrack for some months now and I wished that the fixed version to appeared in the finding's mitigation, so I can see which release resolves the vulnerability without leaving DefectDojo. 

**Related to:** https://github.com/DependencyTrack/dependency-track/issues/5907 and DependencyTrack/dependency-track#1410

Concretely, in `dojo/tools/dependency_track/parser.py`: inside the loop that builds each finding, after the existing component and vulnerability extraction, read the fixed version from the FPF and set it on `Finding.mitigation` (for example "Fixed in 1.2.3") before the `Finding` is constructed. This reuses the existing `mitigation` field and adds no new model or Finding field, so it stays within the contributing scope.

There is a dependency worth stating up front: the FPF does not carry a fixed version today. Emitting it is a separate change on the Dependency-Track side. Nothing is committed there yet, so this issue is here to agree on direction first, in line with the guidance to open an issue before a PR.

**Describe alternatives you've considered**

* Do nothing and rely on the Dependency-Track UI to show the fixed version. That keeps it out of DefectDojo, which is where our triage actually happens.
* Put the full affected version range in mitigation instead of the single fixed version. Less actionable, since the useful part is the concrete "upgrade to" target.
* Add a dedicated structured field on the Finding for the fixed version. Ruled out on purpose, since the contributing guide says a new model or Finding field will not be approved.

**Additional context**

Two things I would like to settle here so the two projects line up:

1. Is a fixed version in the Dependency-Track parser something you would accept, mapped to `Finding.mitigation` as described above?
2. What should the field be named in the FPF payload? I will match that name on the Dependency-Track side, and Dependency-Track would bump `FPF_VERSION` when it adds the field.

If there is interest, I will open the PR here once the field name is agreed and Dependency-Track is emitting it.

webdevred:
Update on direction, this issue is partly outdated now.

[...]

PR #15366, opened 2026-07-27
Set finding mitigation from Dependency Track affected version ranges

webdevred:
**Description**

When Dependency Track pushes findings via the Finding Packaging Format, each vulnerability now carries the affected version ranges per component. DefectDojo dropped them, so a finding told an auditor a component was vulnerable but not which release to move to, leaving the upgrade target to be looked up by hand in the advisory.

This reads those ranges and sets the finding mitigation. For the range the installed component version falls in, it uses versionEndExcluding (the first patched version) to say "Upgrade to X or later", and falls back to "Upgrade to a version after X" from versionEndIncluding when no exact fix version is known. Ranges are matched to the component by purl and compared with packaging, so when a vulnerability spans several release lines the range the installed version actually sits in is the one used. Exact-version entries and component versions that do not parse are left without a mitigation rather than guessing, and findings whose vulnerability has no affected version ranges are unchanged.

This needs the Dependency Track side that emits the ranges (FPF 1.6). It implements #15239 and is the companion to the Dependency Track change that adds the affected version ranges to the FPF export (DependencyTrack/dependency-track#5907).

**Test results**

Added unit tests and FPF fixtures under `unittests/` covering the exact-fix case, the versionEndIncluding wording, multi-range selection (the installed version picks its own range), the exact-version-entry guard, and the unparseable-version fallback. The full `TestDependencyTrackParser` suite passes and the code is Ruff compliant. Also verified end to end by importing a sample FPF export into a running DefectDojo and confirming the mitigation was set on the resulting findings (for example log4j-core 2.14.1 gets "Upgrade to 2.15.0 or later").

**Documentation**

No documentation change needed. This enriches the output of the existing Dependency Track parser (it now sets the mitigation field) and does not add a new parser, setting, or model.

**Checklist**


webdevred:
Sorry, I tried to solve the merge conflict

blakeaowens (approved):

Maffooch (approved):

devGregA (approved):

PR #15374, opened 2026-07-28
fix(finding): clear dangling duplicate_finding refs during bulk delete

Maffooch:
**Description**

`bulk_delete_findings` fails at COMMIT whenever a finding that survives the delete still points at one that was deleted:

```
django.db.utils.IntegrityError: update or delete on table "dojo_finding" violates foreign key
constraint "dojo_finding_duplicate_finding_id_9ff391c5_fk_dojo_finding_id" on table "dojo_finding"
DETAIL:  Key (id)=(...) is still referenced from table "dojo_finding".
```

Root cause: `Finding.duplicate_finding` is a self-FK with `on_delete=DO_NOTHING` (`dojo/finding/models.py:221`), so Django performs no cleanup and Postgres enforces the constraint at COMMIT. `_bulk_delete_findings_internal` calls the cascade walker with `skip_relations={Finding}`, so nothing clears the pointer.

The existing `order_desc=True` ordering only helps when the referrer is inside the deletion set. In practice it often is not:

- transitive duplicate chains (`A → B → C`) — the condition `fix_loop_duplicates` exists to repair. Deleting `B` leaves `C` pointing at it.
- a caller that deletes only part of a cluster — e.g. `async_dupe_delete` trimming excess duplicates, where the `[:DUPE_DELETE_MAX_PER_RUN]` slice in `dojo/tasks.py` can cut a cluster in half.

The scope-based delete path already handles this: `prepare_duplicates_for_delete` nulls outside-scope references with the comment *"prevent FK violation"*. The bare-queryset path never got the same treatment.

**Fix**

Unlink surviving referrers inside each chunk's own transaction, before the chunk is deleted:

```python
 chunk_qs = Finding.objects.filter(id__in=chunk_ids)
 with transaction.atomic():
+    _clear_dangling_duplicate_finding_refs(chunk_ids)
     cascade_delete_related_objects(Finding, chunk_qs, skip_relations={Finding}, skip_m2m_for={Finding})
     execute_delete_sql(chunk_qs)
```

The new helper runs `Finding.objects.filter(duplicate_finding_id__in=chunk_ids).exclude(id__in=chunk_ids).update(duplicate_finding=None, duplicate=False)`. Per chunk rather than once up front, because each chunk commits separately — a single pre-pass would leave later chunks exposed. It is a no-op for callers that already ran `prepare_duplicates_for_delete`, so those paths are unchanged.

The filter is served by the existing `models.Index(fields=["duplicate_finding", "id"])`.

**Test results**

New file `unittests/test_bulk_delete_duplicate_fk.py`, 3 cases. Deferred FKs are validated the way COMMIT would via `connection.check_constraints()`, so the test reproduces the real failure rather than only asserting on resulting state.

| Case | Before | After |
|---|---|---|
| referrer outside the deletion set (transitive chain) | fails with the exact constraint error | passes |
| referrer inside the deletion set (control — already worked) | passes | passes |
| referrers spread across chunks (`chunk_size=1`) | fails with the exact constraint error | passes |

The control case passing before the fix confirms the diagnosis is specific rather than a broad breakage.

No collateral damage, run locally against PostgreSQL:

- `test_duplication_loops`, `test_cascade_delete`, `test_async_delete`, `test_deduplication_logic` — 107 pass, 2 skipped
- `test_prepare_duplicates_for_delete`, `test_finding_helper`, `test_finding_model`, `test_delete_with_endpoints_v3`, `test_importers_deduplication`, `test_utils_deduplication_reopen` — 186 pass, 29 skipped

**Documentation**

No documentation changes needed — this is an internal bug fix with no change to configuration, API surface, or user-facing behaviour. The reasoning is captured in the new helper's docstring and in `_bulk_delete_findings_internal`'s.

**Checklist**

- [x] Make sure to rebase your PR against the very latest `dev`.  *(bugfix — branched off current `bugfix` tip)*
- [ ] Features/Changes should be submitted against the `dev`.  *(n/a — bugfix)*
- [ ] If this is a new feature and not a bug fix, you've included the proper documentation in the docs.  *(n/a — bugfix)*
- [ ] Model changes must include the necessary migrations in the dojo/db_migrations folder.  *(n/a — no model changes)*
- [ ] Add the proper label to categorize your PR.  *(suggested: `bugfix`)*

### DIFF
--- a/dojo/finding/helper.py
+++ b/dojo/finding/helper.py
@@ -819,6 +819,34 @@ def bulk_clear_finding_m2m(finding_qs):
         Notes.objects.filter(id__in=note_ids).delete()
 
 
+def _clear_dangling_duplicate_finding_refs(chunk_ids):
+    """
+    Unlink findings that point at this chunk via duplicate_finding but survive it.
+
+    Finding.duplicate_finding is a self-FK with on_delete=DO_NOTHING, so Django
+    performs no cleanup and Postgres enforces the constraint at COMMIT. The
+    cascade walker is called with skip_relations={Finding}, so nothing else
+    clears the pointer.
+
+    Ordering alone cannot prevent the violation: the referrer is frequently not
+    in the deletion set at all (transitive duplicate chains, or a caller that
+    deletes only part of a cluster - e.g. async_dupe_delete trimming excess
+    duplicates). This mirrors what prepare_duplicates_for_delete already does
+    for scope-based deletes, and is a no-op when that ran first.
+    """
+    unlinked = Finding.objects.filter(
+        duplicate_finding_id__in=chunk_ids,
+    ).exclude(
+        id__in=chunk_ids,
+    ).update(duplicate_finding=None, duplicate=False)
+    if unlinked:
+        logger.debug(
+            "bulk_delete_findings: unlinked %d surviving duplicate_finding references",
+            unlinked,
+        )
+    return unlinked
+
+
 def _bulk_delete_findings_internal(finding_qs, chunk_size=1000, *, order_desc=False):
     """
     Delete findings and all related objects efficiently. Including any related object in Dojo-Pro
@@ -831,6 +859,9 @@ def _bulk_delete_findings_internal(finding_qs, chunk_size=1000, *, order_desc=Fa
     When order_desc is True, findings are processed highest id first (matches
     finding_delete: duplicate_cluster.order_by("-id").delete()) so self-FK
     duplicate chains delete children before parents.
+
+    Before each chunk is deleted, surviving findings that still point at it via
+    duplicate_finding are unlinked; see _clear_dangling_duplicate_finding_refs.
     """
     from dojo.signals import pre_bulk_delete_findings  # noqa: PLC0415 circular import
     from dojo.utils_cascade_delete import (  # noqa: PLC0415 circular import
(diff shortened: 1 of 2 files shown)


Maffooch:
Closing as redundant — #15364 already fixes this, and fixes it better.

Same root cause (the `duplicate_finding` self-FK is `on_delete=DO_NOTHING`, so a survivor pointing into the delete set fails the constraint at COMMIT), found independently. #15364 predates this PR by several hours; I missed it when triaging because I searched commit history but not open PRs. My fault.

The two differ in a way that matters. #15364 re-points the survivor at its chain's first surviving ancestor, preserving the duplicate relationship. This PR just unlinks (`duplicate_finding=None, duplicate=False`), which promotes the survivor to an original and resurrects it as a distinct finding — lossy, and worse behaviour. #15364 also already covers the `DUPE_DELETE_MAX_PER_RUN` truncation case, since its survivor query catches anything pointing into the delete set from outside regardless of cause.

The one thing this PR had that #15364 doesn't is coverage of every `bulk_delete_findings` caller rather than just `async_dupe_delete`. That turns out to be thin: `utils.py:1878` is already preceded by `prepare_duplicates_for_delete`, and the UI bulk delete uses per-object `find.delete()`, so the only genuinely uncovered site is `helper.py:711` (cascade-mode outside-scope deletion). Worth hardening at some point, but as a separate change and using the re-point semantics rather than these — not a reason to keep this PR open.

Two notes for whoever picks up #15364:

- It targets `dev` (3.2.0). If the 3.1.x line needs this sooner, a backport to `bugfix` would be the way, rather than reviving this.
- `helper.py:711` remains a theoretical exposure after #15364 lands.

---

PR #15376, opened 2026-07-28
fix(risk_acceptance): stop the expiration job aborting on an unattached risk acceptance

Maffooch:
**Description**

The `dojo.risk_acceptance.helper.expiration_handler` Celery task aborts with:

```
File "dojo/risk_acceptance/helper.py", line 203, in expiration_handler
    str(risk_acceptance.engagement.product) + ": " + str(risk_acceptance.engagement.name)
AttributeError: 'NoneType' object has no attribute 'product'
```

Root cause: `Risk_Acceptance.engagement` (`dojo/risk_acceptance/models.py:90-96`) is a property over `engagement_set` — the reverse side of `Engagement.risk_acceptance`, a many-to-many used as a one-to-many. It returns `None` when no engagement points at the risk acceptance. Both halves of `expiration_handler` dereferenced it unconditionally while building the notification: `expire_now` at line 54, the heads-up loop at line 203.

**An empty engagement relation is a supported state, not corrupt data.** Several shipped paths produce one:

- `RiskAcceptanceSerializer.create` attaches an engagement **only if** `accepted_findings.exists()` (`dojo/risk_acceptance/api/serializer.py:36-40`). Create a risk acceptance with an expiration date and no findings and it is unattached from birth.
- `_accept_risks`, the bulk accept-risks API (`dojo/risk_acceptance/api/mixins.py:101`), and the SonarQube `WONTFIX` status sync (`dojo/tools/api_sonarqube/updater_from_source.py:101`) both create risk acceptances and never attach an engagement.
- `RiskAcceptanceSerializer.update` has an explicit repair branch for exactly this state, commented *"Handle orphaned risk acceptances"* (`serializer.py:66-70`).
- The same file notes twice that the relation is not the source of truth everywhere: *"This is fine as Pro has its own model + relationshop to track links with engagements."*

The module already concedes the point elsewhere: `get_view_risk_acceptance` (line 216) wraps the identical dereference in `contextlib.suppress(AttributeError)` with the comment *"it does not happen under most circumstances that a risk acceptance does not have engagement"*. The expiration paths never got the same treatment.

**Why the blast radius is larger than one row**

- The exception escapes the `for` loop, so **every risk acceptance later in the batch is left unhandled** — no expiry, no heads-up notification, for unrelated products.
- In the heads-up loop, `expiration_date_warned` is only set *after* `create_notification`, so the offending row is never marked handled. It is re-selected by `get_almost_expired_risk_acceptances_to_handle` and **fails the job again on every subsequent run** until its expiration date passes.
- In `expire_now`, the save at lines 48-50 happens *before* the crash, so the row is already marked expired when the task dies — the notification is silently lost.

**Fix**

Resolve the engagement through a new `engagement_to_notify_about()` helper:

1. Use `risk_acceptance.engagement` when the relation is populated — unchanged behaviour.
2. Otherwise fall back to the engagement of an accepted finding. This is the **same derivation** `RiskAcceptanceSerializer` already uses both to attach an engagement on create and to repair an orphan on update, so it introduces no new notion of ownership.
3. Only when there is neither an engagement nor any findings, log a warning and skip the notification. The notification names a product and reverses a per-engagement URL; neither can be produced from an empty risk acceptance, and a "0 accepted findings will expire" notice has no recipient worth naming.

Everything other than the notification is untouched — the expiry, the finding reactivation and the JIRA comments all still run. The heads-up loop now sets `expiration_date_warned` in every branch, so a risk acceptance that genuinely cannot be notified about stops being re-selected on each run instead of failing the job daily.

The fallback iterates the already-prefetched `accepted_findings` rather than calling `.first()`, so it costs no extra query on the common path.

*Known limitation, deliberately not solved here:* for a risk acceptance whose findings span several engagements, the notification is addressed to the first one. That matches what `RiskAcceptanceSerializer`'s orphan repair already does, and doing better requires ownership information this module does not have. Worth revisiting separately if multi-engagement risk acceptances should fan out notifications.

**Test results**

Three regression tests in `unittests/test_risk_acceptance.py`, one per distinct case. Each drives the real `expiration_handler` with a mix of attached and unattached risk acceptances, patches `create_notification`, and asserts exactly which ones are notified about — and for the fallback cases, that the notification carries the *derived* engagement and its product, not merely that it fired.

| Case | Against unmodified `bugfix` | With this PR |
|---|---|---|
| heads-up, engagement link missing, findings present | `AttributeError` at `helper.py:203` | notified via derived engagement |
| heads-up, no engagement and no findings | `AttributeError` at `helper.py:203` | skipped, still marked warned |
| expiry, engagement link missing, findings present | `AttributeError` at `helper.py:54` | expired + notified via derived engagement |
| `test_expiration_handler` (existing control — all attached) | passes | passes |

All three reproduce the reported traceback at the reported lines before the change. The existing control passing beforehand confirms the diagnosis is specific rather than a broad breakage. Each test also asserts the *rest* of the batch is still processed, which is the part the abort was destroying.

Run locally against PostgreSQL with `manage.py test`:

- `unittests.test_risk_acceptance` — 14 pass (11 existing + 3 new)
- with `test_risk_acceptance_api`, `test_bulk_risk_acceptance_api`, `test_notifications` — 78 pass, 24 skipped

`ruff check --config ruff.toml` clean on both changed files.

**Documentation**

No documentation changes needed — internal bug fix, no change to configuration, API surface, or user-facing behaviour. The reasoning is captured in the new helper's docstring and at the call sites.

**Checklist**

- [x] Make sure to rebase your PR against the very latest `dev`.  *(bugfix — branched off current `bugfix` tip)*
- [ ] Features/Changes should be submitted against the `dev`.  *(n/a — bugfix)*
- [ ] If this is a new feature and not a bug fix, you've included the proper documentation in the docs.  *(n/a — bugfix)*
- [ ] Model changes must include the necessary migrations in the dojo/db_migrations folder.  *(n/a — no model changes)*

### DIFF
--- a/dojo/risk_acceptance/helper.py
+++ b/dojo/risk_acceptance/helper.py
@@ -18,20 +18,35 @@ logger = logging.getLogger(__name__)
 
 def engagement_to_notify_about(risk_acceptance: Risk_Acceptance) -> Engagement | None:
     """
-    Return the engagement backing a risk acceptance, or None when it has none.
+    Return an engagement to address a risk acceptance's expiration notification to.
 
-    `Risk_Acceptance.engagement` is a property over a many-to-many that is used as a
-    one-to-many, so it is None for a risk acceptance no engagement points at. Expiration
-    notifications are addressed to the engagement's product and link to a per-engagement
-    URL, so there is nothing to send when the engagement is missing.
+    `Risk_Acceptance.engagement` is a property over `Engagement.risk_acceptance`, a
+    many-to-many used as a one-to-many. It is empty for a risk acceptance that was not
+    created from an engagement page, which is a supported state rather than corrupt data:
+    `RiskAcceptanceSerializer.create` only attaches an engagement when the risk acceptance
+    already has findings, `_accept_risks` and the SonarQube status sync never attach one at
+    all, and the Pro plugin tracks the association on its own model instead of this relation.
+
+    So fall back to the engagement of an accepted finding - the same derivation
+    `RiskAcceptanceSerializer` uses both to attach an engagement on create and to repair an
+    orphan on update. Only a risk acceptance with neither an engagement nor any findings has
+    nothing to notify about; the notification names a product and links to a per-engagement
+    URL, and neither can be produced from an empty risk acceptance.
     """
     engagement = risk_acceptance.engagement
-    if engagement is None:
-        logger.warning(
-            "risk acceptance %i:%s is not attached to an engagement, skipping its expiration notification",
-            risk_acceptance.id, risk_acceptance,
-        )
-    return engagement
+    if engagement is not None:
+        return engagement
+
+    # iterate rather than .first(), to reuse the accepted_findings prefetch
+    accepted_finding = next(iter(risk_acceptance.accepted_findings.all()), None)
+    if accepted_finding is not None:
+        return accepted_finding.test.engagement
+
+    logger.warning(
+        "risk acceptance %i:%s has no engagement and no findings, skipping its expiration notification",
+        risk_acceptance.id, risk_acceptance,
+    )
+    return None
 
 
 def expire_now(risk_acceptance):
--- a/dojo/risk_acceptance/helper.py
+++ b/dojo/risk_acceptance/helper.py
@@ -9,11 +9,29 @@ from django.utils import timezone
 
 from dojo.celery import app
 from dojo.jira import services as jira_services
-from dojo.models import Dojo_User, Finding, Notes, Risk_Acceptance, System_Settings
+from dojo.models import Dojo_User, Engagement, Finding, Notes, Risk_Acceptance, System_Settings
 from dojo.notifications.helper import create_notification
 from dojo.utils import get_full_url, get_system_setting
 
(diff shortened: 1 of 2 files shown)


Maffooch:
**Related PR:** DefectDojo-Inc/dojo-pro#2068 — *fix(authorization): scope risk acceptances by their accepted findings*

Same root observation, different symptom. Investigating this crash established that an empty `Engagement.risk_acceptance` relation is a **supported state** rather than corrupt data — open source only populates it when the risk acceptance already has findings at creation time, and the Pro plugin tracks the association on its own model.

That turned out to also be an authorization gap on the Pro side: risk acceptances scoped only through their accepted findings were invisible to the very users who own the covered product. The Pro PR fixes that in `_get_authorized_risk_acceptances` and the matching object-level check.

The two are independent — this PR's tests run against unmodified `bugfix`, and the Pro PR's tests run against unmodified open source — but they should be reviewed together, since they encode the same assumption about when that relation may legitimately be empty.

---

blakeaowens (approved):

devGregA (approved):

PR #15382, opened 2026-07-28
Preserve raw CycloneDX VEX analysis on parsed findings

Jino-T:
## Summary

The CycloneDX JSON parser collapses each `vulnerabilities[].analysis` block into Dojo status booleans (`is_mitigated` / `false_p`) and drops `justification` and `response` entirely. This attaches the raw `analysis` dict to the in-memory finding as `finding.unsaved_vex` so downstream consumers can round-trip the full VEX data (`state` / `justification` / `response` / `detail`).

## Why

VEX (Vulnerability Exploitability eXchange) analysis carries more than a mitigated/false-positive flag — the justification and response arrays are meaningful and standardized in CycloneDX 1.7. Preserving the original block lets a consumer re-export a schema-valid VEX document instead of reconstructing a lossy approximation from Dojo status.

## Behavior

`unsaved_vex` is an **inert in-memory attribute** — the parser doesn't persist it and nothing in OSS reads it, so existing behavior is completely unchanged. It follows the same `unsaved_*` convention the importers already use for `unsaved_vulnerability_ids`, `unsaved_endpoints`, etc.: a place for parsed-but-not-yet-persisted data that post-save consumers can pick up.

## Change

A single line in `dojo/tools/cyclonedx/json_parser.py`, set right after the existing `analysis` handling:

```python
finding.unsaved_vex = analysis
```

No schema, model, or migration changes.


rossops:
Closed accidentally due to the bugfix branch getting deleted by some faulty release automation.

rossops:
Closed accidentally due to the bugfix branch getting deleted by some faulty release automation.

blakeaowens (approved):

Maffooch (approved):

devGregA (approved):

PR #15428, opened 2026-07-30
fix(importers): skip child rows buffered for a finding deleted mid-batch

Maffooch:
**Description**

An async reimport fails with a `500` / failed task and this `IntegrityError` whenever a finding it has already processed is deleted before the batch it belongs to is flushed:

```
django.db.utils.IntegrityError: insert or update on table "dojo_vulnerability_id"
violates foreign key constraint "dojo_vulnerability_r_finding_id_c9d59919_fk_dojo_find"
DETAIL:  Key (finding_id)=(...) is not present in table "dojo_finding".
```

Reported twice within four seconds from a production instance on `reimport-scan` (Generic Findings Import), against two different tests of the same product — the shape of a delete racing a batch of reimports rather than a one-off.

The sequence:

1. Vulnerability ids are **not** written as each finding is processed. `store_vulnerability_ids()` (import) and `reconcile_vulnerability_ids()` (reimport) append `Vulnerability_Id` objects to `self.pending_vulnerability_ids`, and `flush_vulnerability_ids()` bulk-inserts them at the batch boundary — up to a thousand findings later.
2. A finding referenced by that buffer can be deleted inside the window. `async_dupe_delete` deletes excess duplicates when `delete_duplicates` is enabled, and a user deleting findings or a delete cascading from a test or engagement does the same.
3. The bulk insert then carries a `finding_id` that is no longer in `dojo_finding`.

Because Django declares its foreign keys `DEFERRABLE INITIALLY DEFERRED`, that insert does not fail where it is issued — PostgreSQL raises it at `COMMIT`. That is why the reported traceback contains **no application frames at all**, only the commit path: the failure surfaces past every handler that still knew what the import was doing. The whole import dies, taking the rest of the batch — findings that were perfectly fine — with it.

`BurpRawRequestResponse` is buffered and flushed through exactly the same pattern (`process_request_response_pairs` → `flush_burp_request_response`) and has the same exposure, so it is fixed alongside rather than left as the next incident.

### The change

Both flushes now filter the buffer through a new `BaseImporter.drop_rows_for_deleted_findings()`, which keeps only the rows whose finding still exists and logs at warning level which finding ids were dropped and how many rows went with them.

Dropping is the correct outcome rather than a suppression: a finding that has been deleted has nothing to own these child rows. This is the same posture `close_old_findings()` already takes — `_sync_close_old_finding_status_fields()` returns only the candidates that still exist, precisely because a candidate can be deleted mid-reimport (#15408). This PR extends that guard to the other place in the same run that holds finding references across a batch.

Deliberately out of scope: why a finding may be deleted mid-run (that is the caller's workflow), and the `pending_vuln_id_deletes` half of the flush, which is a `DELETE ... WHERE finding_id IN (...)` and is unaffected by a missing row.

### Performance impact

One indexed primary-key `SELECT` per flush **that has something to insert**, independent of finding count. Measured across the pinned baselines, every delta is exactly **+1**, and the no-change reimport — where nothing is buffered, so no lookup happens — is **unchanged**. That split is itself the evidence the cost is O(1) per flush and not O(findings):

| | before | after |
|---|---|---|
| `EXPECTED_ZAP_IMPORT_V2` / `_V3` | 287 / 311 | 288 / 312 |
| `EXPECTED_ZAP_REIMPORT_WITH_NEW_V2` / `_V3` | 148 / 177 | 149 / 178 |
| `EXPECTED_ZAP_REIMPORT_NO_CHANGE_V2` / `_V3` | 74 / 86 | 74 / 86 (unchanged) |

In `test_importers_performance.py`, `expected_num_queries1` and `expected_num_queries2` move +1 in all 11 affected cases; `expected_num_queries3` (no-change reimport) and `expected_num_queries4` are untouched. Async task counts are unchanged.

### Impact on the Pro plugin

Checked as part of the acceptance criteria.

- No Pro code change is required for the fix itself. `ProImporter` / `ProReImporter` derive from `DefaultImporter` / `DefaultReImporter` and override neither `flush_vulnerability_ids`, `flush_burp_request_response`, `store_vulnerability_ids` nor `reconcile_vulnerability_ids`, so they inherit the guard — which is what matters here, because the reported traceback runs through the Pro async importer into this file. No Pro module bulk-creates `Vulnerability_Id` on the import path, so there is no second copy of this defect to fix there.
- The new method name does not collide with anything in the Pro importer hierarchy.
- **Pro does need a follow-up commit for its own pinned query counts.** Pro keeps a separate `test_importers_performance.py` with its own `expected_num_queries*` values, which will shift by the same +1 on the buffering paths (and stay unchanged on the no-change reimport) once this merges. Those numbers have to be **measured** in the Pro stack, not carried over from the table above, and this sandbox has no Docker so they could not be produced here. Flagging rather than guessing.
- One Pro observation worth noting separately, not a blocker for this PR: `pro/finding/helpers.py` creates a single `Vulnerability_Id` during EPSS enrichment for a finding held in memory, which is the same class of exposure in a different task. Different code path, different repo, not touched here.

**Test results**

New `unittests/test_importers_deleted_finding_child_rows.py`, 6 tests. The 4 guard tests were confirmed failing on `bugfix` before the change, each with the dangling-reference violation this PR is about:

```
FAILED (failures=1, errors=4)
django.db.utils.IntegrityError: update or delete on table "dojo_finding" violates foreign key
constraint "dojo_vulnerability_r_finding_id_c9d59919_fk_dojo_find" on table "dojo_vulnerability_id"
django.db.utils.IntegrityError: update or delete on table "dojo_finding" violates foreign key
constraint "dojo_burprawrequestr_finding_id_57e40ea9_fk_dojo_find" on table "dojo_burprawrequestresponse"
```

Each test calls `connection.check_constraints()` after the flush — the check PostgreSQL runs at `COMMIT` in production. The test transaction never commits, so without asking for it explicitly the production failure would not appear in a test at all.

Coverage: the import-path flush with a finding deleted after buffering; the reported reimport path through `reconcile_vulnerability_ids`; that a dropped row is not retried by the next batch's flush; the request/response buffer; and two control cases asserting a run where nothing was deleted still writes every row.

Run against PostgreSQL on Python 3.13 (the deferred-constraint behaviour is backend-specific):

- `unittests.test_importers_deleted_finding_child_rows` — **6 passed**
- `unittests.test_importers_performance`, `unittests.test_tag_inheritance_perf` — **35 passed** (26 count assertions failing on the un-updated baselines first, then green on the updated ones)
- `unittests.test_import_reimport`, `test_importers_importer`, `test_importers_closeold`, `test_importers_deduplication`, `test_reimport_prefetch`, `test_import_execution_mode`, `test_duplication_loops` — **283 passed, 1 skipped, 0 failures**
- `ruff check --config ruff.toml` with the pinned 0.15.20 — clean on all four touched files

> Environment note: this sandbox has no Docker, so suites were run through `manage.py test` against a locally provisioned PostgreSQL 16 cluster rather than `run-unittest.sh`, and `scripts/update_performance_test_counts.py` (which shells out to `run-unittest.sh`) could not be used. Every updated count is the value the assertion itself reported, re-verified on a second full run. CI is the authority.

**Documentation**

No documentation change: this restores the documented behaviour of import/reimport (the failure was never an expected outcome) and adds no setting, input, output, or API surface.

**Checklist**

- [x] Bugfixes should be submitted against the `bugfix` branch — based on and targeting `bugfix`.
- [x] Your code is python 3.13 compliant — tests were run on 3.13.
- [ ] Model changes must include the necessary migrations — N/A, no model changes.

Reported internally:
- https://defectdojo-workspace.slack.com/archives/C0BK62Q76P8/p1785386119745169
- https://defectdojo-workspace.slack.com/archives/C0BK62Q76P8/p1785386123906449

https://claude.ai/code/session_01WrYDZdT84SC27Mz53w1Yuj

---
_Generated by [Claude Code](https://claude.ai/code/session_01WrYDZdT84SC27Mz53w1Yuj)_

### DIFF
--- a/dojo/importers/base_importer.py
+++ b/dojo/importers/base_importer.py
@@ -838,9 +838,50 @@ class BaseImporter(ImporterOptions):
                 burpResponseBase64=base64.b64encode(unsaved_response.encode()),
             ))
 
+    def drop_rows_for_deleted_findings(self, rows: list) -> list:
+        """
+        Return only the buffered child rows whose finding still exists.
+
+        Child rows are buffered while a batch of up to a thousand findings is processed
+        and inserted in bulk at the batch boundary, so a finding stays referenced by the
+        buffer for as long as the rest of its batch takes. A finding row can go away
+        inside that window: the excess-duplicate cleanup task deletes findings, and so
+        does a user deleting findings or a delete cascading from a test or engagement.
+        The buffered row then points at a primary key that is no longer in dojo_finding.
+
+        Because Django declares its foreign keys DEFERRABLE INITIALLY DEFERRED, that
+        insert does not fail where it is issued. PostgreSQL raises it at COMMIT, past
+        every handler that knew what the import was doing, so the whole import dies with
+        an IntegrityError naming a database constraint and takes the rest of the batch --
+        findings that were perfectly fine -- with it.
+
+        A finding that is gone has nothing to own these rows, so dropping just those rows
+        is the correct outcome, and it costs one indexed primary-key lookup per flush.
+        Rows whose finding id is not resolvable yet are left alone: bulk_create fills the
+        id in from the related object, and reporting on them is its job, not this guard's.
+        """
+        finding_ids = {row.finding_id for row in rows if row.finding_id is not None}
+        if not finding_ids:
+            return rows
+        live_finding_ids = set(
+            Finding.objects.filter(pk__in=finding_ids).values_list("pk", flat=True),
+        )
+        deleted_finding_ids = finding_ids - live_finding_ids
+        if not deleted_finding_ids:
+            return rows
+        logger.warning(
+            "skipping %s row(s) buffered for %s finding(s) deleted during the import: %s",
+            sum(1 for row in rows if row.finding_id in deleted_finding_ids),
+            len(deleted_finding_ids),
+            sorted(deleted_finding_ids),
+        )
+        return [row for row in rows if row.finding_id not in deleted_finding_ids]
+
     def flush_burp_request_response(self) -> None:
         if self.pending_burp_rr:
-            BurpRawRequestResponse.objects.bulk_create(self.pending_burp_rr, batch_size=1000)
+            rows = self.drop_rows_for_deleted_findings(self.pending_burp_rr)
+            if rows:
+                BurpRawRequestResponse.objects.bulk_create(rows, batch_size=1000)
             self.pending_burp_rr.clear()
 
     def process_locations(
@@ -908,7 +949,9 @@ class BaseImporter(ImporterOptions):
             Vulnerability_Id.objects.filter(finding_id__in=self.pending_vuln_id_deletes).delete()
             self.pending_vuln_id_deletes.clear()
         if self.pending_vulnerability_ids:
-            Vulnerability_Id.objects.bulk_create(self.pending_vulnerability_ids, batch_size=1000)
+            rows = self.drop_rows_for_deleted_findings(self.pending_vulnerability_ids)
+            if rows:
+                Vulnerability_Id.objects.bulk_create(rows, batch_size=1000)
             self.pending_vulnerability_ids.clear()
 
     def process_files(
--- a/unittests/test_importers_deleted_finding_child_rows.py
+++ b/unittests/test_importers_deleted_finding_child_rows.py
@@ -1,0 +1,248 @@
+import base64
+import logging
+
+from django.db import connection
+from django.utils import timezone
+
+from dojo.importers.default_importer import DefaultImporter
+from dojo.importers.default_reimporter import DefaultReImporter
+from dojo.models import (
+    BurpRawRequestResponse,
+    Development_Environment,
+    Engagement,
+    Finding,
+    Product,
+    Product_Type,
+    User,
+    Vulnerability_Id,
+)
+
+from .dojo_test_case import DojoTestCase, get_unit_tests_scans_path
+
+logger = logging.getLogger(__name__)
+
+SCAN_TYPE = "Acunetix Scan"
+SCAN_FILE = "many_findings.xml"
+REQUEST = b"GET / HTTP/1.1"
+RESPONSE = b"HTTP/1.1 200 OK"
+
+
+class TestImportersDeletedFindingChildRows(DojoTestCase):
+
+    """
+    Regression: child rows buffered for a finding deleted mid-batch broke the whole import.
+
+    Vulnerability ids and Burp request/response pairs are not written as each finding is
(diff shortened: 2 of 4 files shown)


blakeaowens (approved):

devGregA (approved):

Maffooch:
### CI failure on `983f09d8` is a pre-existing flake, not this PR

`test-rest-framework (linux/arm64, false)` failed with one test, and it is not in this diff:

```
FAIL: test_api_token_auth_is_rate_limited (unittests.test_apiv2_user.UserTest)
AssertionError: 4 != 3 : api-token-auth should honor the configured rate limit
Ran 4227 tests — FAILED (failures=1, skipped=574)
```

Evidence it is not caused by this branch:

- **Same commit, 3 of 4 unit-test jobs passed.** `linux/amd64 (false)`, `linux/amd64 (true)` and `linux/arm64 (true)` are all green on `983f09d8`; only `arm64 (false)` failed. Identical code, different outcome, so the input is time, not the diff.
- **Different subsystem.** This PR touches `dojo/importers/base_importer.py` and three test files. Nothing in it reaches rate limiting, `dojo/decorators.py`, or `unittests/test_apiv2_user.py`. That test arrived with #15415, which is already on `bugfix`.
- **Passes locally, 3 runs of 3**, against PostgreSQL on Python 3.13.

**Why it flakes.** `dojo_ratelimit` delegates to django-ratelimit, whose `_get_window()` pins the counter to a *fixed* window edge derived from the epoch — `ts - (ts % period) + (crc32(key) % period)` — and that window is part of the cache key. The test fires 8 requests under `3/m` and asserts exactly 3 reach the view. If the window edge falls inside that 8-request loop, the counter resets mid-loop and a 4th request gets through: `4 != 3`. The edge lands at one specific second per minute (jittered per client IP), so the test fails whenever the loop happens to straddle it — more likely on the slower arm64 runner. The assertion is exact-equality against a wall-clock-dependent boundary, which is the defect.

I am **not fixing it here**: it is an unrelated test in an unrelated subsystem, and quietly editing it inside a bugfix PR would hide a real flake behind an unrelated change. It deserves its own PR (waiting for a fresh window before the loop, or asserting `<= 3`). Flagging rather than absorbing it.

I have re-run the failed job. Everything from this PR passed on all four jobs, including the merged query-count baselines.

### On the baseline merge conflict resolution

Independently re-verified locally on the merged tree (`983f09d8`) after `bugfix` brought in #15427: **47 tests, all passing** — `test_importers_performance`, `test_tag_inheritance_perf`, `test_importers_deleted_finding_child_rows` (this PR) and `test_importers_deleted_target` (#15427). So the two guards compose as the merge commit assumed: each contested baseline is `bugfix`'s value plus this branch's `+1`, and the no-change reimport stays flat because nothing is buffered on that path.

---

PR #15447, opened 2026-07-31
fix(product): defer list count annotations

### LINKED ISSUE #15378: Product list page query becomes extremely slow
The Product list view becomes very slow when loading products with findings and location-related annotations enabled.

In my DefectDojo instance, opening the Product/Asset list page results in a database query taking around 66 seconds.

`EXPLAIN ANALYZE` shows that expensive annotations are being evaluated before pagination. The queryset performs correlated subqueries for finding counts, which are executed repeatedly for intermediate rows created by location joins.

Observed query details:
- ~110 products
- ~60,000 `dojo_locationproductreference` records
- ~11,700 findings
- ~600 tests

Relevant query plan observations:
- Execution time: ~66 seconds
- `dojo_test` sequential scan loops: 60,168
- Finding count subqueries executed: 60,168 times
- Shared buffer hits: ~35 million

No application error is shown; the issue is excessive query execution time.

**Steps to reproduce**
1. Populate DefectDojo with a dataset containing a significant number of products, findings, tests, and locations.
2. Enable product location functionality (`V3_FEATURE_LOCATIONS`).
3. Go to **Products**.
4. Load the product list page.
5. Observe that the page takes a long time to load.
6. Capture the SQL query with `EXPLAIN (ANALYZE, BUFFERS)` and observe repeated finding count subquery execution.

**Expected behavior**
The Product list page should load within a reasonable time even with larger datasets.

Expensive calculations such as finding counts and engagement counts should not be executed thousands of times for intermediate rows when only 25 products are displayed per page.

- [X] Docker Compose
- [ ] Kubernetes
- [ ] GoDojo

**Environment information**
 - Docker Compose version: v2.35.1
 - DefectDojo version: v. 3.1.300
 - Database: PostgreSQL 18.4

**Logs**
No application error logs are generated. The issue is database query performance.

kocaemre:

This is a focused bugfix for an open issue: fixes #15378.

**Description**

Defers Product list count annotations until after filtering and pagination unless the request explicitly sorts by `findings_count`.

With `V3_FEATURE_LOCATIONS` enabled, the Product list currently adds finding and location count annotations before pagination. That can make PostgreSQL evaluate expensive correlated finding-count subqueries for intermediate rows created by location joins, even though the page only renders 25 products.

This PR:
- keeps the `findings_count` annotation before pagination only when the Product list is sorted by that field;
- applies `distinct()` to the filtered v3 product queryset before pagination so location joins do not duplicate products;
- moves rendered page-only `findings_count`, `location_host_count`, and `location_count` annotations into `prefetch_for_product()`.

**Test results**

- [x] `.venv/bin/python manage.py test unittests.test_product_list_pagination -v 2`
- [x] `.venv/bin/python -m ruff check --config ruff.toml dojo/product/ui/views.py unittests/test_product_list_pagination.py`
- [x] `.venv/bin/python -m py_compile dojo/product/ui/views.py unittests/test_product_list_pagination.py`
- [x] `/root/.local/share/uv/python/cpython-3.13-linux-x86_64-gnu/bin/python3.13 -m py_compile dojo/product/ui/views.py unittests/test_product_list_pagination.py`
- [x] `git diff --check HEAD~1..HEAD`

The Django test run passed with the existing local warning that `components/node_modules` is missing from `STATICFILES_DIRS`.

**Documentation**

No documentation update needed; this is a Product list query-planning/performance bugfix.

**Checklist**


rossops:
Closed accidentally due to the bugfix branch getting deleted by some faulty release automation.

valentijnscholten (commented):
Thanks for digging into #15378 — the slow query is real and the location annotations are indeed involved. But I don't think this diff fixes it, and I'd like to suggest a much smaller change instead. I compiled the querysets against Django 5.2 to look at the generated SQL.

### The counting side is already covered

Two mechanisms already exist:

1. `get_page_items()` in `dojo/utils.py` calls `get_page_items_and_count(..., do_count=False)`, so there is no extra `total_count` query for this page.
2. Django masks unused annotations out of the Paginator's own `COUNT` query (`Query.get_aggregation()` → `set_annotation_mask(aggregates)`). With only the `findings_count` subquery annotated, `count()` compiles to literally `SELECT COUNT(*) FROM dojo_product` — the subquery is dropped.

So "annotations before pagination are expensive" isn't the mechanism here. A correlated `Subquery` annotation before pagination is nearly free.

### The actual cause is narrower

The two `Count("locations…", distinct=True)` annotations add a `LEFT JOIN` **and** force a `GROUP BY` — and Django puts the correlated findings subquery into that `GROUP BY`, by ordinal:

```sql
GROUP BY "dojo_product"."id", ..., 3   -- 3 = the findings_count subquery
```

That is the 60,168 loops in the issue: the subquery becomes a grouping key, so it is evaluated once per joined location row instead of once per product. The aggregate also makes `has_existing_aggregation` true, which defeats the count-query masking above, so the Paginator's count wraps the same join + group and pays it a second time.

### Why this PR doesn't remove the fan-out

`prefetch_for_product()` already carries six correlated subqueries (`active_engagement_count`, `closed_engagement_count`, `last_engagement_date`, `active_finding_count`, `active_verified_finding_count`, `total_reimport_count`). Moving the location aggregates *into that queryset* reproduces the same shape in the page query, now with ~7 subqueries as grouping keys — and `LIMIT 25` is applied after the grouping, so it doesn't bound the scan:

```sql
SELECT DISTINCT product.*, <7 correlated subqueries>,
       COUNT(DISTINCT loc.host), COUNT(DISTINCT loc.id)
FROM dojo_product LEFT OUTER JOIN dojo_locationproductreference ...
GROUP BY product.id, product.name, 5, 4, ...
ORDER BY product.name LIMIT 25
```

So the PR fixes the count query and leaves the dominant page query fanning out the same way. An `EXPLAIN (ANALYZE, BUFFERS)` on a dataset like the reporter's would confirm, but the mechanism their plan shows is unchanged by this diff.

### Suggested fix

Keep aggregates off the product queryset entirely and count `LocationProductReference` rows per product with subqueries — `build_count_subquery()` in `dojo/query_utils.py` exists for exactly this trap (see the comment in that file), and `dojo/product_type/ui/views.py` already uses it this way. No join, no `GROUP BY`, and the counts can then live pre- or post-pagination without any of the rest of this diff:

```python
location_refs = LocationProductReference.objects.filter(product_id=OuterRef("pk"))
prods = prods.annotate(
    location_count=Coalesce(build_count_subquery(location_refs, group_field="product_id"), Value(0)),
    location_host_count=Coalesce(
        Subquery(
            location_refs.order_by().values("product_id")
            .annotate(c=Count("location__url__host", distinct=True))
            .order_by("product_id").values("c")[:1],
            output_field=IntegerField(),
        ),
        Value(0),
    ),
)
```

Smaller points, if you rework it:

- `product_list_orders_by_findings_count()` hand-parses the `o` parameter, reimplementing what the `OrderingFilter` in `dojo/product/ui/filters.py` already owns. It becomes unnecessary once the location counts are subqueries.
- `findings_count=F("active_finding_count")` emits the identical `COALESCE((SELECT COUNT(...)))` twice in the SELECT list, and in the sort path `findings_count` ends up annotated twice (no error, since it's a model property rather than a field, but it's confusing). Converging the template on one of the two names would be cleaner.
- The unconditional `.distinct()` costs every request with `V3_FEATURE_LOCATIONS` a `SELECT COUNT(*) FROM (SELECT DISTINCT ...)` for pagination. The duplicate-row risk really comes from the host filter in `dojo/filters.py` (`locations__location__url__host__icontains` can match several references per product; the plain `locations__location=` filter can't, thanks to the `unique_location_and_product` constraint), so dedup belongs in those filter methods rather than on every request.
- The tests are `SimpleTestCase` + `MagicMock` asserting the view's internal call order. They don't touch a database, so they can't demonstrate the fix, and they'll break on any harmless refactor. `assertNumQueries` / `CaptureQueriesContext` asserting that the page query has no `GROUP BY`, plus a query test showing a host filter doesn't duplicate products, is what would actually pin the behaviour.

kocaemre:
Thanks for the detailed SQL analysis — I reworked this in the narrower direction you suggested.

Latest commit now:

- keeps `findings_count` as a correlated subquery annotation for ordering/filtering;
- replaces the v3 location `Count(...)` join aggregates with subquery-backed `location_count` / `location_host_count`, using `build_count_subquery()` for `location_count`;
- removes the previous unconditional Product-list `.distinct()`;
- scopes deduplication to the `endpoints__host` filter path, where multiple matching location references can duplicate a product;
- replaces the mock call-order tests with database-backed regression coverage that checks the annotated queryset counts correctly and does not add a top-level Product `GROUP BY` / location-reference join.

Local verification:

```text
DD_DATABASE_HOST=127.0.0.1 DD_DATABASE_PORT=5432 DD_DATABASE_USER=defectdojo DD_DATABASE_PASSWORD=*** DD_DATABASE_NAME=test_defectdojo DD_TEST_DATABASE_NAME=test_defectdojo .venv/bin/python manage.py test unittests.test_product_list_pagination -v 2 --keepdb
# Ran 2 tests in 3.726s — OK

.venv/bin/python -m ruff check --config ruff.toml dojo/product/ui/views.py dojo/filters.py unittests/test_product_list_pagination.py
# All checks passed!

.venv/bin/python -m py_compile dojo/product/ui/views.py dojo/filters.py unittests/test_product_list_pagination.py
git diff --check
```

The test run still reports the existing local `components/node_modules` staticfiles warning, but the focused tests pass.

valentijnscholten (dismissed):
Thanks for tackling the Product-list query planning. The join→subquery conversion for `location_count`/`location_host_count` looks correct and equivalent to the old counts (`Product.locations` is the reverse relation to `LocationProductReference`, and its manager applies no default filter, so the row sets match). Two things should be addressed before merge, plus a description mismatch.

**1. `dojo/filters.py` — the de-duplication fix is incomplete; products still duplicate on the other location filters (blocking)**

`.distinct()` is added only to `filter_endpoints_host_base` (the `endpoints__host` path). But the Product filter also joins the to-many `locations` relation in two other places that have **no** `.distinct()`:

- `location_status = MultipleChoiceFilter(field_name="locations__status", ...)` (`dojo/product/ui/filters.py:103`)
- the `endpoints` filter → `filter_endpoints_base` (`dojo/filters.py`)

So the stated goal ("location joins do not duplicate products") isn't fully met: a product with two locations whose statuses are both in the selected `location_status` set will appear twice, with wrong pagination totals — when filtering by location status or endpoint id **without** an endpoint host. Please dedupe those paths too (or apply `distinct()` once at the view level after filtering, which would cover all of them).

**2. `dojo/filters.py` — `.distinct()` on the shared helper changes the Finding list too, out of scope and untested (blocking)**

`filter_endpoints_host_base` is shared: it's also called by the Finding filter (`dojo/finding/ui/filters.py:124`), so this change alters the Finding list's `endpoints__host` behavior (now de-duplicates findings). That may be a reasonable fix, but it's an unmentioned change to a different list and isn't covered by the added test (only `test_product_list_pagination`). Please either scope the `distinct()` to the product filter, or explicitly confirm + add Finding-list coverage that pagination/ordering/counts don't shift unexpectedly. (Note: all-column `DISTINCT` won't dedupe if the finding queryset carries a to-many-join annotation at that point, so the effect there needs checking either way.)

**Description mismatch (please reconcile)**

The PR description doesn't match the diff: it says it "defers count annotations until after filtering and pagination," keeps `findings_count` before pagination "only when sorted by that field," and "moves annotations into `prefetch_for_product()`." None of that is in the diff — annotations are still applied before pagination (just extracted into helper functions), there's no sort-conditional gating, and nothing moves into `prefetch_for_product()`. Could you update the description to describe the actual change (join→subquery + a filter-level `distinct()`)? It matters for reviewing whether this fully addresses #15378.

kocaemre:
Resolved the `bugfix` merge conflict in `63ef69ba11`.

Resolution notes:
- kept the PR's subquery-backed `annotate_product_findings_count()` / `annotate_product_location_counts()` helpers;
- adopted current `bugfix`'s `locations_enabled()` feature check instead of the older direct setting access;
- preserved the existing endpoint prefetch gating on `locations_enabled()`.

Local verification:
- `python -m py_compile dojo/product/ui/views.py dojo/filters.py unittests/test_product_list_pagination.py` → passed
- `git diff --check` → passed

I also tried the targeted Django test directly:
- `python manage.py test unittests.test_product_list_pagination.ProductListPaginationAnnotationTests -v2 --keepdb`
- blocked in this local environment before running the test because the active Python is 3.11 and current `bugfix` imports `itertools.batched` from `dojo/finding/helper.py`; `itertools.batched` is only available on Python 3.12+.

kocaemre:
Addressed the two blocking review points in `dfb80b619e`.

Changes:
- removed `.distinct()` from the shared `filter_endpoints_host_base()` helper, so the Finding-list helper path is no longer changed by this PR;
- scoped de-duplication to Product filters only by applying `.distinct()` in `ProductFilter.filter_endpoints_host()` and `ProductFilter.filter_endpoints()`;
- changed Product `location_status` to a method filter that applies `.distinct()` for that to-many location join too;
- expanded regression coverage for Product `endpoints__host`, `endpoints`, and `location_status` filter paths, plus a guard showing the shared helper remains non-deduplicating/out-of-scope;
- updated the PR description to match the actual diff: location join-counts become subqueries, and Product-only location filter paths dedupe.

Local verification:
- `python -m py_compile dojo/product/ui/views.py dojo/product/ui/filters.py dojo/filters.py unittests/test_product_list_pagination.py` → passed
- `/root/.local/share/uv/python/cpython-3.12-linux-x86_64-gnu/bin/python3.12 -m py_compile dojo/product/ui/views.py dojo/product/ui/filters.py dojo/filters.py unittests/test_product_list_pagination.py` → passed
- `git diff --check` → passed

Notes:
- `python -m ruff check --config ruff.toml ...` is blocked by the installed local Ruff not recognizing the repo's current `RUF105` rule selector.
- Direct `python manage.py test ...` is blocked on local Python 3.11 because current `bugfix` imports `itertools.batched`; I verified py_compile with the local uv-managed Python 3.12 interpreter, but that interpreter does not have the Django dependencies installed. CI is the authoritative full test run here.

blakeaowens (approved):

Maffooch (approved):

devGregA (approved):

DarkR0ast:
@kocaemre Ty for the fix!

This PR seems to fix the issue. However I noticed a smiliar behaviour when adding an endpoint to a finding and saving the finding, resulting in a similiar long loading time. Maybe this issue is related to the asset loading time issue?

In the postgres logs this is caused by this SQL Statement (The SQL takes **31316.552 ms** to complete):

```
2026-09-07 14:20:31.984 UTC [430] LOG:  duration: 31316.552 ms  statement: SELECT DISTINCT "dojo_product"."id", "dojo_product"."created", "dojo_product"."updated", "dojo_product"."name", "dojo_product"."description", "dojo_product"."product_manager_id", "dojo_product"."technical_contact_id", "dojo_product"."team_manager_id", "dojo_product"."prod_type_id", "dojo_product"."sla_configuration_id", "dojo_product"."tid", "dojo_product"."prod_numeric_grade", "dojo_product"."business_criticality", "dojo_product"."platform", "dojo_product"."lifecycle", "dojo_product"."origin", "dojo_product"."user_records", "dojo_product"."revenue", "dojo_product"."external_audience", "dojo_product"."internet_accessible", "dojo_product"."enable_product_tag_inheritance", "dojo_product"."enable_simple_risk_acceptance", "dojo_product"."enable_full_risk_acceptance", "dojo_product"."disable_sla_breach_notifications", "dojo_product"."async_updating" FROM "dojo_product" LEFT OUTER JOIN "dojo_locationproductreference" ON ("dojo_product"."id" = "dojo_locationproductreference"."product_id") LEFT OUTER JOIN "dojo_engagement" ON ("dojo_product"."id" = "dojo_engagement"."product_id") LEFT OUTER JOIN "dojo_test" ON ("dojo_engagement"."id" = "dojo_test"."engagement_id") LEFT OUTER JOIN "dojo_finding" ON ("dojo_test"."id" = "dojo_finding"."test_id") LEFT OUTER JOIN "dojo_locationfindingreference" ON ("dojo_finding"."id" = "dojo_locationfindingreference"."finding_id") WHERE ("dojo_locationproductreference"."location_id" = 41627 OR "dojo_locationfindingreference"."location_id" = 41627) ORDER BY "dojo_product"."id" ASC
```

PR #15476, opened 2026-08-02
docs(triage): document reachability scoring

devGregA:
## Description

Adds documentation for **reachability** in DefectDojo Pro, which currently has
no user-facing page. It sits in Scoring & Prioritization alongside the CVSS and
EPSS/KEV pages, since reachability is another input to a Finding's priority.

The page covers:

- **The five normalized verdicts** — reachable (runtime), reachable (static),
  potentially reachable, unreachable, unknown — and why normalizing matters
  when different tools use different wording for different things.
- **The rules the feature follows**, which users need in order to trust it:
  unknown never counts against a Finding, unreachable lowers priority but never
  closes a Finding, every verdict shows its source, and verdicts follow
  deduplication.
- **Where verdicts come from**, and that partial coverage is normal and
  expected rather than a misconfiguration.
- **How it affects priority**, including the per-engine scalar (set it to `0`
  to record verdicts without letting them move scores), previewing with the
  simulator, and the reminder to review risk thresholds after enabling it since
  scores shift.
- **What the feature looks like** on a Finding, the Findings list, and an asset.

The feature is beta and off by default; the page says so up front and names
where a superuser enables it.

## Notes

- Text-only, no images.
- Documents only shipped behavior. Capabilities that are not fully released
  are deliberately left out rather than described ahead of availability.

### DIFF
--- a/docs/content/triage_findings/finding_scoring/reachability.md
+++ b/docs/content/triage_findings/finding_scoring/reachability.md
@@ -1,0 +1,122 @@
+---
+title: "Reachability"
+description: "How DefectDojo Pro records whether a Finding's vulnerable code is actually reachable, and how that verdict adjusts priority"
+audience: pro
+weight: 3
+---
+
+A Critical CVE in code your application never calls is not the same risk as
+the same CVE on a live request path. **Reachability** captures that
+difference: DefectDojo Pro records whether each Finding's vulnerable code can
+actually be reached, shows you where that conclusion came from, and feeds it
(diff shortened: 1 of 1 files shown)


devGregA:
Pairs with the Pro change that adds the reachable risk floor and unreachable risk ceiling. Please hold this until that lands so the docs do not describe settings that are not yet available.

blakeaowens (approved):

Maffooch (approved):

PR #15482, opened 2026-08-03
feat(parsers): add fifty-six file parsers for vendors with no importer

devGregA:
Adds fifty-six file parsers for vendors that have no file importer today. These let a user who cannot grant API credentials — air-gapped network, procurement hold, pending security review — upload the same data as a file.

Each scan type is byte-identical to what the corresponding DefectDojo Pro connector reports, so someone who uploads an export *and* later enables the connector gets one set of findings that deduplicate rather than two copies of everything. Every parser has a test asserting its scan type verbatim, and the deduplication settings are copied from each connector's own configuration rather than chosen.

**One parser per vendor, one parser per scan type.** A parser is scoped to a single uploaded run; a connector syncs an account's ongoing state. Where DefectDojo already ships a parser for a vendor, no second one is added here. Where a connector reports *several* scan types from one export — Fleet and Elastic Security — there is one parser per scan type, following the shipped `anchorectl_vulns` / `anchorectl_policies` split, because each scan type has its own deduplication key.

| Parser | Scan type | Shape |
| --- | --- | --- |
| Socket | `Socket - Connectors Import` | supply-chain alerts |
| Lacework | `Lacework - Connectors Import` | container **and** host vulns, mapped differently |
| CrowdStrike Falcon Spotlight | `CrowdStrike:Spotlight - Connectors Import` | host vulns |
| FOSSA | `FOSSA - Connectors Import` | SCA **and** licence/quality issues |
| Endor Labs | `Endor Labs - Connectors Import` | reachability-aware SCA |
| GitGuardian | `GitGuardian - Connectors Import` | secret incidents |
| Codacy | `Codacy - Connectors Import` | SCA, container and DAST via one endpoint |
| DeepSource | `DeepSource - Connectors Import` | analysis issues **and** dependency advisories (GraphQL) |
| Probely | `Probely API Import` | DAST |
| Detectify | `Detectify Scan` | EASM / DAST |
| Beagle Security | `Beagle Security - Connectors Import` | DAST, one finding per occurrence |
| HackerOne | `HackerOne - Connectors Import` | bug-bounty reports (JSON:API) |
| YesWeHack | `YesWeHack - Connectors Import` | bug-bounty reports |
| Intigriti | `Intigriti - Connectors Import` | bug-bounty submissions |
| Quay | `Quay - Connectors Import` | container image scan (Clair) |
| Google Cloud SCC | `Google Cloud SCC - Connectors Import` | cloud findings |
| Fairwinds Insights | `Fairwinds Insights - Connectors Import` | Kubernetes action items |
| AccuKnox | `AccuKnox - Connectors Import` | CNAPP findings |
| Halo Security | `Halo Security - Connectors Import` | external attack surface |
| Nightfall AI | `Nightfall AI - Connectors Import` | DLP violations in SaaS apps |
| Fleet | `Fleet:Vulnerabilities - Connectors Import` | software CVEs per host |
| Fleet | `Fleet:Policies - Connectors Import` | failing compliance policies |
| Elastic Security | `Elastic Security:CNVM - Connectors Import` | workload CVEs |
| Elastic Security | `Elastic Security:Posture - Connectors Import` | failing benchmark rules |
| Elastic Security | `Elastic Security:Detections - Connectors Import` | detection-engine alerts |
| Action1 | `Action1 Scan` | endpoint patch/vulnerability management |
| Datadog Cloud Security | `Datadog Cloud Security` | cloud findings of every kind, one endpoint |
| Escape | `Escape - Connectors Import` | API-security DAST |
| Rapid7 InsightAppSec | `Rapid7 InsightAppSec - Connectors Import` | DAST |
| Intruder | `Intruder API Import` | external vulnerability scanning |
| NowSecure | `NowSecure` | mobile app, static **and** dynamic |
| Vanta | `Vanta Compliance` | compliance controls |
| Wallarm | `Wallarm API Security` | API security |
| Bright Security | `Bright - Connectors Import` | DAST |
| Microsoft Defender for Cloud | `Microsoft Defender for Cloud - Connectors Import` | Azure server **and** container-registry vulns |
| Akto | `Akto Scan` | API security |
| Holm Security | `Holm Security Scan` | network **and** web scanning |
| Klocwork | `Klocwork Scan` | SAST, **NDJSON** input |
| Qwiet AI | `Qwiet Scan` | SAST + reachability-aware SCA |
| Automox | `Automox Scan` | endpoint patch management, patches joined to devices |
| BigID | `BigID Scan` | DSPM cases — counts only, never the data |
| Calico Cloud | `Calico Cloud Image Assurance Scan` | container image scanning |
| Dragos | `Dragos Scan` | OT/ICS asset vulnerabilities |
| Finite State | `Finite State - Connectors Import` | firmware/binary analysis with **VEX** status |
| 42Crunch | `42Crunch - Connectors Import` | API security: OpenAPI audit **and** conformance scan |
| HiddenLayer | `HiddenLayer Model Scan` | ML model scanning (SARIF) |
| NetRise | `NetRise Scan` | firmware analysis (GraphQL Relay) |
| Nozomi Networks | `Nozomi Vantage Scan` | OT/ICS, denormalised per-asset CVEs |
| Ostorlab | `Ostorlab Scan` | mobile **and** web, static or dynamic per scan |
| Parasoft DTP | `Parasoft DTP Scan` | SAST |
| Uptycs | `Uptycs Scan` | one row fans out per CVE |
| CyberArk Certificate Manager | `CyberArk Certificate Manager Scan` | certificate posture, **computed** |
| ManageEngine VMP | `ManageEngine Vulnerability Manager Plus Scan` | patch/vulnerability management |
| Zimperium zScan | `Zimperium zScan` | mobile app scanning (shared SARIF mapping) |
| Group-IB ASM | `Group-IB ASM - Connectors Import` | external attack surface |
| SOOS | `SOOS - Connectors Import` | SCA, SAST, container, SBOM **and** DAST in one API |

**Eight scan types do not follow the `<Vendor> - Connectors Import` pattern** — `Probely API Import`, `Detectify Scan`, `Action1 Scan`, `Datadog Cloud Security`, `Intruder API Import`, `NowSecure`, `Vanta Compliance` and `Wallarm API Security` — so they cannot be derived from the vendor name and had to be copied. Tests assert the derived form is *not* claimed.

Non-obvious dedup config, all copied not chosen: GitGuardian, Probely, HackerOne, YesWeHack, Intigriti, Google Cloud SCC and **Intruder** use the plain `hash_code` algorithm where the others use `unique_id_from_tool_or_hash_code`; **Rapid7 InsightAppSec hashes `unique_id_from_tool` and nothing else** — no title, no severity — because its vulnerability id is stable and a volatile field would split a regraded finding; **Action1 includes `component_version`**, so a partly-patched machine does not merge with an unpatched one; CrowdStrike lists `unique_id_from_tool` among its hash fields *as well as* pairing with that algorithm; Probely, Halo Security and Beagle Security hash `endpoints`, so those parsers must populate the endpoint or the hash covers nothing.

### Three scan types register no dedup config, deliberately

`Finite State - Connectors Import`, `42Crunch - Connectors Import` and `SOOS - Connectors Import` have
**no hashcode field list and no algorithm entry** to copy. "Copied, not chosen" means copying nothing: both use DefectDojo's
default algorithm, which is exactly what those connectors' own findings do today. Choosing hashcode
fields here would change how the connector's findings deduplicate as well, which is not this PR's
call to make — so none of the three defines `get_dedupe_fields`, and tests assert that.

### Exports that need two API calls

Eight vendors split a finding across two endpoints, and a single-call export produces findings missing their most useful fields — or nothing at all. Each parser accepts the second call keyed by id **or** nested on the row, and the docs page says so:

- **Halo Security** — the list row carries the issue, target and status; the description, category, CVEs and PCI flag exist only on a per-issue detail.
- **Nightfall AI** — the violation carries the policy and state; the detections carry the redacted evidence, the confidence and the API-key verdict. The verdict is what raises a violation to Critical, so a violation-only import loses the severity too.
- **DeepSource** — no REST API at all; the file is a saved GraphQL response.
- **Beagle Security** — the report is a JSON string nested inside the API envelope, so both the envelope and the report body are accepted.
- **Action1** — the vulnerability catalogue is separate from the machines each CVE affects, and **the affected endpoint is the finding**: a catalogue entry nothing is running produces nothing. A third call supplies each machine's OS.
- **Intruder** — an issue is the weakness; its occurrences are the targets. Intruder's own issue object carries `occurrences` as a **URL string**, and a test asserts the unexpanded link is not mistaken for data.
- **Vanta** — a finding is a *(test, failing entity)* pair. A test with no failing entity is the control working.
- **Rapid7 InsightAppSec** — a vulnerability is named only by its attack-module *id*; the readable name and prose come from a module-metadata call, without which every finding is titled `InsightAppSec finding`.

### Input shape provenance — worth a sanity-check

Each parser reads the vendor's **wire shape**, taken from the connector's `api`/`client` layer rather than its internal Go structs, so the input is what the vendor's API actually returns.

**Two places where the file contract is not simply the vendor's response:**

- **FOSSA** — one finding per `issue.projects[].id`, to reproduce the connector's per-project tool id. The connector reads the project locator from a product attribute the API response does not carry. Documented on the FOSSA docs page; if anything here looks wrong, this is the part most likely to need changing.
- **Beagle Security** — the connector's unique id is a SHA-256 over the **application token**, which is a parameter of every Beagle API call and is not part of a report. An export carrying the token gets connector-identical ids; otherwise no unique id is set at all, and the hash over title/severity/endpoints is what matches a file import to an API sync. Inventing a token-less id would produce something that deduplicates against nothing.

### No live parity proof yet

These are built against the connector converters and their semantics, **not** validated against a real export from a live tenant. The tests prove each parser matches the converter it mirrors; they do not prove the input shape is obtainable in practice. That check is still outstanding and is the thing most likely to surface a problem.

It already caught one during review: **DeepSource has no REST API at all**, and an earlier draft of this PR read an invented flat `{"occurrences": [...]}` envelope that no DeepSource user could produce. It now reads the real saved GraphQL response (`data.repository`, `edges[].node`), and a test pins that shape. Only the wrapper was wrong — every field mapping already used DeepSource's camelCase names, and 26 of 27 existing tests passed unchanged against the corrected envelope.

Two vendors do not publish the schema at all, and both parsers say so out loud rather than pretending otherwise:

- **Beagle Security** documents its report-level keys and its occurrence block but not the names of the per-finding fields, and its only sample report omits the finding array. The parser reads every field through the same alias table the connector uses, and locates the finding array by name or, failing that, by shape — the first key, in sorted order, whose value is an array of objects.
- **AccuKnox** returns findings whose column names vary by scan type, so the parser probes a candidate list per field, including a `vulnerability__`-prefixed variant.

### Details a reviewer may want to check

- **Probely's severity is an integer** — only `10`/`20`/`30` exist, mapping to Low/Medium/High, and there is no Critical. It is neither a score nor an index.
- **Halo Security's severity is also an integer, but 5 is the highest** — the inverse of a priority number.
- **Fairwinds grades on a 0.0–1.0 float**, with breakpoints at 0.9/0.7/0.4/0.1, and its JSON keys are PascalCase.
- **Fleet grades an unscored CVE as Medium, not Info** — it enriches from the NVD, so a missing score means "not scored yet" rather than "no risk". An explicit zero is still Info, and `null` and `""` mean different things in its own decoder.
- **Elastic's severity label wins over its CVSS score** for CNVM, and an *unrecognised* label falls back to the score. For posture and detections there is no score to fall back on, so an unrecognised label is Medium — Info would understate a failing benchmark rule.
- **Lacework reports `fix_available` as an integer on container rows and a string on host rows.** Testing one shape's type against the other would silently drop every host mitigation.
- **Socket grades alerts `low`/`middle`/`high`/`critical`** — `middle`, not `medium`. Guessing would drop every middle-severity alert to Info.
- **DeepSource needs two severity ladders.** It grades everything `CRITICAL`/`MAJOR`/`MINOR` whatever the issue is, so the category decides: a security issue keeps its grade, a bug-risk issue drops a step. A `secrets`-analyzer hit is Critical regardless.
- **Codacy's vulnerable package is the *last* entry of the dependency chain**, not the first — the first is the project itself.
- **Quay's scanner is Clair**, so vulnerabilities nest under the *features* they affect with capitalised keys, `Defcon1` grades above Critical, and the identity is the advisory id concatenated with the feature name **with no separator** — reproduced exactly, since adding one would break the merge.
- **Google Cloud SCC puts the finding and its resource side by side** in `listFindingsResults`, not nested, so reading the resource off the finding yields nothing.
- **Endor Labs and Intigriti flatten HTML rather than rendering it** — upstream advisory text and researcher prose respectively. Go's `html.EscapeString` is matched byte for byte rather than using Python's `html.escape`, which spells the apostrophe entity differently.
- **HackerOne's API is JSON:API**, so severity, weakness and reporter are *relationships* at `relationships.<name>.data.attributes`. Reading them off the top level leaves every finding at Info with no CWE and no reporter — silently, with no error. A test asserts the raw fixture genuinely does not carry them anywhere else.
- **Status handling is per-vendor and deliberate.** Probely skips fixed/invalid/accepted but keeps `retesting`. Detectify skips patched/false-positive but keeps `accepted_risk` and flags it. Codacy flags a false-positive-ignored item `false_p` but treats "acceptable risk" as a real accepted finding. YesWeHack maps eight workflow states, and an unrecognised one stays **active** so a future state cannot silently close a finding. Intigriti's terse rejection reason is literally `no`, and its top severity tier is `Exceptional`, not `Critical`. Nightfall's `PENDING` is active but *not* verified, and `EXPIRED` is out of scope rather than mitigated — it can no longer see the resource, so it can confirm neither that the data is gone nor that it is still there.
- **Only failing results are findings** for Fleet policies (`response: fail`) and Elastic posture (`result.evaluation: failed`). Both vendors emit a document per evaluation, passed or failed, and both parsers ignore the other kind of document in the same file — a test in each asserts that.
- **Neither static nor dynamic:** CrowdStrike Spotlight and Elastic detections are marked as neither, which is what the connectors do. Spotlight reads the agent's software inventory; a detection is observed activity that no test found. Elastic detections also carry a triage instruction instead of a mitigation, so a triage queue is not read as a remediation backlog.
- **Nightfall imports no sensitive data.** Its API returns redacted detection text only; that is the one text field read from a detection, and a test asserts a hand-added raw value would not reach a finding.
- **Tags are sorted and deduplicated** wherever the connector does it (Fleet, Elastic), because a tag set differing only in order reads as a change on every reimport. Datadog deliberately does **not** sort — it preserves build order — and both behaviours are asserted.
- **Action1's `score` is a word, not a number** (`Critical`/`High`/…), used when `base_severity` is absent. Treating it as numeric would drop every such finding to Info.
- **Wallarm's `threat` is a number *or* a word in the same field**, depending on which API answered, so both ladders are needed; 5 is the most severe. That numeric direction is the one thing Wallarm does not document — copied from the connector and worth confirming live.
- **Datadog's `base_severity` is deliberately ignored**: it is the rule's default before Datadog adjusts for the environment, and the adjusted value is the one worth importing. Its dates are unix **milliseconds**.
- **Intruder records triage by snoozing**, with a reason: false positive and the two risk-acceptance reasons map to flags, and an unrecognised reason leaves the finding inactive with **neither** rather than guessing at the reviewer's decision.
- **NowSecure and Wallarm sort their vulnerability identifiers** and dedupe case-insensitively, where Datadog and Intruder preserve document order. Same shared extractor, two different call sites — asserted per parser rather than assumed.
- **Vanta has no severity scale at all**, so every failing entity is Medium. That makes `component_name` load-bearing: it is the failing resource, and without it two resources failing one control would merge.
- **Defender for Cloud describes one finding two ways**: a container-registry finding puts the vulnerable package under `softwareDetails`, a server finding flattens it into `softwareName`. Reading one shape would leave every finding of the other kind with no component. Its CVE ids are also matched **anchored**, so a reference title like `supersedes CVE-2000-0009` does not contribute another finding's identifier, and an *unfamiliar* resource type is judged by whether it carries a CVE — so a new Defender scanner is not dropped silently.
- **Klocwork's export is NDJSON**, one issue per line — a whole-document JSON parse fails at the second line. Its trailing *summary* line is skipped by key rather than parsed, and a search that matched nothing answers with that line **alone**, which is an empty result rather than a malformed file. Its severity **code** is also the inverse of a score: 1 is Critical.
- **Qwiet keeps the CVE, package URL, CVSS score, CWE and reachability in a list of key/value tag objects**, not fields — looking for fields of those names finds nothing. A dependency finding with `related_findings` is reachable even with no reachability tag, because those findings *are* the traced path.
- **Holm Security's severity name wins and its numeric level is only the fallback** (4 is the most severe), so an unfamiliar name does not become Info while a usable level sits beside it. Its scan *class* — net or web — decides static versus dynamic and is a property of the scan rather than the row, so an export states it.
- **Akto's `apiUrl` is often a bare path**, which is not an endpoint; the connector does not invent a host, and neither does this parser. Its `IGNORED` status is a false positive while `FIXED` is merely inactive.
- **Bright reports one issue against several resources** when the same weakness is reachable from more than one URL, so with no entry point every resource becomes an endpoint.
- **InsightAppSec's evidence is flattened, not rendered** — it is the application's own response to an attack payload. A consequence worth knowing: a payload that is *only* markup flattens to nothing, leaving an empty `Attack value` line. Mirrored rather than corrected.

### Three things CI caught that local tests had been hiding

**1. An endpoint host that the model rejects.** `Endpoint.clean()` accepts a host of letters, digits, dot, hyphen, underscore or plus — or an IP address — and **raises** otherwise, which fails the whole import rather than the one finding. Codacy tripped it: a container item's `affectedTargets` is an image reference, so `registry.example.com/generic-app` went into the host field. It is now parsed into host + path, which is what DefectDojo does with the endpoint string the connector sends. Halo Security and YesWeHack were hand-splitting a URL in a way that left `host:port` in the host field, and both now parse instead.

Every parser that records a host now checks it first and drops the endpoint rather than raising, because these values are free text — a Fleet display name is often `Someone's MacBook`, and an Elastic cloud resource name can be a path. The value still appears in the description, so nothing is lost. Nine tests pin it, including IPv6 hosts, which the pattern rejects but the IP check accepts.

**2. An assertion that could only ever pass one CI leg.** An unset location path is `""` on the URL location model (`CharField(blank=True)`) and `None` on `Endpoint`, so one Detectify test asserting `assertIsNone(path)` passed the `V3_FEATURE_LOCATIONS=False` leg and failed the `True` one. It now asserts falsiness. A sweep of every test file found exactly this one occurrence.

**3. "Neither flag set" is not neutral.** DefectDojo defaults `static_finding` to false and `dynamic_finding` to **true**. NowSecure leaves both alone when it does not recognise an analysis type — correct, because the connector's own findings behave identically once they reach DefectDojo — but the test asserted two falses and the docs page called it "neither flag set". Both now state what actually happens. Forcing two falses would have been the wrong fix: it would make a file import and an API sync disagree about the same finding.

All three had one root cause: the local harness was more forgiving than the Django models. It coerced `""` to `None`, stubbed every boolean as `None`, and its `clean()` stand-in validated nothing. It now mirrors the real field defaults and the real validation, reproduced each CI failure exactly, and surfaced no others across all thirty-three parsers.

### Escalated, not built

- **Docker Scout.** Its connector reads Prometheus metrics (`scout_stream_vulnerabilities`, `scout_policy_evaluated_images`) and emits count summaries; the connector's own description notes that Docker Scout does not expose individual CVEs through its metrics API. A file parser would either import counters as findings or invent per-CVE detail that no export contains, and the metrics are whole-organisation aggregates rather than one run. Flagging rather than resolving the ambiguity by building.

- **Akamai (`AkamaiManagementV2`).** Built far enough to be sure, then stopped, for four reasons that
  compound. Its one scan type covers **three unrelated record shapes** — findings, incidents and
  application issues — each with its own converter and field set, where every other vendor here has one.
  The application-issues path needs a **second per-issue detail call** joined per issue. Its findings
  carry **no `unique_id_from_tool` at all** (the shared builder's `WithUniqueID` is never called) and the
  scan type has no dedup config, so imported findings would match on the legacy hash of title,
  description and severity alone. And that description is built by sections that are **each gated on
  `len(OwaspTags) == 0`** — so a finding's Source and Evidence blocks vanish when it happens to carry no
  OWASP tag, which makes the only available hash input unstable.

  Separately, `UnsavedVulnerabilityIDs()` runs the **CWE** extractor over the description, so CWE
  identifiers land in the vulnerability-ids field rather than CVEs. Mirroring that faithfully would mean
  shipping a parser that files CWEs as CVEs. Those two look like connector defects rather than
  intentional mappings, and they are worth fixing on the connector side first — a parser written to
  mirror them would have to be rewritten immediately afterwards. Flagging rather than resolving the
  ambiguity by building.

### Already covered, deliberately skipped

- **Mend** — DefectDojo already ships a `Mend Scan` parser, so no second parser is added even though the connector reports its own scan type. Same reasoning as the other already-covered vendors (GitHub, Prowler, Sonatype, Cloudflare, Orca, Coverity, Sysdig, Snyk, Veracode, Qualys, Wazuh, Nuclei, OpenVAS, Contrast, Black Duck, Edgescan).

### Follow-ups worth raising against the connectors, not fixed here

- **DeepSource** grades `STYLE`, `DOCUMENTATION` and `COVERAGE` issues as Info and imports them. Mirrored rather than filtered, because parity is what stops findings duplicating — but those are not weaknesses and are probably worth excluding on both sides.
- **Codacy and DeepSource** date a finding *today* when a timestamp will not parse. Mirrored; a reviewer may prefer leaving the date unset.
- **Fleet** reports an EPSS probability, and the connector renders it into the description rather than into the `epss_score` field. Mirrored for parity; both sides would be better off populating the field.
- **Action1 and NowSecure set the CVSS score unconditionally**, so an unscored finding lands as `0.0` rather than unset. Mirrored; a reviewer may prefer leaving it null.
- **Wallarm's mitigation is an exploit example**, not remediation advice — the only remediation-shaped field it has. Useful to a reviewer, but it is a reproduction rather than a fix.
- **InsightAppSec's sanitiser drops markup-only attack payloads** (above). Both sides would be better off escaping the payload instead of flattening it away.
- **Finite State writes a stable `unique_id_from_tool` but registers no `unique_id_from_tool_or_hash_code` algorithm**, so that id does nothing for deduplication today. Left alone here for the reason above; worth registering on the connector side.
- **42Crunch prefixes every identity with the API id**, which a downloaded report does not contain. The parser accepts a wrapper supplying it, and the docs say findings will not deduplicate against synced ones without it — an export that embedded the id would remove the manual step.
- **Automox's `tags()` comment says the severity is tagged; its code does not tag it.** The code is mirrored, not the comment.
- **Calico Cloud, Dragos and Automox set the CVSS score unconditionally**, so an unscored finding lands as `0.0` rather than unset — the same pattern already noted for Action1 and NowSecure.

- **SOOS treats an unrecognised scan type as dynamic**, because the connector reads its static/dynamic lookup table with a Go map access, which yields false for a missing key exactly as it does for the `dast` entry. Mirrored so both paths agree; a new SOOS scan type would arrive as dynamic.
- **CyberArk Certificate Manager renders subject alternative names in a nondeterministic order** — the connector iterates a Go map. The parser sorts them, which is the one deliberate divergence in this PR; the line is not in the deduplication hash, so the two still match.

### Outside the parser directories

Only `HASHCODE_FIELDS_PER_SCANNER` and `DEDUPLICATION_ALGORITHM_PER_PARSER` in `settings.dist.py`, both additive, for fifty-three of the fifty-six scan types (the three above register nothing). Nothing else — `ruff.toml` is unchanged.

### Validation

- **1,450 tests**, green locally with `V3_FEATURE_LOCATIONS` both on and off.
- Every registration is checked against the `dojo/checks.py` dedup-configuration check that arrived with the `dev` merge: no `dojo.E001`, no `dojo.W001`, no split or orphaned registration, and no parser reporting a scan type nobody registered.
- CI has been green across all four Rest Framework Unit Test legs (amd64 and arm64 × `V3_FEATURE_LOCATIONS` true and false); each new batch is re-run before the next is added. Two mode-dependent model behaviours were found this way and are now covered: the locations model fills an unset port in from the protocol (an https URL is port 443 there, while `Endpoint` leaves it `None`), and it stores unset string parts as `""` rather than `None`.
- `ruff check .` clean without modifying `ruff.toml`.
- Fixtures are built from each vendor's documented schema with generic hosts, package names, private-range addresses and placeholder identifiers. No fixture carries a real CVE, a credential-shaped value, or anything host- or account-specific — a scripted audit of the whole diff confirms it.
- Every finding that carries a host or URL records it through the locations idiom, so it survives under either value of `V3_FEATURE_LOCATIONS` rather than being silently dropped under the default.
- The three Elastic parsers share their document walk and ECS asset rendering via the CNVM module, the way the shipped `invicti` parser extends `netsparker`. A test confirms the shared helper class is not itself registered as a parser.


devGregA:
C:/Program Files/Git/test all

devGregA:
C:/Program Files/Git/test all

blakeaowens (approved):

Maffooch (approved):

PR #15486, opened 2026-08-03
fix(tags): order tag-count decrements so concurrent removals cannot deadlock

Maffooch:
**Description**

Two concurrent operations that remove tags from an overlapping tag set could deadlock, aborting a cascade delete partway through:

```
django.db.utils.OperationalError: deadlock detected
DETAIL:  Process A waits for ShareLock on transaction N; blocked by process B.
         Process B waits for ShareLock on transaction M; blocked by process A.
CONTEXT:  while updating tuple (x,y) in relation "dojo_tagulous_product_tags"
```

**Root cause.** `dojo/tags/utils.py` `bulk_remove_all_tags` decrements tag counts with one `UPDATE` per tag row, iterating the aggregate queryset in whatever order it comes back:

```python
affected_tags = through_model.objects.filter(...).values(target_field_name).annotate(num=Count("id"))
for entry in affected_tags:
    tag_model.objects.filter(pk=entry[target_field_name]).update(count=F("count") - entry["num"])
```

Each `UPDATE` takes a row lock, so the iteration order *is* the lock order. Two callers whose tag sets overlap can acquire those locks in opposite orders, and Postgres breaks the cycle by aborting one of them. Ordering by tag id makes the lock sequence identical for every caller, so the cycle cannot form. The ordering also pins the `GROUP BY` to that one column, which a model's default `Meta.ordering` would otherwise silently widen.

**Backstop.** `async_delete_task` also retries when it loses a concurrency race, for the conflicts ordering cannot rule out — a delete overlapping an import, for instance. Classified by SQLSTATE read from the driver exception Django keeps as `__cause__` (`40P01` deadlock_detected, `40001` serialization_failure), not by message text. A statement timeout (`57014`) is not retried, since that would only repeat an already too-slow delete. After 3 retries the original error is raised, so a persistent problem still surfaces. Retries back off exponentially and are offset by the object pk so tasks that collided don't collide again.

The retry is deliberately skipped under eager execution. Verified against celery 5.6.3 that `retry()` under `apply()` re-invokes the body immediately and discards `countdown` — three invocations in 0.04s with `countdown=5` — so for `block_execution` callers it would re-run the whole delete inside the request while the conflicting transaction may still be open. Those callers get the error instead.

**Resumability is partial, and the code now says so.** Steps 1–4 of the delete refetch or re-filter what they remove, so a re-invocation resumes from what's left. Steps 5–6 are not resumable: once the top-level `obj.delete()` commits, a re-invocation finds nothing and returns early, so post-delete work (product grading) is skipped rather than redone. That case logs at WARNING, because a product grade may be left stale.

No new setting — `ASYNC_DELETE_MAX_CONFLICT_RETRIES` is a module constant and `settings.dist.py` is unchanged.

**Test results**

`unittests/test_tag_utils_bulk.py` — `BulkRemoveAllTagsLockOrderTest` asserts the decrements are issued in ascending tag-id order. Tags are attached in an order unrelated to their ids so "whatever the aggregate returns" and "ascending id" cannot coincide by luck. Without the ordering it fails, locking rows `[3, 1, 2]` against tag ids `[1, 2, 3]`.

`unittests/test_async_delete.py` — `TestAsyncDeleteTransientConflictRetry` asserts behaviour rather than mock arguments:

- a deadlocked attempt followed by a real re-invocation leaves the object deleted;
- eager execution surfaces the error and does not retry;
- the post-delete skip on a retry is logged at WARNING, and a first-call miss stays at INFO;
- deadlock and serialization failure are retryable; statement timeout and a SQLSTATE-less `OperationalError` are not;
- a repeating deadlock surfaces the original error once retries are exhausted, and the backoff grows across attempts.

```
test_async_delete, test_tag_utils_bulk, test_cascade_delete,
test_bulk_delete_findings_m2m, test_prepare_duplicates_for_delete,
test_utils, test_delete_with_endpoints_v3          118 tests   OK
```

`ruff check` clean on all changed files (0.15.20, repo `ruff.toml`).

**Documentation**

No documentation change: no user-visible behaviour change and no new setting.

**Downstream check**

`ASYNC_OBJECT_DELETE` is enabled in the Pro plugin, so this path is the one that runs there. Nothing overrides or subclasses `async_delete_task`; the Pro references to it are a Celery argument profile and a test that calls it positionally, and `bind=True` does not change how callers invoke it. `dojo_dispatch_task` does not introspect the task signature. The Pro `post_delete` receiver that reconciles orphaned connector mappings still fires exactly once per object.


valentijnscholten (commented):
Thanks for digging into this. I think the retry wrapper is worth keeping — a deadlock is always possible under concurrency and absorbing it beats failing a delete half-way through. But I'd like us to also fix the lock ordering that produces it, and a few things in the description don't match what the code does.

### The description's premise doesn't hold

> Bulk object deletion dispatches several `async_delete_task` instances that run concurrently.

There's no fan-out anywhere in this path:

- `dojo/finding/helper.py:907-923` — `_bulk_delete_findings_internal` iterates chunks in a plain `for` loop, one `transaction.atomic()` per chunk, strictly sequential. Chunked execution was deliberately kept serial to avoid exactly this class of complexity.
- `dojo/utils.py:1933` — `async_delete.delete()` dispatches exactly one `async_delete_task` per object.
- Every caller deletes a single object per request: `dojo/product/ui/views.py:1073`, `dojo/engagement/ui/views.py:358`, `dojo/test/ui/views.py:289`, `dojo/product_type/ui/views.py:141`, plus the API `destroy()` methods. None of them loop.

So two of these can only overlap via two separate concurrent requests, or a delete overlapping an import. Could you link the actual traceback / report, and say what the worker concurrency was? It matters for the fix: if the partner transaction is an import rather than a second delete, the retry needs to survive a much longer-running conflicting transaction than a 5s backoff assumes.

### The root cause looks deterministically fixable

Following `dojo_tagulous_product_tags` from the traceback: `dojo/utils_cascade_delete.py:195` calls `bulk_remove_all_tags`, and `dojo/tags/utils.py:500-512` decrements the counts one row at a time in unordered aggregation order:

```python
affected_tags = through_model.objects.filter(...).values(target_field_name).annotate(num=Count("id"))
for entry in affected_tags:
    ... .update(count=models.F("count") - entry["num"])
```

`bulk_add_tags` (`dojo/tags/utils.py:421-425`) increments through a single `UPDATE … CASE WHEN`, which takes its locks in scan order. Two transactions touching an overlapping tag set in different row orders is the textbook row-order deadlock, and it needs only one delete plus one concurrent import — not a bulk delete.

An `.order_by("id")` on `affected_tags` (and an ordered lock acquisition on the add side) makes the lock order deterministic and removes this deadlock rather than absorbing it. Worth noting the comment right above that loop: *"Tag counts are not used in DefectDojo but we maintain them to avoid breaking tagulous's internal bookkeeping"* — the contended writes are to a column nothing reads, so there's room to make them cheaper as well.

I'd like to see that land alongside the retry, and the PR description say that the retry is the backstop rather than the fix.

### Issues in the retry itself

1. **The grading contention named in the description isn't in this task.** `dojo/utils.py:1176-1181` — `perform_product_grading` *dispatches* `calculate_grade` rather than running it. When `ASYNC_OBJECT_DELETE` is on, a grading deadlock surfaces inside `calculate_grade`, which this PR doesn't wrap. So "tag bookkeeping and product grading are common to every object in the batch" overstates what the wrapper covers — only the tag half is in scope.

2. **Retry can't recover anything after `obj.delete()`.** The new docstring says "Every step refetches or re-filters what it deletes, so calling this again after a partial failure resumes from whatever is left." That's true for steps 1-4 only. A conflict in step 6 re-enters, hits `obj is None`, logs `already gone, nothing to do` at INFO and returns — the product grade is never recalculated, silently. Either move the post-delete work somewhere the retry can reach, or say in the docstring that only the pre-delete steps are resumable and log the skip at WARNING.

3. **The backoff likely doesn't apply on the synchronous path.** `dojo/celery_dispatch.py:100` falls through to `sig.apply()` when the user has `block_execution`. In eager mode Celery re-applies the signature inline and ignores `countdown`, so those users get up to three immediate re-runs of the whole delete inside the request — precisely when the winning transaction hasn't committed yet. Please verify against celery 5.6.3; if it holds, skip the retry entirely when `self.request.is_eager`.

4. **The tests don't exercise the retry.** All five patch `async_delete_task.retry` with `side_effect=Retry()`, so `test_retry_delay_grows_with_each_attempt` asserts the value of an argument handed to a mock, not that any delay or re-execution occurs. The classification tests are genuinely useful; the retry-mechanics ones would be more convincing driving the real `retry` through `apply()` and asserting the task body ran twice.

5. **Does this need a new setting?** `max_retries=3` on the task decorator, or a module constant next to `ASYNC_DELETE_RETRY_DELAY`, gets the same result without another entry in `settings.dist.py` — and avoids inheriting the `DD_ASYNC_OBEJECT_...` typo's inconsistent neighbour spelling.

6. **Placement.** `is_transient_db_conflict` is generic DB plumbing and `dojo/utils.py` is the module we're actively draining; `dojo/db_utils.py` or similar would be a better home.

The SQLSTATE-over-message-text classification is right, and `psycopg[c]==3.3.4` means `.sqlstate` is there, so that part is sound.

### Branch

This is a `fix(...)` based on `dev`, so it won't ship until the next minor. If it's chasing a deadlock people are hitting now, it wants to be on `bugfix`.

Maffooch:
Thanks — this was a genuinely useful review. Your root-cause analysis was right and it's now the headline of the PR; the retry is demoted to the backstop. Pushed in f18219941a.

### The lock ordering is fixed

`bulk_remove_all_tags` now orders the decrements by tag id, so the row-lock sequence is identical for every caller and the cycle cannot form. `BulkRemoveAllTagsLockOrderTest` attaches tags in an order unrelated to their ids so that "whatever the aggregate returns" and "ascending id" can't coincide by luck. Without the ordering it fails concretely:

```
AssertionError: Lists differ: [3, 1, 2] != [1, 2, 3]
 : tag rows must be locked in ascending id order ... got [3, 1, 2]
   (tag ids: [('zeta-tag', 1), ('alpha-tag', 2), ('mid-tag', 3)])
```

I did **not** add ordered lock acquisition on the add side, and I'd like your call before doing it. `bulk_add_tags`'s count update is a single `UPDATE … CASE WHEN` outside any `atomic()` block, so forcing its order means wrapping it in a transaction and adding a `SELECT … FOR UPDATE ORDER BY pk` — an extra query and extra lock duration on the import hot path, which also moves the expected query counts the performance tests assert. That felt like it wanted to be its own change with its own perf validation rather than a rider here. Ordering the remove side already eliminates remove-vs-remove, which is what the report shows; remove-vs-add is what the retry now covers.

### You're right that I overstated the concurrency, and here's what the report actually says

No fan-out — I checked your citations and they hold: the chunk loop at `dojo/finding/helper.py` is serial, `async_delete.delete()` dispatches one task per object, and every call site deletes a single object. My description implied fan-out inside one request and that was wrong.

What the report does establish, sanitized (it's a customer instance in a private channel, so I can't link it): **7 distinct `async_delete_task` task ids** failing on one tenant inside a 33-second window, every one `OperationalError: deadlock detected` on `dojo_tagulous_product_tags`, with `DETAIL` blocks naming 2–3 backends waiting on each other — e.g. one three-way cycle across pids 245194 / 245196 / 245199 / 245201, then two separate two-way cycles. So the tasks were deadlocking against *each other*, which is remove-vs-remove, from several overlapping single-object deletes rather than one bulk action.

**Worker concurrency: I don't know, and I'm not going to guess.** It isn't in the diagnostics payload. If it turns out the partner is sometimes an import rather than another delete, I agree the 5s base backoff is optimistic — but note that ordering the remove side removes the delete-vs-delete case entirely, which is the one the report shows, so the backoff is now only load-bearing for the mixed case.

### The rest, point by point

1. **Grading contention — correct, removed.** `perform_product_grading` *dispatches* `calculate_grade`, so a grading deadlock lands in that task, not this one. The claim is gone from the description and the docstring.

2. **Post-delete work isn't recoverable — correct, and now loud.** Steps 5–6 aren't resumable; a re-invocation returns early and grading is silently skipped. The docstring now says exactly that, and the retry case logs at WARNING ("already gone on a retry … post-delete work was not redone") while a first-call miss stays at INFO. Both covered by tests.

3. **Eager backoff — confirmed empirically, and the retry is now skipped there.** On celery 5.6.3, `retry()` under `apply()` re-invokes the body immediately and discards `countdown`:

   ```
   celery 5.6.3
   body invocations: 3   retries seen: [0, 1, 2]
   elapsed seconds (countdown=5 x2 would be ~10s if honoured): 0.04
   ```

   So `block_execution` callers would have got up to three back-to-back re-runs inside the request. `self.request.is_eager` now short-circuits to a re-raise.

4. **Tests — reworked, with one caveat I want to flag.** They now assert behaviour: a deadlocked attempt followed by a real re-invocation must leave the object deleted, eager execution must surface the error without retrying, and the WARNING must be emitted. But driving celery's *real* retry to re-execute in-process is incompatible with point 3 — celery's only in-process re-execution path is the eager one, which you asked me to disable, and pushing a request manually makes `retry()` re-raise via `called_directly`. So what's verified here is that the body is resumable and that `retry` is called with a sane delay; genuine broker-backed re-execution is integration territory. I'd rather say that plainly than dress a mock up as end-to-end coverage.

5. **Setting — dropped.** `ASYNC_DELETE_MAX_CONFLICT_RETRIES` is a module constant next to `ASYNC_DELETE_RETRY_DELAY`; `settings.dist.py` is untouched, and I've removed the `settings_changes` label.

6. **Placement — moved** to a new `dojo/db_utils.py`.

### Branch — this is already on `bugfix`

Worth double-checking on your side, because I think the tooling misled you here: the PR's base is `bugfix`, and the commit's parent is exactly the `bugfix` tip.

```
$ git rev-parse HEAD~1            # first commit's parent
60cd10855a8ccb57e388294f8a9f9cdd59e0004b
$ git rev-parse origin/bugfix
60cd10855a8ccb57e388294f8a9f9cdd59e0004b     # identical
```

`bugfix` is an ancestor of `dev` (`dev` is 129 commits ahead), so anything asking "is this commit in `dev`?" answers yes for every `bugfix` commit — which is probably where "based on `dev`" came from. The `fix(...)` prefix is also just conventional-commits, not a branch claim.

---

Maffooch:
Noting a state change so it doesn't pass unremarked: **the base has been retargeted from `bugfix` to `dev`** (now `dev` @ `a32098e1`). My previous comment said the base was `bugfix` — that was accurate when I wrote it; it changed afterwards.

I'm not flipping it back, because there's a defensible reason to want it on `dev` and it's your call, not mine. But two things follow from it that need a decision:

**1. It no longer ships in a patch release.** This is a deadlock aborting cascade deletes on a live instance now, which is the case you made for `bugfix` yourself. On `dev` it waits for the next minor.

**2. The milestone is now inconsistent.** It's set to **3.2.100**, "Weekly patch release" — which can't be satisfied from a `dev` base. Either the base or the milestone wants changing; I've left both alone pending your answer.

If the reason for the move is **scope**, that's fair and it's probably my fault: my second commit widened the change from `dojo/utils.py` into `dojo/tags/utils.py`, and `bulk_remove_all_tags` is shared with the import path. Wanting that to soak in a minor rather than ride a weekly patch is a reasonable read.

If that's it, I'd suggest splitting rather than delaying the whole thing — the two commits are already cleanly separated along exactly that line:

| Commit | Touches | Suggested target |
|---|---|---|
| `df9c557` retry backstop | `dojo/utils.py`, `dojo/db_utils.py` + tests | `bugfix` — absorbs the deadlock, no shared-path change |
| `f182199` lock ordering | `dojo/tags/utils.py` + tests | `dev` — the real fix, but touches the import path |

That gets affected instances relief in the weekly patch while the ordering change takes the longer route. Downside is the retry lands first and the root cause second, so for one release the deadlock is absorbed rather than prevented — which is where we were before your review, and survivable for a release.

Tell me which you want and I'll do it:

- **keep it all on `dev`** — I'll change the milestone to the minor and leave the branch alone;
- **all back on `bugfix`** — I'll retarget;
- **split as above** — I'll open a second PR for `f182199` against `dev` and reduce this one to the retry on `bugfix`.

Separately, unrelated to any of this: the `search_test.py` UI job failed on this run, and it isn't the diff. `docker compose up` couldn't reach Docker Hub — `Get "https://registry-1.docker.io/v2/": context deadline exceeded` while pulling `webhook.endpoint` — so the stack never started and no test ran. Same infrastructure class as the `postgres` 502 that hit `sla_configuration_test.py` on the previous run. I'll re-run the failed jobs once the run finishes.

---

Maffooch:
Thanks for putting it back on `bugfix` and rebasing — that answers my question, and base + milestone (3.2.100) are consistent again.

I re-verified after the rebase rather than assuming it was clean:

- **Content is unchanged.** The rebased commits are byte-identical to what I pushed — `git diff` over both ranges is 553 lines each, differing only in the blob index line, as expected when the base moves.
- **`dojo/utils.py` did move underneath me**, so I checked the interaction. The upstream change is #15430's lint fix (bare `except:` → `except Exception:`) inside `get_period_counts`, nowhere near `async_delete_task`. No semantic overlap.
- **Re-ran the suites against the new base**: 118 tests across `test_async_delete`, `test_tag_utils_bulk`, `test_cascade_delete`, `test_bulk_delete_findings_m2m`, `test_prepare_duplicates_for_delete`, `test_utils`, `test_delete_with_endpoints_v3` — all pass.
- **Re-linted with the newly pinned ruff 0.16.0** (the base bumped it from 0.15.20 in `requirements-lint.txt`; my earlier runs used the old pin). Clean on all five changed files.

One note on what "green" currently means here: with #15528's two-tier checks, the 12 checks on this head are the light PR tier and they're all green, but the Unit tests suite now runs in the merge queue rather than on the PR. The full suite last ran on `f18219941a` — identical content, previous base — and passed 128/128 after a re-run. So the heavy verification for *this* SHA will happen when it's queued.

Also worth flagging, since it cost two re-runs on this PR: #15522 landing the Docker Hub pull retries should stop exactly the failures I hit — a `postgres` 502 and a `webhook.endpoint` timeout, both killing an integration job before any test ran.

Still deliberately not done, from the earlier review: ordered lock acquisition on the tag *add* side. It needs a `SELECT … FOR UPDATE` in a new transaction on the import hot path, which shifts the query counts the performance tests assert, so I'd rather it be its own change with its own perf validation than a rider here. Happy to open that separately if you want it.

---

blakeaowens (approved):

devGregA (approved):

PR #15499, opened 2026-08-03
Scope importer file-attachment reuse to the owning finding

svader0:
Hardening / consistency improvement to scan-import file handling. The importer now reuses an uploaded-file row only when the finding being imported already references that title, and otherwise creates a new row (uniquifying the title on collision), instead of resolving the file title against the instance-wide namespace. Adds a regression test.

No functional change for a re-import of the same finding.

### DIFF
--- a/dojo/importers/base_importer.py
+++ b/dojo/importers/base_importer.py
@@ -1,6 +1,7 @@
 import base64
 import logging
 import time
+from uuid import uuid4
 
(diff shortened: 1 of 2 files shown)


svader0:
Superseded by #15498, which fixes the same issue (scoping uploaded-file titles to their owner and creating a fresh row per attachment). Closing this in favor of the merged change.

PR #15500, opened 2026-08-03
Risk acceptance: keep the requested expiration date, add expire/reinstate API actions, and fix broken links and the SonarQube status sync

devGregA:
**Description**

A group of risk acceptance fixes, plus two new API actions. Bugfix branch: everything here is repairing existing behaviour, and the two new endpoints exist because expiring and reinstating were previously unreachable except by imitating them through `expiration_date`.

**Reinstating discarded the expiration date it was given.** `reinstate()` always overwrote `expiration_date` with today + the configured window, so extending an expired risk acceptance to a specific date produced a different date, and the only way to make one stick was to edit the risk acceptance a second time once it was active again. The documentation had to warn about this. It now keeps the requested date, and falls back to the configured window only when there is nothing to keep: the Reinstate action edits no date and passes back the one already stored, and expiry stamps `expiration_date` with the moment it lapsed. A date that is not in the future gets the same treatment, since the expiry job would immediately lapse it again.

**The default-window fallback never worked.** `get_system_setting(name, default)` substitutes its default only when the attribute is missing, so a null `risk_acceptance_form_default_days` column returned `None` and three call sites went on to compute `relativedelta(days=None)`. They now share `expiration_days()` / `default_expiration_date()`, whose fallback matches the column default of 180 rather than the 90 the code named.

**New: expire and reinstate over the API.**

- `POST /api/v2/risk_acceptance/{id}/expire/` — 400 if it has already expired.
- `POST /api/v2/risk_acceptance/{id}/reinstate/` — 400 if it has not expired. Optional `expiration_date`; without one the configured window applies.

Both accept an optional `reason`, recorded as a note on the risk acceptance together with the acting user, and both use the same authorization as editing the risk acceptance. `risk_unaccept()` gained matching `reason`/`source` arguments that flow into the note it leaves on the finding, and the reimporter now identifies itself: a finding that lost its acceptance because a scan stopped reporting it previously got a note saying only that "a risk exception" was removed, with no indication that a scan rather than a person had done it.

**Fixes to existing behaviour.**

- `get_view_risk_acceptance()` computed a URL and then returned `""` — the `return` was missing — so every audit note written when a finding was added to or removed from a risk acceptance read `... risk acceptance: "Name" ()` with nothing to click. It now also resolves an engagement through the accepted findings, so a risk acceptance that is only reachable that way gets a link instead of nothing.
- The Slack expiration template interpolated `risk_acceptance_url`, which is not in the notification context. The MS Teams one opened by reassigning `url` from a two-argument URL given a single argument — silently `""` under `as url` — discarding the correct URL the notification already passes. Both now use it, in the classic template tree as well.
- `Risk_Acceptance.get_breadcrumbs()` passed the product id where the URL takes an engagement id, so the crumb pointed at an unrelated engagement, and it raised `AttributeError` when the risk acceptance had no engagement.
- The SonarQube status sync called `ra_helper.remove_finding.from_any_risk_acceptance()`, which does not exist, so syncing an OPEN, CONFIRMED, FIXED or FALSE-POSITIVE issue raised `AttributeError` every time. Its WONTFIX branch also never set `risk_accepted` on the finding, and created an unnamed risk acceptance attached to no engagement — a row nobody could find, covering a finding that did not read as accepted.
- Downloading a proof that was never uploaded, or whose file has gone from storage, returned a 500 from `open()`; it is now a 404. Replacing a proof skipped extension validation entirely, unlike uploading one, and now shares the same accept list and `clean_path`.

**Test results**

New tests in `unittests/test_risk_acceptance.py`, `unittests/test_risk_acceptance_api.py`, and a new `unittests/test_api_sonarqube_updater_from_source.py` (the existing SonarQube test covers only transition mapping, not this updater). Each new test was re-run against unmodified source to confirm it fails without the fix — for example the reinstate date test fails with `datetime.date(2027, 1, 30) != datetime.date(2026, 9, 2)`, the audit-note test fails on `'(admin) (1) added this finding to the risk acceptance: "Accept: Unit test" ()'`, and the breadcrumb test fails with `/engagement/1/...` vs `/engagement/2/...`.

`ruff check` is clean, and `manage.py spectacular` emits both new paths with their request components and no new warnings.

The full `unittests` suite was run locally (5077 tests). It reports 269 failures and 1 error, and **all of them reproduce on unmodified `bugfix` in the same container against the same database**, so none are attributable to this PR — my local environment runs open-source settings inside a Pro image against a reused test database, which these particular suites are sensitive to. Accounted for individually rather than in aggregate: `test_parsers` (255, and the 1 error), `test_jira_config_product` (10), `test_filter_asset_tag_labels` (2), `test_product_endpoint_report_scoping` (2) and `test_notifications.test_auditlog_on` (1); re-running each module against a pristine `bugfix` worktree produced the same failures. The suites this PR touches are green: risk acceptance (3 suites), JIRA (2), import/reimport (132 tests), the API surface test, and SonarQube.

One thing worth noting for reviewers: an earlier draft of this change also guarded `reinstate()` against re-accepting findings that are mitigated, false positive or out of scope. That guard was removed, because `update_finding_status` already clears those flags whenever a finding is reactivated, so an active finding cannot carry them and `reinstate()` only touches active findings. The guard was unreachable.

**Documentation**

Both risk acceptance pages updated: the "your chosen date may be overridden / edit it a second time" warning is gone, since it described the bug, and the new endpoints are documented.

**Checklist**

- [x] If this is a new feature and not a bug fix, you've included the proper documentation in the docs as part of this PR.
- [x] Model changes must include the necessary migrations in the dojo/db_migrations folder. (No model changes.)

### DIFF
--- a/docs/content/triage_findings/findings_workflows/OS__risk_acceptance.md
+++ b/docs/content/triage_findings/findings_workflows/OS__risk_acceptance.md
@@ -124,7 +124,14 @@ Moving the date to a date in the past does not immediately expire the Risk Accep
 
 ### What the API exposes
 
-API consumers can observe expiration state on the Risk Acceptance object via the `expiration_date`, `expiration_date_handled`, and `expiration_date_warned` fields.  A Risk Acceptance is "expired" precisely when `expiration_date_handled` is non-null.  When a reinstate happens, both `expiration_date_handled` and `expiration_date_warned` are cleared back to `null`, and `expiration_date` is updated to the reinstate target.
+API consumers can observe expiration state on the Risk Acceptance object via the `expiration_date`, `expiration_date_handled`, and `expiration_date_warned` fields.  A Risk Acceptance is "expired" precisely when `expiration_date_handled` is non-null.  When a reinstate happens, both `expiration_date_handled` and `expiration_date_warned` are cleared back to `null`, and `expiration_date` holds the date you sent — or today + N days when no date was requested.
+
+Expiring and reinstating are also available directly, so you do not have to drive them by editing `expiration_date`:
+
+- `POST /api/v2/risk_acceptance/{id}/expire/` expires it now.  Returns `400` if it has already expired.
+- `POST /api/v2/risk_acceptance/{id}/reinstate/` reinstates an expired one, re-accepting the Findings it covers.  Returns `400` if it has not expired.  Send `expiration_date` to choose how long for; omit it to use today + N days.
+
+Both accept an optional `reason`, which is recorded as a note on the Risk Acceptance along with who performed the action.  Both require the same permission as editing the Risk Acceptance.
 
 ## Risk Acceptance Best Practices 
 
--- a/docs/content/triage_findings/findings_workflows/OS__risk_acceptance.md
+++ b/docs/content/triage_findings/findings_workflows/OS__risk_acceptance.md
@@ -120,7 +120,7 @@ The date you enter is the date that is saved.  The system setting **Risk Accepta
 
 Moving the expiration date to an earlier-but-still-future date has no special behavior — the Risk Acceptance stays active and the new date is saved.
 
-Moving the date to a date in the past does not immediately expire the Risk Acceptance from the edit form; the next periodic expiration job will pick it up and apply the standard expiration behavior.  On an **already-expired** Risk Acceptance a past date is not a reinstatement at all — the next expiration run would simply lapse it again — so the reinstate falls back to today + N days.
+Moving the date to a date in the past does not immediately expire the Risk Acceptance from the edit form; the next periodic expiration job will pick it up and apply the standard expiration behavior.  That holds for an **already-expired** Risk Acceptance too: a past date is still the date you chose, so it is saved as-is and the next expiration run lapses the Risk Acceptance again.
 
 ### What the API exposes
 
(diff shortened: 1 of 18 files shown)


devGregA:
C:/Program Files/Git/test all

devGregA:
Rebased onto `bugfix` to clear the conflicts.

The overlap was substantive rather than textual: #15498 fixed the same reinstate-date bug with the same condition. I kept upstream's version and dropped my duplicate helper, so what remains of that commit is only the null-safe default — `get_system_setting()` substitutes its default when the *attribute* is missing, not when the column is NULL, so a null `risk_acceptance_form_default_days` returned `None` and `relativedelta(days=None)` raised at three call sites.

One behavioural difference had to be resolved rather than merged. My version also required the requested date to be in the future, falling back to the configured window otherwise; #15498 honors any date the caller supplied. Upstream's rule is the better one — falling back would silently override the user's input, which is what this path set out to stop — so I adopted it and rewrote my test and the docs paragraph to describe that, noting the next expiration run simply lapses the acceptance again. My separate API test for the kept date is gone too, since #15498 added that assertion to its own test.

Local re-run after the rebase: 56 tests green across the risk-acceptance, bulk-acceptance and SonarQube suites; `ruff check` clean.

/test all

PR #15503, opened 2026-08-04
fix(dedupe): drop duplicate links whose original was deleted mid-batch

Maffooch:
**Description**

> **Scope narrowed.** This PR originally fixed both directions of one window on the `duplicate_finding` self-FK. The **delete side** has since landed independently as #15510, with a better ancestor walk than this branch carried, so that work has been dropped and #15510 is now the base. What remains is the **dedupe side**, which #15510 does not touch.

The `duplicate_finding` self-FK is `DEFERRABLE INITIALLY DEFERRED`. Batch deduplication matches its originals near the start of a batch and writes the links at the end of it, and the excess-duplicate delete task runs on its own schedule — so an original can be gone by the time the write lands. The bad link is then only rejected at COMMIT, which rolls back the entire batch:

```
django.db.utils.IntegrityError: insert or update on table "dojo_finding" violates foreign key
constraint "dojo_finding_duplicate_finding_id_..."
DETAIL:  Key (duplicate_finding_id)=(N) is not present in table "dojo_finding".
```

Every other finding in that batch loses its deduplication too, and `post_process_findings_batch` fails — all for one unusable link.

**The fix** (`dojo/finding/deduplication.py`)

`_flush_duplicate_changes` now verifies that each matched original still exists immediately before the write and in the same transaction, and drops only the links that cannot be written. Those findings are left exactly as they were, for the next import to match again. The return value narrows to what was actually persisted, so follow-up processing runs only on rows that were written.

**Why not row locks**

This narrows the window to a single transaction rather than closing it. Locking the originals would put a `select_for_update` on every import's dedupe write and create a lock-ordering cycle with the delete task — trading an occasional aborted batch for routine deadlocks on a hot path.

**Pro impact**

Pro calls `_flush_duplicate_changes` from all three of its dedupe branches via `FINDING_DEDUPE_BATCH_METHOD`, so the fix applies there unchanged with no companion PR. The narrowed return value is what Pro's post-flush prioritization consumes, so it now fires only for findings that were persisted.

**Test results**

`unittests/test_dedupe_flush_missing_original.py` — fails on the unfixed code with the production constraint error above. Covers the dropped link, the rest of the batch still persisting, findings carrying a null link, and an empty batch.

Suites run locally against this base, all green:

```
test_dedupe_flush_missing_original, test_bulk_delete_duplicate_references,
test_bulk_delete_findings_m2m, test_prepare_duplicates_for_delete,
test_duplication_loops, test_deduplication_logic, test_cascade_delete,
test_async_delete, test_importers_deduplication      Ran 222 tests — OK (skipped=2)

test_importers_performance                            Ran 11 tests — OK
```

`ruff check .` passes.

**Performance counts**

`expected_num_queries2` moves 94 → 97 and 204 → 207. Three queries per *batch*: the existence check, plus the savepoint pair for the transaction it now shares with the write. Measured by running the suite, not derived — and confirmed to come entirely from this change, since #15510's delete-side work leaves these counts untouched. The transaction also makes the batch write all-or-nothing, where `bulk_update`'s statements previously each committed on their own.

**Documentation**

No user-facing behaviour change: this removes a failure mode from a background task. No docs update needed.

**Checklist**

### DIFF
--- a/dojo/finding/deduplication.py
+++ b/dojo/finding/deduplication.py
@@ -4,6 +4,7 @@ from operator import attrgetter
 
 import hyperlink
 from django.conf import settings
+from django.db import transaction
 from django.db.models import Prefetch
 from django.db.models.query_utils import Q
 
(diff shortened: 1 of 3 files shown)


blakeaowens (approved):

devGregA (approved):

Maffooch:
Heads up before this is re-queued: **the scope changed after the approvals above.**

This PR was dequeued on a merge conflict. Resolving it surfaced that #15510 had landed the **delete-side** half of this fix independently — same diagnosis and same placement, with a better ancestor walk than this branch carried (it follows a duplicate chain a level at a time when the chain runs out of the chunk, where this branch promoted to original in that case).

So I dropped the delete-side change and its test, rebased onto #15510, and narrowed this PR to the **dedupe side** — `_flush_duplicate_changes`, which #15510 does not touch. It went from 5 files to 3:

| File | |
|---|---|
| `dojo/finding/deduplication.py` | +63 |
| `unittests/test_dedupe_flush_missing_original.py` | +201 (new) |
| `unittests/test_importers_performance.py` | ±4 |

Current state: CI green (47/47), `mergeable_state: clean`, verified locally at 222 tests across the dedupe and delete suites (including #15510's new module) plus 11 performance tests.

I have deliberately **not** re-enabled auto-merge. It was enabled before the dequeue, but that was against the larger diff, and the approvals predate the rewrite — worth a fresh look at the narrowed change before it goes back in the queue rather than having it merge on an approval given for different work.

---

Maffooch:
This PR was dequeued on a `CI_FAILURE`. Recording the diagnosis in case it recurs — **the failing test is not related to this change.**

The queue run failed on exactly one test, in the arm64 job:

```
FAIL: test_api_token_auth_is_rate_limited (unittests.test_apiv2_user.UserTest.test_api_token_auth_is_rate_limited)
AssertionError: 6 != 3 : api-token-auth should honor the configured rate limit
Ran 5077 tests — FAILED (failures=1, skipped=495)
```

Why I do not think this is mine:

- **No causal path.** This diff is `dojo/finding/deduplication.py` (`_flush_duplicate_changes`), one new dedupe test module, and two integers in the performance-count assertions. The failing test exercises API token auth throttling. Nothing in the diff touches auth, throttling, caching, or middleware.
- **It passes locally on this branch.** `manage.py test unittests.test_apiv2_user` → 15 tests, OK.
- **It passed on this PR's own head run** and in every earlier full run of this branch.
- **The assertion shape points at throttle state, not logic.** The view was reached 6 times where 3 were expected — the throttle did not engage at all, rather than engaging incorrectly. That is what a rate-limit counter shared across concurrently-running suites looks like, and #15505 recently moved the parallel-safe phase to run in parallel.
- **The queue is failing broadly right now,** on unrelated tests. Of the last ~30 `merge_group` "Unit tests" runs, six failed, across at least two different and unrelated tests — #15563's run, for instance, failed on `test_importers_blank_engagement_status` (`'Not Started' != ''`), not on this one.

I have not re-queued it myself — thanks @Maffooch for putting it back. If it trips on `test_api_token_auth_is_rate_limited` again, the place to look is that test's isolation under parallel execution, not this diff.

---

PR #15504, opened 2026-08-04
ci(migrations): bring the migration graph check to dev

devGregA:
## Description

#15501 added a check that fails fast when a Django app has more than one migration leaf node. It landed on `bugfix`. This is the same commit, cherry-picked to `dev`.

`dev` is where this has actually bitten. Two changes that each add a migration on the same parent are individually consistent, so both go green before merge — the fork exists only in their union, and nothing re-derives the graph afterwards. Until a release merge-back carries #15501 across, the branch most exposed to that is the one without the check.

While the graph is forked, `migrate` refuses to apply anything at all, so every container-based job dies at boot. That surfaces as ~120 red checks, none of which mentions migrations.

Cherry-picked rather than reimplemented, so the two branches stay byte-identical and the next merge-back is a no-op.

## Test results

On this branch:

```
$ python3 scripts/check_migration_leaves.py
dojo: 199 migrations, 90 squashed out, single leaf 0288_backfill_vulnerability_id_entities
$ echo $?
0
```

```
$ python -m pytest unittests/test_migration_leaves.py -q
22 passed in 1.82s
```

`ruff check .` clean with the CI-pinned 0.16.0.

The script is stdlib-only — `ast`, no Django, no database, no settings — so the job is a checkout plus one `python3` call, well under a second.

### DIFF
--- a/.github/workflows/migration-graph.yml
+++ b/.github/workflows/migration-graph.yml
@@ -1,0 +1,40 @@
+name: Migration Graph
+
+# Its own workflow rather than a job inside the lint workflow: this check goes
+# red for a reason that has nothing to do with code style, and it goes red on a
+# branch whose own commits are usually fine, so the name on the PR rollup needs
+# to say what actually broke.
+#
+# Both triggers are load-bearing, because there are two ways to fork the graph
+# and each trigger catches only one of them:
+#
+#   pull_request - actions/checkout resolves a pull_request event to the merge
+#     commit, so this sees the branch *combined with* its base. That is the only
+#     way to catch a migration written against a base branch that had already
+#     moved on: the branch on its own has a single leaf and looks fine.
+#
+#   push - catches the mirror case, where two PRs each add a migration on the
+#     same parent. Both are internally consistent so both go green pre-merge,
+#     and the fork exists only in their union. Nothing else re-derives the graph
+#     after a merge, since the test suite is per-PR.
+#
+# Without this check the failure mode is every container-based job dying at boot
+# with ~120 red checks, none of which says "migration graph" in its name.
+
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
+
+jobs:
+  migration-graph:
+    name: Migration Graph Check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      # The script is stdlib-only (ast, no Django, no database, no settings), so
+      # it needs no setup beyond the checkout and finishes in well under a second.
+      - name: Check for multiple migration leaf nodes
+        run: python3 scripts/check_migration_leaves.py
--- a/ruff.toml
+++ b/ruff.toml
@@ -129,6 +129,10 @@ preview = true
 "dojo/filters.py" = [
     "A003",  # ruff upgrade to 0.13.3
 ]
+"scripts/check_migration_leaves.py" = [
+    "T201",   # print statements are fine for console output scripts
+    "EXE001", # git won't commit the executable flag I don't know why
+]
 "scripts/update_performance_test_counts.py" = [
     "S603",  # subprocess.run without shell=True is safe for this script
     "S604",  # subprocess.run without shell=True is safe for this script
--- a/scripts/check_migration_leaves.py
+++ b/scripts/check_migration_leaves.py
@@ -1,0 +1,333 @@
+#!/usr/bin/env python3
+"""
+Fail when the Django migration graph has more than one leaf node in an app.
+
+``migrate`` refuses to apply *anything* while a graph has two leaves::
+
+    CommandError: Conflicting migrations detected; multiple leaf nodes in the
+    migration graph: (0281_fileupload_title_not_unique,
+    0288_backfill_vulnerability_id_entities in dojo).
+
+Every fresh install and every CI database build off the branch is blocked until
+someone re-parents one of them. There are two ways to arrive there, and the
+workflow that calls this script watches for both:
+
+* Two changes each add a migration on top of the same parent. Each is
+  internally consistent, so both go green pre-merge; the fork exists only in
+  their union, and nothing re-derives the graph after a merge.
+* One change adds a migration on top of a parent that is no longer the tip,
+  because the author's view of the base branch was stale. The branch alone
+  looks fine — the fork appears in the merge with the base.
+
+The failure is expensive out of proportion to the mistake: with no guard it
+surfaces as every container-based job dying at boot, ~120 red checks whose logs
+all have to be opened before one of them names the actual problem.
+
+The script is deliberately standalone: ``ast`` only, no Django import, no
+database, no settings module, no third-party package. That is what lets it run
+as a plain ``python3`` step with nothing but a checkout, in under a second.
+
+Usage::
+
+    python3 scripts/check_migration_leaves.py           # checks dojo
+    python3 scripts/check_migration_leaves.py DIR...    # checks DIRs
+"""
+
+from __future__ import annotations
+
+import argparse
+import ast
+import sys
+from pathlib import Path
+from typing import TYPE_CHECKING, NamedTuple
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+# ``<repo>/scripts/`` -> ``<repo>/dojo/db_migrations``. ``dojo`` is the only
+# Django app in this repo that ships migrations, and its migration modules live
+# in ``db_migrations`` rather than the conventional ``migrations`` (see
+# MIGRATION_MODULES in dojo/settings/settings.dist.py).
+DEFAULT_MIGRATION_DIRS = (Path(__file__).resolve().parents[1] / "dojo" / "db_migrations",)
+
+# Django resolves these against another app's graph, never this one's.
+CROSS_APP_SENTINELS = frozenset({"__first__", "__latest__"})
+
+# ``migrations.swappable_dependency(settings.AUTH_USER_MODEL)`` resolves to whichever
+# app owns the user model — never ``dojo`` — so it cannot affect an intra-app leaf
+# count. It is common enough that reporting it as unreadable would be pure noise.
+KNOWN_CROSS_APP_CALLS = frozenset({"swappable_dependency"})
+
+
+class Reference(NamedTuple):
+
+    """An ``(app_label, migration_name)`` pair as written in a migration file."""
+
+    app_label: str
+    name: str
+
+
+class ParsedMigration(NamedTuple):
+
+    """The graph-relevant attributes of one migration module."""
+
+    name: str
+    path: Path
+    dependencies: tuple[Reference, ...]
+    run_before: tuple[Reference, ...]
+    replaces: tuple[Reference, ...]
+    # Entries we could not read as literal ``("app", "name")`` pairs — e.g.
+    # ``migrations.swappable_dependency(settings.AUTH_USER_MODEL)``, which is
+    # always cross-app and so never affects an intra-app leaf count. Reported
+    # only when the check fails, as context for an otherwise puzzling verdict.
+    opaque: tuple[str, ...]
+
+
+class AppGraph(NamedTuple):
+
+    """The intra-app migration graph derived from one migrations directory."""
+
+    app_label: str
+    directory: Path
+    migrations: dict[str, ParsedMigration]
+    children: dict[str, set[str]]
+    leaves: tuple[str, ...]
+    # ``("0281_x", "0280_missing")``: 0281 depends on a same-app migration that
+    # is not on disk. Django raises NodeNotFoundError for this, and it would
+    # also skew the leaf count here, so it is reported rather than ignored.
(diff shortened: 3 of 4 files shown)


blakeaowens (approved):

Maffooch (approved):

valentijnscholten:
I checked the premise here and it holds up — but I think the implementation is quite a bit larger than the job needs, and I'd like to make the case for trimming it before this lands.

## The premise verifies

I traced the CI chain: `ruff` gates, then `build-docker-containers` builds two platforms, and only then do ~57 integration cases plus rest-framework (2 platforms × 2 feature-flag values), k8s, and performance fan out. The "~120 red checks" figure is credible, and a forked graph really does block all of them, because every container runs `migrate` at boot. The script finishes in ~0.2s on a bare checkout, so the fail-early value is real. No argument with the goal or with cherry-picking to `dev`.

## The squash handling is genuinely load-bearing

Worth stating explicitly, because it's the part that looks skippable and isn't. I wrote a naive leaf counter that ignores `replaces`, and it reports a **false positive** on this repo: `0001_squashed_0090_index_duplicate_finding` shows up as a second leaf, because `0091` depends on `0090` rather than on the squash. Anyone trying to shrink this needs to keep that logic.

## Three features never fire

- **`run_before`** — there is no `run_before` assignment anywhere in `dojo/db_migrations`. The only grep hit is a docstring mention in `0268_release_authorization_to_pro.py`, describing a cross-app pairing that the script filters out anyway (`app_label == "dojo"`). Confirmed via `build_app_graph`: 0 migrations with a `run_before` edge.
- **`dangling`** — 0 entries on the real graph.
- **`opaque`** — 0 reported. All 11 `swappable_dependency` call sites are suppressed by `KNOWN_CROSS_APP_CALLS`, so the machinery that surfaces unreadable entries and the filter that hides them cancel out on every input this repo actually has.

That's ~140 lines of the 178 code lines (333 total, of which ~92 are comments and docstrings) serving cases that don't exist yet. Same imbalance in the 403 lines of tests. The tests themselves are nicely built, by the way — path-based import so `scripts/` needs no `__init__.py`, throwaway fixture trees, and `TestRealDojoGraph` pinning the actual graph is a good touch.

## Django already does this, and we already call it

`MigrationLoader.detect_conflicts()` implements exactly this check, and `makemigrations --check --dry-run` raises on it before doing anything else — which we already run at `docker/entrypoint-unit-tests.sh:55`. So this is a reimplementation whose sole justification is avoiding a Django import. That's a defensible trade for a sub-second job on a bare checkout, but it's worth naming in the PR description, because it carries a divergence risk: if Django's `replaces` resolution semantics shift, a hand-rolled AST parser drifts from what `migrate` actually does, and `TestRealDojoGraph` can't detect that kind of semantic drift.

Separately, and cheaply: when that existing check catches a leaf conflict, the heredoc prints *"You made changes to the models without creating a DB migration for them"* — an active misdiagnosis, since the real cause is a fork. Fixing that message costs almost nothing and helps even with this PR merged. It may also be most of why the failure currently reads as unrelated to migrations.

## A shorter version

Below is a drop-in replacement — 43 lines of code (70 with the docstring, comments, and the full error message). It keeps the squash handling and the `no migrations found (wrong path?)` guard, which is worth keeping: an early version of mine lacked it and reported a confusing empty leaf list when pointed at the wrong directory.

I verified it against the current PR's script on both paths:

- Clean graph: `dojo: 199 migrations, 90 squashed out, single leaf 0288_backfill_vulnerability_id_entities` — byte-identical to this PR's output.
- Forked graph (injected `0289` hanging off `0287` in a scratch copy): both scripts name the same two leaves, `0288_backfill_vulnerability_id_entities, 0289_rival_change`, and both exit 1.
- `ruff check` with the repo's `ruff.toml` on ruff 0.16.0: one `T201`, nothing else.

```python
#!/usr/bin/env python3
"""
Fail when the dojo migration graph has more than one leaf node.

``migrate`` refuses to apply anything at all in that state, so every fresh
install and every container-based CI job is blocked until someone re-parents
one of the leaves. Stdlib-only (``ast``, no Django, no database, no settings)
so it runs as a bare ``python3`` step on nothing but a checkout.
"""
import ast
import sys
from pathlib import Path

# dojo's migration modules live in db_migrations, not the conventional
# migrations (see MIGRATION_MODULES in dojo/settings/settings.dist.py).
MIGRATIONS_DIR = Path(__file__).resolve().parents[1] / "dojo" / "db_migrations"
GRAPH_ATTRS = ("dependencies", "replaces")


def graph_attrs(path):
    """Read literal ("dojo", "name") pairs out of one migration's class body."""
    found = {attr: [] for attr in GRAPH_ATTRS}
    for cls in ast.parse(path.read_text(encoding="utf-8")).body:
        if not (isinstance(cls, ast.ClassDef) and cls.name == "Migration"):
            continue
        for stmt in cls.body:
            if not isinstance(stmt, ast.Assign):
                continue
            for target in (t.id for t in stmt.targets if isinstance(t, ast.Name)):
                if target not in found:
                    continue
                for item in getattr(stmt.value, "elts", []):
                    # Non-literal entries (migrations.swappable_dependency(...))
                    # are always cross-app, so they cannot affect a dojo leaf.
                    if isinstance(item, (ast.Tuple, ast.List)) and len(item.elts) == 2 and all(
                        isinstance(e, ast.Constant) and isinstance(e.value, str) for e in item.elts
                    ):
                        app, name = (e.value for e in item.elts)
                        if app == "dojo":
                            found[target].append(name)
    return found


migrations = {p.stem: graph_attrs(p) for p in sorted(MIGRATIONS_DIR.glob("*.py")) if p.stem != "__init__"}
if not migrations:
    sys.exit(f"{MIGRATIONS_DIR}: no migrations found (wrong path?)")

# A squashed migration takes the place of what it replaces: the replaced nodes
# leave the graph and references to them land on the squash, mirroring a fresh install.
squashed = {old: new for new, a in migrations.items() for old in a["replaces"] if old in migrations}
for old in squashed:
    del migrations[old]

children = {name: set() for name in migrations}
for name, attrs in migrations.items():
    for parent in (squashed.get(d, d) for d in attrs["dependencies"]):
        if parent in children and parent != name:
            children[parent].add(name)

leaves = sorted(name for name, kids in children.items() if not kids)
if len(leaves) > 1:
    sys.exit(
        f"Conflicting migrations detected; multiple leaf nodes in the migration "
        f"graph: ({', '.join(leaves)} in dojo).\n\n"
        "`migrate` refuses to apply anything in this state. Re-parent the later "
        "migration onto the other leaf and renumber it, or add an empty merge "
        "migration depending on both. Check what the base branch actually holds "
        "first -- a stale remote-tracking ref is a common cause.",
    )
print(f"dojo: {len(migrations)} migrations, {len(squashed)} squashed out, single leaf {leaves[0]}")
```

Trade-offs I'm accepting, so they're explicit rather than accidental: it drops multi-directory/argparse support (nothing calls it with arguments), it silently ignores non-literal dependency entries instead of reporting them, and it hardcodes the `dojo` label. If any of those matter later, they're easy to add back at that point. Tests would shrink proportionally — I'd keep the fork, stale-base, squash, and empty-directory cases plus `TestRealDojoGraph`.

## Two small things regardless of which version lands

**`push:` is unfiltered**, so every PR gets two runs of this check (one `push`, one `pull_request`). Scoping it to `branches: [master, dev, bugfix, 'release/**']` still catches the post-merge fork case the comment describes, and halves the runs. It does match `ruff.yml`'s existing pattern, so this is noise-vs-consistency rather than a defect — your call.

**The `EXE001` ignore looks like leftover scaffolding.** The comment says *"git won't commit the executable flag I don't know why"*, but the diff shows `new file mode 100755` — the flag did get committed. On ruff 0.16.0 I couldn't get EXE001 to fire on this file with or without the bit set, so I think that entry can go and only `T201` is needed. (If it ever does come back: `git update-index --chmod=+x`.)

## One thing I went looking for and did not find

Neither Django Dockerfile does `COPY scripts/`, so I expected `SCRIPT_PATH = /app/scripts/check_migration_leaves.py` to blow up at import time in the `django-unittests` image and take the whole suite red. It doesn't: both `docker-compose.override.unit_tests.yml` and `docker-compose.override.unit_tests_cicd.yml` bind-mount `.:/app:z`, so the file is there at runtime in local and CI runs alike. Flagging it only because it's load-bearing and invisible — the test's import works because of a bind mount, not because the image contains the script.

devGregA:
will respond to @valentijnscholten in a min, but merging and will explain follow-up and why

devGregA:
Thanks — this is a better review than the PR deserved, and you're right about most of it. The short version: I agree with the trim, I don't think it should happen in *this* PR, and one property of the long version is worth keeping.

## Why not here

This PR is a cherry-pick of #15501, which is already merged to `bugfix`. The whole reason it's a cherry-pick rather than a reimplementation is so the two branches stay byte-identical and the next release merge-back is a no-op. If the script gets trimmed here, `dev` and `bugfix` diverge on a file that is 780 lines of scaffolding-plus-tests, and the next merge-back conflicts in exactly the place nobody wants to resolve conflicts.

So I'd rather land this as-is to restore parity, then put your version up against `bugfix` and let it reach `dev` the ordinary way. Same end state, one fewer conflict, and the trim gets reviewed on its own merits instead of riding along with a port.

## Where you're straightforwardly right

**The three dead paths.** I confirmed all of it independently: no `run_before` assignment anywhere in `dojo/db_migrations` (the only hit is the docstring in `0268_release_authorization_to_pro.py`), zero dangling entries, and zero opaque ones because `KNOWN_CROSS_APP_CALLS` suppresses all 11 `swappable_dependency` sites — the reporting machinery and the filter do cancel out on every input this repo has. That's generality I wrote for graphs this repo doesn't have, which is the expensive kind.

**`EXE001` is stale, and the comment next to it is simply wrong.** `git ls-tree` reports `100755`, so the bit did commit. That entry is leftover scaffolding from an earlier attempt and the comment should never have shipped.

**The heredoc misdiagnosis is the best thing in this review**, and it's independent of everything else. When `makemigrations --check` catches a fork it currently prints "You made changes to the models without creating a DB migration for them", which sends the reader looking for a missing migration that doesn't exist. I'd go further than you did on the likely impact: that message is probably why forks have historically read as unrelated to migrations even when a check *did* catch them. Worth fixing on its own, ahead of any of this.

**The `SCRIPT_PATH` bind-mount dependency** is a real latent trap. Neither Dockerfile does `COPY scripts/`, so the test passes because of `.:/app:z`, not because the image contains the script. Nothing breaks today, but it breaks the day someone runs the suite from the image alone.

## On "Django already does this"

Agreed that it should be named in the description, and I'll do that. But I'd frame the justification differently than "avoiding a Django import", because I don't think that's the real value.

`makemigrations --check` runs inside the unit-test job, which needs both images built and postgres up. Pushes to `dev` and `bugfix` don't run unit tests at all — `unit-tests.yml` is `pull_request` only. So the case this guard exists for, two individually-green PRs whose union forks the graph, is precisely the case Django's check cannot reach: by the time anything runs `makemigrations`, someone has already opened the next PR and is debugging ~120 red containers.

That also answers your divergence concern, I think. `makemigrations --check` stays exactly where it is, so Django's own `detect_conflicts()` remains the authority in CI. The AST script is a fast pre-filter that runs in more places, not a replacement for it. If the two ever disagree on `replaces` semantics, Django still fails the suite — the script drifting would cost us early warning, not correctness.

## One thing I'd keep from the long version

Your version silently drops non-literal dependency entries, on the reasoning that `swappable_dependency(...)` is always cross-app. That's true today. But the failure mode if it ever isn't is a **false negative**: a dropped `dojo` edge is a parent link that goes missing, which can hide a fork rather than invent one. For a guard, that's the wrong direction to fail in — a false positive gets noticed and fixed in minutes; a false negative just means the guard was quietly off when it mattered.

The full reporting machinery is not the answer to that, and I'm not arguing to keep it. Three lines is: count the entries you couldn't read as a literal `("dojo", "name")` pair, and if that count is non-zero, exit non-zero saying so. Loud and unmissable, no `opaque`/`dangling`/`KNOWN_CROSS_APP_CALLS` apparatus, and today it reports zero on this repo so it costs nothing until it matters.

Everything else in your 43 lines I'd take as written, including keeping the `no migrations found (wrong path?)` guard — and your point about the squash handling matches why that logic is shaped the way it is. I hit the same `0001_squashed_0090_index_duplicate_finding` false positive early on, which is what put it there.

## On `push:`

You're right that it doubles the runs, and I'll filter it. Both triggers are load-bearing for different fork shapes — `pull_request` sees the branch merged with its base, `push` catches the post-merge union — and scoping `push` to `[master, dev, bugfix, 'release/**']` keeps both while halving the runs. Consistency with `ruff.yml` isn't worth paying twice per PR for.

## Plan

1. Merge this as-is, so `dev` and `bugfix` match again.
2. One PR against `bugfix`: your 43-line version plus the unreadable-entry check, tests trimmed to the fork / stale-base / squash / empty-directory cases plus `TestRealDojoGraph`, `EXE001` and its comment removed, `push:` filtered, and the "why not `makemigrations --check`" rationale in the description.
3. The heredoc fix separately, since it stands alone and helps regardless of which version of the script exists.

Happy to do all three, or to hand you (2) if you'd rather land your own version.

PR #15505, opened 2026-08-04
ci: run the parallel-safe unit test phase in parallel

devGregA:
## Description

The unit test job spends nearly all of its time in a single serial process. Measured on a recent green run ([30873644826](https://github.com/DefectDojo/django-DefectDojo/actions/runs/30873644826), job 91881121251), a 15.6 minute job breaks down as:

| Phase | Duration |
|---|---|
| container start + spectacular + `makemigrations --check` | 1.0 min |
| `migrate` (264 migrations) | 1.6 min |
| **phase 1: 4996 tests** | **12.7 min** |
| phase 2: 45 tests, `tag=non-parallel` | 0.1 min |
| phase 3: 7 tests, `tag=transactional` | 0.0 min |

Four of those run per pull request (amd64 and arm64, each with `DD_V3_FEATURE_LOCATIONS` on and off), so phase 1 alone is about 50 minutes of the suite's runner time and, being the longest single job, most of its wall clock.

## Why phase 1 is the safe one

Phase 1 is already defined as the set of tests that do not need to run on their own. The `non-parallel` and `transactional` tags exist precisely to quarantine the suites that interfere with their neighbours, and phase 1 excludes both. Whatever isolation guarantee justifies running those two phases separately is the same guarantee `--parallel` needs, so it is enabled for phase 1 and nowhere else.

The comment this removes — `# --parallel fails on GitHub Actions` — predates that tagging. The concrete failure it refers to is documented three lines below it, on phase 3: a `TransactionTestCase` leaves the database in a state that breaks fixture loading for every test after it (see #13217). That test now runs last, in its own phase. The fix landed; the comment stayed.

## This is a measured experiment, with the criterion stated up front

**Keep** if phase 1 drops materially across green runs with no new flakes. **Revert** otherwise — `DD_TEST_PARALLEL=1` restores single-process behaviour without touching the script, and dropping the flag is a one-line change.

Worker count comes from `nproc` rather than a constant so it follows the runner rather than needing a change when the runner does.

Two things to watch, both stated so that seeing them is not a surprise:

- **New shared-state failures.** These 4996 tests have only ever run serially. If any depend on execution order or on module-level state, parallel workers will find it. That is the experiment.
- **`PYTHONWARNINGS: error`** is set for unit tests, and warnings raised inside worker processes surface differently from warnings in a single process.

Also adds `tblib`, which Django uses to ship tracebacks back from worker processes. Without it a failure under `--parallel` arrives without the traceback that explains it — the worst possible property for a change whose only real risk is new flakes.

## Verification

CI on this pull request *is* the measurement: the same four jobs, same suite, one variable changed. I will post the phase-1 timings from this run against the 12.7 minute baseline above, and act on the criterion rather than argue with it.

Local verification so far: `bash -n` clean, and shellcheck v0.11.0 with the repo's `SHELLCHECK_OPTS` reports nothing.

### DIFF
--- a/docker-compose.override.unit_tests_cicd.yml
+++ b/docker-compose.override.unit_tests_cicd.yml
@@ -20,7 +20,22 @@ services:
       DD_DATABASE_ENGINE: ${DD_DATABASE_ENGINE:-django.db.backends.postgresql}
       DD_DATABASE_HOST: ${DD_DATABASE_HOST:-postgres}
       DD_DATABASE_PORT: ${DD_DATABASE_PORT:-5432}
-      DD_CELERY_BROKER_URL: 'sqla+sqlite:///dojo.celerydb.sqlite'
+      # An in-process broker rather than the default sqlite file. Nothing consumes
+      # the queue in unit tests -- celeryworker and celerybeat are both reset
+      # above -- so a real broker contributes nothing but contention: under
+      # --parallel every worker process opens the same sqlite file and they race
+      # to declare the queue, which fails in whichever loses with
+      # "UNIQUE constraint failed: kombu_queue.name".
+      #
+      # Set through the scheme, because settings.dist.py only assembles a URL from
+      # the parts when DD_CELERY_BROKER_URL is empty. double_slashes=True there
+      # turns these two into exactly "memory://". The entrypoint also unsets
+      # DD_CELERY_BROKER_URL, but blanking it here means the broker does not
+      # depend on that happening -- and it stops the base compose file's
+      # redis://valkey URL, which no unit test has a valkey for, showing through.
+      DD_CELERY_BROKER_URL: ''
+      DD_CELERY_BROKER_SCHEME: 'memory'
+      DD_CELERY_BROKER_PATH: ''
       DD_JIRA_EXTRA_ISSUE_TYPES: 'Vulnerability' # Shouldn't trigger a migration error
       DD_V3_FEATURE_LOCATIONS: ${DD_V3_FEATURE_LOCATIONS:-False}
       # No Redis/valkey in unit tests -> default django cache is LocMemCache.
--- a/docker/entrypoint-unit-tests.sh
+++ b/docker/entrypoint-unit-tests.sh
@@ -81,10 +81,15 @@ echo "------------------------------------------------------------"
 # because they interfere with their neighbours. That is the same property
 # --parallel needs, so it is enabled here and nowhere else.
 #
+# "auto" rather than a count from nproc: Django's get_max_test_processes() falls
+# back to a single process if the multiprocessing start method is one its worker
+# initialisation cannot handle, and only "auto" consults it. An explicit count
+# bypasses that check and fails instead. See the note in manage.py.
+#
 # Set DD_TEST_PARALLEL=1 to go back to a single process -- useful locally when a
 # failure's traceback matters more than the wall clock, and the one-line revert
 # if this turns out to cost more in flakes than it saves in minutes.
-python3 manage.py test unittests --keepdb --no-input --parallel="${DD_TEST_PARALLEL:-$(nproc)}" --exclude-tag="non-parallel" --exclude-tag="transactional" --exclude-tag="performance" || {
+python3 manage.py test unittests --keepdb --no-input --parallel="${DD_TEST_PARALLEL:-auto}" --exclude-tag="non-parallel" --exclude-tag="transactional" --exclude-tag="performance" || {
     exit 1;
 }
 python3 manage.py test unittests --keepdb --no-input --tag="non-parallel" --exclude-tag="performance" || {
--- a/docker/entrypoint-unit-tests.sh
+++ b/docker/entrypoint-unit-tests.sh
@@ -73,14 +73,18 @@ EOF
 
 python3 manage.py migrate
 
-# --parallel fails on GitHub Actions
-#python3 manage.py test unittests -v 3 --no-input --parallel
-
 echo "Unit Tests"
 echo "------------------------------------------------------------"
 
-# Removing parallel and shuffle for now to maintain stability
-python3 manage.py test unittests --keepdb --no-input --exclude-tag="non-parallel" --exclude-tag="transactional" --exclude-tag="performance" || {
+# This phase is already defined as "everything that does not need to run on its
+# own": the two tags excluded here are exactly the suites that were quarantined
+# because they interfere with their neighbours. That is the same property
+# --parallel needs, so it is enabled here and nowhere else.
+#
+# Set DD_TEST_PARALLEL=1 to go back to a single process -- useful locally when a
+# failure's traceback matters more than the wall clock, and the one-line revert
+# if this turns out to cost more in flakes than it saves in minutes.
+python3 manage.py test unittests --keepdb --no-input --parallel="${DD_TEST_PARALLEL:-$(nproc)}" --exclude-tag="non-parallel" --exclude-tag="transactional" --exclude-tag="performance" || {
     exit 1;
 }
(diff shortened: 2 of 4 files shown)


devGregA:
## Two notes while the unit jobs finish

**The one red check is infrastructure, not this change.** `User Interface Tests (tests/engagement_test.py, debian, false)` failed pulling an image from Docker Hub:

```
mailhog Error Get "https://registry-1.docker.io/v2/": net/http: request canceled
while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

That is a Selenium job. This change touches `docker/entrypoint-unit-tests.sh` and `requirements-dev.txt`, neither of which is on that path. I will re-run it once the run finishes — it cannot be re-run while the workflow is still going.

**`--parallel` has a setup cost I should name before the numbers arrive.** Running it locally against a migrated database, Django cloned the test database once per worker:

```
test_defectdojo
test_defectdojo_1
test_defectdojo_2
test_defectdojo_3
test_defectdojo_4
```

That is `CREATE DATABASE ... TEMPLATE`, so it is a file copy of a 205-table schema, N times, before any test runs. It was slow enough to notice on a Docker Desktop volume, though that filesystem is not representative of a CI runner's.

So the honest form of the hypothesis is that phase 1 drops by less than the worker count would suggest, and by an amount reduced by cloning. If the timings come back showing cloning eating most of the gain, the interesting follow-up is `--parallel` combined with the snapshot in #15507, since a restored database is the cheapest possible clone template.

devGregA:
## First run failed, and the reason was not test isolation

All four unit jobs failed identically:

```
Process ForkServerPoolWorker-1:
  File ".../multiprocessing/pool.py", line 114, in worker
    task = get()
  File ".../multiprocessing/queues.py", line 387, in get
    return _ForkingPickler.loads(res)
  File "/app/unittests/authorization/test_authorization.py", line 32, in <module>
    from django.contrib.auth.models import AnonymousUser
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.
```

**Python 3.14 changed the default multiprocessing start method on Linux from `fork` to `forkserver`**, and this image is on 3.14.6. Django's parallel runner supports exactly two: `django/test/runner.py::_init_worker` calls `django.setup()` only when the method is `spawn`, on the reasoning that a *forked* worker inherits an initialised app registry. A forkserver worker gets neither, so it unpickles its subsuite — which imports test modules, which import models — against an empty registry.

Worth noting that **Django does guard against exactly this**, in `get_max_test_processes()`: any start method other than fork or spawn falls back to a single process. But that is only consulted for `--parallel auto`, and passing an explicit count from `nproc` walked straight past it. Had I used `auto` originally, this would have silently run serially — green, and no faster. Which is arguably the worse outcome, since I would have reported "no gain" and been wrong about why.

Fixed by setting `spawn` for the `test` command only, and switching to `--parallel auto` so that a future change of default degrades to serial rather than failing.

Spawn rather than restoring fork: it is the method whose worker initialisation Django actually implements, and forking a process that may already hold threads is what Python is deliberately moving away from.

### Verified before pushing

Against the real image and a restored database:

| Check | Result |
|---|---|
| default start method in the image | `forkserver` (confirms the diagnosis) |
| `get_max_test_processes()` under forkserver | `1` |
| `get_max_test_processes()` with spawn forced | `8` |
| `test_utils` + `test_apiv2_notifications` + `test_adminsite`, `--parallel=2` | `Ran 17 tests`, `OK`, exit 0 |
| `manage.py check` | unaffected — the start method is set only for `test` |

### One cost to watch in the numbers

Spawn pays a full interpreter start plus a complete app import per worker, and DefectDojo is not a small app to import. In my local run that overhead dominated completely — 17 tests took 114 s where the serial equivalent took 11 s — because there were 17 tests to amortise it over. Against a 4996-test, 12.7-minute phase it should amortise fine, but it is a fixed cost that scales with worker count, and it is the thing to look at if the timings come back underwhelming.

The unrelated `mailhog` registry timeout from the first run should also clear on this run.

devGregA:
## Result: phase 1 halved, and the isolation bug was real

With the celery broker moved in-process, **all four unit jobs pass**:

| Job | Before | After |
|---|---|---|
| linux/amd64, v3=true | 17.0 min | **8.58 min** |
| linux/amd64, v3=false | 15.9 min | **8.08 min** |
| linux/arm64, v3=true | 15.7 min | **7.87 min** |
| linux/arm64, v3=false | 15.7 min | **7.27 min** |

Phase 1 itself: **760s → 324s**, a 2.3x speedup on 5044 tests. Roughly **8 minutes off each of four jobs** — about 32 runner-minutes per pull request, and since these are the longest jobs in the suite, close to 8 minutes off its wall clock.

Spawn's per-worker startup cost, which I flagged earlier as the thing that might eat the gain, did not: it is a fixed cost amortised over 5044 tests, and at this scale it disappears into the noise.

### Both failures along the way were worth having

The `--parallel` risk I pre-registered was "new shared-state failures", and one appeared — but not as an order dependency between tests. Saving a `Product_Type` dispatches a celery task from a post_save signal, and the broker was a **sqlite file shared by every worker process**, so they raced to declare the queue and whichever lost took its whole test class down. Any test that saves a `Product_Type` could have drawn that short straw; only one did. Tagging that class `non-parallel` would have made the symptom go away and left the race in place for the next test to touch that code path.

Before that, the first run failed for a reason that had nothing to do with this repository: Python 3.14 moved the default multiprocessing start method to `forkserver`, which Django's parallel runner does not initialise workers for.

### The one red check is not this change

`User Interface Tests (tests/metrics_extended_test.py, debian, false)` failed pulling an image from Docker Hub:

```
webhook.endpoint Error Get "https://registry-1.docker.io/v2/": context deadline exceeded
```

That is the **third** Docker Hub timeout across this branch's runs, on a different test file each time (`engagement_test.py`, `test_type_test.py`, now `metrics_extended_test.py`). Every one of the 113 Selenium jobs pulls `mailhog` and `go-httpbin` from Docker Hub, so the suite plays that lottery 113 times per pull request. Re-run in flight.

That is worth a separate look — either pinning those two images into the build-and-artifact flow the other images already use, or reducing the number of jobs that pull them.

valentijnscholten:
nice that it now seems to work in github actions. in the past it wasn't stable there.

blakeaowens (approved):

Maffooch (approved):

devGregA:
The instability had two identifiable causes, so it's worth recording that "flaky on GitHub Actions" was never weather:

1. **Python 3.14 changed the default multiprocessing start method on Linux from `fork` to `forkserver`**, which Django's parallel runner cannot initialise workers for (`_init_worker` calls `django.setup()` only under `spawn`). Every worker died on `AppRegistryNotReady`. Fixed by forcing `spawn` for the `test` command — and by using `--parallel auto`, so a future default change degrades to serial instead of failing. Notably, an explicit `--parallel N` *bypasses* Django's own `get_max_test_processes()` guard; `auto` is what consults it.

2. **The unit-test celery broker was a sqlite file shared by every worker process.** Saving a `Product_Type` dispatches a task from a `post_save` signal, so workers raced to declare the queue — `UNIQUE constraint failed: kombu_queue.name` — and whichever lost took its test class down. Any test saving a `Product_Type` could draw that straw, which is why it looked random. Unit tests have no consumer (celeryworker/beat are reset in the override), so the broker is now kombu's in-process `memory://`.

Whatever the failures were when it was tried before, they were probably equally specific — parallelism doesn't create flakiness so much as surface shared state that was already there. The two above were the only ones the suite had left, and the tag quarantine (`non-parallel` / `transactional` phases) you all built since is what confined the rest.

Measured effect since merging: phase 1 760s → 324s, whole unit job 15.7–17.0 → 7.3–8.6 min, stable across every run on `bugfix` since.

PR #15506, opened 2026-08-04
ci: cache docker layers for the test image builds

devGregA:
## Description

Every pull request rebuilds all seven test images from scratch. Nothing in the expensive part of those builds changes between runs: the apt install, the `pip3 wheel` pass over `requirements.txt`, and the two pip installs are decided entirely by the Dockerfile and the requirements files. App code is copied in near the end, where it invalidates only cheap layers.

Enable buildx's GitHub Actions cache, scoped per image, os and platform.

`mode=max` is the part that matters. The costliest layer is the `pip3 wheel` in the `build` stage, which is an intermediate stage that never reaches the final image, and the default `mode=min` exports only the final image's layers — it would cache everything except the thing worth caching.

## Expected effect, stated conservatively

Baseline from run [30873644826](https://github.com/DefectDojo/django-DefectDojo/actions/runs/30873644826):

| Image | Build time |
|---|---|
| django, alpine | 3.4 min |
| django, debian | 3.1 min |
| nginx, alpine | 3.5 min |
| integration-tests, debian | 1.8 min |

Build is a hard `needs:` of every test job, so it is on the critical path of the whole suite — but it is also a parallel fan-out, so the saving to look for is in the slowest job, not the sum. Expect a couple of minutes off wall clock, not a transformation. This is worth doing because it is cheap and it compounds with everything else, not because it is dramatic on its own.

## Two decisions worth reviewing

**Only non-`pull_request` runs write to the cache.** A pull request from a fork has no write access to this repository's cache at all. Beyond that, the 10 GB per-repository budget should not be spent by every push to every branch: over budget, GitHub evicts least-recently-used, which is how you end up with a cache that is always cold for the runs that matter.

**Which leaves the question of who does write, and that is what `ci-warm-caches.yml` is for.** GitHub scopes a cache entry to the ref that saved it, and a run may restore only from its own ref, its base branch (`pull_request` events), and the default branch. So a pull request cannot leave behind an entry that another pull request can read — it saves into its own PR scope, which dies with it. Without a run triggered from the release line itself, "cache the docker layers" does nothing at all for the first push of a new branch, which is most branches most of the time.

The warm workflow calls the same reusable build workflow the suite uses, rather than carrying a copy of its build step. Drift between the two would be silent: it would warm a scope nothing reads, and every pull request would quietly go on paying for a full build.

Two new inputs keep the warming run from doing pointless work — it builds `django` only, where the costly layers are, and skips the artifact upload that has no consumer in a warming run. Both default to today's behaviour, so the existing caller in `unit-tests.yml` is unchanged.

Starting with `django` only also keeps this well inside the cache budget, so eviction never reaches the entries being warmed. `nginx` and `integration-tests` can be added once the footprint is observed.

## Verification

This pull request's own build jobs exercise the read path (`cache-from` with nothing to find, which must be a no-op rather than an error). The write path needs a non-pull_request run, so after merge the warm workflow should be dispatched once — `gh workflow run "CI: Warm Caches" --ref bugfix` — and the following pull request is the first that can hit. I will post before/after build timings.

### DIFF
--- a/.github/workflows/build-docker-images-for-testing.yml
+++ b/.github/workflows/build-docker-images-for-testing.yml
@@ -7,11 +7,28 @@ on:
         platform:
           type: string
           default: "linux/amd64"
+        docker-images:
+          type: string
+          default: '["django", "nginx", "integration-tests"]'
+        upload-artifacts:
+          type: boolean
+          default: true
   workflow_call:
     inputs:
         platform:
           type: string
           default: "linux/amd64"
+        # Which images to build, as a JSON array. Callers that only want to
+        # populate the layer cache (ci-warm-caches.yml) narrow this down.
+        docker-images:
+          type: string
+          default: '["django", "nginx", "integration-tests"]'
+        # The image tarballs are what the test jobs consume. A run that exists
+        # only to warm the layer cache has no consumer, and the upload is the
+        # slowest step in the job, so it can be skipped.
+        upload-artifacts:
+          type: boolean
+          default: true
 
 jobs:
   build:
@@ -20,7 +37,7 @@ jobs:
     strategy:
       matrix:
         # integration tests are only build (and run) on debian linux/amd64
-        docker-image: [django, nginx, integration-tests]
+        docker-image: ${{ fromJSON(inputs.docker-images) }}
         os: [alpine, debian]
         platform: ["${{ inputs.platform }}"]
         exclude:
@@ -37,6 +54,7 @@ jobs:
         run: |
           platform=${{ inputs.platform }}
           echo "PLATFORM=${platform//\//-}" >> $GITHUB_ENV
+          echo "BUILD_CACHE_SCOPE=${{ matrix.docker-image }}-${{ matrix.os }}-${platform//\//-}" >> $GITHUB_ENV
           echo $GITHUB_ENV
 
       - name: Checkout
@@ -63,9 +81,22 @@ jobs:
           tags: defectdojo/defectdojo-${{ matrix.docker-image }}:${{ matrix.os }},${{ env.IMAGE_REPOSITORY }}/defectdojo-${{ matrix.docker-image }}:${{ matrix.os }}
           file: Dockerfile.${{ matrix.docker-image }}-${{ matrix.os }}
           outputs: type=docker,dest=${{ matrix.docker-image }}-${{ matrix.os }}-${{ env.PLATFORM }}_img
+          # A miss costs nothing but the lookup, so this is unconditional.
+          cache-from: type=gha,scope=${{ env.BUILD_CACHE_SCOPE }}
+          # mode=max is what makes this worth doing: the expensive layer is the
+          # `pip3 wheel` in the `build` stage, which never reaches the final
+          # image, and the default mode=min would not export it.
+          #
+          # Only non-pull_request runs write. Two reasons: the 10 GB per-repo
+          # cache budget should not be spent by every push to every branch (and
+          # LRU eviction would then throw away the entries that are actually
+          # read), and a pull request from a fork has no write access to this
+          # repository's cache in the first place.
+          cache-to: ${{ github.event_name != 'pull_request' && format('type=gha,mode=max,scope={0}', env.BUILD_CACHE_SCOPE) || '' }}
 
       # export docker images to be used in next jobs below
       - name: Upload image ${{ matrix.docker-image }} as artifact
+        if: inputs.upload-artifacts
         timeout-minutes: 15
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
--- a/.github/workflows/ci-warm-caches.yml
+++ b/.github/workflows/ci-warm-caches.yml
@@ -1,0 +1,70 @@
+name: "CI: Warm Caches"
+
+# Why this workflow exists at all.
+#
+# GitHub scopes every Actions cache entry to the ref that saved it. A run can
+# restore entries saved by its own ref, by its base branch (pull_request events
+# only), and by the repository's default branch -- nothing else. So a pull
+# request cannot leave behind a cache that any *other* pull request can use: it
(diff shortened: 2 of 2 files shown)


valentijnscholten (commented):
First, on the premise: I went looking for whether the hierarchical cache scoping makes `ci-warm-caches.yml` unnecessary, and it doesn't. The hierarchy applies to **reads** only — a run restores from its own ref, its base branch (`pull_request` events), and the default branch — while writes are flat: a PR run saves into `refs/pull/N/merge`, and that scope dies with the PR. Since `unit-tests.yml` triggers only on `pull_request`/`workflow_dispatch`, and `release-x-manual-docker-containers.yml` is `workflow_dispatch`/`workflow_call` only, there is currently no run in this repo that writes into a scope a pull request is allowed to read. So the warming workflow is doing real work, not duplicating a built-in.

Three things I think need changing.

### 1. It warms every build except the slowest one, so the expected wall-clock saving is close to zero

The baseline table has nginx/alpine at 3.5 min, the slowest of the four, and the description correctly argues that because the build is a parallel fan-out the saving to look for is in the slowest job rather than the sum. But `docker-images: '["django"]'` excludes exactly that job, justified as "nginx and integration-tests are comparatively cheap" — which the table contradicts. If django drops to near-nothing and nginx/alpine still takes 3.5 min, the build phase still takes 3.5 min.

Giving them a shared scope wouldn't fix it either. The comment at the top of both alpine Dockerfiles says the build stages should be identical so the Docker cache can be shared, but they have drifted: `Dockerfile.nginx-alpine` does `COPY requirements.txt requirements-dev.txt ./` followed by two `pip3 wheel` runs, where `Dockerfile.django-alpine` copies only `requirements.txt` and runs one. The `apk` block above is identical, but the differing `COPY` changes the layer chain, so nginx's wheel layers can never hit django's. Adding `nginx` to `docker-images` is the small fix; realigning those two build stages is a separate and larger win.

### 2. The nightly cron cannot refresh `dev` or `bugfix`

Scheduled workflows only ever run against the latest commit of the **default branch**, which is `master` here. So the `schedule:` trigger runs with ref `refs/heads/master` and refreshes the master scope, not the release-line scopes the comment says it is protecting from the 7-day eviction. And because this PR is based on `bugfix`, the cron will not fire at all until the file reaches master through a release merge, so the mechanism is inert for the first few weeks after merge.

There is an upside hiding in that, though: the default-branch scope is readable by every run regardless of base, so warming master is the highest-leverage single thing this workflow could do. Warming `dev` and `bugfix` on push still earns its keep, because `requirements.txt` genuinely diverges between the release lines and a master-scoped wheel layer would miss — but that is the real argument for it, and it isn't the one the comment makes.

### 3. It creates a second, duplicate cache namespace for the same layers

`release-x-manual-docker-containers.yml` already uses `type=gha` with scope `${image}-${os}-${platform}-${{ github.head_ref || github.ref_name }}`. Putting the branch name inside the buildx scope is redundant with GitHub's own per-ref isolation and defeats the default-branch fallback. This PR sensibly omits it, but the two schemes now form separate namespaces holding the same layers, which doubles the footprint against the 10 GB per-repo budget — and that budget is what the description is trying to stay inside, since LRU eviction would discard precisely the entries meant to be read. Dropping the branch suffix from the release workflow would unify them.

### Smaller notes

- `timeout-minutes: 15` on the Build step now also has to cover the `mode=max` export on write runs. If it ever trips, the cache silently never gets written and it looks like a flaky build rather than a cache that stopped working.
- Footprint is worth measuring rather than asserting: django alone is four warmed scopes (alpine/debian × amd64/arm64), six once nginx is added, alongside the release workflow's separate scopes and the two `actions/cache` users in `gh-pages.yml` and `validate_docs_build.yml`.
- Two things I checked that are **not** problems: `cache-to: ''` on PR runs is safe, because `build-push-action` parses an empty input as an empty list and omits the flag; and the missing `secrets: inherit` in the warm workflow is harmless, since the build workflow uses no secrets (`IMAGE_REPOSITORY` is derived from `github.repository`).

devGregA:
All three findings verified independently — no argument with any of them, and the first one invalidates this PR's own headline claim.

1. **Warming everything except the slowest build is a ~0 wall-clock change.** My own baseline table says nginx/alpine is the slowest at 3.5 min, and `docker-images: '["django"]'` excludes exactly it. The build phase is a parallel fan-out, so it still takes 3.5 min however fast django gets. The justification comment ("nginx and integration-tests are comparatively cheap") is contradicted by the table two paragraphs above it.

2. **The cron cannot refresh release-line scopes.** Scheduled workflows run against the default branch only, so the `schedule:` trigger would warm master's scope, not `bugfix`/`dev` — and wouldn't fire at all until this file reaches master through a release. The eviction-protection story in the comment was wrong. (Your observation that master-scope warming is the *highest-leverage* variant, since every run can read the default branch's scope, is the right design and needs a file on master — a separate change.)

3. **The duplicate namespace is real.** `release-x-manual-docker-containers.yml` already writes `type=gha` cache with a `-${{ github.head_ref || github.ref_name }}` suffix; that suffix duplicates GitHub's own per-ref scoping and defeats the default-branch fallback, and my un-suffixed scheme doubles the footprint against the same 10 GB budget. Dropping the suffix there unifies the namespaces.

Also taking the buried lead in your point 1: the two alpine build stages have drifted (`COPY requirements.txt requirements-dev.txt` + two wheel runs in nginx vs one in django), which both breaks the "identical so the cache can be shared" comment at the top of those Dockerfiles and explains why nginx is the slowest build. Realigning them is worth more than any cache tuning.

Closing this PR rather than reworking it in place: every one of the three fixes moves where the work happens (nginx into the warm set, the cron into a master-scoped file, the suffix into the release workflow), so the replacement shares almost no diff with this branch. The redo will come as: (a) release-workflow suffix drop + PR-side `cache-from` for all images, (b) the warm workflow rebuilt around your master-scope observation, (c) the alpine build-stage realignment as its own change. Context that changed underneath this PR meanwhile: #15524 cut the suite to ~38 jobs and #15528 tiers PR runs to amd64-only, so the warm set for the light tier is smaller than when this was written.

PR #15507, opened 2026-08-04
ci: restore the migrated database instead of migrating it four times

devGregA:
## Description

> **Stacked on #15506.** This branch is #15506 plus one commit, because it adds a job to the `ci-warm-caches.yml` that #15506 introduces. It targets `bugfix` rather than #15506's branch so that the unit-test workflow actually runs against it — `unit-tests.yml` only triggers for pull requests into `master`, `dev`, `bugfix` and the release lines, and the mechanism here is the kind that has to be verified by CI rather than by reading. **Review the second commit only**; the first is #15506 unchanged. Merge #15506 first and this reduces to that one commit.

Every unit test job starts by applying all 264 migrations to an empty database. Measured on run [30873644826](https://github.com/DefectDojo/django-DefectDojo/actions/runs/30873644826) (job 91881121251) that is **98 seconds**, and it happens four times per pull request — amd64 and arm64, each with `DD_V3_FEATURE_LOCATIONS` on and off — for a result that is identical every time.

Snapshot it instead: `pg_dump --format=custom` after the migrate, cached under a key describing the migration state, and `pg_restore` on the way in.

This works because of something the entrypoint already does. It pre-migrates the database outside the test runner and passes `--keepdb`, so the runner logs `Using existing test database for alias 'default'` and does not migrate again. A restored schema satisfies it in exactly the same way.

## The key hashes models, not just migrations

`requirements.txt`, `dojo/db_migrations/**`, `dojo/models.py`, `dojo/base_models/**`, `dojo/pghistory_models.py`, `dojo/settings/settings.dist.py`, both compose files, and the entrypoint.

Including the **models** is what makes it safe for the entrypoint to skip its `makemigrations --check` guard on a hit: a model change that lacks a migration cannot exist without changing the key, and a changed key means there is no snapshot and the job takes the full guarded path. A snapshot keyed on migrations alone would quietly disable that guard, so the reasoning is written into the script where the skip happens.

`requirements.txt` is in there because third-party apps bring their own migrations.

## Failure modes, and what happens

| If | Then |
|---|---|
| The snapshot is bad or truncated | The restore runs `--single-transaction`, so it either applies completely or rolls back to an empty database, and the miss path takes over. No half-restored schema can reach the tests. |
| `pg_restore` version skew | Both dump and restore run inside the postgres container, whose client matches its own server. The runner's client is never involved — which matters, since the server is 18 and `pg_dump` refuses to dump a server newer than itself. |
| Four jobs race to save the same key | Three lose, which the action reports as a warning, so the step tolerates its own failure. |
| The tests fail | The snapshot is already saved: it is a function of the migration state, not of whether the suite passed. |

Both paths converge before the tests run — restored, or migrated by the prepare step — so the test step always runs with `DD_TEST_DB_MODE=reuse` and there is one code path through it rather than two.

## Why per-job rather than one shared prepare job

A dedicated `prepare-db` job would migrate once on a cold key instead of four times, saving ~5 runner-minutes — but every run, including the warm ones that are the common case, would wait for it before any test job could start. Doing it per job costs nothing on the warm path, which is the path that matters, and adds no cross-job dependency.

## Warming

Snapshots are warmed on the release lines for the same cache-scoping reason as #15506: a pull request cannot leave behind an entry another pull request can read, so without the warm job the first push of a branch always misses. The warm job probes with `lookup-only` first, so the daily run costs seconds unless a merge actually changed the migration state — in which case rebuilding the snapshot doubles as a check that the release line still migrates from empty.

## Expected effect

About 100 seconds off each of four jobs on a hit — roughly 6 runner-minutes per pull request, and rather less in wall clock since the four run concurrently.

Worth being straight about the ratio: this is a modest win next to what the same mechanism is worth on the integration-test matrix, where the initializer re-derives the same database **113 times** at 126 seconds each. I scoped this to the unit tests deliberately, as the smaller and safer place to prove the mechanism; the initializer version is the follow-up, and it needs the key to hash the fixtures too.

## Verification

Local: `docker compose config` confirms `DD_TEST_DB_MODE` reaches the container (`prepare` when set, `full` when unset, so existing behaviour is unchanged); all three modes take the intended branches; shellcheck v0.11.0 clean with the repo's `SHELLCHECK_OPTS`; all four copies of the cache key are byte-identical.

CI is what verifies the round trip. This run is a cold key: it should migrate, dump, and save. A second push should restore, log `Using existing test database`, and finish ~100 seconds faster. I will post both.

### DIFF
--- a/.github/workflows/ci-warm-caches.yml
+++ b/.github/workflows/ci-warm-caches.yml
@@ -1,0 +1,132 @@
+name: "CI: Warm Caches"
+
+# Why this workflow exists at all.
+#
+# GitHub scopes every Actions cache entry to the ref that saved it. A run can
+# restore entries saved by its own ref, by its base branch (pull_request events
+# only), and by the repository's default branch -- nothing else. So a pull
+# request cannot leave behind a cache that any *other* pull request can use: it
+# saves into its own PR scope, and that scope dies with the PR.
+#
+# Without a workflow like this one, caching the migrated database would therefore
+# do nothing at all for the first push of a new branch, which is most branches
+# most of the time. Something has to save an entry into a scope pull requests are
+# allowed to read, and only a run triggered from a release line can do that.
+#
+# Note what is deliberately absent: a `schedule:` trigger. Scheduled workflows
+# only ever run against the default branch, so a cron here would run with ref
+# refs/heads/master and warm the master scope -- not the release-line scopes this
+# is for. Eviction is handled instead by reads: GitHub drops an entry it has not
+# seen used for 7 days, and every pull request based on a release line reads this
+# one, which keeps it alive without a timer.
+#
+# Warming the default-branch scope is worth doing too, since every run can read
+# it regardless of base, but that needs a workflow file on master and is a
+# separate change.
+
+on:
+  push:
+    branches:
+      - bugfix
+      - dev
+    # Exactly the paths the snapshot key hashes -- anything else leaves the key
+    # unchanged, so there would be nothing to warm.
+    paths:
+      - 'requirements.txt'
+      - 'dojo/db_migrations/**'
+      - 'dojo/models.py'
+      - 'dojo/base_models/**'
+      - 'dojo/pghistory_models.py'
+      - 'dojo/settings/settings.dist.py'
+      - 'docker-compose.yml'
+      - 'docker-compose.override.unit_tests_cicd.yml'
+      - 'docker/entrypoint-unit-tests.sh'
+      - '.github/workflows/ci-warm-caches.yml'
+  workflow_dispatch:
+
+concurrency:
+  # Two pushes in quick succession would warm the same key twice.
+  group: warm-caches-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  warm-migrated-db:
+    name: Warm Migrated Database Snapshot
+    # A fork has its own cache scope and its own billing, so there is nothing
+    # here for it to warm. Same guard as release-nightly-dev.yml.
+    if: github.repository == 'DefectDojo/django-DefectDojo'
+    runs-on: ubuntu-latest
+    env:
+      DJANGO_VERSION: debian
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      # KEEP THIS KEY IN SYNC with the two copies in rest-framework-tests.yml.
+      # If they drift, this job warms an entry nothing reads and every pull
+      # request goes on paying the full migrate, with nothing failing to say so.
+      - name: Probe for an existing snapshot
+        id: probe
+        uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: migrated-db.dump
+          key: migrated-db-v1-${{ hashFiles('requirements.txt', 'dojo/db_migrations/**', 'dojo/models.py', 'dojo/base_models/**', 'dojo/pghistory_models.py', 'dojo/settings/settings.dist.py', 'docker-compose.yml', 'docker-compose.override.unit_tests_cicd.yml', 'docker/entrypoint-unit-tests.sh') }}
+          # Do not download it -- whether the key exists is the whole question,
+          # and on most runs the answer is usually yes.
+          lookup-only: true
+
+      - name: Set unit-test mode
+        if: steps.probe.outputs.cache-hit != 'true'
+        run: docker/setEnv.sh unit_tests_cicd
+
+      # Built here rather than pulled from the build job's artifacts: this needs
+      # one image on one platform, and building it directly keeps the job
+      # self-contained instead of coupling it to that job's upload.
+      - name: Build the unit-test image
+        if: steps.probe.outputs.cache-hit != 'true'
+        timeout-minutes: 30
+        run: docker compose build uwsgi
+
+      - name: Start Postgres
+        if: steps.probe.outputs.cache-hit != 'true'
+        run: docker compose up --no-deps -d postgres
+
+      - name: Wait for Postgres to accept connections
+        if: steps.probe.outputs.cache-hit != 'true'
+        run: |
+          for _ in $(seq 1 60); do
+            if docker compose exec -T postgres pg_isready -q -U defectdojo -d test_defectdojo; then
+              exit 0
+            fi
+            sleep 2
+          done
+          echo "Postgres did not become ready in 120s" >&2
+          exit 1
+
+      - name: Migrate and dump
+        if: steps.probe.outputs.cache-hit != 'true'
+        timeout-minutes: 20
+        run: |
+          docker compose run --rm -T --no-deps uwsgi
+          docker compose exec -T postgres pg_dump \
+            --username defectdojo --dbname test_defectdojo --format=custom > migrated-db.dump
+          ls -lh migrated-db.dump
+        env:
+          DD_TEST_DB_MODE: prepare
+
+      - name: Save the snapshot
+        if: steps.probe.outputs.cache-hit != 'true'
+        uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: migrated-db.dump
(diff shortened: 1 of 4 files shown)


devGregA:
## Local verification of the round trip

CI's first run only exercises the cold path (migrate, dump, save), so I ran the restore half locally against the real `unit_tests_cicd` stack — postgres 18.4, the `django-unittests` image, the same commands the workflow issues.

**1. `prepare` mode does exactly its share of the work and no more:**

```
spectacular ran?    0   (expected 0 — skipped in prepare)
makemigrations?     1   ("No changes detected")
migrations applied: 180
tests ran?          0   (expected 0)
Database migrated; DD_TEST_DB_MODE=prepare, so not running any tests.
PREPARE_EXIT=0
```

**2. Dump, then wipe the database completely, then restore with the workflow's exact flags:**

| | |
|---|---|
| tables before dump | 205 |
| `pg_dump --format=custom` | 2.4 s, **956 KB** |
| after `drop database` / `create database` | 0 tables |
| `pg_restore --no-owner --no-privileges --single-transaction` | **11.1 s**, exit 0 |
| tables after restore | 205 |
| `django_migrations` rows | 355 |

**3. The restored database satisfies `--keepdb`** — this is the assumption the whole change rests on, so it is the one worth seeing directly:

```
Preserving test database for alias 'default' ('test_defectdojo')...
Running migrations:
  No migrations to apply.
...
Ran 11 tests in 10.670s
OK
```

Django found the schema complete and applied nothing. Tests then passed against the restored dump.

Two things worth recording from this:

- **11 seconds to restore against 98 seconds to migrate**, so the saving is ~87 s per job rather than the full 98.
- **The snapshot is 956 KB.** I had expected single-digit megabytes. It is schema plus `django_migrations` and nothing else, because the unit-test stack resets the initializer and so seeds no data. That makes the cache footprint here irrelevant next to the docker layer cache in #15506, and it means eviction pressure is not a concern for this entry.

devGregA:
## The mechanism worked and cached nothing

Both runs of this branch restored a miss, migrated, and saved. Green both times. The reason is in a single job's own log:

```
restore  key: migrated-db-v1-99a364db70405620e4eba05edb01240d9ba47604...
save     key: migrated-db-v1-232bdb3eeac1b64c1bea1df5926ad11892f5447...
```

Two different keys, same job, same commit.

`hashFiles` is evaluated per step, and `dojo/db_migrations/**` hashes whatever is on disk when it runs. Migrating imports every migration module, which writes `__pycache__/*.pyc` beside them — and the repository is bind mounted into the container, so those files land in the runner's workspace. The save computed a key describing a tree with 290 more files in it than the tree the restore had asked about.

So the entry was written perfectly, under a key nothing would ever look up. Every run paid the migrate and left behind another orphan.

**Fixed** by computing the key once, straight after checkout and before anything runs Python, and referencing that step's output from both the restore and the save. Same shape, same fix, in `ci-warm-caches.yml` — probe before the build, save after the migrate.

This also retires a comment I had written into both files. "Keep the two copies byte-identical" was the wrong discipline: the copies **were** identical. Identical expressions still produced different values, because the thing they hash changed underneath them. One computation, referenced twice, is the only version of this that cannot drift.

### Worth dwelling on how this presented

The first run reported 153/153 green, every step `success`, and `Cache saved with key ...` in the log. Nothing failed. The feature simply did not work.

The only reason it surfaced is that I re-ran the same commit specifically to watch the warm path, and the timings came back at 16.0–17.1 min — indistinguishable from baseline. Had I merged on the strength of the green tick and the reasoning, this would have sat in the repository indefinitely: no failures, no speedup, and a plausible-looking cache step in every log to suggest otherwise.

Next run on this branch is the real test. Cold key again (the key changes with this commit), so the run after it is the one whose numbers mean anything.

devGregA:
## Verified: the warm path hits, and it is faster

Steps from the warm run (run 30893426071):

```
success  Compute migrated database snapshot key
success  Restore migrated database snapshot
success  Restore snapshot into Postgres      <- ran
skipped  Migrate and snapshot the database   <- no migrate
skipped  Save migrated database snapshot
success  Unit tests
```

And the clock:

| Job | Baseline | Cold (miss) | **Warm (hit)** |
|---|---|---|---|
| amd64, v3=false | 15.9 | 18.48 | **13.30** |
| amd64, v3=true | 17.0 | 17.60 | **15.67** |
| arm64, v3=true | 15.7 | 16.12 | **14.23** |
| arm64, v3=false | 15.7 | 16.63 | **13.62** |
| total | ~64 runner-min | ~66 | **56.8** |

So about 1.5–2.5 minutes off each job, ~7 runner-minutes per pull request. That is the ~100 seconds of migrate plus the skipped `makemigrations` guard, less ~11 seconds to restore.

The cold column is the honest cost of the other case: a run that has to build the snapshot is about a minute slower than doing nothing, because it pays the dump and the upload on top of the migrate. That happens only when the migration state actually changes.

### What this run also settles

The key is now stable across the job — one value, `99a364db…`, used by the restore and the save alike, where before there were two. The `__pycache__` pollution is gone because the key is computed before anything runs Python.

Ready for review. Merge #15506 first and this reduces to its second commit.

One note on sequencing with #15505, which halves these same jobs by parallelising phase 1: the two compose rather than overlap. #15505 attacks the 12.7 minutes of test execution, this attacks the 100 seconds of migrate in front of it. Together the unit job should land in the 6–7 minute range, from 15.7–17.0 today.

valentijnscholten:
Can you give a summary of what happened and the net win? In the past I decided not to make this change as the 4 jobs were running in parallel anyway so preparing a DB was not "supposed to" speed things up. But it kind of sounds some benefit was made?

devGregA:
Good question, and worth answering carefully because **your original reasoning was correct** — this change doesn't beat it, it wins on different axes.

## Your old analysis, restated

The four unit jobs run concurrently, each migrating its own postgres in parallel. Deduplicating the migrate can't shorten the *phase*, because no job waits on another's migrate. That's true, and it's still true. If wall clock of the unit phase under unconstrained concurrency were the only prize, this PR wouldn't be worth its complexity.

## What actually changed, measured

**1. Per-job latency.** Every job individually pays migrate (98s) + the `makemigrations --check` guard. A warm restore is 11s and lets the entrypoint skip the guard safely (the key hashes models as well as migrations, so a missing-migration defect can't exist on a hit). Measured on this PR:

| | baseline | warm hit |
|---|---|---|
| per unit job | 15.7–17.0 min | **13.3–15.7 min** |
| unit phase runner-min | ~64 | **~57** |

All four improve in parallel, so the unit-phase *wall* also drops ~2 min — not because the migrate was deduplicated, but because each copy of it got 87s cheaper.

**2. The proportion changed since you last evaluated this.** After #15505 the unit job is ~7–8 min, so the same ~1.5–2 min saving is now **~20–25% of the job**, not ~10%. Decisions made when the job was 16 min deserve re-checking at 8.

**3. Runner-minutes stopped being free-ish this week.** The 128-job fan-out was saturating org concurrency — two PRs in flight stalled every other run, twice. In a concurrency-constrained world, runner-minutes anywhere convert to queue-drain time everywhere, which is a thing your 4-parallel-jobs analysis correctly ignored back when the org wasn't hitting the ceiling. (#15524's grouping removed most of that pressure; this PR's ~7 min/push is a smaller second helping.)

**4. The forward-looking reason: the snapshot is a building block.** The Selenium jobs pay a 126-second initializer per stack — post-grouping that's still ~29 min/PR of repeated identical work. Restoring an *initialized* snapshot the same way is the natural next step, and this PR is where the mechanism (key discipline, warm workflow, restore-else-prepare fallback) gets proven on the simpler case first.

## What happened along the way (the "summary of what happened" part)

The first version was green and cached nothing: `hashFiles` is evaluated per step, and `migrate` writes `__pycache__/*.pyc` into the bind-mounted `dojo/db_migrations/**`, so the save key differed from the restore key within the same job — every run saved an orphan. 153/153 green, zero effect; only re-running the same commit and seeing baseline timings exposed it. The fix (and the lesson now recorded in both workflow files): compute a cache key once, in its own step, before anything runs Python, and reference that output from both restore and save. "Keep the two inline copies identical" was the wrong discipline — the copies *were* identical and still produced different values.

Honest cost to weigh: a cold-key run (migration state changed) is ~1 min *slower* than baseline, because it pays the dump and upload on top of the migrate. That's rare and self-limiting, but it's real.

Net: ~7 runner-min and ~2 min of unit wall per PR at current shape — modest, honestly below where it ranked when this branch started — plus the proven mechanism for the initializer snapshot, which is worth roughly 4× more.

PR #15515, opened 2026-08-04
fix(dedupe): stop prefetching the deprecated endpoint relation under V3

valentijnscholten:
## Summary

- The move to Locations does not delete `Endpoint` rows, and `Endpoint.__init__` raises `NotImplementedError` once `V3_FEATURE_LOCATIONS` is on (see `Endpoint.allow_endpoint_init`). Any queryset that prefetches the `endpoints` relation therefore hydrates the deprecated model for every surviving row and raises.
- `get_finding_models_for_deduplication` prefetched `endpoints` unconditionally. That put the crash on the batch-deduplication path every import and reimport takes (`post_process_findings_batch`) and on `manage.py dedupe` in batch mode. A fresh V3 instance has no endpoint rows to hydrate, so nothing surfaces it — it only fires on an instance that migrated from endpoints, which is why it went unnoticed.
- Adds `location_prefetch_lookups(prefix="")` to `dojo/location/queries.py`, returning the prefetch lookup for the location model actually in use, and routes the dedupe and rehash querysets through it. This replaces three separate spellings of the same decision: the unconditional prefetch above, the inline `V3_FEATURE_LOCATIONS` branch in `build_candidate_scope_queryset`, and the two duplicated `select_related`/`prefetch_related` blocks in the dedupe command.
- The command's V3 branch previously prefetched `locations` (one hop); it now prefetches `locations__location__url` like the other call sites, since `get_locations()` reads `location.url`.
- `unittests/test_dedupe_with_endpoints_v3.py` covers it, joining the existing `*_endpoints_v3.py` regression family. Verified it catches the original defect: restoring the `endpoints` prefetch fails `test_batch_dedupe_loader_survives_legacy_endpoint_rows` with `NotImplementedError: Endpoint model is deprecated when V3_FEATURE_LOCATIONS is enabled`.
- The `prefix` argument exists because the Pro Tuner's rehash tasks page `EnhancedFinding` and reach the finding through a relation; a follow-up on the Pro side replaces its local copy of this helper with this one.

### DIFF
--- a/dojo/finding/deduplication.py
+++ b/dojo/finding/deduplication.py
@@ -8,6 +8,7 @@ from django.db.models import Prefetch
 from django.db.models.query_utils import Q
 
 from dojo.celery import app
+from dojo.location.queries import location_prefetch_lookups
 from dojo.models import Endpoint_Status, Finding, System_Settings
 from dojo.vulnerability.queries import vulnerability_id_prefetch
 
@@ -18,7 +19,7 @@ deduplicationLogger = logging.getLogger("dojo.specific-loggers.deduplication")
 def get_finding_models_for_deduplication(finding_ids):
     """
     Load findings with optimal prefetching for deduplication operations.
-    This avoids N+1 queries when accessing test, engagement, product, endpoints, and original_finding.
+    This avoids N+1 queries when accessing test, engagement, product, locations, and original_finding.
 
     Args:
         finding_ids: A list of Finding IDs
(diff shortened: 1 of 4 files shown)


valentijnscholten:
Closing as redundant: #15508 already fixes the crash in `get_finding_models_for_deduplication` on the same base branch, and does it more completely — it also updates the query-count expectations in `test_importers_performance.py` and `test_tag_inheritance_perf.py`, which this PR does not, so this one would fail CI on exactly those tests.

What was not duplicated here, and is worth doing separately on the minor line rather than on `bugfix`:

- a shared `location_prefetch_lookups()` helper, so the choice of location lookup has one definition instead of the three it now has — the new ternary in #15508, the `V3_FEATURE_LOCATIONS` branch in `build_candidate_scope_queryset`, and the two duplicated `select_related`/`prefetch_related` blocks in `manage.py dedupe`
- `manage.py dedupe` prefetches `locations` (one hop) where the other call sites use `locations__location__url`; `get_locations()` reads `location.url`, so the one-hop version leaves a query per location reference

The branch is left in place for that follow-up.

PR #15524, opened 2026-08-04
ci: run the Selenium matrix as 10 file groups, not 113 single-file jobs

devGregA:
## The problem, measured

The integration matrix is 56 files × 2 feature-flag values with one compose stack per file: **113 jobs per pull request**. From a green run's job logs, each job spends ~3.5 minutes on fixed cost — artifact download, `docker load` (42s), boot, a **126-second initializer** — and a median of ~42 seconds actually running tests. That is **83% overhead**, ~463 of the suite's ~565 runner-minutes.

The second-order cost is worse than the minutes: the suite fans out to 128 jobs against a concurrency allowance of a few dozen, so **two pull requests in flight stall every other run in the organization**. That was observed directly, twice, this week.

## The change

The same 56 files regrouped into **10 groups of up to six**, run sequentially in one stack per group by `entrypoint-integration-tests.sh` (which now accepts a whitespace-separated list — a loop around exactly the code that ran one file).

| | before | after |
|---|---|---|
| Selenium jobs per PR | 113 | **23** |
| suite total jobs per PR | 128 | **38** |
| Selenium fixed cost | ~396 runner-min | ~80 runner-min |
| suite runner-minutes | ~565 | **~250** |
| concurrency waves to drain | ~6 | ~2 |

Total test time is unchanged — the same tests run in the same order.

## Why this grouping is safe rather than hopeful

The entrypoint's no-filter branch (the `else` at the bottom) has always run **62 files sequentially in one stack** — that is the original, still-maintained execution path. The groups are **contiguous slices of exactly that order**. Every adjacency inside every group already has years of history running back-to-back in a *shared* stack; this change only adds fresh stacks at nine cut points where the legacy path had none.

So the isolation argument runs one way: strictly **more** isolation than the code path that has always existed, never less. Nobody has to reason about whether `dedupe_test` tolerates following `file_test` — it has done so, in the same database, since those files were written.

Two rules for future edits, recorded as comments in the matrix: appending a **new** file to the last group (or its own group) is always safe; reordering files **across** groups is the only thing that needs thought.

## Deliberate details

- **`endpoint_test.py` stays alone** — the `v3_feature_locations: true` exclude matches the exact entry string, so it cannot ride inside a group.
- **Groups fail fast**, matching the legacy branch: a failed UI test can leave state that cascades into everything after it in the stack, and one clean failure names the culprit where five cascading ones bury it. The failing file is always the last `Running:` line in the log, and failure artifacts remain per-group.
- **Test-step timeout 10 → 30 minutes.** The largest groups sum to ~11 minutes of test time on a healthy stack (working back from per-job durations: `finding_test` alone is ~5 minutes of tests). The old 10 would have tripped on group 1 on a normal day. Single-file entries are unaffected; it is a backstop, not a target.
- **Check names get long** (each lists its group's files). None are required contexts — `unit-tests-complete` gates on the workflow result — so nothing depends on the names. If the length offends, short labels can come later; they'd add parsing to the entrypoint for a cosmetic win.
- `tests/zap.py` stays disabled, exactly as it was in the per-file matrix.

## Verification

- New matrix's file set proven **byte-identical** to the old one: set-diff none lost, none gained; 12 entries × 2 values − 1 exclude = 23 jobs.
- Entrypoint loop exercised against a stubbed `python3`: correct order, fail-fast on a middle failure, single-file behaviour unchanged.
- `shellcheck`, `actionlint 1.7.7`, `bash -n`: clean.
- This PR's own run is the live proof: watch the group jobs' durations against the ~4.5-minute single-file baseline, and the whole-suite wall clock.

## Merge-order note

#15522 touches a different hunk of the same workflow (retrying the two Docker Hub pulls). They compose — and this change also *reduces* that flake's surface from 113 pull-lotteries per PR to 23. Merging #15522 first is cleanest; if this lands first, #15522 rebases trivially.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -12,63 +12,37 @@ jobs:
       matrix:
         test-case: [
           "openapi-validatator",
-          "tests/action_history_test.py",
-          "tests/alerts_test.py",
-          "tests/announcement_banner_test.py",
-          "tests/banner_test.py",
-          "tests/base_test_class.py",
-          "tests/benchmark_test.py",
-          "tests/calendar_test.py",
-          "tests/check_various_pages.py",
-          "tests/close_old_findings_dedupe_test.py",
-          "tests/close_old_findings_test.py",
-          "tests/dashboard_test.py",
-          "tests/dedupe_test.py",
-          "tests/endpoint_extended_test.py",
+          # endpoint_test.py stays alone so the v3 exclude below still matches it.
           "tests/endpoint_test.py",
-          "tests/engagement_checklist_test.py",
-          "tests/engagement_export_test.py",
-          "tests/engagement_extended_test.py",
-          "tests/engagement_presets_test.py",
-          "tests/engagement_test.py",
-          "tests/environment_test.py",
-          "tests/false_positive_history_test.py",
-          "tests/file_test.py",
-          "tests/finding_extended_test.py",
-          "tests/finding_group_test.py",
-          "tests/finding_test.py",
-          "tests/login_test.py",
-          "tests/metrics_extended_test.py",
-          "tests/note_type_test.py",
-          "tests/notes_test.py",
-          "tests/notification_webhook_test.py",
-          "tests/notifications_test.py",
-          "tests/object_test.py",
-          "tests/product_member_test.py",
-          "tests/product_metadata_test.py",
-          "tests/product_tag_metrics_test.py",
-          "tests/product_test.py",
-          "tests/product_type_member_test.py",
-          "tests/product_type_test.py",
-          "tests/questionnaire_advanced_test.py",
-          "tests/questionnaire_test.py",
-          "tests/regulations_test.py",
-          "tests/reimport_scan_test.py",
-          "tests/report_builder_test.py",
-          "tests/risk_acceptance_test.py",
-          "tests/search_test.py",
-          "tests/sla_configuration_test.py",
-          "tests/system_settings_test.py",
-          "tests/test_copy_test.py",
-          "tests/test_test.py",
-          "tests/test_type_test.py",
-          "tests/threat_model_test.py",
-          "tests/tool_config.py",
-          "tests/tool_product_test.py",
-          "tests/tool_type_test.py",
-          "tests/user_profile_test.py",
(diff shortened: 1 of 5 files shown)


devGregA:
## Group 3's failure: root-caused, and it was not an order dependency

`CloseOldDedupeTest.test_delete_findings` failed with `NoSuchElementException: [id="no_findings"]`. Tracing it back:

**`ImportAsyncWaitApiTest` leaks ~100 findings.** The two async-dedupe API tests (added in July, #15007) each create a product via `_make_engagement` and import 25 findings into it twice — and `_make_engagement` returns only the engagement id, discarding the product's. Nothing deletes any of it, and those tests run *after* the suite's own `ProductTest("test_delete_product")` cleanup. The file exits leaving two products and ~100 findings behind.

Being newer than the per-file matrix, no test file had ever run after them in a shared stack — until group 3 did, on this PR's first run.

**`test_delete_findings` can't cope with more than a page.** It is explicitly built for the shared-stack world ("removing previous findings..."), but it bulk-deletes only the page it is on, then asserts the global list is empty. With 100 leftovers, the assertion is unsatisfiable — this was a deterministic failure, not a flake. Its post-delete check also probes the old DOM with only the 1-second implicit wait, which a full-page cascade delete can outrun.

**Fixes pushed** (`5045a87da`):
- `ImportAsyncWaitApiTest` registers an `addCleanup` deleting each product it creates (cascades engagement + findings; fires even when the test fails). This restores the invariant the rest of the suite already maintains.
- `test_delete_findings` (both `close_old` copies) now deletes page by page until the empty state appears — bounded at 10 pages so a non-shrinking delete fails loudly — and waits for the page reload after each round instead of racing the redirect. `dedupe_test`'s own copy is deliberately untouched: it runs seven times inside that file between small scenarios and never crosses a page boundary.

**The honest correction to this PR's safety argument:** "every adjacency has years of shared-stack history" is true only for files older than the per-file matrix. Tests added since have none — this was the first such test, and it surfaced on the grouping's first run. I'd argue that cuts in the grouping's favour: a fresh stack per file doesn't test "cleans up after itself" *at all*, and the suite had already lost that property without anyone noticing. The grouped run is what noticed.

Re-running now with the fix; group 3 green on the re-run is the verification.

Issue #14596, opened 2026-03-26
Missing django_linear_migrations package when using CSRF environment variables


Kisaragi-ng:
When setting up django-DefectDojo in master branch to be accessible over https with traefik, I have to add ENV `DD_SESSION_COOKIE_SECURE`, `DD_CSRF_COOKIE_SECURE`, and `DD_CSRF_TRUSTED_ORIGINS`
I draw these conclusion after reading:
- https://github.com/DefectDojo/django-DefectDojo/issues/6329
- https://github.com/DefectDojo/django-DefectDojo/blob/master/docker-compose.override.https.yml
- https://docs.djangoproject.com/en/6.0/ref/settings/#std-setting-CSRF_TRUSTED_ORIGINS

However, after adding those 3 variables i got error from uwsgi during docker compose up instead:
```
uwsgi-1         |   File "/app/manage.py", line 10, in <module>
uwsgi-1         |     execute_from_command_line(sys.argv)
uwsgi-1         |     ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
uwsgi-1         |   File "/usr/local/lib/python3.13/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
uwsgi-1         |     utility.execute()
uwsgi-1         |     ~~~~~~~~~~~~~~~^^
uwsgi-1         |   File "/usr/local/lib/python3.13/site-packages/django/core/management/__init__.py", line 416, in execute
uwsgi-1         |     django.setup()
uwsgi-1         |     ~~~~~~~~~~~~^^
uwsgi-1         |   File "/usr/local/lib/python3.13/site-packages/django/__init__.py", line 24, in setup
uwsgi-1         |     apps.populate(settings.INSTALLED_APPS)
uwsgi-1         |     ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
uwsgi-1         |   File "/usr/local/lib/python3.13/site-packages/django/apps/registry.py", line 91, in populate
uwsgi-1         |     app_config = AppConfig.create(entry)
uwsgi-1         |   File "/usr/local/lib/python3.13/site-packages/django/apps/config.py", line 193, in create
uwsgi-1         |     import_module(entry)
uwsgi-1         |     ~~~~~~~~~~~~~^^^^^^^
uwsgi-1         |   File "/usr/local/lib/python3.13/importlib/__init__.py", line 88, in import_module
[...]

PR #14571, opened 2026-03-22
fix: remove django-linear-migrations

valentijnscholten:
Removes `django-linear-migrations` from the project. The tool caused startup failures when `DD_DEBUG=True` but the package was not installed (e.g. production images). Too much hassle with low gains. Any non-linear migration will be caught on startup by django anyway during unit tests.

PR #15529, opened 2026-08-05
ci: follow-ups from the migration-graph review — push scoping and the misdiagnosing heredoc

devGregA:
Two small items promised in the #15504 review thread, independent of the larger script-trim (which comes separately).

## 1. Scope `push:` to the release lines

`ruff.yml` and `migration-graph.yml` trigger on both bare `push:` and `pull_request`, so every push to a PR branch ran each **twice** on the same commit. Measured while debugging queue starvation: a single PR push spawned 13 workflow runs, these two among the duplicates — and that contributed to exhausting the org's concurrent-job allowance while full suites sat queued.

The push trigger's real job is the release lines, where it's the only thing that checks the **merged result** after a PR lands — for the migration graph that post-merge union is the entire point. Scoping keeps exactly that case; on PR branches the `pull_request` event was the better run anyway (it sees the branch merged with its base, not the branch alone). Matches the pattern `gh-pages.yml` already uses.

## 2. The heredoc that misdiagnosed forked graphs

When `makemigrations --check` fails, both unit-test entrypoints printed one explanation: *"You made changes to the models without creating a DB migration for them."* For the other failure mode — `Conflicting migrations detected; multiple leaf nodes` — that's an active misdiagnosis: no model change of the author's is involved, and the fix is re-parenting, not creating. Anyone hitting a fork was sent hunting for a model change that doesn't exist, which is plausibly most of why forked-graph failures historically read as unrelated to migrations.

The message now names both causes, keyed to the error text Django prints directly above it, with the actual fix for each. Both copies updated (CI + devDocker entrypoints).

## Verification

`shellcheck` and `bash -n` clean on both entrypoints; `actionlint` clean on both workflows; LF endings preserved. The trigger change is observable on this PR itself: its branch pushes should spawn **no** `push`-event runs of ruff or the graph check, while the `pull_request` runs appear as usual.

Credit: both items are from @valentijnscholten's review on #15504.

### DIFF
--- a/.github/workflows/migration-graph.yml
+++ b/.github/workflows/migration-graph.yml
@@ -23,7 +23,18 @@ name: Migration Graph
 
 on:
   workflow_dispatch:
+  # The push trigger exists for the post-merge case described above, and that
+  # case only happens on the release lines -- so it is scoped to them. Unfiltered
+  # it also fired on every push to every PR branch, where it duplicated the
+  # pull_request run of the same commit (and the pull_request run is the better
+  # of the two there: it sees the branch merged with its base).
   push:
+    branches:
+      - master
+      - dev
+      - bugfix
+      - release/**
+      - hotfix/**
   pull_request:
 
 jobs:
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -4,7 +4,16 @@ on:
   workflow_dispatch:
   # called by unit-tests.yml as a gate: nothing else runs until ruff passes
   workflow_call:
+  # push is for the release lines: it is what lints the merged result after a
+  # pull request lands. Unfiltered it also fired on every push to every PR
+  # branch, doubling up with the pull_request trigger below on the same commit.
   push:
+    branches:
+      - master
+      - dev
+      - bugfix
+      - release/**
+      - hotfix/**
   pull_request:
   # A merge queue tests each speculative merge commit and waits for the required
   # status checks to report against it. `ruff-linting` is a required check, so it
--- a/docker/entrypoint-unit-tests-devDocker.sh
+++ b/docker/entrypoint-unit-tests-devDocker.sh
@@ -23,9 +23,21 @@ python3 manage.py makemigrations --no-input --check --dry-run --verbosity 3 || {
 
 ********************************************************************************
 
-You made changes to the models without creating a DB migration for them.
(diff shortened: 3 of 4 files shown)


valentijnscholten:
Thanks for the followup. Even with this check there is still the potential for duplicate leaves to be introduced. If two PRs have passing tests and both introduce a new leave migration, they can be merged shortly after eachother. 

I think we should take another attempt at implementing https://pypi.org/project/django-linear-migrations/ which will prevent this altogether by causing a merge conflict whenever two PRs introduce a new migration and one gets merged.

devGregA:
Agreed on the residual case — this PR's scoping change never touches it. Two individually-green PRs that each add a leaf on the same parent still fork the graph in their union, and the only thing here that addresses that is post-merge detection, which tells you *after* the branch is broken.

Two things are about to change that picture, and they compose rather than compete:

**The merge queue (imminent on `bugfix`) catches the union structurally.** The queue tests each PR on a speculative merge commit that includes whatever landed ahead of it, and the unit jobs run `migrate` — a forked graph dies with `CommandError: Conflicting migrations` inside the required `Unit Tests Complete` gate, so the second PR gets ejected instead of merged. That closes the "merged shortly after each other" window at the cost of one wasted queue cycle.

**`django-linear-migrations` is still worth the retry on top of that**, because it moves the failure from queue-ejection time to git-conflict time: both PRs must edit the same `max_migration.txt` line, so the second author finds out the moment the first PR merges — instantly, locally, with zero CI spend. Cheaper feedback, and it also covers `dev` until the queue extends there.

On the "another attempt" part — I looked up why the first one died, since the retry should not re-die the same way. #14145 added it, #14571 removed it: the app was enabled conditionally (`DD_DEBUG=True`) while the package wasn't installed in production images, so debug-enabled prod-image environments crashed at startup (#14496, #14596). The failure was the *conditional wiring*, not the tool. The retry shape that avoids it: unconditional dependency in `requirements.txt`, unconditionally in `INSTALLED_APPS`, no env-gating anywhere — it's a tiny pure-Python package, and the whole point is that it behaves identically in every environment. One-time cost to flag: every open PR carrying a migration conflicts the day it lands (they all need the `max_migration.txt` bump), so it wants a moment when the migration-PR backlog is small.

Happy to put that PR up unless you'd rather take it — you have the history with the first attempt.

PR #15557, opened 2026-08-06
fix(importers): stop an empty engagement status failing every save

Maffooch:
**Description**

A scan import aborts with a `ValidationError` on a field the import never touched:

```
django.core.exceptions.ValidationError: {'status': ['This field cannot be blank.']}
  File "dojo/importers/default_reimporter.py", line 135, in process_scan
    self.save_without_resurrecting(self.test.engagement)
  File "dojo/importers/base_importer.py", line 687, in save_without_resurrecting
    instance.save()
  File "dojo/base_models/base.py", line 64, in save
    self.full_clean()
```

Nothing is imported. Reported from several instances, all reimports into pre-existing engagements.

### Root cause

`Engagement.status` (`dojo/engagement/models.py`) is declared `null=True` with
`default="Not Started"` but **without** `blank=True`. Those three do not agree:

- the column accepts `NULL`, and DRF derives `allow_null=True` from `null=True`, so an empty
  status can be written;
- `full_clean()` — which `BaseModelWithoutTimeMeta.save()` runs on every save when
  `V3_FEATURE_LOCATIONS` is on — rejects **both** `None` and `""`, because
  `Field.validate()` raises `blank` for anything in `empty_values` when `blank=False`.

So a row with an empty status can exist and can then never be saved again, **by any code
path** — not just an import. That is the defect this fixes.

Empty-status rows are not hypothetical, and this repository ships two of them: engagement 7
in `defect_dojo_sample_data.json` and `defect_dojo_sample_data_locations.json` has
`"status": ""`, so a fresh sample-data install reproduces it. Any write made before the base
model started validating produces the same thing.

### How an import hit it, and how #15512 changed that

`process_scan` saves the test's engagement at the end of a run, and the only engagement field
an import ever sets is `target_end`, for CI/CD engagements (`base_importer.update_timestamps`).
A scan into a blank-status engagement therefore failed on a field it had not set.

**#15512 landed on `bugfix` after this PR was opened** and made that write-back conditional:
the engagement is saved only when `update_timestamps()` actually moved `target_end`
(`engagement_target_end_updated`). That narrows the import path considerably — an engagement
whose `target_end` is already in the future is no longer written back at all, so imports that
used to fail this way now skip the save entirely. This PR is **not** claiming to be the only
thing standing between that traceback and a user.

What #15512 does not do is make the row saveable. A blank-status engagement is still rejected
by every other write — a UI edit, an API `PATCH`, an engagement close — and still by an import
whenever `target_end` does move, which is any CI/CD engagement whose target end has passed.
The root defect is unchanged, so this fix is still needed; its reported symptom is simply
rarer than when it was filed.

### Fix

`Engagement.pre_save_logic()` applies the field's declared default when the status is empty.
An empty value is not one of `ENGAGEMENT_STATUS_CHOICES`, so there is nothing to preserve —
the alternative is a row no code path can save. `pre_save_logic()` runs *before* the
validation step and regardless of whether validation is enabled, so the save proceeds and
the row heals on its next write.

Deliberately not done:

- **No migration.** The column stays nullable and no data is rewritten, so an upgrade does
  not have to scan the engagement table; existing rows heal individually as they are saved.
  A `NOT NULL` migration would also not have covered the reported case, which is `""`, not
  `NULL`.
- **No relaxing of validation to `blank=True`.** That would legitimise a value outside the
  field's own `choices` and leave empty statuses rendering as empty in the UI.
- **No further change to the importer.** An earlier revision of this description argued
  against making the write-back conditional; #15512 has since done exactly that, on
  performance grounds, and this PR simply sits alongside it.

A status that is merely *wrong* is still rejected, and a populated status is left alone.

The two sample-data fixtures are corrected in the same change (engagement 7, and its seeded
pghistory insert event so the seeded history agrees with the seeded row).

**Test results**

New module `unittests/test_importers_blank_engagement_status.py`:

- the reported reimport into an engagement whose status was set behind the model's back —
  parameterized over `""` and `None` — completes, and the status is persisted as the default.
  The engagement is seeded with a past `target_end` so the run genuinely reaches the
  write-back under #15512's condition, and the test asserts that `target_end` actually moved
  **before** it reads anything into the status, so it fails loudly rather than passing for the
  wrong reason if that gate changes again;
- the same heal at model level, which is the lowest level the bug reproduces at, asserting
  the persisted value is one of the declared choices;
- control: a populated status (`"Blocked"`) survives a reimport untouched, so the fallback
  can only fire where the save previously failed;
- an invalid status (`"Nonsense"`) is still rejected (`skip_unless_v3`, since that is what
  makes the base model validate at all);
- the column still permits `NULL`, pinning why the fix is at save time rather than a
  migration.

Every assertion reports the persisted value in its failure message.

Verification: no Docker daemon was available in the environment this was prepared in, so the
Django suite was not run locally — CI is the evidence, and it has already caught two real
mistakes in the test fixture, both fixed in the commits above:

1. seeding `timezone.now()` into `DateField`-backed attributes, which errored all six tests in
   `setUp` on a datetime-vs-date comparison;
2. seeding a `target_end` that never moved, so after #15512 the write-back never ran and both
   reimport cases failed in the merge queue with the status still blank.

Directly verified besides CI:

- the mechanism, end to end, in a minimal standalone project against the pinned Django
  (5.2.15) replicating only `save()` (`pre_save_logic()` then `full_clean()`) and the
  `status` field declaration verbatim. Without the fallback, both `""` and `None` raise
  `{'status': ['This field cannot be blank.']}` — the reported message exactly. With it,
  both save and persist `Not Started`, while `"Nonsense"` still raises
  `{'status': ["Value 'Nonsense' is not a valid choice."]}` and `"Blocked"` is unchanged.
- both fixtures parse and no longer contain a blank engagement status.
- `ruff check` clean at the pinned CI version (0.16.0) on both changed Python files.

**Documentation**

None needed. No setting, endpoint, or documented behaviour changes — an engagement whose
status was empty now saves instead of raising.

**Impact on Dojo Pro**

No Pro-side change is required. Pro's connector reimport (`ConnectorsImporter` →
`ProReImporter`) delegates to this `process_scan`, so it inherits both this fix and #15512's
narrowing. Checked against Pro: nothing subclasses or proxies `Engagement`, nothing overrides
`pre_save_logic()` or `save_without_resurrecting()` for it, and no Pro test asserts a blank
engagement status. One visible consequence: Pro's engagement serializer omits `None` fields
from its payload, so a legacy engagement will now report `status: "Not Started"` where it
previously omitted the key — a healed value rather than a missing one.

**Checklist**

- [x] Rebased on the latest `bugfix`.
- [x] Ruff compliant at the pinned CI version.
- [x] No migration needed (deliberately — see above).
- [x] Tests added to the unit tests.


blakeaowens (approved):

devGregA (approved):

Maffooch:
Ejected from the merge queue on a real failure of mine, not a flake — force-pushed a rebase onto the current `bugfix`, so this needs a fresh look.

**What happened.** #15512 merged after this PR was opened and made the closing write-back conditional: `process_scan` now saves the engagement only when `update_timestamps()` actually moved `target_end` (`engagement_target_end_updated`). My regression test seeded `target_end` at today, so nothing moved, the engagement was never written back, and both reimport cases failed in the queue with the status still blank:

```
AssertionError: 'Not Started' != '' : expected status='Not Started', persisted=''
```

The PR-level run passed because it was built on the older base. The queue caught it exactly as intended.

**What changed.** Test-only — the fix under test is untouched:

- the engagement is seeded with a `target_end` 30 days back, and that is reset in the same statement that blanks the status (the initial import in `setUp` pushes `target_end` out to the scan date, so without the reset there would again be nothing for the reimport to move);
- the test now asserts `target_end` actually moved *before* it reads anything into the status, so if that gate changes again it fails loudly instead of passing for the wrong reason.

**One correction to the description, which I have updated.** I had argued against making the write-back conditional; #15512 has since done exactly that, and it means the reported import symptom is now much rarer — an engagement whose `target_end` is already in the future is no longer written back at all. It does not make the row saveable, though: a blank-status engagement is still rejected by every other write, and by an import whenever `target_end` does move. The root defect and this fix are unchanged; only the claim about how often it surfaces was wrong, and the description no longer makes it.

---

PR #15561, opened 2026-08-06
Add a DISA STIG checklist parser (.ckl and .cklb)

devGregA:
## Description

Adds a parser for **DISA STIG Viewer checklists**, so the items in a checklist that require remediation become trackable findings rather than living in a file on someone's laptop.

Requested by a customer running STIG assessments who wants the open items aged against SLAs, assigned, and reported on alongside everything else.

### Formats

Both checklist formats are accepted through **one** scan type, `DISA STIG Checklist`:

- `.ckl` — XML, written by STIG Viewer 2.x
- `.cklb` — JSON, written by STIG Viewer 3.x

The format is detected from **file content, not the file name**, so a renamed export still imports. Both readers build the same intermediate records and feed one finding builder, so identical content in either format produces identical findings — held in place by a parity test that compares the two sample files field by field.

Checklists with several `iSTIG` blocks (one host assessed against several STIGs) are supported, as are exports from tools other than STIG Viewer, whose JSON omits keys STIG Viewer always writes.

This reads *checklists*. XCCDF benchmark results remain the Openscap parser's job, and a file rooted at `<Benchmark>` is rejected with a message that says so.

### Statuses

All four checklist statuses are imported, so the finding list mirrors the checklist:

| Checklist status | Finding state |
|---|---|
| Open | Active, Verified |
| Not Reviewed | Active, not Verified |
| Not A Finding | Mitigated (inactive) |
| Not Applicable | Out of Scope (inactive) |

Not A Finding and Not Applicable carry `is_mitigated` / `out_of_scope` rather than only `active=False`. That is deliberate and load-bearing: on reimport an incoming finding that is merely inactive is a no-op, so mapping them that way would leave a remediated item sitting open forever. With these flags, reimporting an updated checklist closes items that were fixed and reactivates ones that regressed.

### Identity

`unique_id_from_tool` is the **V-number qualified by the assessed host** (`rhel9.example.mil/V-257777`), with `DEDUPE_ALGO_UNIQUE_ID_FROM_TOOL`.

- The V-number is the stable per-rule id. `Rule_ID` carries a revision suffix (`SV-257777r925318_rule`) that changes with **every STIG release**, so it is recorded in the description for the reader and never used as identity — a STIG version upgrade keeps finding continuity.
- Reimport and cross-test dedupe match on the unique id without ever comparing endpoints, so qualifying it with the host is what keeps one rule failing on two hosts as two findings.
- A `HASHCODE_FIELDS_PER_SCANNER` entry (`vuln_id_from_tool`, `endpoints`) is registered alongside so the *stored* `hash_code` is stable too, rather than falling through to the legacy title+description formula that false-positive history and similar-findings key on.

A checklist with no host name, FQDN or IP recorded falls back to the bare V-number; two such un-keyed checklists in one product cannot be told apart. Documented on the parser page.

### Other mapping notes

- Severity comes from the rule's DISA category (`high`/`medium`/`low`), and the category itself (CAT I/II/III) is kept in Impact, where an assessor's severity override does not move it. An override sets the severity and its reason becomes the severity justification.
- `Check_Content` becomes the steps to reproduce and `Fix_Text` the mitigation.
- Every CCI the rule cites is written on its own line in References. That format is a deliberate seam: DefectDojo Pro's compliance pack reads those CCIs back to map checklist items onto their NIST 800-53 controls.
- An unrecognised status is imported as Not Reviewed with a warning rather than dropped; an item with no V-number has no identity and is skipped.

### `.ckl` / `.cklb` added to the importable file types

Without this, `ImporterFileExtensionValidator` rejects a natively-named checklist before the parser is ever reached.

## Testing

- 16 new parser tests, run with `V3_FEATURE_LOCATIONS` **both on and off**.
- Full native suite, all three CI passes: **6486 / 45 / 7 tests**, no failures attributable to this change. (One pre-existing environment-only failure, `test_add_jira_instance_unknown_host`, which asserts the libc resolver's message text and fails on macOS at any commit.)
- `test_parsers`, `test_factory` and `test_dedupe_config_checks` all pass, so the docs page, sample directory, class naming and the dedupe-config system check are satisfied.
- `ruff check` clean against the repo config with the CI-pinned 0.16.0.
- Validated live against a running stack: imported the `.ckl`, confirmed the four status states and the CCI lines; reimported an edited checklist and confirmed a fixed item closes, a regressed item reactivates, and one dropped from the checklist is closed by absence; imported the `.cklb` twin and confirmed all four findings deduplicate against the `.ckl` originals; imported a second host's checklist and confirmed the same rule stays a separate finding.

### Sample data

Eight synthetic samples covering mixed statuses, a severity override with justification, repeated `CCI_REF` blocks, multiple `iSTIG` blocks, a blank host, an empty checklist, a `.cklb` missing optional keys, a windows-1252 encoded `.ckl`, and a `.cklb` with a UTF-8 BOM. Content is public STIG rule material only.

### DIFF
--- a/docs/content/federal_compliance/compliance_profile.md
+++ b/docs/content/federal_compliance/compliance_profile.md
@@ -49,7 +49,10 @@ Two profile settings are not on the form and are set through the compliance API:
   *do* carry their own control references are mapped from those instead; see
   [Control Coverage](../control_coverage).
 * **Configuration test types** — the test types whose findings are treated as configuration items,
-  which is what drives CM-6 consolidation in the ledger.
+  which is what drives CM-6 consolidation in the ledger. `DISA STIG Checklist` is the usual
+  candidate: a benchmark assessment produces many configuration items at once, and consolidating
+  them keeps the ledger readable. Whether to consolidate is a per-system decision, so no test type
+  is listed here by default.
 
 ## Auditability
 
--- a/docs/content/federal_compliance/control_coverage.md
+++ b/docs/content/federal_compliance/control_coverage.md
@@ -49,8 +49,10 @@ was enabled, backfill them:
 manage.py extract_control_mappings --product <id>
 ```
 
-Use `--all` to scan every active finding instead of one product. The command reports how many
-mappings it created, and it leaves manual and suppressed mappings alone.
+Use `--all` to scan every active finding instead of one product. Both passes — scanner references
+and the CCI crosswalk — run by default, and each reports how many mappings it created. Add
+`--skip-crosswalk` or `--skip-scanner-refs` to run only one. Manual and suppressed mappings are left
+alone either way.
 
 ## Correcting a mapping
 
--- a/docs/content/federal_compliance/control_coverage.md
+++ b/docs/content/federal_compliance/control_coverage.md
@@ -18,9 +18,24 @@ automatically. Among others:
 * **Prowler** writes NIST 800-53 control lists into finding references.
 * **Tenable** plugins carry 800-53 cross-references.
 * **InSpec** and **MITRE SAF** profiles tag their checks with `nist` identifiers.
+* **DISA STIG checklists** cite CCIs, which DefectDojo crosswalks onto 800-53 — see below.
 
 Extraction is grounded in the imported catalog, so an identifier the catalog does not recognize
 never produces a mapping.
+
+### STIG checklists and CCIs
+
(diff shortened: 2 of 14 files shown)


devGregA:
Closing this: the STIG checklist parser belongs in DefectDojo Pro, not open source. Reopening as a Pro change. Apologies for the noise.

PR #14964, opened 2026-06-05
Prefetcher updates

dogboat:
Update and optimize prefetcher capabilities.

[sc-13155]

PR #15570, opened 2026-08-07
Make a private note private

svader0:
Makes the "Private" checkbox on a note actually mean private. Right now it only keeps the note out of reports and Jira; anyone who can see the parent object can still read it.

What changes:

- Notes marked private are only shown to their author. That covers the API (Finding, Test, Engagement and Risk Acceptance endpoints, their notes actions, and generate_report), the notes tab on those pages, and the note popovers and note counts on the findings tables.
- The rule lives in one place, `dojo.notes.helper.visible_notes`, used by the serializer, the views and a new `visible_to` template filter. It was spread across several hand-written filters before, which is how the API and the UI came to disagree in the first place.
- The `private` field gets `help_text` (migration 0290) so the checkbox explains itself, instead of leaving it to the caption underneath. That caption changes from "(will not appear in report)" to "(private to you)", and the docs sentence is updated to match.

Superusers still see every note, and public notes are unchanged.

Two things worth a reviewer's attention:

- Note-type availability deliberately keeps the unfiltered set. A single-use note type that someone else used privately is still taken, so offering it would only fail on save.
- The popovers on the findings tables read the notes relation straight off the model, so filtering in the views did not reach them. They go through the new template filter instead.

**This is a visible behaviour change and needs a release note.** Notes people could previously read will disappear from their view, and note counts will get smaller. Nothing is deleted; those users just cannot see them any more.

### DIFF
--- a/docs/content/triage_findings/findings_workflows/OS__risk_acceptance.md
+++ b/docs/content/triage_findings/findings_workflows/OS__risk_acceptance.md
@@ -71,7 +71,7 @@ To remove a Finding from a Full Risk Acceptance, click the **Remove** button wit
 
 The Full Risk Acceptance's view also includes a table at the bottom for all other Findings from Tests within that Engagement. From there, you may select additional Findings and add them to that Full Risk Acceptance. 
 
-Additionally, there is a Notes function that allows Users to include additional context to the Full Risk Acceptance. All public notes will appear in any Reports that are generated for the Full Risk Acceptance, whereas notes that are toggled as **Private** will not appear in reports. 
+Additionally, there is a Notes function that allows Users to include additional context to the Full Risk Acceptance. All public notes will appear in any Reports that are generated for the Full Risk Acceptance. Notes toggled as **Private** are visible only to their author and to superusers, and are left out of reports. 
 
 Importantly, if a Full Risk Acceptance is deleted entirely, the Findings within will have their status automatically reverted to “Active.”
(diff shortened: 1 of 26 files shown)


svader0:
Context: 

Notes.private has meant two different things at once for a little while. In the past, it's meant 'keep out of reports and Jira' and the UI says 'will not appear in report'. Then #14964 added authorization policies to the prefetch path and needed a rule for notes. It landed on the stricter one, a note is visible only to its author, and the regression test justified it as matching existing UI behaviour. That justification was wrong: every filter(private=False) in the codebase is a report or export path, and the notes tab has always used notes.all(). Report code got read as UI code. The Pro notes rewrite later treated that policy as the spec and built on it.

Nobody changed the caption, the docs, or the serializers at any point, so we ended up shipping both meanings at the same time. One request could answer both ways: GET /api/v2/findings/{id}/?prefetch=notes returned zero notes under the prefetch policy and the private note inline in the same response.

So, this change migrates to the stricter interpretation. The field is called "private," Pro's UI already behaves that way, and a checkbox that does not do what its name says is the more surprising option. The rule now lives in one function that the serializer, the views and the templates all call, so the next place that reads notes gets it by default rather than inventing its own filter.

blakeaowens (approved):

Maffooch (approved):

devGregA (approved):

PR #15573, opened 2026-08-07
Make a private note private in the UI too, and say so

svader0:
Follow-up to #15570, which scoped notes to their author on the API. This does the same for the UI, so the flag means one thing everywhere.

**Stacked on #15570.** The diff below includes that PR's commit until it merges; the second commit is the one to review here.

What changes:

- The notes tab on Findings, Tests, Engagements and Risk Acceptances no longer shows you notes other people marked private.
- Same for the note popover and note counts on the findings tables, which read the relation straight off the model and so were missed by the view-level filter.
- The rule now lives in one place, `dojo.notes.helper.visible_notes`, used by the serializer, the views and a new `visible_to` template filter. It was spread across several hand-written filters before, which is how the API and the UI came to disagree.
- The `private` field gets `help_text` (migration 0290), so the checkbox explains itself instead of relying on the caption underneath. The caption changes from "(will not appear in report)" to "(private to you)", and the docs sentence is updated to match.

Note-type availability deliberately keeps the unfiltered set: a single-use note type someone else used privately is still taken, and offering it would only fail on save.

**This is a visible behaviour change and needs a release note.** Notes people could previously read on these pages will disappear, and note counts will get smaller. Nothing is deleted; those users just cannot see them any more. Superusers are unaffected, and public notes are unchanged.

### DIFF
--- a/docs/content/triage_findings/findings_workflows/OS__risk_acceptance.md
+++ b/docs/content/triage_findings/findings_workflows/OS__risk_acceptance.md
@@ -71,7 +71,7 @@ To remove a Finding from a Full Risk Acceptance, click the **Remove** button wit
 
 The Full Risk Acceptance's view also includes a table at the bottom for all other Findings from Tests within that Engagement. From there, you may select additional Findings and add them to that Full Risk Acceptance. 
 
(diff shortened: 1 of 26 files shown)


svader0:
Folding this into #15570 so the whole change reviews as one PR. Nothing dropped; both commits are on that branch now.

PR #15623, opened 2026-08-11
fix(findings): retry bulk finding delete chunks on transient DB conflicts

Maffooch:
## Description

A bulk finding delete (`POST /findings/bulk_delete/`) of a few hundred findings failed in production with a hard 500:

```
psycopg.errors.DeadlockDetected: deadlock detected
DETAIL:  Process A waits for ShareLock on transaction X; blocked by process B.
         Process B waits for ShareLock on transaction Y; blocked by process A.
CONTEXT:  while updating tuple (...) in relation "dojo_finding"
```

### Root cause

`_bulk_delete_findings_internal` (`dojo/finding/helper.py`) deletes findings one chunk at a time, each chunk in its own `transaction.atomic()`. Within that transaction it **updates** `dojo_finding` rows — clearing M2M through rows and resolving inbound `duplicate_finding` references — before issuing the delete. When a bulk delete overlaps a concurrent import or deduplication that touches the same rows in the opposite order, Postgres detects a deadlock and aborts one of the two transactions (SQLSTATE `40P01`).

The **background** cascade-delete task (`async_delete_task` in `dojo/utils.py`) already treats a deadlock / serialization failure as a lost race worth retrying — the rolled-back work is not wrong, only undone — using `is_transient_db_conflict`. The **synchronous** bulk-delete path had no such backstop, so the aborted chunk surfaced to the API caller as a 500.

### The fix

Wrap each per-chunk atomic block in a bounded retry that re-runs the chunk only on a transient DB conflict (`is_transient_db_conflict`), with exponential backoff. Anything else — a statement timeout, a dropped connection — and a conflict that survives every attempt still propagate unchanged.

This is safe because:
- The delete is already chunked with **per-chunk transactions**, so earlier committed chunks are untouched and only the aborted chunk re-runs.
- `ATOMIC_REQUESTS` is not enabled, so each chunk is its own top-level transaction; a deadlock rolls it back cleanly and a fresh retry can commit (an in-request savepoint retry would not, hence this check matters).

No schema change, so **no migration** is required — this is a pure control-flow change around an existing transaction.

### Pro impact

Dojo-Pro overrides `BULK_DELETE_FINDINGS_METHOD` but its implementation delegates straight to `_bulk_delete_findings_internal`, so the Pro `/api/vue/findings/bulk_delete/` path (the one that hit this in production) inherits the retry with **no companion Pro PR needed**. No public API, serializer, or user-visible behavior changes — a delete that previously 500'd on a race now completes — so no docs change is required.

## Test results

New `unittests/test_bulk_delete_findings_deadlock_retry.py` covers:
- a transient conflict on the first attempt that is retried and then deletes the chunk;
- a non-transient `OperationalError` (e.g. statement timeout) that propagates without retry, leaving the findings intact;
- a persistent transient conflict that gives up after the configured number of retries and raises, deleting nothing.

The tests patch `utils_cascade_delete.execute_delete_sql` (the same seam the existing `test_bulk_delete_findings_m2m` suite uses) to simulate the conflict deterministically, and no-op the backoff `sleep` so they run fast.

`ruff check` passes under the pinned config.

## Checklist

- [x] Give a meaningful name to your PR.
- [x] Your code is Ruff compliant.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CHk1DJ3wrYkSL3E8K7RBug)_

### DIFF
--- a/dojo/finding/helper.py
+++ b/dojo/finding/helper.py
@@ -6,7 +6,7 @@ from itertools import batched
 from time import sleep, strftime
 
 from django.conf import settings
-from django.db import transaction
+from django.db import OperationalError, transaction
 from django.db.models import Count
 from django.db.models.query_utils import Q
 from django.db.models.signals import post_delete, pre_delete
@@ -19,6 +19,7 @@ from fieldsignals import pre_save_changed
 
 import dojo.risk_acceptance.helper as ra_helper
 from dojo.celery import app
+from dojo.db_utils import is_transient_db_conflict
 from dojo.endpoint.utils import endpoint_get_or_create, save_endpoints_to_add
 from dojo.file_uploads.helper import delete_related_files
 from dojo.finding.cwe import finding_cwe_labels
@@ -1028,6 +1029,14 @@ def resolve_inbound_duplicate_references(chunk_ids, delete_scope_ids):
     return len(survivors)
 
 
+# A synchronous bulk delete that loses a concurrency race (deadlock or serialization
+# failure) rolls the offending chunk back and would otherwise surface as a 500 to the
+# caller. Retry that chunk a few times, mirroring async_delete_task's backstop for the
+# background cascade deletes that deterministic lock ordering cannot fully rule out.
+BULK_DELETE_RETRY_DELAY = 0.5  # seconds before the first retry of a chunk; doubled each attempt
+BULK_DELETE_MAX_CONFLICT_RETRIES = 3
+
+
 def _bulk_delete_findings_internal(finding_qs, chunk_size=1000, *, order_desc=False):
     """
     Delete findings and all related objects efficiently. Including any related object in Dojo-Pro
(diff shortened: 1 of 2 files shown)


blakeaowens (approved):

devGregA (approved):

Maffooch:
Dropped from the merge queue on a **flaky UI test**, not a real failure from this change.

- The failing job was `test-user-interface / User Interface Tests (tests/login_test.py tests/alerts_test.py tests/system_settings_test.py …)` — a Selenium UI shard.
- This diff is **backend-only** (a transient-conflict retry inside `_bulk_delete_findings_internal`); it cannot affect login / alerts / system-settings UI flows.
- The **identical shard passed on this PR's head commit** (`494cef7c`) in the normal PR run, and the other PRs batched in the queue at the same time passed the same suite — so the base is healthy and this was a flake in the queue run.

No code change needed. Re-queuing.

---

PR #15656, opened 2026-08-13
perf(finding): drop two dojo_finding indexes a wider index already covers

devGregA:
## What

Django creates a single-column index for every `ForeignKey`. Two of those on `Finding` are redundant, because another index already begins with the same column:

| Redundant index | Already covered by |
|---|---|
| `(test_id)` | nine composites leading with `test_id`, among them `(test_id, active, verified)` and `(test_id, hash_code, duplicate)` |
| `(duplicate_finding_id)` | `(duplicate_finding_id, id)` |

A btree whose columns are a strict prefix of another btree, with the same (here absent) partial predicate, can be answered from the longer index. The shorter one returns nothing for the write cost it adds to every finding insert.

## Why it's worth doing

`dojo_finding` is the busiest table in the schema and currently carries **49 indexes**, every one of which is maintained on each row written. Import is where that is felt — a scan that creates thousands of findings pays all 49 per row.

This PR removes only what is *provably* covered, so there is no judgement call about query patterns involved.

## Why it doesn't weaken the foreign keys

PostgreSQL requires an index on the **referenced** (unique) side of a foreign key, not on the referencing side. The referencing-side index only accelerates cascade deletes and parent-row deletion checks — and for both fields a covering composite remains, so those paths keep an index to use.

## Migration safety

Built with `DROP INDEX CONCURRENTLY` in a non-atomic migration, following the pattern already established by `0280_vulnerability_id_upper_index`, so it takes no exclusive lock on a large `dojo_finding`. `IF EXISTS` makes it idempotent, and the reverse rebuilds concurrently as well, so a downgrade doesn't lock the table either.

The index names in the SQL are Django's deterministic auto-generated names for these two foreign keys (table + column + a hash of both), so they are identical on every install.

## Verification

- `makemigrations dojo --check --dry-run` reports **"No changes detected"** — the migration's state operations match the model, so a later `makemigrations` stays a no-op.
- Applied against a migrated database: `dojo_finding` goes from **49 indexes to 47**, and both named indexes are gone.
- Reversed: both indexes come back.
- `ruff` (repo config) clean.

## Scope

Deliberately narrow. Several single-column **boolean** indexes on the same table look like stronger candidates still — six of them are already covered by `(test_id, …)` composites — but whether an index is actually used is a property of the queries an instance runs, not of the schema. Settling those needs `pg_stat_user_indexes` from a long-running instance, so they are left alone here.

### DIFF
--- a/dojo/db_migrations/0295_drop_redundant_finding_fk_indexes.py
+++ b/dojo/db_migrations/0295_drop_redundant_finding_fk_indexes.py
@@ -1,0 +1,88 @@
+"""Drop two `dojo_finding` indexes that a wider index already covers.
+
+Django creates a single-column index for every ForeignKey. On `Finding` two of those are
+redundant, because another index already begins with the same column:
+
+    (test_id)               -> covered by nine composites that lead with test_id, e.g.
+                               (test_id, active, verified), (test_id, hash_code, duplicate)
+    (duplicate_finding_id)  -> covered by (duplicate_finding_id, id)
+
+A btree whose columns are a strict prefix of another btree, with the same (here, absent)
+partial predicate, can be answered from the longer index, so the shorter one returns
+nothing for the write cost it adds to every finding insert. `dojo_finding` is the busiest
+table in the schema and carries close to fifty indexes, so each one removed is a real
+saving on import.
+
+Dropping the referencing-side index does not weaken the foreign keys. PostgreSQL requires
+an index on the REFERENCED (unique) side of a foreign key, not on the referencing side; the
+referencing-side index only accelerates cascade deletes and parent-row deletion checks, and
+in both cases a covering composite remains.
+
+Built with DROP INDEX CONCURRENTLY (non-atomic migration, following 0280) so it does not
+take an exclusive lock on a large `dojo_finding`. IF EXISTS makes it idempotent, and the
+reverse rebuilds concurrently too, so a downgrade does not lock the table either.
+
+The index names are Django's deterministic auto-generated names for these two foreign keys
+(table + column + a hash of both), so they are identical on every install.
+"""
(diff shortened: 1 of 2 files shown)


devGregA:
Closing this — **production index-usage statistics refute the premise.** Thanks for the patience; better to withdraw it than have it reviewed on a false basis.

I argued that a single-column index whose column is a strict prefix of a wider index returns nothing, because the wider index *can* serve the same lookup. That is true of what the planner is *able* to do, and wrong about what it *chooses* to do.

Measured on a long-running instance (~184k findings, ~6 months of accumulated statistics), the two indexes this PR proposed to drop are among the busiest on the table:

| Index | `idx_scan` | Size |
|---|---|---|
| `dojo_finding_test_id_ef8ec5fc` `(test_id)` | **11,284,300** | 7,936 kB |
| `dojo_finding_duplicate_finding_id_9ff391c5` `(duplicate_finding_id)` | **4,377,532** | 9,184 kB |

The planner prefers them precisely *because* they are narrow: `(test_id)` is 7,936 kB against 35 MB for `(test_id, hash_code, duplicate)`. Dropping them would not remove work, it would push more than 15 million lookups onto indexes roughly four times larger — trading a small write saving for a much larger read cost, and more buffer churn on the hottest table in the schema.

The correct general lesson, which I had backwards: **prefix-redundancy makes an index removable, not useless.** Whether removing it is a win depends on relative size and on how often the narrow path is actually taken, and that is only answerable from `pg_stat_user_indexes` on a real workload.

There *are* genuine candidates on this table — eleven indexes showed zero scans over the same six months, several of them large. I would rather come back with a PR grounded in that data, across more than one instance, than leave this one open.

PR #15664, opened 2026-08-14
fix(tags): order single-finding delete tag-count decrements so concurrent deletes cannot deadlock

Maffooch:
**Description**

A single-object finding delete — `DELETE /api/v2/findings/{id}/` → `Finding.delete()` → `super().delete()` — surfaced a Postgres deadlock to the client:

```
django.db.utils.OperationalError: deadlock detected
DETAIL:  Process A waits for ShareLock on transaction N; blocked by process B.
         Process B waits for ShareLock on transaction M; blocked by process A.
CONTEXT:  while updating tuple (x,y) in relation "dojo_tagulous_finding_tags"
```

**Root cause.** `dojo_tagulous_finding_tags` is the tagulous **tag** table (it holds `count`), so the deadlocked statement is the per-tag `count = count - 1` UPDATE, which takes a row lock on each tag. The synchronous single-object delete does not clear tags itself; it relies on `super().delete()`, whose `pre_delete` fires tagulous's per-object `clear()`. That `clear()` issues one count UPDATE per tag **in the manager's own iteration order**, not in ascending tag-id order. Because the UPDATE order is the lock order, two concurrent single-finding deletes whose tag sets **overlap** can take the same tag-row locks in **opposite orders**, forming a cycle Postgres breaks by aborting one side (SQLSTATE `40P01`).

This is the **single-object twin** of two deadlocks already fixed for the other tag-count paths:
- #15486 made the bulk/cascade remove path (`bulk_remove_all_tags`) decrement counts in a deterministic **ascending tag-id** order, and added an `async_delete_task` retry backstop.
- #15652 gave the import **add** path the same ascending tag-id ordering.

The synchronous `Finding.delete()` path — used by the REST `destroy()` in both open source and the Pro override, which calls `super().destroy()` → `instance.delete()` — was never routed through that shared ordering, so it can still deadlock against a concurrent single delete, bulk delete, or import.

**Fix.** `Finding.delete()` now removes the finding's tags via `bulk_remove_all_tags(Finding, Finding.objects.filter(pk=self.pk))` **before** `super().delete()`. This issues the count decrements in the same ascending tag-id order every other tag-count path already uses, so all callers acquire the tag-row locks in one shared order and the cycle cannot form. Clearing the through rows first also leaves the tagulous `pre_delete` handler nothing to decrement, so tag counts are not double-processed. Behavior is now consistent with the bulk cascade delete, which is the dominant deletion path.

**No migration.** This is a pure ordering change to an existing code path — no model or schema change. A migration was neither generated nor needed; the fix reuses the already-shipped `bulk_remove_all_tags` helper.

**Downstream (Pro) impact.** DefectDojo Pro's `FindingViewSet.destroy()` overrides the OSS one but delegates to `super().destroy()`, which calls the model-level `Finding.delete()` — so the fix covers the Pro delete path with no Pro-side change. Pro's `Finding` `pre_delete` receiver only dispatches integrator notifications and does not touch tag counts. Pro does not subclass `Finding` or override `Finding.delete()`.

**Test results**

Added `FindingDeleteTagLockOrderTest` in `unittests/test_tag_utils_bulk.py`, mirroring the existing `BulkRemoveAllTagsLockOrderTest` (#15486) and `BulkAddTagsToInstancesLockOrderTest` (#15652). Three tags are attached to a finding in an order unrelated to their ids; the test captures the tag rows locked by the count UPDATEs during `finding.delete()` and asserts they are issued in ascending tag-id order. It fails on the old per-object-`clear()` order (name order, `[2,3,1]`) and passes once the delete is routed through `bulk_remove_all_tags` (`[1,2,3]`).

The test is deterministic — it asserts the update/lock ordering, not a live race. As with #15652, a full Postgres-backed suite run could not be stood up in the ephemeral environment used to author this change; validation is being carried through CI.

`ruff check` clean on both changed files (0.16.0, repo `ruff.toml`).

**Documentation**

No documentation change: no user-visible behavior change and no new setting.

**Checklist**

- [x] Meaningful PR name.
- [x] No model changes / no migration.
- [x] Ruff compliant.
- [x] Unit test added covering the ordering guarantee.

---
_Generated by [Claude Code](https://claude.ai/code/session_015i5bbWnQg2wKmQpvD7DRLz)_


blakeaowens (approved):

devGregA (approved):

Maffooch:
CI note: this PR was removed from the merge queue on a single, unrelated unit-test failure in the `test-rest-framework` job:

```
FAIL: unittests.test_apiv2_user.UserTest.test_api_token_auth_is_rate_limited
AssertionError: 4 != 3 : api-token-auth should honor the configured rate limit
```

That is not caused by this change. This PR only reorders tagulous tag-count decrements during `Finding.delete()` (`dojo/finding/models.py`, plus a tag-ordering unit test) — it does not touch API-token auth or throttling. The same `test-rest-framework` suite passed on this PR's own head-commit run, and this looks like a timing/state-dependent flake in the rate-limit counter (it saw one extra request reach the auth view within the window, 4 vs 3).

No code change is warranted here — re-queuing should clear it.

---

Maffooch:
CI note (2nd dequeue): the merge-queue failure this time was the UI job `tests/finding_test.py` in the **`DD_V3_FEATURE_LOCATIONS=true`** matrix variant. That variant runs only in the merge queue (the `pull_request` event runs the `false` variant only, per `integration-tests.yml`), which is why it didn't appear on this PR's own commit runs — where the full unit, rest-framework (true+false), and UI (`false`) suites all pass on both the original and the post-merge commit.

This looks like the same Selenium flakiness the queue is seeing across PRs right now — several other PRs (#15616, #15650, #15671) failed their merge-queue `Unit tests` run and then passed unchanged on re-queue in the same window. The failing variant toggles the V3 Locations feature flag, which is orthogonal to this PR's change (ordering of tagulous tag-count decrements in `Finding.delete()`); nothing here touches locations/endpoints or the UI delete flow.

No code change is warranted — re-queuing should clear it, as it has for the other PRs.

---

Maffooch:
CI note (3rd dequeue): again an unrelated failure, this time in the `test-rest-framework (linux/arm64, V3_FEATURE_LOCATIONS=true)` variant:

```
FAIL: unittests.test_v3_alias_authorized_users_api_authz
      .V3AliasAuthorizedUsersApiPermissionTest.test_asset_edit_cannot_add_authorized_users
AssertionError: 403 != 404 : {"detail":"No Product matches the given query."}
```

That's a V3-alias authorization permission test (a 403-vs-404 object-visibility flake) — orthogonal to this PR, which only reorders tagulous tag-count decrements in `Finding.delete()`.

All three merge-queue dequeues of this PR have hit **different** tests (rate-limit → `finding_test.py` Selenium → this authz test), each in the `V3_FEATURE_LOCATIONS=true`/arm64 matrix that only runs in the queue, and other PRs in the same window show the same fail-then-pass-on-retry behavior. The full unit, rest-framework (true+false), and UI suites all pass on this PR's own commit runs.

This looks like flakiness in the queue's full matrix rather than anything in this change. Re-queuing has eventually landed other PRs; if it keeps bouncing, an admin merge may be the pragmatic path.

---

Maffooch:
CI note (4th dequeue) — update to the picture: this time it's the **same** test as the 3rd dequeue, now on the `linux/arm64, false` variant as well as the earlier `arm64, true`:

```
FAIL: unittests.test_v3_alias_authorized_users_api_authz
      .V3AliasAuthorizedUsersApiPermissionTest.test_asset_edit_cannot_add_authorized_users
AssertionError: 403 != 404 : {"detail":"No Product matches the given query."}
```

So this isn't purely random — `test_asset_edit_cannot_add_authorized_users` is failing **reproducibly on the arm64 matrix** (both `V3_FEATURE_LOCATIONS` values). That test and its fixtures come from `bugfix` and are unrelated to this PR (which only reorders tagulous tag-count decrements in `Finding.delete()`); the merge here doesn't touch it, so it predates/is independent of this change. It will presumably block any PR whose queue batch runs the arm64 leg until the test itself is looked at.

Given that, re-queuing alone likely won't clear it — an admin merge, or a fix/skip of `test_asset_edit_cannot_add_authorized_users` on arm64, is probably the path. Happy to open a separate issue for the arm64 authz-test failure if that's useful.

---

PR #15666, opened 2026-08-14
Enforce the engagement object permission on the questionnaire answer view

svader0:
Hardening / consistency improvement to the engagement-scoped questionnaire views. Aligns the answer view with its siblings so it always applies the object-level edit permission check, and adds a regression test. No functional change for correctly-permissioned users.

### DIFF
--- a/dojo/survey/ui/views.py
+++ b/dojo/survey/ui/views.py
@@ -15,7 +15,6 @@ from django.views import View
 
 from dojo.authorization.authorization import (
     user_has_configuration_permission,
-    user_has_permission,
     user_has_permission_or_403,
 )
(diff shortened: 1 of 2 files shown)


Maffooch (changes requested):
It is a feature to allow surveys to be answered without a login if configured to be available

svader0:
@Maffooch This change should keep surveys answerable without a login. The engagement questionnaire page always needs a login, so with anonymous surveys switched on a logged-out user still gets sent to the login screen. But the setting also drops the permission check there, so any logged-in account can rewrite another team's saved answers. Surely that is not intended?

PR #15705, opened 2026-08-18
fix(dedupe): retry post_process_findings_batch on transient DB deadlock

Maffooch:
**Description**

Fixes a `DeadlockDetected` (Postgres SQLSTATE `40P01`) failure of the
`dojo.finding.helper.post_process_findings_batch` Celery task, observed in production as
a background task crash:

```
psycopg.errors.DeadlockDetected: deadlock detected
DETAIL:  Process A waits for ShareLock on transaction ...; blocked by process B.
         Process B waits for ShareLock on transaction ...; blocked by process A.
CONTEXT:  while updating tuple (...) in relation "dojo_finding"
```

When two `post_process_findings_batch` tasks run concurrently — parallel imports or
connector syncs into the same product — their batch deduplication / false-positive-history
writes update overlapping `dojo_finding` rows. Postgres detects the lock cycle and aborts
one participant. The aborted batch was only rolled back, not computed incorrectly, so the
correct response is to re-run it; instead the task surfaced as a hard failure and the whole
batch lost its deduplication.

**The fix** (`dojo/finding/helper.py`)

The status-changing dedup / false-positive-history write unit is now wrapped in a
transient-conflict retry with exponential backoff, reusing the existing
`is_transient_db_conflict` predicate (SQLSTATE `40P01` deadlock / `40001` serialization
failure). This mirrors the backstop `bulk_delete_findings` already uses for the same class
of conflict on the same table.

- Findings are reloaded on each attempt, so a retry acts on the state the winning
  transaction just committed.
- The non-status-changing follow-ups (issue updater, product grading, JIRA push) run once,
  *after* the retry loop succeeds, so a retry never double-fires them (no duplicate JIRA
  pushes).
- Only transient conflicts are retried. Any other `OperationalError` (e.g. a statement
  timeout) re-raises immediately, as does a conflict that survives every attempt.

**Why retry rather than row locks**

Serializing the batch writes with `select_for_update` on the candidate rows would trade an
occasional aborted batch for routine deadlocks on this hot path (an import's dedup write
and the excess-duplicate delete task would form a lock-ordering cycle). Retry-on-conflict
is the pattern already established in this module for exactly this reason.

**No migration**

This is a pure control-flow change to a background task — no schema change, no data
migration. The only avenue considered and rejected was lock-based serialization, above.

**Pro impact**

Pro overrides deduplication via the `FINDING_DEDUPE_BATCH_METHOD` hook, which is invoked
*inside* `dedupe_batch_of_findings`, i.e. inside the retried block — so the Pro dedup path
is covered by this retry with no companion change. Pro does not override
`post_process_findings_batch` itself, and the JIRA/issue-updater follow-ups it relies on
remain outside the retry, so they still run exactly once.

**Test results**

Added `unittests/test_finding_helper.py::TestPostProcessFindingsBatchDeadlockRetry` with a
simulated deadlock (no real concurrency needed):

- retry-then-succeed: two simulated deadlocks then success completes without raising and
  reloads findings each attempt;
- re-raise after exhausting retries: a persistent deadlock surfaces after
  `initial + MAX_CONFLICT_RETRIES` attempts;
- a non-transient `OperationalError` propagates immediately with no retry/backoff.

The retry-then-succeed test fails on the unfixed code (the first deadlock propagates).

**Documentation**

No user-facing behavior change — this removes a failure mode from a background task. No
docs update needed.

**Checklist**

- [x] Give a meaningful name to your PR.
- [x] Your code is Ruff compliant.

---
_Generated by [Claude Code](https://claude.ai/code/session_0159aVEfX5R8uJtkn6eTxsf7)_

### DIFF
--- a/dojo/finding/helper.py
+++ b/dojo/finding/helper.py
@@ -463,6 +463,17 @@ def post_process_finding_save_internal(finding, dedupe_option=True, rules_option
             jira_services.push(finding.finding_group)
 
 
+# post_process_findings_batch runs the status-changing dedup / false-positive-history
+# writes for a batch of findings. Two of these tasks racing on overlapping dojo_finding
+# rows -- concurrent imports or connector syncs into the same product -- can deadlock
+# (Postgres SQLSTATE 40P01 deadlock_detected): each holds a row the other needs and
+# Postgres aborts one participant. The aborted batch is not wrong, only rolled back, so
+# re-running it is safe. This mirrors bulk_delete_findings' backstop for the same class
+# of transient conflict on the same table.
+POST_PROCESS_BATCH_RETRY_DELAY = 0.5  # seconds before the first retry; doubled each attempt
+POST_PROCESS_BATCH_MAX_CONFLICT_RETRIES = 3
+
+
 @app.task
 def post_process_findings_batch(
     finding_ids,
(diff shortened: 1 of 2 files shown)


blakeaowens (approved):

devGregA (approved):

Maffooch:
The merge queue dequeued this PR on a CI failure that is unrelated to this diff:

- **Unit tests (arm64):** a single failure in `unittests/test_v3_alias_authorized_users_api_authz.py::test_asset_edit_cannot_add_authorized_users` — `AssertionError: 403 != 404 : No Product matches the given query`. That test exercises v3-alias authorized-users API authorization; this PR only changes `dojo/finding/helper.py` (`post_process_findings_batch` deadlock retry) and its unit tests, and does not touch authorization or the products API. The `403` vs `404` split is a test-isolation/ordering artifact, not something this change can affect.
- **test-k8s:** failed at cluster bring-up (`connection to the server localhost:8080 was refused`), i.e. infrastructure, before any test ran.

The same queue "Unit tests" failure hit the two docs-only PRs batched alongside this one in the same window, which confirms it is not originating from this change. This PR's own head checks are all green (including the amd64 Rest Framework unit tests that run the added tests). It's already been re-queued — no code change needed here.

---

PR #15733, opened 2026-08-19
Regulations

dangoelz:
This pull request adds documentation pertaining to Regulations in Pro + OS DefectDojo. 

(I made a minor copy-paste error between the two docs that I only realized after I'd staged, committed, and pushed the initial PR, which I then corrected, so that's why there are two commits.)

I also fixed a separate, prior copy-paste error in the subtitle of the OS Calendar documentation.


paulOsinski (approved):

paulOsinski:
I'm fine with approving this - will be catching up the Pro screenshots this week.

Maffooch (approved):

PR #15756, opened 2026-08-20
docs: restore eight markdown links mangled into NUL bytes

devGregA:
**Description**

Five documentation files contained literal NUL bytes (`0x00`) that broke eight inline markdown links. In every case a placeholder of the form `\0<index>\0` sits exactly where the link destination `](url)` belongs, so the link renders as raw text:

| File | NULs | Broken links |
| --- | --- | --- |
| `connectors/upstream/toolreference.md` | 6 | Semgrep token page, Snyk REST base, Snyk EU REST base |
| `get_started/about/about_defectdojo.md` | 4 | Open-Source demo, Pro demo |
| `get_started/contributing/how-to-write-a-parser.md` | 2 | `acunetix.md` example page |
| `get_started/contributing/documentation.md` | 2 | `http://localhost:1313` |
| `issue_tracking/pro_integration/messaging_connectors.md` | 2 | `https://api.slack.com/apps` |

The NUL bytes had a second effect beyond the eight broken links, and it is the reason this went unnoticed: **`grep` reports each of these files as `Binary file ... matches` and suppresses the matching lines**, so the broken text never shows up in a search. For example, `grep -rn "Snyk REST API"` returned only `Binary file ... matches` rather than the line. `git diff` is affected for three of the five — it renders them as `Binary files differ` — while `toolreference.md` and `how-to-write-a-parser.md` still diff as text because their first NUL byte falls past git's 8000-byte binary-detection window.

**Origin and how the destinations were recovered**

All eight placeholders trace to a single commit, 1cafb09ac7 (#15661). The pattern is consistent with a link-protecting pass that swaps each destination for a NUL-delimited sentinel, rewrites the surrounding prose, and restores the destinations afterwards — with the restore step dropping these eight. Indices are per-file counters (17, 18, 19, 22, 77, 389, 394, 395).

Nothing here is guessed. For each file, the last revision before #15661 still holds the intact text; it was located by walking that file's history and counting NUL bytes per revision until one came back clean. Every repaired line is **byte-identical to its pre-#15661 counterpart**, and all eight restored links turn out to be self-links (display text equals destination), which is a useful independent sanity check.

`toolreference.md` has a second corroborating source: its four translated siblings (`.de`, `.es`, `.fr`, `.ja`) were never touched by #15661 and carry the same markup. They agree with the parent revision on all three destinations. Those four files need no change — 0 NUL bytes each, links intact.

**Scope note**

One pre-existing inconsistency was deliberately left alone, because it predates the corruption and appears identically in all five language files: the Snyk link text reads `/rest` while the destination points at `/v1`. The `/rest` value is the one the docs instruct you to enter in the **Location** field, so only the decorative href is stale. Worth a follow-up, but changing it here would mean an unexplained edit sitting on top of an otherwise byte-exact restore.

**Test results**

- `python -c "d=open('docs/content/connectors/upstream/toolreference.md','rb').read(); print(d.count(b'\0'))"` → `0`
- Repo-wide scan: **0** text files under `docs/` contain NUL bytes (images excluded).
- Regex sweep across all of `docs/content` for markdown link openers whose next two characters are not `](` → **0 hits**, so the defect is fully enumerated rather than partially patched.
- Every repaired line verified byte-identical to its pre-#15661 revision.
- No CR bytes introduced; files stay LF-only per `.gitattributes` (`*.md text eol=lf`).
- `grep` now reads all five files as text.
- Docs rendering is covered by the existing `validate_docs_build` workflow; no local Hugo toolchain was used.

**Documentation**

This PR is the documentation change.


devGregA:
GitHub cannot render a text diff for three of the five files: the **old** blob has a NUL byte inside git's 8000-byte binary-detection window, so they show as `Bin` / `+0/-0` above. Those changes are reproduced here so they can be reviewed without fighting the binary diff.

Each NUL byte is written as `<NUL>` below. In every case the only edit is the placeholder becoming the link destination - nothing else on the line changed.

### `docs/content/get_started/about/about_defectdojo.md`

```diff
@@ line 108 @@
-A running example of DefectDojo (Open-Source Edition) is available at [https://demo.defectdojo.org/<NUL>18<NUL>.
+A running example of DefectDojo (Open-Source Edition) is available at [https://demo.defectdojo.org/](https://demo.defectdojo.org/).
@@ line 113 @@
-[https://pro.demo.defectdojo.com/<NUL>19<NUL>.
+[https://pro.demo.defectdojo.com/](https://pro.demo.defectdojo.com/).
```

### `docs/content/get_started/contributing/documentation.md`

```diff
@@ line 21 @@
-4. Visit [http://localhost:1313<NUL>17<NUL>.
+4. Visit [http://localhost:1313](http://localhost:1313).
```

### `docs/content/issue_tracking/pro_integration/messaging_connectors.md`

```diff
@@ line 38 @@
-1. Go to [https://api.slack.com/apps<NUL>22<NUL> and select **Create New App**, then **From scratch**.
+1. Go to [https://api.slack.com/apps](https://api.slack.com/apps) and select **Create New App**, then **From scratch**.
```

The other two files (`connectors/upstream/toolreference.md`, `get_started/contributing/how-to-write-a-parser.md`) diff as text in the Files tab, because their first NUL byte falls past that 8000-byte window.

blakeaowens (approved):

Maffooch (approved):

PR #15761, opened 2026-08-22
docs: 3.3.x upgrade notes for the three deduplication identity changes

devGregA:
The 3.3 upgrade page currently says "no special instructions". This release changes the stored identity of three scan types, so that is wrong three times over.

`Xeol Parser`, `Checkmarx One Scan` and `Checkmarx Scan detailed` all gain `HASHCODE_FIELDS_PER_SCANNER` registrations in 3.3.0. Findings imported before the upgrade therefore carry a `hash_code` that no longer matches what an import computes afterwards, and left alone the next re-import closes them as absent and recreates them, detaching false positive history and risk acceptances.

The page now follows the shape the 3.2 notes established:

* A table of the three scan types, the new field lists, and why each moved.
* The ledger path: instances with the identity signature ledger enabled (the default from Pro 3.2.300) need no action, with one caveat spelled out: a large never-backfilled instance should run `identity_signatures_backfill` **before** upgrading, because a rehash replaces the identities the backfill exists to record.
* The manual path: the three `manage.py dedupe --parser '...' --hash_code_only` commands, in the same form the 2.4x and 3.2 notes use (and the command name is `dedupe`, matching the correction merged in #15662).

Installs that never imported these scan types are unaffected, and the page says so.

### DIFF
--- a/docs/content/releases/os_upgrading/3.3.md
+++ b/docs/content/releases/os_upgrading/3.3.md
@@ -20,22 +20,28 @@ history and risk acceptances along the way.
 | `Checkmarx One Scan` | `unique_id_from_tool` | The parser already deduplicates on the vendor id. The stored hash previously fell through to the legacy field set, whose title and description are the same volatile text for every result family, so the stored identity moved whenever Checkmarx reworded a finding even though matching did not. |
 | `Checkmarx Scan detailed` | `vuln_id_from_tool`, `file_path`, `line` | The same three fields as its sibling `Checkmarx CxFlow SAST`, which the parser populates on every detailed-mode finding. None of them carries scan text. |
 
-### If the identity signature ledger is enabled
+### DefectDojo Pro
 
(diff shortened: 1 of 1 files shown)


devGregA:
Amended after the default-flip attempt on the Pro side surfaced two inaccuracies in the first cut of this page:

* `identity_signature_matching` does **not** default on in 3.2.300. Only dual-write does. Running the full Pro suite under matching-on failed the affirmative-distinctness suites (disjoint vulnerability ids, untracked version bumps), so matching stays opt-in until the fallback can distinguish an evaluated rejection from an absent identity.
* The drift watch **reports**; it does not repair. The rewritten section sends everyone to the same three `manage.py dedupe` commands, with the Tuner acceptance as the Pro-native equivalent, and keeps the backfill-before-upgrade guidance scoped to instances where matching is actually enabled.

blakeaowens (approved):

Maffooch (approved):

PR #15786, opened 2026-08-24
docs(rules_engine_2): document the SLA configuration and Risk Priority asset actions

svader0:
## Description

Documents two new Rules Engine 2.0 actions: setting an Asset's SLA configuration, and setting its Risk Priority.

Asset support is already documented. This adds the two action nodes to the node reference, and fills two gaps on the Building Rules page:

- **A field reference for Asset items.** An Asset item carries no `finding`, `test` or `engagement` block, so the new table lists what it does carry.
- **A note about Asset type and scope.** The scope picker offers what the Assets list can filter on, and Asset type is not one of those filters. To act on Assets of one type, scope broadly and add an If / Filter condition on `product.asset_type`, which compares the type's code rather than its display label.

Two behaviours are written down because a run reports them and they are otherwise puzzling:

- The SLA action skips an Asset that is already being recalculated, and says it was skipped rather than changed.
- A rule owner needs the matching configuration permission to use either action. Without it the picker is empty.

Pairs with a DefectDojo Pro change that adds the two nodes.

## Test Commands

Docs only. Built locally with Hugo (`npm ci && npm run build`), no errors, and the new headings and anchors resolve in the built HTML.

### DIFF
--- a/docs/content/automation/rules_engine_2/about.md
+++ b/docs/content/automation/rules_engine_2/about.md
@@ -120,7 +120,7 @@ Runs and deliveries are both kept for 180 days by default, then pruned. The prod
 ## Where to go next
 
 * [Building Rules](../building_rules/) covers the editor, triggers, scope, conditions and templates.
-* [Node Reference](../node_reference/) documents all 25 nodes.
+* [Node Reference](../node_reference/) documents all 38 nodes.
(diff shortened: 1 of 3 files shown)


svader0:
Superseded by #15788. This branch was cut from `dev`, but the Pro change it pairs with is based on `bugfix`, so both sides need to be on the same release line.

PR #15809, opened 2026-08-26
fix(api): answer 410 rather than 500 when a v2 route reaches a deprecated Endpoint

svader0:
## The problem

When Locations is enabled, the old `Endpoint` model raises `NotImplementedError` on init. Nothing on the `/api/v2/` surface caught that error. Any route that still touched Endpoint data returned a 500 with no explanation. Every retry wrote an error line into alerting.

## The change

`Endpoint.__init__` now raises `EndpointDeprecatedError`, its own subclass of `NotImplementedError`. The api/v2 exception handler catches that one class and answers `410 Gone`. The body is machine-readable, so a client can branch on `code` instead of parsing prose:

```json
{
  "code": "endpoint_api_sunset",
  "message": "The Endpoint API is not available on instances with Locations enabled. Reads are served from Locations; writes are not available.",
  "replacement": "/api/v2/location/",
  "docs": "https://docs.defectdojo.com/asset_modelling/locations/pro__migrating_from_endpoints/"
}
```

The handler logs this at info, not error.

A plain `NotImplementedError` still returns a 500. Only the new subclass answers 410.

## Tests

Two reported routes get their exact failing requests as regression tests. Those requests are `GET /api/v2/endpoints/?limit=500` and an ordered `GET /api/v2/test_imports/`. Both routes were fixed in 3.2.100, and nothing held them there until now.

A third test sweeps every route registered on the v2 router. It fails if any list route answers 5xx, or if any falls into the deprecation sunset.

One test resolves the configured `EXCEPTION_HANDLER` and calls it, because the other tests call the handler directly.

## What does not change

Writes to `/api/v2/endpoints/` and `/api/v2/endpoint_status/` still return 403. Read compatibility is unchanged. Every other route behaves as before.

## Docs

The migration guide gains one section describing the 410 body and what to do if a read returns it.

### DIFF
--- a/docs/content/asset_modelling/locations/PRO__migrating_from_endpoints.md
+++ b/docs/content/asset_modelling/locations/PRO__migrating_from_endpoints.md
@@ -99,6 +99,21 @@ A few things behave differently from the original Endpoint API:
 - **`endpoint` field on Endpoint_Status.** The legacy `endpoint` field is reconstructed by looking up the matching Asset Reference. In rare cases where a Finding's Asset no longer matches its Location's Asset references, this field may be null.
 - **Pagination and ordering.** Available ordering fields on the read-compat shim are `host`, `product`, `id`, and `active_finding_count`. If your client orders by another field, switch to one of these or move to the new Locations endpoints.
 
+### If a Route Returns 410
+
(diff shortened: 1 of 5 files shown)


svader0:
[sc-14273]

Maffooch (approved):

blakeaowens (approved):

devGregA (approved):

PR #15813, opened 2026-08-26
fix(parser): make xeol severity parsing deterministic without wall clock

### LINKED ISSUE #15651: Xeol parser derives severity from the wall clock, so the same report parses differently over time
### Summary

`XeolParser` derives a finding's `severity` from the wall clock at parse time, so parsing the **same, unchanged** report on two different days can produce findings with different severities. Parser output should be a pure function of its input; here it is a function of the input *and* the calendar.

### Where

`dojo/tools/xeol/parser.py`, lines 66-83:

```python
# Determine severity based on EOL date
severity = "Info"
eol_str = cycle.get("Eol", "")
try:
    eol_date = datetime.strptime(eol_str, "%Y-%m-%d")
    now = datetime.now()
    if eol_date < now:
        delta = now - eol_date
        if delta <= timedelta(weeks=2):
            severity = "Low"
        elif delta <= timedelta(weeks=4):
            severity = "Medium"
        elif delta <= timedelta(weeks=6):
            severity = "High"
        else:
            severity = "Critical"
except Exception:
    severity = "Info"
```

`now` is the only non-input in the expression, and it moves.

### Impact

A component that reached end of life walks through four severities over the six weeks that follow, with no change whatsoever to the scan report:

| Time since the component's EOL date | Severity |
|---|---|
| not yet EOL | Info |
| 0 to 2 weeks | Low |
| 2 to 4 weeks | Medium |
| 4 to 6 weeks | High |
| beyond 6 weeks | Critical (terminal) |

Two consequences:

1. **Severity churn on reimport.** Re-importing an unchanged Xeol report re-grades existing findings. A user watching a dashboard sees EOL findings silently escalate with no upstream event behind the change, and any severity-based SLA or report shifts with it.
2. **Duplicate findings for anyone who hashes severity.** `severity` is a legal entry in `HASHCODE_FIELDS_PER_SCANNER` and is used by a number of scan types today (Anchore, Aqua, Burp and others). Xeol is not in that dict, so it currently falls back to the legacy field set (`title`, `cwe`, `line`, `file_path`, `description`) and its `hash_code` is unaffected. But a user who customizes the hash fields for Xeol to include `severity`, which is a supported configuration, gets a hash that changes on its own schedule: reimport stops matching the stored findings, closes them as absent, and creates duplicates. Rehashing cannot repair that, because the stored rows do not contain the values the parser would write today.

### Reproduction

`unittests/scans/xeol/xeol_one_finding.json` carries `"Eol": "2026-07-02"` for Perl. That sample crossed the six-week boundary at `2026-08-13T00:00` local time of whatever runs the parser:

```
parse on 2026-08-12 -> severity High
parse on 2026-08-14 -> severity Critical
```

Nothing about the file changed.

Note this is also a latent trap in the test corpus: any assertion on that file's severity is time-dependent. `unittests/tools/test_xeol_parser.py` happens not to assert severity for `xeol_one_finding.json` (it only asserts it for `xeol_multiple_findings.json`, whose EOL dates are in 2019 and therefore permanently Critical), which is why CI has not surfaced this.

### Prior art

This is the same class of defect as the Govulncheck parser building its description from `", ".join(set(...))`, where Python's per-process string hash randomization made the description, and therefore the `hash_code`, differ on every import of the same file. That was fixed by sorting the sets in `dojo/tools/govulncheck/parser.py`. The principle applied there is the one at stake here: a parser's output must be reproducible from its input alone.

### Possible fixes

1. **Grade against a timestamp carried in the report** rather than `datetime.now()`, so the result is fixed per file. The xeol JSON `descriptor` block does not appear to carry a scan timestamp today, so this may need an upstream tool change.
2. **Drop the time banding** and assign a single severity to any past-EOL component (with `Info` retained for one not yet EOL). Deterministic, and simple to reason about. This does change severity for existing findings on the next import, so it is worth calling out in release notes.
3. **Keep the banding but compute it from data in the report**, for example the EOL date relative to the artifact's release or latest-release date, both of which are already parsed into the description.

Option 2 is the smallest change that makes the parser deterministic. Happy to open a PR for whichever direction maintainers prefer, since this is a product decision about how EOL findings should be graded rather than a purely mechanical fix.

P6-36-216:
**Description**

Fixes #15651

Removes the non-deterministic `datetime.now()` wall-clock calculations from `XeolParser`. Previously, severity was derived dynamically from the current date relative to EOL dates, causing the same static scan report to escalate in severity on re-import and leading to hash inconsistencies.

Severity is now computed deterministically, findings with an active `Eol` field are assigned `Critical`, whereas components not marked EOL default to `Info`.

**Test results**

- Added an explicit severity assertion (`Critical`) to `test_parse_file_with_one_finding` in `unittests/tools/test_xeol_parser.py`.
- Ran unit tests locally with `python manage.py test unittests.tools.test_xeol_parser` (3/3 tests passing).
- Verified code styling and line length compliance with `flake8`.

**Documentation**

No documentation updates required (parser behavior bug fix).

**Checklist**

### DIFF
--- a/dojo/tools/xeol/parser.py
+++ b/dojo/tools/xeol/parser.py
@@ -2,68 +2,106 @@ import json
 
 from dojo.models import Finding
 
-
 class XeolParser:
     def get_scan_types(self):
-        return ["Xeol Scan"]
+        return ["Xeol Parser"]
 
     def get_label_for_scan_types(self, scan_type):
-        return "Xeol Scan"
+        return "Xeol Parser"
 
     def get_description_for_scan_types(self, scan_type):
-        return "Import Xeol JSON output."
+        return "Import JSON report"
 
     def get_findings(self, file, test):
-        tree = json.load(file)
-        return self.get_items(tree, test)
+        findings = []
+        data = json.load(file)
 
-    def get_items(self, tree, test):
-        items = {}
-        for match in tree.get("matches", []):
+        if not isinstance(data, dict) or "matches" not in data:
+            return findings
+
+        distro = data.get("distro", {})
+        for match in data["matches"]:
+            cycle = match.get("Cycle", {})
             artifact = match.get("artifact", {})
-            cycle = match.get("cycle", {})
 
-            name = artifact.get("name", "")
-            version = artifact.get("version", "")
-            purl = artifact.get("purl", "")
-            cpe = artifact.get("cpe", "")
-
-            title = f"{name} EOL Information"
-            description = (
-                f"**Artifact Name:** {name}\n"
-                f"**Artifact Version:** {version}\n"
-                f"**PURL:** {purl}\n"
-                f"**CPE:** {cpe}\n"
-                f"**Cycle:** {cycle.get('cycle', '')}\n"
-                f"**Release Date:** {cycle.get('releaseDate', '')}\n"
-                f"**EOL Date:** {cycle.get('eol', '')}\n"
-                f"**Latest Release:** {cycle.get('latestRelease', '')}\n"
-                f"**Latest Release Date:** "
-                f"{cycle.get('latestReleaseDate', '')}\n"
+            title = (
+                f"{cycle.get('ProductName', 'Unknown Product')} EOL Information"
             )
 
-            severity = "Info"
-            eol_val = cycle.get("Eol", "")
-            if eol_val:
-                if (isinstance(eol_val, bool) and eol_val) or (
-                    isinstance(eol_val, str) and eol_val.lower() not in {"false", "none", ""}
-                ):
-                    severity = "Critical"
+            description_lines = [
+                f"**Product Name:** {cycle.get('ProductName', 'N/A')}",
+                f"**Release Cycle:** {cycle.get('ReleaseCycle', 'N/A')}",
+                f"**EOL Date:** {cycle.get('Eol', 'N/A')}",
+                f"**Latest Release Date:** {cycle.get('LatestReleaseDate', 'N/A')}",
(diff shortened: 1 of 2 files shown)


dojo/tools/xeol/parser.py
```diff
@@ -1,107 +1,69 @@
...
 
-from dojo.location.feature import locations_enabled
 from dojo.models import Finding
-from dojo.tools.locations import LocationData
 
 
 class XeolParser:
     def get_scan_types(self):
-        return ["Xeol Parser"]
+        return ["Xeol Scan"]
```
Maffooch: Changing the name of the scanner will break automation for anyone using this tool. Please revert

Maffooch (changes requested):
With all the changes to the general structure of the finding, and the default values when fields are not present, deduplication and reimport may not be able to match on findings created with the current version of the parser. 

please back out any other changes that are unrelated to changing the severity parsing

P6-36-216:
> With all the changes to the general structure of the finding, and the default values when fields are not present, deduplication and reimport may not be able to match on findings created with the current version of the parser.
> 
> please back out any other changes that are unrelated to changing the severity parsing

Understood, thanks! I'm working on it. I noticed that some piece of code is formatted across multiple lines, that was originally in one single line (which I think is happening because of my IDE). I'll try and fix that as well.

dojo/tools/xeol/parser.py
P6-36-216: Done! I have removed any/all unrelated changes and tested it. Please let me know if anything else, thanks!

Maffooch (approved):

blakeaowens (approved):

devGregA (approved):

PR #15825, opened 2026-08-28
refactor(importers): one existence lookup for findings deleted mid-import

valentijnscholten:
**Summary**

Follow-up to #15512, which removed an existence `SELECT` from the test/engagement write-back by detecting the deleted row from `save(force_update=True)` instead. This does the same tidying one level down, on the existence lookups the importer still runs for findings.

- `filter_finding_ids_by_existence()` (import-history records) and `BaseImporter.deleted_finding_ids()` (child-row buffers, added by #15428) are the same query written in opposite directions, with one caller each. They become a single chunked `finding_helper.deleted_finding_ids()`; the importer method stays as a delegate so the flush paths keep their override seam.
- Unlike the write-back, these callers cannot let the write report the deletion. They are `bulk_create()` calls, and with `DEFERRABLE INITIALLY DEFERRED` foreign keys the violation only arrives at `COMMIT`, naming a constraint rather than a row. So the lookup stays here — it just stops being duplicated.
- It also gains a bound on its `IN` clause. `update_import_history()` was the one caller running unbounded: the buffer flushes ask about at most a batch, but it is called once per import with every finding it touched, so a large scan put tens of thousands of ids in one statement.
- `close_old_findings()` keeps its own query — existence already falls out of the refresh `_sync_close_old_finding_status_fields()` does anyway, so sharing the helper would only add a round trip. It now shares the chunk size, and the docstring says why.
- No behaviour change, and no query-count change: the pinned baselines in `test_importers_performance.py` and `test_tag_inheritance_perf.py` are untouched.

**Tests**

- New in `test_finding_helper.py`: `TestDeletedFindingIds`, 7 tests — the polarity in both directions, the cases that must cost no query (empty input, all-`None`), an id that never existed, and two on the chunking (query count with `FINDING_EXISTENCE_CHUNK` patched to 2, and the answer being chunk-size independent).
- New in `test_importers_deleted_finding_child_rows.py`: `test_update_import_history_skips_a_finding_deleted_during_the_import`, next to the buffer guards from #15428. It goes there because `UpdateImportHistoryTests` is skipped wholesale, so the import-history pre-check otherwise has no running coverage.
- Updated in `test_update_import_history.py`: 2 patch targets moved to `deleted_finding_ids`. They were already stale — #15600 renamed the function they named, so they would have raised `AttributeError` if that class were ever un-skipped.

Green on PostgreSQL: `test_finding_helper` (25), `test_importers_deleted_finding_child_rows` (11), `test_importers_performance` (11), `test_import_reimport` (132), `test_importers_importer` (50), `test_importers_deduplication` (50), `test_tag_inheritance_perf` (24), `test_importers_closeold` (7).

### DIFF
--- a/dojo/finding/helper.py
+++ b/dojo/finding/helper.py
@@ -186,17 +186,42 @@ def update_finding_status(new_state_finding, user, changed_fields=None):
     new_state_finding.last_status_update = now
 
 
-def filter_finding_ids_by_existence(finding_ids):
-    """
-    Return the subset of the given ids that still exist in the database.
+# Bounds the IN clause of the existence lookup below. An import's result set is not
+# bounded by anything else, so without this a large scan asks the database about tens of
+# thousands of ids in one statement.
+FINDING_EXISTENCE_CHUNK = 1000
 
-    Centralized helper used by importers to avoid FK violations during
-    bulk_create -- e.g. a background async_dupe_delete task removing a finding
-    between when an importer collected it and when it writes a history record.
+
+def deleted_finding_ids(finding_ids) -> set[int]:
     """
+    Of the given finding ids, the ones whose row is no longer in the database.
+
+    The single place that answers "which of these findings are gone", for every caller
+    holding finding references across a window in which a finding can be deleted --
+    import history records, the child-row buffers flushed at an import batch boundary,
+    and anything else that would otherwise insert a dangling reference or re-save a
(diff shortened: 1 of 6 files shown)


valentijnscholten:
A follow up to #15512 which I had already on that branch, but wasn't pushed yet.

blakeaowens (approved):

Maffooch (approved):

devGregA (approved):

